PR #226, opened 2022-05-30
add logic for add-on invoice and webhook

lovrocolic:
Please first review #221 (apply add-ons to customers)

Also, found and fixed one bug in webhooks/base_service

### DIFF
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -5,6 +5,7 @@ class Fee < ApplicationRecord
 
   belongs_to :invoice
   belongs_to :charge, optional: true
+  belongs_to :add_on, optional: true
   belongs_to :subscription
 
   has_one :customer, through: :subscription
@@ -18,15 +19,19 @@ class Fee < ApplicationRecord
   validates :vat_amount_currency, inclusion: { in: currency_list }
   validates :units, numericality: { greated_than_or_equal_to: 0 }
 
-  scope :subscription_kind, -> { where(charge_id: nil) }
+  scope :subscription_kind, -> { where(charge_id: nil, add_on_id: nil) }
   scope :charge_kind, -> { where.not(charge_id: nil) }
 
   def subscription_fee?
-    charge_id.blank?
+    charge_id.blank? && add_on_id.blank?
   end
 
   def charge_fee?
     charge_id.present?
+  end
+
+  def add_on_fee?
+    add_on_id.present?
   end
 
   def compute_vat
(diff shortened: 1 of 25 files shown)


app/services/webhooks/add_on_service.rb
```diff
@@ -0,0 +1,23 @@
...
+    def object_serializer
+      ::V1::InvoiceSerializer.new(
+        object,
+        root_name: 'invoice',
+        includes: %i[customer subscription],
+      )
+    end
+
+    def webhook_type
+      'add_on.created'
```
vincent-pochet, line 20: Since the the webhook is sending an invoice object, I suggest to set `webhook_type` to `invoice.add_on_added`

app/services/webhooks/add_on_service.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def current_organization
+      @current_organization ||= object.organization
+    end
+
+    def object_serializer
+      ::V1::InvoiceSerializer.new(
+        object,
+        root_name: 'invoice',
+        includes: %i[customer subscription],
```
vincent-pochet, line 15: Following my comments about fees, you should also include fees in the webhook

It will require an update of the fee model on:
- https://github.com/getlago/lago-api/blob/main/app/models/fee.rb#L37
- https://github.com/getlago/lago-api/blob/main/app/models/fee.rb#L43
- https://github.com/getlago/lago-api/blob/main/app/models/fee.rb#L49

app/services/invoices/add_on_service.rb
```diff
@@ -0,0 +1,57 @@
...
+    def create
+      ActiveRecord::Base.transaction do
+        invoice = Invoice.find_or_create_by!(
+          subscription: subscription,
+          from_date: date,
+          to_date: date,
+          issuing_date: date,
+        )
+
+        compute_amounts(invoice)
```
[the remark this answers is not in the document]
lovrocolic: added:) Tnx for noticing

app/services/webhooks/add_on_service.rb
lovrocolic: changed

app/services/webhooks/add_on_service.rb
lovrocolic: Tnx for the comments, updated

app/models/fee.rb
```diff
@@ -5,6 +5,7 @@ class Fee < ApplicationRecord
 
   belongs_to :invoice
   belongs_to :charge, optional: true
+  belongs_to :add_on, optional: true
```
vincent-pochet, line 8: Actually, I think the relation should be at the applied_add_on level. A bit like it's done for the coupon with the credit model.

Sorry it wasn't clear in my previous review :/

lovrocolic: update is pushed:)

vincent-pochet (approved):

Issue #242, opened 2022-06-03
[FEAT]: Implement BE for percentage charge model


lovrocolic:
These changes add possibility to charge based on percentage

PR #248, opened 2022-06-06
feat: #242 add basic logic for percentage charge model

lovrocolic:


app/services/charges/charge_models/percentage_service.rb
```diff
@@ -0,0 +1,32 @@
...
+        amount + (value * BigDecimal(fixed_amount_value))
+      end
+
+      # NOTE: FE divides percentage rate with 100 and sends to BE
+      def rate
+        charge.properties['rate']
+      end
+
+      def fixed_amount_value
+        charge.properties['fixed_amount_value'] || 0
```
vincent-pochet, line 24: I suggest to name this one `fixed_amount`. It's more in sync with the naming of other charge model amount fields

app/services/charges/charge_models/percentage_service.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Charges
+  module ChargeModels
+    class PercentageService < Charges::ChargeModels::BaseService
+      protected
+
+      def compute_amount(value)
+        amount = value * BigDecimal(rate)
+
+        return amount if BigDecimal(fixed_amount_value).zero?
```
vincent-pochet, line 11: Since the methods calls 3 times `BigDecimal(fixed_amount_value)` what do you think of adding a method for it?

vincent-pochet: Just a suggestion, but I feel it could be a bit clearer with something like that:

```ruby
def compute_amount(value)
  compute_percent_amount(value) + compute_fixed_amount(value)
end

def compute_rate_amount(value)
  value * rate
end

def compute_fixed_amount(value)
  return fixed_amount if fixed_amount_target == 'all_units'
  
  value * fixed_amount_value
end

def rate
  @rate ||= BigDecimal(charge.properties['rate'])
end

def fixed_amount
  @fixed_amount ||= BigDecimal(charge.properties['fixed_amount_value'] || 0)
end
```

vincent-pochet (commented):
Smalls comments on the form, but the logic is good! :+1:

app/services/charges/charge_models/percentage_service.rb
```diff
@@ -0,0 +1,38 @@
...
+      def compute_amount(value)
+        compute_percentage_amount(value) + compute_fixed_amount(value)
+      end
+
+      def compute_percentage_amount(value)
+        value * rate
+      end
+
+      def compute_fixed_amount(value)
+        return 0 if fixed_amount.zero?
```
vincent-pochet, line 17: You can keep it, but actually if `fixed_amount` is zero, the method will return 0 in both cases :)

vincent-pochet (approved):

app/services/charges/charge_models/percentage_service.rb
lovrocolic: 🤦 removed in other PR!

PR #254, opened 2022-06-10
feat: Create Stripe connection for organizations

### LINKED ISSUE #253: [FEAT]: Create Stripe connection for organizations
Add data model, GraphQL mutations and query to handle connection to Stripe API using API secret key

vincent-pochet:
Fixes #253

Add data model, GraphQL mutations and query to handle connection to Stripe API using API secret key

**NOTE**: This PR is also adding the config to use [Active Record Encryption](https://guides.rubyonrails.org/active_record_encryption.html) for API secrets.
It still have to be configured for staging and production


app/graphql/types/payment_providers/stripe.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      field :id, ID, null: false
+      field :secret_key, String, null: false
+
+      field :create_customers, Boolean, null: false
+      field :send_zero_amount_invoice, Boolean, null: false
+
+      # NOTE: Secret key is a sensitive information. It should not be sent back to the
+      #       front end application. Instead we send an obfuscated value
+      def secret_key
```
jdenquin, line 16: is it something the Front really need to have?

app/models/payment_providers/stripe_provider.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module PaymentProviders
+  class StripeProvider < BaseProvider
```
jdenquin, line 4: just a naming thing, could it be better to have `PaymentProviders::Stripe` only?

spec/models/payment_providers/base_provider_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe PaymentProviders::BaseProvider, type: :model do
+  subject(:provider) { described_class.new(attributes) }
+
+  let(:attributes) do
```
jdenquin, line 8: just a best practice thing, can we put the `let` with block after the one lines `let` ? or does the order matter?

jdenquin (approved):
Small comments, approved, well done!!!

app/models/payment_providers/stripe_provider.rb
jdenquin: yeah I was thinking about this too, fair enough, you win!

PR #265, opened 2022-06-16
feat: enable configuration for webhook attempts

prognostikos:
This change introduces the `LAGO_WEBHOOK_ATTEMPTS` environment variable,
which will be used instead of the default of 3 when attempting to
deliver webhooks.


prognostikos:
I will see if I can fix the specs, on first glance the failures don't seem to be related to this change

vincent-pochet:
Hey @prognostikos ,

Thank you very much for this pull request!

It looks good to me.
For the failing specs, it seems related to the active record encryption that has been introduced later this week. I will have to dig a bit into it to understand what's going on.

vincent-pochet (approved):

Issue #286, opened 2022-06-23
[BUG]: Fix date boundaries during plan upgrade


lovrocolic:
When upgrading a plan and when old plan has payment in arrear, invoice from_date should be fixed.

PR #287, opened 2022-06-23
fix: #286 fix invoice boundaries for termination case

lovrocolic:


app/services/invoices/create_service.rb
```diff
@@ -59,6 +59,19 @@ def from_date
...
                    else
                      raise NotImplementedError
+                   end
+
+      # NOTE: In case we are terminating old plan (paying in arrear),
+      # we should move to the beginning of the billing period
+      if subscription.terminated?
+        @from_date = case subscription.plan.interval.to_sym
+                     when :monthly
+                       Time.zone.at(timestamp).to_date.beginning_of_month
```
lovrocolic, line 69: we already covered the case If it's less than started_at (below)

vincent-pochet: What do you think of moving this logic into a dedicated method?
We start to have a lot of logic in this one

vincent-pochet (approved):
Looks good! :ship: it!

app/services/invoices/create_service.rb
lovrocolic: agree!

jdenquin (approved):

PR #300, opened 2022-06-29
Block registration if LAGO_DISABLE_SIGNUP is set

prognostikos:
Closes https://github.com/getlago/lago/issues/47


vincent-pochet:
The feature as been handled by https://github.com/getlago/lago-front/pull/236

Issue #292, opened 2022-06-27
[FEAT]: Add coupons API endpoints


lovrocolic:
Related to: https://github.com/getlago/lago-product/issues/33

lovrocolic:
POST /api/v1/coupons
PUT /api/v1/coupons/{{coupon-code}}
DELETE /api/v1/coupons/{{coupon-code}}
GET /api/v1/coupons
GET /api/v1/coupons/{{coupon-code}}

PR #308, opened 2022-06-30
feat: #292 coupons API endpoints

lovrocolic:


app/controllers/api/v1/coupons_controller.rb
```diff
@@ -0,0 +1,111 @@
...
+        service = Coupons::CreateService.new
+        result = service.create(
+          **input_params
+            .merge(organization_id: current_organization.id)
+            .to_h
+            .symbolize_keys
+        )
+
+        if result.success?
+          render(
```
vincent-pochet, line 16: Since this exact same code is used in `show`, `create`, `update` and `destroy`, I'm wondering if we could add a private shared `render_coupon` method. What do you think?

vincent-pochet (approved):
:rocket:

app/controllers/api/v1/coupons_controller.rb
lovrocolic: Sounds great to me! Refactored

Issue #293, opened 2022-06-27
[FEAT]: Add Add-ons API-endpoints


lovrocolic:
Related to: https://github.com/getlago/lago-product/issues/33

lovrocolic:
POST /api/v1/add-ons
PUT /api/v1/add-ons/{{add-on-code}}
DELETE /api/v1/add-ons/{{add-on-code}}
GET /api/v1/add-ons
GET /api/v1/add-ons/{{add-on-code}}

PR #309, opened 2022-06-30
feat: #293 add add-ons api endpoints

lovrocolic:


app/controllers/api/v1/add_ons_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+        service = AddOns::CreateService.new
+        result = service.create(
+          **input_params
+            .merge(organization_id: current_organization.id)
+            .to_h
+            .symbolize_keys
+        )
+
+        if result.success?
+          render(
```
vincent-pochet, line 16: Same as https://github.com/getlago/lago-api/pull/308, we probably could add a private shared `render_add_on` method.

vincent-pochet (approved):
:rocket:

app/controllers/api/v1/add_ons_controller.rb
lovrocolic: Refactored :)

Issue #295, opened 2022-06-27
[FEAT]: Add organisations API endpoints


lovrocolic:
Related to: https://github.com/getlago/lago-product/issues/33

lovrocolic:
PUT /api/v1/organizations//{{organization.id}}

PR #310, opened 2022-07-01
feat: #295 add organization update endpoint

lovrocolic:


app/controllers/api/v1/organizations_controller.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class OrganizationsController < Api::BaseController
+      def update
```
vincent-pochet, line 6: Do we plan to add a route to show the current_organization?

vincent-pochet (approved):

app/controllers/api/v1/organizations_controller.rb
lovrocolic: Product decided that we only need update ATM @vincent-pochet

Issue #294, opened 2022-06-27
[FEAT]: Add invoices API endpoints


lovrocolic:
Related to: https://github.com/getlago/lago-product/issues/33

lovrocolic:
GET /api/v1/invoices
GET /api/v1/invoices//{{invoice.id}}

PR #311, opened 2022-07-01
feat: #294 add invoice show and index endpoints

lovrocolic:


app/controllers/api/v1/invoices_controller.rb
```diff
@@ -22,6 +22,35 @@ def update
...
 
+      def show
+        invoice = Invoice.find_by(id: params[:id])
+
+        return not_found_error unless invoice
+
+        render(
+          json: ::V1::InvoiceSerializer.new(
+            invoice,
+            root_name: 'invoice',
```
vincent-pochet, line 33: We should probably include the attached fees (or maybe add an option to include them):

```ruby
:V1::InvoiceSerializer.new(
  invoice,
  root_name: 'invoice',
  includes: %i[fees],
)
```

lovrocolic: Updated, I have talked with Mike also in the morning and we should send customer info as well. Thanks for the catch!

vincent-pochet (approved):

PR #319, opened 2022-07-05
feat: Add PDF invoice generation

jdenquin:
- Add PDF Invoice Generation Template
- Add GraphQL `downloadInvoice` mutation

- API endpoint and logic will be added on an other PR


app/services/invoices/generate_service.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module Invoices
+  class GenerateService < BaseService
+    include ActiveSupport::NumberHelper
+
+    def generate_from_api(invoice_id:)
+      invoice = Invoice.find(invoice_id)
```
lovrocolic, line 8: do we need this method at all? As I can see, it's not working yet :)

lovrocolic (approved):

app/services/invoices/generate_service.rb
vincent-pochet: `Invoice.find` will raise an error, you should use `find_by`

app/services/invoices/generate_service.rb
```diff
@@ -0,0 +1,53 @@
...
+      result
+    end
+
+    private
+
+    def generate_pdf(invoice)
+      template = File.read(Rails.root.join('app/views/templates/invoice.slim'), encoding: 'UTF-8')
+      invoice_html = Slim::Template.new { template }.render(invoice)
+
+      pdf_url = "#{ENV['LAGO_PDF_URL']}/forms/chromium/convert/html"
```
vincent-pochet, line 24: Just a suggestion: 
```ruby
URI.join(ENV['LAGO_PDF_URL'], /forms/chromium/convert/html)
```

vincent-pochet (approved):

PR #325, opened 2022-07-06
feat: api pdf generation

jdenquin:
- Add api endpoint `POST /api/v1/invoices/:id/download` that returns directly the invoice with the `file_url` if it exists or generate it asynchronously
- Add the `invoice.generated` webhook that is sent when the invoice pdf file is generated, and return the invoice with the `file_url`in the payload

Depends on https://github.com/getlago/lago-api/pull/319


lovrocolic (approved):
🚀

PR #337, opened 2022-07-19
WIP: feat: Add Wallets Migration and GraphQL Endpoints

jdenquin:
- Add `wallets`
- Add GraphQL resolvers and mutations


db/migrate/20220718083657_create_wallets.rb
```diff
@@ -0,0 +1,24 @@
...
+class CreateWallets < ActiveRecord::Migration[7.0]
+  def change
+    create_table :wallets, id: :uuid do |t|
+      t.references :customer, type: :uuid, null: false, foreign_key: true, index: true
+
+      t.integer :status, null: false
+      t.string :currency, null: false
+
+      t.string :name
+      t.string :rate_amount, null: false
```
rsempe, line 12: Why we're not using `bigint` here?

app/models/wallet.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class Wallet < ApplicationRecord
```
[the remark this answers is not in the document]
jdenquin: not for now!

db/migrate/20220718083657_create_wallets.rb
jdenquin: we'll use `BigDecimal` since we want to support 5 decimals, it's the best way to avoid strange rouding or behaviours

jdenquin: we store `BigDecimal` as `string`! It's the best way we found to avoid issues, do you have more exp with the `BigDecimal` behaviour?

app/graphql/resolvers/wallet_resolver.rb
```diff
@@ -0,0 +1,22 @@
...
+    include AuthenticableApiUser
+    include RequiredOrganization
+
+    description 'Query a single wallet'
+
+    argument :id, ID, required: true, description: 'Uniq ID of the wallet'
+
+    type Types::Wallets::SingleObject, null: true
+
+    def resolve(id: nil)
```
lovrocolic, line 14: is this necessary if id is required?

lovrocolic (approved):

app/graphql/resolvers/wallet_resolver.rb
jdenquin: id is required but it's a ruby formatting, since it's a named params, it's better to initialize it to `nil`!

PR #339, opened 2022-07-21
feat: adapt logic for multiple plans and extend necessary graphQL and API endpoints

lovrocolic:


lovrocolic:
This PR covers part 1:
- adapt current usage for multiple subscriptions
- extend create subscription endpoint (create multiple active subscription, upgrade from given subscription, downgrade from given subscription) - graphQL and API
- update subscription name endpoints (needed for differentiate multiple subscriptions with the same plan) - graphQL and API
- adapt terminate subscription endpoint - graphQL and API
- add warning if new subscription has different currency that older one

app/models/subscription.rb
```diff
@@ -63,6 +63,15 @@ def pending_start_date
     (created_at.end_of_month + 1.day).to_date
   end
 
+  def next_pending_start_date
+    following_subscription = next_subscription
+
+    return unless following_subscription
+    return unless following_subscription.pending?
+
+    (following_subscription.created_at.end_of_month + 1.day).to_date
```
vincent-pochet, line 72: Just wondering if the `end_of_month` also apply when plan recurrence is yearly?

app/controllers/api/v1/customers_controller.rb
```diff
@@ -24,7 +24,9 @@ def create
 
       def current_usage
         service = Invoices::CustomerUsageService
-          .new(nil, customer_id: params[:customer_id], organization_id: current_organization.id)
+          .new(nil, customer_id: params[:customer_id], subscription_id: params[:subscription_id],
+               organization_id: current_organization.id
+          )
```
vincent-pochet, line 29: Just a suggestion on the styling:

```ruby
          .new(
            nil,
            customer_id: params[:customer_id], 
            subscription_id: params[:subscription_id],
            organization_id: current_organization.id,
          )
```

app/services/invoices/customer_usage_service.rb
```diff
@@ -60,8 +62,14 @@ def customer(customer_id: nil)
...
-    def subscription
-      @subscription ||= customer.active_subscription
+    def find_subscription(subscription_id)
+      return nil unless customer
+
+      active_subscriptions = customer.active_subscriptions
+
+      return nil unless active_subscriptions
+
+      active_subscriptions.find_by(id: subscription_id)
```
vincent-pochet, line 72: Since `customer.active_subscriptions` is an ActiveQuery scope, you should do it that way:

```ruby
customer&.active_subscriptions&.find_by(id: subscription_id)
```

It will run only one DB query instead of two.

app/services/subscriptions/create_service.rb
```diff
@@ -42,21 +46,29 @@ def process_create
...
       result.fail_with_validations!(e.record)
     end
 
-    def current_subscription
-      @current_subscription ||= current_customer.subscriptions.active.first
+    def find_current_subscription(subscription_id)
+      return nil unless current_customer
+      return nil unless subscription_id
+
+      current_customer.subscriptions.active.find_by(id: subscription_id)
```
vincent-pochet, line 64: You could make it shorter by doing:

```ruby
current_customer&.active_subscriptions&.find_by(id: subscription_id)
```

Or

```ruby
Subscription.active.find_by(
  id: subscription_id,
  customer_id: current_customer&.id,
)
```

app/services/subscriptions/create_service.rb
```diff
@@ -42,21 +46,29 @@ def process_create
       return result.fail!('missing_argument', 'unable to find customer') unless current_customer
       return result.fail!('missing_argument', 'plan does not exists') unless current_plan
 
+      if currency_missmatch?(current_customer&.active_subscription&.plan, current_plan)
```
vincent-pochet, line 49: Shouldn't we use `@current_subscription` instead of `current_customer&.active_subscription` here?

app/services/subscriptions/create_service.rb
```diff
@@ -32,6 +34,8 @@ def create(**args)
         organization_id: args[:organization_id],
         id: args[:plan_id]&.strip,
       )
+      @name = args[:name]&.strip
```
vincent-pochet, line 37: Just a suggestion : what about passing the `args` to  the `process_create`?
It would allow you to add the striping logic in only one place, could make it easier to add new arguments in the future

app/services/subscriptions/create_service.rb
```diff
@@ -42,21 +46,29 @@ def process_create
...
+
+      current_customer.subscriptions.active.find_by(id: subscription_id)
     end
 
     def handle_subscription
       return upgrade_subscription if upgrade?
       return downgrade_subscription if downgrade?
 
-      current_subscription || create_subscription
+      create_subscription
```
vincent-pochet, line 71: I fear this change introduce a risk of duplication logic when called from the API.

If for some reason a request is sent twice to the endpoint, we will have two subscription instead of only one in the previous situation. Maybe we could have a constraint on the name or another system to handle idem-potency. What do you think?

app/services/subscriptions/terminate_service.rb
```diff
@@ -9,16 +9,6 @@ def terminate(subscription_id)
       process_terminate(subscription)
     end
 
-    def terminate_from_api(organization:, customer_id:)
-      customer = organization.customers.find_by(customer_id: customer_id)
```
vincent-pochet, line 13: Warning here: The `terminate` method does not scope the subscription search to the organization.
It means that someone could potentially terminate the subscription of a customer outside of its organization.

app/controllers/api/v1/subscriptions_controller.rb
```diff
@@ -38,15 +34,58 @@ def terminate
...
-          validation_errors(result)
+          render_error_response(result)
         end
       end
 
+      def update
+        service = Subscriptions::UpdateService.new
+
+        result = service.update(**{
+          id: params[:id],
```
vincent-pochet, line 45: To prevent malicious users to update subscriptions not attached to their organization, we should pass the `organization_id` as well

spec/requests/api/v1/subscriptions_spec.rb
```diff
@@ -45,13 +47,13 @@
...
 
-  describe 'DELETE /subscriptions/terminate' do
+  describe 'delete /subscriptions/terminate' do
     let(:subscription) { create(:subscription, customer: customer, plan: plan) }
 
     before { subscription }
 
     it 'terminates a subscription' do
-      delete_with_token(organization, "/api/v1/subscriptions?customer_id=#{customer.customer_id}")
+      delete_with_token(organization, "/api/v1/subscriptions?subscription_id=#{subscription.id}")
```
vincent-pochet, line 56: Maybe we could turn this route into `DELETE /api/v1/subscriptions/:id`. What do you think?

vincent-pochet (commented):
Fiew comments, but goog job! :rocket: 

Let me know if you have questions about it

app/models/subscription.rb
lovrocolic: Thanks for the catch! Fixed!

app/controllers/api/v1/customers_controller.rb
lovrocolic: Updated!

app/services/invoices/customer_usage_service.rb
lovrocolic: Updated

app/services/subscriptions/create_service.rb
lovrocolic: updated

app/services/subscriptions/create_service.rb
lovrocolic: we will only have @current_subscription when there is upgrade or downgrade. However, there is condition that all subscriptions assigned to one customer should have the same currency. so that's why I'm fetching the last active one (since this cover upgrade/downgrade + adding new plan to customer)

app/services/subscriptions/create_service.rb
lovrocolic: Can you please explain this one a bit more? :)

app/services/subscriptions/terminate_service.rb
lovrocolic: I got it, fixed! Thanks!

app/controllers/api/v1/subscriptions_controller.rb
lovrocolic: Fixed!

spec/requests/api/v1/subscriptions_spec.rb
lovrocolic: Changed

app/services/subscriptions/create_service.rb
vincent-pochet: Sorry, I wasn't clear :)

My suggestion is to change the signature of the `process_create` method to pass all updatable attributes:

```ruby
def process_create(args)
  #  ...
end
```

app/services/subscriptions/create_service.rb
vincent-pochet: I see :thinking: 
Maybe we need a uniq attribute to handle this then. Like the `transaction_id` we have on event.

app/models/subscription.rb
```diff
@@ -63,7 +63,29 @@ def pending_start_date
...
+    return unless following_subscription.pending?
+
+    last_interval_date(self)
+  end
+
   def already_billed?
     fees.subscription_kind.any?
   end
+
+  def last_interval_date(subscription)
```
vincent-pochet, line 79: Do we need to pass, the subscription, can't we use the `self` one instead?

app/services/invoices/customer_usage_service.rb
```diff
@@ -60,8 +62,10 @@ def customer(customer_id: nil)
       )
     end
 
-    def subscription
-      @subscription ||= customer.active_subscription
+    def find_subscription(subscription_id)
+      return nil unless customer
```
vincent-pochet, line 66: Since you are now using the safe travel operator, you can also remove this guard

app/services/subscriptions/create_service.rb
```diff
@@ -42,21 +46,29 @@ def process_create
...
       result
     rescue ActiveRecord::RecordInvalid => e
       result.fail_with_validations!(e.record)
     end
 
-    def current_subscription
-      @current_subscription ||= current_customer.subscriptions.active.first
+    def find_current_subscription(subscription_id)
+      return nil unless current_customer
+      return nil unless subscription_id
```
vincent-pochet, line 62: Since you are now using the safe travel operator, you can also remove these guards

app/models/subscription.rb
lovrocolic: At first I planned to use this method twice (for self and for next subscription) but since I'm just using it on self for now, changed it :)

app/services/invoices/customer_usage_service.rb
lovrocolic: fixed

app/services/subscriptions/create_service.rb
lovrocolic: fixed

app/services/subscriptions/create_service.rb
lovrocolic: added `unique_id`. I tried with `object_id` but since it's reserved word in rails I lost quite some time until I realised why nothing works properly :)

lovrocolic: I had validation so I thought that it would be enough but you are right, make sense to return existing 😅

PR #343, opened 2022-07-25
Set membership context on each request

rsempe:
## Context

We decided to track memberships and not users on [Segment.com](http://segment.com/) for now, as they should be authenticated to perform the actions.
Concerning the API, we only have the information about the organization, we’ll put tracking on the first created membership.

### Changes

Each time a request is made (cloud or self-hosted), the membership id is now added to the current context.
This information is mandatory for sending tracking events to Segment.

### DIFF
--- a/app/controllers/concerns/trackable.rb
+++ b/app/controllers/concerns/trackable.rb
@@ -1,0 +1,25 @@
+module Trackable
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :set_tracing_information
+  end
+
+  def set_tracing_information
+    CurrentContext.membership = "membership/#{membership_id || 'unidentifiable'}"
+  end
+
+  def membership_id
+    return nil unless current_organization
+
+    # NOTE: When doing requests from the API, we haven't the current user information.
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/trackable.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def set_tracing_information
+    CurrentContext.membership = "membership/#{membership_id || 'unidentifiable'}"
+  end
+
+  def membership_id
+    return nil unless current_organization
+
+    # When doing requests from the API (self-hosted), we haven't the current user information.
+    # In that case, we add tracing information on the first created membership of the organization.
```
vincent-pochet, line 16: Could you prefix the comment with a `NOTE`:

```ruby
# NOTE: When doing....
```

Also, about the comment itself, not sure the `(self-hosted)` apply here, as the logic also apply on cloud API

vincent-pochet (approved):
Looks good! just a small comment

jdenquin (approved):

PR #345, opened 2022-07-26
fix: canceled next subscription when terminating a downgraded one

vincent-pochet:
### Context

1. When customer got an active plan A and we decide to downgrade this plan B, a new line appears and the plan B is in `pending` status.
2. If we terminate the active plan A, the plan B is still in pending in the base and will be active on the next billable period

### Expected behavior

Terminating the active plan A should also cancel the pending plan B

### Changes

Check for presence of pending next subscription when terminating and canceled it

### DIFF
--- a/spec/services/subscriptions/terminate_service_spec.rb
+++ b/spec/services/subscriptions/terminate_service_spec.rb
@@ -46,7 +46,7 @@ RSpec.describe Subscriptions::TerminateService do
 
       before { next_subscription }
 
-      it 'canceled the next subscription' do
+      it 'cancels the next subscription' do
         result = terminate_service.terminate(subscription.id)
 
         aggregate_failures do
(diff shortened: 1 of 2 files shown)


lovrocolic (approved):

spec/services/subscriptions/terminate_service_spec.rb
```diff
@@ -33,6 +33,28 @@
...
+      end
+
+      before { next_subscription }
+
+      it 'canceled the next subscription' do
+        result = terminate_service.terminate(subscription.id)
+
+        aggregate_failures do
+          expect(result).to be_success
+          expect(next_subscription.reload).to be_canceled
```
rsempe, line 54: What about writing it like this?

```ruby
expect { terminate_service.terminate(subscription.id) }
  .to change { next_subscription.reload.status }.from('pending').to('canceled')
```

rsempe (approved):

spec/services/subscriptions/terminate_service_spec.rb
vincent-pochet: @rsempe it good but to be honest, I find it a bit harder to read

PR #347, opened 2022-07-26
feat(segment): Add some tracking on Segment (part 2)

rsempe:
### Context

### Objective

The goal of this PR is to send a tracking event to Segment on:
- Customer creation
- Subscription creation
- Coupon creation
- Coupon assignment
- Addon creation
- Addon assignment

### Segment Screenshots

Capture d’écran 2022-07-26 à 15 07 35
Capture d’écran 2022-07-26 à 15 09 35
Capture d’écran 2022-07-26 à 15 12 00
Capture d’écran 2022-07-26 à 15 23 16
Capture d’écran 2022-07-26 à 15 24 19
Capture d’écran 2022-07-26 à 15 24 46

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -157,6 +157,7 @@ module Subscriptions
     def subscription_type
       return 'downgrade' if downgrade?
       return 'upgrade' if upgrade?
+
       'create'
     end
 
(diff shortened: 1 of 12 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -152,5 +153,26 @@ def pending_subscription?
     def cancel_pending_subscription
       current_subscription.next_subscription.mark_as_canceled!
     end
+
+    def subscription_type
+      return 'downgrade' if downgrade?
+      return 'upgrade' if upgrade?
+      'create'
```
jdenquin, line 160: a small line return before it!

jdenquin (approved):
🚀

lovrocolic (approved):

PR #349, opened 2022-07-27
feat(segment): Identify user on signup and signin

rsempe:
### Context

### Objective

The goal of this PR is to send an identified event to Segment on:
- when the user signs up
- when the user logs in

### Segment Screenshots

Capture d’écran 2022-07-27 à 16 00 21
Capture d’écran 2022-07-27 à 16 00 58

### DIFF
--- a/app/services/users_service.rb
+++ b/app/services/users_service.rb
@@ -4,7 +4,14 @@ class UsersService < BaseService
   def login(email, password)
     result.user = User.find_by(email: email)&.authenticate(password)
     result.token = generate_token if result.user
-    result.fail!('incorrect_login_or_password') unless result.user
+
+    unless result.user
+      result.fail!('incorrect_login_or_password')
+      return result
+    end
+
+    # Note: We're tracking the first membership linked to the user.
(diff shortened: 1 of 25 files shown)


app/services/users_service.rb
```diff
@@ -4,7 +4,14 @@ class UsersService < BaseService
   def login(email, password)
     result.user = User.find_by(email: email)&.authenticate(password)
     result.token = generate_token if result.user
-    result.fail!('incorrect_login_or_password') unless result.user
+
+    unless result.user
+      result.fail!('incorrect_login_or_password')
+      return result
+    end
```
vincent-pochet, line 11: Small tip: `fail!` method is returning the result object so that you can do:

```suggestion
    return result.fali!('incorrect_login_or_password') unless result.user
```

vincent-pochet (approved):
:rocket:

app/services/users_service.rb
rsempe: Nice, thanks 👍

PR #352, opened 2022-07-28
feat: Update Subscription data model to support anniversary date

vincent-pochet:
## Context

## Objective

The objective of this changes is to prepare the introduction of the new billing time:

- Add `billing_time` enum to subscription
  - Possible values: `anniversary` / `calendar`
  - Default value: `calendar` to keep it compatible with existing subscription / API

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -19,9 +19,9 @@ ActiveRecord::Schema[7.0].define(version: 2022_07_28_144707) do
   create_table "active_storage_attachments", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.string "name", null: false
     t.string "record_type", null: false
+    t.uuid "record_id"
     t.uuid "blob_id", null: false
     t.datetime "created_at", null: false
-    t.uuid "record_id"
     t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
   end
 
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -19,11 +19,10 @@
   create_table "active_storage_attachments", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.string "name", null: false
     t.string "record_type", null: false
-    t.uuid "record_id", null: false
     t.uuid "blob_id", null: false
     t.datetime "created_at", null: false
+    t.uuid "record_id"
```
rsempe, line 24: Some changes are not related to this PR, is that expected?

db/schema.rb
```diff
@@ -344,10 +344,10 @@
...
     t.string "currency", null: false
     t.string "name"
-    t.string "rate_amount", null: false
-    t.string "credits_balance", default: "0.00", null: false
-    t.string "balance", default: "0.00", null: false
-    t.string "consumed_credits", default: "0.00", null: false
+    t.decimal "rate_amount", precision: 5, default: "0", null: false
+    t.decimal "credits_balance", precision: 5, default: "0", null: false
+    t.decimal "balance", precision: 5, default: "0", null: false
+    t.decimal "consumed_credits", precision: 5, default: "0", null: false
```
rsempe, line 350: Same here

rsempe (approved):

db/schema.rb
vincent-pochet: :thinking: weird...

I'll try to fix it

PR #354, opened 2022-07-29
feat: Invoice logic refactoring for multiple plans

lovrocolic:


app/services/invoices/create_service.rb
```diff
@@ -2,36 +2,41 @@
...
 
     def create
       ActiveRecord::Base.transaction do
-        invoice = Invoice.find_or_create_by!(
-          subscription: subscription,
-          from_date: from_date,
-          to_date: to_date,
-          charges_from_date: charges_from_date,
-          issuing_date: issuing_date,
+        invoice = Invoice.create!(
```
lovrocolic, line 16: Do you think that is fine to use only `create!` here? SInce we removed some fields I was afraid that we would return wrong invoice in some cases. (e.g. one subscription was terminated and billed for pay in arrear in the same day when invoice for another subscription is billed)

app/services/invoices/create_service.rb
```diff
@@ -2,36 +2,41 @@
...
+          customer: customer,
+          issuing_date: (Time.zone.at(timestamp) - 1.day).to_date,
```
lovrocolic, line 18: I talked with Raffi and we should probably simplify logic for issuing date and use invoice creation date here

rsempe:
Closed in favor of https://github.com/getlago/lago-api/pull/363.

PR #358, opened 2022-08-02
refactor: Use named parameters for BaseService::Result#fail!

rsempe:
### Changes

The goal of this PR is to add named parameters to the `BaseService::Result#fail!` method.

So that we can have:
`result.fail!(code: 'missing_mandatory_param', details: arguments)`
Instead of
`result.fail!('missing_mandatory_param', nil, arguments)`


vincent-pochet (approved):
Nice one!

PR #359, opened 2022-08-02
fix: make Segment optional

prognostikos:
Before this change Rails would not boot if `SEGMENT_WRITE_KEY` was not
set in the environment. Now we re-use the `LAGO_DISABLE_SEGMENT`
environment variable that was already being checked in the Segment jobs
to disable this initializer if it's set.


prognostikos:
Looks like `LAGO_DISABLE_SEGMENT` will also have to be updated in the docker-compose files in getlago/lago as well

rsempe (approved):

vincent-pochet (approved):

rsempe:
Thanks @prognostikos

PR #360, opened 2022-08-02
feat: Add new date service to compute bounds for subscription and charge fees

vincent-pochet:
## Context

## Changes

The objective of this pull request is to introduce a new `Subscriptions::DatesService`.
This service is responsible for computing all bounds for invoice and related subscription and charge fees.

It exposes few methods:
- `from_date`: compute the lower bound of an invoicing period based on a provided timestamp and subscription
- `to_date`: compute the upper bound of an invoicing period based on a provided timestamp and subscription
- `charges_from_date`: compute the lower bound of an invoicing period for charges based on a provided timestamp and subscription. It allows to handle the `bill_charge_monthly` case for yearly plans
- `next_end_of_period`: compute the next end of period for a given subscription and date 

*Note:* The service is not yet used in invoicing services. This will come with a latter change

## Related resources

These follows the previous one https://github.com/getlago/lago-api/pull/352 to update the subscription data model.


app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,304 @@
...
+
+module Subscriptions
+  class DatesService < BaseService
+    def initialize(subscription, timestamp)
+      @subscription = subscription
+
+      # NOTE: timestamp is the billing day
+      #       It should be the end of the billing period + 1 day
+      #       When subscription is terminated, it is the termination day
+      @timestamp = timestamp
```
rsempe, line 11: Is it not simpler to directly send and use the `billing_date` instead of a `timestamp`? 🤔

app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,304 @@
...
+                   when :monthly
+                     monthly_from_date
+                   when :yearly
+                     yearly_from_date
+                   else
+                     raise NotImplementedError
+      end
+
+      # NOTE: On first billing period, subscription might start after the computed start of period
+      #       ei: if we bill on beginning of period, and user registered on the 15th, the invoice should
```
rsempe, line 31: ie:

app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,304 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  class DatesService < BaseService
```
rsempe, line 4: Do we need to inherit from `BaseService`?

app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,304 @@
...
+    def initialize(subscription, timestamp)
+      @subscription = subscription
+
+      # NOTE: timestamp is the billing day
+      #       It should be the end of the billing period + 1 day
+      #       When subscription is terminated, it is the termination day
+      @timestamp = timestamp
+
+      super(nil)
```
rsempe, line 13:
```suggestion
      super
```

app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,304 @@
...
+                 when :yearly
+                   yearly_to_date
+                 else
+                   raise NotImplementedError
+      end
+
+      if subscription.terminated? && @to_date > subscription.terminated_at
+        # NOTE: When subscription is terminated, we cannot generate an invoice for a period after the termination
+        @to_date = if %i[pending active].include?(subscription.next_subscription&.status&.to_sym)
+          subscription.terminated_at.to_date - 1.day
```
rsempe, line 55: Not sure to understand `- 1 day` here.. 🤔 

For instance, if `next_subscription` is active for more than 1 day.

app/services/subscriptions/dates_service.rb
vincent-pochet: Actually, not at all :+1:

app/services/subscriptions/dates_service.rb
vincent-pochet: It is to cover the upgrade case. When upgrading, we will have `previous_subscription.terminated_at = next_subscription.started_at` and we do not want to bill fees twice on both subscriptions.

That  said, you are right, we might have here if terminated_at == started_at on the terminated subscription. I'll fix this one.

*Note:* Keep in mind that fees for the period will be bill on the next plan
(That's something we need to challenge but it is the current behavior)

app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,297 @@
...
+                   when :monthly
+                     monthly_from_date
+                   when :yearly
+                     yearly_from_date
+                   else
+                     raise NotImplementedError
+      end
+
+      # NOTE: On first billing period, subscription might start after the computed start of period
+      #       ei: if we bill on beginning of period, and user registered on the 15th, the invoice should
```
rsempe, line 28: ie:

app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,297 @@
...
+                       billing_date - 1.month
+                     when :yearly
+                       billing_date - 1.year
+                     else
+                       raise NotImplementedError
+      end
+    end
+
+    def subscription_day_name
+      subscription_date.strftime('%A').downcase.to_sym
```
rsempe, line 118: We can probably memoize this

app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,297 @@
...
+    def yearly_from_date
+      if terminated_pay_in_arrear?
+        return subscription.anniversary? ? previous_yearly_anniversary_day(billing_date) : billing_date.beginning_of_year
+      end
+
+      subscription.anniversary? ? previous_yearly_anniversary_day(base_date) : base_date.beginning_of_year
+    end
+
+    def yearly_to_date
+      return from_date.end_of_year if subscription.calendar? || subscription_date.day == 1
```
rsempe, line 171: `day` is checking the first day of the week, I think we want the first day of the year, no?

```suggestion
      return from_date.end_of_year if subscription.calendar? || subscription_date.yday == 1
```

app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,297 @@
...
+
+      build_date(year, month, day)
+    end
+
+    def previous_monthly_anniversary_day(date)
+      year = nil
+      month = nil
+      day = subscription_date.day
+
+      if date.day <= subscription_date.day
```
rsempe, line 185: Same here, we have to use `yday`, no?

```suggestion
      if date.yday <= subscription_date.yday
```

app/services/subscriptions/dates_service.rb
```diff
@@ -0,0 +1,297 @@
...
+      year = date.month < subscription_date.month ? date.year - 1 : date.year
+      month = subscription_date.month
+      day = subscription_date.day
+
+      build_date(year, month, day)
+    end
+
+    # NOTE: Handle leap years and anniversary date > 28
+    def build_date(year, month, day)
+      day_count_in_month = Time.days_in_month(month, year)
```
rsempe, line 206:
```suggestion
      days_count_in_month = Time.days_in_month(month, year)
```

rsempe:
Just a general thought: Do you think extracting that logic into 3 _smaller_ classes (weekly, monthly and yearly) can be interesting?

vincent-pochet:
> Just a general thought: Do you think extracting that logic into 3 _smaller_ classes (weekly, monthly and yearly) can be interesting?

Yeah, good suggestion, it could make it a bit easier to read. Let's try it!

spec/services/subscriptions/dates/monthly_service_spec.rb
```diff
@@ -0,0 +1,286 @@
...
+
+      context 'when billing last month of year' do
+        let(:billing_date) { DateTime.parse('04 Jan 2022') }
+
+        it 'returns the previous month month day' do
+          expect(result).to eq('2022-01-01')
+        end
+      end
+
+      context 'when billing subscription day does not extist in the month' do
```
jdenquin, line 170: small vincent here!

jdenquin (approved):
just a small typo
🚀

rsempe (approved):

PR #362, opened 2022-08-03
fix(multiple_plans): Use subscription for cache key on customer usage

rsempe:
### Context

We want to be able to create multiple plans for a specific customer.

### Changes

The goal of this PR is to fix the name of the cache key to interact with the subscription id instead of the customer id.


jdenquin (approved):
🚀

PR #371, opened 2022-08-09
fix(invoices): Fix permissions for invoices show and download on API

vincent-pochet:
## Context

`GET /api/v1/invoices/:id` and `POST /api/v1/invoices/:id/download` does not apply permissions when looking for the invoice. It should be fixed as it is a potential security hole.

## Changes

Invoices search should be scoped to the current organization


rsempe (approved):
👍

PR #376, opened 2022-08-10
feat: create wallet transactions and generate invoice upon adding paid credits

jdenquin:
## Context

Updating ledger logic

## Description

We create wallet transactions (both for free credits and paid credits) upon creating wallet and during top-up.
For free credits we instantly update wallet's balance (and create settled transaction) and DO NOT create invoice. However for paid credits we create invoice and pending transaction. Balance will be updated only when invoice status become succeeded.

## Related Task

-

## How Has This Been Tested?

Automations tests are added and Tech QA is performed.
Product QA will be performed after completing full feature

## Notice
After discussion with @jdenquin we decided to add polymorphic association on `fee` so that we don't need to add new association with each new invoiceable object.
We will handle old `fee` data through migration in separate task (add fee_type and polymorphic association data)


app/services/wallets/create_service.rb
```diff
@@ -16,6 +18,8 @@ def create(**args)
       result.wallet = wallet
 
       WalletTransactions::CreateJob.perform_later(
```
ansmonjol, line 20: Shall we create them instantly instead of later ?
I'm wondering cause on wallet creation action, we re-fetch the wallet list right after the create and expect the transactions to be present in the UI.
Or is this line creating them with a pending status and "just" executing them later (creating the "real" transaction) later ?

jdenquin: I think it's a different behavior, it's better to have them created async since it's an other business scope, we should definitely think about a websocket :p

ansmonjol: ok! We'll probably add a disclaimer about potential delay in the UI, either in the wallet page or the creation modal.

Thanks for your answer

app/services/invoices/payments/stripe_service.rb
```diff
@@ -148,6 +150,21 @@ def update_invoice_status(status)
         invoice.update!(status: status)
       end
 
+      def update_wallet(status)
+        return unless status == 'succeeded'
+        return unless invoice.invoice_type == 'credit'
```
jdenquin, line 155: this one should be the first return!

app/services/wallet_transactions/create_service.rb
```diff
@@ -5,18 +5,63 @@ class CreateService < BaseService
...
-      # TODO
+    def handle_paid_credits(wallet:, paid_credits:)
+      paid_credits_amount = BigDecimal(paid_credits)
+
+      return if paid_credits_amount.zero?
+
+      wallet_transaction = WalletTransaction.create!(
+        wallet: wallet,
+        transaction_type: :inbound,
+        amount: BigDecimal(wallet.rate_amount) * paid_credits_amount,
```
jdenquin, line 34: `wallet.rate_amount` shoud already bee a decimal no?

app/services/wallets/balance/decrease_service.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Wallets
+  module Balance
+    class DecreaseService < BaseService
```
jdenquin, line 5: I think this service is not used at this step, btw we also should update the `consumed_credits_amount` during this step, wdyt?

app/services/webhooks/paid_credit_service.rb
```diff
@@ -0,0 +1,27 @@
...
+    def object_serializer
+      ::V1::InvoiceSerializer.new(
+        object,
+        root_name: 'invoice',
+        includes: %i[customer fees],
+      )
+    end
+
+    def webhook_type
+      'invoice.paid_credit_added'
```
jdenquin, line 20: is this type okay for you @michaelponrajah ?

jdenquin (commented):
Some comments to check, otherwise good job!

rsempe:
@lovrocolic Is it possible to split this PR into several smaller ones? 🙏 🥲

app/graphql/mutations/wallet_transactions/create.rb
```diff
@@ -13,12 +13,26 @@ class Create < BaseMutation
...
+          paid_credits: args[:paid_credits],
+          granted_credits: args[:granted_credits],
+        )
+
+        result.success? ? result.wallet_transactions : result_error(result)
+      end
+
+      private
+
+      def customer(wallet_id)
```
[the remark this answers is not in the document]
lovrocolic: yep not needed here, fixed

app/graphql/resolvers/coupons_resolver.rb
```diff
@@ -21,7 +21,7 @@ def resolve(ids: nil, page: nil, limit: nil, status: nil)
         .coupons
         .order_by_status_and_expiration
         .page(page)
-        .limit(limit)
+        .per(limit)
```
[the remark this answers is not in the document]
lovrocolic: Ok, will not be incuded

app/models/fee.rb
```diff
@@ -16,47 +17,34 @@ class Fee < ApplicationRecord
   monetize :amount_cents
   monetize :vat_amount_cents
 
+  TYPE = %i[charge add_on subscription credit].freeze
```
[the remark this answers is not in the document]
lovrocolic: fixed

app/services/invoices/payments/stripe_service.rb
lovrocolic: fixed

app/services/wallet_transactions/create_service.rb
lovrocolic: nope, it's string

app/services/wallets/balance/decrease_service.rb
lovrocolic: yep fixed and moved in other PR

PR #383, opened 2022-08-16
feat(free-units): Calculate running total on sum aggregation

rsempe:
## Context

We want to add the ability to:
- Apply a fixed fee per `distinct transaction`
- Apply free units on the `number of distinct transactions` or on `the monetary total amount`

## Description

This is the first step of the feature, the goal of this PR is to calculate running total from sum aggregator service.

The running total of the events is the cumulative sum of the event’s `properties.field_name`.

## Related Task

https://github.com/getlago/lago/issues/92

## How Has This Been Tested?

A full QA will be performed when the feature will be 100% implemented.

### DIFF
--- a/app/services/billable_metrics/aggregations/base_service.rb
+++ b/app/services/billable_metrics/aggregations/base_service.rb
@@ -9,7 +9,7 @@ module BillableMetrics
         @subscription = subscription
       end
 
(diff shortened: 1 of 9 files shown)


jdenquin (approved):
🚀

PR #385, opened 2022-08-16
feat: Add pay in advance logic into date service 

vincent-pochet:
## Context

## Description

Following the previous introduction of a date service (see https://github.com/getlago/lago-api/pull/360) to extract logic to compute fees boundaries, few changes are required in order to use this service in the `Fees::ChargeService` and `Fees::SubscriptionService`:

- `from_date` and `to_date`  should return the boundaries of the coming subscription period when plan is payed in advance. It will kill the need for boundaries rewriting in `Fees::SubscriptionService` (see https://github.com/getlago/lago-api/blob/main/app/services/fees/subscription_service.rb#L152 for example)
- Since `to_date` is also used for charge, a `charges_to_date` is required as charges are always billed on the passed billing period no matter if the plan is payed in advance or in arrear
- a `single_day_price` method is required to compute the cost of a single subscription day based on the plan cost and the period duration
- a `current_usage` flag is required to return forecast dates for charges boundaries

### DIFF
--- a/spec/services/subscriptions/dates/monthly_service_spec.rb
+++ b/spec/services/subscriptions/dates/monthly_service_spec.rb
@@ -53,7 +53,7 @@ RSpec.describe Subscriptions::Dates::MonthlyService, type: :service do
         context 'when plan is pay in advance' do
           let(:pay_in_advance) { true }
 
-          it 'returns the beginning of the month month' do
+          it 'returns the beginning of the month' do
             expect(result).to eq('2022-03-01')
           end
         end
@@ -64,7 +64,7 @@ RSpec.describe Subscriptions::Dates::MonthlyService, type: :service do
       let(:billing_time) { :anniversary }
       let(:billing_date) { DateTime.parse('03 Mar 2022') }
 
-      it 'returns the previous month month day' do
+      it 'returns the day in the previous month day' do
         expect(result).to eq('2022-02-02')
       end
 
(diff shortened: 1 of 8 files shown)


spec/services/subscriptions/dates/monthly_service_spec.rb
```diff
@@ -88,8 +88,8 @@
         context 'when plan is pay in advance' do
           let(:pay_in_advance) { true }
 
-          it 'returns the previous month month day' do
-            expect(result).to eq('2022-02-02')
+          it 'returns the current month month day' do
```
rsempe, line 91: `month month` 😄

spec/services/subscriptions/dates/monthly_service_spec.rb
```diff
@@ -247,6 +252,76 @@
           expect(result).to eq(subscription.started_at.to_date.to_s)
         end
       end
+
+      context 'when plan is pay in advance' do
+        let(:pay_in_advance) { true }
+        let(:subscription_date) { DateTime.parse('02 Feb 2020') }
+
+        it 'returns the start of the previous period' do
+          expect(result).to eq('2022-02-02')
```
rsempe, line 261: Is it correct? It's not the previous period in that case, but the current, no?

rsempe (approved):
LGTM 🚀 - The QA part will be fun! 😄

spec/services/subscriptions/dates/monthly_service_spec.rb
vincent-pochet: @rsempe, thanks for the review!

It looks correct:
We are in the test of `charge_from_date` method. Charges are billed on the finished previous period no matter if the plan is payed in advance or in arrears. 
The current period case would apply for the `from_date` method

PR #388, opened 2022-08-17
fix: Add validation for Plan and Charges currencies

jdenquin:
## Context

We can create a Plan with somes charges that have a different currency from the Plan, it's not correct and can leads to errors

## Description

- Add a validation on charges to check that the currency is the same as the linked plan

## How Has This Been Tested?

- Added some unit tests

### DIFF
--- a/app/models/charge.rb
+++ b/app/models/charge.rb
@@ -22,7 +22,7 @@ class Charge < ApplicationRecord
 
   enum charge_model: CHARGE_MODELS
 
-  validates :amount_currency, inclusion: { in: currency_list }
+  validate :validate_currency
   validate :validate_amount, if: :standard?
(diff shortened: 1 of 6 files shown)


app/models/charge.rb
```diff
@@ -57,4 +57,10 @@ def validate_percentage
 
     validation_result.error.each { |error| errors.add(:properties, error) }
   end
+
+  def validate_currency
+    return if plan.amount_currency == amount_currency
```
rsempe, line 62: Why not directly setting the currency of the charge based on the currency of the plan in that case?

Not sure to understand what we're trying to solve here.. 🤔

jdenquin: the currency can be set in the payload, we may change this if we want to force the currency of the plan, but it's out of the bug scope for me

PR #390, opened 2022-08-18
feat(free-units): Calculate percentage based on fixed amount and free units

rsempe:
## Context

We want to add the ability to:
- Apply a fixed fee per distinct transaction
- Apply free units on the number of distinct transactions or on the monetary total amount

## Description

This is the third step of adding fixed amount and free units to the percentage charge.

The goal of this PR is to:
- Calculate percentage charge based on fixed amount
- Calculate percentage charge based on free units (per aggregation or per events)

## Related Task

https://github.com/getlago/lago/issues/92

## How Has This Been Tested?

A full QA will be performed when the feature will be 100% implemented.

### DIFF
--- a/app/services/charges/charge_models/graduated_service.rb
+++ b/app/services/charges/charge_models/graduated_service.rb
@@ -9,7 +9,7 @@ module Charges
         charge.properties.map(&:with_indifferent_access)
       end
 
-      def compute_amount(value)
+      def compute_amount
         ranges.reduce(0) do |result_amount, range|
           flat_amount = BigDecimal(range[:flat_amount])
           per_unit_amount = BigDecimal(range[:per_unit_amount])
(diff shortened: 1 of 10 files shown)


app/services/charges/charge_models/graduated_service.rb
```diff
@@ -9,18 +9,18 @@ def ranges
...
 
           # NOTE: Add flat amount to the total
           result_amount += flat_amount unless value.zero?
 
-          units = compute_range_units(range[:from_value], range[:to_value], value)
+          units = compute_range_units(range[:from_value], range[:to_value])
           result_amount += units * per_unit_amount
 
-          # NOTE: value is between the bounds of the current range,
+          # NOTE: aggregation_result.aggregation is between the bounds of the current range,
```
vincent-pochet, line 23: What about adding a delegate to units in the base class?

```ruby
delegate :units, to :result
```

This way we could use units in this comments as in any other place where we relies on `value`.

vincent-pochet (approved):
:rocket:

app/services/charges/charge_models/graduated_service.rb
rsempe: Good idea!

PR #392, opened 2022-08-19
Remove plan charge currency

jdenquin:
**Context**

`amount_currency` on `Charges` is misleading and not used, it should always be the same as the plan's currency.

**Description**

- Remove the `amount_currency` for `charges` in payload
- Make `amount_currency` nullable on `charges` before removing it later

### DIFF
--- a/app/serializers/v1/charge_serializer.rb
+++ b/app/serializers/v1/charge_serializer.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-#
+
 module V1
   class ChargeSerializer < ModelSerializer
     def serialize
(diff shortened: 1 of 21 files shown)


app/serializers/v1/charge_serializer.rb
```diff
@@ -1,13 +1,12 @@
...
-#
+
 module V1
   class ChargeSerializer < ModelSerializer
     def serialize
       {
         lago_id: model.id,
         lago_billable_metric_id: model.billable_metric_id,
         created_at: model.created_at.iso8601,
-        amount_currency: model.amount_currency,
```
rsempe, line 10: this seems to be a breaking change

rsempe (approved):

app/serializers/v1/charge_serializer.rb
jdenquin: it is!

vincent-pochet (approved):

PR #394, opened 2022-08-19
feat: Add validation logic for the volume charge model

vincent-pochet:
## Context

We want to introduce a new charge model to charge users based on a consumed volume.

The existing graduated pricing does not fit this need because it creates tier, and each tier has its own price.
The volume price works differently, it creates tiers, but the whole number of units is multiplied by the current tier. Then, **the number of units consumed determine the total price for each unit.**

## Description

This pull request is the first step of the volume pricing charge model.
It adds the logic to validate a charge with a `volume` charge model.

### DIFF
--- a/app/services/charges/validators/volume_service.rb
+++ b/app/services/charges/validators/volume_service.rb
@@ -13,7 +13,8 @@ module Charges
 
         next_from_value = 0
         ranges.each_with_index do |range, index|
-          errors << :invalid_amount unless valid_amounts?(range)
+          errors << :invalid_per_unit_amount unless valid_amount?(range[:per_unit_amount])
+          errors << :invalid_flat_amount unless valid_amount?(range[:flat_amount])
           errors << :invalid_ranges unless valid_bounds?(range, index, next_from_value)
 
           next_from_value = (range[:to_value] || 0) + 1
@@ -30,9 +31,8 @@ module Charges
         charge.properties['ranges']&.map(&:with_indifferent_access)
       end
 
-      def valid_amounts?(range)
-        ::Validators::DecimalAmountService.new(range[:per_unit_amount]).valid_amount? &&
-          ::Validators::DecimalAmountService.new(range[:flat_amount]).valid_amount?
+      def valid_amount?(amount)
+        ::Validators::DecimalAmountService.new(amount).valid_amount?
       end
 
       def valid_bounds?(range, index, next_from_value)
--- a/app/services/charges/validators/volume_service.rb
+++ b/app/services/charges/validators/volume_service.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Charges
+  module Validators
+    class VolumeService < Charges::Validators::BaseService
+      def validate
+        errors = []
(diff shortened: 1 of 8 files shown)


app/services/charges/validators/volume_service.rb
```diff
@@ -0,0 +1,46 @@
...
+          errors << :missing_ranges
+          return result.fail!(code: :invalid_properties, message: errors)
+        end
+
+        next_from_value = 0
+        ranges.each_with_index do |range, index|
+          errors << :invalid_amount unless valid_amounts?(range)
+          errors << :invalid_ranges unless valid_bounds?(range, index, next_from_value)
+
+          next_from_value = (range[:to_value] || 0) + 1
```
rsempe, line 19: So it supposes that volume charge model only works with integers, right?

I'm not even sure that having decimals volume (from 0 to 1.5 for instance) makes any sense... 😅

spec/factories/charge_factory.rb
```diff
@@ -40,5 +40,12 @@
         }
       end
     end
+
+    factory :volume_charge do
+      charge_model { 'volume' }
+      properties do
+        { ranges: [] }
```
rsempe, line 47: Can we define a valid factory?

I mean, here it will return `:missing_ranges` each time we're trying to use that factory, right?

app/services/charges/validators/volume_service.rb
vincent-pochet: Well, we could probably handle decimal bounds but we decided not to for now as it make it more complex for bounds validation

spec/factories/charge_factory.rb
vincent-pochet: Good catch

spec/services/charges/validators/volume_service_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+          },
+          {
+            from_value: 11,
+            to_value: 20,
+            per_unit_amount: '10',
+            flat_amount: '20',
+          },
+          {
+            from_value: 21,
+            to_value: nil,
```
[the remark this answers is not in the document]
vincent-pochet: Maybe I'm wrong, bug I don't think infinity is a valid JSON value whereas nil is. 
We could probably add it as some type of constant, but I think it's better not to deal with different data type for the same field. What do you think?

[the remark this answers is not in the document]
rsempe: I agree - What about having `-1` instead of `nil` then? 🤔

_Anyway, that's not a blocker for me, just wanted to distinguish `infinity` from `nil`_

rsempe (approved):

spec/services/charges/validators/volume_service_spec.rb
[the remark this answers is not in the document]
vincent-pochet: The issue it that having `-1` or another value would require the front to handle it as an exception, that's why I think nil still remain the best "ok" value for this.
But hanks for bringing up this topic :)

PR #400, opened 2022-08-23
feat: create wallet transactions base

lovrocolic:
## Context

Base change for handling wallet transactions

## Description

First step in reviewing


app/services/wallet_transactions/validate_service.rb
```diff
@@ -31,14 +31,12 @@ def valid?
...
 
     def valid_wallet?
-      current_wallet = Wallet.find_by(
-        id: args[:wallet_id],
-        customer_id: args[:customer_id],
-      )
+      organization = Organization.find_by(id: args[:organization_id])
 
-      return 'wallet_not_found' unless current_wallet
+      result.current_wallet = organization.wallets.find_by(id: args[:wallet_id])
```
rsempe, line 36: What about preventing having an exception if organization is not found?

rsempe, line 36: What about avoiding raising an exception if the organization is not found?

app/services/wallet_transactions/create_service.rb
```diff
@@ -5,18 +5,42 @@ class CreateService < BaseService
...
+
+      return if granted_credits_amount.zero?
+
+      WalletTransaction.create!(
+        wallet: wallet,
+        transaction_type: :inbound,
+        amount: wallet.rate_amount * granted_credits_amount,
+        credit_amount: granted_credits_amount,
+        status: :settled,
+        settled_at: Time.zone.now,
```
rsempe, line 40:
```suggestion
        settled_at: Time.current,
```

app/services/wallets/balance/increase_service.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Balance
+    class IncreaseService < BaseService
+      def initialize(wallet:, credits_amount:)
+        super(nil)
+
+        @wallet = wallet
+        @credits_amount = credits_amount
+      end
+
+      def call
```
rsempe, line 13: 😍

app/services/wallets/balance/increase_service.rb
```diff
@@ -0,0 +1,28 @@
...
+        amount = BigDecimal(wallet.rate_amount) * credits_amount
+
+        wallet.update!(
+          balance: BigDecimal(wallet.balance) + amount,
+          credits_balance: BigDecimal(wallet.credits_balance) + credits_amount,
+          last_balance_sync_at: Time.zone.now,
```
rsempe, line 19:
```suggestion
          last_balance_sync_at: Time.current,
```

app/services/wallet_transactions/create_service.rb
```diff
@@ -5,18 +5,42 @@ class CreateService < BaseService
...
+      )
+
+      Wallets::Balance::IncreaseService.new(wallet: wallet, credits_amount: granted_credits_amount).call
```
rsempe, line 43: Do we want to use a transaction for creating the wallet transaction and updating the wallet?

If we have an issue when updating the wallet here, do we still want to persist the wallet transaction?

db/schema.rb
```diff
@@ -359,10 +359,10 @@
...
     t.string "currency", null: false
     t.string "name"
-    t.decimal "rate_amount", precision: 5, default: "0", null: false
-    t.decimal "credits_balance", precision: 5, default: "0", null: false
-    t.decimal "balance", precision: 5, default: "0", null: false
-    t.decimal "consumed_credits", precision: 5, default: "0", null: false
+    t.string "rate_amount", null: false
+    t.string "credits_balance", default: "0.00", null: false
+    t.string "balance", default: "0.00", null: false
+    t.string "consumed_credits", default: "0.00", null: false
```
rsempe, line 365: You need to reset your database to avoid these changes ;-)

spec/jobs/wallet_transactions/create_job_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    allow(wallet_transaction_create_service).to receive(:create)
+
+    described_class.perform_now(
+      organization_id: '123456',
+      wallet_id: '123456',
+      paid_credits: '1.00',
+      granted_credits: '1.00',
+    )
+
+    expect(wallet_transaction_create_service).to have_received(:create)
```
rsempe, line 22: Can we ensure the parameters as well by using `.with(...)`?

spec/services/wallet_transactions/create_service_spec.rb
```diff
@@ -22,7 +22,8 @@
       {
         name: 'New Wallet',
         wallet_id: wallet.id,
-        customer_id: customer.id,
+        customer_id: customer.customer_id,
+        organization_id: organization.id,
```
rsempe, line 26: Can we add some tests concerning the changes about `granted_credits`?

spec/services/wallets/balance/increase_service_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  let(:credits_amount) { BigDecimal('4.5') }
+
+  before { wallet }
+
+  describe '.call' do
+    it 'updates wallet balance' do
+      create_service.call
+
+      expect(wallet.reload.balance).to eq('14.5')
+      expect(wallet.reload.credits_balance).to eq('14.5')
```
rsempe, line 18:
```suggestion
          expect {
            create_service.call
          }.to change { wallet.reload.balance }.from(nil).to('14.5')
          .and change { wallet.credits_balance }.from(nil).to('14.5')
```

spec/graphql/mutations/wallets/create_spec.rb
```diff
@@ -31,7 +31,7 @@
       query: mutation,
       variables: {
         input: {
-          customerId: customer.id,
+          customerId: customer.customer_id,
```
vincent-pochet, line 34: Not sure that it will be easy for front-end to know which customer_id to use :thinking: 

My point is we should limit the usage of this field to public API. Open to discuss that point though :)

app/services/wallet_transactions/validate_service.rb
lovrocolic: I was thinking about that but organisation is probably the only record that is always present if user is logged in. So not really sure if these cases also needs to be covered? WDYT?

app/services/wallet_transactions/create_service.rb
lovrocolic: fixed

app/services/wallets/balance/increase_service.rb
lovrocolic: Fixed (and removed BigDecimal casting once my schema is fixed)

app/services/wallet_transactions/create_service.rb
lovrocolic: Thanks for the catch! Fixed

db/schema.rb
lovrocolic: Fixed

spec/jobs/wallet_transactions/create_job_spec.rb
lovrocolic: Added

spec/services/wallet_transactions/create_service_spec.rb
lovrocolic: Sorry for that, specs are added in the next PR (since this create service is not finished here - paid credits handling is missing here)

spec/graphql/mutations/wallets/create_spec.rb
lovrocolic: We already synced with Alex and it's used on FE for this feature. For me it's also a bit easier to use `customer_id` since I can reuse the same service from api controller in that case :)

app/services/wallet_transactions/validate_service.rb
rsempe: Makes sense 👍

db/schema.rb
```diff
@@ -19,9 +19,9 @@
   create_table "active_storage_attachments", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.string "name", null: false
     t.string "record_type", null: false
-    t.uuid "record_id"
     t.uuid "blob_id", null: false
     t.datetime "created_at", null: false
+    t.uuid "record_id"
```
rsempe, line 24: not sure this change is needed

lovrocolic: Oh this attribute just got reordered somehow after running migration, but nothing really changed

rsempe: can you remove that change from this PR?

lovrocolic: I mean I can manually revert it but not sure about benefits of it? Maybe I misunderstood you 🤔  ...

spec/graphql/mutations/wallets/create_spec.rb
jdenquin: we already use `customer_id` on mutations linked to the customer (create subscriptions etc), maybe it's better to keep it this way for consistency

jdenquin (approved):

rsempe (approved):

spec/graphql/mutations/wallets/create_spec.rb
vincent-pochet: Sorry to bring back this one, but no on the front, we are using id:
- https://github.com/getlago/lago-api/blob/main/app/services/customers/update_service.rb#L6
- https://github.com/getlago/lago-api/blob/main/app/services/subscriptions/create_service.rb#L32

That's why I was bringing the consistency point

PR #402, opened 2022-08-23
feat: apply prepaid credits

lovrocolic:
## Context

Apply prepaid credits on subscription invoices

## Description

- Created `applied_prepaid_credit` collection that creates relation between invoice and wallet transaction
- Apply prepaid credit on invoice
- Update slim template


app/services/invoices/create_service.rb
```diff
@@ -123,12 +124,26 @@ def applied_coupon
...
     def should_create_credit?
       return false if applied_coupon.nil?
 
       applied_coupon.amount_currency == currency
     end
 
+    def should_create_applied_prepaid_credit?
+      return false unless wallet
+      return false unless wallet.active?
+      return false unless BigDecimal(wallet.balance) > 0
```
vincent-pochet, line 142: Rails should handle the data type automatically :)

app/services/invoices/create_service.rb
```diff
@@ -123,12 +124,26 @@ def applied_coupon
...
+
+      true
+    end
```
vincent-pochet, line 145: You could make it a bit shorter:

```ruby
       return false unless wallet&.active?
       
       wallet.balance > 0
    end
```

lovrocolic: refactored

app/services/credits/applied_prepaid_credit_service.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      super(nil)
+    end
+
+    def create
+      return result if already_applied?
+
+      amount_cents = compute_amount
+      amount = compute_amount_from_cents(amount_cents)
+      credit_amount = amount.fdiv(BigDecimal(wallet.rate_amount))
```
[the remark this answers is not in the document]
lovrocolic: fixed

app/services/wallets/balance/decrease_service.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        wallet.update!(
+          balance: wallet.balance - amount,
+          credits_balance: wallet.credits_balance - credits_amount,
+          last_balance_sync_at: Time.zone.now,
+          consumed_credits: wallet.consumed_credits + credits_amount,
+          consumed_amount: wallet.consumed_amount + amount,
+          last_consumed_credit_at: Time.current
+        )
+      end
```
vincent-pochet, line 24: Maybe we should return a result here. What do you think?

```ruby
  result.wallet = wallet
  result
  result.fail_with_validations!(e.record)
```

app/services/credits/applied_prepaid_credit_service.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def compute_amount_from_cents(amount)
+      currency = invoice.amount.currency
+
+      amount.round.fdiv(currency.subunit_to_unit)
+    end
+
+    def balance_cents
+      balance = BigDecimal(wallet.balance)
```
vincent-pochet, line 63: Do we still need this one?

lovrocolic: leftover:) Removed

app/services/wallets/balance/decrease_service.rb
lovrocolic: wallet is actually not used in a caller for now, but added it if that's the standard :)

PR #407, opened 2022-08-24
feat: add Organization Name in Segment Identify

jdenquin:
## Context

We miss the organization name in the segment identify informations

## Description

- Add the organization name in the segment identify traits

### DIFF
--- a/app/jobs/segment_identify_job.rb
+++ b/app/jobs/segment_identify_job.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class SegmentIdentifyJob < ApplicationJob
   queue_as :default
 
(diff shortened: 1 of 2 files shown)


app/jobs/segment_identify_job.rb
```diff
@@ -6,8 +8,13 @@ def perform(membership_id:)
     return if membership_id.nil? || membership_id == 'membership/unidentifiable'
 
     membership = Membership.find(membership_id.sub(/\Amembership\//, ''))
-    traits = { created_at: membership.created_at, hosting_type: hosting_type, version: version }
-    traits.merge!(email: membership.user.email) if hosting_type == 'cloud'
+    traits = {
+      created_at: membership.created_at,
+      hosting_type: hosting_type,
+      version: version,
+      organization_name: membership.organization.name,
```
rsempe, line 15: Are you sure we can track the organization's name on the self-hosted version as well? 🤔 

For instance, the email of the membership is only tracked on the cloud version.

vincent-pochet (approved):
It's OK on the code point of view, a bit trickier from a privacy perspective

PR #409, opened 2022-08-25
fix: use internal customer id for graphql and external in api for wallets

lovrocolic:


app/graphql/mutations/wallets/create.rb
```diff
@@ -23,10 +23,19 @@ def resolve(**args)
...
+            **args
+              .merge(organization_id: current_organization.id)
+              .merge(customer: customer(args[:customer_id]))
+              .except(:customer_id)
+          )
 
         result.success? ? result.wallet : result_error(result)
       end
+
+      def customer(id)
```
jdenquin, line 36: if we can rename it
```ruby
current_customer
```

app/graphql/mutations/wallets/create.rb
```diff
@@ -23,10 +23,19 @@ def resolve(**args)
 
         result = ::Wallets::CreateService
           .new(context[:current_user])
-          .create(**args.merge(organization_id: current_organization.id))
+          .create(
+            **args
+              .merge(organization_id: current_organization.id)
+              .merge(customer: customer(args[:customer_id]))
+              .except(:customer_id)
```
jdenquin, line 30: why we do an except here?

jdenquin (changes requested):
some comments for me!

app/graphql/mutations/wallets/create.rb
lovrocolic: since we only want to send customer object to the service and not the customer_id anymore (by default args contains customer_id)

app/graphql/mutations/wallets/create.rb
lovrocolic: here you are :)

rsempe:
We could have sent the `lago_customer_id` from the API (`app/controllers/api/v1/wallets_controller.rb`) without doing any changes on GraphQL - But works for me as it is!

rsempe (approved):

lovrocolic:
> We could have sent the `lago_customer_id` from the API (`app/controllers/api/v1/wallets_controller.rb`) without doing any changes on GraphQL - But works for me as it is!

Yep, I was thinking about it but in that case we would do the customer query twice :/

jdenquin (approved):

PR #417, opened 2022-08-29
feat: update wallet when status is changed to succeeded from API

lovrocolic:
## Context

Update wallet when invoice status changed to succeeded from the API


app/jobs/prepaid_credit_job.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class PrepaidCreditJob < ApplicationJob
```
jdenquin, line 3: can you namespace it? `Wallets::PrepaidCreditJob` for example?

jdenquin (changes requested):

app/jobs/prepaid_credit_job.rb
lovrocolic: as discussed offline, namespaced it with `Invoices`

app/jobs/invoices/prepaid_credit_job.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Invoices
+  class PrepaidCreditJob < ApplicationJob
```
vincent-pochet, line 4: Looks good!

My only comment would be to suggest to move this business logic into a dedicated service with a name clearer than the job name. Right now, it's not easy to guess what the job is doing by reading its name.

vincent-pochet (approved):

app/jobs/invoices/prepaid_credit_job.rb
lovrocolic: seems reasonable. I also wasn't 100% happy with the naming but in the same job we handle `WalletTransaction` and `Wallet` so I wasn't sure how to name it...
Moved the logic into new service and I have named it `Wallets::ApplyPaidCreditsService`

PR #419, opened 2022-08-30
misc: Use external_id instead of customer_id and subscription_id

rsempe:
## Context

We want to clarify the difference between lago ids and external ids.

Today, we have `customer_id` on customers and `subscription_id|unique_id` on subscriptions.

On the API, the goal is to use external_ids:
- `external_id` or `external_customer_id` for customer
- `external_id` or `external_subscription_id` for subscription

## Description

The goal of this PR is to:
- rename `customer_id` to `external_id` on customers
- rename `subscription_id` to `external_id` on subscriptions
- remove `unique_id` from subscriptions


app/services/subscriptions/create_service.rb
```diff
@@ -40,10 +40,9 @@ def create(**args)
 
       # NOTE: prepare subscription attributes
       @name = args[:name]&.strip
-      @unique_id = SecureRandom.uuid
+      @external_id = SecureRandom.uuid
```
vincent-pochet, line 43: Wondering if this still apply. Should we put the customer external id (Or waiting from the front to provide this value)?

app/services/invoices/customer_usage_service.rb
```diff
@@ -2,20 +2,19 @@
 
 module Invoices
   class CustomerUsageService < BaseService
+    # NOTE: customer_id can reference the lago id or the external id of the customer
+    # NOTE: subscription_id can reference the lago id or the external id of the subscription
```
vincent-pochet, line 6: Merf :/
Something that we might improve in a near future :smile:

app/services/payment_provider_customers/create_service.rb
```diff
@@ -10,7 +10,7 @@ def initialize(customer)
 
     def create_or_update(customer_class:, payment_provider_id:, params:, async: true)
       provider_customer = customer_class.find_or_initialize_by(
-        customer_id: customer.id,
+        customer_id: customer.id, # TODO: Why?
```
vincent-pochet, line 13: :thinking:

vincent-pochet (approved):
:exploding_head:

app/services/payment_provider_customers/create_service.rb
vincent-pochet: `customer_class` will be a `PaymentProviderCustomers::StripeCustomer` so no changes required here. (Appart from a refact maybe... :D )

db/schema.rb
```diff
@@ -327,7 +327,7 @@
     t.uuid "previous_subscription_id"
     t.date "subscription_date"
     t.string "name"
-    t.string "unique_id", null: false
+    t.string "external_id", null: false
```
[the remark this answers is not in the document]
rsempe: You're right - Added in that PR even if not related 👍

app/services/subscriptions/create_service.rb
rsempe: Well - Not sure we want the front to manage external ids and not sure we want to store the external customer id.

For me, I think it's good enough to generate this info from the backend, no?

app/services/invoices/customer_usage_service.rb
rsempe: Yes - That's why I put a note here 😅

app/services/payment_provider_customers/create_service.rb
rsempe: TODO removed 😅

PR #424, opened 2022-08-31
feat: add migration to update existing fee data

lovrocolic:


lib/tasks/fees.rake
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+namespace :fees do
+  desc 'Fill missing fee_type'
+  task fill_fee_type: :environment do
+    Fee.where(fee_type: nil).find_each do |fee|
+      fee_type = if fee.charge_id.blank? && fee.applied_add_on_id.blank?
```
jdenquin, line 7:
```ruby
return fee.add_on! if fee.applied_add_on.present?
return fee.charge! if fee.charge_id.present?

fee.subscription!
```

You can simplify it like this!

jdenquin (changes requested):

lovrocolic:
@rsempe This PR is created to handle fee_type for existing fee records. Also we had to remove null: false in existing migration since exception is raised if there are existing fee records without type (PG::NotNullViolation: ERROR:  column "fee_type" of relation "fees" contains null values).
Polymorphic association will be migrated later since we don't use it anyways for fees that are not type `credit`

rsempe (approved):

jdenquin (approved):

PR #426, opened 2022-09-01
fix(subscriptions): Apply charge fees on subscription upgrade

vincent-pochet:
## Context

With we introduce the anniversary date subscription billing, we made a change in the way we bill charges. 

## Description

Before the feature, the charges were not billed on the terminated subscription, but only at the end of the upgraded subscription period.
We decided to get rid of this behavior because it was a mess for the billing period computation and was kind of a weird behavior, that's why 

Unfortunately, we forgot to remove one condition in the `Invoices::CreateService` to check if charge fees should be generated.
This PR is here to fix this mistake

### DIFF
--- a/app/services/invoices/create_service.rb
+++ b/app/services/invoices/create_service.rb
@@ -103,10 +103,6 @@ module Invoices
     end
 
     def should_create_charge_fees?(invoice, subscription)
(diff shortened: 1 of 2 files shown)


jdenquin (approved):
:rocket:

lovrocolic (approved):

rsempe (approved):

PR #427, opened 2022-09-01
misc: Add external_customer_id on wallet serializer

rsempe:
## Context

We want to clarify the difference between lago ids and external ids.

## Description

The goal of this PR is to:
- return `external_customer_id` on wallet serializer


vincent-pochet (approved):

lovrocolic (approved):
💯

PR #421, opened 2022-08-30
feat(persisted-metrics): Aggregation logic implementation

vincent-pochet:
## Context

The objective of this feature is to introduce a system of persistent billable metrics that do not resume to 0 at the end of a billing period.

## Description

This PR is adding the aggregation logic. It uses the previously introduced `PersistedMetric`.

The logic takes into account:
- persisted metrics added before the period and not terminated or terminated after the period
- persisted metrics added during the period
- persisted metrics removed during the period
- persisted metrics added and removed during the period
- subscription terminated (or upgraded) before the end of the period
- subscription started before the beginning

## Related Task

- PersistedMetric model: https://github.com/getlago/lago-api/pull/414
- New aggregation type: https://github.com/getlago/lago-api/pull/415

PR #429, opened 2022-09-01
feat(persisted-metrics): Create persisted event from event

vincent-pochet:
## Context

## Description

This PR updates the event creation logic to manage as well the persistent metrics:
- Create a persistent metric when receiving an `add` operation
- Remove an existing persistent metric when receiving a `remove` operation
- Validate the persisted metrics requirements before processing the event

**NOTE: We should wait for the merge of https://github.com/getlago/lago-api/pull/419 since `Subscription#unique_id` will be renamed into `Subscription#external_id`**

## Related Task

Aggregation logic: https://github.com/getlago/lago-api/pull/421

### DIFF
--- a/app/services/events/create_batch_service.rb
+++ b/app/services/events/create_batch_service.rb
@@ -13,7 +13,7 @@ module Events
 
     def call(organization:, params:, timestamp:, metadata:)
       customer = organization.subscriptions.find_by(
-        external_id: params[:external_subscription_ids]&.first
+        external_id: params[:external_subscription_ids]&.first,
       )&.customer
 
       Events::ValidateCreationService.call(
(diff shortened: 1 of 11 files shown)


app/services/events/create_batch_service.rb
```diff
@@ -65,5 +66,13 @@ def call(organization:, params:, timestamp:, metadata:)
       result.events = events
       result
     end
+
+    def handle_persisted_metric(event)
+      service = PersistedMetrics::CreateOrUpdateService.new(event)
```
jdenquin, line 71: can we use `persisted_service` or smth like that just for readability?

jdenquin (approved):

rsempe (approved):

PR #431, opened 2022-09-02
feat(persisted-metrics): Add persisted metric breakdown on the invoice

vincent-pochet:
## Context

## Description

The objective of this pull request is to update the PDF invoice in order to render the breakdown of the persisted metrics:
- Items billed on the full period
- Items added during the period
- Items remove during the period
- Items added and then remove during the same period

## Related Task

- Event ingestion logic: https://github.com/getlago/lago-api/pull/429
- Aggregation logic for the recurring persisted metrics: https://github.com/getlago/lago-api/pull/421

### DIFF
--- a/app/services/billable_metrics/aggregations/recurring_count_service.rb
+++ b/app/services/billable_metrics/aggregations/recurring_count_service.rb
@@ -178,7 +178,7 @@ module BillableMetrics
         added_and_removed_list.map do |aggregation|
           OpenStruct.new(
             date: aggregation.first.to_date,
-            action: 'add',
+            action: 'add_and_removed',
             count: aggregation.last,
             duration: (aggregation.second.to_date + 1.day - aggregation.first.to_date).to_i,
             total_duration: period_duration,
--- a/app/services/billable_metrics/aggregations/recurring_count_service.rb
+++ b/app/services/billable_metrics/aggregations/recurring_count_service.rb
@@ -11,6 +11,19 @@ module BillableMetrics
         result
       end
 
+      def breakdown(from_date:, to_date:)
+        @from_date = from_date.to_date
+        @to_date = to_date.to_date
+
+        breakdown = persisted_breakdown
+        breakdown += added_breakdown
+        breakdown += removed_breadown
+        breakdown += added_and_removed_breakdown
+
+        result.breakdown = breakdown.sort_by(&:date)
+        result
+      end
+
       private
 
       attr_reader :from_date, :to_date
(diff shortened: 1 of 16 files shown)


rsempe (approved):

app/services/billable_metrics/aggregations/recurring_count_service.rb
```diff
@@ -11,6 +11,19 @@ def aggregate(from_date:, to_date:, options: {})
...
+      def breakdown(from_date:, to_date:)
+        @from_date = from_date.to_date
+        @to_date = to_date.to_date
+
+        breakdown = persisted_breakdown
+        breakdown += added_breakdown
+        breakdown += removed_breadown
+        breakdown += added_and_removed_breakdown
+
+        result.breakdown = breakdown.sort_by(&:date)
```
lovrocolic, line 23: Probably a bit overkill, but we can get better complexity here (instead of using built-in sort) with custom sorting algorithm since each array is already sorted

spec/models/invoice_spec.rb
```diff
@@ -187,4 +187,32 @@
...
+  end
+
+  describe '#recurring_breakdown' do
+    let(:invoice_subscription) { create(:invoice_subscription) }
+
+    it 'returns the fees of the corresponding invoice_subscription' do
+      invoice = invoice_subscription.invoice
+      subscription = invoice_subscription.subscription
+      billable_metric = create(:recurring_billable_metric, organization: subscription.organization)
+      charge = create(:standard_charge, plan: subscription.plan, billable_metric: billable_metric)
```
lovrocolic, line 212: Just a style comment :D Not sure why in this file in general in several more places we don't create variables with `let` but instead we are using local variables?

lovrocolic (approved):

spec/models/invoice_spec.rb
vincent-pochet: Well, not sure exactly :D
I guess it's because it used only in one single test. But that's true that we  could refact it a bit :+1:

app/services/billable_metrics/aggregations/recurring_count_service.rb
vincent-pochet: Nice one :thinking: 
I think we can keep it as it for now since we will probably have really few persisted events per customer. We can keep it in mind for future performance tweak if required. What do you think?

lovrocolic: Yep make sense. We can improve sorting part from `O(n*logn) -> O(n)` if `n` is the length of the final array. But yeah, it wouldn't be significant if there is small number of elements

PR #432, opened 2022-09-05
feat(persisted-metrics): Rename PersistedMetric into PersistedEvent

vincent-pochet:
## Context

## Description

This PR rename the previously added `PersistedMetrics` into `PersistedEvents` to make it more obvious it's related to events and not a new kind of metric definition


rsempe (approved):
🤗

PR #434, opened 2022-09-05
feat(customers): Add index and show on API

jdenquin:
## Context

We do not have any endpoints to list organization's customer or search for a customer on the API.

## Description

- Add the `/customers` index endpoint
- Add the `/customers/:customer_id` show endpoint

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,7 +16,7 @@ Rails.application.routes.draw do
 
   namespace :api do
     namespace :v1 do
-      resources :customers, param: :external_id, only: %i[create] do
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -16,9 +16,10 @@
 
   namespace :api do
     namespace :v1 do
-      resources :customers, only: %i[create] do
+      resources :customers, only: %i[create index] do
         get :current_usage
       end
+      resources :customers, only: :show, param: :customer_id
```
rsempe, line 22:
```suggestion
      resources :customers, only: :show, param: :external_id
```

rsempe (approved):

PR #436, opened 2022-09-05
Memberships: list and revoke

ansmonjol:
This PR is included in the scope of the "invite memberships" feature.
This is the first part out of three:
1. [ ] Memberships: list and revoke 👈 we are here
2. [ ] Invites: list, create and revoke
3. [ ] Accept invite

---

This PR adds
- Query memberships of an organization
- Revoke a member of an organization


db/migrate/20220905095454_add_membership_status_and_revoked_at.rb
```diff
@@ -0,0 +1,8 @@
+class AddMembershipStatusAndRevokedAt < ActiveRecord::Migration[7.0]
+  def change
+    change_table :memberships do |t|
+      t.integer :status, null: false, default: 0
+      t.date :revoked_at, null: true
```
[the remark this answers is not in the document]
ansmonjol: Value in here will be a `Time.zone.now`, so yes indeed ! 🙏 
I updated the code and re-runned the migration to update the schema

app/models/membership.rb
```diff
@@ -4,5 +4,16 @@ class Membership < ApplicationRecord
...
 
+  STATUSES = [
+    :active,
+    :revoked,
+  ].freeze
+
+  enum status: STATUSES
   enum role: [:admin]
+
+  def mark_as_revoked!(timestamp = Time.zone.now)
```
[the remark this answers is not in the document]
ansmonjol: Is that to prevent cases where timezone is not defined ? 

I can see a lot of `Time.zone.now` usage in the app, should we create a task to update them all ?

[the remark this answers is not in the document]
ansmonjol: Asking because on the output I can see the value is the same

app/services/memberships/revoke_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Memberships
+  class RevokeService < BaseService
+    def revoke(id)
```
rsempe, line 5: Can you create unit tests for this?

app/models/membership.rb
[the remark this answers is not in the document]
rsempe: Can you create unit tests for this? 😅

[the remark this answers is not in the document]
rsempe: `Time.current` is preferred over `Time.zone.now` 😅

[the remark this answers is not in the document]
ansmonjol: Code is updated to use `Time.current`. Also added a test for the membership model!

app/services/memberships/revoke_service.rb
ansmonjol: Done! Won't forget next time ;)

spec/models/membership_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Membership, type: :model do
+  let(:membership) { create(:membership) }
+
+  describe '.mark_as_revoked' do
```
rsempe, line 8:
```suggestion
  describe '#mark_as_revoked' do
```

As this is an instance method ;)

ansmonjol: Fixed

spec/models/membership_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  describe '#mark_as_revoked' do
+    it 'revokes the membership with a Time' do
+      expect(membership.status).to eq('active')
+      expect(membership.revoked_at).to eq(nil)
+
+      membership.mark_as_revoked!
+
+      expect(membership.status).to eq('revoked')
+      expect(membership.revoked_at).to be_a(Time)
```
rsempe, line 16: What about writing it like this?

```suggestion
      freeze_time do
        expect { membership.mark_as_revoked! }
          .to change { membership.reload.status }.from(:active).to(:revoked)
          .and change(membership, :revoked_at).from(nil).to(Time.current)
      end
```

app/services/memberships/revoke_service.rb
rsempe, line 5:
```suggestion
    def call(id)
```

Then we can have `Memberships::RevokeService.call()` instead of `Memberships::RevokeService.revoke()`

app/services/memberships/revoke_service.rb
```diff
@@ -0,0 +1,24 @@
...
+      membership = Membership.find_by(id: id)
+      return result.fail!(code: 'not_found') unless membership
+
+      if result.user.id == membership.user.id
+        return result.fail!(
+          code: 'unprocessable_entity',
+          message: 'revoke_own_membership',
```
rsempe, line 12:
```suggestion
          message: 'trying to revoke my own membership',
```

I think message is more a sentence than a code in the codebase

spec/services/memberships/revoke_service_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Memberships::RevokeService, type: :service do
+  subject(:revoke_service) { described_class.new(membership.user) }
+
+  let(:membership) { create(:membership) }
+
+  describe '#revoke' do
```
rsempe, line 10:
```suggestion
  describe '#call' do
```

app/services/memberships/revoke_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Memberships
+  class RevokeService < BaseService
+    def revoke(id)
+      membership = Membership.find_by(id: id)
+      return result.fail!(code: 'not_found') unless membership
```
rsempe, line 7: What about returning an explicit error code?

```suggestion
      return result.fail!(code: 'membership_not_found') unless membership
```

spec/services/memberships/revoke_service_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        expect(result).not_to be_success
+        expect(result.error).to eq('not_found')
+      end
+    end
+
+    context 'when revoking another membership' do
+      let(:another_membership) { create(:membership) }
+
+      it 'revokes the memembership' do
```
rsempe, line 32:
```suggestion
      it 'revokes the membership' do
```

spec/services/memberships/revoke_service_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        result = revoke_service.revoke(another_membership.id)
+
+        expect(result).to be_success
+        expect(result.membership.id).to eq(another_membership.id)
```
rsempe, line 36: Can we also add an assertion to ensure that the membership is correctly revoked?

app/services/memberships/revoke_service.rb
```diff
@@ -0,0 +1,24 @@
...
+          code: 'unprocessable_entity',
+          message: 'revoke_own_membership',
+        )
+      end
+
+      membership.mark_as_revoked!
+
+      result.membership = membership
+      result
+    rescue ActiveRecord::RecordInvalid => e
```
rsempe, line 20: How is it possible to have an `ActiveRecord::RecordInvalid` error?

spec/models/membership_spec.rb
ansmonjol: Od didn't knew this notation, gonna use it yeah !

My only concern is with `to(Time.current)`. I tested it before but there is a small diff between the value in test and the value stored in the `revoked_at` attribute, hence my check on `.to be_a(Time)`

Do you know an other way to do or is the `to be_a(Time)` check ok ?

rsempe: That's why I used `freeze_time` (https://apidock.com/rails/v6.1.3.1/ActiveSupport/Testing/TimeHelpers/freeze_time) ;)

app/services/memberships/revoke_service.rb
ansmonjol: Is that a convention? I find it more explicit to name the method but happy to change and follow any good practice

app/services/memberships/revoke_service.rb
ansmonjol: Hum indeed, I added a message first but I also wanted to send a detail with the code `revoke_own_membership`

```
return result.fail!(
          code: 'unprocessable_entity',
          message: 'Cannot revoke own membership',
          detail: 'revoke_own_membership',
)
```

The issue I had is that the `detail` was never received by the client, like if some kind of serializer didn't allowed such attribute. So I changed the `message` value to have the value of `detail` 😅

spec/models/membership_spec.rb
ansmonjol: ooohhh ok nice 🙏

app/services/memberships/revoke_service.rb
rsempe: It's not a convention, but I like having a `call` method in all services if the naming is like this.

`Memberships::RevokeService.revoke` is quite redundant, I prefer having `MembershipsService.revoke` or `Memberships::RevokeService.call`

spec/models/membership_spec.rb
ansmonjol: Updated

app/services/memberships/revoke_service.rb
ansmonjol: Updated here and all invocations

app/services/memberships/revoke_service.rb
ansmonjol: Updated

spec/services/memberships/revoke_service_spec.rb
ansmonjol: Ouch, good catch thanks

spec/services/memberships/revoke_service_spec.rb
ansmonjol: Done

app/services/memberships/revoke_service.rb
ansmonjol: Yep makes no sense, removed

rsempe (approved):

PR #437, opened 2022-09-07
feat: add overridden_plan_id to plans and use it as default scope

lovrocolic:
## Context

Currently we are adding feature to allow overriding prices and charges while creating new subscription

## Description

This change adds required DB changes in order to support 'override prices and charges' feature.
Also, we created new scope that will not take into account any 'extended' plans in queries across the app


app/models/plan.rb
```diff
@@ -24,6 +26,8 @@ class Plan < ApplicationRecord
   validates :code, presence: true, uniqueness: { scope: :organization_id }
   validates :amount_currency, inclusion: { in: currency_list }
 
+  scope :default, -> { where(overridden_plan_id: nil) }
```
rsempe, line 29: `default_scope` instead?

lovrocolic: I can switch to default scope but I personally don't like it since it mask behaviour :/

rsempe: I tend to agree 😅 

But I think we can use a different name than `default` then, something more explicit?

lovrocolic: `root`? 😃

vincent-pochet: `parent` makes more sense I think.

`default` could means many many things :smile:

vincent-pochet (commented):
Good job! :rocket: 

Maybe it also worth to add at list one test case somewhere

app/models/plan.rb
jdenquin: what about `base` ?

jdenquin (changes requested):

app/models/plan.rb
lovrocolic: switched to `base`

lovrocolic:
> Good job! 🚀
> 
> Maybe it also worth to add at list one test case somewhere

Added one test in controller:)

vincent-pochet (approved):

app/models/plan.rb
```diff
@@ -24,6 +26,8 @@ class Plan < ApplicationRecord
   validates :code, presence: true, uniqueness: { scope: :organization_id }
   validates :amount_currency, inclusion: { in: currency_list }
 
+  scope :base, -> { where(overridden_plan_id: nil) }
```
rsempe, line 29: `not_overriden`? 😅

app/models/plan.rb
rsempe: `not_overriden`, explicit 😅

PR #439, opened 2022-09-08
feat: add logic to create subscription from overridden plan

lovrocolic:
## Context

Adding core logic for 'override prices and charges' feature

## Description

This change adds logic for override plan feature. In this PR there are graphQL and service changes for mentioned feature.
API changes will be added later.


app/services/subscriptions/override_service.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  class OverrideService < BaseService
+    include ChargeModelAttributesHandler
```
vincent-pochet, line 5: I think we should not use GraphQL helper at this level. Input arguments should be prepared before calling the service.

For example, the signature of the method could be:
```ruby
def call(paln_args:, subscription_args:)
```

app/services/subscriptions/override_service.rb
```diff
@@ -0,0 +1,83 @@
...
+          details: plan_result.error_details,
+        )
+      end
+    end
+
+    def failed_subscription_creation_result(subscription_result)
+      if subscription_result.error_code == 'missing_argument'
+        result.fail!(
+          code: 'not_found',
+          message: 'Some resources have not been found while creating subscription',
```
vincent-pochet, line 65: Front end does not have real need for "human readable messages" as it uses error code for translation.

I'm currently working on an improvement of the error management, but in the meantime, I suggest to just return the sub-service error.

vincent-pochet (commented):
Small comments on argument and error handling, but otherwise the logic looks good :+1:

app/services/subscriptions/override_service.rb
```diff
@@ -0,0 +1,83 @@
...
+
+        result.subscription = subscription_result.subscription
+      end
+
+      result
+    end
+
+    private
+
+    def handle_plan_args(args)
```
jdenquin, line 28: what about `prepare_plan_args` ?

app/services/subscriptions/override_service.rb
```diff
@@ -0,0 +1,83 @@
...
+      params = args[:plan]
+      params[:code] = "#{params[:code]}-#{SecureRandom.uuid}"
+
+      prepare_arguments(**params)
+        .merge(overridden_plan_id: args[:overridden_plan_id])
+        .merge(organization_id: args[:organization_id])
+    end
+
+    def handle_subscription_args(args, plan_id)
```
jdenquin, line 37: and `prepare_subscription_args` ?

app/services/subscriptions/override_service.rb
jdenquin: agree, why we do not just return sub service result?!

spec/graphql/mutations/subscriptions/create_with_override_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+require 'rails_helper'
+
+RSpec.describe Mutations::Subscriptions::CreateWithOverride, type: :graphql do
+  let(:membership) { create(:membership) }
+  let(:organization) { membership.organization }
+  let(:plan) { create(:plan, organization: organization) }
+  let(:customer) { create(:customer, organization: organization) }
+  let(:billable_metric) { create(:billable_metric, organization: organization) }
+  let(:mutation) do
```
jdenquin, line 11: let blocks should be at the end of the let list, and if we can have some new line between blocks! 🙏

spec/graphql/mutations/subscriptions/create_with_override_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+            interval: plan.interval,
+            payInAdvance: plan.pay_in_advance,
+            amountCents: 259,
+            amountCurrency: plan.amount_currency,
+            charges: [
+              {
+                billableMetricId: billable_metric.id,
+                amount: '100.00',
+                chargeModel: 'standard',
+              }
```
jdenquin, line 53: missing `,`

jdenquin (changes requested):

app/services/plans/create_service.rb
```diff
@@ -8,6 +8,7 @@ def create(**args)
         name: args[:name],
         code: args[:code],
         description: args[:description],
+        overridden_plan_id: args[:overridden_plan_id],
```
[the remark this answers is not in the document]
lovrocolic: added new validation service for that

app/services/subscriptions/override_service.rb
lovrocolic: refactored

app/services/subscriptions/override_service.rb
lovrocolic: We mostly use `not_found` if any record could not be found (even though it is e.g. some other record's argument) so that's why I wrapped errors here (to avoid `missing_argument`) error. Also, wanted to add clarification in which part (plan or subscription) error happened. I was aware that FE is using just `key` but I was thinking that API could also use it (later realised I will need new service for API) :) 
It's changed now and I'm just passing errors if it will be refactored in Vincent's change anyways :)

PR #442, opened 2022-09-09
feat(errors): Handle not found error in services API and GraphQL errors

vincent-pochet:
## Context

## Description

The objective of this pull request is to normalize the way service return not found errors.

When a required resource is not found, the services now returns a failed result as it was doing before. Instead of a code and a message, attached to the result, we now build a `BaseService::NotFoundFailure` with a `code` attribute and we attach it as the `error` field of the result object.

- At API controller level, when a service result has a `BaseService::NotFoundFailure`, it will format an HTTP 404 response with the error code in the payload
- At GraphQL level, it formats a `GraphQL::ExecutionError`, with the error code attached in the `extensions` payload

The current list of not found error codes is:
- `add_on_not_found`
- `applied_coupon_not_found`
- `billable_metric_not_found`
- `billable_metrics_not_found`
- `coupon_not_found`
- `customer_not_found`
- `event_not_found`
- `invoice_not_found`
- `payment_provider_not_found`
- `plan_not_found`
- `resource_not_found`
- `stripe_customer_not_found`
- `subscription_not_found`
- `wallet_not_found`
- `wallet_transaction_not_found`

### DIFF
--- a/app/graphql/resolvers/add_on_resolver.rb
+++ b/app/graphql/resolvers/add_on_resolver.rb
@@ -16,7 +16,7 @@ module Resolvers
 
       current_organization.add_ons.find(id)
     rescue ActiveRecord::RecordNotFound
-      not_found_error
+      not_found_error(code: 'add_on_not_found')
     end
   end
 end
--- a/app/graphql/resolvers/add_on_resolver.rb
+++ b/app/graphql/resolvers/add_on_resolver.rb
@@ -16,7 +16,7 @@ module Resolvers
 
       current_organization.add_ons.find(id)
     rescue ActiveRecord::RecordNotFound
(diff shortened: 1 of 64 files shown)


app/graphql/resolvers/add_on_resolver.rb
```diff
@@ -16,7 +16,7 @@ def resolve(id: nil)
 
       current_organization.add_ons.find(id)
     rescue ActiveRecord::RecordNotFound
-      not_found_error
+      not_found_error(code: 'add_on_not_found')
```
rsempe, line 19: What about something like `not_found_error(resource: 'add_on')` instead to not repeat `not_found` each time?

rsempe (approved):

PR #444, opened 2022-09-13
feat(errors): Handle not allowed error in services API and GraphQL errors

vincent-pochet:
## Context

## Description

The objective of this pull request is to normalize the way services return method not allowed errors.

When a service action cannot be performed because the status is invalid or the pre-conditions are not met, the service keeps returning a failed as it was doing before. But instead of a code and a message attached to the result, we now build a `BaseService::MethodNotAllowedFailure` with a code attribute and we attach it as the error field of the result object.

- At API controller level, when a service result has a `BaseService::MethodNotAllowedFailure`, it will format an HTTP 405 response with the error code in the payload
- At GraphQL level, it formats a `GraphQL::ExecutionError`, with the error code attached in the `extensions` payload

The current list error codes is:
- `attached_to_an_active_subscription`
- `attached_to_an_active_customer`

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -45,23 +45,23 @@ module Api
       )
     end
 
-    def forbidden_error(error_result)
+    def method_not_allowed_error(code:)
       render(
         json: {
-          status: 403,
-          error: 'Forbidden',
-          message: error_result.error,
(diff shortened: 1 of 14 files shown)


rsempe (approved):
👏

jdenquin (approved):

PR #438, opened 2022-09-07
Invites - Create, revoke and list

ansmonjol:
This PR is included in the scope of the "invite memberships" feature.
This is the second part out of three:
1. [x] [Memberships: list and revoke](https://github.com/getlago/lago-api/pull/436) 
2. [ ] Invites: list, create and revoke 👈 we are here
3. [ ] Accept invite

---

This PR adds
- the invite model and type
- ability to create invites
- ability to revoke an invite
- ability to list all organization pending invites
- ability to get the informations about one specific invite

Better review commit by commit

PR #447, opened 2022-09-13
Accept invite

ansmonjol:
This PR is included in the scope of the "invite memberships" feature.
This is the last part of the feature.
1. [x] [Memberships: list and revoke](https://github.com/getlago/lago-api/pull/436) 
2. [x] [Invites: list, create and revoke ](https://github.com/getlago/lago-api/pull/438)
3. [ ] Accept invite 👈 we are here

---

This PR adds
- a new create_user_and_membership method in user service
- a new mark_as_accepted method in invite model
- a new accept invite mutation and related service

Better review commit by commit


spec/services/invites/accept_service_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  let(:invite) { create(:invite, organization: organization) }
+  let(:accept_args) do
+    {
+      email: invite.email,
+      password: 'ILoveLago!',
+      token: invite.token,
+    }
+  end
+
+  describe 'accept' do
```
rsempe, line 19:
```suggestion
  describe '#call' do
```

spec/services/invites/accept_service_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    it 'accepts an invite' do
+      accept_service.call(**accept_args)
+
+      invite.reload
+
+      expect(invite).to be_accepted
+      expect(invite.recipient.id).to be_present
```
rsempe, line 26:
```suggestion
      expect { accept_service.call(**accept_args) }
        .to change { invite.reload.status }.from('pending').to('accepted')
        .and change(invite, :membership_id).from(nil).to(...)
```

spec/services/invites/accept_service_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    context 'when invite is already accepted' do
+      it 'returns invite_not_found error' do
+        invite.accepted!
+        invite.reload
+
+        result = accept_service.call(**accept_args)
```
rsempe, line 34: same here

app/services/invites/accept_service.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      ActiveRecord::Base.transaction do
+        result = UsersService.new.register_from_invite(
+          args[:email],
+          args[:password],
+          invite.organization.id,
+        )
+
+        invite.recipient = result.membership || Membership.find_by(user_id: result.user.id)
+        invite.save!
```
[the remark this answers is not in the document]
jdenquin: since you're doing a `mark_as_accepted!` after I think no!

spec/services/invites/accept_service_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        expect(result).not_to be_success
+        expect(result.error_code).to eq('invite_not_found')
+      end
+    end
+
+    context 'without email' do
+      it 'returns an error' do
+        result = accept_service.call(email: nil, password: accept_args[:password], token: accept_args[:token])
+
+        expect(result).not_to be_success
```
rsempe, line 45: can we check the status as well?

app/services/users_service.rb
```diff
@@ -41,6 +34,25 @@ def register(email, password, organization_name)
     result
   end
 
+  def register_from_invite(email, password, organization_id)
+    result.user = User.find_or_initialize_by(email: email)
```
jdenquin, line 38: this should be in a validate service? :p

jdenquin (changes requested):

app/services/invites/accept_service.rb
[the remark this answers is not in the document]
ansmonjol: Ok interesting ! Will remove

spec/services/users_service_spec.rb
```diff
@@ -28,6 +28,24 @@
         }
       )
     end
+
+    it 'creates an organization, user and membership' do
+      result = subject.register('email', 'password', 'organization_name')
```
jdenquin, line 33: RSpec/NamedSubject: Name your test subject if you need to reference it explicitly. (https://rspec.rubystyle.guide/#use-subject, https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/NamedSubject)

app/services/invites/accept_service.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Invites
+  class AcceptService < BaseService
+    def call(**args)
+      invite = Invite.find_by(token: args[:token], status: :pending)
+
+      return result.fail!(code: 'invite_not_found') unless invite
```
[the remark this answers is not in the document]
ansmonjol: Updated ! Thanks

app/services/invites/accept_service.rb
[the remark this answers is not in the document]
ansmonjol: Removed

spec/services/invites/accept_service_spec.rb
ansmonjol: Done !

app/services/invites/accept_service.rb
```diff
@@ -0,0 +1,24 @@
...
+      return result.not_found_failure!(resource: 'invite') unless invite
+
+      ActiveRecord::Base.transaction do
+        result = UsersService.new.register_from_invite(
+          args[:email],
+          args[:password],
+          invite.organization_id,
+        )
+
+        invite.recipient = result.membership || Membership.find_by(user_id: result.user.id)
```
rsempe, line 16: What happens if at the end recipient is `nil`?

app/services/users_service.rb
```diff
@@ -41,6 +34,24 @@ def register(email, password, organization_name)
...
+    if result.user.id
+      result.fail!(code: 'user_already_exists')
+
+      return result
+    end
+
+    ActiveRecord::Base.transaction do
+      result.organization = Organization.find(organization_id)
+      result.token = generate_token
+      create_user_and_membership(result, result.organization, password)
```
rsempe, line 49: you already have `organization` inside `result` :)

app/services/invites/accept_service.rb
ansmonjol: well, it cannot really happen.
During invites the membership is created, and TBH that should be the only condition.
While copying the `register` method and adapting it for invites, I notices there is a condition if the user already exists then it only return the user, hence the `Membership.find_by`.

Does that sounds wrong ?

app/services/users_service.rb
```diff
@@ -49,6 +60,21 @@ def new_token(user)
...
   private
 
+  def create_user_and_membership(result, organization, password)
+    result.user.password = password
+    result.user.save!
+
+    result.membership = Membership.create!(
+      user: result.user,
+      organization: organization,
+      role: :admin,
```
rsempe, line 70: is it not the default role?

app/services/users_service.rb
ansmonjol: What do you mean ? this endpoint is called while the user is not connected so I need to fetch the orga by id

ansmonjol: Oh you mean it's not necessary to pass it as argument ?

app/services/users_service.rb
ansmonjol: indeed, removed

app/services/users_service.rb
```diff
@@ -41,6 +34,24 @@ def register(email, password, organization_name)
...
+  def register_from_invite(email, password, organization_id)
+    result.user = User.find_or_initialize_by(email: email)
+
+    if result.user.id
+      result.fail!(code: 'user_already_exists')
+
+      return result
+    end
+
+    ActiveRecord::Base.transaction do
```
rsempe, line 46: What about moving the transaction inside `create_user_and_membership`method?

app/services/invites/accept_service.rb
ansmonjol: I removed all code on the right of `||` and added a test for this case. Indeed result.membership should always exist

app/services/users_service.rb
ansmonjol: removed

app/services/users_service.rb
ansmonjol: Done!

rsempe (approved):

PR #452, opened 2022-09-14
feat: Migration and listing plan override

lovrocolic:
## Context

Add migration and listing part for plan override feature

## Description

- adding migration (self relation on plans)
- fix queries where we fetch plans

This PR closes https://github.com/getlago/lago-api/pull/437


db/migrate/20220907084504_add_override_plan_to_plans.rb
```diff
@@ -0,0 +1,5 @@
+class AddOverridePlanToPlans < ActiveRecord::Migration[7.0]
```
jdenquin, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

rsempe (approved):

lovrocolic:
This feature specs got reworked so this will be handled in another PR

PR #464, opened 2022-09-15
feat: Return downgrade plan date instead of next and previous external id

rsempe:
When a downgrade is made, `external_id` is copied from the original to the new subscription.
Instead of returning `external_id`, we want to return the downgrade plan date of the subscription in the API.


vincent-pochet (approved):
:rocket:

PR #468, opened 2022-09-16
feat: support new field called parent_id on plan creation

lovrocolic:
## Context

There is a new feature that eases the process of plan creation with pre-filling information from selected plan.

## Description

While creating a new plan based on selected 'parent' plan we still want to keep this relationship in order to keep possibility to group such a plans in the UI


vincent-pochet (approved):
:boom:

PR #469, opened 2022-09-16
feat(customers): Assign customer currency from wallet creation

vincent-pochet:
## Context

Currently, we cannot apply a coupon, an add-on or even prepaid-credits if a customer has no active subscription. This problem happen because customer object does not hold the currency directly, but via the currency of the plan of it’s first active subscription.

It’s a problem when a plan is invoiced in-advance because it can then take up to 1 year to apply the mentioned objects on the first generated invoice.

## Description

The role of this PR is to assign the customer currency when creation a wallet for a customer with no currency

As a side effect it forces us to update the `createWallet` mutation and the `POST /api/v1/wallets` API route to accept a mandatory currency parameter.

It also removes the need for a subscription before creating a wallet

### DIFF
--- a/app/controllers/api/v1/wallets_controller.rb
+++ b/app/controllers/api/v1/wallets_controller.rb
@@ -81,6 +81,7 @@ module Api
         params.require(:wallet).permit(
           :rate_amount,
           :name,
+          :currency,
           :paid_credits,
           :granted_credits,
           :expiration_date,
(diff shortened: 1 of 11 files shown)


lovrocolic (approved):

rsempe (approved):

vincent-pochet:
Closed with: https://github.com/getlago/lago-api/pull/478

PR #470, opened 2022-09-19
fix(billable_metrics): Wrong aggregation type should raise a validation error

vincent-pochet:
## Context

Trying to assign an invalid aggregation type when creating a billable metric leads to an HTTP 500 error when it should returns a 422.

## Description

This behavior is related to the Rails `enum` implementation. When assigning an invalid value to an enum, a ruby `ArgumentError` is raised immediately, making it impossible to validate the provided value with the standard model validations.

In order to fix the issue, the idea is:
- to override the default enum setter, to add filter invalid inputs
- add a validation rule to validate inclusion of the value (or nil) in the list of aggregation types

### DIFF
--- a/spec/models/billable_metric_spec.rb
+++ b/spec/models/billable_metric_spec.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe BillableMetric, type: :model do
+  subject(:billable_metric) { described_class.new }
+
+  describe '#aggregation_type=' do
+    it 'assigns the aggregation type' do
+      billable_metric.aggregation_type = :count_agg
+      billable_metric.valid?
+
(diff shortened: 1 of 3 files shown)


spec/models/billable_metric_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe BillableMetric, type: :model do
+  subject(:billable_metric) { BillableMetric.new }
```
jdenquin, line 6: RSpec/DescribedClass: Use `described_class` instead of `BillableMetric`. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/DescribedClass)

lovrocolic (approved):

PR #473, opened 2022-09-19
feat(customers): Add migration task to populate customer currency

vincent-pochet:
## Context

Currently, we cannot apply a coupon, an add-on or even prepaid-credits if a customer has no active subscription. This problem happen because customer object does not hold the currency directly, but via the currency of the plan of it’s first active subscription.

It’s a problem when a plan is invoiced in-advance because it can then take up to 1 year to apply the mentioned objects on the first generated invoice.

## Description

The role of this PR is to run a migration task to assign customer currency from the active subscription

### DIFF
--- a/app/graphql/types/customers/single_object.rb
+++ b/app/graphql/types/customers/single_object.rb
@@ -22,10 +22,6 @@ module Types
       def applied_add_ons
         object.applied_add_ons.order(created_at: :desc)
       end
-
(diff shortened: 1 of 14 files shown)


lovrocolic (approved):

vincent-pochet:
Closed with: https://github.com/getlago/lago-api/pull/478

PR #474, opened 2022-09-19
fix: prevent revoked users to log in

ansmonjol:
User without an active status were still able to log in the app.

This PR brings a check to prevent them to do so.

I has been agree with product that the returned error should be the same as if the user does not exists

### DIFF
--- a/app/services/users_service.rb
+++ b/app/services/users_service.rb
@@ -3,9 +3,11 @@
 class UsersService < BaseService
   def login(email, password)
     result.user = User.find_by(email: email)&.authenticate(password)
-    result.token = generate_token if result.user
 
     return result.fail!(code: 'incorrect_login_or_password') unless result.user
+    return result.fail!(code: 'incorrect_login_or_password') unless result.user.memberships.first.active?
+
+    result.token = generate_token if result.user
 
     # Note: We're tracking the first membership linked to the user.
     SegmentIdentifyJob.perform_later(membership_id: "membership/#{result.user.memberships.first.id}")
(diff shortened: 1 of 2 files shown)


spec/graphql/mutations/login_user_spec.rb
```diff
@@ -58,4 +55,24 @@
...
+    let(:revoked_membership) { create(:membership, status: :revoked) }
+
+    it 'returns an error' do
+      result = execute_graphql(
+        query: mutation,
+        variables: {
+          input: {
+            email: revoked_membership.user.email,
+            password: 'ILoveLago',
+          }
```
jdenquin, line 69: Style/TrailingCommaInHashLiteral: Put a comma after the last item of a multiline hash.

app/services/users_service.rb
```diff
@@ -3,9 +3,11 @@
 class UsersService < BaseService
   def login(email, password)
     result.user = User.find_by(email: email)&.authenticate(password)
-    result.token = generate_token if result.user
 
     return result.fail!(code: 'incorrect_login_or_password') unless result.user
+    return result.fail!(code: 'incorrect_login_or_password') unless result.user.memberships.first.active?
```
vincent-pochet, line 8: Since order of the membership list is unpredictable, (ordered on id field, but ids are UUIDs) with this logic, if the memberships is revoked, user cannot log-in anymore even if he has other active memberships, I'm not sure it is the intended behaviors.

What about something like:
```suggestion
    return result.fail!(code: 'incorrect_login_or_password') unless result.user.memberships.active.any?
```

ansmonjol: Well, to be correct we should check for the membership of the current organisation 🤔 but I get your point

I know we don't handle multi-orga yet but indeed the any seems more correct.

vincent-pochet: I guess it also handles the re-invite of revoked member

ansmonjol: Went with your solution. Code is updated

PR #477, opened 2022-09-20
feat(errors): More handling of single validation errors

vincent-pochet:
## Context

## Description

The objective of this PR is to attach validation failures to a field when instead of rendering a single validation code.

It has been validated that this change does not require updating the frontend, as it only triggers a generic error.

### DIFF
--- a/app/services/invoices/update_service.rb
+++ b/app/services/invoices/update_service.rb
@@ -5,7 +5,13 @@ class Invoices::UpdateService < BaseService
     invoice = Invoice.find_by(id: invoice_id)
 
     return result.not_found_failure!(resource: 'invoice') if invoice.blank?
-    return result.single_validation_failure!(field: :status, error_code: 'value_is_invalid') unless valid_status?(params[:status])
+
+    unless valid_status?(params[:status])
+      return result.single_validation_failure!(
+        field: :status,
+        error_code: 'value_is_invalid',
+      )
+    end
 
     invoice.status = params[:status] if params.key?(:status)
     invoice.save!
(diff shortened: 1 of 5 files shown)


app/services/invoices/update_service.rb
```diff
@@ -5,7 +5,7 @@ def update_from_api(invoice_id:, params:)
     invoice = Invoice.find_by(id: invoice_id)
 
     return result.not_found_failure!(resource: 'invoice') if invoice.blank?
-    return result.fail!(code: 'invalid_status') unless valid_status?(params[:status])
+    return result.single_validation_failure!(field: :status, error_code: 'value_is_invalid') unless valid_status?(params[:status])
```
jdenquin, line 8: Style/IfUnlessModifier: Modifier form of `unless` makes the line too long. (https://rubystyle.guide#if-as-a-modifier)

jdenquin, line 8: Layout/LineLength: Line is too long. [130/120] (https://rubystyle.guide#max-line-length)

app/services/invoices/update_service.rb
```diff
@@ -1,46 +1,54 @@
...
-    return unless invoice.invoice_type == 'credit'
-    return unless status == 'succeeded'
-
-    Invoices::PrepaidCreditJob.perform_later(invoice)
+module Invoices
+  class UpdateService < BaseService
+    def update_from_api(invoice_id:, params:)
+      invoice = Invoice.find_by(id: invoice_id)
+
+      return result.not_found_failure!(resource: 'invoice') if invoice.blank?
```
jdenquin, line 8: do you think it's the right time to use a validate service here too?

vincent-pochet: For a single message, I think it's a bit overkill

jdenquin: we have two, but this is not the right time to do it!

jdenquin (approved):

PR #480, opened 2022-09-21
fix(errors): Add invite_not_found in error_enum

ansmonjol:
## Context

We recently added invite resource to the API. Hence having an error code `invite_not_found` when the invite is not found in `invite_resolver`
The API usually export all its error code to the GraphQL client so they can have them in an enum

## Description

We add the `invite_not_found` code in the error code enum as it has been forgot during implementation


ansmonjol:
Those error code are handled on frontend side in the graphQL resolver. Closing

PR #481, opened 2022-09-21
fix(subscription): Fix Upgrade Old Subscription Boundaries

jdenquin:
## Context

During a subscription upgrade, the amount calculated for an in advance and a different interval is wrong.
The boudaries passed to calculate the previous subscription are wrong.

## Description

- Use the previous subscription boundaries instead of the new one
- Fix some specs

### DIFF
--- a/app/services/fees/subscription_service.rb
+++ b/app/services/fees/subscription_service.rb
@@ -141,7 +141,7 @@ module Fees
 
       # NOTE: to_date for previous plan might be different from to_date
       #       if plan interval is not the same
-      old_to_date = compute_old_to_date(previous_subscription)
+      old_to_date = compute_old_to_date(previous_subscription, to_date)
 
       if plan.has_trial?
         from_date = to_date + 1.day if subscription.trial_end_date >= to_date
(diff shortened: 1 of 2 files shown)


app/services/fees/subscription_service.rb
```diff
@@ -218,12 +218,14 @@ def single_day_price(target_subscription, optional_from_date: nil)
...
-    def compute_to_date(target_subscription, base_date)
+    def compute_old_to_date(old_subscription, base_date)
       return base_date if plan.pay_in_arrear?
 
+      old_base_date = date_service(old_subscription).to_date
+
       # NOTE: when plan is pay in advance, the_to date should be the
       #       end of the actual period
-      date_service(target_subscription).next_end_of_period(base_date)
+      date_service(old_subscription).next_end_of_period(old_base_date)
```
lovrocolic, line 228: Shouldn't we just call: `date_service(old_subscription).next_end_of_period(boundaries.timestamp)` here?

jdenquin: it can do the job but I'm not sure, I may try this

vincent-pochet (approved):

PR #489, opened 2022-09-23
feat(coupons): recurring part coupons

lovrocolic:
## Context

This change includes frequency logic into coupons.

## Description

We introduce `frequency` and `frequency_duration` fields into coupons.

Applied coupon termination should happen (for recurring coupons) when frequency duration becomes zero, for both fixed amount and percentage coupons.
Also, on recurring coupons we don't have to track remaining amount on coupon anymore. Each time we should just try to use as much as possible from the coupon.


app/services/credits/applied_coupon_service.rb
```diff
@@ -21,7 +21,8 @@ def create
         amount_currency: applied_coupon.amount_currency,
       )
 
-      applied_coupon.mark_as_terminated! if applied_coupon.coupon.percentage? || credit_amount >= remaining_amount
+      applied_coupon.frequency_duration -= 1 if applied_coupon.recurring?
+      should_terminate_applied_coupon?(credit_amount) ? applied_coupon.mark_as_terminated! : save_frequency_duration
```
jdenquin, line 25: I'm generally not a big fan of ternary condition since it's a bit hard to read, can we change a bit this part?

app/services/credits/applied_coupon_service.rb
```diff
@@ -57,5 +56,31 @@ def remaining_amount
...
+    def should_terminate_applied_coupon?(credit_amount)
+      if applied_coupon.once?
+        applied_coupon.coupon.percentage? || credit_amount >= remaining_amount
+      else
+        applied_coupon.frequency_duration <= 0
+      end
+    end
+
+    def save_frequency_duration
+      return if applied_coupon.once?
```
jdenquin, line 69: it's strange to have this test here, we should call the save_frequency_duration only for a recurring coupon, wdyt?

app/services/credits/applied_coupon_service.rb
```diff
@@ -57,5 +56,31 @@ def remaining_amount
...
+
+      applied_coupon.save!
+    end
+
+    def calculate_recurring_amount
```
jdenquin, line 74: can we have only one method `calculate_amount` that take cares of the type etc?

app/services/credits/applied_coupon_service.rb
```diff
@@ -46,9 +47,7 @@ def compute_amount
         return (discounted_value >= invoice.amount_cents) ? invoice.amount_cents : discounted_value.round
       end
 
-      return invoice.amount_cents if remaining_amount > invoice.amount_cents
-
-      remaining_amount
+      applied_coupon.recurring? ? calculate_recurring_amount : calculate_one_time_coupon_amount
```
jdenquin, line 50: same thing for the ternary

spec/services/credits/applied_coupon_service_spec.rb
```diff
@@ -123,5 +123,107 @@
...
+      let(:applied_coupon) do
+        create(:applied_coupon, coupon: coupon, frequency: 'recurring', frequency_duration: 3, amount_cents: 12)
+      end
+
+      it 'creates a credit' do
+        result = credit_service.create
+
+        expect(result).to be_success
+
+        expect(result.credit.amount_cents).to eq(12)
```
jdenquin, line 139: can you aggregate failures here? 🙏

spec/services/credits/applied_coupon_service_spec.rb
```diff
@@ -123,5 +123,107 @@
...
+          frequency: 'recurring',
+          frequency_duration: 3,
+          percentage_rate: 20.00,
+        )
+      end
+
+      it 'creates a credit' do
+        result = credit_service.create
+
+        expect(result).to be_success
```
jdenquin, line 181: same here

spec/services/credits/applied_coupon_service_spec.rb
```diff
@@ -123,5 +123,107 @@
...
+            percentage_rate: 20.00,
+          )
+        end
+
+        it 'creates a credit' do
+          result = credit_service.create
+
+          expect(result).to be_success
+
+          expect(result.credit.amount_cents).to eq(25)
```
jdenquin, line 213: :D

jdenquin (changes requested):

app/services/credits/applied_coupon_service.rb
lovrocolic: Refactored a bit (apply to other comments as well). At first I was wandering that Assignment Branch Condition size would be too high so I had few more helper methods, but at the end it's not...

spec/services/credits/applied_coupon_service_spec.rb
lovrocolic: added aggregation on all places

app/services/credits/applied_coupon_service.rb
```diff
@@ -21,7 +21,12 @@ def create
         amount_currency: applied_coupon.amount_currency,
       )
 
-      applied_coupon.mark_as_terminated! if applied_coupon.coupon.percentage? || credit_amount >= remaining_amount
+      applied_coupon.frequency_duration -= 1 if applied_coupon.recurring?
+      if should_terminate_applied_coupon?(credit_amount)
+        applied_coupon.mark_as_terminated!
+      elsif applied_coupon.recurring?
```
jdenquin, line 27: a simple `else` is enought or I'm missing something?

jdenquin (changes requested):

app/services/credits/applied_coupon_service.rb
lovrocolic: Here are two cases for updating applied coupon. Either we want to mark termination if certain conditions are met or we just want to save updated frequency_duration. However, frequency_duration should be saved only for recurring coupons. We don't want to open DB connection otherwise...

jdenquin (approved):

PR #494, opened 2022-09-27
fix (coupons): refactor coupon and applied coupon validations

lovrocolic:


ansmonjol (approved):

vincent-pochet (approved):

jdenquin (approved):

app/models/coupon.rb
```diff
@@ -33,13 +33,13 @@ class Coupon < ApplicationRecord
   enum coupon_type: COUPON_TYPES
   enum frequency: FREQUENCIES
 
-  monetize :amount_cents
+  monetize :amount_cents, disable_validation: true, allow_nil: true
```
jdenquin, line 36: just for my culture, an `applied_coupon` that is based on a percentage coupon does not have an amount?!

lovrocolic: user can override amount and currency (while applying coupon) only for related coupons that are `fixed_amount`
for coupons that are `percentage` user can override only `percentage_rate`

PR #503, opened 2022-10-03
fix: prevent charging subscription fees twice in the same day

lovrocolic:
This bug fix prevents charging subscription fees twice in the same day when paying is in advance and when subscription is created on the billing day (just before BillingJob running time)

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -74,6 +74,10 @@ class Subscription < ApplicationRecord
     next_subscriptions.not_canceled.order(created_at: :desc).first
   end
 
+  def already_billed_today?
+    fees.subscription_kind.where(created_at: Time.current.beginning_of_day..Time.current.end_of_day).any?
+  end
+
   def already_billed?
     fees.subscription_kind.any?
   end
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -74,10 +74,8 @@ class Subscription < ApplicationRecord
     next_subscriptions.not_canceled.order(created_at: :desc).first
   end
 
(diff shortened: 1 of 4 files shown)


app/services/invoices/create_service.rb
```diff
@@ -82,6 +82,10 @@ def create_charges_fees(invoice, subscription, boundaries)
     end
 
     def should_create_subscription_fee?(subscription)
+      # NOTE: When plan is pay in advance we generate an invoice upon subscription creation
+      # We want to prevent creating subscription fee if subscription creation already happened on billing day
+      return false if subscription.plan.pay_in_advance? && subscription.already_billed_today?
```
vincent-pochet, line 87: Shouldn't we check for the day in the "timestamp" rather than today?

vincent-pochet (commented):
Good approach I think, my only comment would be around the date basis: Today/Timestamp

app/services/invoices/create_service.rb
lovrocolic: oh yes, you are right, thanks for catching! Fixed

app/models/subscription.rb
```diff
@@ -74,6 +74,12 @@ def next_subscription
     next_subscriptions.not_canceled.order(created_at: :desc).first
   end
 
+  def fee_exists?(timestamp)
```
vincent-pochet, line 77: Looks good, but I would just suggest to pass the date in here, rather than the timestamp :)

vincent-pochet (approved):
One last small comment. Otherwise, looks good! :rocket:

PR #506, opened 2022-10-03
feat (subscriptions): add logic to create subscription with given subscription_date

lovrocolic:
## Context

This change allows creating subscription with preselected subscription_date.

## Description

This is the first PR for `subscription on specific date` feature

In this PR only necessary logic for creating subscription on specific date is added.

Upgrade/downgrade; subscription update and converting pending to active subscriptions will be handled in following PRs

### DIFF
--- a/app/services/fees/subscription_service.rb
+++ b/app/services/fees/subscription_service.rb
@@ -84,8 +84,16 @@ module Fees
     #        or when it is payed in advance on an anniversary base
     def should_use_full_amount?
       return true if plan.pay_in_advance? && subscription.anniversary?
+      return true if subscription.fees.subscription_kind.exists?
 
-      subscription.fees.subscription_kind.exists?
+      return true if subscription.started_in_past? && plan.pay_in_advance?
+
+      if subscription.started_in_past? &&
+        subscription.started_at < date_service(subscription).previous_beginning_of_period
+        return true
+      end
+
+      false
     end
 
     def first_subscription_amount
--- a/app/services/fees/subscription_service.rb
+++ b/app/services/fees/subscription_service.rb
@@ -89,7 +89,7 @@ module Fees
       return true if subscription.started_in_past? && plan.pay_in_advance?
 
       if subscription.started_in_past? &&
-        subscription.started_at < date_service(subscription).previous_beginning_of_period
+         subscription.started_at < date_service(subscription).previous_beginning_of_period
         return true
       end
 
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -117,6 +117,8 @@ module Subscriptions
 
       if new_subscription.subscription_date > Time.current.to_date
         new_subscription.pending!
+      elsif new_subscription.subscription_date < Time.current.to_date
+        new_subscription.mark_as_active!(new_subscription.subscription_date.beginning_of_day)
       else
         new_subscription.mark_as_active!
       end
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -2,8 +2,15 @@
 
 module Subscriptions
   class CreateService < BaseService
-    attr_reader :current_customer, :current_plan, :current_subscription, :name, :external_id, :billing_time,
-                :subscription_date
+    attr_reader(
+      :current_customer,
+      :current_plan,
+      :current_subscription,
+      :name,
+      :external_id,
+      :billing_time,
+      :subscription_date,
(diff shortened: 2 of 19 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -2,7 +2,8 @@
 
 module Subscriptions
   class CreateService < BaseService
-    attr_reader :current_customer, :current_plan, :current_subscription, :name, :external_id, :billing_time
+    attr_reader :current_customer, :current_plan, :current_subscription, :name, :external_id, :billing_time,
```
jdenquin, line 5: Layout/MultilineMethodArgumentLineBreaks: Each argument in a multi-line method call must start on a separate line.

app/services/subscriptions/create_service.rb
```diff
@@ -20,6 +28,7 @@ def create_from_api(organization:, params:)
       @name = params[:name]&.strip
       @external_id = params[:external_id]&.strip
       @billing_time = params[:billing_time]
+      @subscription_date = params[:subscription_date] || Time.current.to_date
```
jdenquin, line 31: is it the same as `Time.zone.now` ?

app/services/subscriptions/create_service.rb
```diff
@@ -50,8 +60,7 @@ def create(**args)
     private
 
     def process_create
-      return result.not_found_failure!(resource: 'customer') unless current_customer
-      return result.not_found_failure!(resource: 'plan') unless current_plan
+      return result unless valid?(customer: current_customer, plan: current_plan, subscription_date: subscription_date)
```
jdenquin, line 63: this will open a discussion with @vincent-pochet I think 😂

app/services/subscriptions/create_service.rb
lovrocolic: yep, from the docs: `Returns Time.zone.now when [Time.zone](https://apidock.com/rails/Time/zone/class) or config.time_zone are set, otherwise just returns Time.now.`

app/services/subscriptions/validate_service.rb
```diff
@@ -0,0 +1,39 @@
...
+      return true if args[:plan]
+
+      add_error(field: :plan, error_code: 'plan_not_found')
+    end
+
+    def valid_subscription_date?
+      return true if args[:subscription_date].is_a?(Date)
+      return true if args[:subscription_date].is_a?(String) && Date._strptime(args[:subscription_date]).present?
+
+      add_error(field: :subscription_date, error_code: 'invalid_subscription_date')
```
vincent-pochet, line 36: I suggest to use `invalid_date` as error code, since it's already attached to the `subscription_date` field

app/services/subscriptions/validate_service.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      true
+    end
+
+    private
+
+    def valid_customer?
+      return true if args[:customer]
+
+      add_error(field: :customer, error_code: 'customer_not_found')
```
vincent-pochet, line 23: That's were our current validator implementation is limited (cc @jdenquin :trollface:)

For me if we are not able to find the customer or the plan, we should render a 404 and not a 422.

Maybe we should handle this at the `BaseValidator` level. Let me know if you need help on this topic.

app/services/subscriptions/create_service.rb
vincent-pochet: Done! :trollface:

app/services/subscriptions/validate_service.rb
lovrocolic: I am never sure what to return in such a cases :) 404 makes sense but on the other hand it's 'just' relationship, not the main object as in e.g. GET show 😅 However, refactored it a bit and returning 404 now

app/services/fees/subscription_service.rb
```diff
@@ -84,8 +84,16 @@ def should_compute_upgraded_amount?
     #        or when it is payed in advance on an anniversary base
     def should_use_full_amount?
       return true if plan.pay_in_advance? && subscription.anniversary?
+      return true if subscription.fees.subscription_kind.exists?
 
-      subscription.fees.subscription_kind.exists?
+      return true if subscription.started_in_past? && plan.pay_in_advance?
+
+      if subscription.started_in_past? &&
+        subscription.started_at < date_service(subscription).previous_beginning_of_period
```
jdenquin, line 92: Layout/MultilineOperationIndentation: Align the operands of a condition in an `if` statement spanning multiple lines.

PR #507, opened 2022-10-04
feat(credit_notes): Expose credit notes into GraphQL

vincent-pochet:
## Context

## Description

This PR exposes the customer `creditNotes` into the GraphQL schema. It is attached to the `customerDetails` type

### DIFF
--- a/app/graphql/types/fees/object.rb
+++ b/app/graphql/types/fees/object.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Types
+  module Fees
+    class Object < Types::BaseObject
+      graphql_name 'Fee'
+      implements Types::Invoices::InvoiceItem
+
+      field :vat_amount_cents, GraphQL::Types::BigInt, null: false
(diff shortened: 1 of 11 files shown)


app/graphql/types/fees/object.rb
```diff
@@ -0,0 +1,21 @@
...
+      implements Types::Invoices::InvoiceItem
+
+      field :vat_amount_cents, GraphQL::Types::BigInt, null: false
+      field :vat_amount_currency, Types::CurrencyEnum, null: false
+
+      field :vat_rate, GraphQL::Types::Float, null: true
+      field :units, GraphQL::Types::Float, null: false
+      field :events_count, GraphQL::Types::BigInt, null: true
+
+      def item_type
```
lovrocolic, line 16: is this unused?

vincent-pochet: The field itself is defined in the `Types::Invoices::InvoiceItem`

lovrocolic: 👍

lovrocolic (approved):

PR #508, opened 2022-10-04
activate pending subscriptions

lovrocolic:
## Context

This PR belongs to subscription on specific date feature

## Description

This PR is follow up to the https://github.com/getlago/lago-api/pull/506. When current date is the same as subscription date on pending subscriptions; we should activate them and trigger invoice if paying is in advance

### DIFF
--- a/app/services/subscriptions/activate_service.rb
+++ b/app/services/subscriptions/activate_service.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  class ActivateService < BaseService
+    def activate_all_pending
+      Subscription
+        .pending
+        .where(previous_subscription: nil)
+        .where(subscription_date: Time.current.to_date)
+        .find_each do |subscription|
+          subscription.mark_as_active!
+
+          BillSubscriptionJob.perform_later([subscription], Time.current.to_i) if subscription.plan.pay_in_advance?
+        end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/services/subscriptions/activate_service.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  class ActivateService < BaseService
+    def initialize(timestamp:)
```
jdenquin, line 5: Lint/MissingSuper: Call `super` to initialize state of the parent class.

spec/services/subscriptions/activate_service_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    before do
+      active_subscription
+      pending_subscriptions
+      future_pending_subscriptions
+    end
+
+    it 'activates all pending subscriptions with subscription date set to today' do
+      expect do
+        activate_service.activate_all_pending
+      end.to change(Subscription.pending, :count).by(-3).and change(Subscription.active, :count).by(3)
```
vincent-pochet, line 25:
```suggestion
      expect { activate_service.activate_all_pending }
        .to change(Subscription.pending, :count).by(-3)
        .and change(Subscription.active, :count).by(3)
```

vincent-pochet (approved):
Nice!

Just one suggestion

PR #511, opened 2022-10-07
feat(credit_notes): Download and generate credit note PDF from API

vincent-pochet:
## Context

## Description

This PR adds:
- A new route to download a credit note PDF
- A new webhook `credit_note.generated` sent when the credit note PDF is generated
- The credit note PDF generation logic

### DIFF
--- a/spec/services/credit_notes/generate_service_spec.rb
+++ b/spec/services/credit_notes/generate_service_spec.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe CreditNotes::GenerateService, type: :service do
+  subject(:credit_note_generate_service) { described_class.new }
+
+  let(:organization) { create(:organization, name: 'LAGO') }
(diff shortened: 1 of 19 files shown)


spec/services/credit_notes/generate_service_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  let(:invoice) { create(:invoice, customer: customer) }
+  let(:credit_note) { create(:credit_note, invoice: invoice, customer: customer) }
+  let(:fee) { create(:fee, invoice: invoice) }
+  let(:credit_note_item) { create(:credit_note_item, credit_note: credit_note, fee: fee) }
+
+  let(:pdf_content) do
+    File.read(Rails.root.join('spec/fixtures/blank.pdf'))
+  end
+
+  let(:pdf_generator) { instance_double(Utils::PdfGenerator) }
```
jdenquin, line 19: can we move it with all the one lines let?

jdenquin (approved):

PR #516, opened 2022-10-10
feat(credit_notes): Add logic to download a credit note via GraphQL

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds:
- A new mutation to download a credit note PDF


app/graphql/types/credit_notes/object.rb
```diff
@@ -21,8 +21,7 @@ class Object < Types::BaseObject
       field :created_at, GraphQL::Types::ISO8601DateTime, null: false
       field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
 
-      # TODO: Expose credit note document
-      # field :file_url, String, null: true
+      field :file_url, String, null: true
```
rsempe, line 24: Does it make sense to use a custom `Types::Url` for this, something like this: https://graphql-ruby.org/type_definitions/scalars.html#custom-scalars?

rsempe (approved):

lovrocolic (approved):

app/graphql/types/credit_notes/object.rb
vincent-pochet: I've checked and actually since the type is not really built-in and that this field is not supposed to be sent from the front we do not have real need for it now.

But it could make sens to have it for input types like the webhook URL for example

PR #525, opened 2022-10-10
feat(credit_notes): Apply credit notes to an invoice

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds:
- Changes on the credit note model:
  - New `total_amount` fields: to store the credit note total amount (credited amount + refunded amount)
  - Rename `remaining_amount` fields into `balance_amount`
  - Rename `amount` fields into `credit_amount`
- Logic to apply available credit_notes to a new invoice


app/services/invoices/create_service.rb
```diff
@@ -129,13 +130,21 @@ def applied_coupon
       @applied_coupon = customer.applied_coupons.active.first
     end
 
+    def credit_notes
+      customer.credit_notes.available
```
lovrocolic, line 134: What do you think about memoizing credit notes and sending it to `Credits::CreditNoteService`?

app/models/credit_note.rb
```diff
@@ -15,13 +15,14 @@ class CreditNote < ApplicationRecord
 
   has_one_attached :file
 
-  monetize :amount_cents
-  monetize :remaining_amount_cents
+  monetize :total_amount_cents
+  monetize :credit_amount_cents
+  monetize :balance_amount_cents
```
lovrocolic, line 20: can those fields be nil in certain scenarios? I remember I had bug around nil value and `monetize`...

rsempe (approved):

PR #526, opened 2022-10-10
fix(test): method called should be the same as describe

ansmonjol:
## Context

A user service spec is present in the codebase.

It does have `describe` blocks per methods in the service.

I noticed the `login` block was calling `register` method

## Description

This PR fixes this by calling the `login` method instead.

We need to create a membership first too.

Please note the `login` method accept only 2 args

A test for the `register` method is already present in the file


vincent-pochet (approved):
Good catch! :rocket:

jdenquin (approved):

PR #530, opened 2022-10-11
fix(user_type): only user's organizations with an active membership s…

ansmonjol:
## Context

When querying for the currentUser, we can retrieve it's list of organizations.

This list is used by the Frontend app in order to log the user in the correct organization.

Today the first orga in the array is choose by default.

Recently with the 'invite member' feature, your membership can have a revoked status on an organisation.

## Description

If you were revoked by an organization, the organization was still returned in the `currentUser` query.

This PR fixes that by overriding the `organizations` relationship in the `Types::UserType`

## To go further

This Pr does not fix all issues a user can face.

For instance, if you're invite to 2 different organization, on log-in you will only access the first one you joined.

This can be fixed later with a "switch organization" feature, that appear to be Frontend only.

### DIFF
--- a/spec/graphql/resolvers/current_user_resolver_spec.rb
+++ b/spec/graphql/resolvers/current_user_resolver_spec.rb
@@ -9,6 +9,9 @@ RSpec.describe Resolvers::CurrentUserResolver, type: :graphql do
         currentUser {
           id
           email
+          organizations {
+            id
+          }
         }
       }
     GRAPHQL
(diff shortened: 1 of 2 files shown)


spec/graphql/resolvers/current_user_resolver_spec.rb
```diff
@@ -28,6 +31,24 @@
...
+
+    before { revoked_membership }
+
+    it 'only lists organizations when membership has an active status' do
+      result = execute_graphql(
+        current_user: membership.user,
+        query: query,
+      )
+
+      expect(result['data']['currentUser']['organizations'].length).to eq(1)
```
rsempe, line 48: Can you check that revoked membership is not present instead?

Something like:
`expect(result['data']['currentUser']['organizations']).not_to include(...)`

ansmonjol: I've changed with this code, ok for you ? 

`expect(result['data']['currentUser']['organizations']).not_to include(revoked_membership.organization)`

PR #538, opened 2022-10-13
feat(credit_notes): Create a credit note from subscription upgrade

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds the logic to create a credit note from a subscription upgrade


spec/services/subscriptions/create_service_spec.rb
```diff
@@ -482,6 +482,31 @@
             end
           end
 
+          context 'when old subscription was payed in advabce' do
```
lovrocolic, line 485: `advance` You were testing us here to see if we are going through it carefully :D

lovrocolic (approved):

PR #541, opened 2022-10-14
feat (gocardless): add possibility to create and update gocardless payment provider

lovrocolic:
## Context

OAuth with the Gocardless

## Description

This PR covers only one step in the integration process. At one point Lago FE will create mutation with the auth code so that Lago API can request auth token based on auth code. Given access token needs to be stored in Payment provider object and this token will be used in all other requests to gocardless...


spec/services/payment_providers/gocardless_service_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe PaymentProviders::GocardlessService, type: :service do
+  subject(:gocardless_service) { described_class.new(membership.user) }
```
jdenquin, line 6: RSpec/EmptyLineAfterSubject: Add an empty line after `subject`. (https://rspec.rubystyle.guide/#empty-line-after-let, https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/EmptyLineAfterSubject)

vincent-pochet (approved):
Looks good! 

I have one question though, does GoCardless access tokens expires, and if, do they provide a refresh token (that we should probably store in the DB as well)?

lovrocolic:
> Looks good!
> 
> I have one question though, does GoCardless access tokens expires, and if, do they provide a refresh token (that we should probably store in the DB as well)?

From what I saw during my investigation, they do not provide refresh token and access token does not expire

PR #544, opened 2022-10-17
feat(groups): Ability to fetch billable metric's groups

rsempe:
## Roadmap Task

## Context

## Description

This PR adds the ability to fetch all groups for a specific billable metric.

In a future PR, we'll add the ability to create a charge by defining properties on groups, something like this:
```
group_properties: [
  :group_id,
  { values: {} },
]
```

We should be able to retrieve the id of a group in order to assign values to it for a specific charge.

### DIFF
--- a/spec/requests/api/v1/groups_spec.rb
+++ b/spec/requests/api/v1/groups_spec.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::GroupsController, type: :request do
+  let(:organization) { create(:organization) }
+  let(:billable_metric) { create(:billable_metric, organization: organization) }
+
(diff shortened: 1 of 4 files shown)


spec/requests/api/v1/groups_spec.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::GroupsController, type: :request do
```
jdenquin, line 5: RSpec/FilePath: Spec path should end with `api/v1/groups_controller*_spec.rb`. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FilePath)

vincent-pochet (approved):
Pronto! :trollface:

PR #549, opened 2022-10-19
feat(groups): Ability to create charge with group properties

rsempe:
## Roadmap Task

## Context

## Description

This PR adds the ability to create charge with group properties.


app/controllers/api/v1/plans_controller.rb
```diff
@@ -88,7 +88,16 @@ def input_params
           :trial_period,
           :pay_in_advance,
           :bill_charges_monthly,
-          charges: [:id, :billable_metric_id, :charge_model, { properties: {} }],
+          charges: [
+            :id,
+            :billable_metric_id,
+            :charge_model,
+            { properties: {} },
+            group_properties: [
```
jdenquin, line 96: Style/HashAsLastArrayItem: Wrap hash in `{` and `}`. (https://rubystyle.guide#hash-literal-as-last-array-item)

vincent-pochet (approved):

PR #550, opened 2022-10-19
feat(groups) Ensure that all groups is referenced on group properties

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to:
- Fix selectable groups when one dimension
- Add a validation when creating/updating a charge with group properties

### DIFF
--- a/app/models/charge.rb
+++ b/app/models/charge.rb
@@ -63,9 +63,9 @@ class Charge < ApplicationRecord
   end
 
   def validate_group_properties
-    # Group properties should be set for all the groups of a BM
-    bm_group_ids = billable_metric.groups.pluck(:id).sort
-    gp_group_ids = group_properties.map { |gp| gp[:group_id] }.sort
+    # Group properties should be set for all the selectable groups of a BM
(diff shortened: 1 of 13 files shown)


app/models/charge.rb
```diff
@@ -59,4 +61,12 @@ def validate_charge_model(validator)
       .flatten
       .each { |code| errors.add(:properties, code) }
   end
+
+  def validate_group_properties
+    # Group properties should be set for all the selectable groups of a BM
+    bm_group_ids = billable_metric.selectable_groups.pluck(:id).sort
```
lovrocolic, line 67: It would be more efficient to sort it on DB level (especially if we have right indexes set) but probably it's fine to sort it in this example in memory also (if there are small umber of records)

lovrocolic (approved):

vincent-pochet (approved):
:rocket:

app/models/charge.rb
rsempe: Yeah, I think it's acceptable for now (as we can have only two dimensions) but I agree with you

PR #556, opened 2022-10-24
feat(credit_notes): Add refund fields to credit notes

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds the mandatory database fields to allow a credit note to be handled as a refund

### DIFF
--- a/app/models/credit_note.rb
+++ b/app/models/credit_note.rb
@@ -21,10 +21,12 @@ class CreditNote < ApplicationRecord
   monetize :refund_amount_cents
   monetize :vat_amount_cents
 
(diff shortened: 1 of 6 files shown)


app/models/credit_note.rb
```diff
@@ -21,10 +21,12 @@ class CreditNote < ApplicationRecord
   monetize :refund_amount_cents
   monetize :vat_amount_cents
 
-  CREDIT_STATUS = %i[available consumed].freeze
+  CREDIT_STATUS = %i[available consumed no_credit].freeze
```
rsempe, line 24: hard to understand what `no_credit` means without a note 😬

rsempe (approved):

PR #558, opened 2022-10-24
fix(invoice): organization country

jdenquin:
## Context

The organization country is actually bad on the invoice

## Description

- Fix the organization country


vincent-pochet (approved):
Oups :smile: 

:rocket:

PR #559, opened 2022-10-24
feat(credit_notes): Handle refund amount for credit note creation

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds the logic to handle refund amount when creating a credit note


lovrocolic (approved):
🥇

PR #561, opened 2022-10-25
feat(credit_notes): Create a credit note via GraphQL

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds the GraphQL types and mutations to create a credit note with credit and/or refund


app/graphql/types/credit_notes/object.rb
```diff
@@ -9,7 +9,8 @@ class Object < Types::BaseObject
       field :sequential_id, ID, null: false
       field :number, String, null: false
 
-      field :credit_status, Types::CreditNotes::CreditStatusTypeEnum, null: false
+      field :credit_status, Types::CreditNotes::CreditStatusTypeEnum, null: true
+      field :refund_status, Types::CreditNotes::RefundStatusTypeEnum, null: true
```
rsempe, line 13: Why `null: true` for both statuses?

rsempe (approved):

app/graphql/types/credit_notes/object.rb
vincent-pochet: If `credit_amount_cents` is 0, we have no need for a `credit_status`, same logic for `refund_amount_cents` and `refund_status`

PR #573, opened 2022-10-28
misc: Fix currency to fees from the invoice

rsempe:
We currently have an issue where the invoice currency can be different from the charges currency.

The goal of this PR is to fix that issue.


vincent-pochet (approved):
:ringed_planet:

jdenquin (approved):

PR #575, opened 2022-10-31
feat (gocardless): fetch mandate before payment

lovrocolic:
## Context

Add few improvements related to gocardless feature

## Description

1. Fetch mandate before each payment
2. On UI, we won't be using mandate_id field anymore, so it should be removed
3. Generate webhook secret key (will be used on UI) while generating gocardless access token

### DIFF
--- a/spec/services/invoices/payments/gocardless_service_spec.rb
+++ b/spec/services/invoices/payments/gocardless_service_spec.rb
@@ -8,9 +8,11 @@ RSpec.describe Invoices::Payments::GocardlessService, type: :service do
   let(:customer) { create(:customer) }
   let(:organization) { customer.organization }
   let(:gocardless_payment_provider) { create(:gocardless_provider, organization: organization) }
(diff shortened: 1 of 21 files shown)


spec/services/invoices/payments/gocardless_service_spec.rb
```diff
@@ -28,6 +30,14 @@
 
       allow(GoCardlessPro::Client).to receive(:new)
         .and_return(gocardless_client)
+      allow(gocardless_client).to receive(:mandates)
+        .and_return(gocardless_mandates_service)
+      allow(gocardless_mandates_service).to receive(:list)
+        .and_return(gocardless_list_response)
+      allow(gocardless_list_response).to receive(:records)
+        .and_return([GoCardlessPro::Resources::Mandate.new(
```
jdenquin, line 38: Layout/FirstArrayElementLineBreak: Add a line break before the first element of a multi-line array.

vincent-pochet (approved):

PR #580, opened 2022-10-31
feat(credit_notes): Add void GraphQL mutation for credit notes

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds the void GraphQL mutation to "terminate" a credit note with remaining balance


app/services/credit_notes/void_service.rb
```diff
@@ -0,0 +1,29 @@
...
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: 'credit_note') if credit_note.nil?
+
+      result.credit_note = credit_note
+      return result.not_allowed_failure!(code: 'no_voidable_amount') unless credit_note.voidable?
+
+      credit_note.update!(
```
ansmonjol, line 17: Just a suggestion.

Maybe create a `mark_as_void` or similar method in the credit note model instead ?

ansmonjol (approved):
Just a non blocking comment

app/services/credit_notes/void_service.rb
vincent-pochet: Good catch!

PR #585, opened 2022-11-03
feat(credit_notes): Refund credit note via stripe

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds the stripe refund logic


app/services/credit_notes/refunds/stripe_service.rb
```diff
@@ -0,0 +1,151 @@
...
+      end
+
+      def update_credit_note_status(status)
+        credit_note.update!(refund_status: status)
+      end
+
+      def track_refund_status_changed(status)
+        SegmentTrackJob.perform_later(
+          membership_id: CurrentContext.membership,
+          event: 'refund_status_change',
```
rsempe, line 141: I think we're always using the past tense for naming events on segment, what about `refund_status_changed`?

PR #590, opened 2022-11-04
feat(credit_notes): Void a credit note via API

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds the `PUT /api/v1/credit_notes/:id/void` route to void a credit note via the API

### DIFF
--- a/app/controllers/api/v1/credit_notes_controller.rb
+++ b/app/controllers/api/v1/credit_notes_controller.rb
@@ -77,6 +77,25 @@ module Api
         head(:ok)
       end
 
+      def void
+        credit_note = current_organization.credit_notes.find_by(id: params[:id])
+        return not_found_error(resource: 'credit_note') unless credit_note
+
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/credit_notes_controller.rb
```diff
@@ -77,6 +77,25 @@ def download
         head(:ok)
       end
 
+      def void
+        credit_note = current_organization.credit_notes.find_by(id: params[:id])
```
rsempe, line 81: Does it make sense to have an external id here? 🤔

rsempe (approved):

app/controllers/api/v1/credit_notes_controller.rb
vincent-pochet: Well like invoice it will be mostly created automatically, and it does not belongs to a resource from the customer system, so I don't really think it make sense

PR #595, opened 2022-11-11
fix (gocardless): fix calling invoice without its instance var

lovrocolic:
Private method called `handle_prepaid_credits` is called from method where `invoice` instance variable is present but also from another method where `invoice` instance variable is not present

Also, in this PR few other improvements are added:
1.) gocardless auth site url is handled in model instead using hardcoded string that applies only to sandbox
2.) migration is added that sets sync_with_provider field to true for existing customers (if customer's payment provider is set to stripe)

### DIFF
--- a/lib/tasks/customers.rake
+++ b/lib/tasks/customers.rake
@@ -15,4 +15,20 @@ namespace :customers do
       customer.update!(currency: currencies.first)
     end
   end
+
+  desc 'Set sync_with_provider field for existing customers'
+  task populate_sync_with_provider: :environment do
+    Organization.all.each do |org|
+      next if org&.stripe_payment_provider&.create_customers&.blank?
+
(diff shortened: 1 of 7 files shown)


lib/tasks/customers.rake
```diff
@@ -15,4 +15,20 @@ namespace :customers do
       customer.update!(currency: currencies.first)
     end
   end
+
+  desc 'Set sync_with_provider field for existing customers'
+  task populate_sync_with_provider: :environment do
+    Organization.all.each do |org|
+      next if org&.stripe_payment_provider&.create_customers&.blank?
```
jdenquin, line 22: Rails/SafeNavigationWithBlank: Avoid calling `blank?` with the safe navigation operator in conditionals.

vincent-pochet (approved):

PR #596, opened 2022-11-14
feat: add invoice_credit_notes resolver

ansmonjol:
## Roadmap Task

👉 https://github.com/getlago/lago/issues/59

## Context

We need to fetch credit notes from an invoice details view

Please note this PR is merged into a feature branch

## Description

This PR adds the endpoints for it

### DIFF
--- a/app/graphql/resolvers/invoice_credit_notes_resolver.rb
+++ b/app/graphql/resolvers/invoice_credit_notes_resolver.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Resolvers
+  class InvoiceCreditNotesResolver < Resolvers::BaseResolver
+    include AuthenticableApiUser
+    include RequiredOrganization
+
+    description "Query invoice's credit note"
+
+    argument :invoice_id, ID, required: true, description: 'Uniq ID of the invoice'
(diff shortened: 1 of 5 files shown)


app/graphql/resolvers/invoice_credit_notes_resolver.rb
```diff
@@ -0,0 +1,28 @@
...
+    argument :page, Integer, required: false
+    argument :limit, Integer, required: false
+
+    type Types::CreditNotes::Object.collection_type, null: true
+
+    def resolve(invoice_id: nil, page: nil, limit: nil)
+      validate_organization!
+
+      Invoice.find(invoice_id)
+        .credit_notes
```
rsempe, line 20: Are we sure we want to return all credit notes here?

Even the ones not available or already voided?

rsempe (approved):

app/graphql/resolvers/invoice_credit_notes_resolver.rb
ansmonjol: Good point, checking

ansmonjol: Apparently no ! We need to display all credit notes at this point

PR #598, opened 2022-11-15
feat(customer): add credit notes related data for UI purpose

ansmonjol:
## Roadmap Task

👉 https://github.com/getlago/lago/issues/59

## Context

This PR is part of the credit note feature.

We need the information about customer credit notes (count and remaining credits)

## Description

We add those datas to the customer type, to be returned to frontend app in order to handle UI logic

### DIFF
--- a/app/graphql/types/customers/object.rb
+++ b/app/graphql/types/customers/object.rb
@@ -37,6 +37,14 @@ module Types
       field :has_active_wallet, Boolean, null: false, description: 'Define if a customer has an active wallet'
       field :has_credit_notes, Boolean, null: false, description: 'Define if a customer has any credit note'
       field :active_subscription_count, Integer, null: false, description: 'Number of active subscriptions per customer'
+      field :credit_notes_credits_available_count,
+            Integer,
+            null: false,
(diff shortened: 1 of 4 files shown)


vincent-pochet (approved):

app/graphql/types/customers/object.rb
```diff
@@ -65,6 +73,14 @@ def active_subscription_count
       def can_edit_attributes
         object.editable?
       end
+
+      def credit_notes_credits_available_count
+        object.credit_notes.where('credit_notes.credit_amount_cents > 0').count
```
rsempe, line 78: You're using the plural for `credits_available`but the singular for `credit_amount_cents`, not really consistent, right? 🤔

rsempe (approved):

app/graphql/types/customers/object.rb
ansmonjol: hum, was about the count of them all so I put an `s`, still not convinced ? 😄

rsempe: true! my bad!

PR #602, opened 2022-11-15
feat(timezone): Add timezone to organizations and customers

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR is the first of the feature. It adds two new database columnes:
- `organizations#timezone`
- `customers#timezone`

### DIFF
--- a/spec/models/customer_spec.rb
+++ b/spec/models/customer_spec.rb
@@ -27,6 +27,19 @@ RSpec.describe Customer, type: :model do
       customer.country = ''
       expect(customer).not_to be_valid
     end
+
+    it 'validates the timezone' do
+      expect(customer).to be_valid
+
+      customer.timezone = 'Europe/Paris'
+      expect(customer).to be_valid
+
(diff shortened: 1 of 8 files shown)


lovrocolic (approved):

spec/models/customer_spec.rb
```diff
@@ -27,6 +27,19 @@
...
+      expect(customer).to be_valid
+
+      customer.timezone = 'Europe/Paris'
+      expect(customer).to be_valid
+
+      customer.timezone = 'foo'
+      expect(customer).not_to be_valid
+
+      customer.timezone = nil
+      expect(customer).to be_valid
```
rsempe, line 41: is it not the same as line 32?

rsempe (approved):

spec/models/customer_spec.rb
vincent-pochet: Actually yes :smile:. Good catch!

PR #604, opened 2022-11-16
feat (coupons): allow applying multiple coupons to the customer

lovrocolic:
## Context

This change handles applying multiple coupons to the customer.

## Description

Applying multiple coupons to the customer requires removing validation on the BE side upon apply_coupon creation and updating logic when generating an invoice.

From now on, we will also track `recurring` case with the `frequency_duration_remaining` field so that we don't change originally value that is set on applied coupon object. This field will also be needed in the API endpoint for applied_coupons collection.

### DIFF
--- a/app/services/invoices/create_service.rb
+++ b/app/services/invoices/create_service.rb
@@ -178,7 +178,7 @@ module Invoices
 
     def create_coupon_credit(invoice)
       applied_coupons.each do |applied_coupon|
-        return unless invoice.amount_cents&.positive?
+        break unless invoice.amount_cents&.positive?
 
         next if applied_coupon.coupon.fixed_amount? && applied_coupon.amount_currency != currency
 
--- a/app/services/invoices/create_service.rb
+++ b/app/services/invoices/create_service.rb
@@ -36,7 +36,7 @@ module Invoices
         compute_amounts(invoice)
 
         create_credit_note_credit(invoice) if should_create_credit_note_credit?
-        create_coupon_credit(invoice) if should_create_coupon_credit?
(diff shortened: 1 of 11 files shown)


app/services/invoices/create_service.rb
```diff
@@ -178,13 +177,19 @@ def create_credit_note_credit(invoice)
     end
 
     def create_coupon_credit(invoice)
-      credit_result = Credits::AppliedCouponService.new(
-        invoice: invoice,
-        applied_coupon: applied_coupon,
-      ).create
-      credit_result.throw_error unless credit_result.success?
+      applied_coupons.each do |applied_coupon|
+        return unless invoice.amount_cents&.positive?
```
jdenquin, line 181: Lint/NonLocalExitFromIterator: Non-local exit from iterator, without return value. `next`, `break`, `Array#find`, `Array#any?`, etc. is preferred.

spec/services/credits/applied_coupon_service_spec.rb
```diff
@@ -131,7 +131,14 @@
...
 
       let(:applied_coupon) do
-        create(:applied_coupon, coupon: coupon, frequency: 'recurring', frequency_duration: 3, amount_cents: 12)
+        create(
+          :applied_coupon,
+          coupon: coupon,
+          frequency: 'recurring',
+          frequency_duration: 3,
+          frequency_duration_remaining: 3,
+          amount_cents: 12
```
jdenquin, line 140: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

vincent-pochet (approved):

PR #608, opened 2022-11-17
feat(graph) Add some graphQL fields for UI purpose

ansmonjol:
## Roadmap Task

👉  https://github.com/getlago/lago/issues/{{NUMBER}}

## Context

This PR is part of the credit notes work and will be released on the feature branch

## Description

This PR adds some graphQL fields needed for UI porpose

### DIFF
--- a/app/graphql/types/credit_notes/object.rb
+++ b/app/graphql/types/credit_notes/object.rb
@@ -40,6 +40,7 @@ module Types
 
       field :file_url, String, null: true
 
+      field :customer, Types::Customers::Object, null: false
       field :invoice, Types::Invoices::Object
       field :items, [Types::CreditNoteItems::Object], null: false
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -80,7 +80,7 @@ ActiveRecord::Schema[7.0].define(version: 2022_11_18_084547) do
     t.decimal "percentage_rate", precision: 10, scale: 5
     t.integer "frequency", default: 0, null: false
     t.integer "frequency_duration"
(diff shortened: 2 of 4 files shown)


app/graphql/types/fees/object.rb
```diff
@@ -20,6 +20,12 @@ class Object < Types::BaseObject
 
       field :creditable_amount_cents, Integer, null: false
 
+      field :amount_currency, Types::CurrencyEnum, null: false
+
+      def amount_currency
+        object.currency
```
vincent-pochet, line 26: Actually, this fields is already defined in `/app/graphql/types/invoices/invoice_item.rb` which is implemented by this object

app/graphql/types/credit_notes/object.rb
```diff
@@ -37,13 +37,24 @@ class Object < Types::BaseObject
 
       field :file_url, String, null: true
 
+      field :customer, Types::Customers::Object, null: false
       field :invoice, Types::Invoices::Object
       field :items, [Types::CreditNoteItems::Object], null: false
 
+      field :is_credited, Boolean, null: false
+      field :is_refunded, Boolean, null: false
```
vincent-pochet, line 45: Do you have a need for these fields?

The methods they rely on are defined like this:

```ruby
  def credited?
    credit_amount_cents.positive?
  end

  def refunded?
    refund_amount_cents.positive?
  end
```

and both `refund_amount_cents` and `credit_amount_cents` are already exposed

ansmonjol: I could remove them indeed ! Done

app/graphql/types/fees/object.rb
ansmonjol: indeed, removed

vincent-pochet (approved):

PR #620, opened 2022-11-23
bug: fix customer count wrongly calculated

ansmonjol:
## Context

We display the number of "customer count" on different places in the app.
- Plans list
- Coupons list
- Addons list

This value were wrong at 2 places, as we calculated all active subscription instead

## Description

This PR corrects the value returned by those methods


vincent-pochet (approved):
Nice catch!

Maybe it worth to add some tests

PR #625, opened 2022-11-25
feat(grace_period): Create invoice as draft if applicable grace period

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to create invoice as draft if applicable grace period.

### DIFF
--- a/app/services/invoices/create_service.rb
+++ b/app/services/invoices/create_service.rb
@@ -49,7 +49,9 @@ module Invoices
         result.invoice = invoice
       end
 
-      unless grace_period?
+      if grace_period?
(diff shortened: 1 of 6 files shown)


app/services/invoices/create_service.rb
```diff
@@ -61,10 +64,22 @@ def create
 
     attr_accessor :subscriptions, :timestamp, :customer, :currency
 
+    def issuing_date
+      @issuing_date ||= Time.zone.at(timestamp).to_date
+    end
+
+    def grace_period?
```
lovrocolic, line 71: maybe we should memoize this method since it's called multiple times

lovrocolic (approved):

PR #628, opened 2022-11-25
feat(timezones): Add issuing date in customer timezone on Credit notes

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR updates the logic around credit notes to add an `issuing_date`. This date will be in the customer timezone.


db/migrate/20221125111605_add_issuing_date_to_credit_notes.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_time_literal: true
```
jdenquin, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

rsempe (approved):

PR #634, opened 2022-11-29
feat(timezones): Coupon expiration date

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR updates the coupon to turn `expiration_date` into `expiration_at`, it is keeping the compatibility with the existing fields.


app/services/legacy/coupon_mapper.rb
```diff
@@ -0,0 +1,14 @@
+module Legacy
```
jdenquin, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

lovrocolic (approved):

PR #640, opened 2022-12-01
fix: Trial period set to zero should not be considered as present

vincent-pochet:
## Context

When setting a trial period at `0` the first subscription of a paid in advance calendar plan is wrongly billed at 0 when created on the last day of the period

## Description

The issue is related to a wrong assumption of the behavior of the rails `present?` method:
```ruby
0.0.present?
# => true
```

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -39,7 +39,7 @@ class Plan < ApplicationRecord
   end
 
   def has_trial?
-    trial_period.present?
+    trial_period.present? && trial_period.positive?
   end
 
   # NOTE: Method used to compare plan for upgrade / downgrade on
--- a/spec/models/plan_spec.rb
+++ b/spec/models/plan_spec.rb
@@ -4,10 +4,18 @@ require 'rails_helper'
 
 RSpec.describe Plan, type: :model do
   describe '.has_trial?' do
-    let(:plan) { create(:plan, trial_period: 3) }
+    let(:plan) { build(:plan, trial_period: 3) }
 
     it 'returns true when trial_period' do
       expect(plan).to have_trial
+    end
+
+    context 'when value is 0' do
+      let(:plan) { build(:plan, trial_period: 0) }
+
+      it 'returns false' do
+        expect(plan).not_to have_trial
+      end
     end
   end
 


spec/models/plan_spec.rb
```diff
@@ -4,11 +4,19 @@
...
 
     it 'returns true when trial_period' do
       expect(plan).to have_trial
     end
+
+    context 'when value is 0' do
+      let(:plan) { build(:plan, trial_period: 0) }
+
+      it 'returns false' do
+        expect(plan).to have_trial
```
jdenquin, line 17: Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body beginning. (https://rubystyle.guide#empty-lines-around-bodies)

app/models/plan.rb
```diff
@@ -39,7 +39,7 @@ def deletable?
   end
 
   def has_trial?
-    trial_period.present?
+    trial_period.present? && trial_period.positive?
```
rsempe, line 42: Don't we want to default `trial_period` to `0.0` instead of managing `nil` values?

lovrocolic (approved):

spec/models/plan_spec.rb
rsempe, line 17: That's the inverse, no? 🤔

vincent-pochet: Copy-past... :facepalm:

app/models/plan.rb
ansmonjol, line 42: Ruby question: won't `.positive?` also check `.present?` so we can only keep one condition ?

vincent-pochet: Hmm, good question :thinking:  (however the issue would be the same)

rsempe: `nil.positive?` => `undefined method `positive?' for nil:NilClass` ;)

vincent-pochet: Not exactly:

```ruby
nil.positive?
# => undefined method `positive?' for nil:NilClass (NoMethodError)
```

But it's related to @rsempe point

ansmonjol: oh yeah so we want to make sure it's present 😄 
thanks 👍

rsempe (approved):

PR #644, opened 2022-12-02
feat(timezone): Wallet expiration date

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR updates the wallet to turn `expiration_date` into `expiration_at`, it is keeping the compatibility with the existing fields.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_11_28_132620) do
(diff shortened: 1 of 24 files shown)


lovrocolic (approved):

db/schema.rb
```diff
@@ -328,6 +328,8 @@
     t.uuid "customer_id"
     t.boolean "legacy", default: false, null: false
     t.float "vat_rate"
+    t.bigint "credit_amount_cents", default: 0, null: false
+    t.string "credit_amount_currency"
```
lovrocolic, line 332: does it belong in this PR?

PR #652, opened 2022-12-06
Add autoprovisioning of member & orgs

zobione:
## Context

Lago by design is used to push events using one of its clients to the api server. As a developper, often times, the event collection happens on a different app than Lago. In a modern dev environment where apps are deployed at the same time this can lead to a confusing developer experience. Let's take an example where `app_1` is used to collect billable metrics and `lago_api` is used to store it. A developper will boot both app in its cluster (docker-compose, k8s, ...). Since `app_1` is requesting lago client to be able to work the developper has to inject the lago url as well as the token in then ENV variables of `app_1`. But those credentials only exists when the developer logs in to the newly deployed instance of lago, signs up for an account and get his/her api key, then proceed to update the conf of `app_1` and reboot `app_1`. Then the whole cluster can work. This is a suitable experience when developping with Lago Cloud, but it doesn't scale when using the on premise version.   

## Description

The idea of this pull request is to introduce a new way to provision Lago in a local cluster. This will ensure that the api key is deterministic and can be injected in all dependencies (app_1) at boot time without having to manually fetch it. In order to do so I have introduced an optionnal `Rake task`that is called on:
- start.sh
- start.dev.sh

This task will run as follow:
- Check if the developer has injected the required environment variables to create an org, a user and the api_key of the org
- If none of the parameters are passed the task will do nothing thus following the standard way to deploy Lago
- If the parameters are passed it will create or update the org, user and api_key allowing the developer to be able to log in right away when Lago is deployed.
- If all parameters are not properly passed then it will throw an error to guide the developer in the right direction

This fixes the issues raised above.

### DIFF
--- a/lib/tasks/signup.rake
+++ b/lib/tasks/signup.rake
@@ -1,21 +1,21 @@
 namespace :signup do
   desc "This task seeds lago with an organisation & a user for on premise deployment"
   task seed_organization: :environment do
-    if ENV['CREATE_ORG'].present? && ENV['CREATE_ORG'] == 'true'
+    if ENV['LAGO_CREATE_ORG'].present? && ENV['LAGO_CREATE_ORG'] == 'true'
       pp "starting seeding environment"
-      if ENV['ORG_USER_PASSWORD'].present? && ENV['ORG_USER_EMAIL'].present? && ENV['ORG_NAME'].present?
-        user = User.create_with(password: ENV['ORG_USER_PASSWORD'])
-          .find_or_create_by(email: ENV['ORG_USER_EMAIL'])
-        organization = Organization.find_or_create_by!(name: ENV['ORG_NAME'])
-        if ENV['ORG_API_KEY'].present?
-          organization.api_key = ENV['ORG_API_KEY']
-          organization.save
+      if ENV['LAGO_ORG_USER_PASSWORD'].present? && ENV['LAGO_ORG_USER_EMAIL'].present? && ENV['LAGO_ORG_NAME'].present?
+        user = User.create_with(password: ENV['LAGO_ORG_USER_PASSWORD'])
+          .find_or_create_by!(email: ENV['LAGO_ORG_USER_EMAIL'])
+        organization = Organization.find_or_create_by!(name: ENV['LAGO_ORG_NAME'])
+        if ENV['LAGO_ORG_API_KEY'].present?
+          organization.api_key = ENV['LAGO_ORG_API_KEY']
+          organization.save!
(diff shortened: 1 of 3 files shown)


lib/tasks/signup.rake
```diff
@@ -0,0 +1,23 @@
...
+  task seed_organization: :environment do
+    if ENV['CREATE_ORG'].present? && ENV['CREATE_ORG'] == 'true'
+      pp "starting seeding environment"
+      if ENV['ORG_USER_PASSWORD'].present? && ENV['ORG_USER_EMAIL'].present? && ENV['ORG_NAME'].present?
+        user = User.create_with(password: ENV['ORG_USER_PASSWORD'])
+          .find_or_create_by(email: ENV['ORG_USER_EMAIL'])
+        organization = Organization.find_or_create_by!(name: ENV['ORG_NAME'])
+        if ENV['ORG_API_KEY'].present?
+          organization.api_key = ENV['ORG_API_KEY']
+          organization.save
```
vincent-pochet, line 12: Could you use `organization.save!` to make sure the task fails if in case of issue with the model validation?

lib/tasks/signup.rake
```diff
@@ -0,0 +1,23 @@
+namespace :signup do
+  desc "This task seeds lago with an organisation & a user for on premise deployment"
+  task seed_organization: :environment do
+    if ENV['CREATE_ORG'].present? && ENV['CREATE_ORG'] == 'true'
```
vincent-pochet, line 4: We try to prefix every inner environment variable by `LAGO_` to make sure it does not interfere with other application. Do you mind using the same approach?

lib/tasks/signup.rake
```diff
@@ -0,0 +1,23 @@
...
+      pp "starting seeding environment"
+      if ENV['ORG_USER_PASSWORD'].present? && ENV['ORG_USER_EMAIL'].present? && ENV['ORG_NAME'].present?
+        user = User.create_with(password: ENV['ORG_USER_PASSWORD'])
+          .find_or_create_by(email: ENV['ORG_USER_EMAIL'])
```
vincent-pochet, line 8: Could you use `find_or_create_by!`?

vincent-pochet (commented):
Hello @zobione

Thank you very much for this pull request and the clear description of the change. 
It looks great.

I only have few comments on it but I think it will be good to go :rocket:

lib/tasks/signup.rake
zobione: Should be fixed

PR #653, opened 2022-12-06
feat(timezones): Handle customer and organization timezones

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR is the "Final" one for the feature, it contains all breaking changes that cannot be merged before the feature delivery

### DIFF
--- a/app/services/customers/create_service.rb
+++ b/app/services/customers/create_service.rb
@@ -20,6 +20,7 @@ module Customers
         customer.logo_url = params[:logo_url] if params.key?(:logo_url)
         customer.legal_name = params[:legal_name] if params.key?(:legal_name)
         customer.legal_number = params[:legal_number] if params.key?(:legal_number)
+        customer.timezone = params[:timezone] if params.key?(:timezone)
 
         if params.key?(:currency)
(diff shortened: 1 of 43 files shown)


app/services/customers/create_service.rb
```diff
@@ -21,6 +21,9 @@ def create_from_api(organization:, params:)
         customer.legal_name = params[:legal_name] if params.key?(:legal_name)
         customer.legal_number = params[:legal_number] if params.key?(:legal_number)
 
+        # TODO: Timezone update is turned of for now
```
rsempe, line 24: `off`

spec/services/customers/update_service_spec.rb
```diff
@@ -33,6 +34,7 @@
       updated_customer = result.customer
       aggregate_failures do
         expect(updated_customer.name).to eq('Updated customer name')
+        # expect(updated_customer.timezone).to eq('Europe/Paris') # TODO: Timezone update is turned of for now
```
rsempe, line 37: What about having a naming like `TODO(:timezone)` so that it will be easier for searching?

rsempe (approved):

PR #655, opened 2022-12-06
feat(timezones): Turn subscription#subscription_date into subscription_at

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR turn the `subscription_date` (`date`) field into a `subscription_at` (`datetime`) for subscriptions.
For now, it does not changes the API contract, it will be handled in a new PR

### DIFF
--- a/app/serializers/v1/subscription_serializer.rb
+++ b/app/serializers/v1/subscription_serializer.rb
@@ -12,7 +12,7 @@ module V1
         plan_code: model.plan.code,
         status: model.status,
         billing_time: model.billing_time,
-        subscription_date: model.subscription_date&.iso8601,
+        subscription_date: model.subscription_at&.to_date&.iso8601,
         started_at: model.started_at&.iso8601,
         terminated_at: model.terminated_at&.iso8601,
         canceled_at: model.canceled_at&.iso8601,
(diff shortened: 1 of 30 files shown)


app/services/subscriptions/dates/yearly_service.rb
```diff
@@ -78,9 +78,9 @@ def compute_previous_beginning_of_period(date)
       end
 
       def previous_anniversary_day(date)
-        year = date.month < subscription_date.month ? date.year - 1 : date.year
-        month = subscription_date.month
-        day = subscription_date.day
+        year = date.month < subscription_at.month ? date.year - 1 : date.year
```
jdenquin, line 81: Style/TernaryParentheses: Use parentheses for ternary expressions with complex conditions.

app/serializers/v1/subscription_serializer.rb
```diff
@@ -12,7 +12,7 @@ def serialize
         plan_code: model.plan.code,
         status: model.status,
         billing_time: model.billing_time,
-        subscription_date: model.subscription_date&.iso8601,
+        subscription_date: model.subscription_at&.to_date&.iso8601,
```
lovrocolic, line 15: we are not exposing `subscription_at` here?

vincent-pochet: It will be, but in a next PR, this one is about changing the datatype internaly without breaking the current API/GraphQL

lovrocolic (approved):

PR #660, opened 2022-12-09
feat(timezones): Prevent overlaping of timezone when customer timezone changes

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR ensure that charges billed on a period does not overlap with a previous invoice or that no holes appears between 2 consecutive invoices in case of a change of time zone at customer or organization level.
Timezone will be stored at invoice level to ensure we are not loosing the timezone information

### DIFF
--- a/app/models/invoice_subscription.rb
+++ b/app/models/invoice_subscription.rb
@@ -13,17 +13,16 @@ class InvoiceSubscription < ApplicationRecord
   monetize :subscription_amount_cents, disable_validation: true, allow_nil: true
   monetize :total_amount_cents, disable_validation: true, allow_nil: true
 
-  scope :order_by_charges_to_datetime, lambda {
-    condition = ActiveRecord::Base.sanitize_sql_for_conditions(
-      <<-SQL
-        COALESCE(
-          (invoice_subscriptions.properties->>\'to_datetime\')::timestamp, invoice_subscriptions.created_at
-        ) ASC
-      SQL
-    )
+  scope :order_by_charges_to_datetime,
+        lambda {
+          condition = <<-SQL
+            COALESCE(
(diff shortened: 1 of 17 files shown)


app/models/invoice_subscription.rb
```diff
@@ -13,6 +13,18 @@ class InvoiceSubscription < ApplicationRecord
   monetize :subscription_amount_cents, disable_validation: true, allow_nil: true
   monetize :total_amount_cents, disable_validation: true, allow_nil: true
 
+  scope :order_by_charges_to_datetime, lambda {
```
jdenquin, line 16: Layout/MultilineMethodArgumentLineBreaks: Each argument in a multi-line method call must start on a separate line.

app/models/invoice_subscription.rb
```diff
@@ -13,6 +13,18 @@ class InvoiceSubscription < ApplicationRecord
...
+    condition = ActiveRecord::Base.sanitize_sql_for_conditions(
+      <<-SQL
```
jdenquin, line 18: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

app/models/invoice_subscription.rb
```diff
@@ -21,19 +32,19 @@ def fees
   end
 
   def from_datetime
-    fees_datetime('from_datetime')&.to_datetime
+    properties['from_datetime']&.to_datetime
```
lovrocolic, line 35: I'm not sure if I'm missing something, but I cannot see where does `properties` reference is coming from 🤔

lovrocolic (approved):
I believe you for the specs :D

PR #662, opened 2022-12-12
feat(timezones): Prevent double billing

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR ensure that a subscription cannot is not billed twice on its creation date or the billing date

### DIFF
--- a/app/jobs/bill_subscription_job.rb
+++ b/app/jobs/bill_subscription_job.rb
@@ -5,10 +5,11 @@ class BillSubscriptionJob < ApplicationJob
 
   retry_on Sequenced::SequenceError
 
-  def perform(subscriptions, timestamp)
-    result = Invoices::CreateService.new(
+  def perform(subscriptions, timestamp, invoice_source: :initial)
+    result = Invoices::SubscriptionService.new(
       subscriptions: subscriptions,
       timestamp: timestamp,
+      invoice_source: invoice_source,
     ).create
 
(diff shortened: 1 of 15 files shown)


app/jobs/bill_subscription_job.rb
```diff
@@ -5,10 +5,11 @@ class BillSubscriptionJob < ApplicationJob
 
   retry_on Sequenced::SequenceError
 
-  def perform(subscriptions, timestamp)
-    result = Invoices::CreateService.new(
+  def perform(subscriptions, timestamp, invoice_source: :initial)
+    result = Invoices::SubscriptionService.new(
       subscriptions: subscriptions,
       timestamp: timestamp,
+      invoice_source: invoice_source,
```
lovrocolic, line 12: Is this field there only for debugging purposes? I haven't seen that it's used actually...

lovrocolic (approved):

app/jobs/bill_subscription_job.rb
vincent-pochet: It's used only once [here](https://github.com/getlago/lago-api/pull/662/files#diff-c2fc569d4dc2c01c9cf24a63e8b53ba1b815903360e0fbdda0323fe4d9ada80cR170) and that's the key to make sure the subscription is not billed twice for the same period without impacting `terminate` invoices

PR #663, opened 2022-12-13
feat(credit-notes): Disable credit note creation until license management is ready

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR disable the manual credit note creation for API and GraphQL until we release the license management system

### DIFF
--- a/app/services/credit_notes/create_from_upgrade.rb
+++ b/app/services/credit_notes/create_from_upgrade.rb
@@ -27,6 +27,7 @@ module CreditNotes
           },
         ],
         reason: :order_change,
+        automatic: true,
       ).call
     end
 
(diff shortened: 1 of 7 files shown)


app/services/credit_notes/create_from_upgrade.rb
```diff
@@ -27,6 +27,7 @@ def call
           },
         ],
         reason: :order_change,
+        automatic: true
```
jdenquin, line 30: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

lovrocolic (approved):

PR #669, opened 2022-12-15
fix (coupons): validate expiration_at field on create and update

lovrocolic:
When creating/updating coupon from the API we need to validate expiration_at field. Currently, we don't validate it and coupon can be created even with `expiration_at` field set in the past

### DIFF
--- a/app/services/coupons/validate_service.rb
+++ b/app/services/coupons/validate_service.rb
@@ -26,7 +26,7 @@ module Coupons
 
       return true if expiration_at && expiration_at.to_date >= Time.current.to_date
 
-      add_error(field: :expiration_at, error_code: 'invalid_value')
+      add_error(field: :expiration_at, error_code: 'invalid_date')
 
       false
     end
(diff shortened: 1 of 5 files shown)


app/services/coupons/validate_service.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      expiration_at = if args[:expiration_at].is_a?(Time)
+        args[:expiration_at]
+      elsif args[:expiration_at].is_a?(String) && Date._strptime(args[:expiration_at]).present?
+        Date.strptime(args[:expiration_at])
+      end
+
+      return true if expiration_at && expiration_at.to_date >= Time.current.to_date
+
+      add_error(field: :expiration_at, error_code: 'invalid_value')
```
vincent-pochet, line 29: What about `invalid_date`? It will be a bit more consistent with subscription validation

vincent-pochet (approved):

app/services/coupons/validate_service.rb
lovrocolic: fixed

PR #670, opened 2022-12-16
fix(invoices): race conditions for wbehook

jdenquin:
## Context

- we had a race condition that was sending the webhook for a invoice before closing the transaction DB for it.

## Description

- Call the webhook job after the invoice DB transaction

### DIFF
--- a/app/services/invoices/create_service.rb
+++ b/app/services/invoices/create_service.rb
@@ -45,6 +45,7 @@ module Invoices
     private
 
     attr_accessor :subscriptions, :timestamp, :customer, :currency
+    delegate :invoice, to: :result
 
     def issuing_date
       Time.zone.at(timestamp).in_time_zone(customer.applicable_timezone).to_date
@@ -66,8 +67,4 @@ module Invoices
       )
     end
   end
-
-  private
-
-  delegate :invoice, to: :resul
 end
--- a/app/services/invoices/create_service.rb
+++ b/app/services/invoices/create_service.rb
@@ -12,8 +12,11 @@ module Invoices
     end
 
     def create
+      result = nil
+      invoice = nil
+
       ActiveRecord::Base.transaction do
-        result.invoice = Invoice.create!(
+        invoice = Invoice.create!(
           customer: customer,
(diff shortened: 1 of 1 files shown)


app/services/invoices/create_service.rb
```diff
@@ -32,12 +32,13 @@ def create
...
         ).call
 
-        SendWebhookJob.perform_later(:invoice, invoice) if should_deliver_webhook?
         Invoices::Payments::CreateService.new(invoice).call
         track_invoice_created(invoice)
 
         result
       end
+
+      SendWebhookJob.perform_later(:invoice, invoice) if should_deliver_webhook?
```
vincent-pochet, line 41: I think both `Invoices::Payments::CreateService.new(invoice).call` and `track_invoice_created(invoice)` are in the same situation

And we should keep returning the `result` :)

jdenquin: oh yes sorry for the `result`, I'll move the other 2 also

app/services/invoices/create_service.rb
```diff
@@ -12,6 +12,9 @@ def initialize(subscriptions:, timestamp:)
     end
 
     def create
+      invoice = nil
+      result = nil
```
vincent-pochet, line 16: Actually, self.result is already defined so, I suggest to only do:

```ruby
        result.invoice = Invoice.create!(
   # ...
```
and add a `delegate :invoice, to: :result` in the private section of the class

vincent-pochet (approved):
It works, but allow me to be "punctilious" :smile:

app/services/invoices/create_service.rb
jdenquin: that works for me!

app/services/invoices/create_service.rb
```diff
@@ -66,4 +66,8 @@ def track_invoice_created(invoice)
       )
     end
   end
+
+  private
```
jdenquin, line 70: Lint/UselessAccessModifier: Useless `private` access modifier.

PR #676, opened 2022-12-20
misc: Rename throw_error service method into raise_if_error

vincent-pochet:
## Context

Today result objects expose a `throw_error` that raise a `FailedResult` (or a subclass) if the result is not a success.

```ruby
  def throw_error
    return if success?

    raise(error)
  end
```

Since the naming of this method does not make obvious that it will not raise if the result is a success, the code end-up with things like:

```ruby
raise(result.throw_error) unless result.success?
```
It's seems a bit confusing :exploding_head: 

## Description

This PR simply renames `throw_error` into `raise_if_error!` and clean all the related code.

### DIFF
--- a/app/services/invoices/add_on_service.rb
+++ b/app/services/invoices/add_on_service.rb
@@ -61,8 +61,8 @@ module Invoices
 
     def create_add_on_fee(invoice)
       fee_result = Fees::AddOnService
-        .new(invoice: invoice, applied_add_on: applied_add_on).create
-      raise(fee_result.throw_error) unless fee_result.success?
+        .new(invoice:, applied_add_on:).create
+      fee_result.raise_if_error!
     end
 
     def should_deliver_webhook?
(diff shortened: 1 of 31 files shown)


app/services/invoices/add_on_service.rb
```diff
@@ -61,8 +61,8 @@ def compute_amounts(invoice)
 
     def create_add_on_fee(invoice)
       fee_result = Fees::AddOnService
-        .new(invoice: invoice, applied_add_on: applied_add_on).create
-      raise(fee_result.throw_error) unless fee_result.success?
+        .new(invoice:, applied_add_on:).create
```
vincent-pochet, line 64: New Ruby 3.1 syntax. Rubocop enforces it.

I don't know if we want to keep it like that, let me know :smile:

rsempe (approved):

PR #679, opened 2022-12-20
feat(grace_period): Ability to finalize draft invoices

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to be able to finalize draft invoices via API or graphQL.


app/services/invoices/finalize_service.rb
```diff
@@ -13,7 +13,7 @@ def initialize(invoice:)
 
     def call
       ActiveRecord::Base.transaction do
-        result = Invoices::RefreshDraftService.call(invoice: invoice)
+        result = Invoices::RefreshDraftService.call(invoice:)
```
lovrocolic, line 16: I think you miss a check if invoice is present (in the service level)

lovrocolic (approved):

app/services/invoices/finalize_service.rb
lovrocolic, line 16: IMO we should check the result from the sub-service here and return error result if any

PR #681, opened 2022-12-21
feat (invoices): add possibility to fetch and filter invoices

lovrocolic:
## Context

Provide possibility to fetch and filter invoices

## Description

Add filter by payment status in the GET /invoices endpoint
Add resolver to fetch invoices and filter them by payment status

### DIFF
--- a/app/controllers/api/v1/invoices_controller.rb
+++ b/app/controllers/api/v1/invoices_controller.rb
@@ -31,6 +31,11 @@ module Api
         if params[:external_customer_id]
           invoices = invoices.joins(:customer).where(customers: { external_id: params[:external_customer_id] })
         end
+
+        if valid_payment_status?(params[:payment_status])
+          invoices = invoices.where(payment_status: params[:payment_status])
+        end
+
         invoices = invoices.where(date_from_criteria) if valid_date?(params[:issuing_date_from])
         invoices = invoices.where(date_to_criteria) if valid_date?(params[:issuing_date_to])
         invoices = invoices.order(created_at: :desc)
@@ -88,6 +93,10 @@ module Api
 
       def date_to_criteria
         { issuing_date: ..Date.strptime(params[:issuing_date_to]) }
+      end
+
+      def valid_payment_status?(status)
+        Invoice.payment_statuses.key?(status)
       end
     end
   end
--- a/spec/graphql/resolvers/invoices_resolver_spec.rb
+++ b/spec/graphql/resolvers/invoices_resolver_spec.rb
@@ -49,12 +49,12 @@ RSpec.describe Resolvers::InvoicesResolver, type: :graphql do
   context 'when filtering by succeeded payment status' do
     let(:query) do
       <<~GQL
-      query {
-        invoices(limit: 5, paymentStatus: [succeeded]) {
-          collection { id }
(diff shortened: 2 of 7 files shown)


spec/graphql/resolvers/invoices_resolver_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      expect(invoices_response['metadata']['currentPage']).to eq(1)
+      expect(invoices_response['metadata']['totalCount']).to eq(2)
+    end
+  end
+
+  context 'when filtering by succeeded payment status' do
+    let(:query) do
+      <<~GQL
+      query {
```
jdenquin, line 52: Layout/HeredocIndentation: Use 2 spaces for indentation in a heredoc. (https://rubystyle.guide#squiggly-heredocs)

spec/requests/api/v1/invoices_spec.rb
```diff
@@ -134,5 +134,24 @@
...
+      let(:invoice2) { create(:invoice, customer: customer, payment_status: :failed) }
+      let(:invoice3) { create(:invoice, customer: customer, payment_status: :pending) }
+
+      before do
+        invoice2
+        invoice3
+      end
+
+      it 'returns invoices with correct payment status' do
+        get_with_token(organization, "/api/v1/invoices?payment_status=pending")
```
jdenquin, line 149: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols. (https://rubystyle.guide#consistent-string-literals)

rsempe (approved):

app/controllers/api/v1/invoices_controller.rb
```diff
@@ -31,6 +31,11 @@ def index
         if params[:external_customer_id]
           invoices = invoices.joins(:customer).where(customers: { external_id: params[:external_customer_id] })
         end
+
+        if valid_payment_status?(params[:payment_status])
```
vincent-pochet, line 35: Wondering if all this logic should be part of a query class:

```ruby
class BaseQuery < BaseService
  def initialize(organization)
    @organization = organization

    super
  end

  private

  attr_reader :organization
end

class InvoiceQuery < BaseQuery
  def call(filters)
    invoices = base_scope.xxx

    result.invoices = invoices
    result
  end

  private

  def base_scope
    organization.invoices
  end
end
```

That way you can even handle error with filters, or have shared methods for pagination

In the controller:

```ruby
def index
  query = InvoiceQuery.new(current_organization)
  result = query.call(params.permit(:external_customer_id, :payment_status))

  if result.success?
    render(
          invoices,
          ::V1::InvoiceSerializer,
          collection_name: 'invoices',
          meta: pagination_metadata(invoices),
        ),
      )
  else
    render_error_response(result)
  end
end
```

vincent-pochet (approved):
Looks good, but since we will have more and more filter logic in controller and in GraphQL resolvers, I wonder if we shouldn't find a way to share and delegate some logic.

Note: it could probably wait for a specific PR ;)

app/controllers/api/v1/invoices_controller.rb
lovrocolic: opened a card for it :)

PR #683, opened 2022-12-21
fix(currencies): Fix currency names

vincent-pochet:
## Context

Some currency symbol and name were mismatching. 

## Description

This PR fixes the currency symbol and name matching and also adds the `CLF` currency


rsempe (approved):
🎁

PR #687, opened 2022-12-26
feat(grace_period): Add nullable to invoice_grace_period

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to:
- add nullable to invoice grace period column on customers
- be able to create persisted events in a terminated subscription

### DIFF
--- a/app/models/customer.rb
+++ b/app/models/customer.rb
@@ -35,7 +35,7 @@ class Customer < ApplicationRecord
   validates :country, country_code: true, unless: -> { country.nil? }
   validates :currency, inclusion: { in: currency_list }, allow_nil: true
   validates :external_id, presence: true, uniqueness: { scope: :organization_id }
-  validates :invoice_grace_period, numericality: { greater_than_or_equal_to: 0 }
+  validates :invoice_grace_period, numericality: { greater_than_or_equal_to: 0 }, allow_nil: true
   validates :payment_provider, inclusion: { in: PAYMENT_PROVIDERS }, allow_nil: true
   validates :timezone, timezone: true, allow_nil: true
   validates :vat_rate, numericality: { less_than_or_equal_to: 100, greater_than_or_equal_to: 0 }, allow_nil: true
@@ -69,7 +69,7 @@ class Customer < ApplicationRecord
   end
 
   def applicable_invoice_grace_period
-    return invoice_grace_period if invoice_grace_period.positive?
+    return invoice_grace_period if invoice_grace_period.present?
 
     organization.invoice_grace_period
   end
--- a/app/services/events/validate_creation_service.rb
+++ b/app/services/events/validate_creation_service.rb
@@ -61,9 +61,8 @@ module Events
 
       return invalid_code_error unless valid_code?
 
-      invalid_persisted_event = persisted_event_validation(
-        customer.active_subscriptions.first || organization.subscriptions.find_by(id: params[:subscription_id]),
-      )
+      subscription = organization.subscriptions.find_by(external_id: params[:external_subscription_id])
+      invalid_persisted_event = persisted_event_validation(subscription || customer.active_subscriptions.first)
       return invalid_persisted_event_error(invalid_persisted_event) if invalid_persisted_event.present?
     end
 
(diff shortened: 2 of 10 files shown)


spec/services/events/validate_creation_service_spec.rb
```diff
@@ -227,6 +227,36 @@
...
+            code: billable_metric.code,
+            properties: {
+              billable_metric.field_name => 'ext_1234',
+              'operation_type' => 'add',
+            },
+          }
+        end
+
+        it 'returns no error' do
+
```
jdenquin, line 254: Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body beginning. (https://rubystyle.guide#empty-lines-around-bodies)

app/models/customer.rb
```diff
@@ -69,7 +69,7 @@ def applicable_timezone
   end
 
   def applicable_invoice_grace_period
-    return invoice_grace_period if invoice_grace_period.positive?
+    return invoice_grace_period if invoice_grace_period.present?
```
jdenquin, line 72: we do not need the `positive` test anymore?

app/services/events/validate_creation_service.rb
```diff
@@ -62,7 +62,9 @@ def validate_create
       return invalid_code_error unless valid_code?
 
       invalid_persisted_event = persisted_event_validation(
-        customer.active_subscriptions.first || organization.subscriptions.find_by(id: params[:subscription_id]),
+        organization.subscriptions.find_by(
```
jdenquin, line 65: this is strange by reading it, could we improve it?

spec/models/customer_spec.rb
```diff
@@ -118,7 +118,7 @@
       let(:organization_invoice_grace_period) { 5 }
 
       before do
-        customer.invoice_grace_period = 0
+        customer.invoice_grace_period = nil
```
jdenquin, line 121: if it's null by default, no need to initialize it?

spec/services/events/validate_creation_service_spec.rb
```diff
@@ -227,6 +227,35 @@
...
+        let(:billable_metric) do
+          create(
+            :billable_metric,
+            organization:,
+            aggregation_type: 'recurring_count_agg',
+            field_name: 'item_id',
+          )
+        end
+
+        let(:subscription) { create(:subscription, customer:, organization:, status: :terminated) }
```
jdenquin, line 241: can we put the one line `let` before the blocks?

jdenquin (changes requested):

app/models/customer.rb
rsempe: nope - we just have to check if `invoice_grace_period` is set or not now
previously, we were checking because by default, value was equal to zero

spec/services/events/validate_creation_service_spec.rb
rsempe: why this convention? 😅 - I personally prefer to put them to be coherent when it's used (here for instance, subscription is called when we're using the initializing `params`)

app/services/events/validate_creation_service.rb
rsempe: why strange?
`has_many :subscriptions, through: :customers`

spec/models/customer_spec.rb
rsempe: nope, here `invoice_grace_period` is equal to `3`, see [here](https://github.com/getlago/lago-api/pull/687/files/090c606ccc60595fe1cb2cf146d97035d76ca223#diff-169a4b470e704cdc458507df849294eaa7005d1db666def55bb70f25f5d3a02cL110)

app/services/events/validate_creation_service.rb
jdenquin: not the find_by but we use the result directly into a method call, would be better to put it into a var no?

spec/services/events/validate_creation_service_spec.rb
jdenquin: it's only for readability 😂

app/services/events/validate_creation_service.rb
rsempe: Agreed - Updated

jdenquin (approved):

PR #689, opened 2022-12-27
fix (webhooks): change webhook public key endpoint reponse type

lovrocolic:
## Context

In order to be consistent in API response types, this PR changes response type for endpoint to generate webhook public key

## Description

`api/v1/webhooks/public_key` returned text/plain in the past. It's changed now to application/json

### DIFF
--- a/app/controllers/api/v1/webhooks_controller.rb
+++ b/app/controllers/api/v1/webhooks_controller.rb
@@ -3,7 +3,12 @@
 module Api
   module V1
     class WebhooksController < Api::BaseController
+      # Deprecated method
       def public_key
+        render(plain: Base64.encode64(RsaPublicKey.to_s))
+      end
+
+      def json_public_key
         render(
           json: {
             webhook: {
--- a/app/controllers/api/v1/webhooks_controller.rb
+++ b/app/controllers/api/v1/webhooks_controller.rb
@@ -4,7 +4,14 @@ module Api
   module V1
     class WebhooksController < Api::BaseController
       def public_key
-        render(plain: Base64.encode64(RsaPublicKey.to_s))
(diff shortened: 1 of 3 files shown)


vincent-pochet (approved):
Looks good, but it's a breaking change :boom: 
How do we want to handle this?

lovrocolic:
> Looks good, but it's a breaking change 💥 How do we want to handle this?

Yep, that's why we haven't merged it yet. This one is not an urgent one and secret key is not fetched so often so we should probably be good here... However, it would be good if we can notice users about it... @mathieu-lago Can you maybe help here with notifying users? 🙏

vincent-pochet:
Hey @lovrocolic! I had a discussion with @mathieu-lago and @jdenquin and we think it's better not to add a new breaking change.

To prevent it, we suggest to add a new route `GET /api/v1/webhooks/json_public_key` which will return the new response format and to keep the old one untouched.
We should rely on the json one in the clients and in the doc and keep the old one undocumented and marked as deprecated in the code (and in the coming release note)

Let me know if you need more info

lovrocolic:
> Hey @lovrocolic! I had a discussion with @mathieu-lago and @jdenquin and we think it's better not to add a new breaking change.
> 
> To prevent it, we suggest to add a new route `GET /api/v1/webhooks/json_public_key` which will return the new response format and to keep the old one untouched. We should rely on the json one in the clients and in the doc and keep the old one undocumented and marked as deprecated in the code (and in the coming release note)
> 
> Let me know if you need more info

Hey @vincent-pochet . Make sense to me! Just updated it!

vincent-pochet (approved):
:rocket:

PR #693, opened 2022-12-28
misc(license): Add license support

jdenquin:
## Description

- Add License Support

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -170,8 +170,8 @@ ActiveRecord::Schema[7.0].define(version: 2022_12_19_111209) do
     t.string "refund_vat_amount_currency"
     t.bigint "vat_amount_cents", default: 0, null: false
     t.string "vat_amount_currency"
-    t.date "issuing_date", null: false
     t.datetime "refunded_at"
+    t.date "issuing_date", null: false
     t.index ["customer_id"], name: "index_credit_notes_on_customer_id"
     t.index ["invoice_id"], name: "index_credit_notes_on_invoice_id"
   end
@@ -331,9 +331,9 @@ ActiveRecord::Schema[7.0].define(version: 2022_12_19_111209) do
     t.uuid "customer_id"
     t.boolean "legacy", default: false, null: false
     t.float "vat_rate"
-    t.integer "status", default: 0, null: false
     t.bigint "credit_amount_cents", default: 0, null: false
     t.string "credit_amount_currency"
+    t.integer "status", default: 0, null: false
     t.string "timezone", default: "UTC", null: false
     t.index ["customer_id"], name: "index_invoices_on_customer_id"
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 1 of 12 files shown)


db/schema.rb
```diff
@@ -331,9 +331,9 @@
     t.uuid "customer_id"
     t.boolean "legacy", default: false, null: false
     t.float "vat_rate"
-    t.integer "status", default: 0, null: false
     t.bigint "credit_amount_cents", default: 0, null: false
     t.string "credit_amount_currency"
+    t.integer "status", default: 0, null: false
```
rsempe, line 336: You can rebuild your db to not have this change ;)

jdenquin: ah yes wanted to remove the `schema.rb` from this PR scope

spec/lib/lago_utils/license_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+RSpec.describe LagoUtils::License do
+  subject(:license) { described_class.new(url) }
+
+  let(:url) { 'https://license.lago' }
+
+  before do
+    ENV['LAGO_LICENSE'] = 'test-license'
+  end
+
+  describe '.verify' do
```
rsempe, line 14:
```suggestion
  describe '#verify' do
```

rsempe, line 14:
```suggestion
  describe '#verify' do
```

spec/lib/lago_utils/license_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      it 'sets premium to true' do
+        license.verify
+
+        expect(license).to be_premium
+      end
+    end
+
+    context 'when license is not present' do
+      before do
+        ENV['LAGO_LICENSE'] = nil
```
rsempe, line 36: Are you sure you don't have to set it again after the test?

rsempe (approved):

spec/lib/lago_utils/license_spec.rb
jdenquin: since it only runs for test I pretty sure yes!

lib/lago_utils/lago_utils/license.rb
```diff
@@ -0,0 +1,27 @@
...
+    def initialize(url)
+      @url = url
+      @premium = false
+    end
+
+    def verify
+      return if ENV['LAGO_LICENSE'].blank?
+
+      http_client = LagoHttpClient::Client.new("#{url}/verify/#{ENV['LAGO_LICENSE']}")
+      response = http_client.post({}, [])
```
lovrocolic, line 14: TBH I'm not sure why we don't use HTTP GET here since we just fetch some informations here

jdenquin: that designed in the license service this way, but we can update it with a `GET`, wdyt @rsempe ?

rsempe: Yep, `GET` makes more sense to me as well

jdenquin: let's go for a GET so!

PR #694, opened 2022-12-28
misc: Use Events::ValidateCreationService for validating event params

rsempe:
The goal of this PR is to improve the validations of parameters when creating an event.

### DIFF
--- a/app/services/events/create_service.rb
+++ b/app/services/events/create_service.rb
@@ -2,34 +2,19 @@
 
 module Events
   class CreateService < BaseService
-    ALL_REQUIRED_PARAMS = %i[transaction_id code].freeze
-    ONE_REQUIRED_PARAMS = %i[external_subscription_id external_customer_id].freeze
-
-    def validate_params(params:)
-      params_errors = ALL_REQUIRED_PARAMS.each_with_object({}) do |key, errors|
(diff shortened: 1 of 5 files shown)


app/services/events/validate_creation_service.rb
```diff
@@ -6,12 +6,13 @@ def self.call(...)
       new(...).call
     end
 
-    def initialize(organization:, params:, result:, customer:, batch: false)
+    def initialize(organization:, params:, result:, customer:, batch: false, send_webhook: true)
```
jdenquin, line 9: Metrics/ParameterLists: Avoid parameter lists longer than 5 parameters. [6/5] (https://rubystyle.guide#too-many-params)

app/services/events/create_service.rb
```diff
@@ -77,6 +62,16 @@ def call(organization:, params:, timestamp:, metadata:)
       result
     end
 
+    private
+
+    def customer(organization:, params:)
```
lovrocolic, line 67: I started using e.g. `return @customer if defined? @customer` as a first line so that we also memoize nil values and don't do query each time if customer is not found

rsempe: Good spot 👍

lovrocolic (approved):

PR #698, opened 2022-12-29
fix (events): add validation for invalid event field_name value

lovrocolic:
For billable metric that is `sum_agg` or `max_agg` we need to check if event's field_name value is integer/float.
For invalid values, validation message should be raised and event should not be persisted

### DIFF
--- a/spec/services/events/validate_creation_service_spec.rb
+++ b/spec/services/events/validate_creation_service_spec.rb
@@ -179,6 +179,34 @@ RSpec.describe Events::ValidateCreationService, type: :service do
         end
       end
 
+      context 'when field_name value is not a number' do
+        let(:billable_metric) { create(:sum_billable_metric, organization: organization) }
+        let(:params) do
+          {
+            code: billable_metric.code,
+            external_customer_id: customer.external_id,
+            properties: {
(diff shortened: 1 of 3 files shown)


spec/services/events/validate_creation_service_spec.rb
```diff
@@ -179,6 +179,34 @@
         end
       end
 
+      context 'when field_name value is not a number' do
+        let(:billable_metric) { create(:sum_billable_metric, organization: organization) }
```
jdenquin, line 183: Style/HashSyntax: Omit the hash value. (https://rubystyle.guide#hash-literals)

spec/services/events/validate_creation_service_spec.rb
```diff
@@ -179,6 +179,34 @@
...
+        let(:params) do
+          {
+            code: billable_metric.code,
+            external_customer_id: customer.external_id,
+            properties: {
+              item_id: 'test'
```
jdenquin, line 189: Style/TrailingCommaInHashLiteral: Put a comma after the last item of a multiline hash.

rsempe (approved):

PR #710, opened 2023-01-03
feat (invoices): add issued_at attribute in graphQL response

lovrocolic:
## Context

Add `issued_at` attribute in graphQL response

## Description

UI needs information about timezone so `issuing_date` attribute is not suitable for this case.

This PR adds new attribute called `issued_at` and this attribute is returned in customer timezone. It will be consumed only from the UI

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -120,6 +120,14 @@ class Invoice < ApplicationRecord
     amount + wallet_transaction_amount
   end
 
+  def issued_at
+    issuing_time = created_at.in_time_zone(customer.applicable_timezone)
+
+    return issuing_time unless grace_period?
+
+    issuing_time + customer.applicable_invoice_grace_period.days
+  end
+
   def organization
     customer&.organization
   end
@@ -173,5 +181,9 @@ class Invoice < ApplicationRecord
     formatted_sequential_id = format('%03d', sequential_id)
 
     self.number = "#{customer.slug}-#{formatted_sequential_id}"
+  end
+
+  def grace_period?
+    customer.applicable_invoice_grace_period.positive?
   end
 end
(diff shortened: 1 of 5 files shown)


spec/models/invoice_spec.rb
```diff
@@ -188,6 +188,27 @@
     end
   end
 
+  describe '#issued_at' do
+    let(:customer) { create(:customer, timezone: 'America/Los_Angeles') }
+    let(:invoice) { create(:invoice, customer: customer, created_at: DateTime.parse('2022-11-17 23:34:23')) }
```
jdenquin, line 193: Style/HashSyntax: Omit the hash value. (https://rubystyle.guide#hash-literals)

app/models/invoice.rb
```diff
@@ -120,6 +120,14 @@ def subtotal_before_prepaid_credits
     amount + wallet_transaction_amount
   end
 
+  def issued_at
+    issuing_time = created_at.in_time_zone(customer.applicable_timezone)
+
+    return issuing_time unless grace_period?
+
+    issuing_time + customer.applicable_invoice_grace_period.days
```
rsempe, line 128: This won't work if a customer manually finalizes the invoice before the end of the grace period.
In that case, issuing time should not be equal to this calculation

app/models/invoice.rb
```diff
@@ -120,6 +120,14 @@ def subtotal_before_prepaid_credits
     amount + wallet_transaction_amount
   end
 
+  def issued_at
+    issuing_time = created_at.in_time_zone(customer.applicable_timezone)
```
vincent-pochet, line 124: We should not expose values in customer timezone, all datetime should be exposed as UTC. It's the front-end responsibility to do the conversion into the right timezone.

Invoices have a `timezone` attribute which can be exposed in the GraphQL type to help the front-end with this conversion.

vincent-pochet (commented):
Not sure of the purpose of this field.
Today, `issuing_date` is a "value date" in customer timezone, so I'm not sure what we are trying to solve here.
Happy to discuss it :)

lovrocolic:
@vincent-pochet @rsempe We won't be adding this change :) Closing PR

PR #720, opened 2023-01-06
feat: Expose amount_cents fields as bigint in GraphQL

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/719.

## Description

The PR intends to expose all `*AmountCents` fields as bigint rather than integer to fix issue with large amount.
It will requires the front-end to adapt the way it consumes and pushes amount fields to the backend


app/graphql/mutations/add_ons/create.rb
```diff
@@ -12,7 +12,7 @@ class Create < BaseMutation
       argument :name, String, required: true
       argument :code, String, required: true
       argument :description, String, required: false
-      argument :amount_cents, Integer, required: true
+      argument :amount_cents, GraphQL::Types::BigInt , required: true
```
jdenquin, line 15: Layout/SpaceBeforeComma: Space found before comma.

rsempe (approved):

PR #721, opened 2023-01-06
feat (search): add search capability for billable metrics

lovrocolic:
## Context

Add search capability from the UI for billable metrics


jdenquin:
This should be compared to main only, the last branch will be too big

ansmonjol (approved):

PR #730, opened 2023-01-10
feat(credit_notes): add search capability to GraphQl resolver

ansmonjol:
## Roadmap Task

👉  https://github.com/getlago/lago/issues/156

## Context

It’s hard for our user to find a specific objects when there’s more than 10 objects in a list view.

## Description

This PR adds the ability to search on a specific search query agains CreditNotes `number` and `id`

Note that Ransack needs the ID value to be a varchar in order to perform a `_cont` on it.
This PR adds an helper to transform the ID type during ransack search

### DIFF
--- a/spec/queries/customer_credit_notes_query_spec.rb
+++ b/spec/queries/customer_credit_notes_query_spec.rb
@@ -1,0 +1,145 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe CustomerCreditNotesQuery, type: :query do
+  subject(:customer_credit_notes_query) do
+    described_class.new(organization:)
+  end
+
+  let(:membership) { create(:membership) }
+  let(:organization) { membership.organization }
+  let(:customer) { create(:customer, organization:) }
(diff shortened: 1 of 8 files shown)


spec/queries/customer_credit_notes_query_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+        search_term: 'imthe',
+        customer_id: customer.id,
+        page: 1,
+        limit: 10,
+      )
+
+      returned_ids = result.credit_notes.pluck(:id)
+
+      aggregate_failures do
+        expect(result.credit_notes.count).to eq(3)
```
vincent-pochet, line 56: It looks like we are returning 3 credit notes instead of 1 :)

vincent-pochet (approved):
One small comment in the spec, looks good otherwise :rocket:

spec/queries/customer_credit_notes_query_spec.rb
ansmonjol: Good catch ! I updated this test to search for a patttern that return 2 items that was not returned in other tests.

Udpated

jdenquin (approved):

PR #735, opened 2023-01-11
feat(grace_period): Finalize draft invoices when grace period is zero

rsempe:


vincent-pochet (approved):
The funny one! :smile:

PR #736, opened 2023-01-11
feat(customer_invoices): add search capability to GraphQl resolver

ansmonjol:
## Roadmap Task

👉  https://github.com/getlago/lago/issues/156

## Context

It’s hard for our user to find a specific objects when there’s more than 10 objects in a list view.

## Description

This PR adds the ability to search for customer invoices on 2 attributes

- id
- number


app/queries/customer_invoices_query.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class CustomerInvoicesQuery < BaseQuery
```
vincent-pochet, line 3: Is there a reason to add a specific `CustomerInvoicesQuery`?

It seems we could use the `InvoicesQuery` by adapting the possible filters/query arguments.
What do you think?

ansmonjol: hum yeah we could, just wondering about if the customer_id is passed or not, should we filter on customer ID or find customer.invoices as we do today.

Wondering if there is perf issue with the first approach. Let's discuss about it today

ansmonjol: Updated the code to have one query used for both  invoice resolvers

app/queries/invoices_query.rb
```diff
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 class InvoicesQuery < BaseQuery
-  def call(search_term:, status:, payment_status:, page:, limit:, filters: {}) # rubocop:disable Metrics/ParameterLists
+  def call(search_term:, status:, page:, limit:, filters: {}, customer_id: nil, payment_status: nil) # rubocop:disable Metrics/ParameterLists
     @search_term = search_term
 
-    invoices = base_scope.result.includes(:customer)
+    invoices = base_scope(customer_id:).result.includes(:customer)
```
vincent-pochet, line 7: You could store the `customer_id` into an instance variable `@customer_id` and set it as `attr_reader`. This way you will not have to cascade it to the `search_params` method

app/queries/invoices_query.rb
```diff
@@ -18,20 +19,22 @@ def call(search_term:, status:, payment_status:, page:, limit:, filters: {}) # r
...
-    }
+    {}.tap do |h|
+      h[:m] = 'or'
+      h[:id_cont] = search_term
+      h[:number_cont] = search_term
+      unless customer_id
+        h[:customer_name_cont] = search_term
+        h[:customer_external_id_cont] = search_term
+        h[:customer_email_cont] = search_term
+      end
```
vincent-pochet, line 37: This approach feels a bit more rubist ;) :

```suggestion
    terms = {
      m: 'or',
      id_cont: search_term,
      number_cont: search_term,
    }
    return terms if customer_id.present?
    
    terms.merge(
     customer_name_cont: search_term,
     customer_external_id_cont: search_term,
     customer_email_cont: search_term,
    )
```

spec/queries/customer_invoices_query_spec.rb
```diff
@@ -0,0 +1,160 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe InvoicesQuery, type: :query do
```
vincent-pochet, line 5: The name of the query should match the name of the spec file. Could you also merge this test case into the existing `spec/queries/invoice_query_spec.rb`?

vincent-pochet (commented):
Few comments, but it looks good!

spec/queries/customer_invoices_query_spec.rb
ansmonjol: Good point, done !

app/queries/invoices_query.rb
ansmonjol: code updated

app/queries/invoices_query.rb
ansmonjol: done, using `@customer_id` now

vincent-pochet (approved):

PR #738, opened 2023-01-12
fix(credit-notes): Fix credit note amount for terminate with previous credit note

vincent-pochet:
## Context

We have an issue with the subscription termination in the following case:

1. I apply a pay in advance plan to a customer (subscription A)
2. Invoice is finalized and I applied a credit note covering the 100% of invoice
3. I decided to terminate the subscription A
4. I’ve this error, the terminate does not lead to an invoice creation

It fails with a validation failure:

```
Validation errors: {:amount_cents=>["higher_than_remaining_fee_amount", "higher_than_remaining_invoice_amount"]}
```
The reason is that when terminating the subscription, we try to emit a credit note for the overpaid amount (at prorata of the period duration) targeting the subscription fee of the last invoice. Since this fee has been already re-credited, the validation fails as there is no remaining creditable amount.

## Description

The PR fixes the issue by taking the max creditable amount on the last subscription fee if the pro-rata amount is higher than the remaining amount.

NOTE: this amount could be 0, in that case, no credit note will be emitted.

### DIFF
--- a/app/services/credit_notes/create_from_termination.rb
+++ b/app/services/credit_notes/create_from_termination.rb
@@ -16,9 +16,9 @@ module CreditNotes
       return result unless amount.positive?
 
       # NOTE: if credit notes were already issued on the fee,
-      #       we take the remaining creditable amount
-      amount = last_subscription_fee.creditable_amount_cents if amount > last_subscription_fee.creditable_amount_cents
-      return result if amount.zero?
+      #       we have to deduct them from the prorated amount
+      amount -= last_subscription_fee.credit_note_items.sum(:amount_cents)
+      return result unless amount.positive?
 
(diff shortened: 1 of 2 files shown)


jdenquin (approved):

config/initializers/license.rb
```diff
@@ -5,3 +5,7 @@
 License = LagoUtils::License.new(Rails.application.config.license_url)
 
 License.verify unless Rails.env.test?
+
+if Rails.env.development?
```
jdenquin, line 9: Style/IfUnlessModifier: Favor modifier `if` usage when having a single-line body. Another good alternative is the usage of control flow `&&`/`||`. (https://rubystyle.guide#if-as-a-modifier)

PR #741, opened 2023-01-18
feat(delete_bm): Add deleted_at and setup discard for soft-delete

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to:
- Add `deleted_at` to all affected tables
- Setup discard gem with `deleted_at` column and default scope `kept`
- Ensure code uniqueness validation does not take into account deleted billable metrics

### DIFF
--- a/db/migrate/20230118100324_add_deleted_at_to_billable_metrics.rb
+++ b/db/migrate/20230118100324_add_deleted_at_to_billable_metrics.rb
@@ -15,5 +15,8 @@ class AddDeletedAtToBillableMetrics < ActiveRecord::Migration[7.0]
     add_index :group_properties, :deleted_at
     add_index :events, :deleted_at
     add_index :persisted_events, :deleted_at
+
+    remove_index :billable_metrics, %i[organization_id code]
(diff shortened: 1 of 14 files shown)


db/migrate/20230118100324_add_deleted_at_to_billable_metrics.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class AddDeletedAtToBillableMetrics < ActiveRecord::Migration[7.0]
+  def change
+    add_column :billable_metrics, :deleted_at, :datetime
```
jdenquin, line 5: `index: true` does not work here?

jdenquin (approved):

vincent-pochet (approved):
:rocket:

db/migrate/20230118100324_add_deleted_at_to_billable_metrics.rb
rsempe: Nope - https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/add_column

PR #747, opened 2023-01-19
feat(delete_bm): Expose deleted resources in PDF invoice

vincent-pochet:
## Roadmap Task

## Context

## Description

The goal of this PR is to ensure deleted resource remains visible on PDF invoice

### DIFF
--- a/spec/factories/utils.rb
+++ b/spec/factories/utils.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  trait :discarded do
+    deleted_at { Time.current }
+  end
+end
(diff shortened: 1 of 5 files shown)


spec/factories/utils.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  trait :discarded do
```
rsempe, line 4: what about renaming this to `deleted`?  😬

rsempe (approved):

spec/factories/utils.rb
vincent-pochet: well, sure :)

PR #749, opened 2023-01-20
feat(delete_bm): Ensure deleted resources are exposed in Invoice API endpoint

vincent-pochet:
## Roadmap Task

## Context

## Description

The goal of this PR is to ensure deleted resource attached to a fee remain visible in invoice API endpoint

### DIFF
--- a/app/serializers/v1/fee_serializer.rb
+++ b/app/serializers/v1/fee_serializer.rb
@@ -5,7 +5,7 @@ module V1
     def serialize
       {
         lago_id: model.id,
-        lago_group_id: model.group&.id,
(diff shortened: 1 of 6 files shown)


app/serializers/v1/fee_serializer.rb
```diff
@@ -5,7 +5,7 @@ class FeeSerializer < ModelSerializer
     def serialize
       {
         lago_id: model.id,
-        lago_group_id: model.group&.id,
+        lago_group_id: model.group_id,
```
rsempe, line 8: 😍

rsempe (approved):

PR #750, opened 2023-01-20
feat(delete_bm): Add activeSubscriptionsCount and draftInvoicesCount

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to add on the billable metric object:
- `activeSubscriptionsCount`
- `draftInvoicesCount`

These fields are used to know the impact of deleting billable metric.

### DIFF
--- a/app/graphql/types/billable_metrics/object.rb
+++ b/app/graphql/types/billable_metrics/object.rb
@@ -34,7 +34,12 @@ module Types
       end
 
       def draft_invoices_count
-        object.plans.joins(subscriptions: [:invoices]).merge(Invoice.draft).count
+        object.plans
+          .joins(subscriptions: [:invoices])
+          .merge(Invoice.draft)
+          .select(:invoice_id)
+          .distinct
+          .count
       end
     end
   end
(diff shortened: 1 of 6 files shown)


app/graphql/types/billable_metrics/object.rb
```diff
@@ -25,6 +27,14 @@ def group
       def flat_groups
         object.selectable_groups
       end
+
+      def active_subscriptions_count
+        object.plans.joins(:subscriptions).merge(Subscription.active).count
+      end
+
+      def draft_invoices_count
+        object.plans.joins(subscriptions: [:invoices]).merge(Invoice.draft).count
```
vincent-pochet, line 36: Since we are going through subscription to get invoices don't we have a risk to get duplication in the count if an invoice is linked to multiple subscription for the same plan?

rsempe: Oh seems like a nice catch! 👏  Looking into this!

vincent-pochet (approved):

PR #756, opened 2023-01-20
feat(delete_bm): Discard metric and associations

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to:
- Refactor bm destroy service
- Discard metric and associations (charges, groups, group properties, events and persisted events)

### DIFF
--- a/app/services/billable_metrics/destroy_service.rb
+++ b/app/services/billable_metrics/destroy_service.rb
@@ -14,7 +14,18 @@ module BillableMetrics
     def call
       return result.not_found_failure!(resource: 'billable_metric') unless metric
 
-      metric.destroy!
+      ActiveRecord::Base.transaction do
+        metric.discard!
+        metric.charges.discard_all
+        metric.groups.each do |group|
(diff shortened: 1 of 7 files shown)


app/services/billable_metrics/destroy_service.rb
```diff
@@ -2,24 +2,36 @@
...
     end
 
-    def destroy_from_api(organization:, code:)
-      metric = organization.billable_metrics.find_by(code:)
+    def call
       return result.not_found_failure!(resource: 'billable_metric') unless metric
 
-      metric.destroy!
+      ActiveRecord::Base.transaction do
+        metric.discard
```
vincent-pochet, line 18: Should we use `discard!` to ensure the action is executed?

vincent-pochet (approved):
:rocket:

app/services/billable_metrics/destroy_service.rb
rsempe: good idea!

PR #757, opened 2023-01-20
feat(delete_bm): Track metric and refresh draft invoices

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to:
- Track metric deletion on Segment
- Refresh linked draft invoices

### DIFF
--- a/app/services/billable_metrics/destroy_service.rb
+++ b/app/services/billable_metrics/destroy_service.rb
@@ -14,6 +14,9 @@ module BillableMetrics
     def call
       return result.not_found_failure!(resource: 'billable_metric') unless metric
 
+      draft_invoice_ids = Invoice.draft.joins(plans: [:billable_metrics])
+        .where(billable_metrics: { id: metric.id }).distinct.pluck(:id)
+
       ActiveRecord::Base.transaction do
         metric.discard!
         metric.charges.discard_all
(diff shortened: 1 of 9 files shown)


app/services/billable_metrics/destroy_service.rb
```diff
@@ -14,6 +14,9 @@ def initialize(metric:)
     def call
       return result.not_found_failure!(resource: 'billable_metric') unless metric
 
+      draft_invoice_ids = Invoice.draft.joins(plans: [:billable_metrics])
+        .where(billable_metrics: { id: metric.id }).distinct.pluck(:id)
```
rsempe, line 18: This, to avoid adding `with_deleted` in the has_many associations

app/services/billable_metrics/destroy_service.rb
```diff
@@ -26,12 +29,34 @@ def call
       # NOTE: Discard all related events asynchronously.
       BillableMetrics::DeleteEventsJob.perform_later(metric)
 
+      track_billable_metric_deleted
+
+      # NOTE: Refresh all invoices linked to the billable metric.
+      Invoice.find(draft_invoice_ids).each do |invoice|
+        ::Invoices::RefreshDraftService.call(invoice:)
```
vincent-pochet, line 36: Shouldn't we execute it in job? We might have a lot of draft invoice with the metric and so it could take a while to refresh all of them. What do you think?

vincent-pochet (approved):

app/services/billable_metrics/destroy_service.rb
rsempe: Yeah - I agree - Good point 👍

PR #766, opened 2023-01-26
feat(delete_plan): Add subscription and invoice counters to plans

vincent-pochet:
## Context

Currently, there is no way to delete objects linked to a subscription. This can be a real limitation during PoCs and onboarding, as users have to ping us to delete plans.

The goal of this feature is to be able to soft-delete a plan linked to an active or terminated subscription.

## Description

The goal of this PR is to add two counters in the `V1::PlanSerializer` and `Plan` GraphQL type:
- `active_subscription_count`
- `draft_invoice_count`

### DIFF
--- a/app/graphql/types/plans/object.rb
+++ b/app/graphql/types/plans/object.rb
@@ -26,6 +26,8 @@ module Types
 
       field :charge_count, Integer, null: false, description: 'Number of charges attached to a plan'
       field :customer_count, Integer, null: false, description: 'Number of customers attached to a plan'
+      field :active_subscriptions_count, Integer, null: false
+      field :draft_invoices_count, Integer, null: false
 
       field :can_be_deleted, Boolean, null: false do
         description 'Check if plan is deletable'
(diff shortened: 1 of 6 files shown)


app/graphql/types/plans/object.rb
```diff
@@ -42,6 +44,18 @@ def customer_count
       def can_be_deleted
         object.deletable?
       end
+
+      def active_subscriptions_count
+        object.subscriptions.merge(Subscription.active).count
```
rsempe, line 49:
```suggestion
        object.subscriptions.active.count
```

app/serializers/v1/plan_serializer.rb
```diff
@@ -27,5 +29,18 @@ def serialize
     def charges
       ::CollectionSerializer.new(model.charges, ::V1::ChargeSerializer, collection_name: 'charges').serialize
     end
+
+    def active_subscriptions_count
+      model.subscriptions.merge(Subscription.active).count
```
rsempe, line 34:
```suggestion
      model.subscriptions.active.count
```

rsempe (approved):

app/graphql/types/plans/object.rb
vincent-pochet: :man_facepalming: indeed! Good catch :smile:

PR #775, opened 2023-01-27
feat (coupons): coupon overlapping plans validation

lovrocolic:
## Context

It’s impossible to assign a coupon only for some dedicated plans. It means that coupons is deducted from total invoice no matter the plans.

## Description

This PR adds validation that prevent adding plan limitation only if there is plan limitation already added on previous coupon (that is applied on customer)

### DIFF
--- a/app/services/applied_coupons/create_service.rb
+++ b/app/services/applied_coupons/create_service.rb
@@ -49,6 +49,7 @@ module AppliedCoupons
     def check_preconditions
       return result.not_found_failure!(resource: 'customer') unless customer
       return result.not_found_failure!(resource: 'coupon') unless coupon
+      return result.not_allowed_failure!(code: 'plan_overlapping') if plan_limitation_overlapping?
       return if reusable_coupon?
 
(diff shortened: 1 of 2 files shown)


app/services/applied_coupons/create_service.rb
```diff
@@ -96,6 +97,16 @@ def reusable_coupon?
...
     end
 
+    def plan_limitation_overlapping?
+      return false unless coupon.limited_plans?
+
+      related_plans = coupon.coupon_plans.pluck(:plan_id)
+
+      active_coupons_ids = customer.applied_coupons.active.pluck(:coupon_id)
+
+      CouponPlan.where(coupon_id: active_coupons_ids, plan_id: related_plans).exists?
```
vincent-pochet, line 107: I think you can achieve the same think with one query instead of 3: (To be validated ;) )
```ruby
customer.applied_coupons
  .active
  .joins(coupon: :coupon_plans)
  .where(coupon_plans: { plan_id: coupon.coupon_plans.select(:plan_id) })
  .exist?
```

vincent-pochet (approved):
One comment to optimize the number of DB query, otherwise, looks good!

PR #782, opened 2023-01-31
feat(delete_add_on): Add ability to soft delete an add_on

vincent-pochet:
## Context

## Description

To allow soft deletion of customers, we first need to be able to soft delete add_ons and coupons

The goal of this PR adds a `deleted_at` field to the `add_ons` table and the Discard logic to allow soft deletion

### DIFF
--- a/app/controllers/api/v1/add_ons_controller.rb
+++ b/app/controllers/api/v1/add_ons_controller.rb
@@ -35,11 +35,8 @@ module Api
       end
 
       def destroy
-        service = AddOns::DestroyService.new
-        result = service.destroy_from_api(
-          organization: current_organization,
-          code: params[:code],
-        )
+        add_on = current_organization.add_ons.find_by(code: params[:code])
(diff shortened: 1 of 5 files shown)


app/models/add_on.rb
```diff
@@ -12,11 +14,14 @@ class AddOn < ApplicationRecord
   monetize :amount_cents
 
   validates :name, presence: true
-  validates :code, uniqueness: { scope: :organization_id, allow_nil: false }
+  validates :code,
```
rsempe, line 17: Maybe we can add an example as we did on the others, no?
https://github.com/getlago/lago-api/blob/main/spec/services/plans/create_service_spec.rb#L96

rsempe (approved):

app/models/add_on.rb
vincent-pochet: Yeah indeed, good catch!

PR #785, opened 2023-02-01
feat(delete_coupon): Add ability to soft delete a coupon

rsempe:
## Context

## Description

To allow soft deletion of customers, we first need to be able to soft delete add_ons and coupons

The goal of this PR is to add a `deleted_at` field to the `coupons` table and the Discard logic to allow soft deletion.

### DIFF
--- a/app/graphql/mutations/coupons/destroy.rb
+++ b/app/graphql/mutations/coupons/destroy.rb
@@ -13,7 +13,8 @@ module Mutations
       field :id, ID, null: true
 
       def resolve(id:)
-        result = ::Coupons::DestroyService.new(context[:current_user]).destroy(id)
+        coupon = context[:current_user].coupons.find_by(id:)
+        result = ::Coupons::DestroyService.call(coupon:)
 
         result.success? ? result.coupon : result_error(result)
       end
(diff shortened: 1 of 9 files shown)


app/graphql/mutations/coupons/destroy.rb
```diff
@@ -13,7 +13,8 @@ class Destroy < BaseMutation
       field :id, ID, null: true
 
       def resolve(id:)
-        result = ::Coupons::DestroyService.new(context[:current_user]).destroy(id)
+        coupon = context[:current_user].coupons.find_by(id:)
```
jdenquin, line 16: why do we chose to remove it from the `DestroyService`

jdenquin (approved):

lovrocolic:
Can we please sync about this topic before merging and see if there is any conflict with new coupons logic (plan limitations)?

app/graphql/mutations/coupons/destroy.rb
rsempe: to not have `destroy_from_api` and `destroy` methods anymore ;)

PR #788, opened 2023-02-01
feat (coupons): add DB changes for coupons with plan limitations

lovrocolic:
## Context

It’s impossible to assign a coupon only for some dedicated plans. It means that coupons is deducted from total invoice no matter the plans.

## Description

This PR adds DB changes that are needed for coupons with plan limitations

### DIFF
--- a/db/migrate/20230125104957_create_coupon_plans.rb
+++ b/db/migrate/20230125104957_create_coupon_plans.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateCouponPlans < ActiveRecord::Migration[7.0]
+  def change
+    create_table :coupon_plans, id: :uuid do |t|
+      t.references :coupon, type: :uuid, index: true, null: false, foreign_key: true
+      t.references :plan, type: :uuid, index: true, null: false, foreign_key: true
+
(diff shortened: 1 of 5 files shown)


db/migrate/20230125104957_create_coupon_plans.rb
```diff
@@ -0,0 +1,14 @@
...
+class CreateCouponPlans < ActiveRecord::Migration[7.0]
+  def change
+    create_table :coupon_plans, id: :uuid do |t|
+      t.references :coupon, type: :uuid, index: true, null: false, foreign_key: true
+      t.references :plan, type: :uuid, index: true, null: false, foreign_key: true
+
+      t.timestamps
+    end
+
+    add_column :coupons, :limited_plans, :boolean, default: false, null: false
```
rsempe, line 12: Do we want to merge this one as well?

rsempe (approved):

db/migrate/20230125104957_create_coupon_plans.rb
lovrocolic: I would say yes, it has no affect and it's not exposed. Also it's false by default and that would be default behaviour

PR #791, opened 2023-02-02
fix(payment_provider): Prevent duplicated payment provider customers

vincent-pochet:
## Context

In production, two users have duplicated Stripe customers. The reason why is still unclear, but it leads to unpredictable behaviors when trying to update the stripe customer id.

## Description

To prevent duplication of payment provider customer, this PR adds a validation rule to ensure each customer can have only one PaymentProviderCustomer per type of provider.

### DIFF
--- a/app/models/payment_provider_customers/base_customer.rb
+++ b/app/models/payment_provider_customers/base_customer.rb
@@ -10,6 +10,8 @@ module PaymentProviderCustomers
     has_many :payments
     has_many :refunds, foreign_key: :payment_provider_customer_id
 
+    validates :customer_id, uniqueness: { scope: :type }
+
     def push_to_settings(key:, value:)
       self.settings ||= {}
       settings[key] = value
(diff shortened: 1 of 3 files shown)


jdenquin (approved):

app/models/payment_provider_customers/base_customer.rb
```diff
@@ -10,6 +10,8 @@ class BaseCustomer < ApplicationRecord
     has_many :payments
     has_many :refunds, foreign_key: :payment_provider_customer_id
 
+    validates :customer_id, uniqueness: { scope: :type }
```
jdenquin, line 13: Rails/UniqueValidationWithoutIndex: Uniqueness validation should have a unique index on the database column.

rsempe (approved):

rsempe:
Maybe updating the database index to be unique as well?

PR #793, opened 2023-02-02
feat(delete_customer): Add attributes to allow soft deletion

vincent-pochet:
## Context

## Description

The goal of this PR is to add the `deleted_at` column to the `customers` table, to allow soft deletion of customers

### DIFF
--- a/spec/services/customers/create_service_spec.rb
+++ b/spec/services/customers/create_service_spec.rb
@@ -8,11 +8,12 @@ RSpec.describe Customers::CreateService, type: :service do
   let(:user) { nil }
   let(:membership) { create(:membership) }
   let(:organization) { membership.organization }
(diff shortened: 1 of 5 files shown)


spec/services/customers/create_service_spec.rb
```diff
@@ -8,11 +8,12 @@
   let(:user) { nil }
   let(:membership) { create(:membership) }
   let(:organization) { membership.organization }
+  let(:external_id) { SecureRandom.uuid }
 
   describe 'create_from_api' do
     let(:create_args) do
       {
-        external_id: SecureRandom.uuid,
+        external_id: external_id,
```
jdenquin, line 16: Style/HashSyntax: Omit the hash value. (https://rubystyle.guide#hash-literals)

rsempe (approved):

PR #795, opened 2023-02-03
feat: Restrict add on update if already applied

rsempe:
The goal of this PR is to restrict the update of an add_on if already applied to a subscription.


vincent-pochet (approved):
:rocket:

PR #796, opened 2023-02-03
feat(delete_customer): Keep customer visible from invoice

vincent-pochet:
## Context

## Description

This PR adds two changes:
- It creates a direct relation between `invoice` and `organization` models, rather than going through `customer`
- It ensure soft deleted customers remains visible from an invoice perspective

### DIFF
--- a/spec/jobs/invoices/prepaid_credit_job_spec.rb
+++ b/spec/jobs/invoices/prepaid_credit_job_spec.rb
@@ -3,19 +3,20 @@
 require 'rails_helper'
 
 RSpec.describe Invoices::PrepaidCreditJob, type: :job do
-  let(:invoice) { create(:invoice, customer: customer) }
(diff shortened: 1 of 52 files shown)


spec/jobs/invoices/prepaid_credit_job_spec.rb
```diff
@@ -3,19 +3,20 @@
...
   end
   let(:fee) do
-    create(:fee,
+    create(
+      :fee,
       fee_type: 'credit',
       invoiceable_type: 'WalletTransaction',
       invoiceable_id: wallet_transaction.id,
-      invoice: invoice
+      invoice:
```
jdenquin, line 19: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

rsempe (approved):

PR #801, opened 2023-02-06
feat(delete_customer): Remove delete limitation

vincent-pochet:
## Context

## Description

This PR adds two changes:
- It removes the not allowed error when trying to delete a plan attached to a subscription
- It removes the `canBeDeleted` attribute from the `Customer` GraphQL type

### DIFF
--- a/app/models/customer.rb
+++ b/app/models/customer.rb
@@ -46,14 +46,6 @@ class Customer < ApplicationRecord
   validates :timezone, timezone: true, allow_nil: true
   validates :vat_rate, numericality: { less_than_or_equal_to: 100, greater_than_or_equal_to: 0 }, allow_nil: true
 
-  def attached_to_subscriptions?
-    subscriptions.exists?
-  end
-
-  def deletable?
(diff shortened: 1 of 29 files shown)


ansmonjol (approved):

app/models/customer.rb
```diff
@@ -81,7 +73,7 @@ def applicable_invoice_grace_period
   end
 
   def editable?
-    !attached_to_subscriptions? &&
+    subscriptions.none? &&
       applied_add_ons.none? &&
       applied_coupons.where.not(amount_currency: nil).none? &&
```
rsempe, line 78: Do you know why we have `where.not(amount_currency: nil)` here? 🤔

rsempe (approved):

app/models/customer.rb
vincent-pochet: @rsempe if we assign any resource holding currency to a customer, it becomes impossible to update the currency of the customer.
Regarding coupon, the limitation does not apply to "percentage" coupons as they do not hold a currency

lovrocolic (approved):

PR #805, opened 2023-02-06
feat(webhooks): Add Webhooks Table and Model

jdenquin:
## Context

We want to add an history of webhooks for users to be able to manage them.

## Description

- Add the `webhooks` table and `webhook` model
- Add some seeds

### DIFF
--- a/app/models/webhook.rb
+++ b/app/models/webhook.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Webhook < ApplicationRecord
+  STATUS = %i[pending succeeded failed].freeze
+
+  belongs_to :organization
+  belongs_to :object, polymorphic: true
+
+  enum status: STATUS
+end
--- a/db/migrate/20230206143214_create_webhooks.rb
+++ b/db/migrate/20230206143214_create_webhooks.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class CreateWebhooks < ActiveRecord::Migration[7.0]
+  def change
+    create_table :webhooks, id: :uuid do |t|
+      t.references :organization, index: true
+
+      t.uuid :object_id, null: false
+      t.string :object_type, null: false
+
+      t.integer :status, default: 0
+      t.integer :retries, default: 0
+      t.integer :http_status
+
+      t.string :endpoint
+      t.string :webhook_type
+
+      t.json :payload
+      t.json :response
+
+      t.timestamp :last_retried_at
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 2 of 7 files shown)


db/migrate/20230206143214_create_webhooks.rb
```diff
@@ -0,0 +1,26 @@
...
+class CreateWebhooks < ActiveRecord::Migration[7.0]
+  def change
+    create_table :webhooks, id: :uuid do |t|
+      t.references :organization, index: true
+
+      t.uuid :object_id, null: false
+      t.string :object_type, null: false
+
+      t.integer :status, default: 0
+      t.integer :retries, default: 0
```
rsempe, line 12:
```suggestion
      t.integer :status, default: 0, null: false
      t.integer :retries, default: 0, null: false
```

db/migrate/20230206143214_create_webhooks.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class CreateWebhooks < ActiveRecord::Migration[7.0]
+  def change
+    create_table :webhooks, id: :uuid do |t|
+      t.references :organization, index: true
+
+      t.uuid :object_id, null: false
+      t.string :object_type, null: false
```
rsempe, line 9: No index?

app/models/webhook.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Webhook < ApplicationRecord
+  STATUS = %i[pending succeeded failed].freeze
```
lovrocolic, line 4: Not sure which scenario is covered with `pending`. On first it seems to me that we only need succeeded for 200 status code and `failed` for everything else. Probably I'm missing something:)?

jdenquin: it can happen if we have some wait in the sidekiq's queue, it's the default status

db/migrate/20230206143214_create_webhooks.rb
jdenquin: no real need here, it's just for information more than usage

db/migrate/20230206143214_create_webhooks.rb
jdenquin: since we have a default is it useful?

app/models/webhook.rb
lovrocolic: I see, I was thinking that object will be created inside a job

jdenquin: in facts it is, but I'll prefer to keep the pending state, we never know!

db/migrate/20230206143214_create_webhooks.rb
vincent-pochet: I think yes, default is set only when creating the row, it does not prevent to set the value to null afterwards

vincent-pochet (approved):

rsempe (approved):

PR #807, opened 2023-02-07
feat (locales): Add base setup for localization

lovrocolic:
## Context

For legal reason, some companies had to generate invoices in their own languages. Unfortunately Lago is supporting only one language that do not fit their need.

## Description

This PR adds base DB changes and setup for allowing internalisation on the pdf documents

### DIFF
--- a/app/models/customer.rb
+++ b/app/models/customer.rb
@@ -36,6 +36,8 @@ class Customer < ApplicationRecord
   default_scope -> { kept }
   sequenced scope: ->(customer) { customer.organization.customers.with_discarded }
 
+  enum document_locale: I18n.available_locales
+
   validates :country, country_code: true, unless: -> { country.nil? }
   validates :currency, inclusion: { in: currency_list }, allow_nil: true
   validates :external_id,
--- a/config/application.rb
+++ b/config/application.rb
@@ -22,6 +22,10 @@ module LagoApi
     config.active_record.encryption.deterministic_key = ENV['ENCRYPTION_DETERMINISTIC_KEY']
     config.active_record.encryption.key_derivation_salt = ENV['ENCRYPTION_KEY_DERIVATION_SALT']
 
+    config.i18n.load_path += Dir[Rails.root.join('config', 'locales', '**', '*.{rb,yml}')]
+    config.i18n.available_locales = %i[en fr]
+    config.i18n.default_locale = :en
+
     config.generators do |g|
       g.orm :active_record, primary_key_type: :uuid
     end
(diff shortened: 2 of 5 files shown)


config/application.rb
```diff
@@ -22,6 +22,10 @@ class Application < Rails::Application
     config.active_record.encryption.deterministic_key = ENV['ENCRYPTION_DETERMINISTIC_KEY']
     config.active_record.encryption.key_derivation_salt = ENV['ENCRYPTION_KEY_DERIVATION_SALT']
 
+    config.i18n.load_path += Dir[Rails.root.join('config', 'locales', '**', '*.{rb,yml}')]
```
jdenquin, line 25: Rails/FilePath: Prefer `Rails.root.join('path/to')`.

app/models/customer.rb
```diff
@@ -36,6 +36,8 @@ class Customer < ApplicationRecord
   default_scope -> { kept }
   sequenced scope: ->(customer) { customer.organization.customers.with_discarded }
 
+  enum document_locale: I18n.available_locales
```
vincent-pochet, line 39: I know it might have been covered in the dive-in, but what's the reason behind using an enum here, instead of a plain string type? 

My point is that if the order of `I18n.available_locales` we could end up in a really messed-up situation...
I would personally suggest to use the same approach as with countries and maybe with a custom validator. I think it will also be easier to read in the database.

lovrocolic: I see your point regarding I18n.available_locales reordering (and possible consequences if order changes with adding new locales) since we will have greater control by moving away from enums 👍  However, TBH not sure why not to use ISO language codes for `document_locale` field? If we use string words for language, we will also have to maintain mapping between language string and ISO language code... We can maybe switch from enum to string (with validator) but also use language iso code as value?

vincent-pochet: Oh sure indeed, I was thinking at ISO code, like for countries or currencies :)

lovrocolic: updated

app/validators/language_code_validator.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class LanguageCodeValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    record.errors.add(attribute, :language_code_invalid) unless valid?(value)
```
vincent-pochet, line 5: Could you add the message for this error in `/api/config/locales/en.yml`?

vincent-pochet (approved):
Nice! :rocket:

app/validators/language_code_validator.rb
lovrocolic: I always forget it:) Added!

PR #808, opened 2023-02-07
feat(webhooks): Add Webhooks GQL Resolver

jdenquin:
## Description

- Add the GQL Webhooks Resolver

### DIFF
--- a/app/graphql/resolvers/webhooks_resolver.rb
+++ b/app/graphql/resolvers/webhooks_resolver.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Resolvers
+  class WebhooksResolver < Resolvers::BaseResolver
+    include AuthenticableApiUser
+    include RequiredOrganization
+
+    description 'Query Webhooks'
+
+    argument :page, Integer, required: false
+    argument :limit, Integer, required: false
+    argument :status, Types::Webhooks::StatusEnum, required: false
+
+    type Types::Webhooks::Object.collection_type, null: false
+
+    def resolve(page: nil, limit: nil, status: nil)
+      validate_organization!
+
+      webhooks = current_organization.webhooks
+        .page(page)
+        .per(limit)
+
+      webhooks = webhooks.where(status:) if status.present?
+
+      webhooks.order(updated_at: :desc)
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


app/graphql/resolvers/webhooks_resolver.rb
```diff
@@ -0,0 +1,28 @@
...
+    argument :status, Types::Webhooks::StatusEnum, required: false
+
+    type Types::Webhooks::Object.collection_type, null: false
+
+    def resolve(page: nil, limit: nil, status: nil)
+      validate_organization!
+
+      webhooks = current_organization.webhooks
+        .page(page)
+        .limit(limit)
```
vincent-pochet, line 21: I think what you want here is a `per(limit)` :)

jdenquin: oh yes good catch!

app/graphql/resolvers/webhooks_resolver.rb
```diff
@@ -0,0 +1,28 @@
...
+    def resolve(page: nil, limit: nil, status: nil)
+      validate_organization!
+
+      webhooks = current_organization.webhooks
+        .page(page)
+        .per(limit)
+
+      webhooks = webhooks.where(status:) if status.present?
+
+      webhooks.order(updated_at: :desc)
```
rsempe, line 25: Are we sure we want `updated_at` here? Not `last_retried_at` or `created_at`?

rsempe (approved):

app/graphql/resolvers/webhooks_resolver.rb
jdenquin: since the `last_retried_at` can be `nil`, I think the source of truth is the `updated_at`, wdyt?

app/graphql/resolvers/webhooks_resolver.rb
```diff
@@ -0,0 +1,28 @@
...
+    argument :page, Integer, required: false
+    argument :limit, Integer, required: false
+    argument :status, Types::Webhooks::StatusEnum, required: false
+
+    type Types::Webhooks::Object.collection_type, null: false
+
+    def resolve(page: nil, limit: nil, status: nil)
+      validate_organization!
+
+      webhooks = current_organization.webhooks
```
lovrocolic, line 19: It would be good to switch to query service (WebhooksQuery), especially if there will be API endpoint for fetching webhooks

jdenquin: It will be on an other PR !

app/graphql/resolvers/webhooks_resolver.rb
rsempe: works for me - will see in the future if not :)

PR #809, opened 2023-02-07
feat(webhooks): Add Cleanup Job

jdenquin:
## Context

- We need to delete webhooks that are older than 90 days

## Description
 
- Add a Clock job to remove old webhooks

### DIFF
--- a/app/jobs/clock/webhooks_cleanup_job.rb
+++ b/app/jobs/clock/webhooks_cleanup_job.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Clock
+  class WebhooksCleanupJob < ApplicationJob
+    queue_as 'clock'
+
+    def perform
+      Webhook.where('updated_at < ?', 90.days.ago).destroy_all
+    end
+  end
+end
--- a/spec/jobs/clock/webhooks_cleanup_job_spec.rb
+++ b/spec/jobs/clock/webhooks_cleanup_job_spec.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Clock::WebhooksCleanupJob, job: true do
+  subject(:webhooks_cleanup_job) { described_class }
+
+  describe '.perform' do
+    before do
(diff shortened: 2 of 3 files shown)


vincent-pochet (approved):

app/jobs/clock/webhooks_cleanup_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Clock
+  class WebhooksCleanupJob < ApplicationJob
+    queue_as 'clock'
+
+    def perform
+      Webhook.where('updated_at < ?', 90.days.ago).destroy_all
```
rsempe, line 8: Do we really need to hard-destroy them? 🤔 

Maybe hiding them by an archived status?

spec/jobs/clock/webhooks_cleanup_job_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Clock::WebhooksCleanupJob, job: true do
+  subject { described_class }
```
rsempe, line 6: this line is useless, right? 😅

spec/jobs/clock/webhooks_cleanup_job_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  describe '.perform' do
+    before do
+      create_list(:webhook, 5, :succeeded, updated_at: 95.days.ago)
```
rsempe, line 10: Can we create only one instance here instead of 5? 😅

app/jobs/clock/webhooks_cleanup_job.rb
jdenquin: I think we don't need to keep them, it will possibly be a very large table if we keep everything

spec/jobs/clock/webhooks_cleanup_job_spec.rb
jdenquin: it is, copy paste from an other job 😂

lovrocolic (approved):

PR #810, opened 2023-02-07
feat(delete_customer): Add scenario spec

rsempe:
## Context

## Description

This PR adds a spec scenario for deleting a customer.


vincent-pochet (approved):
Really cool!! :rocket:

PR #814, opened 2023-02-08
feat(webhooks): Add Search capability

jdenquin:
## Context

We want to be able to search webhooks by `id` or `webhook_type`, and also filter by status

## Description
- Add search capability on webhooks


lovrocolic (approved):
🎸

PR #817, opened 2023-02-09
feat (localization): add support for customer documentLocale in graphQL

lovrocolic:
## Context

For legal reason, some companies had to generate invoices in their own languages. Unfortunately Lago is supporting only one language that do not fit their need.

## Description

This PR handles updating customer document locale from the graphQL


app/graphql/types/customers/billing_configuration.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Types
+  module Customers
+    class BillingConfiguration < Types::BaseObject
+      graphql_name 'CustomerBillingConfiguration'
+
+      field :id, ID, null: false
```
vincent-pochet, line 8: What is this ID? Not sure to understand where it comes from :thinking:

vincent-pochet (commented):
Apart from this small question about ID, looks good :rocket:

app/graphql/types/customers/billing_configuration.rb
ansmonjol: this ID is required by the frontend 😕

Thing is, our cache management system uses IDs to manage it's cache updates.

When we define a new type, the cache systems sees those resources as separate models, hence warning us about potential data loss.

To prevent this, I asked Lovro to allow us to request an ID on the type level. Also, note this ID is not present in the input type.

This ID could have the value of the parent (here customer) ID in case having a random one is an issue.
Also, there is no need for this ID to be stored in DB as it's only used bu the cache management system to properly store and update a resource.

Does it make sense and answer your question ?

spec/graphql/mutations/customers/update_spec.rb
```diff
@@ -21,6 +21,7 @@
           canEditAttributes
           invoiceGracePeriod
           providerCustomer { id, providerCustomerId }
+          billingConfiguration { documentLocale }
```
ansmonjol, line 24: Maybe request the ID to make sure it's present 🙏

app/graphql/types/customers/billing_configuration.rb
vincent-pochet: Alright! Thank you for the explanation :)

vincent-pochet (approved):

spec/graphql/mutations/customers/update_spec.rb
```diff
@@ -59,6 +63,8 @@
       expect(result_data['invoiceGracePeriod']).to be_nil
       expect(result_data['providerCustomer']['id']).to be_present
       expect(result_data['providerCustomer']['providerCustomerId']).to eq('cu_12345')
+      expect(result_data['billingConfiguration']['documentLocale']).to eq('fr')
+      expect(result_data['billingConfiguration']['id']).to be
```
jdenquin, line 67: RSpec/Be: Don't use `be` without an argument. (https://rspec.rubystyle.guide/#be-matcher, https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Be)

spec/graphql/mutations/customers/update_spec.rb
lovrocolic: added

PR #820, opened 2023-02-12
fix(specs): Fix a flaky test on subscriptions date

jdenquin:
- Do not use a `now` for this spec


vincent-pochet (approved):
Looks good, I think... :stuck_out_tongue:

PR #823, opened 2023-02-13
fix(webhooks): Organization ref should be an uuid

jdenquin:
- Change `organization_id` from `bigint` to `uuid`

### DIFF
--- a/db/migrate/20230213180133_change_webhooks_organization_id_type.rb
+++ b/db/migrate/20230213180133_change_webhooks_organization_id_type.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class ChangeWebhooksOrganizationIdType < ActiveRecord::Migration[7.0]
+  def change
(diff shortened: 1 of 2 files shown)


db/migrate/20230213180133_change_webhooks_organization_id_type.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ChangeWebhooksOrganizationIdType < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :webhooks, :organization_id
```
jdenquin, line 5: Rails/ReversibleMigration: remove_column(without type) is not reversible. (https://rails.rubystyle.guide#reversible-migration, https://api.rubyonrails.org/classes/ActiveRecord/Migration/CommandRecorder.html)

vincent-pochet (approved):
:trollface:

PR #824, opened 2023-02-14
fix(usage): Round VAT at invoice level

vincent-pochet:
## Context

Lago API performs some math on top of monetary values, to calculate VAT, internal fees, etc.

Since these operations are done using integer amounts in cents and are spread around the codebase, there are some rounding issues and inconsistencies, like off-by-one errors.

## Description

This PR ensure VAT amount computed at invoice level on customer current usage is computed based on fee amount without using rounded value computed at fee level, to prevent rounding issues

### DIFF
--- a/app/services/invoices/customer_usage_service.rb
+++ b/app/services/invoices/customer_usage_service.rb
@@ -98,7 +98,7 @@ module Invoices
 
     def compute_amounts
       invoice.amount_cents = invoice.fees.sum(&:amount_cents)
-      invoice.vat_amount_cents = invoice.fees.sum(&:vat_amount_cents)
+      invoice.vat_amount_cents = invoice.fees.sum { |f| f.amount_cents * f.vat_rate }.fdiv(100).ceil
       invoice.total_amount_cents = invoice.amount_cents + invoice.vat_amount_cents
     end
 
--- a/spec/services/invoices/customer_usage_service_spec.rb
+++ b/spec/services/invoices/customer_usage_service_spec.rb
@@ -122,7 +122,7 @@ RSpec.describe Invoices::CustomerUsageService, type: :service do
         )
       end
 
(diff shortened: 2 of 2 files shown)


app/services/invoices/customer_usage_service.rb
```diff
@@ -98,7 +98,7 @@ def boundaries
 
     def compute_amounts
```
rsempe, line 99: Would be nice to call `Invoices::ComputeAmountsFromFees` no?

spec/services/invoices/customer_usage_service_spec.rb
```diff
@@ -3,258 +3,144 @@
...
-            subscription_at: subscription_at,
-            started_at: started_at,
-            billing_time: :anniversary,
-          )
-        end
-
-        it 'initialize an invoice' do
-          travel_to(current_date) do
-            result = invoice_service.usage
+    it 'initialize an invoice' do
```
rsempe, line 72:
```suggestion
    it 'initializes an invoice' do
```

spec/services/invoices/customer_usage_service_spec.rb
```diff
@@ -3,258 +3,144 @@
...
-          )
-        end
-
-        it 'initialize an invoice' do
-          travel_to(current_date) do
-            result = invoice_service.usage
-
-            aggregate_failures do
-              expect(result).to be_success
+      it 'initialize an invoice' do
```
rsempe, line 125:
```suggestion
      it 'initializes an invoice' do
```

rsempe (approved):

app/services/invoices/customer_usage_service.rb
vincent-pochet: Fees are not persisted in this class, so sadly it's not working

PR #826, opened 2023-02-14
fix(webhooks): Change Webhooks organization reference type

jdenquin:
- use `uuid` for `organization_id` on `webhooks` table

### DIFF
--- a/db/migrate/20230214100638_change_webhooks_organization_id_to_uuid.rb
+++ b/db/migrate/20230214100638_change_webhooks_organization_id_to_uuid.rb
@@ -4,6 +4,6 @@ class ChangeWebhooksOrganizationIdToUuid < ActiveRecord::Migration[7.0]
   def change
     remove_column :webhooks, :organization_id, :bigint
 
-    add_reference :webhooks, :organization, type: :uuid, index: true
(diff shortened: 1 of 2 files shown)


db/migrate/20230214100638_change_webhooks_organization_id_to_uuid.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ChangeWebhooksOrganizationIdToUuid < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :webhooks, :organization_id, :bigint
+
+    add_reference :webhooks, :organization, type: :uuid, index: true
```
rsempe, line 7: `null: false`?

jdenquin: it should not, but if we have some users with `webhooks` already in database, it will break everything

jdenquin, line 7: Rails/NotNullColumn: Do not add a NOT NULL column without a default value.

vincent-pochet (approved):

PR #837, opened 2023-02-16
fix (applied coupons): fix delete applied coupon logic by using lago id as identifier

lovrocolic:
## Description

Currently we can assign multiple coupons of a same type on a customer. That being said we need to remove applied coupon by unique identifier.


rsempe (approved):

rsempe:
Just to not forget to update clients and docs ;)

PR #840, opened 2023-02-16
misc: Merge API and GraphQL Coupons update services

vincent-pochet:
## Description

This PR merges `Coupons::UpdateService#update` and `Coupons::UpdateService#update_from_api` methods into a single one named `Coupons::UpdateService#call`. 

This will reduce the amount of code and test to maintain and ensure the behavior is the same between API and GraphQL

### DIFF
--- a/app/controllers/api/v1/coupons_controller.rb
+++ b/app/controllers/api/v1/coupons_controller.rb
@@ -20,10 +20,10 @@ module Api
       end
 
       def update
-        service = Coupons::UpdateService.new
-        result = service.update_from_api(
-          organization: current_organization,
-          code: params[:code],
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/coupons_controller.rb
```diff
@@ -20,10 +20,13 @@ def create
       end
 
       def update
-        service = Coupons::UpdateService.new
-        result = service.update_from_api(
-          organization: current_organization,
+        coupon = Coupon.find_by(
           code: params[:code],
+          organization_id: current_organization.id,
+        )
```
rsempe, line 26: `current_organization.coupons.find_by(code: params[:code])`?

rsempe (approved):

PR #843, opened 2023-02-20
feat(applied coupons): extend applied coupon serializer

lovrocolic:
## Context

Some clients requested extension for applied coupons serializer. They want to see on which invoices coupon was spent and the amount that is used. This can be achieved by extending credit serializer and returning list of credits with each applied coupon.

### DIFF
--- a/app/serializers/v1/credit_serializer.rb
+++ b/app/serializers/v1/credit_serializer.rb
@@ -16,7 +16,7 @@ module V1
         invoice: {
           lago_id: model.invoice_id,
           payment_status: model.invoice.payment_status,
-        }
+        },
       }
     end
   end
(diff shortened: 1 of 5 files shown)


app/serializers/v1/credit_serializer.rb
```diff
@@ -13,6 +13,10 @@ def serialize
           code: model.item_code,
           name: model.item_name,
         },
+        invoice: {
+          lago_id: model.invoice_id,
+          payment_status: model.invoice.payment_status,
+        }
```
jdenquin, line 19: Style/TrailingCommaInHashLiteral: Put a comma after the last item of a multiline hash.

vincent-pochet (approved):

rsempe (approved):

PR #856, opened 2023-02-22
bug (plans): fix update plan charges condition

lovrocolic:
## Context

We don't want to sanitize charges if they are not passed at all on update request

### DIFF
--- a/app/services/plans/update_service.rb
+++ b/app/services/plans/update_service.rb
@@ -36,7 +36,7 @@ module Plans
       ActiveRecord::Base.transaction do
         plan.save!
 
-        process_charges(plan, params[:charges] || [])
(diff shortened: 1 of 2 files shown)


app/services/plans/update_service.rb
```diff
@@ -36,7 +36,7 @@ def call
       ActiveRecord::Base.transaction do
         plan.save!
 
-        process_charges(plan, params[:charges] || [])
+        process_charges(plan, params[:charges]) unless params[:charges].nil?
```
rsempe, line 39:
```suggestion
        process_charges(plan, params[:charges]) if params[:charges]
```

rsempe (approved):

vincent-pochet (approved):

PR #857, opened 2023-02-23
fix(stripe): Ignore not found refund webhook

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/845

This fix is related to the following error:
```
BaseService::NotFoundFailure
stripe_refund_not_found
```
We have a lot of them in sentry and a lot of related dead jobs.

The issue is that we are receiving stripe webhooks for all refund executed by Stripe, even those not initiated by the receiving lago instance, leading to a failure because the app is unable to find a matching refund.

## Description

Since we are sending the invoice ID in the metadata when initiating the refund, to we should be looking for the presence of this invoice in the database and ignore the hook if the invoice is not found.

### DIFF
--- a/app/services/credit_notes/refunds/stripe_service.rb
+++ b/app/services/credit_notes/refunds/stripe_service.rb
@@ -34,15 +34,15 @@ module CreditNotes
         result
       end
 
-      def update_status(provider_refund_id:, status:)
-        refund = Refund.find_by(provider_refund_id: provider_refund_id)
-        return result.not_found_failure!(resource: 'stripe_refund') unless refund
+      def update_status(provider_refund_id:, status:, metadata: {})
+        refund = Refund.find_by(provider_refund_id:)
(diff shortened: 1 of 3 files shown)


app/services/credit_notes/refunds/stripe_service.rb
```diff
@@ -147,6 +147,16 @@ def track_refund_status_changed(status)
           },
         )
       end
+
+      def handle_missing_refund(metadata)
+        # NOTE: Refund was not initiated by lago
+        return result unless metadata&.key?(:lago_invoice_id)
+
+        # NOTE: Invoice does not belongs to this lago instance
+        return result if Invoice.find_by(id: metadata[:lago_invoice_id]).nil?
```
rsempe, line 156:
```suggestion
        return result unless Invoice.find_by(id: metadata[:lago_invoice_id])
```

rsempe (approved):

PR #862, opened 2023-02-23
feat(webhooks): Resend webhooks

jdenquin:
## Context

We wand to add the capability to resend a failed front end on the GraphQL API

## Description

- Add the resend logic and GraphQL Mutation


app/services/webhooks/base_service.rb
```diff
@@ -5,12 +5,14 @@
 module Webhooks
   # NOTE: Abstract Service, should not be used directly
   class BaseService
-    def initialize(object:, options: {})
+    def initialize(object:, options: {}, webhook_id: nil)
```
lovrocolic, line 8: I guess we should put options as a last param since it's a hash

app/graphql/mutations/webhooks/resend.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Mutations
+  module Webhooks
+    class Resend < BaseMutation
```
rsempe, line 5: What about changing the naming from `Resend` to `Retry` eveyrwhere? 😬 

If I understand correctly, nothing has already been sent as it was a failure. Plus it can be consistent with the naming of `last_retried_at` column and the variable `LAGO_WEBHOOK_ATTEMPTS`.

app/graphql/mutations/webhooks/resend.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      argument :id, ID, required: true
+
+      type Types::Webhooks::Object
+
+      def resolve(id:)
+        validate_organization!
+
+        webhook = current_organization.webhooks.find_by(id:)
+        result = ::Webhooks::ResendService.new(webhook:).call
```
rsempe, line 20:
```suggestion
        result = ::Webhooks::ResendService.call(webhook:)
```

spec/services/webhooks/resend_service_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Webhooks::ResendService, type: :service do
+  subject(:resend_service) { described_class.new(webhook:) }
+
+  let(:webhook) { create(:webhook, :failed) }
+
+  it 'enqueues a SendWebhookJob' do
```
rsempe, line 10: Maybe another example for ensuring that webhook is correctly assigned to the result response?

vincent-pochet (approved):

app/services/webhooks/base_service.rb
```diff
@@ -79,12 +96,13 @@ def issuer
     end
 
     def current_webhook
-      @current_webhook ||= current_organization.webhooks.find_or_initialize_by(
+      @current_webhook ||= Webhook.find_or_initialize_by(
```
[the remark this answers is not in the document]
jdenquin: not for Webhook that are retried, because we don't use the base object

app/graphql/mutations/webhooks/resend.rb
jdenquin: we can but I'll keep the mutation name as `ResendWebhook` if that works for you

rsempe (approved):

PR #866, opened 2023-02-27
feat(instant_charge): Add API route to show a fee

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds a new route to retrieve a single fee via the API (`GET /api/v1/fees/:id`)

### DIFF
--- a/spec/requests/api/v1/fees_spec.rb
+++ b/spec/requests/api/v1/fees_spec.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::FeesController, type: :request do
+  let(:organization) { create(:organization) }
+
+  describe 'GET /fees/:id' do
+    let(:invoice) { create(:invoice, organization:) }
+    let(:fee) { create(:fee, invoice:) }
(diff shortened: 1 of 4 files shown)


spec/requests/api/v1/fees_spec.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::FeesController, type: :request do
```
jdenquin, line 5: RSpec/FilePath: Spec path should end with `api/v1/fees_controller*_spec.rb`. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FilePath)

db/schema.rb
```diff
@@ -112,6 +112,7 @@
     t.integer "charge_model", default: 0, null: false
     t.jsonb "properties", default: "{}", null: false
     t.datetime "deleted_at"
+    t.boolean "instant", default: false, null: false
```
rsempe, line 115: Not related to that PR ;)

rsempe (approved):

PR #867, opened 2023-02-27
feat (customer_metadata): Apply customer metadata on documents and limit metadata per customer

lovrocolic:
## Context

Currently there is no possibility to add additional data related to customer in Lago App

## Description

With this PR, following goals are satisfied:

- We are limiting number of metadata per customer
- We are applying metadata on documents


app/services/customers/update_service.rb
```diff
@@ -6,6 +6,10 @@ def update(**args)
       customer = result.user.customers.find_by(id: args[:id])
       return result.not_found_failure!(resource: 'customer') unless customer
 
+      unless valid_metadata_count?(metadata: args[:metadata])
```
jdenquin, line 9: maybe we can validates this with a validator on the model? wdyt?

lovrocolic: I am still not sure what is the final decision about validator classes, is it a rule of thumb to use it across the project? I know that some of us don't prefer it but some like it 😄 I personally haven't used it here since it is just one condition that needs to be checked on service level but if having validators are style that we are going to follow, I can change it so that we are consistent:)

lovrocolic: @jdenquin I talked about model validation with Vincent and Romain before and we concluded that such a validation is more business related and that we should place it service level as other similar validations. Do you agree with that?

app/services/customers/create_service.rb
```diff
@@ -6,6 +6,10 @@ def create_from_api(organization:, params:)
       customer = organization.customers.find_or_initialize_by(external_id: params[:external_id])
       new_customer = customer.new_record?
 
+      unless valid_metadata_count?(metadata: params[:metadata])
+        return result.not_allowed_failure!(code: 'invalid_number_of_metadata')
```
vincent-pochet, line 10: It suggest to render an unprocessable entity. As it's more a business logic validation than an "ability" to perform the action.
What do you think?

app/services/customers/update_service.rb
```diff
@@ -6,6 +6,10 @@ def update(**args)
...
+        return result.not_allowed_failure!(code: 'invalid_number_of_metadata')
```
vincent-pochet, line 10: Same comment regarding the type of error

app/services/customers/create_service.rb
lovrocolic: 👍 Changed

app/services/customers/update_service.rb
lovrocolic: 👍

vincent-pochet (approved):

PR #874, opened 2023-03-01
fix: update invoice and credit note customer name

lovrocolic:


app/views/templates/credit_note.slim
```diff
@@ -310,7 +310,11 @@ html
           .body-2 = organization.email
         .billing-information-column
           .body-1 = I18n.t('credit_note.credit_to')
-          .body-2 = customer.name
+          .body-2
+            - if customer.legal_name.present?
+              | #{customer.legal_name}
```
jdenquin, line 315: can we also check if it's a blank string?!

jdenquin (changes requested):

jdenquin (approved):

PR #876, opened 2023-03-02
fix(invoices): Fix invoices template when no fees

jdenquin:
- Add `nil` escape for billable metrics on invoice template


lovrocolic (approved):

app/views/templates/invoice.slim
```diff
@@ -553,9 +553,9 @@ html
                       td width="70%"
                         .body-1 = fee.billable_metric.name
```
vincent-pochet, line 554: I have the feeling that this one should be escaped to ;)

PR #890, opened 2023-03-03
feat(instant_charge): Allow fees#invoice_id to be null

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR allows a fee to be created withou an invoice

### DIFF
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -3,7 +3,7 @@
 class Fee < ApplicationRecord
   include Currencies
 
-  belongs_to :invoice
+  belongs_to :invoice, optional: true
   belongs_to :charge, -> { with_discarded }, optional: true
   belongs_to :applied_add_on, optional: true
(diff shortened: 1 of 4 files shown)


db/migrate/20230303153213_allow_nil_invoice_for_fees.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AllowNilInvoiceForFees < ActiveRecord::Migration[7.0]
+  def change
+    change_column_null :fees, :invoice_id, null: true
```
rsempe, line 5: Hum this is already the case no? 😅 
Nothing changed in the schema...

vincent-pochet: Oh yeah indeed! :smile: 

Weird... Fixing it!

rsempe (approved):

PR #892, opened 2023-03-06
feat(instant_charge): Ability to have an instant charge

rsempe:
## Roadmap Task

## Context

## Description

This PR allows to have an instant charge.

### DIFF
--- a/app/controllers/api/v1/plans_controller.rb
+++ b/app/controllers/api/v1/plans_controller.rb
@@ -85,11 +85,16 @@ module Api
             :id,
             :billable_metric_id,
             :charge_model,
-            { properties: {} },
-            group_properties: [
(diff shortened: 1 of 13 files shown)


app/controllers/api/v1/plans_controller.rb
```diff
@@ -85,11 +85,16 @@ def input_params
...
-            ],
+            :instant,
+            {
+              properties: {},
+            },
+            {
+              group_properties: [
+                :group_id,
+                { values: {} },
+              ]
```
jdenquin, line 96: Style/TrailingCommaInHashLiteral: Put a comma after the last item of a multiline hash.

vincent-pochet (approved):

PR #893, opened 2023-03-06
feat(instant_charge): Compute instant aggregation

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds the logic to compute the instant aggregation for each aggregation types impacted by the feature

### DIFF
--- a/app/services/billable_metrics/aggregations/sum_service.rb
+++ b/app/services/billable_metrics/aggregations/sum_service.rb
@@ -4,10 +4,11 @@ module BillableMetrics
   module Aggregations
     class SumService < BillableMetrics::Aggregations::BaseService
       def aggregate(from_datetime:, to_datetime:, options: {})
-        events = events_scope(from_datetime: from_datetime, to_datetime: to_datetime)
+        events = events_scope(from_datetime:, to_datetime:)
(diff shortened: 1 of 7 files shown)


app/services/billable_metrics/aggregations/sum_service.rb
```diff
@@ -48,6 +49,13 @@ def running_total_per_aggregation(events, aggregation)
             accumulator << total += val
           end
       end
+
+      def compute_instant_aggregation
+        return 0 unless event
+        return 0 if event.properties.blank?
```
rsempe, line 55: Maybe returning big decimals as well here?

spec/services/billable_metrics/aggregations/sum_service_spec.rb
```diff
@@ -212,11 +216,42 @@
...
+
+    let(:properties) { { total_count: 12 } }
+
+    it 'assigns an instant aggregation' do
+      result = sum_service.aggregate(from_datetime:, to_datetime:)
+
+      expect(result.instant_aggregation).to eq(12)
+    end
+
+    context 'when event is missing properties' do
```
rsempe, line 247: maybe another exemple with a given properties not matching the expected field_name?

rsempe (approved):

PR #898, opened 2023-03-07
chore(AddOn): remove unused single type

ansmonjol:
## Context

While looking at `single_object` usages, I noticed a GraphQL resolved declared but not used by FE app.

## Description

This PR removed the unused resolver.

There is an associated PR on FE side: https://github.com/getlago/lago-front/pull/881


vincent-pochet (approved):
:metal:

PR #899, opened 2023-03-07
feat(instant_charge): Add instant_event_id to fees to track instant fee source

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds a new `instant_event_id` column to the fees table, in order to keep track of the event leading to the creation of the fee.
The data type is intentionally using `string` over `uuid` because no foreign key will be defined on the relation and we might want to be able to delete old events or to extract event table from the main database in the future

### DIFF
--- a/db/migrate/20230307131524_add_instant_event_id_to_fees.rb
+++ b/db/migrate/20230307131524_add_instant_event_id_to_fees.rb
@@ -2,6 +2,6 @@
 
 class AddInstantEventIdToFees < ActiveRecord::Migration[7.0]
   def change
-    add_column :fees, :instant_event_id, :string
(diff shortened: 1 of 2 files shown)


db/migrate/20230307131524_add_instant_event_id_to_fees.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddInstantEventIdToFees < ActiveRecord::Migration[7.0]
+  def change
+    add_column :fees, :instant_event_id, :string
```
rsempe, line 5:
```suggestion
    add_column :fees, :instant_event_id, :uuid
```

rsempe (approved):

PR #914, opened 2023-03-13
feat(bm): Prevent update if attached to plan

vincent-pochet:
## Context

Plans edition has some issue if you update some properties of a billable metric, especially if you change groups definition or aggregation type :

"After the billable metric edition, new dimensions are created with new group_id that do not match the ones in the plans already defined"

## Description

This PR prevents the update of `code`, `aggregation_type`, `field_name` and `groups` if the billable metric is attached to a plan.

It also exposes a new `plansCount` field in the `BillableMetric` graphql type to let the front-end know if the BM is editable.


rsempe (approved):

spec/serializers/v1/billable_metric_serializer_spec.rb
```diff
@@ -49,4 +50,11 @@
 
     expect(result['billable_metric']['draft_invoices_count']).to eq(1)
   end
+
+  it 'returns the number of plans' do
+    plan = create(:plan, organization: billable_metric.organization)
+    charge = create(:standard_charge, billable_metric:, plan:)
```
jdenquin, line 56: Lint/UselessAssignment: Useless assignment to variable - `charge`. (https://rubystyle.guide#underscore-unused-vars)

PR #916, opened 2023-03-14
fix: Correct percentage calculation for fixed amount

rsempe:
The goal of this PR is to fix an issue we currently have when calculating `free_units_count` for the percentage charge model.

I also added some scenarios in order to improve the coverage of our tests.


vincent-pochet (approved):
:rocket:

jdenquin (approved):

PR #918, opened 2023-03-14
feat(email): Add Organization email settings

jdenquin:
## Context

We want to save emails preferences for organization.

## Description

- Add `email_settings` field on `organizations`
- Support this new field on GraphQL `updateOrganization` mutation
- Support this new field on `PUT /organizations`


db/migrate/20230313145506_add_email_settings_to_organizations.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddEmailSettingsToOrganizations < ActiveRecord::Migration[7.0]
+  def change
+    add_column :organizations, :email_settings, :string, array: true, default: []
```
vincent-pochet, line 5: Since we have a default, shoudn't we set it as **not nullable**?

app/graphql/types/organization_type.rb
```diff
@@ -2,28 +2,34 @@
 
 module Types
   class OrganizationType < Types::BaseObject
+    description 'Organization Type'
+
     field :id, ID, null: false
-    field :name, String, null: false
+
+    field :address_line1, String
```
vincent-pochet, line 9: What about grouping these fields in "logical" block (at least for the address block)

```ruby
    field :id, ID, null: false

    field :legal_name, String
    field :legal_number, String
    field :logo_url, String
    field :name, String, null: false

    field :email, String

    field :api_key, String, null: false
    field :webhook_url, String
    
    field :address_line1, String
    field :address_line2, String
    field :city, String
    field :country, Types::CountryCodeEnum, null: true
    field :state, String
    field :zipcode, String

    field :timezone, Types::TimezoneEnum, null: true
    
```

app/graphql/types/organizations/update_organization_input.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Types
+  module Organizations
+    class UpdateOrganizationInput < BaseInputObject
+      description 'Update Organization input arguments'
+
+      argument :address_line1, String, required: false
```
vincent-pochet, line 8: Same comment about logical block :)

spec/models/organization_spec.rb
```diff
@@ -114,6 +114,18 @@
...
+    it 'is valid with email_settings' do
+      organization.email_settings = ['invoice.finalized', 'credit_note.created']
+
+      expect(organization).to be_valid
+    end
+
+    it 'is invalid with non permitted email_settings value' do
+      organization.email_settings = ['email.not_permitted']
+
+      expect(organization).not_to be_valid
```
rsempe, line 127: what about checking that the error is on `email_settings` instead of just not valid?

vincent-pochet (approved):

PR #921, opened 2023-03-17
feat(instant_charge): Allow instant fee creation without persistence

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR allows the Fees::CreateInstantService to perform the aggregation and compute fees amount without delivering webhook or persisting a fee. We will then use it to compute the expected fees of an event.

### DIFF
--- a/app/services/fees/create_instant_service.rb
+++ b/app/services/fees/create_instant_service.rb
@@ -2,9 +2,10 @@
 
 module Fees
   class CreateInstantService < BaseService
-    def initialize(charge:, event:)
+    def initialize(charge:, event:, estimate: false)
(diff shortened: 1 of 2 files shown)


app/services/fees/create_instant_service.rb
```diff
@@ -2,9 +2,10 @@
 
 module Fees
   class CreateInstantService < BaseService
-    def initialize(charge:, event:)
+    def initialize(charge:, event:, expected: false)
```
rsempe, line 5:
```suggestion
    def initialize(charge:, event:, estimate: false)
```

rsempe (approved):

PR #924, opened 2023-03-20
feat(emails): Add Invoice Finalized Mailer

jdenquin:
## Context

We want to send emails to customers when an invoice is finalized, when the email settings is activated.

## Description

- Add `InvoiceMailer`
- Add the `finalized` mailer and template
- Add Locales
- Add logic to send the email


spec/services/invoices/finalize_service_spec.rb
```diff
@@ -6,6 +6,8 @@
   subject(:finalize_service) { described_class.new(invoice:) }
 
   describe '#call' do
+    around { |test| lago_premium!(&test) }
+
```
jdenquin, line 10: Layout/TrailingWhitespace: Trailing whitespace detected. (https://rubystyle.guide#no-trailing-whitespace)

spec/services/invoices/add_on_service_spec.rb
```diff
@@ -11,6 +11,8 @@
   let(:applied_add_on) { create(:applied_add_on) }
 
   describe 'create' do
+    around { |test| lago_premium!(&test) }
```
vincent-pochet, line 14: What do you think of setting this only when testing the email delivery?

```ruby
describe 'with lago_premium' do
  around { |test| lago_premium!(&test) }

  it 'enqueues an ActionMailer::MailDeliveryJob'
    ...
end
```

That way the premium flag is limited to the smallest unit of tests, and you don't have to turn it off back for the "when license if not premium"

vincent-pochet (commented):
General comment regarding premium helper. I think we should use it only at lowest possible level, because it turning on the flag globally could have unpredictable effect on some specs

app/mailers/invoice_mailer.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def finalized
+    @invoice = params[:invoice]
+    @organization = @invoice.organization
+    @customer = @invoice.customer
+
+    return if @organization.email.blank?
+    return if @customer.email.blank?
+
+    I18n.locale = @customer.preferred_document_locale
```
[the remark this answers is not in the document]
vincent-pochet: Since you moved the logic below, you should remove this line ;)

spec/services/invoices/subscription_service_spec.rb
```diff
@@ -103,6 +105,32 @@
...
+      before { subscription.customer.organization.update!(email_settings: []) }
+
+      it 'does not enqueue an ActionMailer::MailDeliveryJob' do
+        expect do
+          invoice_service.create
+        end.not_to have_enqueued_job(ActionMailer::MailDeliveryJob)
+      end
+    end
+
+    context 'when license is not premium' do
```
vincent-pochet, line 124: What about having the same `with lago_premium` from the other specs? :)

vincent-pochet (approved):

PR #925, opened 2023-03-20
feat (customer portal): add graphql support for generating customer portal url

lovrocolic:
## Context

Customer portal allows end user to visit basic informations like subscription and invoices

## Description

This PR handles generating customer portal url from the graphQL


spec/services/customer_portal/generate_url_service.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe CustomerPortal::GenerateUrlService, type: :service do
```
jdenquin, line 5: RSpec/FilePath: Spec path should end with `customer_portal/generate_url_service*_spec.rb`. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FilePath)

app/graphql/mutations/customer_portal/generate_url.rb
```diff
@@ -0,0 +1,30 @@
...
+      include RequiredOrganization
+
+      graphql_name 'GenerateCustomerPortalUrl'
+      description 'Generate customer portal URL'
+
+      argument :id, ID, required: true
+
+      field :url, String, null: false
+
+      def resolve(**args)
```
vincent-pochet, line 16: Since there is only one input argument you could simply have: `def resolve(id:)`

vincent-pochet (approved):

app/graphql/mutations/customer_portal/generate_url.rb
lovrocolic: ✅

PR #932, opened 2023-03-23
feat(instant_charge): Add and update fees#payment_status

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds new `payment_status` field to `fees` table and ensures this field is always set

### DIFF
--- a/app/services/fees/add_on_service.rb
+++ b/app/services/fees/add_on_service.rb
@@ -21,6 +21,7 @@ module Fees
         invoiceable: applied_add_on,
         vat_rate: customer.applicable_vat_rate,
         units: 1,
+        payment_status: :pending,
       )
 
       new_fee.compute_vat
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -69,6 +69,7 @@ module Fees
         properties: boundaries.to_h,
         events_count: amount_result.count,
         group_id: group&.id,
+        payment_status: :pending,
       )
 
       new_fee.compute_vat
--- a/app/services/invoices/update_service.rb
+++ b/app/services/invoices/update_service.rb
@@ -44,6 +44,7 @@ module Invoices
         handle_prepaid_credits(params[:payment_status])
         track_payment_status_changed
         deliver_webhook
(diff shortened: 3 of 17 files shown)


app/services/fees/charge_service.rb
```diff
@@ -69,6 +69,7 @@ def init_fee(properties:, group: nil)
         properties: boundaries.to_h,
         events_count: amount_result.count,
         group_id: group&.id,
+        payment_status: :pending,
```
rsempe, line 72: that's the default value, we can probably remove that line, no?

app/services/fees/add_on_service.rb
```diff
@@ -21,6 +21,7 @@ def create
         invoiceable: applied_add_on,
         vat_rate: customer.applicable_vat_rate,
         units: 1,
+        payment_status: :pending,
```
rsempe, line 24: that's the default value, we can probably remove that line, no?

app/services/invoices/update_service.rb
```diff
@@ -27,7 +27,10 @@ def call
         )
       end
 
-      invoice.payment_status = params[:payment_status] if params.key?(:payment_status)
+      if params.key?(:payment_status)
+        invoice.payment_status = params[:payment_status]
+        invoice.fees.update_all(payment_status: params[:payment_status]) # rubocop:disable Rails/SkipsModelValidations
```
rsempe, line 32: maybe not using `update_all` here, just in case for future validations, no?

spec/services/fees/add_on_service_spec.rb
```diff
@@ -28,6 +28,7 @@
         expect(created_fee.vat_rate).to eq(20.0)
         expect(created_fee.units).to eq(1)
         expect(created_fee.events_count).to be_nil
+        expect(created_fee.payment_status).to eq('pending')
```
rsempe, line 31:
```suggestion
        expect(created_fee.payment_status).to be_pending
```

rsempe (approved):

app/services/fees/add_on_service.rb
vincent-pochet: I prefer to keep it to make it clear the default value is there and always `pending`

PR #934, opened 2023-03-24
feat(instant_charge): expose more details in fee serializer

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds new fields in the `FeeSerializer`. It will be updated later to add new `payment_status` field


lovrocolic (approved):

app/serializers/v1/fee_serializer.rb
```diff
@@ -10,6 +10,8 @@ def serialize
           type: model.fee_type,
           code: model.item_code,
           name: model.item_name,
+          lago_item_id: model.item_id,
+          lago_item_type: model.item_type,
```
rsempe, line 14: are we sure we want `lago_item_type` and not just `item_type`?

app/serializers/v1/fee_serializer.rb
```diff
@@ -19,6 +21,8 @@ def serialize
         total_amount_currency: model.amount_currency,
         units: model.units,
         events_count: model.events_count,
+        invoice_id: model.invoice_id,
```
rsempe, line 24: `lago_invoice_id`? (not really fan of prepending by `lago_` but that's what we currently have for other ids, no?)

rsempe: For me, the preferable solution would be to prepend by `external_` each time it's not related to a Lago id...

Then, no need to prepend by `lago_` each time... something maybe to think about in a future breaking change refactor 😅

rsempe (approved):

PR #935, opened 2023-03-24
feat(instant_charge): add API end-point to update fee payment_status

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds a new `PUT /api/v1/fees/:id` route to update the payment status of a fee. It delegates the logic to a `Fees::UpdateService`

### DIFF
--- a/spec/requests/api/v1/fees_spec.rb
+++ b/spec/requests/api/v1/fees_spec.rb
@@ -73,4 +73,49 @@ RSpec.describe Api::V1::FeesController, type: :request do
       end
     end
   end
+
+  describe 'PUT /fees/:id' do
+    let(:customer) { create(:customer, organization:) }
+    let(:subscription) { create(:subscription, customer:) }
+    let(:fee) { create(:charge_fee, fee_type: 'instant_charge', subscription:, invoice: nil) }
+
+    let(:update_params) { { payment_status: 'succeeded' } }
+
+    it 'updates the fee' do
+      put_with_token(organization, "/api/v1/fees/#{fee.id}", fee: update_params)
+
+      aggregate_failures do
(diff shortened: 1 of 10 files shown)


lovrocolic (approved):

spec/requests/api/v1/fees_spec.rb
```diff
@@ -73,4 +73,45 @@
...
+        expect(json[:fee][:amount_currency]).to eq(fee.amount_currency)
+        expect(json[:fee][:vat_amount_cents]).to eq(fee.vat_amount_cents)
+        expect(json[:fee][:vat_amount_currency]).to eq(fee.vat_amount_currency)
+        expect(json[:fee][:units]).to eq(fee.units.to_s)
+        expect(json[:fee][:events_count]).to eq(fee.events_count)
+        expect(json[:fee][:payment_status]).to eq(fee.payment_status)
+        expect(json[:fee][:created_at]).to eq(fee.created_at&.iso8601)
+        expect(json[:fee][:succeeded_at]).to eq(fee.succeeded_at&.iso8601)
+        expect(json[:fee][:failed_at]).to eq(fee.failed_at&.iso8601)
+        expect(json[:fee][:refunded_at]).to eq(fee.refunded_at&.iso8601)
```
rsempe, line 105: is that possible to have only one expectation here? 😅  something like `have_attributes(...)` but for your json variable

spec/requests/api/v1/fees_spec.rb
```diff
@@ -73,4 +73,45 @@
...
+      end
+    end
+
+    context 'when fee does not exsits' do
```
rsempe, line 109:
```suggestion
    context 'when fee does not exist' do
```

PR #939, opened 2023-03-27
feat(instant_charge): Add GET /api/v1/fees route

vincent-pochet:
## Roadmap Task

## Context

## Description

This PR adds a new `GET /api/v1/fees` to retrieve fees. It accepts a lot of filter attributes

### DIFF
--- a/app/controllers/api/v1/fees_controller.rb
+++ b/app/controllers/api/v1/fees_controller.rb
@@ -24,10 +24,53 @@ module Api
         end
       end
 
+      def index
+        result = FeesQuery.call(
+          organization: current_organization,
+          pagination: BaseQuery::Pagination.new(
+            page: params[:page],
+            limit: params[:per_page] || PER_PAGE,
+          ),
+          filters: BaseQuery::Filters.new(index_filters),
+        )
+
+        if result.success?
+          render(
+            json: ::CollectionSerializer.new(
+              result.fees,
+              ::V1::FeeSerializer,
+              collection_name: 'fees',
+              meta: pagination_metadata(result.fees),
+            ),
+          )
+        else
+          render_error_response(result)
+        end
+      end
+
       private
 
       def update_params
         params.require(:fee).permit(:payment_status)
+      end
+
+      def index_filters
+        params.permit(
+          :fee_type,
+          :payment_status,
+          :external_subscription_id,
+          :external_customer_id,
+          :billable_metric_code,
+          :currency,
+          :created_at_from,
+          :created_at_to,
+          :failed_at_from,
+          :failed_at_to,
+          :succeeded_at_from,
+          :succeeded_at_to,
+          :refunded_at_from,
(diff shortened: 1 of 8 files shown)


spec/queries/fees_query_spec.rb
```diff
@@ -0,0 +1,381 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe FeesQuery, type: :query do
+  subject(:fees_query) { described_class.new(organization:, pagination:, filters:) }
+
+  let(:organization) { create(:organization,) }
```
jdenquin, line 8: Style/TrailingCommaInArguments: Avoid comma after the last parameter of a method call, unless each item is on its own line. (https://rubystyle.guide#no-trailing-params-comma)

app/controllers/api/v1/fees_controller.rb
```diff
@@ -4,17 +4,74 @@ module Api
...
+          :billable_metric_code,
+          :currency,
+          :created_at_from,
+          :created_at_to,
+          :failed_at_from,
+          :failed_at_to,
+          :succeeded_at_from,
+          :succeeded_at_to,
+          :refunded_at_from,
+          :refunded_at_to,
```
rsempe, line 72: Are we sure we really need to have these filters?

I mean, we can use the combination of `payment_status` and `created_at`, no?

app/queries/fees_query.rb
```diff
@@ -0,0 +1,120 @@
...
+        .raise_if_error!
+    end
+
+    scope.where(fee_type: filters.fee_type)
+  end
+
+  def with_payment_status(scope)
+    unless Fee::PAYMENT_STATUS.include?(filters.payment_status.to_sym)
+      result.single_validation_failure!(field: :payment_status, error_code: 'value_is_invalid')
+        .raise_if_error!
```
rsempe, line 59: Do we really want to fail in that case? 🤔

app/queries/fees_query.rb
```diff
@@ -0,0 +1,120 @@
...
+
+  def with_billable_metric_code(scope)
+    scope.joins(:billable_metric)
+      .where(billable_metric: { code: filters.billable_metric_code })
+  end
+
+  def with_fee_type(scope)
+    unless Fee::FEE_TYPES.include?(filters.fee_type.to_sym)
+      result.single_validation_failure!(field: :fee_type, error_code: 'value_is_invalid')
+        .raise_if_error!
```
rsempe, line 50: same here

rsempe (approved):
👏

lovrocolic (approved):

app/controllers/api/v1/fees_controller.rb
vincent-pochet: Yep it's one of the request of some customers. they want to be able to retrive fees based on their payment|success/failure/refund date, and not only the creation date

PR #942, opened 2023-03-29
feat(password_reset): Add password request, create and reset flows

ansmonjol:
## Context

Currently, users cannot update their password through the UI.

## Description

This PR adds the ability for then to request a password reset and to change their password by themself

Mail looks like this

image


app/services/password_resets/create_service.rb
```diff
@@ -0,0 +1,31 @@
...
+      @user = user
+
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: 'user') if user.blank?
+
+      password_reset = PasswordReset.create!(
+        user_id: user.id,
```
lovrocolic, line 15: just `user:` in enough

app/services/password_resets/create_service.rb
```diff
@@ -0,0 +1,31 @@
...
+        token: SecureRandom.hex(20),
+        expire_at: DateTime.now + 30.minutes,
```
lovrocolic, line 17: I guess you should use `Time.current + 30.minutes`

spec/graphql/mutations/password_resets/reset_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    <<~GQL
+      mutation($input: ResetPasswordInput!) {
+        resetPassword(input: $input) {
+          token
+        }
+      }
+    GQL
+  end
+
+  it 'creates an invite for a new user' do
```
lovrocolic, line 19: title is incorrect, probably leftover:)

app/graphql/mutations/password_resets/reset.rb
```diff
@@ -0,0 +1,23 @@
...
+      graphql_name 'ResetPassword'
+      description 'Reset password for user and log in'
+
+      argument :new_password, String, required: true
+      argument :token, String, required: true
+
+      type Types::Payloads::LoginUserType
+
+      def resolve(new_password:, token:)
+        password_reset = PasswordReset.where('expire_at > ?', DateTime.now).find_by(token:)
```
lovrocolic, line 15: I guess it should be: `PasswordReset.where('expire_at < ?', DateTime.now).find_by(token:)` Probably you also miss test case for this.

app/graphql/resolvers/password_reset_resolver.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Resolvers
+  class PasswordResetResolver < Resolvers::BaseResolver
```
lovrocolic, line 4: Do we need this resolver, where is it used? 🤔

spec/graphql/mutations/password_resets/create_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    <<~GQL
+      mutation($input: CreatePasswordResetInput!) {
+        createPasswordReset(input: $input) {
+          id
+        }
+      }
+    GQL
+  end
+
+  it 'creates an invite for a new user' do
```
lovrocolic, line 19: title is incorrect:)

app/services/password_resets/create_service.rb
ansmonjol: Updated ! Thanks

app/services/password_resets/create_service.rb
ansmonjol: Yeah, read some article in the meantime about why :)

Changed it everywhere it was used in this PR

spec/graphql/mutations/password_resets/reset_spec.rb
ansmonjol: Good catch ! I did moved the query inside the service and added tests for the expirtation 🙏

app/graphql/mutations/password_resets/reset.rb
ansmonjol: Done !

app/graphql/resolvers/password_reset_resolver.rb
ansmonjol: Yeah we need it on the FE side.

It's called when we access the reset password page to prevent using an expired password reset token

spec/graphql/mutations/password_resets/create_spec.rb
ansmonjol: Fixed !

spec/services/password_resets/reset_service_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    let(:membership) { create(:membership, user:) }
+    let(:password_reset) { create(:password_reset, user: membership.user) }
+    let(:reset_args) do
+      {
+        token: password_reset.token,
+        new_password: 'HelloLago!2',
+      }
+    end
+
+    context 'when the reset password service' do
```
lovrocolic, line 19: This context is not needed

spec/services/password_resets/reset_service_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      it 'changes the user password' do
```
lovrocolic, line 20: I would check here if new password is valid, probably using `authenticate` method

spec/services/password_resets/reset_service_spec.rb
ansmonjol: Removed :)

spec/services/password_resets/reset_service_spec.rb
ansmonjol: Done ! Good one

PR #951, opened 2023-04-03
feat: Turn balance float amounts into cents

vincent-pochet:
## Context

In our data model, amounts are usually integers in cents (e.g. `amount_cents` or `vat_amount_cents`).

However, in wallet, amount attributes are defined in float and exposed as string in the two API

This difference makes it difficult for users to use this information for calculation purposes (e.g. cannot easily deduct `amount_cents` from `balance` without converting the attribute into an integer).

## Description

This PR changes turned wallet fields into bigint and adds currency columns to match the amount and currency handling from other models.
- `balance_cents` / `balance_currency`
- `consumed_amount_cents` / `consumed_amount_currenty`

For compatibility purpose, all float fields are still exposed in public API but are deprecated.
This changes also requires an update of the front-end to adapt to the new amount fields

NOTE: the same changes will be applied to wallet_transaction that suffers from the same limitations.

### DIFF
--- a/db/migrate/20230403093407_add_balance_cents_to_wallets.rb
+++ b/db/migrate/20230403093407_add_balance_cents_to_wallets.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+class AddBalanceCentsToWallets < ActiveRecord::Migration[7.0]
+  def change
+    change_table :wallets, bulk: true do |t|
+      t.bigint :balance_cents
+      t.string :balance_currency
+
+      t.bigint :consumed_amount_cents
+      t.string :consumed_amount_currenty
+    end
+
+    Wallet.find_each do |wallet|
+      currency = Money::Currency.new(wallet.attributes['currency'])
+
+      wallet.update!(
+        balance_cents: (wallet.attributes['balance'] * currency.subunit_to_unit).to_i,
+        balance_currency: currency.iso_code,
+        consumed_amount_cents: (wallet.attributes['consumed_amount'] * currency.subunit_to_unit).to_i,
+        consumed_amount_currenty: currency.iso_code,
+      )
+    end
+
+    change_column_default :wallets, :balance_cents, from: nil, to: 0
+    change_column_null :wallets, :balance_cents, false
+    change_column_null :wallets, :balance_currency, false
+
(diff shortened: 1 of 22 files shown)


db/migrate/20230403093407_add_balance_cents_to_wallets.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+class AddBalanceCentsToWallets < ActiveRecord::Migration[7.0]
+  def change
+    add_column :wallets, :balance_cents, :bigint
```
jdenquin, line 5: Rails/BulkChangeTable: You can use `change_table :wallets, bulk: true` to combine alter queries. (https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_table, https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Table.html)

db/migrate/20230403093407_add_balance_cents_to_wallets.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class AddBalanceCentsToWallets < ActiveRecord::Migration[7.0]
+  def change
+    change_table :wallets, bulk: true do |t|
+      t.bigint :balance_cents
+      t.string :wallets, :balance_currency
```
rsempe, line 7:
```suggestion
      t.string :balance_currency
```

no?

db/schema.rb
```diff
@@ -597,19 +597,21 @@
...
     t.decimal "consumed_credits", precision: 30, scale: 5, default: "0.0", null: false
     t.datetime "expiration_at", precision: nil
     t.datetime "last_balance_sync_at", precision: nil
     t.datetime "last_consumed_credit_at", precision: nil
     t.datetime "terminated_at", precision: nil
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.decimal "consumed_amount", precision: 30, scale: 5, default: "0.0"
+    t.bigint "balance_cents", default: 0, null: false
+    t.string "wallets"
```
rsempe, line 611: 😅

vincent-pochet: Fixed! :facepalm:

rsempe (approved):

PR #955, opened 2023-04-05
fix: Do not bill subscription if not active

rsempe:
The goal of this PR is to not bill the subscription if pending.

Here is the following issue we want to fix in that PR:
- when a user is creating a subscription via API
- and plan is pay in advance
- with a subscription date in the future but today
- we don't want to enqueue a `BillSubscriptionJob`

### DIFF
--- a/spec/services/subscriptions/create_service_spec.rb
+++ b/spec/services/subscriptions/create_service_spec.rb
@@ -137,10 +137,19 @@ RSpec.describe Subscriptions::CreateService, type: :service do
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/services/subscriptions/create_service_spec.rb
```diff
@@ -137,6 +137,15 @@
       end
     end
```
jdenquin, line 138: Layout/CommentIndentation: Incorrect indentation detected (column 1 instead of 0).

vincent-pochet (approved):

PR #958, opened 2023-04-06
fix: Do not create an invoice when recurring and no active subscriptions

rsempe:
The goal of this PR is to fix potentially duplicated invoices.

Indeed, when `BillSubscriptionJob` is enqueued but not yet performed (due to the 1-50 minutes wait) and meantime the subscription is terminated, two invoices are created.


vincent-pochet (approved):
:rocket:

PR #967, opened 2023-04-12
feat(billable_metric): allow to filter by aggregation_type

ansmonjol:
## Context

We're refactoring the plan view on FE side.

It does required to have 2 inputs to search for billable metric. One in the "normal" charge section and one in the Instant charges section.

The instant charge section should only retrieve BM according to their `aggregation_type` value.

## Description

This PR updates the BM query to allow to filter by an array of `aggregation_type`

### DIFF
--- a/app/graphql/resolvers/billable_metrics_resolver.rb
+++ b/app/graphql/resolvers/billable_metrics_resolver.rb
@@ -12,9 +12,11 @@ module Resolvers
     argument :limit, Integer, required: false
     argument :search_term, String, required: false
 
+    argument :aggregation_types, [Types::BillableMetrics::AggregationTypeEnum], required: false
+
     type Types::BillableMetrics::Object.collection_type, null: false
 
-    def resolve(ids: nil, page: nil, limit: nil, search_term: nil)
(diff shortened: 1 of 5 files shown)


app/graphql/resolvers/billable_metrics_resolver.rb
```diff
@@ -12,14 +12,17 @@ class BillableMetricsResolver < GraphQL::Schema::Resolver
...
     type Types::BillableMetrics::Object.collection_type, null: false
 
-    def resolve(ids: nil, page: nil, limit: nil, search_term: nil)
+    def resolve(ids: nil, page: nil, limit: nil, search_term: nil, aggregation_types: nil)
       validate_organization!
 
       query = ::BillableMetricsQuery.new(organization: current_organization)
       result = query.call(
         search_term:,
+        aggregation_types:,
```
lovrocolic, line 25: Should we move it in `filters` field? Wdyt?

ansmonjol: yeah tbh I was not sure what to put into filter or not, but yeah it makes more sense in a way ! Will do

PR #979, opened 2023-04-17
feat(coupon): expose invoice version & split PDF

vincent-pochet:
## Context

Since already invoices created with the current implementation will still have to be generated or impacted by credit notes, we need a way to quickly identify which logic was applied when creating the invoice.

## Description

This PR uses the new `invoice.version_number` field to render the right PDF template and exposes it into API and GraphQL schema

### DIFF
--- a/app/graphql/types/invoices/object.rb
+++ b/app/graphql/types/invoices/object.rb
@@ -33,7 +33,7 @@ module Types
       field :created_at, GraphQL::Types::ISO8601DateTime, null: false
       field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
 
-      field :legacy, Boolean, null: false
+      field :version_number, Integer, null: false
 
       field :subscriptions, [Types::Subscriptions::Object]
       field :invoice_subscriptions, [Types::InvoiceSubscription::Object]
--- a/app/serializers/v1/legacy/invoice_serializer.rb
+++ b/app/serializers/v1/legacy/invoice_serializer.rb
@@ -5,11 +5,18 @@ module V1
     class InvoiceSerializer < ModelSerializer
       def serialize
         {
+          legacy:,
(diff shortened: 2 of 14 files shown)


app/serializers/v1/legacy/invoice_serializer.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module V1
+  module Legacy
+    class InvoiceSerializer < ModelSerializer
+      def serialize
+        {
+          legacy: legacy,
```
jdenquin, line 8: Style/HashSyntax: Omit the hash value. (https://rubystyle.guide#hash-literals)

app/graphql/types/invoices/object.rb
```diff
@@ -32,7 +32,9 @@ class Object < Types::BaseObject
       field :created_at, GraphQL::Types::ISO8601DateTime, null: false
       field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
 
+      # NOTE: will be removed when front-end will implement invoice versioning
```
ansmonjol, line 35: Well, we do not use this `legacy` attribute in our code so you can probably remove it already

ansmonjol: What is it used for on your side ? I assume it's for calculation logic, but wondering if it's either a issue on old invoices on UI side, or if it does not have any UI impact at all

vincent-pochet: @ansmonjol 
It's used on old invoice where coupons, and prepaid credit were applied before VAT and to prevent creation of credit notes on this invoices.
It probably means there is an issue with this on the front-end, but since these invoice are now quite old and probably attached to test accounts, I think it's not a big deal...

rsempe (approved):

PR #981, opened 2023-04-17
feat(invoice): Add new coupons_amount_cents to invoice

vincent-pochet:
## Context

## Description

This PR adds a new `coupons_amount_cents` column to the `invoices` table. It will store the sum of coupon credit amount.

### DIFF
--- a/app/services/invoices/calculate_fees_service.rb
+++ b/app/services/invoices/calculate_fees_service.rb
@@ -34,7 +34,7 @@ module Invoices
 
         Invoices::ComputeAmountsFromFees.call(invoice:)
         create_credit_note_credit if should_create_credit_note_credit?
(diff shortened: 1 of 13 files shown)


app/services/invoices/calculate_fees_service.rb
```diff
@@ -165,13 +165,6 @@ def create_credit_note_credit
       refresh_amounts(credit_amount_cents: credit_result.credits.sum(&:amount_cents)) if credit_result.credits
     end
 
-    def create_applied_coupons_credit
```
lovrocolic, line 168: Worked a lot on coupons and haven't noticed that we don't use this one :)

lovrocolic (approved):

PR #983, opened 2023-04-17
feat(invoice): Add new prepaid_credit_amount_cents to invoice

vincent-pochet:
## Context

## Description

This PR adds a new `prepaid_credit_amount_cents` column to the `invoices` table. It will store the sum of wallet transaction amount.

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -28,6 +28,7 @@ class Invoice < ApplicationRecord
   monetize :fees_amount_cents, with_model_currency: :amount_currency
   monetize :coupons_amount_cents, with_model_currency: :amount_currency
   monetize :vat_amount_cents
+  monetize :prepaid_credit_amount_cents, with_model_currency: :amount_currency
   monetize :credit_amount_cents
   monetize :credit_notes_amount_cents, with_model_currency: :amount_currency
   monetize :total_amount_cents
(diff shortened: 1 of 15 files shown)


spec/models/invoice_spec.rb
```diff
@@ -164,42 +164,15 @@
...
-  end
-
   describe '#subtotal_before_prepaid_credits' do
     let(:customer) { create(:customer) }
-    let(:invoice) { create(:invoice, customer:, amount_cents: 555, organization: customer.organization) }
-    let(:wallet) { create(:wallet, customer:, balance: 10.0, credits_balance: 10.0) }
-    let(:wallet_transaction) do
-      create(:wallet_transaction, invoice:, wallet:, amount: 1, credit_amount: 1)
+    let(:invoice) do
+      create(:invoice, customer:, amount_cents: 555, organization: customer.organization, prepaid_credit_amount_cents: 100)
```
jdenquin, line 170: Layout/LineLength: Line is too long. [123/120] (https://rubystyle.guide#max-line-length)

app/models/invoice.rb
```diff
@@ -115,20 +115,8 @@ def subscription_amount_cents
   end
   alias subscription_amount_currency currency
 
-  def wallet_transaction_amount_cents
```
lovrocolic, line 118: Not sure what happened with this logic after removing it?

vincent-pochet: It is updated directly in https://github.com/getlago/lago-api/pull/983/files#diff-79d4322ca48862bf4414fb50260475427521f6d7ff238b38e6411abc57946714R34, so we have not need to recompute it each time

lovrocolic (approved):

db/migrate/20230417140356_add_prepaid_credit_amount_cents_to_invoices.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class AddPrepaidCreditAmountCentsToInvoices < ActiveRecord::Migration[7.0]
+  def change
+    add_column :invoices, :prepaid_credit_amount_cents, :bigint, null: false, default: 0
+
+    reversible do |dir|
+      dir.up do
+        currency_list = WalletTransaction.joins(:wallet).pluck('DISTINCT(wallets.balance_currency)')
+        next if currency_list.blank?
```
jdenquin, line 10: Layout/EmptyLineAfterGuardClause: Add empty line after guard clause.

PR #987, opened 2023-04-19
fix(customers): fix updating data that is sent in billing_configuration payload

lovrocolic:


app/services/customers/create_service.rb
```diff
@@ -176,7 +176,13 @@ def handle_api_billing_configuration(customer, params, new_customer)
 
       return if customer.payment_provider.nil?
 
-      create_or_update_provider_customer(customer, billing)
+      update_provider_customer = (billing || {})[:provider_customer_id].present?
+      update_provider_customer ||=
+        customer.__send__("#{customer.payment_provider}_customer")&.provider_customer_id.present?
+
+      if billing.key?(:payment_provider) && update_provider_customer
```
jdenquin, line 183: Style/GuardClause: Use a guard clause (`return unless billing.key?(:payment_provider) && update_provider_customer`) instead of wrapping the code inside a conditional expression. (https://rubystyle.guide#no-nested-conditionals)

app/services/customers/create_service.rb
```diff
@@ -176,6 +176,12 @@ def handle_api_billing_configuration(customer, params, new_customer)
 
       return if customer.payment_provider.nil?
 
+      update_provider_customer = (billing || {})[:provider_customer_id].present?
+      update_provider_customer ||=
+        customer.__send__("#{customer.payment_provider}_customer")&.provider_customer_id.present?
```
rsempe, line 181: What about extracting this logic and avoiding the meta-programation here?

Something like:
```ruby
customer.provider_customer&.provider_customer_id.present?

def provider_customer # In Customer model?
  if payment_provider == 'stripe'
    stripe_customer
  elsif payment_provider == 'gocardless'
    gocardless_customer
  end
end
```

lovrocolic: updated

rsempe: Cool - we can add a unit test for the model as well no? 😅

rsempe (approved):

app/services/customers/create_service.rb
lovrocolic: Tests, what? 😄 Added test cases!

PR #990, opened 2023-04-20
feat(invoice): Remove invoices#amount_cents and invoices#credit_amount_cents

vincent-pochet:
## Context

## Description

This PR removes `amount_cents` and `credit_amount_cents` from invoices.

It keeps exposing them in InvoiceSerializer not to break exsisting API contract


spec/models/credit_note_spec.rb
```diff
@@ -125,7 +125,7 @@
     end
 
     it 'returns the list of subscription ids' do
-      expect(credit_note.subscription_ids).to eq([subscription_fee.subscription_id, charge_fee.subscription_id])
+      expect(credit_note.subscription_ids).to contain_exactly(subscription_fee.subscription_id, charge_fee.subscription_id)
```
jdenquin, line 128: Layout/LineLength: Line is too long. [123/120] (https://rubystyle.guide#max-line-length)

PR #1016, opened 2023-05-05
feat(tax_rates): Add tax rate graphql resolvers

rsempe:
## Roadmap Task

## Context

## Description

- add the tax rate graphql resolver
- add the tax rates graphql resolver

### DIFF
--- a/app/queries/tax_rates_query.rb
+++ b/app/queries/tax_rates_query.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+class TaxRatesQuery < BaseQuery
+  def call(search_term:, page:, limit:, filters: {})
+    @search_term = search_term
+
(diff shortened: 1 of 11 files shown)


app/queries/tax_rates_query.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class TaxRatesQuery < BaseQuery
```
ansmonjol, line 3: Non blocking comment, 
It's not specified in specs but I'm wondering if we want to add the ability to search by rate value.
Like, I want all 20% tax rates to be returned so I type `20`, WDYT?
Again, just thought about it and don't want to add more work on your plate 😬

ansmonjol (approved):

vincent-pochet (approved):

PR #1025, opened 2023-05-11
feat(tax_rate): Assign a tax rate to an organization via API

vincent-pochet:
## Roadmap Task

## Context

## Description

The goal of this PR is to expose the `TaxRate#applied_by_default` attribute in API in order to turn it on/off for the organization


spec/requests/api/v1/tax_rates_spec.rb
```diff
@@ -12,6 +12,7 @@
         code: 'tax_rate_code',
         value: 20.0,
         description: 'tax_rate_description',
+        applied_by_default: false,
```
rsempe, line 15: maybe we can put `true` here as `false` is the default value?

PR #1035, opened 2023-05-15
hotfix: Fix monthly charges for yearly plan when subscription starts on last day of the month

vincent-pochet:
## Context

An issue have been identified with yearly plan billed in advance with charges billed monthly when the subscription starts on the last day of a month with 31 days.

If  the next month is shorter than the subscription one, the subscription fee is re-billed on the second month...

## Description

The fix is to ensure that in this situation when subscription was started on the last day of the month, we take the last day of current month as billing day.

### DIFF
--- a/spec/scenarios/billing_scenario_spec.rb
+++ b/spec/scenarios/billing_scenario_spec.rb
@@ -258,4 +258,72 @@ describe 'Billing Scenarios', :scenarios, type: :request do
       expect(subscription.reload.invoices.count).to eq(1)
     end
   end
+
+  context 'when plan is billed yearly with monthly charges' do
+    let(:plan) do
+      create(
+        :plan,
+        organization:,
+        amount_cents: 5_000_000,
+        amount_currency: 'GBP',
+        interval: 'yearly',
+        pay_in_advance: true,
+        bill_charges_monthly: true,
+      )
+    end
+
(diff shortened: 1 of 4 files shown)


spec/scenarios/billing_scenario_spec.rb
```diff
@@ -258,4 +258,72 @@
       expect(subscription.reload.invoices.count).to eq(1)
     end
   end
+
+  context 'when plan is billed yearly with monthly charges' do
```
rsempe, line 262: 😍

spec/services/subscriptions/dates/monthly_service_spec.rb
```diff
@@ -130,6 +130,26 @@
...
+        it 'returns the day current billing day' do
+          expect(result).to eq('2021-04-30 00:00:00 UTC')
+        end
+      end
+
+      context 'when plan is in arrear and date is on the last day of month' do
+        let(:billing_at) { DateTime.parse('30 apr 2021') }
+        let(:subscription_at) { DateTime.parse('31 mar 2021') }
+
+        it 'returns the day current billing day' do
```
rsempe, line 149:
```suggestion
        it 'returns the last day of the previous month' do
```
?

rsempe (approved):

PR #1036, opened 2023-05-16
feat(tax_rates): Ability to assign and unassign tax rate via graphql

rsempe:
## Roadmap Task

## Context

## Description

- create applied tax rate via graphql
- destroy applied tax rate via graphql


app/graphql/types/customers/object.rb
```diff
@@ -32,6 +32,7 @@ class Object < Types::BaseObject
       field :zipcode, String, null: true
 
       field :metadata, [Types::Customers::Metadata::Object], null: true
+      field :tax_rates, [Types::TaxRates::Object], null: true
```
vincent-pochet, line 35: Should we expose the `applied_tax_rates`, rather than the `tax_rates`? Otherwise, I'm not sure how the front-end will be able to fetch them (for destroy purpose in particular)

vincent-pochet: Also, maybe we should a request for these attributes somewhere in the specs so that we can validate it serializes well

vincent-pochet (approved):

ansmonjol (approved):

PR #1037, opened 2023-05-16
feat(admin): Add invoice regenerate endpoint

jdenquin:
## Context

We want to be able to regenerate an invoice PDF with the admin endpoints

## Description

- Add the `POST /admin/invoices/:id/regenerate` endpoint and logic

### DIFF
--- a/spec/requests/admin/invoices_spec.rb
+++ b/spec/requests/admin/invoices_spec.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Admin::InvoicesController, type: [:request, :admin] do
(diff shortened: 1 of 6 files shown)


spec/requests/admin/invoices_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Admin::InvoicesController, type: [:request, :admin] do
```
jdenquin, line 5: RSpec/FilePath: Spec path should end with `admin/invoices_controller*_spec.rb`. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FilePath)

PR #1053, opened 2023-05-25
feat (charges): Invoiceable pay in advance charges

lovrocolic:
## Context

Currently it is not possible to pay charges in advance

## Description

This PR covers the case when charge `pay_in_advance` field is set to `true` and `invoiceable` field is also set to `true`. In that case, when event is ingested, new invoice with depending fees needs to be created.


app/services/events/create_service.rb
```diff
@@ -92,17 +100,31 @@ def handle_persisted_event
...
+      @non_invoiceable_charges ||= charges.where(invoiceable: false)
+    end
+
+    def invoiceable_charges
+      @invoiceable_charges ||= charges.where(invoiceable: true)
+    end
+
+    def applicable_event?
+      return true unless billable_metric.sum_agg?
+      return true unless event.properties[billable_metric.field_name]&.to_i.negative?
```
jdenquin, line 121: Lint/SafeNavigationChain: Do not chain ordinary method call after safe navigation operator.

app/serializers/v1/fee_serializer.rb
```diff
@@ -80,5 +80,15 @@ def taxes
     def legacy_values
       ::V1::Legacy::FeeSerializer.new(model).serialize
     end
+
+    def calculate_pay_in_advance
```
vincent-pochet, line 84: What about simply renaming this one `pay_id_advance`?

app/serializers/v1/fee_serializer.rb
```diff
@@ -16,8 +16,8 @@ def serialize
           lago_item_id: model.item_id,
           item_type: model.item_type,
         },
-        pay_in_advance: model.pay_in_advance,
-        invoiceable: model.charge&.invoiceable || false,
+        pay_in_advance: calculate_pay_in_advance,
+        invoiceable: model.charge? ? model.charge&.invoiceable : true,
```
vincent-pochet, line 20: Maybe this one could be exported into a proper `invoiceable` method in this serializer`?

app/services/invoices/create_pay_in_advance_charge_service.rb
```diff
@@ -0,0 +1,133 @@
+# frozen_string_literal: true
+
+module Invoices
+  class CreatePayInAdvanceChargeService < BaseService
+    def initialize(charge:, event:, timestamp:)
+      @charge = charge
+      @event = event
+      @timestamp = timestamp
+
+      super(nil)
```
vincent-pochet, line 10: I think you can get rid of the `nil` argument

app/services/invoices/create_pay_in_advance_charge_service.rb
```diff
@@ -0,0 +1,133 @@
...
+        InvoiceSubscription.create!(
+          invoice:,
+          subscription: event.subscription,
+          recurring: false,
+        )
+
+        create_fees(invoice)
+
+        invoice.fees_amount_cents = invoice.fees.sum(:amount_cents)
+        invoice.sub_total_excluding_taxes_amount_cents = invoice.fees.sum(:amount_cents)
```
vincent-pochet, line 36: Maybe you could reuse `invoice.fees_amount_cents` to reduce the number of query sent to the DB?

vincent-pochet (approved):
Few small comments regarding the code, but looks good :rocket: 

On a more general aspects, I think we are starting to have a lot of duplication on the invoice creation logic across services, and probably we should extract email delivery, segment tracking, credit note and wallet credit creation into proper services.

app/serializers/v1/fee_serializer.rb
lovrocolic: fixed

app/services/invoices/create_pay_in_advance_charge_service.rb
lovrocolic: OFC, nice one 👍

PR #1059, opened 2023-05-26
fix (coupons): adapt applied coupon logic

lovrocolic:


vincent-pochet (approved):
Look good! I was also thinking at adding a validation on the `invoice#total_amount_cents` to make sure it's always greater than or equal to 0 and so that we have no way to emit negative invoices (even if it means failing jobs in some cases). What do you think?

lovrocolic:
> Look good! I was also thinking at adding a validation on the `invoice#total_amount_cents` to make sure it's always greater than or equal to 0 and so that we have no way to emit negative invoices (even if it means failing jobs in some cases). What do you think?

Yep, make sense! Added!

PR #1060, opened 2023-05-26
feat (charges): add new template for pay_in_advance charges

lovrocolic:
This PR adds a new template for invoices that are created with ingesting an event (case `pay_in_advance=true` and `invoiceable=true`)

### DIFF
--- a/app/services/invoices/generate_pdf_service.rb
+++ b/app/services/invoices/generate_pdf_service.rb
@@ -43,6 +43,8 @@ module Invoices
     def template
       if invoice.one_off?
         'invoices/one_off'
+      elsif charge?
+        'invoices/charge'
       else
         "invoices/v#{invoice.version_number}"
       end
@@ -54,6 +56,10 @@ module Invoices
 
     def should_generate_pdf?
       context == 'admin' || invoice.file.blank?
+    end
+
+    def charge?
+      invoice.fees.all? { |f| f.pay_in_advance? }
     end
   end
 end
--- a/app/services/invoices/generate_pdf_service.rb
+++ b/app/services/invoices/generate_pdf_service.rb
@@ -59,7 +59,7 @@ module Invoices
     end
 
     def charge?
-      invoice.fees.all? { |f| f.pay_in_advance? }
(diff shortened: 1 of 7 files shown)


app/services/invoices/generate_pdf_service.rb
```diff
@@ -55,5 +57,9 @@ def should_send_webhook?
     def should_generate_pdf?
       context == 'admin' || invoice.file.blank?
     end
+
+    def charge?
+      invoice.fees.all? { |f| f.pay_in_advance? }
```
jdenquin, line 62: Style/SymbolProc: Pass `&:pay_in_advance?` as an argument to `all?` instead of a block.

app/views/templates/invoices/charge.slim
```diff
@@ -0,0 +1,484 @@
...
+            td.body-2 = I18n.t('invoice.item')
+            td.body-2 style="text-align: left;"
+            td.body-2 style="text-align: left;" = I18n.t('invoice.unit')
+            td.body-2 style="text-align: right;" = I18n.t('invoice.amount')
+          - fees.each do |fee|
+            tr
+              td.body-1 = fee.billable_metric.name
+              td.body-2 style="text-align: left;"
+              td.body-2 style="text-align: left;" = fee.units
+              td.body-2 style="text-align: right;" = fee.amount&.format(format: I18n.t('money.format'), decimal_mark: I18n.t('money.decimal_mark'), thousands_separator: I18n.t('money.thousands_separator'))
```
vincent-pochet, line 426: We are repeating `format(format: I18n.t('money.format'), decimal_mark: I18n.t('money.decimal_mark'), thousands_separator: I18n.t('money.thousands_separator'))` all over the templates. I'm wondering if we could find a way to move it in an helper.

vincent-pochet (approved):
I trust you for the template part :)

Only one small suggestion to be a bit more DRY ;) .

app/views/templates/invoices/charge.slim
lovrocolic: Added small formatter;)

PR #1066, opened 2023-05-29
feat(tax): Return applied taxes via graphql

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to return applied taxes for fees, invoices and credit notes on graphql.

### DIFF
--- a/app/graphql/types/invoices/applied_taxes/object.rb
+++ b/app/graphql/types/invoices/applied_taxes/object.rb
@@ -5,22 +5,9 @@ module Types
     module AppliedTaxes
       class Object < Types::BaseObject
         graphql_name 'InvoiceAppliedTax'
-
-        field :id, ID, null: false
+        implements Types::Taxes::AppliedTax
 
         field :invoice, Types::Invoices::Object, null: false
-        field :tax, Types::Taxes::Object, null: false
-
-        field :tax_code, String, null: false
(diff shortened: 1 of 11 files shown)


ansmonjol (approved):

app/graphql/types/invoices/applied_taxes/object.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Types
+  module Invoices
+    module AppliedTaxes
+      class Object < Types::BaseObject
+        graphql_name 'InvoiceAppliedTax'
+
+        field :id, ID, null: false
```
vincent-pochet, line 9: A possible approach to share some code between all those `AppliedTaxes` could be to use GraphQL interfaces:

```ruby
module Types
  module Taxes
    module AppliedTax
      include Types::BaseInterface

      field ...
    end
  end
end
```

And then use them in the different types:

```ruby
module Types
  module Invoices
    module AppliedTaxes
      class Object < Types::BaseObject
        graphql_name 'InvoiceAppliedTax'
        implements Types::Taxes::AppliedTaxes

        field :invoice, Types::Invoices::Object, null: false
      end
    end
  end
end
```

vincent-pochet (approved):

PR #1074, opened 2023-05-31
fix: Avoid duplicate invoices on anniversary and timezone

rsempe:
When a **timezone** is defined at the customer or organization level
And billing time is **anniversary**
(no matter what is the interval (weekly, monthly, yearly))
An invoice can be duplicated depending when the billing is made.

Example:
Customer timezone: Europe/Paris (UTC +2h)
May 23: Subscription created (weekly anniversary)
May 30: Billing (20h UTC => 22h in timezone) => 1 Invoice created
May 30: Billing (23 UTC => may 31 - 1h in timezone) => 1 Invoice created

How to fix it?
By always looking at the customer applicable timezone when fetching billable subscriptions and the one already billed.


jdenquin (approved):

lovrocolic:
nice one! 🚀 Do you think that we should also add custom validation on the invoice model level? Probably it can check if there is already invoice related to the same subscription for recurring case, created in less than {billing_period} days?

rsempe:
@lovrocolic yeah probably, that's something we can think about.. not really easy as it depends on timezone, but yeah 👍

PR #1081, opened 2023-06-02
feat(applied_coupon): expose coupon name in serializer

ansmonjol:
## Context

The `applied_coupon` object does not include `coupon_name`, so we need to use `coupon_code` and perform `GET /api/v1/coupons/:code` to retrieve the name of the applied coupon, which is not convenient.

## Description

This PR adds the `coupon_name` directly to the `applied_coupon_serializer`


rsempe (approved):

rsempe:
I think we need to update API clients and doc, no?

PR #1083, opened 2023-06-02
feat: Rewrite billing service in pure SQL

vincent-pochet:
## Context

The current implementation of the `BillingService` is quite complex as it mixes Rails (Arel) query syntax with SQL, making it hard to understand and maintain.

## Description

This PR completely rewrites the query in pure SQL:
- it takes advantage of the Common Table Expressions (CTE) approach to simplify the previous implementation
- it rewrites the previous logic to deal with special anniversary cases (leap year, month durations...) from ruby to SQL
- it adds comments in the query to explain the different conditions and filters


rsempe (approved):
👏

PR #1090, opened 2023-06-06
feat (billable metrics): add support for recurring field

lovrocolic:
## Context

Currently it is not possible to charge events period over period for `sum_gg` and `unique_count_agg`

## Description

This PR adds `recurring` field on billable metric. This field is allowed only for `sum_agg` and `unique_count_agg` and it determines if persisted events would be charges or not.


app/graphql/mutations/billable_metrics/create.rb
```diff
@@ -15,6 +15,7 @@ class Create < BaseMutation
       argument :field_name, String, required: false
       argument :group, GraphQL::Types::JSON, required: false
       argument :name, String, required: true
+      argument :recurring, Boolean, required: false
```
jdenquin, line 18: GraphQL/ExtractInputType: Consider moving arguments to a new input type

app/models/billable_metric.rb
```diff
@@ -24,6 +24,8 @@ class BillableMetric < ApplicationRecord
 
   enum aggregation_type: AGGREGATION_TYPES
 
+  validate :validate_recurring, if: -> { count_agg? || max_agg? || recurring_count_agg? }
```
vincent-pochet, line 27: It fill a bit weird to have the preconditions splitter between this lambda and the validation method. What do you think of moving all conditions into the method?

lovrocolic: 👍

vincent-pochet (approved):

PR #1094, opened 2023-06-08
feat (recurring-bm): Update sum aggregation

lovrocolic:
## Context

Currently it is not possible to charge events period over period for `sum_gg` and `unique_count_agg`

## Description

This PR adds new logic in sum aggregation.  Base aggregation logic is updated regarding recurring field. Also, new logic is added for pay_in_advance case where we take into account maximum is current period.

### DIFF
--- a/app/services/billable_metrics/aggregations/sum_service.rb
+++ b/app/services/billable_metrics/aggregations/sum_service.rb
@@ -68,15 +68,17 @@ module BillableMetrics
 
         old_max = BigDecimal(previous_event.metadata['max_aggregation'])
 
-        if current_aggregation > old_max
+        result = if current_aggregation > old_max
           update_event_metadata(current_aggregation:, max_aggregation: current_aggregation)
 
-          return BigDecimal(current_aggregation - old_max)
+          current_aggregation - old_max
         else
           update_event_metadata(current_aggregation:, max_aggregation: old_max)
 
-          return BigDecimal(0)
+          0
         end
+
+        BigDecimal(result)
       end
 
       private
--- a/app/services/billable_metrics/aggregations/sum_service.rb
+++ b/app/services/billable_metrics/aggregations/sum_service.rb
@@ -4,7 +4,9 @@ module BillableMetrics
   module Aggregations
     class SumService < BillableMetrics::Aggregations::BaseService
       def aggregate(from_datetime:, to_datetime:, options: {})
-        events = events_scope(from_datetime:, to_datetime:)
+        charges_from_date = billable_metric.recurring? ? subscription.started_at : from_datetime
+
+        events = events_scope(from_datetime: charges_from_date, to_datetime:)
           .where("#{sanitized_field_name} IS NOT NULL")
 
         result.aggregation = events.sum("(#{sanitized_field_name})::numeric")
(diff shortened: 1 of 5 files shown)


app/services/billable_metrics/aggregations/sum_service.rb
```diff
@@ -54,7 +56,56 @@ def compute_pay_in_advance_aggregation
...
-        BigDecimal(event.properties.fetch(billable_metric.field_name, 0).to_s)
+        unless previous_event
+          value = event.properties.fetch(billable_metric.field_name, 0).to_s
+          update_event_metadata(current_aggregation: value, max_aggregation: value)
+
+          return BigDecimal(value)
+        end
+
+        current_aggregation = BigDecimal(previous_event.metadata['current_aggregation']) +
+          BigDecimal(event.properties.fetch(billable_metric.field_name, 0).to_s)
```
jdenquin, line 67: Layout/MultilineOperationIndentation: Align the operands of an expression in an assignment spanning multiple lines.

app/services/billable_metrics/aggregations/sum_service.rb
```diff
@@ -54,7 +56,56 @@ def compute_pay_in_advance_aggregation
...
+      def date_service
+        @date_service ||= Subscriptions::DatesService.new_instance(
+          subscription,
+          Time.current,
+          current_usage: true,
+        )
+      end
+
+      def previous_event
+        @previous_event ||= begin
```
jdenquin, line 95: Style/RedundantBegin: Redundant `begin` block detected. (https://rubystyle.guide#begin-implicit)

spec/services/billable_metrics/aggregations/sum_service_spec.rb
```diff
@@ -225,23 +267,80 @@
...
+          customer:,
+          subscription:,
+          timestamp: to_datetime - 3.days,
+          properties: {
+            total_count: -6,
+          },
+          metadata: {
+            current_aggregation: '4',
+            max_aggregation: '10',
+          }
```
jdenquin, line 305: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

app/services/billable_metrics/aggregations/sum_service.rb
```diff
@@ -54,7 +56,55 @@ def compute_pay_in_advance_aggregation
...
+
+        BigDecimal(result)
+      end
+
+      private
+
+      def date_service
+        @date_service ||= Subscriptions::DatesService.new_instance(
+          subscription,
+          Time.current,
```
vincent-pochet, line 89: We should not use `Time.current` at this level as it will prevent us from replaying the aggregation with a past date (for example to retry a failed job, of refreshing an invoice).

Do you think we could get this value injected into the aggregator at init?

app/services/billable_metrics/aggregations/sum_service.rb
```diff
@@ -54,7 +56,55 @@ def compute_pay_in_advance_aggregation
...
+          current_usage: true,
+        )
+      end
+
+      def previous_event
```
vincent-pochet, line 94: Could you add a comment explaining the goal of this method? I have the feeling that it will soon become unclear for people in the future :smile:

app/services/billable_metrics/aggregations/sum_service.rb
```diff
@@ -54,7 +56,55 @@ def compute_pay_in_advance_aggregation
...
+        @previous_event ||=
+          events_scope(from_datetime: date_service.charges_from_datetime, to_datetime: date_service.charges_to_datetime)
+            .where("#{sanitized_field_name} IS NOT NULL")
+            .where.not(id: event.id)
+            .order(created_at: :desc)
+            .first
+      end
+
+      def update_event_metadata(current_aggregation: nil, max_aggregation: nil)
```
vincent-pochet, line 103: I don't think this service should update the event. It's purpose is only to aggregate, nothing more.
I suggest you attach these properties to the result object, and update the event in the calling service.

spec/scenarios/pay_in_advance_charges_spec.rb
```diff
@@ -193,6 +193,98 @@
     end
   end
 
+  describe 'with sum_agg / standard' do
+    let(:aggregation_type) { 'sum_agg' }
+    let(:field_name) { 'amount' }
+
+    it 'creates an pay_in_advance fee' do
```
vincent-pochet, line 200:
```suggestion
    it 'creates a pay_in_advance fee' do
```

app/services/billable_metrics/aggregations/sum_service.rb
lovrocolic: added `event.timestamp`

app/services/billable_metrics/aggregations/sum_service.rb
lovrocolic: added;)

app/services/billable_metrics/aggregations/sum_service.rb
lovrocolic: Updated

app/services/fees/create_pay_in_advance_service.rb
```diff
@@ -126,5 +131,17 @@ def deliver_webhooks
 
       result.fees.each { |f| SendWebhookJob.perform_later('fee.created', f) }
     end
+
+    def update_event_metadata(aggregation_result:)
+      unless aggregation_result.current_aggregation.nil?
```
vincent-pochet, line 136: I see a potential issue with attaching these details to the event:

The `current_aggregation` and `max_aggregation` properties as relative to the subscription (via the boundaries). 
Since the same billable metric could be used on different plans having different duration and so depend to multiple subscriptions, the values stored at event level could we wrong.
I think we should store these values on an entity attached to the subscription and not on the event itself.

vincent-pochet (approved):

PR #1107, opened 2023-06-12
feat(stripe) send webhook when user is created

ansmonjol:
This PR adds a new feature.

Once a new customer with Stripe as a payment provider is added to the app, via UI or API, we send a webhook with a checkout session url.

This checkout session will be used to store customer payment information on Stripe in order to ease future payments.

This URL is sent on the `customer.checkout_url_generated` webhook event.

### DIFF
--- a/app/services/payment_provider_customers/stripe_service.rb
+++ b/app/services/payment_provider_customers/stripe_service.rb
@@ -2,6 +2,8 @@
 
 module PaymentProviderCustomers
   class StripeService < BaseService
+    CHECKOUT_SUCCESS_URL = 'https://www.getlago.com'
+
     def initialize(stripe_customer = nil)
       @stripe_customer = stripe_customer
 
@@ -20,6 +22,7 @@ module PaymentProviderCustomers
       )
 
       deliver_success_webhook
+      PaymentProviderCustomers::StripeCheckoutUrlJob.perform_later(stripe_customer)
 
       result.stripe_customer = stripe_customer
       result
@@ -72,6 +75,30 @@ module PaymentProviderCustomers
       result.single_validation_failure!(field: :payment_method_id, error_code: 'value_is_invalid')
     end
 
+    def generate_checkout_url
+      return unless customer.organization.webhook_url?
+
+      res = Stripe::Checkout::Session.create(
+        checkout_link_params,
+        {
+          api_key:,
+        },
+      )
+
+      result.checkout_url = res['url']
+
+      SendWebhookJob.perform_later(
+        'customer.checkout_url_generated',
+        customer,
+        checkout_url: result.checkout_url,
+      )
+
+      result
+    rescue Stripe::InvalidRequestError, Stripe::PermissionError => e
+      deliver_error_webhook(e)
+      result
+    end
+
     private
 
     attr_accessor :stripe_customer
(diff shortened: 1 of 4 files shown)


app/services/payment_provider_customers/stripe_service.rb
```diff
@@ -72,6 +73,23 @@ def check_payment_method(payment_method_id)
       result.single_validation_failure!(field: :payment_method_id, error_code: 'value_is_invalid')
     end
 
+    def generate_checkout_url
+      res = Stripe::Checkout::Session.create(
```
vincent-pochet, line 77: - We usualy tries to always return a `result` object within the service classes.

What about something like:

```ruby
response = Stripe::Checkout::Session.create(
  # ...

result.url = response['url']
result
```

- Are we sure that the response from Stripe is always a `success`? Meaning do we have to check for some status in the response?

ansmonjol: About the first point, I do understand however I'm wondering if that would be needed in such a case, as the URL only needs to be used in the webhook and not returned into the result at all. So I'm not 100% we should do it 🤔

About the 2nd, indeed it can be a failure.
I'll check the doc but will probably add a `rescue` to `deliver_error_webhook` as we do in this same file in other methods. I guess it should cover the failure case. There is no failure  status or such we can have in the response body.

app/services/payment_provider_customers/stripe_service.rb
```diff
@@ -86,6 +104,15 @@ def api_key
       organization.stripe_payment_provider.secret_key
     end
 
+    def checkout_link_params
+      {
+        success_url: 'https://www.getlago.com',
```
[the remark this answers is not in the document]
ansmonjol: It's the URL customer will be redirected to after checkout success.

Indeed that's odd and for now, we agreed with product to do so until complaints. And for now, we do not have an existing constant for such or other occurrences in the app.
Maybe it should be defined on top of the file with an explicit constant correct?

app/services/payment_provider_customers/stripe_service.rb
ansmonjol: I've updated the code with
- a result returned, containing a `checkout_url` attribute
- a rescue on the method that uses Stripe client

app/services/payment_provider_customers/stripe_service.rb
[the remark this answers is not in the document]
ansmonjol: Are you ok with letting it like this today? I assume later we would like to add a success page on either the app, the website or s3 (static file) to improve experience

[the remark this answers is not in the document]
vincent-pochet: If it's intended then fine with me :)

Well ideally I think it could be a setting somewhere but meanwhile maybe turning it into a constant is a good first step.

app/services/payment_provider_customers/stripe_service.rb
```diff
@@ -72,6 +73,30 @@ def check_payment_method(payment_method_id)
...
+      SendWebhookJob.perform_later(
+        'customer.checkout_url_generated',
+        customer,
+        checkout_url: result.checkout_url,
+      )
+
+      result
+    rescue Stripe::InvalidRequestError, Stripe::PermissionError => e
+      deliver_error_webhook(e)
+      nil
```
vincent-pochet, line 97: You should keep return `result` even without url, otherwise it will fail when calling `fail_with_error!` in the job.

app/services/payment_provider_customers/stripe_service.rb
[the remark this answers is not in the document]
ansmonjol: I've set it in the class on top of the file. Tested and it works

app/services/payment_provider_customers/stripe_service.rb
ansmonjol: ok! Updated

vincent-pochet (approved):

PR #1115, opened 2023-06-15
feat (billable metrics): extend billable metric query service

lovrocolic:
In order to support UI experience for recurring billable metric, it has to be enabled to filter billable metrics by `recurring` field

### DIFF
--- a/app/graphql/resolvers/billable_metrics_resolver.rb
+++ b/app/graphql/resolvers/billable_metrics_resolver.rb
@@ -10,13 +10,21 @@ module Resolvers
     argument :ids, [String], required: false, description: 'List of plan ID to fetch'
     argument :limit, Integer, required: false
     argument :page, Integer, required: false
+    argument :recurring, Boolean, required: false
     argument :search_term, String, required: false
 
     argument :aggregation_types, [Types::BillableMetrics::AggregationTypeEnum], required: false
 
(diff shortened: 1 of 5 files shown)


app/graphql/resolvers/billable_metrics_resolver.rb
```diff
@@ -10,13 +10,21 @@ class BillableMetricsResolver < GraphQL::Schema::Resolver
...
     argument :page, Integer, required: false
+    argument :recurring, Boolean, required: false
     argument :search_term, String, required: false
 
     argument :aggregation_types, [Types::BillableMetrics::AggregationTypeEnum], required: false
 
     type Types::BillableMetrics::Object.collection_type, null: false
 
-    def resolve(ids: nil, page: nil, limit: nil, search_term: nil, aggregation_types: nil)
+    def resolve( # rubocop:disable Metrics/ParameterLists
```
vincent-pochet, line 20: Maybe instead of disabling the linter, you could turn it into `def resolve(**args)`

And change the call to the query with:

```ruby
result = ::BillableMetricsQuery.new(organization: current_organization).call(
  search_term: args[:search_term],
  page: args[:page],
  limit: args[:limit],
  filters: args.slice(%i[ids recurring aggregation_types]),
)
```

vincent-pochet (approved):
Small comment, look good otherwise :)

PR #1136, opened 2023-06-27
feat (charges): pass charge to aggregation services

lovrocolic:
## Context

As part of charges paid in advance we want to allow users to select if they want to allow proration or not

## Description

This PR passes `charge` object to aggregation services instead of billable metric. In aggregation services we need access to `charge` object since prorated field (but also pay_in_advance) is attached on it. However we can also fetch billable metric directly from charge object.

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -112,7 +112,7 @@ class Invoice < ApplicationRecord
 
   def recurring_breakdown(fee)
     result = BillableMetrics::Aggregations::RecurringCountService.new(
-      billable_metric: fee.charge.billable_metric,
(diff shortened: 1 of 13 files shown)


vincent-pochet (commented):
It feels very weird to have aggregation scoped to billable metric and then to pass a charge as main argument.

How to plan to apply the pro-rata?

Maybe I missing something but my feeling is that we aggregation should not take care of the pro-rata logic, it should only aggregate the events between boundaries (or count persisted) and then the charge should use the aggregated result and apply pro-rata on it.
However, I'm not saying it's easy with the current approach :D
What do you think?

PR #1138, opened 2023-06-28
Display true up fees and groups on credit note template

rsempe:
The goal of this PR is to update the credit note template in order to display:
- true up fees
- dimensions


vincent-pochet (approved):
Well, I'm sure it's alright :sweat_smile:

PR #1153, opened 2023-07-05
feat(taxes): Ability to assign and unassign taxes to a plan

rsempe:
## Roadmap Task

## Context

After the delivery of the "multiple taxes" feature https://github.com/getlago/lago-api/pull/1104, we now want to be able to define taxes at plans or charge levels

## Description

This PR permits to assign and unassign taxes to a specific plan.

### DIFF
--- a/app/controllers/api/v1/plans/applied_taxes_controller.rb
+++ b/app/controllers/api/v1/plans/applied_taxes_controller.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Plans
+      class AppliedTaxesController < Api::BaseController
+        def create
+          tax = current_organization.taxes.find_by(code: input_params[:tax_code])
+          result = ::Plans::AppliedTaxes::CreateService.call(plan:, tax:)
+          if result.success?
+            render(json: ::V1::Plans::AppliedTaxSerializer.new(result.applied_tax, root_name: 'applied_tax'))
+          else
+            render_error_response(result)
+          end
+        end
+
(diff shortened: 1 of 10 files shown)


spec/requests/api/v1/plans/applied_taxes_spec.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::Plans::AppliedTaxesController, type: :request do
```
jdenquin, line 5: RSpec/FilePath: Spec path should end with `api/v1/plans/applied_taxes_controller*_spec.rb`. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FilePath)

app/controllers/api/v1/plans/applied_taxes_controller.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Plans
+      class AppliedTaxesController < Api::BaseController
+        def create
+          return not_found_error(resource: 'plan') unless plan
```
vincent-pochet, line 8: Since we already handle the "not_found" case in the create/destroy service, I'm wondering if we could get rid of these ones in the controller?

The `render_error_response` should take care of it automatically. What do you think?

rsempe: Good idea 👍

PR #1157, opened 2023-07-06
feat: cancel pending subscription

ansmonjol:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/retrieve-andor-cancel-pending-subscriptions

## Context

We want to allow canceling a pending subscription, either because they are planned in the future alone or due to a downgrade

## Description

This PR allows such action

Better review commit-by-commit

### DIFF
--- a/app/controllers/api/v1/subscriptions_controller.rb
+++ b/app/controllers/api/v1/subscriptions_controller.rb
@@ -119,7 +119,7 @@ module Api
       end
 
       def index_filters
-        params.permit(:external_customer_id, :plan_code)
+        params.permit(:external_customer_id, :plan_code, status: [])
       end
     end
   end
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/subscriptions_controller.rb
```diff
@@ -37,7 +37,13 @@ def create
 
       # NOTE: We can't destroy a subscription, it will terminate it
       def terminate
-        subscription = current_organization.subscriptions.active.find_by(external_id: params[:external_id])
+        subscription = current_organization.subscriptions.where(external_id: params[:external_id])
```
vincent-pochet, line 40: Maybe since it is an intermediary query variable, you could name it `query`:

```ruby
query = current_organization.subscriptions.where(external_id: params[:external_id])
query = params[:status] == 'pending' ? query.pending : query.active
subscription = query.first
```

vincent-pochet (approved):

app/controllers/api/v1/subscriptions_controller.rb
ansmonjol: Done something similar thanks for the proposal love it.

Just har to keep the `if...ifelse` for rubocop to not complain but renames the first variable to query

PR #1167, opened 2023-07-12
misc(puma): use web concurrency

jdenquin:


config/puma.rb
```diff
@@ -32,7 +32,7 @@
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
+workers ENV.fetch("WEB_CONCURRENCY") { 1 }
```
jdenquin, line 35: Style/RedundantFetchBlock: Use `fetch("WEB_CONCURRENCY", 1)` instead of `fetch("WEB_CONCURRENCY") { 1 }`. (https://github.com/JuanitoFatas/fast-ruby#hashfetch-with-argument-vs-hashfetch--block-code)

jdenquin, line 35: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols. (https://rubystyle.guide#consistent-string-literals)

rsempe (approved):

PR #1176, opened 2023-07-17
fix(event): Reduce noise with duplicated event errors

vincent-pochet:
## Context

Today, a lot of `Events::CreateJob` Sidekiq jobs are failling with two similar kind of errors:
- `BaseService::ValidationFailure`: Validation errors: {:transaction_id=>["value_already_exist"]}
- `ActiveRecord::RecordNotUnique`: PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "index_events_on_subscription_id_and_transaction_id" 

These two errors are actually the same and are related to the validation rules around transaction_id/subscription_id unicity to prevent duplicated events.

The `Events::CreateService` has been initially built to return the event if it already exists with the transaction_id for the same subscription.

## Description

Since these errors generating a lot of noise in the dead job queue and since there is nothing we can do from a debug perspective, this PR tries to improve the current behaviour.

With this change, if one of the `BaseService::ValidationFailure` or `ActiveRecord::RecordNotUnique` is raised:
- we first look again in DB to make sure that the same event was not created between the enqueuing of the job and the processing.
- if it is found, we just return it, preventing the jobs from ending up in the dead queue
- if not found, we notify the emitter via a webhook and we return a validation error, in this case it will keep failling into the dead queue for further investigation.


app/services/events/validate_creation_service.rb
```diff
@@ -57,9 +57,10 @@ def validate_create
       return invalid_customer_error if params[:external_customer_id] && !customer
       return missing_subscription_error if params[:external_subscription_id].blank? && subscriptions.count > 1
       return missing_subscription_error if subscriptions.empty?
-      if params[:external_subscription_id] && !subscriptions.pluck(:external_id).include?(params[:external_subscription_id])
+      if params[:external_subscription_id] && subscriptions.pluck(:external_id).exclude?(params[:external_subscription_id])
```
jdenquin, line 60: Layout/LineLength: Line is too long. [123/120] (https://rubystyle.guide#max-line-length)

rsempe (approved):

PR #1179, opened 2023-07-18
fix (subscription): fix billing boundaries for termination case

lovrocolic:
## Context

Fix billing boundaries for termination case if there is customer timezone

## Description

If customer timezone is set, in case of termination wrong number of days was calculated and therefore proration coefficient was incorrect

### DIFF
--- a/spec/models/subscription_spec.rb
+++ b/spec/models/subscription_spec.rb
@@ -165,7 +165,7 @@ RSpec.describe Subscription, type: :model do
           plan:,
           external_id: 'sub_id',
           customer: previous_subscription.customer,
-          )
+        )
       end
       let(:previous_subscription) do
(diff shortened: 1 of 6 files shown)


spec/models/subscription_spec.rb
```diff
@@ -135,6 +135,53 @@
...
+    context 'with a previous subscription' do
+      let(:subscription) do
+        create(
+          :active_subscription,
+          previous_subscription:,
+          started_at: Time.zone.yesterday,
+          plan:,
+          external_id: 'sub_id',
+          customer: previous_subscription.customer,
+          )
```
jdenquin, line 168: Layout/ClosingParenthesisIndentation: Indent `)` to column 8 (not 10)

vincent-pochet (approved):
:rocket:

PR #1183, opened 2023-07-20
feat(taxes): Ability to assign taxes on plan creation

rsempe:
## Roadmap Task

## Context

After the delivery of the "multiple taxes" feature https://github.com/getlago/lago-api/pull/1104, we now want to be able to define taxes at plans or charge levels

## Description

The goal of this PR is to use an array of `tax_codes` to assign or unassign taxes on plan.


app/graphql/mutations/plans/create.rb
```diff
@@ -18,6 +18,7 @@ class Create < BaseMutation
       argument :name, String, required: true
       argument :parent_id, ID, required: false
       argument :pay_in_advance, Boolean, required: true
+      argument :tax_codes, [String], required: false
```
jdenquin, line 21: GraphQL/ExtractInputType: Consider moving arguments to a new input type

app/graphql/mutations/plans/update.rb
```diff
@@ -17,6 +17,7 @@ class Update < BaseMutation
       argument :interval, Types::Plans::IntervalEnum, required: true
       argument :name, String, required: true
       argument :pay_in_advance, Boolean, required: true
+      argument :tax_codes, [String], required: false
```
jdenquin, line 20: GraphQL/ExtractInputType: Consider moving arguments to a new input type

vincent-pochet (approved):
:rocket:

app/graphql/mutations/plans/create.rb
ansmonjol, line 21: can we have this named `applied_taxes` and be an array of `Types::Taxes::Object`? So I can retrieve the data directly using graphql response instead of performing a call to retrieve tax information such as the rate? 
I don't think we have a `GET /taxes/:id` endpoint

ansmonjol, line 21: Why not receive an array `Types::Taxes::Object`, as this is what the pan query returns?

rsempe: to be consistent with the API... is that a blocker for you?

ansmonjol (approved):

PR #1185, opened 2023-07-20
feat (aggregations): migrate old aggregation data

lovrocolic:
## Context

As part of new aggregation logic, we need to migrate old aggregation data.

### DIFF
--- a/db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
+++ b/db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
@@ -29,6 +29,7 @@ class MigrateRecurringCountAndUniqueCountAggregation < ActiveRecord::Migration[7
           INNER JOIN subscriptions ON subscriptions.id = events.subscription_id
           INNER JOIN billable_metrics ON billable_metrics.code = events.code
           WHERE events.deleted_at IS NULL
+            AND events.properties ->> billable_metrics.field_name IS NOT NULL
             AND billable_metrics.aggregation_type = 3
             AND billable_metrics.organization_id = events.organization_id
           ORDER BY event_timestamp ASC;
--- a/db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
+++ b/db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
@@ -8,6 +8,7 @@ class MigrateRecurringCountAndUniqueCountAggregation < ActiveRecord::Migration[7
   class Event < ApplicationRecord; end
   class Plan < ApplicationRecord; end
   class Subscription < ApplicationRecord; end
+  class QuantifiedEvent < ApplicationRecord; end
 
   def change
     reversible do |dir|
@@ -21,26 +22,25 @@ class MigrateRecurringCountAndUniqueCountAggregation < ActiveRecord::Migration[7
             events.customer_id AS customer_id,
             events.properties AS properties,
             events.timestamp AS event_timestamp,
-            subscriptions.external_id AS subscription_external_id
+            subscriptions.external_id AS subscription_external_id,
+            billable_metrics.id AS billable_metric_id,
+            billable_metrics.field_name AS field_name
           FROM events
           INNER JOIN subscriptions ON subscriptions.id = events.subscription_id
           INNER JOIN billable_metrics ON billable_metrics.code = events.code
-          WHERE events.deleted_at IS NULL AND billable_metrics.aggregation_type = 3
+          WHERE events.deleted_at IS NULL
+            AND billable_metrics.aggregation_type = 3
+            AND billable_metrics.organization_id = events.organization_id
           ORDER BY event_timestamp ASC;
         SQL
 
         ApplicationRecord.connection.select_all(sql).each_with_object({}) do |row, _result|
-          matching_billable_metric = BillableMetric.find_by(
-            code: row['code'],
-            organization_id: row['organization_id'],
-          )
-
           existing_quantified_event =
             QuantifiedEvent.where(
               customer_id: row['customer_id'],
-              billable_metric_id: matching_billable_metric.id,
+              billable_metric_id: row['billable_metric_id'],
               external_subscription_id: row['subscription_external_id'],
-              external_id: JSON.parse(row['properties'])[matching_billable_metric.field_name.to_s],
+              external_id: JSON.parse(row['properties'])[row['field_name'].to_s],
             ).where(removed_at: nil).any?
 
           # There can only be one quantified event for certain external_id which guarantees uniqueness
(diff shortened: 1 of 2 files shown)


db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
```diff
@@ -0,0 +1,151 @@
...
+        # All charges that are related to recurring_count_agg should be prorated = true
+        sql = <<-SQL
+          SELECT
+            charges.id AS charge_id
+          FROM charges
+          INNER JOIN billable_metrics ON billable_metrics.id = charges.billable_metric_id
+          WHERE billable_metrics.aggregation_type = 4;
+        SQL
+        ApplicationRecord.connection.select_all(sql).each_with_object({}) do |row, _result|
+          charge = Charge.find_by(id: row['charge_id'])
```
vincent-pochet, line 72: We should avoid as much as possible using rails models inside migration since we add a lot of issues with this in the past.

For example this one could be written in pure SQL:

```ruby
execute <<-SQL
  WITH recurring_count_charges AS (
    SELECT charges.id AS charge_id
    FROM charges
      INNER JOIN billable_metrics ON billable_metrics.id = charges.billable_metric_id
    WHERE billable_metrics.aggregation_type = 4 -- recurring_count_agg
  )

  UPDATE charges
  SET prorated = true
  FROM recurring_count_charges
  WHERE recurring_count_charges.charge_id = charges.id
SQL
```

db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
```diff
@@ -0,0 +1,151 @@
...
+
+        # Recurring count agg billable metrics should be recurring = true and aggregation type should be unique count
+        sql = <<-SQL
+          SELECT
+            billable_metrics.id AS billable_metric_id
+          FROM billable_metrics
+          WHERE billable_metrics.aggregation_type = 4;
+        SQL
+        ApplicationRecord.connection.select_all(sql).each_with_object({}) do |row, _result|
+          billable_metric = BillableMetric.find_by(id: row['billable_metric_id'])
```
vincent-pochet, line 87: Same comment, this one could be turned into a single UPDATE SQL query

db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
```diff
@@ -0,0 +1,151 @@
...
+            subscriptions.external_id AS subscription_external_id
+          FROM events
+          INNER JOIN subscriptions ON subscriptions.id = events.subscription_id
+          INNER JOIN plans ON plans.id = subscriptions.plan_id
+          INNER JOIN charges ON charges.deleted_at IS NULL AND charges.plan_id = plans.id
+          INNER JOIN billable_metrics on billable_metrics.code = events.code
+          WHERE events.deleted_at IS NULL AND charges.pay_in_advance = TRUE AND billable_metrics.aggregation_type = 3;
+        SQL
+        ApplicationRecord.connection.select_all(sql).each_with_object({}) do |row, _result|
+          event = Event.find_by(id: row['event_id'])
```
vincent-pochet, line 143: Same comment

db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
```diff
@@ -0,0 +1,151 @@
...
+            subscriptions.external_id AS subscription_external_id
+          FROM events
+          INNER JOIN subscriptions ON subscriptions.id = events.subscription_id
+          INNER JOIN plans ON plans.id = subscriptions.plan_id
+          INNER JOIN charges ON charges.deleted_at IS NULL AND charges.plan_id = plans.id
+          INNER JOIN billable_metrics on billable_metrics.code = events.code
+          WHERE events.deleted_at IS NULL AND charges.pay_in_advance = TRUE AND billable_metrics.aggregation_type = 1;
+        SQL
+        ApplicationRecord.connection.select_all(sql).each_with_object({}) do |row, _result|
+          billable_metric = BillableMetric.find_by(
```
vincent-pochet, line 112: Maybe a bit trickier, but I think it would also be way more efficient in a single SQL update

db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
```diff
@@ -0,0 +1,151 @@
...
+        SQL
+
+        ApplicationRecord.connection.select_all(sql).each_with_object({}) do |row, _result|
+          matching_billable_metric = BillableMetric.find_by(
+            code: row['code'],
+            organization_id: row['organization_id'],
+          )
+
+          existing_quantified_event =
+            QuantifiedEvent.where(
```
vincent-pochet, line 39: You could probably filter those ones directly in the query to reduce the number of calls on the database.

db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
```diff
@@ -0,0 +1,151 @@
...
+              customer_id: row['customer_id'],
+              billable_metric_id: matching_billable_metric.id,
+              external_subscription_id: row['subscription_external_id'],
+              external_id: JSON.parse(row['properties'])[matching_billable_metric.field_name.to_s],
+            ).where(removed_at: nil).any?
+
+          # There can only be one quantified event for certain external_id which guarantees uniqueness
+          next if existing_quantified_event
+
+          quantified_event = QuantifiedEvent.create!(
```
vincent-pochet, line 49: Since here we cannot avoid using the model, I would suggest to redefine it in the migration, to prevent issue in the future if we change validations, attributes or anything else on the model.

vincent-pochet (commented):
A lot of comment, sorry about that :confused: , but I really think we should avoid using `ActiveRecord` as much as possible in the migrations, especially for update purpose.

Let me know if you need an hand to rewrite them into pure SQL

db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
lovrocolic: update this part based on our conversation. This condition left as it is but removed BM query in the loop

db/migrate/20230719100256_migrate_recurring_count_and_unique_count_aggregation.rb
lovrocolic: updated

lovrocolic:
> A lot of comment, sorry about that 😕 , but I really think we should avoid using `ActiveRecord` as much as possible in the migrations, especially for update purpose.
> 
> Let me know if you need an hand to rewrite them into pure SQL

Updated migration, tried it locally and seems that it's working fine 🤞

jdenquin (approved):

jdenquin:
Can we run it on staging main with some old data and make sure that everything is okay? 🙏

lovrocolic:
> Can we run it on staging main with some old data and make sure that everything is okay? 🙏

@jdenquin - @sarkissianraffi already prepared test data on staging, sure, we will first try it there 🤞 I also tested it on my set of data locally

PR #1191, opened 2023-07-25
feat(webhooks): Webhook endpoints update

jdenquin:
## Description

- Add `PUT` endpoint for `webhook_endpoints`

### DIFF
--- a/app/controllers/api/v1/webhook_endpoints_controller.rb
+++ b/app/controllers/api/v1/webhook_endpoints_controller.rb
@@ -11,11 +11,23 @@ module Api
 
         result = service.call
 
-        if result.success?
-          render_webhook_endpoint(result.webhook_endpoint)
-        else
-          render_error_response(result)
-        end
+        return render_webhook_endpoint(result.webhook_endpoint) if result.success?
+
+        render_error_response(result)
+      end
+
+      def update
+        service = ::WebhookEndpoints::UpdateService.new(
+          id: params[:id],
+          organization: current_organization,
+          params: update_params,
+        )
+
+        result = service.call
+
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/webhook_endpoints_controller.rb
```diff
@@ -11,11 +11,23 @@ def create
...
+          id: params[:id],
+          organization: current_organization,
+          params: update_params,
+        )
+
+        result = service.call
+
+        return render_webhook_endpoint(result.webhook_endpoint) if result.success?
+
+        render_error_response(result) unless result.success?
```
rsempe, line 30:
```suggestion
        render_error_response(result)
```

app/services/webhook_endpoints/update_service.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      super
+    end
+
+    def call
+      webhook_endpoint = organization.webhook_endpoints.find_by(id:)
+
+      return result.not_found_failure!(resource: 'webhook_endpoint') if webhook_endpoint.blank?
+
+      webhook_endpoint.update(
```
rsempe, line 18:
```suggestion
      webhook_endpoint.update!(
```

config/routes.rb
```diff
@@ -61,7 +61,7 @@
 
       put '/organizations', to: 'organizations#update'
 
-      resources :webhook_endpoints, only: %i[create index show destroy]
+      resources :webhook_endpoints
```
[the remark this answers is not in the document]
jdenquin: since we are on api only it should not?!

lovrocolic (approved):

PR #1203, opened 2023-07-31
fix (subscriptions): do not raise duplicated_invoice error for yearly plan case

lovrocolic:
## Context

If plan interval is `yearly` and `bill_charges_monthly` is `true` we do not want to raise `duplicated_invoice` error since on multiple invoices we can have the same `from_datetime` and `to_datetime` values but different `charges_from_datetime` and `charges_to_datetime`


rsempe (approved):
🤞 🚀

PR #1206, opened 2023-07-31
fix(register): Do not permit signup when disabled

jdenquin:
## Context

https://github.com/getlago/lago/issues/220

## Description

- Do not permit register (signup) when its disabled with `LAGO_SIGNUP_DISABLED` env var is set to `true`

### DIFF
--- a/app/services/users_service.rb
+++ b/app/services/users_service.rb
@@ -17,6 +17,8 @@ class UsersService < BaseService
   end
 
   def register(email, password, organization_name)
+    return result.not_allowed_failure!(code: 'signup is disabled') if ENV.fetch('LAGO_SIGNUP_DISABLED', 'false') == 'true'
+
     result.user = User.find_or_initialize_by(email:)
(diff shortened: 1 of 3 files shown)


app/services/users_service.rb
```diff
@@ -17,6 +17,8 @@ def login(email, password)
   end
 
   def register(email, password, organization_name)
+    return result.not_allowed_failure!(code: 'signup is disabled') if ENV.fetch('LAGO_SIGNUP_DISABLED', 'false') == 'true'
```
jdenquin, line 20: Style/IfUnlessModifier: Modifier form of `if` makes the line too long. (https://rubystyle.guide#if-as-a-modifier)

jdenquin, line 20: Layout/LineLength: Line is too long. [122/120] (https://rubystyle.guide#max-line-length)

rsempe (approved):

PR #1210, opened 2023-08-01
fix(fees): handle precise_coupon_amount on fees when its value is greater than fees amount_cents

lovrocolic:
## Description

handle `precise_coupon_amount` on fees when its value is greater than fees `amount_cents`


rsempe (approved):
🤖

PR #1213, opened 2023-08-02
fix (charges): do not bill recurring pay in advance charges upon termination

lovrocolic:
## Description

When billable metric is recurring and charge is pay in advance, we do not want to bill such charges upon subscription termination

### DIFF
--- a/spec/services/invoices/calculate_fees_service_spec.rb
+++ b/spec/services/invoices/calculate_fees_service_spec.rb
@@ -237,7 +237,7 @@ RSpec.describe Invoices::CalculateFeesService, type: :service do
         end
       end
 
-      context 'charges are pay in advance and billable metric is recurring' do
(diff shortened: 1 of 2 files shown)


spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -237,6 +237,32 @@
         end
       end
 
+      context 'charges are pay in advance and billable metric is recurring' do
```
jdenquin, line 240: RSpec/ContextWording: Context description should match /^when\b/, /^with\b/, or /^without\b/. (https://rspec.rubystyle.guide/#context-descriptions, https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ContextWording)

rsempe (approved):

PR #1228, opened 2023-08-08
fix (charges): fix condition that determines when to bill charges

lovrocolic:
**Cases when we don’t want to bill the charges (i.e. don’t include them in subscription invoice):**

1. pay_in_advance = true AND recurring = false
2. pay_in_advance = true AND recurring = true AND subscription IS terminated (upgrade / old subscription)
3. pay_in_advance = true AND recurring = true AND subscription IS terminated (regular termination)
4. pay_in_advance = false AND prorated = false AND recurring = true AND subscription IS terminated (upgrade / old subscription)

**IN THIS CASE WE WANT TO BILL → BUG**

1. pay_in_advance = false AND prorated = false AND recurring = true AND subscription IS terminated (regular termination) → BILL THIS ONE ⚠️

### DIFF
--- a/app/services/invoices/calculate_fees_service.rb
+++ b/app/services/invoices/calculate_fees_service.rb
@@ -135,8 +135,8 @@ module Invoices
     def should_not_create_charge_fee?(charge, subscription)
       if charge.pay_in_advance?
         condition = charge.billable_metric.recurring? &&
-          subscription.terminated? &&
-          (subscription.upgraded? || subscription.next_subscription.nil?)
+                    subscription.terminated? &&
+                    (subscription.upgraded? || subscription.next_subscription.nil?)
 
         return condition
       end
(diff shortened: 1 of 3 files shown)


app/services/invoices/calculate_fees_service.rb
```diff
@@ -133,10 +133,19 @@ def create_charges_fees(subscription, boundaries)
     end
 
     def should_not_create_charge_fee?(charge, subscription)
-      (charge.pay_in_advance? || !charge.prorated?) &&
-        charge.billable_metric.recurring? &&
+      if charge.pay_in_advance?
+        condition = charge.billable_metric.recurring? &&
+          subscription.terminated? &&
+          (subscription.upgraded? || subscription.next_subscription.nil?)
```
jdenquin, line 139: Layout/MultilineOperationIndentation: Align the operands of an expression in an assignment spanning multiple lines.

vincent-pochet (approved):
Looks good to me. (I trust you :D)

What do you think of adding some scenario to covers the different cases?

PR #1240, opened 2023-08-14
misc: Recurring count migration is missing ruby extenssion

vincent-pochet:
## Context

The `MigrateRecurringCountAndUniqueCountAggregation` migration is missing the `.rb ` file extenssion.

NOTE: it as been executed anyway on the production server

## Description

This PR simply fix the name by adding the correct ruby file extension to the file.


lovrocolic:
As I can see it was renamed from .rb -> without extension 3 days ago (probably by mistake) so yeah as you said at the time of running it was executed correctly since it had extension then :)

lovrocolic (approved):

PR #1243, opened 2023-08-14
fix (invoice): add charges in advance interval

lovrocolic:
## Context

When plan is pay in arrear and charges are pay in advance we need to send to the UI correct charges interval

### DIFF
--- a/app/graphql/types/invoice_subscription/object.rb
+++ b/app/graphql/types/invoice_subscription/object.rb
@@ -40,8 +40,8 @@ module Types
 
         @should_use_in_advance_charges_interval =
           object.fees.charge_kind.any? &&
-            object.subscription.plan.charges.where(pay_in_advance: true).any? &&
-            !object.subscription.plan.pay_in_advance?
+          object.subscription.plan.charges.where(pay_in_advance: true).any? &&
+          !object.subscription.plan.pay_in_advance?
       end
 
       def charge_pay_in_advance_interval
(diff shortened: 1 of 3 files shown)


ansmonjol (approved):

app/graphql/types/invoice_subscription/object.rb
```diff
@@ -16,8 +16,38 @@ class Object < Types::BaseObject
...
+
+        charge_pay_in_advance_interval[:charges_to_date]
+      end
+
+      def should_use_in_advance_charges_interval
+        return @should_use_in_advance_charges_interval if defined? @should_use_in_advance_charges_interval
+
+        @should_use_in_advance_charges_interval =
+          object.fees.charge_kind.any? &&
+            object.subscription.plan.charges.where(pay_in_advance: true).any? &&
```
jdenquin, line 43: Layout/MultilineOperationIndentation: Align the operands of an expression in an assignment spanning multiple lines.

app/graphql/types/invoice_subscription/object.rb
```diff
@@ -16,8 +16,38 @@ class Object < Types::BaseObject
...
+            !object.subscription.plan.pay_in_advance?
```
jdenquin, line 44: Layout/MultilineOperationIndentation: Align the operands of an expression in an assignment spanning multiple lines.

PR #1264, opened 2023-08-22
feat(charge): Add per_event_aggregation to aggregators

vincent-pochet:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/apply-price-floorcap-on-transactions

## Context

Fintech users want to be able to cap a transaction to a maximum amount or to flour it to a minimum amount.

This feature will apply this logic to the `percentage` charge model.

## Description

This PR adds a new `per_event_aggregation` method to the billable metric aggregator services.
This method will later be used for the min/max amount per transaction computation


rsempe (approved):
lgtm

PR #1280, opened 2023-08-30
feat (subscriptions): add job that terminates ended subscriptions

lovrocolic:
## Context

Currently it is not possible to terminate subscriptions automatically

## Description

In this PR new job that terminates ended subscriptions has been added


vincent-pochet (approved):
:rocket:

PR #1282, opened 2023-08-31
feat(add_on): Remove routes and services for applied add on creation

vincent-pochet:
## Context

Applied add on have been deprecated since it has been replaced by the invoice creation route.

## Description

This PR removes routes, services and serializers logic related to the applied add-ons


rsempe (approved):
🙌 🙌

PR #1283, opened 2023-08-31
feat (subscriptions): add job that alerts users about ending subscriptions

lovrocolic:
## Context

Currently it is not possible to terminate subscriptions automatically

## Description

This PR adds a job that sends webhook 15 days before subscription ending date

### DIFF
--- a/app/jobs/clock/subscriptions_to_be_terminated_job.rb
+++ b/app/jobs/clock/subscriptions_to_be_terminated_job.rb
@@ -8,13 +8,13 @@ module Clock
       Subscription
         .joins(customer: :organization)
         .joins('left join webhooks on subscriptions.id = webhooks.object_id and '\
-               "webhooks.webhook_type = 'subscription.reaching_termination'")
+               "webhooks.webhook_type = 'subscription.termination_alert'")
         .active
         .where("DATE(#{Subscription.ending_at_in_timezone_sql}) = ?", (Time.current + 15.days).to_date)
         .where('webhooks.id IS NULL')
         .find_each do |subscription|
           if subscription.customer.organization.webhook_endpoints.any?
-            SendWebhookJob.perform_later('subscription.reaching_termination', subscription)
+            SendWebhookJob.perform_later('subscription.termination_alert', subscription)
           end
         end
     end
--- a/app/jobs/clock/subscriptions_to_be_terminated_job.rb
+++ b/app/jobs/clock/subscriptions_to_be_terminated_job.rb
@@ -8,13 +8,13 @@ module Clock
       Subscription
         .joins(customer: :organization)
         .joins('left join webhooks on subscriptions.id = webhooks.object_id and '\
-               "webhooks.webhook_type = 'alert.subscription_to_be_terminated' and webhooks.status = 1")
+               "webhooks.webhook_type = 'subscription.reaching_termination'")
         .active
         .where("DATE(#{Subscription.ending_at_in_timezone_sql}) = ?", (Time.current + 15.days).to_date)
         .where('webhooks.id IS NULL')
         .find_each do |subscription|
           if subscription.customer.organization.webhook_endpoints.any?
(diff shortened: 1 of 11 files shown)


app/jobs/clock/subscriptions_to_be_terminated_job.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module Clock
+  class SubscriptionsToBeTerminatedJob < ApplicationJob
+    queue_as 'clock'
+
+    def perform
+      Subscription
+        .joins(customer: :organization)
+        .joins('left join webhooks on subscriptions.id = webhooks.object_id and '\
+               "webhooks.webhook_type = 'alert.subscription_to_be_terminated' and webhooks.status = 1"
```
jdenquin, line 11: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

app/jobs/clock/subscriptions_to_be_terminated_job.rb
```diff
@@ -0,0 +1,23 @@
...
+        )
```
jdenquin, line 12: Layout/ClosingParenthesisIndentation: Align `)` with `(`.

spec/jobs/send_webhook_job_spec.rb
```diff
@@ -338,6 +338,28 @@
...
   end
 
+  context 'when webhook_type is alert.subscription_to_be_terminated' do
+    let(:webhook_service) { instance_double(Webhooks::Subscriptions::TerminatingSoonService) }
+    let(:subscription) { create(:subscription) }
+
+    before do
+      allow(Webhooks::Subscriptions::TerminatingSoonService).to receive(:new)
+       .with(object: subscription, options: {}, webhook_id: nil)
+       .and_return(webhook_service)
```
jdenquin, line 348: Layout/MultilineMethodCallIndentation: Use 2 (not 1) spaces for indenting an expression spanning multiple lines.

app/jobs/send_webhook_job.rb
```diff
@@ -8,6 +8,7 @@ class SendWebhookJob < ApplicationJob
   retry_on ActiveJob::DeserializationError, wait: :exponentially_longer, attempts: 6
 
   WEBHOOK_SERVICES = {
+    'alert.subscription_to_be_terminated' => Webhooks::Subscriptions::TerminatingSoonService,
```
vincent-pochet, line 11: The webhook should be scoped to the resource type it is alerting about, here `subscription`.

What about something like `subscription.reaching_termination`.

In addition, I think we are doing ourself a favor by using the same name for both the webhook service and the event name :). So here, `Webhooks::Subscriptions::ReachingTerminationService`

app/jobs/clock/subscriptions_to_be_terminated_job.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module Clock
+  class SubscriptionsToBeTerminatedJob < ApplicationJob
+    queue_as 'clock'
+
+    def perform
+      Subscription
+        .joins(customer: :organization)
+        .joins('left join webhooks on subscriptions.id = webhooks.object_id and '\
+               "webhooks.webhook_type = 'alert.subscription_to_be_terminated' and webhooks.status = 1")
```
vincent-pochet, line 11: Since we have an auto-retry system on the webhooks delivery. Should we really check the status?
The risk here is that we send twice or more the same webhook if the first attempt was not a success. 

My point is that we should let the webhook engine responsible for ensuring the right delivery of the webhook, and try to push only one alert at this step. What do you think?

vincent-pochet (commented):
Small comments, looks great otherwise ;)

app/jobs/clock/subscriptions_to_be_terminated_job.rb
lovrocolic: make sense 👍

PR #1290, opened 2023-09-05
feat(weighted-sum): Add new aggregation type to billable metric

vincent-pochet:
## Context

For metrics that are charged upon a time frame, Lago is not able to calculate correctly the charge. For instance, imagine you want to price the number of GB/seconds used by a server. Currently, you can either calculate the number of seconds used, or the number of Gigabytes used, but not both at the same time.

## Description

This PR adds:
- The new `weighted_sum_agg` value in `BillableMetric::AGGREGATION_TYPES` enumeration
- A new `BillableMetric::WEIGHTED_INTERVAL` enumeration


app/models/billable_metric.rb
```diff
@@ -20,9 +20,13 @@ class BillableMetric < ApplicationRecord
...
     recurring_count_agg
+    weighted_sum_agg
   ].freeze
 
+  WEIGHTED_INTERVAL = { seconds: 'seconds' }.freeze
+
   enum aggregation_type: AGGREGATION_TYPES
+  enum weighted_interval: WEIGHTED_INTERVAL
 
   validate :validate_recurring
```
lovrocolic, line 31: Is this new BM compatible with recurring? If no, we should update this validation

vincent-pochet: Yes, it will allow recurring as well

lovrocolic (approved):

PR #1292, opened 2023-09-05
feat(weighted-sum): Add aggregation logic

vincent-pochet:
## Context

For metrics that are charged upon a time frame, Lago is not able to calculate correctly the charge. For instance, imagine you want to price the number of GB/seconds used by a server. Currently, you can either calculate the number of seconds used, or the number of Gigabytes used, but not both at the same time.

## Description

This PR adds:
- the aggregation logic for the weighted sum billable metrics.
- it persists the units in `quantified_events` table when the billable metric is recurring
- it stores the units before the application of the `weigthed_interval` to be able to display it on the invoice

### DIFF
--- a/app/services/billable_metrics/aggregations/weighted_sum_service.rb
+++ b/app/services/billable_metrics/aggregations/weighted_sum_service.rb
@@ -4,15 +4,19 @@ module BillableMetrics
   module Aggregations
     class WeightedSumService < BillableMetrics::Aggregations::BaseService
       def aggregate(options: {})
+        events = fetch_events(from_datetime:, to_datetime:)
+
         result.aggregation = compute_aggregation
-        result.current_usage_units = 0
-        result.count = 0
-        result.pay_in_advance_aggregation = BigDecimal(0)
-        result.options = { running_total: 0 }
+        result.count = events.count
+        result.options = options
         result
       end
 
       private
+
+      def fetch_events(from_datetime:, to_datetime:)
+        events_scope(from_datetime:, to_datetime:).where("#{sanitized_field_name} IS NOT NULL")
+      end
 
       def compute_aggregation
         # query_result = ActiveRecord::Base.connection.select_all(aggregation_sql)
@@ -30,7 +34,7 @@ module BillableMetrics
             (#{initial_value})
             UNION
             (#{
-              events_scope(from_datetime:, to_datetime:)
+              fetch_events(from_datetime:, to_datetime:)
                 .select("timestamp, (#{sanitized_field_name})::numeric AS difference")
                 .to_sql
             })
--- a/app/services/billable_metrics/aggregations/weighted_sum_service.rb
+++ b/app/services/billable_metrics/aggregations/weighted_sum_service.rb
@@ -19,6 +19,10 @@ module BillableMetrics
         result
       end
 
+      def breakdown
+        ActiveRecord::Base.connection.select_all(breakdown_sql).to_a
+      end
+
       private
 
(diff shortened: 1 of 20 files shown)


app/models/charge.rb
```diff
@@ -87,11 +87,16 @@ def validate_charge_model(validator)
...
 
   # NOTE: An pay_in_advance charge cannot be created in the following cases:
-  # - billable metric aggregation type is max_agg or recurring_count_agg
+  # - billable metric aggregation type is max_agg, recurring_count_agg or weighted_sum_agg
   # - charge model is volume
   def validate_pay_in_advance
     return unless pay_in_advance?
-    return unless billable_metric.recurring_count_agg? || billable_metric.max_agg? || volume?
+
+    unless %w[recurring_count_agg max_add weighted_sum_agg].include?(billable_metric.aggregation_type) ||
```
lovrocolic, line 95: should be `max_agg` :) I fixed it while resolving merge conflicts. Maybe you can rebase merge this branch?

lovrocolic:
Should we add prorated aggregation service since `recurring` can be true for this BM type?

vincent-pochet:
@lovrocolic that's a fair question.
I'll check, but i'm not sure "prorated" would mean something in this context since here aggregation is already prorated on the period duration

lovrocolic:
> @lovrocolic that's a fair question. I'll check, but i'm not sure "prorated" would mean something in this context since here aggregation is already prorated on the period duration

Yep, I am not sure for this context, but if prorated field cannot be true in this case, maybe we can update depending validation in charge model?

app/services/billable_metrics/aggregations/weighted_sum_service.rb
```diff
@@ -4,14 +4,110 @@ module BillableMetrics
...
+            VALUES (timestamp without time zone '#{to_datetime}', 0)
+          ) AS t(timestamp, difference)
+        SQL
+      end
+
+      def latest_value
+        quantified_event = QuantifiedEvent
+          .where(billable_metric_id: billable_metric.id)
+          .where(customer_id: subscription.customer_id)
+          .where(external_subscription_id: subscription.external_id)
```
lovrocolic, line 104: should we have something similar hare as for `unique_count`?
```
quantified_events = if billable_metric.recurring?
          quantified_events.where(external_subscription_id: subscription.external_id)
        else
          quantified_events.joins(:events).where(events: { subscription_id: subscription.id })
        end
```

app/services/charges/charge_models/base_service.rb
```diff
@@ -20,6 +20,8 @@ def apply
         result.full_units_number = aggregation_result.full_units_number
         result.count = aggregation_result.count
         result.amount = compute_amount
+        result.total_aggregated_units = aggregation_result.total_aggregated_units if aggregation_result.total_aggregated_units
```
jdenquin, line 23: Style/IfUnlessModifier: Modifier form of `if` makes the line too long. (https://rubystyle.guide#if-as-a-modifier)

jdenquin, line 23: Layout/LineLength: Line is too long. [126/120] (https://rubystyle.guide#max-line-length)

app/services/billable_metrics/aggregations/weighted_sum_service.rb
vincent-pochet: plus, we want to transfers the consumption from one subscription to another in case of upgrade/downgrade

lovrocolic (approved):

PR #1296, opened 2023-09-07
fix (subscriptions): terminate ended subscriptions condition regarding timezone

lovrocolic:
## Description

Fix timezone condition in job that terminates ending subscription.


app/services/utils/timezone_service.rb
```diff
@@ -12,5 +12,11 @@ def self.date_in_customer_timezone_sql(customer, value)
 
       "#{sanitized_field_name}::timestamptz AT TIME ZONE '#{sanitized_timezone}'"
     end
+
+    def self.at_time_zone
```
vincent-pochet, line 16: :raised_hands:

vincent-pochet (approved):

PR #1300, opened 2023-09-08
feat (localization): fix translations for italian

binc0:
## Context

I corrected some translations of the first translation I provided after reviewing of the context of each element.

## Description

Some translation have been reviewed according to their effective context.


vincent-pochet (approved):
Thank you very much for these transalations @binc0 

It will be part of the next release.

PR #1309, opened 2023-09-12
misc(fix-pronto) Pin rugged gem version to 1.6.3

ivannovosad:
## Context

Pronto actions fail after updating rugged gem to 1.7.1


vincent-pochet (approved):
Nice catch! :rocket:

PR #1317, opened 2023-09-15
fix(stripe): Prevent new payment attempt when invoice is not in the right status

vincent-pochet:
## Context

We are receiving A LOT of `ActiveJob::Uniqueness::JobNotUnique` on ` PaymentProviders::Stripe::HandleEventJob`.

It happens when we tries to reprocess pending invoices for a customer after receiving a Stripe `setup_intent.succeeded`.

## Description

This PR is an attempt to reduce the noise in the Sidekiq dead queue by ensuring only finalized and ready for processing invoices are being retried.


rsempe (approved):
🙌

PR #1324, opened 2023-09-19
feat(override-plan): Add GET single subscription endpoint

rsempe:
## Context

We want to improve the way we are overriding a plan. Without working as a duplicate, by adding the logic for managing parent and children of plans.

## Description

The goal of this PR is to:
- add the endpoint `GET /api/v1/subscriptions/:external_id`
- add the GraphQL resolver for fetching a single subscription

### DIFF
--- a/app/controllers/api/v1/subscriptions_controller.rb
+++ b/app/controllers/api/v1/subscriptions_controller.rb
@@ -67,7 +67,10 @@ module Api
       end
 
       def show
-        subscription = current_organization.subscriptions.find_by(external_id: params[:external_id])
+        subscription = current_organization.subscriptions.find_by(
+          external_id: params[:external_id],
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/subscriptions_controller.rb
```diff
@@ -70,17 +60,19 @@ def update
...
-            ),
-          )
+          render_subscription(result.subscription)
         else
           render_error_response(result)
         end
       end
 
+      def show
+        subscription = current_organization.subscriptions.find_by(external_id: params[:external_id])
```
vincent-pochet, line 70: `external_id` could belongs to multiple subscription. Should we return only the active one? (and maybe have an ability to retrieve a terminated?)

vincent-pochet (approved):

PR #1338, opened 2023-09-21
feat(override-plan): Take into consideration children for customers_count

rsempe:
## Context

We want to improve the way we are overriding a plan. Without working as a duplicate, by adding the logic for managing parent and children of plans.

## Description

the goal of this PR is to include customers using overridden plans on `customers_count` of the parent plan.

### DIFF
--- a/app/graphql/types/plans/object.rb
+++ b/app/graphql/types/plans/object.rb
@@ -41,7 +41,12 @@ module Types
       end
 
       def customers_count
-        object.subscriptions.active.select(:customer_id).distinct.count
+        customers_count = object.subscriptions.active.select(:customer_id).distinct.count
+        return customers_count unless object.children
+
+        customers_count + model.children.joins(:subscriptions).where(
+          subscriptions: { status: :active },
+        ).select('distinct(customer_id)').count
       end
 
       def subscriptions_count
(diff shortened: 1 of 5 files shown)


app/graphql/types/plans/object.rb
```diff
@@ -41,7 +41,10 @@ def charges_count
       end
 
       def customers_count
-        object.subscriptions.active.select(:customer_id).distinct.count
+        customers_count = object.subscriptions.active.select(:customer_id).distinct.count
+        return customers_count unless object.children
+
+        customers_count + object.children.sum { |c| c.subscriptions.active.select(:customer_id).distinct.count }
```
vincent-pochet, line 47: Do we have a way to do it in a single DB query? Here we will be doing one query per child plan :/

app/serializers/v1/plan_serializer.rb
```diff
@@ -38,6 +39,13 @@ def charges
       ).serialize
     end
 
+    def customers_count
+      customers_count = model.subscriptions.active.select(:customer_id).distinct.count
+      return customers_count unless model.children
+
+      customers_count + model.children.sum { |c| c.subscriptions.active.select(:customer_id).distinct.count }
```
vincent-pochet, line 46: Same comment.
Maybe it could be good as well to share this logic in the model?

vincent-pochet (approved):

PR #1354, opened 2023-09-26
feat(event): Add external customer_id and external_subscription_id

vincent-pochet:
## Context

On the path for high volume improvements, we need to adapt the structure of the events table for better performance.

## Description

This PR adds `events#external_customer_id` and `events#external_subscription_id fields`. Event creations services are also updated in order to fill these fields with the appropriate value.

Note: A DB migration will come right after to populate theses fields for existing events

### DIFF
--- a/db/migrate/20230926132500_add_external_ids_to_events.rb
+++ b/db/migrate/20230926132500_add_external_ids_to_events.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddExternalIdsToEvents < ActiveRecord::Migration[7.0]
+  def change
+    change_table :events, bulk: true do |t|
+      t.string :external_customer_id
+      t.string :external_subscription_id
+    end
+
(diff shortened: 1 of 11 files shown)


db/migrate/20230926132500_add_external_ids_to_events.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddExternalIdsToEvents < ActiveRecord::Migration[7.0]
+  def change
+    add_column :events, :external_customer_id, :string
```
jdenquin, line 5: Rails/BulkChangeTable: You can use `change_table :events, bulk: true` to combine alter queries. (https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_table, https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Table.html)

jdenquin (approved):

PR #1366, opened 2023-10-03
hotfix: Subscription upgrade

vincent-pochet:
## Context

Some issues were recently reported with subscription upgrade with subscriptions attached to a plan payed in advance,

After investigation it appears that the issue is related to some jobs enqueued during the database transaction dealing with the upgrade process. As the jobs might be performed by sidekiq before the commit of the transaction, the result is unpredictable (subscription status in particular might not be the expected one) and the issue hard to reproduce.

## Description

The fix is to add all jobs in a queue and perform them only when the DB transaction lock is released.

For that purpose a new reusable utility concern was created named `Utils::TransactionalJobs`. It exposes two methods: `perform_later` to push the job into a queue and `perform_pending_jobs` to run all jobs when the transaction is released.

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -2,6 +2,8 @@
 
 module Subscriptions
   class CreateService < BaseService
+    include Utils::TransactionalJobs
+
     def initialize(customer:, plan:, params:)
       super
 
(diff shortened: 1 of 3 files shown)


vincent-pochet:
@jdenquin indeed, that would be great for some scenarios!

PR #1391, opened 2023-10-12
Add entries to swedish and spanish locales, fix translations in pdf

ivannovosad:
## Description

Add entries to swedish and spanish locales, fix translations in pdf.

### DIFF
--- a/app/views/templates/credit_note.slim
+++ b/app/views/templates/credit_note.slim
@@ -429,7 +429,7 @@ html
                 tr
                   td.body-1
                     - if item.fee.true_up_parent_fee_id?
-                      | #{I18n.t('invoice.true_up_metric', metric: item.fee.invoice_name)}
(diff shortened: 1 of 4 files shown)


app/views/templates/credit_note.slim
```diff
@@ -429,7 +429,7 @@ html
                 tr
                   td.body-1
                     - if item.fee.true_up_parent_fee_id?
-                      | #{I18n.t('invoice.true_up_metric', metric: item.fee.invoice_name)}
+                      | #{I18n.t('invoice.true_up_metric', metric: item.fee.true_up_parent_fee.invoice_name)}
```
ansmonjol, line 432: This change seems not to be related to this PR objective, does it?
Is that only a rebase issue maybe?

ansmonjol: discussed together, it's expected

PR #1394, opened 2023-10-12
feat(event): Ensure only one active subscription is active in post-processing

vincent-pochet:
## Context

On the path for high volume improvements, we need to adapt the structure of the events table for better performance.

## Description

This PR add a check on the number of active subscriptions when post-processing an event. If multiple subscriptions are identified, the event is not attached to any subscription.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -114,10 +114,10 @@ class Subscription < ApplicationRecord
 
   def validate_external_id
     return unless active?
+    return unless organization.subscriptions.active.exists?(external_id:)
 
     # NOTE: We want unique external id per organization.
-    used_ids = organization.subscriptions.active.pluck(:external_id)
-    errors.add(:external_id, :value_already_exist) if used_ids&.include?(external_id)
+    errors.add(:external_id, :value_already_exist)
   end
 
   def downgrade_plan_date
--- a/app/services/events/post_process_service.rb
+++ b/app/services/events/post_process_service.rb
@@ -11,8 +11,15 @@ module Events
     def call
       event.customer_id = customer&.id
       event.external_customer_id ||= customer&.external_id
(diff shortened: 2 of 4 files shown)


app/models/subscription.rb
```diff
@@ -116,8 +116,9 @@ def validate_external_id
     return unless active?
 
     # NOTE: We want unique external id per organization.
-    used_ids = organization.subscriptions.active.pluck(:external_id)
-    errors.add(:external_id, :value_already_exist) if used_ids&.include?(external_id)
+    if organization.subscriptions.active.exists?(external_id:)
```
jdenquin, line 119: Style/GuardClause: Use a guard clause (`return unless organization.subscriptions.active.exists?(external_id:)`) instead of wrapping the code inside a conditional expression. (https://rubystyle.guide#no-nested-conditionals)

jdenquin, line 119: Style/IfUnlessModifier: Favor modifier `if` usage when having a single-line body. Another good alternative is the usage of control flow `&&`/`||`. (https://rubystyle.guide#if-as-a-modifier)

app/services/events/post_process_service.rb
```diff
@@ -11,8 +11,15 @@ def initialize(event:)
     def call
       event.customer_id = customer&.id
       event.external_customer_id ||= customer&.external_id
-      event.subscription_id = subscriptions.first&.id
-      event.external_subscription_id ||= subscriptions.first&.external_id
+
+      # NOTE: prevent subscription if more than 1 subscription is active
+      #       if multiple terminated matches the timestamp, takes the most recent
+      if subscriptions.select(&:active?).count <= 1
```
jdenquin, line 17: Performance/Count: Use `count` instead of `select...count`.

jdenquin (approved):

PR #1399, opened 2023-10-16
feat: Send started and terminated subscription webhooks

rsempe:
The goal of this PR is to send `subscription.started` and `subscription.terminated` on upgrade and downgrade.

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -147,21 +147,9 @@ module Subscriptions
 
       # NOTE: When upgrading, the new subscription becomes active immediatly
       #       The previous one must be terminated
-      current_subscription.mark_as_terminated!
+      Subscriptions::TerminateService.call(subscription: current_subscription)
       new_subscription.mark_as_active!
-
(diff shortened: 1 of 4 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -147,21 +147,9 @@ def upgrade_subscription
 
       # NOTE: When upgrading, the new subscription becomes active immediatly
       #       The previous one must be terminated
-      current_subscription.mark_as_terminated!
+      Subscriptions::TerminateService.call(subscription: current_subscription)
```
vincent-pochet, line 150: We have to take care of Sidekiq jobs as we are in a database transaction context.

We had issue lately with billing jobs on terminated subscription processed before the commit of the DB transaction, leading to wrong calculation on the invoice as the job was not aware the subscription was terminated.

That' why jobs were store in a temporary queue with `perform_later` private method, and called only after the transaction commit. See https://github.com/getlago/lago-api/pull/1366 for more details

vincent-pochet (approved):

PR #1404, opened 2023-10-17
feat(event): Basic clickhouse database config

vincent-pochet:
## Context

On the path for high volume improvements, we are introducing the clickhouse database engine.

## Description

This PR is the first of the feature. It introduces a basic clickhouse config/model for the development environment.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem 'oauth2'
 gem 'rack-cors'
 
 # Database
+gem 'clickhouse-activerecord'
 gem 'discard', '~> 1.2'
 gem 'kaminari-activerecord'
 gem 'paper_trail'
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -28,6 +28,7 @@ gem 'paper_trail'
 gem 'pg'
 gem 'ransack', '~> 4.0.0'
 gem 'with_advisory_lock'
+gem 'clickhouse-activerecord'
```
jdenquin, line 31: Bundler/OrderedGems: Gems should be sorted in an alphabetical order within their section of the Gemfile. Gem `clickhouse-activerecord` should appear before `with_advisory_lock`.

app/models/clickhouse/event_raw.rb
```diff
@@ -0,0 +1,5 @@
+module Clickhouse
```
jdenquin, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

db/clickhouse_schema.rb
```diff
@@ -0,0 +1,15 @@
+# This file is auto-generated from the current state of the database. Instead
```
jdenquin, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

PR #1405, opened 2023-10-19
Setup PSP redirect_url

ivannovosad:
## Context

End customers are redirected to [www.getlago.com](http://www.getlago.com/) after doing a payment intent on our users payment flow.

## Description

Add redirect urls where the customers are redirected to after performing a payment intent.


app/graphql/mutations/payment_providers/adyen/create.rb
```diff
@@ -0,0 +1,16 @@
...
+module Mutations
+  module PaymentProviders
+    module Adyen
+      class Create < Base
+        graphql_name 'AddAdyenPaymentProvider'
+        description 'Add Adyen payment provider'
+
+        input_object_class Types::PaymentProviders::AdyenInput
+
+        type Types::PaymentProviders::Adyen
```
lovrocolic, line 12: maybe this can also be moved into parent class

app/services/payment_provider_customers/stripe_service.rb
```diff
@@ -134,13 +134,18 @@ def api_key
 
     def checkout_link_params
       {
-        success_url: CHECKOUT_SUCCESS_URL,
```
lovrocolic, line 137: maybe we can remove this const now

lovrocolic (approved):
Few questions but looks good 👌

app/validators/adyen_url_validator.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AdyenUrlValidator < ActiveModel::EachValidator
```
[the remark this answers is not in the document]
ivannovosad: Yes, we're using it [here](https://github.com/getlago/lago-api/blob/feat-psp-redirect-url/app/models/payment_providers/adyen_provider.rb#L8)

app/graphql/mutations/payment_providers/adyen/create.rb
ivannovosad: That is how I originally wanted to implement it @lovrocolic 
But the create and update mutations use different input/argument. Also when I move the type to their parent class it autogenerates the return type of the mutation:

```
          "kind": "OBJECT",
          "name": "AddAdyenPaymentProviderPayload",
          "description": "Autogenerated return type of AddAdyenPaymentProvider",

```

lovrocolic (approved):

PR #1408, opened 2023-10-24
misc(stripe-unauthorized): Rescue Stripe::AuthenticationError error

ivannovosad:
## Context

When Stripe hasn’t been set up in the settings, or the Stripe API key is invalid, and the user tries to create a customer account through integration, the API returns a 500.

## Description

Return a 401 error with a message that Stripe authentication failed.


vincent-pochet (approved):
:rocket:

PR #1414, opened 2023-10-25
fix (date service): update guard clause in charges_from_datetime method

lovrocolic:
This small fix is preventing raising error if we are in first billing period and if there is no generated invoices yet.

### DIFF
--- a/app/services/subscriptions/dates_service.rb
+++ b/app/services/subscriptions/dates_service.rb
@@ -63,7 +63,7 @@ module Subscriptions
       # NOTE: If customer applicable timezone changes during a billing period, there is a risk to double count events
       #       or to miss some. To prevent it, we have to ensure that invoice bounds does not overlap or that there is no
       #       hole bewtween a charges_from_datetime and the charges_to_datetime of the previous period
-      if timezone_has_changed?
+      if timezone_has_changed? && previous_charge_to_datetime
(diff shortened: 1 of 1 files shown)


app/services/subscriptions/dates_service.rb
```diff
@@ -151,7 +151,7 @@ def timezone_has_changed?
     end
 
     def previous_charge_to_datetime
-      return if last_invoice_subscription.blank?
+      return nil if last_invoice_subscription.blank?
```
ivannovosad, line 154: This change is not necessary, we can keep the original which does the same thing:

`return if last_invoice_subscription.blank?`

spec/services/subscriptions/dates/monthly_service_spec.rb
```diff
@@ -310,6 +310,18 @@
             expect(result).to match_datetime('2022-02-01 00:00:00')
           end
         end
+
+        context 'when timezone has changed and there is no invoices generated in the past' do
```
ivannovosad, line 314:
```suggestion
        context 'when timezone has changed and there are no invoices generated in the past' do
```

ivannovosad (changes requested):
Looks good to me, I just added some suggestions.

ivannovosad (approved):

PR #1416, opened 2023-10-26
fix(event): Handle ActiveRecord::RecordNotUnique error

vincent-pochet:
## Context

Since move to post processing of events, we are receiving some `ActiveRecord::RecordNotUnique` in sidekiq jobs.

As these error are only related to duplicated events, we should not see them in dead queue.

## Description

This PR handle the error, sends a webhook back to the organization and just ignore the event.


app/services/events/post_process_service.rb
```diff
@@ -38,6 +38,10 @@ def call
     rescue ActiveRecord::RecordInvalid => e
       delivor_error_webhook(error: e.record.errors.messages)
 
+      result
```
jdenquin, line 41: Lint/UselessAssignment: Useless assignment to variable - `e`. (https://rubystyle.guide#underscore-unused-vars)

lovrocolic (approved):

PR #1417, opened 2023-10-26
fix(adyen-no-payment-methods): Send an error webhook when no payment methods are active

ivannovosad:
## Context

When receiving an error from Adyen (422 Error), we do not log it and we do not return a webhook with the error. We should send a webhook with the error, so the user know about it.

### DIFF
--- a/app/services/payment_provider_customers/adyen_service.rb
+++ b/app/services/payment_provider_customers/adyen_service.rb
@@ -12,7 +12,10 @@ module PaymentProviderCustomers
       result.adyen_customer = adyen_customer
       return result if adyen_customer.provider_customer_id?
 
-      result.checkout_url = generate_checkout_url.checkout_url
+      checkout_url_result = generate_checkout_url
+      return result unless checkout_url_result.success?
+
+      result.checkout_url = checkout_url_result.checkout_url
       result
     end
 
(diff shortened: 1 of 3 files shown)


app/services/payment_provider_customers/adyen_service.rb
```diff
@@ -12,14 +12,20 @@ def create
       result.adyen_customer = adyen_customer
       return result if adyen_customer.provider_customer_id?
 
-      result.checkout_url = generate_checkout_url.checkout_url
+      checkout_url_result = generate_checkout_url
+      return result unless checkout_url_result.success?
+
+      result.checkout_url = checkout_url_result.checkout_url
```
lovrocolic, line 18: As I can see checkout_url has already been attached to result in the `generate_checkout_url` method.

IMO `generate_checkout_url` can just return `url` (since the `result` is the same one across the class).
On line 27 maybe we can just have `return nil unless result.success?`

On lines 15,16 maybe we can just have:
```
checkout_url = generate_checkout_url
return result unless result.success?

result.checkout_url = checkout_url
```

ivannovosad: @lovrocolic I see your point, but `generate_checkout_url ` is also used [here](https://github.com/getlago/lago-api/blob/8dc4d45a796886ee20db91e8a321067836078660/app/jobs/payment_provider_customers/adyen_checkout_url_job.rb#L11). And it needs to contain the result object itself for that case.

vincent-pochet (approved):

PR #1420, opened 2023-10-27
fix (customers): fix customer create flow when sync_with_provider is true

lovrocolic:
If `sync_with_provider` is true, there is no payment provider for customer but payload is missing payment_provider attribute the error is raised.

We need to add guard clause for that case


vincent-pochet (approved):

ivannovosad (approved):

jdenquin (approved):

jdenquin:
🚀

PR #1422, opened 2023-10-27
feat(billable-metric): Use Factory pattern to initialize aggregators

vincent-pochet:
## Context

With the coming high usage event ingestion, aggregator classes will have to be adapted to target the right event store (PG or Clickhouse) depending on the context.

To make it easier, and since current Aggregator initialization is quite complex, a small cleanup has to be performed in order to simplify the logic.

## Description

This PR takes advantage of the factory pattern to initialize the right aggregator for a charge.

It will also soon be responsible for passing the right Event store

### DIFF
--- a/app/services/billable_metrics/aggregation_factory.rb
+++ b/app/services/billable_metrics/aggregation_factory.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+
+module BillableMetrics
+  class AggregationFactory
+    def self.new_instance(charge:, current_usage: false, **attributes)
+      aggregator_class(charge, current_usage).new(
+        billable_metric: charge.billable_metric,
+        **attributes,
+      )
+    end
+
+    def self.aggregator_class(charge, current_usage)
+      case charge.billable_metric.aggregation_type.to_sym
+      when :count_agg
+        BillableMetrics::Aggregations::CountService
+
+      when :latest_agg
+        raise(NotImplementedError) if charge.pay_in_advance? && !current_usage
+
+        BillableMetrics::Aggregations::LatestService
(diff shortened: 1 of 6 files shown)


app/services/billable_metrics/aggregation_factory.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+module BillableMetrics
+  class AggregationFactory
+    def self.new_instance(charge:, current_usage: false, **attributes)
+      aggregator_class(charge, current_usage).new(
+        billable_metric: charge.billable_metric,
+        **attributes
```
jdenquin, line 8: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

app/services/billable_metrics/aggregation_factory.rb
```diff
@@ -0,0 +1,58 @@
...
+      )
+    end
+
+    private
```
jdenquin, line 12: Lint/UselessAccessModifier: Useless `private` access modifier.

app/services/billable_metrics/aggregation_factory.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def self.aggregator_class(charge, current_usage)
```
jdenquin, line 14: Lint/IneffectiveAccessModifier: `private` (on line 12) does not make singleton methods private. Use `private_class_method` or `private` inside a `class << self` block instead.

rsempe (approved):

PR #1423, opened 2023-10-30
fix(customer): Allign timezone validation with GraphQL list

vincent-pochet:
## Context

We have some issues with some customers having timezone not accepted by the GraphQL Timezone enum, leading to an issue with the front end application when trying to fetch the customers. 

It happens because the timezone list used for validation (full TZ list), is not the same as the one used on GraphQL/front (Rails simplified list). If means that some customers could be created via API with a timezone not recognized by the front.

## Description

This PR ensures that both paths uses the same TZ list (the rails simpliefied one)


app/validators/timezone_validator.rb
```diff
@@ -8,6 +8,8 @@ def validate_each(record, attribute, value)
   protected
 
   def valid?(value)
-    value && ActiveSupport::TimeZone[value].present?
+    return true if value == 'UTC'
+
+    value && ActiveSupport::TimeZone::MAPPING.values.include?(value)
```
jdenquin, line 13: Performance/InefficientHashSearch: Use `#value?` instead of `#values.include?`. (https://github.com/JuanitoFatas/fast-ruby#hashkey-instead-of-hashkeysinclude-code)

rsempe (approved):

app/validators/timezone_validator.rb
```diff
@@ -8,6 +8,8 @@ def validate_each(record, attribute, value)
   protected
 
   def valid?(value)
-    value && ActiveSupport::TimeZone[value].present?
+    return true if value == 'UTC'
+
+    value && ActiveSupport::TimeZone::MAPPING.value?(value)
```
ivannovosad, line 13:
```suggestion
    value == 'UTC' || ActiveSupport::TimeZone::MAPPING.value?(value)
```

ivannovosad (approved):

PR #1427, opened 2023-10-31
feat(clickhouse): Add events_raw_mv to fill events_raw

vincent-pochet:
## Context

On the path for high volume improvements, we are introducing the clickhouse database engine.

## Description

This PR changes turns `events_raw#properties` field type from `string` to `Map(String, String)` and adds the `events_raw_mv` to fill the `events_raw`


db/clickhouse_migrate/20231030163703_create_events_raw_mv.rb
```diff
@@ -0,0 +1,17 @@
+class CreateEventsRawMv < ActiveRecord::Migration[7.0]
```
jdenquin, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

jdenquin (approved):

PR #1431, opened 2023-11-02
feat(default-currency): Add organization default currency

ivannovosad:
## Context

To display revenue graphs, we need to know what is the default currency of an organization. This default currency will be used as a filter when landing on this dashboard section.

## Description

This PR adds `default_currency` column to `organizations` table.

It also adds the field to organization graphql type and update input, serializer and the organization update service.

### DIFF
--- a/app/serializers/v1/organization_serializer.rb
+++ b/app/serializers/v1/organization_serializer.rb
@@ -6,6 +6,7 @@ module V1
       payload = {
         lago_id: model.id,
         name: model.name,
+        default_currency: model.default_currency,
         created_at: model.created_at.iso8601,
         webhook_url: webhook_urls.first.to_s,
         webhook_urls:,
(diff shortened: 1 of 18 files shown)


app/serializers/v1/organization_serializer.rb
```diff
@@ -52,8 +53,6 @@ def taxes
       ).serialize
     end
 
-    private
```
rsempe, line 55: why?

ivannovosad: @rsempe There's already `private` [here](https://github.com/getlago/lago-api/pull/1431/files/6487a62bfb855fae48009ced48c560baa70b8aaa#diff-d777209c00ddebab1911162ee907d61ac10c1f7ab087ab53f837e5d97bea9d13L33)

PR #1433, opened 2023-11-02
fix(invoice): Fix create invoice request spec

ivannovosad:
## Context

Create invoice request spec was failing sometimes.

## Description

Addon fees were rarely created in an opposite order, we're now finding the fee by code to be sure to get the correct one.


rsempe (approved):

vincent-pochet (approved):
:heart:

PR #1443, opened 2023-11-07
misc: Include group properties when overriding a plan

rsempe:
The goal of this PR is to avoid having N+1 queries when overriding a plan with group properties.

### DIFF
--- a/app/services/plans/override_service.rb
+++ b/app/services/plans/override_service.rb
@@ -29,7 +29,7 @@ module Plans
           return taxes_result unless taxes_result.success?
         end
 
-        plan.charges.each do |charge|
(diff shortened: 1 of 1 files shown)


app/services/plans/override_service.rb
```diff
@@ -29,7 +29,7 @@ def call
           return taxes_result unless taxes_result.success?
         end
 
-        plan.charges.each do |charge|
+        plan.charges.includes(:group_properties).each do |charge|
```
jdenquin, line 32: Rails/FindEach: Use `find_each` instead of `each`. (https://rails.rubystyle.guide#find-each)

vincent-pochet (approved):

PR #1447, opened 2023-11-07
fix (events): add migration for incorrectly discarded events

lovrocolic:
## Description

This migration undiscards all events that got incorrectly discarded and that affect future invoices

### DIFF
--- a/db/migrate/20231107110809_undiscard_incorrectly_deleted_events.rb
+++ b/db/migrate/20231107110809_undiscard_incorrectly_deleted_events.rb
@@ -9,14 +9,12 @@ class UndiscardIncorrectlyDeletedEvents < ActiveRecord::Migration[7.0]
             SELECT events.id AS event_id
             FROM events
             INNER JOIN subscriptions ON subscriptions.id = events.subscription_id
-            INNER JOIN plans ON plans.id = subscriptions.plan_id
             INNER JOIN billable_metrics ON billable_metrics.code = events.code
-            WHERE (events.timestamp::timestamp(0) >= '2023-10-06')
+            WHERE (events.timestamp::timestamp(0) >= '2022-11-06')
               AND events.deleted_at IS NOT NULL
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):

db/migrate/20231107110809_undiscard_incorrectly_deleted_events.rb
```diff
@@ -0,0 +1,30 @@
...
+            FROM events
+            INNER JOIN subscriptions ON subscriptions.id = events.subscription_id
+            INNER JOIN plans ON plans.id = subscriptions.plan_id
+            INNER JOIN billable_metrics ON billable_metrics.code = events.code
+            WHERE (events.timestamp::timestamp(0) >= '2023-10-06')
+              AND events.deleted_at IS NOT NULL
+              AND events.deleted_at > billable_metrics.created_at
+              AND billable_metrics.deleted_at IS NULL
+              AND subscriptions.status IN (0, 1) -- pending and active subscriptions
+              AND plans.interval IN (0, 1) -- weekly and monthly intervals (there is no quarterly and yearly subs)
```
rsempe, line 19: we don't really know for OSS 😬

vincent-pochet: Good catch!

vincent-pochet:
One thing we could do, since we know the request might take some time to perform on the US cloud cluster, is to perform the migration manually in the rails console (and to inject the version in the DB schema_migration table) before deploying the migration, so that is does block the app startup.

db/migrate/20231107110809_undiscard_incorrectly_deleted_events.rb
lovrocolic: Oh yes 👍 Hmm if we remove this condition probably we also have to take into account all events from the beginning of the year (if there is some yearly plan in OSS)

PR #1448, opened 2023-11-07
fix(credit-note): Fix credit note estimate taxe rate

vincent-pochet:
## Context

Since it's a complicated task to create a credit note with coupons, multiple tax rates and other complex computation, we recently introduce a `CreditNotes::EstimateService` and the corresponding API/GraphQL endpoint to interact with it.

However, it appears that the computation is not always correct when different taxes are applied to an invoice.

## Description

This PR tries the fix this situation by fixing the computed tax rate on the credit note and by exposing the `precise_taxes_amount_cents` into the GraphQL query.

### DIFF
--- a/app/graphql/types/credit_notes/estimate.rb
+++ b/app/graphql/types/credit_notes/estimate.rb
@@ -11,10 +11,12 @@ module Types
       field :coupons_adjustment_amount_cents, GraphQL::Types::BigInt, null: false
       field :max_creditable_amount_cents, GraphQL::Types::BigInt, method: :credit_amount_cents, null: false
       field :max_refundable_amount_cents, GraphQL::Types::BigInt, method: :refund_amount_cents, null: false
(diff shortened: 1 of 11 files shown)


app/graphql/types/credit_notes/estimate.rb
```diff
@@ -13,6 +13,7 @@ class Estimate < Types::BaseObject
       field :max_refundable_amount_cents, GraphQL::Types::BigInt, method: :refund_amount_cents, null: false
       field :sub_total_excluding_taxes_amount_cents, GraphQL::Types::BigInt, null: false
       field :taxes_amount_cents, GraphQL::Types::BigInt, null: false
+      field :precise_taxes_amount_cents, GraphQL::Types::Float, null: false
```
jdenquin, line 16: GraphQL/OrderedFields: Fields should be sorted in an alphabetical order within their section. Field `precise_taxes_amount_cents` should appear before `taxes_amount_cents`.

rsempe (approved):

PR #1452, opened 2023-11-09
feat(events): Add Materialized view for post validation

vincent-pochet:
## Context

Since we no longer validate event ingestion, users are flying blind to understand which events have been ingested but not used for billing.

Ideally, our users should receive warnings when issues are detected during event ingestion.

## Description

This PR is the first step of the feature.
It adds a new Clock job (not activated yet) that will be executed every hour to:
- Refresh a materialized view with the events from the last terminated hour
- (later) enqueue a validation job for each organization that have received at least one event

### DIFF
--- a/app/jobs/clock/events_validation_job.rb
+++ b/app/jobs/clock/events_validation_job.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Clock
+  class EventsValidationJob < ApplicationJob
+    queue_as 'clock'
+
+    def perform
+      # NOTE: refresh the last hour events materialized view
+      Scenic.database.refresh_materialized_view(
+        Events::LastHourMv.table_name,
+        concurrently: false,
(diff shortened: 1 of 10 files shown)


app/jobs/clock/events_validation_job.rb
```diff
@@ -0,0 +1,18 @@
...
+module Clock
+  class EventsValidationJob < ApplicationJob
+    queue_as 'clock'
+
+    def perform
+      # NOTE: refresh the last hour events materialized view
+      Scenic.database.refresh_materialized_view(
+        Events::LastHourMv.table_name,
+        concurrently: false,
+        cascade: false
```
jdenquin, line 12: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

jdenquin (approved):

PR #1459, opened 2023-11-13
fix: Ability to update charge from subscription

rsempe:
Currently, updating a charge from a subscription is not working because `billable_metric_id` is missing.
The goal of this PR is to fix it by accepting `billableMetricId`.

### DIFF
--- a/app/graphql/types/subscriptions/charge_overrides_input.rb
+++ b/app/graphql/types/subscriptions/charge_overrides_input.rb
@@ -3,6 +3,7 @@
 module Types
   module Subscriptions
     class ChargeOverridesInput < Types::BaseInputObject
+      argument :billable_metric_id, ID, required: true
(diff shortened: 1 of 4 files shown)


ansmonjol (approved):

app/graphql/types/subscriptions/charge_overrides_input.rb
```diff
@@ -4,6 +4,7 @@ module Types
   module Subscriptions
     class ChargeOverridesInput < Types::BaseInputObject
       argument :id, ID, required: true
+      argument :billable_metric_id, ID, required: true
```
jdenquin, line 7: GraphQL/OrderedArguments: Arguments should be sorted in an alphabetical order within their section. Field `billable_metric_id` should appear before `id`.

PR #1467, opened 2023-11-14
Analytics mrr

ivannovosad:
## Context

No revenue calculation is done with Lago.

## Description

This PR adds MRR model, controller, serializer and graphql resources.


app/controllers/api/v1/analytics/mrrs_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def filters
+          {
+            currency: params[:currency]&.upcase,
+            months: params[:months].to_i,
+          }
+        end
+
+        def authorize
+          return forbidden_error(code: 'premium_feature') unless License.premium?
```
jdenquin, line 31: Style/RedundantReturn: Redundant `return` detected. (https://rubystyle.guide#no-explicit-return)

app/controllers/api/v1/analytics/mrrs_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def filters
+          {
+            currency: params[:currency]&.upcase,
+            months: params[:months].to_i,
+          }
+        end
+
+        def authorize
+          forbidden_error(code: 'premium_feature') unless License.premium?
```
lovrocolic, line 31: I think that it would be good to follow common pattern here. Usually license check is placed in service, so that we can simply return `result.forbidden_failure!`. In that case error code will be set to `feature_unavailable` by default. In controller you can simply check if service result is success and return r`ender_error_response(result)` if it is not successful

lovrocolic:
I have just one more general comment regarding naming (sorry for not mentioning it during dive-in). IMO using `Mrr` in bunch of places does not sound clear enough. Not a blocker, but maybe we should consider using full versions of name... What do you think? Did you have other options in mind?

app/controllers/api/v1/analytics/mrrs_controller.rb
ivannovosad: @lovrocolic 
cc: @ansmonjol 
Lets move it to a service but I'd recommend to do it in a separate refactoring PR after this one's merged. 
This change would require 2 other (already merged) PRs to follow this as well.

ivannovosad:
@lovrocolic We can use `MonthlyRecurringRevenue`. Then we'll have to change it also in Open API and all the API clients.

app/controllers/api/v1/analytics/mrrs_controller.rb
ivannovosad: @lovrocolic I added the service in this PR but I'll another PR to fix the rest.

lovrocolic (approved):

PR #1473, opened 2023-11-17
fix(subscriptions): fix termination boundaries

lovrocolic:
## Description

When anniversery period and when `terminated_at` is the same as `started_at`, termination boundaries haven't been calculated correctly.

This PR fixes this issue and adds scenario test for this case.

### DIFF
--- a/app/services/invoices/calculate_fees_service.rb
+++ b/app/services/invoices/calculate_fees_service.rb
@@ -296,6 +296,8 @@ module Invoices
       current_time = Time.zone.at(timestamp)
       current_time_in_timezone = Time.zone.at(timestamp).in_time_zone(customer.applicable_timezone)
 
+      return boundaries if (current_time_in_timezone - 1.day) < subscription.started_at
+
       dates_service = Subscriptions::DatesService.new_instance(duplicate, current_time - 1.day, current_usage: true)
 
       return boundaries if current_time_in_timezone < dates_service.charges_to_datetime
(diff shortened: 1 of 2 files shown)


spec/scenarios/subscriptions/terminate_ended_spec.rb
```diff
@@ -178,6 +178,46 @@
...
+              subscription_at: subscription_at.iso8601,
+              ending_at: nil,
+            },
+          )
+
+          subscription = customer.subscriptions.first
+          expect(subscription).to be_active
+        end
+
+        Organization.update_all(webhook_url: nil) # rubocop:disable Rails/SkipsModelValidationss
```
jdenquin, line 201: Rails/SkipsModelValidations: Avoid using `update_all` because it skips validations. (https://guides.rubyonrails.org/active_record_validations.html#skipping-validations)

ivannovosad (approved):

app/services/invoices/calculate_fees_service.rb
```diff
@@ -296,6 +296,8 @@ def termination_boundaries(subscription, boundaries)
       current_time = Time.zone.at(timestamp)
       current_time_in_timezone = Time.zone.at(timestamp).in_time_zone(customer.applicable_timezone)
 
+      return boundaries if (current_time_in_timezone - 1.day) < subscription.started_at
```
rsempe, line 299: can we add a comment to explain it here? 
not sure to understand next time I'll see this line 😅

rsempe (approved):

vincent-pochet (approved):

PR #1475, opened 2023-11-17
feat(analytics): Refactor controllers and services

ivannovosad:
## Context

Analytics controllers are not using services to handle authorization.

## Description

This PR adds services to handle authorization for analytics controllers.


lovrocolic (approved):
🚀

PR #1485, opened 2023-11-23
fix(adyen): Fix adyen service when card is invalid

ivannovosad:
## Context

Creating adyen payment fails as a result of changing payment method to a card with invalid number.

## Description

This PR takes care of not raising an error when that happens.

### DIFF
--- a/app/services/invoices/payments/adyen_service.rb
+++ b/app/services/invoices/payments/adyen_service.rb
@@ -98,12 +98,8 @@ module Invoices
           Lago::Adyen::Params.new(payment_method_params).to_h,
         ).response
 
-        if (payment_method_id = result['storedPaymentMethods']&.first&.dig('id'))
-          customer.adyen_customer.update!(payment_method_id:)
-        end
-      rescue Adyen::AdyenError => e
-        deliver_error_webhook(e)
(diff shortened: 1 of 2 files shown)


app/services/invoices/payments/adyen_service.rb
```diff
@@ -98,12 +98,8 @@ def update_payment_method_id
           Lago::Adyen::Params.new(payment_method_params).to_h,
         ).response
 
-        if (payment_method_id = result['storedPaymentMethods']&.first&.dig('id'))
-          customer.adyen_customer.update!(payment_method_id:)
-        end
-      rescue Adyen::AdyenError => e
-        deliver_error_webhook(e)
```
jdenquin, line 105: we do not send the webhook anymore?

ivannovosad: We do it here instead: https://github.com/getlago/lago-api/pull/1485/files/85e33c7334cf0f76a8b4355ea65006a059a0de39#diff-43fd964e38ba8c7b757a07adaae23508fbece49763232b081bc36136adb041deR110

jdenquin (approved):

PR #1490, opened 2023-11-24
fix(checkout-url): Fix when the checkout url should be generated

ivannovosad:
## Context

Cannot link customers to existing Adyen shoppers

## Description

This PR fixes a bug when the checkout url is generated and the customer is linked to a payment provider customer.


vincent-pochet (approved):
:rocket:

PR #1495, opened 2023-11-27
feat(invoice-numbering): Invoice numbering db changes

lovrocolic:
## Context

Currently Lago does not provide solution for updating document numbers. Also, `per_customer` is the only approach in creating incremental numbers. With this feature, it will be allowed to update one part of document number. Also, it will be possible to choose between `per_customer` and `per_organization` numbering 

## Description

This PR adds needed DB fields and migration.


db/migrate/20231123105540_add_fields_for_new_document_numbering.rb
```diff
@@ -0,0 +1,36 @@
+class AddFieldsForNewDocumentNumbering < ActiveRecord::Migration[7.0]
```
jdenquin, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

vincent-pochet (approved):

PR #1503, opened 2023-11-30
feat(taxes): Add EU VAT Taxe management field on Organizations

jdenquin:
## Context

- Part 1 of the EU VAT auto management
- We want to support VAT autodetection for european organizations.

## Description

- Add the `eu_tax_management` field on `organizations`

### DIFF
--- a/app/graphql/types/organizations/billing_configuration.rb
+++ b/app/graphql/types/organizations/billing_configuration.rb
@@ -6,6 +6,7 @@ module Types
       graphql_name 'OrganizationBillingConfiguration'
 
       field :document_locale, String
+      field :eu_tax_management, Boolean
       field :id, ID, null: false
       field :invoice_footer, String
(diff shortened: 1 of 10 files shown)


ansmonjol (approved):

app/graphql/types/organizations/billing_configuration.rb
```diff
@@ -6,6 +6,7 @@ class BillingConfiguration < Types::BaseObject
       graphql_name 'OrganizationBillingConfiguration'
 
       field :document_locale, String
+      field :eu_tax_management, Boolean
```
vincent-pochet, line 9: We are trying to get rid of the "billing_configuration" as it's not a real object. Shouldn't we move the attribute directly at organization level?

Also, I would suggest to name it: `eu_customer_taxes_autodetection`

vincent-pochet (approved):

PR #1508, opened 2023-12-01
fix: reduce lock scope for sequence_id generation

vincent-pochet:
## Context

On billing day, a lot of `BillSubscriptionJob` are failing with `Sequenced::SequenceError: Unable to aquire lock on the database`

It appears this issue is related to a lock created for ensuring the uniqueness of the `sequential_id`. The lock uses `invoices_lock` as name and so is shared between all invoices, locking the entire table...

## Description

Since the `lock` logic is scoped at customer level, it would make more sense to use the `customer_id` in the lock key, to lock only invoices attached to the customer

### DIFF
--- a/app/models/concerns/sequenced.rb
+++ b/app/models/concerns/sequenced.rb
@@ -18,7 +18,7 @@ module Sequenced
 
     def generate_sequential_id
       result = self.class.with_advisory_lock(
-        "#{self.class.name.underscore}_lock",
+        lock_key_value,
         transaction: true,
         timeout_seconds: 10.seconds,
       ) do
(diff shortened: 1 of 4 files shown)


app/models/concerns/sequenced.rb
```diff
@@ -55,6 +60,14 @@ def sequenced_options=(options)
     def sequenced_options
       @sequenced_options
     end
+
+    def sequenced_lock_key=(lock_key)
+      @sequenced_lock_key = lock_key
```
jdenquin, line 65: ThreadSafety/InstanceVariableInClassMethod: Avoid instance variables in class methods.

app/models/concerns/sequenced.rb
```diff
@@ -55,6 +60,14 @@ def sequenced_options=(options)
...
+    end
+
+    def sequenced_lock_key
+      @sequenced_lock_key
```
jdenquin, line 69: ThreadSafety/InstanceVariableInClassMethod: Avoid instance variables in class methods.

jdenquin (approved):

PR #1509, opened 2023-12-01
feat (document-numbering): add new logic for switching between two document numbering types

lovrocolic:
## Context

Currently only one document numbering logic is present. However with this feature there will be possibility to choose between `per_customer` and `per_organization` type. Also, it will be possible to change numbering type at any time.

## Description

This PR adds main logic that enables switching between two numbering types


app/models/concerns/sequenced.rb
```diff
@@ -5,8 +5,10 @@ module Sequenced
 
   included do
     scope :with_sequential_id, -> { where.not(sequential_id: nil) }
+    scope :with_org_sequential_id, -> { where.not(organization_sequential_id: nil) }
 
     before_save :ensure_sequential_id
+    before_save :ensure_organization_sequential_id, if: -> { self.class.to_s == 'Invoice' }
```
jdenquin, line 11: Style/ClassEqualityComparison: Use `instance_of?(::Invoice)` instead of comparing classes. (https://rubystyle.guide#instance-of-vs-class-comparison)

app/models/concerns/sequenced.rb
```diff
@@ -16,6 +18,12 @@ def ensure_sequential_id
       self.sequential_id = generate_sequential_id
     end
 
+    def ensure_organization_sequential_id
+      return if organization_sequential_id.present? && organization_sequential_id > 0
```
jdenquin, line 22: Style/NumericPredicate: Use `organization_sequential_id.positive?` instead of `organization_sequential_id > 0`. (https://rubystyle.guide#predicate-methods)

app/models/invoice.rb
```diff
@@ -68,7 +68,8 @@ class Invoice < ApplicationRecord
     end
   end
 
-  sequenced scope: ->(invoice) { invoice.customer.invoices }
+  sequenced scope: ->(invoice) { invoice.customer.invoices },
+            organization_scope: ->(invoice) { invoice.organization.invoices.where(created_at: Time.now.all_month) }
```
jdenquin, line 72: Rails/TimeZone: Do not use `Time.now` without zone. Use one of `Time.zone.now`, `Time.current`, `Time.now.in_time_zone`, `Time.now.utc`, `Time.now.getlocal`, `Time.now.xmlschema`, `Time.now.iso8601`, `Time.now.jisx0301`, `Time.now.rfc3339`, `Time.now.httpdate`, `Time.now.to_i`, `Time.now.to_f` instead. (https://rails.rubystyle.guide#time, http://danilenko.org/2012/7/6/rails_timezones)

spec/scenarios/invoices/invoice_numbering_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+            ORG-1-002-003
+            ORG-1-003-003
+            ORG-1-202310-001
+            ORG-1-202310-002
+            ORG-1-202310-003
+            ORG-1-001-005
+            ORG-1-002-005
+            ORG-1-003-005
+            ORG-1-002-006
+          ]
```
jdenquin, line 178: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

app/models/concerns/sequenced.rb
```diff
@@ -5,8 +5,10 @@ module Sequenced
 
   included do
     scope :with_sequential_id, -> { where.not(sequential_id: nil) }
+    scope :with_org_sequential_id, -> { where.not(organization_sequential_id: nil) }
 
     before_save :ensure_sequential_id
+    before_save :ensure_organization_sequential_id, if: -> { self.instance_of?(::Invoice) }
```
jdenquin, line 11: Style/RedundantSelf: Redundant `self` detected. (https://rubystyle.guide#no-self-unless-required)

app/models/organization.rb
```diff
@@ -46,6 +46,7 @@ class Organization < ApplicationRecord
   enum document_numbering: DOCUMENT_NUMBERINGS
 
   before_create :generate_api_key
+  before_update -> { self.document_number_prefix = document_number_prefix.upcase }
```
vincent-pochet, line 49: Could you explain the reason behind this? I'm not sure to understand why it should be before each update.

app/services/organizations/update_service.rb
```diff
@@ -10,6 +10,13 @@ def initialize(organization:, params:)
     end
 
     def call
+      if params.key?(:document_number_prefix) && !params[:document_number_prefix].to_s.length.between?(1, 10)
```
vincent-pochet, line 13: Is there a reason for not adding this as a validation on the organization model?

app/models/concerns/sequenced.rb
```diff
@@ -32,20 +40,50 @@ def generate_sequential_id
...
-      # NOTE: If the application was unable to aquire the lock, the block returns false
-      raise(SequenceError, 'Unable to aquire lock on the database') unless result
+      # NOTE: If the application was unable to acquire the lock, the block returns false
+      raise(SequenceError, 'Unable to acquire lock on the database') unless result
+
+      result
+    end
+
+    def generate_organization_sequential_id
+      result = Invoice.with_advisory_lock(
```
vincent-pochet, line 50: This logic seems specific to the Invoice model, and not related to the exciting sequence logic. What do you think of adding it directly in the model? This module is shared between multiple models and is generic, I think it should not contain specific model logic

app/models/concerns/sequenced.rb
```diff
@@ -32,20 +40,50 @@ def generate_sequential_id
...
+        'invoice_lock',
```
vincent-pochet, line 51: I guess we will end with the issue we have faced today. This will lock the entire Invoice table for all invoice generation. Any chance we could scope the lock key to the organization?

app/models/organization.rb
lovrocolic: We don't allow small cases and this field is editable:) I guess other option is moving it to service level...

app/services/organizations/update_service.rb
lovrocolic: yep, This field is initially set with `after_create` hook and in order to set this field correctly to default value that we currently use, we need record's ID (that's why we are using after create). If we add it on model level, this validation will raise error with each creation since validation will be called even before after_create hook is triggered...

app/models/concerns/sequenced.rb
lovrocolic: make sense 👍

app/services/organizations/update_service.rb
vincent-pochet: I see.

For update only validation, it's possible to use the `on` option on the validates instruction:

```ruby
validates ..., on: :update`
```

lovrocolic: Cool, thanks, forgot on `on `option

PR #1514, opened 2023-12-04
feat (invoice-numbering): Add Logic for generating organization sequential id

lovrocolic:
## Context

Before, it was not possible to adapt document number. With this feature it will be enabled to switch between `per_customer` and `per_organization` document numbering logic. Also, it will be possible to edit number prefix.

## Description

This PR adds main logic for generating organization sequantial id


app/models/invoice.rb
```diff
@@ -248,8 +249,49 @@ def void_invoice!
...
+  end
+
+  def ensure_organization_sequential_id
+    return if organization_sequential_id.present? && organization_sequential_id.positive?
+
+    self.organization_sequential_id = generate_organization_sequential_id
+  end
+
+  def generate_organization_sequential_id
+    organization_sequence_scope = self.organization.invoices.where(created_at: Time.now.utc.all_month)
```
jdenquin, line 271: Style/RedundantSelf: Redundant `self` detected. (https://rubystyle.guide#no-self-unless-required)

PR #1496, opened 2023-11-27
feat(document-numbering): add support for new document numbering fields

lovrocolic:
## Context

Currently Lago does not provide solution for updating document numbers. Also, `per_customer` is the only approach in creating incremental numbers. With this feature, it will be allowed to update one part of document number. Also, it will be possible to choose between `per_customer` and `per_organization` numbering 

## Description

This PR adds API and graphql support for new document numbering fields.
It also sets by default `document_number_prefix` as combination of organization name and id.

PR #1515, opened 2023-12-04
feat (invoice-numbering): Add new document numbering logic

lovrocolic:
## Context

Before, it was not possible to adapt document number. With this feature it will be enabled to switch between `per_customer` and `per_organization` document numbering logic. Also, it will be possible to edit number prefix.

## Description

This PR is feature branch that includes all individual PRs:
- https://github.com/getlago/lago-api/pull/1495
- https://github.com/getlago/lago-api/pull/1496
- https://github.com/getlago/lago-api/pull/1514
- https://github.com/getlago/lago-api/pull/1516


app/models/invoice.rb
```diff
@@ -248,8 +249,50 @@ def void_invoice!
   def ensure_number
     return if number.present?
 
-    formatted_sequential_id = format('%03d', sequential_id)
+    if organization.document_numbering.to_s == 'per_customer'
```
jdenquin, line 252: since it's an enum you can do `organization.document_numbering.per_customer?`

app/models/invoice.rb
```diff
@@ -248,8 +249,50 @@ def void_invoice!
...
+      customer_slug = "#{organization.document_number_prefix}-#{format('%03d', customer.sequential_id)}"
```
jdenquin, line 253: can you put a `NOTE` comment to explain whats the format we wait here?

app/models/invoice.rb
```diff
@@ -248,8 +249,50 @@ def void_invoice!
...
+      formatted_sequential_id = format('%03d', sequential_id)
 
-    self.number = "#{customer.slug}-#{formatted_sequential_id}"
+      self.number = "#{customer_slug}-#{formatted_sequential_id}"
+    else
+      org_formatted_sequential_id = format('%03d', organization_sequential_id)
+
+      self.number =
+        "#{organization.document_number_prefix}-#{Time.now.utc.strftime('%Y%m')}-#{org_formatted_sequential_id}"
```
jdenquin, line 261: you can use some intermediate var to avoid this strange line return

app/models/invoice.rb
```diff
@@ -248,8 +249,50 @@ def void_invoice!
...
+  def ensure_organization_sequential_id
+    return if organization_sequential_id.present? && organization_sequential_id.positive?
+
+    self.organization_sequential_id = generate_organization_sequential_id
+  end
+
+  def generate_organization_sequential_id
+    organization_sequence_scope = organization.invoices.where(created_at: Time.now.utc.all_month)
+
+    result = Invoice.with_advisory_lock(
```
jdenquin, line 274: could this code be moved into the sequenced concern?

app/models/invoice.rb
```diff
@@ -248,8 +249,50 @@ def void_invoice!
...
+      timeout_seconds: 10.seconds,
+    ) do
+      org_sequential_id = organization_sequence_scope
+        .where.not(organization_sequential_id: nil)
+        .order(organization_sequential_id: :desc)
+        .limit(1)
+        .pick(:organization_sequential_id)
+      org_sequential_id ||= 0
+
+      loop do
```
jdenquin, line 286: small comment? I do not understand what we do here, and a loop is really dangerous

app/models/organization.rb
```diff
@@ -94,6 +108,10 @@ def generate_api_key
     self.api_key = SecureRandom.uuid
   end
 
+  def generate_document_number_prefix
+    update!(document_number_prefix: "#{name.first(3).upcase}-#{id.last(4).upcase}")
```
jdenquin, line 112: small comment to precise what the format ?

spec/scenarios/invoices/invoice_numbering_spec.rb
```diff
@@ -0,0 +1,182 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'Invoice Numbering Scenario', :scenarios, type: :request, transaction: false do
```
jdenquin, line 5: nice!

db/migrate/20231123105540_add_fields_for_new_document_numbering.rb
```diff
@@ -0,0 +1,38 @@
...
+    reversible do |dir|
+      dir.up do
+        execute <<-SQL
+          UPDATE organizations
+          SET document_number_prefix = UPPER(CAST(SUBSTRING(name, 1, 3) AS VARCHAR)) || '-' || UPPER(CAST(RIGHT(id::TEXT, 4) AS VARCHAR))
+          WHERE document_number_prefix IS NULL;
+        SQL
+
+        execute <<-SQL
+					WITH ordered_organization_invoices AS (
```
jdenquin, line 20: I have a strange indentation, can you check on your side?

jdenquin (changes requested):
Great work! I think we can improve the model code, it's a bit confusing and heavy to read (and it will also be the case to maintain or debug it). Maybe we can split it (use the sequenced concern)

app/models/invoice.rb
lovrocolic: Added

app/models/invoice.rb
lovrocolic: Added

app/models/invoice.rb
lovrocolic: As described offline, if you agree, I would rather keep this code portion in invoice model since sequenced concern is also included in few other models (credit note & customer) where this logic is not needed

app/models/invoice.rb
lovrocolic: Added

app/models/organization.rb
lovrocolic: Added

db/migrate/20231123105540_add_fields_for_new_document_numbering.rb
lovrocolic: Hmm, in my IDE looks good 🤔

app/models/invoice.rb
jdenquin: Okay, maybe it will be good to improve this part later (but not so late)

jdenquin (approved):

PR #1519, opened 2023-12-06
feat(stripe): Update stripe customer billing address

ivannovosad:
## Context

When a customer is updated in Lago, we should update the information sent to Stripe.

## Description

This PR updates Stripe customers after their details are updated in Lago.

### DIFF
--- a/app/services/payment_provider_customers/update_service.rb
+++ b/app/services/payment_provider_customers/update_service.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module PaymentProviderCustomers
+  class UpdateService < BaseService
+    def initialize(customer)
+      @customer = customer
+
+      super(nil)
+    end
+
+    def call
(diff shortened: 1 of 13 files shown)


app/services/payment_provider_customers/update_service.rb
```diff
@@ -0,0 +1,25 @@
...
+module PaymentProviderCustomers
+  class UpdateService < BaseService
+    def initialize(customer)
+      @customer = customer
+
+      super(nil)
+    end
+
+    def call
+      result = "PaymentProviderCustomers::#{provider_name}Service".constantize.new(customer.provider_customer).update
```
vincent-pochet, line 12: What about using the factory design pattern here rather than meta-prog?

We already have a `PaymentProviderCustomers::Factory` that can be used as an example.

The reason behind this is that meta prog tends to make code maintenance harder on the long run.

vincent-pochet (approved):
Just one comment, but looks good to me!

PR #1520, opened 2023-12-07
fix(graduated-prorated-charge-model): Fix graduated prorated initial state

lovrocolic:
## Context

Usually, for graduated prorated charge model, in order to fetch recurring value for each billing period, we take it from the last invoice's fee that belongs to certain charge. Mentioned logic also ignores all events contained in recurring value that are ingested after generation of previous invoice.

## Description

If there are not generated invoices yet, we still want to include past events, even if they got ingested after subscription creation with the past timestamp. This PR fixes this case.

### DIFF
--- a/app/services/billable_metrics/prorated_aggregations/sum_service.rb
+++ b/app/services/billable_metrics/prorated_aggregations/sum_service.rb
@@ -100,16 +100,13 @@ module BillableMetrics
 
       def recurring_value
         previous_charge_fee_units = previous_charge_fee&.units
-
         return previous_charge_fee_units if previous_charge_fee_units
 
         query = persisted_query
-          .select("SUM((CAST(#{sanitized_field_name} AS FLOAT)) * (#{persisted_pro_rata}))::numeric").to_sql
-        recurring_value_before_first_fee = ActiveRecord::Base.connection.execute(query).first['sum']
+          .select("SUM(((#{sanitized_field_name})::numeric) * (#{persisted_pro_rata}))::numeric").to_sql
+        recurring_value_before_first_fee = ActiveRecord::Base.connection.select_one(query)['sum']
 
-        return nil unless recurring_value_before_first_fee
-
-        (recurring_value_before_first_fee <= 0) ? nil : recurring_value_before_first_fee
+        ((recurring_value_before_first_fee || 0) <= 0) ? nil : recurring_value_before_first_fee
       end
     end
   end
--- a/app/services/billable_metrics/prorated_aggregations/sum_service.rb
+++ b/app/services/billable_metrics/prorated_aggregations/sum_service.rb
@@ -54,8 +54,6 @@ module BillableMetrics
 
       protected
 
-      attr_reader :options
-
       def compute_aggregation
         ActiveRecord::Base.connection.execute(aggregation_query).first['aggregation_result']
       end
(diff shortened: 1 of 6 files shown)


spec/scenarios/charge_models/prorated_graduated_spec.rb
```diff
@@ -194,6 +194,112 @@
...
+              ],
+            },
+          )
+
+          travel_to(DateTime.new(2023, 12, 2)) do
+            create_event(
+              {
+                code: billable_metric.code,
+                transaction_id: SecureRandom.uuid,
+                timestamp: 1699336493, ## November 2023
```
jdenquin, line 242: Style/NumericLiterals: Use underscores(_) as thousands separator and separate every 3 digits with them. (https://rubystyle.guide#underscores-in-numerics)

lovrocolic:
> Looks good to me. I just have a general question regarding this logic of initial value. In the percentage charge model, we use this per_event_aggregation, to compute the the per transaction logic (min/max), and I am wondering if this "initial value" logic is not messing a bit with the computation as it will consider this initial value as a transaction.

`prorated` option is allowed only for standard, volume and graduated charge models. In this case per_event_aggregation is updated in `BillableMetrics::ProratedAggregations::BaseService` and it is used only for graduated charge model

app/services/billable_metrics/prorated_aggregations/sum_service.rb
```diff
@@ -87,6 +97,20 @@ def duration_ratio_sql(from, to)
...
+
+        return previous_charge_fee_units if previous_charge_fee_units
+
+        query = persisted_query
+          .select("SUM((CAST(#{sanitized_field_name} AS FLOAT)) * (#{persisted_pro_rata}))::numeric").to_sql
+        recurring_value_before_first_fee = ActiveRecord::Base.connection.execute(query).first['sum']
+
+        return nil unless recurring_value_before_first_fee
+
+        recurring_value_before_first_fee <= 0 ? nil : recurring_value_before_first_fee
```
jdenquin, line 112: Style/TernaryParentheses: Use parentheses for ternary expressions with complex conditions.

app/services/billable_metrics/prorated_aggregations/unique_count_service.rb
```diff
@@ -196,6 +196,15 @@ def added_and_removed_list
...
         end
       end
+
+      def recurring_value
+        previous_charge_fee_units = previous_charge_fee&.units
+
+        return previous_charge_fee_units if previous_charge_fee_units
+
+        recurring_value_before_first_fee = prorated_persisted_query.sum("(#{persisted_pro_rata})::numeric")
+        recurring_value_before_first_fee <= 0 ? nil : recurring_value_before_first_fee
```
jdenquin, line 206: Style/TernaryParentheses: Use parentheses for ternary expressions with complex conditions.

spec/services/billable_metrics/prorated_aggregations/unique_count_service_spec.rb
```diff
@@ -535,5 +537,41 @@
         expect(result.event_prorated_aggregation.map { |el| el.ceil(5) }).to eq([first, first, second])
       end
     end
+
+    context 'with multiple events added in the period and with one added and removed during period' do
```
jdenquin, line 541: RSpec/RepeatedExampleGroupDescription: Repeated context block description on line(s) [504] (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/RepeatedExampleGroupDescription)

PR #1532, opened 2023-12-12
feat(taxes): Customer EU VAT autodetection

jdenquin:
## Context

- Based on the customer country or tax identification number, we want to assign the right tax code to the customer.

## Description

- Add `Customers::EuAutoTaxesService`  that detect and return the right tax code for the customer.
- Trigger this service call in customer create and update


app/services/customers/update_service.rb
```diff
@@ -74,6 +74,13 @@ def update(**args)
       ActiveRecord::Base.transaction do
         customer.save!
 
+        if customer.organization.eu_tax_management
+          eu_tax_code = Customers::EuAutoTaxesService.call(customer:)
+
+          args[:tax_codes] ||= []
```
rsempe, line 80: I think it will remove taxes for an existing customer if `tax_codes` is not passed but `eu_tax_management` is activated 😬

jdenquin: yes, it's the actual behavior we want, we may change this later, the doc will focus on this cc @sarkissianraffi

vincent-pochet (approved):

PR #1536, opened 2023-12-12
fix(event): Pay in advance with missing or invalid group

vincent-pochet:
## Context

An issue has recently been identified with the pay in advance aggregation and group:

```
Steps to reproduce:

1. Create a billable metric with groups;
2. Create a plan with a charge paid in advance linked to the billable metric;
3. Create a subscription;
4. Send events that don’t include the expected group properties → Lago generates a fee even if the group value is invalid or the expected group key is missing.
```

It is a side effect of the default properties at charge level recently introduced.

## Description

This PR changes the check for group on billable metrics by checking the selectable groups rather than the group properties at charge level.

### DIFF
--- a/app/services/fees/create_pay_in_advance_service.rb
+++ b/app/services/fees/create_pay_in_advance_service.rb
@@ -15,7 +15,7 @@ module Fees
       fees = []
 
       ActiveRecord::Base.transaction do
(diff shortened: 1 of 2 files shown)


rsempe (approved):
good spot 👏

PR #1540, opened 2023-12-13
fix: date intervals in invoice template and fee payload

lovrocolic:
## Context

In invoice template pay in advance charge fees haven't been grouped correctly and they were also incorrectly calculated in fee serializer

## Description

This PR fix both cases

### DIFF
--- a/app/serializers/v1/fee_serializer.rb
+++ b/app/serializers/v1/fee_serializer.rb
@@ -52,6 +52,18 @@ module V1
     private
 
     def date_boundaries
+      if model.charge? && !model.pay_in_advance? && model.charge.pay_in_advance?
+        subscription = model.subscription
+        invoice = model.invoice
+        timestamp = invoice.invoice_subscription(subscription.id).timestamp
+        interval = invoice.charge_pay_in_advance_interval(timestamp, subscription)
+
+        return {
+          from_date: interval[:charges_from_date]&.to_datetime&.iso8601,
+          to_date: interval[:charges_to_date]&.to_datetime&.end_of_day&.iso8601,
+        }
+      end
+
       {
         from_date:,
         to_date:,
--- a/app/views/templates/invoices/v4/_subscription_details.slim
+++ b/app/views/templates/invoices/v4/_subscription_details.slim
@@ -19,10 +19,34 @@
           td.body-2 == TaxHelper.applied_taxes(invoice_subscription(subscription.id).subscription_fee)
           td.body-2 = MoneyHelper.format(invoice_subscription(subscription.id).subscription_amount)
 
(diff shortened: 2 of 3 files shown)


app/serializers/v1/fee_serializer.rb
```diff
@@ -52,6 +52,18 @@ def serialize
     private
 
     def date_boundaries
+      if model.charge? && !model.pay_in_advance? && model.charge.pay_in_advance?
```
vincent-pochet, line 55: I might be misunderstanding this one, but since we are in the `FeeSerializer`, I think that model is a fee, and I'm not sure it as a `pay_in_advance` attribute. Should it be `model.subscription.pay_in_advance?`

app/serializers/v1/fee_serializer.rb
```diff
@@ -52,6 +52,18 @@ def serialize
...
+        subscription = model.subscription
+        invoice = model.invoice
+        timestamp = invoice.invoice_subscription(subscription.id).timestamp
+        interval = invoice.charge_pay_in_advance_interval(timestamp, subscription)
+
+        return {
+          from_date: interval[:charges_from_date]&.to_datetime&.iso8601,
```
vincent-pochet, line 62: On a global perspective, I'm wondering if we should store the correct boundaries in the fee (or in new attributes `charge_in_advance_from|to_datetime` of the invoice subscription)

app/serializers/v1/fee_serializer.rb
lovrocolic: Seems strange but actually it is good :) Fee can also be pay in advance for the event pay in advance invoice (bot the one at the end of the billing period). These boundaries should be only recalculated for charges that are pay in advance...

app/serializers/v1/fee_serializer.rb
lovrocolic: Yep that is a good question, but since pay_in_advance charges can be recurring, existing boundaries are correct in terms of aggregation (since for recurring that are pay in advance we still need to generate invoice at the end of the period and use existing boundaries). But in terms of UI / api payload presentation, those are not correct. Do you think we should have both persisted?

vincent-pochet: Yes, that was my point. Having to recompute it each time, makes me think we should store it, mainly for performance and maintainability (pretty sure we will forget to use the right boundaries in some other places)

vincent-pochet (approved):

app/serializers/v1/fee_serializer.rb
lovrocolic: Writing this task in my to-do list

PR #1543, opened 2023-12-14
fix(invoice): Add missing unique index on invoice organization_sequential ids

vincent-pochet:
## Context

Invoice table is missing the index to ensure that `organization_sequential_id` is unique in the scope of the `organization_id`

## Description

This PR is adding the index

### DIFF
--- a/db/migrate/20231214133638_add_unique_index_on_invoice_organization_sequential_id.rb
+++ b/db/migrate/20231214133638_add_unique_index_on_invoice_organization_sequential_id.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddUniqueIndexOnInvoiceOrganizationSequentialId < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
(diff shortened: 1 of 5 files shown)


db/migrate/20231214133638_add_unique_index_on_invoice_organization_sequential_id.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddUniqueIndexOnInvoiceOrganizationSequentialId < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :invoices,
+              "organization_id, organization_sequential_id, date_trunc('month'::text, created_at)",
+              name: 'unique_organization_sequential_id',
+              unique: true
```
lovrocolic, line 10: I think that we should add: `where: 'organization_sequential_id IS NOT NULL'` if we will skip this generation for each invoice

PR #1545, opened 2023-12-14
Add ability to set Google Cloud Storage credentials file path by environment variable

edwardmp:
## Context

Lago by default looks at a file on disk to load Google Cloud credentials, which is not a convenient option for us right now since our platform only allows putting secret files at a certain path, and Lago hardcodes it as located in the Rails root.

This PR adds a env var to allow overriding this configuration.

## Description

Uses the value of the env var `GCS_KEYFILE_JSON_PATH` or falls back if the env var isn't set, falls back to pre-existing behavior.

### DIFF
--- a/config/storage.yml
+++ b/config/storage.yml
@@ -25,6 +25,6 @@ amazon_compatible_endpoint:
 
 google:
   service: GCS
-  credentials: <%= ENV.fetch('GCS_KEYFILE_JSON_PATH', Rails.root.join('gcs_keyfile.json')) %>
+  credentials: <%= ENV.fetch('LAGO_GCS_KEYFILE_JSON_PATH', Rails.root.join('gcs_keyfile.json')) %>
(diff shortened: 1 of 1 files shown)


edwardmp:
Not sure why Pronto always fails for my PRs 🤷

jdenquin:
I think it does not like you 😂 , no it's a token issue!

vincent-pochet (commented):
Looks good! Thank you for this PR.

To keep the consistence, with other ENV variable, what do you think of prefixing this one with `LAGO_`?

edwardmp:
@vincent-pochet makes total sense! Added the prefix

vincent-pochet (approved):

vincent-pochet:
Thank you again for this PR @edwardmp 

It's now merged in main and will be part of the next release, probably next week.

PR #1553, opened 2023-12-18
feat(one-off-invoices): use new invoice generation service for creating one-off invoices

lovrocolic:
## Context

We recently added new service that handles invoice creation.

## Description

Adapt creation of one-off invoices with new invoice generation service.

### DIFF
--- a/app/services/invoices/create_service.rb
+++ b/app/services/invoices/create_service.rb
@@ -16,29 +16,16 @@ module Invoices
       return result.not_found_failure!(resource: 'fees') if fees.blank?
       return result.not_found_failure!(resource: 'add_on') unless add_ons.count == add_on_identifiers.count
 
-      invoice = nil
-
       ActiveRecord::Base.transaction do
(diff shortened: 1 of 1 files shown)


vincent-pochet (approved):
Looks good to me.
I was wondering if we should apply the same logic as in other services. It's not necessary dictated by a performance need as the service is quite simple, but more to have an unified way to handle invoices in all services dealing with it (maybe something we could extract into a concern for example). Curious of you thought about it.

lovrocolic:
> Looks good to me. I was wondering if we should apply the same logic as in other services. It's not necessary dictated by a performance need as the service is quite simple, but more to have an unified way to handle invoices in all services dealing with it (maybe something we could extract into a concern for example). Curious of you thought about it.

If I understood you correctly, you would prefer to still process this service in a job, which would apparently allow us to use the same structure in this service?
Make sense, but I will have to check if UI expects returned invoice...

PR #1557, opened 2023-12-19
feat(unit-price): Add rate to percentage amount details

rsempe:
The goal of this PR is to remove `per_unit_amount` for adding `rate` for percentage and gradutaed_percentage charge models.


db/migrate/20231218170631_add_rate_to_percentage_amount_details.rb
```diff
@@ -0,0 +1,21 @@
...
+    belongs_to :charge, -> { with_discarded }, optional: true
+  end
+
+  def up
+    percentage_fees = Fee.joins(:charge).merge(Charge.percentage).where.not(amount_details: nil)
+
+    percentage_fees.find_each do |fee|
+      fee.update!(
+        amount_details: fee.amount_details.except!('per_unit_amount').merge(
+          rate: BigDecimal(fee.charge.properties['rate'].to_s)
```
jdenquin, line 14: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

db/migrate/20231218170631_add_rate_to_percentage_amount_details.rb
```diff
@@ -0,0 +1,21 @@
...
+        )
```
jdenquin, line 15: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

db/migrate/20231218170631_add_rate_to_percentage_amount_details.rb
```diff
@@ -0,0 +1,36 @@
...
+        amount_details: fee.amount_details.except!('per_unit_amount').merge(
+          rate: BigDecimal(fee.charge.properties['rate'].to_s),
+        ),
+      )
+    end
+
+    graduated_percentage_fees = Fee.joins(:charge).merge(Charge.graduated_percentage).where.not(amount_details: nil)
+
+    graduated_percentage_fees.find_each do |fee|
+      fee.amount_details['graduated_percentage_ranges'] = fee.amount_details['graduated_percentage_ranges'].tap do |ranges|
```
jdenquin, line 22: Layout/LineLength: Line is too long. [123/120] (https://rubystyle.guide#max-line-length)

vincent-pochet (approved):
:rocket:

app/services/charges/amount_details/range_graduated_percentage_service.rb
```diff
@@ -38,11 +38,7 @@ def flat_unit_amount
       end
 
       def rate
-        @rate ||= BigDecimal(range[:rate])
-      end
-
-      def per_unit_amount
-        @per_unit_amount ||= units.zero? ? BigDecimal(0) : per_unit_total_amount.fdiv(units)
+        @rate ||= BigDecimal(range[:rate].to_s)
```
ivannovosad, line 41: Can `range[:rate]` be `nil`?

if yes, then we can do something like: 

```suggestion
        @rate ||= BigDecimal(range[:rate].to_s.presence || 0)
```

spec/services/charges/amount_details/range_graduated_percentage_service_spec.rb
```diff
@@ -21,7 +21,7 @@
         from_value: 0,
         to_value: 10,
         flat_unit_amount: 2,
-        per_unit_amount: '0.02',
+        rate: 2,
```
ivannovosad, line 24: Lets add a context with `rate: nil` in case it can happen.

ivannovosad (changes requested):

app/services/charges/amount_details/range_graduated_percentage_service.rb
rsempe: At this point, `rate` should not be `nil`.

In that case, I prefer raising an error instead of putting `0` as the value

ivannovosad (approved):

PR #1558, opened 2023-12-20
fix: update index for organization_sequential_id

lovrocolic:
## Context

Existing index was taking into account only month, but we should also take year into account so that index can be really unique.

### DIFF
--- a/db/migrate/20231220115621_update_organization_sequential_id_index.rb
+++ b/db/migrate/20231220115621_update_organization_sequential_id_index.rb
@@ -1,0 +1,10 @@
+class UpdateOrganizationSequentialIdIndex < ActiveRecord::Migration[7.0]
+  def change
+    remove_index :invoices, name: :unique_organization_sequential_id
+    add_index :invoices,
+              "organization_id, organization_sequential_id, (date_trunc('month', created_at)::date)",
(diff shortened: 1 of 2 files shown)


db/migrate/20231220115621_update_organization_sequential_id_index.rb
```diff
@@ -0,0 +1,10 @@
+class UpdateOrganizationSequentialIdIndex < ActiveRecord::Migration[7.0]
```
jdenquin, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

db/migrate/20231220115621_update_organization_sequential_id_index.rb
```diff
@@ -0,0 +1,10 @@
...
+  def change
+    remove_index :invoices, name: :unique_organization_sequential_id
```
jdenquin, line 3: Rails/ReversibleMigration: remove_index(without column) is not reversible. (https://rails.rubystyle.guide#reversible-migration, https://api.rubyonrails.org/classes/ActiveRecord/Migration/CommandRecorder.html)

rsempe (approved):

vincent-pochet (approved):

PR #1559, opened 2023-12-20
feat(invoices): Do not send invoice when amount is zero

jdenquin:


app/mailers/invoice_mailer.rb
```diff
@@ -10,6 +10,7 @@ def finalized
 
     return if @organization.email.blank?
     return if @customer.email.blank?
+    return if @invoice.fees_amount_cents.zero?
```
ivannovosad, line 13: Can we add a spec for this case?

jdenquin: Yes sorry it's draft I forgot to put it in draft

jdenquin: done @ivannovosad !

vincent-pochet (approved):

ivannovosad (approved):

PR #1560, opened 2023-12-20
feat (invoices): add db changes for edit draft invoice feature

lovrocolic:
## Context

Currently it is not possible to edit units and amounts on draft invoice fees.

## Description

This feature adds DB changes need for edit draft invoice feature

### DIFF
--- a/app/models/adjusted_fee.rb
+++ b/app/models/adjusted_fee.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AdjustedFee < ApplicationRecord
+  belongs_to :invoice
+  belongs_to :fee, optional: true
+
+  enum fee_type: Fee::FEE_TYPES
+end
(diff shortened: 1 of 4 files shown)


app/models/adjusted_fee.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AdjustedFee < ApplicationRecord
+  belongs_to :invoice
+  belongs_to :fee, optional: true
```
ivannovosad, line 5: Why is the fee optional?

vincent-pochet (approved):

app/models/adjusted_fee.rb
lovrocolic: The reason behind was that I wanted to ensure that this object can exist without fee (in scenario when refresh job is called and fee got destroyed)

PR #1566, opened 2023-12-22
Set network time to Redis

TronTram:
## Why
- Fix error: 
image


config/environments/production.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/integer/time'
+require 'sprockets/railtie'
```
jdenquin, line 4: we do not need `sprockets` for production

config/credentials/production.yml.enc
jdenquin, line 1: You can remove this file

Dockerfile
jdenquin, line 1: We do not want to include the development libs for the image we publish, you can use the `Dockerfile.dev` for this!

.gitignore
```diff
@@ -36,3 +36,5 @@ db/events_schema.rb
 LAGO_VERSION
 
 **/.DS_Store
+
+/config/credentials/production.key
```
jdenquin, line 40: You can remote it, we do not use those credentials

jdenquin (commented):
Thanks for contributing, I put some comments here!

TronTram:
> Thanks for contributing, I put some comments here!

thanks you so much.
 it's my mistake, it should merge to my fork repo.

PR #1572, opened 2023-12-26
feat(taxes): Add EU VAT terms

jdenquin:
- Add EU VAT Tax term on invoices templates


lovrocolic (approved):
Do we need to include this partial for credit notes as well?

jdenquin:
> Do we need to include this partial for credit notes as well?

@michaelponrajah

PR #1588, opened 2024-01-05
misc: Add ready_to_be_refreshed to invoices

rsempe:
The goal of this refactor is to refresh draft invoices asynchronously.

We now want to use a clock and not a “true real-time” when an invoice needs to be automatically refreshed.

Because with the current solution, we can have duplicate refreshes for the same invoice at the same time.


app/jobs/clock/refresh_draft_invoices_job.rb
```diff
@@ -0,0 +1,13 @@
...
+  class RefreshDraftInvoicesJob < ApplicationJob
+    queue_as 'clock'
+
+    def perform
+      Invoice.ready_to_be_refreshed.find_each do |invoice|
+        Invoices::RefreshDraftJob.perform_later(invoice:)
+      end
+    end
+  end
+end
```
jdenquin, line 13: Layout/TrailingEmptyLines: Final newline missing. (https://rubystyle.guide#newline-eof)

spec/jobs/clock/refresh_draft_invoices_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Clock::RefreshDraftInvoicesJob, job: true do
```
jdenquin, line 5: RSpec/FilePath: Spec path should end with `clock/refresh_draft_invoices_job*_spec.rb`. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FilePath)

vincent-pochet (approved):

PR #1592, opened 2024-01-09
misc: refactor money with precision helper method

lovrocolic:
## Context

Currently Money gem does not allow setting `default_infinite_precision` property per class instance. 

## Description

This PR cleans existing logic with the new class that extends Money. New class should be used only when `default_infinite_precision` is needed to be true. This approach is more safe and does not require setting and resetting property on each usage.

### DIFF
--- a/app/services/utils/money_with_precision.rb
+++ b/app/services/utils/money_with_precision.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module Utils
+  class MoneyWithPrecision < Money
+    self.default_infinite_precision = true
+  end
+end
--- a/app/views/helpers/money_helper.rb
+++ b/app/views/helpers/money_helper.rb
@@ -10,16 +10,10 @@ class MoneyHelper
   end
 
   def self.format_with_precision(amount_cents, currency)
-    Money.default_infinite_precision = true
-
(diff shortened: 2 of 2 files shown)


ivannovosad (approved):

app/services/utils/money_with_precision.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Utils
+  class MoneyWithPrecision < Money
+    self.default_infinite_precision = true
```
jdenquin, line 5: this will add the `infinite_precision` for the whole instance, on some value it will not match what we expect because we do not want the precise value but the rounded one

jdenquin (changes requested):

app/services/utils/money_with_precision.rb
rsempe: In that case, `MoneyWithPrecision` class, the whole instance needs to be with precision ;)

rsempe (approved):

app/services/utils/money_with_precision.rb
lovrocolic: yep, only MoneyWithPrecision will have this property set. Money class will be untouched.

jdenquin (approved):

PR #1598, opened 2024-01-12
feat(events): Batch Event endpoint

jdenquin:
## Context

- We want to support multiple events ingestion with one API call.

## Description

- Update the `/events/batch` endpoint that was deprecated
- Add the support of 100 event at a time
- Add a simple validation
- It follows the same process as the single event ingestion

### DIFF
--- a/app/controllers/api/v1/events_controller.rb
+++ b/app/controllers/api/v1/events_controller.rb
@@ -94,14 +94,15 @@ module Api
 
       def batch_params
         params
-          .require(:event)
           .permit(
-            :transaction_id,
-            :external_customer_id,
-            :code,
-            :timestamp,
-            external_subscription_ids: [],
-            properties: {},
+            events: [
+              :transaction_id,
+              :external_customer_id,
+              :code,
+              :timestamp,
+              :external_subscription_id,
+              properties: {},
+            ],
           )
       end
 
--- a/app/controllers/api/v1/events_controller.rb
+++ b/app/controllers/api/v1/events_controller.rb
@@ -23,22 +23,25 @@ module Api
         end
       end
 
-      # DEPRECATED
       def batch
-        validate_result = Events::CreateBatchService.new.validate_params(
+        result = ::Events::CreateBatchService.call(
           organization: current_organization,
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/events_controller.rb
```diff
@@ -94,15 +97,16 @@ def create_params
...
-            external_subscription_ids: [],
-            properties: {},
-          )
+            events: [
+              :transaction_id,
+              :external_customer_id,
+              :code,
+              :timestamp,
+              :external_subscription_id,
+              properties: {},
```
jdenquin, line 107: Style/HashAsLastArrayItem: Wrap hash in `{` and `}`. (https://rubystyle.guide#hash-literal-as-last-array-item)

app/services/events/validate_creation_service.rb
```diff
@@ -1,56 +1,26 @@
 # frozen_string_literal: true
 
-# frozen_string_literal: true
-
 module Events
   class ValidateCreationService
```
rsempe, line 4: Why do we still need this class? 🤔

app/services/events/create_batch_service.rb
```diff
@@ -1,98 +1,82 @@
...
-      return unless persisted_service.matching_billable_metric?
+    def post_validate_events
+      ActiveRecord::Base.transaction do
+        result.events.each do |event|
+          event.save!
 
-      service_result = persisted_service.call
-      service_result.raise_if_error!
+          produce_kafka_event(event)
+          Events::PostProcessJob.perform_later(event:)
```
vincent-pochet, line 60: Jobs enqueued in a DB transaction :boom:

app/services/events/validate_creation_service.rb
vincent-pochet: I think it is still used for the "fees#estimate"

jdenquin: yes it's still used by the `fees#estimate` endpoint

vincent-pochet (approved):

PR #1606, opened 2024-01-16
feat(filter): Create or update service for billable metric filters

vincent-pochet:
## Context

## Description

This PR adds a new `BillableMetricFilters::CreateOrUpdateBatchService` to manage billable metric filters

### DIFF
--- a/app/services/billable_metrics/update_service.rb
+++ b/app/services/billable_metrics/update_service.rb
@@ -15,9 +15,18 @@ module BillableMetrics
       billable_metric.name = params[:name] if params.key?(:name)
       billable_metric.description = params[:description] if params.key?(:description)
 
-      if params.key?(:group)
-        group_result = update_groups(billable_metric, params[:group])
-        return group_result if group_result.error
+      ActiveRecord::Base.transaction do
+        if params.key?(:group)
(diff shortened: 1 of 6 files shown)


app/services/billable_metrics/update_service.rb
```diff
@@ -44,12 +53,10 @@ def call
...
-      ActiveRecord::Base.transaction do
-        Groups::CreateOrUpdateBatchService.call(
-          billable_metric: metric,
-          group_params: group_params.with_indifferent_access,
-        )
-      end
+      Groups::CreateOrUpdateBatchService.call(
+        billable_metric: metric,
+        group_params: group_params.with_indifferent_access,
+      )
```
rsempe, line 59:
```suggestion
      ).raise_if_error!
```

rsempe (approved):
👍

app/services/billable_metrics/update_service.rb
vincent-pochet: No, the error result is handled differently here. I agree we should use `raise_if_error!` but since the group related logic will be deprecated, I don't want to invest time in refactoring the way it's working now

rsempe: Allright!

PR #1613, opened 2024-01-17
feat(minimum-spending): Add serializers

ivannovosad:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/set-minimum-spend-for-subscriptions

## Context

As a user, I want to define the minimum spend that applies to the customer's plan. Consider the following monthly plan:

- Subscription fee = $50 per month (in arrears);
- Usage-based charge = $10 per unit per month (in arrears); and
- Monthly minimum spend = $100.

If at the end of the month, the customer has consumed 3 units, their total invoice will be: $50 + 3 * $10 = $80 + $20 (true-up fee) = $100.

## Description

In this PR we're adding serializers.

### DIFF
--- a/spec/serializers/v1/commitment_serializer_spec.rb
+++ b/spec/serializers/v1/commitment_serializer_spec.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ::V1::CommitmentSerializer do
+  subject(:serializer) do
+    described_class.new(commitment, root_name: 'commitment', includes: %i[taxes])
+  end
+
+  let(:commitment) { create(:commitment) }
+  let(:tax) { create(:tax, organization: commitment.plan.organization) }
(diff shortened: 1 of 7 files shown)


spec/serializers/v1/commitment_serializer_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  let(:tax) { create(:tax, organization: commitment.plan.organization) }
+  let(:commitment_applied_tax) { create(:commitment_applied_tax, commitment:, tax:) }
+
+  before { commitment_applied_tax }
+
+  it 'serializes the object' do
+    result = JSON.parse(serializer.to_json)
+
+    aggregate_failures do
+      expect(result['commitment']['lago_id']).to eq(commitment.id)
```
rsempe, line 20: can we refactor this to have only one expectation?

something like `expect(result['commitment']).to include(...)`

rsempe (approved):

PR #1621, opened 2024-01-20
Update Dockerfile to fix health check reliant on curl binary

zeeshanmuhammad:
Healthcheck in https://github.com/getlago/lago/blob/main/docker-compose.yml for lago-api relies on curl binary existing in lago-api docker image (it wasn't included). Docker currently floods log files with a failure message:

`Health check for container x error: OCI runtime exec failed: exec failed: unable to start container process: exec: \"curl\": executable file not found in $PATH: unknown"`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /app
 COPY ./Gemfile /app/Gemfile
 COPY ./Gemfile.lock /app/Gemfile.lock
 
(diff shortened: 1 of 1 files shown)


vincent-pochet:
Thank you @zeeshanmuhammad  for this contribution

It's indeed a regression. I'm merging the PR so that it can be released in the next version.

PR #1622, opened 2024-01-22
feat(minimum-spending): Add controllers for plan commitments

ivannovosad:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/set-minimum-spend-for-subscriptions

## Context

As a user, I want to define the minimum spend that applies to the customer's plan. Consider the following monthly plan:

- Subscription fee = $50 per month (in arrears);
- Usage-based charge = $10 per unit per month (in arrears); and
- Monthly minimum spend = $100.

If at the end of the month, the customer has consumed 3 units, their total invoice will be: $50 + 3 * $10 = $80 + $20 (true-up fee) = $100.

## Description

In this PR we're updating plan create/update services for creating and updating plan's minimum commitment. Also the controllers are updated to support it.

### DIFF
--- a/app/services/plans/update_service.rb
+++ b/app/services/plans/update_service.rb
@@ -43,9 +43,10 @@ module Plans
         end
 
         process_charges(plan, params[:charges]) if params[:charges]
+        process_minimum_commitment(plan, params[:minimum_commitment]) if params[:minimum_commitment] && License.premium?
       end
 
-      result.plan = plan
+      result.plan = plan.reload
       result
     rescue ActiveRecord::RecordInvalid => e
       result.record_validation_failure!(record: e.record)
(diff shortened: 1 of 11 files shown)


app/services/plans/update_service.rb
```diff
@@ -97,6 +98,24 @@ def charge_model(params)
       model
     end
 
+    def process_minimum_commitment(plan, params)
+      minimum_commitment = plan.minimum_commitment || Commitment.new(plan:, commitment_type: 'minimum_commitment')
```
lovrocolic, line 102: As I can see this constant is used multiple times. Can we maybe use this one based on enum value, or create a constant for it?

app/services/plans/update_service.rb
```diff
@@ -97,6 +98,24 @@ def charge_model(params)
...
+
+      minimum_commitment.amount_cents = params[:amount_cents]
+      minimum_commitment.invoice_display_name = params[:invoice_display_name]
+      minimum_commitment.save!
```
lovrocolic, line 106: I think the logic here is not valid. If Commitment already exists and `invoice_display_name`  is already present, but if it is not sent in input parameters, it will be set to nil here. In update service (update mode) we should not set param that is not passed in params

spec/services/plans/update_service_spec.rb
```diff
@@ -213,6 +224,102 @@
...
+        context 'when license is not premium' do
+          it 'does not update minimum commitment' do
+            result = plans_service.call
+
+            expect(result.plan.minimum_commitment.amount_cents).not_to eq(update_args[:amount_cents])
+          end
+        end
+      end
+    end
+
```
lovrocolic, line 322: Can you please add a test case where not all params are present and check that the service has not changed this attribute?

PR #1624, opened 2024-01-23
feat(filter): Add grouped_by logic on events aggregation

vincent-pochet:
## Context

## Description

This PR updates the event stores to add support for grouped aggregation on:
- `count`
- `max`
- `sum`
- `last`

### DIFF
--- a/app/services/events/stores/clickhouse_store.rb
+++ b/app/services/events/stores/clickhouse_store.rb
@@ -92,6 +92,14 @@ module Events
         events.maximum(Arel.sql(sanitized_numeric_property))
       end
 
+      def grouped_max
+        events
+          .reorder(nil)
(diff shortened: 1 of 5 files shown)


lovrocolic (approved):

app/services/events/stores/clickhouse_store.rb
```diff
@@ -61,17 +61,75 @@ def count
...
+          with events as (#{cte_sql})
+
+          select
+            DISTINCT ON (#{group_names}) #{group_names},
+            property
+          from events
+          ORDER BY #{group_names}, events.timestamp DESC
+        SQL
+
+        byebug
```
jdenquin, line 128: Lint/Debugger: Remove debugger entry point `byebug`.

PR #1625, opened 2024-01-23
feat(group-by): add filters argument to aggregator classes

vincent-pochet:
## Context

## Description

This PR adds a `filters` argument on the aggregation services to replace the `group` and `event` ones and will allow new filters for the next steps of the feature


vincent-pochet:
Closed by https://github.com/getlago/lago-api/pull/1620

PR #1628, opened 2024-01-24
feat(payment-url): generate payment url for stripe and adyen

lovrocolic:
## Context

Some customers don’t have payment methods that accept recurring payments (ie: card payments in India) and currently it is not possible to fetch one-time payment link when invoice payment status is failed.

## Description

This PR adds a logic for stripe and adyen in order to generate one time payment link


jdenquin (approved):
LGTM

ivannovosad (approved):

PR #1633, opened 2024-01-25
fix(payment-methods): Fix updating payment provider payment methods

ivannovosad:
## Context

It is impossible to remove sepa_debit payment method from a customer billing configuration via API.

## Description

This PR fixes Customer API to be able to update/remove payment methods from customer billing configuration.

### DIFF
--- a/app/services/customers/create_service.rb
+++ b/app/services/customers/create_service.rb
@@ -237,10 +237,10 @@ module Customers
 
       return unless update_provider_customer
 
+      create_or_update_provider_customer(customer, billing)
+
       if customer.provider_customer&.provider_customer_id
         PaymentProviderCustomers::UpdateService.call(customer)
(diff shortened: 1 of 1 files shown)


app/services/customers/create_service.rb
```diff
@@ -237,10 +237,10 @@ def handle_api_billing_configuration(customer, params, new_customer)
 
       return unless update_provider_customer
 
+      create_or_update_provider_customer(customer, billing)
+
       if customer.provider_customer&.provider_customer_id
```
jdenquin, line 242: is this if still needed?

ivannovosad: It is and yes, we should refactor that part in a followup pull request.

jdenquin (approved):

PR #1634, opened 2024-01-25
feat(grouped-by) Adapt sum aggregation to return grouped aggregation

vincent-pochet:
## Context

## Description

This PR applies the grouped_by logic to in arrears charges with sum aggregation. It creates a fee per aggregated group.

### DIFF
--- a/spec/services/fees/charge_service_spec.rb
+++ b/spec/services/fees/charge_service_spec.rb
@@ -43,7 +43,6 @@ RSpec.describe Fees::ChargeService do
       billable_metric:,
       properties: {
         amount: '20',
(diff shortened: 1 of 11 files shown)


spec/services/fees/charge_service_spec.rb
```diff
@@ -67,6 +66,71 @@
         )
       end
```
lovrocolic, line 68: Can we maybe add one case here with combination of `grouped_by` and `adjusted_fee`?

lovrocolic (approved):

rsempe (approved):

PR #1644, opened 2024-01-29
feat(payment-url): handle incoming webhooks

lovrocolic:
## Context

Currently we are not able to generate one-time payment url

## Description

This PR handles incoming webhooks where invoice status is changed according to the checkout status

### DIFF
--- a/app/services/payment_providers/adyen_service.rb
+++ b/app/services/payment_providers/adyen_service.rb
@@ -129,7 +129,7 @@ module PaymentProviders
 
     def update_payment_status(event, payment_type)
       provider_payment_id = event['pspReference']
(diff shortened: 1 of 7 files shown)


app/services/payment_providers/adyen_service.rb
```diff
@@ -116,5 +124,15 @@ def reattach_provider_customers(organization_id:, adyen_provider:)
           c.update(payment_provider_id: adyen_provider.id)
         end
     end
+
+    private
+
+    def update_payment_status(event, payment_type)
+      provider_payment_id = event['pspReference']
+      status = event['success'] == 'true' ? 'succeeded' : 'failed'
```
jdenquin, line 132: Style/TernaryParentheses: Use parentheses for ternary expressions with complex conditions.

PR #1646, opened 2024-01-29
feat(clickhouse): Clickhouse migration paths

jdenquin:
## Description

- Add Clickhouse migration paths
- Fix timestamp for events

### DIFF
--- a/app/services/events/create_service.rb
+++ b/app/services/events/create_service.rb
@@ -49,7 +49,7 @@ module Events
           external_customer_id: event.external_customer_id,
           external_subscription_id: event.external_subscription_id,
           transaction_id: event.transaction_id,
-          timestamp: event.timestamp.to_i,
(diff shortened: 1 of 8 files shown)


app/services/events/create_service.rb
```diff
@@ -49,7 +49,7 @@ def produce_kafka_event(event)
           external_customer_id: event.external_customer_id,
           external_subscription_id: event.external_subscription_id,
           transaction_id: event.transaction_id,
-          timestamp: event.timestamp,
+          timestamp: event.timestamp.to_i,
```
vincent-pochet, line 52: It might be an issue for the weighted sum aggregation as we need to handle events with millisecond granularity

I think you could use `to_f`, but we have to check if the kafka topic handle it correctly

jdenquin: updated!

rsempe (approved):

PR #1653, opened 2024-01-31
feat (document-template): Adjust credit note and invoice templates

lovrocolic:
## Context

Existing PDF templates miss support for `grouped_by` feature. Also, currently, we always display pay in arrear fees first and after that pay in advance fees. However, we need to change it in a way that subscription fee is always on the top, no matter if pay in arrear or pay in advance. Basically, fees ordering should use logic that was implemented before so this part has been reverted.

## Description

This PR covers:
- support for grouped_by feature
- reorder fees

### DIFF
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -104,7 +104,7 @@ class Fee < ApplicationRecord
     return '' unless charge?
     return '' if charge.properties['grouped_by'].blank?
 
-    ' • ' + grouped_by.values.join(' • ')
+    " • #{grouped_by.values.join(' • ')}"
   end
 
   def invoice_sorting_clause
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -100,6 +100,23 @@ class Fee < ApplicationRecord
     charge&.group_properties&.find_by(group:)&.invoice_display_name || group&.name
   end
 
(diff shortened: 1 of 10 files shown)


app/models/fee.rb
```diff
@@ -100,6 +100,23 @@ def group_name
     charge&.group_properties&.find_by(group:)&.invoice_display_name || group&.name
   end
 
+  def grouped_by_display
+    return '' unless charge?
+    return '' if charge.properties['grouped_by'].blank?
+
+    ' • ' + grouped_by.values.join(' • ')
```
jdenquin, line 107: Style/StringConcatenation: Prefer string interpolation to string concatenation. (https://rubystyle.guide#string-interpolation)

app/models/fee.rb
```diff
@@ -100,6 +100,23 @@ def group_name
     charge&.group_properties&.find_by(group:)&.invoice_display_name || group&.name
   end
 
+  def grouped_by_display
```
jdenquin, line 103: can we use helpers in views instead of put this in the model? 🙏

lovrocolic: fixed

jdenquin (approved):

PR #1659, opened 2024-02-01
feat(events) Modify events batch length limit

odnobit:
Added the ability to modify the value of MAX_LENGTH for events batching

## Context

 I've added support for configuring the maximum length of events batching by introducing the `LAGO_EVENTS_BATCH_MAX_LENGTH` env variable.

## Description

Now, users can easily customize the batch length according to their specific requirements without modifying the codebase directly.

### DIFF
--- a/app/services/events/create_batch_service.rb
+++ b/app/services/events/create_batch_service.rb
@@ -2,7 +2,7 @@
 
 module Events
   class CreateBatchService < BaseService
-    MAX_LENGTH = 100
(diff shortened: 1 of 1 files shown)


jdenquin (approved):

jdenquin:
thanks for it @odnobit, just be warned that increasing batch quantity will also add load on your workers and database!

odnobit:
> thanks for it @odnobit, just be warned that increasing batch quantity will also add load on your workers and database!

thank you!
yeah, sure.
but it's the admin's pain :)

PR #1667, opened 2024-02-04
misc(html-invoices): Add route for showing invoices as html

ivannovosad:
## Description

Add route for rendering invoices as html pages in development environment: 

`GET /admin/invoices/:id`

### DIFF
--- a/app/controllers/admin/invoices_controller.rb
+++ b/app/controllers/admin/invoices_controller.rb
@@ -2,8 +2,6 @@
 
 module Admin
   class InvoicesController < BaseController
-    skip_before_action :authenticate, only: :show
-
     def regenerate
       result = ::Invoices::GeneratePdfService.call(invoice:, context: 'admin')
 
--- a/app/controllers/admin/invoices_controller.rb
+++ b/app/controllers/admin/invoices_controller.rb
@@ -2,12 +2,20 @@
 
 module Admin
   class InvoicesController < BaseController
+    skip_before_action :authenticate, only: :show
+
     def regenerate
       result = ::Invoices::GeneratePdfService.call(invoice:, context: 'admin')
 
       return render_error_response(result) unless result.success?
 
       head(:ok)
+    end
+
+    def show
+      service = ::Invoices::GeneratePdfService.new(invoice:)
+
+      render(html: service.render_html)
     end
 
     private
--- a/app/services/invoices/generate_pdf_service.rb
+++ b/app/services/invoices/generate_pdf_service.rb
@@ -21,6 +21,13 @@ module Invoices
       result
     end
 
+    # rubocop:disable Rails/OutputSafety
+    def render_html
+      pdf_service = Utils::PdfGenerator.new(template:, context: invoice)
+      pdf_service.render_html.html_safe
+    end
+    # rubocop:enable Rails/OutputSafety
+
     private
 
     attr_reader :invoice, :context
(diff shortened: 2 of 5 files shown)


app/controllers/admin/invoices_controller.rb
```diff
@@ -10,6 +12,12 @@ def regenerate
       head(:ok)
     end
 
+    def show
+      service = ::Invoices::GeneratePdfService.new(invoice:)
+
+      render html: service.render_html
```
jdenquin, line 18: Style/MethodCallWithArgsParentheses: Use parentheses for method calls with arguments. (https://rubystyle.guide#method-invocation-parens)

app/services/invoices/generate_pdf_service.rb
```diff
@@ -21,6 +21,11 @@ def call
       result
     end
 
+    def render_html
+      pdf_service = Utils::PdfGenerator.new(template:, context: invoice)
+      pdf_service.render_html.html_safe
```
jdenquin, line 26: Rails/OutputSafety: Tagging a string as html safe may be a security risk.

app/services/invoices/generate_pdf_service.rb
```diff
@@ -21,6 +21,13 @@ def call
       result
     end
 
+    # rubocop:disable Rails/OutputSafety
+    def render_html
+      pdf_service = Utils::PdfGenerator.new(template:, context: invoice)
+      pdf_service.render_html.html_safe
+    end
+    #rubocop:enable Rails/OutputSafety
```
jdenquin, line 29: Layout/LeadingCommentSpace: Missing space after `#`. (https://rubystyle.guide#hash-space)

config/routes.rb
```diff
@@ -92,9 +92,11 @@
   namespace :admin do
     resources :memberships, only: %i[create]
     resources :organizations, only: %i[update]
-    resources :invoices do
+    resources :invoices, except: :show do
       post :regenerate, on: :member
     end
+
+    get '/invoices/:id', to: 'invoices#show' if Rails.env.development?
```
vincent-pochet, line 99: What about having a `dev_tools` namespace in the routes?
I think it's better to not mix the concerns. It could be used in the future to add any other rendering logic, like credit notes, or email templates.
What do you think?

rsempe:
I personally still think that we can use the rails console easily and creating a route is not really useful :/ 
But if you think it can be useful for you guys, why not...

```
Invoices::GeneratePdfService.new(invoice:)
invoice.file.url
```

vincent-pochet:
> I personally still think that we can use the rails console easily and creating a route is not really useful :/ But if you think it can be useful for you guys, why not...
> 
> ```
> Invoices::GeneratePdfService.new(invoice:)
> invoice.file.url
> ```

Sure it does work, but you have to purge the file each time you update something and retrieve the link to download the PDF, with the web page, you can just "live debug" without event having the docker PDF instance running

jdenquin:
I think both are useful, PDF rendering with Gotenberg is not the same as the browser one sometimes!

config/routes.rb
ivannovosad: That's a good point, I'll change the route.

app/controllers/admin/invoices_controller.rb
```diff
@@ -2,6 +2,8 @@
 
 module Admin
   class InvoicesController < BaseController
+    skip_before_action :authenticate, only: :show
```
vincent-pochet, line 5: I guess you could remove the changes in this file :)

vincent-pochet (approved):
Looks good!

PR #1671, opened 2024-02-05
fix (invoice-number): use organization timezone in per org numbering

lovrocolic:
## Context

When generating invoice number in `per_organization` mode, we should use organization timezone.

## Description

`organization_sequential_id` usually have following pattern: `prefix-202402-001`. Middle part and sequential id should be generated in the organization scope and based on existing month. However, month scope should take organization timezone into account.

### DIFF
--- a/db/migrate/20240205160647_remove_organization_sequential_id_index_from_invoices.rb
+++ b/db/migrate/20240205160647_remove_organization_sequential_id_index_from_invoices.rb
@@ -2,6 +2,8 @@
 
 class RemoveOrganizationSequentialIdIndexFromInvoices < ActiveRecord::Migration[7.0]
   def change
-    remove_index :invoices, name: 'unique_organization_sequential_id'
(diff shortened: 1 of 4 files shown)


db/migrate/20240205160647_remove_organization_sequential_id_index_from_invoices.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemoveOrganizationSequentialIdIndexFromInvoices < ActiveRecord::Migration[7.0]
+  def change
+    remove_index :invoices, name: 'unique_organization_sequential_id'
```
jdenquin, line 5: Rails/ReversibleMigration: remove_index(without column) is not reversible. (https://rails.rubystyle.guide#reversible-migration, https://api.rubyonrails.org/classes/ActiveRecord/Migration/CommandRecorder.html)

PR #1676, opened 2024-02-06
feat(events-batch): Handle errors in events batch

odnobit:
## Context

Currently, the system returns an error for the entire batch if any of the events fail validation.
imho, this behavior is unfair.

## Description

Implemented validation error handling for events batch processing: now returns error only if all events fail validation; allows operation if a subset of events have validation errors and return list of accepted events.
For example: if the user has sent 3 events in a batch and 1 event fails validation, it returns as follows: `{"events":[{$valid_event0}, {$valid_event1}]}`

What do you think about this change?

### DIFF
--- a/app/services/events/create_batch_service.rb
+++ b/app/services/events/create_batch_service.rb
@@ -20,7 +20,7 @@ module Events
 
       validate_events
 
-      return result.validation_failure!(errors: result.errors) if result.errors.present?
+      return result.validation_failure!(errors: result.errors) if result.events.size() == 0
 
       post_validate_events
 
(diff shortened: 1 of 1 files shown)


jdenquin:
Hello @odnobit 

It was a choice on our side to return an error if any of the event is invalid.
Maybe the best solution is to return a 200 with both event ingested and invalid events

odnobit:
> Hello @odnobit
> 
> It was a choice on our side to return an error if any of the event is invalid. Maybe the best solution is to return a 200 with both event ingested and invalid events

Hello!
Ok, I will implement that

jdenquin:
@odnobit any news on your side regarding this change?

PR #1677, opened 2024-02-06
fix(invoice-pdf): handle invoice pdf template issues

lovrocolic:
## Context

There are few smaller issues in PDF invoice template

## Description

This PR fixes:

- customer email list -> This list is actually string with elements separated by comma. HTML by default break string on multiple lines (when needed) upon empty space detection, so we need to ensure that email list string is correctly formatted.
- removes unnecessary word in multiple subscriptions breakdown page
- displays adjusted subscription fee display name

### DIFF
--- a/app/views/templates/invoices/v4.slim
+++ b/app/views/templates/invoices/v4.slim
@@ -439,7 +439,7 @@ html
               = customer.city
           .body-2 = customer.state
           .body-2 = ISO3166::Country.new(customer.country)&.common_name
-          .body-2 = customer.email
+          .body-2 = customer.email.gsub(/,\s*/, ', ')
           - if customer.tax_identification_number.present?
             .body-2 = I18n.t('invoice.tax_identification_number', tax_identification_number: customer.tax_identification_number)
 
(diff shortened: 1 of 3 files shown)


app/views/templates/invoices/v4.slim
```diff
@@ -439,7 +439,7 @@ html
               = customer.city
           .body-2 = customer.state
           .body-2 = ISO3166::Country.new(customer.country)&.common_name
-          .body-2 = customer.email
+          .body-2 = customer.email.gsub(/,\s*/, ', ')
```
vincent-pochet, line 442: Do we have invalid emails? If yes, should we fix them at DB level?

vincent-pochet (approved):

app/views/templates/invoices/v4.slim
lovrocolic: Nope, DB attribute is valid. This is just formatting needed for HTML so that it can break this string to multiple lines correctly (upon space detection), when string is really long...

PR #1682, opened 2024-02-08
fix(customers): Fix provider emails

jdenquin:
## Context

When using a list of emails for customer, it creates error on providers side.

## Description

- Split the email and take the first one in the list.

### DIFF
--- a/app/services/payment_provider_customers/adyen_service.rb
+++ b/app/services/payment_provider_customers/adyen_service.rb
@@ -117,7 +117,7 @@ module PaymentProviderCustomers
         recurringProcessingModel: 'UnscheduledCardOnFile',
         expiresAt: Time.current + 69.days,
       }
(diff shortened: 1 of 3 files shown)


vincent-pochet (approved):

app/services/payment_provider_customers/adyen_service.rb
```diff
@@ -117,7 +117,7 @@ def payment_link_params
         recurringProcessingModel: 'UnscheduledCardOnFile',
         expiresAt: Time.current + 69.days,
       }
-      prms[:shopperEmail] = customer.email if customer.email
+      prms[:shopperEmail] = customer.email.split(',').first if customer.email
```
lovrocolic, line 120: since customer.email is something that we receive from the UI, probably it would be good if we also remove all whitespaces (if any):
`prms[:shopperEmail] = customer.email.strip.split(',').first if customer.email`

PR #1699, opened 2024-02-15
fix (charge models): adjust volume charge model logic

lovrocolic:
## Context

If aggregation amount is the number between two tiers, currently the error is thrown.

## Description

Tiers:
0 -> 5
6 -> 10
11 -> nil

If aggregation result is 5.5, we should apply second tier values.

### DIFF
--- a/app/services/charges/charge_models/volume_service.rb
+++ b/app/services/charges/charge_models/volume_service.rb
@@ -50,17 +50,9 @@ module Charges
       end
 
       def matching_range
-        @matching_range ||= ranges.each_with_index do |range, index|
-          is_matching = if index.zero?
-            range[:from_value] <= number_of_units && (!range[:to_value] || number_of_units <= range[:to_value])
-          else
-            (range[:from_value] - 1) < number_of_units && (!range[:to_value] || number_of_units <= range[:to_value])
-          end
-
-          return range if is_matching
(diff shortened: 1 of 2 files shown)


app/services/charges/charge_models/volume_service.rb
```diff
@@ -50,9 +50,17 @@ def per_unit_total_amount
...
-        @matching_range ||= ranges.find do |range|
-          range[:from_value] <= number_of_units && (!range[:to_value] || number_of_units <= range[:to_value])
+        @matching_range ||= ranges.each_with_index do |range, index|
+          is_matching = if index.zero?
+            range[:from_value] <= number_of_units && (!range[:to_value] || number_of_units <= range[:to_value])
+          else
+            (range[:from_value] - 1) < number_of_units && (!range[:to_value] || number_of_units <= range[:to_value])
+          end
+
+          return range if is_matching
```
vincent-pochet, line 60: It seems it could work with something like:
```suggestion
        @matching_range ||= ranges.find do |range|
          range[:from_value] <= number_of_units.ceil && (!range[:to_value] || number_of_units <= range[:to_value])
```

lovrocolic: Actualy yes, I over-engineered it 😄

vincent-pochet (approved):

PR #1703, opened 2024-02-16
fix (jobs-inside-transaction): add setup for after commit everywhere and fix customer.vies_check webhook

lovrocolic:
## Context

Common scenario in Lago is that we have to trigger active job inside DB transaction. SInce job tries to fetch and use record that is saved inside transaction, we often end up with `ActiveJob::DeserializationError` if job is processed before transaction commit command.

## Description

We usually handle this issue by setting wait parameter to 3 seconds upon calling active job. 

This PR sets up `ActiveCommitEverywhere` gem so that we can handle this issue without a 3 seconds hack.

Also, as part of this PR new library is used to correctly call `customer.vies_check` webhook.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem 'oauth2'
 gem 'rack-cors'
 
 # Database
+gem 'after_commit_everywhere'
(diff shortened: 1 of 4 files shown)


vincent-pochet (approved):
Nice! :rocket:

ivannovosad (approved):

PR #1711, opened 2024-02-21
fix(subscriptions): fix edit subscription attached to parent plan

lovrocolic:
## Context

Subscription edition works in a way that it changes plan's values. It works fine if subscription is attached to overridden plan.
However, if subscription is attached to parent plan it will change also values on all other subscriptions.

### DIFF
--- a/app/services/subscriptions/update_service.rb
+++ b/app/services/subscriptions/update_service.rb
@@ -71,7 +71,7 @@ module Subscriptions
       else
         Plans::OverrideService.call(
           plan: current_plan,
-          params: params[:plan_overrides].to_h.with_indifferent_access
+          params: params[:plan_overrides].to_h.with_indifferent_access,
         )
       end
     end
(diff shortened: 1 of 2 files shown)


app/services/subscriptions/update_service.rb
```diff
@@ -61,6 +60,22 @@ def process_subscription_at_change(subscription)
...
+
+      if current_plan.parent_id
+        Plans::UpdateService.call(
+          plan: current_plan,
+          params: params[:plan_overrides].to_h.with_indifferent_access,
+        )
+      else
+        Plans::OverrideService.call(
+          plan: current_plan,
+          params: params[:plan_overrides].to_h.with_indifferent_access
```
jdenquin, line 74: Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)

vincent-pochet (approved):

PR #1715, opened 2024-02-22
Add .tool-versions for asdf with ruby 3.2.3

dorianmariecom:
This is needed to have asdf run the correct version of ruby

## Context

I was trying to setup the project and ran `bundle`, but I got:

```
Your Ruby version is 3.3.0, but your Gemfile specified 3.2.3
```

## Description

The alternative is using `~/.asdfrc` with:

```
legacy_version_file = yes
```


jdenquin:
thanks for it @dorianmariecom !
BTW we recommend to use our dev env config : https://github.com/getlago/lago/wiki/Development-Environment
we all use the same and its easier to manage for us!

PR #1717, opened 2024-02-22
require a SEGMENT_WRITE_KEY

dorianmariecom:
Otherwise the app doesn't start with:

Failure/Error: require_relative '../config/environment'

ArgumentError:
  Write key must be initialized

## Context

When doing `rspec` I would get an error.

## Description

Solution was to have:

```
SEGMENT_WRITE_KEY=1
```

or

```
LAGO_DISABLE_SEGMENT=true
```

### DIFF
--- a/config/initializers/analytics_ruby.rb
+++ b/config/initializers/analytics_ruby.rb
@@ -17,7 +17,7 @@ unless ENV['LAGO_DISABLE_SEGMENT'] == 'true'
 
   SEGMENT_CLIENT = Segment::Analytics.new(
     {
-      write_key: ENV['SEGMENT_WRITE_KEY'],
+      write_key: ENV.fetch('SEGMENT_WRITE_KEY', 'changeme'),
       on_error: proc { |status, msg| Sentry.capture_exception(SegmentError.new(status, msg)) },
       stub: Rails.env.development? || Rails.env.test?,
     },


config/initializers/analytics_ruby.rb
```diff
@@ -17,7 +17,7 @@ def initialize(status, error_message)
 
   SEGMENT_CLIENT = Segment::Analytics.new(
     {
-      write_key: ENV['SEGMENT_WRITE_KEY'],
+      write_key: ENV.fetch('SEGMENT_WRITE_KEY'),
```
vincent-pochet, line 20: We do not want SEGMENT_WRITE_KEY to be a requirement for the application. It used by the lago team to monitor the usage of the tool. It should not be required to perform the specs

I would recommend to use a fallback value instead:

```suggestion
      write_key: ENV.fetch('SEGMENT_WRITE_KEY', 'change_me'),
```

vincent-pochet (commented):
Thank you for you PR @dorianmariecom 

I have one comment regarding the proposed changed

config/initializers/analytics_ruby.rb
dorianmariecom: Done, I used `changeme` to be consistent with the rest of the app:

```
.env.dist:DATABASE_URL=postgresql://lago:changeme@localhost:5432/lago_test
config/database.yml:    password: changeme
config/database.yml:    password: changeme
config/initializers/analytics_ruby.rb:      write_key: ENV.fetch('SEGMENT_WRITE_KEY', 'changeme'),
```

vincent-pochet (approved):

PR #1719, opened 2024-02-22
update db/clickhouse_schema.rb and db/schema.rb with default generated schemas

dorianmariecom:
Those are the schemas generated by docker-compose up

## Context

I was developing on the app when the schemas were updated


vincent-pochet:
Closing as we do not want to commit the kafka secrets inside the `clickhouse_schema.rb`

PR #1721, opened 2024-02-23
Add .env.development and .env.test for local setup

dorianmariecom:
This makes local development easier

I was trying to setup the api locally natively but I had more success with the docker-compose approach


vincent-pochet:
@dorianmariecom Thank you for this PR

Since the application has multiple components and databases, we are encouraging people to use a docker/docker-compose approach as it simplify greatly the setup, especially for dev/test setup.
Note sure we want to encourage people to follow a native approach for now as it would bring more issues with dependencies, versions, specific setup and so on

jdenquin:
I can confirm we do not want to encourage the local installation without Docker for now

PR #1726, opened 2024-02-26
feat(edit-plan-amount-cents): Allow editing plan amount cents

lovrocolic:
## Context

Currently it is not possible to edit plan amount cents on edit plan page and on edit subscription page.

## Description

We want to allow users to edit plan/subscription amount cents so that they don't need to create new plan with each change.

Also, if downgraded plan's amount cents is updated to the value that is less than the one in downgraded plan, pending subscription should be destroyed.

### DIFF
--- a/app/services/plans/update_service.rb
+++ b/app/services/plans/update_service.rb
@@ -206,8 +206,8 @@ module Plans
     def process_downgraded_subscriptions
       return unless plan.subscriptions.active.exists?
 
-      Subscription.where(previous_subscription_id: plan.subscriptions.active, status: :pending).find_each do |sub|
-        sub.destroy! if plan.amount_cents < sub.plan.amount_cents
+      Subscription.where(previous_subscription: plan.subscriptions.active, status: :pending).find_each do |sub|
+        sub.mark_as_canceled! if plan.amount_cents < sub.plan.amount_cents
       end
     end
   end
--- a/app/services/plans/update_service.rb
+++ b/app/services/plans/update_service.rb
@@ -10,10 +10,12 @@ module Plans
 
     def call
       return result.not_found_failure!(resource: 'plan') unless plan
+      old_amount_cents = plan.amount_cents
 
       plan.name = params[:name] if params.key?(:name)
       plan.invoice_display_name = params[:invoice_display_name] if params.key?(:invoice_display_name)
       plan.description = params[:description] if params.key?(:description)
+      plan.amount_cents = params[:amount_cents] if params.key?(:amount_cents)
 
       # NOTE: Only name and description are editable if plan
       #       is attached to subscriptions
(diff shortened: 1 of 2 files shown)


app/services/plans/update_service.rb
```diff
@@ -197,5 +199,16 @@ def discard_charge!(charge)
...
       Invoice.where(id: draft_invoice_ids).update_all(ready_to_be_refreshed: true) # rubocop:disable Rails/SkipsModelValidations
     end
+
+    # NOTE: We should remove pending subscriptions
+    #       if plan has been downgraded but amount cents became less than downgraded value. This pending subscription
+    #       is not relevant in this case and downgrade should be ignored
+    def process_downgraded_subscriptions
+      return unless plan.subscriptions.active.exists?
+
+      Subscription.where(previous_subscription_id: plan.subscriptions.active, status: :pending).each do |sub|
```
jdenquin, line 209: Rails/FindEach: Use `find_each` instead of `each`. (https://rails.rubystyle.guide#find-each)

app/services/plans/update_service.rb
```diff
@@ -197,5 +199,16 @@ def discard_charge!(charge)
...
       Invoice.where(id: draft_invoice_ids).update_all(ready_to_be_refreshed: true) # rubocop:disable Rails/SkipsModelValidations
     end
+
+    # NOTE: We should remove pending subscriptions
+    #       if plan has been downgraded but amount cents became less than downgraded value. This pending subscription
+    #       is not relevant in this case and downgrade should be ignored
+    def process_downgraded_subscriptions
+      return unless plan.subscriptions.active.exists?
+
+      Subscription.where(previous_subscription_id: plan.subscriptions.active, status: :pending).find_each do |sub|
```
vincent-pochet, line 209: Shouldn't it be?
```suggestion
      Subscription.where(previous_subscription_id: plan.subscriptions.active.select('id'), status: :pending).find_each do |sub|
```
or 
```suggestion
      Subscription.where(previous_subscription: plan.subscriptions.active, status: :pending).find_each do |sub|
```


It seems to work but it feels a bit weird :D

app/services/plans/update_service.rb
```diff
@@ -197,5 +199,16 @@ def discard_charge!(charge)
...
+        sub.destroy! if plan.amount_cents < sub.plan.amount_cents
```
vincent-pochet, line 210: Any reason for destroying rather than canceling?

vincent-pochet (approved):

app/services/plans/update_service.rb
lovrocolic: Yep, using cancel status now, wasn't aware of it and it wasn't part of specs:)

PR #1728, opened 2024-02-27
feat(filter): Add validations for charge filters and values

vincent-pochet:
## Context

## Description

This PR adds the logic to validate `charge_filters#properties` and `charge_filter_values#value`


app/models/charge_filter_value.rb
```diff
@@ -9,6 +9,15 @@ class ChargeFilterValue < ApplicationRecord
...
 
   validates :value, presence: true
+  validate :validate_value
 
   default_scope -> { kept }
+
+  private
+
+  def validate_value
+    return if billable_metric_filter&.values&.include?(value) # rubocop:disable Performance/InefficientHashSearch
```
rsempe, line 19:
```suggestion
    return if billable_metric_filter.values.include?(value) # rubocop:disable Performance/InefficientHashSearch
```

rsempe (approved):

PR #1734, opened 2024-02-29
fix(credit-notes): recalculate credit note item upon draft invoice refresh

lovrocolic:
## Context

When draft invoice fee got changed, we do not recalculate attached credit note item value. It has to change because end-user could end up with invalid credit note amount that would be later applied.

## Description

This PR handles credit note recalculation upon fee change.

In prior PR - https://github.com/getlago/lago-api/pull/1733 scenario tests for this case were added. The goal of this change is to cover credit note item recalculation so that scenario tests passes with expected values.

### DIFF
--- a/app/services/credit_notes/refresh_draft_service.rb
+++ b/app/services/credit_notes/refresh_draft_service.rb
@@ -2,9 +2,10 @@
 
 module CreditNotes
   class RefreshDraftService < BaseService
-    def initialize(credit_note:, fee:)
+    def initialize(credit_note:, fee:, old_fee_values:)
       @credit_note = credit_note
       @fee = fee
+      @old_fee_values = old_fee_values
 
       super
     end
@@ -14,7 +15,16 @@ module CreditNotes
       return result unless credit_note.draft?
 
       credit_note.applied_taxes.destroy_all
-      credit_note.items.update_all(fee_id: fee.id) # rubocop:disable Rails/SkipsModelValidations
+      credit_note.items.each do |item|
+        item.fee_id = fee.id
+
+        if old_fee_values.any?
+          old_entry = old_fee_values.find { |h| h[:credit_note_item_id] == item.id }
+          item.precise_amount_cents = calculate_item_value(item, old_entry[:fee_amount_cents]) if old_entry
+        end
+
+        item.save(validate: false) # rubocop:disable Rails/SkipsModelValidations
+      end
 
       taxes_result = CreditNotes::ApplyTaxesService.call(
         invoice: fee.invoice,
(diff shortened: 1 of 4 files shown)


app/services/credit_notes/refresh_draft_service.rb
```diff
@@ -45,6 +55,10 @@ def call
 
     private
 
-    attr_accessor :credit_note, :fee
+    attr_accessor :credit_note, :fee, :old_fee_values
+
+    def calculate_item_value(item, old_fee_amount_cents)
+      (item.precise_amount_cents.fdiv(old_fee_amount_cents)) * fee.amount_cents
```
jdenquin, line 61: Style/RedundantParentheses: Don't use parentheses around a method call.

app/services/credit_notes/refresh_draft_service.rb
```diff
@@ -14,7 +15,16 @@ def call
...
-      credit_note.items.update_all(fee_id: fee.id) # rubocop:disable Rails/SkipsModelValidations
+      credit_note.items.each do |item|
+        item.fee_id = fee.id
+
+        if old_fee_values.any?
+          old_entry = old_fee_values.find { |h| h[:credit_note_item_id] == item.id }
+          item.precise_amount_cents = calculate_item_value(item, old_entry[:fee_amount_cents]) if old_entry
+        end
+
+        item.save(validate: false) # rubocop:disable Rails/SkipsModelValidations
```
vincent-pochet, line 26: Any reason to skip the validation since now we are changing the values?

app/services/credit_notes/refresh_draft_service.rb
vincent-pochet: Could you explain this computation? It's not clear what it's doing

app/services/credit_notes/refresh_draft_service.rb
rsempe, line 26: why saving without validations?

app/services/credit_notes/refresh_draft_service.rb
```diff
@@ -45,6 +55,10 @@ def call
 
     private
 
-    attr_accessor :credit_note, :fee
+    attr_accessor :credit_note, :fee, :old_fee_values
+
+    def calculate_item_value(item, old_fee_amount_cents)
+      item.precise_amount_cents.fdiv(old_fee_amount_cents) * fee.amount_cents
```
vincent-pochet, line 61: Could you explain this computation? It's not clear what it's doing

lovrocolic: Added a comment in the code. Basically we didn't change credit note item value with refresh. However it needs to be recalculated based on the ration between old fee value and new fee value

app/services/credit_notes/refresh_draft_service.rb
lovrocolic: We were skipping validations before with `update_all` so I just kept behaviour. Added validation trigger!

lovrocolic: We were skipping validations before with update_all so I just kept behaviour. Added validation trigger!

vincent-pochet (approved):

PR #1736, opened 2024-03-01
fix: Deliver email with SendEmailJob

vincent-pochet:
## Context

This PR is an attempt to handle retry and errors with email deliveries.

## Description

The PR changes the jobs used to deliver email from the default `ActionMailer::MailDeliveryJob` to an app specific `SendEmailJob`. The new job simply inherit from the first one to keep the delivery behavior

The goal is to allow us:
- To use a custom sidekiq queue rather than the `default` one
- To handle error catching and retry (`ActiveJob::DeserializationError` and `LagoHttpClient::HttpError` for now)

### DIFF
--- a/app/jobs/send_email_job.rb
+++ b/app/jobs/send_email_job.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class SendEmailJob < ActionMailer::MailDeliveryJob
+  queue_as 'mailers'
+
+  retry_on ActiveJob::DeserializationError, wait: :exponentially_longer, attempts: 6
+  retry_on LagoHttpClient::HttpError, wait: :exponentially_longer, attempts: 6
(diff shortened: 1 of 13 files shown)


app/jobs/send_email_job.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class SendEmailJob < ActionMailer::MailDeliveryJob
+  queue_as 'webhook'
```
rsempe, line 4: why `webhook` here?

vincent-pochet: Well, I wanted to move out of the `default` one. But maybe we should add an `email` one

rsempe (approved):

rsempe (approved):

PR #1741, opened 2024-03-01
misc: Remove wait time for subscription billing

vincent-pochet:
## Description

This PR removes the random wait time when enqueuing the recurring subscription billing, in order to let Sidekiq responsible for balancing the load over the workers. This should help to reduce the number of dead jobs that occurs during the billing day


rsempe (approved):
👍

PR #1754, opened 2024-03-06
fix(terminate-subscription): update automatic credit note creation logic

lovrocolic:
## Context

When subscription auto-termination feature is used by setting `ending_at`, there is a chance that unnecessary credit note can be created. Described scenario can happen if `ending_at` is set on billing day and if termination job is processed before billing job on that day. In that case, since invoice has not been created, we should not issue a credit note.

## Description

This PR handles described case. New logic that checks the case is added and also specs have been updated and added.


lovrocolic:
For this change new PR will be opened

PR #1763, opened 2024-03-07
fix: cover edge case when pending subscriptions need to be canceled

lovrocolic:
## Context

This edge case is originally part of update-plan-subscription-fee feature. If plan that got updated has pending subscription whose new subscription fee is greater that the one from original plan, pending subscription needs to be canceled..

## Description

This PR covers the described edge case and adds related specs

### DIFF
--- a/app/services/plans/update_service.rb
+++ b/app/services/plans/update_service.rb
@@ -47,6 +47,7 @@ module Plans
         process_charges(plan, params[:charges]) if params[:charges]
         process_minimum_commitment(plan, params[:minimum_commitment]) if params[:minimum_commitment] && License.premium?
         process_downgraded_subscriptions if old_amount_cents != plan.amount_cents
+        process_pending_subscriptions if old_amount_cents != plan.amount_cents
       end
 
       result.plan = plan.reload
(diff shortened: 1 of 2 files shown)


spec/services/plans/update_service_spec.rb
```diff
@@ -171,6 +171,37 @@
           end
         end
       end
+
+      context 'when there are pending subscriptions which are not relevant after the amount cents update' do
```
jdenquin, line 175: RSpec/RepeatedExampleGroupDescription: Repeated context block description on line(s) [155] (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/RepeatedExampleGroupDescription)

app/services/plans/update_service.rb
```diff
@@ -47,6 +47,7 @@ def call
         process_charges(plan, params[:charges]) if params[:charges]
         process_minimum_commitment(plan, params[:minimum_commitment]) if params[:minimum_commitment] && License.premium?
         process_downgraded_subscriptions if old_amount_cents != plan.amount_cents
+        process_pending_subscriptions if old_amount_cents != plan.amount_cents
```
vincent-pochet, line 50: Since the condition is the same, maybe you could group them:
```suggestion
        if old_amount_cents != plan.amount_cents
          process_downgraded_subscriptions
          process_pending_subscriptions
        end
```

vincent-pochet (approved):

PR #1780, opened 2024-03-13
feat: adding scheduler monitoring for sentry if enabled

inventionlabsSydney:
## Context

Sentry scheduled tasks can easily be monitored by Sentry crons checkin functions.
This allows us to know fairly immediately whether a sidekiq task has stopped processing.

## Description

ENV variable "SENTRY_ENABLE_CRONS" will activate a mixin cron checkin for sentry within the activejobs.
These mixins are preconfigured to match the clockwork functions found in clock.rb

No customer installation changes required.

Dependencies for sentry have been lifted to:
```
-gem 'sentry-rails', '~> 5.12.0'
-gem 'sentry-ruby', '~> 5.12.0'
-gem 'sentry-sidekiq', '~> 5.12.0'
+gem 'sentry-rails', '~> 5.16', '>= 5.16.1'
+gem 'sentry-ruby', '~> 5.16', '>= 5.16.1'
+gem 'sentry-sidekiq', '~> 5.16', '>= 5.16.1'
```


app/jobs/clock/activate_subscriptions_job.rb
```diff
@@ -2,10 +2,16 @@
 
 module Clock
   class ActivateSubscriptionsJob < ApplicationJob
+    include Sentry::Cron::MonitorCheckIns
+
     queue_as 'clock'
 
     unique :until_executed, on_conflict: :log
 
+    if ENV['SENTRY_ENABLE_CRONS']
```
jdenquin, line 11: it would be awesome to create a concern for it that take the slug in params, wdyt?

jdenquin (commented):
Very good idea for observability, thanks for your contribution!

inventionlabsSydney:
@jdenquin What's with Pronto failing? can you please lend a hand.

jdenquin:
> @jdenquin What's with Pronto failing? can you please lend a hand.

It's a token issue with PR coming from an external contributor, I have to fix this!

inventionlabsSydney:
@jdenquin  I've made a few updates to this:

1. I have moved the mixin to a concern and then just used that as a basic module to include a perform override, this then allows us to put the config of the monitor in the clock.rb directly, and using the super placed above it ensures that the sentry monitor only checks in if the task actually performs without exception.
2. There's a docker-compose test file which I use to run unittests, this is more just to not pollute my dev environment but would be useful to others who wish to isolate their dev work to docker containers.

If you could please review this that would be great!

clock.rb
```diff
@@ -19,51 +19,75 @@ module Clockwork
   # NOTE: All clocks run every hour to take customer timezones into account
 
   every(5.minutes, 'schedule:activate_subscriptions') do
-    Clock::ActivateSubscriptionsJob.perform_later
+    Clock::ActivateSubscriptionsJob.perform_later(
+        :slug => "activate_subscriptions", :cron => "*/5 * * * *"
```
jdenquin, line 23: you can use this format

```ruby
Clock::ActivateSubscriptionsJob.perform_later(slug: 'activate_subscriptions, cron: '*/5 * * * *')
```

vincent-pochet (approved):
Really good contribution, thank you very much @inventionlabsSydney!

Could you move chages related to the test env (Makefile and docker-compose.test.yml) in a separate pull request, as they are not directly related to the sentry scheduler monitoring logic?

vincent-pochet:
Also, it seems that the specs are not running due to a small syntax error

inventionlabsSydney:
@vincent-pochet you'll have to forgive me, I am still just getting used to working in RoR again. Having spent the last ten years in python!

Will make the relevant changes and commit back.

vincent-pochet:
> @vincent-pochet you'll have to forgive me, I am still just getting used to working in RoR again. Having spent the last ten years in python!
> 
> Will make the relevant changes and commit back.

No worries!
And thank you for this great contribution :)

vincent-pochet:
Closing in favor of https://github.com/getlago/lago-api/pull/2109

jdenquin:
@inventionlabsSydney 👋 , we have some issue with SentryCron, its loaded on each job even if sentry is not configured and we're disabled sentry if no key is provided (so it can break everything for many users).
Do you have a value to uses it? whats different from monitoring sidekiq dead jobs?

inventionlabsSydney:
Hi,
I think it can probably be deprecated out if monitoring of sidekiq's is fine.
The primary issue here is ensuring any error's raised get caught into sentry, if you've integrated a better solution for this, then fine.
Else we should look at properly disabling it for users not using sentry.

jdenquin:
This is what I did here : https://github.com/getlago/lago-api/pull/3366
Your main issue was to know when a clock job is failing? in any case, its a Sidekiq job, if it fails with our configuration, it's in dead job and it triggers an exception

PR #1781, opened 2024-03-13
feat: adding SENTRY_ENVIRONMENT to sentry config

inventionlabsSydney:
## Roadmap Task

👉  [Add support for SENTRY_ENVIRONMENT environment variable](https://getlago.canny.io/developer-experience/p/add-support-for-sentryenvironment-environment-variable)

## Context

Sentry allows environments to be specified in their projects so you can filter issues in production, staging, development, etc.

## Description

This change allows users to use env variable 'SENTRY_ENVIRONMENT' to set whatever name of this installations environment in sentry. If nothing is specified it will default to 'production'

No changes required unless you wish to change the env to something you have already setup.

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -5,5 +5,6 @@ if ENV['SENTRY_DSN']
     config.dsn = ENV['SENTRY_DSN']
     config.breadcrumbs_logger = %i[active_support_logger http_logger]
     config.traces_sample_rate = 0
+    config.environment = ENV['SENTRY_ENVIRONMENT'] || "production"
   end
 end


config/initializers/sentry.rb
```diff
@@ -5,5 +5,6 @@
     config.dsn = ENV['SENTRY_DSN']
     config.breadcrumbs_logger = %i[active_support_logger http_logger]
     config.traces_sample_rate = 0
+    config.environment = ENV['SENTRY_ENVIRONMENT'] || "production"
```
jdenquin, line 8: sentry is using `RAILS_ENV` value, this change would break actual configurations, maybe we can just fall back to `Rails.env`

inventionlabsSydney: Ah, this was an oversight on my behalf, you're right. I will fix this now.

jdenquin (approved):

PR #1793, opened 2024-03-15
feat(filter): Accept multiple values on a charge_filter_value

vincent-pochet:
## Context

## Description

This PR turns `change_filter_values#value` into `change_filter_values#values` to accept multiple values for a single filter.

### DIFF
--- a/spec/models/charge_filter_value_spec.rb
+++ b/spec/models/charge_filter_value_spec.rb
@@ -10,29 +10,38 @@ RSpec.describe ChargeFilterValue, type: :model do
   it { is_expected.to belong_to(:charge_filter) }
   it { is_expected.to belong_to(:billable_metric_filter) }
 
-  it { is_expected.to validate_presence_of(:value) }
+  it { is_expected.to validate_presence_of(:values) }
 
-  describe '#valdiate_value' do
+  describe '#valdiate_values' do
     subject(:charge_filter_value) do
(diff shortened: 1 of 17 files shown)


spec/models/charge_filter_value_spec.rb
```diff
@@ -10,29 +10,29 @@
   it { is_expected.to belong_to(:charge_filter) }
   it { is_expected.to belong_to(:billable_metric_filter) }
 
-  it { is_expected.to validate_presence_of(:value) }
+  it { is_expected.to validate_presence_of(:values) }
 
   describe '#valdiate_value' do
```
ivannovosad, line 15:
```suggestion
  describe '#validate_value' do
```

spec/models/charge_filter_value_spec.rb
```diff
@@ -10,29 +10,29 @@
...
 
     let(:billable_metric_filter) { create(:billable_metric_filter) }
-    let(:value) { billable_metric_filter.values.first }
+    let(:values) { [billable_metric_filter.values.first] }
 
     it { expect(charge_filter_value).to be_valid }
 
     context 'when value is not included in billable_metric_filter values' do
-      let(:value) { 'invalid_value' }
+      let(:values) { ['invalid_value'] }
```
ivannovosad, line 26: We can add a spec with an empty array as well.

ivannovosad (approved):

PR #1794, opened 2024-03-15
fix (date-boundaries): yearly plan boundaries for anniversary interval were incorrect

lovrocolic:
## Context

For current usage, when interval is `anniversary` and plan is `yearly`, we haven't calculated boundaries correctly.

## Description

This fix handles logic for finding starting boundary for yearly plan when interval is anniversary.

### DIFF
--- a/app/services/subscriptions/dates/yearly_service.rb
+++ b/app/services/subscriptions/dates/yearly_service.rb
@@ -103,7 +103,7 @@ module Subscriptions
 
       def period_started_in_last_year?(date)
         return true if date.month < subscription_at.month
-        return true if (date.month == subscription_at.month) && (subscription_at.day > date.day)
+        return true if (date.month == subscription_at.month) && (date.day < subscription_at.day)
 
         false
       end
--- a/app/services/subscriptions/dates/yearly_service.rb
+++ b/app/services/subscriptions/dates/yearly_service.rb
@@ -102,8 +102,8 @@ module Subscriptions
       end
 
       def period_started_in_last_year?(date)
-        return true if (date.month < subscription_at.month)
-        return true if (date.month == subscription_at.month)  && (subscription_at.day > date.day)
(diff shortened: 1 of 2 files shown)


app/services/subscriptions/dates/yearly_service.rb
```diff
@@ -100,6 +100,13 @@ def compute_charges_duration(from_date:)
 
         compute_duration(from_date:)
       end
+
+      def period_started_in_last_year?(date)
+        return true if (date.month < subscription_at.month)
```
jdenquin, line 105: Style/ParenthesesAroundCondition: Don't use parentheses around the condition of an `if`. (https://rubystyle.guide#no-parens-around-condition)

app/services/subscriptions/dates/yearly_service.rb
```diff
@@ -100,6 +100,13 @@ def compute_charges_duration(from_date:)
...
+        return true if (date.month == subscription_at.month)  && (subscription_at.day > date.day)
```
jdenquin, line 106: Layout/ExtraSpacing: Unnecessary spacing detected.

app/services/subscriptions/dates/yearly_service.rb
```diff
@@ -100,6 +100,13 @@ def compute_charges_duration(from_date:)
 
         compute_duration(from_date:)
       end
+
+      def period_started_in_last_year?(date)
+        return true if date.month < subscription_at.month
+        return true if (date.month == subscription_at.month) && (subscription_at.day > date.day)
```
ivannovosad, line 106:
```suggestion
        return true if (date.month == subscription_at.month) && (date.day < subscription_at.day)
```

Looks good! 
Here's a small suggestion so that the line reads the same way as the one above

ivannovosad (approved):

vincent-pochet (approved):

PR #1795, opened 2024-03-18
fix(payment-methods): Fix provider payment methods

ivannovosad:
## Context

Stripe payment methods were automatically overridden by Lago.

**Steps to reproduce:**

Create a customer via API with stripe as the payment provider and "provider_payment_methods": ["card","sepa_debit"];

Update the customer account via API by sending only external_id and a random field, such as city (do not send anything related to billing_configuration);

In the API response, you can see that Lago automatically removed "sepa_debit" from the list of available payment methods.

Expected behaviour: when updating a customer, only the attributes included in the request payload should be modified.

## Description

This PR fixes updating provider customer specifically the payment methods when no payment method is sent.

### DIFF
--- a/app/services/payment_provider_customers/create_service.rb
+++ b/app/services/payment_provider_customers/create_service.rb
@@ -20,16 +20,7 @@ module PaymentProviderCustomers
         provider_customer.sync_with_provider = params[:sync_with_provider].presence
       end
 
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):
:+1:

PR #1796, opened 2024-03-18
misc: Replace each with find_each in plan and customer delete services

vincent-pochet:
## Context

This PR intends to replace some calls to `each` with `find_each` when looping over DB resources in Plans and Customers deletion services, in order to reduce the memory usage when dealing with a lot of resources.


ivannovosad (approved):
🚀

We can replace also this one: https://github.com/getlago/lago-api/blob/fix-plans-destroy-job/app/services/coupons/destroy_service.rb#L17

Issue #1812, opened 2024-03-25
[BUG]: estimate fees API create an entry in database.


ccfiel:
**Describe the bug**
Calling estimate fees API https://docs.getlago.com/api-reference/events/estimated-fee will create an event entry in the database. The expected flow gives the estimated fee based on the event parameter and it should not create a database entry.

**To Reproduce**

See the video to reproduce the behavior

https://github.com/getlago/lago-api/assets/1788360/f137a155-3d3d-40f0-913c-5c3585b17785

**Expected behavior**
After calling the estimate fees API it should NOT create a database entry.

**Version**
- Lago: v1.0.0
-

vincent-pochet:
Thank you for reporting this @ccfiel 
The fix have been merged into main, it will be part of the next release.

PR #1813, opened 2024-03-25
Refactor event transaction in estimate_pay_in_advance_service.rb

ccfiel:
Fix issue:

https://github.com/getlago/lago-api/issues/1812

### DIFF
--- a/app/services/fees/estimate_pay_in_advance_service.rb
+++ b/app/services/fees/estimate_pay_in_advance_service.rb
@@ -20,7 +20,7 @@ module Fees
 
       fees = []
 
-      event.transaction do
+      EventsRecord.transaction do
         event.save!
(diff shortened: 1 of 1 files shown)


app/services/fees/estimate_pay_in_advance_service.rb
```diff
@@ -20,10 +20,10 @@ def call
 
       fees = []
 
-      ActiveRecord::Base.transaction do
+      event.transaction do
```
vincent-pochet, line 23: To keep the consistency with other DB transaction, could you use `EventsRecord` instead?

```suggestion
      EventsRecord.transaction do
```

vincent-pochet (commented):
Thank you @ccfiel for this fix!

PR #1818, opened 2024-03-27
feat(disputes): Add dispute columns to invoices

ivannovosad:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/dispute-notification

## Context

PSP integration should listen to disputes and chargebacks to update payment status.

## Description

This PR adds a migration that adds `payment_dispute_lost_at` column to `invoices` table.
It also handles validations in `Invoice` model with and model specs as well.
It adds this field to `InvoiceSerializer`.

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -11,7 +11,6 @@ class Invoice < ApplicationRecord
 
   before_save :ensure_organization_sequential_id, if: -> { organization.per_organization? }
   before_save :ensure_number
-  before_save :assign_payment_dispute_lost_at
 
   belongs_to :customer, -> { with_discarded }
   belongs_to :organization
@@ -266,6 +265,18 @@ class Invoice < ApplicationRecord
     subscription_from != charges_from && subscription_to != charges_to
   end
 
+  def mark_as_disputed!(timestamp = Time.current)
+    self.payment_dispute_lost_at ||= timestamp
+    self.payment_disputed = true
+    save!
+  end
+
+  def mark_as_not_disputed!
+    self.payment_dispute_lost_at = nil
+    self.payment_disputed = false
+    save!
+  end
+
   private
 
   def should_assign_sequential_id?
@@ -339,10 +350,6 @@ class Invoice < ApplicationRecord
     raise(SequenceError, 'Unable to acquire lock on the database') unless result
 
     result
-  end
-
-  def assign_payment_dispute_lost_at
-    self.payment_dispute_lost_at = DateTime.current if !payment_disputed_was && payment_disputed
   end
 
   def switched_from_customer_numbering?
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -96,8 +96,7 @@ class Invoice < ApplicationRecord
   validates :issuing_date, :currency, presence: true
   validates :timezone, timezone: true, allow_nil: true
   validates :total_amount_cents, numericality: { greater_than_or_equal_to: 0 }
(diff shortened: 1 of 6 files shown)


app/models/invoice.rb
```diff
@@ -338,6 +341,10 @@ def generate_organization_sequential_id
     result
   end
 
+  def assign_payment_dispute_lost_at
```
lovrocolic, line 344: Can we rather set `payment_dispute_lost_at` using the common approach in the Lago app:
```
def mark_as_disputed!(timestamp = Time.current)
    self.payment_dispute_lost_at ||= timestamp
    self.payment_disputed = true
  end
```
in terms of readability, I would rather avoid callbacks if possible. This method can be called wherever we need it.

app/models/invoice.rb
```diff
@@ -96,6 +96,8 @@ class Invoice < ApplicationRecord
   validates :issuing_date, :currency, presence: true
   validates :timezone, timezone: true, allow_nil: true
   validates :total_amount_cents, numericality: { greater_than_or_equal_to: 0 }
+  validates :payment_disputed, inclusion: { in: [false] }, if: :voided?
+  validates :payment_disputed, inclusion: { in: [false, true] }, unless: :voided?
```
lovrocolic, line 100: I think this second validation is unnecessary. First should be enough I think

ivannovosad: The second validation (as well as the first one) prevents `payment_disputed` to be valid when `nil`. 

```
invoice = FactoryBot.create(:invoice)
irb(main):014:0> invoice.payment_disputed = nil
=> nil
irb(main):015:0> invoice.valid?
=> false
```

app/models/invoice.rb
```diff
@@ -263,6 +265,18 @@ def different_boundaries_for_subscription_and_charges(subscription)
     subscription_from != charges_from && subscription_to != charges_to
   end
 
+  def mark_as_disputed!(timestamp = Time.current)
+    self.payment_dispute_lost_at ||= timestamp
+    self.payment_disputed = true
```
julienbourdeau, line 270: Is there a sceneario where you have `payment_disputed = true` but `payment_dispute_lost_at = nil?`? Can we have disputed payment that you don't lose or if we "win", we `mark_as_not_disputed!`?

I'm thinking the `payment_disputed` could be a timestamp `payment_disputed_at` instead of a boolean, like we do for `discarded_at` for instance. If the lost_at can be different, we can have 2 timestamps, other only keep `payment_disputed_at`.

I'm reading the spec now, sorry I didn't review it earlier 🙏

db/schema.rb
```diff
@@ -481,8 +481,8 @@
     t.bigint "unit_amount_cents", default: 0, null: false
     t.boolean "pay_in_advance", default: false, null: false
     t.decimal "precise_coupons_amount_cents", precision: 30, scale: 5, default: "0.0", null: false
-    t.decimal "total_aggregated_units"
     t.string "invoice_display_name"
+    t.decimal "total_aggregated_units"
```
rsempe, line 485: not related to your PR

app/models/invoice.rb
lovrocolic: @ivannovosad I am not sure we usually add such validations regarding nil value, especially if this field is set internally. However, if you prefer to add it, I would rather use `exclusion: { in: [nil] }`

ivannovosad: I removed the `payment_disputed` completely, we only have `payment_dispute_lost_at` now

julienbourdeau (approved):

vincent-pochet (approved):

PR #1831, opened 2024-03-29
feat(filter): Handle adjusted fees and improve ignored filters

vincent-pochet:
## Context

## Description

This PR:
- Adds the logic to handle `AdjustedFee` with filters
- Fixes the behavior with filters and `grouped_by`
- Improve the handling of `ignored_filters` with `ALL_VALUES`

### DIFF
--- a/app/serializers/v1/customers/charge_usage_serializer.rb
+++ b/app/serializers/v1/customers/charge_usage_serializer.rb
@@ -48,6 +48,8 @@ module V1
       end
 
       def filters(fees)
+        return [] unless fees.first.charge&.filters&.any?
+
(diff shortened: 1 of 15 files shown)


app/serializers/v1/customers/charge_usage_serializer.rb
```diff
@@ -48,6 +48,8 @@ def groups(fees)
       end
 
       def filters(fees)
+        return [] unless fees.first.charge&.filters&.any?
```
rsempe, line 51: maybe extracting this into the `Fee` model to avoid the duplication?

rsempe (approved):
🤯 👏

PR #1833, opened 2024-03-29
feat(filter): GraphQL and BillableMetricFilter improvements

vincent-pochet:
## Context

## Description

This PR:
- Improve the GraphQl API for billable metric and charge filter
- Makes sure that the edited ChargeFilter and ChargeFilterValue keeps their order after edition

### DIFF
--- a/app/graphql/types/customers/usage/charge_filter.rb
+++ b/app/graphql/types/customers/usage/charge_filter.rb
@@ -6,7 +6,7 @@ module Types
       class ChargeFilter < Types::BaseObject
         graphql_name 'ChargeFilterUsage'
 
-        field :id, ID, null: false, method: :charge_filter_id
+        field :id, ID, null: true, method: :charge_filter_id
 
         field :amount_cents, GraphQL::Types::BigInt, null: false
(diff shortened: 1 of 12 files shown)


app/graphql/types/customers/usage/charge_filter.rb
```diff
@@ -6,7 +6,7 @@ module Usage
       class ChargeFilter < Types::BaseObject
         graphql_name 'ChargeFilterUsage'
 
-        field :id, ID, null: false, method: :charge_filter_id
+        field :id, ID, null: true, method: :charge_filter_id
```
rsempe, line 9: why?

rsempe (approved):

app/graphql/types/customers/usage/charge_filter.rb
vincent-pochet: For the default values (the values attached to the charge itself), we are generating a fee without a persisted charge filter. Since the front-end need to know that this fee is the one matching the default values, and some other attributes (units, unit_amount...), we are returning a "virtual" charge filter with ID = nil

PR #1839, opened 2024-04-02
chore(env): define ActiveRecord encryption config

julienbourdeau:
## Description

Defining configuration here rather than in env vars makes it easier to run tests outside Docker.

When running tests, do we need actual encryption value? We could put the values from `.github/workflows/spec.yml` inside the config if they matter but I don't think they do.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -19,12 +19,17 @@ Rails.application.configure do
   config.action_dispatch.show_exceptions = false
   config.action_controller.allow_forgery_protection = false
   config.active_storage.service = :test
+
   config.action_mailer.perform_caching = false
   config.action_mailer.delivery_method = :test
+
+  config.active_record.encryption.primary_key = 'test'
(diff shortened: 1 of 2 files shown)


config/environments/test.rb
```diff
@@ -19,12 +19,17 @@
...
   config.action_mailer.delivery_method = :test
+
+  config.active_record.encryption.primary_key = 'test'
+  config.active_record.encryption.deterministic_key = 'test'
+  config.active_record.encryption.key_derivation_salt = 'test'
+
   config.active_support.deprecation = :stderr
   config.active_support.disallowed_deprecation = :raise
   config.active_support.disallowed_deprecation_warnings = []
-  config.action_mailer.delivery_method = :test
```
julienbourdeau, line 27: This is just a duplicate of line 23

vincent-pochet (approved):
Nice one!

PR #1840, opened 2024-04-02
fix(invoices): Fix boundaries when in advance anniversary

vincent-pochet:
## Context

An issue have been identified when a subscription is billed in advance on a monthly basis if the subscription starts on a 30 and the billing month is shorter than the subscription month

Subscription_at = 30th of January, billed on an anniversary basis
Plan: Monthly, billed in advance

On the billing day of February 29th, the following boundaries were returned: 
- From datetime: 2024-01-30 00:00:00.000000000
- To datetime: 2024-02-28T59:59.999999000 UTC +00:00

The expected boundaries are:
- From datetime: 2024-02-29 00:00:00.000000000
- To datetime: 2024-03-29T59:59.999999000 UTC +00:00

## Description

This PR ensure that the boundaries are correct

### DIFF
--- a/app/services/subscriptions/dates/monthly_service.rb
+++ b/app/services/subscriptions/dates/monthly_service.rb
@@ -104,9 +104,9 @@ module Subscriptions
         year = nil
         month = nil
 
-        # NOTE: in anniversary mode if both subscription date and current date are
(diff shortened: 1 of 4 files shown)


rsempe (approved):
👏

PR #1842, opened 2024-04-03
feat (integrations): add base model and db setup for integrations

lovrocolic:
## Context

Lago currently does not support accounting integrations.

## Description

This PR adds basic model and DB setup for supporting accounting integrations

### DIFF
--- a/app/models/integrations/base_integration.rb
+++ b/app/models/integrations/base_integration.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Integrations
+  class BaseIntegration < ApplicationRecord
+    include PaperTrailTraceable
+
+    self.table_name = 'integrations'
+
+    belongs_to :organization
+
+    encrypts :secrets
+
+    validates :code, uniqueness: { scope: :organization_id }
+    validates :name, presence: true
+
+    def secrets_json
+      JSON.parse(secrets || '{}')
+    end
+
+    def push_to_secrets(key:, value:)
+      self.secrets = secrets_json.merge(key => value).to_json
+    end
+
+    def get_from_secrets(key)
+      secrets_json[key.to_s]
+    end
+
+    def push_to_settings(key:, value:)
+      self.settings ||= {}
+      settings[key] = value
+    end
+
+    def get_from_settings(key)
+      (settings || {})[key]
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


app/models/integrations/netsuite_integration.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Integrations
+  class NetsuiteIntegration < BaseIntegration
+
```
jdenquin, line 5: Layout/EmptyLinesAroundClassBody: Extra empty line detected at class body beginning. (https://rubystyle.guide#empty-lines-around-bodies)

app/models/integrations/base_integration.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def push_to_settings(key:, value:)
+      self.settings ||= {}
+      settings[key] = value
+    end
+
+    def get_from_settings(key)
+      (settings || {})[key]
+    end
```
ivannovosad, line 35: We could extract these methods to concerns since they are also defined in `PaymentProviders::BaseProvider`:

```
module SettingsStorable
  extend ActiveSupport::Concern
  
  def push_to_settings(key:, value:)
    self.settings ||= {}
    settings[key] = value
  end

  def get_from_settings(key)
    (settings || {})[key]
  end
end
```

```
module SecretsStorable
  extend ActiveSupport::Concern
  
  included do
    encrypts :secrets
  end
  
  def secrets_json
   JSON.parse(secrets || '{}')
  end

  def push_to_secrets(key:, value:)
    self.secrets = secrets_json.merge(key => value).to_json
  end

  def get_from_secrets(key)
    secrets_json[key.to_s]
  end
end

```

spec/models/integrations/netsuite_integration_spec.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Integrations::NetsuiteIntegration, type: :model do
+  subject(:netsuite_integration) { build(:netsuite_integration, attributes) }
+
+  let(:attributes) {}
```
ivannovosad, line 8: We can remove `attributes` here.

ivannovosad (changes requested):

app/models/integrations/base_integration.rb
lovrocolic: 👌

ivannovosad (approved):

PR #1821, opened 2024-03-27
feat(trial): Introduce Subscriptions::FreeTrialBillingService

julienbourdeau:
## Context

Following https://github.com/getlago/lago-api/pull/1820, we need a job find all subscriptions which has a trial which just ended.

## Description

This introduce a new service to find all subscription with a trial period ending today. The job will run every hour to follow customer timezones. 

We introduce a new timestamp in subscriptions to know if the trial has ended. This ensure we only run the job once. Without this flag, we would retrieve the same subscription 24 times and dispatch the webhook 24 times. The invoice can't be used to know if the job was run because we only change plans with `pay_in_advance = true`.

This adds the fourth definition of the `at_time_zone` method. I'm hoping to refactor this later.

**Note that the job doesn't run yet**. The following PR will add it to the scheduler and generate invoice and send the webhooks.

### Migration

The data is migrated for all subscriptions with a trial that has already ended. This is done to ensure a minimum of consistency in the data.

The logic for `initial_started_at` is implemented to avoid the N+1 queries triggered by the original method.

I didn't reuse the SQL query from the service because `find_by_sql` returns an array, not a collection so you can't load the data in batch (like with `find_each`). We'd have to load ALL subscription in the array 🙈

Alternatively, we can probably write an SQL query that update all data at once, without using the model. It would be fun but I'm not sure it's worth the effort.

PR #1847, opened 2024-04-04
feat(trial): Invoice subscription fee at the end of the trial period

julienbourdeau:
## Context

When a subscription has a trial period, we want to invoice the customer at the end of the trial, not at the beginning.

Following #1820 and #1821

## Description

* Disable billing when creating a new subscription if there is a trial
* Disable billing when activating a pending subscription if there is a trial
* Run the `Subscriptions::FreeTrialBillingService` introduced in #1821 every hour
* Migrate data for existing subscription with trial already ended
* Ensure we don't bill customers if an invoice was created when the subscription started (for sub created before this feature)

### Free trial reminder

Keep in mind that if you're trial ends on day X, this day is *NOT FREE*.
If the 10 days free trial start on the 10th at 14:00, it will end on the 20th at 14:00. You get the 10th to the 19th for free. Day 20 is your first pay day even if you're still in trial for a few hours.

This was not changed in this PR or other related PRs.

## Note

* Introduction of `invoice.skip_charge`
* Refactor of charge fees for first invoice, see: https://github.com/getlago/lago-api/pull/1836
* Handle grace_period
* Handle minimum_commitment

### DIFF
--- a/app/services/subscriptions/free_trial_billing_service.rb
+++ b/app/services/subscriptions/free_trial_billing_service.rb
@@ -13,10 +13,10 @@ module Subscriptions
         if subscription.plan_pay_in_advance &&
            !subscription.was_already_billed_today &&
            !already_billed_on_day_one?(subscription)
-          BillSubscriptionJob.perform_later([subscription], timestamp, skip_charges: true)
+          BillSubscriptionJob.perform_later([subscription], timestamp, recurring: false, skip_charges: true)
         end
 
-        subscription.update!(trial_ended_at: subscription.trial_end_date_from_query)
+        subscription.update!(trial_ended_at: subscription.trial_end_utc_date_from_query)
 
         SendWebhookJob.perform_later('subscription.trial_ended', subscription)
       end
@@ -42,7 +42,7 @@ module Subscriptions
         SELECT DISTINCT
           plans.pay_in_advance AS plan_pay_in_advance,
           already_billed_today.invoiced_count > 0 AS was_already_billed_today,
-          #{trial_end_date} as trial_end_date_from_query,
+          #{trial_end_date} as trial_end_utc_date_from_query,
           subscriptions.*
         FROM
           subscriptions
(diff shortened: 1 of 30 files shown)


spec/services/subscriptions/free_trial_billing_service_spec.rb
```diff
@@ -5,8 +5,10 @@
 RSpec.describe Subscriptions::FreeTrialBillingService, type: :service do
   subject(:service) { described_class.new(timestamp:) }
 
+  before { travel_to timestamp }
```
vincent-pochet, line 8: Curious about this one. Is it reseted  after the spec file is executed?

vincent-pochet (approved):

app/services/subscriptions/free_trial_billing_service.rb
```diff
@@ -10,36 +10,52 @@ def initialize(timestamp: Time.current)
...
           INNER JOIN initial_started_at ON initial_started_at.external_id = subscriptions.external_id
           INNER JOIN customers ON subscriptions.customer_id = customers.id
           INNER JOIN organizations ON customers.organization_id = organizations.id
+          LEFT JOIN already_billed_today ON already_billed_today.subscription_id = subscriptions.id
         WHERE
           subscriptions.status = 1
           AND subscriptions.trial_ended_at IS NULL
-          AND DATE(initial_started_at#{at_time_zone} + plans.trial_period * INTERVAL '1 day') = DATE('#{timestamp}'#{at_time_zone})
+          AND #{trial_end_date + at_time_zone} <= '#{timestamp}'#{at_time_zone}
+          AND #{trial_end_date + at_time_zone} > '#{timestamp - 2.hours}'#{at_time_zone}
```
julienbourdeau, line 58: Maybe we should get rid of this. Any subsciptions that terminated and doesn't have `trial_ended_at` should be treated.

spec/services/subscriptions/free_trial_billing_service_spec.rb
julienbourdeau: I was expecting it to yes, but now I wonder if it was just because the `rails_helper` in my previous codebase was doing it 🤔 

In this case, I removed it because I think it was a left over of an initial version.

app/services/subscriptions/free_trial_billing_service.rb
```diff
@@ -31,7 +31,13 @@ def call
     # during the customer trial period
     # Unfortunately, this introduces an N+1 query
     def already_billed_on_day_one?(subscription)
-      subscription.invoice_subscriptions.where(timestamp: subscription.started_at.all_day).exists?
+      Fee.subscription_kind.where(
+        invoice_id: subscription.invoice_subscriptions.select('invoices.id').joins(:invoice).where(
+          'invoices.invoice_type' => :subscription,
+          'invoices.status' => %i[draft finalized],
+          timestamp: subscription.started_at.all_day,
```
vincent-pochet, line 38: Not new, but I'm not sure it exactly right if customer timezone is not UTC :thinking:

julienbourdeau: Very good point, I keep thinking about this and it's still unclear 🙈

PR #1850, opened 2024-04-05
feat(integrations-mutations): Add mutations for netsuite integration

ivannovosad:
## Context

## Description

This PR adds support for graphql mutations related to integrations models


app/models/integrations/netsuite_integration.rb
```diff
@@ -35,5 +35,37 @@ def client_id=(value)
     def client_id
       get_from_settings('client_id')
     end
+
+    def sync_credit_notes=(value)
```
vincent-pochet, line 39: Note for future improvements: maybe we should thins at a way to define setting accessors in a more concise way, for example via a new concern or something similar.

```ruby
include SettingAccessors

setting_accessors :sync_invoices, :sync_credit_notes
```

vincent-pochet (approved):

app/models/integrations/netsuite_integration.rb
ivannovosad: That's a good idea! I added those methods to concerns.

So this is possible now in models:

```
    settings_accessors :client_id, :account_id
    secrets_accessors :connection_id, :client_secret
```

vincent-pochet (approved):

PR #1852, opened 2024-04-08
feat (accounting-integrations): add integration destroy mutation

lovrocolic:
## Context

Currently, Lago does not support accounting integrations

## Description

This PR is part of the connection part. It provides destroy mutation so that integration can be destroyed from the UI


app/services/integrations/destroy_service.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Integrations
+  class DestroyService < BaseService
+    def destroy(id:)
+      integration = Integrations::BaseIntegration.find_by(
+        id:,
+        organization_id: result.user.organization_ids,
```
vincent-pochet, line 8: Should we inject the integration to destroy, rather than relying on this?

app/services/integrations/destroy_service.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Integrations
+  class DestroyService < BaseService
+    def destroy(id:)
```
vincent-pochet, line 5: What about using the `call` approach, and passing the arguments in the initializer?

ivannovosad (commented):
I added one suggestion so that the destroy mutation follows the logic of other destroy mutations in the project.

ivannovosad (approved):

PR #1861, opened 2024-04-11
chore(config): Move Sidekiq config to initializer

julienbourdeau:
## Description

Very small improvements:

* Move sidekiq session config to initializer rather than `routes.rb`.
* Only `require 'sidekiq/web'` if we're going to mount it

See https://github.com/getlago/lago-api/pull/1071/


ivannovosad (approved):

vincent-pochet (approved):
Nice!

PR #1868, opened 2024-04-15
fix(invoice): Prevent duplicated non-recurring invoices

vincent-pochet:
## Context

In some cases, due to race conditions, duplicated non-recurring invoices are created by the system, (for example when the clock job stars a subscription)

Today, only recurring invoices are protected against the duplication via a unique index on `recurring`, `charges_from_datetime` and `charges_to_datetime` columns of the `invoice_subscriptions` table

## Description

This PR tries to address the problem by
- Adding a new `invoicing_reason` on the `invoice_subscriptions` table as an enum accepting `subscription_started`, `subscription_terminated`, `in_advance_charge` and `subscription_periodic` values
  - It will replace the `recurring` column in the future
- Adding two new unique indexes on the `invoice_subscriptions` table
  - subscription_id, invoice_type where invoice_type = 'started'
  - subscription_id, invoice_type where invoice_type = 'terminated'
- Populating these fields depending of the context

### DIFF
--- a/app/models/invoice_subscription.rb
+++ b/app/models/invoice_subscription.rb
@@ -13,6 +13,15 @@ class InvoiceSubscription < ApplicationRecord
   monetize :subscription_amount_cents, disable_validation: true, allow_nil: true
   monetize :total_amount_cents, disable_validation: true, allow_nil: true
 
+  INVOICING_REASONS = {
+    subscription_starting: 'subscription_starting',
+    subscription_periodic: 'subscription_periodic',
+    subscription_terminating: 'subscription_terminating',
+    in_advance_charge: 'in_advance_charge',
+  }.freeze
+
+  enum invoicing_reason: INVOICING_REASONS
+
   scope :order_by_charges_to_datetime,
         lambda {
           condition = <<-SQL
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -117,7 +117,14 @@ module Subscriptions
         # NOTE: Since job is launched from inside a db transaction
         #       we must wait for it to be committed before processing the job.
         #       We do not set offset anymore but instead retry jobs
-        after_commit { BillSubscriptionJob.perform_later([new_subscription], Time.zone.now.to_i, skip_charges: true) }
+        after_commit do
+          BillSubscriptionJob.perform_later(
+            [new_subscription],
+            Time.zone.now.to_i,
+            invoicing_reason: :subscription_starting,
+            skip_charges: true,
+          )
+        end
       end
 
       if new_subscription.active?
@@ -162,7 +169,13 @@ module Subscriptions
         # NOTE: Since job is launched from inside a db transaction
         #       we must wait for it to be committed before processing the job
         #       We do not set offset anymore but instead retry jobs
-        after_commit { BillSubscriptionJob.perform_later(billable_subscriptions, Time.zone.now.to_i + 1.second) }
+        after_commit do
+          BillSubscriptionJob.perform_later(
+            billable_subscriptions,
+            Time.zone.now.to_i + 1.second,
+            invoicing_reason: :upgrading,
+          )
+        end
       end
 
       new_subscription
--- a/app/services/subscriptions/terminate_service.rb
+++ b/app/services/subscriptions/terminate_service.rb
@@ -70,7 +70,7 @@ module Subscriptions
       else
         [subscription]
       end
-      BillSubscriptionJob.perform_later(billable_subscriptions, timestamp)
+      BillSubscriptionJob.perform_later(billable_subscriptions, timestamp, invoicing_reason: :upgrading)
 
       SendWebhookJob.perform_later('subscription.terminated', subscription)
       SendWebhookJob.perform_later('subscription.started', next_subscription)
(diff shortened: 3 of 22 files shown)


app/models/invoice_subscription.rb
```diff
@@ -13,6 +13,15 @@ class InvoiceSubscription < ApplicationRecord
   monetize :subscription_amount_cents, disable_validation: true, allow_nil: true
   monetize :total_amount_cents, disable_validation: true, allow_nil: true
 
+  INVOICE_TYPE = {
+    starting: 'starting',
+    periodic: 'periodic',
+    terminating: 'terminating',
+    charge: 'charge',
```
julienbourdeau, line 20: starting, periodic, terminating are kind of related to a lifecycle. I find that `charge` stands out and I dunno what to expect in this invoice.

app/services/subscriptions/create_service.rb
```diff
@@ -2,8 +2,6 @@
 
 module Subscriptions
   class CreateService < BaseService
-    include Utils::TransactionalJobs
```
julienbourdeau, line 5: Really cool refactor! I think this was only used here so we can get rid of the concern definition too. If this PR takes a while for discussion (because I think it's a significant change) we could extract this part 🙏

app/services/subscriptions/terminate_service.rb
```diff
@@ -70,7 +70,7 @@ def terminate_and_start_next(timestamp:)
       else
         [subscription]
       end
-      BillSubscriptionJob.perform_later(billable_subscriptions, timestamp)
+      BillSubscriptionJob.perform_later(billable_subscriptions, timestamp, invoice_type: :upgrading)
```
julienbourdeau, line 73: Here billable_subscriptions typically has 2 subscriptions. The one you just ended and the new one you upgraded too. I'd expect 2 different types. The old one would be `terminating` and the new one would be `starting`.

If there is only one type per group of subscription, then maybe the type should be on Invoice rather than InvoiceSubscription 🤔

julienbourdeau (commented):
I'm happy to specify this "type" further but I'm not sure yet about the current types proposed and especially the relation between `invoices.invoice_type` and `invoice_subscriptions.invoice_type`.

julienbourdeau:
Maybe we could improve the "duplication protection" here:

https://github.com/getlago/lago-api/blob/1efbd79c976de91ec5d8aa13299e2b4dc5130ba7/app/services/invoices/create_invoice_subscription_service.rb#L16-L21

For some reason, it only works for recurring invoices

https://github.com/getlago/lago-api/blob/1efbd79c976de91ec5d8aa13299e2b4dc5130ba7/app/services/invoices/create_invoice_subscription_service.rb#L60-L68

app/models/invoice_subscription.rb
```diff
@@ -13,6 +13,15 @@ class InvoiceSubscription < ApplicationRecord
   monetize :subscription_amount_cents, disable_validation: true, allow_nil: true
   monetize :total_amount_cents, disable_validation: true, allow_nil: true
 
+  INVOICE_TYPE = {
```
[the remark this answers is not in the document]
vincent-pochet: It will not be a real persisted type. Basically it means that the invoice itself can be related to two distinct subscription, one newly started and an other one terminated

vincent-pochet:
> Maybe we could improve the "duplication protection" here:
> 
> https://github.com/getlago/lago-api/blob/1efbd79c976de91ec5d8aa13299e2b4dc5130ba7/app/services/invoices/create_invoice_subscription_service.rb#L16-L21
> 
> For some reason, it only works for recurring invoices
> 
> https://github.com/getlago/lago-api/blob/1efbd79c976de91ec5d8aa13299e2b4dc5130ba7/app/services/invoices/create_invoice_subscription_service.rb#L60-L68

@julienbourdeau 
Indeed, it's enforcing only the recurring invoice duplication, because today, we have no easy way to prevent duplication of invoice for starting or terminating subscription. That's why I'm adding the new enum to have more granularity on the index.
One merged, it will allow the removal of the `recurring` as the `periodic` value of the enum is here for the same purpose.

app/models/invoice_subscription.rb
vincent-pochet: I agree the naming is confusing. I'm currently turning it this way:

```ruby
  INVOICING_REASONS = {
    subscription_starting: 'subscription_starting',
    subscription_periodic: 'subscription_periodic',
    subscription_terminating: 'subscription_terminating',
    in_advance_charge: 'in_advance_charge',
  }.freeze
```

app/services/subscriptions/terminate_service.rb
vincent-pochet: Actually an invoice can be generated for:
- A single starting subscription
- A single terminating subscription
- One or multiple periodic invoice subscriptions
- A single terminating subscription and a starting subscription.

What we want to enforce:
- No duplication for the same boundaries on periodic invoices (done using the recurring flag)
- No duplication for the start of a single subscription
- No duplication for the termination of a single subscription
Those two lasts are the ones targeted by this PR.

The flag/type cannot be at invoice level, because of the n-n relation between subscriptions and invoices.

Let imagine the following case:
- A subscription upgraded
- As a consequence the subscription is terminated and a new one is started
- An invoice is generated with the fees for the terminated one and the subscription fees to the new one
- Because of some race condition a terminating invoice is emitted for the initial invoice 
  -> it should fail, but today it's not and we end-up with two invoices with the same boundaries

rsempe (approved):
nice 👍

PR #1869, opened 2024-04-15
feat (accounting-integrations): add logic for calling nango sync endpoint

lovrocolic:
## Context

## Description

This PR adds base service for communicating with Lago's accounting integrations aggregator - Nango.

Also, the sync job is added. Lago needs to call Nango sync endpoint upon integration object creation. Sync is needed for Nango so that they can pull data from accounting provider.
Nango's enpoint structure will be the same for all other providers.

### DIFF
--- a/app/services/integrations/aggregator/base_service.rb
+++ b/app/services/integrations/aggregator/base_service.rb
@@ -5,7 +5,7 @@ require 'lago_http_client'
 module Integrations
   module Aggregator
     class BaseService < BaseService
-      BASE_URL = 'https://api.nango.dev/'.freeze
+      BASE_URL = 'https://api.nango.dev/'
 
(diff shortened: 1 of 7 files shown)


app/services/integrations/aggregator/base_service.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      def http_client
+        LagoHttpClient::Client.new(endpoint_url)
+      end
+
+      def endpoint_url
+        "#{BASE_URL}#{action_path}"
+      end
+
+      def generate_headers
```
ivannovosad, line 32: I'd call it just headers, but it's up to you.

```suggestion
      def headers
```

ivannovosad (approved):
Looks good!

PR #1871, opened 2024-04-15
feat(okta): Add Okta Authorize mutation and logic

jdenquin:
## Description

- Add `oktaAuthorize` mutation
- Returns the okta authorize url if a configured domain match the user email


app/services/auth/okta_service.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Auth
+  class OktaService < BaseService
+    def authorize(email:)
```
vincent-pochet, line 5: What about having something like

```ruby
module Auth
  module Okta
    class AuthorizeService
      def initialize(email:)
      ...

      def call
```

It will allow us to keep the `call` approach in the mutation: `Auth::Okta::AuthorizeService.call(email:)`

app/services/auth/okta_service.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      initialize_state
+
+      params = {
+        client_id: okta_integration.client_id,
+        response_type: 'code',
+        response_mode: 'query',
+        scope: 'openid profile email',
+        redirect_uri: "#{ENV['LAGO_FRONT_URL']}/auth/okta/callback",
+        state: CurrentContext.okta_state,
```
vincent-pochet, line 22: Is there a way to pass this context in parameters / response, rather than relying on the context?

jdenquin: I need to keep the state for the next call from front end to check if it's the same coming back from Okta, we can discuss it tomorrow!

app/services/auth/okta_service.rb
jdenquin: I would need to refact `Auth::GoogleService` as well

app/graphql/mutations/auth/okta/authorize.rb
```diff
@@ -0,0 +1,20 @@
...
+    module Okta
+      class Authorize < BaseMutation
+        graphql_name 'OktaAuthorize'
+
+        argument :email, String, required: true
+
+        type Types::Auth::Okta::Authorize
+
+        def resolve(email:)
+          result = ::Auth::Okta::AuthorizeService.new(email:).call
```
ivannovosad, line 14:
```suggestion
          result = ::Auth::Okta::AuthorizeService.call(email:)
```

ivannovosad (commented):
I'm adding a suggestion with preferred way of calling services.

app/services/auth/okta/authorize_service.rb
```diff
@@ -0,0 +1,59 @@
...
+      attr_reader :email
+
+      def initialize_state
+        redis_config = { url: ENV['REDIS_URL'] }
+        redis_config.merge({ password: ENV['REDIS_PASSWORD'] }) if ENV['REDIS_PASSWORD'].present?
+        redis_client = ::Redis.new(url: ENV['REDIS_URL'])
+
+        redis_client.set(state, email)
+        redis_client.expire(state, 90)
+      end
```
julienbourdeau, line 52: Any reason we can't use `Rails.cache` here? I'm surprised to see the redis client initiated in the service. That would remove the need for the `mock_redis ` gem too. What am I missing?

```suggestion
      def initialize_state
        Rails.cache.write(state, email, expires_in: 90.seconds)
      end
```

jdenquin: I think it's better yes

jdenquin: done @julienbourdeau

PR #1872, opened 2024-04-16
poc: Add feature flags capability

julienbourdeau:
## Introduction

This PR adds the ability to use [Feature Flag](https://launchdarkly.com/blog/what-are-feature-flags/). It's been discussed in the team that we will need it at some point. This PR is a POC for discussion. I wanted to show what it takes to add Flipper and what it looks like. I'm open to other solutions if the team doesn't like it.

It adds the [Flipper gem](https://www.flippercloud.io/docs/introduction) to the project with the Redis driver. It doesn't rely on any external services.

Feature flags can be enabled globally or for specific organizations.

### Feature flags, not plan flags

There are really (at least) two kind of "feature flags":

* Feature flags that are temporary to release a feature temporarily. They are enabled progressively but should ultimately be removed. They are managed by admins, the user cannot turn on and off these flags (except if you want to, like for a beta).

* Feature flags that depends on the plan. You can upgrade to high plans to enabled new feature that are gated in lower plans. The users can usually enabled/disable them by changing plan themselves.  In Lago it's `org.premium?` for now.

Note that this is also different from permissions!

In the future we could have things like the following.

```ruby
def can_use_feature_x?(org, user)
  License.premium? && FeatureFlag.enabled?(:x, actor: org) && user.can?('view:x')
end
```

## How to use

In your code, you can check if a flag is enabled for an actor. You can also use the console to enable it.

```ruby
FeatureFlag.enabled? :feature1
# => false

FeatureFlag.disabled? :feature1
# => true

FeatureFlag.enabled? :feature1, actor: organization1
# => true

FeatureFlag.enabled? :feature1, actor: another_org
# => false
```

### In Redis

CleanShot 2024-04-16 at 10 57 41@2x

### Frontend

In GraphQL, you can retrieve an object with all the feature flags and whether they are enabled or not for the current organization.

```js
{
  feature1: true,
  feature2: false
}
```

In a typical SAAS app, we could filter out the disabled flags to keep them secret and not leak them to the frontend but Lago is open source, there is nothing to hide.

### Using a hash to define all the feature flags.

Flipper can create feature flags on the fly but I didn't really like it workin with it before, that's why I used a defined hardcoded list and `ensure_feature_flag_exists!` in dev mode.
This allows us to easily type the object in GraphQL and can be added to the documentation.

### `FeatureFlag.sync!`

This method removed or add the flags from Redis based on the definition hash. It should run after deploys, just like `rails db:migrate`.

### Going Further: Add the UI (not part of the POC)

There is a UI where you can add flag to an actor or enable/disable. Today, we don't have a good way to load protected admin tools so the UI is not part of this POC.

We could rely on Rack and basic auth, but I'd rather discuss this seperately.

```ruby
flipper_app = Flipper::UI.app do |builder|
  builder.use Rack::Auth::Basic do |username, password|
    username == ENV['ADMIN_USER'] && password == ENV['ADMIN_PASSWORD']
  end
end
mount flipper_app, at: '/feature-flags'
```

CleanShot 2024-04-16 at 09 24 32@2x

CleanShot 2024-04-16 at 09 24 19@2x

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -81,7 +81,7 @@ gem 'karafka'
 gem 'valvat', require: false
 
 # Feature flags
-gem 'flipper'
+gem 'flipper', '1.3.0'
 gem 'flipper-redis'
 
 group :development, :test, :staging do
--- a/Gemfile
+++ b/Gemfile
@@ -80,6 +80,10 @@ gem 'karafka'
 # Taxes
 gem 'valvat', require: false
 
+# Feature flags
+gem 'flipper'
+gem 'flipper-redis'
+
 group :development, :test, :staging do
   gem 'factory_bot_rails'
   gem 'faker'
--- a/Gemfile
+++ b/Gemfile
@@ -83,6 +83,7 @@ gem 'valvat', require: false
 # Feature flags
 gem 'flipper'
 gem 'flipper-redis'
+gem 'flipper-ui'
 
 group :development, :test, :staging do
   gem 'factory_bot_rails'
--- a/Gemfile
+++ b/Gemfile
@@ -83,7 +83,6 @@ gem 'valvat', require: false
 # Feature flags
 gem 'flipper'
 gem 'flipper-redis'
-gem 'flipper-ui'
 
 group :development, :test, :staging do
   gem 'factory_bot_rails'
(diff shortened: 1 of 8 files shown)


julienbourdeau:
Let's close this one for now, we'll see if we need them later. I'm still convinced Flipper is the best way to handle feature flags with Rails.

PR #1878, opened 2024-04-16
feat(accounting-integrations): Send restlet endpoint to nango

lovrocolic:
## Context

## Description

This PR handles following cases:
- When connection is created or updated Lago needs to send restlet endpoint URL to aggregator (Nango)
- Also, this PR removes job that triggers Nango sync upon connection creation. We realised that it can be done automatically by providing extra configuration in Nango. However, until the whole integration is done, we will leave related code in the codebase, just in case.

### DIFF
--- a/spec/services/integrations/aggregator/send_restlet_endpoint_service_spec.rb
+++ b/spec/services/integrations/aggregator/send_restlet_endpoint_service_spec.rb
@@ -21,13 +21,15 @@ RSpec.describe Integrations::Aggregator::SendRestletEndpointService do
       integration.save!
     end
 
-    it 'successfully calls sync endpoint' do
+    it 'successfully sends restlet endpoint' do
       send_restlet_endpoint_service.call
 
-      expect(LagoHttpClient::Client).to have_received(:new)
-        .with(endpoint)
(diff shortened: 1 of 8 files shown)


vincent-pochet (approved):

spec/services/integrations/aggregator/send_restlet_endpoint_service_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      integration.save!
+    end
+
+    it 'successfully calls sync endpoint' do
+      send_restlet_endpoint_service.call
+
+      expect(LagoHttpClient::Client).to have_received(:new)
+        .with(endpoint)
+      expect(lago_client).to have_received(:post_with_response) do |payload|
+        expect(payload[:restletEndpoint]).to eq('https://example.com')
```
ivannovosad, line 30:
```suggestion
      aggregate_failures do
        expect(LagoHttpClient::Client).to have_received(:new)
          .with(endpoint)
        expect(lago_client).to have_received(:post_with_response) do |payload|
          expect(payload[:restletEndpoint]).to eq('https://example.com')
      end    
```

ivannovosad (approved):

PR #1879, opened 2024-04-16
feat(integrations-mapping): Add integration mapping services

ivannovosad:
## Roadmap Task

## Context

## Description

This Pull request adds services for netsuite integration mappings.

### DIFF
--- a/app/services/integration_collection_mappings/netsuite/create_service.rb
+++ b/app/services/integration_collection_mappings/netsuite/create_service.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module IntegrationCollectionMappings
+  module Netsuite
+    class CreateService < BaseService
+      def call(**args)
+        integration_collection_mapping = IntegrationCollectionMappings::NetsuiteCollectionMapping.new(
+          integration_id: args[:integration_id],
+          mapping_type: args[:mapping_type],
+        )
+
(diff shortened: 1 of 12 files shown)


app/services/integration_collection_mappings/netsuite/create_service.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module IntegrationCollectionMappings
+  module Netsuite
+    class CreateService < BaseService
+      def call(**args)
```
julienbourdeau, line 6: I'm a little surprised why CreateService looks different from UpdateService.
* CreateService: no constructor and a call method with `**args`
* UpdateService: a constructor with a `params:` arg and a call method with no args

Unfortunately there is no real use of the service so I'm not sure how that will be used. It could be consistent with the rest of the codebase but I'm a little surprised looking at this diff only.

julienbourdeau (approved):

app/services/integration_collection_mappings/netsuite/create_service.rb
ivannovosad: The Create and Update services will be used in the following PR which will contain the corresponding GrapQL mutations.

It follows the convention as the existing create and update services. 
We can refactor that so it's more consistent (e.g.: `args`  vs `params`) but it's not in the scope of this pull request.

PR #1880, opened 2024-04-16
POC: Role-based Access Control exploration

julienbourdeau:
## Roadmap Task

## Context

Lago doesn't have any permission system, we're adding a Role-based Access Control (RBAC) brick.

## Description

Proof of concept.


lib/tasks/permissions.rake
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+namespace :permissions do
+  desc 'Generate YAML from CSV'
```
julienbourdeau, line 6: Temporary command until we settle on the final permission definitions.
The CSV file won't make it to the final version either.

PR #1884, opened 2024-04-17
feat(integrations-mapping): Add integration mapping GraphQL mutations

ivannovosad:
## Roadmap Task

## Context

## Description

This Pull request adds GraphQL mutations for netsuite integration and collection mappings.


app/graphql/mutations/integration_collection_mappings/destroy.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      field :id, ID, null: true
+
+      def resolve(id:)
+        integration_collection_mapping = ::IntegrationCollectionMappings::BaseCollectionMapping
+          .joins(:integration)
+          .where(id:)
+          .where(integration: { organization: current_organization }).first
+
+        return not_found_error(resource: 'integration_collection_mapping') unless integration_collection_mapping
```
vincent-pochet, line 22: It looks like this logic is already handled in the destroy service. Any reason for having it here as well?

app/graphql/mutations/integration_mappings/destroy.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      field :id, ID, null: true
+
+      def resolve(id:)
+        integration_mapping = ::IntegrationMappings::BaseMapping
+          .joins(:integration)
+          .where(id:)
+          .where(integration: { organization: current_organization }).first
+
+        return not_found_error(resource: 'integration_mapping') unless integration_mapping
```
vincent-pochet, line 22: Same comment regarding the check for presence in the destroy service

vincent-pochet (approved):

app/graphql/mutations/integration_collection_mappings/netsuite/create.rb
```diff
@@ -0,0 +1,27 @@
...
+      class Create < BaseMutation
+        include AuthenticableApiUser
+        include RequiredOrganization
+
+        graphql_name 'CreateNetsuiteIntegrationCollectionMapping'
+        description 'Create Netsuite integration collection mapping'
+
+        input_object_class Types::IntegrationCollectionMappings::Netsuite::CreateInput
+
+        type Types::IntegrationCollectionMappings::Netsuite::Object
```
lovrocolic, line 15: I am not sure why do we have `netsuite` specific object to return. IMO we should use generic object which can be used for other integrations as well

app/graphql/mutations/integration_collection_mappings/netsuite/update.rb
```diff
@@ -0,0 +1,32 @@
...
+      class Update < BaseMutation
+        include AuthenticableApiUser
+        include RequiredOrganization
+
+        graphql_name 'UpdateNetsuiteIntegrationCollectionMapping'
+        description 'Update Netsuite integration mapping'
+
+        input_object_class Types::IntegrationCollectionMappings::Netsuite::UpdateInput
+
+        type Types::IntegrationCollectionMappings::Netsuite::Object
```
lovrocolic, line 15: Same comment as above

PR #1886, opened 2024-04-18
misc(settings-and-secrets): Add settings and secrets accessors to payment providers and customers

ivannovosad:
## Context

Since we have `SecretsStorable` and `SettingsStorable` we can call methods like `settings_accessors :webhook_secret, :success_redirect_url` in the models so that we don't have to implement each getter and setter.

## Description

This PR includes `SecretsStorable` and `SettingsStorable` to payment providers and payment provider customers base models and then calls `settings_accessors` and `secret_accessors` when needed.


vincent-pochet (approved):
Nice!

PR #1896, opened 2024-04-19
feat(custom_agg): Add Ruby Sandbox to allow ruby execution in a safe environment

vincent-pochet:
## Context

Some of our customers have express a need for aggregation and charge models that does not fit into the current logic in place in Lago, like for example having a single aggregation for 2 different properties, but make the price per unit changing based on the position of the event.

This feature aims to propose a way to build custom aggregation logic for specific cases.

## Description

This PR adds the Ruby Sandbox that will be used to execute the custom aggregation method in a next PR


lovrocolic (approved):

lovrocolic:
🚀

PR #1901, opened 2024-04-22
Add integration items resolver

ivannovosad:
## Context

## Description

This PR adds GraphQL types and resolvers for fetching integration items.

### DIFF
--- a/app/queries/integration_items_query.rb
+++ b/app/queries/integration_items_query.rb
@@ -19,7 +19,7 @@ class IntegrationItemsQuery < BaseQuery
   attr_reader :search_term
 
   def base_scope
-    IntegrationItem.ransack(search_params)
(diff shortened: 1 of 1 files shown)


app/queries/integration_items_query.rb
```diff
@@ -0,0 +1,35 @@
...
+    result.integration_items = integration_items
+    result
+  end
+
+  private
+
+  attr_reader :search_term
+
+  def base_scope
+    IntegrationItem.ransack(search_params)
```
lovrocolic, line 22: We have to include organization in the query as well. Also, I would prefer having `integration_id` here as well, so that in the main method we can only have filters + pagination

ivannovosad: I added the organization.

lovrocolic (approved):

PR #1903, opened 2024-04-22
feat(wallet): Set transaction status to wallet transaction

rsempe:
## Context

We want to differentiate a free wallet transaction vs. a paid wallet transaction.
And be able to void some wallet transactions.

## Description

This PR aims to add a `transaction_status` field for wallet transactions.
This field can be set to `paid`, `offered`, or `voided` (not used yet).

### DIFF
--- a/app/controllers/api/v1/wallet_transactions_controller.rb
+++ b/app/controllers/api/v1/wallet_transactions_controller.rb
@@ -4,8 +4,7 @@ module Api
   module V1
     class WalletTransactionsController < Api::BaseController
       def create
-        service = WalletTransactions::CreateService.new
-        result = service.create(
+        result = WalletTransactions::CreateService.new.call(
           organization_id: current_organization.id,
           wallet_id: input_params[:wallet_id],
           paid_credits: input_params[:paid_credits],
--- a/app/graphql/mutations/wallet_transactions/create.rb
+++ b/app/graphql/mutations/wallet_transactions/create.rb
@@ -16,7 +16,7 @@ module Mutations
       type Types::WalletTransactions::Object.collection_type
 
       def resolve(**args)
-        result = ::WalletTransactions::CreateService.new.create(
+        result = ::WalletTransactions::CreateService.new(context[:current_user]).call(
           organization_id: current_organization.id,
           wallet_id: args[:wallet_id],
           paid_credits: args[:paid_credits],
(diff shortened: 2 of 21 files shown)


app/controllers/api/v1/wallet_transactions_controller.rb
```diff
@@ -4,8 +4,7 @@ module Api
   module V1
     class WalletTransactionsController < Api::BaseController
       def create
-        service = WalletTransactions::CreateService.new
-        result = service.create(
+        result = WalletTransactions::CreateService.new.call(
```
ivannovosad, line 7: Can we shorten it to just:

```suggestion
        result = WalletTransactions::CreateService.call(
```

app/graphql/mutations/wallet_transactions/create.rb
```diff
@@ -16,7 +16,7 @@ class Create < BaseMutation
       type Types::WalletTransactions::Object.collection_type
 
       def resolve(**args)
-        result = ::WalletTransactions::CreateService.new.create(
+        result = ::WalletTransactions::CreateService.new(context[:current_user]).call(
```
ivannovosad, line 19: Same comment as above: 
```suggestion
        result = ::WalletTransactions::CreateService.call(context[:current_user]
```

spec/serializers/v1/wallet_transaction_serializer_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      expect(result['wallet_transaction']).to include(
+        'lago_id' => wallet_transaction.id,
+        'lago_wallet_id' => wallet_transaction.wallet_id,
+        'status' => wallet_transaction.status,
+        'transaction_status' => wallet_transaction.transaction_status,
+        'transaction_type' => wallet_transaction.transaction_type,
+        'amount' => wallet_transaction.amount.to_s,
+        'credit_amount' => wallet_transaction.credit_amount.to_s,
+        'settled_at' => wallet_transaction.settled_at&.iso8601,
+        'created_at' => wallet_transaction.created_at&.iso8601,
```
ivannovosad, line 25:
```suggestion
        'created_at' => wallet_transaction.created_at.iso8601,
```

ivannovosad (changes requested):

app/controllers/api/v1/wallet_transactions_controller.rb
rsempe: Yes, that's the plan, but it requires several changes due to the fact that we have to refactor `WalletTransactions::ValidateService`.

So if that's ok for you, will do it in a future PR ;)

ivannovosad (approved):

PR #1912, opened 2024-04-23
feat(gql): expose Memberships to UserType (to retrieve upcoming permissions)

julienbourdeau:
## Context

Role-based permissions are coming to Lago.

## Description

This PR introduces an proper RoleEnum for `membership.role` attribute. Today there is only admin but more roles are coming.

This is preliminary work to help the frontend retrieve permissions and role for the current user for each organization through memberships. The `organizations` relation is still available for backward compatibility reasons.

### Before

```gql
fragment CurrentUserInfos on User {
  id
  email
  premium
  organizations {
    id
    name
    logoUrl
  }
}
```

### After

```gql
fragment CurrentUserInfos on User {
  id
  email
  premium
  memberships {
    id
    role
    permissions.  <--- upcoming
    organization {
      id
      name
      logoUrl
    }
  }
}
```

### DIFF
--- a/app/graphql/resolvers/memberships_resolver.rb
+++ b/app/graphql/resolvers/memberships_resolver.rb
@@ -15,6 +15,7 @@ module Resolvers
     def resolve(page: nil, limit: nil)
       current_organization
         .memberships
+        .includes(:user)
         .active
         .page(page)
         .per(limit)
--- a/app/graphql/types/membership_type.rb
+++ b/app/graphql/types/membership_type.rb
@@ -2,13 +2,17 @@
 
 module Types
   class MembershipType < Types::BaseObject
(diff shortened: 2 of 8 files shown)


app/graphql/resolvers/memberships_resolver.rb
```diff
@@ -15,6 +15,7 @@ class MembershipsResolver < GraphQL::Schema::Resolver
     def resolve(page: nil, limit: nil)
       current_organization
         .memberships
+        .includes(:user)
```
julienbourdeau, line 18: https://github.com/getlago/lago-front/blob/7486e2849511b771bc43b6f9d392c298ed6140e2/src/components/settings/members/MembershipItem.tsx#L15-L20

PR #1919, opened 2024-04-23
refactor(gql): Refactor Organization types

julienbourdeau:
## Description

We recently started using `SafeOrganizationType` to expose organization fields in relationships. I believe we should default to this type. For example [here](https://github.com/getlago/lago-api/blob/91a1e85d99a3d5d0a9ee748424fe44b07b9ffc13/app/graphql/types/billable_metrics/object.rb#L10), [here](https://github.com/getlago/lago-api/blob/91a1e85d99a3d5d0a9ee748424fe44b07b9ffc13/app/graphql/types/add_ons/object.rb#L9) or [here](https://github.com/getlago/lago-api/blob/91a1e85d99a3d5d0a9ee748424fe44b07b9ffc13/app/graphql/types/plans/object.rb#L9).

Instead of changing all types of `field :organization`, I propose to rename `SafeOrganizationType` to `Organization` and use a new `CurrentOrganization` type to expose all fields of the organization.


vincent-pochet (approved):
Nice!

PR #1926, opened 2024-04-24
feat(rbac): Introduce permission system

julienbourdeau:
Currently, all Lago users are admin. They can perform all actions and access every piece of data.

This PR introduces the base needed to add permissions all over the codebase.

## Overview

- **users** have roles per **organization**, hence **membership** have `roles`
- **roles** have `permissions`
- app always checks for `permissions` (as much as possible), not `roles`

### 3 roles, no custom roles

Lago defines 3 roles by default

* Admin
* Account Manager
* Finance & Analyst

These roles exist for every organization. It's not possible to create custom roles. It's not possible to customize the permissions of these roles.

### Examples

```ruby
# ❌ Bad
user.has_role? :admin, organization
user.admin? organization
user.manager? organization

# ✅ Good
user.has_permission? "organization:info:edit", organization
user.can? "organization:info:edit", organization
```

## Implementation

* The role and their perimissions are defined in yaml (`app/config/permissions/`)
  * `definition.yml` defines ALL permissions and their default value
  * `role-finance.yml` and `role-manager.yml` overrides the default permissions for their respective roles

Each roles has a list of permission associated stored in a hash. The hash is instantiated and frozen when the app is booted. Checking permission is simply checking the value in a flat hash.

```ruby
# Checking `addons:create` for account manager.
MANAGER_PERMISSIONS_HASH['addons:create']
```

Permissions are *per user per organization*, hence they are part of the Membership model. Usually, it's called directly from users. If you're in a loop, you can eager load `user.memberships`.

```ruby
user.can? 'addons:create', organization: org
```

### GraphQL

The permissions **of the current user on the current organization** are passed in the GraphQL context. We don't need to call `user.can?`, we can directly check the `context['permissions']` hash.

⚠️ If there is no current organization, the permissions hash will be empty.

## Usage

In this step, the permissions are only added to the GraphQL. The REST API will come late when we introduce api key scopes.

### Protecting a Type field when retrieving data

Every used can retrieve the current organization but some fields will require a some permissions. **If you request a field you don't have access to, you'll simply get a null value, not an error.**

```diff
module Types
  class CurrentOrganizationType < Types::BaseObject
    description 'Current Organization Type'

    field :id, ID, null: false
    field :name, String, null: false

    ...

-    field :api_key, String
+    field :api_key, String, permission: 'developer:view'

-    field :webhook_url, String
+    field :webhook_url, String, permissions: ['developer:view', 'something:else']

    ...
  end
end
```

### Protecting a mutation

If the entire mutation requires a role to be executed, you can define the required permission in a constant.

```diff
module Mutations
  module AddOns
    class Create < BaseMutation
      include AuthenticableApiUser
      include RequiredOrganization

+      REQUIRED_PERMISSION = 'addons:create'

      graphql_name 'CreateAddOn'
      description 'Creates a new add-on'

      input_object_class Types::AddOns::CreateInput
      type Types::AddOns::Object

      def resolve(**args)
        ...
      end
    end
  end
end
```

### Protecting only a field

In some situation, a user can update some fields of a type but not all of them. They must be able to access the mutation so the previous solution doesn't work. In this case, we rely on the input object to restrict the list of arguments.

In this case, the app will loop over the arguments and remove the arguments doesn't have the required permission.

```diff
module Types
  module AddOns
    class UpdateInput < Types::BaseInputObject
      graphql_name 'UpdateAddOnInput'

      argument :id, ID, required: true
      argument :description, String, required: false
      ... 
-     argument :tax_codes, [String], required: false
+     argument :tax_codes, [String], required: false, permission: 'addons:update:tax'
    end
  end
end
```

### DIFF
--- a/app/config/permissions/definition.yml
+++ b/app/config/permissions/definition.yml
@@ -1,0 +1,84 @@
+---
+analytics:
+  view:
+billable_metrics:
+  view:
+  create:
+  update:
+  delete:
+plans:
+  view: true
+  create:
+  update:
+  delete:
+addons:
+  view: true
+  create:
+  update:
+  delete:
+coupons:
+  view: true
+  create:
+  update:
+  delete:
+  attach:
+  detach:
+customers:
+  view: true
+  create:
+  update:
+  delete:
+subscriptions:
+  create:
+  update:
+  delete:
+wallets:
+  create:
+  update:
+  top_up:
+  terminate:
+invoices:
+  view: true
+  send: true
+  update: true
+  void: true
+  create:
+draft_invoices:
+  update: true
+credit_notes:
+  view: true
+  create: true
+  void: true
+customer_settings:
+  view: true
+  edit:
+    tax_rates: true
+    payment_terms: true
+    grace_period: true
+    lang: true
+developers:
+  view:
+  keys:
+    view:
+organization:
+  view:
+  edit:
+  invoices:
+    view:
+    edit:
+  taxes:
+    view:
+    edit:
+  emails:
+    view:
+    edit:
+  integrations:
+    view:
+    create:
+    update:
+    delete:
+  members:
+    view:
+    create:
+    update:
+    delete:
--- a/app/config/permissions/role-finance.yml
+++ b/app/config/permissions/role-finance.yml
@@ -1,0 +1,64 @@
+
+---
+analytics:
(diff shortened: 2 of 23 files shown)


vincent-pochet (approved):
:rocket:

PR #1948, opened 2024-04-26
feat (integration-items): Clear integration items before refetch

lovrocolic:
## Context

## Description

This PR updates existing logic to re-fetch integration items.

Before each re-fetch we need to destroy all existing records

### DIFF
--- a/app/graphql/mutations/integration_items/fetch_items.rb
+++ b/app/graphql/mutations/integration_items/fetch_items.rb
@@ -20,6 +20,8 @@ module Mutations
 
         ::Integrations::Aggregator::SyncService.call(integration:)
 
+        integration.integration_items.where(item_type: :standard).destroy_all
+
         result = ::Integrations::Aggregator::ItemsService.call(integration:)
 
         result.success? ? result.items : result_error(result)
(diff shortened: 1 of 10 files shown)


app/graphql/mutations/integration_items/fetch_items.rb
```diff
@@ -20,6 +20,8 @@ def resolve(**args)
 
         ::Integrations::Aggregator::SyncService.call(integration:)
 
+        integration.integration_items.where(item_type: :standard).destroy_all
```
vincent-pochet, line 23: Should it be part of the sync service?

app/graphql/mutations/integration_items/fetch_tax_items.rb
```diff
@@ -20,6 +20,8 @@ def resolve(**args)
 
         ::Integrations::Aggregator::SyncService.call(integration:)
 
+        integration.integration_items.where(item_type: :tax).destroy_all
```
vincent-pochet, line 23: Same question, as it feels weird to have "business logic" in the mutation

app/graphql/mutations/integration_items/fetch_items.rb
lovrocolic: Actually sync service should just trigger various syncs on Nango side. I moved this line to the items service where items are actually fetched

app/graphql/mutations/integration_items/fetch_tax_items.rb
lovrocolic: Moved it to the items service

vincent-pochet (approved):

PR #1949, opened 2024-04-26
feat(integrations): Add contact service

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds service for fetch a contact by ID from Nango.


ivannovosad:
After a conversation with @sarkissianraffi we concluded that this service is no longer needed.
The subsidiary_id is not required anymore when updating Netsuite contacts.

PR #1941, opened 2024-04-25
feat(rbac): Add permissions to mutations

julienbourdeau:
This is a followup of this beautiful PR: #1926 

Adding permissions to Mutations

- used `organization:integrations:*` for everything payment providers
- used `invoices:update` for Adjusted Fees create/delete

PR #1952, opened 2024-04-29
feat(graphql): Rework GraphQl concerns

julienbourdeau:
## Context

Granular permissions are being added to Lago

## Description

A first PR added all the permission system foundations to Lago (https://github.com/getlago/lago-api/pull/1926). A second PR added permissions to all GraphQL mutations (https://github.com/getlago/lago-api/pull/1941).

- As I'm working on adding it to Resolvers, I realized I needed the exact the thing as mutations so I extracted some logic into `CanRequirePermissions` concern.
- Not all resolvers were extending our custom BaseResolvers, now they do.
- As I was testing my new concern, I added tests to `AuthenticableApiUser` and `RequiredOrganization` concerns

When I was editing tests for mutations, I add to edit MANY tests re-testing _"without current user"_ and _"without current organization"_. I believe those tests are redundant and we can now simply tests of the Mutation/Resolver includes the correct concern. The behavior is tested once in the concern. See e48109da for single example.

Recently @josemoyab introduced [a great refactor](https://github.com/getlago/lago-api/pull/1853) of `RequiredOrganization` to avoid calling `validate_organization!` in every `resolve` method. I'm taking this one step further by removing the meta programming and using the `ready?` method to be consistent with other concerns.

I'll deal with `AuthenticableCustomerPortalUser` concern separately as this PR is already huge and it doesn't affect my work on RBAC.

### About execution order

Each `ready?` method will be executed in order, starting by the last concern added, all the way to the first concern added.

```ruby
class CreateThing < BaseMutation
    include AuthenticableApiUser
    include RequiredOrganization

    ...
end
```

In the above snippet, we'll call `RequiredOrganization#ready?` first, then `AuthenticableApiUser#ready?`. In this case, you'll never get to the `AuthenticableApiUser` error because `RequiredOrganization#ready?` also checks that the current_user is part of the current_organization. Without a current_user, this check will never pass.

**This is the reason why I had to edit all mutation tests in the PR**. Previously, the `AuthenticableApiUser#ready?` would run first, then `resolve` would automatically call `current_organization!`.

### Nota Bene

We don't need to add dedicated tests case for current_user and curent_organization in every new mutation or resolver.

### DIFF
--- a/app/controllers/graphql_controller.rb
+++ b/app/controllers/graphql_controller.rb
@@ -26,7 +26,9 @@ class GraphqlController < ApplicationController
       current_organization:,
       customer_portal_user:,
       request:,
-      permissions: current_user&.memberships&.find_by(organization: current_organization)&.permissions_hash || {},
+      permissions:
+        current_user&.memberships&.find_by(organization: current_organization)&.permissions_hash ||
+        Permission::EMPTY_PERMISSIONS_HASH,
     }
     result = LagoApiSchema.execute(query, variables:, context:, operation_name:)
     render(json: result)
--- a/app/graphql/concerns/can_require_permissions.rb
+++ b/app/graphql/concerns/can_require_permissions.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module CanRequirePermissions
+  extend ActiveSupport::Concern
+
+  private
+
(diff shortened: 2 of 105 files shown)


vincent-pochet (approved):
:rocket:

PR #1958, opened 2024-04-29
feat(okta): add accept invite logic

jdenquin:
## Context

- We want to be able to signup with an invite by using Okta.

## Description

- Add logic for Okta and accept invite

### DIFF
--- a/app/services/auth/okta/accept_invite_service.rb
+++ b/app/services/auth/okta/accept_invite_service.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module Auth
+  module Okta
+    class AcceptInviteService < BaseService
+      def initialize(invite_token:, code:, state:)
+        @invite_token = invite_token
+        @code = code
+        @state = state
+
+        super
+      end
+
+      def call
+        check_state
+        check_okta_integration(result.email)
+        check_invite(result.email)
+        query_okta_access_token
+        check_userinfo(result.email)
+
+        Invites::AcceptService.new.call(
+          invite: result.invite,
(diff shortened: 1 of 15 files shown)


app/services/auth/okta/accept_invite_service.rb
```diff
@@ -0,0 +1,39 @@
...
+        query_okta_access_token
+        check_userinfo(result.email)
+
+        Invites::AcceptService.new.call(
+          invite: result.invite,
+          email: result.email,
+          token: invite_token,
+          password: SecureRandom.hex,
+        )
+      rescue ValidationError => e
```
lovrocolic, line 27: As I can see, we should also rescue `Auth::Okta::ValidationError`

julienbourdeau (approved):

app/services/auth/okta/accept_invite_service.rb
julienbourdeau: Good point. I would expect that it would rescue `Auth::Okta::ValidationError` here since we're in `Auth::Okta` module 🤔

jdenquin: exactly the same @julienbourdeau since we are in the right module it should call by default `Auth::Okta::ValidationError`

lovrocolic: I see now. In that case I guess `check_state` should just raise `ValidationError`?

jdenquin: yes good catch

lovrocolic (approved):

PR #1967, opened 2024-04-30
feat(integrations): Add netsuite_mappable to addons and billable metrics

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds `netsuite_mappable` to addons and billable metrics.

### DIFF
--- a/app/graphql/types/add_ons/object.rb
+++ b/app/graphql/types/add_ons/object.rb
@@ -25,6 +25,8 @@ module Types
 
       field :taxes, [Types::Taxes::Object]
 
+      field :integration_mappings, [Types::IntegrationMappings::Netsuite::Object], null: true
+
       def customers_count
         object.applied_add_ons.select(:customer_id).distinct.count
       end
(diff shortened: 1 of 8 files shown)


app/graphql/types/add_ons/object.rb
```diff
@@ -25,6 +25,8 @@ class Object < Types::BaseObject
 
       field :taxes, [Types::Taxes::Object]
 
+      field :netsuite_mapping, Types::IntegrationMappings::Netsuite::Object, null: true
```
lovrocolic, line 28: I have few comments here:)
1.) can we please introduce integration_mapping instead of netsuite_mapping (I am doing it also in another branch)
2.) can you explain why one add_on will have only one integration_mapping? There can be many connections so I will say that it should be `has_many` relation

app/models/add_on.rb
```diff
@@ -16,6 +16,8 @@ class AddOn < ApplicationRecord
   has_many :applied_taxes, class_name: 'AddOn::AppliedTax', dependent: :destroy
   has_many :taxes, through: :applied_taxes
 
+  has_one :netsuite_mapping, class_name: 'IntegrationMappings::NetsuiteMapping', as: :mappable
```
lovrocolic, line 19: Should it be `has_many`?

PR #1970, opened 2024-04-30
feat(custom_agg): Add a new scenario and fix aggregation properties

vincent-pochet:
## Context

Some of our customers have express a need for aggregation and charge models that does not fit into the current logic in place in Lago, like for example having a single aggregation for 2 different properties, but make the price per unit changing based on the position of the event.

This feature aims to propose a way to build custom aggregation logic for specific cases.

## Description

This PR add a new scenario for a complex custom aggregation involving filters.
It also makes sure that the applied aggregation properties are the one from the filter

### DIFF
--- a/app/services/billable_metrics/aggregations/custom_service.rb
+++ b/app/services/billable_metrics/aggregations/custom_service.rb
@@ -74,6 +74,8 @@ module BillableMetrics
       private
 
       def custom_properties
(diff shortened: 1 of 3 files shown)


julienbourdeau (approved):
Even GitHub wants to hide the diff :D

PR #1971, opened 2024-04-30
fix (accounting-integrations): fix namespaces for integration mapping

lovrocolic:
## Context

## Description

For Integration mappings, we should have similar structure across all current and future integrations since mapping data is not something that should be affected by the end integration.

That being said, this PR fixes all namespaces that include `netsuite` for intagration mappings


app/queries/integration_mappings/mappings_query.rb
```diff
@@ -1,15 +1,15 @@
 # frozen_string_literal: true
 
 module IntegrationMappings
-  class NetsuiteMappingsQuery < BaseQuery
+  class MappingsQuery < BaseQuery
```
ivannovosad, line 4: What about moving it up one level - similar to integration_integration_items_query:

`class IntegrationMappingsQuery`

app/graphql/resolvers/integration_mappings/mappings_resolver.rb
```diff
@@ -2,7 +2,7 @@
 
 module Resolvers
   module IntegrationMappings
-    class NetsuiteMappingsResolver < Resolvers::BaseResolver
+    class MappingsResolver < Resolvers::BaseResolver
```
ivannovosad, line 5: Similar comment as the one for query - what about having just:

```
module Resolvers
  class IntegrationMappingsResolver
  ...
```

(and also `IntegrationMappingResolver`)

ivannovosad (changes requested):

ivannovosad (approved):

PR #1975, opened 2024-05-01
misc(indices): Add indices to several database tables

ivannovosad:
## Context

There were performance issues when selecting from multiple database tables while generating invoices.

## Description

In this PR we're adding indices to multiple DB tables.


vincent-pochet:
Enum fields have very low cardinality, so I think that adding indexes only on these field will not bring a huge performance boost.
I would suggest instead to check for slow queries and to try to add indexes on multiple fields, involving the enum and probably the `deleted_at` columns.

Let's talk about it if you want :)

ivannovosad:
@vincent-pochet 
Ok, I'm closing this one.

PR #1978, opened 2024-05-02
feat(rbac): Add premissions on Organization (resolver and mutation)

julienbourdeau:
## Context

Granular permissions are being added to Lago.

## Description

Following #1926,  #1941, #1952 and #1969, this PR add permissions restriction to Organization-related endpoints.

For organization, both mutations and resolvers are accessible without permissions but many fields/arguments are restricted. 

**Line length: Remember that soon, with StandardRb, there is no line length restrictions** 😬

### DIFF
--- a/app/graphql/types/current_organization_type.rb
+++ b/app/graphql/types/current_organization_type.rb
@@ -7,25 +7,25 @@ module Types
     field :id, ID, null: false
     field :logo_url, String
     field :name, String, null: false
-    field :timezone, Types::TimezoneEnum, null: true
+    field :timezone, Types::TimezoneEnum
 
     field :default_currency, Types::CurrencyEnum, null: false
     field :email, String
 
     field :legal_name, String
     field :legal_number, String
-    field :tax_identification_number, String, null: true
+    field :tax_identification_number, String, permission: 'organization:taxes:view'
 
     field :address_line1, String
     field :address_line2, String
     field :city, String
-    field :country, Types::CountryCodeEnum, null: true
+    field :country, Types::CountryCodeEnum
     field :net_payment_term, Integer, null: false
     field :state, String
     field :zipcode, String
 
-    field :api_key, String, null: false
-    field :webhook_url, String
+    field :api_key, String, permission: 'developers:keys:manage'
+    field :webhook_url, String, permission: 'developers:manage'
 
     field :document_number_prefix, String, null: false
     field :document_numbering, Types::Organizations::DocumentNumberingEnum, null: false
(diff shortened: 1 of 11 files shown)


app/graphql/types/current_organization_type.rb
```diff
@@ -39,13 +39,13 @@ class CurrentOrganizationType < Types::BaseObject
...
-    field :taxes, [Types::Taxes::Object], null: true, resolver: Resolvers::TaxesResolver
+    field :billing_configuration, Types::Organizations::BillingConfiguration, permission: 'organization:invoices:view'
+    field :email_settings, [Types::Organizations::EmailSettingsEnum], permission: 'organization:emails:view'
+    field :taxes, [Types::Taxes::Object], resolver: Resolvers::TaxesResolver, permission: 'organization:taxes:view'
 
-    field :adyen_payment_providers, [Types::PaymentProviders::Adyen], null: true
-    field :gocardless_payment_providers, [Types::PaymentProviders::Gocardless], null: true
-    field :stripe_payment_providers, [Types::PaymentProviders::Stripe], null: true
+    field :adyen_payment_providers, [Types::PaymentProviders::Adyen], permission: 'organization:integrations:view'
+    field :gocardless_payment_providers, [Types::PaymentProviders::Gocardless], permission: 'organization:integrations:view' # rubocop:disable Layout/LineLength
```
vincent-pochet, line 47: Why not putting the definition on multiple lines?

app/graphql/types/organizations/update_organization_input.rb
```diff
@@ -29,8 +29,8 @@ class UpdateOrganizationInput < BaseInputObject
       argument :document_number_prefix, String, required: false
       argument :document_numbering, Types::Organizations::DocumentNumberingEnum, required: false
 
-      argument :billing_configuration, Types::Organizations::BillingConfigurationInput, required: false
-      argument :email_settings, [Types::Organizations::EmailSettingsEnum], required: false
+      argument :billing_configuration, Types::Organizations::BillingConfigurationInput, required: false, permission: 'organization:invoices:view' # rubocop:disable Layout/LineLength
```
vincent-pochet, line 32: Same question regarding line length

vincent-pochet (approved):

app/graphql/types/organizations/update_organization_input.rb
julienbourdeau: Here it's a longer, it could be multiline but I prefer to keep the `argument` flow 🙏

vincent-pochet: Well, I think we should avoid using `rubocop:disable` as much as possible and I don't think it's an issue to have multiple lines here.

That said, not a blocker for me :)

julienbourdeau: Yes, I agree it should be some sort of last resort. If we didn't decide to move to standardrb, I think I'd have broken the line down.

PR #1982, opened 2024-05-03
misc: Setup StandardRB rules by default

rsempe:
The goal of this PR is to setup by default the [StandardRB base rules](https://github.com/standardrb/standard/blob/main/config/base.yml).


julienbourdeau (approved):
Super exited to run `rubocop` automatically in IntelliJ 🚀

vincent-pochet (approved):

PR #1984, opened 2024-05-03
feat(integrations): Create customer

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds services for creating and updating integration customers. These services will be used by GraphQL mutations and API controllers.

### DIFF
--- a/app/services/integration_customers/update_service.rb
+++ b/app/services/integration_customers/update_service.rb
@@ -13,13 +13,13 @@ module IntegrationCustomers
 
       return result.not_found_failure!(resource: 'integration_customer') unless integration_customer
 
-      if sync_with_provider
+      if external_customer_id.present?
+        integration_customer.update!(external_customer_id:)
+      elsif sync_with_provider
         update_result = Integrations::Aggregator::Contacts::UpdateService.call(integration:, integration_customer:)
         integration_customer.update!(subsidiary_id:)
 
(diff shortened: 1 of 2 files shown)


app/services/integration_customers/create_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    def initialize(params:, integration:, customer:)
+      @customer = customer
+      super(params:, integration:)
+    end
+
+    def call
+      result = super
+      return result if result.error
+
+      res = if sync_with_provider
```
lovrocolic, line 14: What should be the flow if `sync_with_provider` is true and `external_customer_id` is given?
In this case, I think we should just use `external_customer_id` and create integration customer on Lago side.
That being said, probably we should have:
```
res = if external_customer_id.present?
    link_customer!
elsif sync_with_provider
    sync_customer!
end
```

app/services/integration_customers/update_service.rb
```diff
@@ -0,0 +1,35 @@
...
+      super(params:, integration:)
+    end
+
+    def call
+      result = super
+      return result if result.error
+
+      return result.not_found_failure!(resource: 'integration_customer') unless integration_customer
+
+      if sync_with_provider
```
lovrocolic, line 16: same comment as in create service

lovrocolic (approved):

PR #1985, opened 2024-05-03
fix(stripe): Handle IdempotencyError when creating customer

vincent-pochet:
## Context

This PR is an attempt to handle `Stripe::IdempotencyError` happening when a customer creation via API failed and does not properly save the `provider_customer_id` on the `PayementProviderCustoner::StripCustomer` instance.

The issue appears in some rare case of race conditions, but it puts the customer in an instable state, as it cannot be synced again on Stripe due to idempotency protection.

## Description

In case of `Stripe::IdempotencyError`, the service not tries to fetch the StripeCustomer from the API using the customer email. If only one matches the email, it is returned by the service and its ID is saved in the `provider_customer_id` field.


julienbourdeau (approved):
TIL: `idempotency key` request param 👌

PR #1987, opened 2024-05-06
misc: Stop using quantified events

vincent-pochet:
## Context

The `QuantifiedEvent` model was initially introduced  to cache data for unique count aggregation.
This aggregation type was refactored to rely only on the event table.

In parallel, a model was added to store cached aggregated data for in advance charges/aggregation, the `CachedAggregation`

Today, the weighted sum aggregation, when used in recurring mode is using the `QuantifiedEvent` model to keep the aggregation state from a period to an other without recomputing the aggregation from zero.

Since `CachedAggregation` and `QuantifiedEvent` share the same purpose, and since `QuantifiedEvent` remains only for weighted sum, it was decided to merge the two model, in order to remove the `QuantifiedEvent` model in a future release.

## Description

This PR, changes all logic relying on `QuantifiedEvent`  in favor of `CachedAggregation`


rsempe (approved):
🙌 🙌

PR #1990, opened 2024-05-06
feat(accounting-integrations): add strategy service for creating or updating integration customers

lovrocolic:
## Context

## Description

This PR handles strategy service for accounting customers.

This service will be called from `customers/create_service.rb` and `customers/update_service.rb`.

The main goal of this service is to determine if we should:
- skip syncing accounting customer
- create integration customer
- update integration customer (remove one)

Based on decision we call appropriate job so that the whole logic of accounting customer is extracted and can be retried. Accounting customer does not need to be in the same DB transaction and the main goal was that the whole process which handles accounting customer does not add extra complexity in customer create/update services.

### DIFF
--- a/app/jobs/integration_customers/create_job.rb
+++ b/app/jobs/integration_customers/create_job.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module IntegrationCustomers
+  class CreateJob < ApplicationJob
+    queue_as 'integrations'
+
+    retry_on LagoHttpClient::HttpError, wait: :exponentially_longer, attempts: 3
+
+    def perform(integration_customer_params:, integration:, customer:)
+      result = IntegrationCustomers::CreateService.call(params: integration_customer_params, integration:, customer:)
+      result.raise_if_error!
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


app/jobs/integration_customers/create_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module IntegrationCustomers
+  class CreateJob < ApplicationJob
+    queue_as 'integrations'
+
+    retry_on LagoHttpClient::HttpError, wait: :exponentially_longer, attempts: 3
+
+    def perform(integration_customer_params:, integration:, customer:)
+      result = IntegrationCustomers::CreateService.call(params: integration_customer_params, integration:, customer:)
```
julienbourdeau, line 10: I'm surprised to see `CreateJob` and `UpdateJob` call `CreateService` and `UpdateService` but no reference to the new `CreateOrUpdateService` 🤔

julienbourdeau (approved):

app/jobs/integration_customers/create_job.rb
lovrocolic: That was the goal:) Because of it, new logic can be merged and we will just open new PR later with the tiny change where `CreateOrUpdate` is actually called. That approach allowed us not to carry many branches. Instead all new logic got extracted and merged and later we will just call new logic where needed.

ivannovosad (approved):

PR #1993, opened 2024-05-07
feat(webhook): Break down SendWebhookJob into 2 jobs (one creates the models, the other sends HTTP)

julienbourdeau:
## Description

Last year we added the possibility to add multiple webhook endpoints. It means currently one `SendWebhookJob` can make up to 5 HTTP calls.

This PR break down the `Webhook::BaseService`. Before the service was responsible for genereating the `webhook` model (generating payload, attach to WebhookEndpoint...) AND making the HTTP calls.
Now, the service create a model and dispatch a new job using this new webhook model. Normally, the HTTP part is where the error happen and the retrying wasn't too obvious. We had to carry the options and the object which we didn't need since the webhook model was already saved

Unintended benefit: I also had to refactor the test and make them much smaller. **Notice that it's a separate commit to help with review.**

Other things I'd like to do with webhook:
- Add `retrying` status. I believe `failed` and `retrying` are different.
- Make application a proper relationship `through:` instead of a `delegate`
- A way to subscribe to webhook (see next point)
- Remove `organization.webhook_endpoint` column
- Move `WEBHOOK_SERVICES` hash from job to Webhook model

### Why now?

I was originally looking at how we can let users "subscribe" to only a subset of webhooks. This would help our customers receive less calls and us make way less http requests. With the current code, the retry strategy wasn't too obvious. This PR makes the next step very easy

```diff
current_organization.webhook_endpoints.each do |webhook_endpoint|
+ next unless webhook_endpoint.subscribed_to webhook_type
  webhook = create_webhook(webhook_endpoint, payload)
  SendHttpWebhookJob.perform_later(webhook)
end
```

### DIFF
--- a/app/jobs/send_http_webhook_job.rb
+++ b/app/jobs/send_http_webhook_job.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class SendHttpWebhookJob < ApplicationJob
+  def perform(webhook)
+    Webhooks::SendHttpService.call(webhook:)
+  end
+end
--- a/app/jobs/send_webhook_job.rb
+++ b/app/jobs/send_webhook_job.rb
@@ -43,6 +43,13 @@ class SendWebhookJob < ApplicationJob
   def perform(webhook_type, object, options = {}, webhook_id = nil)
     raise(NotImplementedError) unless WEBHOOK_SERVICES.include?(webhook_type)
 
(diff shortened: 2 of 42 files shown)


vincent-pochet (approved):
:+1:

julienbourdeau:
Let's merge and deploy Monday 05/20.

PR #1994, opened 2024-05-07
misc: Enable argument alignment on Rubocop

rsempe:
As we want to migrate to StandardRB, the goal of this PR is to enable argument alignment on Rubocop.


julienbourdeau (approved):
Hide whitespaces and there is nothing to review 🤣

CleanShot 2024-05-07 at 10 28 04@2x

PR #1997, opened 2024-05-07
fix(events): update event subscription method

lovrocolic:
## Context

Sometimes we receive `Error: NoMethodError: undefined method 'plan' for nil` error in `Fees::CreatePayInAdvanceJob`

We need to update `subscription` method in `event` model.

## Description

If the precision of `terminated_at` is higher than seconds which are used in condition (e.g., milliseconds), it could result in a discrepancy where the truncated timestamp used in the condition might not match exactly with the provided timestamp.

Based on investigation and various tests, it seems that using `millisecond` precision solves the issue

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -50,9 +50,9 @@ class Event < EventsRecord
     end
 
     scope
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):
:rocket: 
Good catch!

PR #2002, opened 2024-05-13
fix(integration-customer): Fix create service

ivannovosad:
## Roadmap Task

## Context

## Description

This PR fixes create service for integration customers when the params are empty.

### DIFF
--- a/app/services/customers/update_service.rb
+++ b/app/services/customers/update_service.rb
@@ -16,6 +16,8 @@ module Customers
       end
 
       old_payment_provider = customer.payment_provider
+      # TODO:
+      # old_provider_customer = customer.provider_customer
       ActiveRecord::Base.transaction do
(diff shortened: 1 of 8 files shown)


vincent-pochet (approved):

app/services/customers/update_service.rb
```diff
@@ -75,7 +76,14 @@ def update(**args)
       customer.external_id = args[:external_id] if customer.editable? && args.key?(:external_id)
 
       ActiveRecord::Base.transaction do
+        # TODO:
+        # if args[:payment_provider].nil? && args[:payment_provider_code].present?
```
lovrocolic, line 80: I would just add one extra condition: `if old_provider_customer && args[:payment_provider].nil? && args[:payment_provider_code].present?`, to prevent UndefinedMethod on nil class errors, just in case

lovrocolic (approved):

PR #2008, opened 2024-05-14
feat(integrations): Add integration resources models

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds integration resource models, factories and specs


app/models/invoice.rb
```diff
@@ -11,6 +11,10 @@ class Invoice < ApplicationRecord
 
   before_save :ensure_organization_sequential_id, if: -> { organization.per_organization? }
```
lovrocolic, line 12: Can we please move the code in the invoice model (everything except has_many definition) into separate PR that would be in the draft status until we ship and QA the feature? I would like to avoid calling any new code before the real QA of the feature. Probably we can continue using workflow as before where we would merge only dead code which is not called in any way :)

app/models/invoice.rb
```diff
@@ -11,6 +11,10 @@ class Invoice < ApplicationRecord
...
   before_save :ensure_number
+  after_create :trigger_invoice_sync, if: :finalized?
+  after_update :trigger_invoice_sync, if: -> { status_updated_to_finalized? }
+  after_create :trigger_sales_order_sync, if: :finalized?
+  after_update :trigger_sales_order_sync, if: -> { status_updated_to_finalized? }
```
lovrocolic, line 17: I think that we can use just `after_save` here and use the same check that we have for setting invoice number (`status_changed_to_finalized?`). Invoice number is also set only when the invoice is finalized

ivannovosad: The check was not working in after_save:

```
rspec ./spec/models/invoice_spec.rb:327 # Invoice#save with an existing record when status changed to finalized with integration customer when sync invoices is true when sync sales orders is true enqueues invoice create job
rspec ./spec/models/invoice_spec.rb:331 # Invoice#save with an existing record when status changed to finalized with integration customer when sync invoices is true when sync sales orders is true enqueues sales order create job
rspec ./spec/models/invoice_spec.rb:339 # Invoice#save with an existing record when status changed to finalized with integration customer when sync invoices is true when sync sales orders is false enqueues invoice create job
rspec ./spec/models/invoice_spec.rb:359 # Invoice#save with an existing record when status changed to finalized with integration customer when sync invoices is false when sync sales orders is true enqueues sales order create job
```

ivannovosad: In case of invoice number it is in the before_save calback:

```
before_save :ensure_number
```

app/models/invoice.rb
ivannovosad: I created a separate [PR](https://github.com/getlago/lago-api/pull/2012)

PR #2009, opened 2024-05-14
feat(wallet): Rename RecurringTransactionRule#rule_type to trigger

rsempe:
## Context

After the release of real-time wallet version 1, some customers have expressed concerns about odd or missing behaviors.

## Description

The goal of this PR is to rename `Wallet#rule_type` into `Wallet#trigger`.

A new column `method` has been added as well to not create another migration in the next PR.

### DIFF
--- a/db/migrate/20240514072741_add_method_to_recurring_transaction_rules.rb
+++ b/db/migrate/20240514072741_add_method_to_recurring_transaction_rules.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddMethodToRecurringTransactionRules < ActiveRecord::Migration[7.0]
+  def change
+    rename_column :recurring_transaction_rules, :rule_type, :trigger
+    add_column :recurring_transaction_rules, :method, :integer, null: false, default: 0
+  end
+end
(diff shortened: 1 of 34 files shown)


db/migrate/20240514072741_add_method_to_recurring_transaction_rules.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddMethodToRecurringTransactionRules < ActiveRecord::Migration[7.0]
+  def change
+    rename_column :recurring_transaction_rules, :rule_type, :trigger
+    add_column :recurring_transaction_rules, :method, :integer, null: false, default: 0
```
julienbourdeau, line 6: I don't see where method is used in the PR. Is it a leftover of a previous implementation that didn't make it?

julienbourdeau (approved):

db/migrate/20240514072741_add_method_to_recurring_transaction_rules.rb
rsempe: It was in the description, but yep this new column `action` has been added as well to not create another migration in the next PR

julienbourdeau: Oh yeah I missed it. The new column seem to be named `method` tho, not `action`.

rsempe: Oh my bad 🤦 😅

PR #2012, opened 2024-05-16
Integrations invoices models only

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds integration resource models, factories and specs

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -282,22 +282,18 @@ class Invoice < ApplicationRecord
     save!
   end
 
+  def should_sync_invoice?
+    finalized? && customer.integration_customers.any? { |c| c.integration.sync_invoices }
+  end
+
+  def should_sync_sales_order?
+    finalized? && customer.integration_customers.any? { |c| c.integration.sync_sales_orders }
+  end
+
   private
 
   def should_assign_sequential_id?
     status_changed_to_finalized?
-  end
-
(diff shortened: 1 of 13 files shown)


app/models/invoice.rb
```diff
@@ -287,6 +288,18 @@ def should_assign_sequential_id?
     status_changed_to_finalized?
   end
 
+  def trigger_invoice_sync
```
vincent-pochet, line 291: Not sure we should trigger business logic from the model. What do you think of having this called from the services?

ivannovosad: Here in the [followup PR](https://github.com/getlago/lago-api/pull/2013/files) you can see that those methods should be called after save, so in this case I think it belongs in the model.

vincent-pochet: That's my point :). Having callbacks in model to sync with external services is usually not a good practice.

I would suggest to add the logic in the service dealing with the create/update of the invoice, or in last resorts into an observer

vincent-pochet: Happy to discuss it if you want :)

ivannovosad: I changed it so the model checks if it's possible to sync, then all the services that handle saving invoices will need to call `should_sync_invoice?` and `should_sync_sales_order?` and enquee the jobs.

vincent-pochet (approved):

PR #2030, opened 2024-05-17
fix: Attempt to fix flacky test in unique count scenario

vincent-pochet:


julienbourdeau (approved):
Thanks! I guess we will only know in a few weeks if it doesn't show up again 😀

vincent-pochet:
> Thanks! I guess we will only know in a few weeks if it doesn't show up again 😀

"Attempt" :D

PR #2032, opened 2024-05-17
fix(gql): Missing billing_configuration solution for OrganizationType

julienbourdeau:
## Description

I recently refactored GraphQL `OrganizationType` https://github.com/getlago/lago-api/pull/1919

`billing_configuration` is exposed in both Types but I forgot the resolution method in on type.

### DIFF
--- a/app/graphql/types/base_organization_type.rb
+++ b/app/graphql/types/base_organization_type.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Types
+  class BaseOrganizationType < BaseObject
+    def billing_configuration
+      {
+        id: "#{object&.id}-c0nf", # Each nested object needs ID so that appollo cache system can work properly
+        vat_rate: object&.vat_rate,
+        invoice_footer: object&.invoice_footer,
+        invoice_grace_period: object&.invoice_grace_period,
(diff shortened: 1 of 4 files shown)


app/graphql/types/base_organization_type.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Types
+  class BaseOrganizationType < BaseObject
```
vincent-pochet, line 4: Question: today organization types are located at the root of the `Types` module, mainly for legacy reasons.
Since almost all other types have their own sub-modules, should we use this PR as an opportunity to refact it?

vincent-pochet (approved):

app/graphql/types/base_organization_type.rb
ansmonjol: Agree with Vincent here. 
But as this PR fixes a production issue, can we perform the refactor after its merge?
Unless Julien already did it in the meantime, if that makes sense

ansmonjol (approved):

app/graphql/types/base_organization_type.rb
julienbourdeau: I will definitely move it to it's own subfolder!

PR #2043, opened 2024-05-20
feat(integrations): Add graphql for syncing invoices

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds a mutation for syncing invoices manually in the UI


spec/graphql/mutations/integrations/sync_invoice_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  before { subject }
+
+  it_behaves_like 'requires current user'
+  it_behaves_like 'requires current organization'
+  it_behaves_like 'requires permission', 'organization:integrations:update'
+
+  it 'syncs an invoice' do
+    # expect(::Integrations::Aggregator::Invoices::CreateJob).to have_received(:perform_later).with(invoice:)
```
lovrocolic, line 41: I am not sure why is this commented out?

app/graphql/types/integrations/sync_invoice_input.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  module Integrations
```
lovrocolic, line 4: As I can see, we don't need this class

app/graphql/mutations/integrations/sync_invoice.rb
```diff
@@ -0,0 +1,25 @@
...
+      graphql_name 'SyncIntegrationInvoice'
+      description 'Sync integration invoice'
+
+      argument :invoice_id, ID, required: true
+
+      type Types::IntegrationItems::Object.collection_type, null: false
+
+      def resolve(**args)
+        ::Integrations::Aggregator::Invoices::CreateJob.perform_later(invoice:)
+
```
lovrocolic, line 20: I guess, we should first find invoice by `invoice_id`. Here, we don't have invoice object at all

spec/graphql/mutations/integrations/sync_invoice_spec.rb
ivannovosad: Fixed. Jobs were on another branch at the time this was written.

app/graphql/types/integrations/sync_invoice_input.rb
ivannovosad: I moved the argument to the input and we're using this input now.

app/graphql/mutations/integrations/sync_invoice.rb
```diff
@@ -0,0 +1,29 @@
...
+      description 'Sync integration invoice'
+
+      input_object_class Types::Integrations::SyncInvoiceInput
+
+      field :invoice_id, ID, null: true
+
+      def resolve(**args)
+        invoice = current_organization.invoices.find_by(id: args[:invoice_id])
+
+        ::Integrations::Aggregator::Invoices::CreateJob.perform_later(invoice:)
```
vincent-pochet, line 21: I'm wondering if we could introduce a `call_async` logic into some services that just need to call an external service:

```ruby
module Integrations
  module Aggregator
    module Invoices
      class CreateService

# ...
        def call_async
          return result.not_found_failure!(resource: 'invoice') unless invoice

          ::Integrations::Aggregator::Invoices::CreateJob.perform_later(invoice:)
          result.invoice_id = invoice.id
          result
        end
```

It brings 3 advantages:
- keeping the presence check for the invoice and render "NOT FOUND" if applicable
- building a proper result inside the service and not in the mutation
- keeping classic `result.success? ? result.invoice_id : result_error(result)` result handing in the mutation

In the future, if we see more of this, we could even implement the pattern at service level

lovrocolic (approved):
Looks good, just one comment!

PR #2055, opened 2024-05-22
chore(tests): Add cache metadata to configure Rails.cache driver

julienbourdeau:
## Description

I was writing tests with redis cache driver and realized we didn't have an easy way to set the drivers in tests.
This first PR introduce a simple way to pass metadata to your test and rspec will ensure that the correct cache driver is set.

It avoids mocking `Rails.cache`. 

In a following PR, I'll add the Redis driver (because I need it).

Note that it can be set on `describe` or `example`.

```rb
RSpec.describe Invites::AcceptService cache: :memory do

  context 'with cache' do
    it do
      # Rails.cache is using the memory driver
    end
  end

  context 'without cache', cache: :null do
    it do
      # Rails.cache is using the null driver
    end
  end
end
```

### DIFF
--- a/spec/graphql/mutations/auth/okta/accept_invite_spec.rb
+++ b/spec/graphql/mutations/auth/okta/accept_invite_spec.rb
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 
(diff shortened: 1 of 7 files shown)


vincent-pochet (approved):
Nice! :rocket:

PR #2058, opened 2024-05-23
feat(integrations): return external_id in invoice graphql object

lovrocolic:
## Context

## Description

This PR adds `external_id` to the invoice object so that proper URL on integration provider can be displayed in the UI

### DIFF
--- a/app/graphql/types/invoices/object.rb
+++ b/app/graphql/types/invoices/object.rb
@@ -52,6 +52,7 @@ module Types
       field :invoice_subscriptions, [Types::InvoiceSubscription::Object]
       field :subscriptions, [Types::Subscriptions::Object]
 
+      field :integration_external_id, String, null: true
       field :integration_syncable, GraphQL::Types::Boolean, null: false
 
       def applied_taxes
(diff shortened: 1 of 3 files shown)


app/graphql/types/invoices/object.rb
```diff
@@ -52,6 +52,7 @@ class Object < Types::BaseObject
       field :invoice_subscriptions, [Types::InvoiceSubscription::Object]
       field :subscriptions, [Types::Subscriptions::Object]
 
+      field :integration_external_id, String, null: true
```
ivannovosad, line 55: Can we do a small change so that it matches how the other external fields are named?

```suggestion
      field :external_integration_id, String, null: true
```

app/graphql/types/invoices/object.rb
```diff
@@ -62,6 +63,19 @@ def integration_syncable
         object.should_sync_invoice? &&
           object.integration_resources.where(resource_type: 'invoice', syncable_type: 'Invoice').none?
       end
+
+      def integration_external_id
```
ivannovosad, line 67:
```suggestion
      def external_integration_id
```

ivannovosad (changes requested):

ivannovosad (approved):

PR #2065, opened 2024-05-23
feat(wallet): Add granted and paid credits on recurring transaction rule

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to be able to create recurring transaction rule by passing `granted_credits` and `paid_credits` fields (without taking the ones defined on the wallet object).

### DIFF
--- a/app/services/wallets/create_service.rb
+++ b/app/services/wallets/create_service.rb
@@ -2,40 +2,45 @@
 
 module Wallets
   class CreateService < BaseService
-    def create(args)
-      return result unless valid?(**args)
+    def initialize(params:)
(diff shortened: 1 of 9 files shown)


app/services/wallets/create_service.rb
```diff
@@ -2,45 +2,45 @@
 
 module Wallets
   class CreateService < BaseService
-    def create(args)
```
vincent-pochet, line 5: One of the last survivor of the pre-call era 🪦 :clap:

vincent-pochet (approved):

PR #2073, opened 2024-05-27
feat(clickhouse): Rewrite count query without CTE

vincent-pochet:
## Description

This PR is trying to improve query performance for the count aggregation on Clickhouse queries by removing the CTE in favor of a pure select query.

It also takes advantage of the `uniqExact` method to replace `COUNT(DISTINC`

### DIFF
--- a/db/clickhouse_migrate/20231024084411_create_events_raw.rb
+++ b/db/clickhouse_migrate/20231024084411_create_events_raw.rb
@@ -2,7 +2,7 @@ class CreateEventsRaw < ActiveRecord::Migration[7.0]
   def change
     options = <<-SQL
       MergeTree
-      ORDER BY (organization_id, external_subscription_id, code, timestamp)
+      ORDER BY (organization_id, external_subscription_id, code, transaction_id, timestamp)
     SQL
 
     create_table :events_raw, id: false, options: do |t|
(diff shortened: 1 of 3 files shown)


rsempe (approved):

db/clickhouse_migrate/20231024084411_create_events_raw.rb
```diff
@@ -2,7 +2,7 @@ class CreateEventsRaw < ActiveRecord::Migration[7.0]
   def change
     options = <<-SQL
       MergeTree
-      ORDER BY (organization_id, external_subscription_id, code, timestamp)
+      ORDER BY (organization_id, external_subscription_id, code, transaction_id, timestamp)
```
julienbourdeau, line 5: I'm surprised to see transaction_id here. I expect it would be UUID or this kind of random string. Does it help with the deduplication?

julienbourdeau (approved):

db/clickhouse_migrate/20231024084411_create_events_raw.rb
vincent-pochet: @julienbourdeau  Exactly, Clickhouse uses the order by as the primary key.
Adding the transaction_id here ensure that a transaction_id cannot appears more than once in a query result

jdenquin (approved):

PR #2074, opened 2024-05-27
feat(wallet): Remove duplication on wallet rules validation

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to:

- Add validation on credits for recurring transaction rules
- Remove duplication on wallet rules validation

### DIFF
--- a/app/services/wallets/recurring_transaction_rules/validate_service.rb
+++ b/app/services/wallets/recurring_transaction_rules/validate_service.rb
@@ -11,6 +11,7 @@ module Wallets
       def call
         return false unless valid_trigger?
         return false unless valid_method?
+        return false unless valid_credits?
 
         true
       end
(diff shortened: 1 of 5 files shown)


vincent-pochet (approved):

app/services/wallets/recurring_transaction_rules/validate_service.rb
```diff
@@ -40,11 +41,18 @@ def valid_threshold_trigger?
       end
 
       def valid_method?
-        (method == "target") ? valid_target_method? : true
+        (method == "target") ? valid_decimal?(params[:target_ongoing_balance]) : true
```
julienbourdeau, line 44: In other methods you seem to return early instead, which might be clearer 🤷‍♂️ Just a suggestion

Maybe `valid_credits?` can be simplified too. `unless` combined with `||` isn't very easy, at least for me ^^

```suggestion
        return valid_decimal?(params[:target_ongoing_balance]) if method == "target"
        
        true
```

julienbourdeau (approved):

PR #2082, opened 2024-05-29
fix(subscription): Pending subscription cancel status

brunomiguelpinto:
## Context

This pull request addresses the issue where pending subscriptions are incorrectly marked as terminated when canceled. The goal is to ensure that pending subscriptions are marked as canceled, without incorrectly triggering termination-related webhooks.

## Description

When an active subscription is terminated, it should be marked as terminated, but when a pending subscription is deleted, it should be marked as canceled


lovrocolic (approved):
nice! 🚀 Can you please build review app for this one?

jdenquin (approved):

PR #2087, opened 2024-05-29
feat(fees): Add billing_at arg to CreatePayInAdvance flow

julienbourdeau:
## Context

When a charge is `invoiceable: true, pay_in_advance: true` an invoice is created every time an event is received. If the  Billable Metrics attached is `recurring: true`, each month this charge will be added to the subscription invoice.

If the charge is `invoiceable: false, pay_in_advance: true` a fee is created every time an even is received (but no invoice). If the  Billable Metrics attached is `recurring: true`, nothing happens each months.

We want to change this behavior and create a new Fee each month.

## Description

Each time a new fee is created for the new period, we'll continue to attach it to the same event that triggered it in the first place.

In this PR, we extract the "timestamp" into an arg so on following months, we can use the correct billing date instead of relying on the time of the event.

Without this change, the boundaries of any fee would always be the boundaries computed at the time of the event.

### DIFF
--- a/app/services/fees/create_pay_in_advance_service.rb
+++ b/app/services/fees/create_pay_in_advance_service.rb
@@ -2,9 +2,10 @@
 
 module Fees
   class CreatePayInAdvanceService < BaseService
-    def initialize(charge:, event:, estimate: false)
+    def initialize(charge:, event:, billing_at: event.timestamp, estimate: false)
       @charge = charge
       @event = event
+      @billing_at = billing_at
       @estimate = estimate
 
       super
@@ -33,7 +34,7 @@ module Fees
 
     private
 
-    attr_reader :charge, :event, :estimate
+    attr_reader :charge, :event, :billing_at, :estimate
 
     delegate :billable_metric, to: :charge
     delegate :subscription, :customer, to: :event
@@ -89,7 +90,7 @@ module Fees
     def date_service
       @date_service ||= Subscriptions::DatesService.new_instance(
         subscription,
-        event.timestamp,
+        billing_at,
         current_usage: true
       )
     end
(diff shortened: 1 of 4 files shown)


app/services/fees/create_pay_in_advance_service.rb
```diff
@@ -101,7 +102,7 @@ def boundaries
         charges_from_datetime: date_service.charges_from_datetime,
         charges_to_datetime: date_service.charges_to_datetime,
         charges_duration: date_service.charges_duration_in_days,
-        timestamp: Time.current
+        timestamp: Time.current # QUESTION: should we use billing_at ?
```
julienbourdeau, line 105: I believe `Time.current` was used as a shortcut. Usually in boundaries, the timestamp is "when the thing was billed". It should have been billed at `billing_at` but the execution can be delayed for technical reason (sidekiq queue being late, job failed and is tried later...).

Or am I wrong and the timestamp should reflect "the time it was actually executed"?

vincent-pochet: @julienbourdeau you are right, it should definitely be `billing_ag`

app/services/fees/create_pay_in_advance_service.rb
```diff
@@ -135,7 +136,7 @@ def cache_aggregation_result(aggregation_result:, charge_filter:)
       CachedAggregation.create!(
         organization_id: event.organization_id,
         event_id: event.id,
-        timestamp: event.timestamp,
+        timestamp: billing_at, # QUESTION: Should we use billing_at ?
```
vincent-pochet, line 139: Same, it's better to use billing_at

vincent-pochet (approved):

PR #2089, opened 2024-05-29
misc(clickhouse): Define pool number

jdenquin:
## Description

- Define a pool number for Clickhouse adapter

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -84,6 +84,7 @@ production:
     database: <%= ENV['LAGO_CLICKHOUSE_DATABASE'] %>
     host: <%= ENV['LAGO_CLICKHOUSE_HOST'] %>
     port: <%= ENV.fetch('LAGO_CLICKHOUSE_PORT', 8123) %>
+    pool: <%= ENV.fetch('DATABASE_POOL', 10) %>
     username: <%= ENV['LAGO_CLICKHOUSE_USERNAME'] %>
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -84,6 +84,7 @@ production:
     database: <%= ENV['LAGO_CLICKHOUSE_DATABASE'] %>
     host: <%= ENV['LAGO_CLICKHOUSE_HOST'] %>
     port: <%= ENV.fetch('LAGO_CLICKHOUSE_PORT', 8123) %>
+    pool: <%= ENV.fetch('DATABASE_POOL', 10) %>
```
julienbourdeau, line 87: I'm wondering if we necessarily want the same value for PG and ClickHouse 🤔 

How do you feel about `LAGO_CLICKHOUSE_POOL`, I know we already have a lot of of env vars 🙈 

```suggestion
    pool: <%= ENV.fetch('LAGO_CLICKHOUSE_POOL', 10) %>
```

julienbourdeau (approved):

config/database.yml
jdenquin: we use clickhouse in HTTP, so it's not really a POOL, it's just for AR to be happy when we have a lot of requests, adding one more env var here is useless I think

brunomiguelpinto (approved):

PR #2090, opened 2024-05-29
feat(fees): Create recurring pay in advance fees on renewal

julienbourdeau:
## Context

When a charge is `invoiceable: true, pay_in_advance: true` an invoice is created every time an event is received. If the Billable Metrics attached is `recurring: true`, each month this charge will be added to the subscription invoice.

If the charge is `invoiceable: false, pay_in_advance: true` a fee is created every time an even is received (but no invoice). If the Billable Metrics attached is `recurring: true`, nothing happens each months.

We want to change this behavior and create a new Fee each month.

#### Current Behavior

1. **Invoiceable Charges:**
   - When `invoiceable: true` and `pay_in_advance: true`:
     - An invoice is created every time an event is received.
     - If the Billable Metrics attached are `recurring: true`, this charge is added to the subscription invoice each month.

2. **Non-Invoiceable Charges:**
   - When `invoiceable: false` and `pay_in_advance: true`:
     - A fee is created every time an event is received (but no invoice is generated).
     - If the Billable Metrics attached are `recurring: true`, nothing happens each month.

#### New Behavior
1. **Invoiceable Charges:**
   - No changes. Invoiceable charges will continue to function as before.
   
2. **Non-Invoiceable Charges:**
   - When `invoiceable: false` and`pay_in_advance: true`:
    - A fee will still be created every time an event is received (with no invoice generated).
    - If the Billable Metrics attached are `recurring: true`, a new fee will now be created each month, ensuring consistent billing for recurring metrics.

#### Change

We have updated the system to ensure that recurring non-invoiceable charges generate a new fee each month.

#### Details of Implementation

1. **Job Dispatch:**
   - We introduced a job `BillRecurringPayInAdvanceChargesJob` which is dispatched in the `app/services/subscriptions/billing_service.rb` to handle the creation of recurring fees each month.
   - This job is scheduled to run at the beginning of each billing cycle.

2. **Reusing Existing Code:**
   - The job uses the existing `Fees::CreateRecurringPayInAdvanceService` to handle the logic for creating recurring fees.

3. **Trigger Conditions:**
   - The job is triggered under the following conditions:
     - Subscriptions must be active.
     - The charge must have `invoiceable: false` and `pay_in_advance: true`.
     - The Billable Metrics attached to the charge must be `recurring: true`.

4. **Using Previous Fee and Event:**
   - The service checks the most recent fee created for the charge.
   - It retrieves the associated event from the last fee (`pay_in_advance_event_id`).
   - If an event is found, a new fee is created for the current billing period based on the previous event..

### DIFF
--- a/app/jobs/bill_recurring_pay_in_advance_charges_job.rb
+++ b/app/jobs/bill_recurring_pay_in_advance_charges_job.rb
@@ -5,8 +5,5 @@ class BillRecurringPayInAdvanceChargesJob < ApplicationJob
 
   def perform(subscriptions, timestamp)
     Fees::CreateRecurringPayInAdvanceService.call(subscriptions:, billing_at: Time.zone.at(timestamp))
-
-    # TODO: Should we retry the job if it fails?
-    # result.raise_if_error!
   end
 end
--- a/app/jobs/bill_recurring_pay_in_advance_charges_job.rb
+++ b/app/jobs/bill_recurring_pay_in_advance_charges_job.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class BillRecurringPayInAdvanceChargesJob < ApplicationJob
+  queue_as 'billing'
+
+  def perform(subscriptions, timestamp)
+    Fees::CreateRecurringPayInAdvanceService.call(subscriptions:, billing_at: Time.zone.at(timestamp))
+
+    # TODO: Should we retry the job if it fails?
+    # result.raise_if_error!
+  end
+end
--- a/app/services/fees/create_recurring_pay_in_advance_service.rb
+++ b/app/services/fees/create_recurring_pay_in_advance_service.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Fees
+  class CreateRecurringPayInAdvanceService < BaseService
+    def initialize(subscriptions:, billing_at:)
+      @subscriptions = subscriptions
+      @billing_at = billing_at
+
+      super
+    end
+
+    def call
+      plan_ids = subscriptions.select(&:active?).map(&:plan_id).uniq
+
+      Charge.joins(:billable_metric)
+        .where(plan_id: plan_ids, pay_in_advance: true, invoiceable: false)
+        .where(billable_metrics: {recurring: true})
+        .find_each do |charge|
+        last_fee = charge.fees.order(created_at: :desc).first
+        event = Event.find_by(id: last_fee&.pay_in_advance_event_id)
+
+        if event
+          Fees::CreatePayInAdvanceJob.perform_later(charge:, event:, billing_at:)
+        end
+      end
+
+      result
+    end
+
+    private
+
+    attr_reader :subscriptions, :billing_at
+  end
+end
--- a/app/services/fees/create_recurring_pay_in_advance_service.rb
+++ b/app/services/fees/create_recurring_pay_in_advance_service.rb
@@ -12,10 +12,7 @@ module Fees
     def call
       plan_ids = subscriptions.select(&:active?).map(&:plan_id).uniq
 
-      Charge.joins(:billable_metric)
-        .where(plan_id: plan_ids, pay_in_advance: true, invoiceable: false)
-        .where(billable_metrics: {recurring: true})
-        .find_each do |charge|
+      eligible_charges(plan_ids).find_each do |charge|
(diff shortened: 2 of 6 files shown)


app/services/fees/create_recurring_pay_in_advance_service.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def call
+      plan_ids = subscriptions.select(&:active?).map(&:plan_id).uniq
+
+      Charge.joins(:billable_metric)
+        .where(plan_id: plan_ids, pay_in_advance: true, invoiceable: false)
+        .where(billable_metrics: {recurring: true})
+        .find_each do |charge|
+        last_fee = charge.fees.order(created_at: :desc).first
+        event = Event.find_by(id: last_fee&.pay_in_advance_event_id)
```
vincent-pochet, line 20: We will have to adapt this condition for in advance handling in clickhouse

app/jobs/bill_recurring_pay_in_advance_charges_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class BillRecurringPayInAdvanceChargesJob < ApplicationJob
+  queue_as 'billing'
+
+  def perform(subscriptions, timestamp)
+    Fees::CreateRecurringPayInAdvanceService.call(subscriptions:, billing_at: Time.zone.at(timestamp))
+
+    # TODO: Should we retry the job if it fails?
+    # result.raise_if_error!
```
vincent-pochet, line 10: the raise_if_error is here to make sure we are not silencing any problems. Since the service is built with the `billing_at` input, and should use a DB transaction, I guess we are sage to raise and retry if mandatory.
What do you think?

app/services/fees/create_recurring_pay_in_advance_service.rb
julienbourdeau: I'm not too sure what you would need to change, I think all of this is happening in Rails

brunomiguelpinto: do we need to change something on our side? or will this be done in another pr related to the clubhouse?

app/jobs/bill_recurring_pay_in_advance_charges_job.rb
brunomiguelpinto: since the service enqueues a job and verifies if we have the data needed, does retrying make sense?

app/services/fees/create_recurring_pay_in_advance_service.rb
vincent-pochet: **Let's handle it with clickhouse.**

In Clickhouse, events does not have a unique ID like the one in PG, and so we will not be able to link in advance events to fees like we are doing here

vincent-pochet (approved):

app/services/fees/create_recurring_pay_in_advance_service.rb
```diff
@@ -0,0 +1,35 @@
...
+      plan_ids = subscriptions.select(&:active?).map(&:plan_id).uniq
+
+      eligible_charges(plan_ids).find_each do |charge|
+        last_fee = charge.fees.order(created_at: :desc).first
+        event = Event.find_by(id: last_fee&.pay_in_advance_event_id)
+
+        if event
+          Fees::CreatePayInAdvanceJob.perform_later(charge:, event:, billing_at:)
+        end
+      end
```
vincent-pochet, line 22: Could you return the `result` event if no property is attached to it? It's just to keep the common pattern ;)

app/services/fees/create_recurring_pay_in_advance_service.rb
julienbourdeau: Ah! I wasn't aware of that. The `Fees::CreatePayInAdvanceService` is really based on the event so we'll definitely need to pass it an event.

It's probably possible to refactor the service not depend on event but directly on some attributes (subscriptions, customers, date...) but it's not easy. Maybe easier than other solutions tho.

julienbourdeau:
After further discussion, we realize it's not what we want to build. We need to aggregate events on the full past period.

app/services/fees/create_recurring_pay_in_advance_service.rb
julienbourdeau: Excellent news @vincent-pochet. This is not at all how we want this to work 🙈 The actual version won't need the event.

PR #2093, opened 2024-05-30
feat(invoice): Hide subscription fee when 0

vincent-pochet:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/hide-fees-at-zero

## Context

Users do not want to display subscription fee on invoice when the amount is 0.

## Description

This PR ensures that the subscription fee is hidden on the invoice except when no other charge fees are visible as well

### DIFF
--- a/app/views/helpers/fee_display_helper.rb
+++ b/app/views/helpers/fee_display_helper.rb
@@ -8,4 +8,11 @@ class FeeDisplayHelper
 
     " • #{fee.grouped_by.values.compact.join(" • ")}"
   end
+
+  def self.should_display_subscription_fee?(invoice_subscription)
+    return false if invoice_subscription.blank?
+    return true if invoice_subscription.charge_amount_cents.zero?
+
+    invoice_subscription.subscription_amount_cents.positive?
+  end
 end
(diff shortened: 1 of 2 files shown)


app/views/helpers/fee_display_helper.rb
```diff
@@ -8,4 +8,11 @@ def self.grouped_by_display(fee)
 
     " • #{fee.grouped_by.values.compact.join(" • ")}"
   end
+
+  def self.should_display_subscription_fee?(invoice_subscription)
+    return false if invoice_subscription.blank?
+    return true if invoice_subscription.charge_amount_cents.zero?
```
julienbourdeau, line 14: If I understand correct, this would display the subscription fee is the amount is zero 🤔 which seems to be the opposite of what we're trying to do.

```suggestion
    return false if invoice_subscription.charge_amount_cents.zero?
```

vincent-pochet: @julienbourdeau  no, actually it is an exception to display the subscription if there is no other charge fee to display

julienbourdeau: Oh yeah I see 👌

julienbourdeau (approved):

PR #2098, opened 2024-05-30
feat(wallet): Add started_at to recurring_transaction_rules

rsempe:
## Roadmap Task

## Context

## Description

The goal of this PR is to add `started_at` to `recurring_transaction_rules`.


vincent-pochet (approved):
:rocket:

PR #2116, opened 2024-06-03
fix(tests): Clear enqueued jobs before each test run

brunomiguelpinto:
## Context

To ensure that our tests run reliably and do not interfere with each other, it's necessary to clear any enqueued jobs before each test. This prevents flaky tests caused by jobs being carried over from previous tests.

## Description

This PR adds a configuration to clear enqueued jobs before each test run. Specifically, it adds the following to the test setup:

```ruby
config.before(:each) do
  ActiveJob::Base.queue_adapter.enqueued_jobs.clear
end

### DIFF
--- a/spec/services/subscriptions/create_service_spec.rb
+++ b/spec/services/subscriptions/create_service_spec.rb
@@ -447,8 +447,7 @@ RSpec.describe Subscriptions::CreateService, type: :service do
             let(:old_plan) { create(:plan, amount_cents: 100, organization:, pay_in_advance: false) }
 
             it 'enqueues a job to bill the existing subscription' do
-              create_service.call
-              expect(BillSubscriptionJob).to have_been_enqueued.at_least(1).times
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: spec/services/subscriptions/create_service_spec.rb [hunk 1]
```diff
@@ -447,8 +447,7 @@
             let(:old_plan) { create(:plan, amount_cents: 100, organization:, pay_in_advance: false) }
 
             it 'enqueues a job to bill the existing subscription' do
-              create_service.call
-              expect(BillSubscriptionJob).to have_been_enqueued.at_least(1).times
+              expect { create_service.call }.to have_enqueued_job(BillSubscriptionJob)
```
julienbourdeau, line 450: Did you edit these tests because they were failing with your change in the setup 🤔

julienbourdeau (approved):

PR #2121, opened 2024-06-03
feat(pdf): Ensure Invoice PDF is created before sending emails and webhooks

julienbourdeau:
### Context
Separate PDF generation into a different Worker and Queue (:pdf). 

### Description
Move PDF generation to a distinct worker and queue.  
Change the process to generate the PDF and send the email 

> [!CAUTION]
> This PR introduces a new sidekiq queue named `pdf`.
> Other than the sidekiq config here, what else should be updated?

### Changes Introduced

- **New Job Implementation**: Introduced a new job `GeneratePdfAndNotifyJob` to handle the PDF generation and email notification.
- **Mailer Invocation Update**: Modified the invoicing process to utilize the new job, ensuring the email is sent only after the PDF is successfully generated.
- **Test Configuration**: Added a `PdfHelper` module and updated RSpec configuration to stub PDF generation in tests tagged with `:scenarios`, reducing repetitive stubbing code in individual tests.

## Webhook Events

> [!IMPORTANT]  
> Stop listening to `invoice.created` to request the PDF creation via the API. Instead, listen to `invoice.generated` event, which always has the PDF file url.

### `invoice.created` webhook

This webhook is sent after the invoice object is created, in the webhook payload you get an `invoice` object but the PDF link is not set yet. The PDF was not generated yet.

Until now, users would typically call back our API to request the PDF generation. **This is now deprecated!**

Instead, wait for the `invoice.generated` webhook.

### `invoice.generated` webhook

This webhook is sent when the invoice PDF is generated. Until now, we generated the PDF only if you requested it via the API, via the UI or if the invoice was sent over email. If you don't send invoice over email, you would rarely receive this event.

Now, **the PDF is generated every time an Invoice model is created**. It's now recommended to listent this event if you want the PDF file URL to be part of the webhook payload.

| Webhook name | Desc |
|----------------:|--------|
| `invoice.created` | This webhook indicates that the invoice has been created, but the PDF is not yet ready. It's delivered fast after the invoice creation. |
| `invoice.generated` | This webhook indicates that the PDF has been generated. You receive the Invoice object WITH the PDF URL. This webhook is sent a bit after `invoice.created` because generating PDF is typically a bit longer. |

### DIFF
--- a/spec/services/invoices/add_on_service_spec.rb
+++ b/spec/services/invoices/add_on_service_spec.rb
@@ -54,28 +54,28 @@ RSpec.describe Invoices::AddOnService, type: :service do
       end.to have_enqueued_job(SendWebhookJob)
     end
 
-    it 'does not enqueue an SendEmailJob' do
+    it 'enqueue an GeneratePdfAndNotifyJob with email false' do
       expect do
         invoice_service.create
-      end.not_to have_enqueued_job(SendEmailJob)
+      end.to have_enqueued_job(Invoices::GeneratePdfAndNotifyJob).with(hash_including(email: false))
     end
 
     context 'with lago_premium' do
       around { |test| lago_premium!(&test) }
 
-      it 'enqueues an SendEmailJob' do
+      it 'enqueues an GeneratePdfAndNotifyJob with email true' do
         expect do
           invoice_service.create
-        end.to have_enqueued_job(SendEmailJob)
+        end.to have_enqueued_job(Invoices::GeneratePdfAndNotifyJob).with(hash_including(email: true))
       end
 
       context 'when organization does not have right email settings' do
         before { applied_add_on.customer.organization.update!(email_settings: []) }
 
-        it 'does not enqueue an SendEmailJob' do
+        it 'enqueue an GeneratePdfAndNotifyJob with email false' do
           expect do
             invoice_service.create
-          end.not_to have_enqueued_job(SendEmailJob)
+          end.to have_enqueued_job(Invoices::GeneratePdfAndNotifyJob).with(hash_including(email: false))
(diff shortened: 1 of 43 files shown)


julienbourdeau:
Closed in favor of https://github.com/getlago/lago-api/pull/2127

spec/services/invoices/add_on_service_spec.rb
```diff
@@ -54,28 +54,28 @@
       end.to have_enqueued_job(SendWebhookJob)
     end
 
-    it 'does not enqueue an SendEmailJob' do
+    it 'enqueue an GeneratePdfAndNotifyJob with email false' do
       expect do
         invoice_service.create
-      end.not_to have_enqueued_job(SendEmailJob)
+      end.to have_enqueued_job(Invoices::GeneratePdfAndNotifyJob).with(hash_including(email: false))
```
ancorcruz, line 60: I have seen none of the tests that expects the new `GeneratePdfAndNotifyJob` being queued checks the invoice is passed as an argument. Would you consider to add the invoice argument expectation at least in the happy path?

ancorcruz (approved):
Looks good to me.

I have just left one comment for your consideration.

vincent-pochet (approved):

spec/services/invoices/add_on_service_spec.rb
brunomiguelpinto: Thank you for the suggestion. The primary goal of our test is to verify that the GeneratePdfAndNotifyJob is enqueued with the correct email flag. I see the importance of passing the correct invoice as an argument; however, accessing the invoice within the current test setup is not straightforward.

PR #2122, opened 2024-06-03
feat(integrations): Sync payments job

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds `Integrations::Aggregator::Payments::CreateJob`

### DIFF
--- a/app/services/integrations/aggregator/payments/create_service.rb
+++ b/app/services/integrations/aggregator/payments/create_service.rb
@@ -44,7 +44,7 @@ module Integrations
         def call_async
           return result.not_found_failure!(resource: 'payment') unless payment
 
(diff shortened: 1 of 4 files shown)


app/services/integrations/aggregator/payments/create_service.rb
```diff
@@ -44,7 +44,7 @@ def call
         def call_async
           return result.not_found_failure!(resource: 'payment') unless payment
 
-          ::Integrations::Aggregator::PAyments::CreateJob.perform_later(payment:)
+          ::Integrations::Aggregator::Payments::CreateJob.perform_later(payment:)
```
lovrocolic, line 47: Shouldn't we add this line in separate PR since this is actually calling integration. I would suggest having everything except this line so that we can merge this PR

PR #2128, opened 2024-06-04
feat(billing): Add billing_at arg to BillingService

julienbourdeau:
## Description

If you need to re-run billing in the past (like if the clock didn't run) or if you want to test on staging envs, it's easier to just pass `billing_at` to the billing service.

It's also more inline with what we do in other similar services.

```ruby
Subscriptions::BillingService.call 3.days.from_now
```

Today, we do it with `timecop`

```ruby
require 'timecop'

Timecop.freeze(DateTime.new(2024, 6, 20, 10)) do
  Subscriptions::BillingService.call
end
```

### DIFF
--- a/app/services/subscriptions/billing_service.rb
+++ b/app/services/subscriptions/billing_service.rb
@@ -2,9 +2,8 @@
 
 module Subscriptions
   class BillingService < BaseService
-    def call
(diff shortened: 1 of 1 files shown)


vincent-pochet (approved):
Nice little one! :)

PR #2132, opened 2024-06-05
feat(test): Add not invoiceable in_advance test

andrasszommer:
## Context

I created a scenario for my specific in_advance, prorated `invoiceable: false` use-case.

## Description

⚠️ **This merge request is not intended to be pulled, it's exists for demo purposes.**

### DIFF
--- a/spec/scenarios/commitments/minimum/in_advance/anniversary/not_invoiceable_spec.rb
+++ b/spec/scenarios/commitments/minimum/in_advance/anniversary/not_invoiceable_spec.rb
@@ -1,0 +1,128 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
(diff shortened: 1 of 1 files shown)


julienbourdeau:
I'm working on non invoiceable in_advance charges right now. Any test scenario is more than welcome! 🙌 🚀

andrasszommer:
The only reason is why i'm not sure about this test, is because there may be a chance that Lago is "not wanting" to work like the way I think a non invoiceable in_advance billing should work. Please check the attached issue to understand my use-case.

jdenquin:
@andrasszommer this has been covered by the feature @julienbourdeau worked on, it's released and available!

I'm closing this one so!

PR #2136, opened 2024-06-06
feat(anrok-integration): refactor create integration customer service

lovrocolic:
## Context

Lago does not support tax tool integrations, but support for it is currently being implemented.

## Description

This PR adds needed changes in create integration service.

Anrok does not sync customer right away. Instead, customer is synced when the first document is sent to the anrok.

Based on requirements from above, we changed a bit generic create service so that it works based on specific integration needs.


ivannovosad (approved):
Lokks good! Maybe we can also add some specs for the factory.

PR #2139, opened 2024-06-06
feat(invoice): Introduce charge_in_advance invoice status

julienbourdeau:
## Description

When a plan has a `pay_in_advance` charge, we create an invoice every time we receive an event related to this charge.

Today, these invoices are `invoice_type: :subscription`. This PR create a new type of invoice named `charge_in_advance`.

* This allows us to differentiate them from `subscription` invoice which is created at the beginning of a new period.
* The same webhook are triggered but the invoice objects now has a different status.
* On the first day of the period, the subscription invoice might have only charge fees (pay in arrears) but no subscription fee. It's still a `subscription` invoice because it happened "on billing day". The type is more related to "when and how" it happens than what fees it holds.
* ~Existing invoices will *NOT* be migrated and will continue to be have `subscription` type forever.~ I think we need to migrate the data otherwise we can never safely rely on this type and it becomes useless to even introduce it.

I believe we recently introduce the `charge_in_advance` argument to the `Invoices::CreateGeneratingService` because this type was lacking. See https://github.com/getlago/lago-api/pull/1800

### Why now?

We're looking new invoices for few new feature:
* progressing invoicing 
* "non invoiceable fees" invoice (we'll find a better name)
and I think this type is missing.

### Breaking change?

Because the `invoice_type` received in the webook will change, we might consider this as a breaking change.
How should we handle this? I believe it's acceptable since it will be part of the changelog for selfhosted and we can warn cloud users before this is released.

What do you think? What am I missing?

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -48,7 +48,7 @@ class Invoice < ApplicationRecord
     allow_nil: true,
     with_model_currency: :currency
 
-  INVOICE_TYPES = %i[subscription add_on credit one_off].freeze
+  INVOICE_TYPES = %i[subscription add_on credit one_off charge_in_advance].freeze
   PAYMENT_STATUS = %i[pending succeeded failed].freeze
   STATUS = %i[draft finalized voided generating].freeze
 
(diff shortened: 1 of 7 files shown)


julienbourdeau:
I'm convinced this should be done but let's get the new `invoiceable: false` feature done before looking into this.

PR #2140, opened 2024-06-06
chore(spec): Use `awesome_print` to debug tests

julienbourdeau:
## Description

I use `pp` a lot to print variables or messages when I'm debugging with tests. Unfortunately, it's printing only one line without colors in my tests. It's all pretty in the console. I must be missing something but one solution I like is to use the `awesome_print` gem.

Am I the only one with this problem?

### BEFORE

CleanShot 2024-06-06 at 15 27 47@2x

### AFTER

CleanShot 2024-06-06 at 15 26 26@2x

### BONUS

You can go and uncomment one line to easily see where you left your `pp`. 
Especially useful when you started editing gems 😬 

CleanShot 2024-06-06 at 15 26 41@2x

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -99,6 +99,7 @@ group :development, :test do
   gem 'rspec-rails'
   gem 'simplecov', require: false
   gem 'webmock'
+  gem 'awesome_print', require: false
 end
 
 group :test do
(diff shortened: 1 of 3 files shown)


vincent-pochet (approved):
:+1:

brunomiguelpinto (approved):

PR #2143, opened 2024-06-06
feat(events): Refactor PayInAdvance Billable Metric validation

julienbourdeau:
## Description

Whenever a event associated with PayInAdvance charge is recieved, we create a Fee for this event.

The validation was performed only after we create the Fee for `invoiceable: false` charges creating some disparities if the charge is invoiceable or not.

I realized that there is already a validation on the aggregation in the Charge model, which I refactored.

~I have removed all the validation in the PostProcess events because, you're not supposed to be able to create invalid charge (see Charge model validation) and if you send events without the `properties[bm.field_name]`, I'm not sure it should fail silently.~ It creates a fee even without valid properties so I'm keeping the condition as-is, but earlier in the function.

### DIFF
--- a/app/services/events/post_process_service.rb
+++ b/app/services/events/post_process_service.rb
@@ -30,11 +30,11 @@ module Events
       result.event = event
       result
     rescue ActiveRecord::RecordInvalid => e
-      delivor_error_webhook(error: e.record.errors.messages)
+      deliver_error_webhook(error: e.record.errors.messages)
 
       result
     rescue ActiveRecord::RecordNotUnique
-      delivor_error_webhook(error: {transaction_id: ['value_already_exist']})
+      deliver_error_webhook(error: {transaction_id: ['value_already_exist']})
 
       result
     end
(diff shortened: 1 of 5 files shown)


brunomiguelpinto (dismissed):

app/services/events/post_process_service.rb
```diff
@@ -97,16 +97,16 @@ def expire_cached_charges(subscriptions)
     def handle_pay_in_advance
       return unless billable_metric
 
-      charges.where(invoiceable: false).find_each do |charge|
-        Fees::CreatePayInAdvanceJob.perform_later(charge:, event:)
-      end
-
       # NOTE: ensure event is processable
       processable_event = billable_metric.count_agg? ||
```
julienbourdeau, line 101: We might want to add `unique_count_agg?` here 🤔

julienbourdeau: `unique_count_agg` needs the `event.properties[billable_metric.field_name]` to be set, that's why it's not here!!±

vincent-pochet (approved):
:rocket:

PR #2153, opened 2024-06-10
misc(billable_metric): preload filters and groups in index route

vincent-pochet:
## Context

The billable metric index route appears to be slow in API in some cases.

## Description

This PR tries to speed it up a bit by preloading attached `filters` and `groups` . 

NOTE: Billable metric serializer is also exposing three counters: (`active_subscriptions_count`, `draft_invoices_count` and `        plans_count`), that are slowing down this route. A discussion will be opened about it soon.

### DIFF
--- a/app/models/billable_metric.rb
+++ b/app/models/billable_metric.rb
@@ -60,7 +60,7 @@ class BillableMetric < ApplicationRecord
   end
 
   def active_groups
-    scope = groups.order(created_at: :asc)
+    scope = groups
     scope = scope.with_discarded if discarded?
     scope
   end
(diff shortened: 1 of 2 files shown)


app/models/billable_metric.rb
```diff
@@ -60,7 +60,7 @@ def aggregation_type=(value)
   end
 
   def active_groups
-    scope = groups.order(created_at: :asc)
+    scope = groups
```
vincent-pochet, line 63: Note: This change as no impact anymore on the application as groups are not used anywhere in the app.

julienbourdeau (approved):

PR #2160, opened 2024-06-10
feat(xero): Add models

ivannovosad:
## Roadmap Task

## Context

Currently Lago does not support Xero integration.

## Description

This PR adds mapping and collection mapping models for Xero integration.


spec/models/integration_collection_mappings/netsuite_collection_mapping_spec.rb
```diff
@@ -6,7 +6,7 @@
   subject(:mapping) { build(:netsuite_collection_mapping) }
 
   let(:mapping_types) do
-    %i[fallback_item coupon subscription_fee minimum_commitment tax prepaid_credit credit_note]
+    %i[fallback_item coupon subscription_fee minimum_commitment tax prepaid_credit credit_note account]
```
ancorcruz, line 9: why did you change `netsuite` mapping types? as the feature is about adding a new integration for Xero I wouldn't expect other integrations to change.

spec/models/integration_collection_mappings/xero_collection_mapping_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  it { is_expected.to define_enum_for(:mapping_type).with_values(mapping_types) }
+
+  describe '#external_id' do
+    let(:external_id) { SecureRandom.uuid }
+
+    it 'assigns and retrieve a setting' do
+      mapping.external_id = external_id
+      expect(mapping.external_id).to eq(external_id)
+    end
+  end
```
ancorcruz, line 21: I don't quite understand why this test is needed. It tests that the active record model has an attribute and knows how to set and get its value, that is it.

The same for the rest of the spec file and the xero_mapping_spec file below.

ancorcruz (approved):
I left a few questions about the code, otherwise looks good to me.

spec/models/integration_collection_mappings/netsuite_collection_mapping_spec.rb
ivannovosad: Because this example fails otherwise https://github.com/getlago/lago-api/pull/2160/files/3295acbb1ddf92507a094d78246577df3db169ad#diff-5be09cab6e233e232c33fa3e68fc77d9d3d7f266d41062841f0505d3a83b8f3eL12

I'm moving it to spec/models/integration_collection_mappings/base_collection_mapping_spec.rb

spec/models/integration_collection_mappings/xero_collection_mapping_spec.rb
ivannovosad: Yes, but it's a jsonb attribute defined by `settings_accessors :external_id` so it tests if it is defined.

ancorcruz (approved):
:shipit:

PR #2161, opened 2024-06-10
Fix exception handling

cstar:
## Roadmap Task

## Context

Clock is broken in exception handling

## Description

Appropriate naming of variables


julienbourdeau (approved):
Looking at the blame, it seems that it's been this way for a while 😬

PR #2169, opened 2024-06-12
fix(subscriptions): Fix incorrect `error_details` when creating a subscription without `external_customer_id`

ancorcruz:
## Context

When a user tries to create a subscription via API without providing external_customer_id, the API returns a 422 error.

This is the expected behaviour, but the reason is incorrect: it says "external_id": ["value is mandatory"] instead of "external_customer_id": ["value is mandatory"].

## Description

Add guard clause to `Subscriptions::CreateService` to ensure `external_customer_id` is provider or return a validation error with the right details as result.


jdenquin (approved):

jdenquin:
LGTM :rocket: Let's wait for a second Review and go!

julienbourdeau (approved):

PR #2170, opened 2024-06-12
chore(segment): Dedup track_invoice_created function

julienbourdeau:
## Description

I was about to copy and paste the definition of `track_invoice_created(invoice)` for a feature I'm working on.

I realized it's already defined 6 times, the exact same way.

I'd like to extract this to a helper, what do you think? I also refactored `track_refund_status_changed` just to see how it files to put these functions into a helper Module.


app/services/credit_notes/refunds/adyen_service.rb
```diff
@@ -30,7 +30,7 @@ def create
         refund.save!
 
         update_credit_note_status(refund.status)
-        track_refund_status_changed(refund.status)
+        Utils::Track.refund_status_changed(refund.status, credit_note.id, organization.id)
```
rsempe, line 33:
```suggestion
        Utils::SegmentTrack.refund_status_changed(refund.status, credit_note.id, organization.id)
```

rsempe: what about using named parameters as well?

rsempe (approved):

app/services/credit_notes/refunds/adyen_service.rb
julienbourdeau: I renamed the class.

I'm not sure about named arguments here, I find that it adds a lot fo noise:
```rb
something(refund.status, credit_note.id, organization.id)
# VS
something(status: refund.status, credit_note_id: credit_note.id, organization_id: organization.id
```

PR #2172, opened 2024-06-13
chore(invoices): Remove duplicated invoice amounts computations

julienbourdeau:
## Description

When calculating fees with `Invoices:: CalculateFeesService` we sum all fees to add the total amount to the invoice.

It turned out that this computation is done in `Invoices::ComputeAmountsFromFees` line 49.

https://github.com/getlago/lago-api/blob/df93e6fe75a860f9fbf482d6211fa18c9a7dec52/app/services/invoices/compute_amounts_from_fees.rb#L19-L23

### DIFF
--- a/app/services/invoices/calculate_fees_service.rb
+++ b/app/services/invoices/calculate_fees_service.rb
@@ -41,10 +41,6 @@ module Invoices
           end
         end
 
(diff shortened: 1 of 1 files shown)


julienbourdeau:
It turned out `should_create_coupon_credit?` relies on `invoice.fees_amount_cents` so it needs to be set before. It broke enough that this change is not worth it.

PR #2175, opened 2024-06-13
feat(integrations): support array of integration customers

lovrocolic:
## Context

Currently Lago app is being extended with accounting and tax integrations

## Description

For accounting integrations, we could have only **one** integration customer per customer create/update request.

However, with tax integrations, Lago should support multiple integration customers per request. That being said, Lago should be able to create integration customer object related to tax provider and integration customer object for accounting provider.

This PR handles described scenario and supports array of integration customers from both GQL and API.

This update is NOT a breaking change since released accounting integration is premium add-on and is not being used in production yet.

### DIFF
--- a/app/services/customers/create_service.rb
+++ b/app/services/customers/create_service.rb
@@ -188,9 +188,7 @@ module Customers
 
       input_types = integration_customers&.map { |c| c.to_h.deep_symbolize_keys }&.map { |c| c[:integration_type] }
 
-      return true if input_types.length == input_types.uniq.length
-
-      false
+      input_types.length == input_types.uniq.length
     end
 
     def create_metadata(customer:, args:)
--- a/app/services/customers/create_service.rb
+++ b/app/services/customers/create_service.rb
@@ -75,7 +75,7 @@ module Customers
       result.customer = customer.reload
 
       IntegrationCustomers::CreateOrUpdateService.call(
-        integration_customer_params: params[:integration_customer],
+        integration_customers: params[:integration_customers],
         customer: result.customer,
         new_customer:
       )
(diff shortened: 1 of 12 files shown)


spec/services/integration_customers/create_or_update_service_spec.rb
```diff
@@ -81,6 +83,31 @@
         expect { service_call }.to have_enqueued_job(IntegrationCustomers::CreateJob)
       end
 
+      context 'with multiple integration customers' do
+        let(:integration_anrok) { create(:anrok_integration, organization:) }
+        let(:integration_customers) do
```
ivannovosad, line 88: What happens when somebody tries to add multiple netsuite customers at the same time? Should we add a validation to prevent that?

lovrocolic: We have `validates :customer_id, uniqueness: {scope: :type}` this validation on model level and DB level (index unique validation) so I think we are safe. From UI it is not possible to send multiple objects of the same type. From the API it is possible but it should result with error.

app/services/customers/create_service.rb
```diff
@@ -176,6 +183,16 @@ def valid_metadata_count?(metadata:)
...
     end
 
+    def valid_integration_customers_count?(integration_customers:)
+      return true if integration_customers.blank?
+
+      input_types = integration_customers&.map { |c| c.to_h.deep_symbolize_keys }&.map { |c| c[:integration_type] }
+
+      return true if input_types.length == input_types.uniq.length
+
+      false
```
ivannovosad, line 193:
```suggestion
      input_types.length == input_types.uniq.length
```

ivannovosad (approved):

PR #2179, opened 2024-06-14
misc: move InvoicesQuery params as filters

ancorcruz:
## Context

Before adding more filters to the  `InvoicesQuery` object and making harder to work with it, this change moves the existing filters to the filters hash.

## Description

Move existing filters to filters hash in `InvoicesQuery` object
Add some documentation to make easier to discover the available filters in the query object.

### DIFF
--- a/app/queries/invoices_query.rb
+++ b/app/queries/invoices_query.rb
@@ -1,20 +1,20 @@
 # frozen_string_literal: true
 
 class InvoicesQuery < BaseQuery
-  def call(search_term:, status:, page:, limit:, filters: {}, customer_id: nil, payment_status: nil, payment_dispute_lost: nil, payment_overdue: nil) # rubocop:disable Metrics/ParameterLists
+  def call(search_term:, page:, limit:, filters: {}, payment_overdue: nil)
     @search_term = search_term
-    @customer_id = customer_id
+    @customer_id = filters[:customer_id]
     @filters = filters
 
     invoices = base_scope.result.includes(:customer)
     invoices = invoices.where(currency: filters[:currency]) if filters[:currency]
     invoices = with_customer_external_id(invoices) if filters[:customer_external_id]
-    invoices = invoices.where(customer_id:) if customer_id.present?
+    invoices = invoices.where(customer_id: filters[:customer_id]) if filters[:customer_id].present?
     invoices = invoices.where(invoice_type: filters[:invoice_type]) if filters[:invoice_type]
     invoices = with_issuing_date_range(invoices) if filters[:issuing_date_from] || filters[:issuing_date_to]
-    invoices = invoices.where(status:) if status.present?
-    invoices = invoices.where(payment_status:) if payment_status.present?
-    invoices = invoices.where.not(payment_dispute_lost_at: nil) unless payment_dispute_lost.nil?
+    invoices = invoices.where(status: filters[:status]) if filters[:status].present?
+    invoices = invoices.where(payment_status: filters[:payment_status]) if filters[:payment_status].present?
+    invoices = invoices.where.not(payment_dispute_lost_at: nil) if filters[:payment_dispute_lost]
     invoices = invoices.where(payment_overdue:) if payment_overdue.present?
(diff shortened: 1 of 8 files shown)


app/queries/invoices_query.rb
```diff
@@ -1,16 +1,51 @@
...
+  # @param limit [Integer] the number of invoices per page
+  # @param filters [Hash] the filters to apply to the query
+  # @option filters [String] :customer_id the ID of the customer to filter by
+  # @option filters [Array<String>] :ids an array of invoice IDs to filter by
+  # @option filters [String] :status the status of invoices to filter by
+  # @option filters [String] :payment_status the payment status of invoices to filter by
+  # @option filters [Boolean] :payment_dispute_lost whether to include invoices with lost payment disputes
+  #
+  # @return [Result] the result of the query with filtered invoices
+  def call(search_term:, page:, limit:, filters: {})
```
vincent-pochet, line 39: Maybe we could use it as an opportunity to migrate this query to the new format we are trying to push.

See https://github.com/getlago/lago-api/blob/main/app/controllers/api/v1/subscriptions_controller.rb#L91 for an example.

Regarding the documentation comment, I'm not sure this brings that much value to the code. Those classes are not intended to be used in external/public application, and I think the code should be self describing as much as possible.
For me I think having tons of comments brings more problems than it solves because the cost to maintain it up to date is usually high.

Since it is just my own opinion, maybe that is something that we should discuss with the team

ancorcruz: @vincent-pochet thanks for the heads up with the new format.

I was wondering why not passing the filters as a hash to the Query object then instantiate the Filter object from the Query one. IMO the controller in the example knows to much as BaseQuery::Filters feels like an internal component of the query objects.

This is what I have in mind...

```ruby
  # controller
      def index
        result = SubscriptionsQuery.call(
          organization: current_organization,
          pagination: BaseQuery::Pagination.new(
            page: params[:page],
            limit: params[:per_page] || PER_PAGE
          ),
          filters: index_filters      # this is a hash, controller does not care about how the Filters object is built.
        )

        #...
      end

  # BaseQuery object
  def initialize(organization:, pagination: Pagination.new, filters: {})
    @organization = organization
    @pagination = pagination
    @filters = Filters.new(filters)

    super
  end
```
  
I can take it and do the required updates across controllers/resolvers/query objects if we agree.

What do you think? @getlago/back

ancorcruz: ah, the same could be applied for the pagination object.

vincent-pochet: yes it makes sense!

julienbourdeau: Personally I agree with Vincent on documentation, I find that it's often outdated and unmaintained.

About the hash/Filter object, I like that it keeps the signature small but actually, you have no idea what arguments are available unless you dig into the method, whereas named argument give you a good hints of what's available.

I think we have similar issue with some services, [where you get get `**args` as a parameter](https://github.com/getlago/lago-api/blob/82b37840d5d8cd36d6112e5bba879a98c9f74c71/app/services/invites/create_service.rb#L5).

I kinda like how the graphql part does it with input objects but it's very verbose.

Similar to what you're proposing, each Query could (should?) define their query object.

```ruby
# BaseQuery object
class BaseQuery
  def initialize(organization:, pagination: {}, filters: {})
    @organization = organization
    @pagination = pagination # same with pagination
    @filters = make_filters(filters)

    super
  end

  private

  def default_filters
    {}
  end

  def make_filters(filters)
    raise 'Invalid filters' unless valid_filters?(filters)
    Filters.new(filters)
  end
end

# InvoicesQuery object
class InvoicesQuery < BaseQuery
  def call
    # ...
  end

  private

  def default_filters
    {
      start_date: some_default_value_if_needed,
    }
  end

  def valid_filters?(filters)
    format = {
      customer_id: ID,
      start_date: Time,
      customer_name: String,
    }
    validate_hash_magically(filters, format)
    
    # or something like this :D 
    # filters.keys.sort == valid_keys.sort
  end
end
```

PR #2180, opened 2024-06-14
feat (anrok-integration): add support for connection_id

lovrocolic:
## Context

Currently Lago does not support tax integrations

## Description

When Lago create integration object, `connection_id` also needs to be stored. This is identifier in Nango (integrations aggregator) and it will be used in all requests from Lago's BE.


vincent-pochet (approved):

spec/models/integrations/anrok_integration_spec.rb
```diff
@@ -19,4 +19,11 @@
       expect(anrok_integration.api_key).to eq('123abc456')
     end
   end
+
+  describe '.connection_id' do
```
ivannovosad, line 23: We can add a spec to validate its presence as well.

ivannovosad (approved):

PR #2181, opened 2024-06-14
fix(service): Use call_later for RetryAllJob and raise if error

julienbourdeau:
## Description

Insignificant but `result.throw_error` doesn't exist.

@ivannovosad and @vincent-pochet recently introduced `call_async` in services so why not use it instead of `call_later`.


vincent-pochet (approved):
:rocket:

ivannovosad (approved):

ancorcruz (approved):

PR #2182, opened 2024-06-17
misc(integrations): Remove connection_id from Integration types

ivannovosad:
## Context

connectionId is not needed to be exposed to the UI after a connection is created or updated.

## Description

This PR removes `connectionId` from `NetsuiteIntegration` and `XeroIntegration`


ansmonjol:
So, is that PR still relevant. IIRC this attribute is still needed for mapping part

ivannovosad:
@lovrocolic Please confirm if we can close this.

PR #2184, opened 2024-06-17
misc: extend redis timeout property

lovrocolic:
## Context

Lago uses Redis for storing sidekiq jobs and its operational data.

## Description

In order to avoid `RedisClient::ReadTimeoutError` we can increase `redis` timeout property. By default our redis version sets this property to `1s`.

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -3,6 +3,7 @@
 redis_config = {
   url: ENV['REDIS_URL'],
   pool_timeout: 5,
+  timeout: 5,
   ssl_params: {
     verify_mode: OpenSSL::SSL::VERIFY_NONE
   }


jdenquin (approved):

julienbourdeau:
5 sec seems like a very long time to read from Redis. I think here we want it to be fairly short so it doesn't hang for too long 🤔

How often do we get the `RedisClient::ReadTimeoutError`? Maybe there is some connectivity issue and instead of dropping and getting this error after 1 sec, we'll now have it after 5 sec.

lovrocolic:
> 5 sec seems like a very long time to read from Redis. I think here we want it to be fairly short so it doesn't hang for too long 🤔
> 
> How often do we get the `RedisClient::ReadTimeoutError`? Maybe there is some connectivity issue and instead of dropping and getting this error after 1 sec, we'll now have it after 5 sec.

`RedisClient::ReadTimeoutError` is received when there is delay in answer from redis server and shouldn't be related to connection itself. For connection pool we already set it to 5s. As I can see, we receive this error from time to time and it seemed fine to me to set it to 5s, since that was the default value of the previous redis version.

ancorcruz (approved):

jdenquin:
> 5 sec seems like a very long time to read from Redis. I think here we want it to be fairly short so it doesn't hang for too long 🤔
> 
> How often do we get the `RedisClient::ReadTimeoutError`? Maybe there is some connectivity issue and instead of dropping and getting this error after 1 sec, we'll now have it after 5 sec.

we get it when we have an high load sometimes, 1s is pretty low, there is no connectivity issue at all monitored when it happen, just a higher load on Redis

PR #2186, opened 2024-06-18
misc(sidekiq): Add new monitoring system for k8s liveness probe

jdenquin:
## Description

- `sidekiqmon` cause a lot of issues depending on Redis configuration
- Uses a very simple approach to check Sidekiq to Redis networking


vincent-pochet (approved):

julienbourdeau (approved):
😳 😁

PR #2190, opened 2024-06-18
fix (draft- invoices): refresh draft invoice only if there is active subscription

lovrocolic:
## Context

Prevent refreshing draft invoices which don't have at least one active subscription

## Description

If invoice is in draft status but subscriptions related to it are terminated, we don't want to refresh such an invoice each 5 minutes

### DIFF
--- a/app/jobs/clock/refresh_draft_invoices_job.rb
+++ b/app/jobs/clock/refresh_draft_invoices_job.rb
@@ -5,9 +5,7 @@ module Clock
     queue_as 'clock'
 
     def perform
-      Invoice.ready_to_be_refreshed.find_each do |invoice|
-        next unless invoice.subscriptions.pluck(:status).include?('active')
-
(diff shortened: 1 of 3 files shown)


app/jobs/clock/refresh_draft_invoices_job.rb
```diff
@@ -6,6 +6,8 @@ class RefreshDraftInvoicesJob < ApplicationJob
 
     def perform
       Invoice.ready_to_be_refreshed.find_each do |invoice|
```
vincent-pochet, line 8: Maybe we should check the condition directly in a DB query. Since we could have a lot of invoice to refresh, we might want to avoid performing one query per invoice. What do you think?

lovrocolic: Makes sense! Updated

vincent-pochet (approved):

PR #2191, opened 2024-06-19
fix(draft-invoice): fix invoice reason logic for draft invoices

lovrocolic:
## Context

When refreshing invoice, that is result of the upgrade operation, we didn't calculate correctly invoice reason. If there are multiple subscriptions, old logic could lead to fetching invalid reason and consequently to the DB index uniqueness error.

## Description

This PR handles described scenario and adds related tests.


jdenquin (approved):

vincent-pochet (approved):
:rocket:

PR #2194, opened 2024-06-20
fix(integration-customers): fix incorrectly cached variables

lovrocolic:
## Context

Lago is currently adding more integrations; accounting and tax ones

## Description

Integration customer logic was originally written for one `integration_customer` object per `customer`. Recently the logic changed so that we support array in the input side.

However, we haven't changed memoization logic. The issue was that only the first object got cached and all subsequent params hashes got linked to the invalid `@integration` and `@integration_customer` instance variables.

This PR fixes described problem.

### DIFF
--- a/app/services/integration_customers/create_or_update_service.rb
+++ b/app/services/integration_customers/create_or_update_service.rb
@@ -54,6 +54,7 @@ module IntegrationCustomers
     def remove_integration_customer?
       !new_customer &&
         integration_customer &&
+        integration_customer.external_customer_id &&
         integration_customer_params[:external_customer_id].blank?
     end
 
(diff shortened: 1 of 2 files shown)


app/services/integration_customers/create_or_update_service.rb
```diff
@@ -63,19 +64,29 @@ def skip_creating_integration_customer?
     end
 
     def integration
-      return @integration if defined? @integration
+      type = Integrations::BaseIntegration.integration_type(integration_customer_params[:integration_type])
+
+      if defined? @integration
+        return @integration if @integration&.type == type
+      end
```
ivannovosad, line 71:
```suggestion
      return @integration if defined?(@integration) && @integration&.type == type
```

app/services/integration_customers/create_or_update_service.rb
```diff
@@ -63,19 +64,29 @@ def skip_creating_integration_customer?
...
       @integration = Integrations::BaseIntegration.find_by(type:, code:)
     end
 
     def integration_customer
-      @integration_customer ||= IntegrationCustomers::BaseCustomer.find_by(integration:, customer:)
+      type = IntegrationCustomers::BaseCustomer.customer_type(integration_customer_params[:integration_type])
+
+      if defined? @integration_customer
+        return @integration_customer if @integration_customer&.type == type
+      end
```
ivannovosad, line 87:
```suggestion
      return @integration_customer if defined?(@integration_customer) && @integration_customer&.type == type
```

ivannovosad (changes requested):

ivannovosad (approved):
There's a rubocop warning, but I'm approving it.

PR #2204, opened 2024-06-24
feat(lib): Move VersionService to LagoUtils and initialize at boot time

julienbourdeau:
## Description

wip


lib/lago_utils/lago_utils/version.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module LagoUtils
+  class Version
+    VERSION_FILE = Rails.root.join('LAGO_VERSION')
```
julienbourdeau, line 5: I'd like to avoid using `Rails` in the `lib` folder, even if we already [do it for VAT Rate](https://github.com/getlago/lago-api/blob/ccb17cb7e1189e51a676702166a0af5d1e2fc44d/lib/lago_eu_vat/lago_eu_vat/rate.rb#L6).

vincent-pochet (approved):

PR #2206, opened 2024-06-24
feat(export invoices): Extend InvoicesQuery with new filters

ancorcruz:
## Roadmap Task

## Context

## Description

This change includes new filters to query invoices by:

* invoice_type
* currency
* issuing_date (from and to range)
* customer external_id

also add database indices for all of them for performance reasons.

and drop support for filtering by ids as this filter is not in use.


db/migrate/20240624115536_add_indices_to_invoices.rb
```diff
@@ -0,0 +1,10 @@
+class AddIndicesToInvoices < ActiveRecord::Migration[7.0]
+  def change
+    add_index :invoices, :currency
+    add_index :invoices, :issuing_date
+    add_index :invoices, :invoice_type
+    add_index :invoices, :status
+    add_index :invoices, :payment_status
+    add_index :invoices, :payment_dispute_lost_at
```
julienbourdeau, line 8: I'm not sure these index would be used often if we combine conditions, which I think would be pretty often 🤔

ancorcruz: I was expecting that postgres will use one of the indices or use Bitmap Index Scans to combine them. While having a combined index is more performant that option is not feasible for our query filters.

db/migrate/20240624115536_add_indices_to_invoices.rb
```diff
@@ -0,0 +1,10 @@
+class AddIndicesToInvoices < ActiveRecord::Migration[7.0]
+  def change
+    add_index :invoices, :currency
+    add_index :invoices, :issuing_date
+    add_index :invoices, :invoice_type
+    add_index :invoices, :status
+    add_index :invoices, :payment_status
```
vincent-pochet, line 7: Indexes on `*status` should be avoided I think, the cardinality is very low so it is not really efficient but it could make senses to combine them with an ID field lie `organization_id` or `customer_id` depending on context.

spec/queries/invoices_query_spec.rb
```diff
@@ -39,12 +41,14 @@
...
       customer: customer_first,
-      number: '3333333333'
+      number: '3333333333',
+      issuing_date: 3.weeks.ago
     )
   end
   let(:invoice_fourth) do
     create(
       :invoice,
+      :usd,
```
vincent-pochet, line 51: What about setting the currency, rather than using a trait here?

db/migrate/20240624115536_add_indices_to_invoices.rb
ancorcruz: I'm dropping all the new indices.

vincent-pochet (approved):

PR #2207, opened 2024-06-25
feat(xero): Add payments services

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds services and payloads for Xero payments. It also modifies some of the existing ones.


app/services/integrations/aggregator/payments/create_service.rb
```diff
@@ -38,7 +46,7 @@ def call
 
           deliver_error_webhook(customer:, code:, message:)
 
-          raise e
+          raise e if e.error_code.to_i >= 500
```
vincent-pochet, line 49: Maybe we should make sure to return a result (empty, I guess). What do you think?
```suggestion
          return result if e.error_code.to_i < 500
          
          raise e
```

vincent-pochet (approved):

jdenquin (approved):

app/services/integrations/aggregator/payments/payloads/anrok.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
```
lovrocolic, line 1: This class is not needed for Anrok

ivannovosad: I'm removing it.

lovrocolic (approved):

PR #2217, opened 2024-06-26
feat(anrok-integration): add support for customer shipping address on the API side

lovrocolic:
## Context

Currently Lago is adding integration with Tax provider - Anrok.

## Description

In some countries taxes can differ based on where the product is used and not based on customer's main location.
That's why shipping address is introduced and this PR adds support for it on the API side.

### DIFF
--- a/app/serializers/v1/customer_serializer.rb
+++ b/app/serializers/v1/customer_serializer.rb
@@ -29,7 +29,8 @@ module V1
         applicable_timezone: model.applicable_timezone,
         net_payment_term: model.net_payment_term,
         external_salesforce_id: model.external_salesforce_id,
-        billing_configuration:
+        billing_configuration:,
+        shipping_address:
       }.merge(legacy_values.except(:billing_configuration))
 
       payload = payload.merge(metadata)
@@ -73,6 +74,17 @@ module V1
       end
 
       configuration
+    end
+
+    def shipping_address
+      {
+        address_line1: model.shipping_address_line1,
+        address_line2: model.shipping_address_line2,
+        city: model.shipping_city,
+        zipcode: model.shipping_zipcode,
+        state: model.shipping_state,
+        country: model.shipping_country
+      }
     end
 
     def legacy_values
--- a/app/serializers/v1/customer_serializer.rb
+++ b/app/serializers/v1/customer_serializer.rb
@@ -30,7 +30,7 @@ module V1
         net_payment_term: model.net_payment_term,
         external_salesforce_id: model.external_salesforce_id,
         billing_configuration:,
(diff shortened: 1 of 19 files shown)


app/services/customers/create_service.rb
```diff
@@ -7,6 +7,7 @@ class CreateService < BaseService
     def create_from_api(organization:, params:)
       customer = organization.customers.find_or_initialize_by(external_id: params[:external_id])
       new_customer = customer.new_record?
+      shipping_address = params.key?(:shipping_address) ? params[:shipping_address] : {}
```
ancorcruz, line 10: would you consider this shorter and concise way?

```suggestion
      shipping_address = params[:shipping_address] || {}
```

app/serializers/v1/customer_serializer.rb
```diff
@@ -75,6 +76,17 @@ def billing_configuration
       configuration
     end
 
+    def shipping_address
```
vincent-pochet, line 79: Could you add a test  for this new field :pray: :wink:

vincent-pochet (approved):

db/migrate/20240626094521_add_shipping_address_to_customers.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddShippingAddressToCustomers < ActiveRecord::Migration[7.1]
+  change_table :customers, bulk: true do |t|
```
ancorcruz, line 4: Could customers have more than one shipping address?

Is there any chance a customer change their address in the middle of an invoice period or something and the taxes are not properly applied?

app/services/customers/create_service.rb
ancorcruz, line 10: would you consider this shorter and concise way?

```suggestion
      shipping_address = params[:shipping_address] || {}
```

ancorcruz (approved):
Looks good.

app/serializers/v1/customer_serializer.rb
lovrocolic: done

app/services/customers/create_service.rb
lovrocolic: updated

db/migrate/20240626094521_add_shipping_address_to_customers.rb
lovrocolic: Based on current requirements, customer can have only one shipping address.

Sure, if shipping address got updated, there is a chance that fetched taxes would also changed. However, each organization is responsible for it and I am not sure we should have any extra validation here

PR #2218, opened 2024-06-26
feat(export invoices): Add GraphQL invoices filters

ancorcruz:
## Roadmap Task

## Context

## Description

This change includes new filters to the GraphQL invoices resolver to query invoices by:

* invoice_type
* currency
* issuing_date (from and to range)
* customer external_id

Also drops support for filtering by ids as this filter is not in use.


julienbourdeau (approved):

vincent-pochet (approved):
:rocket:

PR #2220, opened 2024-06-26
feat(anrok-integrations): add gql support for shipping address

lovrocolic:
## Context

Currently Lago is adding integration with Tax provider - Anrok.

## Description

In some countries taxes can differ based on where the product is used and not based on customer's main location.
That's why shipping address is introduced and this PR adds support for it on the GQL side.

### DIFF
--- a/app/graphql/types/customers/address.rb
+++ b/app/graphql/types/customers/address.rb
@@ -7,9 +7,9 @@ module Types
 
       field :address_line1, String, null: true
       field :address_line2, String, null: true
+      field :city, String, null: true
       field :country, Types::CountryCodeEnum, null: true
       field :state, String, null: true
-      field :city, String, null: true
       field :zipcode, String, null: true
     end
   end
--- a/app/graphql/types/customers/address_input.rb
+++ b/app/graphql/types/customers/address_input.rb
@@ -7,9 +7,9 @@ module Types
 
       argument :address_line1, String, required: false
       argument :address_line2, String, required: false
+      argument :city, String, required: false
       argument :country, Types::CountryCodeEnum, required: false
       argument :state, String, required: false
-      argument :city, String, required: false
       argument :zipcode, String, required: false
     end
   end
--- a/schema.json
+++ b/schema.json
@@ -12477,6 +12477,18 @@
               "deprecationReason": null
             },
             {
+              "name": "city",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "defaultValue": null,
+              "isDeprecated": false,
+              "deprecationReason": null
+            },
+            {
               "name": "country",
               "description": null,
               "type": {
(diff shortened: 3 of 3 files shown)


app/graphql/types/customers/object.rb
```diff
@@ -124,6 +126,17 @@ def billing_configuration
           document_locale: object&.document_locale
         }
       end
+
+      def shipping_address
```
vincent-pochet, line 130: Maybe this part could be extracted into a method on the customer model as it seems it will be also used on the serializer side

app/graphql/types/customers/shipping_address_input.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Types
+  module Customers
+    class ShippingAddressInput < BaseInputObject
```
vincent-pochet, line 5: Same here, what about an `AddressInput` type?

app/graphql/types/customers/shipping_address.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Types
+  module Customers
+    class ShippingAddress < Types::BaseObject
```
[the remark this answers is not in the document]
lovrocolic: Changed, using address interface now

app/graphql/types/customers/object.rb
lovrocolic: Moved on the model

app/graphql/types/customers/shipping_address_input.rb
lovrocolic: TBH, I haven't found approach like gql interface for the input type. Did you have anything specific in mind?

vincent-pochet: My point was to ask, do we need a specific `ShippingAddress`? It could just be an `Address` that could be reused in multiple places.

app/graphql/types/customers/object.rb
```diff
@@ -28,11 +27,11 @@ class Object < Types::BaseObject
       field :payment_provider, Types::PaymentProviders::ProviderTypeEnum, null: true
       field :payment_provider_code, String, null: true
       field :phone, String, null: true
-      field :state, String, null: true
       field :tax_identification_number, String, null: true
       field :timezone, Types::TimezoneEnum, null: true
       field :url, String, null: true
-      field :zipcode, String, null: true
+
+      field :shipping_address, Types::Customers::ShippingAddress, null: true
```
vincent-pochet, line 34: In my mind it was something like:

```suggestion
      field :shipping_address, Types::Customers::Address, null: true
```

But not sure it could be used easily for the base address

app/graphql/types/customers/shipping_address_input.rb
lovrocolic: Updated

vincent-pochet (approved):

PR #2109, opened 2024-05-31
misc: Add scheduler monitoring for sentry if enabled

vincent-pochet:
## NOTE:
This PR is extracted from https://github.com/getlago/lago-api/pull/1780 in order to only apply the sentry related logic

## Context

Sentry scheduled tasks can easily be monitored by Sentry crons checkin functions.
This allows us to know fairly immediately whether a sidekiq task has stopped processing.

## Description

ENV variable "SENTRY_ENABLE_CRONS" will activate a mixin cron checkin for sentry within the activejobs.
These mixins are preconfigured to match the clockwork functions found in clock.rb

No customer installation changes required.

Dependencies for sentry have been lifted to:

```diff
-gem 'sentry-rails', '~> 5.12.0'
-gem 'sentry-ruby', '~> 5.12.0'
-gem 'sentry-sidekiq', '~> 5.12.0'
+gem 'sentry-rails', '~> 5.17.0'
+gem 'sentry-ruby', '~> 5.17.0'
+gem 'sentry-sidekiq',  '~> 5.17.0'
```

PR #2225, opened 2024-06-27
misc(sidekiq): New approach for cron monitoring

vincent-pochet:
## Context

This PR is a new approach to try to finally add cron monitoring on Sidekiq.
The logic was first introduced in https://github.com/getlago/lago-api/pull/2109 and https://github.com/getlago/lago-api/pull/1780

## Description

The new approach uses `ActiveJob.set` to pass cron arguments to the job, in order to avoid the issues we faced lately with the initial approach

### DIFF
--- a/app/jobs/clock/activate_subscriptions_job.rb
+++ b/app/jobs/clock/activate_subscriptions_job.rb
@@ -2,6 +2,8 @@
 
 module Clock
   class ActivateSubscriptionsJob < ApplicationJob
+    include SentryCronConcern
+
     queue_as 'clock'
 
     unique :until_executed, on_conflict: :log
--- a/app/jobs/clock/create_interval_wallet_transactions_job.rb
+++ b/app/jobs/clock/create_interval_wallet_transactions_job.rb
@@ -2,6 +2,8 @@
 
 module Clock
   class CreateIntervalWalletTransactionsJob < ApplicationJob
+    include SentryCronConcern
+
     queue_as 'clock'
 
     def perform
--- a/app/jobs/clock/events_validation_job.rb
+++ b/app/jobs/clock/events_validation_job.rb
@@ -2,6 +2,8 @@
 
 module Clock
   class EventsValidationJob < ApplicationJob
+    include SentryCronConcern
+
     queue_as 'clock'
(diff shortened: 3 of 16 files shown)


julienbourdeau:
Just some food for thoughts, since these jobs are all in the `Clock` module and they are only called via `clock.rb`, we could store the sentry config directly in the job instead of using the class variable.
It could be a method or some constants.

```diff
module Clock
  class ActivateSubscriptionsJob < ApplicationJob
    include SentryCronConcern
    queue_as 'clock'

    unique :until_executed, on_conflict: :log

+    SENTRY_SLUG = :lago_activate_subscriptions
+    SENTRY_SCHEDULE = '*/5 * * * *'

    def perform
      Subscriptions::ActivateService.new(timestamp: Time.current.to_i).activate_all_pending
    end
    
+    def sentry
+      {slug: 'lago_activate_subscriptions', cron: '*/5 * * * *'}
+    end
  end
end
```

The drawback is that if you call the job manually, it would mess with the monitoring data 📊

julienbourdeau (approved):

vincent-pochet:
> Just some food for thoughts, since these jobs are all in the `Clock` module and they are only called via `clock.rb`, we could store the sentry config directly in the job instead of using the class variable. It could be a method or some constants.
> 
> ```diff
> module Clock
>   class ActivateSubscriptionsJob < ApplicationJob
>     include SentryCronConcern
>     queue_as 'clock'
> 
>     unique :until_executed, on_conflict: :log
> 
> +    SENTRY_SLUG = :lago_activate_subscriptions
> +    SENTRY_SCHEDULE = '*/5 * * * *'
> 
>     def perform
>       Subscriptions::ActivateService.new(timestamp: Time.current.to_i).activate_all_pending
>     end
>     
> +    def sentry
> +      {slug: 'lago_activate_subscriptions', cron: '*/5 * * * *'}
> +    end
>   end
> end
> ```
> 
> The drawback is that if you call the job manually, it would mess with the monitoring data 📊

There is also a risk of de-sync with what is configured in the clock file itself. The risk still exists in the current approach but at least it done in the same place

PR #2226, opened 2024-06-27
feat(export invoices): Create data export

ancorcruz:
## Roadmap Task

## Context

## Description

This change modifies the data_exports database table adding references to memberships and organizations (drops reference to users) to ensure we can identify the user that initiates the export request but also the organization we are exporting the data from.

Also, this introduces the mailer to notify the user the export is ready, a service to create to create the data export record and queue the job to process the export, and enables action mailer previews to try emails in local in a more convenient way opening the email in the url: https://api.lago.dev/rails/mailers/data_export_mailer/completed.html?locale=en

Monosnap Mailer Preview for data_export_mailer#completed 2024-07-01 09-16-29


app/jobs/data_exports/export_resources_job.rb
```diff
@@ -0,0 +1,9 @@
+module DataExports
+  class ExportResourcesJob < ApplicationJob
+    queue_as :data_export
```
julienbourdeau, line 3: Does this queue already exists?

julienbourdeau (approved):

app/jobs/data_exports/export_resources_job.rb
ancorcruz: no, this is a new queue.

I would not mix this kind of jobs with other jobs as the time building the files could be long enough to add delays to the queue processing.

PR #2239, opened 2024-07-02
feat: Add Discard Handling for AuthenticationError

brunomiguelpinto:
### Context:
In this pull request, we change the error handling within the StripeCreateJob by discarding jobs that encounter Stripe::AuthenticationError. This change ensures that we prevent creating noise 

### Description:
- When a `Stripe::AuthenticationError` is raised, it indicates an issue with the authentication credentials. Since this is a non-recoverable error in the job context, retries are unnecessary.
- Customers are already notified of such errors through a webhook, making additional retries redundant and potentially confusing for the customer.

#### Additional Context:
- The `update_payment_method_id` method in the related service sends a webhook notification when any `Stripe::StripeError` is raised, ensuring customers are informed of issues promptly.

Screenshot 2024-07-02 at 18 30 10

### DIFF
--- a/app/services/invoices/payments/stripe_service.rb
+++ b/app/services/invoices/payments/stripe_service.rb
@@ -52,11 +52,11 @@ module Invoices
         result.payment = payment
         result
       rescue Stripe::AuthenticationError, Stripe::CardError, Stripe::InvalidRequestError, Stripe::PermissionError => e
-        deliver_error_webhook(e)
         # NOTE: Do not mark the invoice as failed if the amount is too small for Stripe
         #       For now we keep it as pending, the user can still update it manually
         return if e.code == 'amount_too_small'
 
+        deliver_error_webhook(e)
         update_invoice_payment_status(payment_status: :failed, deliver_webhook: false)
         nil
       rescue Stripe::RateLimitError, Stripe::APIConnectionError => e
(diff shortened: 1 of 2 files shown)


julienbourdeau (approved):
TIL `discard_on`

I always rescued and ignore in perform. I like `discard_on` much better.

vincent-pochet:
Some other errors are handled here:
https://github.com/getlago/lago-api/blob/main/app/services/invoices/payments/stripe_service.rb#L193

Not sure but I think we should do the same things here rather than just silencing the error.
- Optionally (since it seems already OK): Notifying the owner of the account that its stripe account is not correctly configured 
- Setting the payment status to failed

app/services/invoices/payments/stripe_service.rb
```diff
@@ -51,6 +51,17 @@ def create
...
+        # NOTE: Do not mark the invoice as failed if the amount is too small for Stripe
+        #       For now we keep it as pending, the user can still update it manually
+        return if e.code == 'amount_too_small'
+
+        deliver_error_webhook(e)
+        update_invoice_payment_status(payment_status: :failed, deliver_webhook: false)
+        nil
+      rescue Stripe::RateLimitError, Stripe::APIConnectionError => e
+        deliver_error_webhook(e)
+        raise
```
vincent-pochet, line 64: Don't we want to keep handling `Stripe::StripeError` and just "ignore" the `Stripe::RateLimitError`, `Stripe::APIConnectionError`?

```suggestion
      rescue Stripe::RateLimitError, Stripe::APIConnectionError => e
        raise # Let the auto-retry process do its own job
      rescue Stripe::StripeError => e
        deliver_error_webhook(e)
        raise
       end
```

vincent-pochet (approved):

PR #2242, opened 2024-07-03
fix(netsuite-integration): limit netsuite payload rate

lovrocolic:
## Description

Netsuite only support decimals of to 14 digits if there is decimal point. This PR adapts this use-case

### DIFF
--- a/app/services/integrations/aggregator/invoices/payloads/netsuite.rb
+++ b/app/services/integrations/aggregator/invoices/payloads/netsuite.rb
@@ -5,6 +5,8 @@ module Integrations
     module Invoices
       module Payloads
         class Netsuite < BasePayload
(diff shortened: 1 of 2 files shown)


app/services/integrations/aggregator/invoices/payloads/netsuite.rb
```diff
@@ -88,6 +88,18 @@ def discounts
 
             output
           end
+
+          def limited_rate(precise_unit_amount)
+            unit_amount_str = precise_unit_amount.to_s
+
+            return precise_unit_amount if unit_amount_str.length <= 15
```
vincent-pochet, line 95: Could you turn this magic number into a constant so that we know what it means ;)

vincent-pochet (approved):

PR #2243, opened 2024-07-03
fix(wallet): Prevent from having duplicate Invoices::PrepaidCreditJob

rsempe:
## Description

We've recently seen an issue where we receive the `succeeded` webhook from Stripe several times at the same moment (some milliseconds interval) on some wallet transactions.

The goal of this PR is to add a unique lock on the `Invoices::PrepaidCreditJob` job to not impact the balance twice.


julienbourdeau (approved):
Hours of debugging, fixed in 1 line 😄 🚀

PR #2244, opened 2024-07-03
fix(trial): Handle timezone correct to bill subscription on billing day

julienbourdeau:
## Context

When using Trial period, the subscription is only billed at the end of the trial.

## Description

If the trial ends on billing day, we should invoice the subscription for the upcoming month. This part works but it didn't handle timezone. It creates issues with billing happens the day before in customer timezone


rsempe (approved):
👍

PR #2245, opened 2024-07-03
feat(export invoices): add mutation to request data export for invoices

ancorcruz:
## Roadmap Task

## Context

## Description

This change adds a mutation to the graphQL API to request an invoices data export. Returns the export id and the status.


app/graphql/mutations/data_exports/invoices/create.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Mutations
+  module DataExports
+    module Invoices
+      class Create < BaseMutation
+        include AuthenticableApiUser
+        include RequiredOrganization
+
+        REQUIRED_PERMISSION = 'invoices:create'
```
ancorcruz, line 10: still investigating how the permissions work. The feature should be available for all roles...

julienbourdeau: I'd recommend creating `invoices:export`. You can add to the [permissions definitions.yml](https://github.com/getlago/lago-api/blob/main/app/config/permissions/definition.yml#L40-L45) file.

When set to true here, it's available to all roles.

```
invoices:
  - ...
  - export: true
```

ancorcruz: thanks

julienbourdeau (approved):

vincent-pochet (approved):

PR #2257, opened 2024-07-08
misc: Remove deprecated group handling

vincent-pochet:
## Context

This PR is part of the process of replacing groups with filters

## Description

This PR removes the handling of groups in all part of the application, leaving only filters.

For now `groups` and `group_properties` database tables and the related models are not removed. This final step will be done in a later release to keep a way for rolling back in case of emergency


rsempe (approved):
🎢

PR #2260, opened 2024-07-08
fix(dates-service): Fix terminated dates service for min commitment calculations

ivannovosad:
## Context

Minimum commitment calculation was using incorrect dates service in when refreshing draft invoices for terminated subscriptions.

## Description

Fix terminated dates service to be able to skip calling `matching_invoice_subscription?` when creating the dates service.

### DIFF
--- a/app/services/commitments/dates_service.rb
+++ b/app/services/commitments/dates_service.rb
@@ -31,7 +31,8 @@ module Commitments
       Subscriptions::TerminatedDatesService.new(
         subscription: invoice_subscription.subscription,
         invoice: invoice_subscription.invoice,
-        date_service: ds
+        date_service: ds,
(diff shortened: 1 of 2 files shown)


julienbourdeau:
Can you share more details about what's going on? If you can add an example that would be great.
I don't get what the original bug is and why it's fixed this way. 

The orignal `matching_invoice_subscription?(subscription, new_dates_service) ? date_service : new_dates_service` is already hard to understand without your fix.

Maybe a scenario test would help too. 🙏

ivannovosad:
The root cause of the bug is that an incorrect dates service is initiated.

Then when the subscription is terminated it returns incorrect `dates_service.previous_beginning_of_period` which causes `all_invoice_subscriptions` to be empty.
And later on `all_invoice_subscriptions.first.from_datetime` is called on `nil`.

So in this PR we're initiating a different dates service.

PR #2264, opened 2024-07-08
fix: Normalize CSV Comparison in Tests to Ignore Quotes

brunomiguelpinto:
#### Description:
This PR addresses an issue with the `.get_all_as_csv` test in the `Deprecation` model where the presence of quotes around `organization.name` caused the test to fail. The test has been updated to normalize both the expected and actual CSV strings by removing quotes before comparison, ensuring the comparison focuses on the data content rather than the format.

### DIFF
--- a/app/graphql/types/data_exports/invoices/filters_input.rb
+++ b/app/graphql/types/data_exports/invoices/filters_input.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Types
   module DataExports
     module Invoices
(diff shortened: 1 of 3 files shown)


brunomiguelpinto:
we can also change on the factory the name to remove the ""
name { Faker::Company.name.gsub(/[,"']/, '') } # Remove commas, double quotes, and single quotes

brunomiguelpinto:
def csv_safe(value)
    # Enclose the value in double quotes if it contains a comma or double quote
    if value.include?(',') || value.include?('"')
      value = value.gsub('"', '""') # Escape double quotes by doubling them
      "\"#{value}\""
    else
      value
    end
  end
  
  we can create a csv helper too

ancorcruz (approved):

PR #2267, opened 2024-07-09
feat(applied-taxes): add logic for creating applied taxes based on external provider

lovrocolic:
## Context

Currently Anrok - tax tool integration is being added

## Description

This PR created applied taxes objects in Lago based on fetched Anrok taxes.

Two new services are added: `Fees::ApplyProviderTaxesService` and `Invoices::ApplyProviderTaxesService`.

This logic follows existing behaviour in `Fees::ApplyTaxesService` and `Invoices::ApplyTaxesService` services. The only difference is that source of taxes is not Lago app anymore but external provider.

New services are not used yet.

### DIFF
--- a/app/services/fees/apply_provider_taxes_service.rb
+++ b/app/services/fees/apply_provider_taxes_service.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module Fees
+  class ApplyProviderTaxesService < BaseService
+    def initialize(fee:, fee_taxes:)
+      @fee = fee
+      @fee_taxes = fee_taxes
+
+      super
+    end
+
+    def call
+      result.applied_taxes = []
+      return result if fee.applied_taxes.any?
+
+      applied_taxes_amount_cents = 0
+      applied_taxes_rate = 0
+
+      fee_taxes.tax_breakdown.each do |tax|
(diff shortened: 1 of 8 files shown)


app/services/fees/apply_provider_taxes_service.rb
```diff
@@ -0,0 +1,53 @@
...
+        applied_tax.amount_cents = tax_amount_cents.round
+        applied_tax.save! if fee.persisted?
+
+        applied_taxes_amount_cents += tax_amount_cents
+        applied_taxes_rate += tax_rate
+
+        result.applied_taxes << applied_tax
+      end
+
+      fee.taxes_amount_cents = applied_taxes_amount_cents.round
```
annvelents, line 41: I'm wondering, wouldn't it potentially lead to not matching sum of applied taxes with fee.taxes_amount_cents and if it can be a problem 
(because for applied tax we're rounding each element, but for tax_amount_cents we're rounding the sum)

lovrocolic: I think both round calls are pretty extra since we should receive amount in cents (and it cents there should be no decimals). I just added it as a guard but we should be good IMO

vincent-pochet (approved):

PR #2269, opened 2024-07-10
fix (integration-customers): adapt integration customer queries

lovrocolic:
## Context

Improve integration customer queries

## Description

Currently Accounting integrations, Netsuite and Xero, are supported in Lago but tax integration Anrok is currently being implemented.

This PR improves existing integration customers queries. Customer can have attached ONLY one accounting integration and ONLY one tax integration, but we want to avoid case that tax integration customer is returned in accounting context.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_07_06_204557) do
+ActiveRecord::Schema[7.1].define(version: 2024_07_08_195226) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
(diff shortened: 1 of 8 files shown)


db/schema.rb
```diff
@@ -535,7 +535,7 @@
 
   create_table "fees_taxes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.uuid "fee_id", null: false
-    t.uuid "tax_id", null: false
+    t.uuid "tax_id"
```
ivannovosad, line 538: Should these schema changes be in this PR?

db/schema.rb
```diff
@@ -741,7 +741,7 @@
 
   create_table "invoices_taxes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.uuid "invoice_id", null: false
-    t.uuid "tax_id", null: false
+    t.uuid "tax_id"
```
ivannovosad, line 744: Same here. Should these schema changes be in this PR?

ivannovosad (approved):

PR #2283, opened 2024-07-12
feat: Add partial index to fees on charge_id and invoice_id

brunomiguelpinto:
**Description**:

This PR adds a partial index to the fees table on the charge_id and invoice_id columns, with a condition to exclude rows where deleted_at is not NULL

### DIFF
--- a/db/migrate/20240712090133_add_index_to_fees.rb
+++ b/db/migrate/20240712090133_add_index_to_fees.rb
@@ -5,7 +5,6 @@ class AddIndexToFees < ActiveRecord::Migration[7.1]
 
   def change
     add_index :fees, [:charge_id, :invoice_id],
-      name: 'index_fees_on_charge_id_and_invoice_id',
       where: 'deleted_at IS NULL',
       algorithm: :concurrently,
       if_not_exists: true


jdenquin (approved):

db/migrate/20240712090133_add_index_to_fees.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddIndexToFees < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    unless index_exists?(:fees, [:charge_id, :invoice_id], name: 'index_fees_on_charge_id_and_invoice_id', where: 'deleted_at IS NULL')
+      add_index :fees, [:charge_id, :invoice_id],
+        name: 'index_fees_on_charge_id_and_invoice_id',
+        where: 'deleted_at IS NULL'
```
jdenquin, line 10: I Think you can use `if_not_exists: true`
Also I think you missed the algorithm for the index creation

jdenquin (changes requested):

ivannovosad (approved):

rsempe (approved):

jdenquin (approved):

PR #2288, opened 2024-07-16
POC: Custom configuration for model

julienbourdeau:
# Summary

Instead of constantly adding new columns for very specific feature, we could add a `config` object to models and store a json object in the database. It avoid running so many migrations and it's easy to configure defaults. 
It can be turned into a proper object with methods, instead of dealing with a hash.

**Not everything should go there, there are still plenty of good reasons to use a proper dedicated column.**

### Pros

* No database migration to add new fields easily
* Postgres is good enough to make queries in `jsonb` columns
* Same specific object (like `Wallet::Config`) for many similar model

### Cons

* Not as nice as a proper ActiveRecord attribute
* Could be abused and used for the wrong reasons

### Custom `Config` object

Using a dedicated `Wallet::Confg` here allows us to create custom helper method to avoid using the hash.
It also avoid for the forever issue of symbol vs string in the hash. Remember that once loaded from the database (via `JSON.parse`, all symbols are turned to strings.

```
wt.config.invoice_require_successful_payment?
# VS
wt.config["invoice.require_successful_payment"]
```

### Query

If you often need to retrieve all models via a field, like `Wallet.where(something: '234')` it's probably not the best to put it in the config. If you sometimes add some where statement on the config, Postgres should handle that well.

```ruby
WalletTransaction.where(
  wallet_id: id,
  status: :settled
).where(
  "config->>'invoice.require_successful_payment' = ?", true
).count
```

### DottedHash

`DottedHash` was introduced for Permissions. It flattens a complex hash to a 1-level hash with dots separating the keys.

```ruby
h = {
    invoice: {
        retry: true,
        notify: [:email, :slack]
    },
    taxes: :skip
}

dotted_h = {
    "invoice.retry" => true, 
    "invoice.notify" => [:email, :slack], 
    "taxes" => :skip
}
```

# Example

CleanShot 2024-07-16 at 10 56 49@2x

CleanShot 2024-07-16 at 10 58 09@2x

CleanShot 2024-07-16 at 10 58 19@2x

# Other possible improvements

* We could store some sort of schema along with defaults to generate the GrapgQL types
* Add data validation (using the same schema?)

Actually the best evolution is that Config becomes a proper `ActiveRecord::Model` with attribute, validation, and such, but backed by one json column, not a full table. 😎 

In this case, we should get rid of the DottedHash thing 👍 

* https://evilmartians.com/chronicles/wrapping-json-based-active-record-attributes-with-classes
* https://github.com/palkan/store_attribute
* https://github.com/madeintandem/jsonb_accessor

### DIFF
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class Config
+  def self.load(value)
+    new(value || {})
+  end
+
+  def self.dump(value)
+    DottedHash.new(value.to_h).to_h
+  end
+
+  def initialize(hash)
+    @hash = DottedHash.new(default).merge(DottedHash.new(hash)).to_h
+  end
+
+  def [](key)
+    @hash[key]
+  end
+
+  def to_h
+    @hash.dup
+  end
+
+  def inspect
+    "(#{self.class.name}) \n\t #{to_h.inspect}\n"
+  end
+
+  def default
+    {}
+  end
+
+  def reset
+    @hash = DottedHash.new(default).to_h
+  end
+end
--- a/app/models/recurring_transaction_rule.rb
+++ b/app/models/recurring_transaction_rule.rb
@@ -5,6 +5,8 @@ class RecurringTransactionRule < ApplicationRecord
 
   belongs_to :wallet
 
+  serialize :config, coder: ::Wallet::Config
+
   INTERVALS = [
     :weekly,
     :monthly,
--- a/app/models/wallet.rb
+++ b/app/models/wallet.rb
@@ -1,6 +1,24 @@
 # frozen_string_literal: true
 
 class Wallet < ApplicationRecord
+  class Config < ::Config
+    def invoice_require_successful_payment?
+      @hash["invoice.require_successful_payment"]
+    end
+
+    def invoice_require_successful_payment=(value)
+      @hash["invoice.require_successful_payment"] = value
+    end
+
+    def default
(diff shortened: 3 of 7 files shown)


jdenquin:
IMO it's a good idea, but I'm a bit afraid about the performances, querying huge table on PG with jsonb is really really expansive (even with gin indexes) since he has to parse all the json fields, so making a request on a jsonb properties may have an impact. We can't index a jsonb properties as we index a "real" column.
So it may be useful for some cases, if you don't have to base queries on it.

vincent-pochet:
We already have something similar with the `SettingsStorable` concern, it's a bit less "fancy" as it only exposes accessors, but I think it could be extended. You can see it in action on `Integrations` and `PaymentProviders` sub models

julienbourdeau:
@jdenquin Yeah, I 100% agree if you need to often query on these fields, they don't belong in the config. I'm afraid that this will be abused. I meant to say that it's possible to query if you need to migrate the data for instance, but it should be rare.

@vincent-pochet Very good point. I'm adding a "solution 3" to the dive-in where we simply use this concern. I'd need at least add a way to handle defaults.

julienbourdeau:
I like really like this but for my feature, I think a regular active model column is a better choice. I don't want to rush it.
I'm thinking https://github.com/palkan/store_attribute is probably all we need. Maybe we should consider using it for our settings storable 🤔 

This PR will stay as a POC for future reference.

PR #2289, opened 2024-07-16
feat: Add Upgrade Verification Task

brunomiguelpinto:
> **Warning: Placement of the Versions File**
> 
> For the new Rake task to function correctly, the `versions.yml` file needs to be placed in an accessible and secure 
 location within our project structure.
> Where should we put it?  

### Description

This PR introduces a new Rake task under the `upgrade` namespace that verifies the current system's readiness for an upgrade and outlines the necessary migration paths. The task performs checks on database migrations and background jobs to ensure the system is prepared for an upgrade. The task compares the current version against the target versions and determines if a direct upgrade is possible or if a migration path is needed.

### Key Changes:

1. **New Rake Task:** `upgrade:verify`
   - This task checks the system's readiness for an upgrade by ensuring all necessary migrations and background jobs have been completed.

2. **Subtasks:**
   - `check_migrations`: Verifies if all migrations for the current version have been run.
   - `check_background_jobs`: Ensures that there are no pending jobs in the 'background_migration' queue.

### Example Output

Running the task with the command `rake upgrade:verify` will produce output such as:

```
rake upgrade:verify
Your current version is 1.3. The latest version is 1.8.
You need to upgrade. Here is the migration path:
To upgrade to version 1.5, you need to run the following migrations:
  - migration_1
  - migration_2
To upgrade to version 1.6, you need to run the following migrations:
  - migration_3
To upgrade to version 1.8, you need to run the following migrations:
  - migration_4
```


jdenquin:
how can we make it usable for OSS?

config/versions.yml
```diff
@@ -0,0 +1,13 @@
...
+  - version: 1.5.0
+    migrations: []
+  - version: 1.5.1
+    migrations: []
+  - version: 1.6.0
+    migrations: []
+  - version: 1.6.1
+    migrations:
+      - 20240603095841
+      - 20240628083830
```
brunomiguelpinto, line 11: removed a field

config/versions.yml
```diff
@@ -0,0 +1,13 @@
+versions:
+  - version: 1.5.0
+    migrations: []
+  - version: 1.5.1
+    migrations: []
+  - version: 1.6.0
+    migrations: []
+  - version: 1.6.1
+    migrations:
+      - 20240603095841
```
brunomiguelpinto, line 10: desecrated field and start using a new one

vincent-pochet (approved):

lib/tasks/upgrade_verification.rake
```diff
@@ -0,0 +1,127 @@
...
+      Gem::Version.new(versions['versions'].last['version'])
+    else
+      # Use the LAGO_VERSION constant in other environments
+      Gem::Version.new(LAGO_VERSION.number)
+    end
+  end
+
+  def load_versions_data
+    # todo check were we can store this
+    uri = URI("https://raw.githubusercontent.com/getlago/lago-api/ca9a4af8dadf12fd41f574517ae7d49d2f15d100/config/to_versions.yml")
```
vincent-pochet, line 77: I think this will be:
```suggestion
    uri = URI("https://github.com/getlago/lago-api/tree/main/config/to_versions.yml")
```

Once merge of course. This way it will always point to the one in the main branch, that is supposed to be the most up to date

PR #2290, opened 2024-07-16
fix(xero): Fix syncing xero accounts

ivannovosad:
## Roadmap Task

## Context

## Description

This PR removes the accounts resolver and adds FetchAccounts mutation instead.


app/services/integrations/aggregator/accounts_service.rb
```diff
@@ -3,20 +3,38 @@
 module Integrations
   module Aggregator
     class AccountsService < BaseService
+      LIMIT = 450
+      MAX_SUBSEQUENT_REQUESTS = 15
```
julienbourdeau, line 7: I'm not sure what an "account" is in this context. I'm assuming that it's very unexpected to have so many?

julienbourdeau (approved):

PR #2296, opened 2024-07-17
feat(anrok): Add failed status to invoice

annvelents:
## Context
https://www.notion.so/getlago/Dive-in-Calculate-taxes-info-for-finalized-invoices-subscription-58481b4c8a5c40b18cf86b8a800cf49d?pvs=4#255f83ec5c6a47b283ee5afda6d13df1
As part of processing errors from tax_provider, we want to be able to indicate that invoice failed to be generated

## Description

Added new status for invoices: `failed`
Added prefix to payment_statuses enum methods: `payment`

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -50,7 +50,7 @@ class Invoice < ApplicationRecord
 
   INVOICE_TYPES = %i[subscription add_on credit one_off advance_charges].freeze
   PAYMENT_STATUS = %i[pending succeeded failed].freeze
-  STATUS = %i[draft finalized voided generating].freeze
+  STATUS = %i[draft finalized voided generating status_failed].freeze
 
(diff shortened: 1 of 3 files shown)


julienbourdeau:
I'm *very happy* to see the `payment_` prefix on payment status! 🤩  I'd have expected it to be harder to be honest.

Would you mind doing only this change as a first PR? 🙏  Regardless of what happened to the `failed` status (if it's renamed or whatever), I think this would be a great improvement.

annvelents:
@julienbourdeau yes, makes sense, as it's quite a big change
(also actually while looking for all places to change this prefix I was thinking if we should do the same for fees, because we have exactly the same enum there)

julienbourdeau:
I think we absolutely should do it for fees too! 🚀  Probably in 2 separate PR.

lovrocolic (approved):

PR #2297, opened 2024-07-17
feat(webhook): Always dispatch webhook job

julienbourdeau:
## Description

Currently, we often make a query to the DB to figure if we should enqueue a job to send the webhooks.

As discussed during our tech meeting about a month ago, we're moving to always enqueuing a job. The job will simply do nothing if there is no webhook_endpoint configured for the given organization.

It doesn't show in the the diff but the orgs without endpoints won't receive any webhook.

https://github.com/getlago/lago-api/blob/b615f63a405efe81990436667f99c6ee34b408e6/app/services/webhooks/base_service.rb#L24-L27

### Benefits

* Centralized logic of _"should we send webhook?"_ in the job (easier if want to [subscribe to only a subset of webhooks](https://github.com/getlago/lago-api/pull/1993))
* Saves a query during the main tasks

### Down sides

* Enqueues more jobs (very small jobs)

### DIFF
--- a/app/jobs/clock/subscriptions_to_be_terminated_job.rb
+++ b/app/jobs/clock/subscriptions_to_be_terminated_job.rb
@@ -18,9 +18,7 @@ module Clock
         )
         .where('webhooks.id IS NULL OR webhooks.created_at::date != ?', Time.current.to_date)
         .find_each do |subscription|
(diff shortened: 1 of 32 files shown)


vincent-pochet (approved):
:rocket:

PR #2298, opened 2024-07-17
fix(rake): Skip gsub if no variable is defined

julienbourdeau:
## Description

When not env variable is defined we end up with `gsub('', '*****')` which result in a segmentation fault. 😅

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -20,9 +20,13 @@ namespace :db do
     migration_file = 'db/clickhouse_schema.rb'
     text = File.read(migration_file)
 
-    text = text.gsub(ENV.fetch('LAGO_KAFKA_BOOTSTRAP_SERVERS', ''), '*****')
-    text = text.gsub(ENV.fetch('LAGO_KAFKA_RAW_EVENTS_TOPIC', ''), '*****')
-    text = text.gsub(ENV.fetch('LAGO_KAFKA_CLICKHOUSE_CONSUMER_GROUP', ''), '*****')
+    [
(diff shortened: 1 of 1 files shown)


brunomiguelpinto (approved):

lib/tasks/db.rake
```diff
@@ -20,9 +20,13 @@ namespace :db do
     migration_file = 'db/clickhouse_schema.rb'
     text = File.read(migration_file)
 
-    text = text.gsub(ENV.fetch('LAGO_KAFKA_BOOTSTRAP_SERVERS', ''), '*****')
-    text = text.gsub(ENV.fetch('LAGO_KAFKA_RAW_EVENTS_TOPIC', ''), '*****')
-    text = text.gsub(ENV.fetch('LAGO_KAFKA_CLICKHOUSE_CONSUMER_GROUP', ''), '*****')
+    [
+      ENV.fetch('LAGO_KAFKA_BOOTSTRAP_SERVERS', nil),
+      ENV.fetch('LAGO_KAFKA_RAW_EVENTS_TOPIC', nil),
+      ENV.fetch('LAGO_KAFKA_CLICKHOUSE_CONSUMER_GROUP', nil)
```
vincent-pochet, line 26: Since nil is the default, you can avoid the fetch:
```suggestion
      ENV['LAGO_KAFKA_BOOTSTRAP_SERVERS'],
      ENV['LAGO_KAFKA_RAW_EVENTS_TOPIC'],
      ENV['LAGO_KAFKA_CLICKHOUSE_CONSUMER_GROUP']
```

julienbourdeau:
File was removed by https://github.com/getlago/lago-api/pull/2309

PR #2299, opened 2024-07-18
Feat(invoice.payment_status): add prefix to invoice payment status enum

annvelents:
As part of implementation error status for invoices, we got conflict between current `failed?` method on invoice that comes from payment_status and new `failed?` method that comes from invoice status itself, so it makes sense to make payment statuses explicitly payment's  statuses, introducing this prefix to payment_status enum, so it will be
`paymen_failed` instead (and other methods from the enum)
Also it will make it clearer that `invoice.pending?` does not mean that the invoice is pending because of something, instead we'll have now `invoice.payment_pending?` which is again more explicit

### DIFF
--- a/app/jobs/clock/mark_invoices_as_payment_overdue_job.rb
+++ b/app/jobs/clock/mark_invoices_as_payment_overdue_job.rb
@@ -9,7 +9,7 @@ module Clock
     def perform
       Invoice
         .finalized
-        .not_succeeded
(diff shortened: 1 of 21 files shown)


julienbourdeau (approved):
🙌

lovrocolic (approved):

rsempe (approved):

PR #2302, opened 2024-07-18
chore(services): Rename FinalizeService to RefreshDraftAndFinalizeService

julienbourdeau:
## Description

As we're considering adding more Invoice state, we should specify what the current `FinalizeService` does. It not only finalize it, it takes a draft invoice, refresh it and finalize it.

We're also changing a little bit how invoice is set in result and reloaded, hoping to make it a little bit clearer.

### DIFF
--- a/app/jobs/invoices/finalize_job.rb
+++ b/app/jobs/invoices/finalize_job.rb
@@ -5,7 +5,7 @@ module Invoices
     queue_as 'invoices'
 
     def perform(invoice)
-      Invoices::FinalizeService.call(invoice:)
+      Invoices::RefreshDraftAndFinalizeService.call(invoice:)
     end
   end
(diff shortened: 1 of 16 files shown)


app/jobs/invoices/finalize_job.rb
```diff
@@ -5,7 +5,7 @@ class FinalizeJob < ApplicationJob
     queue_as 'invoices'
 
     def perform(invoice)
```
julienbourdeau, line 7: The Job itself is not renamed (yet) because renaming jobs is not so easy with self-hosted users. We believe it's not worth the effort.

lovrocolic (approved):

brunomiguelpinto (approved):

PR #2305, opened 2024-07-18
feat(anrok): add migration for IntegrationErrorDetails

annvelents:
## Roadmap Task

https://www.notion.so/getlago/Dive-in-Calculate-taxes-info-for-finalized-invoices-subscription-58481b4c8a5c40b18cf86b8a800cf49d?pvs=4#62871d933e9745abb548f8c03bdb55da

## Context

As part of implementation logic to store errors received from Anrok through Nango we need a new model that would be used to store the errors

## Description

Added migration to create table for the new model

### DIFF
--- a/db/migrate/20240718105718_create_integration_error_details.rb
+++ b/db/migrate/20240718105718_create_integration_error_details.rb
@@ -4,7 +4,7 @@ class CreateIntegrationErrorDetails < ActiveRecord::Migration[7.1]
   def change
     create_table :integration_error_details, id: :uuid do |t|
       t.references :owner, type: :uuid, polymorphic: true, null: false, index: true
-      t.references :integration, type: :uuid, polymorphic: true, index: true, null: false
+      t.references :error_producer, type: :uuid, polymorphic: true, index: true, null: false
       t.jsonb :details, null: false, default: {}
       t.datetime :deleted_at, index: true
 
--- a/db/migrate/20240718105718_create_integration_error_details.rb
+++ b/db/migrate/20240718105718_create_integration_error_details.rb
@@ -2,9 +2,9 @@
 
 class CreateIntegrationErrorDetails < ActiveRecord::Migration[7.1]
   def change
-    create_table :integration_error_details, id: :uuid do |t|
+    create_table :error_details, id: :uuid do |t|
       t.references :owner, type: :uuid, polymorphic: true, null: false, index: true
-      t.references :error_producer, type: :uuid, polymorphic: true, index: true, null: false
+      t.references :integration, type: :uuid, polymorphic: true, index: true, null: true
       t.jsonb :details, null: false, default: {}
       t.datetime :deleted_at, index: true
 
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):

julienbourdeau:
I know we discussed how this model is only related to integration and is not meant to be used for all other errors.
I understand the point of wanting to control the model and the table schema (other errors won't have `integration_id` but my need `payment_provider_id` for instance).

I think we should use a delegated type here.
This would make it very easy to list all the errors for a model `invoices.error_details` would give you an ActiveRecord relation where the models are sometimes `IntegrationErrorDetails` or `PaymentProvderErrorDetails` or some other specific error.

Without delegated types, to know if there is an error on an Invoice, you'd need to check multiple relations `inv.payment_provider_error_details.any? && inv.integration_error_details.any? && inv.else_error_details.any?`

What do you think?

annvelents:
Hey @julienbourdeau !
Thank you for pointing this out again, and actually after reading your dive-in i see how it makes sense, extending it now 😁

db/migrate/20240718105718_create_integration_error_details.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateIntegrationErrorDetails < ActiveRecord::Migration[7.1]
+  def change
+    create_table :integration_error_details, id: :uuid do |t|
+      t.references :owner, type: :uuid, polymorphic: true, null: false, index: true
+      t.references :error_producer, type: :uuid, polymorphic: true, index: true, null: false
```
annvelents, line 7: I was deciding between STI, delegated_type and polymorphism
Initially i ended up with polymorphic `:integration` as it would have highlighted that payment_provider and integration are the same instance, and later, when we'd do the refactoring we'd get rid of this relation at all (having simple integration_id).
(also because integration error should have the same behaviour for different types of integrations and the same fields)
But then I thought about making this model even more generic, if we're talking not only about payment_providers and integrations, but if suddenly we have something else, and in this case something, where the error happened is something that created that error, so then this is the error producer.
(at the same time I'm still thinking that technically all external integrations should follow pattern of integrations, and would be perfect to end up with only integration_id...)

annvelents: (but then should it be something like `third_party_error_details`?

julienbourdeau: Hey Anna thanks a lot for thinking this through and caring.

I'd go as far as calling the table `error_details` and let anyone add their error there. 
It makes sense that most errors would come from 3rd parties (because our errors are supposed to be handled 😄) so I'd add a polymorphic relation to integration **but make it optional.**

```rb
create_table :error_details, id: :uuid do |t|
  t.references :owner, type: :uuid, polymorphic: true, null: false, index: true
  t.references :integration, type: :uuid, polymorphic: true, index: true, null: true
  # ...
end
```

I also strongly believe we should migration the `payment_providers` table content to `integrations` and keep using the same PaymentProvider classes/code. But let's talk about this separately. 👍

annvelents:
After more discussion with @julienbourdeau we decided to end up with more generic table name - error_details, but less generic link to integration, which will be optional.
As having an external service triggering an error in our system might be a sign that this should follow the integrations interface, but at the same time having this relation optional will allow us to save even internal errors if they require user to make an action

annvelents:
after discussion with @julienbourdeau we decided that having reference to the organization will be benefitial as not always owner will have a direct link to the organization, and we might need possibility to query all error for an organization.
Also added error_code, as error_details suggest having explicit information about the error, while error_code will be enum that also would help us querying generic errors

julienbourdeau (approved):
Great! 🤩

PR #2307, opened 2024-07-18
fix(export-invoices): allow no pagination for Query objects

ancorcruz:
## Roadmap Task

👉 https://getlago.canny.io/feature-requests/p/ability-to-export-data-from-the-user-interface

## Context

When exporting data (eg: invoices) we use query objects to fetch the records to export. However, the current implementation of query objects paginates to first page and limit results to 25 or 100 (depends on implementation) when no pagination information is provided.

We need the query object to allow no pagination and fetch all resources matching the search / filter criteria when no pagination is provided.

## Description

This change allows client code to call Query object with no pagination information to avoid pagination defaults (page: 1, limit: 25).

Also, it makes `BaseQuery::Pagination` internal knowledge of the query objects. As query objects expect pagination as a hash `{page: 1, limit: 50}`. The same will be applied to `BaseQuery::Filters` in a coming PR.

Updates the Query objects implementing the latest approach and updates `InvoicesQuery` object to allow no pagination and fix the CSV data export issues.

### DIFF
--- a/app/queries/base_query.rb
+++ b/app/queries/base_query.rb
@@ -1,19 +1,13 @@
 # frozen_string_literal: true
 
 class BaseQuery < BaseService
-  PER_PAGE = 100
-
-  Pagination = Struct.new(:page, :limit, keyword_init: true) do
-    def initialize(page: 0, limit: PER_PAGE)
-      super
-    end
-  end
+  Pagination = Struct.new(:page, :limit, keyword_init: true)
 
   class Filters < OpenStruct; end
 
-  def initialize(organization:, pagination: Pagination.new, filters: Filters.new)
+  def initialize(organization:, pagination: nil, filters: Filters.new)
     @organization = organization
-    @pagination = pagination
+    @pagination_params = pagination
     @filters = filters
 
     super
@@ -21,9 +15,20 @@ class BaseQuery < BaseService
 
   private
 
-  attr_reader :organization, :pagination, :filters
+  attr_reader :organization, :pagination_params, :filters
+
+  def pagination
+    return if pagination_params.blank?
+
+    @pagination ||= Pagination.new(
+      page: pagination_params[:page],
+      limit: pagination_params[:limit]
+    )
+  end
 
   def paginate(scope)
+    return scope unless pagination
+
     scope.page(pagination.page).per(pagination.limit)
   end
 
--- a/app/queries/base_query.rb
+++ b/app/queries/base_query.rb
@@ -2,7 +2,7 @@
 
 class BaseQuery < BaseService
   # nil values force Kaminari to apply its default values for page and limit.
-  DEFAULT_PAGINATION_PARAMS = { page: nil, limit: nil }
+  DEFAULT_PAGINATION_PARAMS = {page: nil, limit: nil}
 
   Pagination = Struct.new(:page, :limit, keyword_init: true)
 
--- a/app/queries/base_query.rb
+++ b/app/queries/base_query.rb
@@ -1,11 +1,14 @@
 # frozen_string_literal: true
 
 class BaseQuery < BaseService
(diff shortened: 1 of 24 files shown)


app/queries/base_query.rb
```diff
@@ -1,29 +1,34 @@
...
 
 class BaseQuery < BaseService
-  PER_PAGE = 100
-
-  Pagination = Struct.new(:page, :limit, keyword_init: true) do
-    def initialize(page: 0, limit: PER_PAGE)
-      super
-    end
-  end
+  Pagination = Struct.new(:page, :limit, keyword_init: true)
```
ancorcruz, line 4: this `PER_PAGE` default was not use and all the time it was using kaminari per page default (25) as limit was provided with `nil` value.

Also, first page is `1` instead of `0`, kaminari also defaults `nil` as first page.

hence, this code wasn't needed.

app/queries/base_query.rb
```diff
@@ -1,29 +1,34 @@
...
+    return if pagination_params.blank?
+
+    @pagination ||= Pagination.new(
+      page: pagination_params[:page],
+      limit: pagination_params[:limit]
+    )
+  end
 
   def paginate(scope)
+    return scope unless pagination
```
ancorcruz, line 30: no pagination.

vincent-pochet (approved):
Looks good to me :+1: (Really nice for the hash input format rather than the weird object) 

Maybe just one question:
Forcing the pagination was an easy way to limit memory consumption, especially on models with a lot of records (no max is forced, witch is an issue in itself...), so it means that now we have to be extra-cautious when opening queries without pagination.
I understand the need for exporters, but do you think it would have been possible to keep pagination and process records "by batch"?

ancorcruz:
@vincent-pochet I got your concern, for export there should be no issue as the Query object returns an active relation object, not the actual records, and we iterate in batches to produce the CSV file.

However, as you pointed, with this implementation we must take care of include pagination or we could fetch too many records at once. In order to mitigate this we could have pagination by default and allow no pagination as an option... the method signature could be as follow;

```ruby
  DEFAULT_PAGINATION_PARAMS = { page: 1, limit: 25 }
  # or using nil values to let kaminari defaults apply.
  # DEFAULT_PAGINATION_PARAMS = { page: nil, limit: nil }
  
  def initialize(organization:, pagination: DEFAULT_PAGINATION_PARAMS, filters: Filters.new)
```

everything else will remain as it is in this PR.

Then when no pagination is required the query object will be instantiated with `pagination: nil` argument. However, when no pagination is provided it will be added by default keeping our query objects safe of taking too much memory or impact app performance.

What do you think?

ancorcruz:
BTW, this PR is related to https://github.com/getlago/lago-api/pull/2179

vincent-pochet:
> @vincent-pochet I got your concern, for export there should be no issue as the Query object returns an active relation object, not the actual records, and we iterate in batches to produce the CSV file.
> 
> However, as you pointed, with this implementation we must take care of include pagination or we could fetch too many records at once. In order to mitigate this we could have pagination by default and allow no pagination as an option... the method signature could be as follow;
> 
> ```ruby
>   DEFAULT_PAGINATION_PARAMS = { page: 1, limit: 25 }
>   # or using nil values to let kaminari defaults apply.
>   # DEFAULT_PAGINATION_PARAMS = { page: nil, limit: nil }
>   
>   def initialize(organization:, pagination: DEFAULT_PAGINATION_PARAMS, filters: Filters.new)
> ```
> 
> everything else will remain as it is in this PR.
> 
> Then when no pagination is required the query object will be instantiated with `pagination: nil` argument. However, when no pagination is provided it will be added by default keeping our query objects safe of taking too much memory or impact app performance.
> 
> What do you think?

@ancorcruz looks like a perfect way to address the issue :+1:

PR #2311, opened 2024-07-19
feat(anrok): add model for error_details

annvelents:
## Context

As part of implementation logic to store errors received from Anrok through Nango we need a new model that would be used to store the errors we got from integrations we have

## Description

Added the model itself after the table has been added


app/models/error_detail.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class ErrorDetail < ApplicationRecord
+  include Discard::Model
+  self.discard_column = :deleted_at
+
+  belongs_to :integration, polymorphic: true, optional: true
+  belongs_to :owner, polymorphic: true
+  belongs_to :organization
+
+  ERROR_CODES = %w[not_provided]
```
julienbourdeau, line 11: 🙌

spec/factories/error_details.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :error_detail do
+    organization
+    association :owner, factory: %i[invoice].sample
+    association :integration, factory: :anrok_integration
```
julienbourdeau, line 7: Not sure it matters but I'd keep it empty by default and add a trait or a dedicated `anrok_error_detail` factory I think.

julienbourdeau (approved):

spec/factories/error_details.rb
ivannovosad: Or like this: 

```suggestion
    association :integration, factory: %i[netsuite_integration okta_integration xero_integration anrok_integration].sample
```

app/models/error_detail.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ErrorDetail < ApplicationRecord
+  include Discard::Model
+  self.discard_column = :deleted_at
+
+  belongs_to :integration, polymorphic: true, optional: true
```
lovrocolic, line 7: relationship to integration should not be polymorphic. For integrations we use STI, and relationship to the integration is actually relationship to the `Integrations::BaseIntegration` model

lovrocolic (changes requested):
Nice one! Please take a look on the comment regarding integration association

app/models/error_detail.rb
lovrocolic, line 11: So the goal is to extend this list with all possible error codes that will be used?

spec/factories/error_details.rb
annvelents: 👍

app/models/error_detail.rb
annvelents: the thing is that we also want to store payment_providers here, and we chose :integration as polymorphic relation to highlight that if something in our codebase is triggering integration error, most likely it should follow pattern of integrations

app/models/error_detail.rb
annvelents: yes 😁

app/models/error_detail.rb
lovrocolic: I think that the answer lies in the decision how we are going to use it: [https://github.com/getlago/lago-api/pull/2316/files#r1687997757](url)
1. If we will have something generic in `error_code` like `tax_error` or `payment_provider_error`, then we can completely remove this association I guess. It is easy to get the integration source: `invoice.customer.anrok_customer.integration` and we already have organization and owner relationship. In that case for `details` we will have `{tax_error: 'xyz'}`. If multiple errors are attached to the one invoice we can easily filter them by `error_code`.
2. If specific error code will be written in `error_code` field (like `taxDateTooFarInFuture`), in that case maybe it is useful to have integration relationship and I guess it can stay polymorphic, even I don't see in the near future we will use it for sth other than integration. In this case, at least for Anrok, we won't be using details field but it is fine I guess

annvelents: yep, just clarified: an organization can have together multiple connections with both xero and netsuite, but on the customer level we'll have only one

lovrocolic: exactly, on customer level you cannot be connected to multiple tax providers, because you need only one source of truth. For payment providers it should work the same. One customer cannot have multiple payment providers attached (that would mean that he/she would be billed multiple times)

app/models/error_detail.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class ErrorDetail < ApplicationRecord
+  include Discard::Model
+  self.discard_column = :deleted_at
+
+  belongs_to :owner, polymorphic: true
+  belongs_to :organization
+
+  ERROR_CODES = %w[not_provided tax_error]
+  enum error_code: ERROR_CODES.zip(ERROR_CODES).to_h
```
lovrocolic, line 11: Just realised that string DB field is used to be an enum. Natively in Rails enums are stored as integers and that would allow us to avoid such syntax. Is there any additional reason for using it this way? If not, I would prefer using enums as we use everywhere else across the app:)

annvelents: with enums - we don't know how much this codes will grow and how often we'll need to add them, because every time we're adding a new error type, we'll need to update the enum table.
If it's okay I'm happy to update 😁

app/models/integrations/anrok_integration.rb
```diff
@@ -2,6 +2,10 @@
 
 module Integrations
   class AnrokIntegration < BaseIntegration
+    has_many :error_details, -> { where({error_details: {error_code: 'tax_error'}}) },
```
lovrocolic, line 5: I think we don't need it in this direction since it would return all error details included in all invoices, I think that having access on invoice to fetch error details is enough

annvelents: I thought we might need it when showing all errors on the integration page 🤔

app/models/error_detail.rb
lovrocolic: If I haven't missed something, In both cases, we will only have to update `ERROR_CODES` constant. It is just matter how we store it and how enum definition looks like:)

app/models/integrations/anrok_integration.rb
lovrocolic: I think we will display only number of failed errors, but we can leave it for now

db/migrate/20240724085326_change_error_code_type_on_error_details.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class ChangeErrorCodeTypeOnErrorDetails < ActiveRecord::Migration[7.1]
```
lovrocolic, line 3: Is there any chance we can merge those two migrations? We haven't deployed it anywhere so it is fine to do it in one file

annvelents: 💪

PR #2315, opened 2024-07-22
feat(invoice): Introduce visible/invisible invoice status

julienbourdeau:
## Description

We're splitting invoices statuses into 2 categories:
* **Visible:** show up in the UI (via graphQL) and available via the rest API
* **Invisible:** cannot be retrieved via API or UI, never trigger webhook. They are purely technical, waiting to move to a visible status

### DIFF
--- a/app/graphql/types/invoices/status_type_enum.rb
+++ b/app/graphql/types/invoices/status_type_enum.rb
@@ -5,7 +5,7 @@ module Types
     class StatusTypeEnum < Types::BaseEnum
       graphql_name 'InvoiceStatusTypeEnum'
 
-      Invoice::STATUS.reject { |k| k == :generating }.each do |type|
+      Invoice::VISIBLE_STATUS.keys.each do |type|
         value type
       end
     end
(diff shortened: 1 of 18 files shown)


app/graphql/types/invoices/status_type_enum.rb
```diff
@@ -5,7 +5,7 @@ module Invoices
     class StatusTypeEnum < Types::BaseEnum
       graphql_name 'InvoiceStatusTypeEnum'
 
-      Invoice::STATUS.reject { |k| k == :generating }.each do |type|
+      Invoice::VISIBLE_STATUS.keys.each do |type|
```
rsempe, line 8: 😍

rsempe (approved):

app/graphql/types/invoices/status_type_enum.rb
julienbourdeau: Yeah, I think it's really what we're trying to communicate when we exclude `generating`

jdenquin (approved):

julienbourdeau:
TODO: make sure this is included if lovro merges first: https://github.com/getlago/lago-api/pull/2322/files#diff-244f5e3c3e029c9ff001da2140f7dc2073260370b4fb65d330951d9a22be96d3R158

PR #2316, opened 2024-07-22
Feat(anrok): create integration error details service

annvelents:
## Context

As part of implementation logic to store errors received from Anrok through Nango we need a new model that would be used to store the errors we got from integrations we have. This PR creates service to create this error

## Description

Added `IntegrationErrorDetails::CreateService`


app/services/error_details/create_service.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module ErrorDetails
+  class CreateService < BaseService
+    def initialize(params:, owner:, organization:, integration: nil)
+      super(params:, integration:, owner:, organization:)
+    end
+
+    def call
+      result = super
+      return result if result.error
```
julienbourdeau, line 11: I think in general we use `result.success?` instead of looking at the error. Especially because you're model here is an error, it gets extra confusing 😁

app/services/error_details/create_service.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      res = create_error_details!
+      return res if res&.error
+
+      res
```
julienbourdeau, line 16: Here we return `res` in any case. If you wrote it this way because you know you're going to add code after the early return in a following PR, please ignore this comment!

```suggestion
      create_error_details!
```

julienbourdeau (approved):

app/services/error_details/create_service.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module ErrorDetails
+  class CreateService < BaseService
+    def initialize(params:, owner:, organization:, integration: nil)
```
lovrocolic, line 5: I think we can remove this constructor since no new code is added inside (compared to the parent)

app/services/error_details/create_service.rb
```diff
@@ -0,0 +1,36 @@
...
+      res = create_error_details!
+      return res if res&.error
+
+      res
+    end
+
+    private
+
+    def create_error_details!
+      new_error = ErrorDetail.create(
```
lovrocolic, line 22: can you use `create!` here

app/services/error_details/create_service.rb
```diff
@@ -0,0 +1,36 @@
...
+        integration:,
+        owner:,
+        organization:,
+        error_code: params[:error_code],
+        details: params[:details]
+      )
+
+      result.error_details = new_error
+      result
+    rescue ArgumentError => e
```
lovrocolic, line 32: Can you rescue `ActiveRecord::RecordInvalid`. This one should be raises if there are validation errors

spec/services/error_details/create_service_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+  let(:integration) { create(:anrok_integration, organization:) }
+  let(:owner) { create(:invoice, organization:, customer:) }
+
+  describe '#call' do
+    subject(:service_call) { described_class.call(params:, integration:, owner:, organization:) }
+
+    let(:params) do
+      {
+        error_code: 'not_provided',
+        details: {'error_code' => 'taxDateTooFarInFuture'}
```
lovrocolic, line 18: I am curious how did you design it? It this case we will receive `taxDateTooFarInFuture` from the Anrok but shouldn't we save it as `error_code`? If not, what should we set there?

spec/services/error_details/create_service_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+          result = service_call
+
+          aggregate_failures do
+            expect(result).to be_success
+            expect(result.error_details.owner_id).to eq(owner.id)
+            expect(result.error_details.owner_type).to eq(owner.class.to_s)
+            expect(result.error_details.integration_id).to eq(integration.id)
+            expect(result.error_details.integration.class.to_s).to eq(integration.class.to_s)
+            expect(result.error_details.organization_id).to eq(organization.id)
+            expect(result.error_details.details).to eq(params[:details])
```
lovrocolic, line 38: can you please test error_code field?

app/services/error_details/create_service.rb
annvelents: yep, makes sense 🙈

spec/services/error_details/create_service_spec.rb
annvelents: originally I thought to store everything in the details and not have this separate error_code.
But then I saw that when having more generic error that can store any errors, we might want to query some specific errors for an organization, and querying by jsonb field will not be that convenient. 
so the error_code got extracted to the model level
but then 
1) we might receive more information about the error that can be stored in the jsonb field
2) I was thinking we can decide it when creating the error, but specifically for anrok we can set error_code = 'tax_error' (or just keep the error_code in the error_code and keep details empty)

app/services/error_details/base_service.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def call
+      result.not_found_failure!(resource: 'owner') unless owner
+      result.not_found_failure!(resource: 'organization') unless organization
+      result
+    end
+
+    private
+
+    attr_reader :params, :integration, :owner, :organization
```
lovrocolic, line 21: integration is extra here

annvelents: 🥇

lovrocolic (approved):

PR #2325, opened 2024-07-23
refactor(wallet): change ApplyPaidCreditsService argument

julienbourdeau:
## Description

As we're working on some wallet improvements, I was looking at this service. I'd like to make some modification:

* `*Service.call` methods should return a result as much as possible to avoid calling `raise_if_error!` on nil
* The Service is under `Wallets` namespace and only need a wallet transaction so I change the params. The invoice was not needed. I think this makes the test more straight forward too.

### DIFF
--- a/app/jobs/invoices/prepaid_credit_job.rb
+++ b/app/jobs/invoices/prepaid_credit_job.rb
@@ -8,7 +8,7 @@ module Invoices
 
     def perform(invoice)
       wallet_transaction = invoice.fees.find_by(fee_type: 'credit')&.invoiceable
-      Wallets::ApplyPaidCreditsService.new.call(wallet_transaction)
+      Wallets::ApplyPaidCreditsService.call(wallet_transaction:)
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/jobs/invoices/prepaid_credit_job.rb
```diff
@@ -7,7 +7,8 @@ class PrepaidCreditJob < ApplicationJob
     unique :until_executed, on_conflict: :log
 
     def perform(invoice)
-      Wallets::ApplyPaidCreditsService.new.call(invoice)
+      wallet_transaction = invoice.fees.find_by(fee_type: 'credit')&.invoiceable
+      Wallets::ApplyPaidCreditsService.new.call(wallet_transaction)
```
rsempe, line 11:
```suggestion
      Wallets::ApplyPaidCreditsService.call(wallet_transaction)
```

rsempe (approved):

app/jobs/invoices/prepaid_credit_job.rb
julienbourdeau: I had to reverse this commit because the service accepts the argument in the `call` method. If you want to be able to call it this way, you must accept the argument in the `initialize`

julienbourdeau: I refactored the service in the end ^^

PR #2326, opened 2024-07-23
feat(wallet): Add `invoice_require_successful_payment` attributes on wallet-related models

julienbourdeau:
## Description

We're introducing a way to topup your wallet without an invoice. The invoice is only generated after the payment is made.

Internally, when the wallet transaction is initialized, we'll create [an `open` invoice](https://github.com/getlago/lago-api/pull/2330), which is invisible and doesn't have a number yet. This invoice can be paid. Once it's paid it's going to be finalized. (see #2336)

This PR introduce a parameter to control this behavior.

* When a `WalletTransaction` is created, the `invoice_require_successful_payment` attribute can be set explicitly.
 * If not set, it follows the `Wallet` configuration
* When a wallet `RecurringTransactionRule` is created, the `invoice_require_successful_payment` attribute can be set explicitly.
  * If not set, it follows the `Wallet` configuration
* When a `WalletTransaction` is created from a `RecurringTransactionRule`, the invoice_require_successful_payment` is always read from the rule.
* All three models default to `false` in the DB

### DIFF
--- a/app/controllers/api/v1/wallet_transactions_controller.rb
+++ b/app/controllers/api/v1/wallet_transactions_controller.rb
@@ -53,7 +53,8 @@ module Api
           :wallet_id,
           :paid_credits,
           :granted_credits,
-          :voided_credits
+          :voided_credits,
+          :invoice_require_successful_payment
         )
       end
     end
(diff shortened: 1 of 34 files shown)


brunomiguelpinto (approved):

ancorcruz (approved):
LGTM!

PR #2313, opened 2024-07-19
misc-refactor-query-object-filters

ancorcruz:
## Context

TBD

## Description

TBD

PR #2327, opened 2024-07-23
Fix export invoices query object filters

ancorcruz:
## Roadmap Task

👉 https://getlago.canny.io/feature-requests/p/ability-to-export-data-from-the-user-interface

## Description

This change include a number of fixes related to query objects in general, the InvoicesQuery objects filters and how the invoices CSV export service uses the InvoicesQuery;

* Fixes query objects date parsing capabilities as original implementation was able to parse datetimes and failed with dates.
* Fix InvoicesQuery filters to ensure it works when keys are strings or symbols using the Filters object.
* Fix DataExports::CSV::Invoices service to use properly the current InvoicesQuery interface as it was missing some filters that are currently outside of the filters hash (will be fixed in https://github.com/getlago/lago-api/pull/2313)

### DIFF
--- a/app/queries/invoices_query.rb
+++ b/app/queries/invoices_query.rb
@@ -4,14 +4,14 @@ class InvoicesQuery < BaseQuery
   def call(search_term: nil, status: nil, filters: {}, customer_id: nil, payment_status: nil, payment_dispute_lost: nil, payment_overdue: nil) # rubocop:disable Metrics/ParameterLists
     @search_term = search_term
     @customer_id = customer_id
-    @filters = filters
+    @filters = filters = BaseQuery::Filters.new(filters)
 
     invoices = base_scope.result.includes(:customer)
-    invoices = invoices.where(currency: filters[:currency]) if filters[:currency]
-    invoices = with_customer_external_id(invoices) if filters[:customer_external_id]
+    invoices = invoices.where(currency: filters.currency) if filters.currency
+    invoices = with_customer_external_id(invoices) if filters.customer_external_id
     invoices = invoices.where(customer_id:) if customer_id.present?
-    invoices = invoices.where(invoice_type: filters[:invoice_type]) if filters[:invoice_type]
-    invoices = with_issuing_date_range(invoices) if filters[:issuing_date_from] || filters[:issuing_date_to]
+    invoices = invoices.where(invoice_type: filters.invoice_type) if filters.invoice_type
+    invoices = with_issuing_date_range(invoices) if filters.issuing_date_from || filters.issuing_date_to
     invoices = invoices.where(status:) if status.present?
     invoices = invoices.where(payment_status:) if payment_status.present?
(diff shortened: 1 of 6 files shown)


app/queries/invoices_query.rb
```diff
@@ -4,14 +4,14 @@ class InvoicesQuery < BaseQuery
   def call(search_term: nil, status: nil, filters: {}, customer_id: nil, payment_status: nil, payment_dispute_lost: nil, payment_overdue: nil) # rubocop:disable Metrics/ParameterLists
     @search_term = search_term
     @customer_id = customer_id
-    @filters = filters
+    @filters = filters = BaseQuery::Filters.new(filters)
```
ivannovosad, line 7:
```suggestion
    @filters = BaseQuery::Filters.new(filters)
```

app/services/data_exports/csv/invoices.rb
```diff
@@ -82,14 +82,23 @@ def headers
...
-        status = resource_query["status"]
-        filters = resource_query.except("search_term", "status")
+        filters = resource_query.except(
+          "search_term",
+          "customer_id",
+          "payment_dispute_lost",
+          "payment_overdue",
+          "payment_status",
+          "status"
+        )
```
ivannovosad, line 92:
```suggestion
        filters = resource_query.except(
          'search_term',
          'customer_id',
          'payment_dispute_lost',
          'payment_overdue',
          'payment_status',
          'status'
        )
```

app/services/data_exports/csv/invoices.rb
```diff
@@ -82,14 +82,23 @@ def headers
...
-          search_term:,
-          status:,
-          filters:
+          search_term: resource_query["search_term"],
+          filters:,
+          customer_id: resource_query["customer_id"],
+          payment_dispute_lost: resource_query["payment_dispute_lost"],
+          payment_overdue: resource_query["payment_overdue"],
+          payment_status: resource_query["payment_status"],
+          status: resource_query["status"]
```
ivannovosad, line 101:
```suggestion
          search_term: resource_query['search_term'],
          filters:,
          customer_id: resource_query['customer_id'],
          payment_dispute_lost: resource_query['payment_dispute_lost'],
          payment_overdue: resource_query['payment_overdue'],
          payment_status: resource_query['payment_status'],
          status: resource_query['status']
```

ivannovosad (changes requested):

app/queries/invoices_query.rb
ancorcruz: This is a temporary step until this PR with query objects refactor gets merged (https://github.com/getlago/lago-api/pull/2313).

I would keep it this way to be able to use `filters` within the `call` method instead of change them all to `@filters` (other wise it will use the method `filters` param that is a hash) as the goal after the refactor would be use the filters from the new Query object design...

ivannovosad: I understand, thanks for the explanation.

ivannovosad (approved):

PR #2328, opened 2024-07-24
misc(devex) - Add support for asdf. expand .gitignore with .zed/ and config/master.key

nudded:
## Description

Small DevEx improvements.

* Adds support for using [asdf](https://asdf-vm.com) as ruby version manager. 
* Also ignore .zed/ for users of [Zed](zed.dev).
* ignores `config/master.key`


julienbourdeau (approved):
I remember being pretty confused with the master.key when I installed my dev env

rsempe (approved):

PR #2329, opened 2024-07-24
Feat: anrok - add error detail to invoice serialization

annvelents:
## Context

As part of implementation logic to store errors received from Anrok through Nango we need to serialize error_details to the FE and API

## Description

Added Error Details to serializers


ancorcruz (approved):
LGTM!

app/graphql/types/error_details/object.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Types
+  module ErrorDetails
+    class Object < Types::BaseObject
+      graphql_name 'ErrorDetail'
+
+      field :error_code, Types::ErrorDetails::ErrorCodesEnum, null: false
+      field :error_details, GraphQL::Types::JSON, null: true
```
lovrocolic, line 9: What do you think of making it string? It would be easier for FE to fetch it and it follows structure used on other similar places (e.g. netsuite object). If we will have new key in error details, we will just expose it as new field in this object...

annvelents: ah, sure!
Was thinking about making it a string, but found Jsons at some places, so thought it's more common to use Jsons 🙈

lovrocolic: yep, we also use JSON somewhere, but in those cases that field is flexible and can have various keys. In our case (until we extend error details) it is just string that is describing error code. If doable, I prefer defining type to avoid errors related to it :D

PR #2331, opened 2024-07-24
fix: GoCardless payment creation exception when no mandate available

ancorcruz:
GoCardless payment creation does not raise error when no mandate available

Instead of raising an error when a mandate is not available at GoCardless to create a new payment,

We notify the client with a meaningful error message, flag the invoice as payment failed and return a failure result.

## Context

Given a customer does not have a GoCardless mandate available to process an invoice payment, `Invoices::Payments::GocardlessCreateJob` was raising an exception, moving the job to sidekiq dead jobs queue and notifying the client with not helpful information `"One of your parameters was incorrectly typed"`.

```
job_name: "Invoices::Payments::GocardlessCreateJob"
error_class: "GoCardlessPro::InvalidApiUsageError"
error_message: "One of your parameters was incorrectly typed"
```

## Description

This change fixes this issue exiting the job with a service failure result, notifying the client pointing the customer has no available mandate for the payment and updates the invoice as payment failed.

### DIFF
--- a/app/services/invoices/payments/gocardless_service.rb
+++ b/app/services/invoices/payments/gocardless_service.rb
@@ -6,7 +6,7 @@ module Invoices
       include Customers::PaymentProviderFinder
 
       class MandateNotFoundError < StandardError
-        DEFAULT_MESSAGE = 'No mandate avaiable for payment'
+        DEFAULT_MESSAGE = 'No mandate available for payment'
 
         attr_reader :code
 
--- a/app/services/invoices/payments/gocardless_service.rb
+++ b/app/services/invoices/payments/gocardless_service.rb
@@ -6,14 +6,15 @@ module Invoices
       include Customers::PaymentProviderFinder
 
       class MandateNotFoundError < StandardError
-        DEFAULT_MESSAGE = 'No mandate available for payment'
-
-        attr_reader :code
+        DEFAULT_MESSAGE = "No mandate available for payment"
+        ERROR_CODE = "no_mandate_error"
 
         def initialize(msg = DEFAULT_MESSAGE)
-          @code = '999'
-
           super
+        end
+
+        def code
+          ERROR_CODE
         end
       end
 
--- a/app/services/invoices/payments/gocardless_service.rb
+++ b/app/services/invoices/payments/gocardless_service.rb
@@ -5,6 +5,18 @@ module Invoices
     class GocardlessService < BaseService
       include Customers::PaymentProviderFinder
 
+      class MandateNotFoundError < StandardError
+        DEFAULT_MESSAGE = 'No mandate avaiable for payment'
+
+        attr_reader :code
+
+        def initialize(msg = DEFAULT_MESSAGE)
+          @code = '999'
+
+          super
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


app/services/invoices/payments/gocardless_service.rb
```diff
@@ -5,6 +5,18 @@ module Payments
     class GocardlessService < BaseService
       include Customers::PaymentProviderFinder
 
+      class MandateNotFoundError < StandardError
+        DEFAULT_MESSAGE = 'No mandate avaiable for payment'
+
+        attr_reader :code
+
+        def initialize(msg = DEFAULT_MESSAGE)
+          @code = '999'
```
ancorcruz, line 14: no idea what error code use here...

ancorcruz:
@michaelponrajah @sarkissianraffi What do you think about the error message?

app/services/invoices/payments/gocardless_service.rb
```diff
@@ -5,6 +5,18 @@ module Payments
     class GocardlessService < BaseService
       include Customers::PaymentProviderFinder
 
+      class MandateNotFoundError < StandardError
+        DEFAULT_MESSAGE = 'No mandate avaiable for payment'
```
ivannovosad, line 9:
```suggestion
        DEFAULT_MESSAGE = 'No mandate available for payment'
```

ivannovosad (changes requested):

app/services/invoices/payments/gocardless_service.rb
ancorcruz: Thanks for catching the typo.

app/services/invoices/payments/gocardless_service.rb
```diff
@@ -5,6 +5,18 @@ module Payments
     class GocardlessService < BaseService
       include Customers::PaymentProviderFinder
 
+      class MandateNotFoundError < StandardError
+        DEFAULT_MESSAGE = 'No mandate available for payment'
+
+        attr_reader :code
+
+        def initialize(msg = DEFAULT_MESSAGE)
+          @code = '999'
```
ivannovosad, line 14: We can use something like `No mandate error` or `no_mandate_error`.

spec/services/invoices/payments/gocardless_service_spec.rb
```diff
@@ -174,6 +174,44 @@
...
+          .and_raise(GoCardlessPro::Error.new('code' => 'code', 'message' => 'error'))
+      end
+
+      it 'delivers an error webhook' do
+        result = gocardless_service.create
+
+        aggregate_failures do
+          expect(result).not_to be_success
+          expect(result.error).to be_a(BaseService::ServiceFailure)
+          expect(result.error.code).to eq('999')
```
ivannovosad, line 196: In case of changing the code we'll need to change it here as well.

ivannovosad (changes requested):

app/services/invoices/payments/gocardless_service.rb
ancorcruz: looks good

julienbourdeau (approved):

PR #2332, opened 2024-07-25
misc: add category to integration

annvelents:
## Context

We have integrations of different categories, but now define type of an integration is possible only by the name of integrations, so if we know that integration type is NetsuiteIntegration or XeroIntegration, then we can define that this is an accounting integration. Instead we can store category on the integration itself, which will allow us to have more generic components on the FE, find corresponding integration for an error (when error type is "accounting" we will know which integrations we need to look to be set up on a client)

## Description

Added string :category to integrations table and serialized it to FE

### DIFF
--- a/app/models/credit_note.rb
+++ b/app/models/credit_note.rb
@@ -106,7 +106,7 @@ class CreditNote < ApplicationRecord
   end
 
   def should_sync_credit_note?
-    finalized? && customer.integration_customers.accounting_kind.any? { |c| c.integration.sync_credit_notes }
+    finalized? && customer.accounting_customer.integration.sync_credit_notes
   end
 
   def voidable?
--- a/app/models/credit_note.rb
+++ b/app/models/credit_note.rb
@@ -106,7 +106,7 @@ class CreditNote < ApplicationRecord
   end
 
   def should_sync_credit_note?
-    finalized? && customer.accounting_customer.integration.sync_credit_notes
+    finalized? && customer.accounting_customer&.integration&.sync_credit_notes
   end
 
   def voidable?
--- a/app/models/credit_note.rb
+++ b/app/models/credit_note.rb
@@ -106,7 +106,7 @@ class CreditNote < ApplicationRecord
   end
 
   def should_sync_credit_note?
-    finalized? && customer.accounting_customer&.integration&.sync_credit_notes
+    finalized? & customer.accounting_customer&.integration&.sync_credit_notes
   end
 
   def voidable?
--- a/app/models/customer.rb
+++ b/app/models/customer.rb
@@ -71,6 +71,8 @@ class Customer < ApplicationRecord
   validates :payment_provider, inclusion: {in: PAYMENT_PROVIDERS}, allow_nil: true
   validates :timezone, timezone: true, allow_nil: true
 
+  alias :provider_customer :payment_provider_customer
+
   def self.ransackable_attributes(_auth_object = nil)
     %w[id name external_id email]
   end
@@ -113,10 +115,6 @@ class Customer < ApplicationRecord
     return document_locale.to_sym if document_locale?
 
     organization.document_locale.to_sym
-  end
-
-  def provider_customer
-    payment_provider_customer
   end
 
   def shipping_address
--- a/app/models/customer.rb
+++ b/app/models/customer.rb
@@ -71,7 +71,7 @@ class Customer < ApplicationRecord
   validates :payment_provider, inclusion: {in: PAYMENT_PROVIDERS}, allow_nil: true
   validates :timezone, timezone: true, allow_nil: true
 
-  alias :provider_customer :payment_provider_customer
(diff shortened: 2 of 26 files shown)


db/migrate/20240724154417_add_category_to_integrations.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def up
+    add_column :integrations, :category, :integer
+    add_index :integrations, :category
+
+    Integrations::AnrokIntegration.update_all(category: 'tax_provider')
+    Integrations::NetsuiteIntegration.update_all(category: 'accounting')
+    Integrations::XeroIntegration.update_all(category: 'accounting')
+    Integrations::OktaIntegration.update_all(category: 'system')
```
annvelents, line 29: Run it locally on 4000 integrations - took 0.04s

db/migrate/20240724154417_add_category_to_integrations.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def up
+    add_column :integrations, :category, :integer
+    add_index :integrations, :category
+
+    Integrations::AnrokIntegration.update_all(category: 'tax_provider') # rubocop:disable Rails/SkipsModelValidations
+    Integrations::NetsuiteIntegration.update_all(category: 'accounting') # rubocop:disable Rails/SkipsModelValidations
+    Integrations::XeroIntegration.update_all(category: 'accounting') # rubocop:disable Rails/SkipsModelValidations
+    Integrations::OktaIntegration.update_all(category: 'system') # rubocop:disable Rails/SkipsModelValidations
```
annvelents, line 31: Run it locally on 4000 integrations - took 0.04s

ivannovosad (commented):
I'm thinking that if we don't necessarily need the category for BE feature/code at the moment to leave this PR open and lets get back to it later.

Or is it handy to have for some FE stuff? @ansmonjol @keellyp

annvelents:
actually I accidentally found more places that can be refactored using the concept of categories, moving this PR to draft and will update

app/models/credit_note.rb
```diff
@@ -106,7 +106,7 @@ def add_on_items
   end
 
   def should_sync_credit_note?
-    finalized? && customer.integration_customers.accounting_kind.any? { |c| c.integration.sync_credit_notes }
+    finalized? & customer.accounting_customer&.integration&.sync_credit_notes
```
annvelents, line 109: updated to `bitwise and` because `logical and` returns value if it exist instead of true/false
So in our case if second half returns nil, 
&& - returns nil for the whole operation
& - returns false for the whole operation,
no matter what first argument was

app/models/customer.rb
```diff
@@ -105,14 +116,7 @@ def preferred_document_locale
   end
 
   def provider_customer
```
julienbourdeau, line 118: I assume you're keeping this method because it's used all over the place?

app/models/credit_note.rb
julienbourdeau: would `&& customer.accounting_customer&.integration&.sync_credit_notes.present?` work here? I have nothing against "bitwise and" but without you're comment I'm not sure I'd have even noticed ^^

julienbourdeau (approved):

app/models/customer.rb
annvelents: yeah, becasue there are quite a lot of places we're using it 🙈 
but actually i think declaring this method as alias will be better 😅 
thank you for putting my attention here

lovrocolic:
> I'm thinking that if we don't necessarily need the category for BE feature/code at the moment to leave this PR open and lets get back to it later.
> 
> Or is it handy to have for some FE stuff? @ansmonjol @keellyp

@ansmonjol @keellyp Do you think that having category would be handy on the FE side?

My feeling regarding this refactoring is that too many things got changed at once and it would require proper QA before getting merged.
However, it would be pretty complex to test everything at once (payment providers flow, accounting integration flow...), so I would prefer splitting this PR in few parts:
- Adding category field if that makes sense for the FE team
- Refactor payment providers
- Refactor accounting/tax integrations
@annvelents What do you think? Does it make sense to you?

annvelents:
@ansmonjol the original reason was to make it easier to show error for integration, because we were going to associate errors with integrations, and when displaying errors having this doesn't seem right 🙈 
"if (integration.type == 'AnrokIntegration' || integration.type == 'FutureTaxProvider') { puts 'tax error' }" (but this association was changed on the BE side, so it doesnt solve this problem anymore )
then I also looked at how currently integrations are displayed at the FE and realised it will be helpful for you to have categories - to display integrations page, because you have categories technically, but they are just static fields
also because then for example, while connecting customers with integrations, you'll be able to pull integrations with type "accounting" and iterate through them. As well as for validations: we will need only check that customer should have one integration with a specific category

annvelents:
I'm using it as base for another one I'm still working on.. But this one actually can be closed, thank you

PR #2337, opened 2024-07-26
fix(tax-integrations): add small improvements to the retry service

lovrocolic:
## Context

Currently Lago is implementing tax integrations

## Description

This PR adds small improvements to the retry service

### DIFF
--- a/app/services/invoices/retry_service.rb
+++ b/app/services/invoices/retry_service.rb
@@ -15,13 +15,12 @@ module Invoices
       taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateService.call(invoice:, fees: invoice.fees)
 
       unless taxes_result.success?
-        return result.service_failure!(code: 'taxes', message: 'fetching failed')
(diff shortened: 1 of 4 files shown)


annvelents (approved):

app/services/invoices/retry_service.rb
```diff
@@ -78,10 +78,13 @@ def credit_notes
     end
 
     def should_create_credit_note_credit?
+      return false if invoice.one_off?
+
       credit_notes.any?
     end
 
     def should_create_applied_prepaid_credit?
+      return false if invoice.one_off?
```
julienbourdeau, line 87: I had a quick and I don't see this condition on the first payment. It seems that you can pay the first time with credit but not when retrying?

lovrocolic: Actually credit notes / prepaid credits are not applied at all on one off invoices. We don't use those credits in main service so those should be skipped also when retrying

PR #2338, opened 2024-07-26
Feat clean expired topup invoices

brunomiguelpinto:
## Description

This PR introduces a scheduled cleanup job that automatically deletes pending wallet transactions and corresponding unpaid open invoices that are older than 90 days.

### DIFF
--- a/app/jobs/clock/expired_invoices_cleanup_job.rb
+++ b/app/jobs/clock/expired_invoices_cleanup_job.rb
@@ -1,0 +1,13 @@
+# app/jobs/expired_invoices_cleanup_job.rb
+class ExpiredInvoicesCleanupJob < ApplicationJob
+  queue_as :default
+
+  def perform()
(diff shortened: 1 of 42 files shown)


app/jobs/clock/expired_invoices_cleanup_job.rb
```diff
@@ -0,0 +1,13 @@
+# app/jobs/expired_invoices_cleanup_job.rb
+class ExpiredInvoicesCleanupJob < ApplicationJob
+  queue_as :default
+
+  def perform()
+    expired_invoices = Invoice.open.where('created_at < ?', 90.days.ago)
```
brunomiguelpinto, line 6: open credit 
and 
the payment failed or pending

brunomiguelpinto: wallets older 90 invoices pending  and the new flag
grab the invoice id ->

PR #2340, opened 2024-07-26
fix: invoices query payment_overdue filter

ancorcruz:
## Context

Given `payment_overdue` filter is false, `InvoicesQuery` ignores the filter and return all invoices without consider their `payment_overdue` value, while it should only return `payment_overdue` == false invoices.

## Description

This change fixes the filter by providing the proper scoping base on the value of the filter.

### DIFF
--- a/spec/queries/invoices_query_spec.rb
+++ b/spec/queries/invoices_query_spec.rb
@@ -241,7 +241,39 @@ RSpec.describe InvoicesQuery, type: :query do
         payment_overdue: true
       )
 
-      expect(result.invoices.pluck(:id)).to eq([invoice_third.id])
+      returned_ids = result.invoices.pluck(:id)
+
+      aggregate_failures do
+        expect(returned_ids.count).to eq(1)
+        expect(returned_ids).not_to include(invoice_first.id)
+        expect(returned_ids).not_to include(invoice_second.id)
+        expect(returned_ids).to include(invoice_third.id)
+        expect(returned_ids).not_to include(invoice_fourth.id)
+        expect(returned_ids).not_to include(invoice_fifth.id)
+        expect(returned_ids).not_to include(invoice_sixth.id)
+      end
+    end
+  end
+
+  context 'when filtering by payment overdue false' do
+    it 'returns expected invoices' do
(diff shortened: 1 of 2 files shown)


spec/queries/invoices_query_spec.rb
```diff
@@ -241,7 +241,39 @@
...
+
+      aggregate_failures do
+        expect(returned_ids.count).to eq(1)
+        expect(returned_ids).not_to include(invoice_first.id)
+        expect(returned_ids).not_to include(invoice_second.id)
+        expect(returned_ids).to include(invoice_third.id)
+        expect(returned_ids).not_to include(invoice_fourth.id)
+        expect(returned_ids).not_to include(invoice_fifth.id)
+        expect(returned_ids).not_to include(invoice_sixth.id)
+      end
```
rsempe, line 254:
```suggestion
      expect(returned_ids).to contain_exactly(invoice_third.id)
```

spec/queries/invoices_query_spec.rb
```diff
@@ -241,7 +241,39 @@
...
+
+      aggregate_failures do
+        expect(returned_ids.count).to eq(5)
+        expect(returned_ids).to include(invoice_first.id)
+        expect(returned_ids).to include(invoice_second.id)
+        expect(returned_ids).not_to include(invoice_third.id)
+        expect(returned_ids).to include(invoice_fourth.id)
+        expect(returned_ids).to include(invoice_fifth.id)
+        expect(returned_ids).to include(invoice_sixth.id)
+      end
```
rsempe, line 276:
```suggestion
      expect(returned_ids).to contain_exactly(
        invoice_first.id,
        invoice_second.id,
        invoice_fourth.id,
        invoice_fifth.id,
        invoice_sixth.id
      )
```

rsempe (approved):
👍

spec/queries/invoices_query_spec.rb
ancorcruz: I found harder to identify which one is missing or should not be included with `contain_exactly`. The test is more verbose but this is a good trade off.

PR #2341, opened 2024-07-26
Feat(anrok): add handling discarded errors when retrying invoice

annvelents:
## Context

When retrying an invoice we want to mark all previous errors as discarded, so if the invoice succeeded these errors will not be shown

## Description

- Update deleted_at for existing tax_error error_details  when retrying fetch taxes for the invoice
- Create new errors when retrying invoice fetching taxes failed
- Serialize only error_details that are not discarded

### DIFF
--- a/app/services/invoices/retry_service.rb
+++ b/app/services/invoices/retry_service.rb
@@ -16,7 +16,8 @@ module Invoices
       taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateService.call(invoice:, fees: invoice.fees)
 
       unless taxes_result.success?
-        return result.validation_failure!(errors: {tax_error: [taxes_result.error.code]})
+        create_error_detail(taxes_result.error.code)
+        return result.service_failure!(code: 'tax_error', message: taxes_result.error.code)
       end
 
       provider_taxes = taxes_result.fees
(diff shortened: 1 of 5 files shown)


app/services/invoices/retry_service.rb
```diff
@@ -12,10 +12,12 @@ def call
       return result.not_found_failure!(resource: 'invoice') unless invoice
       return result.not_allowed_failure!(code: 'invalid_status') unless invoice.failed?
 
+      invoice.error_details.tax_error.kept.update_all(deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
       taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateService.call(invoice:, fees: invoice.fees)
 
       unless taxes_result.success?
-        return result.validation_failure!(errors: {tax_error: [taxes_result.error.code]})
+        create_error_detail(taxes_result.error.code)
+        return result.service_failure!(code: 'tax_error', message: taxes_result.error.code)
```
lovrocolic, line 20: We should return `return result.validation_failure!(errors: {tax_error: [taxes_result.error.code]})` here. Even the source of this error is external service, this `RetryService` returns errors directly to the caller. On GQL side and api side we expose only several types of errors. In this case it makes sense to return 422. Please take a look at `execution_error_responder.rb`. Without returning validation error this code would return 500 to the GQL

annvelents: ahaaaa, thank you! 👍 
checking it now

lovrocolic (approved):

PR #2345, opened 2024-07-29
misc(events): Add new optimized index

jdenquin:
## Context

- We have pretty bad performances on the `events` table when we want to retreive events for aggregations and before computing it.

## Description

- Add the missing optimized index on `external_subscription_id`, `code`  and `timestamp` and including `organization_id` and `properties`

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_07_23_150304) do
+ActiveRecord::Schema[7.1].define(version: 2024_07_29_154334) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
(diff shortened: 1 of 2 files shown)


brunomiguelpinto (approved):

ancorcruz (approved):

db/schema.rb
```diff
@@ -755,6 +756,7 @@
     t.index ["organization_id"], name: "index_invoices_on_organization_id"
     t.index ["payment_overdue"], name: "index_invoices_on_payment_overdue"
     t.index ["sequential_id"], name: "index_invoices_on_sequential_id"
+    t.index ["status"], name: "index_invoices_on_status"
```
rsempe, line 759: why this?

rsempe (approved):

db/schema.rb
jdenquin: I added it last week and I think it was missing in the schema!

nudded (approved):

annvelents (approved):

PR #2346, opened 2024-07-29
feat(dunning): Introduce PayableGroup model

rsempe:


db/migrate/20240729151049_add_payable_type_and_payable_id_to_payments.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddPayableTypeAndPayableIdToPayments < ActiveRecord::Migration[7.1]
+  def change
+    add_column :payments, :payable_type, :string, null: false, default: "Invoice"
+    rename_column :payments, :invoice_id, :payable_id
```
rsempe, line 6: add_column

app/models/payable_group.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class PayableGroup < ApplicationRecord
+  include PaperTrailTraceable
+
+  has_many :invoices
+  has_many :payments, as: :payable
+
+  def customer
+    invoices.first&.customer
```
annvelents, line 10: since we have reference to a customer in the table, do we still need it?

rsempe: good spot 👍

annvelents (approved):

PR #2347, opened 2024-07-30
misc(migrations) - Add extra check to make sure schema.rb is up to date.

nudded:
## Description

Make sure schema.rb is up to date with latest migrations.


nudded:
> Interesting. My approach was to check git after the migration were executed. https://github.com/getlago/lago-api/pull/2272/files

I've tried this as well, but by default the schema.rb is not dumped when running the migrations. (And when debugging with `act` it does not do a checkout, so using simple diff was easier).

PR #2349, opened 2024-07-31
misc(devex) - Add annotations to model files

nudded:
## Description

Add annotations to model files and setup Rails to automatically annotate when running a migration.


ancorcruz (approved):
I prefer the annotations on top but this is great anyways...

PR #2350, opened 2024-07-31
feat(netsuite): Add lago customer link to netsuite

ivannovosad:
## Context

By creating or updating a customer from Lago to NetSuite, a client needs us to send a custom field with the hyperlink of our Lago customer.

## Description

Add `custentity_lago_customer_link` to payload body when saving a netsuite customer.


lovrocolic (approved):
🚀

PR #2358, opened 2024-08-01
Feat(anrok): add finalize_all_invoices job and service

annvelents:
## Context

We want to provide an additional functionality to finalize all draft invoices for organization. To do so we needed invoices/finalize_all_job and invoices/finalize_batch_service

## Description

Added job and service to finalize all draft invoices for organization and mutation to trigger the job


lovrocolic (approved):

app/services/invoices/finalize_batch_service.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      result
+    end
+
+    def call(invoice_ids)
+      processed_invoices = []
+      Invoice.where(id: invoice_ids).find_each do |invoice|
+        result = Invoices::RefreshDraftAndFinalizeService.new(invoice:).call
+
+        return result unless result.success?
```
julienbourdeau, line 24: I think we should continue to process all the invoices and not fail after the first error.
A buggy invoice would prevent the batch to ever finish.

I'd return 2 list of invoices in the final result and set error if there was any error

```rb
result.processed_invoices = processed_invoices.compact
result.failed_invoices = failed.compact

result.fail_with_error!(error: :something)

result
```

app/services/invoices/finalize_batch_service.rb
```diff
@@ -0,0 +1,42 @@
...
+module Invoices
+  class FinalizeBatchService < BaseService
+    def initialize(organization:)
+      @organization = organization
+
+      super
+    end
+
+    def call_async
+      Invoices::FinalizeAllJob.perform_later(organization:, invoice_ids: invoices.ids)
```
julienbourdeau, line 12: What's the reason to finalize in a loop rather than enqueue single jobs to process in parallel.

For sure one benefit will be to avoid the Sequence error 😄  (see https://github.com/getlago/lago-api/pull/2364) but I'm wondering.

app/services/invoices/finalize_batch_service.rb
annvelents: that's a tricky one, because if we failed with an integration, possible that it means that the integration is not set up correctly, and therefore we'll fail with all further fetches as well...
And anrok is slow in responding + they have their limits... 🤔  so in this case it might be better to be sure that the error is fixed before continuing

app/services/invoices/finalize_batch_service.rb
annvelents: to be honest, I was following the pattern we have with `invoices/retry_batch_service` and `invoices/payments/retry_batch_service`, and exactly because we already have the pattern I'd prefer follow it for now, and if we decide to move to parallel, it'd be better to change them alltogether 🙈

julienbourdeau: That's a good reason 😄 👌

PR #2362, opened 2024-08-02
perf(wallets): Improve job enqueuing

julienbourdeau:
## Context

Every 5 minutes, we enqueues a few thousands of jobs to update the wallet balance.

## Description

I'm proposing a few improvements to how we enqueue all the `Wallets::RefreshOngoingBalanceJob`s.

### 1. Use `select(:id)`

Instead of retrieving all the wallet columns and hydrated a full models, we'll only retrieve the ID because all we need is to be able to generate a Global ID for each model. When the job is executed, it retrieves the full model.

CleanShot 2024-08-02 at 10 31 10@2x

### 2. Use `perform_all_later`

Since we updated recently to Rails 7.1, we can now [use `perform_all_later`](https://blog.saeloun.com/2023/11/27/rails-7-1-adds-active-job-perform-all-later/). [Official docs](https://edgeguides.rubyonrails.org/active_job_basics.html#bulk-enqueuing).

With `perform_all_later`, a 1000 wallets to refresh will write 10 times to Redis, enqueuing 100 jobs each time.
With `perform_later`,  a 1000 wallets to refresh will write 1000 times to Redis, enqueuing 1 job at a time.

### DIFF
--- a/app/jobs/clock/refresh_wallets_ongoing_balance_job.rb
+++ b/app/jobs/clock/refresh_wallets_ongoing_balance_job.rb
@@ -14,7 +14,7 @@ module Clock
       jobs = []
       batch_size = 100
 
-      Wallet.active.select(:id).find_each do |wallet|
+      Wallet.active.where(updated_at: ..1.minute.ago).select(:id).find_each do |wallet|
         jobs << Wallets::RefreshOngoingBalanceJob.new(wallet)
 
         if jobs.size >= batch_size
--- a/app/jobs/clock/refresh_wallets_ongoing_balance_job.rb
+++ b/app/jobs/clock/refresh_wallets_ongoing_balance_job.rb
@@ -11,9 +11,19 @@ module Clock
     def perform
       return unless License.premium?
 
-      Wallet.active.find_each do |wallet|
-        Wallets::RefreshOngoingBalanceJob.perform_later(wallet)
+      jobs = []
+      batch_size = 100
+
+      Wallet.active.select(:id).find_each do |wallet|
+        jobs << Wallets::RefreshOngoingBalanceJob.new(wallet)
+
+        if jobs.size >= batch_size
+          ActiveJob.perform_all_later(jobs)
+          jobs = []
+        end
       end
+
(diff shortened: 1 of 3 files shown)


jdenquin (approved):

lovrocolic (approved):

brunomiguelpinto (approved):

app/jobs/clock/refresh_wallets_ongoing_balance_job.rb
```diff
@@ -11,9 +11,19 @@ class RefreshWalletsOngoingBalanceJob < ApplicationJob
...
+      Wallet.active.select(:id).find_each do |wallet|
+        jobs << Wallets::RefreshOngoingBalanceJob.new(wallet)
+
+        if jobs.size >= batch_size
+          ActiveJob.perform_all_later(jobs)
+          jobs = []
+        end
       end
+
+      ActiveJob.perform_all_later(jobs)
```
ivannovosad, line 26:
```suggestion
      Wallet.active.select(:id).in_batches(of: 100) do |wallets|
        jobs = wallets.map { |wallet| Wallets::RefreshOngoingBalanceJob.new(wallet) }

        ActiveJob.perform_all_later(jobs)
      end
```

app/jobs/wallets/refresh_ongoing_balance_job.rb
```diff
@@ -4,9 +4,14 @@ module Wallets
...
     queue_as 'wallets'
 
-    unique :until_executed, on_conflict: :log
+    unique :while_executing, on_conflict: :log
 
     def perform(wallet)
+      # If the wallet was updated recently, it's probably because this same job was executed
+      # Since it runs every 5.minutes, even if the was was modified for other reasons and
+      # the balance wasn't updated, it will be updated in the next run.
+      return if wallet.updated_at < 1.minute.ago
```
nudded, line 13: This should be `updated_at > 1.minute.ago` right? 🤔

julienbourdeau:
To be explored later

PR #2366, opened 2024-08-02
chore(billing): Add logs on BillingService

julienbourdeau:
## Description

Add a log to count how many customers are billed at every `BillingService` run.

Note that we don't use `find_each` here because `find_by_sql` always returns all the results.

### DIFF
--- a/app/services/subscriptions/billing_service.rb
+++ b/app/services/subscriptions/billing_service.rb
@@ -5,7 +5,11 @@ module Subscriptions
     def call(billing_at: Time.current)
       @today = billing_at
 
-      billable_subscriptions.group_by(&:customer_id).each do |_customer_id, customer_subscriptions|
+      billable_subscriptions_per_customer = billable_subscriptions.group_by(&:customer_id)
+      total = billable_subscriptions_per_customer.count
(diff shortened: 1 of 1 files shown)


app/services/subscriptions/billing_service.rb
```diff
@@ -5,7 +5,11 @@ class BillingService < BaseService
     def call(billing_at: Time.current)
       @today = billing_at
 
-      billable_subscriptions.group_by(&:customer_id).each do |_customer_id, customer_subscriptions|
+      billable_subscriptions_per_customer = billable_subscriptions.group_by(&:customer_id)
+      total = billable_subscriptions_per_customer.count
+      Rails.logger.info "[#{self.class.name}] Billing #{total} customers"
```
nudded, line 10: Perhaps a suggestion to wait for the tracing improvements (https://github.com/getlago/lago-api/pull/2357) so you could extend the current span with this attribute? 🤔 (that should make graphing afterwards easier)

julienbourdeau: I'd 1000x prefer a metric available to make a bar chart! 👍 
It's rediscuss this when I return from holidays (in 2 weeks!)

PR #2368, opened 2024-08-02
feat(tresholds): Add models

ivannovosad:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/progressive-billing

## Context

There's always a risk of customers not paying invoices generated at the end of a billing period. It would be beneficial to bill customers at given thresholds (units vs. amount) rather than waiting until the end of the period. This approach would allow for removing customer access if invoices are not paid and prevent having a highest amount of unpaid invoices.

## Description

This PR adds migration and model for progressive billing tresholds.


db/migrate/20240802115017_create_progressive_billing_tresholds.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.string :treshold_display_name
+      t.bigint :amount_cents, null: false
+      t.string :amount_currency, null: false
+      t.boolean :recurring, null: false, default: false
+
+      t.timestamps
+    end
+
+    add_index :progressive_billing_tresholds, %i[amount_cents plan_id recurring], unique: true
+    add_index :progressive_billing_tresholds, %i[recurring plan_id], unique: true, where: "recurring is true"
```
nudded, line 16: I thought it was better to have high-cardinality columns first, especially as `recurring` will be true always. 🤔

nudded (approved):

PR #2373, opened 2024-08-05
fix(wrong-month-abbreviation): use month full names instead

ancorcruz:
## Context

 Invoices and emails have been sent with dates with month abbreviated,
 however the abbreviation was wrong for some months and languages, also
 the format was off in some cases.

related: https://github.com/getlago/lago/issues/386

 ## Description

 This change fixes this issue changing the date format with month full
 name.

 Also, correct the month name abbreviations

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -10,19 +10,18 @@ fr:
     - Ven
     - Sam
     abbr_month_names:
-    -
-    - Jan
-    - Fév
-    - Mar
-    - Avr
-    - Mai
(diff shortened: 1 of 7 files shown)


config/locales/fr.yml
```diff
@@ -10,19 +10,18 @@ fr:
     - Ven
     - Sam
     abbr_month_names:
```
rsempe, line 12: are we sure we want them as lower case?

rsempe (approved):

PR #2374, opened 2024-08-05
fix: Bill subscription when created with subscription date in the past

rsempe:
## Context

When a subscription is created with a subscription date in the past, for a paid-in-advance plan, we're not generating any invoices before the next billing cycle.

## Description

The goal of this PR is to be able to generate an invoice for the pay-in-advance subscription fee immediately when the subscription is created.

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -91,7 +91,9 @@ module Subscriptions
     end
 
     def should_be_billed_today?(sub)
-      sub.active? && sub.subscription_at.today? && plan.pay_in_advance? && !sub.in_trial_period?
+      return false unless sub.subscription_at.today? || sub.subscription_at.past?
+
+      sub.active? && plan.pay_in_advance? && !sub.in_trial_period?
     end
 
     def create_subscription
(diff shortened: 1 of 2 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -91,7 +91,9 @@ def downgrade?
     end
 
     def should_be_billed_today?(sub)
-      sub.active? && sub.subscription_at.today? && plan.pay_in_advance? && !sub.in_trial_period?
+      return false unless sub.subscription_at.today? || sub.subscription_at.past?
```
vincent-pochet, line 94: I might miss some context here, but in my mind the current behavior was intentional, to allow migration existing subscriptions from an external tool into Lago without billing them.

nudded: When playing around with Lago and pay in advance plans I also noticed this and we have dedicated documentation for this behaviour as well. 

> If the plan includes a subscription fee to be paid in advance, it will be considered as already paid for the current period.

rsempe: yep - that's something in discussion - closing the PR for now

rsempe:
Closed in favor of https://github.com/getlago/lago-api/pull/2388

PR #2377, opened 2024-08-06
feat: add the source to the serializer

brunomiguelpinto:
## Context

This PR introduces the functionality to attach metadata to wallet transactions to differentiate between automated and manual top-ups. Metadata can only be added through the API during the creation of a wallet or when topping up a wallet with a new transaction.

## Description

1. **Metadata on Wallet Creation:** Users can now assign metadata to a transaction at the point of wallet creation. This metadata is exclusive to the wallet transaction triggered during the creation process.
   
2. **Metadata on Wallet Top-Up:** Metadata can also be included when topping up a wallet to provide additional transaction details.

**Limitations:**
- Metadata is not searchable.
- The number of metadata entries per transaction is capped at the same limit as customer-level metadata.
- Metadata can be set only through API calls, not through the UI.
- Applicable for both initial wallet creation and subsequent top-up transactions.

### DIFF
--- a/app/services/wallets/create_interval_wallet_transactions_service.rb
+++ b/app/services/wallets/create_interval_wallet_transactions_service.rb
@@ -6,7 +6,7 @@ module Wallets
       recurring_transaction_rules.each do |recurring_transaction_rule|
         wallet = recurring_transaction_rule.wallet
 
-        WalletTransactions::CreateJob.perform_later(
+        job_params = {
           organization_id: wallet.organization.id,
           params: {
             wallet_id: wallet.id,
@@ -15,7 +15,10 @@ module Wallets
             source: :interval,
             invoice_requires_successful_payment: recurring_transaction_rule.invoice_requires_successful_payment?
           }
-        )
+        }
+        job_params[:params][:metadata] = recurring_transaction_rule.metadata unless recurring_transaction_rule.metadata.empty?
+
+        WalletTransactions::CreateJob.perform_later(job_params)
       end
     end
 
--- a/app/services/wallets/threshold_top_up_service.rb
+++ b/app/services/wallets/threshold_top_up_service.rb
@@ -12,16 +12,19 @@ module Wallets
       return if wallet.credits_ongoing_balance > threshold_rule.threshold_credits
       return if (pending_transactions_amount + wallet.credits_ongoing_balance) > threshold_rule.threshold_credits
 
-      WalletTransactions::CreateJob.set(wait: 2.seconds).perform_later(
+      job_params = {
         organization_id: wallet.organization.id,
         params: {
           wallet_id: wallet.id,
-          paid_credits:,
-          granted_credits:,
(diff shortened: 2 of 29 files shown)


app/services/wallets/create_interval_wallet_transactions_service.rb
```diff
@@ -15,7 +15,10 @@ def call
             source: :interval,
             invoice_requires_successful_payment: recurring_transaction_rule.invoice_requires_successful_payment?
           }
-        )
+        }
+        job_params[:params][:metadata] = recurring_transaction_rule.metadata unless recurring_transaction_rule.metadata.empty?
+
+        WalletTransactions::CreateJob.perform_later(job_params)
```
ancorcruz, line 21: why not? is the key not included when value is empty to avoid updating the tests?

```
  WalletTransactions::CreateJob.perform_later(
          organization_id: wallet.organization.id,
          params: {
            wallet_id: wallet.id,
            paid_credits: paid_credits(recurring_transaction_rule),
            granted_credits: granted_credits(recurring_transaction_rule),
            source: :interval,
            invoice_requires_successful_payment: recurring_transaction_rule.invoice_requires_successful_payment?,
            metadata: recurring_transaction_rule.metadata
          }
  )
```

is there an issue if `metadata` key is included with an empty hash as value?

app/services/wallets/threshold_top_up_service.rb
```diff
@@ -12,16 +12,19 @@ def call
...
-          paid_credits:,
-          granted_credits:,
+          paid_credits: paid_credits,
+          granted_credits: granted_credits,
           source: :threshold,
           invoice_requires_successful_payment: threshold_rule.invoice_requires_successful_payment?
         }
-      )
+      }
+      job_params[:params][:metadata] = threshold_rule.metadata unless threshold_rule.metadata.empty?
```
ancorcruz, line 25: the same question here....

spec/services/wallets/recurring_transaction_rules/validate_service_spec.rb
```diff
@@ -17,6 +17,20 @@
...
+      let(:params) do
+        {
+          trigger: "interval",
+          interval: "weekly",
+          metadata: {'key' => {'nested_key' => 'value'}}
+        }
+      end
+
+      it "returns false" do
+        expect(validate_service.call).to be_falsey
```
ancorcruz, line 30: `be_falsey` will match with nil and false... as this service returns only true or false I would set the expectation to false instead of falsey...

```suggestion
        expect(validate_service.call).to eq false
```

ancorcruz (approved):
Good stuff, I have left a few questions.

app/services/wallets/create_interval_wallet_transactions_service.rb
brunomiguelpinto: no problem

PR #2378, opened 2024-08-06
Feat(anrok): add tax fetching and error processing on draft invoices

annvelents:
## Context

Added error processing on draft invoices, finalization of draft invoices with tax error details

## Description

Added:
- Fetch taxes from anrok/create_draft_invoice service if calculating fees and applying taxes for draft invoices
- Fetch taxes from anrok/create_invoice service if calculating fees and applying taxes for draft invoices that is being finalized
- From refresh_draft_and_finalize service if there was a tax error we now update invoice to failed state and do not send any notifications
- Return from finalize_service if it was a tax_error while fetching taxes
- Reset all invoice values when refreshing it - not only delete values, but also set 0 to computed values, so when we fail we have 0s instead of previous calculated values


app/services/invoices/calculate_fees_service.rb
```diff
@@ -47,14 +47,15 @@ def call
...
 
           unless taxes_result.success?
             create_error_detail(taxes_result.error.code)
 
-            # Draft invoice needs to stay in draft status
-            invoice.failed! if invoice.finalized?
+            # only fail invoices that are finalizing
+            invoice.failed! if finalizing_invoice?(invoice)
 
+            invoice.save
```
lovrocolic, line 58: This line is not needed since `invoice.failed!` saves invoice

app/services/invoices/refresh_draft_service.rb
```diff
@@ -95,5 +91,25 @@ def invoice_credit_note_items
...
+      invoice.applied_taxes.destroy_all
+      invoice.error_details.discard_all
+
+      invoice.taxes_amount_cents = 0
+      invoice.total_amount_cents = 0
+      invoice.taxes_rate = 0
+      invoice.fees_amount_cents = 0
+      invoice.sub_total_excluding_taxes_amount_cents = 0
+      invoice.sub_total_including_taxes_amount_cents = 0
+      invoice.save
```
lovrocolic, line 112: Do we really need last 7 lines? This is extra DB call but all those fields will be populated with new values anyway, once the refresh is done

app/services/invoices/refresh_draft_service.rb
```diff
@@ -60,6 +55,7 @@ def call
           recurring:,
           context:
         )
+        return calculate_result if tax_error?(calculate_result.error)
```
lovrocolic, line 58: shouldn't we return error result after iterating over credit notes?

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -181,6 +183,54 @@
...
+          context 'when no context is passed' do
+            let(:endpoint) { 'https://api.nango.dev/v1/anrok/draft_invoices' }
+
+            it 'creates fees' do
+              result = invoice_service.call
+
+              aggregate_failures do
+                expect(result).to be_success
+
+                expect(result.invoice.fees_amount_cents).to eq(100)
```
lovrocolic, line 208: Can you please either alter response from Anrok and set different numbers or add new expectation to check which endpoint is called?

spec/services/invoices/refresh_draft_and_finalize_service_spec.rb
```diff
@@ -30,7 +30,6 @@
       )
```
lovrocolic, line 30: Can you please also update `FinalizeJob` so that we don't generate dead job for each tax error

app/services/invoices/refresh_draft_and_finalize_service.rb
```diff
@@ -12,8 +12,12 @@ def call
       return result unless invoice.draft?
```
lovrocolic, line 12: Can you please also update `FinalizeJob` and `RefreshDraftJob` so that we don't generate dead job for each tax error

app/services/invoices/calculate_fees_service.rb
annvelents: it will be called only for finalizing_invoices, so in case we're refreshing a draft invoice, it won't be saved, and as result new fee sums will not be applied

app/services/invoices/refresh_draft_service.rb
annvelents: yes, because if we have an invoice where everything is calculated, including taxes, then we were refreshing an invoice and got an error - we won't be able to recalculate taxes, and without this nullifying it will keep old values

app/services/invoices/refresh_draft_service.rb
annvelents: Because you had concerns with credit notes, I thought it'll be better to return before 🙈 
can move it after since anyway we'll be implementing logic on credit notes 👍

app/services/invoices/calculate_fees_service.rb
lovrocolic: okay, just call invoice.save! please

app/services/invoices/refresh_draft_service.rb
lovrocolic: for now both works, but please add note if it will stay as it is

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: ah, right, we have allow in the before block, not expect 🤦‍♀️

app/services/invoices/refresh_draft_and_finalize_service.rb
annvelents: 🥇

app/services/invoices/refresh_draft_service.rb
lovrocolic: Hmm, this will work but not sure if this extra save is needed. Even if old values stay attached, we won't display it on UI I think (we display "-" for each attribute that can't be calculated. But ok, maybe can be zero if someone fetched it from the API

annvelents: by the way, what do you thing about having `nil`s instead of `0`s there?

annvelents: nah, sorry, forget about it - we have default 0s  🤦‍♀️

app/services/invoices/refresh_draft_service.rb
annvelents: updated

app/jobs/invoices/finalize_all_job.rb
```diff
@@ -6,7 +6,13 @@ class FinalizeAllJob < ApplicationJob
 
     def perform(organization:, invoice_ids:)
       result = Invoices::FinalizeBatchService.new(organization:).call(invoice_ids)
-      result.raise_if_error!
+      result.raise_if_error! unless tax_error?(result.error)
```
lovrocolic, line 9: Error returned from finalize service is validation error so you have to check it a bit different: `result.error.messages.dig(:tax_error)`. Also, we need it in `refresh_draft_job.rb` and `finalize_job.rb`

annvelents: fixed :)

app/services/invoices/apply_provider_taxes_service.rb
```diff
@@ -116,7 +116,11 @@ def fees_amount_cents(tax)
     end
 
     def fetch_provider_taxes_result
-      taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateService.call(invoice:)
+      taxes_result = if invoice.finalized?
```
lovrocolic, line 119: I think this does not work. In most cases we call `invoice.finalized!` after we call `Invoices::ComputeAmountsFromFees` (please take a look at `Invoices::CreatePayInAdvanceChargeService`). That being said, invoice will be in `generating` status and  `Integrations::Aggregator::Taxes::Invoices::CreateDraftService` will be called instead of `Integrations::Aggregator::Taxes::Invoices::CreateService`. IMO we should probably have one path just for draft invoices and other path for everything else

lovrocolic (approved):

PR #2380, opened 2024-08-07
feat(tresholds): Add create and update services

ivannovosad:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/progressive-billing

## Context

There's always a risk of customers not paying invoices generated at the end of a billing period. It would be beneficial to bill customers at given thresholds (units vs. amount) rather than waiting until the end of the period. This approach would allow for removing customer access if invoices are not paid and prevent having a highest amount of unpaid invoices.

## Description

This PR adds and modifies services that handle creating, updating and overriding plans/subscriptions with usage tresholds. 
Also the (treshold -> threshold) typo in models and serializers was fixed.


app/services/plans/create_service.rb
```diff
@@ -33,6 +33,14 @@ def create(args)
           taxes_result.raise_if_error!
         end
 
+        if args[:usage_thresholds].present? &&
+            License.premium? &&
```
nudded, line 37: How do you feel about something like: `License.supports_progressive_billing?(plan.organization)`

app/services/plans/create_service.rb
```diff
@@ -77,6 +85,17 @@ def create_commitment(plan, args, commitment_type)
       )
     end
 
+    def create_usage_threshold(plan, args)
+      usage_threshold = plan.usage_thresholds.new(
+        threshold_display_name: args[:threshold_display_name],
+        amount_cents: args[:amount_cents],
+        amount_currency: args[:amount_currency],
```
nudded, line 92: Do we allow specifying a different currency from the plan's currency?

ivannovosad: Yes we do, it was discussed [here](https://www.notion.so/Dive-in-Progressive-billing-tresholds-d9b176432ea8406ebc8a6af8103e358c?d=0acb194edf0b4a0b99f4072045c73ddd&pvs=4#47f09c60261445b9ad84c8d7fc5432bb).

app/services/plans/create_service.rb
ivannovosad: @nudded That's a good point which we already talked about a couple of time in the past.

We discussed using feature flags (e.g. [LaunchDarkly](https://launchdarkly.com/feature-flags-ruby/)) for this purpose with @jdenquin but since it was not decided yet, we can keep it like this.
That might affect implementation of `License` as well. 

Any opinions @jdenquin @lovrocolic @vincent-pochet

app/services/plans/create_service.rb
nudded: Okay 👍 I've moved the discussion to the dive-in, allowing thresholds with different currencies would complicate the calculation. We should discuss with product I think on how to proceed :)

ivannovosad: We'll always use plan's currency for tresholds.

app/services/plans/update_service.rb
```diff
@@ -74,6 +82,18 @@ def bill_charges_monthly?
       params[:bill_charges_monthly] || false
     end
 
+    def create_usage_threshold(plan, params)
+      usage_threshold = plan.usage_thresholds.new(
+        threshold_display_name: params[:threshold_display_name],
+        amount_cents: params[:amount_cents],
+        amount_currency: plan.amount_currency,
```
nudded, line 89: Another alternative is to do like we do in the `Charge` model: 

``` 
monetize :amount_cents, with_currency: ->(threshold) { threshold.plan.amount_currency }
```

But this works as well of course 👍

nudded (approved):

app/services/plans/update_service.rb
ivannovosad: I think this is even better in the end...

 - no need to store `amount_cents` in the usage_tresholds
 - risking potential conflict in having saved different currencies in `plans` and `usage_tresholds`
 
@nudded @vincent-pochet @lovrocolic @jdenquin

vincent-pochet (approved):

PR #2386, opened 2024-08-08
feat (tax-integrations): add new service for syncing void invoice with anrok

lovrocolic:
## Context

Currently integration with tax provider Anrok is being added

## Description

This service adds logic for sending new request to Nango. This action notifies Anrok that invoice has been voided in Lago.

### DIFF
--- a/app/services/integrations/aggregator/taxes/invoices/void_service.rb
+++ b/app/services/integrations/aggregator/taxes/invoices/void_service.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Integrations
+  module Aggregator
+    module Taxes
+      module Invoices
+        class VoidService < BaseService
+          def action_path
+            "v1/#{provider}/void_invoices"
+          end
+
+          def call
+            return result unless integration
+            return result unless integration.type == 'Integrations::AnrokIntegration'
+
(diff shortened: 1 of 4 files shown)


ancorcruz (approved):

app/services/integrations/aggregator/taxes/invoices/void_service.rb
```diff
@@ -0,0 +1,42 @@
...
+    module Taxes
+      module Invoices
+        class VoidService < BaseService
+          def action_path
+            "v1/#{provider}/void_invoices"
+          end
+
+          def call
+            return result unless integration
+            return result unless integration.type == 'Integrations::AnrokIntegration'
```
annvelents, line 14: I'm wondering, since we do this check in every service that is inherited from BaseService, can we move it to init in base?

annvelents (approved):

app/services/integrations/aggregator/taxes/invoices/void_service.rb
lovrocolic: 🤔 Probably raising an error in initailizer could work, even not my style preference, but returning result in initializer can affect how this object is called.
Probably if those checks list will grow, we can add one helper method in parent class:)

PR #2387, opened 2024-08-08
feature(ProgressiveBilling) - Add lifetime usage model

nudded:
## Context

## Description

Adds a LifetimeUsage model. sets up links between Subscription and LifetimeUsage.

### DIFF
--- a/app/models/lifetime_usage.rb
+++ b/app/models/lifetime_usage.rb
@@ -6,23 +6,14 @@ class LifetimeUsage < ApplicationRecord
   self.discard_column = :deleted_at
 
   belongs_to :organization
-
-  validates :currency, inclusion: {in: currency_list}
+  belongs_to :subscription
 
   validates :current_usage_amount_cents, numericality: {greater_than_or_equal_to: 0}
   validates :invoiced_usage_amount_cents, numericality: {greater_than_or_equal_to: 0}
 
   monetize :current_usage_amount_cents,
     :invoiced_usage_amount_cents,
-    with_model_currency: :currency
+    with_currency: ->(lifetime_usage) { lifetime_usage.subscription.plan.amount_currency }
 
   default_scope -> { kept }
-
-  def subscription
-    Subscription.active
-      .joins(:organization)
-      .where(organizations: {id: organization_id})
-      .where(external_id: external_subscription_id)
-      .sole
-  end
 end
--- a/app/models/lifetime_usage.rb
+++ b/app/models/lifetime_usage.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class LifetimeUsage < ApplicationRecord
+  include Currencies
+  include Discard::Model
+  self.discard_column = :deleted_at
+
+  belongs_to :organization
+
+  validates :currency, inclusion: {in: currency_list}
+
+  validates :current_usage_amount_cents, numericality: {greater_than_or_equal_to: 0}
+  validates :invoiced_usage_amount_cents, numericality: {greater_than_or_equal_to: 0}
+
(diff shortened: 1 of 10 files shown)


spec/models/lifetime_usage_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  it { is_expected.to belong_to(:organization) }
+  it { is_expected.to monetize(:current_usage_amount_cents) }
+  it { is_expected.to monetize(:invoiced_usage_amount_cents) }
+
+  describe 'default scope' do
+    let(:deleted_lifetime_usage) { create(:lifetime_usage, :deleted) }
+
+    it "only returns non-deleted lifetime-usage objects" do
+      expect(LifetimeUsage.all).to eq([lifetime_usage])
```
ancorcruz, line 16: be aware the `deleted_lifetime_usage` record hasn't been created, `let` is lazy and only executes the code when is called. Same for the next line.

you can add the following code to ensure the records are created to properly setup the test

```
before do
  lifetime_usage
  deleted_lifetime_usage
end
```

spec/models/lifetime_usage_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      expect do
+        LifetimeUsage.create(organization: lifetime_usage.organization,
+          external_subscripton_id: lifetime_usage.external_subscription_id)
+      end.to raise_error
+    end
+  end
+
+  describe "#subscription" do
+    it "returns the subscription" do
+      expected_subscription = Subscription.active.where(external_id: lifetime_usage.external_subscription_id).sole
```
ancorcruz, line 50: didn't know about `#sole`... thanks.

app/models/lifetime_usage.rb
```diff
@@ -0,0 +1,28 @@
...
+    with_model_currency: :currency
+
+  default_scope -> { kept }
+
+  def subscription
+    Subscription.active
+      .joins(:organization)
+      .where(organizations: {id: organization_id})
+      .where(external_id: external_subscription_id)
+      .sole
```
ancorcruz, line 26: do we want this method to raise an exception if there is no subscription or more than one matching subscription?

ancorcruz (changes requested):
left one comment and one question...

app/models/lifetime_usage.rb
vincent-pochet: Yes, it is expected. Since we will delete the lifetime usage along with the subscription, we are expecting a lifetime usage to always belongs to an existing active subscription

spec/models/lifetime_usage_spec.rb
nudded: Yeah don’t understand, I remember writing `let!` here. But will make sure it’s created in advance.

ancorcruz: maybe rubocop does not like `let!`

nudded: Appears it was just an oversight, fixed it.

spec/models/lifetime_usage_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe LifetimeUsage, type: :model do
+  subject(:lifetime_usage) { create(:lifetime_usage) }
+
+  it { is_expected.to belong_to(:organization) }
```
ancorcruz, line 8:
```suggestion
  it { is_expected.to belong_to(:organization) }
  it { is_expected.to belong_to(:subscription) }
```

ancorcruz (approved):
LGTM

PR #2394, opened 2024-08-09
feat(ProgressiveBilling): Flag invoiced lifetime usage for refresh

vincent-pochet:
## Context

## Description

This PR flags `LifetimeUsage` attached to subscription for refresh (by setting `recalculate_invoiced_usage` to `true`) when:
- Activating a subscription
- Finalizing an invoice
- Voiding an invoice


app/services/invoices/refresh_draft_and_finalize_service.rb
```diff
@@ -68,5 +70,9 @@ def should_deliver_email?
       License.premium? &&
         invoice.organization.email_settings.include?('invoice.finalized')
     end
+
+    def flag_lifetime_usage
+      LifetimeUsages::FlagRefreshFromInvoiceService.call(invoice:).raise_if_error!
```
vincent-pochet, line 75: Looking into it. I had to use plural form for module to avoid conflict with model class

nudded: Same, I think we can just live with that.

app/services/lifetime_usages/flag_refresh_from_invoice_service.rb
```diff
@@ -0,0 +1,39 @@
...
+      @invoice = invoice
+      super
+    end
+
+    def call
+      return result unless invoice.subscription?
+      return result unless invoice.finalized? || invoice.voided?
+      return result unless has_plan_usage_thresholds?
+
+      result.lifecycle_usages = []
```
nudded, line 15:
```suggestion
      result.lifetime_usages = []
```

app/services/lifetime_usages/flag_refresh_from_invoice_service.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      invoice.subscriptions.each do |subscription|
+        lifetime_usage = subscription.lifetime_usage
+        lifetime_usage ||= subscription.build_lifetime_usage(organization: subscription.organization)
+        lifetime_usage.recalculate_invoiced_usage = true
+        lifetime_usage.save!
+
+        result.lifecycle_usages << lifetime_usage
```
nudded, line 23:
```suggestion
        result.lifetime_usages << lifetime_usage
```

app/services/lifetime_usages/flag_refresh_from_invoice_service.rb
```diff
@@ -0,0 +1,39 @@
...
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    end
+
+    private
+
+    attr_reader :invoice
+
+    def has_plan_usage_thresholds?
+      invoice.subscriptions.any? { |s| s.plan.usage_thresholds.any? }
```
nudded, line 36: not sure if it really matters, but doing it like this might also lead to unnecessary work performed for all subscriptions if one of them has thresholds. But this is likely premature optimization, so also happy to keep as is.

nudded (approved):

PR #2395, opened 2024-08-09
feat(ProgressiveBilling): Flag usage refresh for event

vincent-pochet:
## Context

## Description

This PR flags `LifetimeUsage` attached to subscription for refresh (by setting `recalculate_current_usage` to `true`) when post processing an event


app/services/lifetime_usages/flag_refresh_from_subscription_service.rb
```diff
@@ -0,0 +1,29 @@
...
+    def initialize(subscription:)
+      @subscription = subscription
+      super
+    end
+
+    def call
+      return result unless subscription.active?
+
+      lifetime_usage = subscription.lifetime_usage
+      lifetime_usage ||= subscription.build_lifetime_usage(organization: subscription.organization)
```
nudded, line 14: I'm wondering, this will mean we'll track the usage for all active subscriptions. Should we try and limit it only to those that have thresholds defined?

vincent-pochet: @nudded To avoid overloading the workers and DB, maybe we should indeed check for the presence of thresholds.
Updating it!

app/services/lifetime_usages/flag_refresh_from_subscription_service.rb
```diff
@@ -0,0 +1,30 @@
...
+    def call
+      return result unless subscription.active?
+      return result unless subscription.plan.usage_thresholds.any?
+
+      lifetime_usage = subscription.lifetime_usage
+      lifetime_usage ||= subscription.build_lifetime_usage(organization: subscription.organization)
+      lifetime_usage.recalculate_current_usage = true
+      lifetime_usage.save!
+
+      result.lifecycle_usage = lifetime_usage
```
nudded, line 19:
```suggestion
      result.lifetime_usage = lifetime_usage
```

nudded (approved):

PR #2402, opened 2024-08-12
perf(active_storage) - add index that went missing when the record_id column changed type

nudded:
the `record_id` column was removed and re-added as a UUID (https://github.com/getlago/lago-api/commit/c36ce47a7ddf7c53eecfb9211c38c2bbe0ab6fe8) . This inadvertently removed the index that ActiveStorage adds when setting up the table. 

This MR adds that index again.


vincent-pochet (approved):
Good catch!

PR #2409, opened 2024-08-13
fix(bulk_events) - Add #blank? check to `Events::CreateBatchService`

nudded:
## Description

Adds an extra sanity check before any logic is called in the `Events::CreateBatchService`


vincent-pochet (approved):
:rocket:

PR #2410, opened 2024-08-13
perf - fix N+1 queries on #index actions

nudded:
## Description

Many #index actions include extra relations when serializing. If we don't include the associations we'll have N+1 queries in overview pages making them unnecessarily slow.


vincent-pochet (approved):
Great!

jdenquin (approved):

PR #2413, opened 2024-08-13
feat(ProgressiveBilling): create an invoice for a threshold

vincent-pochet:
## Context

## Description

This PR adds the services to create an invoice with a fee when a specific `usage_threshold` is reach by a subscription.

It creates:
- A `progressive_billing` invoice
- An `invoice_subscription` with the right period boundaries
- A `progressive_billing` fee

It also handles:
- PDF generation
- Tax computation
- Email delivery
- Tracking on Segment
- Payment
- Sync with integrations

### DIFF
--- a/app/services/fees/create_from_usage_threshold_service.rb
+++ b/app/services/fees/create_from_usage_threshold_service.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+module Fees
+  class CreateFromUsageThresholdService < BaseService
+    def initialize(usage_threshold:, invoice:, amount_cents:)
+      @usage_threshold = usage_threshold
+      @invoice = invoice
+      @amount_cents = amount_cents
+
+      super
+    end
+
+    def call
+      fee = Fee.new(
+        subscription: invoice.subscriptions.first,
+        invoice:,
+        usage_threshold:,
+        invoice_display_name: usage_threshold.threshold_display_name,
+        invoiceable: usage_threshold,
+        amount_cents: amount_cents,
+        amount_currency: invoice.currency,
+        fee_type: :progressive_billing,
+        units:,
+        unit_amount_cents: unit_amount_cents,
+        payment_status: :pending,
+        taxes_amount_cents: 0,
+        properties: {
+          charges_from_datetime: invoice.invoice_subscriptions.first.charges_from_datetime,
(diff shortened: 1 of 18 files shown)


app/services/fees/create_from_usage_threshold_service.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def call
+      fee = Fee.new(
+        subscription: invoice.subscriptions.first,
+        invoice:,
+        usage_threshold:,
+        invoice_display_name: usage_threshold.threshold_display_name,
+        invoiceable: usage_threshold,
+        amount_cents: usage_threshold.amount_cents,
```
nudded, line 19: Is this the right amount? 🤔 

Should this not take into account the already invoiced_usage of the lifetime_usage?

app/services/fees/create_from_usage_threshold_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    private
+
+    attr_reader :usage_threshold, :invoice, :amount_cents
+
+    def units
+      return 1 unless usage_threshold.recurring?
+
+      amount_cents.fdiv(usage_threshold.amount_cents)
```
nudded, line 50: more a question than a remark, but this means we can have for example 0.8 units? (I think that's fine, just wondering if this will be confusing :smile)

vincent-pochet: Theoretically yes, but actually the values:
```ruby
units = (total_lifetime_usage_amount_cents - invoiced_threshold_amount_cents) / usage_threshold.amount_cents
units * usage_threshold.amount_cents
``` 
so it should always be an integer

app/services/invoices/progressive_billing_service.rb
```diff
@@ -0,0 +1,104 @@
...
+          .call(invoice:, subscriptions: [subscription], timestamp:, invoicing_reason: :progressive_billing)
+          .raise_if_error!
+      end
+      invoice_result.raise_if_error!
+
+      @invoice = invoice_result.invoice
+    end
+
+    def create_threshold_fees
+      usage_thresholds.each do |usage_threshold|
```
nudded, line 61: nit: would it be more clear to the reader to explicitly sort them again? (somewhere in this service, not specifically here)

nudded (approved):

PR #2422, opened 2024-08-14
feat (tax-integrations): add mutation for retrying void operation in tax provider

lovrocolic:
## Context

Currently Lago is adding new integration with tax provider Anrok

## Description

This PR adds possibility to re-sync voided invoice in tax provider (if previous sync failed)


app/graphql/mutations/invoices/retry_tax_provider_voiding.rb
```diff
@@ -0,0 +1,25 @@
...
+      REQUIRED_PERMISSION = 'invoices:update'
+
+      description 'Retry voided invoice sync'
+
+      argument :id, ID, required: true
+
+      type Types::Invoices::Object
+
+      def resolve(**args)
+        invoice = current_organization.invoices.visible.find_by(id: args[:id])
```
rsempe, line 18: should we use `visible` here?
voiding a `draft` or already `voided` invoice does not make sense, right?

ancorcruz: `visible` also includes `voided`, `draft` and `generating` are the only ones excluded from `visible`

ancorcruz (approved):

PR #2431, opened 2024-08-19
feat: add email validation

brunomiguelpinto:
## Context

This pull request changes Customers::CreateService adding email validation to ensure only valid email addresses are accepted when creating or updating customer records.

This is particularly important because we sync these emails with Stripe, which enforces strict email validation. Without this check, invalid emails could cause the Stripe synchronization process to fail

### DIFF
--- a/app/services/customers/create_service.rb
+++ b/app/services/customers/create_service.rb
@@ -9,6 +9,13 @@ module Customers
       new_customer = customer.new_record?
       shipping_address = params[:shipping_address] ||= {}
 
+      unless valid_email?(params[:email])
+        return result.single_validation_failure!(
+          field: :email,
+          error_code: 'invalid_format'
+        )
+      end
+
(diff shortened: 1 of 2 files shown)


app/services/customers/create_service.rb
```diff
@@ -187,6 +194,13 @@ def create(**args)
 
     private
 
+    def valid_email?(email)
+      return true if email.nil?
+
+      email_regex = /\A[\w+\-.]+@[a-z\d\-.]+\.[a-z]+\z/i
```
julienbourdeau, line 200: Where the regex from?
I think we should copy the regex from Devise: https://github.com/heartcombo/devise/blob/72884642f5700439cc96ac560ee19a44af5a2d45/lib/devise.rb#L114-L118

brunomiguelpinto: changed

vincent-pochet (approved):

PR #2432, opened 2024-08-19
feat(ProgressiveBilling): Update PDF template

vincent-pochet:
## Context

## Description

This PR allows generating the PDF for progressive billing invoices

### DIFF
--- a/app/services/invoices/progressive_billing_service.rb
+++ b/app/services/invoices/progressive_billing_service.rb
@@ -99,7 +99,7 @@ module Invoices
       @invoiced_amount_cents ||= subscription.invoices
         .finalized
         .where(invoice_type: %w[subscription progressive_billing])
-        .sum(:fees_amount_cents)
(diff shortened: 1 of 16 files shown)


app/services/invoices/progressive_billing_service.rb
```diff
@@ -99,7 +99,7 @@ def invoiced_amount_cents
       @invoiced_amount_cents ||= subscription.invoices
         .finalized
         .where(invoice_type: %w[subscription progressive_billing])
-        .sum(:fees_amount_cents)
+        .sum { |invoice| invoice.fees.charge.sum(:amount_cents) }
```
nudded, line 102: `invoice.fees.charge` Is it correct to only look at `charge` fees? (e.g. we create dedicated `progressive_billing` fees for the progressive invoices, I thought)

vincent-pochet: Good catch, I will update it!

nudded (approved):

PR #2435, opened 2024-08-19
feat(ProgressiveBilling) - add Clock::RefreshLifetimeUsagesJob + LifetimeUsages::RecalculateAndCheckService

nudded:
## Context

## Description

Adds: 
- Clock::RefreshLifetimeUsagesJob
  * fetches all usages to recalculate and enqueues the RefreshJob
- LifetimeUsages::RecalculateAndCheckJob
  * calls the RecalculateAndCheckService for this lifetime_usage, guarantees only 1 running at a time.
- LifetimeUsages::RecalculateAndCheckService
  * recalculates the usage
  * checks if there are thresholds that have been passed
  * send webhooks if needed
  * creates an invoice if needed


app/jobs/lifetime_usages/recalculate_and_check_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module LifetimeUsages
+  class RecalculateAndCheckJob < ApplicationJob
+    unique :until_executed, on_conflict: :log
```
vincent-pochet, line 5: Should we use the `default` queue. I was thinking at using the `billing` one, so that is does not take precedence over `mailers` and `provider`, what do you think?

vincent-pochet (approved):

app/jobs/lifetime_usages/recalculate_and_check_job.rb
nudded: I had not considered that, I'll move it to the `billing` queue 👍

PR #2441, opened 2024-08-20
feat(dunning): Create payment request

rsempe:
## Context

## Description

The goal of this PR is to:

- create the payable group for the overdue invoices
- create the payment request for the payable group
- send the payment_request.created webhook

on the payment request creation service

### DIFF
--- a/app/services/payment_requests/create_service.rb
+++ b/app/services/payment_requests/create_service.rb
@@ -17,14 +17,21 @@ module PaymentRequests
         payable_group = customer.payable_groups.create!(organization:)
         invoices.each { |i| i.update!(payable_group:) }
 
-        # TODO: Create payment request for the payable group
+        # NOTE: Create payment request for the payable group
+        payment_request = payable_group.payment_requests.create!(
+          organization:,
+          customer:,
+          amount_cents: invoices.sum(:total_amount_cents),
+          amount_currency: invoices.first.currency,
+          email:
+        )
 
         # TODO: Send payment_request.created webhook
 
         # TODO: When payment provider is set: Create payment intent for the overdue invoices
         # TODO: When payment provider is not set: Send email to the customer
 
-        # result.payment_request = payment_request
+        result.payment_request = payment_request
       end
 
       result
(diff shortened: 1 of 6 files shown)


app/services/payment_requests/create_service.rb
```diff
@@ -10,19 +10,34 @@ def initialize(organization:, params:)
...
+      return result.not_found_failure!(resource: "invoice") if invoices.empty?
 
-        # TODO: Create payment request for the payable group
+      # TODO: Return error if invoices are not overdue
 
-        # TODO: Send payment_request.created webhook
+      ActiveRecord::Base.transaction do
+        # NOTE: Create payable group for the overdue invoices
+        payable_group = customer.payable_groups.create!(organization:)
+        invoices.each { |i| i.update!(payable_group:) }
```
ancorcruz, line 23: would you expect any side effect or validation to happen here. if not we could use `update_all` to perform a unique SQL statement in order to update all the invoices at once.

what do you think?

app/services/payment_requests/create_service.rb
```diff
@@ -10,19 +10,34 @@ def initialize(organization:, params:)
...
+        payment_request = payable_group.payment_requests.create!(
+          organization:,
+          customer:,
+          amount_cents: invoices.sum(:total_amount_cents),
+          amount_currency: invoices.first.currency,
+          email:
+        )
+
+        # NOTE: Send payment_request.created webhook
+        SendWebhookJob.perform_later("payment_request.created", payment_request)
```
ancorcruz, line 35: we are not testing the `SendWebhookJob` is queued.

ancorcruz (approved):
I have left 2 comments but it looks good to me!

app/services/payment_requests/create_service.rb
rsempe: Good spot 👍

app/services/payment_requests/create_service.rb
rsempe: Yeah - I was hesitant but agreed let's update all the invoices at once 👍

PR #2445, opened 2024-08-21
feat(dunning): Prevent creating payment request when not overdue

rsempe:
## Context

## Description

The goal of this PR is to prevent the creation of a payment request:

- when not premium
- when not having `dunning` add-on feature
- when invoices are not overdue

### DIFF
--- a/app/services/payment_requests/create_service.rb
+++ b/app/services/payment_requests/create_service.rb
@@ -10,7 +10,9 @@ module PaymentRequests
     end
 
     def call
-      # TODO: Return error unless premium license
+      unless License.premium? && organization.premium_integrations.include?("dunning")
+        return result.not_allowed_failure!(code: "premium_addon_feature_missing")
+      end
 
       return result.not_found_failure!(resource: "customer") unless customer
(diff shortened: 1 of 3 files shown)


app/services/payment_requests/create_service.rb
```diff
@@ -10,12 +10,16 @@ def initialize(organization:, params:)
...
-      # TODO: Return error unless premium license
+      unless License.premium? && organization.premium_integrations.include?("dunning")
+        return result.not_allowed_failure!(code: "premium_addon_feature_missing")
+      end
 
       return result.not_found_failure!(resource: "customer") unless customer
       return result.not_found_failure!(resource: "invoice") if invoices.empty?
 
-      # TODO: Return error if invoices are not overdue
+      if invoices.where(payment_overdue: false).exists?
```
ancorcruz, line 20:
```suggestion
      if invoices.not.payment_overdue.exists?
```
Would you use the existing scope?

Or, if you prefer not to, exists? accepts conditions like

```
      if invoices.exists?(payment_overdue: false)
```

ancorcruz (approved):

PR #2451, opened 2024-08-21
misc(graphql): Update dependencies

jdenquin:
- Due to some CVE and old dependencies, we need to update our `graphql` gem
- No breaking changes
- The only adaptation was around our custom `permissions`


db/schema.rb
```diff
@@ -757,6 +757,7 @@
     t.boolean "payment_overdue", default: false
     t.uuid "payable_group_id"
     t.bigint "negative_amount_cents", default: 0, null: false
+    t.bigint "progressive_billing_credit_amount_cents", default: 0, null: false
```
jdenquin, line 760: I don't know why I have this change on a fresh DB, maybe @nudded can help?

nudded: That should not be possible 🤔 i will check

rsempe (approved):

PR #2459, opened 2024-08-22
feat(dunning): Prevent creation of payment request

rsempe:
## Context

## Description

The goal of this PR is to return an error when trying to create a payment request with:
- invoices having different currencies
- invoices not ready for payment processing

### DIFF
--- a/app/services/payment_requests/create_service.rb
+++ b/app/services/payment_requests/create_service.rb
@@ -10,23 +10,15 @@ module PaymentRequests
     end
 
     def call
-      unless License.premium? && organization.premium_integrations.include?("dunning")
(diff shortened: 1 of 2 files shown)


app/services/payment_requests/create_service.rb
```diff
@@ -47,6 +39,34 @@ def call
 
     attr_reader :organization, :params
 
+    def check_preconditions
+      # NOTE: Prevent creation of payment request if:
+      # - the organization does not have the premium dunning integration
+      # - the customer does not exist
```
ancorcruz, line 45:
```suggestion
      # - the customer does not exist
      # - there are no invoices
```

ancorcruz (approved):
LGTM,

there is a precondition missed in the comment...

PR #2461, opened 2024-08-22
feat(ProgressiveBilling): Bill current usage

vincent-pochet:
## Context

## Description

This PR refactors the `Invoices::ProgressiveBillingService` to create current usage fees, instead of a single progressive billings one per threshold.

Note: The logic to apply progressive billing credits to the invoice in order to avoid double billing will be handled in a separate pull request as the current one is already complex. The PDF template will also be adapted separatly


app/services/lifetime_usages/recalculate_and_check_service.rb
```diff
@@ -31,7 +31,7 @@ def progressive_billed_amount
       subscription.invoices
         .finalized
         .progressive_billing
-        .sum { |invoice| invoice.fees.progressive_billing.sum(:amount_cents) }
```
nudded, line 34: should we remove the `progressive_billing` fee type if we're no longer using it?

& 

can we change this to just be `fees_amount_cents`?

nudded (approved):

PR #2462, opened 2024-08-22
feat(invoices): Ability to skip 0 invoices

brunomiguelpinto:
## Context

We currently finalize all invoices, including those with a total amount of $0, leading to unnecessary clutter and potential confusion. To improve flexibility and reduce unnecessary finalizations, organizations will now have the option to decide whether $0 invoices should be finalized or not. This setting can be configured at the organization level, with the option to override it for individual customers.

## Description

This PR introduces a new service, `TransitionToFinalStatusService`, which checks the `finalize_zero_amount_invoice` setting to decide whether to `finalize` an invoice or move it to the `closed` state after checking the customer and organization setting ( if the invoice `fees_amount_cents` is 0 ) 

#### Organization Setting

Added a new parameter `finalize_zero_amount_invoice` to the organization settings:

```json
{
  "organization": {
    "finalize_zero_amount_invoice": "false" // default is true
  }
}
```

#### Customer Setting

Added a new parameter `finalize_zero_amount_invoice` to the customer settings to allow overriding the organization-level setting:

```json
{
  "customer": {
    "finalize_zero_amount_invoice": "finalize" // default value is "inherit"
  }
}
```

**Possible Values:**
- **`inherit`**: (Default) The setting follows the organization-level configuration.
- **`finalize`**: Invoices are issued and finalized, even if the total amount is $0.
- **`skip`**: Invoices with a total amount of $0 are not finalized.

#### `TransitionToFinalStatusService`

The service checks the `finalize_zero_amount_invoice` setting to determine whether to finalize an invoice or move it to the `closed` state after evaluating the customer, organization, and invoice amount.
  
#### Invoice State: `closed`

- Introduces a `closed` state for invoices with a total amount of $0.
- Invoices in this state:
  - Will not be assigned a number.
  - Will not appear in the application.
  - Will only exist in the database.

#### Service Integration

The new `TransitionToFinalStatusService` was integrated into various points where invoices are finalized:
- `Invoices::AdvanceChargesService`
- `Invoices::CreateOneOffService`
- `Invoices::CreatePayInAdvanceChargeService`
- `Invoices::RefreshDraftAndFinalizeService`
- `Invoices::SubscriptionService`

### DIFF
--- a/app/services/invoices/check_transition_to_finalized_service.rb
+++ b/app/services/invoices/check_transition_to_finalized_service.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Invoices
+  class CheckTransitionToFinalizedService < BaseService
+    def initialize(invoice:)
+      @invoice = invoice
+      @customer = @invoice.customer
+      @organization = @customer.organization
+      super
+    end
+
+    def call
+      if should_finalize_invoice?
+        invoice.update(status: :finalized)
+      else
+        invoice.update(status: :closed)
+      end
+      result.invoice = invoice
+      result
+    end
+
+    private
+
+    attr_reader :invoice, :customer, :organization
+
+    def should_finalize_invoice?
+      return true unless invoice.fees_amount_cents.zero?
+      customer_setting = customer.finalize_zero_amount_invoice
+      if customer_setting == 'inherit'
+        organization.finalize_zero_amount_invoice
+      else
+        customer_setting == 'finalize'
+      end
+    end
+  end
+end
--- a/app/services/invoices/check_transition_to_finalized_service.rb
+++ b/app/services/invoices/check_transition_to_finalized_service.rb
@@ -10,10 +10,10 @@ module Invoices
     end
 
     def call
-      if should_finalize_invoice?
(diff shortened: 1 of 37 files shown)


app/services/invoices/check_transition_to_finalized_service.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Invoices
+  class CheckTransitionToFinalizedService < BaseService
```
brunomiguelpinto, line 4: i think CheckInvoiceFinalStatusService explains better what this does what do you think?

rsempe (approved):

app/services/invoices/check_transition_to_finalized_service.rb
annvelents: Hey!
sorry, after our discussion I still have some doubts  if instead of check it should be `Find` or `Assign`, or `Set`, or `Transition` (`TransitionInvoiceToFinalStatus`, wdyt?)... because `Check` sounds as if it should return true or false...

annvelents (approved):

PR #2463, opened 2024-08-22
feat(ProgressiveBilling): Apply credit to progressive billing invoices

vincent-pochet:
## Context

## Description

This PR applies credits from the progressive_billing invoices to every new progressive billing invoice in the same period. It also makes sure that credits are well applied to fees and taken into account when computing taxes

### DIFF
--- a/app/services/invoices/progressive_billing_service.rb
+++ b/app/services/invoices/progressive_billing_service.rb
@@ -17,9 +17,11 @@ module Invoices
 
         invoice.fees_amount_cents = invoice.fees.sum(:amount_cents)
         invoice.sub_total_excluding_taxes_amount_cents = invoice.fees_amount_cents
-        Credits::AppliedCouponsService.call(invoice:) if invoice.fees_amount_cents&.positive?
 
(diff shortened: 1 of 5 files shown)


app/services/invoices/progressive_billing_service.rb
```diff
@@ -17,9 +17,14 @@ def call
 
         invoice.fees_amount_cents = invoice.fees.sum(:amount_cents)
         invoice.sub_total_excluding_taxes_amount_cents = invoice.fees_amount_cents
-        Credits::AppliedCouponsService.call(invoice:) if invoice.fees_amount_cents&.positive?
+
+        if invoice.fees_amount_cents.positive?
+          Credits::ProgressiveBillingService.call(invoice:)
```
nudded, line 22: 🤔 when/why would we not end up crediting the previously invoiced usage?

vincent-pochet: Good catch, it should always g through the Credits services :+1:

nudded (approved):

PR #2468, opened 2024-08-23
feat (tax-integrations): handle current usage with tax provider

lovrocolic:
## Context

Currently Lago is implementing integration with tax provider Anrok

## Description

This PR adds logic to use tax provider taxes when fetching current usage


app/services/invoices/customer_usage_service.rb
```diff
@@ -136,6 +141,28 @@ def compute_amounts
       invoice.total_amount_cents = invoice.fees_amount_cents + invoice.taxes_amount_cents
     end
 
+    def compute_amounts_with_provider_taxes
+      invoice.fees_amount_cents = invoice.fees.sum(&:amount_cents)
+
+      taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateDraftService.call(invoice:, fees: invoice.fees)
```
lovrocolic, line 147: @vincent-pochet @rsempe Ongoing balance job is calling current usage each 5 minutes. Do you think that we should consider caching Anrok result for current usage and set expiration after e.g. 12 hours? Taxes can vary if something changes in the setup but I am also not sure if it is good idea to ping Anrok that often from ongoing balance job... 🤔

vincent-pochet: @lovrocolic Indeed I think it would be great to cache it if possible

lovrocolic: Added caching that expires after 1 day

ivannovosad (approved):

PR #2471, opened 2024-08-23
feat(dunning): Create stripe payment

ancorcruz:
## Context

## Description

The goal of this PR is to create a stripe payment for customers this payment provider.

### DIFF
--- a/app/services/payment_requests/payments/stripe_service.rb
+++ b/app/services/payment_requests/payments/stripe_service.rb
@@ -1,0 +1,214 @@
+# frozen_string_literal: true
+
+module PaymentRequests
+  module Payments
+    class StripeService < BaseService
+      include Customers::PaymentProviderFinder
+
+      PENDING_STATUSES = %w[processing requires_capture requires_action requires_confirmation requires_payment_method]
+        .freeze
+      SUCCESS_STATUSES = %w[succeeded].freeze
+      FAILED_STATUSES = %w[canceled].freeze
+
+      def initialize(payable = nil)
+        @payable = payable
+
+        super(nil)
+      end
+
+      def create
+        result.payable = payable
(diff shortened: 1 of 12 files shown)


app/services/payment_requests/payments/stripe_service.rb
```diff
@@ -0,0 +1,224 @@
...
+
+      PENDING_STATUSES = %w[processing requires_capture requires_action requires_confirmation requires_payment_method]
+        .freeze
+      SUCCESS_STATUSES = %w[succeeded].freeze
+      FAILED_STATUSES = %w[canceled].freeze
+
+      def initialize(payable = nil)
+        @payable = payable
+
+        super(nil)
```
rsempe, line 16:
```suggestion
        super
```

app/services/payment_requests/payments/stripe_service.rb
```diff
@@ -0,0 +1,224 @@
...
+    class StripeService < BaseService
+      include Customers::PaymentProviderFinder
+
+      PENDING_STATUSES = %w[processing requires_capture requires_action requires_confirmation requires_payment_method]
+        .freeze
+      SUCCESS_STATUSES = %w[succeeded].freeze
+      FAILED_STATUSES = %w[canceled].freeze
+
+      def initialize(payable = nil)
+        @payable = payable
```
rsempe, line 14: 😍

rsempe (approved):

app/services/payment_requests/payments/stripe_service.rb
ancorcruz: that would call initialize in BaseService that expects current_user as argument and will pass the payable object instead. This requires `super(nil)`.

PR #2473, opened 2024-08-23
Feat(anrok): add fields to graphql response on applied taxes

annvelents:
## Context

On FE to be able to show differently taxes applied on all line items vs on few only within the invoice, we need to specify on the applied_tax if it's applicable on the whole invoice
and to be able to show different translation for the taxes depending on their code, we want to send tax_cod with applied tax as well

## Description

- added tax_code
- and applied_on_whole_invoice

to Invoice::AppliedTax graphql type

### DIFF
--- a/app/graphql/types/invoices/applied_taxes/object.rb
+++ b/app/graphql/types/invoices/applied_taxes/object.rb
@@ -7,9 +7,10 @@ module Types
         graphql_name 'InvoiceAppliedTax'
         implements Types::Taxes::AppliedTax
 
-        field :invoice, Types::Invoices::Object, null: false
-
+        field :applied_on_whole_invoice?, GraphQL::Types::Boolean, null: false
         field :fees_amount_cents, GraphQL::Types::BigInt, null: false
+        field :invoice, Types::Invoices::Object, null: false
+        field :tax_code, String, null: false
       end
     end
   end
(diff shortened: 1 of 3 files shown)


schema.graphql
```diff
@@ -4068,6 +4068,7 @@ type Invoice {
 type InvoiceAppliedTax implements AppliedTax {
   amountCents: BigInt!
   amountCurrency: CurrencyEnum!
+  appliedOnWholeInvoice?: Boolean!
```
rsempe, line 4071: is the schema up-to-date? I can't see `taxCodes` for instance 🤔

annvelents: hm
that's interesting as I added both fields and run the command to regenerate the schema...
will fix it shortly

annvelents: thanks for noticing!

annvelents: found it, thanks!
the InvoiceAppliedTax implements AppliedTax, which has taxCode, so we don't need to add it on the InvoiceAppliedTax level, cleaned 😁

app/graphql/types/invoices/applied_taxes/object.rb
```diff
@@ -7,9 +7,9 @@ class Object < Types::BaseObject
         graphql_name 'InvoiceAppliedTax'
         implements Types::Taxes::AppliedTax
 
-        field :invoice, Types::Invoices::Object, null: false
-
+        field :applied_on_whole_invoice?, GraphQL::Types::Boolean, null: false
```
lovrocolic, line 10: This is just definition for TS method on FE (it turns into `applidOnWholeInvoice`) so I would return questionmark at the end of the method since this is mostly rails convention and not TS or JS convention

rsempe (approved):

PR #2474, opened 2024-08-23
feat(dunning): Create Adyen payment

ancorcruz:
## Context

## Description

The goal of this PR is to create a Adyen payment for customers this payment provider.

### DIFF
--- a/app/models/payment_request.rb
+++ b/app/models/payment_request.rb
@@ -24,6 +24,11 @@ class PaymentRequest < ApplicationRecord
   def invoice_ids
     applied_invoices.pluck(:invoice_id)
   end
+
+  def increment_payment_attempts!
+    increment(:payment_attempts)
+    save!
+  end
 end
 
 # == Schema Information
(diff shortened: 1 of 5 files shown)


app/models/payment_request.rb
```diff
@@ -24,6 +24,11 @@ class PaymentRequest < ApplicationRecord
   def invoice_ids
     applied_invoices.pluck(:invoice_id)
   end
+
+  def increment_payment_attempts!
+    increment(:payment_attempts)
+    save!
```
rsempe, line 30: why not directly?
```suggestion
    increment!(:payment_attempts)
```

app/services/payment_requests/payments/adyen_service.rb
```diff
@@ -0,0 +1,171 @@
...
+          amount_currency: payable.currency.upcase,
+          provider_payment_id: res.response['pspReference'],
+          status: res.response['resultCode']
+        )
+        payment.save!
+
+        payable_payment_status = payable_payment_status(payment.status)
+        update_payable_payment_status(payment_status: payable_payment_status)
+        update_invoices_payment_status(payment_status: payable_payment_status)
+
```
rsempe, line 50: what about doing that in a transaction?

rsempe (approved):

app/models/payment_request.rb
ancorcruz: rubocop does not allow `increment!` because it skips ActiveRecord validations, callbacks, etc.

PR #2475, opened 2024-08-23
feat(ProgressiveBilling) - automated CreditNote generation at the periodic invoice if needed

nudded:
## Context

## Description

Whenever we credit already invoiced progressive billing invoices at the periodic invoice, we could end up in a situation that leaves more to credit than invoicable. At that point we'll create credit notes for past progressive billing invoices.

This adds a `CreditNotes::CreateFromProgressiveBillingInvoice` service that will take care of distributing the amount to credit over all fees.

Also updates `debug` gem and allows for remote debugging when `ENV['RUBY_DEBUG_PORT']` is set


app/services/credit_notes/create_from_progressive_billing_invoice.rb
```diff
@@ -0,0 +1,61 @@
...
+    def call
+      return result unless amount.positive?
+      return result.forbidden_failure! unless progressive_billing_invoice.progressive_billing?
+
+      # Important to call this method as it modifies @amount if needed
+      items = calculate_items!
+
+      CreditNotes::CreateService.new(
+        invoice: progressive_billing_invoice,
+        credit_amount_cents: amount,
```
vincent-pochet, line 22: In this case, the amount must take taxes into account.
You can check for `CreditNotes::CreateFromTermination#creditable_amount_cents` that applies a similar logic

PR #2478, opened 2024-08-26
feat (tax-integrations): Use Anrok taxes for one off invoices

lovrocolic:
## Context

Currently integration with tax provider Anrok is being implemented.

## Description

This PR uses all existing Anrok logic on one-off invoices. 

If customer is connected to Anrok, its taxes should be used. All services and Anrok core logic is already implemented and merged on main. This is just logic that consumes it for one-off invoices

### DIFF
--- a/app/graphql/mutations/integrations/anrok/fetch_draft_invoice_taxes.rb
+++ b/app/graphql/mutations/integrations/anrok/fetch_draft_invoice_taxes.rb
@@ -26,7 +26,7 @@ module Mutations
             fees: fees(args)
           ).call
 
-          result.success? ? result.fees : validation_error(messages: {tax_error: result.error.code})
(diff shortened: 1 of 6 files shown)


app/graphql/mutations/integrations/anrok/fetch_draft_invoice_taxes.rb
```diff
@@ -26,7 +26,7 @@ def resolve(**args)
             fees: fees(args)
           ).call
 
-          result.success? ? result.fees : result_error(result)
+          result.success? ? result.fees : validation_error(messages: {tax_error: [result.error.code]})
```
ivannovosad, line 29: Do we have a mutation spec for this change?

ivannovosad (approved):

PR #2486, opened 2024-08-27
feat (tax-integrations): Use Anrok taxes in subscription invoices flow

lovrocolic:
## Context

Currently integration with tax provider Anrok is being implemented.

## Description

This PR uses all existing Anrok logic on subscription and draft invoices. 

If customer is connected to Anrok, its taxes should be used. All services and Anrok core logic is already implemented and merged on main. This is just logic that consumes it for subscription and draft invoices

### DIFF
--- a/app/jobs/bill_subscription_job.rb
+++ b/app/jobs/bill_subscription_job.rb
@@ -14,6 +14,9 @@ class BillSubscriptionJob < ApplicationJob
       skip_charges:
     )
     return if result.success?
+    # NOTE: We don't want a dead job for failed invoice due to the tax reason.
+    #       This invoice should be in failed status and can be retried.
+    return if result.error.messages.dig(:tax_error)
 
     # If the invoice was passed as an argument, it means the job was already retried (see end of function)
     result.raise_if_error! if invoice
--- a/app/services/fees/commitments/minimum/create_service.rb
+++ b/app/services/fees/commitments/minimum/create_service.rb
@@ -35,9 +35,6 @@ module Fees
             taxes_amount_cents: 0
           )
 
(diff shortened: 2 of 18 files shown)


app/services/fees/commitments/minimum/create_service.rb
```diff
@@ -35,9 +35,6 @@ def call
             taxes_amount_cents: 0
           )
 
-          taxes_result = Fees::ApplyTaxesService.call(fee: new_fee)
```
ivannovosad, line 38: Are we not applying taxes on commitments anymore?

app/jobs/bill_subscription_job.rb
```diff
@@ -14,6 +14,9 @@ def perform(subscriptions, timestamp, invoicing_reason:, invoice: nil, skip_char
       skip_charges:
     )
     return if result.success?
+    # NOTE: We don't want a dead job for failed invoice due to the tax reason.
+    #       This invoice should be in failed status and can be retried.
+    return if result.error.messages.dig(:tax_error)
```
ivannovosad, line 19: Do we add a spec for this case?

app/services/fees/commitments/minimum/create_service.rb
lovrocolic: I removed it since usually taxes are applied only on one place where we need to determine if it is related to tax provider or not. It worked fine, but since taxes were applied in minimum commitment service, tax application was skipped here: https://github.com/getlago/lago-api/blob/7141f81087734e5dca1332884d073b8cf3ae3453/app/services/invoices/compute_amounts_from_fees.rb#L14. With this change, as far as I see, the logic has not changed, just tax application is moved to the right service. Wdyt?

ivannovosad (approved):

PR #2496, opened 2024-08-27
Fix(anrok): return only number of failed invoices

annvelents:
When returning amount of failed invoices for the anrok integration, filter invoices with only status 'failed'

### DIFF
--- a/app/graphql/types/integrations/anrok.rb
+++ b/app/graphql/types/integrations/anrok.rb
@@ -30,7 +30,8 @@ module Types
       end
 
       def failed_invoices_count
-        Invoice.where(id: object.error_details.where(owner_type: 'Invoice').select(:owner_id)).failed.count
+        Invoice.where(organization_id: object.organization_id, status: 'failed')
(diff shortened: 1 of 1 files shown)


app/graphql/types/integrations/anrok.rb
```diff
@@ -30,7 +30,7 @@ def external_account_id
       end
 
       def failed_invoices_count
-        object.error_details.where(owner_type: 'Invoice').kept.count
+        Invoice.where(id: object.error_details.where(owner_type: 'Invoice').select(:owner_id)).failed.count
```
lovrocolic, line 33: can't we make it easier without sub-query? e.g. `object.organization.invoices.failed.count`?

lovrocolic (approved):

app/graphql/types/integrations/anrok.rb
annvelents: My concern was that potentially not all invoices can be failed because of tax error (in future). Having this we're sure that we're showing only the failed invoices due to errors with this integration, and anyway AR transforms it all into one query only

PR #2498, opened 2024-08-28
feat(dunning): Extend payment request response

rsempe:
## Roadmap Task

## Context

## Description

The goal of this change is to add:

- customer
- invoices
- payment_status

to the payment request response.

ℹ️ `payment_attempts` and `ready_for_payment_processing` are not returned for invoices, so we decided to not return them for a payment request as well.


ancorcruz (approved):
LGTM!

PR #2512, opened 2024-08-29
Feat(anrok): add enumed_tax_code to tax breakdown serialization

annvelents:
to display addon taxes the same way when creating a one-off invoice as when we show fee taxes in all the other invoices, FE needs enumed_tax_code to be sent in the same format as for Invoice::AppliedTax (which is usually created when creating an invoice)

### DIFF
--- a/app/graphql/types/integrations/anrok_objects/breakdown_object.rb
+++ b/app/graphql/types/integrations/anrok_objects/breakdown_object.rb
@@ -6,6 +6,7 @@ module Types
       class BreakdownObject < Types::BaseObject
         graphql_name 'AnrokBreakdownObject'
 
+        field :enumed_tax_code, Types::Invoices::AppliedTaxes::WholeInvoiceApplicableTaxCodeEnum, null: true
         field :name, String, null: true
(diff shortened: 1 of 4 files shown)


ansmonjol (approved):

app/graphql/types/integrations/anrok_objects/breakdown_object.rb
```diff
@@ -14,6 +17,14 @@ class BreakdownObject < Types::BaseObject
         def rate
           BigDecimal(object.rate)
         end
+
+        def tax_code
+          object.name&.parameterize(separator: '_')
```
ancorcruz, line 22: would you consider memoize the tax_code, as it is currently calculated twice at `enumed_tax_code` method?

```suggestion
          @tax_code ||= object.name&.parameterize(separator: '_')
```

ancorcruz (approved):
LGTM

app/graphql/types/integrations/anrok_objects/breakdown_object.rb
annvelents: ah, right, thank you! 🙈

PR #2521, opened 2024-08-30
feat(dunning): Add update_payment_status for stripe payments

ancorcruz:
## Roadmap Task

👉

 ## Context

 ## Description

The goal of this change is to update the status of stripe payments for payment requests.

### DIFF
--- a/app/services/payment_requests/payments/stripe_service.rb
+++ b/app/services/payment_requests/payments/stripe_service.rb
@@ -47,11 +47,11 @@ module PaymentRequests
           payable_payment_status = payable_payment_status(payment.status)
           update_payable_payment_status(
             payment_status: payable_payment_status,
-            processing: payment.status == 'processing'
+            processing: payment.status == "processing"
           )
           update_invoices_payment_status(
             payment_status: payable_payment_status,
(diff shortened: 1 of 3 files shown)


app/services/payment_requests/payments/stripe_service.rb
```diff
@@ -82,13 +82,39 @@ def generate_payment_url
...
+        result.payment = payment
+        result.payable = payment.payable
+        return result if payment.payable.payment_succeeded?
+
+        payment.update!(status:)
+
+        processing = status == "processing"
+        payment_status = payable_payment_status(status)
+        update_payable_payment_status(payment_status:, processing:)
+        update_invoices_payment_status(payment_status:, processing:)
```
rsempe, line 113: what about doing it inside a `ActiveRecord::Base.transaction`?

rsempe (approved):

app/services/payment_requests/payments/stripe_service.rb
ancorcruz: I would rather leave as it is... these methods sent webhooks, if wrapping them within a transaction the db change would be rolled back while webhook will be sent to users...

PR #2532, opened 2024-09-03
misc: prevent applying taxes for current usage where it is not necessary

lovrocolic:
## Context

In current usage, taxes are relevant only in API context and in Ongoing Wallet Balance Job. In other cases, tax computation should be skipped.


nudded (approved):

vincent-pochet (approved):
:rocket:

PR #2539, opened 2024-09-05
fix: Compare terminated_at to timestamp for terminated subscription

rsempe:
## Issue

After terminating a subscription, when there are several draft invoices linked to the same subscription (i.e. several invoice subscriptions) and we refresh the oldest one, the corresponding invoice subscription is updated with the boundaries of the most recent invoice subscription.

## Objective

The goal of this PR is to check the termination date of the subscription and compare it with the given timestamp, in addition of checking if the subscription is terminated.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -166,7 +166,10 @@ class Subscription < ApplicationRecord
   end
 
   def terminated_at?(timestamp)
-    terminated? && terminated_at <= timestamp
+    return false unless terminated?
+    return false if terminated_at.nil? || timestamp.nil?
+
+    terminated_at <= timestamp
   end
 end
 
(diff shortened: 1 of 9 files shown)


app/services/subscriptions/dates/monthly_service.rb
```diff
@@ -12,7 +12,7 @@ def compute_from_date(date = base_date)
       end
 
       def compute_charges_from_date
-        if terminated?
+        if subscription.terminated_at?(billing_at)
```
vincent-pochet, line 15: Here it was using the method defined as:
```ruby
def terminated?
  subscription.terminated? && !subscription.next_subscription
end
```

Maybe we should keep the `&& !subscription.next_subscription` as well (meaning upgrade this `terminated?` itself) what do you think?

rsempe: ohh good spot 👍 - thanks!

PR #2543, opened 2024-09-06
misc(cleanup) Fix billing service call & introduce rubocop ensuring code consistency

nudded:
## Description

We've built several subclasses of BaseService that don't adhere to the common pattern of having an `initialize` method with arguments and a `call` method without arguments. In an effort to standardize this pattern, a rubocop is added that validates that the call method definition on subclasses of BaseService does not have any arguments.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,7 @@ require:
   - rubocop-rails
   - rubocop-thread_safety
   - rubocop-graphql
+  - ./lib/cops/service_call_cop.rb
 
 inherit_gem:
   standard: config/base.yml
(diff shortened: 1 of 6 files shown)


brunomiguelpinto (approved):

rsempe (approved):

annvelents (approved):

ancorcruz (approved):
looks great, funny times ahead updating old services.

nudded:
> looks great, funny times ahead updating old services.

I debated adding the rubocop ignore rule to all existing violations, could still be the easiest way forward 🤔

PR #2553, opened 2024-09-09
feat: Add firstname and last name to queries

brunomiguelpinto:
## Context
We have enhanced the search functionality for both Customers and Invoices to allow searches by `firstname` and `lastname` fields, in addition to the existing name field.

### DIFF
--- a/spec/queries/customers_query_spec.rb
+++ b/spec/queries/customers_query_spec.rb
@@ -88,16 +88,16 @@ RSpec.describe CustomersQuery, type: :query do
   end
 
   context 'when searching for lastname "Doe"' do
-    let(:search_term) { 'Doe' }
+    let(:search_term) { 'Johnson' }
 
     it 'returns only one customer' do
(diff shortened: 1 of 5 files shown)


spec/queries/customers_query_spec.rb
```diff
@@ -71,4 +71,34 @@
...
+        expect(returned_ids.count).to eq(1)
+        expect(returned_ids).to include(customer_second.id)
+        expect(returned_ids).not_to include(customer_first.id)
+        expect(returned_ids).not_to include(customer_third.id)
+      end
+    end
+  end
+
+  context 'when searching for lastname "Doe"' do
+    let(:search_term) { 'Johnson' }
```
ivannovosad, line 91:
```suggestion
  context 'when searching for lastname "Johnson"' do
    let(:search_term) { 'Johnson' }
```

ivannovosad (changes requested):

PR #2561, opened 2024-09-10
fix(subscription): Make sure terminated invoice always comes with the right boundaries

vincent-pochet:
## Context

Some termination invoices failed to refresh or are generated with the wrong boundaries.
This is related to a small mismatch with the time comparison when looking if the subscription was terminated before the invoice timestamp.

## Description

This PR fixes the `subscription#terminated_at?` comparison so that it always compare time with the second precision, to avoid side effects related to rounding when persisting `ActiveRecord` objects.

A lot of `DateTime` objects are also converted into proper `Time.zone` to fix potential issues around time management as DateTime should not be used in Lago's context (see https://docs.datadoghq.com/code_analysis/static_analysis_rules/ruby-best-practices/no-datetime/)

### DIFF
--- a/app/services/invoices/create_invoice_subscription_service.rb
+++ b/app/services/invoices/create_invoice_subscription_service.rb
@@ -133,7 +133,7 @@ module Invoices
       # NOTE: upgrading is used as a not persisted reason as it means
       #       one subscription starting and a second one terminating
       return invoicing_reason if invoicing_reason.to_sym != :upgrading
-      return :subscription_terminating if subscription.terminated? && subscription.terminated_at <= timestamp
+      return :subscription_terminating if subscription.terminated_at?(timestamp)
 
       :subscription_starting
     end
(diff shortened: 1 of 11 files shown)


app/services/invoices/create_invoice_subscription_service.rb
```diff
@@ -133,7 +133,7 @@ def invoicing_reason_for_subscription(subscription)
       # NOTE: upgrading is used as a not persisted reason as it means
       #       one subscription starting and a second one terminating
       return invoicing_reason if invoicing_reason.to_sym != :upgrading
-      return :subscription_terminating if subscription.terminated? && subscription.terminated_at <= timestamp
+      return :subscription_terminating if subscription.terminated? && subscription.terminated_at?(timestamp)
```
nudded, line 136: you can remove the `subscription.terminated?` check if you use this method I believe.

vincent-pochet: :facepalm: thanks!

PR #2565, opened 2024-09-10
misc: Remove months restriction from gross revenues

rsempe:
## Roadmap Task

 ## Context

 ## Description

The goal of this change is to not setting `months: 12` each time we're fetching gross revenues from graphql.


ancorcruz (approved):
LGTM!

PR #2566, opened 2024-09-10
misc: create a random id on generated plain object

ansmonjol:
Create a random ID on a plain object created “on the flight”, to ease some FE logic

### DIFF
--- a/app/graphql/types/customers/usage/charge.rb
+++ b/app/graphql/types/customers/usage/charge.rb
@@ -8,12 +8,17 @@ module Types
 
         field :amount_cents, GraphQL::Types::BigInt, null: false
         field :events_count, Integer, null: false
+        field :id, ID, null: false
         field :units, GraphQL::Types::Float, null: false
 
         field :billable_metric, Types::BillableMetrics::Object, null: false
         field :charge, Types::Charges::Object, null: false
         field :filters, [Types::Customers::Usage::ChargeFilter], null: true
         field :grouped_usage, [Types::Customers::Usage::GroupedUsage], null: false
+
+        def id
+          SecureRandom.uuid
+        end
 
         def units
           object.map { |f| BigDecimal(f.units) }.sum
--- a/app/graphql/types/customers/usage/grouped_usage.rb
+++ b/app/graphql/types/customers/usage/grouped_usage.rb
@@ -8,10 +8,15 @@ module Types
 
         field :amount_cents, GraphQL::Types::BigInt, null: false
         field :events_count, Integer, null: false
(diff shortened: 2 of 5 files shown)


app/graphql/types/customers/usage/charge.rb
```diff
@@ -8,13 +8,18 @@ class Charge < Types::BaseObject
 
         field :amount_cents, GraphQL::Types::BigInt, null: false
         field :events_count, Integer, null: false
+        field :id, ID, null: false
```
ivannovosad, line 11: @ansmonjol Can we add the new field to the [spec](https://github.com/getlago/lago-api/blob/main/spec/graphql/types/customers/usage/charge_spec.rb#L8) as well, please?

ivannovosad (changes requested):

app/graphql/types/customers/usage/charge.rb
ivannovosad: I just added it.

ivannovosad (approved):

lovrocolic (approved):

app/graphql/types/customers/usage/charge.rb
```diff
@@ -8,13 +8,18 @@ class Charge < Types::BaseObject
...
         field :units, GraphQL::Types::Float, null: false
 
         field :billable_metric, Types::BillableMetrics::Object, null: false
         field :charge, Types::Charges::Object, null: false
         field :filters, [Types::Customers::Usage::ChargeFilter], null: true
         field :grouped_usage, [Types::Customers::Usage::GroupedUsage], null: false
 
+        def id
+          SecureRandom.uuid
```
nudded, line 20: I think just `Random.uuid` would be good enough here? SecureRandom has some overhead to make sure it's secure, is that needed for this use-case?

app/graphql/types/customers/usage/charge.rb
ansmonjol: Thanks a lot!!

app/graphql/types/customers/usage/charge.rb
ansmonjol: Well, I just need a unique ID for each items, it does not matter if the ID changes every time I fetch the data.

Sounds like `Random.uuid` can potentially create twice or more the same ID?
This object is received in an array with potentially many others.

nudded: it's not possible for Random.uuid to generate the same ID. However, my concerns were based on outdated information and it shouldn't matter 👍

PR #2571, opened 2024-09-11
feat (tax-integrations): support endpoint for reporting credit notes in tax provider

lovrocolic:
## Context

Currently Lago is implementing integration with tax provider Anrok

## Description

This PR adds base service which sends credit note to the Anrok and creates transaction there.

### DIFF
--- a/app/services/integrations/aggregator/taxes/credit_notes/create_service.rb
+++ b/app/services/integrations/aggregator/taxes/credit_notes/create_service.rb
@@ -38,46 +38,12 @@ module Integrations
           attr_reader :credit_note
 
           def payload
-            [
-              {
-                'id' => "cn_#{credit_note.id}",
-                'issuing_date' => credit_note.issuing_date,
-                'currency' => credit_note.currency,
-                'contact' => {
-                  'external_id' => integration_customer&.external_customer_id || customer.external_id,
-                  'name' => customer.name,
-                  'address_line_1' => customer.shipping_address_line1 || customer.address_line1,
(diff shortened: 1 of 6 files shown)


app/services/integrations/aggregator/taxes/credit_notes/create_service.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+module Integrations
+  module Aggregator
+    module Taxes
+      module CreditNotes
+        class CreateService < Integrations::Aggregator::Taxes::Invoices::BaseService
```
annvelents, line 7: I thought we want to have CreditNotes service to be separate from Invoice service, and in this case shouldn't we have a separate Base service for CreditNotes? or if it will be reused, extract the base from invoices, and have `Integrations::Aggregator::Taxes::BaseService`?

lovrocolic: Well, the point was to have a dedicated service for credit notes so that we can centralise credit note logic and put it from the start in the right place. I wouldn't say it is problematic to inherit from invoices base service since single-responsibility principle is still satisfied because the main goal was just to have this dedicated service. However, if you want, we can also structure a bit different parent classes. Since everything is actually inherited from `Integrations::Aggregator::BaseService` maybe we can upgrade a bit the full stack of inheritance (i know few other inheritance related parts that can be updated), but since it could be bigger change, let's do it as the last step of this story but before the QA. How does it sound?

annvelents (approved):

PR #2576, opened 2024-09-12
fix(invoice): avoid failure on error not related to taxes

vincent-pochet:
## Description

This PR is fixing a regression recently introduced by the external tax management feature. A test was added to handle `tax_error` and was testing a specific error code on a service failure, but not all `BaseService::FailedResult` are responding to the `error` method

### DIFF
--- a/app/services/invoices/subscription_service.rb
+++ b/app/services/invoices/subscription_service.rb
@@ -139,7 +139,10 @@ module Invoices
     end
 
     def tax_error?(fee_result)
-      !fee_result.success? && fee_result&.error&.code == 'tax_error'
+      return false if fee_result.success?
+      return false unless fee_result.error.is_a?(BaseService::ServiceFailure)
+
+      fee_result.error.code == 'tax_error'
     end
   end
 end


app/services/invoices/subscription_service.rb
```diff
@@ -139,7 +139,10 @@ def flag_lifetime_usage_for_refresh
     end
 
     def tax_error?(fee_result)
-      !fee_result.success? && fee_result&.error&.code == 'tax_error'
+      return false if fee_result.success?
```
nudded, line 142: There's 7 implementors of `tax_error?` and looks like a total of 3 different implementations. Maybe a simple module that's included in BaseService? Or a class method somewhere? 🤔 Just so we can align all implementations maybe? what do you think?

nudded (approved):

app/services/invoices/subscription_service.rb
vincent-pochet: I agree. This will be tacled in a dedicated pull request as it might require some deeper changes

PR #2577, opened 2024-09-12
feat(tax-integration): add logic for reporting credit note to the tax provider

lovrocolic:
## Context

Currently Lago is working on integration with tax provider Anrok.

## Description

This PR covers sending credit notes to the tax provider and handling tax provider errors

### DIFF
--- a/spec/services/credit_notes/provider_taxes/report_service_spec.rb
+++ b/spec/services/credit_notes/provider_taxes/report_service_spec.rb
@@ -146,7 +146,7 @@ RSpec.describe CreditNotes::ProviderTaxes::ReportService, type: :service do
         aggregate_failures do
           expect(credit_note.error_details.tax_error.order(created_at: :asc).last.id).not_to eql(old_error_id)
           expect(credit_note.error_details.tax_error.count).to be(1)
-          expect(credit_note.error_details.tax_error.order(created_at: :asc).last.discarded?).to be(false)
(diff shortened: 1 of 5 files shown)


spec/services/credit_notes/provider_taxes/report_service_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+
+      it 'resolves old tax error and creates new one' do
+        old_error_id = credit_note.reload.error_details.order(created_at: :asc).last.id
+
+        report_service.call
+
+        aggregate_failures do
+          expect(credit_note.error_details.tax_error.order(created_at: :asc).last.id).not_to eql(old_error_id)
+          expect(credit_note.error_details.tax_error.count).to be(1)
+          expect(credit_note.error_details.tax_error.order(created_at: :asc).last.discarded?).to be(false)
```
ancorcruz, line 149: I think this is more idiomatic

```suggestion
          expect(credit_note.error_details.tax_error.order(created_at: :asc).last).not_to be_discarded
```

However I wondering if it won't be better to just set the expectation the `ErrorDetails::CreateService` is called with the given parameters as you are testing here the implementation of this service (in the whole test scenario). What do you think?

ancorcruz (approved):
LGTM!

spec/services/credit_notes/provider_taxes/report_service_spec.rb
lovrocolic: updated

PR #2578, opened 2024-09-12
misc: refactor tax provider inheritance for the aggregator scope

lovrocolic:
## Context

Currently Lago is working on integration with tax provider Anrok

## Description

This PR refactors inheritance inside the aggregator scope

### DIFF
--- a/app/services/integrations/aggregator/taxes/credit_notes/create_service.rb
+++ b/app/services/integrations/aggregator/taxes/credit_notes/create_service.rb
@@ -8,7 +8,7 @@ module Integrations
           def initialize(credit_note:)
             @credit_note = credit_note
 
-            super(integration:)
+            super()
           end
 
           def action_path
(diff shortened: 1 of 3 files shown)


app/services/integrations/aggregator/taxes/credit_notes/create_service.rb
```diff
@@ -4,11 +4,11 @@ module Integrations
   module Aggregator
     module Taxes
       module CreditNotes
-        class CreateService < Integrations::Aggregator::Taxes::Invoices::BaseService
+        class CreateService < Integrations::Aggregator::Taxes::BaseService
           def initialize(credit_note:)
             @credit_note = credit_note
 
-            super(invoice: credit_note.invoice)
+            super(integration:)
```
annvelents, line 11: since integration is defined at base service, what do you think if here we will call super without arguments, and will call 
`super(integration:)` form the `Integrations::Aggregator::Taxes::BaseService`, where it's defined?

annvelents: (it's just a matter of taste, so if you don't like it feel free to ignore)

lovrocolic: For me it is the same, but if you have this preference, here it is:)

annvelents (approved):

PR #2583, opened 2024-09-12
Fix: save fees with taxes_rate when creating one off invoice

annvelents:
When creating one off invoice and applying taxes on fee, we need to save fee, otherwise calculated values on fees are set, but not saved

### DIFF
--- a/app/services/fees/one_off_service.rb
+++ b/app/services/fees/one_off_service.rb
@@ -109,6 +109,7 @@ module Fees
 
         res = Fees::ApplyProviderTaxesService.call(fee:, fee_taxes:)
         res.raise_if_error!
+        fee.save
       end
 
       taxes_result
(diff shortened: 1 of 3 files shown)


brunomiguelpinto (approved):

app/services/fees/one_off_service.rb
```diff
@@ -109,6 +109,7 @@ def apply_provider_taxes(fees_result)
 
         res = Fees::ApplyProviderTaxesService.call(fee:, fee_taxes:)
         res.raise_if_error!
+        fee.save
```
lovrocolic, line 112: Can you please explain which data are not stored on fee? If we need it, can you please use `save! `to run validations?

annvelents: not saved taxes_rate on fee was causing a bug. Will update to `save!`

PR #2584, opened 2024-09-12
fix(subscription): fix tax error digging in all the places

vincent-pochet:
## Context

Error when not raised anymore by the `BillSubscrpitionJob`

## Description

The error happens because `[]` is `true`


nudded (approved):
🚀

PR #2590, opened 2024-09-17
misc: Split graphql definitions into two schemas

rsempe:


config/routes.rb
```diff
@@ -4,9 +4,13 @@
   mount Sidekiq::Web, at: '/sidekiq' if defined? Sidekiq::Web
   mount Karafka::Web::App, at: '/karafka' if ENV['KARAFKA_WEB']
 
-  mount GraphiQL::Rails::Engine, at: '/graphiql', graphql_path: '/graphql' if Rails.env.development?
+  if Rails.env.development?
+    mount GraphiQL::Rails::Engine, at: '/customer_portal_graphiql', as: 'customer_portal_graphiql_rails', graphql_path: '/customer_portal_graphql'
+    mount GraphiQL::Rails::Engine, at: '/api_graphiql', as: 'api_graphiql_rails', graphql_path: '/api_graphql'
+  end
 
-  post '/graphql', to: 'graphql#execute'
```
vincent-pochet, line 9: You might want to keep this route temporary, otherwise the deploy will break the front-end completely :boom:

Rakefile
```diff
@@ -3,9 +3,19 @@
...
-require 'graphql/rake_task'
+require_relative "config/application"
+require "graphql/rake_task"
 
 Rails.application.load_tasks
 
-GraphQL::RakeTask.new(schema_name: 'LagoApiSchema')
+GraphQL::RakeTask.new(
+  schema_name: "ApiSchema",
+  idl_outfile: "api_schema.graphql",
```
vincent-pochet, line 13: What do you thing of using something like `graphql_schemas/[api|customer_portal].[graphql|json]`?

vincent-pochet (commented):
Nice!

Rakefile
rsempe: Good spot - Done

config/routes.rb
rsempe: True 👍

app/graphql/types/customer_portal_mutation_type.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Types
+  class CustomerPortalMutationType < Types::BaseObject
```
ancorcruz, line 4: should we namespace it instead? (some fore query, etc)

```suggestion
  module CustomerPortal
    class MutationType < Types::BaseObject
```

ancorcruz (approved):
LGTM!

just wondering if we should namespace CustomerPortal as it is a different application to isolate it from the authenticate api.

config/routes.rb
```diff
@@ -4,9 +4,16 @@
...
   mount Karafka::Web::App, at: '/karafka' if ENV['KARAFKA_WEB']
 
-  mount GraphiQL::Rails::Engine, at: '/graphiql', graphql_path: '/graphql' if Rails.env.development?
+  if Rails.env.development?
+    mount GraphiQL::Rails::Engine, at: '/customer_portal_graphiql', as: 'customer_portal_graphiql_rails', graphql_path: '/customer_portal_graphql'
+    mount GraphiQL::Rails::Engine, at: '/api_graphiql', as: 'api_graphiql_rails', graphql_path: '/api_graphql'
+  end
+
+  # TODO: Remove this deprecated route.
+  post '/graphql', to: 'graphql/api#execute'
```
ancorcruz, line 13: this change will require sync with FE to update customer portal requests, isn't it?

The current FE implementation will send request to this graphQL endpoint, and will be routed to the API namespace queries/mutations. Then, with the changes in this PR, as an admin accessing the customer portal will work as you have the GraphQL cookie with your privileged admin user access. However, I guess as a real customer accessing the portal with the url (incognito windows) it will fail as the new API controller does not handle the customer portal user variable.

This is related to the previous comments in the query and mutation types for the API namespace.

app/graphql/types/api/mutation_type.rb
```diff
@@ -0,0 +1,150 @@
...
+      field :destroy_plan, mutation: Mutations::Plans::Destroy
+      field :update_plan, mutation: Mutations::Plans::Update
+
+      field :create_customer, mutation: Mutations::Customers::Create
+      field :destroy_customer, mutation: Mutations::Customers::Destroy
+      field :update_customer, mutation: Mutations::Customers::Update
+      field :update_customer_invoice_grace_period, mutation: Mutations::Customers::UpdateInvoiceGracePeriod
+
+      field :download_customer_portal_invoice, mutation: Mutations::CustomerPortal::DownloadInvoice
+      field :generate_customer_portal_url, mutation: Mutations::CustomerPortal::GenerateUrl
```
ancorcruz, line 28: Could we move this mutation out of the CustomerPortal namespace as this is a CustomerPortal url generation but from the admin POV? hence, not part of the CustomerPortal

app/graphql/types/api/mutation_type.rb
```diff
@@ -0,0 +1,150 @@
...
+      field :create_plan, mutation: Mutations::Plans::Create
+      field :destroy_plan, mutation: Mutations::Plans::Destroy
+      field :update_plan, mutation: Mutations::Plans::Update
+
+      field :create_customer, mutation: Mutations::Customers::Create
+      field :destroy_customer, mutation: Mutations::Customers::Destroy
+      field :update_customer, mutation: Mutations::Customers::Update
+      field :update_customer_invoice_grace_period, mutation: Mutations::Customers::UpdateInvoiceGracePeriod
+
+      field :download_customer_portal_invoice, mutation: Mutations::CustomerPortal::DownloadInvoice
```
ancorcruz, line 27: Shouldn't this one be moved to the customer portal mutation type? (delete it from here)

app/graphql/types/api/query_type.rb
```diff
@@ -0,0 +1,71 @@
...
+      field :credit_note_estimate, resolver: Resolvers::CreditNotes::EstimateResolver
+      field :credit_notes, resolver: Resolvers::CreditNotesResolver
+      field :current_version, resolver: Resolvers::VersionResolver
+      field :customer, resolver: Resolvers::CustomerResolver
+      field :customer_invoices, resolver: Resolvers::Customers::InvoicesResolver
+      field :customer_portal_invoice_collections, resolver: Resolvers::CustomerPortal::Analytics::InvoiceCollectionsResolver
+      field :customer_portal_invoices, resolver: Resolvers::CustomerPortal::InvoicesResolver
+      field :customer_portal_organization, resolver: Resolvers::CustomerPortal::OrganizationResolver
+      field :customer_portal_overdue_balances, resolver: Resolvers::CustomerPortal::Analytics::OverdueBalancesResolver
+      field :customer_portal_user, resolver: Resolvers::CustomerPortal::CustomerResolver
```
ancorcruz, line 28: Same as previous comments in the mutation side, shouldn't all this customer portal queries be moved to the customer portal query type? (delete them from here)

ancorcruz (changes requested):
Looks great but I have some concerns about the change and FE, expressed comments...

ancorcruz:
I think the migration will require a smoother and progressive change:

- [ ] 1. keep the original implementation of the graphQL controller as it is... while introducing the 2 new endpoints
- [ ] 2. API namespace will remain as is the current implementation, while the CustomerPortal namespace will have the current customer portal mutations, queries, and types.
- [ ] 3. All new CustomerPortal stuff should be included in its namespace only.
- [ ] 4. Frontend needs to migrate customer portal requests to the new endpoint (customer portal graphql endpoint)
- [ ] 5. Once Frontend is ready, we can remove the customer portal mutations and queries from the API namespace.
- [ ] 6. Then we can route old graphql endpoint to the new API namespaced controller and drop the old controller.
- [ ] 7. FE needs to migrate admin graphql queries to the new API endpoint.
- [ ] 8. Once FE is ready, we can drop the old graphql route.

Something to consider in order to require less FE changes would be keep the old graphQL endpoint route for the API namespace instead of introducing a new one for it. We will change the controller it is routed to but for FE nothing changed. This removes steps 7 and 8 above.

WDYT?

app/graphql/types/api/mutation_type.rb
rsempe: it has been pushed back here with the rebase - removing 👍

app/graphql/types/api/mutation_type.rb
rsempe: yep sure - but in a separate PR ;)

PR #2597, opened 2024-09-17
feat: add hubspot connection services

brunomiguelpinto:
## Roadmap Task

## Context

Integration with HubSpot to import customers, subscriptions and invoices.

## Description

Add `Hubspot::CreateService` and `Hubspot::UpdateService`  services and modify existing ones where necessary to support hubspot.

### DIFF
--- a/app/services/integrations/aggregator/send_private_app_token_service.rb
+++ b/app/services/integrations/aggregator/send_private_app_token_service.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Integrations
+  module Aggregator
+    class SendPrivateAppTokenService < BaseService
+      def action_path
+        "connection/#{integration.connection_id}/metadata"
+      end
+
+      def call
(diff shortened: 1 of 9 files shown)


app/services/integrations/aggregator/send_private_app_token_service.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Integrations
+  module Aggregator
+    class SendPrivateAppTokenService < BaseService
```
ivannovosad, line 5: Can we also add spec for this service, pls?

ivannovosad (changes requested):

app/services/integrations/aggregator/send_private_app_token_service.rb
brunomiguelpinto: Sure forgot that one 👍

ivannovosad (approved):

PR #2599, opened 2024-09-18
feat(cp-improv): Customer Portal expose and edit customer info

ancorcruz:
## Roadmap Task

👉 https://getlago.canny.io/feature-requests/p/add-features-to-the-customer-portal

 ## Context

The current customer portal already exposes some customer information. However, it is missing some important pieces of customer information. Also, we want to allow customer to edit their own billing information.

 ## Description

This change improves the customer access to their own billing information within the customer portal and reduce organization support requirements to maintain customer billing information up to date, allowing customer to update their own billing information.

Also, it increases customer data security as not all fields could be retrieved from the customer portal GraphQL requests. Only the data required for the customer portal.


app/graphql/types/customer_portal/customers/object.rb
```diff
@@ -0,0 +1,42 @@
...
+        field :city, String, null: true
+        field :country, Types::CountryCodeEnum, null: true
+        field :state, String, null: true
+        field :zipcode, String, null: true
+
+        field :shipping_address, Types::Customers::Address, null: true
+
+        def billing_configuration
+          {
+            id: "#{object&.id}-c0nf",
```
rsempe, line 35: why this?

ancorcruz: 🤷‍♂️ I don't know, this is already in the admin site customers object type... not sure what does it means...

https://github.com/getlago/lago-api/blob/037290f94a6c25ce797cbad7c1af01c7aac0c64b/app/graphql/types/customers/object.rb#L135-L138

ancorcruz: @lovrocolic you introduced this change with the commit https://github.com/getlago/lago-api/commit/ce24c19fa267c0d14457042ddece5fafeced228e

Do you remember why?

app/graphql/types/customer_portal/customers/object.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Types
+  module CustomerPortal
+    module Customers
+      class Object < Types::BaseObject
```
rsempe, line 6: Why not using the customer object directly?

app/services/customer_portal/customer_update_service.rb
```diff
@@ -0,0 +1,67 @@
...
+        customer.shipping_state = shipping_address[:state] if shipping_address.key?(:state)
+        customer.shipping_country = shipping_address[:country]&.upcase if shipping_address.key?(:country)
+
+        customer.save!
+        customer.reload
+
+        tax_codes = []
+        if customer.organization.eu_tax_management
+          # This service does not return a 'result' object but a string
+          tax_codes << Customers::EuAutoTaxesService.call(customer:)
```
rsempe, line 46: are we sure we have to handle this even in the customer portal update?

rsempe (approved):

app/services/customer_portal/customer_update_service.rb
ancorcruz: I think so, based on this thread...

https://www.notion.so/Spec-Improve-Customer-Portal-Expose-customer-information-and-edit-customer-information-0ffef63110d2804883aceea71e7c14dc?d=21b49a90e2854ae88164a21ef4047708&pvs=4#c892e96bc4a5413186bdb9b6fdc64ed9

app/graphql/types/customer_portal/customers/object.rb
ancorcruz: looks like frontend requires an id, for some reason it can't be random and we do not have a DB id as this information is defined within the customer record.

https://github.com/getlago/lago-api/pull/817#discussion_r1102645644

PR #2600, opened 2024-09-18
fix (accounting-integrations): send webhook for incomplete payload

lovrocolic:
## Context

For invalid payload due to incomplete configuration, we raise error (500) instead of silently notifying users via webhooks.

## Description

This PR covers described scenario


ivannovosad (approved):
🚀

PR #2602, opened 2024-09-19
feat(tax-invoices): add required invoice title for some countries

lovrocolic:
## Context

In some countries, invoices must include specific information when taxes are included. One of these obligations is to display `Tax invoice` in the header of the document, instead of `Invoice`.

## Description

This PR handles described use-case. Currently we are handling only following country codes: `AU`, `AE`, `NZ` and `ID`.

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -325,7 +325,7 @@ class Invoice < ApplicationRecord
   def document_invoice_name
     return I18n.t('invoice.prepaid_credit_invoice') if credit?
 
-    return I18n.t('invoice.document_tax_name') if %w[AU AE ID].include?(organization.country)
+    return I18n.t('invoice.document_tax_name') if %w[AU AE ID NZ].include?(organization.country)
 
     I18n.t('invoice.document_name')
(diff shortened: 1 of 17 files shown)


app/models/invoice.rb
```diff
@@ -322,6 +322,14 @@ def should_sync_sales_order?
       end
   end
 
+  def document_invoice_name
+    return I18n.t('invoice.prepaid_credit_invoice') if credit?
+
+    return I18n.t('invoice.document_tax_name') if %w[AU AE ID NZ].include?(organization.country)
```
nudded, line 328: what about cases where the customer is in that country?

lovrocolic: This spec is covering just organization country and we will see if it satisfy customer needs. In general, there are bunch of other conditions, like customer country or thresholds which can define if to present tax invoice or not. However, we will tackle it in another iteration if there will be need for it...

nudded: okay 👍

nudded (approved):

PR #2603, opened 2024-09-19
misc: Split graphql definitions into two schemas

rsempe:

This first PR allows to keep the original implementation of the graphQL controller as it is... while introducing the 2 new endpoints.


app/graphql/types/api/mutation_type.rb
```diff
@@ -0,0 +1,149 @@
...
+      field :create_plan, mutation: Mutations::Plans::Create
+      field :destroy_plan, mutation: Mutations::Plans::Destroy
+      field :update_plan, mutation: Mutations::Plans::Update
+
+      field :create_customer, mutation: Mutations::Customers::Create
+      field :destroy_customer, mutation: Mutations::Customers::Destroy
+      field :update_customer, mutation: Mutations::Customers::Update
+      field :update_customer_invoice_grace_period, mutation: Mutations::Customers::UpdateInvoiceGracePeriod
+
+      field :generate_customer_portal_url, mutation: Mutations::CustomerPortal::GenerateUrl
```
ancorcruz, line 27: Could we consider move this mutation out of the `CustomerPortal` namespace?

What do you think about `Mutations::GenerateCustomerPortalUrl`?

ancorcruz (approved):
LGTM!

vincent-pochet (approved):

PR #2608, opened 2024-09-20
misc: Remove deprecated graphql schema

rsempe:

This PR remove deprecated graphql endpoint and schema.


config/routes.rb
```diff
@@ -4,9 +4,6 @@
   mount Sidekiq::Web, at: '/sidekiq' if defined? Sidekiq::Web
   mount Karafka::Web::App, at: '/karafka' if ENV['KARAFKA_WEB']
 
-  # TODO(graphql_schema): This route is deprecated and should be removed.
-  post '/graphql', to: 'graphql#execute'
```
nudded, line 8: Can only be merged once the frontend uses the new endpoint ✔️ Think about how to handle the release for OSS? e.g. this will incur a bit of downtime of the frontend depending on upgrade order. Maybe a mention in the release notes is enough 🤔

nudded: (this note is just here so we don't accidentally merge it)

rsempe: good point - as discussed together, let's keep the old graphql schema available for now

PR #2613, opened 2024-09-20
feat(DynamicPricing) - Add dynamic charge model & validator

nudded:
## Context

AI, CPaaS and Fintech customers can apply a price to a unit that fluctuates over time. Currently Lago does not support this usecase.

## Description

Adds `dynamic` as an option for the possible Charge Models. Also makes sure you can only use it in combination with Sum AGG


app/services/charges/validators/dynamic_service.rb
```diff
@@ -0,0 +1,22 @@
...
+        super
+      end
+
+      private
+
+      def validate_billable_metric
+        # Only sum aggregation is compatible with Dynamic Pricing for now
+        return if charge.billable_metric.sum_agg?
+
+        add_error(field: :billable_metric, error_code: 'invalid_billable_metric_value')
```
vincent-pochet, line 18: For custom aggregation, the validation error is done that way on field `charge_model` with error code `invalid_aggregation_type_or_charge_model`, what do you think of using the same format?

app/services/charges/charge_models/dynamic_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Charges
+  module ChargeModels
+    class DynamicService < Charges::ChargeModels::BaseService
+      protected
+
+      def compute_amount
+        total_units = aggregation_result.full_units_number || units
+        return 0 if total_units.zero?
```
vincent-pochet, line 10: :+1:

app/services/charges/validators/dynamic_service.rb
nudded: I copied https://github.com/getlago/lago-api/blob/fc0e89db025fec4b745763c8ae5c5749f7a4214c/app/services/charges/validators/graduated_percentage_service.rb#L32 this approach. 

The annoying part is that the `validate_charge_model` method strips the field information and moves it all to `:properties`, but having a dedicated validation class is nice. I'll try changing it to how we do it for CustomAggregation :)

nudded: Fixed!

vincent-pochet (approved):

vincent-pochet (approved):

PR #2617, opened 2024-09-23
Feat(anrok): call report credit note job when creating a credit note

annvelents:
When creating a credit note we should schedule a job to report credit note creation to tax provider is customer has tax_provider set up;

When reporting credit note, the reported amount should be taken from the items amounts

### DIFF
--- a/app/services/credit_notes/create_service.rb
+++ b/app/services/credit_notes/create_service.rb
@@ -60,6 +60,7 @@ module CreditNotes
         deliver_webhook
         deliver_email
         handle_refund if should_handle_refund?
+        report_to_tax_provider if should_report_to_tax_provider?
 
         if credit_note.should_sync_credit_note?
           Integrations::Aggregator::CreditNotes::CreateJob.perform_later(credit_note:)
@@ -175,6 +176,12 @@ module CreditNotes
       invoice_payment.present?
     end
 
+    def should_report_to_tax_provider?
+      return false unless credit_note.finalized?
+
+      credit_note.customer.anrok_customer.present?
+    end
+
     def invoice_payment
       @invoice_payment ||= credit_note.invoice.payments.order(created_at: :desc).first
     end
@@ -188,6 +195,10 @@ module CreditNotes
       when PaymentProviders::AdyenProvider
         CreditNotes::Refunds::AdyenCreateJob.perform_later(credit_note)
       end
+    end
+
+    def report_to_tax_provider
+      CreditNotes::ProviderTaxes::ReportJob.perform_later(credit_note:)
     end
 
     def compute_amounts_and_taxes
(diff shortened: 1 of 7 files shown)


app/services/credit_notes/create_service.rb
```diff
@@ -175,6 +176,12 @@ def should_handle_refund?
       invoice_payment.present?
     end
 
+    def should_report_to_tax_provider?
```
lovrocolic, line 179: This check is unnecessary here. This snippet is wrapped under `credit_note.finalized?` and check `customer.anrok_customer` is present inside the job so that we avoid this query in the service

annvelents: 👍

spec/services/credit_notes/create_service_spec.rb
```diff
@@ -135,6 +135,20 @@
       let(:service_call) { create_service.call }
     end
 
+    context 'when customer has tax_provider set up' do
```
lovrocolic, line 138: We can avoid mocking in this test:
```
it 'sync with tax provider' do
      expect do
        create_service.call
      end.to have_enqueued_job(CreditNotes::ProviderTaxes::ReportJob)
    end
```

app/services/integrations/aggregator/taxes/credit_notes/payload.rb
```diff
@@ -51,7 +51,7 @@ def cn_item(item)
             {
               'item_id' => fee.item_id,
               'item_code' => mapped_item.external_id,
-              'amount_cents' => (fee.sub_total_excluding_taxes_amount_cents&.to_i || 0) * -1
+              'amount_cents' => (item.amount_cents&.to_i || 0) * -1
```
lovrocolic, line 54: I am pretty sure we should send value without coupons. Can you double check please?

annvelents: we can issue credit note for the amount less than the original fee.
image
in this case the value that is credited for this exact fee is stored in the credit_note_item.amount_cents.
so we should take the anrok details from the fee, but the exact amount that is being send as credit_note is stored in the item.
As well as sum of the credit note amount can be less then the invoice sum, even if we partially return all fees from the invoice

lovrocolic: Previous (Existing) line seems invalid, probably that is some leftover from invoice logic. Good catch!

However, It seems to me that new logic is also not valid. Item amount cents include value without coupons included in that calculation. It seems to me that we should also take coupons into account. You can try to reproduce the case with applying coupons on customer -> issue invoice -> issue CN. I am not sure if this case is covered

annvelents: I see. Didn't know that we apply coupons proportionally in the credit notes.
Will add the calculation logic then

annvelents: added a method similar to the one we have in `credit_note` and `fee` that proportionally calculates item.sub_total_excluding_taxes_amount_cents, thank you!

PR #2628, opened 2024-09-26
feat(tax-integrations): Cover tax deduction case for tax integrations

lovrocolic:
## Context

Lago recently launched integration with tax provider Anrok.

## Description

In some scenarios tax is not applied on the sum of all fees (with coupon applied).

Instead, there are scenarios where tax base is less than the sum of the fees (with coupon applied) and it depends mostly on the country (e.g. Texas in the USA).

This PR is covering described case. `tax_base_rate` is included in tax logic and also used to display correct base in the UI/PDF.

### DIFF
--- a/app/models/invoice/applied_tax.rb
+++ b/app/models/invoice/applied_tax.rb
@@ -11,6 +11,7 @@ class Invoice
 
     monetize :amount_cents,
       :fees_amount_cents,
+      :taxable_base_amount_cents,
       with_model_currency: :amount_currency
 
     validates :amount_cents, numericality: {greater_than_or_equal_to: 0}
@@ -21,6 +22,14 @@ class Invoice
     def applied_on_whole_invoice?
       TAX_CODES_APPLICABLE_ON_WHOLE_INVOICE.include?(tax_code)
     end
+
+    def taxable_base_amount_cents
+      base_amount = self[:taxable_base_amount_cents]
+
+      return fees_amount_cents if base_amount.blank? || base_amount.zero?
+
+      base_amount
+    end
   end
 end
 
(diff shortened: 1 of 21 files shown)


db/migrate/20240924114730_add_taxes_deduction_rate_fields_to_fees_and_applied_taxes.rb
```diff
@@ -0,0 +1,15 @@
...
+  def change
+    safety_assured do
+      change_table :fees do |t|
+        t.float :taxes_base_rate, default: 1.0, null: false
+      end
+
+      change_table :invoices_taxes do |t|
+        t.bigint :taxable_base_amount_cents, default: 0, null: false
+      end
+    end
```
vincent-pochet, line 13: What about the following to avoid `safety_assured`?

```suggestion
    add_column :fees, :taxes_base_rate, :float, default: 0.0, null: false
    add_column :invoices_taxes, :taxable_base_amount_cents, :float, default: 0.0, null: false
```

lovrocolic: updated

app/models/invoice/applied_tax.rb
```diff
@@ -21,25 +22,34 @@ class AppliedTax < ApplicationRecord
     def applied_on_whole_invoice?
       TAX_CODES_APPLICABLE_ON_WHOLE_INVOICE.include?(tax_code)
     end
+
+    def taxable_base_amount_cents
```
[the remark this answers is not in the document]
lovrocolic: changed 👍

vincent-pochet (approved):

PR #2634, opened 2024-09-26
feat: added translations

brunomiguelpinto:
## Context

When an invoice is generated for **paid_in_advance fees** (invoice_type: `advance_charges`), the title of the invoice currently displays "Invoice" and the last line of the PDF shows "Total due." However, since this invoice is already paid, this is misleading for the users.

The goal is to provide more accurate labeling in the invoice title and the total section.

### Changes:
   - For invoices of type `advance_charges`, the title should now display "Paid invoice" instead of "Invoice."   
   - For these invoices, instead of displaying "Total due," we now show "Already paid."
   - on the due date -> now appears already paid without date 

### Changes made in the PDF:
- Conditional logic has been added to check if the invoice type is `advance_charges` and select the correct label

image

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -325,9 +325,12 @@ class Invoice < ApplicationRecord
   def document_invoice_name
     return I18n.t('invoice.prepaid_credit_invoice') if credit?
 
-    return I18n.t('invoice.paid_invoice') if advance_charges?
+    if %w[AU AE ID NZ].include?(organization.country)
+      return I18n.t('invoice.paid_tax_invoice') if advance_charges?
+      return I18n.t('invoice.document_tax_name')
+    end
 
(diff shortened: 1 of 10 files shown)


app/models/invoice.rb
```diff
@@ -325,6 +325,8 @@ def should_sync_sales_order?
   def document_invoice_name
     return I18n.t('invoice.prepaid_credit_invoice') if credit?
 
+    return I18n.t('invoice.paid_invoice') if advance_charges?
```
lovrocolic, line 328: Prepaid credit invoice does not include tax, but charge one does.
For some countries mentioning `tax` is mandatory (line 330), so we probably also need it for advance charge invoice. Maybe @mathieu-lago has idea how to display it properly?

annvelents (approved):

PR #2636, opened 2024-09-27
feat(customer_portal): Add usage resolver to customer portal

rsempe:
## Roadmap Task

👉  [https://getlago.canny.io/feature-requests/p/add-features-to-the-customer-portal](https://getlago.canny.io/feature-requests/p/add-features-to-the-customer-portal)

## Context

We got feedback that this portal is too limited in terms of features and information. Why?

**1st reason:** the number of information displayed is limited

**2nd reason:** the end user cannot take actions out of it (change info, add payment method)

**3rd reason:** the UI and display is not customizable

## Description

The goal of this PR is to return customer usage from the customer portal.

### DIFF
--- a/app/graphql/resolvers/customer_portal/customers/usage_resolver.rb
+++ b/app/graphql/resolvers/customer_portal/customers/usage_resolver.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module Resolvers
+  module CustomerPortal
+    module Customers
+      class UsageResolver < Resolvers::BaseResolver
+        include AuthenticableCustomerPortalUser
+
+        description "Query the usage of the customer on the current billing period"
+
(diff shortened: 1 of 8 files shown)


ancorcruz (approved):
This new resolver spec create a lots of active record object, I understand some are not avoidable, like organizations, customer, etc... However, I'm wondering if we could reduce some, for example there are 8 (4+4) events created with `create_list` for every test case and 3 more with create list within the filter context...

Creating records in DB is needed but also makes our tests slower, we must keep balance, if you could reduce the number of created event records with `create_list` while keeping the test coverage it would be great.

LGTM. 👏

rsempe:
@ancorcruz Yep - I just took the same spec as we have for the existing usage resolver to ensure nothing is broken for customer portal - But yep happy to improve them as well 👍

PR #2641, opened 2024-09-30
misc: Refact some services to the call pattern

vincent-pochet:
## Context

Some services inherinting from the `BaseService` are not using the `SERVCE_CLASS.call` pattern. They are mostly old services that were created before the introduction of the pattern in the project.

## Description

To increase consistency in the code base, this PR turns some old services to the pattern.
The impacted services are:
- `AddOns::CreateService`
- `BillableMetrics::CreateService`
- `Coupons::CreateService`
- `AddOns::CreateService`

More refactor will follow as other services have still to be converted.

### DIFF
--- a/app/services/add_ons/create_service.rb
+++ b/app/services/add_ons/create_service.rb
@@ -2,7 +2,12 @@
 
 module AddOns
   class CreateService < BaseService
-    def create(**args)
+    def initialize(args)
+      @args = args
+      super
+    end
+
+    def call
       ActiveRecord::Base.transaction do
         add_on = AddOn.create!(
           organization_id: args[:organization_id],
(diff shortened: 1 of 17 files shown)


app/services/add_ons/create_service.rb
```diff
@@ -2,7 +2,12 @@
 
 module AddOns
   class CreateService < BaseService
-    def create(**args)
+    def initialize(args)
+      @args = args
+      super
```
lovrocolic, line 7: Hmm just realised that we mostly call just `super` which actually pass current attributes to the parent constructor. Shouldn't we use `super()` here?

lovrocolic (approved):

app/services/add_ons/create_service.rb
vincent-pochet: Well, actually the base service initializer is
```ruby
def initialize(current_user = nil)
    @result = Result.new
    @source = CurrentContext&.source
    result.user = current_user.is_a?(User) ? current_user : nil
  end
```

So it does not really matter, `args` will be assigned to the `current_user` argument and then ignored... The main goal in the end, is actually to get rid of this `current_user` params as it is some old legacy logic that should not be there anymore.

PR #2642, opened 2024-09-30
feat(hubspot): Add new attributes to hubspot integration model

brunomiguelpinto:
## Context

The goal of this change is the deployment of custom properties and objects in HubSpot when a new integration is created via Lago. Specifically, we need to create the necessary fields for both native HubSpot objects (`Companies` and `Contacts`) and create custom objects (`Subscriptions` and `Invoices`)  in the user's HubSpot account. 

## Description

This PR introduces a new job, `SyncCustomObjectsAndPropertiesJob`, which orchestrates the setup of custom objects and property synchronization in HubSpot when a new integration is created via Lago. The job does not handle sending the actual records, and this will be addressed in a future PR.

version control has been implemented to track the deployment state of these objects and properties. This allows us to know what has been deployed and to trigger a sync when sending new records if the schema has been updated.

### DIFF
--- a/app/services/integrations/hubspot/create_service.rb
+++ b/app/services/integrations/hubspot/create_service.rb
@@ -32,6 +32,7 @@ module Integrations
 
         if integration.type == 'Integrations::HubspotIntegration'
           Integrations::Aggregator::SendPrivateAppTokenJob.perform_later(integration:)
+          Integrations::Aggregator::SyncCustomObjectsAndPropertiesJob.perform_later(integration:)
         end
 
         result.integration = integration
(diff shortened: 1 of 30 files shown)


app/services/integrations/hubspot/create_service.rb
```diff
@@ -32,6 +32,7 @@ def call
 
         if integration.type == 'Integrations::HubspotIntegration'
           Integrations::Aggregator::SendPrivateAppTokenJob.perform_later(integration:)
+          Integrations::Aggregator::SyncCustomObjectsAndPropertiesJob.perform_later(integration:)
```
brunomiguelpinto, line 35: move these jobs to hubspot folder?

PR #2643, opened 2024-09-30
Fix: adjust tax precise amount when issuing credit notes

annvelents:
## Context

We have this issue that when we issue Invoice, we apply tax on sum of items,
but when we issue a credit note, we apply tax on each item separately.
as result, some values can be rounded in deffirent direction which result in not matching sum.

Example that our client had (all values are amount_cents):
```
Invoice 001
charge01 = 68_33
charge 02 = 68_33
charge03 = 57_50
charge04 = 85_00
```

if we apply 20% taxes on the sum of charges, we get:
```
(68_33 + 68_33 + 57_50 + 85_00) * 1.2 = 334_99,2
```
but because it is in cents, when issuing the invoice, the value is rounded to `334_99`

Next, when we create credit notes for this invoice, and we create a separate CN per each item:
```
CN01 = 68_33 * 1.2 = 8199.6 => 82_00
CN02 = 57_50 * 1.2 = 69_00
CN03 = 85_00 * 1.2 = 102_00
CN04 = 68_33 * 1.2 = 8199.6 = 82_00
```
Sum of these credit notes is `335_00` which is more than the invoice total, and we fail vthe validation, despite the fees amount is correct

## Description

To help this situation, we can operate adjustments that were made when credit notes are issued.
So when we're creating CN1-3, we simply create credit notes.
when we create CN4, we are subtracting  all rounding adjustments that were made for all credit notes for this invoice. in our case, we will need to subtract `0.4` cents from precise taxes amount, and then for the last credit note
`CN04 = 68_33 * 1.2 = 8199.6 -0.4 = 8199.2 => 8199`
and the values are matching

also removed validation on credit_note_item total amount, as it tries to predict taxes, but they are adjusted further in the credit note creation service

### DIFF
--- a/app/services/credit_notes/estimate_service.rb
+++ b/app/services/credit_notes/estimate_service.rb
@@ -23,14 +23,10 @@ module CreditNotes
         balance_amount_currency: invoice.currency
       )
 
-      # byebug
       validate_items
       return result unless result.success?
 
       compute_amounts_and_taxes
-      valid_credit_note?
-      # byebug
-      return result unless result.success?
 
       result.credit_note = credit_note
       result
--- a/app/services/credit_notes/estimate_service.rb
+++ b/app/services/credit_notes/estimate_service.rb
@@ -23,10 +23,14 @@ module CreditNotes
         balance_amount_currency: invoice.currency
       )
 
+      # byebug
       validate_items
       return result unless result.success?
 
       compute_amounts_and_taxes
+      valid_credit_note?
+      # byebug
+      return result unless result.success?
 
       result.credit_note = credit_note
       result
@@ -58,7 +62,12 @@ module CreditNotes
       end
     end
 
+    def valid_credit_note?
+      CreditNotes::ValidateService.new(result, item: credit_note).valid?
+    end
+
     def valid_item?(item)
+      # byebug
       CreditNotes::ValidateItemService.new(result, item:).valid?
     end
 
@@ -71,7 +80,9 @@ module CreditNotes
       credit_note.precise_coupons_adjustment_amount_cents = taxes_result.coupons_adjustment_amount_cents
       credit_note.coupons_adjustment_amount_cents = taxes_result.coupons_adjustment_amount_cents.round
       credit_note.precise_taxes_amount_cents = taxes_result.taxes_amount_cents
-      credit_note.taxes_amount_cents = taxes_result.taxes_amount_cents.round
+      # adjust_credit_note_tax_precise_rounding(taxes_result) if credit_note_for_all_remaining_amount?
+
+      credit_note.taxes_amount_cents = credit_note.precise_taxes_amount_cents.round
       credit_note.taxes_rate = taxes_result.taxes_rate
 
       taxes_result.applied_taxes.each { |applied_tax| credit_note.applied_taxes << applied_tax }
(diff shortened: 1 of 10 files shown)


app/services/credit_notes/estimate_service.rb
```diff
@@ -71,24 +75,39 @@ def compute_amounts_and_taxes
...
+    end
+
+    def all_rounding_tax_adjustments
+      credit_note.invoice.credit_notes.sum(&:taxes_rounding_adjustment)
+    end
+
     def compute_refundable_amount
       credit_note.refund_amount_cents = credit_note.credit_amount_cents
 
+      # invoice.refundable_amount_cents is incorrect - in our case it returns 8200, but it should be 8199...
```
ancorcruz, line 110: ?

spec/scenarios/credit_note_spec.rb
```diff
@@ -336,4 +336,547 @@
...
+            refund_amount_cents: 81_99,
+            items: [
+              {
+                fee_id: fees[1].id,
+                amount_cents: 68_33
+              }
+            ]
+          )
+
+          credit_note = invoice.credit_notes.order(:created_at).last
```
ancorcruz, line 438: I don't know a lot about credit notes but as I understand we create one credit note by invoice fee, then, the adjustment is done in the last one, what would happen if the last one is voided? I guess the sums still match but the voided credit not won't match the invoice fee... maybe I'm just mixing things.

ancorcruz (approved):
LGTM!

spec/services/credit_notes/validate_item_service_spec.rb
```diff
@@ -96,19 +96,20 @@
...
+    #
+    #   it 'fails the validation' do
+    #     aggregate_failures do
+    #       expect(validator).not_to be_valid
+    #
+    #       expect(result.error).to be_a(BaseService::ValidationFailure)
+    #       expect(result.error.messages[:amount_cents]).to eq(['higher_than_remaining_invoice_amount'])
+    #     end
+    #   end
+    # end
```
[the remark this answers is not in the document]
annvelents: 😱 have no idea how I forgot this 🤦‍♀️ 
I think I was so happy with the last test finally ready and working that I forgot to look through the PR 😅 
thank you, cleaning everything up!

app/services/credit_notes/validate_item_service.rb
```diff
@@ -7,7 +7,9 @@ def valid?
 
       valid_item_amount?
       valid_individual_amount?
-      valid_global_amount?
+      # we don't save taxes on the credit_note item and we'll adjust them when creating credit notes, so it makes sense
+      # to check taxes on credit note level, where taxes are calculated and stored.
+      # valid_global_amount?
```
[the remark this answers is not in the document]
annvelents: 💯

spec/scenarios/credit_note_spec.rb
annvelents: discussed it at slack, leaving here for the history:
I checked the algorithm behind voiding credit notes - we update their status and balance_amount, while the amounts do not change.
So if we void the last credit note, we won't be able to issue new credit_notes, but previous ones will keep their numbers, and that first CN for amount 81_99.6 will keep it's rounding to 82_00 , which is still mathematically correct.
the issue that we're solving here is the sum of all credit notes should not be bigger than invoice.total_amount, while technically on the level of each credit_note calculation even the current rounding is correct.

PR #2645, opened 2024-10-01
misc: Refactor Customer::UpdateService

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/2641 and keeps refactoring some old service to take advantage of the `call` pattern and to remove the `user` arguments used in `BaseService#initialize`. 

## Description

This PR:
- Extract the `Customers::UpdateService#update_currency` method into a dedicated `Customers::UpdateCurrencyService` and adds specs
- Turn the `Customers::UpdateService#update` method into a proper `Customers::UpdateService#call` one


app/graphql/mutations/customers/update.rb
```diff
@@ -21,7 +21,8 @@ class Update < BaseMutation
       type Types::Customers::Object
 
       def resolve(**args)
-        result = ::Customers::UpdateService.new(context[:current_user]).update(**args)
+        customer = context[:current_user].customers.find_by(id: args[:id])
+        result = ::Customers::UpdateService.call(customer:, args:)
```
nudded, line 25: Just a thought, but could we leverage `raise_if_error` here as well?

vincent-pochet: @nudded 

It would require a change in the way we treat the result:

```ruby
def resolve(**args)
  ::Customers::UpdateService.call(customer:, args:).raise_if_error!.customer
rescue BaseService::FailedResult
  result_error(result)
end
```

It's a personal opinion, but I don't feel it makes it easier to read.

nudded: oh seeing it written down like that I tend to agree @vincent-pochet 👍 (It would only make sense if we captured that error somewhere generically perhaps. But agreed to keep this PR small :)

nudded (approved):

PR #2646, opened 2024-10-01
Fix: closed invoices with generated number

annvelents:
**context** 
Before we were setting invoice status to `finalized` or `draft` before calculating fees (because before calculating fees we don't know if it will be 0-invoice or no, and as result we don't know if it will be `:finalized` or `:closed`)

but inside calculate_fees_service we save the invoice, so if it's in `finalized` status we would generate number for this invoice, which is not needed for `closed` invoices

as result we should set invoice to `finalized` or to `closed` after we calculated fees.

so the definition if the process is finalization or  drafting of the invoice should be done not only by status, but also by the context.

 **what was done** 

- instead of setting invoice status before calculating fees, set up context
- define if the process is finalization for applying credit notes the same way we define it for tax reporting

### DIFF
--- a/app/services/invoices/calculate_fees_service.rb
+++ b/app/services/invoices/calculate_fees_service.rb
@@ -324,7 +324,6 @@ module Invoices
 
     def not_in_finalizing_process?
       !finalizing_invoice?
-      # (invoice.draft? || invoice.voided?) && context != :finalize
     end
 
     def in_trial_period_not_ending_today?(subscription, timestamp)
--- a/app/services/invoices/calculate_fees_service.rb
+++ b/app/services/invoices/calculate_fees_service.rb
@@ -361,7 +361,7 @@ module Invoices
     end
 
     def finalizing_invoice?
-      context == :finalize || Invoice::COMPLETE_INVOICE_STATUSES.include?(invoice.status)
+      context == :finalize || Invoice::GENERATED_INVOICE_STATUSES.include?(invoice.status)
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/services/invoices/subscription_service.rb
```diff
@@ -27,14 +27,14 @@ def call
       result.invoice = invoice
 
       fee_result = ActiveRecord::Base.transaction do
-        invoice.status = invoice_status
+        context = grace_period? ? :draft : :finalize
         fee_result = Invoices::CalculateFeesService.call(
```
annvelents, line 31: before calling this service we cannot say if it will be `closed` or `finalized` invoice, as we need the calculated number to decide
inside this service we save invoice. If invoice is in status `finalized`, we would assign number to this invoice, and updating it then to `closed` will not "delete" invoice number

app/services/invoices/calculate_fees_service.rb
```diff
@@ -323,7 +323,8 @@ def refresh_amounts(credit_amount_cents:)
     end
 
     def not_in_finalizing_process?
-      (invoice.draft? || invoice.voided?) && context != :finalize
+      !finalizing_invoice?
+      # (invoice.draft? || invoice.voided?) && context != :finalize
```
annvelents, line 327: by the previous logic if we send 
- an invoice that is not in draft status and 
- context != :finalize,
we'd consider it as `is_in_finalizing_process`, because first condition returns false, and operator `&&` would return `false`

rsempe (approved):

brunomiguelpinto (approved):

PR #2650, opened 2024-10-02
misc(filters-contracts): Define contracts to validate query objects filters

ancorcruz:
## Context

Currently query objects receive a filters hash, the filters are not validated and used to build the SQL query agains Lago DB.

Also, developers need to read the query object implementation just to be aware of the available filters.

## Description

This is a PoC introduces contracts to validate query objects filters, validating the filter provides a following benefits;

1. Ensure each query object filters are correctly validated before executing the query.

2. Decouple filter validation logic form the query objects.

3. Makes explicit to developers which filters are available without having to dig into the query object implementation details.

`dry-validation` is added as dependency.

Bear in mind, this first contract is meant as an example an covers a simple scenario, there are other query objects with many and more complex filters available.


app/queries/base_query.rb
```diff
@@ -22,6 +22,19 @@ def initialize(organization:, pagination: DEFAULT_PAGINATION_PARAMS, filters: {}
 
   attr_reader :organization, :pagination_params, :filters, :search_term, :order
 
+  def validate_filters
+    contract_class_name = "#{self.class.name}FiltersContract"
+    contract = Object.const_get(contract_class_name).new
```
vincent-pochet, line 27: Just a thought, do you think we could find a way to pass or assign the Contract class to the query class, to avoid doing some magic to load it?

vincent-pochet (approved):
Really cool! I like how it fit with the `Result` pattern :+1:

PR #2664, opened 2024-10-08
fix (wallet-race-condition): Retry wallet update upon race condition

lovrocolic:
## Context

Recently, feature for updating wallet balance upon successful payment was introduced.

## Description

If multiple paid credits are issues, there is a chance that after successful payment processing we will receive all events from payment provider in the same request.

Multiple events will schedule multiple jobs for increasing wallet balance and each job is updating the same resource so there is a high chance of race condition.

This PR introduced optimistic locking approach with 1 additional retry if multiple processes are updating the same resource.

### DIFF
--- a/app/services/wallets/balance/increase_service.rb
+++ b/app/services/wallets/balance/increase_service.rb
@@ -3,31 +3,46 @@
 module Wallets
   module Balance
     class IncreaseService < BaseService
+      MAX_RETRIES = 1
+
       def initialize(wallet:, credits_amount:, reset_consumed_credits: false)
         super
 
         @wallet = wallet
         @credits_amount = credits_amount
         @reset_consumed_credits = reset_consumed_credits
+        @retries = 0
       end
 
       def call
-        currency = wallet.balance.currency
-        amount_cents = wallet.rate_amount * credits_amount * currency.subunit_to_unit
+        ActiveRecord::Base.transaction do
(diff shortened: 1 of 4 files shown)


jdenquin (approved):

nudded (approved):

app/services/wallets/balance/increase_service.rb
```diff
@@ -3,34 +3,49 @@
...
+          Wallets::Balance::RefreshOngoingService.call(wallet:)
         end
 
-        wallet.update!(update_params)
-        Wallets::Balance::RefreshOngoingService.call(wallet:)
-
         result.wallet = wallet
         result
+      rescue ActiveRecord::StaleObjectError
+        @retries += 1
```
nudded, line 40: should we maybe retry more than once?

lovrocolic: after extensive tests, also realised 1 is not enough:)

app/services/wallets/balance/increase_service.rb
```diff
@@ -3,34 +3,51 @@
...
+
+        if @retries <= MAX_RETRIES
+          sleep(0.5)
```
lovrocolic, line 43: @jdenquin @nudded Without it it didn't work for more that cca 10 subsequent processes

PR #2667, opened 2024-10-09
feat(hubspot): Add associations, attributes to customer models to support hubspot integration

ivannovosad:
## Roadmap Task

## Description

This PR adds models, associations and attributes needed for hubspot customers.
It also adds logic to the custoemr serializer so that it can return correct customer type.

### DIFF
--- a/spec/models/integration_customers/hubspot_customer_spec.rb
+++ b/spec/models/integration_customers/hubspot_customer_spec.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe IntegrationCustomers::HubspotCustomer, type: :model do
+  subject(:hubspot_customer) { build(:hubspot_customer) }
+
+  describe '#targeted_object' do
+    let(:targeted_object) { Integrations::HubspotIntegration::TARGETED_OBJECTS.sample }
+
+    it 'assigns and retrieve a setting' do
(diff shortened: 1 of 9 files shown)


vincent-pochet (approved):

spec/models/integration_customers/hubspot_customer_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+require 'rails_helper'
+
+RSpec.describe IntegrationCustomers::HubspotCustomer, type: :model do
+  subject(:hubspot_customer) { build(:hubspot_customer) }
+
+  describe '#targeted_object' do
+    let(:targeted_object) { Integrations::HubspotIntegration::TARGETED_OBJECTS.sample }
+
+    it 'assigns and retrieve a setting' do
```
vincent-pochet, line 11: No blocker here, just an improvement suggestion, I think we could use a shared example to test all these assessors

ivannovosad: In this particular case only Hubspot customers have `targeted_object` and `email` settings.

lovrocolic (approved):

spec/models/integration_customers/hubspot_customer_spec.rb
vincent-pochet: @ivannovosad my point was about settings_accessors in general. ;)

PR #2669, opened 2024-10-10
fix (wallets): improve handling rc upon wallet update

lovrocolic:
## Context

This PR improves how we handle race condition scenario upon wallet update

### DIFF
--- a/app/jobs/invoices/prepaid_credit_job.rb
+++ b/app/jobs/invoices/prepaid_credit_job.rb
@@ -4,6 +4,7 @@ module Invoices
   class PrepaidCreditJob < ApplicationJob
     queue_as 'wallets'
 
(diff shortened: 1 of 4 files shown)


ancorcruz (approved):

app/jobs/invoices/prepaid_credit_job.rb
```diff
@@ -4,6 +4,7 @@ module Invoices
   class PrepaidCreditJob < ApplicationJob
     queue_as 'wallets'
 
+    retry_on ActiveRecord::StaleObjectError, wait: :polynomially_longer, attempts: 6
```
ancorcruz, line 7: would you consider adding a test case to ensure the jobs are retried when the job fails with this kind of error?

PR #2674, opened 2024-10-11
Feat hubspot customers services

ivannovosad:
## Roadmap Task

## Context

When creating or updating customers in Lago, we must sync them to Hubspot.
Depending on if Lago customer is either company or individual, we need to call different services (Nango endpoints).

## Description

This PR implements services that handle syncing the customers either as companies or contacts to Hubspot via Nango.


app/services/integration_customers/factory.rb
```diff
@@ -2,8 +2,8 @@
 
 module IntegrationCustomers
   class Factory
-    def self.new_instance(integration:, customer:, subsidiary_id:)
-      service_class(integration).new(integration:, customer:, subsidiary_id:)
+    def self.new_instance(integration:, customer:, subsidiary_id:, **params)
```
lovrocolic, line 5: Probably we could get rid of `subsidiary_id` since it is just used for netsuite. But maybe we can do it in another PR.
I think it should not be part of constructor for xero or Anrok and it should be part of params and used only for netsuite...

app/services/integration_customers/hubspot_service.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def create
+      create_result = create_service_class.call(
+        integration:,
+        customer:,
+        subsidiary_id: nil
+      )
+
+      return create_result if create_result.error
```
lovrocolic, line 21: I would rather use `unless create_result.success?` since we have this bool variable

app/services/integration_customers/create_service.rb
```diff
@@ -49,6 +51,12 @@ def link_customer!
 
       if integration&.type&.to_s == 'Integrations::NetsuiteIntegration'
         new_integration_customer.subsidiary_id = subsidiary_id
+        new_integration_customer.save!
```
lovrocolic, line 54: before we didn't save `subsidiary_id`?

app/services/integration_customers/update_service.rb
```diff
@@ -14,15 +14,13 @@ def call
...
-      integration_customer.update!(external_customer_id:) if external_customer_id.present?
+      integration_customer.external_customer_id = external_customer_id if external_customer_id.present?
+      integration_customer.targeted_object = targeted_object if targeted_object.present?
+      integration_customer.save!
 
-      if sync_with_provider
-        integration_customer.subsidiary_id = subsidiary_id if subsidiary_id.present?
-
-        update_result = Integrations::Aggregator::Contacts::UpdateService.call(integration:, integration_customer:)
+      if integration_customer.external_customer_id.present?
```
lovrocolic, line 21: Why don't we store subsidiary_id anymore?

app/services/integration_customers/hubspot_service.rb
ivannovosad: It is implemented exactly like this in all the other intergations. So we should change it there as well.

app/services/integration_customers/factory.rb
ivannovosad: Yes, but I didn't want to do the refactorgin on thsi branch as that would affect more classes than necessary for this Hubspot feature.

app/services/integration_customers/create_service.rb
ivannovosad: It seems we were saving only when syncing the customer with the sync checkbox checked. When linking to an existing Netsuite customer we were not saving the subsidiary.

app/services/integration_customers/hubspot_service.rb
ivannovosad: Whay do you think? To leave is as is now, and refactor all the services in a separate PR or do it now? I'd choose another PR.

app/services/integration_customers/update_service.rb
ivannovosad: Because we don't need to in case of updating a customer. If the customer was succesfully created the subsidiary was saved already.

lovrocolic (approved):

PR #2678, opened 2024-10-11
misc(usage): Scope usage caching to the charge filter level

vincent-pochet:
## Context

Today caching for usage data is scope to the charge.
When an event is received for a specific billable metric code will invalidate the cache for all charges attached to it for the customer's plan. The cache will leave until the end of the billing period or until a new event is received for the same billable metric code and subscription.
This logic avoid recomputing the complete charge aggregation every time the customer usage is requested.

This logic was fine until the introduction of charge filters.
Aggregation for filters are done one by one for every filter, and stored in cache using the same charge cache key (ie: all filtered aggregations are grouped inside the same charge cached bucket). When an event matching a filter is received it invalidate the cache for all filters, meaning that when fetching the usage we have to resegregate it for every single filter even if nothing new was received for the majority of them.

## Description

This PR refactors some parts of the usage computation and adds a new caching logic to scope the caching up to the charge filter when applicable and keeping the actual caching when no filters are attached to the charge.

### DIFF
--- a/app/services/events/post_process_service.rb
+++ b/app/services/events/post_process_service.rb
@@ -84,10 +84,13 @@ module Events
       charges = billable_metric.charges
         .joins(:plan)
         .where(plans: {id: active_subscription.map(&:plan_id)})
+        .includes(filters: {values: :billable_metric_filter})
 
       charges.each do |charge|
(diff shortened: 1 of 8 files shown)


rsempe (approved):
🔥

PR #2690, opened 2024-10-14
feat(stripe): Improve support of Indian card

jdenquin:
- Add new logic for payment from Indian card to support 3DS auth
- Send a new webhook when a payment requires a 3DS auth
- All Indian payments are done with `off_session: false`
- Add a new `provider_payment_data` to support extra informations from providers about the payment


config/environments/development.rb
```diff
@@ -57,6 +57,7 @@
   config.hosts << 'api.lago.dev'
   config.hosts << 'api'
   config.hosts << 'lago.ngrok.dev'
+  config.hosts << 'flying-enhanced-puma.ngrok-free.app'
```
nudded, line 60: can be removed?

nudded (approved):

PR #2692, opened 2024-10-15
feat(dunning): Add DunningCampaignsResolver

rsempe:
This pull request introduces a new feature to manage and query dunning campaigns within an organization. The most important changes include adding new permissions, creating a resolver and query for dunning campaigns, and updating the GraphQL schema to support these queries.

### GraphQL Resolvers and Types:

* [`app/graphql/resolvers/dunning_campaigns_resolver.rb`](diffhunk://#diff-d45ebb2f5c75cc98d69a0d641723cff50a2a7034d18de7561d16ff45c71680c3R1-R38): Created a new resolver for querying dunning campaigns with appropriate permissions and arguments.
* [`app/graphql/types/query_type.rb`](diffhunk://#diff-f091357fe3bc9b42153dde4873030274dd9aa16c9cf86f689fee229ca9558b8cR35): Added `dunning_campaigns` field to the `QueryType` to use the new resolver.

### Model and Query Enhancements:

* [`app/models/dunning_campaign.rb`](diffhunk://#diff-3b7eb2d6a06a5b47d979ac5e42118e603975d556bac67a32d1372160bd29c7cdR6-R7): Added `ORDERS` constant and `ransackable_attributes` method for sorting and searching. [[1]](diffhunk://#diff-3b7eb2d6a06a5b47d979ac5e42118e603975d556bac67a32d1372160bd29c7cdR6-R7) [[2]](diffhunk://#diff-3b7eb2d6a06a5b47d979ac5e42118e603975d556bac67a32d1372160bd29c7cdR16-R19)
* [`app/queries/dunning_campaigns_query.rb`](diffhunk://#diff-0d6920857152058222fe3b254236d3b9f001691a8101247633146b92cbabec9cR1-R40): Created a new query class to handle filtering, sorting, and pagination of dunning campaigns.

### DIFF
--- a/spec/queries/dunning_campaigns_query_spec.rb
+++ b/spec/queries/dunning_campaigns_query_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe DunningCampaignsQuery, type: :query do
     create(:dunning_campaign, organization:, name: "defgh", code: "11", applied_to_organization: true)
   end
   let(:dunning_campaign_second) do
-    create(:dunning_campaign, organization:, name: "abcde", code: "22", applied_to_organization: true)
+    create(:dunning_campaign, organization:, name: "abcde", code: "22", applied_to_organization: false)
   end
 
   let(:dunning_campaign_third) do
@@ -45,7 +45,7 @@ RSpec.describe DunningCampaignsQuery, type: :query do
   context "with pagination" do
     let(:pagination) { {page: 2, limit: 2} }
 
-    it "applies the pagination" do
+    it "applies the pagination", :aggregate_failures do
       aggregate_failures do
         expect(result).to be_success
         expect(result.dunning_campaigns.count).to eq(1)
@@ -66,11 +66,19 @@ RSpec.describe DunningCampaignsQuery, type: :query do
     end
   end
 
-  context "with a filter on applied by default" do
+  context "with applied_to_organization is false" do
     let(:filters) { {applied_to_organization: false} }
 
-    it "returns only one dunning campaign" do
-      expect(result.dunning_campaigns).to eq([dunning_campaign_third])
+    it "returns second and third campaigns" do
+      expect(result.dunning_campaigns).to eq([dunning_campaign_second, dunning_campaign_third])
+    end
+  end
+
+  context "with applied_to_organization is true" do
+    let(:filters) { {applied_to_organization: true} }
+
+    it "returns only the first dunning campaign" do
+      expect(result.dunning_campaigns).to eq([dunning_campaign_first])
     end
   end
 
--- a/spec/queries/dunning_campaigns_query_spec.rb
+++ b/spec/queries/dunning_campaigns_query_spec.rb
@@ -1,0 +1,86 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe DunningCampaignsQuery, type: :query do
+  subject(:result) do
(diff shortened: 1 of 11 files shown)


spec/queries/dunning_campaigns_query_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    it "applies the pagination" do
+      aggregate_failures do
+        expect(result).to be_success
+        expect(result.dunning_campaigns.count).to eq(1)
+        expect(result.dunning_campaigns.current_page).to eq(2)
+        expect(result.dunning_campaigns.prev_page).to eq(1)
+        expect(result.dunning_campaigns.next_page).to be_nil
+        expect(result.dunning_campaigns.total_pages).to eq(2)
+        expect(result.dunning_campaigns.total_count).to eq(3)
+      end
```
ancorcruz, line 57:
```suggestion
    it "applies the pagination", : aggregate_failures do
      expect(result).to be_success
      expect(result.dunning_campaigns.count).to eq(1)
      expect(result.dunning_campaigns.current_page).to eq(2)
      expect(result.dunning_campaigns.prev_page).to eq(1)
      expect(result.dunning_campaigns.next_page).to be_nil
      expect(result.dunning_campaigns.total_pages).to eq(2)
      expect(result.dunning_campaigns.total_count).to eq(3)
```

spec/queries/dunning_campaigns_query_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  let(:search_term) { nil }
+  let(:filters) { nil }
+  let(:order) { nil }
+  let(:membership) { create(:membership) }
+  let(:organization) { membership.organization }
+  let(:dunning_campaign_first) do
+    create(:dunning_campaign, organization:, name: "defgh", code: "11", applied_to_organization: true)
+  end
+  let(:dunning_campaign_second) do
+    create(:dunning_campaign, organization:, name: "abcde", code: "22", applied_to_organization: true)
```
ancorcruz, line 20: I have seen there are two campaigns set as applied to organization... this should not happen, afaik, we could be one and only one campaign applied to the organization.

We should add a unique index on dunning_campaigns table with organization_id and applied_to_organzation: true to avoid this to happen. What do you think?

spec/queries/dunning_campaigns_query_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  context "when searching for /de/ term" do
+    let(:search_term) { "de" }
+
+    it "returns only two dunning campaigns" do
+      expect(result.dunning_campaigns).to eq([dunning_campaign_second, dunning_campaign_first])
+    end
+  end
+
+  context "with a filter on applied by default" do
```
ancorcruz, line 69: could we also test the scenario of filter `applied_to_organization: true`?

ancorcruz (approved):
looks good, left a few comments.

spec/queries/dunning_campaigns_query_spec.rb
rsempe: Well, yep it's not possible right now - I haven't created the index because I have the feeling that we could apply multiple campaigns in the future.

But maybe the pragmatic approach is to add the unique index now and remove it when/if needed in the future.

rsempe: Yep - Let's do this 👍

PR #2698, opened 2024-10-16
feat(dunning): Add dunning campaign to customer, database changes

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

The goal of this change is to create set the relationship between customers and dunning campaigns.

A customer could have one dunning campaign applied, or, inherit its organization default campaign (none applied) or, be excluded from dunning at all (with a boolean flag).

Also, when deleting a dunning campaign, all customers with the campaign assigned must be reset.

### DIFF
--- a/app/models/dunning_campaign.rb
+++ b/app/models/dunning_campaign.rb
@@ -6,7 +6,10 @@ class DunningCampaign < ApplicationRecord
   ORDERS = %w[name code].freeze
 
   belongs_to :organization
+
   has_many :thresholds, class_name: "DunningCampaignThreshold", dependent: :destroy
+  has_many :customers, foreign_key: :applied_dunning_campaign_id, dependent: :nullify
+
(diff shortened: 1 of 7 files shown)


app/models/dunning_campaign.rb
```diff
@@ -4,7 +4,10 @@ class DunningCampaign < ApplicationRecord
   include PaperTrailTraceable
 
   belongs_to :organization
+
   has_many :thresholds, class_name: "DunningCampaignThreshold", dependent: :destroy
+  has_many :customers, foreign_key: :applied_dunning_campaign_id, dependent: :nullify
```
ancorcruz, line 9: I'm not 100% sure about setting nullify as dependent strategy here... it makes sense but we also need to reset the customer dunning campaign attempts counter to zero and nullify the last attempt run at timestamp....

rsempe (approved):

PR #2699, opened 2024-10-16
feat(clickhouse): new column for clickhouse events store on org

jdenquin:
- add a new field `clickhouse_events_store`  on `organizations`
- `clickhouse_aggregation` will be dropped later

### DIFF
--- a/app/services/invoices/customer_usage_service.rb
+++ b/app/services/invoices/customer_usage_service.rb
@@ -86,7 +86,7 @@ module Invoices
     end
 
     def charge_usage(charge)
-      return charge_usage_without_cache(charge) if organization.clickhouse_aggregation?
+      return charge_usage_without_cache(charge) if organization.clickhouse_events_store?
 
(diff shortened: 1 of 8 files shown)


vincent-pochet (approved):

app/services/invoices/customer_usage_service.rb
```diff
@@ -86,7 +86,7 @@ def add_charge_fees
     end
 
     def charge_usage(charge)
-      return charge_usage_without_cache(charge) if organization.clickhouse_aggregation?
```
nudded, line 89: Will we support all aggregations in CH? I was thinking it was also good to check if the billable metric supports CH?

jdenquin: all `billable_metrics`  should support Clickhouse yes, it's difficult to have an hybrid event store (between PG and Clickhouse) at the same time, mainly due to ingestion

nudded (approved):

PR #2701, opened 2024-10-17
feat(cascade-plan-updates): cascade subscription fee update

lovrocolic:
## Context

Currently when there is update on parent plan, changes are not cascaded to children.

## Description

This PR handles cascade of subscription fee update to all related children. Cascade should happen only if child plan has not been already updated.

### DIFF
--- a/app/jobs/plans/update_subscription_fee_job.rb
+++ b/app/jobs/plans/update_subscription_fee_job.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Plans
+  class UpdateSubscriptionFeeJob < ApplicationJob
+    queue_as 'default'
+
+    def perform(plan:, amount_cents:)
+      Plans::UpdateSubscriptionFeeService.call(plan:, amount_cents:)
+    end
+  end
+end
--- a/app/services/plans/update_service.rb
+++ b/app/services/plans/update_service.rb
@@ -91,16 +91,16 @@ module Plans
 
     def cascade_subscription_fee_update(old_amount_cents)
       return unless cascade?
-      return if plan.children.empty?
       return if old_amount_cents == plan.amount_cents
+      return if plan.children.empty?
 
       plan.children.where(amount_cents: old_amount_cents).find_each do |p|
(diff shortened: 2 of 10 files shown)


app/jobs/plans/update_subscription_fee_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Plans
+  class UpdateSubscriptionFeeJob < ApplicationJob
```
vincent-pochet, line 4: Since `Fee` refers to another resource, what to do you thing of naming this job `UpdateAmoutJob`?

app/services/plans/update_service.rb
```diff
@@ -87,6 +89,20 @@ def bill_charges_monthly?
       params[:bill_charges_monthly] || false
     end
 
+    def cascade_subscription_fee_update(old_amount_cents)
+      return unless cascade?
+      return if plan.children.empty?
+      return if old_amount_cents == plan.amount_cents
```
vincent-pochet, line 95: Maybe we could move this check as the first one in the method as it would avoid calling the db for the `plan.children.empty?` if no update is required

app/services/plans/update_service.rb
```diff
@@ -87,6 +89,20 @@ def bill_charges_monthly?
...
+
+      plan.children.where(amount_cents: old_amount_cents).find_each do |p|
+        Plans::UpdateSubscriptionFeeJob.perform_later(plan: p, amount_cents: plan.amount_cents)
+      end
+    end
+
+    def cascade?
+      params[:cascade_updates].present? && params[:cascade_updates].to_s == 'true'
```
vincent-pochet, line 103:
```suggestion
      ActiveModel::Type::Boolean.new.cast(params[:cascade_updates])
```

lovrocolic: Thanks!

app/services/plans/update_service.rb
lovrocolic: Make sense, moved up!

vincent-pochet (approved):

PR #2707, opened 2024-10-17
fix(customer): Fix create/update via API when finalize_zero_amount_invoice is nil

vincent-pochet:
## Context

Creation of customer via the REST API is failing if a nil value is provided for the `finalize_zero_amount_invoice` field (when called via the Ruby client for example)

The raised error is a `ActiveRecord::NotNullViolation`

## Description

This PR defaults to the default value (`inherit`) when it's the case

### DIFF
--- a/spec/services/customers/create_service_spec.rb
+++ b/spec/services/customers/create_service_spec.rb
@@ -246,6 +246,29 @@ RSpec.describe Customers::CreateService, type: :service do
         end
       end
 
+      context 'with nil value for finalize_zero_amount_invoice' do
+        let(:create_args) do
+          {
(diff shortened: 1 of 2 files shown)


nudded (approved):

spec/services/customers/create_service_spec.rb
```diff
@@ -246,6 +246,29 @@
         end
       end
 
+      context 'with nill value for finalize_zero_amount_invoice' do
```
rsempe, line 249: `null` or `nil` 😅

rsempe (approved):

PR #2714, opened 2024-10-18
fix(adyen): Improve webhook handling with missing code

vincent-pochet:
## Context

This PR is related to the `BaseService::ServiceFailure - payment_provider_code_missing: Code is missing` raised by the `WebhooksController#adyen` endpoint.

The API is raising a lot of these error that are related to a wrong configuration of the Adyen webhook URL.
When an organization has multiple Adyen account configured, the webhook call must contain a code parameter. When this value is missing the call just end with an `HTTP 500` error.

## Description

Since nothing can be do to remediate the situation on Lago's side and it is generating a lot of noise, this PR handle this kind of error to keep notifying Adyen that the webhook was not succesfully handled (Bad Request response) but is avoiding the `HTTP 500` error and the related Sentry issue


annvelents (approved):
LGTM, I would only maybe add a test on this change of the error code when handling payment_provider_failure from `payment_provider_code_missing` to `webhook_error`

PR #2717, opened 2024-10-21
feat(dunning): Update dunning campaign config for customers

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

Add auto dunning campaign fields to customers object type.

Update customers auto dunning campaign config through customers update mutation.

`exclude_from_dunning_campaign` has higher priority than the applied campaign and if both config arguments are provided at update, exclusion true overwrites the applied campaign.


annvelents (approved):
🏅

PR #2718, opened 2024-10-21
feat(daily_usage): Add DailyUsage model

vincent-pochet:
## Context

This PR is part of the Usage Revenue and unit.

Today, Lago does not offer a way to retrieve customer usage with a granularity lower than the billing period (via invoices and fees). On the other end, it is possible to get the current usage for a customer/subscription, but this usage is just a “snapshot” of the usage a the current time. 

## Description

This PR is the first part of a feature implementing a daily usage monitoring. Is simply introduce a new `DailyUsage` model to store the computed daily usage.


annvelents (approved):
🚀

PR #2725, opened 2024-10-22
feat(cascade-plan-charges): add logic for cascading charge update

lovrocolic:
## Context

## Description

This PR handles the case for cascading charge update case


annvelents (approved):
LGTM, I would only suggest to add a test with `cascade: true` to `Charges::UpdateService`

lovrocolic:
> LGTM, I would only suggest to add a test with `cascade: true` to `Charges::UpdateService`

Added this test case

PR #2726, opened 2024-10-22
feat(cascade-plan-changes): add api and GQL support for cascade feature

lovrocolic:
## Context

## Description

This PR adds `cascade_updates` attribute that allows user to skip cascade action.

Also, this PR includes `has_overridden_plans` attribute on GQL plan object.

### DIFF
--- a/app/controllers/api/v1/plans_controller.rb
+++ b/app/controllers/api/v1/plans_controller.rb
@@ -17,7 +17,13 @@ module Api
 
       def update
         plan = current_organization.plans.parents.find_by(code: params[:code])
-        result = ::Plans::UpdateService.call(plan:, params: input_params.to_h.deep_symbolize_keys)
+
+        update_params = input_params.to_h.deep_symbolize_keys
+        if params[:plan].key?(:cascade_updates)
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/plans_controller.rb
```diff
@@ -17,7 +17,13 @@ def create
 
       def update
         plan = current_organization.plans.parents.find_by(code: params[:code])
-        result = ::Plans::UpdateService.call(plan:, params: input_params.to_h.deep_symbolize_keys)
+
+        update_params = input_params.to_h.deep_symbolize_keys
+        if params[:plan].key?(:cascade_updates)
+          update_params[:cascade_updates] = params.require(:plan).permit(:cascade_updates)[:cascade_updates]
+        end
```
vincent-pochet, line 24: What's the reason for handling this at this level rather than in the `input_params`?

lovrocolic: This param is needed only for `update` action. Another option was to have some logic in `input_params` method...

annvelents (approved):

PR #2733, opened 2024-10-23
feat(dunning): update customer to fallback to organization dunning

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

Customer update service accepts `applied_dunning_campaign_id` as nil to unset its applied campaign and fallback to organization's default dunning campaign.

### DIFF
--- a/app/services/customers/update_service.rb
+++ b/app/services/customers/update_service.rb
@@ -96,7 +96,7 @@ module Customers
 
       if customer.organization.auto_dunning_enabled?
         if args.key?(:applied_dunning_campaign_id)
-          dunning_campaign = DunningCampaign.find(args[:applied_dunning_campaign_id])
+          dunning_campaign = DunningCampaign.find_by(id: args[:applied_dunning_campaign_id])
           customer.applied_dunning_campaign = dunning_campaign
           customer.exclude_from_dunning_campaign = false
         end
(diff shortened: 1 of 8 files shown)


app/services/customers/update_service.rb
```diff
@@ -96,7 +96,7 @@ def call
 
       if customer.organization.auto_dunning_enabled?
         if args.key?(:applied_dunning_campaign_id)
-          dunning_campaign = DunningCampaign.find(args[:applied_dunning_campaign_id])
+          dunning_campaign = DunningCampaign.find_by(id: args[:applied_dunning_campaign_id])
```
annvelents, line 99: do we need to change from `find` to `find_by` because `args[:applied_dunning_campaign_id]` can be nil?
(the if checks only if the key exist, so I guess we can send `args = {applied_dunning_campaign_id: nil}`)
but in this case, if key present, but the value is nil (or something different, but we can't find dunning campaign by the provided id, so we will silently not find dunning_campaign and set dunning_campaign to nil), do we still want to update:
`customer.exclude_from_dunning_campaign = false`?

annvelents: in this case: is it possible that the organization won't have `applied_dunning_campaign`?

vincent-pochet (approved):

app/services/customers/update_service.rb
ancorcruz: Thanks for challenging this piece of code.

I'm going to update it as it will set it as nil if the campaign is not found... but the expected result was to accept nil to unset the campaign.

PR #2734, opened 2024-10-23
feat(api key management) Extract api key model

floganz:
## Context

One tiny step towards advanced API key management.

## Description

Introduces new model `ApiKey` that will replace `api_key` column on the `Organization` model during next steps.
Migrate existing values for `api_key` to the dedicated model.
Adjust REST API authorization to use lookup for ApiKey model and temporary keep previous lookup for Organization model.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -27,8 +27,9 @@ module Api
       true
     end
 
-    def current_organization(api_key = nil)
-      @current_organization ||= Organization.find_by(api_key:)
+    def current_organization(value = nil)
+      @current_organization ||=
+        ApiKey.find_by(value:)&.organization || Organization.find_by(api_key: value)
     end
 
     def set_context_source
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -27,9 +27,8 @@ module Api
       true
     end
 
-    def current_organization(value = nil)
-      @current_organization ||=
-        ApiKey.find_by(value:)&.organization || Organization.find_by(api_key: value)
+    def current_organization(api_key_value = nil)
+      @current_organization ||= ApiKey.find_by(value: api_key_value)&.organization
     end
 
     def set_context_source
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+class ApiKey < ApplicationRecord
+  include PaperTrailTraceable
+
+  belongs_to :organization
+
+  before_create :set_value
+
+  def generate_value
+    value = SecureRandom.uuid
+    api_key = ApiKey.find_by(value:)
+
+    return generate_value if api_key.present?
+
+    value
+  end
+
+  private
+
+  def set_value
+    self.value = generate_value
+  end
+end
+
+# == Schema Information
+#
+# Table name: api_keys
+#
+#  id              :uuid             not null, primary key
+#  value           :string           not null
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  organization_id :uuid             not null
+#
+# Indexes
+#
+#  index_api_keys_on_organization_id  (organization_id)
+#  index_api_keys_on_value            (value) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (organization_id => organizations.id)
+#
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -7,19 +7,16 @@ class ApiKey < ApplicationRecord
 
   before_create :set_value
 
-  def generate_value
-    value = SecureRandom.uuid
-    api_key = ApiKey.find_by(value:)
-
-    return generate_value if api_key.present?
-
-    value
-  end
+  validates :value, uniqueness: true
+  validates :value, presence: true, on: :update
 
   private
 
   def set_value
-    self.value = generate_value
+    loop do
+      self.value = SecureRandom.uuid
+      break unless self.class.exists?(value:)
+    end
   end
 end
 
--- a/app/services/organizations/create_service.rb
+++ b/app/services/organizations/create_service.rb
@@ -8,7 +8,9 @@ module Organizations
     end
 
     def call
-      organization = Organization.new(params)
+      organization = Organization.new(
+        params.slice(:name, :document_numbering)
+      )
 
       ActiveRecord::Base.transaction do
         organization.save!
--- a/app/services/organizations/create_service.rb
+++ b/app/services/organizations/create_service.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Organizations
+  class CreateService < BaseService
+    def initialize(params)
+      @params = params
+      super
+    end
+
+    def call
+      organization = Organization.new(params)
+
+      ActiveRecord::Base.transaction do
+        organization.save!
+        organization.api_keys.create!
+      end
+
+      result.organization = organization
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    end
+
+    private
+
+    attr_reader :params
+  end
+end
--- a/app/services/users_service.rb
+++ b/app/services/users_service.rb
@@ -29,7 +29,10 @@ class UsersService < BaseService
 
     ActiveRecord::Base.transaction do
       result.user = User.create!(email:, password:)
-      result.organization = Organization.create!(name: organization_name, document_numbering: 'per_organization')
+
+      result.organization = Organizations::CreateService
+        .call(name: organization_name, document_numbering: 'per_organization')
+        .organization
 
       result.membership = Membership.create!(
         user: result.user,
--- a/app/services/users_service.rb
+++ b/app/services/users_service.rb
@@ -32,6 +32,7 @@ class UsersService < BaseService
 
       result.organization = Organizations::CreateService
         .call(name: organization_name, document_numbering: 'per_organization')
+        .raise_if_error!
         .organization
 
       result.membership = Membership.create!(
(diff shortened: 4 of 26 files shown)


app/controllers/api/base_controller.rb
```diff
@@ -27,8 +27,9 @@ def authenticate
       true
     end
 
-    def current_organization(api_key = nil)
-      @current_organization ||= Organization.find_by(api_key:)
+    def current_organization(value = nil)
+      @current_organization ||=
+        ApiKey.find_by(value:)&.organization || Organization.find_by(api_key: value)
```
ancorcruz, line 32:
```suggestion
    def current_organization(api_key = nil)
      @current_organization ||=
        ApiKey.find_by(value: api_key)&.organization || Organization.find_by(api_key:)
```

I would keep `api_key` as name of the signature param as it express what it is at first glance without reading the implementation.

app/models/api_key.rb
```diff
@@ -0,0 +1,43 @@
...
+    return generate_value if api_key.present?
+
+    value
+  end
+
+  private
+
+  def set_value
+    self.value = generate_value
+  end
```
ancorcruz, line 23: I would avoid recursion in Ruby when possible and instead of `find_by` use `exists?` for better performance.

also the method could be a private one, is there any reason to expose it a public api of the model?

```suggestion
  before_create :set_key_value
  
  private

  def set_key_value
    begin
      self.value = SecureRandom.uuid 
    end while self.class.exists?(value: self.value)
  end
```

This should do the trick.

app/services/organizations/create_service.rb
```diff
@@ -0,0 +1,28 @@
...
+
+module Organizations
+  class CreateService < BaseService
+    def initialize(params)
+      @params = params
+      super
+    end
+
+    def call
+      organization = Organization.new(params)
```
ancorcruz, line 11: would you avoid mass assignment here and list each attribute assigned to the organization like in the original code?

db/migrate/20241021140054_create_api_keys.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class CreateApiKeys < ActiveRecord::Migration[7.1]
+  def up
+    create_table :api_keys, id: :uuid do |t|
+      t.references :organization, type: :uuid, null: false, foreign_key: true
+
+      t.string :value, null: false
```
ancorcruz, line 8: would you also add a unique index scoped to the organization_id?

db/migrate/20241021140054_create_api_keys.rb
```diff
@@ -0,0 +1,41 @@
...
+      t.timestamps
+    end
+
+    safety_assured do
+      execute <<-SQL
+      INSERT INTO api_keys (value, organization_id, created_at, updated_at)
+      SELECT organizations.api_key, organizations.id, organizations.created_at, organizations.created_at
+      FROM organizations
+      SQL
+    end
```
ancorcruz, line 19: Do you also plan to drop the api_key column in organizations table at some point?

app/models/api_key.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+class ApiKey < ApplicationRecord
+  include PaperTrailTraceable
+
+  belongs_to :organization
+
```
ancorcruz, line 7: I would also add a validation of value presence here? what do you think?

spec/factories/organizations.rb
```diff
@@ -13,7 +13,11 @@
     end
 
     after(:create) do |organization, evaluator|
-      organization.webhook_endpoints.create(webhook_url: evaluator.webhook_url)
+      if evaluator.webhook_url
+        organization.webhook_endpoints.create!(webhook_url: evaluator.webhook_url)
+      end
+
+      organization.api_keys.create!
```
ancorcruz, line 20: could you instead of creating the api_keys here add it to the organizations factory?

spec/models/api_key_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    let(:unique_value) { SecureRandom.uuid }
+
+    before do
+      allow(SecureRandom).to receive(:uuid).and_return(used_value, unique_value)
+    end
+
+    it 'returns unique value between all ApiKeys' do
+      expect(subject).to eq unique_value
+    end
+  end
```
ancorcruz, line 60: I don't see the need of testing these two methods (in my previous comment I have combined both into a single private method) as long as the `save` behavior is tested.

spec/models/api_key_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+
+    context 'with a persisted record' do
+      let(:api_key) { create(:api_key) }
+
+      it 'does not call #set_value' do
+        expect(api_key).not_to have_received(:set_value)
+      end
+    end
```
ancorcruz, line 30: I would avoid stubbing when possible...
here you can just check an uuid is assigned to a new records when saved,
it does not change when the record is already persisted and have one assigned...
and it handles the collision with an existing record with the same value (here you will have to stub SecureRandom I guess).

ancorcruz (changes requested):
Good stuff, I have left you a few comments.

Let me know if you would like to have a chat about my comments.

app/services/users_service.rb
```diff
@@ -29,7 +29,10 @@ def register(email, password, organization_name)
 
     ActiveRecord::Base.transaction do
       result.user = User.create!(email:, password:)
-      result.organization = Organization.create!(name: organization_name, document_numbering: 'per_organization')
+
+      result.organization = Organizations::CreateService
+        .call(name: organization_name, document_numbering: 'per_organization')
+        .organization
```
vincent-pochet, line 35: To raise the new service error (just in case):

```suggestion
      result.organization = Organizations::CreateService
        .call(name: organization_name, document_numbering: 'per_organization')
        .raise_if_error!
        .organization
```

db/migrate/20241021140054_create_api_keys.rb
vincent-pochet: :+1: 

An index on `organization_id` is also mandatory since we are doing `organization.api_keys.first` in multiple places

spec/models/organization_spec.rb
```diff
@@ -138,11 +139,43 @@
...
 
-  describe 'Callbacks' do
-    it 'generates the api key' do
-      organization.save!
+  describe '#save' do
+    subject { organization.save! }
 
-      expect(organization.api_key).to be_present
+    before do
+      allow(organization).to receive(:generate_document_number_prefix).and_call_original # rubocop:disable RSpec/SubjectStub
```
vincent-pochet, line 146: Stubbing private method is not a good idea (that's why the rubocop rule is here :) )
I think it's better to check that after a save, the expected values are set on the model.

The point is that most of the time specs should not be based on the inner implementation, but on the expected result otherwise the test might be passing even if the what is tested is not working as expected

vincent-pochet (changes requested):

db/migrate/20241021140054_create_api_keys.rb
floganz: Adding unique index completely make sense. It'd be global rather then scoped by `organization_id` since we need by ApiKey identify only 1 organization.

However, I missed something with @vincent-pochet comment. Index on `organization_id` already in place as part of `t.references`. Maybe you mean some other combination for index?

db/migrate/20241021140054_create_api_keys.rb
floganz: Yes, that will be removed at some point.

spec/models/organization_spec.rb
floganz: Regarding rubocop, there're no warning about stubbing that private method. Maybe we need to check our Rubocop config?

The disabled cop `RSpec/SubjectStub` has a bit different purpose – it ensures that we do not stub thing that we're testing. 
In this case it has false positive match since subject in my case is different, testing that method is called by when another action is done allowed behaviour.

db/migrate/20241021140054_create_api_keys.rb
vincent-pochet: Ah indeed, forget my remark, the index is here. Sorry for the noise ;)

app/models/api_key.rb
floganz: I’ve tried to do add it but it doesn’t look great.

We cannot add `validates :value, presence: true`  with a `before_create` callback that sets the value. The value for new records will be always missing.
If we change callback to `before_validation` we can add a validation but it’s not possible to test, there will no ability to set `nil` value and receive an error that value is missing because of callback. With such setup the validation is a redundancy.

The most reasonable way of adding it by only applying it to update action. But honestly, I'd skip it at all for this particular case.
Let me know if you think I missed something.

floganz:
@ancorcruz @vincent-pochet Thanks for your reviews!
I've pushed changes, can you take another look?

vincent-pochet (commented):
Looks great! :rocket: 

Just one comment around the fallback for the lookup of the API key at organization level in the base controller

app/controllers/api/base_controller.rb
```diff
@@ -27,8 +27,8 @@ def authenticate
       true
     end
 
-    def current_organization(api_key = nil)
-      @current_organization ||= Organization.find_by(api_key:)
+    def current_organization(api_key_value = nil)
+      @current_organization ||= ApiKey.find_by(value: api_key_value)&.organization
```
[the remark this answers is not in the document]
floganz: Had a chat with Vincent about this. We agreed that it could be done but there’s no big reason for it because with the next step for api key management feature (rotating a key) it might create more confusion then benefits. Since old key shouldn't give access anymore.

ancorcruz (approved):
LGTM. thanks!

PR #2738, opened 2024-10-24
add migration to set parent_ids to charges

annvelents:
## Context

We want to populate parent_ids for charges. To define a parent for a charge, we're looking into plan's parent plan and trying to find a charge with matching charge_model and properties. If parent plan has more than one matching charge, we claim that we're not able to define a parent charge

## Description

added a migration that:
- goes through children plans and loads their charges and parents with their charges
- for each charge of children plan it tries to select charge from parent plan with matching charge model and charge properties
- if only one matching charge is found, update child charge with parent charge (without callbacks)

added a rake task to seed test data, so we can deploy it onto staging, create any number of plans with children, select to not save parent_ids for their charges, run the migration code and check the result

### DIFF
--- a/db/migrate/20241022144437_add_parent_to_charges_from_plan_parent.rb
+++ b/db/migrate/20241022144437_add_parent_to_charges_from_plan_parent.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+#
+class AddParentToChargesFromPlanParent < ActiveRecord::Migration[7.1]
+  def change
+    now = Time.current
+    Plan.where.not(parent_id: nil).includes(:charges, parent: :charges).find_each do |plan|
+      parent = plan.parent
+      next if plan.charges.empty? || parent.charges.empty?
+
+      plan.charges.each do |child_charge|
+        parent_charges = parent.charges.where(charge_model: child_charge.charge_model, properties: child_charge.properties)
+        next if parent_charges.length != 1
+
+        child_charge.update(parent_id: parent_charges[0].id)
+      end
+    end
+    puts '=' * 80
+    puts 'Migration took: ' + (Time.current - now).to_s
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20241022144437_add_parent_to_charges_from_plan_parent.rb
```diff
@@ -0,0 +1,25 @@
...
+        parent_charges = parent.charges.select do |charge|
+          charge.charge_model == child_charge.charge_model && charge.properties == child_charge.properties
+        end
+        next if parent_charges.length != 1
+
+        child_charge.update_columns(parent_id: parent_charges[0].id)
+      end
+    end
+    puts '=' * 80
+    puts 'Migration took: ' + (Time.current - now).to_s + ' seconds'
```
annvelents, line 20: Running locally on 10_000 plans and 136_000 charges took 13 secs

db/migrate/20241022144437_add_parent_to_charges_from_plan_parent.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class AddParentToChargesFromPlanParent < ActiveRecord::Migration[7.1]
+  def up
+    now = Time.current
+    Plan.where.not(parent_id: nil).includes(:charges, parent: :charges).find_each do |plan|
```
vincent-pochet, line 6: To avoid issues with future fresh install if the Plan model definition changes, you should redefine the Plan model and its relation directly in this migration class. It will make sure that the Plan model you are using is loaded with the expected attributes.

You can refer to https://github.com/getlago/lago-api/blob/main/db/migrate/20240603095841_refresh_cached_aggregations.rb if you need an example

annvelents: ah, right
thanks!

vincent-pochet (approved):

vincent-pochet (approved):

lovrocolic (approved):

PR #2750, opened 2024-10-29
misc(clickhouse): Use official clickhouse-activerecord gem

vincent-pochet:
## Context

Two pull requests opened on the `clickhouse-active` records repository were merged lately.
- https://github.com/PNixx/clickhouse-activerecord/pull/158
- https://github.com/PNixx/clickhouse-activerecord/pull/169

Since now all mandatory pieces that we require to use it in production are here, we can use the official gem rather than the fork.

## Description

This PR replaces `gem 'clickhouse-activerecord', git: 'https://github.com/getlago/clickhouse-activerecord.git'` with `gem "clickhouse-activerecord", "~> 1.2.0"`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,134 +1,134 @@
 # frozen_string_literal: true
 
-source 'https://rubygems.org'
+source "https://rubygems.org"
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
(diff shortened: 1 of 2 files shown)


nudded (approved):
🚀

PR #2757, opened 2024-10-31
feat(dunning): find eligible customers and applied dunning campaign

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

This change adds the dunning campaign trigger to find eligible customers and applied campaign threshold to run the next attempt.

### DIFF
--- a/app/jobs/dunning_campaigns/bulk_process_job.rb
+++ b/app/jobs/dunning_campaigns/bulk_process_job.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-module Dunning
-  class ProcessCampaignsJob < ApplicationJob
+module DunningCampaigns
+  class BulkProcessJob < ApplicationJob
     queue_as :default
 
     def perform
--- a/app/services/dunning_campaigns/bulk_process_service.rb
+++ b/app/services/dunning_campaigns/bulk_process_service.rb
@@ -2,5 +2,55 @@
 
 module DunningCampaigns
   class BulkProcessService < BaseService
+    # Find all eligible customers
+    # Find applicable dunning campaign
+    # next step?
+    # Queue job for next step; customer, campaign
+
+    def call
+      return result unless License.premium?
+
+      eligible_customers.find_each do |customer|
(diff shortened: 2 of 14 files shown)


app/jobs/dunning_campaigns/bulk_process_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module DunningCampaigns
+  class BulkProcessJob < ApplicationJob
```
rsempe, line 4: is this job going to be called from another location?

just wondering if it's not simpler to directly call `DunningCampaigns::BulkProcessService.call` from the clock job

app/services/dunning_campaigns/bulk_process_service.rb
```diff
@@ -0,0 +1,75 @@
...
+      result
+    end
+
+    private
+
+    def eligible_customers
+      Customer
+        .joins(:organization)
+        .where(exclude_from_dunning_campaign: false)
+        .where("organizations.premium_integrations @> ARRAY[?]::varchar[]", ['auto_dunning'])
```
rsempe, line 21: Has the name `auto_dunning` been validated by the product team?

rsempe (approved):
👏

app/jobs/dunning_campaigns/bulk_process_job.rb
ancorcruz: as running this service will take longer with a growing number of organizations and customers, IMO it is better to free clock worker and just queue this job an let the normal sidekiq workers take care of it. We can handle queue priority, and more stuff that is not available with clock.

app/services/dunning_campaigns/bulk_process_service.rb
ancorcruz: can't remember, is this name visible for users?

frontend is using it... I'll check with product.

Thanks

app/jobs/dunning_campaigns/bulk_process_job.rb
rsempe: fair enough

PR #2758, opened 2024-10-31
chore(currencies): Add BHD support

jdenquin:
https://github.com/getlago/lago/issues/422


vincent-pochet (approved):
:+1: 
You have to regenerate the GraphQL schema as well to expose it.

PR #2761, opened 2024-10-31
feat(plan-charges-cascading): Support charge filters cascading

lovrocolic:
## Context

## Description

This PR supports cascading of charge filters. There are several rules:

- If main attribute that supports cascading is not true -> cascading is disabled
- Charge filter properties are cascaded only if the same properties are not already overridden in the child filter
- We need to ensure that all filters that are specific to the child are not changed in any way (extra filters that don't exist on parent)
- When performing cascading, we are basically at the same time updating parent filters and children filters. However, we need to ensure that logic which handles cascading on children filter uses parent filter attributes **before** update for comparison. That's why we pass parent filter attributes before any update to the children logic.

### DIFF
--- a/app/services/charge_filters/create_or_update_batch_service.rb
+++ b/app/services/charge_filters/create_or_update_batch_service.rb
@@ -2,9 +2,13 @@
 
 module ChargeFilters
   class CreateOrUpdateBatchService < BaseService
-    def initialize(charge:, filters_params:)
+    def initialize(charge:, filters_params:, options: {})
       @charge = charge
       @filters_params = filters_params
+      @options = options
+      @cascade_updates = options[:cascade]
+      @parent_filters_attributes = options[:parent_filters] || []
+      @parent_filters = ChargeFilter.with_discarded.where(id: parent_filters_attributes.map { |f| f['id']})
 
       super
     end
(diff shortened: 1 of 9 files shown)


app/services/charge_filters/create_or_update_batch_service.rb
```diff
@@ -2,9 +2,13 @@
...
 module ChargeFilters
   class CreateOrUpdateBatchService < BaseService
-    def initialize(charge:, filters_params:)
+    def initialize(charge:, filters_params:, options: {})
       @charge = charge
       @filters_params = filters_params
+      @options = options
+      @cascade_updates = options[:cascade]
+      @parent_filters_attributes = options[:parent_filters] || []
+      @parent_filters = ChargeFilter.with_discarded.where(id: parent_filters_attributes.map { |f| f['id'] })
```
vincent-pochet, line 11:
```suggestion
      @parent_filters = if parent_filters_attributes.blank?
         ChargeFilter.none
       else
         ChargeFilter.with_discarded.where(id: parent_filters_attributes.map { |f| f['id'] })
        end
```

app/services/plans/update_service.rb
```diff
@@ -126,8 +126,16 @@ def cascade_charge_update(charge, payload_charge)
...
-        if child_charge && charge.equal_properties?(child_charge)
-          Charges::UpdateJob.perform_later(charge: child_charge, params: payload_charge, cascade: true)
+        if child_charge
+          Charges::UpdateJob.perform_later(
+            charge: child_charge,
+            params: payload_charge,
+            options: {
+              cascade: true,
+              parent_filters: charge.filters.map(&:attributes),
+              equal_properties: charge.equal_properties?(child_charge)
```
vincent-pochet, line 136: Maybe we should rename the option attribute `cascade_options`, to clearly defined what is its purpose.

vincent-pochet (approved):
Two comments, looks good otherwise :rocket:

PR #2771, opened 2024-11-01
Feat add ability to rotate api key

floganz:
## Context

Another step toward API key management

## Description

Decouple HMAC secret from API keys.
Add ability to instantly rotate API key.


vincent-pochet (approved):
Looks great! :rocket: 

Not a blocker here, but do not hesitate to split pull request in small pieces as it helps with review.
For example, the api_key => hmac migration could be done in a separate PR.

PR #2777, opened 2024-11-05
fix: Ensure DataExports::CombinePartsJob uniqueness

ancorcruz:
This should fix a race condition when multiple data export parts are finished very close in time resulting into multiple CombinePartsJob queued, hence active storage fails because of duplicated key.


nudded (approved):
nice catch! 👍

jdenquin (approved):

PR #2779, opened 2024-11-05
feat(api-key-management): Track when API key was last used

floganz:
## Description

Add ability to track when API key was last used

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -8,31 +8,45 @@ module Api
 
     before_action :authenticate
     before_action :set_context_source
+    before_action :track_api_key_usage
     include Trackable
 
     rescue_from ActionController::ParameterMissing, with: :bad_request_error
 
     private
 
+    attr_reader :current_api_key, :current_organization
+
     def authenticate
-      auth_header = request.headers['Authorization']
+      return unauthorized_error unless auth_token
 
-      return unauthorized_error unless auth_header
+      @current_api_key = ApiKey.active.find_by(value: auth_token)
 
-      api_key = auth_header.split(' ').second
+      return unauthorized_error unless current_api_key
 
-      return unauthorized_error unless api_key
-      return unauthorized_error unless current_organization(api_key)
-
+      @current_organization = current_api_key.organization
       true
     end
 
-    def current_organization(api_key_value = nil)
-      @current_organization ||= ApiKey.active.find_by(value: api_key_value)&.organization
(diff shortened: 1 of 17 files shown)


jdenquin:
@floganz can you rename your PR to follow the same convention that we have? 🙏

app/services/api_keys/track_usage_service.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module ApiKeys
+  class TrackUsageService < BaseService
+    def call
+      ApiKey.find_each do |api_key|
```
jdenquin, line 6: This is a bit ambitious!
You'll run over all the `ApiKey` event the one that are not used.
Is it better to loop over whats in cache?

floganz: Indeed, that looks heavy but I still think that it’s acceptable option.

In order to do it another way (lookup in cache) we either need to know all keys we’re interested in which basically loading every api key to build cache keys. Or our cache store should have some pattern based lookup which will depend on particular cache store used which might vary for self hosted clients. Which is also solvable but that bring additional complexity to the solution.

Considering that all our clients have to use API I suppose that majority of key always in use. Maybe not each minute but each hour.

Speaking of speed improvements we can do 2 optimizations:
- Fetch API keys ids and iterate over it, looking for a cache key rather then loading whole active record object. That should reduce database load but it will have the same load on cache.
- Run this job less frequent (current each 5 minutes). E.g. each 1 hour.

What do you think?

jdenquin: I understand!

I would say the best option is to delay it yes, but is it aligned with what product defined?

nudded (approved):

app/controllers/api/base_controller.rb
```diff
@@ -8,31 +8,45 @@ class BaseController < ApplicationController
...
-    def authenticate
-      auth_header = request.headers['Authorization']
+    attr_reader :current_api_key, :current_organization
 
-      return unauthorized_error unless auth_header
+    def authenticate
+      return unauthorized_error unless auth_token
 
-      api_key = auth_header.split(' ').second
+      @current_api_key = ApiKey.active.find_by(value: auth_token)
```
nudded, line 23: I think you'll tackle it in a future PR, but `.active` imho should be a default scope

floganz: Exactly.

PR #2793, opened 2024-11-07
feat(api-key-management): Add name to the API key model

floganz:
## Description

Add name to the API key model.
Add default scope to the API key.

### DIFF
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -10,7 +10,7 @@ class ApiKey < ApplicationRecord
   validates :value, uniqueness: true
   validates :value, presence: true, on: :update
 
-  scope :active, -> { where('expires_at IS NULL OR expires_at > ?', Time.current) }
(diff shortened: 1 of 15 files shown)


app/models/api_key.rb
```diff
@@ -10,7 +10,7 @@ class ApiKey < ApplicationRecord
   validates :value, uniqueness: true
   validates :value, presence: true, on: :update
 
-  scope :active, -> { where('expires_at IS NULL OR expires_at > ?', Time.current) }
```
nudded, line 13: I think you can keep the scope and maybe also add a `expired` scope? 🤔 and then just use .active in the default_scope? (I prefer not thinking and having the name `active` is really easy to understand.)

nudded (approved):
👍

app/models/api_key.rb
floganz: Don't mind keeping `.active` but I'd prefer to delay adding `.expired` until it will be needed in some feature.

PR #2808, opened 2024-11-12
misc(stripe): Refact payment method update webhooks

vincent-pochet:
## Context

This PR is related to the remaining ` PaymentProviders::Stripe::HandleEventJob` dead jobs with `BaseService::NotFoundFailure: stripe_customer_not_found`  error

## Description

The main goal of the PR is to refactor the handling of the `setup_intent.succeeded` and `customer.updated` to move the logic in a dedicated webhook service (in the objective of killing the unmaintainable `PaymentProviderCustomers::StripeService`).

It also fixes a case in `setup_intent.succeeded` when the webhook has a `customer` with null value and no Lago ids in the metadata. Since in this specif case the webhook does not target a lago resource, the handling should not raise an error but should be ignored.

### DIFF
--- a/app/services/payment_provider_customers/stripe/update_payment_method_service.rb
+++ b/app/services/payment_provider_customers/stripe/update_payment_method_service.rb
@@ -1,0 +1,43 @@
+# frozen_String_literal: true
+
+module PaymentProviderCustomers
+  module Stripe
+    class UpdatePaymentMethodService < BaseService
+      def initialize(stripe_customer:, payment_method_id:)
+        @stripe_customer = stripe_customer
+        @payment_method_id = payment_method_id
+
+        super
+      end
+
+      def call
+        return result.not_found_failure!(resource: 'stripe_customer') unless stripe_customer
+
+        stripe_customer.payment_method_id = payment_method_id
+        stripe_customer.save!
+
+        reprocess_pending_invoices
+
+        result.stripe_customer = stripe_customer
+        result
+      end
+
+      private
+
+      attr_reader :stripe_customer, :payment_method_id
+
+      delegate :customer, to: :stripe_customer
+
+      def reprocess_pending_invoices
+        invoices = customer.invoices
+          .payment_pending
+          .where(ready_for_payment_processing: true)
+          .where(status: 'finalized')
+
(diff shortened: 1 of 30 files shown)


app/services/payment_provider_customers/stripe/update_payment_method_service.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      def reprocess_pending_invoices
+        invoices = customer.invoices
+          .payment_pending
+          .where(ready_for_payment_processing: true)
+          .where(status: 'finalized')
+
+        invoices.find_each do |invoice|
+          Invoices::Payments::StripeCreateJob.perform_later(invoice)
+        rescue ActiveJob::Uniqueness::JobNotUnique
```
nudded, line 39: I'm not sure this job will raise an error when enqueued twice (`on_conflict: log`)

app/services/payment_providers/webhooks/stripe/setup_intent_succeeded_service.rb
```diff
@@ -0,0 +1,58 @@
...
+          result.stripe_customer = stripe_customer
+          result
+        rescue ::Stripe::PermissionError => e
+          result.service_failure!(code: 'stripe_error', message: e.message)
+        end
+
+        private
+
+        def stripe_customer
+          @stripe_customer ||= PaymentProviderCustomers::StripeCustomer
```
nudded, line 30: this feels too complicated of a query to have to perform here, should we move this to the model instead? (e.g. someone will forget to join to the organization_id at some point)

nudded: And I just noticed it's duplicated in some places as well, I think that's a clear sign it should be a scope or finder method 😄

nudded (changes requested):
Nice!

Left some small comments

app/services/payment_provider_customers/stripe/update_payment_method_service.rb
vincent-pochet: Good catch! It was changed in the job. Updating!

nudded (approved):

PR #2818, opened 2024-11-14
fix(netsuite): Aggregator jobs should be unique

ivannovosad:
## Context

While syncing invoices to Netsuite, some invoices were sent twice

## Description

Add `unique :until_executed` to aggregator jobs so that the resources are only sync'd once.

### DIFF
--- a/app/jobs/integrations/aggregator/credit_notes/create_job.rb
+++ b/app/jobs/integrations/aggregator/credit_notes/create_job.rb
@@ -6,7 +6,7 @@ module Integrations
       class CreateJob < ApplicationJob
         queue_as 'integrations'
 
-        unique :until_executed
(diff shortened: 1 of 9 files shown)


app/jobs/integrations/aggregator/credit_notes/create_job.rb
```diff
@@ -6,6 +6,8 @@ module CreditNotes
       class CreateJob < ApplicationJob
         queue_as 'integrations'
 
+        unique :until_executed
```
vincent-pochet, line 9: I guess you should use the `on_conflict: :log` option, to avoid errors when calling duplicated `perform_later` 
```suggestion
        unique :until_executed, on_conflict: :log
```

vincent-pochet (approved):

PR #2820, opened 2024-11-14
fix(adyen): Handle Adyen::AuthenticationError when creating the customer

vincent-pochet:
## Description

This Pull Request adds the handling of `Adyen::AuthenticationError` in the `PaymentProviderCustomers::AdyenService#create` method.
It will keep delivering an error webhook, but will not raise the authentication error anymore as nothing can be handled at Lago's level.

NOTE: In a future, the related `PaymentProvider` should marked as failed/disconnected/invalid and the error logged to make it visible in the web application and in the API

### DIFF
--- a/app/services/payment_provider_customers/adyen_service.rb
+++ b/app/services/payment_provider_customers/adyen_service.rb
@@ -20,6 +20,11 @@ module PaymentProviderCustomers
 
       result.checkout_url = checkout_url_result.checkout_url
       result
+    rescue Adyen::AuthenticationError
+      # NOTE: Authentication errors will be sent to the account owner with a webhook.
+      #       Since nothing can be done on Lago's side, we should not raise the error.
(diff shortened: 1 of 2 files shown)


app/services/payment_provider_customers/adyen_service.rb
```diff
@@ -20,6 +20,11 @@ def create
 
       result.checkout_url = checkout_url_result.checkout_url
       result
+    rescue Adyen::AuthenticationError
```
nudded, line 23: arguably this should be `call` instead of `create`, You probably thought the same and decided it was too much work, just wanted to have that confirmed :D

vincent-pochet: @nudded exactly :D
I started to refact it and... it was A LOT of changes... haha.
I'm working on a proposal to tackle all of those refactors once for all...

PR #2823, opened 2024-11-14
fix(subscription upgrade): Upgrade subscriptions previously considered downgrade on amount increase

ancorcruz:
## Context

Currently, when a user updates the subscription fee amount for a plan associated with a pending downgrade (scheduled to activate at the end of the billing period), the system does not treat the pending subscription as an upgrade. As a result, the new subscription is not activated, and the previous one is not terminated.

## Description

This change extracts the subscription upgrade logic into a dedicated service, ensuring it can be consistently applied across flows that trigger subscription upgrades. Currently, this includes plan updates and initial subscription creation.

An additional edge case, involving plan subscription overrides, is not addressed in this PR. Handling this scenario introduces extra technical and product complexities that require further consideration.

### DIFF
--- a/app/services/plans/update_service.rb
+++ b/app/services/plans/update_service.rb
@@ -268,12 +268,21 @@ module Plans
       end
     end
 
-    # NOTE: We should remove pending subscriptions
-    #       if plan has been downgraded but amount cents of pending plan became higher than original plan.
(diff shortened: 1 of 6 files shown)


nudded (approved):

vincent-pochet (approved):
Great one :rocket:

PR #2825, opened 2024-11-15
fix(stripe): Handle missing payment

vincent-pochet:
## Context

Some `PaymentProviders::Stripe::HandleEventJob` with 'payment_intent.payment_failed' or 'payment_intent.succeeded' event type are failing with `BaseService::NotFoundFailure - stripe_payment_not_found`.

The webhook contains metadata of lago's invoices but the related `Payment` cannot be found in the database. It is probably because the job creating the stripe payment was killed allong with the worker running it under heavy load.

## Description

This PR changes the logic to re-create the `Payment` record when missing using the data from the webhook object. It will allow to update the `invoice.payment_status` accordingly

### DIFF
--- a/app/models/payment_providers/stripe_provider.rb
+++ b/app/models/payment_providers/stripe_provider.rb
@@ -2,6 +2,8 @@
 
 module PaymentProviders
   class StripeProvider < BaseProvider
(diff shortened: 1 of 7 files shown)


nudded (approved):
nice!

PR #2826, opened 2024-11-15
fix(netsuite): Fix possible duplicate integration resources creation

ivannovosad:
## Context

When the Netsuite jobs that create invoices, credit note and paymetns were enqueued at the same time, they sometimes created duplicate integration resources.

## Description

This PR fixes this issue by checking if the resource already exists in our DB.


jdenquin (approved):
Should we QA this on staging before?

vincent-pochet (approved):

PR #2828, opened 2024-11-15
Fix: allow to edit fees with more than 2 decimal price

annvelents:
## Context

currently when an organization is editing fee, they can enter only 2 digits after the comma for new unit_amount. Instead, they want to be able to enter as many digits as they want as it already works for unit amount in plans.

## Description

Added unit_precise_amount_cents to adjusted fee;
Added this new field into consideration when creating fees;
Added a migration to populate unit_precise_amount_cents for currently existing adjusted fees

### DIFF
--- a/app/services/adjusted_fees/create_service.rb
+++ b/app/services/adjusted_fees/create_service.rb
@@ -17,7 +17,7 @@ module AdjustedFees
       charge = fee.charge
       return result.validation_failure!(errors: {charge: ['invalid_charge_model']}) if disabled_charge_model?(charge)
 
-      unit_precise_amount_cents = (params[:unit_precise_amount].presence || 0).to_f * fee.amount.currency.subunit_to_unit
+      unit_precise_amount_cents = params[:unit_precise_amount].to_f * fee.amount.currency.subunit_to_unit
       adjusted_fee = AdjustedFee.new(
         fee:,
         invoice: fee.invoice,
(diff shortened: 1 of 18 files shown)


db/migrate/20241118103032_fulfill_adjusted_fee_unit_precise_amount_cents.rb
```diff
@@ -0,0 +1,17 @@
...
+class FulfillAdjustedFeeUnitPreciseAmountCents < ActiveRecord::Migration[7.1]
+  class AdjustedFee < ApplicationRecord
+    belongs_to :invoice
+  end
+
+  class Invoice < ApplicationRecord
+    enum status: {draft: 0}
+  end
+
+  def change
```
floganz, line 12: Does this migration should work the same when we run it up and down(rollback)?

app/services/adjusted_fees/create_service.rb
```diff
@@ -17,22 +17,23 @@ def call
       charge = fee.charge
       return result.validation_failure!(errors: {charge: ['invalid_charge_model']}) if disabled_charge_model?(charge)
 
+      unit_precise_amount_cents = (params[:unit_precise_amount].presence || 0).to_f * fee.amount.currency.subunit_to_unit
```
floganz, line 20: I guess we can simplify this a bit.
`.to_f` converts both `nil` and `""` to "0.0" which seems what you're trying to achieve.

```suggestion
      unit_precise_amount_cents = params[:unit_precise_amount].to_f * fee.amount.currency.subunit_to_unit
```

floganz (approved):

db/migrate/20241118103032_fulfill_adjusted_fee_unit_precise_amount_cents.rb
annvelents: oops, yes forgot about this

PR #2835, opened 2024-11-20
feat(dunning): Mutation to delete a dunning campaign

rsempe:
## Roadmap Task

 ## Context

 ## Description

This change adds a new mutation to graphql api to delete a dunning campaign.


app/services/dunning_campaigns/destroy_service.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def call
+      return result.not_found_failure!(resource: "dunning_campaign") unless dunning_campaign
+      return result.forbidden_failure! unless dunning_campaign.organization.auto_dunning_enabled?
+
+      ActiveRecord::Base.transaction do
+        dunning_campaign.discard!
+        dunning_campaign.thresholds.discard_all
+
+        # TODO: Reset counters for customers that were in the dunning campaign
```
ancorcruz, line 19: counter cache column is defined in the dunning_campaigns table, hence, there is nothing to do here, no reset required.

Just drop the comment please.

ancorcruz (approved):
looks great!

app/services/dunning_campaigns/destroy_service.rb
rsempe: true for `customers_count` 👍 

but we need to reset `last_dunning_campaign_attempt` and `last_dunning_campaign_attempt_at` at the customer level

PR #2837, opened 2024-11-20
bug(Coupon) - Use advisory lock to make sure coupon application is correct

nudded:
## Description

Coupon deduction was not resistant to concurrency issues. This PR introduces safe-guards so coupons are applied correctly.


vincent-pochet (approved):
Nice one :rocket: 
I'm wondering if we could build a generic pattern based on this approach. We probably have other places were this could help.

PR #2838, opened 2024-11-20
improve(CreditNote::FromTermination) - Use creditable_amount_cents rather than duplicating the logic

nudded:
## Description

Rather than duplicating the logic for calculating how much can be credited, we can rely on the `creditable_amount_cents` on Fee.


brunomiguelpinto (approved):

vincent-pochet (approved):

nudded:
For future reference: this fix is not correct, we agreed to always deduct already issued credit notes.

PR #2839, opened 2024-11-21
misc(throttling): Add concurrent jobs throttling config

ivannovosad:
## Context

Fix external apps rate limits or concurrency limits

## Description

This PR adds sidekiq-throttled gem and configures some aggregator jobs to have a limit of max 5 concurrent jobs.
It also adds a new "Throttled" tab in Sidekiq UI.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem "puma", "~> 6.4"
 gem "rails", "~> 7.1.3.4"
 gem "redis"
 gem "sidekiq"
+gem "sidekiq-throttled", '1.4.0' # '1.5.0' was losing some jobs
 gem "dry-validation"
 
 # Security
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem "rails", "~> 7.1.3.4"
 gem "redis"
 gem "sidekiq"
 gem "sidekiq-throttled", '1.4.0' # '1.5.0' was losing some jobs
+gem "throttling"
 gem "dry-validation"
 
 # Security
--- a/app/jobs/integration_customers/create_job.rb
+++ b/app/jobs/integration_customers/create_job.rb
@@ -5,6 +5,7 @@ module IntegrationCustomers
     queue_as 'integrations'
 
     retry_on LagoHttpClient::HttpError, wait: :polynomially_longer, attempts: 3
+    retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 100
 
     def perform(integration_customer_params:, integration:, customer:)
       result = IntegrationCustomers::CreateService.call(params: integration_customer_params, integration:, customer:)
--- a/app/jobs/integration_customers/create_job.rb
+++ b/app/jobs/integration_customers/create_job.rb
@@ -5,7 +5,7 @@ module IntegrationCustomers
     queue_as 'integrations'
 
     retry_on LagoHttpClient::HttpError, wait: :polynomially_longer, attempts: 3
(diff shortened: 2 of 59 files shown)


app/jobs/integration_customers/create_job.rb
```diff
@@ -5,6 +5,7 @@ class CreateJob < ApplicationJob
     queue_as 'integrations'
 
     retry_on LagoHttpClient::HttpError, wait: :polynomially_longer, attempts: 3
+    retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 100
```
nudded, line 8: 100 is a bit excessive. sidekiq uses 25 retries as a default, 25 seems more sensible 😄

app/services/integrations/aggregator/base_service.rb
```diff
@@ -50,6 +50,26 @@ def provider_key
         end
       end
 
+      def throttle!(*providers)
```
nudded, line 53: this is currently not being used to check anything correct?

ivannovosad: It is used in the aggregator services, [here](https://github.com/getlago/lago-api/pull/2839/files#diff-150de67c3e1df586da34867a50a0b0dda236f499eceb3dec3cf2df9174f4bdf2R17) for example.

Gemfile
```diff
@@ -15,6 +15,8 @@ gem "puma", "~> 6.4"
 gem "rails", "~> 7.1.3.4"
 gem "redis"
 gem "sidekiq"
+gem "sidekiq-throttled", '1.4.0' # '1.5.0' was losing some jobs
```
nudded, line 18: This is quite an ominous comment :smile:

app/jobs/integration_customers/create_job.rb
```diff
@@ -5,6 +5,7 @@ class CreateJob < ApplicationJob
     queue_as 'integrations'
 
     retry_on LagoHttpClient::HttpError, wait: :polynomially_longer, attempts: 3
+    retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
```
nudded, line 8: I feel like this is generic enough that we could consider moving it to `ApplicationJob` or even the `ConcurrencyThrottlable` module

app/services/integrations/aggregator/base_service.rb
```diff
@@ -50,6 +50,26 @@ def provider_key
...
+        end
+      end
+
+      def throttle_key
+        # Hubspot and Xero calls are throttled globally, others are throttled per api key or token
+        case provider
+        when 'netsuite'
+          integration.token_id
+        when 'anrok'
+          integration.api_key
```
nudded, line 67: This could potentially store api_keys in Redis, maybe that's a security issue? 🤔

nudded (approved):

Gemfile
ivannovosad: Yes, but the gem will be not needed after we migrate to Solid Queue.

app/jobs/integration_customers/create_job.rb
ivannovosad: Now it is in the `BaseService` where all the other erros are defined as well. 
We can define it it the `throttling.rb` initializer, since it is just:

```
class ThrottlingError < StandardError; end
```

I wouldn't put it in `ApplicationJob ` because it can be raised even outside of a job. It is dependent on redis not sidekiq (or active job)

app/services/integrations/aggregator/base_service.rb
ivannovosad: Good point!

PR #2846, opened 2024-11-21
misc: Add missing require in DailyUsages::FillHistoryService

vincent-pochet:
## Context

Related to https://github.com/getlago/lago-api/pull/2843

## Description

We missed to move the `require "timecop"` :facepalm: ...


app/services/daily_usages/fill_history_service.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "timecop"
```
nudded, line 3: `timecop` is not installed in production.

PR #2849, opened 2024-11-22
misc: Activate aggregate_failures by default on specs

rsempe:
RSpec::Core provides metadata integration for aggregating failures by default on RSpec.

Why not activating it, instead of putting everywhere `:aggregate_failures` :) 

The legacy calls could be removed later.


vincent-pochet (approved):

nudded (approved):
Oh 👍

PR #2852, opened 2024-11-22
fix: Include draft invoices inside wallet's ongoing balance

rsempe:
The goal of this PR is to include **draft invoices** in the **ongoing** balance.


app/services/wallets/balance/update_ongoing_service.rb
```diff
@@ -58,12 +58,16 @@ def credits_ongoing_usage_balance
...
       def ongoing_balance_cents
-        wallet.balance_cents - total_usage_amount_cents + pay_in_advance_usage_amount_cents
+        wallet.balance_cents - total_usage_amount_cents - draft_invoices_total_amount_cents + pay_in_advance_usage_amount_cents
       end
 
       def credits_ongoing_balance
         ongoing_balance_cents.to_f.fdiv(currency.subunit_to_unit).fdiv(wallet.rate_amount)
       end
+
+      def draft_invoices_total_amount_cents
```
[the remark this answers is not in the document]
rsempe: yep good spot 👍

app/services/invoices/refresh_draft_service.rb
```diff
@@ -31,6 +31,7 @@ def call
 
       ActiveRecord::Base.transaction do
         invoice.update!(ready_to_be_refreshed: false) if invoice.ready_to_be_refreshed?
+        invoice.customer.flag_wallets_for_refresh
```
nudded, line 34: This will be executed every 5 minutes for every draft invoice. Is there a simple way to only flag them if something has actually been updated?

app/services/invoices/refresh_draft_service.rb
```diff
@@ -69,6 +69,7 @@ def call
         calculate_result.raise_if_error!
 
         flag_lifetime_usage_for_refresh
```
nudded, line 71: I think we can add the `flag_lifetime_usage_for_refresh` to the if body as well right?

nudded (approved):

PR #2853, opened 2024-11-22
feat(api-key-management): Add permissions to the ApiKey

floganz:
## Context

Preparation step for implementing endpoint permissions for api keys.

## Description

Define permissions on each api key, fill with default access rights.

### DIFF
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -3,19 +3,67 @@
 class ApiKey < ApplicationRecord
   include PaperTrailTraceable
 
+  RESOURCES = %w[
+    add_on analytic billable_metric coupon applied_coupon credit_note customer_usage
+    customer fee invoice organization payment_request plan subscription lifetime_usage
+    tax wallet wallet_transaction webhook_endpoint
+  ].freeze
+
+  MODES = %w[read write].freeze
+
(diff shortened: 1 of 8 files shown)


app/models/api_key.rb
```diff
@@ -3,20 +3,68 @@
...
+    permissions[resource].include?(mode)
+  end
+
+  def self.default_permissions
+    RESOURCES.index_with { MODES.dup }
+  end
+
   private
 
+  def permissions_keys_compliance
```
nudded, line 43: This means that adding a resource would invalidate all api keys right?

floganz: Didn't think about this way.

I'd probably expect that if we add new resource we need think about permissions and update all API keys to reflect new changes. If we didn't update API key permissions that would indeed make all of them invalid which shouldn't prevent from using the key but probably will block update/rotate actions until new resource added to the key.

We can relax validation to allow resource be missing in a list which make it less scary adding new resources later. `#permit?` then need to be adjusted so that missing resource always return no access.

nudded: that sounds like a good idea 👍

nudded (approved):

vincent-pochet (approved):

PR #2857, opened 2024-11-25
fix(unique-jobs): Unique jobs were not enqueuing jobs with keyword args

ivannovosad:
## Context

There is a activejob-uniqueness bug that does not enqueue integration jobs when using named arguments.
This PR fixes that.

### DIFF
--- a/app/jobs/fees/create_pay_in_advance_job.rb
+++ b/app/jobs/fees/create_pay_in_advance_job.rb
@@ -4,7 +4,9 @@ module Fees
   class CreatePayInAdvanceJob < ApplicationJob
     queue_as :default
 
-    def perform(charge:, event:, billing_at: nil)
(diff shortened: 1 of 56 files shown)


app/jobs/fees/create_pay_in_advance_job.rb
```diff
@@ -4,7 +4,9 @@ module Fees
   class CreatePayInAdvanceJob < ApplicationJob
     queue_as :default
 
-    def perform(charge:, event:, billing_at: nil)
+    unique :until_executed, on_conflict: :log
+
+    def perform(charge, event, billing_at = nil)
```
nudded, line 9: @ivannovosad why is this needed?

nudded: (the removal of keyword arguments that is)

PR #2861, opened 2024-11-25
feat(dunning): reset customer dunning campaign last attempt data on threshold changes

ancorcruz:
## Roadmap

 ## Context

 ## Description

When a dunning campaign threshold changes, it resets customers last_dunning_campaign_attempt fields with the updated campaign applicable (through direct association or falling back to it through the organization default) when the thresholds does not match anymore (currency and/or amount_cents changed or deletion of the threshold).

### DIFF
--- a/app/services/dunning_campaigns/update_service.rb
+++ b/app/services/dunning_campaigns/update_service.rb
@@ -15,22 +15,9 @@ module DunningCampaigns
       return result.not_found_failure!(resource: "dunning_campaign") unless dunning_campaign
 
       ActiveRecord::Base.transaction do
-        dunning_campaign.name = params[:name] if params.key?(:name)
-        dunning_campaign.code = params[:code] if params.key?(:code)
-        dunning_campaign.description = params[:description] if params.key?(:description)
-        dunning_campaign.days_between_attempts = params[:days_between_attempts] if params.key?(:days_between_attempts)
-        dunning_campaign.max_attempts = params[:max_attempts] if params.key?(:max_attempts)
-
+        update_dunning_campaign_attributes
         handle_thresholds if params.key?(:thresholds)
-
-        unless params[:applied_to_organization].nil?
-          organization.dunning_campaigns.applied_to_organization
-            .update_all(applied_to_organization: false) # rubocop:disable Rails/SkipsModelValidations
-
-          organization.reset_customers_last_dunning_campaign_attempt
-
-          dunning_campaign.applied_to_organization = params[:applied_to_organization]
-        end
+        handle_applied_to_organization_update if params.key?(:applied_to_organization)
 
         dunning_campaign.save!
       end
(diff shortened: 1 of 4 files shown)


app/services/dunning_campaigns/update_service.rb
```diff
@@ -45,21 +32,84 @@ def call
 
     attr_reader :dunning_campaign, :organization, :params
 
+    def update_dunning_campaign_attributes
```
rsempe, line 35: do we really need this method?

i mean, we can directly call `assign_attributes`, no? :)

app/services/dunning_campaigns/update_service.rb
```diff
@@ -45,21 +32,84 @@ def call
...
+          unless threshold_matches
+            customer.update!(
+              last_dunning_campaign_attempt: 0,
+              last_dunning_campaign_attempt_at: nil
+            )
+          end
+        end
+    end
+
+    def handle_applied_to_organization_update
```
rsempe, line 102: 👍

app/services/dunning_campaigns/update_service.rb
```diff
@@ -45,21 +32,84 @@ def call
...
+      customers_applied_campaign = organization
+        .customers
+        .with_dunning_campaign_not_completed
+        .where(applied_dunning_campaign: dunning_campaign)
+
+      customers_fallback_campaign = organization
+        .customers
+        .falling_back_to_default_dunning_campaign
+        .with_dunning_campaign_not_completed
+        .where(dunning_campaign.applied_to_organization ? nil : "1 = 0")
```
rsempe, line 81: not sure to understand 🤔

app/services/dunning_campaigns/update_service.rb
```diff
@@ -45,21 +32,84 @@ def call
...
 
+    def update_dunning_campaign_attributes
+      dunning_campaign.assign_attributes(permitted_attributes)
+    end
+
+    def permitted_attributes
+      params.slice(:name, :code, :description, :days_between_attempts, :max_attempts)
+    end
+
     def handle_thresholds
```
rsempe, line 43: what about extracting this logic into a separate service `DunningCampaignThresholds::CreateOrUpdateBatchService` or something like this?

as it's quite complex, can be a bit easier to test and maintain, no?

rsempe (approved):

app/services/dunning_campaigns/update_service.rb
ancorcruz: it can be removed, this was a left over from initial query implementation, thanks for raising it.

app/services/dunning_campaigns/update_service.rb
ancorcruz: lets do it in a separate PR, need to unblock this for the next PR...

app/services/dunning_campaigns/update_service.rb
ancorcruz: true, it does not add much value. I'll change it.

PR #2862, opened 2024-11-25
fix(unique-jobs): Fix on retry in unique jobs

ivannovosad:
## Context

`retry_on` does not work with `until_executed` strategy.
https://github.com/veeqo/activejob-uniqueness/issues/75

## Description

This PR updates all the jobs that are affected with this issue  - a combination of `retry_on` and `unique :until_executed`

### DIFF
--- a/app/jobs/integrations/aggregator/subscriptions/crm/create_customer_association_job.rb
+++ b/app/jobs/integrations/aggregator/subscriptions/crm/create_customer_association_job.rb
@@ -7,8 +7,6 @@ module Integrations
         class CreateCustomerAssociationJob < ApplicationJob
           queue_as 'integrations'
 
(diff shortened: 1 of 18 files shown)


app/jobs/integrations/aggregator/subscriptions/crm/create_customer_association_job.rb
```diff
@@ -7,8 +7,6 @@ module Crm
         class CreateCustomerAssociationJob < ApplicationJob
           queue_as 'integrations'
 
-          unique :until_executed, on_conflict: :log
```
nudded, line 10: probably want to keep this as well? 🤔

brunomiguelpinto (approved):

nudded (approved):

PR #2863, opened 2024-11-25
Bug(Events) - make sure to cleanup clickhouse events after billable metric removal

nudded:
## description

We recently fixed the case where events were not correctly removed for new-style subscriptions (using `external_subscription_id`). We forgot about clickhouse events, this PR fixes that and will also remove those.


jdenquin:
CH hates deletion, but that make sense!
A deletion can be pretty CPU intensive because or reworking on granules

jdenquin (approved):

PR #2864, opened 2024-11-25
feat(salesforce): Fill IntegrationCustomer on salesforce integration 

brunomiguelpinto:
## Context

This PR introduces the logic to handle the IntegrationCustomer association specifically for Salesforce integrations. It ensures that Salesforce-related IntegrationCustomer records are processed synchronously since we don't need to reach a provider on this case and we need the response of the IntegrationCustomer on the serializer ( on customer create and update )

### DIFF
--- a/app/services/customers/update_service.rb
+++ b/app/services/customers/update_service.rb
@@ -148,7 +148,7 @@ module Customers
 
       IntegrationCustomers::CreateOrUpdateService.call(
         integration_customers: args[:integration_customers],
-        customer: result.customer,
+        customer: customer,
         new_customer: false
       )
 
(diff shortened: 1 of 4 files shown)


app/services/customers/update_service.rb
```diff
@@ -146,13 +146,13 @@ def call
         update_result.raise_if_error!
       end
 
-      result.customer = customer
-
       IntegrationCustomers::CreateOrUpdateService.call(
         integration_customers: args[:integration_customers],
-        customer: result.customer,
+        customer: customer,
```
nudded, line 151: you can align both calls to this service perhaps?

if you're passing in `customer` you can simply drop it:

```
.call(..
customer:,
...)
```

and end with `result.customer.reload` ?

nudded (approved):

PR #2887, opened 2024-11-27
feat(salesforce): Add create customer without external id

brunomiguelpinto:
## Context

Add the ability to create a new salesforce customer without passing the external customer id so it can be filled later by the salesforce integration

Screenshot 2024-11-27 at 7 11 54 PM

### DIFF
--- a/app/services/integration_customers/factory.rb
+++ b/app/services/integration_customers/factory.rb
@@ -16,6 +16,8 @@ module IntegrationCustomers
         IntegrationCustomers::XeroService
       when 'Integrations::HubspotIntegration'
         IntegrationCustomers::HubspotService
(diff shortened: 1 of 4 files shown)


app/services/integration_customers/factory.rb
```diff
@@ -16,6 +16,8 @@ def self.service_class(integration)
         IntegrationCustomers::XeroService
       when 'Integrations::HubspotIntegration'
         IntegrationCustomers::HubspotService
+      when 'Integrations::SalesforceIntegration'
```
ivannovosad, line 19: We can also add it to the factory spec.

ivannovosad (changes requested):

ivannovosad (approved):

PR #2890, opened 2024-11-28
feat(dunning): Remove dunning_campaign_completed flag

rsempe:
## Roadmap

 ## Context

 ## Description

The goal of this PR is to remove the flag `dunning_campaign_completed` on customers.
We don't need this complexity and prefer retrying the dunning process when number of attempts is changed.

### DIFF
--- a/app/services/dunning_campaigns/process_attempt_service.rb
+++ b/app/services/dunning_campaigns/process_attempt_service.rb
@@ -16,6 +16,7 @@ module DunningCampaigns
       return result unless applicable_dunning_campaign?
       return result unless dunning_campaign_threshold_reached?
       return result unless days_between_attempts_passed?
+      return result if max_attempts_reached?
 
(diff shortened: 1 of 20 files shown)


app/services/dunning_campaigns/process_attempt_service.rb
```diff
@@ -16,7 +16,6 @@ def call
       return result unless applicable_dunning_campaign?
       return result unless dunning_campaign_threshold_reached?
       return result unless days_between_attempts_passed?
-      return result if customer.dunning_campaign_completed?
```
ancorcruz, line 19: we still need to check if the max attempts has been reached here...

please revert the changes in this commit:

https://github.com/getlago/lago-api/pull/2889/commits/0d72019ad0625f36c332bb53f167e7c77f57a5c1

ancorcruz (changes requested):
looks good. thanks.

PR #2900, opened 2024-12-02
feat(api-key-management): Use permissions defined by API key

floganz:
## Context

Switch to using defined by API key permissions in case when organization has `api_permissions` premium integration.

## Description

Use API key defined permissions for API.
Add shared example for easier permission testing in request specs.


vincent-pochet (commented):
It looks great :+1:

vincent-pochet (approved):
Great one!

PR #2906, opened 2024-12-03
Fix: progressive billing invoice for subscription period

annvelents:
## Context

because invoices are billed after the billing period, invoice_subscription.timestamp would contain date that is after the period, as result `Subscriptions::ProgressiveBilledAmount` might return a progressive billing invoice associated to the next billing period

## Description

Changed the timestamp sent to `Subscriptions::ProgressiveBilledAmount`
Added scenario tests


vincent-pochet (approved):
:rocket:

PR #2909, opened 2024-12-03
Fix: greater_than usage

annvelents:
## Context

Credit Note Item model allows amount_cents to be 0; Because of roundings, applied_coupon.amount_cents can also be 0

### DIFF
--- a/app/services/credit_notes/validate_item_service.rb
+++ b/app/services/credit_notes/validate_item_service.rb
@@ -53,9 +53,9 @@ module CreditNotes
       false
     end
 
-    # NOTE: Check if item amount is >= 0 (CreditNoteItem.amount_cents has numericality validation: greater_than_or_equal_to: 0)
+    # NOTE: Check if item amount is positive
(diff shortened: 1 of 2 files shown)


floganz (approved):

app/services/credit_notes/validate_item_service.rb
```diff
@@ -53,9 +53,9 @@ def valid_fee?
       false
     end
 
-    # NOTE: Check if item amount is positive
+    # NOTE: Check if item amount is >= 0 (CreditNoteItem.amount_cents has numericality validation: greater_than_or_equal_to: 0)
```
nudded, line 56: Ha! We both found this out independently, A fix for this line has already been merged :)

nudded (approved):

app/services/credit_notes/validate_item_service.rb
annvelents: nooooice
will cancel this change then 😁

PR #2910, opened 2024-12-03
feat(daily-usage): Compute final usage for terminated subscriptions

vincent-pochet:
This PR makes sure that the final usage is computed for terminated subscriptions.
It also changes the `usage_diff` behavior for the first `daily_usage` by setting an empty JSON object rather than the same value as the `usage` field

### DIFF
--- a/app/services/daily_usages/compute_service.rb
+++ b/app/services/daily_usages/compute_service.rb
@@ -47,6 +47,8 @@ module DailyUsages
     end
 
     def existing_daily_usage
+      return nil if subscription.terminated_at?(timestamp)
+
       @existing_daily_usage ||= DailyUsage.refreshed_at_in_timezone(timestamp)
         .find_by(subscription_id: subscription.id)
     end
--- a/app/services/daily_usages/compute_service.rb
+++ b/app/services/daily_usages/compute_service.rb
@@ -9,6 +9,11 @@ module DailyUsages
     end
 
     def call
+      if subscription_billing_day?
+        # Usage on billing day will be computed using the periodic invoice as we cannont rely on the caching mechanism
+        return result
+      end
+
       if existing_daily_usage.present?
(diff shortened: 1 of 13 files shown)


app/services/daily_usages/compute_service.rb
```diff
@@ -47,6 +47,8 @@ def current_usage
     end
 
     def existing_daily_usage
+      return nil if subscription.terminated?
```
nudded, line 50: just wondering If it's important to pass in a timestamp here? 🤔

app/services/daily_usages/fill_history_service.rb
```diff
@@ -56,6 +56,10 @@ def call
         end
       end
 
+      if subscription.terminated?
```
nudded, line 59: Same concern as above :)

vincent-pochet: Here we are fine as the `to` value in the loop is `(subscription.terminated_at || Time.current).to_date` of if the subscription is terminated, we have to compute the value for the termination day

app/services/daily_usages/compute_service.rb
```diff
@@ -54,5 +71,17 @@ def existing_daily_usage
     def diff_usage(daily_usage)
       DailyUsages::ComputeDiffService.call!(daily_usage:).usage_diff
     end
+
+    def subscription_billing_day?
+      date_in_timezone = timestamp.in_time_zone(customer.applicable_timezone).to_date
```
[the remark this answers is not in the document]
vincent-pochet: `date_in_timezone == previous_billing_date_in_timezone` :slightly_smiling_face:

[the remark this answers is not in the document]
annvelents: 🤦‍♀️😆 
Thank you!

annvelents (approved):

PR #2912, opened 2024-12-04
Feat: invoice custom section create service

annvelents:
## Context

Service to create and select for organization invoice custom section

### DIFF
--- a/app/services/invoice_custom_sections/select_service.rb
+++ b/app/services/invoice_custom_sections/select_service.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module InvoiceCustomSections
+  class SelectService < BaseService
+    def initialize(section:, organization:)
+      @section = section
+      @organization = organization
+      super
+    end
+
+    def call
+    end
+
+    private
+
+    attr_reader :section, :organization
+  end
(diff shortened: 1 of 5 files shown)


app/services/invoice_custom_sections/select_service.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module InvoiceCustomSections
+  class SelectService < BaseService
+    def initialize(section:, organization:)
```
floganz, line 5: Does this service really need `organization` as input parameter?

`InvoiceCustomSection` has `belongs_to :organization` association and we can relay on it, simplify this service by removing `if organization` checks and also eliminating issue of potential assigning section from one organization to another.

floganz (approved):

app/services/invoice_custom_sections/select_service.rb
annvelents: yep, we need it: we're going to introduce also selections for customers as well, because an invoice_custom_section can be not selected for the organization, but selected for a customer (when customer does not follow organization selection)

vincent-pochet (approved):

PR #2920, opened 2024-12-06
Feat: invoice_custom_sections query resolver

annvelents:
## Context

Next step in implementing invoice custom sections - allow FE to query invoice_custom_sections. Added InvoiceCustomSectionsResolver

### DIFF
--- a/app/graphql/resolvers/invoice_custom_sections_resolver.rb
+++ b/app/graphql/resolvers/invoice_custom_sections_resolver.rb
@@ -19,8 +19,6 @@ module Resolvers
         Arel.sql('CASE WHEN invoice_custom_section_selections.id IS NOT NULL THEN 0 ELSE 1 END'),
         :name
       ).page(page).per(limit)
-    rescue ActiveRecord::RecordNotFound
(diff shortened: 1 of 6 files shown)


app/graphql/resolvers/invoice_custom_sections_resolver.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    type Types::InvoiceCustomSections::Object.collection_type, null: true
+
+    def resolve(page: nil, limit: nil)
+      current_organization.invoice_custom_sections.left_outer_joins(:invoice_custom_section_selections).order(
+        Arel.sql('CASE WHEN invoice_custom_section_selections.id IS NOT NULL THEN 0 ELSE 1 END'),
+        :name
+      ).page(page).per(limit)
+    rescue ActiveRecord::RecordNotFound
+      not_found_error(resource: 'invoice')
```
ancorcruz, line 23: what raises this exception? there is not `find` for the invoice resource.

annvelents: oops, you're right. Thank you!

PR #2928, opened 2024-12-09
fix (grace_period): refactor how issuing date is set

lovrocolic:
## Context

When billing job is triggered in the past, `issuing_date` is not set correctly.

## Description

`issuing_date` is set based on the billing job timestamp and grace period.

However, when we check if it is correct time to finalize invoice or in services that set grace period we calculate issuing date again based on `invoice.created_at`. If `invoice.created_at` is different than billing day timestamp, we could end-up with invalid issuing date. This especially affects when the draft invoice is finalized.

@vincent-pochet @rsempe Can you see any downside of this approach?

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -99,14 +99,13 @@ class Invoice < ApplicationRecord
   scope :ready_to_be_refreshed, -> { where(ready_to_be_refreshed: true) }
   scope :ready_to_be_finalized,
     lambda {
-      date = <<-SQL
-            (
-              invoices.created_at +
-              COALESCE(customers.invoice_grace_period, organizations.invoice_grace_period) * INTERVAL '1 DAY'
-            )
-      SQL
-
-      draft.joins(:customer, :organization).where("#{Arel.sql(date)} < ?", Time.current)
+      draft
+        .joins(customer: :organization)
+        .where(
+          "issuing_date#{Utils::Timezone.at_time_zone_sql} <= " \
+          "DATE(?#{Utils::Timezone.at_time_zone_sql})",
+          Time.current
+        )
     }
 
   scope :created_before,
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -97,16 +97,7 @@ class Invoice < ApplicationRecord
   scope :invisible, -> { where(status: INVISIBLE_STATUS.keys) }
   scope :with_generated_number, -> { where(status: %w[finalized voided]) }
   scope :ready_to_be_refreshed, -> { where(ready_to_be_refreshed: true) }
-  scope :ready_to_be_finalized,
-    lambda {
-      draft
-        .joins(customer: :organization)
-        .where(
-          "issuing_date#{Utils::Timezone.at_time_zone_sql} <= " \
-          "DATE(?#{Utils::Timezone.at_time_zone_sql})",
-          Time.current
-        )
-    }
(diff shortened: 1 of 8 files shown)


app/models/invoice.rb
```diff
@@ -99,14 +99,13 @@ class Invoice < ApplicationRecord
...
     lambda {
-      date = <<-SQL
-            (
-              invoices.created_at +
-              COALESCE(customers.invoice_grace_period, organizations.invoice_grace_period) * INTERVAL '1 DAY'
-            )
-      SQL
-
-      draft.joins(:customer, :organization).where("#{Arel.sql(date)} < ?", Time.current)
+      draft
```
lovrocolic, line 102: before we haven't used customer timezone in this scope

app/models/invoice.rb
```diff
@@ -99,14 +99,13 @@ class Invoice < ApplicationRecord
...
+        .joins(customer: :organization)
```
nudded, line 103: no need to join anymore.

lovrocolic: I think it is needed for the timezone part. We need to fetch timezone there

nudded: ah right 👍

app/models/invoice.rb
vincent-pochet: issuing date is defined like this in the `/app/services/invoices/create_generating_service.rb`

```ruby
    def issuing_date
      date = datetime.in_time_zone(customer.applicable_timezone).to_date
      return date if !grace_period? || charge_in_advance

      date + customer.applicable_invoice_grace_period.days
    end
```

it means that it is technically already a date expressed in the customer timezone.
Maybe I'm wrong, but I don't think we need to do the timezone shift again.

That said, maybe a comment would help future us reading this scope and asking themselves about it :smile:

vincent-pochet (commented):
Looks good! Just a comment regarding the timezone in the scope.

vincent-pochet (approved):

app/services/customers/update_invoice_grace_period_service.rb
```diff
@@ -4,27 +4,29 @@ module Customers
...
+      old_grace_period = customer.invoice_grace_period.to_i
+      grace_period_diff = grace_period - old_grace_period
+
+      if grace_period != old_grace_period
         customer.invoice_grace_period = grace_period
         customer.save!
 
+        # NOTE: Update issuing_date on draft invoices.
+        customer.invoices.draft.each do |invoice|
+          invoice.issuing_date = invoice.issuing_date + grace_period_diff.days
```
nudded, line 21: what happens if the issuing_date moves into the past here?

app/models/invoice.rb
lovrocolic: Actually you are right! Isuuing date is already converted to customer timezone and this part should not be converted!
However I also wanted initially to convert server time in customer timezone so that we include certain invoice for finalization in correct moment. However, after discussion with Mike we concluded that we should rather not change current behaviour and I am not using any timezone shift now. E.g. server time is shifted also in job that terminates invoices based on ending_at and I wanted to follow same pattern here.

app/services/customers/update_invoice_grace_period_service.rb
nudded: Yeah, it's fine as long as we keep the `<=` in the `ready_to_be_finalized` scope, I think 👍

PR #2930, opened 2024-12-09
misc(payment): Refact payment enqueue

vincent-pochet:
## Context

This PR is the first step of the refactor of the payment processing.

- **Avoid double payment at all cost**

## Description

This PR:
- Rename `Invoices::Payments::CreateService#call` into `Invoices::Payments::CreateService#call_async` as the method only enqueue a payment processing job.
- Add a new `Invoices::Payments::CreateJob`
- Add a new definition to `Invoices::Payments::CreateService#call` to delegate logic to payment processor services

In a next step this last method will be updated to create a payment before delegating the payment processing to the providers and to ensure that no processing payment already exists.


rsempe (approved):
👍

annvelents (approved):
🚀

PR #2931, opened 2024-12-09
Feat: invoice custom sections - discard service

annvelents:
## Context

A Service to destroy invoice custom sections

## Description

As now we have another mechanism of deselecting invoice custom sections, it makes sense to extract deselect_for_all_usages service into their own service and extract the deselect services into their namespace


app/services/invoice_custom_sections/deselect/for_organization_service.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module InvoiceCustomSections
+  module Deselect
```
nudded, line 4: not sure if there's a need for a `Deselect` module 🤔 

I don't particularly mind either way, so just a nitpick.

nudded (approved):

app/services/invoice_custom_sections/deselect/for_organization_service.rb
annvelents: thank you, it actually is refactored in one of the further PRs 🙈

PR #2934, opened 2024-12-10
fix: Ensure consistent order of results on list responses (frontend and API)

ancorcruz:
## Context

Some endpoints like `GET /api/v1/credit_notes`  provide duplicate results on pagination as created_at is not accurate enough to effectively order resources by itself.

## Description

This change updates all API/v1 controllers to use query objects to retrieve records for the `GET / (index)` endpoints. Also, apply default ordering by created_at :desc and id :asc to ensure records stay in their position when going through pages.

Query classes specs are re-styled to the current rules we are following.

### DIFF
--- a/app/queries/events_query.rb
+++ b/app/queries/events_query.rb
@@ -5,9 +5,9 @@ class EventsQuery < BaseQuery
     events = organization.clickhouse_events_store? ? Clickhouse::EventsRaw : Event
     events = events.where(organization_id: organization.id)
     events = paginate(events)
+
     events = events.order(timestamp: :desc) unless organization.clickhouse_events_store?
     events = events.order(ingested_at: :desc) if organization.clickhouse_events_store?
-    events = apply_consistent_ordering(events)
 
     events = with_code(events) if filters.code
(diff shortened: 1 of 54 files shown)


app/queries/events_query.rb
```diff
@@ -5,9 +5,9 @@ def call
     events = organization.clickhouse_events_store? ? Clickhouse::EventsRaw : Event
     events = events.where(organization_id: organization.id)
     events = paginate(events)
-
     events = events.order(timestamp: :desc) unless organization.clickhouse_events_store?
     events = events.order(ingested_at: :desc) if organization.clickhouse_events_store?
+    events = apply_consistent_ordering(events)
```
vincent-pochet, line 10: In that case, we are already ordering the events in the previous lines (depending on the event store configured at organization level), we should not reapply any other ordering as no index is covering this correctly and those one have a huge number of rows

Also, if the event store is clickhouse, the table does not have an ID column.
It shows that a test case covering this is missing :D

vincent-pochet (approved):
Great one!

Just one comment regarding the event endpoint as it is a pretty sensitive one

PR #2941, opened 2024-12-11
fix: update regrouped invoice pdf with percentage details on fees

lovrocolic:
## Context

When charge has filters, fee details are not displayed for percentage charge model in the invoice PDF

## Description

This PR ensures that fee details are properly displayed in the invoice PDF

### DIFF
--- a/app/views/templates/invoices/v4/charge.slim
+++ b/app/views/templates/invoices/v4/charge.slim
@@ -456,7 +456,7 @@ html
           - fees.order(:succeeded_at, :created_at).each do |fee|
             tr
               - if fee.charge.percentage? && fee.amount_details.present?
-                - if fee.charge.basic_rate_percentage?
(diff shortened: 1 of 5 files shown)


app/views/templates/invoices/v4/charge.slim
```diff
@@ -456,7 +456,7 @@ html
           - fees.order(:succeeded_at, :created_at).each do |fee|
             tr
               - if fee.charge.percentage? && fee.amount_details.present?
-                - if fee.charge.basic_rate_percentage?
+                - if fee.basic_rate_percentage?
```
rsempe, line 459: Can we remove `Charge#basic_rate_percentage?` method maybe?

lovrocolic: Removed 👍

rsempe (approved):

PR #2948, opened 2024-12-12
misc: Remove Fee#{charge|subscription|commitment}_kind scope

rsempe:


app/graphql/types/invoice_subscription/object.rb
```diff
@@ -39,7 +39,7 @@ def should_use_in_advance_charges_interval
         return @should_use_in_advance_charges_interval if defined? @should_use_in_advance_charges_interval
 
         @should_use_in_advance_charges_interval =
-          object.fees.charge_kind.any? &&
+          object.fees.charge.any? &&
```
annvelents, line 42: I love the removing of the scope since we have enum for this, but personally for me having plain `charge` is slightly confusing because fee can belong to charge or subscription, and when i see `fees.charge`, I'm thinking about `fee.charge`, so I'm expecting it to return charge, and not fees 🙈 
What do you think if we add a prefix to the enum and use `fee.charge_type`:
``` 
 enum fee_type: FEE_TYPES, _prefix: :type
 ```
but at the same time I'm calling them myself "charge fee, subscription fee", so I'm totally fine with having `fees.charge` as well 😁

annvelents (approved):

app/graphql/types/invoice_subscription/object.rb
rsempe: Thanks @annvelents - Let's maybe try without the type prefix, and we can change it in the future if better ;)

PR #2959, opened 2024-12-13
misc(payment): Refactor payment creation on provider

vincent-pochet:
## Context

This PR follow https://github.com/getlago/lago-api/pull/2930 and https://github.com/getlago/lago-api/pull/2935 in the refactor of the payment processing.

- Avoid double payment at all cost

## Description

This PR:
- Extract the payment creation logic from `Invoices::Payments::*Service#call` into dedicated services in `PaymentProviders::*::Payments::CreateService#call`. These services are only responsible for the payment creation itself and the related error handling
- Extract webhook deliveries, invoice payment status handling and integration notification in from each provider services in the main `Invoices::Payments::CreateService`

In the next step this last service will be updated to create a payment before delegating the payment processing to the providers and to ensure that no processing payment already exists.

### DIFF
--- a/app/services/invoices/payments/gocardless_service.rb
+++ b/app/services/invoices/payments/gocardless_service.rb
@@ -5,19 +5,6 @@ module Invoices
     class GocardlessService < BaseService
       include Customers::PaymentProviderFinder
 
-      class MandateNotFoundError < StandardError
-        DEFAULT_MESSAGE = "No mandate available for payment"
-        ERROR_CODE = "no_mandate_error"
-
-        def initialize(msg = DEFAULT_MESSAGE)
-          super
-        end
-
-        def code
-          ERROR_CODE
-        end
-      end
-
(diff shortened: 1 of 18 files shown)


app/services/invoices/payments/gocardless_service.rb
```diff
@@ -30,51 +17,12 @@ def initialize(invoice = nil)
...
-      rescue GoCardlessPro::Error, GoCardlessPro::ValidationError => e
-        deliver_error_webhook(e)
-        update_invoice_payment_status(payment_status: :failed, deliver_webhook: false)
-
-        if e.is_a?(GoCardlessPro::ValidationError)
-          result
-        else
-          raise
-        end
+        # Removed...
```
nudded, line 20: can you comment as to why you're keeping the method? 🤔

lovrocolic (approved):

PR #2960, opened 2024-12-13
Feat (invoice_custom_sections):  customer update service

annvelents:
## Description

Added service to manage invoice_custom_sections on customer level


app/services/customers/update_invoice_custom_sections_service.rb
```diff
@@ -0,0 +1,43 @@
...
+    attr_reader :customer, :section_ids
+
+    def organization
+      @organization ||= customer.organization
+    end
+
+    def assign_organization_sections
+      # Note: when inheriting organization's selections, customer shouldn't have their selected sections
+      customer.selected_invoice_custom_sections = []
+    end
```
ancorcruz, line 37: What about if the customer selected custom sections are the same as the organization by coincidence but the user expect to not change them if the organization one changes? This use case is not feasible with this implementation.

What about having a flag to know when the customer should inherit from organization or have a custom selection?

app/services/organizations/deselect_invoice_custom_section_service.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    private
+
+    attr_reader :section, :organization
+
+    def deselect_for_organization
+      return unless organization.selected_invoice_custom_sections.include?(section)
+
+      InvoiceCustomSectionSelection.where(
+        organization_id: organization.id, invoice_custom_section_id: section.id
```
ancorcruz, line 24: does the `organization_id` here represent this section is applied to the organization?

this is a little bit confusing as we use organization_id for multi tenancy purposes... I would go for a `applied_to_organization` flag instead...

ancorcruz (commented):
I have left a few comments worth reviewing.

app/services/customers/update_service.rb
```diff
@@ -203,5 +205,17 @@ def applied_dunning_campaign
...
+    def manage_invoice_custom_sections
+      if args.key?(:skip_invoice_custom_sections)
+        customer.selected_invoice_custom_sections = [] if args[:skip_invoice_custom_sections]
+        customer.skip_invoice_custom_sections = args[:skip_invoice_custom_sections]
+      end
+
+      if args.key?(:selected_invoice_custom_section_ids)
+        customer.skip_invoice_custom_sections = false
+        UpdateInvoiceCustomSectionsService.call(customer:, section_ids: args[:selected_invoice_custom_section_ids])
+      end
```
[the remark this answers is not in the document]
annvelents: It's a good question, I thought  it's okay to override, but actually validation error sounds better...
yep, I'll add a validation error

app/services/customers/update_invoice_custom_sections_service.rb
annvelents: you're absolutely right. Actually further with the FE we agreed to send empty array when customer follows organization's selection, forgot about this piece. thank you!

app/services/organizations/deselect_invoice_custom_section_service.rb
annvelents: not exactly, the `organization_id` represents belonging to an organization, as organization can have multiple `invoice_custom_sections`, but only few of them will be selected on the organization level, and others (and the same) might be selected on the customer level

annvelents: and these invoice_custom_section_selections are exactly serving the purpose of choosing an invoice_custom_section for the organization or for the customer

spec/services/customers/manage_invoice_custom_sections_service_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      it 'raises an error' do
+        result = service.call
+        expect(result).not_to be_success
+        expect(result.error).to be_a(BaseService::ValidationFailure)
+        expect(result.error.message).to include('skip_sections_and_selected_ids_sent_together')
+      end
+    end
+
+    context 'when updating selected_invoice_custom_sections' do
+      context 'when section_ids match customer\'s applicable sections' do
```
ancorcruz, line 42: for future reference, our preferred string delimiter is double quotes instead of single quotes, using double quotes won't require to escape the apostrophe character here.

spec/services/customers/manage_invoice_custom_sections_service_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      end
+
+      context 'when setting invoice_custom_sections_ids when previously customer had skip_invoice_custom_sections' do
+        let(:section_ids) { [] }
+
+        before { customer.update(skip_invoice_custom_sections: true) }
+
+        it 'sets skip_invoice_custom_sections to false' do
+          service.call
+          expect(customer.reload.skip_invoice_custom_sections).to be_falsey
```
ancorcruz, line 89: here we want to ensure the value is `false` not falsey (nil evaluates as falsey for example)

```suggestion
          expect(customer.reload.skip_invoice_custom_sections).to eq false
```

spec/services/organizations/deselect_invoice_custom_section_service_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  describe '#call' do
+    subject(:service_result) { described_class.call(section:) }
+
+    let(:organization) { create(:organization) }
+    let(:section) { create(:invoice_custom_section, organization:) }
+
+    context 'when section is not selected for the organization' do
+      it 'selects the section for the organization' do
+        expect(service_result).to be_success
+      end
```
ancorcruz, line 15: does it just return success? there is no other change?

spec/services/organizations/deselect_invoice_custom_section_service_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    context 'when section is selected for the organization' do
+      before do
+        organization.selected_invoice_custom_sections << section
+      end
+
+      it 'deselects the section for the organization' do
+        expect { service_result }.to change(organization.selected_invoice_custom_sections, :count).by(-1)
+        expect(organization.invoice_custom_sections).to include(section)
```
ancorcruz, line 25: Could we add the expectation of the service_result to be  success?

```suggestion
        expect { service_result }.to change(organization.selected_invoice_custom_sections, :count).by(-1)
        expect(organization.invoice_custom_sections).to include(section)
        expect(service_result).to be_success
```

ancorcruz (approved):
LGTM!

just a few comments

PR #2963, opened 2024-12-16
misc(export-credit-notes): Refactor services and tests

floganz:
## Context

Refactor and preparations to expand existing data export solution with new export types.

## Description

Extract common logic into a new `BaseCsvService`.
Refactor tests.

### DIFF
--- a/app/mailers/data_export_mailer.rb
+++ b/app/mailers/data_export_mailer.rb
@@ -6,7 +6,6 @@ class DataExportMailer < ApplicationMailer
     @resource_type = @data_export.resource_type.humanize.downcase
     user = @data_export.user
 
-    return if @data_export.file.blank?
     return if @data_export.expired?
     return unless @data_export.completed?
 
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -7,19 +7,16 @@ class DataExport < ApplicationRecord
 
   belongs_to :organization
   belongs_to :membership
+  has_one :user, through: :membership
 
   has_one_attached :file
 
   has_many :data_export_parts
 
+  enum :format, EXPORT_FORMATS, validate: true
+  enum :status, STATUSES, validate: true
+
   validates :resource_type, presence: true
-  validates :format, presence: true, inclusion: {in: EXPORT_FORMATS}
-  validates :status, presence: true, inclusion: {in: STATUSES}
-
-  enum format: EXPORT_FORMATS
-  enum status: STATUSES
-
-  delegate :user, to: :membership
 
   def processing!
     update!(status: 'processing', started_at: Time.zone.now)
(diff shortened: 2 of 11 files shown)


app/models/data_export.rb
```diff
@@ -7,19 +7,17 @@ class DataExport < ApplicationRecord
...
   belongs_to :membership
+  has_one :user, through: :membership
 
   has_one_attached :file
 
   has_many :data_export_parts
 
-  validates :resource_type, presence: true
-  validates :format, presence: true, inclusion: {in: EXPORT_FORMATS}
-  validates :status, presence: true, inclusion: {in: STATUSES}
```
floganz, line 17: These validations were not tested and in fact they don't even have change to work because of specific to enum logic.
```ruby
irb(main):003> data_export = DataExport.new => #<DataExport:0x0000ffff5a6eed80
irb(main):004> data_export.status = 'no-existing-one'
(irb):4:in `<main>': 'no-existing-one' is not a valid status (ArgumentError)

          raise ArgumentError, "'#{value}' is not a valid #{name}"
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

With suggested changes
```ruby
irb(main):001> data_export = DataExport.new => #<DataExport:0x0000ffff7f9b5648
irb(main):002> data_export.status = 'no-existing-one'
=> "no-existing-one"
irb(main):003> data_export.valid?
=> false
irb(main):004> data_export.errors.as_json
=> 
{:organization=>["relation_must_exist"],
 :membership=>["relation_must_exist"],
 :format=>["value_is_invalid"],
 :status=>["value_is_invalid"],
 :resource_type=>["value_is_mandatory"]}
```

app/mailers/data_export_mailer.rb
```diff
@@ -6,7 +6,6 @@ def completed
     @resource_type = @data_export.resource_type.humanize.downcase
     user = @data_export.user
 
-    return if @data_export.file.blank?
```
floganz, line 9: I think this should be part of model logic, `completed` status should require a file to be present.
The mailer should know less about data export internals.

app/services/data_exports/csv/invoice_fees.rb
```diff
@@ -96,7 +46,55 @@ def self.headers
...
+          serialized_fee[:total_amount_currency],
+          serialized_fee[:units],
+          serialized_fee[:precise_unit_amount],
+          serialized_fee[:taxes_amount_cents],
+          serialized_fee[:total_amount_cents]
+        ]
+      end
+
+      def collection
+        Fee
```
vincent-pochet, line 85: This could loads a lot of records, this is why it's done per invoice bases, we have memory issues in the past.

vincent-pochet (commented):
It looks good, just a comment regarding the record loading as it could result in memory issues

app/services/data_exports/csv/invoice_fees.rb
floganz: I've reverted these changes to keep original approach. Not ideal but from performance standpoint should be the same.

vincent-pochet (approved):
Nice one!

PR #2964, opened 2024-12-16
feat(anrok): add new tax related statuses on invoice

lovrocolic:
## Context

Currently Anrok calls are sync and performed during invoice generation

## Description

The goal of this improvement is to make calls to Anrok in dedicated job so that we can implement throttling for rate limit issues.

This PR adds necessary DB changes

### DIFF
--- a/db/migrate/20241216140931_add_tax_status_to_invoices.rb
+++ b/db/migrate/20241216140931_add_tax_status_to_invoices.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddTaxStatusToInvoices < ActiveRecord::Migration[7.1]
+  def change
+    add_column :invoices, :tax_status, :integer, null: false, default: 1
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrate/20241216140931_add_tax_status_to_invoices.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddTaxStatusToInvoices < ActiveRecord::Migration[7.1]
+  def change
+    add_column :invoices, :tax_status, :integer, null: false, default: 1
```
vincent-pochet, line 5: What do you think of using an enum type here? (using `create_enum`)

vincent-pochet (approved):

db/migrate/20241216140931_add_tax_status_to_invoices.rb
lovrocolic: Updated

PR #2969, opened 2024-12-17
Misc: queries optimisation for refresh service

annvelents:


app/jobs/clock/finalize_invoices_job.rb
```diff
@@ -9,7 +9,7 @@ class FinalizeInvoicesJob < ApplicationJob
     unique :until_executed, on_conflict: :log
 
     def perform
-      Invoice.ready_to_be_finalized.find_each do |invoice|
+      Invoice.ready_to_be_finalized.includes(:credit_notes).find_each do |invoice|
```
nudded, line 12: How does this include help? 🤔

annvelents: inside `FinlizeJob` we pass this invoice to `Invoices::RefreshDraftAndFinalizeService`, where there we're passing it further to `RefreshDraftService`, where finally we're going through 
```
        cn_subscription_ids = invoice.credit_notes.map do |cn|
          {credit_note_id: cn.id, subscription_id: cn.fees.pick(:subscription_id)}
        end
```
and actually in addition to the credit notes we can also preload fees... 🤔

nudded: but the inclusion here does nothing right? we load the invoice again in that job?

db/migrate/20241217120924_add_indexes_to_invoices.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddIndexesToInvoices < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :invoices, :ready_to_be_refreshed, where: "ready_to_be_refreshed", algorithm: :concurrently
```
nudded, line 7: I'd be explicit `where: "ready_to_be_refreshed = true"`

app/models/invoice.rb
```diff
@@ -105,7 +105,7 @@ class Invoice < ApplicationRecord
   scope :visible, -> { where(status: VISIBLE_STATUS.keys) }
   scope :invisible, -> { where(status: INVISIBLE_STATUS.keys) }
   scope :with_generated_number, -> { where(status: %w[finalized voided]) }
-  scope :ready_to_be_refreshed, -> { where(ready_to_be_refreshed: true) }
+  scope :ready_to_be_refreshed, -> { draft.where(ready_to_be_refreshed: true) }
```
nudded, line 108: Definitely need a spec for this scope I think

nudded (approved):

PR #2971, opened 2024-12-17
chore(scalability): Add dedicated clock worker

jdenquin:
- Clock job should not impact other queues
- Adding a dedicated clock worker

### DIFF
--- a/config/sidekiq/sidekiq_clock.yml
+++ b/config/sidekiq/sidekiq_clock.yml
@@ -2,7 +2,7 @@ concurrency: 10
 timeout: 25
 retry: 1
 queues:
-  - clock
+  - clock_worker
 
 production:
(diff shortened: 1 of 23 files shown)


annvelents (approved):

config/sidekiq/sidekiq_clock.yml
```diff
@@ -0,0 +1,10 @@
+concurrency: 10
+timeout: 25
+retry: 1
+queues:
+  - clock
+
+production:
+  concurrency: <%= ENV.fetch('SIDEKIQ_CONCURRENCY', 5) %>
```
nudded, line 8: hmm, how do you feel about a dedicated clock concurrency variable?

jdenquin: since its for one worker, each worker has his own env var, do you have something else in mind?

nudded (approved):

PR #2972, opened 2024-12-17
fix: Persist inbound stripe webhooks

ancorcruz:
## Context

Currently, when our application receives webhooks from third-party services (e.g., Stripe, GoCardless, Adyen), their payloads are immediately queued in Sidekiq for background processing. While this approach works in most cases, it introduces a critical reliability issue, especially in self-hosted environments where Sidekiq may lose jobs if workers are abruptly terminated or overloaded.

This unreliability has led to issues like pending transactions that are never resolved despite receiving valid webhooks. For example, a reported issue involved a wallet credit purchase where the payment settled on Stripe, the webhook was received, but the transaction remained pending indefinitely until manually resolved.

To address this, we aim to ensure that no webhook payload is lost and processing remains consistent, even under adverse conditions.

## Description

This PR introduces a robust mechanism for handling inbound webhooks by persisting them in the database before processing. Key changes include:

### Database Persistence:
* Webhooks are now stored in a new `inbound_webhooks` table immediately upon receipt. This ensures all webhook payloads are safely persisted in a reliable system before processing.

### Background Processing:
* A new `InboundWebhooks::CreateService` is introduced to persist the webhook and enqueue it for processing.
* A new `InboundWebhooks::ProcessJob` Sidekiq worker processes the persisted webhooks, invoking the `InboundWebhooks::ProcessService`
* A new `InboundWebhooks::ProcessService` processes the persisted webhooks invoking the appropriate payment provider service.

### Retry Mechanisms:
* Failed webhooks are marked as such and can be retried manually or automatically via a clock job.
* A cleanup job is introduced to remove processed webhook records older than 1 month.

### Controller Update:
* The `WebhooksController` actions (stripe, adyen, and gocardless) now use the InboundWebhooks::CreateService and respond with 200 OK once the webhook is persisted successfully.

### Backward Compatibility:
* No webhooks will be missed during the transition. The change is designed to integrate seamlessly into both cloud and self-hosted environments.

### Key Changes
* Added `InboundWebhook` ActiveRecord model to persist webhook payloads with fields such as `source`, `event_type`, `payload`, `status`.
* Implemented `InboundWebhooks::CreateService` to handle webhook persistence and Sidekiq job enqueueing.
* Added `InboundWebhooks::ProcessJob` to process persisted webhooks from Sidekiq queue.
* Added `InboundWebhooks::ProcessService` to call the respective provider's handle incoming webhook service.
* Updated `PaymentProviders::Stripe::HandleIncomingWebhookService` to process webhooks from the database instead of relying solely on direct Sidekiq queues.
* Updated `WebhooksController#stripe` action to integrate with the new service and workflow.
* Introduced a clock job for:
1. Retrying pending webhooks older than 2 hours.
2. Cleaning up processed webhooks older than 90 days.

### Benefits
**Reliability:** Ensures no webhook is lost, even if Sidekiq workers fail.
**Resilience:** Provides a retry mechanism for failed webhook processing.
**Transparency:** Makes webhook processing traceable with statuses (pending, processing, processed, failed).

### DIFF
--- a/app/jobs/clock/inbound_webhooks_retry_job.rb
+++ b/app/jobs/clock/inbound_webhooks_retry_job.rb
@@ -7,10 +7,7 @@ module Clock
     queue_as "clock"
 
     def perform
-      InboundWebhook
-        .where(status: ["pending", "processing"])
-        .where("updated_at < ?", 2.hours.ago)
-        .find_each do |inbound_webhook|
+      InboundWebhook.retriable.find_each do |inbound_webhook|
         InboundWebhooks::ProcessJob.perform_later(inbound_webhook:)
       end
     end
--- a/app/jobs/clock/inbound_webhooks_retry_job.rb
+++ b/app/jobs/clock/inbound_webhooks_retry_job.rb
@@ -4,10 +4,13 @@ module Clock
   class InboundWebhooksRetryJob < ApplicationJob
     include SentryCronConcern
 
-    queue_as 'clock'
+    queue_as "clock"
 
     def perform
-      InboundWebhook.failed.find_each do |inbound_webhook|
+      InboundWebhook
+        .where(status: ["pending", "processing"])
+        .where("updated_at < ?", 12.hours.ago)
+        .find_each do |inbound_webhook|
         InboundWebhooks::ProcessJob.perform_later(inbound_webhook:)
       end
     end
--- a/app/jobs/clock/inbound_webhooks_retry_job.rb
+++ b/app/jobs/clock/inbound_webhooks_retry_job.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Clock
+  class InboundWebhooksRetryJob < ApplicationJob
+    include SentryCronConcern
+
+    queue_as 'clock'
+
+    def perform
+      InboundWebhook.failed.find_each do |inbound_webhook|
+        InboundWebhooks::ProcessJob.perform_later(inbound_webhook:)
+      end
+    end
+  end
+end
--- a/app/jobs/clock/inbound_webhooks_retry_job.rb
+++ b/app/jobs/clock/inbound_webhooks_retry_job.rb
@@ -9,7 +9,7 @@ module Clock
     def perform
       InboundWebhook
         .where(status: ["pending", "processing"])
-        .where("updated_at < ?", 12.hours.ago)
+        .where("updated_at < ?", 2.hours.ago)
         .find_each do |inbound_webhook|
         InboundWebhooks::ProcessJob.perform_later(inbound_webhook:)
       end
--- a/app/jobs/inbound_webhooks/process_job.rb
+++ b/app/jobs/inbound_webhooks/process_job.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module InboundWebhooks
+  class ProcessJob < ApplicationJob
+    queue_as :default
+
+    def perform(inbound_webhook:)
+      InboundWebhooks::ProcessService.call(inbound_webhook:).raise_if_error!
+    end
+  end
+end
--- a/app/jobs/inbound_webhooks/process_job.rb
+++ b/app/jobs/inbound_webhooks/process_job.rb
@@ -5,7 +5,7 @@ module InboundWebhooks
     queue_as :default
 
     def perform(inbound_webhook:)
-      InboundWebhooks::ProcessService.call(inbound_webhook:).raise_if_error!
+      InboundWebhooks::ProcessService.call!(inbound_webhook:)
     end
   end
 end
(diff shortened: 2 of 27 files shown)


app/jobs/clock/inbound_webhooks_retry_job.rb
```diff
@@ -0,0 +1,18 @@
...
+
+module Clock
+  class InboundWebhooksRetryJob < ApplicationJob
+    include SentryCronConcern
+
+    queue_as "clock"
+
+    def perform
+      InboundWebhook
+        .where(status: ["pending", "processing"])
```
nudded, line 11: Hmmmm, I think the retry should be a bit more laid back, if we do this every 15 minutes, when we have a large backlog, the queue will be quickly filled with jobs no?

app/services/inbound_webhooks/process_service.rb
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    def initialize(inbound_webhook:)
+      @inbound_webhook = inbound_webhook
+
+      super
+    end
+
+    def call
+      inbound_webhook.processing!
```
nudded, line 16: definitely need some precondition checking here I think, what if the web hook is already in processing/ processed?

app/jobs/clock/inbound_webhooks_retry_job.rb
ancorcruz: we could use the unique config to ensure the jobs are unique or touch the records here after queueing them to refresh the updated_at and avoid the job being queued more than once...

The logic behind "shorter" times to retry is that we expect webhooks to be something immediate... wait for 24 hours is too long. What do you think?

app/services/inbound_webhooks/process_service.rb
ancorcruz: yes, thanks for spotting... I am considering return if the status is processed or failed and let pending and processing continue.

The reason behind letting processing continue is if a sidekiq worker is killed while processing the webhook it will be in processing status...

what do you think?

app/jobs/clock/inbound_webhooks_retry_job.rb
nudded: yes, agreed, web hooks should be processed quickly. Perhaps it makes sense to store a retry count on the web hook? And we can skip web hooks that have ben retried an amount of time?

app/jobs/inbound_webhooks/process_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module InboundWebhooks
+  class ProcessJob < ApplicationJob
+    queue_as :default
+
+    def perform(inbound_webhook:)
+      InboundWebhooks::ProcessService.call(inbound_webhook:).raise_if_error!
```
vincent-pochet, line 8:
```suggestion
      InboundWebhooks::ProcessService.call!(inbound_webhook:)
```

app/services/inbound_webhooks/create_service.rb
```diff
@@ -0,0 +1,40 @@
...
+      inbound_webhook = InboundWebhook.create!(
+        organization_id:,
+        source: webhook_source,
+        code:,
+        payload:,
+        signature:,
+        event_type:
+      )
+
+      after_commit do
```
vincent-pochet, line 26: It looks like we are not relying on transaction right? So I guess this `after_commit` block is not mandatory

vincent-pochet (approved):
:raised_hands:

app/services/inbound_webhooks/create_service.rb
ancorcruz: this ensures the job is not added to the queue and tried to be processed until the DB transaction is committed. This will avoid race condition between sidekiq and postgres where sidekiq picks the job but it is unable to find the resource in DB.

app/models/inbound_webhook.rb
```diff
@@ -0,0 +1,51 @@
...
+    processing: "processing",
+    processed: "processed",
+    failed: "failed"
+  }
+
+  enum :status, STATUSES
+
+  scope :retriable, -> { reprocessable.or(old_pending) }
+  scope :reprocessable, -> { processing.where("processing_at <= ?", WEBHOOK_PROCESSING_WINDOW.ago) }
+  scope :old_pending, -> { pending.where("created_at <= ?", WEBHOOK_PROCESSING_WINDOW.ago) }
```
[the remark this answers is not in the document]
ancorcruz: add two combined indices [status, processing_at] and [status, created_at] should do it.

app/services/inbound_webhooks/create_service.rb
vincent-pochet: I agree, but here `InboundWebhooks::CreateService` is called directly from the controller and no transaction is created in it so the job will always be called after the InboundWebhook creation transaction.
That's why I was asking. No a major issue anyway ;)

vincent-pochet (approved):

julienbourdeau:
Really cool work, thank you!

PR #2973, opened 2024-12-17
feat(credit-notes-export): Add services for simple credit note export

floganz:
## Description

Adding services for credit notes export

### DIFF
--- a/spec/services/data_exports/csv/credit_notes_spec.rb
+++ b/spec/services/data_exports/csv/credit_notes_spec.rb
@@ -38,7 +38,7 @@ RSpec.describe DataExports::Csv::CreditNotes do
           credit_note.credit_amount_cents,
           credit_note.balance_amount_cents,
           credit_note.refund_amount_cents,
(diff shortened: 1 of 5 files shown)


spec/services/data_exports/csv/credit_notes_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    before { create(:credit_note) }
+
+    after do
+      file = result.csv_file
+      file.close
+      File.unlink(file.path)
+    end
+
+    it "adds serialized credit notes to csv", :aggregate_failures do
```
nudded, line 54: :aggregate_failures is enabled by default I think

nudded (approved):
nice!

spec/services/data_exports/csv/credit_notes_spec.rb
floganz: I'll merge this one as is but in following PR will double check if it's needed.
Thanks for brining attention to this.

PR #2978, opened 2024-12-18
Misc(events): add timestamp validation for create event

annvelents:
## Description

A check for event timestamp as it should be sent in seconds


vincent-pochet (approved):
Nice one :+1:

PR #2982, opened 2024-12-19
feat(credit-note-export): Add new filters options to credit notes list endpoint

floganz:
## Description

Added more filter options to credit notes list endpoint

### DIFF
--- a/app/controllers/api/v1/credit_notes_controller.rb
+++ b/app/controllers/api/v1/credit_notes_controller.rb
@@ -103,9 +103,9 @@ module Api
           filters: {
             currency: params[:currency],
             customer_external_id: params[:external_customer_id],
-            reason: select_valid_reasons(params[:reason]),
-            credit_status: select_valid_credit_statuses(params[:credit_status]),
-            refund_status: select_valid_refund_statuses(params[:refund_status]),
+            reason: params[:reason],
+            credit_status: params[:credit_status],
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/credit_notes_controller.rb
```diff
@@ -99,8 +99,18 @@ def index
             page: params[:page],
             limit: params[:per_page] || PER_PAGE
           },
+          search_term: params[:search_term],
           filters: {
-            customer_external_id: params[:external_customer_id]
+            currency: params[:currency],
+            customer_external_id: params[:external_customer_id],
+            reason: select_valid_reasons(params[:reason]),
+            credit_status: select_valid_credit_statuses(params[:credit_status]),
```
vincent-pochet, line 107: More a question:
I'm wondering if we should validate/clean the arguments in the query itself rather than in the controller. It could allow us to return errors in some cases if we want to "validate" input arguments and it could be performed no matter where the query is called.
What do you think?

vincent-pochet (approved):
Looks, good. I only have one question regarding the cleanup of the filter arguments

app/controllers/api/v1/credit_notes_controller.rb
floganz: I also thought about it, since we're not returning any errors in controllers it might be beneficial to move that logic to query object itself. 👍

PR #2986, opened 2024-12-19
misc(PaymentRequest): Apply payment idempotency refactor

vincent-pochet:
## Context

This PR follow https://github.com/getlago/lago-api/pull/2962 in the refactor of the payment processing. It applied the logic that was applied to payment for invoices to payment requests

- Avoid double payment at all cost

## Description

This PR applies the following logic for payment processing of payment requests:
- Creates a payment (or retrieve the pending one in case of retry) 
- Delegate the payment processing to the dedicated `PaymentProviders::*::Payments::CreateService` responsible only for the payment creation and the error handling (including idempotency). The payment provider payment creation logic will be shared between types of payment and will rely only on the Lago Payment record


app/jobs/invoices/payments/adyen_create_job.rb
```diff
@@ -10,7 +10,7 @@ class AdyenCreateJob < ApplicationJob
       retry_on Faraday::ConnectionFailed, wait: :polynomially_longer, attempts: 6
 
       def perform(invoice)
-        # NOTE: Legacy job, kept only to avoid existing jobs
+        # NOTE: Legacy job, kept only to avoid faileure with existing jobs
```
rsempe, line 13: `failure` 😅

rsempe (approved):

brunomiguelpinto (approved):

PR #2990, opened 2024-12-20
feat: centralize determine payment status

brunomiguelpinto:
## Context

This Pull Request is to backfill the payable_payment_status field for the Payment model and to centralize the status per provider 

## Description

This Pull Request introduces a reusable method determine_payment_status to the BaseProvider class, centralizing the logic for payment status determination. It also adds provider-specific status constants (PENDING_STATUSES, SUCCESS_STATUSES, FAILED_STATUSES)

and adds a migration to do the backfilling and the payable_payment_status

### DIFF
--- a/app/models/payment_providers/base_provider.rb
+++ b/app/models/payment_providers/base_provider.rb
@@ -32,7 +32,7 @@ module PaymentProviders
       return :succeeded if self.class::SUCCESS_STATUSES.include?(payment_status)
       return :failed if self.class::FAILED_STATUSES.include?(payment_status)
 
-      nil
+      payment_status
     end
   end
 end
--- a/app/models/payment_providers/base_provider.rb
+++ b/app/models/payment_providers/base_provider.rb
@@ -26,6 +26,14 @@ module PaymentProviders
     validates :name, presence: true
 
     settings_accessors :webhook_secret, :success_redirect_url
+
+    def determine_payment_status(payment_status)
+      return :pending if self.class::PENDING_STATUSES.include?(payment_status)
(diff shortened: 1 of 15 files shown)


app/models/payment_providers/base_provider.rb
```diff
@@ -26,6 +26,14 @@ class BaseProvider < ApplicationRecord
     validates :name, presence: true
 
     settings_accessors :webhook_secret, :success_redirect_url
+
+    def determine_payment_status(payment_status)
+      return :pending if self.class::PENDING_STATUSES.include?(payment_status)
+      return :succeeded if self.class::SUCCESS_STATUSES.include?(payment_status)
+      return :failed if self.class::FAILED_STATUSES.include?(payment_status)
+
+      nil
```
brunomiguelpinto, line 35: nil or payment_status if not nil this will break when creating the payment but it should break if we don't know the state ?

app/models/payment_providers/base_provider.rb
```diff
@@ -26,6 +26,14 @@ class BaseProvider < ApplicationRecord
     validates :name, presence: true
 
     settings_accessors :webhook_secret, :success_redirect_url
+
+    def determine_payment_status(payment_status)
+      return :pending if self.class::PENDING_STATUSES.include?(payment_status)
+      return :succeeded if self.class::SUCCESS_STATUSES.include?(payment_status)
+      return :failed if self.class::FAILED_STATUSES.include?(payment_status)
+
+      payment_status
```
brunomiguelpinto, line 35: nil or payment_status. If not nil, this will break when creating the payment, but it should break if we don't know the status?

vincent-pochet (approved):

PR #2999, opened 2024-12-23
fix(wallet): Validate currency

vincent-pochet:
## Context

This PR is related to 
```
ActiveRecord::NotNullViolation

PG::NotNullViolation: ERROR:  null value in column "balance_currency" of relation "wallets" violates not-null constraint (ActiveRecord::NotNullViolation)
```
The error is raised when creating a wallet with a null currency if the related customer does not have a currency.

## Description

The PR
- Add a validation on `wallet#currency` making sure it is included in the accepted currency list.
- Avoid updating the customer currency with a nil value

### DIFF
--- a/app/models/wallet.rb
+++ b/app/models/wallet.rb
@@ -2,6 +2,7 @@
 
 class Wallet < ApplicationRecord
   include PaperTrailTraceable
+  include Currencies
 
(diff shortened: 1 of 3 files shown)


rsempe (approved):
nice!

brunomiguelpinto (approved):

PR #3002, opened 2024-12-24
feat(anrok): make api calls async in dedicated service

lovrocolic:
## Context

Currently Anrok calls are sync and performed during invoice generation

## Description

The goal of this improvement is to make calls to Anrok in dedicated job so that we can implement throttling for rate limit issues.

This PR changes all places where Anrok in used in sync way. This is the last PR that is extension of:
- https://github.com/getlago/lago-api/pull/2984
- https://github.com/getlago/lago-api/pull/2968
- https://github.com/getlago/lago-api/pull/2964

### DIFF
--- a/app/jobs/fees/create_pay_in_advance_job.rb
+++ b/app/jobs/fees/create_pay_in_advance_job.rb
@@ -2,7 +2,11 @@
 
 module Fees
   class CreatePayInAdvanceJob < ApplicationJob
+    include ConcurrencyThrottlable
+
     queue_as :default
+
+    retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
 
     unique :until_executed, on_conflict: :log
 
--- a/app/jobs/fees/create_pay_in_advance_job.rb
+++ b/app/jobs/fees/create_pay_in_advance_job.rb
@@ -2,8 +2,6 @@
 
 module Fees
   class CreatePayInAdvanceJob < ApplicationJob
-    include ConcurrencyThrottlable
-
     queue_as :default
 
     retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
--- a/app/jobs/invoices/create_pay_in_advance_charge_job.rb
+++ b/app/jobs/invoices/create_pay_in_advance_charge_job.rb
@@ -2,8 +2,6 @@
 
 module Invoices
   class CreatePayInAdvanceChargeJob < ApplicationJob
-    include ConcurrencyThrottlable
-
     queue_as do
       if ActiveModel::Type::Boolean.new.cast(ENV['SIDEKIQ_BILLING'])
         :billing
--- a/app/jobs/invoices/create_pay_in_advance_charge_job.rb
+++ b/app/jobs/invoices/create_pay_in_advance_charge_job.rb
@@ -2,6 +2,8 @@
 
 module Invoices
   class CreatePayInAdvanceChargeJob < ApplicationJob
+    include ConcurrencyThrottlable
+
     queue_as do
       if ActiveModel::Type::Boolean.new.cast(ENV['SIDEKIQ_BILLING'])
         :billing
@@ -11,6 +13,7 @@ module Invoices
     end
 
     retry_on Sequenced::SequenceError
+    retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
 
     unique :until_executed, on_conflict: :log
 
--- a/app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
+++ b/app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
@@ -3,7 +3,11 @@
 module Invoices
   module ProviderTaxes
     class PullTaxesAndApplyJob < ApplicationJob
+      include ConcurrencyThrottlable
+
       queue_as 'integrations'
+
+      retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
 
       def perform(invoice:)
         Invoices::ProviderTaxes::PullTaxesAndApplyService.call(invoice:)
--- a/app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
+++ b/app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
@@ -3,8 +3,6 @@
 module Invoices
   module ProviderTaxes
     class PullTaxesAndApplyJob < ApplicationJob
-      include ConcurrencyThrottlable
-
       queue_as 'integrations'
 
       retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
--- a/app/services/invoices/compute_taxes_and_totals_service.rb
+++ b/app/services/invoices/compute_taxes_and_totals_service.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Invoices
+  class ComputeTaxesAndTotalsService < BaseService
+    def initialize(invoice:, finalizing: true)
+      @invoice = invoice
+      @finalizing = finalizing
+
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: 'invoice') unless invoice
+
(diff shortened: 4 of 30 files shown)


spec/graphql/mutations/invoices/finalize_spec.rb
```diff
@@ -44,4 +45,35 @@
...
+  context 'with tax provider' do
+    let(:integration) { create(:anrok_integration, organization:) }
+    let(:integration_customer) { create(:anrok_customer, integration:, customer:) }
+
+    before do
+      integration_customer
+    end
+
+    it 'returns pending invoice' do
+      freeze_time do
```
ancorcruz, line 58: do you need to `freeze_time`? there are no time checks

spec/graphql/mutations/invoices/finalize_spec.rb
```diff
@@ -44,4 +45,35 @@
...
+          permissions: required_permission,
+          query: mutation,
+          variables: {
+            input: {id: invoice.id}
+          }
+        )
+
+        result_data = result['data']['finalizeInvoice']
+
+        aggregate_failures do
```
ancorcruz, line 71: we aggregate_failures by default, hence, this is not needed anymore.

ancorcruz (approved):
I have a few questions... approved anyway.

app/services/invoices/progressive_billing_service.rb
```diff
@@ -22,19 +22,10 @@ def call
         Credits::ProgressiveBillingService.call(invoice:)
         Credits::AppliedCouponsService.call(invoice:)
 
-        if customer_provider_taxation?
-          taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateService.call(invoice:, fees: invoice.fees)
+        totals_result = Invoices::ComputeTaxesAndTotalsService.call(invoice:)
+        return totals_result if !totals_result.success? && totals_result.error.is_a?(BaseService::UnknownTaxFailure)
```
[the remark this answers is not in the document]
lovrocolic: Maybe naming is confusing, but I couldn't find better one:) The idea is:
- if unknown error happens, we don't want to raise error since it would rollback all DB changes
- instead we just want to catch this state and leave invoice in `pending` status
- Async job for fetching taxes will process `pending` invoice and put it either in `finalized` or `failed` (if there are tax errors like invalid product mapping)

app/services/invoices/subscription_service.rb
```diff
@@ -34,7 +34,7 @@ def call
           context:
         )
 
-        set_invoice_generated_status unless invoice.failed?
+        set_invoice_generated_status unless invoice.pending?
```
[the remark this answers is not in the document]
lovrocolic: Before we were fetching taxes is sync mode (in the same step when we generated invoice) so invoice could be `failed` (failed status is only related to tax errors).
With new changes (converting fetching taxes to be async) it cannot be `failed` here, only `pending`. However background job that fetch taxes will put invoice either in `finalized` (`draft`) or `failed`

app/services/invoices/provider_taxes/pull_taxes_and_apply_service.rb
```diff
@@ -34,18 +34,19 @@ def call
         provider_taxes = taxes_result.fees
 
         ActiveRecord::Base.transaction do
+          unless invoice.draft?
+            invoice.issuing_date = issuing_date
+            invoice.payment_due_date = payment_due_date
```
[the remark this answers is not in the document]
lovrocolic: you are right, draft invoice can already have this informations. However, when we are in the finalize mode, we need to overwrite those informations since finalization can happen also before original issuing date (if it is triggered manually)

app/services/invoices/refresh_draft_and_finalize_service.rb
```diff
@@ -15,7 +15,7 @@ def call
       ActiveRecord::Base.transaction do
         invoice.issuing_date = issuing_date
         refresh_result = Invoices::RefreshDraftService.call(invoice:, context: :finalize)
-        if tax_error?(refresh_result.error)
```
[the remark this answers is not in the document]
lovrocolic: Nice catch, will remove it

spec/graphql/mutations/invoices/finalize_spec.rb
lovrocolic: updated

spec/graphql/mutations/invoices/finalize_spec.rb
lovrocolic: updated

app/services/invoices/compute_taxes_and_totals_service.rb
```diff
@@ -0,0 +1,38 @@
...
+    def call
+      return result.not_found_failure!(resource: 'invoice') unless invoice
+
+      if customer_provider_taxation?
+        invoice.status = 'pending' if finalizing
+        invoice.tax_status = 'pending'
+        invoice.save!
+        after_commit { Invoices::ProviderTaxes::PullTaxesAndApplyJob.perform_later(invoice:) }
+
+        return result.unknown_tax_failure!(code: 'tax_error', message: 'unknown taxes')
```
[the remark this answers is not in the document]
ancorcruz: but then this is the expected behavior not a failure, isn't it?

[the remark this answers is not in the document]
lovrocolic: Invoice is not valid without taxes so that's why I wanted to mark this case with this failure. Probably it is just style preference. Other option could be that I check wherever I need if `invoice.tax_status = 'pending'` or something similar:)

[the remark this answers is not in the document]
ancorcruz: I don't want to block the change as you mentioned this is urgent and should be released before the next billing run... you can ignore my comments and go ahead.

However, I would avoid calling a failure some that is the normal behaviour of the service... this is for me a success with pending tax status as you mentioned. Failure should be used to indicate real failures like validation failures, etc.

app/jobs/fees/create_pay_in_advance_job.rb
```diff
@@ -2,8 +2,12 @@
 
 module Fees
   class CreatePayInAdvanceJob < ApplicationJob
+    include ConcurrencyThrottlable
```
ivannovosad, line 5: Do we need to restrict it to only 5 concurrent calls? If not we can remove this line.
Anrok doesn't have concurrency restrictions, only 10 requests per second per API key.

app/jobs/invoices/create_pay_in_advance_charge_job.rb
```diff
@@ -2,6 +2,8 @@
 
 module Invoices
   class CreatePayInAdvanceChargeJob < ApplicationJob
+    include ConcurrencyThrottlable
```
ivannovosad, line 5: Same as above:

Do we need to restrict it to only 5 concurrent calls? If not we can remove this line.
Anrok doesn't have concurrency restrictions, only 10 requests per second per API key.

app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
```diff
@@ -3,8 +3,12 @@
 module Invoices
   module ProviderTaxes
     class PullTaxesAndApplyJob < ApplicationJob
+      include ConcurrencyThrottlable
```
ivannovosad, line 6: Here as well:

Do we need to restrict it to only 5 concurrent calls? If not we can remove this line.
Anrok doesn't have concurrency restrictions, only 10 requests per second per API key.

lovrocolic: removed

app/jobs/invoices/create_pay_in_advance_charge_job.rb
lovrocolic: removed

app/jobs/fees/create_pay_in_advance_job.rb
lovrocolic: removed

PR #3003, opened 2024-12-24
feat: Add payment type and reference

brunomiguelpinto:
## Context

This PR lays the groundwork for implementing manual/partial payment functionality in our system. It introduces necessary changes to the payments and invoices tables to enable the tracking and management of these new payment types and statuses.

## Description

Migration: Add `payment_type` and reference columns to payments

- Added a new `payment_type` column as an ENUM with values provider and manual.
- Added a `reference` column to store additional metadata for payments.
- Backfilled existing records, defaulting payment_type to provider.
- Set payment_type to NOT NULL with a default value of provider.

Migration: Add `total_paid_amount_cents` column to invoices

- Added a `total_paid_amount_cents` column to track the cumulative amount paid toward an invoice.
- Backfilled existing records with a default value of 0.
- Set total_paid_amount_cents to NOT NULL with a default value of 0.


db/migrate/20241224141116_add_payment_type_and_reference_to_payments.rb
```diff
@@ -0,0 +1,37 @@
...
+
+class AddPaymentTypeAndReferenceToPayments < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def up
+    create_enum :payment_type, %w[provider manual]
+
+    safety_assured do
+      change_table :payments, bulk: true do |t|
+        t.column :payment_type, :enum, enum_type: 'payment_type', null: true
```
annvelents, line 11: will we need index on this column?

brunomiguelpinto: Yes good catch I will add it

db/migrate/20241224142141_add_total_paid_amount_cents_to_invoices.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class AddTotalPaidAmountCentsToInvoices < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def up
+    add_column :invoices, :total_paid_amount_cents, :bigint, null: true, default: 0
```
annvelents, line 7: will in the future we update the existing paid invoices with real values? 🤔

brunomiguelpinto: Yes I will do a migration to fill the values i was thinking on another pr but can be on this one

brunomiguelpinto: I need to do the migration later, not on this PR, because I will sync some invoices. However, new ones will have a value of 0 and will only be properly updated when we change the code later. So, when we change the code, we should perform the migration at the same time. For now, just adding the field is fine; we won’t use it yet.

this way I don't have to migrate the data 2 times

ivannovosad (approved):

db/migrate/20241224142141_add_total_paid_amount_cents_to_invoices.rb
```diff
@@ -0,0 +1,27 @@
...
+    # Backfill
+    Invoice.in_batches(of: 10_000).each do |batch|
+      batch.update_all(total_paid_amount_cents: 0) # rubocop:disable Rails/SkipsModelValidations
```
lovrocolic, line 10: Do we need this if default is zero?

brunomiguelpinto: i should be without the default to not lock the db I will remove it

spec/models/payment_spec.rb
```diff
@@ -3,13 +3,72 @@
...
+          it 'does not add an error' do
+            expect(errors.where(:reference, :present)).not_to be_present
+          end
+        end
+      end
+
+      context 'when payment type is manual' do
+        let(:payment_type) { 'manual' }
+
+        context 'when reference is present' do
```
ivannovosad, line 52: Lets add one more context here:

```
        context 'when reference is not present' do
          it 'adds an error' do
            expect(errors.where(:reference, :present)).to be_present
          end
        end
```

PR #3004, opened 2024-12-24
misc: Add more tests

floganz:
## Description

Add Fuubar gem
Change `FactoryBot.create` to `FactoryBot.build` where possible
Add missing tests for PaperTrail
Add tests to some associations and validations


annvelents (approved):
👏 Also nice work getting rid of creates!

PR #3007, opened 2024-12-26
Feat(invoice_custom_sections):  api controllers

annvelents:
## Context

API controllers to support invoice_custom_sections functionality

## Description

Added InvoiceCustomSectionsController and updated customers controller with two new actions


app/serializers/v1/invoice_custom_section_serializer.rb
```diff
@@ -0,0 +1,20 @@
...
+        lago_id: model.id,
+        code: model.code,
+        name: model.name,
+        description: model.description,
+        details: model.details,
+        display_name: model.display_name,
+        selected_for_organization: model.selected_for_organization?,
+        organization: {
+          lago_id: model.organization_id
+        }
```
vincent-pochet, line 16: Do we plan to add more organization details in this serializer? If no I would suggest to flatten it (not a blocker ;) )

app/services/customers/manage_invoice_custom_sections_service.rb
```diff
@@ -35,14 +38,22 @@ def call
 
     private
 
-    attr_reader :customer, :section_ids, :skip_invoice_custom_sections
+    attr_reader :customer, :section_ids, :skip_invoice_custom_sections, :section_codes
+
+    def raise_double_selection
```
vincent-pochet, line 43: Not sure of the naming here as `result.validation_failure!` is not raising an error but only marking the result object as failed and adding the error detail to it.

app/services/customers/manage_invoice_custom_sections_service.rb
```diff
@@ -2,27 +2,30 @@
...
+      @section_codes = section_codes
       @skip_invoice_custom_sections = skip_invoice_custom_sections
 
       super
     end
 
     def call
       return result.not_found_failure!(resource: "customer") unless customer
-      raise_invalid_params if skip_invoice_custom_sections && section_ids.present?
+      raise_double_selection if !section_ids.nil? && !section_codes.nil?
```
vincent-pochet, line 16: You might need to return the result here (or am I missing something?)

app/serializers/v1/invoice_custom_section_serializer.rb
annvelents: yep, makes perfect sense, updating, thank you!

app/services/customers/manage_invoice_custom_sections_service.rb
annvelents: oops, yes, thanks

app/services/customers/manage_invoice_custom_sections_service.rb
annvelents: 🙌

app/serializers/v1/invoice_custom_section_serializer.rb
annvelents: (updated it in the last PR of the chain: https://github.com/getlago/lago-api/pull/3013/files)

vincent-pochet (approved):

PR #3012, opened 2024-12-30
fix: Recurring fee not displayed on invoices after upgrade

ancorcruz:
## Context

Steps to reproduce the issue:

1. Create a recurring metric;

2. Create a plan with a charge linked to the recurring metric - the charge should be:
* prorated = true;
* pay_in_advance = true;
* invoiceable = false;
* regroup_paid_fees = invoice;

3. Create a subscription with a start date in the past and send an event in the past (e.g. 1 December) -> you receive a fee.created webhook;

4. Duplicate the original plan and only modify the price of the charge (higher price);

5. Trigger an upgrade of the existing subscription by assigning the new plan;

6. Mark the fee in step 3 as succeeded via POST api/v1/fees/:lago_id

7. Terminate the subscription -> the fee is not included in the end-of-period invoice.

Expected behaviour: the fee should be included in the end-of-period invoice, even if the original Lago subscription to which it belongs is no longer active.

## Description

This change fetches all subscriptions with status active or terminated of a given organization with a matching external id to invoice their paid in advance fees.

### DIFF
--- a/app/services/subscriptions/terminate_service.rb
+++ b/app/services/subscriptions/terminate_service.rb
@@ -111,15 +111,29 @@ module Subscriptions
           subscription.terminated_at,
           invoicing_reason: :subscription_terminating
         )
-        BillNonInvoiceableFeesJob.set(wait: 2.seconds).perform_later([subscription], subscription.terminated_at)
+        BillNonInvoiceableFeesJob.set(wait: 2.seconds).perform_later(
+          invoiceable_subscriptions.to_a,
+          subscription.terminated_at
+        )
       else
         BillSubscriptionJob.perform_now(
           [subscription],
           subscription.terminated_at,
           invoicing_reason: :subscription_terminating
         )
-        BillNonInvoiceableFeesJob.perform_now([subscription], subscription.terminated_at)
+        BillNonInvoiceableFeesJob.perform_now(
+          invoiceable_subscriptions.to_a,
+          subscription.terminated_at
+        )
       end
+    end
+
+    def invoiceable_subscriptions
+      # Expect all subscriptions belogns to the same organization
+      @invoiceable_subscriptions ||= subscription.organization.subscriptions.where(
+        external_id: subscription.external_id,
+        status: [:active, :terminated]
+      )
     end
 
     # NOTE: If subscription is terminated automatically by setting ending_at, there is a chance that this service will
--- a/app/services/subscriptions/terminate_service.rb
+++ b/app/services/subscriptions/terminate_service.rb
@@ -111,10 +111,7 @@ module Subscriptions
           subscription.terminated_at,
           invoicing_reason: :subscription_terminating
         )
-        BillNonInvoiceableFeesJob.set(wait: 2.seconds).perform_later(
-          invoiceable_subscriptions.to_a,
-          subscription.terminated_at
-        )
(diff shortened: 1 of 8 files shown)


app/services/subscriptions/terminate_service.rb
```diff
@@ -111,17 +111,31 @@ def bill_subscription
           subscription.terminated_at,
           invoicing_reason: :subscription_terminating
         )
-        BillNonInvoiceableFeesJob.set(wait: 2.seconds).perform_later([subscription], subscription.terminated_at)
+        BillNonInvoiceableFeesJob.set(wait: 2.seconds).perform_later(
```
nudded, line 114: should this wait not just be a `after_commit`? 🤔

app/services/subscriptions/terminate_service.rb
```diff
@@ -111,17 +111,31 @@ def bill_subscription
...
+        BillNonInvoiceableFeesJob.perform_now(
+          invoiceable_subscriptions.to_a,
+          subscription.terminated_at
+        )
       end
     end
 
+    def invoiceable_subscriptions
+      # Expect all subscriptions belogns to the same organization
+      @invoiceable_subscriptions ||= subscription.organization.subscriptions.where(
```
nudded, line 133: This might return multiple terminated subscriptions for this external_id, is that expected?

ancorcruz: yes, this is the fix...

for non invoiceable fees we need to wait until the is payment succeeded to generate the invoice. This could happen at whatever time after the fee was created (for example after a few months or even years). The invoice will be generated once the fee is payment succeeded and the subscription is billed because of termination, deletion or end of billing period.

The scenario we are fixing here is....

1. Subscription A is created.
2. A fee is generated.
3. Subscription B upgrades subscription A
4. Fee is marked as payment succeeded
5. Subscription B is terminated or end of billing period is reached

At this point the fee should be invoiced however the fee belongs to the subscription A, not the subscription being terminated at this time that is subscription  B.

This finds all related subscriptions to the subscription being terminated and generates invoices for all non invoiceable fees....

Also, writing this comment I realised the scenario of the end of billing period should be checked  as this is only about termination.

nudded: yes, but you can also have Subscription C, D and E that were already terminated. this query returns all of those, no?

app/services/subscriptions/terminate_service.rb
ancorcruz: That is a good question... queuing these jobs is not wrapped within a DB transaction, the same for another job within this class... maybe a left over from a older implementation or just a fix for a race condition sidekiq vs Postgres.

nudded: okay, but either it's committed and we don't need the wait or it's not and we just need to use `after_commit` no?

ancorcruz: I have rolled back the changes in this file, adding this to the backlog of pending improvements as this is a bugfix and we need to release it ASAP.

app/services/subscriptions/terminate_service.rb
ancorcruz: Yes, I have move this to the advance charges service.

lovrocolic (approved):

PR #3013, opened 2024-12-30
Feat(invoice_custom_sections): use custom sections in pdf

annvelents:
Use applied invoice custom sections in PDFs and include them in invoice payload
Also includes fixes found after QA


app/graphql/resolvers/invoice_custom_sections_resolver.rb
```diff
@@ -15,10 +15,13 @@ class InvoiceCustomSectionsResolver < Resolvers::BaseResolver
...
-      current_organization.invoice_custom_sections.left_outer_joins(:invoice_custom_section_selections).order(
-        Arel.sql('CASE WHEN invoice_custom_section_selections.id IS NOT NULL THEN 0 ELSE 1 END'),
-        :name
-      ).page(page).per(limit)
+      current_organization.invoice_custom_sections
+        .joins('LEFT JOIN invoice_custom_section_selections ON invoice_custom_sections.id = invoice_custom_section_selections.invoice_custom_section_id
+                AND invoice_custom_section_selections.customer_id is NULL')
+        .order(
+          Arel.sql('CASE WHEN invoice_custom_section_selections.id IS NOT NULL THEN 0 ELSE 1 END'),
+          :name
```
vincent-pochet, line 23: No blocker here but maybe this should be turned into a scope. What do you think?

vincent-pochet (approved):

app/graphql/resolvers/invoice_custom_sections_resolver.rb
annvelents: I think it's not needed, because this sorting - by selected for organization, and then alphabetically - only needed in the graphql response... I can't see where else it will be useful... or do you mean it will be better for readability?
(in all other places we're usually using the selected sections and only sort them alphabetically)

PR #3017, opened 2025-01-02
fix(locks): Stop enqueuing duplicates job because of default lock TTL

jdenquin:


nudded (approved):

rsempe (approved):

app/jobs/wallets/refresh_ongoing_balance_job.rb
```diff
@@ -4,7 +4,7 @@ module Wallets
   class RefreshOngoingBalanceJob < ApplicationJob
     queue_as 'wallets'
 
-    unique :until_executed, on_conflict: :log
+    unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
```
rsempe, line 7: not sure about 12 hours here.. maybe too long, no?

jdenquin: whats the potential issue?

nudded: one potential issue is with workers dying. that will leave a wallet unable to be refreshed for 12 hours.

nudded: activejob-uniqueness uses Redis to store the lock, if the worker dies we cannot clean up that state.

jdenquin: okay lets discuss it later on how to manage those cases!

PR #3018, opened 2025-01-02
feat: Add Conditional Integration Customers to use on resync webhook

brunomiguelpinto:
## Context

The ResyncedService requires the inclusion of integration_customers in the serialized Customer object as part of the webhook payload. However, this inclusion should only happen in specific cases, like in the ResyncedService, and not affect other serializers or services. 

we use this webhook to do the salesforce integration that why we need the customer integration

### DIFF
--- a/app/services/webhooks/invoices/resynced_service.rb
+++ b/app/services/webhooks/invoices/resynced_service.rb
@@ -13,7 +13,7 @@ module Webhooks
         ::V1::InvoiceSerializer.new(
           object,
           root_name: 'invoice',
-          includes: %i[customer subscriptions fees credits applied_taxes]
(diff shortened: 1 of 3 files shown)


app/services/webhooks/invoices/resynced_service.rb
```diff
@@ -13,7 +13,7 @@ def object_serializer
         ::V1::InvoiceSerializer.new(
           object,
           root_name: 'invoice',
-          includes: %i[customer subscriptions fees credits applied_taxes]
```
nudded, line 16: maybe a small spec to make sure this field is part of the resynced web hook?

nudded (approved):

PR #3020, opened 2025-01-03
misc(fee): Delete zero amount finalized fees

vincent-pochet:
## Description

This PR removes the charge fees attached to finalized or voided invoices and having `amount = 0`.

In a close future, these fees will not be generated anymore, and the logic built around it will be refactored (adjusted fees, daily_usage computation...)

### DIFF
--- a/db/migrate/20250103124802_drop_zero_amount_fees.rb
+++ b/db/migrate/20250103124802_drop_zero_amount_fees.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class DropZeroAmountFees < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  class FeesTax < ApplicationRecord; end
+
+  class Fee < ApplicationRecord; end
+
+  def change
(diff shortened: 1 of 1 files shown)


db/migrate/20250103124802_drop_zero_amount_fees.rb
```diff
@@ -0,0 +1,29 @@
...
+class DropZeroAmountFees < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    sql = <<~SQL
+      DELETE FROM fees
+      USING invoices
+      WHERE
+        fees.invoice_id = invoices.id
+        AND invoices.status IN (1, 2) -- finalized or voided
```
annvelents, line 12: shall there also be `closed: 6`?

vincent-pochet: Yes, we can add it indeed. Good catch!

rsempe (approved):

nudded (approved):

PR #3028, opened 2025-01-07
fix: Update provider taxes rescue logic

lovrocolic:
For provider taxes, we do not want to create dead job for all errors related to Anrok configuration. For everything else it is fine to have dead job

### DIFF
--- a/app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
+++ b/app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
@@ -8,7 +8,7 @@ module Invoices
       retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
 
       def perform(invoice:)
-        Invoices::ProviderTaxes::PullTaxesAndApplyService.call!(invoice:)
+        Invoices::ProviderTaxes::PullTaxesAndApplyService.call(invoice:)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):

app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
```diff
@@ -8,7 +8,7 @@ class PullTaxesAndApplyJob < ApplicationJob
       retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
 
       def perform(invoice:)
-        Invoices::ProviderTaxes::PullTaxesAndApplyService.call!(invoice:)
+        Invoices::ProviderTaxes::PullTaxesAndApplyService.call(invoice:)
```
vincent-pochet, line 11: This is actually swallowing the error. It feels a bit risky.

Shouldn't we handle the error cases in the service to avoid loosing important errors?

lovrocolic: Maybe I am missing something, but with new logic, all errors related to taxes will be stored in the DB and user can handle them. Everything else will be raised since we don't rescue anymore 🤔

lovrocolic: One more update:)

PR #3032, opened 2025-01-07
feat(invoice): Avoid generating 0 amount charge fees

vincent-pochet:
## Description

This PR follow https://github.com/getlago/lago-api/pull/3020 the goal is to stop generating empty fees (0 units, amount and events without adjustments or true up fees).

The reason behind this change is scalability as for now, depending on the plan setups, it appears that 98% of generated fees does not have an amount...

**NOTE:** a new feature will improve the edit fees feature since this PR will prevent creation of adjustment fees from the UI if no event have been received for a specific billable metric on a period.

### DIFF
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -2,14 +2,15 @@
 
 module Fees
   class ChargeService < BaseService
-    def initialize(invoice:, charge:, subscription:, boundaries:, current_usage: false, cache_middleware: nil, bypass_aggregation: false)
+    def initialize(invoice:, charge:, subscription:, boundaries:, context: nil, cache_middleware: nil, bypass_aggregation: false)
       @invoice = invoice
       @charge = charge
       @subscription = subscription
       @boundaries = OpenStruct.new(boundaries)
       @currency = subscription.plan.amount.currency
 
-      @current_usage = current_usage
+      @context = context
+      @current_usage = context == :current_usage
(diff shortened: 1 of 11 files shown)


app/services/fees/charge_service.rb
```diff
@@ -167,6 +173,15 @@ def init_fee(amount_result, properties:, charge_filter:)
       new_fee
     end
 
+    def should_persit_fee?(fee)
+      return true if context == :recurring
+      return true if fee.units != 0 || fee.amount_cents != 0 || fee.events_count != 0
+      return true if adjusted_fee(charge_filter: fee.charge_filter, grouped_by: fee.grouped_by).present?
+      return true if fee.true_up_parent_fee.present?
+
+      result.fees.any? { |f| f.true_up_parent_fee == fee }
```
nudded, line 182: This specific line feels weird 🤔 can we not rely on `result.fees`?

app/services/fees/charge_service.rb
```diff
@@ -37,7 +38,12 @@ def call
 
       ActiveRecord::Base.transaction do
         result.fees.each do |fee|
```
nudded, line 40: Perhaps it's more expressive to do `result.fees.filter! {|f| should_persist_fee(f)}` before this loop? 

(specifically to avoid confusion around `next result.fees.delete(fee)`)

nudded (approved):

PR #3040, opened 2025-01-09
misc(wallet): Refactor wallet transaction related jobs queuing

vincent-pochet:
## Context
Queuing of wallet transaction related jobs is ideal as jobs as a low priority, meaning that under heavy load, they are a huge delay might happen between enqueue and  processing. Since it is user facing, we need to improve it.

## Description

This PR:
- Adds a new `high_priority` queue that should be used for all jobs requiring an "immediate" processing
- Moves the `Invoices::PrepaidCreditJob` from the `wallet` to the new `high_priority` queue
- Moves the `WalletTransactions::CreateJob` from the `wallet` to the new `high_priority` queue
- Moves `Wallets::RefreshOngoingBalanceJob` into the low priority queue

Note no more jobs will relies on the `wallet` queue, but this queue is kept fro now in the config to avoid issue with existing enqueued jobs.

### DIFF
--- a/spec/services/invoices/advance_charges_service_spec.rb
+++ b/spec/services/invoices/advance_charges_service_spec.rb
@@ -160,6 +160,7 @@ RSpec.describe Invoices::AdvanceChargesService, type: :service do
         create(
           :fee,
           :succeeded,
+          succeeded_at: fee_boundaries[:charges_to_datetime] - 2.days,
           invoice_id: nil,
           subscription: subscription_2,
           amount_cents: 999,
--- a/app/jobs/invoices/prepaid_credit_job.rb
+++ b/app/jobs/invoices/prepaid_credit_job.rb
@@ -2,7 +2,7 @@
 
 module Invoices
   class PrepaidCreditJob < ApplicationJob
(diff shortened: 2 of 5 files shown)


brunomiguelpinto (approved):

nudded (approved):
🚀

spec/services/invoices/advance_charges_service_spec.rb
```diff
@@ -160,6 +160,7 @@ def fee_boundaries
         create(
           :fee,
           :succeeded,
+          succeeded_at: fee_boundaries[:charges_to_datetime] - 2.days,
```
annvelents, line 163: Nice! 👏

annvelents (approved):

spec/services/invoices/advance_charges_service_spec.rb
ancorcruz: @vincent-pochet why `-2.days`? Is it arbitrary or does it have a logic behind it?

ancorcruz (approved):
LGTM

PR #3055, opened 2025-01-15
fix(invoice): Fix updating invoice payment status from succeeded

ivannovosad:
## Context

Currently there is a possibility to update invoice payment status from succeeded to other status.

## Description

This PR fixes `Invoices::UpdateService` that raises an error when attempting to change invoice's `succeeded` payment status.

### DIFF
--- a/app/services/invoices/update_service.rb
+++ b/app/services/invoices/update_service.rb
@@ -35,6 +35,10 @@ module Invoices
         return result.not_allowed_failure!(code: 'payment_status_update_on_draft_invoice')
       end
 
+      if old_payment_status.to_s == 'succeeded' && params.key?(:payment_status) && params[:payment_status].to_s != 'succeeded'
+        return result.not_allowed_failure!(code: 'payment_status_update_on_succeeded_invoice')
+      end
+
       if params.key?(:ready_for_payment_processing) && !invoice.voided?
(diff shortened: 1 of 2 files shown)


nudded (approved):

ivannovosad:
This should be merged only when we forbid this transition (`succeeded` -> `...` ) in the UI as well.

app/services/invoices/update_service.rb
```diff
@@ -35,6 +35,10 @@ def call
         return result.not_allowed_failure!(code: 'payment_status_update_on_draft_invoice')
       end
 
+      if old_payment_status.to_s == 'succeeded' && params.key?(:payment_status) && params[:payment_status].to_s != 'succeeded'
```
vincent-pochet, line 38: I checked the file and I think we could improve a bit the logic of payment status update:

```ruby
old_payment_status = invoice.payment_status
if params.key?(:payment_status) && old_payment_status.to_s != params.key?(:payment_status).to_s
  return result.not_allowed_failure!(code: 'payment_status_update_on_draft_invoice') if invoice.draft?
  return result.not_allowed_failure!(code: 'payment_status_update_on_succeeded_invoice') if invoice.payment_succeeded?

  invoice.payment_status = params[:payment_status]
end
```

PR #3064, opened 2025-01-16
feat: Expose billing periods in invoice serializer

vincent-pochet:
## Context

This PR is related to the rollback of the  zero amount removal (See https://github.com/getlago/lago-api/pull/3059)

It appears that some customers are relying on the fees on some invoice webhooks to retrieve the billing period info. 

## Description

This PR makes sure that the billing period info are always sent in webhooks / exposed in the API when, to ensure that users can rely on it even if no fees have been created


nudded (approved):
🚀

PR #3069, opened 2025-01-17
Add taxes to recreated recurring pay in advance fees

floganz:
## Context

Currently we had an issue that pay in advance, non-invoiceable and recurring fees are missing taxes when recreated at next billing period for subscription. This PR partially fixes it by applying defined in Lago taxes.
There will be another part addressing taxes via Anrok provider.

## Description

Optionally add Lago-defined taxes for charge fees.

### DIFF
--- a/app/services/invoices/calculate_fees_service.rb
+++ b/app/services/invoices/calculate_fees_service.rb
@@ -189,7 +189,7 @@ module Invoices
           charge:,
           subscription:,
           boundaries:,
-          apply_taxes: !invoice.customer.anrok_customer.present?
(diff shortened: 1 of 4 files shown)


app/services/invoices/calculate_fees_service.rb
```diff
@@ -165,6 +165,8 @@ def should_create_recurring_non_invoiceable_fees?(subscription)
       true
     end
 
+    require "pry"
```
lovrocolic, line 168: I guess we don't need it

lovrocolic (approved):

PR #3073, opened 2025-01-17
feat(invoice-preview): add logic for applying coupons on preview invoice

lovrocolic:
## Context

Preview feature enables fetching first Lago invoice. This invoice is not persisted and is calculated on the fly.

## Description

This PR adds service for applying coupons on preview invoice. The goal was not to affect heavily main service for applying coupons on invoices.

In order not to repeat the code, some common logic is extracted  s that it can be used on both places

### DIFF
--- a/app/models/applied_coupon.rb
+++ b/app/models/applied_coupon.rb
@@ -32,6 +32,13 @@ class AppliedCoupon < ApplicationRecord
     self.terminated_at ||= timestamp
     terminated!
   end
+
+  def remaining_amount
+    return @remaining_amount if defined?(@remaining_amount)
+
+    already_applied_amount = credits.sum(&:amount_cents)
+    @remaining_amount = amount_cents - already_applied_amount
+  end
 end
 
 # == Schema Information
--- a/app/services/coupons/preview_service.rb
+++ b/app/services/coupons/preview_service.rb
@@ -26,7 +26,6 @@ module Coupons
         base_amount_cents = base_amount_cents(applied_coupon, fees)
         credit = add_credit(applied_coupon, fees, base_amount_cents)
 
-        applied_coupon.credits << credit
         result.credits << credit
         invoice.credits << credit
       end
--- a/app/services/coupons/preview_service.rb
+++ b/app/services/coupons/preview_service.rb
@@ -1,0 +1,93 @@
+# frozen_string_literal: true
+
+module Coupons
+  class PreviewService < BaseService
+    def initialize(invoice:, applied_coupons:)
+      @invoice = invoice
+      @applied_coupons = applied_coupons
+
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: 'invoice') unless invoice
+      return result.not_found_failure!(resource: 'applied_coupons') unless applied_coupons
+
+      result.credits = []
+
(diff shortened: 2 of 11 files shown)


spec/services/invoices/preview_service_spec.rb
```diff
@@ -66,6 +66,39 @@
...
+            amount_cents: 2,
+            amount_currency: plan.amount_currency
+          )
+        end
+
+        it 'creates preview invoice for 2 days with applied coupons' do
+          travel_to(timestamp) do
+            result = described_class.new(customer:, subscription:, applied_coupons: [applied_coupon]).call
+
+            aggregate_failures do
```
ancorcruz, line 84: no need to aggregate failures. already aggregates by default.

app/services/coupons/preview_service.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    def add_credit(applied_coupon, fees, base_amount_cents)
+      credit_amount = AppliedCoupons::AmountService.call(applied_coupon:, base_amount_cents:).amount
+      new_credit = Credit.new(
+        invoice:,
+        applied_coupon:,
+        amount_cents: credit_amount,
+        amount_currency: invoice.currency,
+        before_taxes: true
+      )
```
ancorcruz, line 49: when is the credit saved?

ancorcruz (changes requested):
left a few comments...

app/models/applied_coupon.rb
```diff
@@ -32,6 +32,13 @@ def mark_as_terminated!(timestamp = Time.zone.now)
     self.terminated_at ||= timestamp
     terminated!
   end
+
+  def remaining_amount
+    return @remaining_amount if defined?(@remaining_amount)
+
+    already_applied_amount = credits.sum(&:amount_cents)
+    @remaining_amount = amount_cents - already_applied_amount
```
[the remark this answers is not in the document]
lovrocolic: I am using version with defined? in order to memoize nil values as well which cannot be achieved with `||=`

app/services/coupons/preview_service.rb
lovrocolic: For preview invoice we compute everything on the fly. That is the main difference between this service and service that is called on real invoice. There are also several other differences (frequency or remaining amount is not tracked...)

spec/models/fee_spec.rb
```diff
@@ -403,6 +403,14 @@
     end
   end
 
+  describe 'compute_precise_credit_amount_cents' do
```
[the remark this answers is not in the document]
lovrocolic: added

spec/services/invoices/preview_service_spec.rb
lovrocolic: fixed on all places

app/services/coupons/preview_service.rb
ancorcruz: then it is expected to not persist the credits instantiated here.

ancorcruz (approved):
Looks great thanks for addressing all the feedback. 🚀

app/services/coupons/preview_service.rb
lovrocolic: exactly, this is expected not to store credit

PR #3075, opened 2025-01-20
feat(segment): Remove 8 unused events

julienbourdeau:
## Context

Lago sends events to Segment for analytics. 

## Description

As we use these analytics, we realized many events aren't useful so let's remove them.


brunomiguelpinto (approved):

vincent-pochet (approved):
image
❤️

PR #3079, opened 2025-01-20
chore(fees): Add unique index on event transaction id

jdenquin:
- To prevent duplicate invoicing for in advance charges, we add a unique index based on the `transaction_id` from the event

### DIFF
--- a/db/migrate/20250120151959_add_unique_event_index_on_fees.rb
+++ b/db/migrate/20250120151959_add_unique_event_index_on_fees.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddUniqueEventIndexOnFees < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :fees, %i[pay_in_advance_event_transaction_id charge_id charge_filter_id], unique: true,
+      where: "created_at > '#{Time.current}'", algorithm: :concurrently
+  end
+end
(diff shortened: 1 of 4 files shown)


db/migrate/20250120151959_add_unique_event_index_on_fees.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddUniqueEventIndexOnFees < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :fees, %i[pay_in_advance_event_transaction_id charge_id charge_filter_id], unique: true,
+      where: "created_at > '#{Time.current}'", algorithm: :concurrently
```
nudded, line 8: Why the added created at? NULL values don’t matter I think 🤔

vincent-pochet, line 8: We have to add `pay_in_advance = 't' and pay_in_advance_event_transaction_id is not null` as well I guess

vincent-pochet: @nudded We have duplicated values in the database, without this condition, the index will fail to create

jdenquin: also the migration CI is failing because the timestamp is different, do you know how to handle it?

nudded: Could we clear the `pay_in_advance_event_transaction_id` for one of the duplicates?

vincent-pochet: We are relying on it for many things in API and PDF templates, and since the fees were pushed to the customers for now we cannot really remove it :/

vincent-pochet (approved):

PR #3083, opened 2025-01-21
misc(adjusted_fee): Allow fee adjustement without pre-existing fee

vincent-pochet:
## Context

As a consequence of the removal of the 0 units fees removal from both DB and invoicing logic, we lost the ability to adjust fees without amount / units as the feature was expecting a fee to be present. This feature is a new approach to bring this functionality back.

## Description

This features refactors the `AdjustedFees::CreateService` and the `Mutations::AdjustedFees::Create` mutation to allow the creation of adjusted fees without pre-existing fees via the GraphQL API

### DIFF
--- a/app/graphql/types/invoices/object.rb
+++ b/app/graphql/types/invoices/object.rb
@@ -41,6 +41,8 @@ module Types
       field :payment_due_date, GraphQL::Types::ISO8601Date, null: false
       field :payment_overdue, Boolean, null: false
 
+      field :all_charges_have_fees, Boolean, null: false, method: :all_charges_have_fees?
+
       field :created_at, GraphQL::Types::ISO8601DateTime, null: false
       field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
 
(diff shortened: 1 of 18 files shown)


app/graphql/types/invoices/object.rb
```diff
@@ -40,6 +40,8 @@ class Object < Types::BaseObject
       field :payment_due_date, GraphQL::Types::ISO8601Date, null: false
       field :payment_overdue, Boolean, null: false
 
+      field :all_charges_have_fees, Boolean, null: false, method: :all_charges_have_fees?
```
julienbourdeau, line 43: Is this for the frontend to know if they should show/hide the "Add" button?

julienbourdeau (approved):
It's interesting that the service and mutation were called Create while it was really an Update!

app/graphql/types/invoices/object.rb
vincent-pochet: Yes exactly! @julienbourdeau

PR #3084, opened 2025-01-21
feat(invoice-preview): add preview support for wallet and credit note credits

lovrocolic:
## Context

Preview feature enables fetching first Lago invoice. This invoice is not persisted and is calculated on the fly.

## Description

This PR adds logic for applying wallet and credit note credits on preview invoice.

### DIFF
--- a/app/services/invoices/preview_service.rb
+++ b/app/services/invoices/preview_service.rb
@@ -107,6 +107,41 @@ module Invoices
       invoice.total_amount_cents = (
         invoice.sub_total_including_taxes_amount_cents - invoice.credit_notes_amount_cents
       )
+
+      create_credit_note_credits
+      create_applied_prepaid_credits
+    end
+
+    def create_credit_note_credits
+      credit_result = Credits::CreditNoteService.call(invoice:, context: :preview)
+      credit_result.raise_if_error!
+
+      refresh_amounts(credit_amount_cents: credit_result.credits.sum(&:amount_cents)) if credit_result.credits
+    end
+
+    def create_applied_prepaid_credits
+      return unless customer.persisted?
+      return unless wallet&.active?
+      return unless invoice.total_amount_cents&.positive?
+      return unless wallet.balance.positive?
+
+      amount_cents = if wallet.balance_cents <= invoice.total_amount_cents
+        wallet.balance_cents
+      else
+        invoice.total_amount_cents
+      end
+      invoice.prepaid_credit_amount_cents += amount_cents
+      refresh_amounts(credit_amount_cents: amount_cents)
+    end
+
+    def refresh_amounts(credit_amount_cents:)
+      invoice.total_amount_cents -= credit_amount_cents
+    end
+
+    def wallet
+      return @wallet if defined? @wallet
+
+      @wallet = customer.wallets&.active&.first
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/services/invoices/preview_service.rb
```diff
@@ -107,6 +107,41 @@ def compute_tax_and_totals
...
+    end
+
+    def refresh_amounts(credit_amount_cents:)
+      invoice.total_amount_cents -= credit_amount_cents
+    end
+
+    def wallet
+      return @wallet if defined? @wallet
+
+      @wallet = customer.wallets&.active&.first
```
floganz, line 144: I guess we might not need safe navigation here.
`customer.wallets.active` should return at least empty array on which we can call `.first` without worries.

app/services/invoices/preview_service.rb
```diff
@@ -107,6 +107,41 @@ def compute_tax_and_totals
...
+      else
+        invoice.total_amount_cents
+      end
+      invoice.prepaid_credit_amount_cents += amount_cents
+      refresh_amounts(credit_amount_cents: amount_cents)
+    end
+
+    def refresh_amounts(credit_amount_cents:)
+      invoice.total_amount_cents -= credit_amount_cents
+    end
```
floganz, line 139: Wdyt if we'll remove this method and do this subtraction inline?
This method does not improve readability or isolate complex logic but requires to guess when you see first time what it's doing.

spec/services/invoices/preview_service_spec.rb
```diff
@@ -93,6 +93,82 @@
...
+            expect(result.invoice.credit_notes_amount_cents).to eq(2)
+            expect(result.invoice.total_amount_cents).to eq(7)
+          end
+        end
+      end
+
+      context 'with wallet credits' do
+        let(:wallet) { build(:wallet, customer:, balance: '0.03', credits_balance: '0.03') }
+
+        before { wallet }
```
floganz, line 135: This `let` is not used anywhere, if we just need to create this data why not do it directly in the before block?

app/services/invoices/preview_service.rb
```diff
@@ -107,6 +107,41 @@ def compute_tax_and_totals
...
+
+      create_credit_note_credits
+      create_applied_prepaid_credits
+    end
+
+    def create_credit_note_credits
+      credit_result = Credits::CreditNoteService.call(invoice:, context: :preview)
+      credit_result.raise_if_error!
+
+      refresh_amounts(credit_amount_cents: credit_result.credits.sum(&:amount_cents)) if credit_result.credits
```
floganz, line 119: What the point into having this `if`?

`Credits::CreditNoteService` will always return `credits` on `result` either empty array of array of credits both will be `true`. If there's an error – it will raise on line above.
Otherwise we can always deduct `credit_result.credits.sum(&:amount_cents)` from total, if there're not credits `.sum(&:amount_cents)` will return 0.

app/services/invoices/preview_service.rb
```diff
@@ -107,6 +107,41 @@ def compute_tax_and_totals
...
+    end
+
+    def create_applied_prepaid_credits
+      return unless customer.persisted?
+      return unless wallet&.active?
```
floganz, line 124: Do we really need check that `wallet` is active? We already perform a search in `wallets.active` scope.

floganz (approved):
👍 
I've some ideas how we can simplify the code if you agree with them.

spec/services/invoices/preview_service_spec.rb
lovrocolic: This before block is common for the both context blocks, that's why I would rather keep it as it is

floganz: I missed that. Let's keep it.

PR #3088, opened 2025-01-21
fix(payment): Handle payment and invoice status update

vincent-pochet:
# Context

This PR is related to the recent changes introduced by https://github.com/getlago/lago-api/pull/2962 and https://github.com/getlago/lago-api/pull/2986 to rely on payment provider idem-potency logic

A new `payable_payment_status` was added to the Payment model, but this status is not updated when a webhook is received from a payment provider

# Description

This PR makes sure that the payment, payment request or all other payment related statuses are updated accordingly

### DIFF
--- a/app/services/invoices/payments/adyen_service.rb
+++ b/app/services/invoices/payments/adyen_service.rb
@@ -24,10 +24,15 @@ module Invoices
         result.invoice = payment.payable
         return result if payment.payable.payment_succeeded?
 
-        payment.update!(status:)
+        payment.status = status
 
-        invoice_payment_status = payment.payment_provider&.determine_payment_status(status)
-        update_invoice_payment_status(payment_status: invoice_payment_status)
+        payable_payment_status = payment.payment_provider&.determine_payment_status(payment.status)
+        if Payment::PAYABLE_PAYMENT_STATUS.include?(payable_payment_status)
+          payment.payable_payment_status = payable_payment_status
+        end
+        payment.save!
+
+        update_invoice_payment_status(payment_status: payable_payment_status)
 
         result
       rescue BaseService::FailedResult => e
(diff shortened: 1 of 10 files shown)


app/services/invoices/payments/adyen_service.rb
```diff
@@ -24,10 +24,15 @@ def update_payment_status(provider_payment_id:, status:, metadata: {})
...
 
-        payment.update!(status:)
+        payment.status = status
 
-        invoice_payment_status = payment.payment_provider&.determine_payment_status(status)
-        update_invoice_payment_status(payment_status: invoice_payment_status)
+        payable_payment_status = payment.payment_provider&.determine_payment_status(payment.status)
+        if payment.class::PAYABLE_PAYMENT_STATUS.include?(payable_payment_status)
+          payment.payable_payment_status = payable_payment_status
+        end
```
julienbourdeau, line 32: I find this section confusing.
Is it expected that we have multiple payment class so we need to use `payment.class::` over `Payment::` ?

If `determine_payment_status` doesn't return a element of PAYABLE_PAYMENT_STATUS, is it valid? If feels like we just dont update the payment status but continue like it's not important 🤔 

Is `determine_payment_status` expected to often return something else than `failed`, `succeeded`, `processing` or we just return the unknown status but it shouldn't happen too much.

vincent-pochet: @julienbourdeau Good catch for the `payment.class` :+1: It is now fixed.

Regarding the `determine_payment_status`, I don't want to touch the current implementation as it is used in many different places. It's clear that it is something that should be improved to at least return an error or a specific `invalid` status.

If the status is unexpected, it will fails to update the invoice status and will raise a `ValidationFailure`,  it is handled in the `update_invoice_payment_status` and we have test coverage for that.
The payment will indeed be updated with the received status but without the `payable_payment_status`.
That said, I think it never happened :D

I started to refactor the handling of payment processing, to extract every action in dedicated services, but it is still an ongoing task that will take some time.

julienbourdeau: Thank you for the details 👌

julienbourdeau (approved):

PR #3091, opened 2025-01-22
feat(invoice-preview): Add invoices preview context service

floganz:
## Context

Part of preview invoice feature.

## Description

Add service that contains all logic that transforms input params into customer, subscription and applied coupons needed for invoice preview.

### DIFF
--- a/app/services/invoices/preview_context_service.rb
+++ b/app/services/invoices/preview_context_service.rb
@@ -1,0 +1,107 @@
+# frozen_string_literal: true
+
+module Invoices
+  class PreviewContextService < BaseService
+    def initialize(organization, params:)
+      @organization = organization
+      @params = params.presence || {}
+      super(nil)
+    end
+
+    def call
+      result.customer = find_or_build_customer
+      result.subscription = build_subscription
+      result.applied_coupons = find_or_build_applied_coupons
+      result
+    rescue ActiveRecord::RecordNotFound => exception
+      result.not_found_failure!(resource: exception.model.demodulize.underscore)
+    ensure
+      result
+    end
+
+    private
+
+    attr_reader :params, :organization
+
+    def find_or_build_customer
+      customer_params = params[:customer] || {}
+
+      if customer_params.key?(:external_id)
+        organization.customers.find_by!(external_id: customer_params[:external_id])
+      else
+        Customer.new(
+          name: customer_params[:name],
+          organization: organization,
+          tax_identification_number: customer_params[:tax_identification_number],
+          currency: customer_params[:currency],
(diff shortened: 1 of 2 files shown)


app/services/invoices/preview_context_service.rb
```diff
@@ -0,0 +1,107 @@
...
+      organization.plans.find_by!(code: params[:plan_code])
+    end
+
+    def find_or_build_applied_coupons
+      applied_coupons = result.customer.applied_coupons.presence
+
+      applied_coupons || Array(params[:coupons]).map do |coupon_attr|
+        coupon = coupon_attr.key?(:code) && organization.coupons.find_by(code: coupon_attr[:code])
+        coupon || Coupon.new(coupon_attr)
+      end.map { |coupon| AppliedCoupon.new(coupon:, customer: result.customer) }
```
lovrocolic, line 89: We also need to assign all coupon attributes on applied coupon. Please take a look at `applied_coupons/create_service.rb`

floganz: What should we do in case if params hash were provided
```
{
   code: "coupon_preview",
   name: "coupon_preview",
   coupon_type: "percentage",
   amount_cents: 1200,
   amount_currency: "EUR",
   percentage_rate: 1
 }
```
And coupon with this code exists. Should we use coupon's values from DB to fill on applied coupon or params provided by user?

lovrocolic: Hmm, probably that wouldn't be common case, but maybe it makes sense to first use applied coupons from DB (per specs). If there are no applied coupons in DB, we should build it based on params

lovrocolic: Usually you cannot override in the app applied coupons, you can only override coupon values when you create applied coupons, so IMO it makes sense to use values from the DB first

app/services/invoices/preview_context_service.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+
+module Invoices
+  class PreviewContextService < BaseService
+    def initialize(organization, params:)
+      @organization = organization
+      @params = params.presence || {}
+      super(nil)
```
vincent-pochet, line 8: This should be enough:
```suggestion
      super
```

vincent-pochet (approved):
:rocket:

app/services/invoices/preview_context_service.rb
floganz: That was needed since `.initialize` accepted not only keyword parameters. I've made those consistent and removed `(nil)`.

PR #3098, opened 2025-01-23
Fix(rev-share): fix bugs found during QA

annvelents:
## Context

There were small issues found during the QA

## Description

Fixed invoice numbering switching from customer-based to per-organization
Do not allow edit customers via API if they are not editable

### DIFF
--- a/app/services/customers/create_service.rb
+++ b/app/services/customers/create_service.rb
@@ -57,7 +57,7 @@ module Customers
         customer.lastname = params[:lastname] if params.key?(:lastname)
         customer.customer_type = params[:customer_type] if params.key?(:customer_type)
 
-        if customer.organization.revenue_share_enabled?
+        if customer.organization.revenue_share_enabled? && customer.editable?
           customer.account_type = params[:account_type] if params.key?(:account_type)
           customer.exclude_from_dunning_campaign = customer.partner_account?
         end
--- a/app/services/customers/create_service.rb
+++ b/app/services/customers/create_service.rb
@@ -174,7 +174,7 @@ module Customers
         customer_type: args[:customer_type]
       )
 
-      if customer&.organization&.revenue_share_enabled? && customer&.editable?
+      if customer&.organization&.revenue_share_enabled?
         customer.account_type = args[:account_type] if args.key?(:account_type)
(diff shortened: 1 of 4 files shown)


brunomiguelpinto (approved):

app/services/customers/create_service.rb
```diff
@@ -57,7 +57,7 @@ def create_from_api(organization:, params:)
         customer.lastname = params[:lastname] if params.key?(:lastname)
         customer.customer_type = params[:customer_type] if params.key?(:customer_type)
 
-        if customer.organization.revenue_share_enabled?
+        if customer.organization.revenue_share_enabled? && customer.editable?
```
ancorcruz, line 60: I would also include the "editable?" check in the other public method `create` around line 178 of this same file, and also cover it with tests...

spec/scenarios/invoices/invoice_numbering_spec.rb
```diff
@@ -746,4 +746,119 @@
...
+        invoices = organization.invoices.order(created_at: :desc).limit(3)
+        sequential_ids = invoices.pluck(:sequential_id)
+        organization_sequential_ids = invoices.pluck(:organization_sequential_id)
+        numbers = invoices.pluck(:number)
+
+        expect(sequential_ids).to match_array([5, 5, 5])
+        expect(organization_sequential_ids).to match_array([0, 0, 0])
+        expect(numbers).to match_array(%w[ORG-11-001-005 ORG-11-002-005 ORG-11-003-005])
+      end
+    end
```
ancorcruz, line 861: wasn't the QA return the other way around... first start with invoice numbering per organization, then change to per customer, and back again per organization????

ancorcruz (approved):
Good stuff, I have left you a few comments.

spec/scenarios/invoices/invoice_numbering_spec.rb
annvelents: The problem in QA was switching from per customer to per organization, as partner's are always per_customer, but counted separately. That's why I added couple rounds of billing per customer, then switch to organization to see this switch working, and back to customer... but I will add one more switch back to organization 👍

PR #3104, opened 2025-01-24
fix(payment): Ensure payment status is checkd as a string

vincent-pochet:
## Description

This PR is related to https://github.com/getlago/lago-api/pull/3088 it ensure that none of the payment `status`, `payable_payment_status` and related payables's `payment_status` are updated if an unknown status is received from a payment provider


ivannovosad (approved):
🚀

PR #3107, opened 2025-01-27
feat(stripe): Save Payment Method in Customer when paying invoice with checkout url

julienbourdeau:
## Context

Recently we added the possibility to pay an invoice with a stripe checkout url. The payment method used wasn't saved in stripe. This change will allow the customer to reuse this method next time.

## Description

When using the `/customers/:external_customer_id/checkout_url` you setup a payment method so the payment method is associated with the customer in Stripe.

When using `/invoices/:lago_id/payment_url` the payment method is used but not associated. This PR makes sure the payment method used is savec with the customer.

Related docs: https://docs.stripe.com/api/checkout/sessions/create#create_checkout_session-payment_intent_data-setup_future_usage

For now, we don't support offsession paiement in India so we don't et this parameter in case the customer is in India.

Note that **it's different** from saving your card with Link (as shown below).

CleanShot 2025-01-27 at 15 12 32@2x

### DIFF
--- a/app/services/invoices/payments/stripe_service.rb
+++ b/app/services/invoices/payments/stripe_service.rb
@@ -136,6 +136,7 @@ module Invoices
           payment_method_types: customer.stripe_customer.provider_payment_methods,
           payment_intent_data: {
             description:,
+            setup_future_usage: off_session? ? "off_session" : nil, # save payment method for future use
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):
:+1:

PR #3109, opened 2025-01-27
misc(billing): Split billing job per organization

vincent-pochet:
## Description

Today, the `Subscriptions::BillingService`, called by the `Clock::SubscriptionsBillerJob` fetches all subscription that have to be billed at a certain moment of time.

Since the number of subscription to bill is growing quickly, this could lead to a very large number of subscription to load in certain condition (mainly on the first day of each month as most of subscription are billed monthly on a calendar basis)

This PR renames the `Subscriptions::BillingService` into `Subscriptions::OrganizationBillingService` and adds a new `organization` argument, to loop over the billable subscription of an organization instead of all of the application.

The `Clock::SubscriptionsBillerJob` is also updated to perform one job per organization

### DIFF
--- a/app/jobs/clock/subscriptions_biller_job.rb
+++ b/app/jobs/clock/subscriptions_biller_job.rb
@@ -5,7 +5,7 @@ module Clock
     include SentryCronConcern
 
     queue_as do
-      if ActiveModel::Type::Boolean.new.cast(ENV['SIDEKIQ_CLOCK'])
+      if ActiveModel::Type::Boolean.new.cast(ENV["SIDEKIQ_CLOCK"])
         :clock_worker
       else
         :clock
@@ -13,7 +13,9 @@ module Clock
     end
 
     def perform
-      Subscriptions::BillingService.call
+      Organization.find_each do |organization|
+        Subscriptions::OrganizationBillingJob.perform_later(organization:)
+      end
     end
   end
 end
--- a/app/jobs/subscriptions/organization_billing_job.rb
+++ b/app/jobs/subscriptions/organization_billing_job.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  class OrganizationBillingJob < ApplicationJob
+    queue_as do
+      if ActiveModel::Type::Boolean.new.cast(ENV["SIDEKIQ_CLOCK"])
+        :clock_worker
+      else
+        :clock
+      end
+    end
+
+    unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
+
+    def perform(organization:)
+      Subscriptions::OrganizationBillingService.call!(organization:)
+    end
+  end
+end
--- a/app/services/subscriptions/organization_billing_service.rb
+++ b/app/services/subscriptions/organization_billing_service.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 module Subscriptions
-  class BillingService < BaseService
-    def initialize(billing_at: Time.current)
+  class OrganizationBillingService < BaseService
+    def initialize(organization:, billing_at: Time.current)
+      @organization = organization
       @today = billing_at
 
       super
@@ -29,11 +30,13 @@ module Subscriptions
 
         BillNonInvoiceableFeesJob.perform_later(billing_subscriptions, today)
       end
+
+      result
     end
 
     private
 
-    attr_reader :today
+    attr_reader :today, :organization
 
     # NOTE: Retrieve list of subscriptions that should be billed today
     def billable_subscriptions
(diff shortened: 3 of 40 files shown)


app/jobs/clock/subscriptions_biller_job.rb
```diff
@@ -5,15 +5,17 @@ class SubscriptionsBillerJob < ApplicationJob
...
         :clock_worker
       else
         :clock
       end
     end
 
     def perform
-      Subscriptions::BillingService.call
+      Organization.find_each do |organization|
+        Subscriptions::OrganizationBillingJob.perform_later(organization:)
```
rsempe, line 17: what about:
```suggestion
        Organizations::BillSubscriptionsJob.perform_later(organization:)
```

app/jobs/subscriptions/organization_billing_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  class OrganizationBillingJob < ApplicationJob
+    queue_as :default
+
+    def perform(organization:)
+      Subscriptions::OrganizationBillingService.call!(organization:)
```
rsempe, line 8: and:
```suggestion
      Organizations::BillSubscriptionsService.call!(organization:)
```

rsempe (approved):
nice one 👍

app/services/subscriptions/organization_billing_service.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 module Subscriptions
-  class BillingService < BaseService
```
julienbourdeau, line 4: Renaming (or changing args of) the job class will make all already enqueued jobs fail.

Maybe it's fine because this queue is usually empty (process quucky and I'm sure nobody upgrades of the first of the month but I want to confirm we're ok with it.

Keeping the job around is not necessarily a great solution either.

spec/scenarios/commitments/minimum/in_advance/anniversary/monthly_spec.rb
```diff
@@ -130,7 +130,7 @@
         }
       )
 
-      Subscriptions::BillingService.new.call
+      Subscriptions::OrganizationBillingService.call(organization:)
       perform_all_enqueued_jobs
```
julienbourdeau, line 134: I was expecting a lot more jobs to use the `perform_billing` helper. No need to change this tho.

julienbourdeau (approved):
Excellent improvement! 🚀

app/jobs/subscriptions/organization_billing_job.rb
vincent-pochet: I though about it, but since the main entities we are dealing with in the service are subscriptions and not the organization (just a filter), I decided to keep the namespace

app/services/subscriptions/organization_billing_service.rb
vincent-pochet: Yes indeed. Since the job is retried every hour by the clock anyway, I think in that specific case, we can skip the "safe renaming approach"

spec/scenarios/commitments/minimum/in_advance/anniversary/monthly_spec.rb
vincent-pochet: Haa, good catch! I think I will replace the calls with this. It's a good opportunity to improve the specs :slightly_smiling_face:

app/jobs/subscriptions/organization_billing_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  class OrganizationBillingJob < ApplicationJob
+    queue_as :default
```
nudded, line 5: is `default` the right queue for this job? 🤔

vincent-pochet: @nudded 

Previously, the logic identifying the subscription to bill was done in the `clock` queue (`Clock::SubscriptionsBillerJob` was performing the `Subscriptions::BillingService.call` dirrectly). For now it only enqueues a job per organization in the `default` queue.
I don't want to add it in the `billing` queue as it has to have a higher priority that the `BillSubscriptionJob` to make sure it will not be stuck for hours every 1st days of the month

Do you think it should be pushed to the `high_priority` queue?, or to a new one (`medium_high_priority`)?

nudded: hmmm, perhaps these new organization jobs could also be clock jobs? They are run every hour right? 🤔 
(that would put them on clock or clock_worker queue).

To me the enqueuing of these jobs feels important, I don't want them to be stuck anywhere, so I'm inclined to even consider high_priority for them? (e.g. this one should be done fast and quick, the billing can be done asynchronous on the billing queue)

vincent-pochet: I'm not against high priority as I agree, it has to be performed quickly

PR #3115, opened 2025-01-28
fix(current usage): fix current usage taxes

lovrocolic:
Initially, when fetching taxes from Anrok we used `fee.item_id` (`add_on_id`, `billable_metric_id`, `subscription_id`) as main identifier that helps us map received taxes with certain fee on Lago side.

First version worked well for real invoices, but also for current usage where we don't have ID persisted on fee.

However, there was a case where multiple charges are related to the same billable_metric (especially when one fee amount is zero and another is positive) and we were not able to perform mapping correctly: https://github.com/getlago/lago-api/pull/2842

However, one regression bug was produced with latest changes. Since `fee.id` is nil for current usage, taxes are not applied correctly in some scenarios.

This PR fixes described case

### DIFF
--- a/app/services/integrations/aggregator/taxes/invoices/payload.rb
+++ b/app/services/integrations/aggregator/taxes/invoices/payload.rb
@@ -47,7 +47,7 @@ module Integrations
             mapped_item ||= empty_struct
 
             {
-              'item_id' => fee.id,
+              'item_id' => fee.id || fee.item_id,
               'item_code' => mapped_item.external_id,
               'amount_cents' => fee.sub_total_excluding_taxes_amount_cents&.to_i
             }
--- a/app/services/invoices/customer_usage_service.rb
+++ b/app/services/invoices/customer_usage_service.rb
@@ -153,8 +153,6 @@ module Invoices
       taxes_result = Rails.cache.read(provider_taxes_cache_key)
 
       unless taxes_result
-        invoice.fees.each { |f| f.id = SecureRandom.uuid }
-
         # Call the service if the cache is empty
         taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateDraftService.call(invoice:, fees: invoice.fees)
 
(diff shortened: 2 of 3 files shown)


app/services/invoices/customer_usage_service.rb
```diff
@@ -165,7 +165,9 @@ def compute_amounts_with_provider_taxes
       result.fees_taxes = taxes_result.fees
 
       invoice.fees.each do |fee|
-        fee_taxes = result.fees_taxes.find { |item| item.item_id == fee.id }
+        fee_taxes = result.fees_taxes.find do |item|
+          (item.item_id == fee.item_id) && (item.amount_cents.to_i == fee.sub_total_excluding_taxes_amount_cents&.to_i)
```
nudded, line 169: would it not be possible to generate some ids on the fees the can be used for mapping? This approach feels very fragile. 

Just using unique IDs will solve all edge cases 🤔

lovrocolic: Refactored 👍

app/services/integrations/aggregator/taxes/invoices/payload.rb
```diff
@@ -47,7 +47,7 @@ def fee_item(fee)
             mapped_item ||= empty_struct
 
             {
-              'item_id' => fee.id,
+              'item_id' => fee.id || fee.item_id,
```
lovrocolic, line 50: `|| fee.item_id` part is still needed for one-off part where taxes are applied on the UI form (unrelated to invoice generation). In that case, `item_id` is identifier which UI uses to find matching add_on in the form (add_od_id is familiar data on FE side). This issue is also regression so `fee.id || fee.item_id` covers all cases now without any FE adjustments and bigger refactorings

app/services/invoices/customer_usage_service.rb
lovrocolic: Just realised that with generating uuid on the fly, we use possibility to cache tax provider result :/ We need to use some combination of attributes on the fee if we want to keep caching available and not ping Anrok too often for current usage. So I reverted my latest commit where ID is set to `SecureRandom.uuid` on the fly
Even if current solution seems a bit hacky it covers well current usage cases + everything else should work as it is.
Let me know if you have something else in mind

nudded: Okay, that's a good point 👍

nudded (approved):

PR #3121, opened 2025-01-30
Feat/terminate orga add orga id credit notes

ancorcruz:
## Context

This change makes it easier and more feasible to implement a mechanism for terminating an organization and deleting all its data. By explicitly associating `CreditNote` records with `Organization`, we improve data consistency and simplify tenant-scoped deletions.  

Beyond the primary goal, this change also benefits the multi-tenant application in several ways:  
- Improves data integrity by making the organization-tenant relationship explicit.  
- Enables more efficient queries and filtering at the organization level.  
- Aligns `CreditNote` with other models that are already explicitly tied to `Organization`.  

## Description
- Added an `organization_id` column to `credit_notes`, with an index and foreign key constraint.  
- Updated `CreditNotes::CreateService` to set `organization_id` when creating a `CreditNote`.
- Refactored event tracking to reference `credit_note.organization_id` directly.  
- Updated specs to assert the presence of `organization_id`.  

### Impact
- Existing `CreditNote` records will need backfilling (handled separately if needed).  
- Queries can now leverage `organization_id` for better performance.  
- Future work on organization deletion is now more straightforward.  

### Next up: 
- backfilling
- marking the column `not_null`

### DIFF
--- a/app/services/credit_notes/create_service.rb
+++ b/app/services/credit_notes/create_service.rb
@@ -23,6 +23,7 @@ module CreditNotes
 
       ActiveRecord::Base.transaction do
         result.credit_note = CreditNote.create!(
+          organization_id: invoice.organization_id,
           customer: invoice.customer,
           invoice:,
           issuing_date:,
@@ -155,7 +156,7 @@ module CreditNotes
         membership_id: CurrentContext.membership,
         event: 'credit_note_issued',
         properties: {
-          organization_id: credit_note.organization.id,
+          organization_id: credit_note.organization_id,
           credit_note_id: credit_note.id,
           invoice_id: credit_note.invoice_id,
           credit_note_method: types
--- a/db/migrate/20250130101122_add_organization_id_to_credit_notes.rb
+++ b/db/migrate/20250130101122_add_organization_id_to_credit_notes.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddOrganizationIdToCreditNotes < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :credit_notes, :organization, type: :uuid, index: {algorithm: :concurrently}
+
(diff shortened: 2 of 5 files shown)


db/migrate/20250130101122_add_organization_id_to_credit_notes.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddOrganizationIdToCreditNotes < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :credit_notes, :organization, type: :uuid, index: {algorithm: :concurrently}
+
+    add_foreign_key :credit_notes, :organizations, validate: false
```
nudded, line 9: I think every statement needs to be its own migration.

app/services/credit_notes/create_service.rb
```diff
@@ -23,6 +23,7 @@ def call
 
       ActiveRecord::Base.transaction do
         result.credit_note = CreditNote.create!(
+          organization_id: invoice.organization_id,
```
nudded, line 26: There is also the EstimateService where a credit note is created.

ancorcruz: yep, thanks. I have missed it as it calls `CreditNote.new` instead of `.create`.

db/migrate/20250130101122_add_organization_id_to_credit_notes.rb
ancorcruz: okay, not sure about this, strong migrations has not complained so far.... why it needs to be in various migrations?

nudded: the reason would be to limit the amount of statements that are executed without transaction. I don't think it matters a lot, but I believe it's good practice to split the migration into multiple files :)

PR #3122, opened 2025-01-30
misc: Introduce BaseResult to get rid of OpenStruct

vincent-pochet:
## Context

As a requirement to enable YJIT compiler for Ruby, we need to get rid of OpenStruct as it leads to a huge memory consumption.

At the same time, we are lacking visibility on what properties are attached to result objects returned by services.

## Description

This PR introduces a new `BaseResult` result object, that should be used a a base in every services to define the attributes of a Result:
```ruby
module SomeModule
  class SomeService < BaseService
    Result = BaseResult[:attr_name1, :attr_name2]
```

To keep compatibility until all services are migrated to this approach, the `BaseService` is still relying on a `LegacyResult` using OpenStruct as a parent class.

### DIFF
--- a/app/services/add_ons/apply_taxes_service.rb
+++ b/app/services/add_ons/apply_taxes_service.rb
@@ -2,6 +2,8 @@
 
 module AddOns
   class ApplyTaxesService < BaseService
+    Result = BaseResult[:applied_taxes]
+
     def initialize(add_on:, tax_codes:)
       @add_on = add_on
       @tax_codes = tax_codes
--- a/app/services/base_result.rb
+++ b/app/services/base_result.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+class BaseResult
+  def self.[](*attributes)
+    Class.new(BaseResult) { attr_accessor(*attributes) }
+  end
+
+  attr_reader :error
+
+  def initialize
+    @failure = false
+    @error = nil
+  end
+
+  def failure?
+    failure
+  end
+
+  def success?
+    !failure
+  end
+
+  def fail_with_error!(error)
+    @failure = true
+    @error = error
+
+    self
+  end
+
+  def not_found_failure!(resource:)
+    fail_with_error!(BaseService::NotFoundFailure.new(self, resource:))
+  end
+
+  def not_allowed_failure!(code:)
+    fail_with_error!(BaseService::MethodNotAllowedFailure.new(self, code:))
+  end
+
+  def record_validation_failure!(record:)
+    validation_failure!(errors: record.errors.messages)
+  end
+
+  def validation_failure!(errors:)
+    fail_with_error!(BaseService::ValidationFailure.new(self, messages: errors))
+  end
+
+  def single_validation_failure!(error_code:, field: :base)
+    validation_failure!(errors: {field.to_sym => [error_code]})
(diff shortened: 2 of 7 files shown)


app/services/add_ons/apply_taxes_service.rb
```diff
@@ -2,6 +2,8 @@
 
 module AddOns
   class ApplyTaxesService < BaseService
+    Result = BaseResult.define_with_attributes(:applied_taxes)
```
nudded, line 5: nitpick: I'd rather have `Result = BaseResult[:applied_taxes]`   as syntax? (I think the `define_with_attributes is a bit much and shorter would be nicer)

julienbourdeau:
Awesome! Being able to see what's expected in the result will be a tremendous help.

Just thinking out loud about the syntax. How do you feel about using [`ActiveModel::Attributes`](https://api.rubyonrails.org/classes/ActiveModel/Attributes.html)?

```ruby
module AddOns
  class ApplyTaxesService < BaseService
    class Result < BaseResult
      attribute :add_on
      attribute :something, default: true
    end

    def call
      # 
    end
  end
end
```

app/services/add_ons/apply_taxes_service.rb
julienbourdeau: It's not nitpicking, it's important to find the best syntax 😆 That's why we love Ruby

vincent-pochet: I agree, it's nicer, I will try it :+1:

vincent-pochet:
> Awesome! Being able to see what's expected in the result will be a tremendous help.
> 
> Just thinking out loud about the syntax. How do you feel about using [`ActiveModel::Attributes`](https://api.rubyonrails.org/classes/ActiveModel/Attributes.html)?
> 
> ```ruby
> module AddOns
>   class ApplyTaxesService < BaseService
>     class Result < BaseResult
>       attribute :add_on
>       attribute :something, default: true
>     end
> 
>     def call
>       # 
>     end
>   end
> end
> 
> def BaseResult
>   include ActiveModel::Attributes
> 
>   # ...
> end
> ```

I like the idea, but I see to problems with that:
1. It leads to a "heavier" definitions, and I would prefer keeping a short syntax (more a matter or personal preference here :slightly_smiling_face: )
2. ActiveModel::Attributes is built for basic, but most of our return objects are model instances or array of model instances, so I'm not sure it would bring much value (+ we have no need for validation as it must be handled at service level)

app/services/base_result.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+class BaseResult
+  def self.[](*attributes)
+    Class.new(BaseResult) { attr_accessor(*attributes) }
```
nudded, line 5: Considered using `Struct` for this vs just class? 🤔

vincent-pochet: Struct does not offers setters, so it would make it harder to use as it would require us to pass the result attributes when creating the instance of the result.

julienbourdeau (approved):
🙌 🙌 🙌

annvelents (approved):

jdenquin (approved):

nudded (approved):

PR #3123, opened 2025-01-30
feat(wallet): New `wallet.created` webhooks

julienbourdeau:
## Context

Wallet operations trigger 3 webhooks today: 
* `wallet.depleted_ongoing_balance`
* `wallet_transaction.updated`
* `wallet_transaction.payment_failure`.

We want to add the following: 
* `wallet.created`
* `wallet.updated`
* `wallet.terminated`

https://docs.getlago.com/api-reference/webhooks/messages#wallets-and-prepaid-credits

### `wallet.created`

Notice that the webhook job is dispatched by `WalletTransactions::CreateJob` so the wallet sent in the webhook has the initial balance instead of 0. Unlike the API which returns 0.

The `WalletTransactions::CreateJob` is always dispatch even if the `WalletTransactions::CreateService` might do nothing.

### `wallet.terminated`

The webhook is set only of the wallet just got terminated. If the service is called on a terminated service, no webhook is sent.

### `wallet.updated`

This webhook is dispatched when 
* an attribute of the wallet is updated (name, expiration date or invoice_requires_successful_payment config)
* a recurring transaction rule is updated because recurring **rules are part of a wallet**, unlike wallet_transactions which are "related to a wallet" (and have their own webhooks)
* the balance is increased or decreased

**No webhook is sent when the ongoing balance is updated** because it would trigger too many events. Tracking ongoing balance is part of the "real time usage" effort.

### DIFF
--- a/spec/support/queues_helper.rb
+++ b/spec/support/queues_helper.rb
@@ -8,4 +8,16 @@ module QueuesHelper
       :webhook
     end
   end
+
+  # This performs any enqueued-jobs, and continues doing so until the queue is empty.
+  # Lots of the jobs enqueue other jobs as part of their work, and this ensures that
+  # everything that's supposed to happen, happens.
+  #
+  # ⚠️ Notice that `have_been_enqueued` does not work with perform_all_enqueued_jobs because of the loop.
+  def perform_all_enqueued_jobs
+    until enqueued_jobs.empty?
+      perform_enqueued_jobs
+      Sidekiq::Worker.drain_all
+    end
+  end
 end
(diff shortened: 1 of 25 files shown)


spec/support/queues_helper.rb
```diff
@@ -8,4 +8,17 @@ def webhook_queue
...
     end
   end
+
+  # This performs any enqueued-jobs, and continues doing so until the queue is empty.
+  # Lots of the jobs enqueue other jobs as part of their work, and this ensures that
+  # everything that's supposed to happen, happens.
+  #
+  # ⚠️ Notice that `have_been_enqueued` might not work with perform_all_enqueued_jobs
+  # because it's only aware of the last run of the loop.
+  def perform_all_enqueued_jobs
```
julienbourdeau, line 18: I moved `perform_all_enqueued_jobs` outside `ScenariosHelper` since I'm using it outside the scenario folder.

julienbourdeau:
closing because `/` in branch name prevent me from deploying staging env 😬

PR #3124, opened 2025-01-31
Allow to override existing customer attributes

floganz:
## Context

Adjustments to invoice preview input params.

## Description

Validate that "coupons" params are array.
Allow to override customer attributes.
Get rid of `shipping_address` wrapper.

### DIFF
--- a/app/controllers/api/v1/invoices_controller.rb
+++ b/app/controllers/api/v1/invoices_controller.rb
@@ -203,6 +203,16 @@ module Api
       end
 
       def preview
+        if preview_params[:coupons] && !preview_params[:coupons].is_a?(Array)
+          return render(
+            json: {
+              status: 400,
+              error: "coupons_must_be_an_array"
+            },
+            status: :bad_request
+          )
+        end
+
         result = Invoices::PreviewContextService.call(
           organization: current_organization,
           params: preview_params.to_h.deep_symbolize_keys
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/invoices_controller.rb
```diff
@@ -281,14 +291,12 @@ def preview_params
...
-              :zipcode,
-              :state,
-              :country
-            ],
+            :address_line1,
+            :address_line2,
+            :city,
+            :zipcode,
+            :state,
+            :country,
```
julienbourdeau, line 299: As far as I understand, `address_line1` and `shipping_address_line1` are 2 separate attributes, so her we're not "removing the nesting", we change from shipping address to main address. Correct?

julienbourdeau (approved):

app/controllers/api/v1/invoices_controller.rb
```diff
@@ -281,14 +291,12 @@ def preview_params
             :tax_identification_number,
             :currency,
             :timezone,
-            shipping_address: [
```
lovrocolic, line 284: Don't we want to keep `shipping_address` as well? Please take. look at this file: https://github.com/getlago/lago-api/blob/main/app/services/integrations/aggregator/taxes/invoices/payload.rb

lovrocolic (approved):

PR #3125, opened 2025-01-31
Feat(manual payments): Support for Partial/Manual Payments

brunomiguelpinto:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/log-partial-payments

## Context

Currently, Lago only supports three invoice payment statuses: pending, succeeded, and failed, without any distinction for partial payments. This causes issues for users tracking invoices paid partially via bank transfers or manual payments.

The lack of a partial payment status forces users to track partially paid invoices manually through metadata or external tools like Slack or spreadsheets, making it harder to manage overdue balances and follow-up during the dunning process.

Additionally:
	•	Manual payments (bank transfers, cash, etc.) were not properly handled because they required a provider_payment_id.
	•	Users had no visibility into how much of the invoice was covered by payments (total_paid_amount_cents), making overdue_balance calculations inaccurate.

## Description

Users can log manual payments directly in the system, which automatically updates the invoice’s total_paid_amount_cents. Invoice statuses transition seamlessly: pending → partial → succeeded as payments progress. Additionally, overdue balance calculations, invoice serializers, and analytics now factor in partial payments, ensuring more accurate reporting and a clearer dunning process.


app/services/payment_requests/create_service.rb
```diff
@@ -18,7 +18,7 @@ def call
         customer.payment_requests.create!(
           organization:,
           dunning_campaign:,
-          amount_cents: invoices.sum(:total_amount_cents),
+          amount_cents: invoices.sum('total_amount_cents - total_paid_amount_cents'),
```
julienbourdeau: Are you sure this works? TIL

julienbourdeau (approved):

app/services/payments/manual_create_service.rb
```diff
@@ -57,9 +55,9 @@ def invoice
     end
 
     def check_preconditions
-      return result.forbidden_failure! unless License.premium?
       return result.not_found_failure!(resource: "invoice") unless invoice
-      return result.forbidden_failure! unless invoice.organization.premium_integrations.include?('manual_payments')
+      return if invoice.invoice_type == "advance_charges"
```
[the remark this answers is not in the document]
brunomiguelpinto: We don't need to check the rest when invoiced.invoice_type == "advance_charges"

app/services/invoices/advance_charges_service.rb
```diff
@@ -24,6 +24,7 @@ def call
 
       if invoice && !invoice.closed?
         SendWebhookJob.perform_later("invoice.created", invoice)
+        create_manual_payment(invoice)
```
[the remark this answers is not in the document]
brunomiguelpinto: yes

app/services/payment_requests/create_service.rb
brunomiguelpinto: we tested payment requests during the qa we didn't found any issues on this

app/services/payment_requests/payments/adyen_service.rb
```diff
@@ -51,6 +52,7 @@ def update_payment_status(provider_payment_id:, status:, metadata: {})
 
         update_payable_payment_status(payment_status: payable_payment_status)
         update_invoices_payment_status(payment_status: payable_payment_status)
+        update_invoices_paid_amount_cents(payment_status: payable_payment_status)
```
vincent-pochet, line 55: Out of the scope of this PR, but seeing this code repeated 5 times clearly shows that we should refactor these services to factorize the common logic in a root service and extract the logic specific to the payment provider in a sub services (in a similar way of what was done for payment creation)

app/services/payments/manual_create_service.rb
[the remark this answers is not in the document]
vincent-pochet: Then you should return an empty result (without setting the payment/invoice) so that it does not break the contract of the service

vincent-pochet (approved):
It looks good.
For future review, I would deeply suggest to split these kinds of feature in multiple independent sub-PRs. It is really complex to do a good review on a PR with more than 100 updated files

db/migrate/20250212123207_backfill_invoices_and_payments.rb
```diff
@@ -0,0 +1,37 @@
...
+    update_payments
+  end
+
+  def down
+  end
+
+  private
+
+  def update_invoices
+    Invoice.where(payment_status: 1).update_all("total_paid_amount_cents = total_amount_cents") # rubocop:disable Rails/SkipsModelValidations
```
vincent-pochet, line 17: I'm wondering if we should redefine the models with the required relations and constants inside this migration to avoid future migration errors on fresh install if we rename/remove or change some models

brunomiguelpinto: yes makes sense i will pass the models with the required relations and constants inside this migration

PR #3126, opened 2025-01-31
fix(invoice-preview): disable preview feature for freemium users

lovrocolic:
Invoice preview should be enabled only for premium users

### DIFF
--- a/spec/services/invoices/preview_service_spec.rb
+++ b/spec/services/invoices/preview_service_spec.rb
@@ -26,123 +26,44 @@ RSpec.describe Invoices::PreviewService, type: :service, cache: :memory do
 
     before { tax }
 
-    context 'when customer does not exist' do
-      it 'returns an error' do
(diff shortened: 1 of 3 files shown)


spec/services/invoices/preview_service_spec.rb
```diff
@@ -26,123 +26,44 @@
...
-        expect(result).not_to be_success
-        expect(result.error.error_code).to eq('subscription_not_found')
-      end
-    end
-
-    context 'with calendar billing' do
-      it 'creates preview invoice for 2 days' do
-        # Two days should be billed, Mar 30 and Mar 31
-
+    context 'with Lago freemium' do
```
floganz, line 29: Probably it's typo – it's context where it free version.

floganz (approved):

spec/services/invoices/preview_service_spec.rb
lovrocolic: This is how we call free version:)

PR #3132, opened 2025-02-04
chore(versions): Bump redis related gems and configure reconnection in redlock

nudded:
## Description

We're sometimes getting errors related to Redis connections. based on https://github.com/redis-rb/redis-client/issues/116#issuecomment-1534406659 we shoud configure reconnection attempts to prevent these.


vincent-pochet (approved):
:+1:

PR #3133, opened 2025-02-04
feat(wallet): New `wallet.(created|updated|terminated)` webhooks

julienbourdeau:
## Context

Wallet operations trigger 3 webhooks today: 
* `wallet.depleted_ongoing_balance`
* `wallet_transaction.updated`
* `wallet_transaction.payment_failure`.

We want to add the following: 
* `wallet.created`
* `wallet.updated`
* `wallet.terminated`

https://docs.getlago.com/api-reference/webhooks/messages#wallets-and-prepaid-credits

### `wallet.created`

~Notice that the webhook job is dispatched by `WalletTransactions::CreateJob` so the wallet sent in the webhook has the initial balance instead of 0. Unlike the API which returns 0.~
~The `WalletTransactions::CreateJob` is always dispatch even if the `WalletTransactions::CreateService` might do nothing.~

Notice that the webhook job is dispatched as soon as the wallet is created so the initial balance can be 0. The API always returns balance = 0 when creating a wallet.

The initial balance is set async by the `WalletTransactions::CreateJob` job and it runs in the `high_priority` queue.

> [!NOTE]
Creating a wallet transaction to set the initial balance will also send a `wallet.updated` webhook. It's possible to receive this one **before** the `wallet.created`.

We still send the wallet.created because you can create a wallet without initial transaction but **we should never rely on the order**.

The `WalletTransactions::CreateJob` is always dispatch even if the `WalletTransactions::CreateService` might do nothing.

### `wallet.terminated`

The webhook is set only of the wallet just got terminated. If the service is called on a terminated service, no webhook is sent.

### `wallet.updated`

This webhook is dispatched when 
* an attribute of the wallet is updated (name, expiration date or invoice_requires_successful_payment config)
* a recurring transaction rule is updated because recurring **rules are part of a wallet**, unlike wallet_transactions which are "related to a wallet" (and have their own webhooks)
* the balance is increased or decreased

**No webhook is sent when the ongoing balance is updated** because it would trigger too many events. Tracking ongoing balance is part of the "real time usage" effort.

### DIFF
--- a/app/jobs/wallet_transactions/create_job.rb
+++ b/app/jobs/wallet_transactions/create_job.rb
@@ -4,9 +4,13 @@ module WalletTransactions
   class CreateJob < ApplicationJob
     queue_as 'high_priority'
 
-    def perform(organization_id:, params:)
+    def perform(organization_id:, params:, new_wallet: false)
       organization = Organization.find(organization_id)
       WalletTransactions::CreateService.call(organization:, params:)
+
+      if new_wallet
+        SendWebhookJob.perform_later('wallet.created', Wallet.find_by(id: params[:wallet_id]))
+      end
     end
   end
 end
--- a/app/jobs/wallet_transactions/create_job.rb
+++ b/app/jobs/wallet_transactions/create_job.rb
@@ -4,13 +4,9 @@ module WalletTransactions
   class CreateJob < ApplicationJob
     queue_as 'high_priority'
 
-    def perform(organization_id:, params:, new_wallet: false)
+    def perform(organization_id:, params:)
       organization = Organization.find(organization_id)
       WalletTransactions::CreateService.call!(organization:, params:)
-
-      if new_wallet
-        SendWebhookJob.perform_later('wallet.created', Wallet.find_by(id: params[:wallet_id]))
-      end
     end
   end
 end
--- a/app/jobs/wallet_transactions/create_job.rb
+++ b/app/jobs/wallet_transactions/create_job.rb
@@ -6,7 +6,7 @@ module WalletTransactions
 
     def perform(organization_id:, params:, new_wallet: false)
       organization = Organization.find(organization_id)
-      WalletTransactions::CreateService.call(organization:, params:)
+      WalletTransactions::CreateService.call!(organization:, params:)
 
       if new_wallet
         SendWebhookJob.perform_later('wallet.created', Wallet.find_by(id: params[:wallet_id]))
--- a/spec/services/wallet_transactions/create_service_spec.rb
+++ b/spec/services/wallet_transactions/create_service_spec.rb
@@ -76,7 +76,7 @@ RSpec.describe WalletTransactions::CreateService, type: :service do
 
     it 'enqueues a SendWebhookJob for each wallet transaction' do
       expect do
-        create_service.call
+        create_service
       end.to have_enqueued_job(SendWebhookJob).thrice.with('wallet_transaction.created', WalletTransaction)
     end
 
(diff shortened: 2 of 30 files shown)


spec/factories/wallets.rb
```diff
@@ -13,6 +13,7 @@
 
     trait :terminated do
       status { 'terminated' }
+      terminated_at { DateTime.now }
```
vincent-pochet, line 16:
```suggestion
      terminated_at { Time.current }
```

app/jobs/wallet_transactions/create_job.rb
```diff
@@ -4,9 +4,13 @@ module WalletTransactions
   class CreateJob < ApplicationJob
     queue_as 'high_priority'
 
-    def perform(organization_id:, params:)
+    def perform(organization_id:, params:, new_wallet: false)
       organization = Organization.find(organization_id)
       WalletTransactions::CreateService.call(organization:, params:)
```
vincent-pochet, line 9: It looks like we are not testing the result of the service in this job:
```suggestion
      WalletTransactions::CreateService.call!(organization:, params:)
```

Any reason not to add the webhook logic inside the service? It feels a bit weird to add business logic in the job class

vincent-pochet (approved):

app/jobs/wallet_transactions/create_job.rb
julienbourdeau: Yeah, it's not obvious from the diff but when a wallet is created, there is often an initial amount, either paid or offered (granted) and [this first balance is added to the wallet in a wallet_transaction](https://github.com/getlago/lago-api/pull/3133/files#diff-12e7d1a7ede943998b91fa6f76498376d687459fc3a0ec6ded0019f35c1b6477R42-R43) created async.

If I dispatch the webhook in `Wallets::CreateService` the balance may be zero or be correct, depending on which jobs is executed first. I think it's best to send the webhook when the initial amount is correct.

That's why I tried to explain in the PR description 😝

spec/factories/wallets.rb
julienbourdeau: Thanks! The rubocop rule only catches `Time.now` but not `DateTime.now`

app/jobs/wallet_transactions/create_job.rb
vincent-pochet: Ok, but I don't see the diff with calling the `SendWebhookJob` as the last step of the `WalletTransactions::CreateService.call`. 
If it is async anyway, it should not change much thing :thinking:

julienbourdeau: > Any reason not to add the webhook logic inside the service?
I was surprised too 🤷‍♂️  I'll merge your suggestion

Ok I think I understand, you'd prefer to have the `new_wallet` carried over the WalletTransaction::CreateService so the logic live in the service, not in the job? Actually I agree 👍 👍

spec/services/wallet_transactions/create_service_spec.rb
```diff
@@ -76,7 +76,7 @@
 
     it 'enqueues a SendWebhookJob for each wallet transaction' do
       expect do
-        create_service.call
+        create_service
```
julienbourdeau, line 79: `create_service` is actually the result here. `.call` worked because the result was an openstruct but now it's a proper BaseResult and breaks 😎

https://github.com/getlago/lago-api/blob/4712e76cce61f8aee375caebc5d32da643fb3273/spec/services/wallet_transactions/create_service_spec.rb#L6

app/jobs/wallet_transactions/create_job.rb
julienbourdeau: Actually I have reconsidered and updated the description

vincent-pochet (approved):

PR #3134, opened 2025-02-04
fix: prevent calling eu tax service when not needed

lovrocolic:
This PR tries to reduce calling EU tax service in certain conditions.

The goal is NOT to call the service if customer is updated, there are existing EU taxes and tax relevant customer attributes haven't changed

### DIFF
--- a/app/services/customers/eu_auto_taxes_service.rb
+++ b/app/services/customers/eu_auto_taxes_service.rb
@@ -4,14 +4,18 @@ require 'valvat'
 
 module Customers
   class EuAutoTaxesService < BaseService
-    def initialize(customer:)
+    def initialize(customer:, new_record:, changed_attributes:)
       @customer = customer
       @organization_country_code = customer.organization.country
+      @new_record = new_record
+      @changed_attributes = changed_attributes
 
       super
     end
 
     def call
+      return nil unless should_apply_eu_taxes?
+
       customer_vies = vies_check
 
       return process_vies_tax(customer_vies) if customer_vies.present?
@@ -21,7 +25,7 @@ module Customers
 
     private
 
-    attr_reader :customer, :organization_country_code
+    attr_reader :customer, :organization_country_code, :changed_attributes, :new_record
 
     def vies_check
       vies_check = Valvat.new(customer.tax_identification_number).exists?(detail: true)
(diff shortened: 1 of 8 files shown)


app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -4,14 +4,18 @@
 
 module Customers
   class EuAutoTaxesService < BaseService
-    def initialize(customer:)
+    def initialize(customer:, new_record:, changed_attributes:)
```
julienbourdeau, line 7: I automatically assumed `changed_attributes` was a list of attributes and maybe even their before/after values.

I'd rename `changed_attributes` to something like `tax_attributes_changed` (something ending with _changed).

I like `new_record` because it matches the active model `new_record?` method.

app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -61,5 +65,13 @@ def applicable_tax_exceptions(country_code:)
     def eu_country_exceptions(country_code:)
       @eu_country_exceptions ||= LagoEuVat::Rate.new.country_rates(country_code:)[:exceptions]
     end
+
+    def should_apply_eu_taxes?
+      return false unless customer.organization.eu_tax_management
+
+      non_existing_eu_taxes = customer.taxes.where('code ILIKE ?', 'lago_eu%').none?
+
+      new_record || non_existing_eu_taxes || changed_attributes
```
julienbourdeau, line 74: I'm not too sure about this condition. If the record is new but the country doesn't have eu taxes, it would return `true`. I'd be expecting that if the country has no eu_tax, it doesn't matter if it's new or attribute has changed?

Maybe this could be 2 different checks 🤔 

```ruby
def call
  return result unless should_apply_eu_taxes?
  return result if customer.taxes.where('code ILIKE ?', 'lago_eu%').none?

  #...
end

def should_apply_eu_taxes?
  return false unless customer.organization.eu_tax_management
  new_record || changed_attributes
end
```

What do you think?

app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -4,14 +4,18 @@
...
       @customer = customer
       @organization_country_code = customer.organization.country
+      @new_record = new_record
+      @changed_attributes = changed_attributes
 
       super
     end
 
     def call
+      return nil unless should_apply_eu_taxes?
```
julienbourdeau, line 17: Ideally, we want to avoid returning nil and always return `result` so the calling method can safely call `result.success?` when any service is called.

app/services/customers/eu_auto_taxes_service.rb
lovrocolic: changed attribute name 👍

app/services/customers/eu_auto_taxes_service.rb
lovrocolic: `nil` was returned since this service originally didn't work with result as a response (instead only string was returned).
But yep, probably this is the good timing to change it so refactored it:)

PR #3140, opened 2025-02-05
fix(invoice-preview): Prevent customer updates during invoice preview

floganz:
## Context

During invoice preview it's possible to override customer attributes only for preview invoice. Customer integrations appears to be saved to DB when it shouldn't.

## Description

Change the way to build customer integrations in preview to avoid persisting.
Improve test to cover this case.

### DIFF
--- a/app/services/invoices/preview_context_service.rb
+++ b/app/services/invoices/preview_context_service.rb
@@ -53,7 +53,8 @@ module Invoices
         shipping_country: customer_params.dig(:shipping_address, :country)
       )
 
-      customer.integration_customers = Array(customer_params[:integration_customers]).map do |integration_params|
+      Array(customer_params[:integration_customers]).map do |integration_params|
+        IntegrationCustomers::BaseCustomer.customer_type(integration_params[:integration_type])
         build_customer_integration(customer, integration_params)
       end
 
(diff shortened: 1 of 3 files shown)


app/services/invoices/preview_context_service.rb
```diff
@@ -53,7 +53,8 @@ def find_or_build_customer
         shipping_country: customer_params.dig(:shipping_address, :country)
       )
 
-      customer.integration_customers = Array(customer_params[:integration_customers]).map do |integration_params|
+      Array(customer_params[:integration_customers]).map do |integration_params|
+        IntegrationCustomers::BaseCustomer.customer_type(integration_params[:integration_type])
```
floganz, line 57: Basically this call used to validate that `integration_params[:integration_type]` has a valid value since that method raises if value is not known.

app/services/invoices/preview_service.rb
```diff
@@ -204,7 +204,7 @@ def fetch_provider_taxes
     end
 
     def provider_taxation?
-      customer.integration_customers&.find { |ic| ic.type == 'IntegrationCustomers::AnrokCustomer' }
```
lovrocolic, line 207: Can you also check `Integrations::Aggregator::Taxes::BaseService` class. It seems we are reading from `customer.integration_customers` there as well

lovrocolic: Maybe we can use there `customer.integration_customers` if integration customer is persisted to leave the current flow as it is and only adapt in-memory path

floganz: Let me check.

lovrocolic (approved):

PR #3141, opened 2025-02-05
Fix(rev-share): do not display logo in credit note PDF

annvelents:
## Context

removed Logo from self-billed credit note PDF

### DIFF
--- a/app/views/templates/credit_notes/self_billed.slim
+++ b/app/views/templates/credit_notes/self_billed.slim
@@ -11,8 +11,6 @@ html
     .wrapper
       .mb-24
         h1.credit-note-title = I18n.t('credit_note.document_name')
-        - if organization.logo.present?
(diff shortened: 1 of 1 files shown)


app/views/templates/credit_notes/self_billed.slim
```diff
@@ -11,8 +11,6 @@ html
     .wrapper
       .mb-24
         h1.credit-note-title = I18n.t('credit_note.document_name')
-        - if organization.logo.present?
```
brunomiguelpinto, line 14: this should be an option. with settings? or just remove it?

annvelents: for self-billed credit note we should not display any logo, because they are issued from customer to the organization, and customer can't have logo

ancorcruz (approved):

brunomiguelpinto (approved):

PR #3142, opened 2025-02-05
fix(anrok): Remove cache for Current Usage

julienbourdeau:
## Context

When calling `GET /customers/:cust_id/current_usage?external_subscription_id=:sub_id` Lago computes the current usage

## Description

The taxes are computed by a thrid party (Anrok through Nango). We originally implemented a cache to avoir http calls to third party but it comes with drawbacks. We recently [tried to improve this](https://github.com/getlago/lago-api/pull/3115).
After investigation we believe it's best to remove the cache.

- [x] Remove Anrok cache
- [x] Introduce new `item_key` to reconcile fees when they are not persisted
- [x] Add new `apply_taxes` params to the API
  - If you get an error due to third party, you can call again without taxes
  - In case you only need pre-tax amounts (like for b2b), passing `apply_taxes: false` will be more performant
  - https://github.com/getlago/lago-openapi/pull/332

### DIFF
--- a/app/services/integrations/aggregator/taxes/invoices/payload.rb
+++ b/app/services/integrations/aggregator/taxes/invoices/payload.rb
@@ -47,6 +47,7 @@ module Integrations
             mapped_item ||= empty_struct
 
             {
+              'item_key' => fee.item_key,
               'item_id' => fee.id || fee.item_id,
               'item_code' => mapped_item.external_id,
               'amount_cents' => fee.sub_total_excluding_taxes_amount_cents&.to_i
--- a/app/controllers/api/v1/customers/usage_controller.rb
+++ b/app/controllers/api/v1/customers/usage_controller.rb
@@ -5,11 +5,13 @@ module Api
     module Customers
       class UsageController < Api::BaseController
         def current
+          apply_taxes = ActiveModel::Type::Boolean.new.cast(params.fetch(:apply_taxes, true))
           result = ::Invoices::CustomerUsageService
             .with_external_ids(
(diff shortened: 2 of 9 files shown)


app/services/integrations/aggregator/taxes/invoices/payload.rb
```diff
@@ -47,6 +47,7 @@ def fee_item(fee)
             mapped_item ||= empty_struct
 
             {
+              'item_key' => fee.item_key,
```
nudded, line 50: I think this does not need to be a new key right? anrok will use item_id for uniqueness and fee.item_key is always unique?

lovrocolic (approved):

app/services/integrations/aggregator/taxes/invoices/payload.rb
julienbourdeau: I'd rather not touch `item_id` for now because it's used in the frontend. `item_id` is usually unique but not necessarily. If you have 2 charge fees relying on the same billable metric, the `item_id` will be the same.
This is what we wanted to fix here: https://github.com/getlago/lago-api/pull/2842

But sometimes, the fee isn't persisted (here or when createing a one-off invoice from the frontend. The frontend expects the item_id to be as defined in `Fee.item_id`.

So I'm introducing this new key. Ideally, both should be merged at some point to have something the frontend can use (not like `object_id`).

nudded: okay, that all sounds really reasonable to me :)

nudded (approved):

PR #3151, opened 2025-02-07
feat(data): Add revenue streams resolver

rsempe:
**Context**

A lot of users ain’t using our analytics feature because it’s not detailed enough. Thus, we are revamping the Analytics dashboard.

**Description**

This pull request introduces a new feature to query revenue streams of an organization using GraphQL.
It requests the `lago-data` api service and returns revenue streams for an organization.


app/services/analytics/base_service.rb
```diff
@@ -1,15 +1,36 @@
...
       super()
     end
 
     private
 
     attr_reader :organization, :filters, :records
+
+    def http_client
+      LagoHttpClient::Client.new(endpoint_url)
+    end
```
julienbourdeau, line 20: Should we keep the client in memory to share it between request. So not in the service instance but somewhere.
Obviously, this kind of optimization can be done later.

app/services/analytics/base_service.rb
```diff
@@ -1,15 +1,36 @@
 # frozen_string_literal: true
 
+require "lago_http_client"
```
julienbourdeau, line 3: How do you feel about making a new BaseService for analutics services calling the new data API? Could be in the same module/folder or even a separate one if you prefer.
I think it would help to differentiate the services that calls the external API and those using PG.


```ruby
module Analytics
  class RevenueStreamsService < BaseDataApiService   # <--- makes http calls
    # ...
  end
end

module Analytics
  class InvoicedUsagesService < BaseService   # <--- can be renamed too 
    # ...
  end
end
```

julienbourdeau (approved):

app/services/analytics/base_service.rb
rsempe: yes - that makes sense - will do it 👍

PR #3152, opened 2025-02-07
fix(invoice): Add GraphQL flag to know editable status

vincent-pochet:
## Context

An issue was reported with creation of adjusted fees following the recent changes related to the removal of zero amount fees from draft invoices.

In case of `subscription_starting` invoice_subscription, adjustment creation is ignored because the invoice refresh is not creating charge fees. This behavior is expected as no charge fees are created when starting a subscription.
Updating the back-end to allow such edition would require a lot of changes with high risk of unexpected side effect.

## Description

This PR adds a new `AcceptNewChargeFees` boolean flag to the `InvoiceSubscription` GraphQL object to let the front-end know if the add new fee button should be displayed or not

### DIFF
--- a/app/graphql/types/invoice_subscriptions/object.rb
+++ b/app/graphql/types/invoice_subscriptions/object.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 module Types
-  module InvoiceSubscription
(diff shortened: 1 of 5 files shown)


julienbourdeau (approved):
great solution 🙌

PR #3153, opened 2025-02-07
feat(regroup_fee): Sum fees to get invoice totals instead of recomputing taxes

julienbourdeau:
## Context

Using Fees with the following attributes, you will get an invoice at the end of the period that regroups all PAID fees.
* pay_in_advance: true
* invoiceable: false
* regroup_paid_fees: 'invoice'

## Description

The taxes on these fees are computed for each fee. The final invoice should show the amount of taxes actually paid, not the amount of taxes that would have been paid if the fees were paid at once in a single invoice invoice.

**Example**

Fee 1:
* amount_cents: 12
* taxes: 20%
* taxes_amount_cents: 2.4 rounded to 2

Fee 2:
* amount_cents: 12
* taxes: 20%
* taxes_amount_cents: 2.4 rounded to 2

Final invoice:
* 2 fees 
* fees_amount_cents: 24 (12 * 2)
* taxes_amount_cents: **4 if you sum the existing fees, but 20% of 24 is 4.8, rounded to 5**
* total_amount_cents: 28 with this PR (actually paid taxes) but 29 before. There is 1 cent invoiced but not collected.

### DIFF
--- a/app/services/fees/create_pay_in_advance_service.rb
+++ b/app/services/fees/create_pay_in_advance_service.rb
@@ -2,6 +2,8 @@
 
 module Fees
   class CreatePayInAdvanceService < BaseService
+    Result = BaseResult[:fees]
+
     def initialize(charge:, event:, billing_at: nil, estimate: false)
       @charge = charge
       @event = Events::CommonFactory.new_instance(source: event)
@@ -58,22 +60,22 @@ module Fees
 
         cache_aggregation_result(aggregation_result:, charge_filter:)
 
-        result = apply_charge_model(aggregation_result:, properties:)
-        unit_amount_cents = result.unit_amount * subscription.plan.amount.currency.subunit_to_unit
+        charge_model_result = apply_charge_model(aggregation_result:, properties:)
+        unit_amount_cents = charge_model_result.unit_amount * subscription.plan.amount.currency.subunit_to_unit
 
         fee = Fee.new(
           subscription:,
           charge:,
           organization_id: organization.id,
-          amount_cents: result.amount,
-          precise_amount_cents: result.precise_amount,
+          amount_cents: charge_model_result.amount,
+          precise_amount_cents: charge_model_result.precise_amount,
           amount_currency: subscription.plan.amount_currency,
           fee_type: :charge,
           invoiceable: charge,
-          units: result.units,
-          total_aggregated_units: result.units,
+          units: charge_model_result.units,
+          total_aggregated_units: charge_model_result.units,
           properties: boundaries,
-          events_count: result.count,
+          events_count: charge_model_result.count,
           charge_filter_id: charge_filter&.id,
           pay_in_advance_event_id: event.id,
           pay_in_advance_event_transaction_id: event.transaction_id,
@@ -82,9 +84,9 @@ module Fees
           taxes_amount_cents: 0,
           taxes_precise_amount_cents: 0.to_d,
           unit_amount_cents:,
-          precise_unit_amount: result.unit_amount,
+          precise_unit_amount: charge_model_result.unit_amount,
           grouped_by: format_grouped_by,
-          amount_details: result.amount_details || {}
+          amount_details: charge_model_result.amount_details || {}
         )
 
         unless customer_provider_taxation?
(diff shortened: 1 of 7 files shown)


app/services/fees/create_pay_in_advance_service.rb
```diff
@@ -2,6 +2,8 @@
 
 module Fees
   class CreatePayInAdvanceService < BaseService
+    Result = BaseResult[:fees]
```
vincent-pochet, line 5: :heart:

app/services/invoices/advance_charges_service.rb
```diff
@@ -79,7 +79,16 @@ def create_group_invoice
           raise ActiveRecord::Rollback
         end
 
-        Invoices::ComputeAmountsFromFees.call(invoice:)
+        # NOTE: We don't want to use Invoices::ComputeAmountsFromFees here
+        #       because it would recompute taxes from pre-tax values. All Fees are already paid
+        #       this invoice should show how much taxes were paid in total.
+        # NOTE: progressing billing, coupons and credit notes are not supported here
+        invoice.fees_amount_cents = invoice.fees.sum(&:amount_cents)
```
vincent-pochet, line 86: Just one though, should we take coupons amount into account as well?

spec/scenarios/invoices/advance_charges_spec.rb
```diff
@@ -46,7 +46,8 @@ def send_card_event!(item_id = SecureRandom.uuid)
         travel_to(DateTime.new(2024, 6, 10 + i, 10)) do
           send_card_event! "card_#{i}"
           expect(subscription.fees.charge.where(invoice_id: nil).count).to eq(i)
-          expect(subscription.fees.charge.order(created_at: :desc).first.amount_cents).to eq((21 - i) * 100)
+          # pp ((bm_amount * (30 - 10 - (i - 1)) / 30) * 100).round
```
vincent-pochet, line 49: You can remove this one I guess :D

vincent-pochet (approved):

spec/scenarios/invoices/advance_charges_spec.rb
julienbourdeau: Oops! 😄

app/services/invoices/advance_charges_service.rb
julienbourdeau: I'll make sure Matt and the customer are aware of it. If we want to take care of it, I'd do it in a separate PR because I'll need some more spec scenario to understand 😬

app/services/invoices/aggregate_amounts_and_taxes_from_fees.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Invoices
+  class AggregateAmountsAndTaxesFromFees < BaseService
+    Result = BaseResult
+
+    def initialize(invoice:)
+      @invoice = invoice
+
+      raise ArgumentError.new("invoice type must be `advance_charges`") unless invoice.advance_charges?
```
julienbourdeau, line 10: Errors returned in the results are error that can happen. Here I want to ensure that if another dev wants to use this service with another type of invoice, they are warned the service wasn't designed for this and they should make changes.

vincent-pochet: I think it is still a precondition, I would stick to returning a failed result.
It should then be raised/handled by the caller.  We already have many similar situations

The risk with using "ArgumentError" or other built-in ruby error is that for some reason, someone could catch the error type in higher services with some custom handling logic.

app/services/invoices/aggregate_amounts_and_taxes_from_fees.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Invoices
+  class AggregateAmountsAndTaxesFromFees < BaseService
+    Result = BaseResult
```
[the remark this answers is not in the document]
julienbourdeau: Agree!

app/services/invoices/aggregate_amounts_and_taxes_from_fees.rb
julienbourdeau: Yes, I wasn't too sure about this myself

PR #3154, opened 2025-02-07
bug(GracePeriod) - Make grace period update async for increased reliability

nudded:
## Description

The grace period update service was not built to deal with many (>100k) draft invoices. This means that some issuing_dates could be set incorrectly if the request timed out. This PR moves the update to an asynchronous job.

For idempotency reasons it's important to store the applied_grace_period on the invoice, this way we can be certain that we're not running the job twice.

### DIFF
--- a/app/services/invoices/update_grace_period_from_organization_service.rb
+++ b/app/services/invoices/update_grace_period_from_organization_service.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Invoices
+  class UpdateGracePeriodFromOrganizationService < BaseService
+    Result = BaseResult[:invoice]
+
+    def initialize(invoice:, old_grace_period:)
+      @invoice = invoice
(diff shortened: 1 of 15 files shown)


vincent-pochet (approved):

app/services/invoices/update_grace_period_from_organization_service.rb
```diff
@@ -0,0 +1,37 @@
...
+    def initialize(invoice:, old_grace_period:)
+      @invoice = invoice
+      @old_grace_period = old_grace_period
+      super
+    end
+
+    def call
+      result.invoice = invoice
+      # only update issuing_date when there is no override on customer
+      return result if invoice.customer.invoice_grace_period.present?
```
julienbourdeau, line 16: Excellent catch!

julienbourdeau (approved):

PR #3156, opened 2025-02-07
fix: Remove tax cache from invoice preview

lovrocolic:
Some clients want to override address in preview feature. Address is one of the parameters that determine tax and we decided to always fetch fresh taxes in preview feature

### DIFF
--- a/app/services/invoices/preview_service.rb
+++ b/app/services/invoices/preview_service.rb
@@ -172,35 +172,13 @@ module Invoices
       apply_zero_tax
     end
 
-    def provider_taxes_cache_key
-      [
-        "preview-taxes",
-        customer.id,
-        subscription.plan.updated_at.iso8601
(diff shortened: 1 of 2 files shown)


julienbourdeau (approved):

app/services/invoices/preview_service.rb
```diff
@@ -172,35 +172,13 @@ def apply_provider_taxes
...
-        subscription.plan.updated_at.iso8601
-      ].join("/")
-    end
-
     def apply_zero_tax
       invoice.taxes_amount_cents = 0
       invoice.taxes_rate = 0
     end
 
     def fetch_provider_taxes
```
floganz, line 180: I guess we can even remove this method, there's no much difference if we call it directly or via the method.

floganz (approved):

PR #3173, opened 2025-02-12
feat(invoice-preview): add base multiple subscription setup

lovrocolic:
## Context

With invoice preview feature, customers will have possibility to check how next invoice will look like

## Description

This is the first PR in second part that covers persisted subscriptions. Here is the setup for multiple subscription and there is added validation that ensures that multiple subscriptions has the same currency.

Since part 1 is already released, second part PRs will be created from part 2 feature branch: `ftr-invoice-preview-two`

### DIFF
--- a/app/services/invoices/preview_context_service.rb
+++ b/app/services/invoices/preview_context_service.rb
@@ -2,6 +2,8 @@
 
 module Invoices
   class PreviewContextService < BaseService
+    Result = BaseResult[:customer, :subscriptions, :applied_coupons]
+
     def initialize(organization:, params:)
       @organization = organization
       @params = params.presence || {}
@@ -10,7 +12,7 @@ module Invoices
 
     def call
       result.customer = find_or_build_customer
-      result.subscription = build_subscription
+      result.subscriptions = find_or_build_subscriptions
       result.applied_coupons = find_or_build_applied_coupons
       result
     rescue ActiveRecord::RecordNotFound => exception
@@ -86,6 +88,10 @@ module Invoices
         created_at: params[:subscription_at].presence || Time.current,
         updated_at: Time.current
       )
+    end
+
+    def find_or_build_subscriptions
+      [build_subscription]
     end
 
     def plan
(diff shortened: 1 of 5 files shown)


floganz (approved):
👍

app/services/invoices/preview_context_service.rb
```diff
@@ -88,6 +90,10 @@ def build_subscription
       )
     end
 
+    def find_or_build_subscriptions
```
julienbourdeau, line 93: I'm not sure with the name, I don't see where it "find", it seems to always "build".
Maybe `build_subscriptions` (plural) is enough but you mihgt have other things in mind coming.

app/services/invoices/preview_service.rb
```diff
@@ -13,13 +15,14 @@ def initialize(customer:, subscription:, applied_coupons: [])
...
       return result.not_found_failure!(resource: 'customer') unless customer
-      return result.not_found_failure!(resource: 'subscription') unless subscription
+      return result.not_found_failure!(resource: 'subscription') if subscriptions.empty?
+      return result unless currencies_aligned?
 
       @invoice = Invoice.new(
         organization: customer.organization,
         customer:,
         invoice_type: :subscription,
-        currency: subscription.plan&.amount_currency,
```
julienbourdeau, line 22: I'm always scared that this `&` was there for a good reason I don't know 🙈 maybe it wasn't tho!

julienbourdeau (approved):

app/services/invoices/preview_context_service.rb
lovrocolic: Yep, sorry, maybe this PR is hard to track since it is just setup for multiple subscriptions and not a final implementation. That's why it goes in feature branch.
Currently we just added support for array but we don't really support multiple subscriptions that are passed as parameters.
In the next PRs we will extend this method and add logic where subscriptions can be either loaded from the DB (based on the external_ids from the params) or we will build subscription object as before

app/services/invoices/preview_service.rb
lovrocolic: We have validation for it in a plan so it should be always there :)

PR #3174, opened 2025-02-12
fix(fee-estimation): fix rounding when small amounts provided

floganz:
## Context

Percentage charges have rounding issues with smaller units and decimal rates.
E.g this code
```ruby
amount = units * rate.fdiv(100)
```
produces rounding issues.
```ruby
rate = BigDecimal("2.9") # service always converts it to BigDecimal
units = 4

units * rate.fdiv(100) # => 0.11599999999999999
units * rate.fdiv(100).class # => Float
units * rate / 100 # => 0.116
(units * rate / 100).class # => BigDecimal
```

The issue that `.fdiv(100)` returns result as Float, in our case downgrading from BigDecimal to Float.
In changes cases we always have BigDecimal, so it's safe to use `/` instead of `fdiv`.
Btw, another way to address this issue by using `.fdiv(BigDecimal("100"))` which for me looks horrible.

## Description

Adjusted percentage calculation to return expected results with smaller amount.s
Adjusted tests to cover changes.

### DIFF
--- a/app/services/charges/amount_details/range_graduated_percentage_service.rb
+++ b/app/services/charges/amount_details/range_graduated_percentage_service.rb
@@ -42,7 +42,7 @@ module Charges
       end
 
       def per_unit_total_amount
-        @per_unit_total_amount ||= (units * rate).fdiv(100)
+        @per_unit_total_amount ||= units * rate / 100
       end
 
       def total_with_flat_amount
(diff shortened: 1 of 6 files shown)


ivannovosad (approved):

julienbourdeau (approved):
TIL!

PR #3179, opened 2025-02-13
chore(rubocop): Disable or fix last errors

julienbourdeau:
As I'm fixing the quotes for string literals, there are other errors showing up so let's fix those. I'm going to merge this one first before updating the quote PRs.

Many of them are about old migrations that I don't think we should change. I simply disabled the rule for them.

`Cops/ServiceCallCop` was renamed `Lago/ServiceCall` and I excluded the services failing. It's not so easy to fix in this kind of cosmetic PR but it should be done one day.

> 3234 files inspected, no offenses detected

CleanShot 2025-02-13 at 13 58 18@2x

### DIFF
--- a/app/services/dunning_campaigns/update_service.rb
+++ b/app/services/dunning_campaigns/update_service.rb
@@ -37,8 +37,6 @@ module DunningCampaigns
     end
 
     def handle_thresholds
-      thresholds_updated = false
-
       input_threshold_ids = params[:thresholds].map { |t| t[:id] }.compact
 
       # Delete thresholds not included in the payload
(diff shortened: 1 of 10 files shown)


app/services/dunning_campaigns/update_service.rb
```diff
@@ -37,8 +37,6 @@ def permitted_attributes
     end
 
     def handle_thresholds
-      thresholds_updated = false
```
julienbourdeau, line 40: Reassigned line 48

vincent-pochet (approved):

PR #3206, opened 2025-02-18
feat(invoice-preview): add charges support

lovrocolic:
## Context

Preview feature allows to see first invoice preview. It works for both non-existing subscriptions but also for active subscriptions.

## Description

This PR adds support for charges for subscriptions that are persisted in the DB.

### DIFF
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -35,7 +35,7 @@ module Fees
           precise_amount_cents: result.fees.sum(&:precise_amount_cents)
         )
       end
-      return result unless result.success?
+      return result if !result.success? || context == :invoice_preview
 
       ActiveRecord::Base.transaction do
         result.fees.reject! { |f| !should_persit_fee?(f, result.fees) }
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -35,10 +35,12 @@ module Fees
           precise_amount_cents: result.fees.sum(&:precise_amount_cents)
         )
       end
-      return result if !result.success? || context == :invoice_preview
+      return result unless result.success?
 
       ActiveRecord::Base.transaction do
         result.fees.reject! { |f| !should_persit_fee?(f, result.fees) }
+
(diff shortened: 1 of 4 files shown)


floganz (approved):

app/services/fees/charge_service.rb
```diff
@@ -35,7 +35,7 @@ def call
           precise_amount_cents: result.fees.sum(&:precise_amount_cents)
         )
       end
-      return result unless result.success?
+      return result if !result.success? || context == :invoice_preview
```
vincent-pochet, line 38: Do we want to render zero-amount fees in the preview? If we escape here, the service will return the full list of fees, including the zero amount ones

app/services/invoices/preview_service.rb
```diff
@@ -124,16 +125,45 @@ def payment_due_date
 
     def add_subscription_fees
       invoice.fees = subscriptions.map do |subscription|
-        fee = Fees::SubscriptionService.call(
+        Fees::SubscriptionService.call(
           invoice:,
           subscription:,
           boundaries: boundaries(subscription),
           context: :preview
         ).raise_if_error!.fee
```
vincent-pochet, line 133:
```suggestion
        Fees::SubscriptionService.call!(
          invoice:,
          subscription:,
          boundaries: boundaries(subscription),
          context: :preview
        ).fee
```

vincent-pochet (approved):

app/services/fees/charge_service.rb
lovrocolic: Hmm good question? I guess both options are fine. The idea was to escape here so that we still create true up fee if needed, but indeed, zero fees will be included.
Do you suggest to exclude all zero amount fees from the array before returning?

vincent-pochet: Maybe it's better to be consistent with the latest changes. What do you think?

lovrocolic: okay, moved `return` after `result.fees.reject! { |f| !should_persit_fee?(f, result.fees) }` line 👍

PR #3209, opened 2025-02-18
feat(gql): Introduce ObfuscatedStringType

julienbourdeau:
## Context

For api keys or token, the graphql exposes a string with dots and only shows the last 3 characters.

## Description

This PR introduces a custom `ObfuscatedStringType` to our graphql schema so you don't have to redeclare it every time you need to hide a sensitive information.

CleanShot 2025-02-19 at 14 38 10@2x


vincent-pochet (approved):
Looks great!

Just make sure to warn the front-end team as it will break some stuff :smile:

PR #3210, opened 2025-02-18
feat(dev): Add console helpers for local development

julienbourdeau:
## Description

This PR adds some helpers that I like to use locally. I use Hooli and Gavin's user account when developing to 2 methods let's you retrieve the org and the user easily.

The third one let's you delete all the webhooks for the hooli work so when testing, you can get rid of all previous webhooks for some clarity.


annvelents (approved):
🚀🚀🚀

PR #3211, opened 2025-02-18
Misc: migrate invoice_errors to error_details

annvelents:
## Description

Migrated InvoiceError into ErrorDetails with code: 'invoice_generation_error', while keeping the whole logic for invoice_error:
- invoice can have only one  `invoice_generation_error`
- error_detail keeps all data for invocie, subscriptions, error, backtrace


vincent-pochet (approved):
Looks good to me

PR #3215, opened 2025-02-19
feat(api): Introduce new error

julienbourdeau:
## Context

As I was working on [adding crypto](https://github.com/getlago/lago-api/pull/3214), I realized if the checkout URL failed, we just return an 200 response where the checkout url is incorrect.

## Description

I sometimes find that the API swallow to many errors. You receive a webhook with all the details but the API will simply return a cryptic error code. 

Here I'm thinking about adding a new kind of failure for Payment Provider. 

A contributor recently introduced `thirdparty_errors` because there was no easy way to return error from another service. I don't like the name too much, we usually use "integrations" or "provider" rather than third parties but more importantly I think it's too limited to return only a message.

Please share anything you have in mind. I'm not too sure we should merge the PR in this state but I'd like the API to return more errors from payment provider.

**Error before**

CleanShot 2025-02-18 at 11 46 01@2x

**Error after**

CleanShot 2025-02-18 at 12 39 48@2x

### DIFF
--- a/app/services/base_service.rb
+++ b/app/services/base_service.rb
@@ -93,6 +93,20 @@ class BaseService
     end
   end
 
+  class PaymentProviderFailure < FailedResult
+    attr_reader :status, :details, :payment_provider, :payment_provider_code
+
+    def initialize(result, message:, status:, details:,
+      payment_provider:, payment_provider_code:)
+      @status = status
+      @details = details
+      @payment_provider = payment_provider
+      @payment_provider_code = payment_provider_code
+
+      super(result, "#{@payment_provider} (#{@payment_provider_code}): #{message}")
+    end
+  end
+
   class ThirdPartyFailure < FailedResult
     attr_reader :third_party, :error_message
 
--- a/app/services/payment_provider_customers/stripe_service.rb
+++ b/app/services/payment_provider_customers/stripe_service.rb
@@ -84,7 +84,7 @@ module PaymentProviderCustomers
       result
     rescue ::Stripe::InvalidRequestError, ::Stripe::PermissionError => e
       deliver_error_webhook(e)
-      result
+      fail! e
     rescue ::Stripe::AuthenticationError => e
       deliver_error_webhook(e)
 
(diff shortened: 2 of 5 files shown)


spec/services/payment_provider_customers/stripe_service_spec.rb
```diff
@@ -500,6 +496,25 @@
           ).on_queue(webhook_queue)
       end
     end
+
+    context "when stripe raises an invalid request error" do
+      before do
+        stub_request(:post, %r{https://api.stripe.com/v1/checkout/sessions}).to_return(status: 400, body: {
```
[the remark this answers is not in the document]
julienbourdeau: vvery good point!

app/services/base_service.rb
```diff
@@ -93,6 +93,20 @@ def initialize(result, message:)
     end
   end
 
+  class PaymentProviderFailure < FailedResult
```
vincent-pochet, line 96: Since we are relying on many integration, I would suggest to rename the `ThirdPartyFailure` into `IntegrationFailure` and allow it to expose the initial error from the provider.

In many places in the project we have different ways of dealing with provider error, I'm not sure that adding a new one will simplify it :smile:

app/services/payment_provider_customers/stripe_service.rb
```diff
@@ -203,6 +203,21 @@ def deliver_error_webhook(stripe_error)
       )
     end
 
+    def fail!(stripe_error)
+      result.fail_with_error!(PaymentProviderFailure.new(
```
vincent-pochet, line 207: To keep the "easy" way, you could rely on `result.provider_error!(...)`. That way, no need to build yourself the instance of the error at this stage

app/services/payment_provider_customers/stripe_service.rb
```diff
@@ -84,7 +84,7 @@ def generate_checkout_url(send_webhook: true)
       result
     rescue ::Stripe::InvalidRequestError, ::Stripe::PermissionError => e
       deliver_error_webhook(e)
-      result
+      fail! e
```
vincent-pochet, line 87: Oups... Indeed, this one in this scope should not be swallowed... :+1:

vincent-pochet (commented):
Nice to put this topic on the table. I would just suggest to build a dedicated way to handle errors for all providers rather than adding a new logic just for payment providers.

My dream is to have a single way to handle failures for integrations:
- Webhooks (with have 10 differents error webhooks, with different signatures...)
- Error storing for display purpose in the front-end or the API, like it has been done for tax providers
- API responses for sync actions
- ...

Happy to discuss about it :)

app/services/payment_provider_customers/stripe_service.rb
julienbourdeau: I'll fix this separately so we can get it fixed today

julienbourdeau:
Closed in favor of #3247 and #3238

PR #3217, opened 2025-02-19
feat(multi-entities): Add billing entity model

ancorcruz:
## Roadmap Task

## Context

## Description

Add billing entity model as a starting point to kick off this feature development.


annvelents (approved):
🚀

PR #3218, opened 2025-02-19
feat(mailers): Add from email premium integration

jdenquin:
- Support organization email defined as `from_email` for all customers emails
- Add new premium integration `from_email`

### DIFF
--- a/app/mailers/credit_note_mailer.rb
+++ b/app/mailers/credit_note_mailer.rb
@@ -16,7 +16,6 @@ class CreditNoteMailer < ApplicationMailer
       attachments["credit_note-#{@credit_note.number}.pdf"] = file.read
     end
 
-    byebug
     from_email = if @organization.from_email_enabled?
       @organization.email
     else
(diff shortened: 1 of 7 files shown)


app/mailers/credit_note_mailer.rb
```diff
@@ -16,10 +16,16 @@ def created
       attachments["credit_note-#{@credit_note.number}.pdf"] = file.read
     end
 
+    from_email = if @organization.from_email_enabled?
+      @organization.email
+    else
+      ENV["LAGO_FROM_EMAIL"]
+    end
```
julienbourdeau, line 23: I'd move this to the organization model to avoid duplication

```ruby
def from_email_address
    if from_email_enabled?
      email
    else
      ENV["LAGO_FROM_EMAIL"]
    end
end
```

julienbourdeau (approved):

app/mailers/credit_note_mailer.rb
jdenquin: done!

PR #3224, opened 2025-02-20
Feat(multi-entities): add references to and from billing_entity

annvelents:
## Context

After we created the billing_entity model itself, we want to add references to it from different model that should be separated by billing entities

## Description

Added references to billing_entities
Added creation of default billing_entity when creating an organization
Updated tests

### DIFF
--- a/spec/factories/billing_entities.rb
+++ b/spec/factories/billing_entities.rb
@@ -9,6 +9,9 @@ FactoryBot.define do
     email { Faker::Internet.email }
     email_settings { ["invoice.finalized", "credit_note.created"] }
 
+    # TODO: remove this magic after is_default field is deleted
+    # we need firstly to set is_default on the current billing_entity,
+    # so when creating organization we won't need to create a default billing entity
     after :build do |billing_entity, values|
       billing_entity.is_default = true if values.organization&.billing_entities&.where(is_default: true).blank?
       billing_entity.organization = build(:organization, billing_entities: [billing_entity]) if values.organization.blank?
--- a/spec/factories/billing_entities.rb
+++ b/spec/factories/billing_entities.rb
@@ -2,13 +2,17 @@
 
 FactoryBot.define do
   factory :billing_entity do
-    organization
     name { Faker::Company.name }
     code { "entity_#{SecureRandom.uuid}" }
     default_currency { "USD" }
 
     email { Faker::Internet.email }
     email_settings { ["invoice.finalized", "credit_note.created"] }
+
+    after :build do |billing_entity, values|
+      billing_entity.is_default = true if values.organization&.billing_entities&.where(is_default: true).blank?
+      billing_entity.organization = build(:organization, billing_entities: [billing_entity]) if values.organization.blank?
+    end
 
     trait :default do
       is_default { true }
--- a/spec/factories/customers.rb
+++ b/spec/factories/customers.rb
@@ -22,6 +22,10 @@ FactoryBot.define do
     legal_number { Faker::Company.duns_number }
     currency { "EUR" }
 
+    after :build do |customer, values|
+      customer.billing_entity = values.organization.default_billing_entity if customer.billing_entity.nil?
+    end
+
     trait :with_shipping_address do
       shipping_address_line1 { Faker::Address.street_address }
       shipping_address_line2 { Faker::Address.secondary_address }
--- a/spec/factories/customers.rb
+++ b/spec/factories/customers.rb
@@ -23,7 +23,7 @@ FactoryBot.define do
     currency { "EUR" }
 
     after :build do |customer, values|
-      customer.billing_entity = values.organization.default_billing_entity if customer.billing_entity.nil?
+      customer.billing_entity = values.organization.default_billing_entity unless customer.billing_entity
     end
 
     trait :with_shipping_address do
--- a/spec/factories/invoices.rb
+++ b/spec/factories/invoices.rb
@@ -12,6 +12,10 @@ FactoryBot.define do
 
     organization_sequential_id { rand(1_000_000) }
 
+    after :build do |invoice, values|
+      invoice.billing_entity = values.organization.default_billing_entity if invoice.billing_entity.nil?
+    end
+
     trait :draft do
       status { :draft }
     end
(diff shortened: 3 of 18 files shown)


spec/factories/billing_entities.rb
```diff
@@ -2,14 +2,18 @@
...
-    organization
     name { Faker::Company.name }
     code { "entity_#{SecureRandom.uuid}" }
     default_currency { "USD" }
 
     email { Faker::Internet.email }
     email_settings { ["invoice.finalized", "credit_note.created"] }
 
+    after :build do |billing_entity, values|
+      billing_entity.is_default = true if values.organization&.billing_entities&.where(is_default: true).blank?
```
ancorcruz, line 13: Good point.

However for a test case I would not rely on setting `is_default` automatically and use the trait to specifically do so. Otherwise we could run into unexpected/hidden behaviours because the `is_default` is set to true and we don not expect it... also this will ignore the fact of archived entities, etc.

spec/factories/customers.rb
```diff
@@ -22,6 +22,10 @@
     legal_number { Faker::Company.duns_number }
     currency { "EUR" }
 
+    after :build do |customer, values|
+      customer.billing_entity = values.organization.default_billing_entity if customer.billing_entity.nil?
```
ancorcruz, line 26:
```suggestion
      customer.billing_entity = values.organization.default_billing_entity unless customer.billing_entity
```

spec/factories/invoices.rb
```diff
@@ -12,6 +12,10 @@
 
     organization_sequential_id { rand(1_000_000) }
 
+    after :build do |invoice, values|
+      invoice.billing_entity = values.organization.default_billing_entity if invoice.billing_entity.nil?
```
ancorcruz, line 16: let's write idiomatic code 🙌 

```suggestion
      invoice.billing_entity = values.organization.default_billing_entity unless invoice.billing_entity
```

spec/models/billing_entity_spec.rb
```diff
@@ -9,13 +9,19 @@
 
   it { is_expected.to belong_to(:organization) }
 
+  it { is_expected.to have_many(:customers) }
+  it { is_expected.to have_many(:invoices) }
+  it { is_expected.to have_many(:invoice_custom_section_selections) }
+  it { is_expected.to have_many(:selected_invoice_custom_sections) }
```
ancorcruz, line 15:
```suggestion
  it { is_expected.to have_many(:selected_invoice_custom_sections).through(:invoice_custom_section_selections) }
```

spec/models/billing_entity_spec.rb
```diff
@@ -9,13 +9,19 @@
...
+  it { is_expected.to have_many(:fees) }
+
```
ancorcruz, line 17: it would be great if also add the other relationships here...

```  
  has_many :subscriptions, through: :customers
  has_many :wallets, through: :customers
  has_many :wallet_transactions, through: :wallets
  has_many :credit_notes, through: :invoices

  has_one :applied_dunning_campaign, class_name: "DunningCampaign"
```

spec/models/billing_entity_spec.rb
```diff
@@ -9,13 +9,19 @@
...
   describe "is_default validation" do
     let(:organization) { create :organization }
 
     it "validates uniqueness of organization_id for is_default excluding deleted and archived records" do
-      deleted_record = create(:billing_entity, :default, :deleted, organization:)
+      # by default an organization is built with a default billing entity
+      expect(organization.default_billing_entity.discard!).to be true
```
ancorcruz, line 23: what if the organization does not have a default billing entity? This is feasible as the billing entity could be discarded or archived...

ancorcruz (changes requested):
Good stuff, left a few comments.

spec/factories/billing_entities.rb
```diff
@@ -2,14 +2,18 @@
...
+      billing_entity.organization = build(:organization, billing_entities: [billing_entity]) if values.organization.blank?
```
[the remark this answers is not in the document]
annvelents: unfortunately I had to do it, because having an organization defined in the "build" process of the model means that when the model is being build, we'll build an organization. 
When building an organization we'll check if the organization have a default billing_entity. the billing_entity we were originally creating, is not yet default, so we'll also build a default billing_entity for this organization, and then we'll get back to the original billing_entity, which now belongs to an organization that already have one default billing_entity. So instead of creating one billing_entity that belong to an organization, and logically should be default billing entity for this organization. we're ending up with an additional billing_entity and the created one is not default.

while with current code, if we create a billing entity by itself, we'll create an organization and mark this billing_entity as default
if organization is passed, we'll assign this organization to the billing entity
if the passed organization does not have a default billing_entity, the newly created billing_entity will be set as default

spec/factories/billing_entities.rb
annvelents: my idea was that since most of the cases in production and test-cases as well per each organization we'll have only one billing_entity and for the correct flow it should be created, it will make test set-up easier to always create an organization with default billing entity. so when we're creating a billing entity, if it's the only entity of an organization, it should be default....

However if we're creating explicitly exactly billing_entity, it makes sense that the person who's writing the test is aware of billing_entities architecture...  and might choose is_default/no themselves...

yep, I'll remove the magic from here

spec/models/billing_entity_spec.rb
annvelents: it's exactly the case, that in most of the cases, except when it's exactly a test case for billing_entity, we will and should have a default billing entity...
Shall we discuss it offline?

spec/factories/billing_entities.rb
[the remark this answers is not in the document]
annvelents: discussed it offline, left a comment that it's needed temporary and will need to be deleted after the final implementation complete

spec/factories/billing_entities.rb
annvelents: discussed it offline, left a comment that it's needed temporary and will need to be deleted after the final implementation complete

spec/models/billing_entity_spec.rb
annvelents: discussed it offline, left a comment that it's needed temporary and will need to be deleted after the final implementation complete

ancorcruz (approved):
🚀

PR #3228, opened 2025-02-20
feat(stripe): Stop listening to `charge.succeeded` webhook

julienbourdeau:
## Description

Lago listens to many webhooks from Stripe, including `charge.succeeded` and `payment_intent.succeeded`.

Charges are the Stripe "plumbing way" of handling payment. We don't use the Charge API, we always use the PaymentIntent API (the recommended way).

Every time a payment is succeeded, we receive BOTH `charge.succeeded` and `payment_intent.succeeded`, we  end up with 2 jobs trying to update the same payment status: [which create some race conditions](https://github.com/getlago/lago-api/pull/3223).

Note that Lago instance already listing to `charge.succeeded` will continue to receive the webhooks because I didn't update the stripe configuration but Lago won't do anything with it. We can update the stripe setup to stop receiving them by calling `PaymentProviders::Stripe::RegisterWebhookService.call`.

CleanShot 2025-02-20 at 12 31 26@2x

https://docs.stripe.com/api/events


vincent-pochet (approved):
:raised_hands:

rsempe (approved):
👍 👍

PR #3229, opened 2025-02-20
feat(payment-receipts): Add migration and models

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds/modifies models for payment receipts.
It also adds a migration containing a trigger for generating per-customer sequences that are stored in payment receipt `number` column.


app/models/payment_receipt.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class PaymentReceipt < ApplicationRecord
+  belongs_to :payment
+
+  scope :for_organization, lambda { |organization|
```
vincent-pochet, line 6: What about adding the `organization_id` to this model?
It would removes the need for this complex query.

db/migrate/20250220180112_create_payment_receipts.rb
```diff
@@ -0,0 +1,72 @@
...
+        $$ LANGUAGE plpgsql;
+      SQL
+
+      execute <<-SQL
+        CREATE TRIGGER before_payment_receipt_insert
+        BEFORE INSERT ON payment_receipts
+        FOR EACH ROW
+        EXECUTE FUNCTION set_payment_receipt_number();
+      SQL
+    end
```
vincent-pochet, line 60: Can you move the function and trigger creation in a dedicated migration?
I think we should keep migrations changes as atomic as possible

vincent-pochet (commented):
Two comments regarding the DB logic itself, otherwise, it looks good! :+1:

db/migrate/20250220180112_create_payment_receipts.rb
ivannovosad: I split the migrations into 3 separate ones

vincent-pochet (approved):

PR #3236, opened 2025-02-21
Feat stripe bank transfers

brunomiguelpinto:
## Context
Lago does not currently support Bank Transfer payments for users receiving payouts through Stripe’s customer balance. This PR adds support for this payment method to accommodate use cases where Bank Transfers are required.

## Description
Added support for Bank Transfer payments when using Stripe’s customer balance.
Updated relevant service logic to handle this payment method.

### DIFF
--- a/app/services/payment_providers/stripe/payments/create_service.rb
+++ b/app/services/payment_providers/stripe/payments/create_service.rb
@@ -171,8 +171,8 @@ module PaymentProviders
         # to permit 3D secure authentication
         # https://docs.stripe.com/india-recurring-payments
         def off_session?
-          return false if customer.country == "IN"
-          return false if customer.stripe_customer.provider_payment_methods.include?("customer_balance")
+          return false if invoice.customer.country == "IN"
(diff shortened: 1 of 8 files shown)


app/services/payment_providers/stripe/payments/create_service.rb
```diff
@@ -139,6 +171,7 @@ def success_redirect_url
         # to permit 3D secure authentication
         # https://docs.stripe.com/india-recurring-payments
         def off_session?
+          return false if provider_customer.provider_payment_methods == ["customer_balance"]
           invoice.customer.country != "IN"
```
julienbourdeau, line 175: Nitpick but for consistency I'd prefer if it looked like `off_session?` in StripeService

But, now I'm wondering why I didn't update this with crypto 🤔 

https://github.com/getlago/lago-api/blob/aba70bdb294c9fcd6ccbc36f9ded1b4155f2b5ba/app/services/invoices/payments/stripe_service.rb#L212-L217

julienbourdeau (approved):

PR #3238, opened 2025-02-21
misc(payment): Factorise logic for payment_url generation

vincent-pochet:
## Context

A lot of business logic is duplicated in all services related to the generation of the payment url. 

## Description

This PR:
- extract the `should_process_payment?` and `deliver_error_webhook` from `Invoices::Adyen|Cashfree|StripeService` and  `PaymentRequests::Adyen|Cashfree|StripeService` into `Invoices|PaymentRequests::GeneratePaymentUrlService` to avoid duplication
- Handles provider errors and returns a ThirdPartyFailure with the detail of the error, allowing a better handling of these failures (See discussions on https://github.com/getlago/lago-api/pull/3215, about a potential renaming of this type of service failure)


rsempe (approved):
🙌

julienbourdeau (approved):

julienbourdeau:
Excellent! This is a great improvment so we should get it in.
I'll share some personal opinions that can be addressed now or later.

1. I'm not a fan of the "Third Party" terminology because we use "Provider" everywhere else.
2. For external services error, I'd love to have a free form object too where we can just forward more details from the provider (see screenshot)
3. Ideally, I'd like the provider name to be set in the base service. Something like:

```ruby
class StripeService < BaseService
  PROVIDER_NAME = "Stripe"
end

class BaseResult
  def thridpartyerror(...)
    fail_with_error(provider: PROVIDER_NAME, ...)
  end
end
```

CleanShot 2025-02-18 at 12 39 48@2x

vincent-pochet:
> Excellent! This is a great improvment so we should get it in. I'll share some personal opinions that can be addressed now or later.
> 
>     1. I'm not a fan of the "Third Party" terminology because we use "Provider" everywhere else.
> 
>     2. For external services error, I'd love to have a free form object too where we can just forward more details from the provider (see screenshot)
> 
>     3. Ideally, I'd like the provider name to be set in the base service. Something like:
> 
> 
> ```ruby
> class StripeService < BaseService
>   PROVIDER_NAME = "Stripe"
> end
> 
> class BaseResult
>   def thridpartyerror(...)
>     fail_with_error(provider: PROVIDER_NAME, ...)
>   end
> end
> ```
> 
> CleanShot 2025-02-18 at 12 39 48@2x

Thank you for the review!

1- I agree that `Third Party` is not really good. As discussed, I think we should use either `Provider` or `Integration` instead, to keep the terminologies we already have in the project. I will not change it in this PR, but I keep it in mind for a new one :) 
2- Same, I agree, I would like to have a single common way to handle all provider errors, in both APIs and webhooks with an easy way of attaching the initial error so that it can we exposed. Even from us, I think it would make debugging and development easier.
3- Good point! Updating it :+1:

PR #3242, opened 2025-02-24
feat(auth): Add user last login informations

jdenquin:
- Add users `last_login_at` and `last_login_method` 
- Impact all auth services for this change
- Require `password` only on users creation for validations (this made us unable to update a user)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -4,6 +4,14 @@ class User < ApplicationRecord
   include PaperTrailTraceable
   has_secure_password
 
+  LOGIN_METHODS = [
+    :email,
+    :google,
+    :okta
+  ].freeze
+
+  enum :last_login_method, LOGIN_METHODS
+
   has_many :password_resets
 
   has_many :memberships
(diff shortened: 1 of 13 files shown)


app/services/users_service.rb
```diff
@@ -1,22 +1,25 @@
...
 
     unless result.user.present? && result.user.memberships&.active&.any?
       return result.single_validation_failure!(error_code: "incorrect_login_or_password")
     end
 
-    result.token = generate_token if result.user
+    if result.user
+      result.token = generate_token if result.user
+      result.user.touch_last_login!(method) if result.user
+    end
```
rsempe, line 14:
```suggestion
    if result.user
      result.token = generate_token
      result.user.touch_last_login!(method)
    end
```

app/models/user.rb
```diff
@@ -19,20 +27,26 @@ class User < ApplicationRecord
   has_many :subscriptions, through: :customers
 
   validates :email, presence: true
-  validates :password, presence: true
+  validates :password, presence: true, on: :create
```
nudded, line 30: We don't want to allow empty passwords on update I think?

jdenquin: this will not override password if its nil? (since the password digest should stay)

nudded: how can this validation fail on touch?

jdenquin: I updated this with 

```
validates :password, presence: true, if: lambda { new_record? || !password.nil? }
```

so we check that the password is not nil and not empty if its a new record and if its an update, we do not validate if password is nil (so will not be updated)

vincent-pochet (approved):

PR #3246, opened 2025-02-24
feat(payment-receipts): Add payment receipts controllers

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds controllers, serializers and queries for payment receipts.

It adds these API endpoints:

```
GET /api/v1/payment_receipts/:id
GET /api/v1/payment_receipts
GET /api/v1/payment_receipts?invoice_id=:invoice_id
```


app/contracts/queries/payment_receipts_query_filters_contract.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Queries
+  class PaymentReceiptsQueryFiltersContract < Dry::Validation::Contract
+    UUID_REGEX = /\A[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\z/
```
julienbourdeau, line 5: This REGEX is already defined in `PaymentsQueryFiltersContract`. I think it should be extracted somewhere. I'm not sure where to be honest.
Maybe a new `app/support/regex.rb` and we can move `EmailValidator::EMAIL_REGEXP` there too.

db/migrate/20250224192149_add_payment_receipt_permission_to_api_keys.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddPaymentReceiptPermissionToApiKeys < ActiveRecord::Migration[7.1]
+  def up
+    ApiKey.update_all(permissions: ApiKey.default_permissions) # rubocop:disable Rails/SkipsModelValidations
```
julienbourdeau, line 5: Is the Api Key permissions feature not fully released yet so nobody can update their key permissions?

app/config/permissions/definition.yml
```diff
@@ -106,3 +106,5 @@ payment_requests:
 payments:
   view: true
   create:
+payment_receipts:
+  view: true
```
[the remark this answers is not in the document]
ivannovosad: This was causing issues in the controller - the `resource_name` method

db/migrate/20250224192149_add_payment_receipt_permission_to_api_keys.rb
ivannovosad: If they call the endpoint without having the premium integration enabled, it just returns an empty result.

julienbourdeau: But if you have the integration and you have keys with customer permissions, like a key that can only send events, it seems that you're going to over override all permissions to all keys to set the default.

So I'm wondering if all keys in the DB have default because it's not possible to customize yet 🤔

ivannovosad: We already ran this type of migration a couple of times. Lets double check this with @floganz

floganz: That migration was OK during development phase with no customers using it.
This feature is now live and some keys might have custom permissions and such action would override it.

We can have two cases:
1. addon is not enabled.
2. addon is enabled and customer might edit or not permissions per key.

For 1st case we don't even need to update the data (logic will allow everything until addon enabled). Once addon will be enabled customer will check or uncheck what ever permission they want.
For 2nd case it's unclear what to do. From one side we can add new permission to their existing permissions. But that not obvious, .e.g. they have a key with read only permissions. Adding new permission with read & write would be wrong.

Summing up, I'd suggest in case when we add new resources – do not update exiting keys.
Wdyt @julienbourdeau @ivannovosad? Does everything above makes sense for you?

ivannovosad: @floganz Thank you for the explanation, I'm removing the migration.
cc: @julienbourdeau

app/config/permissions/definition.yml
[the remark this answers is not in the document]
julienbourdeau: Yeah unfortunately, we tend to use the resource_name for the serializer root name AND for the permissions. We can keep it but I'm afraid we'll start having too many permissions soon.

julienbourdeau (approved):

app/config/permissions/definition.yml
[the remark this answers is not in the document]
ivannovosad: I see, lets address this issue in a separate feature.

PR #3247, opened 2025-02-24
fix(stripe): Better errors on Stripe customer checkout url

julienbourdeau:
## Description

Some stripe errors were not returned by the API when creating a customer checkout url.

This PR also split the payment methods into 2 categories. You cannot create a cutomer checkout url if all their available payment methods are not able to be setup. (this will be very useful for https://github.com/getlago/lago-api/pull/3236)

### DIFF
--- a/spec/models/payment_provider_customers/stripe_customer_spec.rb
+++ b/spec/models/payment_provider_customers/stripe_customer_spec.rb
@@ -30,6 +30,14 @@ RSpec.describe PaymentProviderCustomers::StripeCustomer, type: :model do
     end
   end
 
+  describe "#provider_payment_methods_for_setup" do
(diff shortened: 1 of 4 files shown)


spec/models/payment_provider_customers/stripe_customer_spec.rb
```diff
@@ -30,6 +30,14 @@
     end
   end
 
+  describe "#provider_payment_methods_for_setup" do
```
ivannovosad, line 33:
```suggestion
  describe "#provider_payment_methods_with_setup" do
```

ivannovosad (changes requested):

brunomiguelpinto (approved):

PR #3248, opened 2025-02-24
feat(payment-receipts): Add pdf templates

ivannovosad:
## Roadmap Task

## Context

## Description

This PR adds pdf templates for payment receipts - both for invoices and payment requests.
It also add webhooks to be sent after generating or creating a receipt.


db/migrate/20250224192149_add_payment_receipt_permission_to_api_keys.rb
```diff
@@ -0,0 +1,13 @@
...
+class AddPaymentReceiptPermissionToApiKeys < ActiveRecord::Migration[7.1]
+  def up
+    ApiKey.update_all(permissions: ApiKey.default_permissions) # rubocop:disable Rails/SkipsModelValidations
+  end
+
+  def down
+    ApiKey.update_all( # rubocop:disable Rails/SkipsModelValidations
+      permissions: ApiKey.default_permissions.without("payment_receipt")
+    )
+  end
```
julienbourdeau, line 12: I thought we didn't want to do this because if a keys only have `events:write` for instance, we would modify it and give it all the permissions.

spec/models/payment_receipt_spec.rb
```diff
@@ -7,4 +7,23 @@
 
   it { is_expected.to belong_to(:payment) }
   it { is_expected.to belong_to(:organization) }
+  it { is_expected.to have_one_attached(:file) }
```
julienbourdeau, line 10: I think we should start saving on `it` since our test suite is getting pretty long.
Obviously here it's not an issue with 3 its but I'm saying this as a general rule, we should try to refactor when we can.
See https://github.com/getlago/lago-api/pull/3167
```suggestion
  it do
   expect(subject).to ...
   ...
 end
```

julienbourdeau (approved):

PR #3251, opened 2025-02-25
Feat(multi-entities): add services to update billing_entity

annvelents:
## Context

We need a service to update billing_entity data that is doing exactly everything that is currently done when updating an organization.
Since in the future we'll be storing grace period, net_payment term only on billing_entity, we also need services that will be updating these values on invoices, but we shouldn't trigger invoices updates now

## Description

Added services:
- `billing_entities/update`
- `billing_entities/update_invoice_grace_period`
- `billing_entities/update_invoice_payment_due_date`
- `invoices/update_all_invoice_grace_period_from_billing_entity`
- `invoices/update_grace_period_from_billing_entity`

Also added jobs:
- `app/jobs/invoices/update_all_invoice_grace_period_from_billing_entity_job.rb`
- `app/jobs/invoices/update_grace_period_from_billing_entity_job.rb`

For now the jobs are not triggered and update invoice update services are not triggered

### DIFF
--- a/app/models/billing_entity.rb
+++ b/app/models/billing_entity.rb
@@ -65,6 +65,29 @@ class BillingEntity < ApplicationRecord
 
   after_create :generate_document_number_prefix
 
+  def document_number_prefix=(value)
+    super(value&.upcase)
+  end
+
+  def logo_url
+    return if logo.blank?
+
+    Rails.application.routes.url_helpers.rails_blob_url(logo, host: ENV["LAGO_API_URL"])
+  end
+
+  def base64_logo
+    return if logo.blank?
+
+    logo.blob.open do |tempfile|
+      data = tempfile.read
+      Base64.encode64(data)
+    end
+  end
+
+  def eu_vat_eligible?
+    country && LagoEuVat::Rate.new.countries_code.include?(country)
+  end
+
   private
 
   def generate_document_number_prefix
--- a/app/services/billing_entities/update_invoice_payment_due_date_service.rb
+++ b/app/services/billing_entities/update_invoice_payment_due_date_service.rb
@@ -2,6 +2,8 @@
 
 module BillingEntities
   class UpdateInvoicePaymentDueDateService < BaseService
+    Result = BaseResult[:billing_entity]
+
     def initialize(billing_entity:, net_payment_term:)
       @billing_entity = billing_entity
       @net_payment_term = net_payment_term
@@ -10,18 +12,18 @@ module BillingEntities
 
     def call
       ActiveRecord::Base.transaction do
-        # NOTE: Update payment_due_date if net_payment_term changed
-
+        # NOTE: Set payment_due_date if net_payment_term changed
         if billing_entity.net_payment_term != net_payment_term
           billing_entity.net_payment_term = net_payment_term
 
-          # update only invoices, where the customer does not have a setting
-          billing_entity.invoices.includes(:customer).draft.find_each do |invoice|
-            # the customer has a setting of their own, no update needed.
-            next unless invoice.customer.net_payment_term.nil?
-
-            invoice.update!(net_payment_term:, payment_due_date: invoice_payment_due_date(invoice))
-          end
+          # TODO: uncomment when billing_entity is the source of truth
+          # # update only invoices, where the customer does not have a setting
+          # billing_entity.invoices.includes(:customer).draft.find_each do |invoice|
+          #   # the customer has a setting of their own, no update needed.
+          #   next unless invoice.customer.net_payment_term.nil?
+          #
+          #   invoice.update!(net_payment_term:, payment_due_date: invoice_payment_due_date(invoice))
+          # end
         end
 
         result.billing_entity = billing_entity
--- a/app/services/billing_entities/update_invoice_payment_due_date_service.rb
+++ b/app/services/billing_entities/update_invoice_payment_due_date_service.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module BillingEntities
+  class UpdateInvoicePaymentDueDateService < BaseService
+    def initialize(billing_entity:, net_payment_term:)
+      @billing_entity = billing_entity
+      @net_payment_term = net_payment_term
+      super
+    end
+
+    def call
+      ActiveRecord::Base.transaction do
+        # NOTE: Update payment_due_date if net_payment_term changed
+
+        if billing_entity.net_payment_term != net_payment_term
+          billing_entity.net_payment_term = net_payment_term
+
+          # update only invoices, where the customer does not have a setting
+          billing_entity.invoices.includes(:customer).draft.find_each do |invoice|
+            # the customer has a setting of their own, no update needed.
+            next unless invoice.customer.net_payment_term.nil?
+
+            invoice.update!(net_payment_term:, payment_due_date: invoice_payment_due_date(invoice))
+          end
+        end
+
+        result.billing_entity = billing_entity
+        result
+      end
+    end
+
+    private
+
+    attr_reader :billing_entity, :net_payment_term
+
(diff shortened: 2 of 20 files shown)


app/models/billing_entity.rb
```diff
@@ -65,6 +65,29 @@ class BillingEntity < ApplicationRecord
...
+
+    logo.blob.open do |tempfile|
+      data = tempfile.read
+      Base64.encode64(data)
+    end
+  end
+
+  def eu_vat_eligible?
+    country && LagoEuVat::Rate.new.countries_code.include?(country)
+  end
```
ancorcruz, line 89: as this is public interface of the billing entity model... could we have it covered with tests please?

spec/services/billing_entities/update_invoice_grace_period_service_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  let(:grace_period) { 2 }
+
+  describe "#call" do
+    # let(:invoice_to_be_finalized) do
+    #   create(:invoice, status: :draft, customer:, issuing_date: DateTime.parse("19 Jun 2022").to_date, organization:)
+    # end
+    #
+    # let(:invoice_to_not_be_finalized) do
+    #   create(:invoice, status: :draft, customer:, issuing_date: DateTime.parse("21 Jun 2022").to_date, organization:)
+    # end
```
ancorcruz, line 21: I know this is commented code for later, however, I would expect the code to be correct... should the created invoices be linked to the billing entity here?

spec/services/billing_entities/update_invoice_grace_period_service_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    #         .to(DateTime.parse("23 Jun 2022"))
+    #         .and change { invoice_to_be_finalized.reload.issuing_date }
+    #         .to(DateTime.parse("21 Jun 2022"))
+    #         .and change { invoice_to_not_be_finalized.reload.payment_due_date }
+    #         .to(DateTime.parse("23 Jun 2022"))
+    #     end
+    #   end
+    #
+    #   context "when customer has net_payment_term" do
+    #     let(:customer) { create(:customer, organization:, net_payment_term: 3) }
```
ancorcruz, line 64: same here... shouldn't the customer be linked to the billing entity or does it happen automagically by the factory?

spec/services/billing_entities/update_invoice_grace_period_service_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    #         expect { perform_enqueued_jobs { update_service.call } }.to change { invoice_to_not_be_finalized.reload.issuing_date }
+    #           .to(DateTime.parse("23 Jun 2022"))
+    #           .and change { invoice_to_not_be_finalized.reload.payment_due_date }
+    #           .to(DateTime.parse("26 Jun 2022"))
+    #       end
+    #     end
+    #   end
+    #
+    #   context "when grace_period is the same as the current one" do
+    #     let(:grace_period) { organization.invoice_grace_period }
```
ancorcruz, line 79: Should this one be about the billing entity invoice grace period instead?

```suggestion
    #     let(:grace_period) { billing_entity.invoice_grace_period }
```

app/services/billing_entities/update_invoice_payment_due_date_service.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module BillingEntities
+  class UpdateInvoicePaymentDueDateService < BaseService
+    def initialize(billing_entity:, net_payment_term:)
```
ancorcruz, line 5:
```suggestion
    Result = BaseResult[:billing_entity]

    def initialize(billing_entity:, net_payment_term:)
```

app/services/billing_entities/update_invoice_payment_due_date_service.rb
```diff
@@ -0,0 +1,42 @@
...
+      super
+    end
+
+    def call
+      ActiveRecord::Base.transaction do
+        # NOTE: Update payment_due_date if net_payment_term changed
+
+        if billing_entity.net_payment_term != net_payment_term
+          billing_entity.net_payment_term = net_payment_term
+          billing_entity.save!
```
ancorcruz, line 17: reading the old organization service I realized it assigns the new net payment term but does not save the record  changes (as you do here)... this looks like a bug to me. I'll raise it to the team.

spec/services/billing_entities/update_invoice_payment_due_date_service_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  describe "#call" do
+    it "updates invoice payment_due_date" do
+      expect { update_service.call }.to change { billing_entity.reload.net_payment_term }
+        .from(0).to(30)
+    end
+
+    # TODO: uncomment when we start updating invoices
+    # let(:draft_invoice) do
+    #   create(:invoice, status: :draft, customer:, organization:, issuing_date: DateTime.parse("21 Jun 2022"))
```
ancorcruz, line 22: should the invoice be linked to the billing entity as well as the organization?

app/services/billing_entities/update_service.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+
+      assign_premium_attributes
+      handle_base64_logo if params.key?(:logo)
+
+      billing_entity.save!
+
+      # I guess we dont need this...
+      # ApiKeys::CacheService.expire_all_cache(billing_entity)
+
```
ancorcruz, line 61: if we don't need this, just delete the code. It could be added if we decide to manage api keys at billing entity level.

spec/services/billing_entities/update_service_spec.rb
```diff
@@ -0,0 +1,254 @@
...
+        invoice_grace_period:
+      }
+    }
+  end
+
+  describe "#call" do
+    it "updates the billing_entity" do
+      result = update_service.call
+
+      aggregate_failures do
```
ancorcruz, line 44: aggregate_failures is now applied by default. We don't need to include it. This applies to all other test in this file.

ancorcruz (changes requested):
Looks great. I have left a few questions and comments...

spec/services/billing_entities/update_invoice_grace_period_service_spec.rb
annvelents: it will be connected, because when we create invoice, passing organization, we assign the invoice to the default billing_entity for this organization. Though, you're right, it's better to do it explicitly

app/models/billing_entity.rb
annvelents: yep, great point, thank you.
Was focusing on the services and forgot athat I also update the model 🤦‍♀️

spec/services/billing_entities/update_invoice_grace_period_service_spec.rb
annvelents: yep, the same.
(because we have a lot of places where they are created exactly this way, to not break everything I've added assigning billing_entity unless it's passed)
but yes, will add it explicitly

spec/services/billing_entities/update_invoice_grace_period_service_spec.rb
annvelents: 🤦‍♀️ yes

spec/services/billing_entities/update_service_spec.rb
annvelents: 🙏

app/services/billing_entities/update_service.rb
annvelents: 🚀

ancorcruz (approved):
All looks good... just one comment.

PR #3252, opened 2025-02-25
feat(multi-entities): Add multi_entities premium integrations with two tiers...

ancorcruz:
## Roadmap Task

## Context

## Description

`multi_entities_pro` -> allows 2 billing entities
`multi_entities_enterprise` -> allows unlimited billing entities

Also add the means to check if the organzation can create new entities not reaching already the max number of billing entities per integration tier.


annvelents (approved):
🚀

PR #3257, opened 2025-02-26
feat(invoice-preview): Add invoice preview with simulated termination date on subscription

floganz:
## Context

Allow to simulate subscription termination for invoice preview. This part only accepts new param and prepare correct subscription object for `PreviewService` which responsible for all invoice generation.

## Description

Add support for `terminated_at` timestamp.
Add validation that only 1 subscription could be terminated.
Add validation for provided timestamp (valid + not in past).
Extract subscription building logic into dedicated class for better separation of concerns.

### DIFF
--- a/app/services/invoices/preview/subscriptions_service.rb
+++ b/app/services/invoices/preview/subscriptions_service.rb
@@ -41,6 +41,7 @@ module Invoices
 
         customer_subscriptions.map do |subscription|
           subscription.terminated_at = terminated_at
+          subscription.status = :terminated
           subscription
         end
       end
--- a/app/services/invoices/preview/subscriptions_service.rb
+++ b/app/services/invoices/preview/subscriptions_service.rb
@@ -1,0 +1,113 @@
+# frozen_string_literal: true
+
+module Invoices
+  module Preview
+    class SubscriptionsService < BaseService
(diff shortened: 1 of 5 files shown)


app/services/invoices/preview/subscriptions_service.rb
```diff
@@ -0,0 +1,114 @@
...
+            error_code: "cannot_be_in_past",
+            field: :terminated_at
+          )
+        end
+
+        result.success?
+      end
+
+      def parsed_terminated_at
+        Time.zone.parse(terminated_at)
```
vincent-pochet, line 95: `Time.zone.parse` in itselft is not 100% reliable when it comes to validating dates...
We have `Utils::Datetime.valid_format?(terminated_at)` to improve it.

```
irb> Time.zone.parse('ff')
=> nil
irb> Time.zone.parse('2023')
(irb):3:in `<main>': argument out of range (ArgumentError)
irb> Time.zone.parse('12')
=> Wed, 12 Feb 2025 00:00:00.000000000 UTC +00:00
```

If that helps, we could even add a method doing both the validation and the parsing

vincent-pochet (approved):

lovrocolic (approved):

PR #3260, opened 2025-02-26
bug(WalletTransaction) - round to right amount of digits for wallet transaction creation

nudded:


brunomiguelpinto (approved):
should we update what was previously created with the "wrong decimal cases"?

vincent-pochet (approved):

ancorcruz (approved):
@nudded Would this change fix the one cent off wallet incorrect balance vs wallet transactions?

vincent-pochet (approved):
Great refactor! :ship:

PR #3262, opened 2025-02-26
feat(stripe): Pre-authorize payment before creating subscription

julienbourdeau:
## Context

When creating a subscription, the invoice and the payment are done asynchronously. We're introducing a way to make a pre-authorization on the credit card to mitigate payment failure.

## Description

**The feature should be released in BETA** until we're satisfied with the request/response shape and we added support for other payment provider. Only Stripe is supported now.

CleanShot 2025-02-26 at 11 54 13@2x

### Current situation

When creating a subscription, we'll handle the creation, termination of current subscription and update of future subscriptions if needed.

Then in async, we create the invoice, which in async, will collect the payment. The payment can fail but the subscription is already returned and active.

### Solution to mitigate payment issues

Before creating a subscription, we can create a pre-authorization on the customer cards. This will make a payment, lock the amount on the customer account but not actually take away the money (called capture). The amount and currency are passed to the API by the user.

This will ensure the card is valid and working.

Note that: 
* the total due amount is unknown at the moment of the subscription creation (only know after invoice is finalized)
* the payment intent created cannot be reused because the final amount could be higher than the capturable amount.
* the payment intent is cancelled **right away but asynchronously**.
* This is why it's recommended to **only do it with very small amounts** (like $1). This won't protect agains "insuffisant fund" but will definitely help reduce the number of error
  * If you lock too much money on the customer's account, I beleive we'll actually increase the number of "insuffisant fund" error.
  * example: If you need lock $20, then charge $22 before the initial $20 are released, you know need $42 available

> [!WARNING]  
> Nothing is saved in the database (no new `Payment` type for instance)
> The job directly calls the stripe API, it's not wrapped in a Service
> Only Stripe is supported
>
> **This is intentional** because I believe we should test this feature on beta before we go further with the implementation.

## TODO

- [x] Initial implement with Stripe
- [ ] Decide on API (parameter names and response)
- [x] Add tests

CleanShot 2025-02-26 at 14 55 50@2x
CleanShot 2025-02-26 at 14 57 29@2x

### DIFF
--- a/app/controllers/api/v1/subscriptions_controller.rb
+++ b/app/controllers/api/v1/subscriptions_controller.rb
@@ -37,8 +37,8 @@ module Api
           end
 
           result = PaymentProviders::Stripe::Payments::AuthorizeService.call(
-            amount: params[:authorization][:amount],
-            currency: params[:authorization][:currency],
+            amount: params[:authorization].fetch(:amount_cents),
+            currency: params[:authorization].fetch(:amount_currency),
             provider_customer: customer.provider_customer,
             metadata: {plan_code: create_params[:plan_code]},
             unique_id: request.request_id
--- a/app/controllers/api/v1/subscriptions_controller.rb
+++ b/app/controllers/api/v1/subscriptions_controller.rb
@@ -4,10 +4,54 @@ module Api
   module V1
     class SubscriptionsController < Api::BaseController
       def create
+        response = {}
+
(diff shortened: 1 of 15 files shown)


annvelents (approved):
LGTM 🚀

vincent-pochet (approved):
Looks good as a POC :+1:

PR #3266, opened 2025-02-27
Feat(multientities):remove usage of is_default flag on billing_entities

annvelents:
## Context

We decided to proceed without default billing entity. Instead we'll be treating the first active billing entity of organization as default if an organization has only one billing entity, otherwise we act as if organization has no default billing entity and forcing client to send the billing_entity_id

### DIFF
--- a/spec/factories/customers.rb
+++ b/spec/factories/customers.rb
@@ -23,7 +23,7 @@ FactoryBot.define do
     currency { "EUR" }
 
     after :build do |customer, values|
-      customer.billing_entity = values.organization.default_billing_entity unless customer.billing_entity
+      customer.billing_entity = values.organization&.default_billing_entity unless customer.billing_entity
     end
 
     trait :with_shipping_address do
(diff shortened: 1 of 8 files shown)


julienbourdeau (approved):

spec/factories/customers.rb
```diff
@@ -23,7 +23,7 @@
     currency { "EUR" }
 
     after :build do |customer, values|
-      customer.billing_entity = values.organization.default_billing_entity unless customer.billing_entity
+      customer.billing_entity = values.organization&.default_billing_entity unless customer.billing_entity
```
ancorcruz, line 26:
```suggestion
      customer.billing_entity ||= values.organization&.default_billing_entity
```

spec/factories/invoices.rb
```diff
@@ -13,7 +13,7 @@
     organization_sequential_id { rand(1_000_000) }
 
     after :build do |invoice, values|
-      invoice.billing_entity = values.organization.default_billing_entity unless invoice.billing_entity
+      invoice.billing_entity = values.organization&.default_billing_entity unless invoice.billing_entity
```
ancorcruz, line 16:
```suggestion
      invoice.billing_entity ||= values.organization&.default_billing_entity
```

ancorcruz (approved):
Looks good to me.

PR #3267, opened 2025-02-27
Feat(multi-entities): migration to remove is_default from billing_entities

annvelents:
## Context

Column drop should be happening in a separate commit


ancorcruz (approved):
LGTM

PR #3272, opened 2025-02-28
chore(logs): Improve logging to json

jdenquin:
New Active Job logging format 

```
{:event=>"perform", :status=>"error", :job=>"Invoices::GeneratePdfAndNotifyJob", :duration=>528.95, :job_id=>"11c160df-7edb-45fe-9ab5-31c7db66e97e", :queue=>"invoices", :exception=>{:class=>"Socket::ResolutionError", :message=>"Failed to open TCP connection to pdf:3000 (getaddrinfo: Name or service not known)"}}
```

### DIFF
--- a/lib/active_job/json_log_subscriber.rb
+++ b/lib/active_job/json_log_subscriber.rb
@@ -1,11 +1,13 @@
-# frozen_string_literal
+# frozen_string_literal: true
 
 require "active_support/subscriber"
 require "active_job/log_subscriber"
 
 module ActiveJob
   class JsonLogSubscriber < ActiveSupport::LogSubscriber # :nodoc:
+    # rubocop:disable ThreadSafety/ClassAndModuleAttributes
     class_attribute :backtrace_cleaner, default: ActiveSupport::BacktraceCleaner.new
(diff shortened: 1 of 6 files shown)


nudded (approved):

lib/active_job/json_log_subscriber.rb
```diff
@@ -0,0 +1,267 @@
...
+      job = event.payload[:job]
+      ex = event.payload[:exception_object] || job.enqueue_error
+
+      if ex
+        error do
+          {
+            event: "enqueue",
+            status: "error",
+            job: job.class.name,
+            queue: job.queue_name,
```
julienbourdeau, line 22: Is there a reason to not have the arguments here? 🤔 
and for other logs on error

```suggestion
            queue: job.queue_name,
            arguments: args_info(job)
```

julienbourdeau (approved):

lib/active_job/json_log_subscriber.rb
jdenquin: yes! we have the dead job for this! btw default AJ logger does not provides arguments on error

PR #3275, opened 2025-02-28
feat(invoice-preview): add upgrade logic

lovrocolic:
## Context

Preview feature enables to see invoice breakdown on the fly, without storing any object in the DB

## Description

This PR covers support for upgraded invoices. For this specific case, correct boundaries are fetched and certain fees skipped if not applicable.

### DIFF
--- a/app/services/invoices/preview_service.rb
+++ b/app/services/invoices/preview_service.rb
@@ -302,7 +302,7 @@ module Invoices
       charge.billable_metric.recurring? &&
         subscription.terminated? &&
         subscription.upgraded? &&
-        charge.charge_included_in_next_subscription?(subscription)
+        charge.included_in_next_subscription?(subscription)
     end
   end
 end
--- a/app/services/invoices/preview_service.rb
+++ b/app/services/invoices/preview_service.rb
@@ -138,12 +138,12 @@ module Invoices
       invoice.fees = subscriptions
         .filter { |subscription| should_create_subscription_fee?(subscription) }
         .map do |subscription|
-          Fees::SubscriptionService.call!(
-            invoice:,
-            subscription:,
-            boundaries: boundaries(subscription),
-            context: :preview
-          ).fee
+        Fees::SubscriptionService.call!(
+          invoice:,
+          subscription:,
+          boundaries: boundaries(subscription),
+          context: :preview
+        ).fee
       end
     end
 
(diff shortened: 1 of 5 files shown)


app/services/invoices/preview_service.rb
```diff
@@ -289,10 +280,10 @@ def provider_taxation?
       customer.integration_customers.find { |ic| ic.type == "IntegrationCustomers::AnrokCustomer" }
     end
 
-    def should_create_subscription_fee?
+    def should_create_subscription_fee?(subscription)
       return true if subscription_context == :default
 
-      subscription_context == :terminated && first_subscription.plan.pay_in_arrear?
+      subscription.terminated? == subscription.plan.pay_in_arrear?
```
floganz, line 286: Just do double check: `should_create_subscription_fee` should return `true` in 2 cases:
- terminated subscription and plan pay in advance
- non-terminated subscription and plan is not pay in advance?

app/services/invoices/preview_service.rb
```diff
@@ -86,15 +85,6 @@ def billing_times_aligned?
       true
     end
 
-    def terminate_allowed?
```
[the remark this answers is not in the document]
floganz: Agreed to skip this validation in this service, we'll do that in `PreviewContextService`.

app/services/invoices/preview_service.rb
lovrocolic: I would say this is true.
For context that is not terminated we always return `true`.
However, terminated `context` can happen:
- when one subscription is terminated
- with upgrade or downgrade

In case of downgrade or upgrade new subscription needs to be billed, only if it is pay in advance

spec/services/invoices/preview_service_spec.rb
```diff
@@ -513,7 +535,7 @@
                 expect(result.invoice.sub_total_excluding_taxes_amount_cents).to eq(19)
                 expect(result.invoice.taxes_amount_cents).to eq(10)
                 expect(result.invoice.sub_total_including_taxes_amount_cents).to eq(29)
-                expect(result.invoice.total_amount_cents).to eq(29)
+                expect(result.invoice.total_amount_cents).to eq(20)
```
lovrocolic, line 538: Any idea why CN is here 9?

floganz: It was after rebase on feature branch where generating credit note for terminated subscription was merged.
It should be affect of having credit note.

floganz: I've checked the numbers, those looks correct to me.

According to test setup we create subscription with 1 EUR/month price. It's created March 28th and calendar billing. Since plan pay in advance, we're creating invoice for 4 days of March (3.22 cents per day) = 12.90.

Then when we simulate subscription termination on March 29th. Then half of that amount that should be refunded (for March 30 and 31) which is ~ 6.45 + taxes on that amount 3 = 9.45 which is rounded to 9.

floganz (approved):

PR #3279, opened 2025-03-03
Feat(multientities): add billing_entity code uniqueness validation

annvelents:
## Context

When creating or updating a billing entities, the code should be unique within the organization for not deleted and not archived records

## Description

added uniqueness index and validation for code with organization_id


ancorcruz (approved):
LGTM

PR #3295, opened 2025-03-05
Feat(multientiies): create billing_entity that follows organization

annvelents:
## Context

For each organization we need a billing_entity to be created and to be shadowing the organization data

## Description

- Changed definition of what is billing_entity for an organization
- When updating organization, update also it's default billing_entity
- Create billing_entity per each organization


app/services/organizations/update_service.rb
```diff
@@ -33,40 +33,49 @@ def call
...
-      handle_base64_logo if params.key?(:logo)
-
-      organization.save!
-
       ApiKeys::CacheService.expire_all_cache(organization)
 
       result.organization = organization
       result
+    rescue BaseService::NotFoundFailure => _
+      result.not_found_failure!(resource: "default_billing_entity")
```
ancorcruz, line 74: this could hide some other resource not be found.

db/migrate/20250304163656_create_billing_entity_per_each_organization.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+class CreateBillingEntityPerEachOrganization < ActiveRecord::Migration[7.1]
```
ancorcruz, line 3:
```suggestion
class CreateBillingEntityPerEachOrganization < ActiveRecord::Migration[7.2]
```

ancorcruz (approved):
LGTM

PR #3298, opened 2025-03-05
fix: make json parsing more resilient

lovrocolic:
## Context

When parsing json and converting to ruby hash, we want to be more resilient to parsing errors

## Description

This PR returns empty hash, for invalid json

### DIFF
--- a/lib/lago_http_client/lago_http_client/http_error.rb
+++ b/lib/lago_http_client/lago_http_client/http_error.rb
@@ -16,6 +16,8 @@ module LagoHttpClient
 
     def json_message
       JSON.parse(error_body)
+    rescue JSON::ParserError
+      {}
     end
   end
 end


lib/lago_http_client/lago_http_client/http_error.rb
```diff
@@ -16,6 +16,8 @@ def message
 
     def json_message
       JSON.parse(error_body)
+    rescue JSON::ParserError
+      {}
```
julienbourdeau, line 20: Maybe we should catch the error in sentry and log it, similar to what we do there.
Similar to what we do there: [/clock.rb#L13-L16](https://github.com/getlago/lago-api/blob/aab15479d4631717f429146086905a6275dc952b/clock.rb#L13-L16)

```suggestion
    rescue JSON::ParserError => e
      Rails.logger.error("Received invalid json body: #{error_body}") # or just e.message
      Sentry.capture_exception(e)
      {}
```

julienbourdeau (approved):

lib/lago_http_client/lago_http_client/http_error.rb
julienbourdeau: Actually, I thought we might be missing errors but here the error will be handled anyway, it's ust that we try to deserialise some json that isn't json. Sorry ignore me!

PR #3305, opened 2025-03-07
fix(invoice-preview): Add charge duration to boundaries during preview generation

floganz:
## Context

Invoice preview generation logic was missing charge duration property in boundaries which lead to invalid calculations on charge fees.

## Description

Add `charges_duration` to boundaries and adjust tests.

### DIFF
--- a/spec/services/invoices/preview_service_spec.rb
+++ b/spec/services/invoices/preview_service_spec.rb
@@ -550,9 +550,8 @@ RSpec.describe Invoices::PreviewService, type: :service, cache: :memory do
           end
 
           context "with charge fees" do
-            let(:billable_metric) do
-              create(:billable_metric, aggregation_type: "count_agg")
-            end
+            let(:billable_metric) { create(:unique_count_billable_metric) }
+
             let(:charge) do
               create(
                 :standard_charge,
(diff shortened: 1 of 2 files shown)


spec/services/invoices/preview_service_spec.rb
```diff
@@ -588,11 +587,11 @@
                 expect(result.invoice.fees.length).to eq(2)
                 expect(result.invoice.invoice_type).to eq("subscription")
                 expect(result.invoice.issuing_date.to_s).to eq("2024-04-30")
-                expect(result.invoice.fees_amount_cents).to eq(2632)
-                expect(result.invoice.sub_total_excluding_taxes_amount_cents).to eq(2632)
-                expect(result.invoice.taxes_amount_cents).to eq(1316)
-                expect(result.invoice.sub_total_including_taxes_amount_cents).to eq(3948)
-                expect(result.invoice.total_amount_cents).to eq(3948)
+                expect(result.invoice.fees_amount_cents).to eq(1366)
```
lovrocolic, line 590: why we have different results if charge is not prorated it test case?

lovrocolic (approved):

PR #3308, opened 2025-03-07
fix(dunning): adapt refund logic with dunning

lovrocolic:
## Context

Refund logic does not work with dunning campaign

## Description

Currently credit note refund logic works only if payment intent is created with the initial invoice and if it is successful.

However, with dunning campaign, initial payment can be failed, but there can be separate payment request which covers invoice.

This case was not supported and with refund we were linking invalid failed payment.

This PR fixes described case.

### DIFF
--- a/app/services/credit_notes/refunds/stripe_service.rb
+++ b/app/services/credit_notes/refunds/stripe_service.rb
@@ -81,16 +81,16 @@ module CreditNotes
       def payment
         return @payment if defined?(@payment)
 
-        @payment = if credit_note.invoice.payments.where(status: 'succeeded').present?
-          credit_note.invoice.payments.where(status: 'succeeded').order(created_at: :desc).first
+        @payment = if credit_note.invoice.payments.where(status: "succeeded").present?
+          credit_note.invoice.payments.where(status: "succeeded").order(created_at: :desc).first
         else
-          Payment.where(payable_type: 'PaymentRequest')
+          Payment.where(payable_type: "PaymentRequest")
             .joins("INNER JOIN invoices_payment_requests ON invoices_payment_requests.payment_request_id = payments.payable_id")
             .joins("INNER JOIN payment_requests ON payment_requests.id = invoices_payment_requests.payment_request_id")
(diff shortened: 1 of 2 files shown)


app/services/credit_notes/refunds/stripe_service.rb
```diff
@@ -79,7 +79,21 @@ def should_process_refund?
       end
 
       def payment
-        @payment ||= credit_note.invoice.payments.order(created_at: :desc).first
+        return @payment if defined?(@payment)
+
+        @payment = if credit_note.invoice.payments.where(status: "succeeded").present?
+          credit_note.invoice.payments.where(status: "succeeded").order(created_at: :desc).first
```
ancorcruz, line 85: payment `status` attribute is the raw status from the provider and we should not rely on its content for our logic.... 

should we use `payable_payment_status` instead?

app/services/credit_notes/refunds/stripe_service.rb
```diff
@@ -79,7 +79,21 @@ def should_process_refund?
...
+        else
+          Payment.where(payable_type: "PaymentRequest")
+            .joins("INNER JOIN invoices_payment_requests ON invoices_payment_requests.payment_request_id = payments.payable_id")
+            .joins("INNER JOIN payment_requests ON payment_requests.id = invoices_payment_requests.payment_request_id")
+            .where("invoices_payment_requests.invoice_id = ?", credit_note.invoice_id)
+            .where(payments: {status: "succeeded"})
```
ancorcruz, line 91: same here...

ancorcruz (commented):
left some comments.

app/services/credit_notes/refunds/stripe_service.rb
lovrocolic: Updated 👍

app/services/credit_notes/refunds/stripe_service.rb
lovrocolic: updated

ancorcruz (approved):

PR #3313, opened 2025-03-10
feat(wallet): Add Expiration Support to RecurringTransactionRules

brunomiguelpinto:
## Context

Currently, RecurringTransactionRules do not have an expiration mechanism, meaning they continue indefinitely unless manually terminated. This PR introduces an expiration_at field to allow rules to expire automatically at a specified date, improving control over recurring transactions and reducing manual intervention.

## Description

Schema Changes

To support expiration, updated the recurring_transaction_rules table by adding:
expiration_at: Defines when the rule should automatically expire.
terminated_at: Stores the timestamp of when a rule was terminated.
status: Enum to track whether a rule is active or terminated.

Additionally, an index was added to expiration_at to optimize query performance in the new scopes.

Model Updates

Added STATUSES Enum:

```ruby
STATUSES = [
  :active,
  :terminated
].freeze


```

New Scopes for Rule Expiration Handling:

```ruby
scope :active,
scope :eligible_for_termination,
scope :expired
```

Added Background Job for Automatic Rule Termination _TerminateRecurringTransactionRulesJob_ to the clock worker, which periodically checks for and terminates expired rules.

Service Updates

- All services handling RecurringTransactionRules have been updated to respect the new expiration logic:
- Ensuring Only Active Rules are Processed
```
create_interval_wallet_transactions_service.rb
threshold_top_up_service.rb
```
- These services now check whether a rule is active before applying it. 
- when a Wallet is Terminated added the the logic to terminate and the rules  
- Now when updating  the rules we don't delete the old ones now they are terminated


app/models/recurring_transaction_rule.rb
```diff
@@ -25,20 +30,36 @@ class RecurringTransactionRule < ApplicationRecord
...
   enum :method, METHODS
   enum :trigger, TRIGGERS
+  enum :status, STATUSES
+
+  def mark_as_terminated!(timestamp = Time.zone.now)
+    self.terminated_at ||= timestamp
+    terminated!
+  end
+
+  scope :active, -> { where(status: statuses[:active]).where("expiration_at IS NULL OR expiration_at > ?", Time.current) }
```
julienbourdeau, line 40: Where is `statuses` from? is it a feature of `enum`? 😳 I didn't know!

spec/models/recurring_transaction_rule_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  describe "associations" do
+    it { is_expected.to belong_to(:wallet) }
+  end
+
+  describe "enums" do
+    it { is_expected.to define_enum_for(:interval).with_values(%i[weekly monthly quarterly yearly]) }
+    it { is_expected.to define_enum_for(:method).with_values(%i[fixed target]) }
+    it { is_expected.to define_enum_for(:trigger).with_values(%i[interval threshold]) }
+    it { is_expected.to define_enum_for(:status).with_values(%i[active terminated]) }
+  end
```
julienbourdeau, line 15: I know it's very few it blocks here but as our test suite is getting slower (and bigger), I think we should be careful and create les examples when testing the same context

```suggestion
  it do
    expect(described_class).to belong_to(:wallet)
    expect(described_class).to define_enum_for(:interval).with_values(%i[weekly monthly quarterly yearly])
    expect(described_class).to define_enum_for(:method).with_values(%i[fixed target])
    expect(described_class).to define_enum_for(:trigger).with_values(%i[interval threshold])
    expect(described_class).to define_enum_for(:status).with_values(%i[active terminated])
  end
```

spec/models/recurring_transaction_rule_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        expect(described_class.eligible_for_termination).to contain_exactly(expired_rule)
+      end
+    end
+
+    describe ".expired" do
+      it "returns all rules that have expired" do
+        expect(described_class.expired).to contain_exactly(expired_rule, terminated_rule)
+      end
+    end
+  end
```
julienbourdeau, line 40: Same here, I think this should be a single it since you create all 4 items in each.

spec/requests/api/v1/wallets_controller_spec.rb
```diff
@@ -358,6 +389,40 @@
...
+              }
+            ]
+          }
+        end
+
+        it "updates the rule" do
+          subject
+
+          recurring_rules = json[:wallet][:recurring_transaction_rules]
+          aggregate_failures do
```
julienbourdeau, line 416: We have enabled `aggregate_failures` globally so we avoid using it in new tests. No need to change here, just letting you know.

https://github.com/getlago/lago-api/blob/bd2391e6a650ba9d3a562f20fd6eb24b0f130037/spec/spec_helper.rb#L93-L97

julienbourdeau (approved):

app/models/recurring_transaction_rule.rb
brunomiguelpinto: yes its from the enum

spec/requests/api/v1/wallets_controller_spec.rb
brunomiguelpinto: ah yes totally forgot this

PR #3318, opened 2025-03-11
Feat invoice preview minor improvements

floganz:
## Description

Attach credits from credit notes to the preview invoice.
Switch `Subscription#next_subscription` to use array iteration for record finding.
Update logic for simulated subscription termination to also cancel next subscription.

### DIFF
--- a/app/services/invoices/preview/subscription_plan_change_service.rb
+++ b/app/services/invoices/preview/subscription_plan_change_service.rb
@@ -35,6 +35,13 @@ module Invoices
         current_subscription.terminated_at = termination_date
         current_subscription.status = :terminated
 
+        if current_subscription.next_subscription
+          current_subscription.next_subscription.assign_attributes(
+            status: :canceled,
+            canceled_at: Time.current
+          )
+        end
+
         current_subscription
       end
 
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -117,7 +117,7 @@ class Subscription < ApplicationRecord
   end
 
   def next_subscription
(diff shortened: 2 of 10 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -151,12 +151,11 @@ def downgrade_subscription
 
       # NOTE: When downgrading a subscription, we keep the current one active
       #       until the next billing day. The new subscription will become active at this date
-      Subscription.create!(
+      current_subscription.next_subscriptions.create!(
         customer:,
         plan: params.key?(:plan_overrides) ? override_plan(plan) : plan,
         name:,
         external_id: current_subscription.external_id,
-        previous_subscription_id: current_subscription.id,
```
lovrocolic, line 159: why do we remove this association?

floganz: It's now created via `current_subscription.next_subscriptions.create!` which assigns that automatically, compared to `Subscription.create!` previously where we need manually pass it.

app/services/invoices/preview/subscription_plan_change_service.rb
```diff
@@ -35,6 +35,11 @@ def terminated_current_subscription
         current_subscription.terminated_at = termination_date
         current_subscription.status = :terminated
 
+        current_subscription.next_subscription&.assign_attributes(
```
lovrocolic, line 38: I am not sure we need cancelation in preview flow TBH

lovrocolic (approved):

PR #3320, opened 2025-03-11
feat(preauth): Get latest payment method from Stripe if not found locally

julienbourdeau:
## Description

Payments in Lago rely on Stripe to figure out the payment method, we don't pass it explicitly. For payment authorization, we must pass the payment method explicitly.

When a payment method is added to a stripe user, we receive a `setup_intent.succeeded` and save this payment method in Lago. 

It's possible that when creating a new subscription and making a pre-auth call, the webhook wasn't process yet. In this case, we'll try to find a payment method directly on Stripe.

First, we try the default payment.
Then, the most recent card.
Otherwise, fail.

### DIFF
--- a/app/services/payment_provider_customers/stripe/retrieve_latest_payment_method_service.rb
+++ b/app/services/payment_provider_customers/stripe/retrieve_latest_payment_method_service.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module PaymentProviderCustomers
+  module Stripe
+    class RetrieveLatestPaymentMethodService < BaseService
+      Result = BaseResult[:payment_method_id]
+
+      def initialize(provider_customer:)
+        @provider_customer = provider_customer
+        super
+      end
+
+      def call
+        # We use limit: 10 just in case for some (wrong) reason the customer has a very high number of payment method
+        list = ::Stripe::Customer.list_payment_methods(provider_customer.provider_customer_id, {limit: 10}, api_key:)
+        result.payment_method_id = list.data.filter { _1.type == "card" }.max_by { _1.created }.id
+        result
+      end
+
+      private
+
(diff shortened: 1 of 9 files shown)


app/services/payment_provider_customers/stripe/retrieve_latest_payment_method_service.rb
```diff
@@ -0,0 +1,53 @@
...
+        @provider_customer = provider_customer
+        super
+      end
+
+      def call
+        # First, we try to get the customer default payment method
+        payment_method_id = begin
+          customer = ::Stripe::Customer.retrieve(provider_customer.provider_customer_id, request_options)
+          customer["invoice_settings"]["default_payment_method"]
+        rescue
```
vincent-pochet, line 18: I get why we are swallowing the errors, but maybe you could add a message so that it will remain clear for future readers :)

app/services/payment_provider_customers/stripe/retrieve_latest_payment_method_service.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+
+      private
+
+      attr_reader :provider_customer
+
+      def request_options
+        {
+          api_key:,
+          stripe_version: "2024-09-30.acacia"
```
vincent-pochet, line 44: Should'nt it be a constant?

vincent-pochet (approved):

app/services/payment_provider_customers/stripe/retrieve_latest_payment_method_service.rb
julienbourdeau: It should be! Ideally it would be the constant added in this PR but I'd like to upgrade our version before I fix it.
I'm using the version of the customer testing this beta feature.

PR #3321, opened 2025-03-11
misc(refactor): Refactor Customers::CreateService

ancorcruz:
## Context

We are going to need to change services signature in order to include billing entity information on customer creation, then we are taking the opportunity to refactor this old service following the new style and breaking it into two separate services prior to multi entities development.

## Description

Refactor `Customers::CreateService` breaking it in two services with the standard interface (`.call`)...

* `Customers::CreateService` remains for graphQL flow
* `Customers::CreateFromApiService` with the code required for the credate/update Api flow.


vincent-pochet (approved):
:rocket:

PR #3322, opened 2025-03-11
feat(multi-entities): Update customers create, update and upsert services to require billing entity code param

ancorcruz:
## Roadmap Task

  ## Context

  ## Description

On customer creation, edition or upsert (create/update from rest API request) assign or update its billing entity reference.

### DIFF
--- a/app/services/customers/update_service.rb
+++ b/app/services/customers/update_service.rb
@@ -98,7 +98,7 @@ module Customers
       #       billing_entity_id
       if customer.editable?
         customer.external_id = args[:external_id] if args.key?(:external_id)
-        customer.billing_entity = billing_entity if billing_entity
+        customer.billing_entity = billing_entity if args.key?(:billing_entity_code)
 
         if organization.revenue_share_enabled?
           customer.account_type = args[:account_type] if args.key?(:account_type)
@@ -195,9 +195,7 @@ module Customers
     def_delegators :customer, :organization
 
     def billing_entity
-      return unless args[:billing_entity_id]
-
-      @billing_entity ||= organization.billing_entities.find(args[:billing_entity_id])
+      @billing_entity ||= organization.billing_entities.active.find_by!(code: args[:billing_entity_code])
     end
 
     def valid_metadata_count?(metadata:)
--- a/app/services/customers/update_service.rb
+++ b/app/services/customers/update_service.rb
@@ -195,7 +195,7 @@ module Customers
     def_delegators :customer, :organization
 
     def billing_entity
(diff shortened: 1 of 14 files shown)


app/services/customers/update_service.rb
```diff
@@ -190,6 +194,10 @@ def call
     attr_reader :customer, :args
     def_delegators :customer, :organization
 
+    def billing_entity
+      @billing_entity ||= organization.billing_entities.active.find_by!(code: args[:billing_entity_code])
```
annvelents, line 198: I'm thinking: should we update billing_entity relation on organization, to not have to specify `active` each time... because usually we DO have only active...
```
  has_many :billing_entities, -> { where(archived_at: nil) }
  has_many :all_billing_entities, class_name: "BillingEntity"
```

app/models/billing_entity.rb
```diff
@@ -40,6 +40,7 @@ class BillingEntity < ApplicationRecord
   enum :document_numbering, DOCUMENT_NUMBERINGS
 
   default_scope -> { kept }
+  scope :active, -> { kept.where(archived_at: nil) }
```
annvelents, line 43: I think we don't need `kept` here

app/services/customers/upsert_from_api_service.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 module Customers
-  class CreateFromApiService < BaseService
+  # Upsert (creates or updates) a customer from an API request
+  class UpsertFromApiService < BaseService
```
annvelents, line 5: 🚀 was thinking to suggest it when looking at your previous PR

app/services/customers/upsert_from_api_service.rb
```diff
@@ -145,7 +151,7 @@ def call
 
     private
 
-    attr_reader :organization, :params
+    attr_reader :billing_entity, :organization, :params
```
annvelents, line 154: shall we define belling_entity as an instance variable?
and also we can then define it on init...

spec/services/customers/create_service_spec.rb
```diff
@@ -62,6 +63,29 @@
     expect(SendWebhookJob).to have_received(:perform_later).with("customer.created", result.customer)
   end
 
+  context "when organization has multiple billing entities" do
```
annvelents, line 66: this scenario shouldn't happen, but we have this error raise, what do you think about adding a scenario when an organization has no active billing_entities
or better:
a code for discarded or archived billing_entity is provided

ancorcruz: done added for create, update and upsert from api

annvelents (approved):
LGTM 🚀

PR #3285, opened 2025-03-03
fix(boundaries): Ensure correct boundaries for in_advance_charge_periodic invoices

vincent-pochet:
## Context

Include relevant motivation and context.

## Description

Describe your changes in detail.

List any dependencies that are required.

PR #3327, opened 2025-03-12
feat(regroup_fee): Use current month boundaries for AdvanceCharges invoices

julienbourdeau:
## Context

`advance_charges` invoices will regroup all **non invoiceable fees paid within the last period**. 

Invoice billing periods are normally related to the fee periods but in this case, you can have a fee created in June paid only in November so we want to recompute the periods.

The second issue is that the `DateService` is designed to work active or "recently terminated" subscription. In this case we try to compute dates for a fees long after the subscription is terminated which results in wrong dates (`to` date being before `from` date for instance). This bug is NOT fixed. Even if it was, the Date service would still compute date related to the subscription when the fee is created which is **not what we want in this specific case**. 

## Description

This fixed is inspired by this first attempt by @vincent-pochet : https://github.com/getlago/lago-api/pull/3285

Note that this invoice only has charge fees by design, so subscription fee.

For `advance_charges` and only for these invoices, I created a new service to create the `InvoiceSubscription` model. We grab the most recent subscription to compute correct boundaries and apply them to all InvoiceSubscriptions.

This invoice is special! See other corner case recently fixed: https://github.com/getlago/lago-api/pull/3153

### DIFF
--- a/app/services/invoices/create_advance_charges_invoice_subscription_service.rb
+++ b/app/services/invoices/create_advance_charges_invoice_subscription_service.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Invoices
+  class CreateAdvanceChargesInvoiceSubscriptionService < BaseService
+    Result = BaseResult
+
+    def initialize(invoice:, timestamp:, subscriptions:)
+      @invoice = invoice
+      @timestamp = timestamp
+      @subscriptions = subscriptions
+
+      super
+    end
+
+    # Since the `advance_charges` invoice only have charges by design,
+    # we apply the `charges_(from|to)_date for both charges and subscriptions period
+    # See https://github.com/getlago/lago-api/pull/3327 for details
+    def call
+      latest_subscription = subscriptions.max_by(&:started_at)
+      boundaries = calculate_boundaries(latest_subscription)
+
+      subscriptions.each do |subscription|
+        invoice.invoice_subscriptions << InvoiceSubscription.create!(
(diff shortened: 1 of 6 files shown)


app/services/invoices/create_advance_charges_invoice_subscription_service.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      result
+    end
+
+    private
+
+    attr_reader :invoice, :timestamp, :subscriptions
+
+    def calculate_boundaries(subscription)
+      date_service = Subscriptions::DatesService.new_instance(subscription, timestamp, current_usage: false)
```
vincent-pochet, line 44: One, thing you should check with this is what happen if the subscription is terminated? 
It's a common case after an upgrade, for example. One of the issue we had was that we ended-up with weird boundaries as the date service is not designed to compute values after the termination date

vincent-pochet (approved):
Looks great. Just one question regarding terminated subscriptions

app/services/invoices/create_advance_charges_invoice_subscription_service.rb
julienbourdeau: If the last subscription is terminated it will trigger billing, so we're fine.

But if the **last subscription** was terminated long time ago and suddenly some fees are marked as paid, they will never be added to any invoice because. There are no event to trigger billing in this case.
This could happen if the `subscription.external_id` changed for instance.

PR #3331, opened 2025-03-13
Reapply: Add in advance charge consumer, to replace sidekiq hack (#3325)

vincent-pochet:
This reverts commit bd2391e6a650ba9d3a562f20fd6eb24b0f130037 and reapply https://github.com/getlago/lago-api/commit/ff8e73976cce45174e2bf8703e9b426a6f0c8f73 with fixes

The `Karafka::Web.enable!` was called for every application, no matter the presence of the correct configuration. This PR ensures it is turned on only when expected


lib/karafka/lago_monitor.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Karafka::LagoMonitor < ::Karafka::Instrumentation::Monitor
+  TRACEABLE_EVENTS = %w[
+    consumer.consumed
+  ].freeze
+
+  def instrument(event_id, payload = EMPTY_HASH, &block)
```
julienbourdeau, line 8: Where is `EMPTY_HASH` defined? I don't see it in our codebase so I assume we're using it from another gem. We should probably have one. I needed it for permissions too

https://github.com/getlago/lago-api/blob/bc3ebfc462b9e2d6e7dbf2bba5b7da2dce9cad29/app/models/permission.rb#L10

lib/karafka/lago_monitor.rb
```diff
@@ -0,0 +1,13 @@
...
+    return super unless TRACEABLE_EVENTS.include?(event_id)
```
julienbourdeau, line 9: Probably not important here because there is only one item but I general for this kind of test we should rely on a hash.

```ruby
TRACEABLE_EVENTS = {
  "consumer.consumed" => true
}.freeze

if TRACEABLE_EVENTS[event_id]
  # ...
end

```

karafka.rb
```diff
@@ -26,37 +26,38 @@ class KarafkaApp < Karafka::App
...
-
 Karafka::Process.tags.add(:application_name, "lago-api")
+
+Karafka::Web.setup do |config|
+  # Set this to false in all apps except one
+  config.processing.active = ENV["LAGO_KARAFKA_PROCESSING"] if ENV["LAGO_KARAFKA_PROCESSING"].present?
+  config.ui.sessions.secret = ENV["LAGO_KARAFKA_WEB_SECRET"] if ENV["LAGO_KARAFKA_WEB_SECRET"].present?
+end
+
+Karafka::Web.enable! if ENV["LAGO_KARAFKA_WEB"].present?
```
julienbourdeau, line 63: This entire file is called automatically by the gem, right? There is nothing explicite in our config/initializers. Just want to be sure I understand.

julienbourdeau (approved):

karafka.rb
vincent-pochet: Yes, running `exec bundle exec karafka server` will call this file

lib/karafka/lago_monitor.rb
vincent-pochet: It's part of the gem (https://github.com/karafka/karafka/blob/3fd32d63b2f2d78321055c3e8b2b22d0e7ca3fbd/lib/karafka/processing/inline_insights/tracker.rb#L37)
I just used an example from the Karafka documentation

lib/karafka/lago_monitor.rb
vincent-pochet: It was also copied from the doc. For simple list like this one, I  will keep the current implementation (just a matter of personal preference to be honest :D).
I'm not even sure we will monitor other kind of events in a neat future...

PR #3336, opened 2025-03-14
Fix units rounding

ivannovosad:
## Description

Fix units rounding in invoice templates - round to 6 significant digits.


app/views/helpers/rounding_helper.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class RoundingHelper
+  def self.round_decimal_part(num, decimal_sig_figs = 6)
```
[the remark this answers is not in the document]
nudded: (see `MoneyHelper.format_with_precision`)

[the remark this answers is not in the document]
nudded:
```
amount_cents = if amount_cents < 1
      BigDecimal("%.6g" % amount_cents)
    else
      amount_cents.round(6)
    end
```
 in case you only need significant digits when lower than 1.

app/views/helpers/rounding_helper.rb
```diff
@@ -0,0 +1,12 @@
...
+    return BigDecimal("%.#{decimal_sig_figs}g" % num).to_s if num.abs < 1
+
+    rounded_decimal_part = BigDecimal(num.to_s).frac.round(decimal_sig_figs)
```
nudded, line 7: can you comment on why this is required? 🤔  this feels very convoluted

ivannovosad: If `rounded_decimal_part` is zero, we need to call `rounded_number.to_i.to_s` instead of just `rounded_number.to_s`.

In this example (which would fail otherwise):

```
expect(helper.round_decimal_part(123.000000000456789)).to eq("123")
```

we want the result to equal `123` instead of `123.0`

nudded: Okay, I would argue that the following code is cleaner:

```
rounded = BigDecimal(num.to_s).round(decimal_sig_figs)
rounded.frac.zero? ? rounded.to_i.to_s : rounded.to_s
```

(e.g. I find the rounding of the fractional part confusing, but maybe that's just me)

nudded (approved):

app/views/helpers/rounding_helper.rb
ivannovosad: I pushed the changes

PR #3338, opened 2025-03-14
Fix/wallet ballance with progressive billing

annvelents:
We had a bug calculating wallet balance when progressive_billing threshold was reached:
To reproduce: 

-Create a plan with progressive billing threshold ($7k)

-Purchase wallet credits ($7600)

-Send usage event to go above the PB threshold ($8k)

-Receive invoice (wallet credits are subtracted, so I owe $400)

-Check the wallet ongoing balance (shows -$8k): this is the issue, it should say 0 here

image

after fixing this bug, so that progressive_billing_invoices behave the same way as pay_in_advance charges, we noticed weird behaviour of these charges:

they are excluded from ongoing ballance, but are still shown in the current usage:
image
image

Fixed to not include billed usage in the ongoing balance:
image

And the last thing: when calculating current usage, taxes that are already included in pay_in_advance invoices were handled a way that they were wrongly affecting ongoing balance:
image
(on the screenshot the usage is 30$ + 10% tax, so when they were wrongly calculated we were showing not existing usage)

with fix:
image


app/services/invoices/customer_usage_service.rb
```diff
@@ -171,6 +173,26 @@ def compute_amounts_with_provider_taxes
       invoice.total_amount_cents = invoice.fees_amount_cents + invoice.taxes_amount_cents
     end
 
+    # Progressively billed invoices ae created in the subscription boundaries, but because they
+    # do not have paid_in_advance charges, we can not substitute them in the usage (as paid_in_advance_fees).
+    # We need to add the credits that were progressively billed within the same boundaries
+    def add_progressively_billed_usage
```
nudded, line 179: I think you should use the `Subscriptions::ProgressiveBilledAmount` service.

(I think the logic here is wrong when there's 2 progressive invoices in the past month, e.g it should only be the last one taken into account. But the aforementioned service takes care of everything. )

vincent-pochet (approved):

PR #3341, opened 2025-03-17
misc(sentry): Add headers to sentry

ivannovosad:
## Description

This PR adds headers to Sentry which will help us to debug requests made to external integrations services.

### DIFF
--- a/app/services/data_api/mrrs/plans_service.rb
+++ b/app/services/data_api/mrrs/plans_service.rb
@@ -17,7 +17,7 @@ module DataApi
       private
 
       def action_path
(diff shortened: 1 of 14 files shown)


config/application.rb
```diff
@@ -15,7 +15,7 @@ class Application < Rails::Application
 
     # TODO: Should be turned to false
     config.add_autoload_paths_to_load_path = true
-    # config.autoload_lib(ignore: %w[task])
```
lovrocolic, line 18: should we just add `middleware` without removing `task`?

PR #3343, opened 2025-03-17
fix(tax-provider): do not call tax provider when invoice is zero amount or without fees

lovrocolic:
## Context

When invoice does not have any fee attached we don't want to make a request to the tax provider. In addition, if invoice `fees_amount_cents` is zero but customer property for finalizing zero amount invoices is set to `skip`, we also don't want to sync such invoice to Anrok.

## Description

In described scenario, expected flow should include using Lago defined taxes, which will anyway result in zero amounts even if Lago taxes are defined.

However, tax provider users usually does not define Lago taxes.

### DIFF
--- a/app/services/invoices/compute_amounts_from_fees.rb
+++ b/app/services/invoices/compute_amounts_from_fees.rb
@@ -12,7 +12,7 @@ module Invoices
     def call
       if should_apply_fee_taxes?
         invoice.fees.each do |fee|
-          taxes_result = if provider_taxes && customer_provider_taxation?
+          taxes_result = if provider_taxes && customer_provider_taxation? && invoice.should_apply_provider_tax?
             Fees::ApplyProviderTaxesService.call(fee:, fee_taxes: fee_taxes(fee))
           else
             Fees::ApplyTaxesService.call(fee:)
--- a/app/services/invoices/compute_amounts_from_fees.rb
+++ b/app/services/invoices/compute_amounts_from_fees.rb
@@ -29,7 +29,7 @@ module Invoices
         invoice.fees_amount_cents - invoice.progressive_billing_credit_amount_cents - invoice.coupons_amount_cents
       )
 
-      taxes_result = if customer_provider_taxation?
+      taxes_result = if customer_provider_taxation? && invoice.fees.any?
         Invoices::ApplyProviderTaxesService.call(invoice:, provider_taxes:)
       else
         Invoices::ApplyTaxesService.call(invoice:)
--- a/app/services/invoices/compute_amounts_from_fees.rb
+++ b/app/services/invoices/compute_amounts_from_fees.rb
@@ -29,7 +29,7 @@ module Invoices
         invoice.fees_amount_cents - invoice.progressive_billing_credit_amount_cents - invoice.coupons_amount_cents
       )
 
-      taxes_result = if customer_provider_taxation? && invoice.fees.any?
+      taxes_result = if customer_provider_taxation? && invoice.should_apply_provider_tax?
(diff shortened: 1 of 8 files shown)


app/services/invoices/compute_amounts_from_fees.rb
```diff
@@ -29,7 +29,7 @@ def call
         invoice.fees_amount_cents - invoice.progressive_billing_credit_amount_cents - invoice.coupons_amount_cents
       )
 
-      taxes_result = if customer_provider_taxation?
+      taxes_result = if customer_provider_taxation? && invoice.fees.any?
```
nudded, line 32: is the fall through here logical? e.g. we'd call `Invoices::ApplyTaxesService` when there are no fees?

lovrocolic: I see the point but the goal of this PR was just to skip fetching taxes for `invoice.fees = []` case. In that case, we want to use the flow where all other invoices end-up and where Lago taxes are used (to avoid Anrok to raise validation error for invoices without fees).
This PR ws aiming to fix only this case and not to change any logic for invoices without fees in general.

nudded: Okay, just double checking, I think that is good enough 👍

nudded (approved):

spec/models/invoice_spec.rb
```diff
@@ -1765,4 +1765,45 @@
       end
     end
   end
+
+  describe "should_apply_provider_tax?" do
+    let(:invoice) { create(:invoice, :subscription, subscriptions: [subscription]) }
+    let(:plan) { create(:plan) }
+    let(:subscription) { create(:subscription, plan:) }
+
+    let(:charge) { create(:standard_charge, plan:) }
```
floganz, line 1774:
```suggestion
    let(:invoice) { create(:invoice, :subscription) }
    let(:charge) { create(:standard_charge, plan: invoice.subscriptions.first.plan) }
```

Should be the same.

spec/models/invoice_spec.rb
```diff
@@ -1765,4 +1765,45 @@
...
+    before { charge_fee }
+
+    context "when there are fees with non zero amount" do
+      it { expect(invoice).to be_should_apply_provider_tax }
+    end
+
+    context "when there are no fees" do
+      let(:charge_fee) { nil }
+
+      it { expect(invoice).not_to be_should_apply_provider_tax }
```
floganz, line 1786: A matter of taste but maybe you'll like it also:
Add `subject { invoice.should_apply_provider_tax? }` right after `describe`.

Then change it to such format.
```suggestion
      it { expect(subject).to be false }
```

or `it { expect(subject).to be true }`.

If at any moment we'll decide to rename this method we'd only need to change it once in `subject` instead of all tests.

spec/models/invoice_spec.rb
```diff
@@ -1765,4 +1765,45 @@
...
+    let(:plan) { create(:plan) }
+    let(:subscription) { create(:subscription, plan:) }
+
+    let(:charge) { create(:standard_charge, plan:) }
+    let(:charge_fee) { create(:charge_fee, charge:, invoice:, amount: 100) }
+
+    before { charge_fee }
+
+    context "when there are fees with non zero amount" do
+      it { expect(invoice).to be_should_apply_provider_tax }
```
floganz, line 1780: Another a matter of taste thing: keep contexts symmetry.

```suggestion
    before { charge_fee }

    context "when there are fees with non zero amount" do
      let(:charge_fee) { create(:charge_fee, charge:, invoice:, amount: 100) }

      it { expect(invoice).to be_should_apply_provider_tax }
```

floganz (approved):

PR #3344, opened 2025-03-17
misc(http-client): Add response headers to http error

ivannovosad:
## Description

This PR adds response headers to `LagoHttpClient::HttpError` which will help us to debug requests made to external integrations services.

### DIFF
--- a/lib/lago_http_client/lago_http_client/client.rb
+++ b/lib/lago_http_client/lago_http_client/client.rb
@@ -114,7 +114,7 @@ module LagoHttpClient
     attr_reader :http_client
 
     def raise_error(response)
-      raise(::LagoHttpClient::HttpError.new(response.code, response.body, uri))
+      raise(::LagoHttpClient::HttpError.new(response.code, response.body, uri, response_headers: response.to_hash))
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/lago_http_client/lago_http_client/client.rb
```diff
@@ -114,7 +114,7 @@ def get(headers: {}, params: nil, body: nil)
     attr_reader :http_client
 
     def raise_error(response)
-      raise(::LagoHttpClient::HttpError.new(response.code, response.body, uri))
+      raise(::LagoHttpClient::HttpError.new(response.code, response.body, uri, response_headers: response.to_hash))
```
lovrocolic, line 117: can't we just fetch headers? I guess we are sending whole object as headers...

lovrocolic (approved):

lib/lago_http_client/lago_http_client/client.rb
ivannovosad: I did an improvement, but I'm not sure what exactly you mean by fetching just headers?

lovrocolic (approved):

PR #3348, opened 2025-03-18
feat(pdf): Add env var to disable generating pdfs

ivannovosad:
## Context

Some clients don't need pdfs of invoices, credit notes and payment receipts.

## Description

This PR adds and ENV var to disable generation pdfs: `LAGO_DISABLE_PDF_GENERATION`

### DIFF
--- a/app/services/credit_notes/generate_service.rb
+++ b/app/services/credit_notes/generate_service.rb
@@ -12,9 +12,10 @@ module CreditNotes
     def call
       return result.not_found_failure!(resource: "credit_note") if credit_note.blank? || !credit_note.finalized?
 
-      generate_pdf(credit_note) if should_generate_pdf?
-
-      SendWebhookJob.perform_later("credit_note.generated", credit_note)
+      if should_generate_pdf?
+        generate_pdf(credit_note)
+        SendWebhookJob.perform_later("credit_note.generated", credit_note)
+      end
 
       result.credit_note = credit_note
       result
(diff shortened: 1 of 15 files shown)


app/services/credit_notes/generate_service.rb
```diff
@@ -12,9 +12,10 @@ def initialize(credit_note:, context: nil)
     def call
       return result.not_found_failure!(resource: "credit_note") if credit_note.blank? || !credit_note.finalized?
 
-      generate_pdf(credit_note) if should_generate_pdf?
-
-      SendWebhookJob.perform_later("credit_note.generated", credit_note)
+      if should_generate_pdf?
+        generate_pdf(credit_note)
+        SendWebhookJob.perform_later("credit_note.generated", credit_note)
```
nudded, line 17: This webhook should be sent regardless of the `should_generate_pdf?` no?

vincent-pochet (approved):

app/services/credit_notes/generate_service.rb
julienbourdeau: this webhook is really about the PDF generation. the only difference between `credit_note.generated` and `credit_note.created` is that the generated has the link to the PDF so I think we shouldn't send it.

julienbourdeau (commented):
I wonder if this should be a settings per organization, similar to sending invoice by email.

The customer requesting it is self hosting so this solution works for them but if it was per organization other customer who don't need our PDF could use it (helping our infra).

app/services/credit_notes/generate_service.rb
ivannovosad: I'd say not send it as well, what's your thoughts @vincent-pochet @brunomiguelpinto ?

nudded: ah, this is really a weird webhook name :D I forgot we also have `created`. Fine to skip in that case!

PR #3349, opened 2025-03-18
Misc dump sql

ivannovosad:
## Description

This PR handles the switch from `schema.rb` dump to `structure.sql` which is needed for storing procedures, triggers, etc.


vincent-pochet (approved):

Dockerfile.dev
```diff
@@ -3,7 +3,7 @@ FROM ruby:3.3.6-slim
 WORKDIR /app
 
 RUN apt update -qq && \
-  apt install nodejs build-essential curl git pkg-config libpq-dev libclang-dev -y && \
+  apt install nodejs build-essential curl git pkg-config libpq-dev libclang-dev postgresql-client -y && \
```
nudded, line 6: I think, it's good practice to be explicit about the version of pg-dump we're installing (out of experience, there can be incompatibility if you're using pg-dump versions of newer postgresql versions) and match it to the version of PG we're using locally.

julienbourdeau (approved):

PR #3350, opened 2025-03-18
chore(results): Introduce Result classes for app/queries folder

julienbourdeau:
## Context

We're migrating away from `OpenStruct`.

## Description

This adds a custom `Result` class for all Query classes in `app/queries`. By design, it's kinda expected that there is only one thing returned in the result. 

Although, I'm surprised to see the pagination returned only in `PastUsageQuery`. 🤔

### DIFF
--- a/app/queries/add_ons_query.rb
+++ b/app/queries/add_ons_query.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AddOnsQuery < BaseQuery
+  Result = BaseResult[:add_ons]
+
   def call
     add_ons = base_scope.result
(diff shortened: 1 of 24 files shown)


vincent-pochet (approved):

vincent-pochet:
@julienbourdeau all the collection are paginated via kaminari, so the pagination properties are attached directly to the result attributes. See https://github.com/getlago/lago-api/blob/main/app/controllers/concerns/pagination.rb#L12

julienbourdeau:
Great! I didn't take the time to dig deeper but it makes sense. I see now why `PastUsageQuery` is different.

PR #3351, opened 2025-03-18
Fix(wallet-balance): Fail if failed to fetch full current usage while updating wallet

annvelents:
This PR fixes a bug:
- when updating wallet balance
- we fail to fetch taxes
- result of `Invoices::CustomerUsageService` is not checked if it's success
- we update wallet without correctly calculated current usage invoice (total_amount_cents in this case is 0)

After the fix:
- when updating wallet balance
- we fail to fetch taxes
- we return a failed result from the service `Wallets::Balance::RefreshOngoingService`
- Job `Wallets::RefreshOngoingBalance` raises an exception

### DIFF
--- a/app/jobs/wallets/refresh_ongoing_balance_job.rb
+++ b/app/jobs/wallets/refresh_ongoing_balance_job.rb
@@ -11,7 +11,7 @@ module Wallets
     def perform(wallet)
       return unless wallet.ready_to_be_refreshed?
 
-      Wallets::Balance::RefreshOngoingService.call(wallet:)
+      Wallets::Balance::RefreshOngoingService.call!(wallet:)
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/jobs/wallets/refresh_ongoing_balance_job.rb
```diff
@@ -11,7 +11,7 @@ class RefreshOngoingBalanceJob < ApplicationJob
     def perform(wallet)
       return unless wallet.ready_to_be_refreshed?
 
-      Wallets::Balance::RefreshOngoingService.call(wallet:)
+      Wallets::Balance::RefreshOngoingService.call(wallet:).raise_if_error!
```
vincent-pochet, line 14:
```suggestion
      Wallets::Balance::RefreshOngoingService.call!(wallet:)
```

vincent-pochet (approved):

julienbourdeau (approved):

PR #3354, opened 2025-03-19
feat(dunning): Add BCC emails addresses

julienbourdeau:
## Description

When creating a dunning campaign, you can now add a list of email addresses to pass in the `bcc` field. This can help accounting team keeping track of payment requests.

In GraphQL, the field is an array of string, each string is a valid email address. It's NOT a single string containing  comma-separated email addresses (like we might do in other places.

The dunning campaign is attached to the payment request so I only had to check it in the Mailer. The rest is basically model + gql boilerplate.


app/validators/email_array_validator.rb
```diff
@@ -0,0 +1,22 @@
...
+class EmailArrayValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    unless value.is_a? Array
+      record.errors.add(attribute, "must be an array")
+      return
+    end
+
+    value.each_with_index do |email, index|
+      unless valid? email
+        record.errors.add(attribute, "value #{email} at position #{index} is not a valid email address")
```
vincent-pochet, line 12: It might be a bit more complex to add an error code here :thinking: 

maybe `invalid_email_format["#{index}"]: "#{email}"` ?

vincent-pochet (approved):

PR #3356, opened 2025-03-19
refactor(eu_vat): store json in class const (so it's only read once at boot)

julienbourdeau:
## Description

As I was updating the rates (see #3352) I noticed we could improve the `LagoEuVat::Rate` class.

Instead of reading the json file every time we initialize the class, we now load it once when Rails is eager loading at boot.

The 2 methods are now class methods too.


ancorcruz (approved):
LGTM 🚀

vincent-pochet (approved):

PR #3358, opened 2025-03-19
misc: Upgrade Rails version to 8.0.X

ancorcruz:
Requires upgrade `clickhouse-activerecord` dependency to 1.3.x


spec/services/events/calculate_expression_service_spec.rb
```diff
@@ -7,7 +7,7 @@
     subject(:service_call) { described_class.new(organization: organization, event: event).call }
 
     let(:organization) { create(:organization) }
-    let(:event) { create(:event, organization: organization, timestamp: Time.current.to_i, code: code, properties: properties) }
```
julienbourdeau, line 10: Any idea what's going on here? Is it related to the upgrade of the gem?

PR #3363, opened 2025-03-21
Fix(multientities): ensure seeded organization is created with billing_entity

annvelents:
Because in the seeds we create an Organization directly with ActiveRecord, not through the service, we have to manually create a BillingEntity for it


julienbourdeau (approved):
We should probably use the service "one day" 😄

annvelents:
⬆️  ticket on the refactoring board 🚀

PR #3367, opened 2025-03-21
spec(dunning): Add scenario tests for dunning

julienbourdeau:
This PR adds a new scenario test for dunning campaign.

The interesting part are:
* mocking stripe at the HTTP level 
* simulating the reception of the Stripe webhooks
* testing emails sent
* testing Lago webhook sent by "logging" them rathen than mocking job

The new jobs in `spec/support/jobs/` extend `ActiveJob::Base` instead of `ApplicationJob` because they are NOT jobs from the applications but fake jobs only meant to be run in spec to help mock behavior.

### DIFF
--- a/spec/scenarios/dunning/campaign_v1_spec.rb
+++ b/spec/scenarios/dunning/campaign_v1_spec.rb
@@ -1,0 +1,205 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "Dunning Campaign v1", :scenarios, type: :request do
+  let(:webhook_url) { "https://test.co/lago" }
+  let(:organization) do
+    create(:organization,
+      name: "JC AI",
+      premium_integrations: %w[auto_dunning],
+      email_settings: [],
+      webhook_url:)
+  end
+
+  let(:dunning_campaign) do
+    create(:dunning_campaign, organization:,
+      applied_to_organization: true, max_attempts: 2, days_between_attempts: 2)
+  end
+  let(:dunning_campaign_threshold) do
+    create(:dunning_campaign_threshold, dunning_campaign:, amount_cents: 150_00, currency: "EUR")
+  end
+  let(:stripe_cus_id) { "cus_123456789" }
+  let(:stripe_pm_id) { "pm_123456" }
+
+  let(:stripe_provider) { create(:stripe_provider, organization:) }
+  let(:customer) { create(:customer, organization:, payment_provider: :stripe, payment_provider_code: stripe_provider.code, net_payment_term: 2) }
+  let(:stripe_customer) { create(:stripe_customer, customer:, payment_provider: stripe_provider, provider_customer_id: stripe_cus_id) }
+
+  let(:external_subscription_id) { "sub_overdue-dunning-campaign-v1" }
+  let(:plan) { create(:plan, organization:, pay_in_advance: true, amount_cents: 149_00) }
+  let!(:addon) { create(:add_on, organization:) }
+
+  let(:webhooks_sent) { [] }
+
+  let(:stripe_customer_response) do
+    File.read("spec/fixtures/stripe/customer_retrieve_response.json")
+  end
+  # TODO: this is part of another PR coming soon https://github.com/getlago/lago-api/pull/3345
+  let(:stripe_payment_method_response) do
+    {
+      id: stripe_pm_id,
+      object: "payment_method",
+      allow_redisplay: "always",
+      billing_details: {
+        address: {
+          city: nil,
+          country: "FR",
+          line1: nil,
+          line2: nil,
+          postal_code: nil,
+          state: nil
+        },
+        email: "awdawd@desf.com",
+        name: "Testing Stripe",
+        phone: nil
+      },
+      card: {
+        brand: "visa",
+        checks: {
+          address_line1_check: nil,
+          address_postal_code_check: nil,
+          cvc_check: "pass"
+        },
+        country: "US",
+        display_brand: "visa",
+        exp_month: 12,
+        exp_year: 2028,
+        fingerprint: "8TOiB4cGytYxCweY",
+        funding: "credit",
+        generated_from: nil,
+        last4: "4242",
+        networks: {
+          available: [
+            "visa"
+          ],
+          preferred: nil
+        },
+        regulated_status: "unregulated",
+        three_d_secure_usage: {
+          supported: true
+        },
+        wallet: nil
+      },
+      created: 1741878064,
+      customer: stripe_cus_id,
+      livemode: false,
+      metadata: {},
+      type: "card"
+    }
+  end
+  let(:stripe_payment_intent_response) do
+    File.read("spec/fixtures/stripe/payment_intent_failed_card_declined.json")
+  end
+
+  # TODO: make it a test metadata `:scenarios, type: :request, premium: true`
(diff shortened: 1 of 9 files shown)


spec/scenarios/dunning/campaign_v1_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+          JSON.parse(request_signature.body)
+        end
+
+        Jobs::MockStripeWebhookEventJob.perform_later(
+          organization,
+          request_body_hash,
+          JSON.parse(response.body)
+        )
+      end
+    end
```
julienbourdeau, line 140: Every time a payment intent is created on Stripe, we enqueue en job to simulate the Stripe webhook being received.
It's a new job in `spec/` so we can build mock some data and build the payload.

spec/scenarios/dunning/campaign_v1_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+
+    travel_to(DateTime.new(2025, 1, 7, 10)) do
+      perform_overdue_balance_update
+      expect(customer.invoices.one_off.sole).to be_payment_overdue
+
+      expect(ActionMailer::Base.deliveries.count).to eq(0)
+      perform_dunning
+      # NOTE: Email is sent twice: first synchronously after Stripe response, and then when the webhook is received
+      expect(ActionMailer::Base.deliveries.count).to eq(2)
+      expect(ActionMailer::Base.deliveries.map(&:subject)).to all eq "Your overdue balance from JC AI"
```
julienbourdeau, line 196: Without mocking the webhook, it's almost impossible to notice that.

spec/support/jobs/mock_stripe_webhook_event_job.rb
```diff
@@ -0,0 +1,49 @@
...
+      @organization = organization
+      @request_body = request_body.deep_symbolize_keys
+      @response_body = response_body.deep_symbolize_keys
+
+      @status = @response_body[:error] ? "failed" : "succeeded"
+
+      # NOTE: Update Payment in our database because `PaymentRequests::Payments::StripeService.update_payment_status`
+      #       relies on Payment.find_by(provider_payment_id: stripe_payment.id)
+      lago_payment_id = @request_body[:metadata][:lago_payment_id]
+      Payment.find(lago_payment_id).update!(provider_payment_id: payment_intent_id)
```
julienbourdeau, line 15: Made a lot easier by #3346

spec/scenarios/dunning/campaign_v1_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+
+  before do
+    stub_pdf_generation
+    stripe_customer
+    dunning_campaign_threshold
+
+    stub_request(:post, webhook_url).with do |req|
+      webhooks_sent << JSON.parse(req.body)
+      true
+    end.and_return(status: 200)
```
julienbourdeau, line 112: We log Lago webhook sent so we can verify all the correct webhook were sent.
If we like it, this should be refactored to the ScenarioHelper to be reused.

**Example of expectations**
```ruby
expect(webhooks_sent.map { _1["webhook_type"] }).to eq(%w[
  subscription.started
  invoice.created
  invoice.generated
  invoice.payment_failure
])
```

spec/scenarios/dunning/campaign_v1_spec.rb
ancorcruz: it is expected we send the same email twice or is it a defect?

spec/scenarios/dunning/campaign_v1_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+  let(:stripe_cus_id) { "cus_123456789" }
+  let(:stripe_pm_id) { "pm_123456" }
+
+  let(:stripe_provider) { create(:stripe_provider, organization:) }
+  let(:customer) { create(:customer, organization:, payment_provider: :stripe, payment_provider_code: stripe_provider.code, net_payment_term: 2) }
+  let(:stripe_customer) { create(:stripe_customer, customer:, payment_provider: stripe_provider, provider_customer_id: stripe_cus_id) }
+
+  let(:external_subscription_id) { "sub_overdue-dunning-campaign-v1" }
+  let(:plan) { create(:plan, organization:, pay_in_advance: true, amount_cents: 149_00) }
+  let!(:addon) { create(:add_on, organization:) }
```
ancorcruz, line 31: do we need to create the add-on with the bang (!)... or wait until it is referenced... anyway, for `dunning_campaign_threshold` it is created in the before block... I would prefer to use one approach or the other but mixing both could be confusing overtime.

spec/scenarios/dunning/campaign_v1_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+        Jobs::MockStripeWebhookEventJob.perform_later(
+          organization,
+          request_body_hash,
+          JSON.parse(response.body)
+        )
+      end
+    end
+  end
+
+  it do
```
ancorcruz, line 143: could we have a description for the scenario?

spec/scenarios/dunning/campaign_v1_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+    end
+
+    # The next day nothing happens
+    travel_to(DateTime.new(2025, 1, 10, 10)) do
+      perform_overdue_balance_update
+      perform_dunning
+      expect(ActionMailer::Base.deliveries.count).to eq(4)
+
+      expect(customer.payment_requests.reload.map(&:amount_cents)).to all eq 155_00
+    end
```
ancorcruz, line 223: the comment leads to confusion as something happen after 2 days, another dunning attempt is processed, and new emails are sent. Still not paid.

spec/scenarios/dunning/campaign_v1_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+
+    # This is over
+    travel_to(DateTime.new(2025, 1, 13, 13)) do
+      perform_overdue_balance_update
+      perform_dunning
+      expect(ActionMailer::Base.deliveries.count).to eq(4) # Nothing new
+    end
```
ancorcruz, line 230: it may be worth to change the comment to mention the dunning campaign is over and no new attempt is processed after 2 days from last because campaign is set to max attempts 2... it will make a lot easier to understand the test without having to check the objects, etc.

ancorcruz (approved):
Looks great... left a few comments.

spec/scenarios/dunning/campaign_v1_spec.rb
julienbourdeau: I think it's not expected but we rarely tests with webhook so it's hard to realize. I noticed it as I worked on many stripe feature recently and I wanted to confirm with this test scenario.

spec/scenarios/dunning/campaign_v1_spec.rb
julienbourdeau: I can rewrite the comment if you prefer

spec/scenarios/dunning/campaign_v1_spec.rb
julienbourdeau: Indeed! I moved it to the generation of the one_off invoice because it's not an important piece we reuse 👍

PR #3373, opened 2025-03-24
feat(LifetimeUsage): Add lifetime_usage premium integration 

nudded:
## Description

Some customers want the lifetime usage of a subscription to always be refreshed for all subscriptions. (Even if the plan does not include progressive billing thresholds). This new premium integration enables that.

### DIFF
--- a/app/services/lifetime_usages/flag_refresh_from_invoice_service.rb
+++ b/app/services/lifetime_usages/flag_refresh_from_invoice_service.rb
@@ -9,7 +9,7 @@ module LifetimeUsages
 
     def call
       return result unless invoice.subscription?
-      return result unless has_plan_usage_thresholds?
+      return result unless should_flag_refresh_from_invoice?
 
       result.lifetime_usages = []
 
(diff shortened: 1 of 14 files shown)


app/services/lifetime_usages/flag_refresh_from_invoice_service.rb
```diff
@@ -31,8 +31,8 @@ def call
 
     attr_reader :invoice
 
-    def has_plan_usage_thresholds?
-      invoice.subscriptions.any? { |s| s.plan.usage_thresholds.any? }
+    def should_flag_refresh_from_invoice?
+      invoice.organization.lifetime_usage_enabled? || invoice.subscriptions.any? { |s| s.plan.usage_thresholds.any? }
```
julienbourdeau, line 35: It was already like this and it's not the purpose of this PR so we can discuss it elsewhere but I think there is a double N+1 so how do you feat about this:

```diff
- invoice.subscriptions.any? { |s| s.plan.usage_thresholds.any? }
+ UsageThreshold.where(plan_id: invoice.subscriptions.map(&:plan_id).any?
```

julienbourdeau (approved):

app/services/lifetime_usages/flag_refresh_from_invoice_service.rb
nudded: I don't disagree, but we're likely only going to have a couple of subscriptions per invoice, I don't think it's worth it to optimize this one just yet. (given that the new version is not as readable)

PR #3377, opened 2025-03-25
feat(multi-entities): Organizations' first billing entity matches its id with the organization_id

ancorcruz:
## Roadmap Task

👉 https://getlago.canny.io/feature-requests/p/support-billing-from-multiple-entities

 ## Context

 ## Description

in order to facilitate the migration to multi entities, the first billing entity id for each organization will match the organization id. This will allow us migrate data in a performant way and reduces the risks of migration issues.

Also, a migration is included to:
1. update all billing entity record ids not matching the organization id
2. update all customer's billing_entity_id not matching the organization_id

### DIFF
--- a/db/migrate/20250325145324_assign_customers_to_billing_entities.rb
+++ b/db/migrate/20250325145324_assign_customers_to_billing_entities.rb
@@ -1,13 +1,17 @@
+# frozen_string_literal: true
+
 class AssignCustomersToBillingEntities < ActiveRecord::Migration[7.2]
   def change
     # NOTE: ensure first billing entity has the same id as the organization to ease the migration to multi entities.
     BillingEntity.where("id != organization_id").find_in_batches(batch_size: 1000) do |batch|
-      BillingEntity.where(id: batch.pluck(:id)).update_all("id = organization_id")
+      BillingEntity.where(id: batch.pluck(:id))
+        .update_all("id = organization_id") # rubocop:disable Rails/SkipsModelValidations
     end
 
     # NOTE: Update all customers to have the same billing_entity_id as organization_id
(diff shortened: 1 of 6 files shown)


db/migrate/20250325145324_assign_customers_to_billing_entities.rb
```diff
@@ -0,0 +1,17 @@
...
+class AssignCustomersToBillingEntities < ActiveRecord::Migration[7.2]
+  def change
+    # NOTE: ensure first billing entity has the same id as the organization to ease the migration to multi entities.
+    BillingEntity.where("id != organization_id").find_in_batches(batch_size: 1000) do |batch|
+      BillingEntity.where(id: batch.pluck(:id))
+        .update_all("id = organization_id") # rubocop:disable Rails/SkipsModelValidations
+    end
+
+    # NOTE: Update all customers to have the same billing_entity_id as organization_id
+    Customer.where("billing_entity_id != organization_id").or(Customer.where(billing_entity_id: nil)).find_in_batches(batch_size: 1000) do |batch|
```
annvelents, line 12: I think we can omit 
`.or(Customer.where(billing_entity_id: nil))`, because if billing_entity_id is nil, it will not be equal to organization_id

annvelents (approved):
👏

db/migrate/20250325145324_assign_customers_to_billing_entities.rb
ancorcruz: we need it, SQL follows three-valued logic (true, false, and unknown). The key reason rows where billing_entity_id is NULL and organization_id is not NULL are excluded is that:

* Any comparison involving NULL results in unknown (not true or false).
* The WHERE clause only includes rows where the condition evaluates to true.
* Since NULL != b evaluates to unknown, those rows are filtered out.

PR #3381, opened 2025-03-26
feat(stripe): Update payment method in a job

julienbourdeau:
## Description

I was hesitant to make it async but it's better for retries.

Note that `update_payment_status` might raise a not found error if the webhook arrived before the `payment._provider_payment_id` is updated in the DB, which will retry the HandleEventJob. 

So when `UpdatePaymentMethodDataJob` is dispatched, we're guaranteed that payment is updated... until the implementation of `update_payment_status` changes 😄

### DIFF
--- a/spec/jobs/payments/update_payment_method_data_job_spec.rb
+++ b/spec/jobs/payments/update_payment_method_data_job_spec.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Payments::UpdatePaymentMethodDataJob, type: :job do
+  let(:provider_payment_id) { "pi_123" }
+  let(:provider_payment_method_id) { "pm_001" }
+
+  it "calls the create service" do
+    create(:payment, provider_payment_id:)
+
+    allow(Payments::UpdatePaymentMethodDataService)
+      .to receive(:call!).with(payment: Payment, provider_payment_method_id:).and_return(BaseService::Result.new)
+
+    described_class.perform_now(provider_payment_id: "pi_123", provider_payment_method_id:)
+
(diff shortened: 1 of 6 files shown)


spec/jobs/payments/update_payment_method_data_job_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  it "calls the create service" do
+    create(:payment, provider_payment_id:)
+
+    allow(Payments::UpdatePaymentMethodDataService)
+      .to receive(:call!).with(payment: Payment, provider_payment_method_id:).and_return(BaseService::Result.new)
+
+    described_class.perform_now(provider_payment_id: "pi_123", provider_payment_method_id:)
+
+    expect(Payments::UpdatePaymentMethodDataService).to have_received(:call!)
+  end
```
ivannovosad, line 18: Can we add a context when payment is not found?

ivannovosad (dismissed):

jdenquin:
> ## Description
> I was hesitant to make it async but it's better for retries.
> 
> Note that `update_payment_status` might raise a not found error if the webhook arrived before the `payment._provider_payment_id` is updated in the DB, which will retry the HandleEventJob.
> 
> So when `UpdatePaymentMethodDataJob` is dispatched, we're guaranteed that payment is updated... until the implementation of `update_payment_status` changes 😄

With a job you can still have the payment not updated, it may be rare but it can still happen no?

julienbourdeau:
@jdenquin At this point it shouldn't happen because the code right before enqueuing this job will already fail if the payment is not updated.

jdenquin (changes requested):
Can we discuss about this one?

jdenquin (approved):

PR #3386, opened 2025-03-27
tests: improve GraphQLHelper to set CurrentContext.source to graphql

nudded:


vincent-pochet (approved):

julienbourdeau:
I believe we should reset the context after each rspec example 🤔 I'll check it out

PR #3394, opened 2025-03-27
feat(multientities): Assign billing entity sequential id to invoices

ancorcruz:
## Roadmap Task

## Context

## Description

Fix the invoices index on `billing_entity_sequential_id` to ensure its uniqueness of scoped to the `billing_entity_id`.

Also updates all existing invoices `billing_entity_sequential_id` to match `organization_sequential_id` value or `nil` when this one is `zero`.

The approach used to assign the `billing_entity_sequential_id` on invoices is:

- assign the sequential id to all invoices only when the numbering is set to per_billing_entity. This strategy reduced the database lock to the minimum needed.

- the sequential id is assigned as a before save callback, I have tried to use the AASM event callbacks but looks like we do not use AASM to finalize invoices but we assign the invoice status manually or through the rails enum magic methods.

- skip the sequential id if it is already assigned.

- skip the sequential id if invoice is self billed (for partner customers).

- Then, as we could not rely on the `organization_sequential_id` for existing invoices (if numbering strategy is `"per_customer"` a `zero` is assigned). We fix data within a migration to ensure if the billing_entity has per_billing_entity document numbering the latest finalized invoice will get the right billing_entity_sequential_id.

- When changing document numbering strategy from per_customer to per_billing_entity we count the number of non self billed invoices belonging to the billing entity with document number assigned, and update the billing_entity_sequential_id with the count.

This has been tested in local with concurrent invoice finalization to ensure we introduce the minimal impact to the database and concurrent performance.

Also, this change introduces a `finalized_at` timestamp on invoices table. This should make support investigations and future implementation changes easier as we could workout the actual invoice finalization order whatever scope is needed.


app/models/invoice.rb
```diff
@@ -423,6 +424,30 @@ def ensure_number
...
+    return unless status_changed_to_finalized?
+
+    attempts = 0
+    max_attempts = 10
+
+    begin
+      attempts += 1
+
+      self.class.transaction do
+        last_sequential_id = billing_entity.invoices.non_self_billed.with_generated_number.count
```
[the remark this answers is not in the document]
ancorcruz: DB has unique index on billing_entity_sequential_id, hence, under high concurrency DB will ensure uniqueness.

db/migrate/20250327130155_remove_default_billing_entity_sequential_id_on_invoices.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class RemoveDefaultBillingEntitySequentialIdOnInvoices < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    change_column_default :invoices, :billing_entity_sequential_id, from: 0, to: nil
+
+    Invoice.in_batches(of: 1000) do |batch|
+      batch.update_all( # rubocop:disable Rails/SkipsModelValidations
+        "billing_entity_sequential_id = CASE WHEN organization_sequential_id = 0 THEN NULL ELSE organization_sequential_id END"
```
annvelents, line 11: shouldn't we set billing_entity_sequential_id to 0 in this case?

annvelents: nvm

app/models/invoice.rb
ancorcruz, line 442: count is expensive, use MAX instead...

db/migrate/20250327130155_remove_default_billing_entity_sequential_id_on_invoices.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class RemoveDefaultBillingEntitySequentialIdOnInvoices < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
```
ancorcruz, line 6: ensure we always have an invoice with MAX billing_sequential_id assigned instead of nil if the organization is on per_customer.

app/services/billing_entities/update_invoice_numbering_service.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module BillingEntities
+  class UpdateInvoiceNumberingService < BaseService
```
annvelents, line 4: I'm thinking: since it doesn't save, should it be "ChangeInvoiceNumberingService", or something like this to have less confusion

spec/scenarios/invoices/invoice_numbering_spec.rb
```diff
@@ -851,12 +905,189 @@
...
 
         expect(sequential_ids).to match_array([6, 6, 6])
         expect(organization_sequential_ids).to match_array([11, 12, 0])
+        expect(billing_entity_sequential_ids).to match_array([11, 12, nil])
         expect(numbers).to match_array(%w[ORG-11-202312-011 ORG-11-202312-012 ORG-11-003-006])
       end
     end
   end
+
+  context "with multiple billing entities" do
```
annvelents, line 919: it's a very cool test 😎

annvelents (approved):
👏

PR #3405, opened 2025-03-28
misc(GraphQL): add deletedAt on plans object

ansmonjol:
This attribute is needed for Frontend to build logic on top


app/graphql/types/plans/object.rb
```diff
@@ -33,6 +33,7 @@ class Object < Types::BaseObject
       field :has_subscriptions, Boolean, null: false
 
       field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+      field :deleted_at, GraphQL::Types::ISO8601DateTime, null: true
```
vincent-pochet, line 36: We probably also have a spec file. Could you update it accordingly?

vincent-pochet (approved):

PR #3411, opened 2025-03-31
Feat(multientities): serialize billing_entity_code in API

annvelents:
## Context

Include billing_entity code in:
- invoice serialized object
- customer serialized object
- credit_note serialized object

## Note:
needed to update also `Invoices::PreviewContextService` as customer should always be build with billing entity (the `null: false` on billing_entity_id will be added later)

### DIFF
--- a/app/serializers/v1/billing_entity_serializer.rb
+++ b/app/serializers/v1/billing_entity_serializer.rb
@@ -26,14 +26,6 @@ module V1
         document_number_prefix: model.document_number_prefix,
         tax_identification_number: model.tax_identification_number,
         finalize_zero_amount_invoice: model.finalize_zero_amount_invoice,
-        billing_configuration:
-      }
-    end
-
-    private
-
-    def billing_configuration
-      {
         invoice_footer: model.invoice_footer,
         invoice_grace_period: model.invoice_grace_period,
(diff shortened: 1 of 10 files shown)


app/serializers/v1/billing_entity_serializer.rb
```diff
@@ -26,14 +26,6 @@ def serialize
         document_number_prefix: model.document_number_prefix,
         tax_identification_number: model.tax_identification_number,
         finalize_zero_amount_invoice: model.finalize_zero_amount_invoice,
-        billing_configuration:
```
vincent-pochet, line 29: It's a breaking change in the API contract. Is it expected?

(I agree that it does look better :slightly_smiling_face: )

vincent-pochet (approved):

app/serializers/v1/billing_entity_serializer.rb
annvelents: it's not a breaking change because it's a completely new model 😁 
we're leaving organization as is, but for the new billing entities we'll be serializing billing_configuration the new way

vincent-pochet: Oh indeed, sorry for the noise :D

PR #3414, opened 2025-03-31
 fix(services): return valid Result in AutoGenerateService

julienbourdeau:
Sometimes you don't need the result of the service to return any data but all services MUST return a result.

How do you feel about an `EmptyResult` class. It avoid creating the same class multiple times.

Any idea for a better name?

### DIFF
--- a/app/services/taxes/auto_generate_service.rb
+++ b/app/services/taxes/auto_generate_service.rb
@@ -2,6 +2,8 @@
 
 module Taxes
   class AutoGenerateService < BaseService
+    Result = BaseResult
+
     def initialize(organization:)
       @organization = organization
       super
@@ -13,6 +15,8 @@ module Taxes
       end
 
       create_generic_taxes
+
+      result
     end
 
     private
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):

app/services/invoices/create_advance_charges_invoice_subscription_service.rb
```diff
@@ -2,7 +2,7 @@
 
 module Invoices
   class CreateAdvanceChargesInvoiceSubscriptionService < BaseService
-    Result = BaseResult
+    Result = EmptyResult
```
vincent-pochet, line 5: Actually, this is the same as
```suggestion
    Result = BaseResult
```

rsempe (approved):

app/services/invoices/create_advance_charges_invoice_subscription_service.rb
julienbourdeau: This is absolutely exactly the same 🤦‍♂️ Thank you!

PR #3417, opened 2025-03-31
Feat(multientities): add billing_entity to invoices#preview

annvelents:
## Context

When querying `invoices#preview` endpoint, a billing_entity_code might be provided - in this case we should create a preview for customer from this billing_entity. Otherwise a default billing_entity should be used.

Since customer.external_id stays uniq within org, if customer.external_id is provided, we're ignoring billing_entity when looking for a customer


ancorcruz (approved):
🚀

PR #3420, opened 2025-03-31
Feat/multientities 4 graphql taxes

annvelents:
## Context

Added graphql endpoints to apply and remove taxes


app/graphql/resolvers/billing_entity_taxes_resolver.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Resolvers
+  class BillingEntityTaxesResolver < Resolvers::BaseResolver
```
ancorcruz, line 4: we are missing specs for this resolver.

ancorcruz (approved):
LGTM!

app/graphql/resolvers/billing_entity_taxes_resolver.rb
annvelents: Added 🙌

PR #3421, opened 2025-04-01
Fix(multientities): create customer with billing_entity when creating a subscription

annvelents:
## Context

We missed that when creating a subscription there is also a possibility to create a customer by external_id. This customer is created without using the service and some our clients are using this endpoint creating customers

## Description

When customer, returned by `Customer.find_or_initialize_by` doesn't have a billing_entity_id,
billing_entity_id is assigned to a default one for the organization or to the provided in the request


spec/services/daily_usages/compute_service_spec.rb
```diff
@@ -134,11 +134,13 @@
...
       context "when subscription is terminated after the timestamp" do
+        # Ensure that we terminate he subscription and ingest event at the same billing period
+        let(:termination_date) { Time.current.beginning_of_month - 2.days }
         let(:subscription) do
-          create(:subscription, :terminated, :calendar, customer:, plan:, started_at: 1.year.ago)
+          create(:subscription, :terminated, :calendar, customer:, plan:, started_at: 1.year.ago, terminated_at: termination_date)
         end
 
-        let(:timestamp) { subscription.terminated_at - 1.day }
+        let(:timestamp) { termination_date - 1.day }
```
julienbourdeau, line 143: Is this related to the change?

julienbourdeau (approved):

nudded (approved):

spec/requests/api/v1/subscriptions_controller_spec.rb
```diff
@@ -123,6 +129,72 @@
...
+              billing_entity_code: billing_entity.id
+            }
+          end
+
+          it "uses the customer ignoring billing_entity" do
+            expect { subject }.not_to change(Customer, :count)
+
+            customer.reload
+            expect(customer.billing_entity).to eq(organization.default_billing_entity)
+          end
```
ancorcruz, line 195: How do we handle this if we introduce billing entity api keys and the customer's billing entity is outside of the api key scope? This is not for now but something to consider.

ancorcruz (approved):
LGTM

spec/requests/api/v1/subscriptions_controller_spec.rb
annvelents: it's a good question... Yes, we;ll have to think on it when we introduce billing_entity api keys

PR #3422, opened 2025-04-01
Fix Invoices::PrepaidCreditJob queueing to ensure uniqueness

ancorcruz:
Invoices::PrepaidCreditJob must but unique to avoid incorrect wallet balances... a bug was introduced where payment_status is a string value while by default it is a symbol. This mess up the uniqueness of the job
  and caused duplicated processing, hence incorrect wallet balances.

### DIFF
--- a/app/services/invoices/update_service.rb
+++ b/app/services/invoices/update_service.rb
@@ -82,7 +82,7 @@ module Invoices
       return unless invoice.invoice_type&.to_sym == :credit
       return unless %i[succeeded failed].include?(payment_status.to_sym)
 
-      Invoices::PrepaidCreditJob.perform_later(invoice, payment_status)
+      Invoices::PrepaidCreditJob.perform_later(invoice, payment_status.to_sym)
     end
 
     def valid_metadata_count?(metadata:)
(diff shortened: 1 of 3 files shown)


nudded (approved):

app/services/invoices/update_service.rb
```diff
@@ -82,7 +82,7 @@ def handle_prepaid_credits(payment_status)
       return unless invoice.invoice_type&.to_sym == :credit
       return unless %i[succeeded failed].include?(payment_status.to_sym)
 
-      Invoices::PrepaidCreditJob.perform_later(invoice, payment_status)
+      Invoices::PrepaidCreditJob.perform_later(invoice, payment_status.to_sym)
```
vincent-pochet, line 85: What about handling this with a `lock_key_arguments` defined in the job itself?

It would reduce the risk a forget about it :)

vincent-pochet: Note: that it could be tackled later

ancorcruz: I will add it as well, thanks.

ancorcruz: It is interesting that job uniqueness mechanism does not consider the default value of the second argument and only consider the actual arguments passed in the perform call.

PR #3425, opened 2025-04-01
Fix(multientities): add billing_entity_id to cache key for analytics

annvelents:
## Context

When caching analytics, billing_entity_id should be taken into account


ancorcruz (approved):
LGTM

PR #3426, opened 2025-04-01
fix(bug): Stripe payment intent creation payload

ancorcruz:
when payload includes payment_method_data, it is not allowed to include payment_method as well.


brunomiguelpinto (approved):
Ah yes good catch

vincent-pochet (approved):

julienbourdeau (approved):

PR #3427, opened 2025-04-01
Feat(multientities): Add support of billing_entity filter in invoices query

annvelents:
## Context

We need a possibility to filter invoices by billing_entity in Graphql resolver and in api endpoint

## Description

Added billing_entity param to invoice_resolver and invoice#index action
Added billing_entity_id filter in Invoices query


ancorcruz (approved):
loos good

PR #3428, opened 2025-04-01
Feat(multientities): Add support of billing_entity filter in customers query 

annvelents:
## Context

We need a possibility to filter customers by billing_entity in Graphql resolver and in api endpoint

## Description

Added billing_entity param to customers_resolver and customers#index action
Added billing_entity_id filter in Custoemers query

EDIT:
When adding contract, noticed on the design that we should accept array of billing_entities, changed the code
- to accept billing_entity_codes and billing_entity_ids for customers
- to accept billing_entity_codes and billing_entity_ids for invoices

### DIFF
--- a/app/contracts/queries/customers_query_filters_contract.rb
+++ b/app/contracts/queries/customers_query_filters_contract.rb
@@ -5,6 +5,7 @@ module Queries
     params do
       required(:filters).hash do
         optional(:account_type).array(:string, included_in?: Customer::ACCOUNT_TYPES.values)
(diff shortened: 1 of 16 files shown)


ancorcruz (changes requested):
Looks good but the this query object has a contract... I'm wondering why the customers query object did not fail as there is one filter not covered in the contract.... could you take a look and also update the contract please?

annvelents:
oops, I only checked it on invoice 🤦‍♀️  and forgot to check on other models. Thanks, will update it 🙇

PR #3429, opened 2025-04-01
Feat(multientities): Add support of billing_entity filter in credit_notes query

annvelents:
## Context

We need a possibility to filter credit_notes by billing_entity in Graphql resolver and in api endpoint

## Description

Added billing_entity param to credit_notes_resolver and credit_notes#index action
Added billing_entity_id filter in CreditNotes query

### DIFF
--- a/app/controllers/api/v1/credit_notes_controller.rb
+++ b/app/controllers/api/v1/credit_notes_controller.rb
@@ -94,7 +94,7 @@ module Api
 
       def index
         billing_entities = current_organization.billing_entities.where(code: params[:billing_entity_codes]) if params[:billing_entity_codes].present?
-        return not_found_error(resource: "billing_entity") if params[:billing_entity_codes].present? && billing_entities.blank?
+        return not_found_error(resource: "billing_entity") if params[:billing_entity_codes].present? && billing_entities.count != params[:billing_entity_codes].count
 
         result = CreditNotesQuery.call(
           organization: current_organization,
--- a/app/controllers/api/v1/credit_notes_controller.rb
+++ b/app/controllers/api/v1/credit_notes_controller.rb
@@ -93,6 +93,7 @@ module Api
       end
 
       def index
+        billing_entity = current_organization.billing_entities.find_by!(code: params[:billing_entity_code]) if params[:billing_entity_code].present?
         result = CreditNotesQuery.call(
           organization: current_organization,
           pagination: {
@@ -103,6 +104,7 @@ module Api
           filters: {
             amount_from: params[:amount_from],
             amount_to: params[:amount_to],
+            billing_entity_id: billing_entity&.id,
             credit_status: params[:credit_status],
             currency: params[:currency],
             customer_external_id: params[:external_customer_id],
@@ -128,6 +130,8 @@ module Api
         else
           render_error_response(result)
         end
+      rescue ActiveRecord::RecordNotFound => e
+        not_found_error(resource: e.model)
       end
 
       def estimate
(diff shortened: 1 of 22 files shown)


spec/queries/credit_notes_query_spec.rb
```diff
@@ -362,6 +362,20 @@
     end
   end
 
+  context "when billing entity id filter applied" do
+    let(:billing_entity) { create(:billing_entity, organization:) }
+    let(:filters) { {billing_entity_id: billing_entity.id} }
+
+    let!(:matching_credit_note) { create(:credit_note, customer:, invoice: create(:invoice, billing_entity:)) }
```
ancorcruz, line 369: we don't use `let!` lets do let a use the `before` block

spec/requests/api/v1/credit_notes_controller_spec.rb
```diff
@@ -458,6 +458,32 @@
         expect(json[:credit_notes].pluck(:lago_id)).to contain_exactly matching_credit_note.id
       end
     end
+
+    context "with billing entity code filter" do
+      let(:params) { {billing_entity_code: billing_entity.code} }
+      let(:billing_entity) { create(:billing_entity, organization:) }
+      let!(:matching_credit_note) { create(:credit_note, customer:, invoice: create(:invoice, billing_entity:)) }
```
ancorcruz, line 465: we don't use `let!` use `let` and `before` please.

ancorcruz (approved):
Looks good.

spec/queries/credit_notes_query_spec.rb
annvelents: interestingly, in this spec they exactly were used, that's why I copied this approach

spec/requests/api/v1/credit_notes_controller_spec.rb
annvelents: and I think, the same is here...
Will update to not use `let!`

app/controllers/api/v1/credit_notes_controller.rb
```diff
@@ -93,6 +93,9 @@ def void
       end
 
       def index
+        billing_entities = current_organization.billing_entities.where(code: params[:billing_entity_codes]) if params[:billing_entity_codes].present?
+        return not_found_error(resource: "billing_entity") if params[:billing_entity_codes].present? && billing_entities.count != params[:billing_entity_codes].count
```
ancorcruz, line 97: Do we want to fail or return results as long as there is at least one match?

spec/queries/credit_notes_query_spec.rb
```diff
@@ -362,6 +362,24 @@
...
+    before do
+      matching_credit_note
+      other_credit_note
+    end
+
+    it "returns credit notes with matching billing entity id" do
+      expect(result).to be_success
+      expect(result.credit_notes.pluck(:id)).to contain_exactly matching_credit_note.id
+    end
+  end
```
ancorcruz, line 381: could we haven scenario when we have invoices for more than one billing entity?

app/controllers/api/v1/credit_notes_controller.rb
annvelents: hah, initially I had exactly this logic, but then I decided that since for not existing we return `not_found`, it will be better if we will return `not_found` in this case as well, so users will know that they sent something wrong

spec/queries/credit_notes_query_spec.rb
annvelents: done 🙌

PR #3431, opened 2025-04-02
Feat(multientities): migration to populate billing_entity_id on fees

annvelents:
## Context

As we're adding billing_entity to be an instance responsible for billing, we want all billing-related data to be assigned to a billing_entity.
Also we will allow to delete billing_entities, so having billing_entity_id on fees will be beneficial

Note: locally with 600k fees the migration took 51 seconds


app/jobs/migrate/populate_fees_with_billing_entity_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Migrate
+  class PopulateFeesWithBillingEntityJob < ApplicationJob
+    queue_as :low_priority
+
+    def perform
+      Fee.where(billing_entity_id: nil).find_in_batches(batch_size: 5000) do |batch|
+        PopulateBatchFeesWithBillingEntityJob.perform_later(batch.pluck(:id))
+      end
```
ancorcruz, line 10: what about instead of queueing jobs with 5k uuids as params do a self queueing job...

```
        BATCH_SIZE = 5_000
        
	def perform
          batch = Fee.unscoped.where(billing_entity_id: nil).limit(BATCH_SIZE)
	  
	  if batch.exists?
	    batch.update_all("billing_entity_id = organization_id") # rubocop:disable Rails/SkipsModelValidations

            # Queue the next batch
            self.class.perform_later
          end
	end

ancorcruz (approved):
LGTM!

PR #3432, opened 2025-04-02
Chore(fees): add migration to populate fees with organization_id

annvelents:
## Context

We've added an organization_id to fees, but since it is a very large table, they should be populated in background jobs
Because some fees are missing invoice_id, some are missing subscription_id, we've implemented two algorithms:
- PopulateFeesWithOrganizationFromInvoice
- PopulateFeesWithOrganizationFromSubscription

These jobs should be triggered manually and a migration guide should be provided for self-hosted clients

Locally 600K Fees were processed in ~150s
On US cluster we have 3.5M fees without org_id
On EU cluster we have ~2.3M fees without org_id


annvelents:
TODO:
1) we have fees that do not belong to an invoice - they can be populated through subscriptions
2) check sidekiq serialization limits

app/jobs/migrate/populate_fees_with_organization_id_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Migrate
+  class PopulateFeesWithOrganizationIdJob < ApplicationJob
+    queue_as :low_priority
+
+    def perform
+      Fee.where(organization_id: nil).find_in_batches(batch_size: 1000) do |batch|
+        PopulateBatchFeesWithOrganizationIdJob.perform_later(batch.pluck(:id))
```
ancorcruz, line 9: could this be an issue on Redis side if we have to queue millions of uuids with memory and JSON serialization perfomance?

wouldn't be safer to run the migration as follows (one single migration job, no uuid queueing)?

```
        BATCH_SIZE = 1_000
        
        class Fee < ApplicationRecord
          belongs_to :invoice, optional: true
        end

	def perform
          batch = Fee.unscoped.where(organization_id: nil).joins(:invoice).limit(BATCH_SIZE)
	  
	  if batch.exists?
	    batch.update_all("organization_id = invoices.organization_id, billing_entity_id = invoices.organization_id")

            # Queue the next batch
            self.class.perform_later
          end
	end
```

app/jobs/migrate/populate_batch_fees_with_organization_id_job.rb
```diff
@@ -0,0 +1,25 @@
...
+    def perform(batch_ids)
+      Fee.includes(:organization).where(id: batch_ids, organization_id: nil).find_each do |fee|
+        Fee.no_touching do
+          Fee.where(id: fee.id).update_all(
+            organization_id: fee.organization.id,
+            billing_entity_id: fee.organization.id
+          )
+        end
+      end
+    end
```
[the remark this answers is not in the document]
annvelents: yaaay, missed this 😓 thank you!

app/jobs/migrate/populate_fees_with_organization_id_job.rb
annvelents: ah, that's amazing idea!
Thank you!

ancorcruz (approved):
🚀

annvelents:
@ancorcruz , sure
Yes, we decided to force it to be run manually, so we have more clear picture on what and when is happening

PR #3434, opened 2025-04-02
Feat(multientities): update discarded customers to be assigned to a billing entity

annvelents:
## Context

When data-exports are created, they include invoices that might  belong to descarded customers. When these customers are serialized, these customers should have billing_entity_id

Also Adding null: false for  customers table will require consistent data

Note: this job is done as a migration, because the original Customers migration was done as a migration. Also the amount of discarded customers is really low - ~5k on US, and they already were updated manually


ancorcruz (approved):
LGTM

PR #3435, opened 2025-04-02
Populate invoice_id for wallet transactions

floganz:
## Context

Wallet transaction uses invoice for payments but that invoice isn't assigned to wallet transaction despite such column.
Preparation step for new endpoint which will generate payment url for transactions with failed payments.

## Description

Assign invoice to wallet transaction.
Populate invoice reference for existing wallet transaction using migration.


db/migrate/20250402113844_fill_missing_invoice_id_on_wallet_transactions.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class FillMissingInvoiceIdOnWalletTransactions < ActiveRecord::Migration[7.2]
+  def up
+    safety_assured do
+      execute <<~SQL.squish
+        UPDATE wallet_transactions wt
```
nudded, line 7: have you ran this on production already? Does it take a long time?

vincent-pochet (approved):

PR #3436, opened 2025-04-02
fix(invoices): improve organization numbering performances

jdenquin:
- The way we were assigning organization sequential id was not efficient and doing a lot of queries
- We removed the `switched_from_customer_numbering?`
  - This should be done when the user update the organization to `per_organization` numbering

```ruby
# Instead of doing
organization
  .invoices
  .non_self_billed
   .where.not(organization_sequential_id: 0)
   .order(organization_sequential_id: :desc)
   .limit(1)
   .pick(:organization_sequential_id) || 0

# Just do
organization
  .invoices
  .non_self_billed
  .maximum(:organization_sequential_id) || 0
``` 

- Change the organization update logic when we change the document numbering

```ruby
if params.key?(:document_numbering)
  Organizations::UpdateInvoiceNumberingService.call(
    organization:,
    document_numbering: params[:document_numbering]
   )
end
```

- Create a new service `Organizations::UpdateInvoiceNumberingService` that will assign the last org sequential id to the last invoice created when changed to `per_organization`. So we don't need to check for all invoices whats the count.
- Use api calls to update the organization in the scenarios so it covers all the cases

⚠️ Before deploying : 

- [x] Make sure that no organizations are in a middle state (upgrading from `per_customer` to `per_organization`), if its the case, assign the last organization sequential id to the last invoice
- [x] QA has to be done by @nudded @ancorcruz

### DIFF
--- a/app/services/organizations/update_invoice_numbering_service.rb
+++ b/app/services/organizations/update_invoice_numbering_service.rb
@@ -16,8 +16,12 @@ module Organizations
       return result if organization.document_numbering == document_numbering
 
       if organization.per_customer? && document_numbering == "per_organization"
-        organization_invoices_count = organization.invoices.non_self_billed.with_generated_number.count
-        organization.invoices.order(created_at: :desc).first&.update!(organization_sequential_id: organization_invoices_count)
+        last_invoice = organization.invoices.non_self_billed.with_generated_number.order(created_at: :desc).first
+
+        if last_invoice
+            organization_invoices_count = organization.invoices.non_self_billed.with_generated_number.count
+            last_invoice.update!(organization_sequential_id: organization_invoices_count)
+        end
       end
 
       organization.document_numbering = document_numbering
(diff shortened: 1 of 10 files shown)


ancorcruz (approved):
🚀

vincent-pochet (approved):

vincent-pochet (approved):

app/services/organizations/update_invoice_numbering_service.rb
```diff
@@ -0,0 +1,36 @@
...
+      super
+    end
+
+    def call
+      result.organization = organization
+
+      return result if organization.document_numbering == document_numbering
+
+      if organization.per_customer? && document_numbering == "per_organization"
+        last_invoice = organization.invoices.non_self_billed.with_generated_number.order(created_at: :desc).first
```
ancorcruz, line 19: there is one edge scenario here we reproduce in QA, it is actually not affecting document numbering.

with `per_organization` creates invoices in `draft`:

`invoice_1`
`invoice_2`
`invoice_3`

then finalize `invoice_3` -> this one gets `organization_sequential_id = 1`

switch to `per_customer`

then finalize the other invoices, both get `organization_sequential_id = 0`

switch back to `per_organization`.

here the last invoice created is `invoice_3`, hence its `organization_sequential_id` is overwritten and changes from `1` to `3`.

The next invoice finalized will get a good `organization_sequential_id` and the invoice number will be fine.

PR #3437, opened 2025-04-03
Feat(multientities): graphql create and update user with billing entity

annvelents:
## Context

Update customer accepts billing_entity_id

## Description

Updated graphql inputs for create_customer and update_customer to receive billing_entity_id and send it to the services


ancorcruz (approved):
LGTM!

PR #3440, opened 2025-04-03
fix(invoices): Ensure no organization are in a middle state

ancorcruz:
Add a migration to queue a job.

This job will ensure no organization are in a middle state upgrading from `per_customer` to `per_organization`, if this is the case, assign the non self billed with generated number invoices count to the last invoice. This will ensure next invoice will get the right organization_sequential_id.

using a job instead of running the data migration within the actual migration has two benefits;
1. does not block the release with a long running migration.
2. no need to redefine all models scopes and relationships.

with one drawback;
We can not ensure the migration was done on time... but for this case it should be fine-ish


vincent-pochet (approved):

app/jobs/database_migrations/fix_invoices_organization_sequential_id_job.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def perform
+      Organization.per_organization.find_each do |organization|
+        last_organization_sequential_id = organization.invoices.maximum(:organization_sequential_id) || 0
+        invoices_count = organization.invoices.non_self_billed.with_generated_number.count
+
+        next if last_organization_sequential_id == invoices_count
+
+        last_invoice = organization.invoices.non_self_billed.with_generated_number.order(created_at: :desc).limit(1)
+        last_invoice.update_all(organization_sequential_id: invoices_count) # rubocop:disable Rails/SkipsModelValidations
```
nudded, line 15: no need for update_all here, it's only 1 invoice.

nudded (approved):

app/jobs/database_migrations/fix_invoices_organization_sequential_id_job.rb
ancorcruz: that is true but as it does not make a difference I would keep it as it is to release the fix today.

PR #3443, opened 2025-04-03
feat (avalara-integration): add GQL support for creating avalara connection

lovrocolic:
## Context

Currently, integration with tax provider Avalara is being implemented

## Description

This PR adds graphQL support for creating Avalara connection


app/graphql/types/integrations/avalara/create_input.rb
ivannovosad, line 1: Should we create an input spec as well?

app/services/integrations/avalara/create_service.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Integrations
+  module Avalara
+    class CreateService < BaseService
```
ivannovosad, line 5:
```suggestion
    class CreateService < BaseService
      Result = BaseResult[:integration]
```

ivannovosad (changes requested):

app/services/integrations/avalara/create_service.rb
lovrocolic: updated

ivannovosad (approved):

PR #3448, opened 2025-04-06
Feat/multientities 3 customers billing entity constraints

annvelents:
## Roadmap Task

## Context

## Description


db/migrate/20250402151930_validate_add_non_null_to_customers_billing_entity_id.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ValidateAddNonNullToCustomersBillingEntityId < ActiveRecord::Migration[7.2]
+  def up
+    validate_check_constraint :customers, name: "customers_billing_entity_id_null"
```
julienbourdeau, line 5: The goal here is to ensure that there is no null value currently in the db, right?

julienbourdeau (approved):

db/migrate/20250402151930_validate_add_non_null_to_customers_billing_entity_id.rb
annvelents: yep. Firstly we're adding the check,
then checking that there are no nulls, and after this setting null.
In case we firstly run this validate_check_constraint, when we set the null,
we separately perform
1) validate_check
2) set_null

So set_null doesn't have to validate null, and for the DB locking these are separate operations.
this was the resourceful output from strong_migrations:
```
Setting NOT NULL on an existing column blocks reads and writes while every row is checked.
Instead, add a check constraint and validate it in a separate migration.
```

PR #3456, opened 2025-04-07
feat: Add section_type to InvoiceCustomSections to support different section types

brunomiguelpinto:
## Context

Currently, users have no way to include bank transfer instructions or other automatic content directly on invoices.

To support this in a clean and scalable way, we’re introducing a section_type enum to distinguish between different types of invoice custom sections `manual` and `system_generated`.

This PR focuses only on introducing the structure needed to support that functionality in future changes.

## Description

This PR introduces the following changes:

- Added a new section_type enum column to invoice_custom_sections:
  - manual: user-defined sections (default)
  - system_generated: will be used for auto-generated content (e.g., Stripe funding instructions)
- Updated InvoiceCustomSection model with enum declaration and default value.
- Added two scoped associations in Organization:
  - manual_invoice_custom_sections
  - system_generated_invoice_custom_sections
- Added corresponding model specs to ensure proper behavior of the enum and scoped associations.

No logic yet for auto-generating or assigning system-generated sections — that will come in a follow-up PR.

### DIFF
--- a/app/graphql/types/customers/object.rb
+++ b/app/graphql/types/customers/object.rb
@@ -94,7 +94,7 @@ module Types
 
       field :finalize_zero_amount_invoice, Types::Customers::FinalizeZeroAmountInvoiceEnum, null: true, description: "Options for handling invoices with a zero total amount."
 
-      field :applicable_invoice_custom_sections, [Types::InvoiceCustomSections::Object], null: true, description: "Invoice custom sections applicable to the customer"
+      field :configurable_invoice_custom_sections, [Types::InvoiceCustomSections::Object], null: true, description: "Invoice custom sections manually configured by the customer"
       field :has_overwritten_invoice_custom_sections_selection, Boolean, null: true, description: "Define if the customer has custom invoice custom sections selection"
       field :skip_invoice_custom_sections, Boolean, null: true, description: "Skip invoice custom sections for the customer"
 
@@ -157,7 +157,7 @@ module Types
       end
 
       def has_overwritten_invoice_custom_sections_selection
-        !object.skip_invoice_custom_sections && object.selected_invoice_custom_sections.any?
+        !object.skip_invoice_custom_sections && object.manual_selected_invoice_custom_sections.any?
       end
     end
   end
--- a/app/graphql/types/customers/object.rb
+++ b/app/graphql/types/customers/object.rb
@@ -94,7 +94,7 @@ module Types
 
       field :finalize_zero_amount_invoice, Types::Customers::FinalizeZeroAmountInvoiceEnum, null: true, description: "Options for handling invoices with a zero total amount."
 
-      field :configurable_invoice_custom_sections, [Types::InvoiceCustomSections::Object], null: true, description: "Invoice custom sections manually configured by the customer"
+      field :configurable_invoice_custom_sections, [Types::InvoiceCustomSections::Object], null: true, description: "Invoice custom sections manually configured for the customer"
       field :has_overwritten_invoice_custom_sections_selection, Boolean, null: true, description: "Define if the customer has custom invoice custom sections selection"
       field :skip_invoice_custom_sections, Boolean, null: true, description: "Skip invoice custom sections for the customer"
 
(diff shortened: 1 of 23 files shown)


app/graphql/types/customers/object.rb
```diff
@@ -94,7 +94,7 @@ class Object < Types::BaseObject
 
       field :finalize_zero_amount_invoice, Types::Customers::FinalizeZeroAmountInvoiceEnum, null: true, description: "Options for handling invoices with a zero total amount."
 
-      field :applicable_invoice_custom_sections, [Types::InvoiceCustomSections::Object], null: true, description: "Invoice custom sections applicable to the customer"
+      field :configurable_invoice_custom_sections, [Types::InvoiceCustomSections::Object], null: true, description: "Invoice custom sections applicable to the customer"
```
annvelents, line 97: can we add here "selected manually", to show this difference here 🙈 
because I feel these fields might be confusing
or "Invoice custom sections , where section_type is manual..."

app/models/customer.rb
```diff
@@ -142,9 +149,16 @@ def applicable_net_payment_term
   end
 
   def applicable_invoice_custom_sections
+    manual_ids = configurable_invoice_custom_sections.map(&:id)
+    system_ids = system_generated_invoice_custom_sections.ids
+
+    InvoiceCustomSection.where(id: manual_ids + system_ids)
+  end
+
+  def configurable_invoice_custom_sections
```
annvelents, line 158: and maybe we can add a comment here to help other people to understand the difference between `configurable_invoice_custom_sections` and `applicable_invoice_custom_sections`
For example
applicable are the ones that should be applied in the invoice
configurable are manually selected on the customer or on the organization, if not selected on customer

app/graphql/types/customers/object.rb
brunomiguelpinto: Ah yes in the description ? Changing it

app/models/customer.rb
brunomiguelpinto: Good ideia I will put a small comment

app/models/invoice_custom_section.rb
```diff
@@ -7,6 +7,10 @@ class InvoiceCustomSection < ApplicationRecord
   belongs_to :organization
   has_many :invoice_custom_section_selections, dependent: :destroy
 
+  SECTION_TYPES = {manual: "manual", system_generated: "system_generated"}.freeze
+  enum :section_type, SECTION_TYPES, default: :manual, prefix: :section_type
+  attribute :payment_type, :string
```
annvelents, line 12: do we need this here? 🤔

db/migrate/20250403110833_add_section_type_to_invoice_custom_sections.rb
```diff
@@ -0,0 +1,39 @@
...
+    safety_assured do
+      execute <<~SQL
+        ALTER TABLE invoice_custom_sections ALTER COLUMN section_type SET DEFAULT 'manual';
+      SQL
+      execute <<~SQL
+        ALTER TABLE invoice_custom_sections ALTER COLUMN section_type SET NOT NULL;
+      SQL
+    end
+
+    add_index :invoice_custom_sections, :section_type, algorithm: :concurrently
```
annvelents, line 27: can we split this migration?

spec/graphql/resolvers/invoice_custom_sections_resolver_spec.rb
```diff
@@ -25,7 +25,9 @@
       create(:invoice_custom_section, organization:, name: "c"),
       create(:invoice_custom_section, organization:, name: "a"),
       create(:invoice_custom_section, organization:, name: "z"),
-      create(:invoice_custom_section, organization:, name: "n")
+      create(:invoice_custom_section, organization:, name: "n"),
+      create(:invoice_custom_section, organization:, name: "not show", section_type: "system_generated"),
+      create(:invoice_custom_section, organization:, name: "not show2", section_type: "system_generated")
```
annvelents, line 30: I think we can create only one system generated, can't we?

annvelents (commented):
looks good! Just a small comment

spec/graphql/resolvers/invoice_custom_sections_resolver_spec.rb
brunomiguelpinto: yes

PR #3457, opened 2025-04-07
feat(error): Add original_error to FailedResult

julienbourdeau:
## Context

The system swallows too much details about original error.

## Description

When introducing payment pre-auth when creating subscription, I introduced a new `ProviderError` which adds `error_details` to the API response. This was done in isolation just for this feature.

I'm trying to bring this `error_details` to the webhooks too.


app/services/base_service.rb
```diff
@@ -4,13 +4,32 @@ class BaseService
...
+
+    def original_error_details
+      if original_error.is_a?(::Stripe::StripeError)
+        stripe_error_details(original_error)
+      end
+    end
+
+    private
+
+    def stripe_error_details(err)
```
vincent-pochet, line 24: In my opinion this should not be handled in the generic failed result, but passed when creating the result itself (maybe via a decorator?).
I'm pretty sure that otherwise we will will soon have a LOT of error specific handling in this class

julienbourdeau: > we will will soon have a LOT of error specific handling in this class

Yes probably, but I don't see this as an issue ^^

julienbourdeau: Refactoring! 📈

vincent-pochet (approved):

app/serializers/v1/errors/error_serializer_factory.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module V1
+  module Errors
+    class ErrorSerializerFactory
+      def self.new_instance(error)
+        if error.is_a?(::Stripe::StripeError)
+          V1::Errors::StripeErrorSerializer.new(error)
+        else
+          raise NotImplementedError
```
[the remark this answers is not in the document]
julienbourdeau: The message is already sent outside the "details" so I think it's unexpected to not have a customer serializer when using this feature. 

But yeah it's probably best to get a not very useful message than a 500 ^^

Updating

julienbourdeau:
merging this monday

PR #3458, opened 2025-04-07
feat: Add Stripe funding instructions as system-generated invoice custom sections

brunomiguelpinto:
## Context

This PR builds on the groundwork introduced in the previous [PR](https://github.com/getlago/lago-api/pull/3456), where we added the section_type enum to distinguish between manual and system-generated invoice custom sections.

Now, we’re introducing the logic that automatically fetches bank transfer instructions from Stripe and injects them into invoices via a system_generated section — enabling customers to see accurate payment details directly on the invoice when bank transfer is selected.

## Description

This PR introduces the following:
- New background service that:
- Retrieves bank transfer instructions from Stripe for a given customer.
- Generates a system_generated InvoiceCustomSection with the Stripe funding details.
- Ensures the section is associated with the correct invoice and doesn’t duplicate if it already exists.
- Introduced logic to determine whether bank transfer is the selected payment method and conditionally trigger the generation.
- Ensures system-generated sections are included via applicable_invoice_custom_sections, alongside any manual sections.

image

### DIFF
--- a/app/jobs/payment_provider_customers/stripe_sync_funding_instructions_job.rb
+++ b/app/jobs/payment_provider_customers/stripe_sync_funding_instructions_job.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module PaymentProviderCustomers
+  class StripeSyncFundingInstructionsJob < ApplicationJob
+    queue_as :providers
+
+    retry_on ::Stripe::APIConnectionError, wait: :polynomially_longer, attempts: 6
+    retry_on ::Stripe::APIError, wait: :polynomially_longer, attempts: 6
+    retry_on ::Stripe::RateLimitError, wait: :polynomially_longer, attempts: 6
+    retry_on ActiveJob::DeserializationError
+
+    def perform(stripe_customer)
+      result = PaymentProviderCustomers::Stripe::SyncFundingInstructionsService.new(stripe_customer).call
+      result.raise_if_error!
+    rescue BaseService::UnauthorizedFailure => e
+      Rails.logger.warn(e.message)
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


app/jobs/payment_provider_customers/stripe_sync_funding_instructions_job.rb
```diff
@@ -0,0 +1,19 @@
...
+    retry_on ::Stripe::APIConnectionError, wait: :polynomially_longer, attempts: 6
+    retry_on ::Stripe::APIError, wait: :polynomially_longer, attempts: 6
+    retry_on ::Stripe::RateLimitError, wait: :polynomially_longer, attempts: 6
+    retry_on ActiveJob::DeserializationError
+
+    def perform(stripe_customer)
+      result = PaymentProviderCustomers::Stripe::SyncFundingInstructionsService.new(stripe_customer).call
+      result.raise_if_error!
+    rescue BaseService::UnauthorizedFailure => e
+      Rails.logger.warn(e.message)
```
nudded, line 16: there's nothing we can do in this case? feels weird to just log and mark the job as success?

brunomiguelpinto: api/app/jobs/payment_provider_customers/stripe_create_job.rb:12
api/app/jobs/payment_provider_customers/stripe_checkout_url_job.rb:17

We are doing the same on these places but in this case I agree to it would make sense to send something to the user, like a webhook or something, so they know something is wrong with the key

nudded: Okay if the logic is similar to other places, should be fine

app/services/payment_provider_customers/stripe/sync_funding_instructions_service.rb
```diff
@@ -0,0 +1,131 @@
...
+          )
+        end
+
+        {type: "eu_bank_transfer", eu_bank_transfer: {country: customer_country.upcase}}
+      end
+
+      def customer_currency
+        currency = customer.currency || customer.organization.default_currency
+
+        if currency.blank?
```
nudded, line 102: given that `default_currency` can never be blank, this check is not needed I think.

nudded (approved):

PR #3463, opened 2025-04-09
feat(avalara-integration): add support for company code

lovrocolic:
## Context

Currently Lago is working on integration with tax provider Avalara

## Description

This PR adds `company_code` support that was initially skipped during connection part.

Company code would be needed for fetching company ID, that is expected parameter in several Avalara requests.

### DIFF
--- a/app/services/integrations/avalara/update_service.rb
+++ b/app/services/integrations/avalara/update_service.rb
@@ -21,8 +21,6 @@ module Integrations
 
         integration.name = params[:name] if params.key?(:name)
         integration.code = params[:code] if params.key?(:code)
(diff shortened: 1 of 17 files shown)


app/services/integrations/avalara/update_service.rb
```diff
@@ -21,8 +21,6 @@ def call
 
         integration.name = params[:name] if params.key?(:name)
         integration.code = params[:code] if params.key?(:code)
```
ivannovosad, line 23: Should it be possible to update the `company_code` as well?

lovrocolic: nope, `company_code` should not be editable per current designs

ivannovosad (approved):

PR #3475, opened 2025-04-10
Feat(multientities): Services to apply and remove taxes on billing entity

annvelents:
## Context

Taxes will be created on organization, but applied on billing_entity, so we need to create services to apply and remove taxes

### DIFF
--- a/app/services/billing_entities/taxes/remove_taxes_service.rb
+++ b/app/services/billing_entities/taxes/remove_taxes_service.rb
@@ -2,8 +2,8 @@
 
 module BillingEntities
   module Taxes
-    class ApplyTaxesService < BaseService
-      Result = BaseResult[:applied_taxes, :taxes_to_apply]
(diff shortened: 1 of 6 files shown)


app/services/billing_entities/taxes/remove_taxes_service.rb
```diff
@@ -0,0 +1,44 @@
...
+      def find_taxes_to_remove
+        result.taxes_to_remove = organization.taxes.where(code: tax_codes)
+        if result.taxes_to_remove.count != tax_codes.count
+          result.not_found_failure!(resource: "tax")
+        end
+      end
+
+      def remove_taxes
+        @billing_entity.applied_taxes.where(tax: @taxes).destroy_all
+      end
```
ancorcruz, line 41: this method is not in usage.

ancorcruz (commented):
LGTM!

PR #3477, opened 2025-04-11
bug: Add zeitwerk check to the linters

nudded:
## Description

Run `rails zeitwerk:check` in production rails env to make sure all code can be loaded in production.


vincent-pochet (approved):
:rocket:

PR #3478, opened 2025-04-11
Feat(multientities): update old taxes services

annvelents:
## Context

The taxes services we had before, should create or delete applied taxes on default_billing_entity

## Description

Updated
- `Taxes::CreateService`
- `Taxes::UpdateService`
- `Taxes::DestroyService`

### DIFF
--- a/app/services/taxes/create_service.rb
+++ b/app/services/taxes/create_service.rb
@@ -20,6 +20,8 @@ module Taxes
       tax.applied_to_organization = params[:applied_to_organization] if params.key?(:applied_to_organization)
       tax.save!
 
+      apply_taxes_on_billing_entity if params[:applied_to_organization]
+
       result.tax = tax
       result
     rescue ActiveRecord::RecordInvalid => e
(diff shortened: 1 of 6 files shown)


app/services/taxes/destroy_service.rb
```diff
@@ -33,5 +34,10 @@ def draft_invoice_ids
         .draft
         .pluck(:id)
     end
+
+    def destroy_applied_taxes
+      billing_entity = tax.organization.default_billing_entity
+      BillingEntities::Taxes::RemoveTaxesService.call(billing_entity:, tax_codes: [tax.code])
```
vincent-pochet, line 40: Don't we want to raise/catch potential failures coming from the this service?

app/services/taxes/create_service.rb
```diff
@@ -29,5 +31,12 @@ def call
     private
 
     attr_reader :organization, :params
+
+    def apply_taxes_on_billing_entity
+      billing_entity = organization.default_billing_entity
+      if params[:applied_to_organization]
+        BillingEntities::Taxes::ApplyTaxesService.call(billing_entity:, tax_codes: [params[:code]])
```
vincent-pochet, line 38: Same question :)

app/services/taxes/destroy_service.rb
annvelents: we kinda shouldn't fail here, because the only error risen from `billing_entities/taxes` services is tax not found, but this is already processed at the beginning of method `call`...
The only other thing that can fail will be failing to save `BillingEntity::AppliedDax` object, where we have no validations anyway

vincent-pochet: Alright! Thank you for the explanation :+1:

vincent-pochet (approved):

PR #3484, opened 2025-04-14
Fix(multientities): one more seed org creation with billing_entity

annvelents:
## Context

Found one more place where we create an organization without billing_entity

### DIFF
--- a/lib/tasks/signup.rake
+++ b/lib/tasks/signup.rake
@@ -14,7 +14,8 @@ namespace :signup do
       user = User.create_with(password: ENV["LAGO_ORG_USER_PASSWORD"])
         .find_or_create_by!(email: ENV["LAGO_ORG_USER_EMAIL"])
       organization = Organization.find_or_create_by!(name: ENV["LAGO_ORG_NAME"])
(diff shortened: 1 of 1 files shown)


lib/tasks/signup.rake
```diff
@@ -14,6 +14,7 @@ namespace :signup do
       user = User.create_with(password: ENV["LAGO_ORG_USER_PASSWORD"])
         .find_or_create_by!(email: ENV["LAGO_ORG_USER_EMAIL"])
       organization = Organization.find_or_create_by!(name: ENV["LAGO_ORG_NAME"])
+      billing_entity = BillingEntity.find_or_create_by!(organization:, name: ENV["LAGO_ORG_NAME"], code: ENV["LAGO_ORG_NAME"].parameterize)
```
ancorcruz, line 17: For consistency lets ensure this first billing entity has the same id as the organization.

```suggestion
      billing_entity = BillingEntity.find_or_create_by!(id: organization.id, organization:, name: ENV["LAGO_ORG_NAME"], code: ENV["LAGO_ORG_NAME"].parameterize)
```

ancorcruz (changes requested):
to ensure we can still apply migrations as we have been doing lets ensure the billing entity id is equal to the organization id.

ancorcruz (approved):

PR #3490, opened 2025-04-14
feat(jobs): Make `Payments::UpdatePaymentMethodDataJob` unique

julienbourdeau:
## Description

With Lago you can use the same Stripe account for multiple organization. Maybe you have a second organization for another team or used a staging enviroment.

Let's say you have ORG_A where you set up stripe. You fill the API key and Lago will dispatch the `RegisterWebhookJob` to regsiter a Webhook Endpoint via the API with this endpoint url:

https://api.getlago.com/webhooks/stripe/#{uuid of ORG_A}?code=#{provider code}"

Now, you create a new organization for a different product or so. Same company using the same Stripe with 2 organizations.

The same Stripe account can be set up in ORG_B. Lago will set up a new webhook endpoint to this endpoint url:

https://api.getlago.com/webhooks/stripe/#{org_B.id}?code=#{code}"

Whenever something happens in ORG_A, like updating a customer for instance, Stripe will sent the webhook to EACH ENDPOINT set up in the account. Lago receives 2 webhooks, one for each url listed above. In your Stripe dashboard you'd see something like this:

CleanShot 2025-04-14 at 16 06 37@2x

--- 

This is not new and unrelated to the `UpdatePaymentMethodDataJob` but we want to avoid doing multple times the same work.

Note that the stripe webhooks arrive very fast so they are basically processed **all at the same time in parallel**.

💡 📈  Maybe one thing we can do is add the Stripe `event_id` to `inbound_webhooks` table with a unique index. This would guarantee that the webhook is processed only once.

### DIFF
--- a/app/jobs/payments/update_payment_method_data_job.rb
+++ b/app/jobs/payments/update_payment_method_data_job.rb
@@ -11,6 +11,11 @@ module Payments
 
     def perform(payment:, provider_payment_method_id:)
       ::Payments::UpdatePaymentMethodDataService.call!(payment:, provider_payment_method_id:)
+
+      # Now that the payment method is saved in the payment, we generate the PaymentReceipt
+      if payment.customer.organization.issue_receipts_enabled?
+        PaymentReceipts::CreateJob.perform_later(payment)
+      end
     end
   end
 end
--- a/app/jobs/payments/update_payment_method_data_job.rb
+++ b/app/jobs/payments/update_payment_method_data_job.rb
@@ -4,6 +4,9 @@ module Payments
   class UpdatePaymentMethodDataJob < ApplicationJob
     queue_as "default"
 
+    # NOTE: Even if the perform method is protected against running this job multple time
+    unique :until_executed
+
     retry_on ::Stripe::RateLimitError, wait: :polynomially_longer, attempts: 5
 
     def perform(payment:, provider_payment_method_id:)
--- a/app/jobs/payments/update_payment_method_data_job.rb
+++ b/app/jobs/payments/update_payment_method_data_job.rb
@@ -4,7 +4,8 @@ module Payments
   class UpdatePaymentMethodDataJob < ApplicationJob
     queue_as "default"
 
-    # NOTE: Even if the perform method is protected against running this job multple time
+    # NOTE: Even if the service is protected against running multiple time, this job must be unique.
+    #       https://github.com/getlago/lago-api/pull/3490
     unique :until_executed
 
     retry_on ::Stripe::RateLimitError, wait: :polynomially_longer, attempts: 5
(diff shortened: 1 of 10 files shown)


app/services/payments/update_payment_method_data_service.rb
```diff
@@ -4,15 +4,21 @@ module Payments
   class UpdatePaymentMethodDataService < BaseService
     Result = BaseResult[:payment]
 
-    def initialize(payment:, provider_payment_method_id:)
+    def initialize(payment:, provider_payment_method_id:, force: false)
```
vincent-pochet, line 7: Not sure to understand this `force` argument. When are we planning to call it with `force: true`?

julienbourdeau: Yeah I hesitated with this one.

I feel weird calling `UpdateThisThing` and the service doesn't update it. 
Originally, I had put the check in the job:

```ruby 
def perform(...)
  if payment.provider_payment_method_id != provider_payment_method_id
    ::Payments::UpdatePaymentMethodDataService.call!(payment:, provider_payment_method_id:)
  end
end
```

I thought I'd put in the service but then the service is weird. I liked the force option but maybe a better alternative is to rename the job:

* `SetPaymentMethodDataService`
* `AddPaymentMethodDataService`
* `CreatePaymentMethodDataService`
* ?

What do you think?

app/jobs/payments/update_payment_method_data_job.rb
```diff
@@ -4,10 +4,19 @@ module Payments
...
+    # NOTE: Even if the service is protected against running multiple time, this job must be unique.
+    #       https://github.com/getlago/lago-api/pull/3490
+    unique :until_executed
+
     retry_on ::Stripe::RateLimitError, wait: :polynomially_longer, attempts: 5
 
     def perform(payment:, provider_payment_method_id:)
       ::Payments::UpdatePaymentMethodDataService.call!(payment:, provider_payment_method_id:)
+
+      # Now that the payment method is saved in the payment, we generate the PaymentReceipt
```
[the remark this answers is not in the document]
julienbourdeau: I believe the service shouldn't know about this. When I call `UpdatePaymentMethod` I don't expect a payment receipt to be generated. The two aren't related. 
Ideally, [I'd love a good way to chain jobs](https://laravel.com/docs/12.x/queues#job-chaining) but so far I believe it's best way in Rails.

It feels a little silly to have a dedicated service to enqueue the job 😞 

```ruby
class SetPaymentMethodDataAndGeneratePaymentReceiptService
  def call
    SetPaymentMethodDataService.call(payment)
    GeneratePaymentReceiptJob.perform_async(payment)

    result
  end
end
```

**The issue might be the job name.** We do this for [GeneratePdfAndNotify](https://github.com/getlago/lago-api/blob/main/app/jobs/invoices/generate_pdf_and_notify_job.rb) but the job name is clearer.

vincent-pochet (approved):

app/services/payments/update_payment_method_data_service.rb
julienbourdeau: removed force and renamed!

julienbourdeau:
Also see this PR for related code changes: https://github.com/getlago/lago-api/pull/3511

PR #3494, opened 2025-04-15
feat(Idempotency): introduce IdempotencyRecord

nudded:
## Description

IdempotencyRecords are a new way to enforce idempotency on the DB level where you can't have all data in a single table.
(if you can enforce it with a uniqueness constraint, that approach is better, this should only be used if you need data across tables!) 

This PR adds the groundwork for this feature, next step is to use this for progressive billing.

### DIFF
--- a/app/services/idempotency_records/key_service.rb
+++ b/app/services/idempotency_records/key_service.rb
@@ -10,14 +10,14 @@ module IdempotencyRecords
     KEY_VERSION = "v1"
     SEPARATOR = "|"
 
-    def initialize(*key_parts)
+    def initialize(**key_parts)
       @key_parts = key_parts
 
       super()
     end
 
     def call
-      string_to_digest = key_parts.map(&:to_s).join(SEPARATOR)
+      string_to_digest = key_parts.sort.map { |k, v| "#{k}#{v}" }.join(SEPARATOR)
       result.idempotency_key = Digest::SHA256.digest("#{KEY_VERSION}#{SEPARATOR}#{string_to_digest}")
       result
     end
--- a/app/support/idempotency.rb
+++ b/app/support/idempotency.rb
@@ -20,13 +20,13 @@ class Idempotency
     attr_accessor :idempotent_resources
 
     def initialize
-      @idempotent_resources = Hash.new { |k, v| k[v] = [] }
+      @idempotent_resources = Hash.new { |k, v| k[v] = {} }
     end
 
     def ensure_idempotent!
       idempotent_resources.each do |resource, values|
         # generate idempotency key for this resource
-        idempotency_key = IdempotencyRecords::KeyService.call!(*values).idempotency_key
+        idempotency_key = IdempotencyRecords::KeyService.call!(**values).idempotency_key
 
         # try and generate a resource
         result = IdempotencyRecords::CreateService.call(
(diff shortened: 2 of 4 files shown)


spec/services/idempotency_records/key_service_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    context "with multiple key_parts" do
+      let(:key_parts) { ["key1", "key2"] }
+
+      it "returns the same value when called twice" do
+        expect(result.idempotency_key).to eq(described_class.call(*key_parts).idempotency_key)
+      end
+
+      it "does not return the same value if the order of the key_parts changes" do
+        expect(result.idempotency_key).not_to eq(described_class.call(*key_parts_reversed).idempotency_key)
+      end
```
julienbourdeau, line 33: I'm wondering if this is what we want. I fee like using `[payable.id, payable.amount]` should return the same key as `[payable.amount, payable.id]` 🤔

app/support/idempotency.rb
```diff
@@ -0,0 +1,98 @@
...
+  thread_mattr_accessor :current_transaction
+
+  class IdempotencyError < StandardError; end
+
+  # Represents a transaction context for an idempotent operation
+  class Transaction
+    attr_accessor :idempotent_resources
+
+    def initialize
+      @idempotent_resources = Hash.new { |k, v| k[v] = [] }
```
julienbourdeau, line 23: TIL! I can't believe I didn't know about this way of initializing Hash!

app/services/idempotency_records/key_service.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def initialize(*key_parts)
+      @key_parts = key_parts
+
+      super()
+    end
+
+    def call
+      string_to_digest = key_parts.map(&:to_s).join(SEPARATOR)
+      result.idempotency_key = Digest::SHA256.digest("#{KEY_VERSION}#{SEPARATOR}#{string_to_digest}")
```
julienbourdeau, line 21: I understand that we like services but I feel like this should really be on the model. The constant for version and separator especially.

julienbourdeau (approved):

spec/services/idempotency_records/key_service_spec.rb
nudded: Huh, interesting viewpoint. How would you implement that though? (just sorting? 🤔 )

app/support/idempotency.rb
nudded: ah yes, common pitfall is to do `Hash.new([])` which means the same array will be re-used for _all_ keys.

app/services/idempotency_records/key_service.rb
nudded: The model does not store the key parts though?

nudded: @julienbourdeau I'm thinking this has a chance of causing some unexpected bugs (like imagine using booleans for these values, order then is suddenly really important). Do you see any downside to requiring that the order is also important?

spec/services/idempotency_records/key_service_spec.rb
julienbourdeau: Yes, sorting the array before joining.

app/services/idempotency_records/key_service.rb
julienbourdeau: I didn't mean to store them but just to have the constant and a "build" method on the model. This is really a key feature of the model, usually services do stuff "around" the model.
This is not really important tho, it just felt weird on this one but feel free to ignore.

```
def build_key(key_parts)
  Digest::SHA256.digest("#{KEY_VERSION}#{SEPARATOR}#{key_parts.map(&:to_s).join(SEPARATOR)}")
end
```

spec/services/idempotency_records/key_service_spec.rb
julienbourdeau: Indeed I didn't think about boolean 🤔  Sorting is probably not possible then.

I was thinking about 2 services, doing similar things. If the args are in a different order it won't prevent duplication but it's probably very edge case. It's probably better to know that the order matters so we're careful.

nudded: I think I'll just use a hash instead? then the order is no longer important and it's easy to think about and use.

julienbourdeau (approved):

PR #3500, opened 2025-04-15
feat(multientities): Handle eu tax management at billing entity level

ancorcruz:
## Roadmap Task

## Context

## Description

Introduce a new service to standardize the management of EU taxes from billing entity creates and updates.

Organization update still changes its attribute but not handles the tax creation anymore. This attribute will be removed at some point.

### DIFF
--- a/spec/services/billing_entities/update_service_spec.rb
+++ b/spec/services/billing_entities/update_service_spec.rb
@@ -208,7 +208,7 @@ RSpec.describe BillingEntities::UpdateService do
 
           expect(result).not_to be_success
           expect(result.error).to be_a(BaseService::ValidationFailure)
-          expect(result.error.messages).to eq({eu_tax_management: ["org_must_be_in_eu"]})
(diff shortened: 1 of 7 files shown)


spec/services/billing_entities/update_service_spec.rb
```diff
@@ -208,7 +208,7 @@
 
           expect(result).not_to be_success
           expect(result.error).to be_a(BaseService::ValidationFailure)
-          expect(result.error.messages).to eq({eu_tax_management: ["org_must_be_in_eu"]})
+          expect(result.error.messages).to eq({eu_tax_management: ["billing_entity_must_be_in_eu"]})
```
annvelents, line 211: can we please add a test here that Taxes::AutoGenerate / ChangeEuTaxManagement service is called for eu billing_entity?

ancorcruz: it was already covered with tests, anyway I have took then chance to simplify the tests.

annvelents (approved):

PR #3502, opened 2025-04-16
feat(audit_logs): Create Clickhouse::ActivityLog model

rsempe:
## Roadmap Task

👉  [https://getlago.canny.io/feature-requests/p/access-audit-logs](https://getlago.canny.io/feature-requests/p/access-audit-logs)

## Context

Our users want to have visibility on what’s happening in Lago.

Examples:
- *A plan and prices have been updated: When? By whom? What?*
- *A subscription has been deleted: When? By whom? What?*

## Description

The goal of this PR is to introduce a new Clickhouse table `activity_logs` for storing all activity events from our customers.

We decided to put logs on Clickhouse instead of PG.
This is the first type of logs, but we want to add api and security logs in the future.

### DIFF
--- a/app/models/clickhouse/activity_log.rb
+++ b/app/models/clickhouse/activity_log.rb
@@ -6,6 +6,9 @@ module Clickhouse
     self.primary_key = nil
 
     belongs_to :organization
+    belongs_to :resource, polymorphic: true
+
+    before_save :ensure_activity_id
 
     def user
       organization.users.find_by(id: user_id)
@@ -23,8 +26,10 @@ module Clickhouse
       organization.subscriptions.find_by(external_id: external_subscription_id)
     end
 
-    def resource
-      resource_type.constantize.find_by(id: resource_id)
+    private
+
+    def ensure_activity_id
+      self.activity_id = SecureRandom.uuid if activity_id.blank?
     end
   end
 end
--- a/app/models/clickhouse/activity_log.rb
+++ b/app/models/clickhouse/activity_log.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module Clickhouse
+  class ActivityLog < BaseRecord
+    self.table_name = "activity_logs"
+    self.primary_key = nil
+
+    belongs_to :organization
+
+    def user
+      organization.users.find_by(id: user_id)
+    end
+
+    def api_key
+      organization.api_keys.find_by(id: api_key_id)
+    end
+
+    def customer
+      organization.customers.find_by(external_id: external_customer_id)
+    end
+
+    def subscription
+      organization.subscriptions.find_by(external_id: external_subscription_id)
+    end
+
+    def resource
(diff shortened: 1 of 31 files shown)


app/models/clickhouse/activity_log.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module Clickhouse
+  class ActivityLog < BaseRecord
+    self.table_name = "activity_logs"
+    self.primary_key = nil
+
```
mariohd, line 7: Should we add the enum values here in this class?

```suggestion
      SOURCES = [
        :api,
        :front,
        :system
      ].freeze
      
      enum :activity_source, SOURCES
```

app/models/clickhouse/activity_log.rb
```diff
@@ -0,0 +1,60 @@
...
+end
+
+# == Schema Information
+#
+# Table name: activity_logs
+#
+#  activity_source          :Enum8('api' = 1, not null
+#  activity_type            :string           not null
+#  logged_at                :datetime         not null
+#  resource                 :string
```
mariohd, line 49: it has `resource` and `resource_id`. 

What will be saved in the `resource` field?

rsempe: It's the `object` from paper trail - But yeah, that's confusing, you're right.

app/models/clickhouse/activity_log.rb
rsempe: Not sure we want this for clickhouse, but let's see

app/models/clickhouse/activity_log.rb
rsempe: let's go for `activity_object` and `activity_object_changes` 👍

app/models/clickhouse/activity_log.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module Clickhouse
+  class ActivityLog < BaseRecord
+    self.table_name = "activity_logs"
+    self.primary_key = nil
+
+    before_save :ensure_activity_id
+
+    def organization
```
vincent-pochet, line 10: I think you can define it as a `belongs_to`

app/models/clickhouse/activity_log.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def customer
+      organization.customers.find_by(external_id: external_customer_id)
+    end
+
+    def subscription
+      organization.subscriptions.find_by(external_id: external_subscription_id)
+    end
+
+    def resource
```
vincent-pochet, line 30: Same here, you might be able to use a polymorphic belongs_to

vincent-pochet (approved):
Looks, good! Just two (non blocking) suggestion :slightly_smiling_face:

mariohd (approved):

PR #3518, opened 2025-04-22
feat(payment-urls): Store payment url

floganz:
## Context

Currently invoice and wallet_transactions endpoints `payment_url` returns new payment link each time endpoint is hit. That creates high risk for double payments. With this changes we'll return the same payment link during 24 hours from first request.

## Description

Add `PaymentIntent` model for storing payment urls from payment providers.
Add service to fetch existing active url or create new one.
Adjust all payment providers to have the same expiration time on payment links.


julienbourdeau (commented):
I'm a little confused about what this PaymentIntent fixes. I feel like the important piece is the `idempotency_key` and we could build the key with the `payable.id` and `amount` for instance. 🤔

julienbourdeau:
Or maybe we could rely on the upcoming IdempotencyRecord! https://github.com/getlago/lago-api/pull/3494

julienbourdeau:
Alternative solution discussed would be to use `Rails.cache`.

```rb
def generate_payment_url
  Rails.cache.fetch("payurl-#{invoice.id}", expires_in: 24.hours - 1.minute) do
    do_generate_payment_url
  end
end
```

I see there a [`race_condition_ttl` option documented](https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html) but  never used it.

app/models/payment_intent.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class PaymentIntent < ApplicationRecord
+  STATUSES = [:active, :expired].freeze
+
+  belongs_to :invoice
```
julienbourdeau, line 6: We should probably be reusing this for other payable with payment url (like `PaymentRequest` created by dunning). So I think it should directly use a polymorphic `belongs_to :payable` relation here.

The name `PaymentIntent` also makes me think that we'll be using it for more stuff than payment_url.

If you wanted to keep it very focus, which I understand, maybe we can find a more specific name for the mode `InvoicePaymentUrl`

app/services/invoices/payments/moneyhash_service.rb
```diff
@@ -132,6 +136,7 @@ def payment_url_params
           customer: invoice.customer.moneyhash_customer.provider_customer_id,
           webhook_url: moneyhash_payment_provider.webhook_end_point,
           merchant_initiated: false,
+          expires_after_seconds: payment_intent.expires_at.to_i,
```
julienbourdeau, line 139: I believe here we need a number of second (time difference) and we're passing a timestamps.

https://docs.moneyhash.io/reference/payment-intent-v11

julienbourdeau (approved):
Hey @floganz I know this was a lot of work (and testing for different payment providers) so I don't want to block this any longer.

I have left a few last comments.

Later, I think we should add a cleaning job that remove old expired entries.

app/services/invoices/payments/moneyhash_service.rb
brunomiguelpinto: something like expires_after_seconds: (payment_intent.expires_at - Time.current).to_i

brunomiguelpinto: or create an helper that gives you this already calculated on the payment_intent

brunomiguelpinto (approved):

app/services/invoices/payments/moneyhash_service.rb
floganz: Indeed, very weird but it really seconds from now. Verified it on practice.

app/models/payment_intent.rb
floganz: That might be an option to use this model with `PaymentRequest` if there're similar issues. Do not know about any plans that change that part.

At this point that might used, might be not. I'd prefer to avoid changing it in advance without clear priorities. Rename of association shouldn't take a lot of time if we'll need it at some point.
Speaking of future, it could be that we'll extend this model with more intent-specific logic.

Wdyt?

ancorcruz:
I'm late but... PaymentIntent could be also applicable to payment requests (dunning) as payable object...

PR #3520, opened 2025-04-22
feat: change the void flow

brunomiguelpinto:
## Roadmap Task

## Context

## Description


app/services/coupons/void_and_restore_applied_coupon_service.rb
```diff
@@ -0,0 +1,56 @@
...
+      @applied_coupon = credit.applied_coupon
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "applied_coupon") if applied_coupon.nil?
+      return result if unlimited_usage?
+
+      ActiveRecord::Base.transaction do
+        applied_coupon.lock do
```
nudded, line 16: `lock do` creates an implicit transaction so no need for the explicit one

PR #3521, opened 2025-04-23
feat: Add create_org helper to Rails console

brunomiguelpinto:
## Context

This process usually involves calling the Organizations::CreateService and then manually creating an admin invite.
To streamline this, we’re adding a new helper method to Rails::ConsoleMethods.

## Description

This PR adds a create_organization(org_name:, email:) method to the Rails::ConsoleMethods module. It:
Creates an organization using Organizations::CreateService
Creates an admin invite for the specified email via Invites::CreateService
Prints the invitation URL in the console

Also returns the organization and invite_url in a hash for further use.

### DIFF
--- a/config/initializers/console.rb
+++ b/config/initializers/console.rb
@@ -71,4 +71,22 @@ module Rails::ConsoleMethods
       puts "Invoice #{id} has been successfully deleted." # rubocop:disable Rails/Output
     end
   end
+
+  def create_organization(org_name:, email:)
+    organization = Organizations::CreateService
+      .call(name: org_name, document_numbering: "per_organization")
+      .raise_if_error!
+      .organization
+
+    result = Invites::CreateService.call(
+      current_organization: organization,
+      email: email,
+      role: :admin
+    )
+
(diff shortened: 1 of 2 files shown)


config/initializers/console.rb
```diff
@@ -71,4 +71,22 @@ def hard_delete_invoice(id)
...
+      .raise_if_error!
+      .organization
+
+    result = Invites::CreateService.call(
+      current_organization: organization,
+      email: email,
+      role: :admin
+    )
+
+    url = "https://app.getlago.com/invitation/#{result.invite.token}"
```
nudded, line 87: this url is not correct on EU, but no big deal 👍

nudded (approved):

config/initializers/console.rb
julienbourdeau, line 87: I think for the URL you need to use the ENV var to be sure to have the url of the current cluster (and use it locally).
Maybe the URL could be added to the service too and have `result.invite_url`.

PR #3522, opened 2025-04-23
Feat(multientities): payment generation switch to entity

annvelents:
## Context

When creating payments, the info should be taken from the billing_entity

## Description

- added billing_entity into payment description and metadata


ancorcruz (approved):
LGTM

PR #3510, opened 2025-04-17
feat(alerting): Introduce Alerting 🔔

julienbourdeau:

PR #3528, opened 2025-04-24
refactor(billing): Breakdown RecalculateAndCheckService to inject current_usage

julienbourdeau:
## Description

[Alerting](https://github.com/getlago/lago-api/pull/3510) is coming and we need to calculate CurrentUsage only once  per subscription for all services.
This PR breaks down `RecalculateAndCheckService` into 2 services:

* `CalculateService`: was already there but current_usage can be injected (not used yet)
* `CheckThresholdsService`: The same as RecalculateAndCheckService except I removed the inner call to `CalculateService`.

### DIFF
--- a/app/jobs/lifetime_usages/recalculate_and_check_job.rb
+++ b/app/jobs/lifetime_usages/recalculate_and_check_job.rb
@@ -13,7 +13,8 @@ module LifetimeUsages
     unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
 
     def perform(lifetime_usage)
-      LifetimeUsages::RecalculateAndCheckService.call(lifetime_usage:)
+      LifetimeUsages::CalculateService.call!(lifetime_usage:)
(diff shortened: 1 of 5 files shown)


app/jobs/lifetime_usages/recalculate_and_check_job.rb
```diff
@@ -13,7 +13,8 @@ class RecalculateAndCheckJob < ApplicationJob
     unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
 
     def perform(lifetime_usage)
-      LifetimeUsages::RecalculateAndCheckService.call(lifetime_usage:)
+      LifetimeUsages::CalculateService.call(lifetime_usage:)
```
julienbourdeau, line 16: to be completely consistent with before, this should be `call!` I think

```suggestion
      LifetimeUsages::CalculateService.call!(lifetime_usage:)
```

nudded (approved):

PR #3535, opened 2025-04-25
feat(alerting): Introduce SubscriptionActivity

julienbourdeau:
## Context

First big step toward Alerting feature

## Description

Follows: https://github.com/getlago/lago-api/pull/3528

* Every time an event is received, we insert a SubscriptionActivity in the db, unique per subscription
* Every 5 minutes, we recompute `current_usage` for all active subscription
  * If org has progressive billings, we check the thresholds
  * If org is using alerting, we'll check the alert at the same time (later
* Stop using `lifetime_usage.recalculate_current_usage` but the job keeps looking at this column until there are not more jobs to process


app/services/usage_monitoring/process_all_subscription_activities_service.rb
```diff
@@ -0,0 +1,18 @@
...
+
+module UsageMonitoring
+  class ProcessAllSubscriptionActivitiesService < BaseService
+    Result = BaseResult
+
+    def call
+      # NOTE: If we need to handle different delays per organization, this would be done here.
+      #     This is why it's a dedicated service and not just done in the job
+
+      Organization.with_activity_tracking.select(:id).find_each do |organization|
```
nudded, line 11: I don't see a reason for this "funky" pattern. I'd rather be explicit and fetch with: `.pluck(:id).each` 

There's hardly any reason to use `find_each` here anyway.

app/services/usage_monitoring/process_subscription_activity_service.rb
```diff
@@ -0,0 +1,54 @@
...
+        LifetimeUsages::CalculateService.call!(lifetime_usage:, current_usage:)
+      end
+
+      if organization.progressive_billing_enabled?
+        LifetimeUsages::CheckThresholdsService.call(lifetime_usage:)
+      end
+
+      # TODO: Add Alerting here
+
+      subscription_activity.delete
```
nudded, line 26: You'll end up creating lots of bloat on this table, keep in mind that it will need to be `vacuumed` and also `repacked` often.

app/services/usage_monitoring/process_organization_subscription_activities_service.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      organization.subscription_activities.where(enqueued: false).select(:id).in_batches(of: BATCH_SIZE) do |batch|
+        jobs = []
+        batch.each do |subscription_activity|
+          jobs << ProcessSubscriptionActivityJob.new(subscription_activity.id)
+        end
+
+        ActiveRecord::Base.transaction do
+          batch.update_all(enqueued: true, enqueued_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
+          after_commit { ActiveJob.perform_all_later(jobs) }
```
nudded, line 27: `perform_all_later` does bypass the AJ-uniqueness I think, (it's not configured for this job, but perhaps you could end up enqueuing lots of jobs here)

app/services/events/post_process_service.rb
```diff
@@ -95,8 +95,11 @@ def expire_cached_charges(subscriptions)
       end
     end
 
-    def flag_refresh_from_subscription
-      subscriptions.select(&:active?).each { |s| LifetimeUsages::FlagRefreshFromSubscriptionService.new(subscription: s).call }
+    def track_subscription_activity!
+      UsageMonitoring::TrackSubscriptionActivityService.call(
+        organization:,
+        subscription_ids: subscriptions.select(&:active?).pluck(:id)
```
nudded, line 101: the `.pluck` here seems to indicate a DB query, (given there's a `select(&:active?)` it's not, so perhaps it's better to just write `map`? 🤔

app/services/lifetime_usages/calculate_service.rb
```diff
@@ -19,18 +19,13 @@ def call
...
         lifetime_usage.invoiced_usage_amount_refreshed_at = Time.current
-        lifetime_usage.current_usage_amount_refreshed_at = Time.current
       end
 
-      if lifetime_usage.recalculate_current_usage
-        lifetime_usage.current_usage_amount_cents = calculate_current_usage_amount_cents
-        lifetime_usage.recalculate_current_usage = false
-        lifetime_usage.current_usage_amount_refreshed_at = Time.current
-      end
+      lifetime_usage.current_usage_amount_cents = calculate_current_usage_amount_cents
```
nudded, line 26: oh wow, that's a nice simplification 👏

app/services/usage_monitoring/process_subscription_activity_service.rb
julienbourdeau: Very good point. Who's going to do this now 😬

app/services/usage_monitoring/process_organization_subscription_activities_service.rb
julienbourdeau: Yeah, I didn't use activejob-uniqueness on purpose. With the `enqueued` flag, I don't expect duplicate jobs and if there is, the `subscription_activity_id` won't be found so the job will just exit.

nudded: Also one more potential problem is that you can have a job enqueued that dies (for some reason) and so you can end up with an `enqueued` record preventing any further processing for that subscription. Probably wise to setup a safeguard for that case as well.

app/services/usage_monitoring/process_all_subscription_activities_service.rb
julienbourdeau: Yeah, I can remove the `select` it's silly optimisation 👍

app/services/usage_monitoring/track_subscription_activity_service.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def call
+      return unless organization.tracks_subscription_activity?
+
+      activities = []
+      subscription_ids.each do |id|
+        activities << {organization_id: organization.id, subscription_id: id}
+      end
+
+      UsageMonitoring::SubscriptionActivity.insert_all(activities, unique_by: :idx_subscription_unique) # rubocop:disable Rails/SkipsModelValidations
```
julienbourdeau, line 21: I'll rework this to not insert activity if there is no usage_threshold or alerts !

app/services/events/post_process_service.rb
julienbourdeau: I have reworked the service and it takes a full model now

app/services/subscriptions/flag_refreshed_service.rb
```diff
@@ -29,10 +29,10 @@ def flag_wallets_for_refresh
         .update_all(ready_to_be_refreshed: true) # rubocop:disable Rails/SkipsModelValidations
     end
 
-    def flag_lifetime_usage_for_refresh
-      Subscription.where(id: subscription_id).includes(:lifetime_usage, plan: :usage_thresholds).find_each do
```
julienbourdeau, line 33: I have removed the eager loading for a single model and add the `has_many through:` on subscription so instead of loading plan and usage_threshold we get a "any? query" (shown below). Also, we might not even query if one of the earlier return condition is met.

```sql
SELECT
	1 AS one
FROM
	"usage_thresholds"
	INNER JOIN "plans" ON "usage_thresholds"."plan_id" = "plans"."id"
WHERE
	"plans"."id" = $1
LIMIT
	$2;
```

app/services/usage_monitoring/process_all_subscription_activities_service.rb
julienbourdeau: Alerting is not going to be a "premium integration", so this should just be `Organization.all` in the end.
And indeed, no need to "batch" since Organization is always expected to be small-ish.

app/services/usage_monitoring/process_organization_subscription_activities_service.rb
```diff
@@ -0,0 +1,41 @@
...
+        end
```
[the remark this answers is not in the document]
julienbourdeau: 🤔

app/models/usage_monitoring/subscription_activity.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module UsageMonitoring
+  class SubscriptionActivity < ApplicationRecord
```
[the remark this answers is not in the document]
nudded: okay, no biggie

app/jobs/usage_monitoring/process_organization_subscription_activities_job.rb
```diff
@@ -0,0 +1,18 @@
...
+  class ProcessOrganizationSubscriptionActivitiesJob < ApplicationJob
+    unique :until_executed, on_conflict: :log
+
+    def perform(organization_id)
+      return unless License.premium?
+
+      organization = Organization.find(organization_id)
+      result = UsageMonitoring::ProcessOrganizationSubscriptionActivitiesService.call!(organization:)
+
+      Rails.logger.info(
```
vincent-pochet, line 13: Not a blocker, but since job enqueuing is always logged anyway, do we really need this one?

app/services/usage_monitoring/process_organization_subscription_activities_service.rb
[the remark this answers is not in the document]
vincent-pochet: Sure copilot, sure sure.... :smile:

app/services/usage_monitoring/process_subscription_activity_service.rb
```diff
@@ -0,0 +1,54 @@
...
+        with_cache: true
+      ).usage
+    end
+
+    def find_or_create_lifetime_usage
+      subscription.lifetime_usage || begin
+        ltu = subscription.build_lifetime_usage(organization:)
+        ltu.save!
+        ltu
+      end
```
vincent-pochet, line 51: (just a suggestion ;) )
```suggestion
      subscription.lifetime_usage || subscription.create_lifetime_usage!(organization:)
```

app/services/usage_monitoring/track_subscription_activity_service.rb
```diff
@@ -0,0 +1,42 @@
...
+module UsageMonitoring
+  class TrackSubscriptionActivityService < BaseService
+    Result = BaseResult
+
+    # NOTE: The organization can be passed to avoid loading it from the subscription
+    #       If not passed, it's lazy loaded from the subscription
+    def initialize(subscription:, organization: nil)
+      @subscription = subscription
+      @organization = organization
+      super()
```
vincent-pochet, line 12:
```suggestion
      super
```

vincent-pochet (approved):
Looks good :+1:

nudded (approved):

app/services/usage_monitoring/process_subscription_activity_service.rb
julienbourdeau: yes 👌 
It was done this way so I didn't even think 😬

PR #3536, opened 2025-04-25
chore(organizations): Add Organization ref on add_ons_taxes

jdenquin:
- Reference organization on `add_ons_taxes`

### DIFF
--- a/app/jobs/database_migrations/populate_add_ons_taxes_with_organization_job.rb
+++ b/app/jobs/database_migrations/populate_add_ons_taxes_with_organization_job.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module DatabaseMigrations
+  class PopulateAddOnsTaxesWithOrganizationJob < ApplicationJob
+    queue_as :low_priority
+    unique :until_executed
+
+    BATCH_SIZE = 1000
+
+    def perform(batch_number = 1)
+      batch = AddOn::AppliedTax.unscoped
+        .where(organization_id: nil)
+        .limit(BATCH_SIZE)
+
+      if batch.exists?
+        # rubocop:disable Rails/SkipsModelValidations
+        batch.update_all("organization_id = (SELECT organization_id FROM add_ons WHERE add_ons.id = add_ons_taxes.add_on_id)")
+        # rubocop:enable Rails/SkipsModelValidations
+
+        # Queue the next batch
+        self.class.perform_later(batch_number + 1)
+      else
+        Rails.logger.info("Finished the execution")
+      end
+    end
+
+    def lock_key_arguments
+      [arguments]
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


db/migrate/20250425102555_validates_add_ons_taxes_organizations_foreign_key.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ValidatesAddOnsTaxesOrganizationsForeignKey < ActiveRecord::Migration[7.2]
+  def change
+    validate_foreign_key :add_ons_taxes, :organizations
```
julienbourdeau, line 5: Can we enforce the validation before the data migration (the job) is over 🤔

app/jobs/database_migrations/populate_add_ons_taxes_with_organization_job.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      if batch.exists?
+        # rubocop:disable Rails/SkipsModelValidations
+        batch.update_all("organization_id = (SELECT organization_id FROM add_ons WHERE add_ons.id = add_ons_taxes.add_on_id)")
+        # rubocop:enable Rails/SkipsModelValidations
+
+        # Queue the next batch
+        self.class.perform_later(batch_number + 1)
+      else
+        Rails.logger.info("Finished the execution")
```
julienbourdeau, line 23: Kind of unrelated to your PR but seeing this, I'm assuming that the job name is already attached to the logs.

I'm [adding a log here](https://github.com/getlago/lago-api/pull/3535/files#diff-3f25050769ae05d8823748244efb284d1b17b6d5a3f0cb76f55a0345b5844a87R13) and I think I'm doing it wrong then.

db/migrate/20250425102555_validates_add_ons_taxes_organizations_foreign_key.rb
nudded: this is only needed because adding a Foreign key with `validate:false` is quicker and requires less locks. Once the foreign key is validated it will be impossible to insert invalid data. (strong-migrations suggests to split up these migrations). So no data migration needs to take place before validating the foreign key afaik.

julienbourdeau: I realize I was confusing validating that it's a foreign key and validating that it's not nullable 🤦‍♂️ 
Thank you!

app/services/add_ons/apply_taxes_service.rb
```diff
@@ -20,7 +20,7 @@ def call
       ).destroy_all
 
       result.applied_taxes = tax_codes.map do |tax_code|
-        add_on.applied_taxes.find_or_create_by!(tax: taxes.find_by(code: tax_code))
+        add_on.applied_taxes.find_or_create_by!(tax: taxes.find_by(code: tax_code), organization: add_on.organization)
```
vincent-pochet, line 23: The `find_or_create_by` my lead to an error if the organization_id is not yet backfilled.

To avoid this you could do:
```suggestion
        add_on.applied_taxes.create_with(organization_id: add_on.organization_id).find_or_create_by!(tax: taxes.find_by(code: tax_code))
```

vincent-pochet (approved):

PR #3537, opened 2025-04-25
misc(segment): Track only active memberships during login

floganz:
## Context

Currently during login we can add any of user's memberships to segment, even revoked ones.
This PR ensures that membership we'll mention in segment will be always active. 

## Description

Ensure only active membership submitted to segment.
Add tests.

### DIFF
--- a/app/services/users_service.rb
+++ b/app/services/users_service.rb
@@ -4,14 +4,15 @@ class UsersService < BaseService
   def login(email, password)
     result.user = User.find_by(email:)&.authenticate(password)
 
-    unless result.user.present? && result.user.memberships&.active&.any?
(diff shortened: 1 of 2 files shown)


app/services/users_service.rb
```diff
@@ -4,14 +4,15 @@ class UsersService < BaseService
   def login(email, password)
     result.user = User.find_by(email:)&.authenticate(password)
 
-    unless result.user.present? && result.user.memberships&.active&.any?
+    unless result.user.present? && result.user.memberships.active.any?
```
julienbourdeau, line 7: 🙌

julienbourdeau (approved):

PR #3538, opened 2025-04-25
chore(rubocop): Disable Rails/SkipsModelValidations

julienbourdeau:
## Description

1. We rely on `insert_all` and such in many places and for good reasons
2. Every body in this team know this skips the model validation and callbacks (_that's the point!_)
3. A good rubocop rule shouldn't be disabled 56 times 😬 

If anybody wants to keep it, I'm happy to close this PR and keep adding the magic comment wherever we use these methods.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -80,6 +80,9 @@ RSpec/NestedGroups:
 RSpec/NamedSubject:
   Enabled: false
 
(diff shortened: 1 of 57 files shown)


julienbourdeau:
After discussion, we like being explicite about it!

PR #3546, opened 2025-04-28
fix(dispute-lost): Do do call lose dispute service for payment requests

ivannovosad:
## Context

We're processing payment dispute won or lost events even for payments that belong to payment requests not just invoices.
In that case the service fails because payment requests don't have the method defined:

```
undefined method `mark_as_dispute_lost!' for an instance of PaymentRequest (NoMethodError)
```

## Description

This PR fixes that by calling the method only for invoices.


vincent-pochet (approved):
:+1:

PR #3547, opened 2025-04-28
Feat(multientities): use organizations default_currency and timezone from the default billing_entity

annvelents:
## Context

Since we're deprecating these fields on organization, we should be taking them from the default billing_entity when sending to the frontend. 
The analytics still should collect data for different currencies, but on the FE they're sending us default currency of org to filter the values

## Description

Redefined default_currency and timezone on org
changed the way we retrieve default_billing_entity to not have to reload the association


app/models/organization.rb
```diff
@@ -59,7 +59,6 @@ class Organization < ApplicationRecord
   has_one :salesforce_integration, class_name: "Integrations::SalesforceIntegration"
 
   has_one :applied_dunning_campaign, -> { where(applied_to_organization: true) }, class_name: "DunningCampaign"
-  has_one :default_billing_entity, -> { active.order(created_at: :asc) }, class_name: "BillingEntity"
```
annvelents, line 62: I needed to change this one from association to method, because the association remains cached, while method is always reloaded. Now in tests we create a billing entity while creation organization, and since we have an additional logic for default_currency and timezone, it slows creation of organization for micro-seconds, but that's enough for rails to create them with less delay, and in order to get org's default_billing_entity in tests, the organization should be reloaded. Instead, having it always reloaded will be more safe

annvelents: at the same time, technically we will remove this additional logic... so maybe we can keep the scop, but add reload in after create for organization factory...

spec/models/customer_spec.rb
```diff
@@ -616,12 +616,14 @@
 
     let(:organization) { create(:organization, timezone: "America/Los_Angeles") }
```
ancorcruz, line 617: we can remove the timezone for the organization to avoid confusion when reading the specs.

```suggestion
    let(:organization) { create(:organization) } 
```

spec/models/customer_spec.rb
```diff
@@ -616,12 +616,14 @@
...
+
     it "has helper to get dates in timezones" do
-      aggregate_failures do
-        expect(customer.created_at.to_s).to eq("2022-11-17 23:34:23 UTC")
-        expect(customer.created_at_in_customer_timezone.to_s).to eq("2022-11-18 00:34:23 +0100")
-        expect(customer.created_at_in_organization_timezone.to_s).to eq("2022-11-17 15:34:23 -0800")
-      end
+      expect(customer.created_at.to_s).to eq("2022-11-17 23:34:23 UTC")
+      expect(customer.created_at_in_customer_timezone.to_s).to eq("2022-11-18 00:34:23 +0100")
+      expect(customer.created_at_in_organization_timezone.to_s).to eq("2022-11-17 15:34:23 -0800")
```
ancorcruz, line 626: shall we create a "on billing entity timezone" helper?

spec/models/organization_spec.rb
```diff
@@ -428,4 +428,32 @@
...
+  describe "default_currency" do
+    let(:organization) { create(:organization, default_currency: "USD") }
+    let(:billing_entity) { create(:billing_entity, organization:, default_currency: "EUR") }
+
+    before do
+      organization.default_billing_entity.update(default_currency: "GBP")
+      billing_entity
+    end
+
+    it "ignores existing record on organization and uses value from default_billing_entity" do
```
ancorcruz, line 441:
```suggestion
    it "ignores existing value in organization and uses value from default_billing_entity" do
```

spec/models/organization_spec.rb
```diff
@@ -428,4 +428,32 @@
...
+  describe "timezone" do
+    let(:organization) { create(:organization, timezone: "UTC") }
+    let(:billing_entity) { create(:billing_entity, organization:, timezone: "America/New_York") }
+
+    before do
+      organization.default_billing_entity.update(timezone: "Europe/London")
+      billing_entity
+    end
+
+    it "ignores existing record on organization and uses value from default_billing_entity" do
```
ancorcruz, line 455:
```suggestion
    it "ignores existing value in organization and uses value from default_billing_entity" do
```

ancorcruz (approved):
looks great, thanks.

ancorcruz (approved):
LGTM!!!!

PR #3550, opened 2025-04-28
Feat(multientiites): add billing entities to data exports

annvelents:
## Context

When creating a data export, if an organization has multiple entities, we want to show the billing_entity_code in the csv export

## Description

- Added billing_entity_code into filters input for invoices and credit notes
- Added an optional column to the data export csv, if an organization has multiple billing_entities


ancorcruz (approved):
looks great!

PR #3554, opened 2025-04-28
feat(alerting): Introduce Alerting 🔔

julienbourdeau:
* Introduce `UsageMonitoring::Alert` model with STI. Each type of alert get it's own child class
  * The specific alert class is meant to be able to find it's "current value" in a given object
  * We do this to compute `current_usage` only once and pass it to all alerts
* Process Alerts when process SubscriptionActivity, together with progressive billing
  * Ensure having one alert breaking keep the other running
  * Ensure having progressive billing keep the alert running
* Create, Update, Destroy services
* GraphQL layer


app/serializers/v1/usage_monitoring/triggered_alert_serializer.rb
```diff
@@ -0,0 +1,24 @@
...
+          current_value: model.current_value,
+          previous_value: model.previous_value,
+          crossed_thresholds: model.crossed_thresholds,
+          triggered_at: model.triggered_at.iso8601
+        }
+
+        # if model.alert.billable_metric
+        # TODO: Billable metric
+        # end
+      end
```
julienbourdeau, line 21: We probably want to know what charge has matched 🤔

db/migrate/20250520080000_create_usage_monitoring_alerts.rb
```diff
@@ -0,0 +1,57 @@
...
+      t.timestamps
+
+      t.index %w[subscription_external_id organization_id alert_type],
+        unique: true,
+        name: "idx_alerts_unique_per_type_per_subscription",
+        where: "(billable_metric_id IS NULL AND deleted_at IS NULL)"
+      t.index %w[subscription_external_id organization_id alert_type billable_metric_id],
+        unique: true,
+        name: "idx_alerts_unique_per_type_per_subscription_with_bm",
+        where: "(billable_metric_id IS NOT NULL AND deleted_at IS NULL)"
```
vincent-pochet, line 26: More a question:

I'm wondering if a single index is not enough:
```ruby
t.index %w[subscription_external_id organization_id alert_type billable_metric_id],
  unique: true,
  name: "idx_alerts_unique_per_type_per_subscription",
  where: "deleted_at IS NULL"
```

app/graphql/mutations/usage_monitoring/alerts/update.rb
```diff
@@ -0,0 +1,33 @@
...
+        type Types::UsageMonitoring::Alerts::Object
+
+        def resolve(**args)
+          alert = current_organization.alerts.find(args[:id])
+          result = ::UsageMonitoring::UpdateAlertService.call(
+            alert:,
+            params: args,
+            billable_metric: args[:billable_metric_id] ?
+              current_organization.billable_metrics.find(args[:billable_metric_id]) :
+              nil
```
vincent-pochet, line 25: Same comment regarding the billable_metric

app/models/usage_monitoring/alert.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    def find_value(current_metrics)
+      raise NotImplementedError
+    end
+
+    private
+
+    def billable_metric_when_required
+      if billable_metric_id.blank? && BILLABLE_METRIC_TYPES.include?(alert_type)
+        errors.add(:billable_metric_id, "is required for `#{alert_type}` alert type")
```
vincent-pochet, line 97: We are usually using error_codes rather than english sentences.
The reason behind this is that it allows API users to map these codes with proper translations / handling on their side

app/services/usage_monitoring/create_alert_service.rb
```diff
@@ -0,0 +1,58 @@
...
+      @subscription = subscription
+      @params = params
+      @billable_metric = billable_metric
+      super
+    end
+
+    def call
+      [:code, :alert_type, :thresholds].each do |field|
+        if params[field].blank?
+          return result.single_validation_failure!(field:, error_code: "#{field}_must_be_present")
```
vincent-pochet, line 18: any reason to validate this in the service rather than adding validation on the model?

With the current approach if the 3 fields are missing you will have an error only for `code`

app/models/usage_monitoring/alert.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    def subscription
+      @subscription ||= organization
+        .subscriptions
+        .active
+        .order(started_at: :desc)
+        .find_by(external_id: subscription_external_id)
+    end
+
+    def find_thresholds_crossed(current)
```
vincent-pochet, line 50: I think this should be in the in the `ProcessAlertService`, it is coupled with the service logic and it is not really re-usable as it requires the `current_usage` to be injected

app/graphql/mutations/usage_monitoring/alerts/create.rb
```diff
@@ -0,0 +1,33 @@
...
+        type Types::UsageMonitoring::Alerts::Object
+
+        def resolve(**args)
+          result = ::UsageMonitoring::CreateAlertService.call(
+            organization: current_organization,
+            subscription: current_organization.subscriptions.find(args[:subscription_id]),
+            params: args,
+            billable_metric: args[:billable_metric_id] ?
+              current_organization.billable_metrics.find(args[:billable_metric_id]) :
+              nil
```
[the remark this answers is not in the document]
julienbourdeau: I actually had to handle a lot of things in the [rest api](https://github.com/getlago/lago-api/pull/3600) indeed!

app/models/usage_monitoring/alert.rb
julienbourdeau: Here only the current value (a number) is injected, no the `current_usage` object.

app/serializers/v1/usage_monitoring/triggered_alert_serializer.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module V1
+  module UsageMonitoring
+    class TriggeredAlertSerializer < ModelSerializer
+      def serialize
+        {
+          lago_id: model.id,
+          lago_alert_id: model.alert.id,
+          lago_subscription_id: model.subscription_id,
+          lago_billable_metric_id: model.alert.billable_metric_id,
```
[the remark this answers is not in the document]
julienbourdeau: Absolutely! I totally overlooked that

db/migrate/20250520080000_create_usage_monitoring_alerts.rb
julienbourdeau: That's what I thought first! Unfortunately, PG will consider **null as distinct values**.
For example, the following two lines are considered different because `null != null`:

| subscription_external_id | organization_id | alert_type | billable_metric_id |
|--------------------------|-----------------|------------|--------------------|
| 12                       | 1               | usage      | null               |
| 12                       | 1               | usage      | null               |

vincent-pochet: Ah yeah you are right indeed. Ok, good for me then :)

app/services/usage_monitoring/create_alert_service.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    attr_reader :organization, :subscription, :params
+
+    def billable_metric
+      @billable_metric ||= if params[:billable_metric]
+        params[:billable_metric]
+      elsif params[:billable_metric_id]
+        BillableMetric.find_by(id: params[:billable_metric_id])
+      elsif params[:billable_metric_code]
+        BillableMetric.find_by(code: params[:billable_metric_code])
```
vincent-pochet, line 62: This one should be scoped to the organization.

Also, maybe the service should return `result.not_found_error` when a billable metric reference is provided but the BM is not found

app/services/usage_monitoring/update_alert_service.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    private
+
+    attr_reader :alert, :params
+
+    def billable_metric
+      @billable_metric ||= if params[:billable_metric_id]
+        BillableMetric.find_by(id: params[:billable_metric_id])
+      elsif params[:billable_metric_code]
+        BillableMetric.find_by(code: params[:billable_metric_code])
```
vincent-pochet, line 43: Same remark around `code` lookup and not found

app/models/usage_monitoring/alert.rb
vincent-pochet: Ok I see.
Not a blocker for me, but I have the feeling that since this method is only useful in a single service as part of its "core" purpose, the code should be there

vincent-pochet (approved):

app/models/usage_monitoring/alert.rb
julienbourdeau: Noted! I'll see how it goes when I add the next two types of alerts

PR #3564, opened 2025-04-30
fix(Analytics): Check if Sentry is defined

nudded:


lovrocolic (approved):

julienbourdeau:
I had the same issue recently so I think we should wrap it into our own ErrorReporting (or something)

https://github.com/getlago/lago-api/blob/e9669ecc4125ad76bec32486c46e174bb9fd3b65/app/jobs/usage_monitoring/process_subscription_activity_job.rb#L13

PR #3565, opened 2025-04-30
chore(spec): Ensure all tables have organization_id column

julienbourdeau:
## Context

We're adding `organization_id` to all tables.

## Description

In order not to forget to add new `organization_id` to new tables, I'm adding this spec which queres the db schema.
After all tables are migrated, the expectation will be updated to `.to be_empty`.

Remove the table name every time you migrate one.

### DIFF
--- a/spec/organization_id_column_spec.rb
+++ b/spec/organization_id_column_spec.rb
@@ -1,0 +1,96 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+Rspec.describe "All tables must have an organization_id" do
+  let(:internal_tables) do
+    %w[
+      active_storage_attachments
+      active_storage_blobs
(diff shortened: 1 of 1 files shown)


spec/organization_id_column_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+Rspec.describe "All tables must have an organization_id" do
+  let(:internal_tables) do
+    %w[
+      active_storage_attachments
+      active_storage_blobs
+      active_storage_variant_records
+      ar_internal_metadata
+      schema_migrations
+      versions
```
julienbourdeau, line 13: Actually this is PaperTrail table, we could add it there too I think

nudded (approved):

PR #3566, opened 2025-04-30
Fix(multientities): fix payment receipt pdf generation

annvelents:
## Context

Fixing a misspelling in payment receipt pdf generation

### DIFF
--- a/app/views/templates/payment_receipts/v1.slim
+++ b/app/views/templates/payment_receipts/v1.slim
@@ -15,7 +15,7 @@ html
       .mb-24
         h1.invoice-title = I18n.t('payment_receipt.document_name')
         - if @billing_entity.logo.present?
-          img.header-logo src="data:#{@biling_entity.logo.content_type};base64,#{@biling_entity.base64_logo}"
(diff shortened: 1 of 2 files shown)


app/views/templates/payment_receipts/v1.slim
```diff
@@ -15,7 +15,7 @@ html
       .mb-24
         h1.invoice-title = I18n.t('payment_receipt.document_name')
         - if @billing_entity.logo.present?
-          img.header-logo src="data:#{@biling_entity.logo.content_type};base64,#{@biling_entity.base64_logo}"
+          img.header-logo src="data:#{@billing_entity.logo.content_type};base64,#{@billing_entity.base64_logo}"
```
floganz, line 18: Do we have any test for pdfs? Looks like easy to catch with tests issue.

floganz (approved):

app/views/templates/payment_receipts/v1.slim
annvelents: actually, yes, we do! I saw `stub_pdf_generation` in the service test and thought, we skip it completely, but seems it only stubs the file write.
Adding the test :)

PR #3571, opened 2025-05-01
Fix(multientities): invoice numbering - double invoices lock

annvelents:


spec/models/invoice_spec.rb
```diff
@@ -169,7 +169,7 @@
 
         expect(invoice).to be_valid
         expect(invoice.sequential_id).to eq(3)
-        expect(invoice.billing_entity_sequential_id).to eq(3)
+        # expect(invoice.billing_entity_sequential_id).to eq(3)
```
ancorcruz, line 172: Instead of commenting this assertion, could we set the billing entity numbering to per_b_e?

ancorcruz (approved):

ancorcruz:
We will fix the numbering later when replacing the orga seq id... lets avoid double counting here.

PR #3578, opened 2025-05-05
feat(Idempotency): Detect early returns and raise an error

nudded:
## Description

detect when an early return happens inside the block. If this happens, we need to raise an error!


vincent-pochet (approved):
:rocket:

PR #3580, opened 2025-05-05
fix: Fix resolver to not include pending downgraded  subscriptions

ivannovosad:
## Description

Fix the subscriptions resolver to not return the downgraded pending subscriptions.

### DIFF
--- a/app/graphql/resolvers/customers/subscriptions_resolver.rb
+++ b/app/graphql/resolvers/customers/subscriptions_resolver.rb
@@ -21,7 +21,7 @@ module Resolvers
         statuses = status
         subscriptions = object.subscriptions
 
-        return subscriptions.order(created_at: :desc) if statuses.blank?
+        return subscriptions.order(created_at: :desc).reject { |s| s.pending? && s.downgraded? } if statuses.blank?
 
         # if requested statuses do not include pending ones we should just perform filtering by given statuses
         return subscriptions.where(status: statuses).order(created_at: :desc) unless statuses&.include?("pending")
--- a/app/graphql/resolvers/customers/subscriptions_resolver.rb
+++ b/app/graphql/resolvers/customers/subscriptions_resolver.rb
@@ -21,7 +21,7 @@ module Resolvers
         statuses = status
         subscriptions = object.subscriptions
 
-        return subscriptions.order(created_at: :desc).reject { |s| s.pending? && s.downgraded? } if statuses.blank?
+        return subscriptions.order(created_at: :desc).reject { |s| s.pending? && s.previous_subscription&.downgraded? } if statuses.blank?
 
         # if requested statuses do not include pending ones we should just perform filtering by given statuses
         return subscriptions.where(status: statuses).order(created_at: :desc) unless statuses&.include?("pending")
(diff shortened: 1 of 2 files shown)


ancorcruz (approved):
LGTM.

could we add a test case to avoid regressions?

app/graphql/resolvers/customers/subscriptions_resolver.rb
```diff
@@ -21,7 +21,7 @@ def resolve(status: nil)
         statuses = status
         subscriptions = object.subscriptions
 
-        return subscriptions.order(created_at: :desc) if statuses.blank?
+        return subscriptions.order(created_at: :desc).reject { |s| s.pending? && s.downgraded? } if statuses.blank?
```
ancorcruz, line 24:
```suggestion
        return subscriptions.order(created_at: :desc).reject { |s| s.pending? && s.upgraded? } if statuses.blank?
```

spec/graphql/resolvers/customers/subscriptions_resolver_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+    customer
+  end
+
+  it_behaves_like "requires permission", "customers:view"
+
+  describe "when no status filter is provided" do
+    let!(:active_subscription) { create(:subscription, customer:, plan:, status: "active") }
+    let!(:terminated_subscription) { create(:subscription, customer:, plan:, status: "terminated") }
+    let!(:pending_subscription) { create(:subscription, customer:, plan:, status: "pending", started_at: 1.day.from_now) }
+    let!(:downgraded_pending) { create(:subscription, customer:, plan:, status: "pending") }
```
ancorcruz, line 40: we prefer `let` and referencing the objects in the before block instead of `let!`

spec/graphql/resolvers/customers/subscriptions_resolver_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+    let(:active_from_downgrade) { create(:subscription, customer:, plan:, status: "active", previous_subscription: downgraded_pending) }
```
ancorcruz, line 41:
```suggestion
    let!(:downgraded_pending) { create(:subscription, customer:, plan:, status: "pending", previous_subscription: active_from_downgrade) }
    let(:active_from_downgrade) { create(:subscription, customer:, plan:, status: "active") }
```

or use `next_subscriptions` instead...

spec/graphql/resolvers/customers/subscriptions_resolver_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+      result = execute_graphql(
+        current_user: membership.user,
+        current_organization: organization,
+        permissions: required_permission,
+        query:,
+        variables: {customerId: customer.id}
+      )
+      response = result["data"]["customer"]["subscriptions"]
+
+      aggregate_failures do
```
ancorcruz, line 57: no need to aggregate failures as we do it by default.

ancorcruz (changes requested):
there are a few style questions to fix in the spec but also the setup is wrong for the no filter scenario...

ancorcruz (approved):
LGTM thanks for adding the resolver tests

PR #3598, opened 2025-05-06
fix(multientities):  fulfil billing entities net_payment_term

annvelents:
## Context

When creating billing_entities per organization somehow we missed the net_payment_term 😱 

## Description

Added a migration to update billing_entities with the net_payment_term


ancorcruz (approved):
LGTM!

PR #3602, opened 2025-05-07
fix(multientities):  wrong document numbering

annvelents:
## Context

When creating billing entity per organization, we generate it's document number the same way as we generate it for org: first 3 letters of name + last 3 letters of id. That's why when we create a first billing_entity per organization, it inherits org's ID.
But at some point we were not populating the id of billing entity, as result generated document_number_prefix was different from the organization (the id part).
Later billing_entities ids were fixed to always match organizations, but we didn't fix the document_number_prefix (on US cluster there are 4 orgs with this issue, 0 on EU).
This PR finds these BEs and fixes them

## Description

Find default billing_entities per organizations, where billing_entity.documnet_number_prefix != organization.document_number_prefix and fix them to have org's document_number_prefix

### DIFF
--- a/db/migrate/20250507110137_fix_billing_entity_document_numbering_prefix.rb
+++ b/db/migrate/20250507110137_fix_billing_entity_document_numbering_prefix.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class FixBillingEntityDocumentNumberingPrefix < ActiveRecord::Migration[8.0]
+  class Organization < ApplicationRecord
+    has_many :billing_entities, foreign_key: :organization_id
+  end
+
+  class BillingEntity < ApplicationRecord
+    belongs_to :organization, foreign_key: :organization_id
+  end
+
+  def up
+    # Find all billing entities that need to be updated
+    billing_entities = Organization
+      .joins(:billing_entities)
+      .where(billing_entities: { archived_at: nil })
+      .where('organizations.document_number_prefix != billing_entities.document_number_prefix')
+      .where('billing_entities.id = organizations.id')
+      .select('billing_entities.id, organizations.document_number_prefix')
+
+    # Update each billing entity with its organization's document number prefix
+    billing_entities.each do |be|
+      BillingEntity.where(id: be.id).update_all(document_number_prefix: be.document_number_prefix)
+    end
+  end
+
+  def down
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20250507110137_fix_billing_entity_document_numbering_prefix.rb
```diff
@@ -0,0 +1,31 @@
...
+      .where("organizations.document_number_prefix != billing_entities.document_number_prefix")
+      .where("billing_entities.id = organizations.id")
+      .select("billing_entities.id, organizations.document_number_prefix")
+
+    # Update each billing entity with its organization's document number prefix
+    billing_entities.each do |be|
+      # rubocop:disable Rails/SkipsModelValidations
+      BillingEntity.where(id: be.id).update_all(document_number_prefix: be.document_number_prefix)
+      # rubocop:enable Rails/SkipsModelValidations
+    end
```
ancorcruz, line 26: I think this migration will leave the billing entity document number prefix as it was...

perhaps it can be done as

```suggestion
    # rubocop:disable Rails/SkipsModelValidations
    BillingEntity
      .joins(:organizations)
      .where("billing_entities.id = billing_entities.organization_id")
      .where("billing_entities.document_number_prefix != organizations.document_number_prefix")
      .update_all("billing_entities.document_number_prefix = organization.document_number_prefix")
    # rubocop:enable Rails/SkipsModelValidations
```

maybe we can even remove `      .where("billing_entities.document_number_prefix != organizations.document_number_prefix")` and update them all.

ancorcruz (changes requested):

db/migrate/20250507110137_fix_billing_entity_document_numbering_prefix.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def up
+    # rubocop:disable Rails/SkipsModelValidations
+    BillingEntity
+      .joins(:organizations)
+      .where("billing_entities.id = billing_entities.organization_id")
+      .where("billing_entities.document_number_prefix != organizations.document_number_prefix")
+      .update_all("billing_entities.document_number_prefix = organizations.document_number_prefix")
+    # rubocop:enable Rails/SkipsModelValidations
+  end
```
ancorcruz, line 20:
```suggestion
  def up
    # rubocop:disable Rails/SkipsModelValidations
    BillingEntity.unscoped.update_all(<<~SQL)
      document_number_prefix = organizations.document_number_prefix
      FROM organizations
      WHERE organizations.id = billing_entities.organization_id
          AND billing_entities.document_number_prefix != organizations.document_number_prefix
    SQL
    # rubocop:enable Rails/SkipsModelValidations
  end
```

ancorcruz (changes requested):
this should work.

ancorcruz (approved):

PR #3603, opened 2025-05-07
Fix(multientities):do not set eu_tax_management if not provided

annvelents:
## Context

when params do not include eu_tax_management, this field should be set to default => false, but because we were always calling the service that sets eu_tax_management to value from params, if value is missing, it becomes nil instead of false

## Description

Only set eu_tax_management if the value is sent in params


rsempe (approved):

julienbourdeau (approved):
One day this will be not nullable :D

PR #3604, opened 2025-05-08
feat(avalara-integration): Add void invoice flow for avalara

lovrocolic:
## Context

Lago is currently adding integration with tax provider Avalara.

## Description

This PR handles void invoice flow.

When invoice is voided in Lago it needs to be reported in the tax provider.

The only exception is when Avalara transaction is in the locked state. In that case, it is not possible to change transaction in any way, so we need to create negative transaction which will act as the refund.

### DIFF
--- a/app/jobs/invoices/provider_taxes/void_job.rb
+++ b/app/jobs/invoices/provider_taxes/void_job.rb
@@ -6,7 +6,7 @@ module Invoices
       queue_as "integrations"
 
       def perform(invoice:)
-        return unless invoice.customer.anrok_customer
+        return unless invoice.customer.tax_customer
 
         # NOTE: We don't want to raise error here.
         #       If sync fails, invoice would be marked and retry option would be available in the UI
--- a/app/services/integrations/aggregator/taxes/invoices/create_service.rb
+++ b/app/services/integrations/aggregator/taxes/invoices/create_service.rb
@@ -45,7 +45,9 @@ module Integrations
 
             invoice_data = payload_body.first
             invoice_data["id"] = invoice.id
-            invoice_data["type"] = "salesInvoice" if integration.type.to_s == "Integrations::AvalaraIntegration"
+            if integration.type.to_s == "Integrations::AvalaraIntegration"
+              invoice_data["type"] = invoice.voided? ? "returnInvoice" : "salesInvoice"
+            end
 
             [invoice_data]
           end
--- a/app/services/integrations/aggregator/taxes/invoices/payloads/avalara.rb
+++ b/app/services/integrations/aggregator/taxes/invoices/payloads/avalara.rb
@@ -76,7 +76,7 @@ module Integrations
             def item_amount(fee)
               amount = fee.sub_total_excluding_taxes_amount_cents&.to_i&.fdiv(fee.amount.currency.subunit_to_unit)
 
-              amount = amount * -1 if invoice.voided?
+              amount *= -1 if invoice.voided?
 
               amount
             end
--- a/app/services/integrations/aggregator/taxes/invoices/payloads/avalara.rb
+++ b/app/services/integrations/aggregator/taxes/invoices/payloads/avalara.rb
@@ -61,7 +61,7 @@ module Integrations
                 "item_id" => fee.id || fee.item_id,
                 "item_code" => mapped_item.external_id,
                 "unit" => fee.units,
-                "amount" => fee.sub_total_excluding_taxes_amount_cents&.to_i&.fdiv(fee.amount.currency.subunit_to_unit)
+                "amount" => item_amount(fee)
               }
             end
 
@@ -71,6 +71,14 @@ module Integrations
 
             def empty_struct
               @empty_struct ||= OpenStruct.new
+            end
+
+            def item_amount(fee)
+              amount = fee.sub_total_excluding_taxes_amount_cents&.to_i&.fdiv(fee.amount.currency.subunit_to_unit)
+
+              amount = amount * -1 if invoice.voided?
+
+              amount
             end
           end
         end
--- a/app/services/integrations/aggregator/taxes/invoices/void_service.rb
+++ b/app/services/integrations/aggregator/taxes/invoices/void_service.rb
@@ -31,16 +31,17 @@ module Integrations
           private
 
           def payload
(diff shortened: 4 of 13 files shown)


app/jobs/invoices/provider_taxes/void_job.rb
```diff
@@ -6,7 +6,7 @@ class VoidJob < ApplicationJob
       queue_as "integrations"
 
       def perform(invoice:)
-        return unless invoice.customer.anrok_customer
+        return unless invoice.customer.tax_customer
```
ancorcruz, line 9: could we also cover this into the job spec? right it has two scenarios anrok customer, not anrok customer. At least update the description of the spec to let the reader know it is not only anrok customer but tax customer...

spec/services/integrations/aggregator/taxes/invoices/create_service_spec.rb
```diff
@@ -217,6 +217,61 @@
...
+            ]
+          end
+
+          before { invoice.voided! }
+
+          it "returns fees for valid request payload" do
+            result = service_call
+
+            expect(result).to be_success
+          end
```
ancorcruz, line 273: should this one also check the returned fees not only the result was a success?

app/services/integrations/aggregator/taxes/invoices/create_service.rb
```diff
@@ -45,7 +45,9 @@ def payload
 
             invoice_data = payload_body.first
             invoice_data["id"] = invoice.id
-            invoice_data["type"] = "salesInvoice" if integration.type.to_s == "Integrations::AvalaraIntegration"
+            if integration.type.to_s == "Integrations::AvalaraIntegration"
+              invoice_data["type"] = invoice.voided? ? "returnInvoice" : "salesInvoice"
```
ancorcruz, line 49: this branching is not covered in the service test

app/services/integrations/aggregator/taxes/invoices/payloads/avalara.rb
```diff
@@ -72,6 +72,14 @@ def fee_item(fee)
             def empty_struct
               @empty_struct ||= OpenStruct.new
             end
+
+            def item_amount(fee)
+              amount = fee.sub_total_excluding_taxes_amount_cents&.to_i&.fdiv(fee.amount.currency.subunit_to_unit)
+
+              amount *= -1 if invoice.voided?
```
ancorcruz, line 79:
```suggestion
              amount &&= -amount if invoice.voided?
```

or just this if amount always present....

`amount = -amount if invoice.voided?`

is this any clearer?

app/services/integrations/aggregator/taxes/invoices/void_service.rb
```diff
@@ -31,11 +31,20 @@ def call
...
+                }
+              ]
+            else
+              [
+                {
+                  "company_code" => integration.company_code,
+                  "id" => invoice.id
+                }
+              ]
+            end
```
ancorcruz, line 47: is this a general case of does it only applies to Avalara? if only applies to Avalara I would use a case statement instead of if/else otherwise it might require more work adding a 3rd aggregator integration.

spec/services/integrations/aggregator/taxes/invoices/void_service_spec.rb
```diff
@@ -78,6 +78,45 @@
...
+
+        before do
+          allow(lago_client).to receive(:post_with_response).with(params, headers).and_return(response)
+          allow(response).to receive(:body).and_return(body)
+        end
+
+        it "returns invoice_id" do
+          result = service_call
+
+          aggregate_failures do
```
ancorcruz, line 113: `aggregate_failures` is not required as we do it by default in all tests.

spec/services/invoices/provider_taxes/void_service_spec.rb
```diff
@@ -109,6 +109,19 @@
       end
     end
 
+    context "when invoice is not voided" do
+      before { invoice.finalized! }
+
+      it "returns an error" do
+        result = void_service.call
+
+        aggregate_failures do
```
ancorcruz, line 118: same here, this is not required anymore.

spec/services/invoices/provider_taxes/void_service_spec.rb
```diff
@@ -190,5 +203,44 @@
...
+
+      it "resolves old tax error and creates new one" do
+        old_error_id = invoice.reload.error_details.last.id
+
+        void_service.call
+
+        aggregate_failures do
+          expect(invoice.error_details.tax_voiding_error.last.id).not_to eql(old_error_id)
+          expect(invoice.error_details.tax_voiding_error.count).to be(1)
+          expect(invoice.error_details.tax_voiding_error.order(created_at: :asc).last.discarded?).to be(false)
```
ancorcruz, line 241:
```suggestion
          expect(invoice.error_details.tax_voiding_error.order(created_at: :asc).last).not_to be_discarded
```

this can be more idiomatic.

ancorcruz (approved):
Looks great.

I just left a few comments for your consideration.

app/jobs/invoices/provider_taxes/void_job.rb
lovrocolic: added one more test case 👍

app/services/integrations/aggregator/taxes/invoices/create_service.rb
lovrocolic: This is private method so I haven't tested it directly, but in the spec I am providing mocked response based on expected payload (lines 156 and 226 in spec) and there are two different expected payloads, one for each case

spec/services/integrations/aggregator/taxes/invoices/void_service_spec.rb
lovrocolic: Updated

spec/services/invoices/provider_taxes/void_service_spec.rb
lovrocolic: Updated

spec/services/invoices/provider_taxes/void_service_spec.rb
lovrocolic: Updated

PR #3605, opened 2025-05-08
Feat(multientities): return billing_entity_sequential_id to invoices

annvelents:
## Context

In order to switch to invoice numbering generation via billing_entities, invoices should have billing_entity_sequential_id populated. that's why we're updating:
1) when generating invoice's organization_sequential_id, put the same value to the billing_entity_sequential_id
2) when switching on org from per_customer to per_organization, fulfil billing_entity_sequential_id as well as organization_sequential_id

## Description

- Added job to fulfil billing_entity_sequential_id on invoices
- updated sequential_id generation to also fill billing_entity_sequential_id with the generated value
- updated service responsible for sequential_id generation when switching document numbering on organization


app/jobs/database_migrations/populate_invoices_billing_entity_sequential_id_job.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class PopulateInvoicesBillingEntitySequentialIdJob < ApplicationJob
+  queue_as :low_priority
+  unique :until_executed
+
+  BATCH_SIZE = 1000
+
+  def perform(batch_number = 1)
+    batch = Invoice.where(billing_entity_sequential_id: nil).where.not(organization_sequential_id: nil).limit(BATCH_SIZE)
```
ancorcruz, line 10:
```suggestion
    batch = Invoice
      .where(billing_entity_sequential_id != organization_sequential_id)
      .where(organization_sequential_id != 0)
      .limit(BATCH_SIZE)
```

PR #3616, opened 2025-05-12
Fix(multientities): return be sequential id

annvelents:
## Context

In order to switch to invoice numbering generation via billing_entities, invoices should have billing_entity_sequential_id populated. that's why we're updating:
1) when generating invoice's organization_sequential_id, put the same value to the billing_entity_sequential_id
2) when switching on org from per_customer to per_organization, fulfil billing_entity_sequential_id as well as organization_sequential_id

## Description

- Added job to fulfil billing_entity_sequential_id on invoices
- updated sequential_id generation to also fill billing_entity_sequential_id with the generated value
- updated service responsible for sequential_id generation when switching document numbering on organization


ancorcruz (approved):
LGTM

PR #3641, opened 2025-05-14
feat: Add webhook resolver

rsempe:
The goal of this PR is to introduce a WebhookResolver.
It's useful for the frontend in order to select a specific webhook.

### DIFF
--- a/app/graphql/resolvers/webhook_resolver.rb
+++ b/app/graphql/resolvers/webhook_resolver.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Resolvers
+  class WebhookResolver < Resolvers::BaseResolver
+    include AuthenticableApiUser
+    include RequiredOrganization
+
+    REQUIRED_PERMISSION = "developers:manage"
+
+    description "Query a webhook"
+
+    argument :id, ID, required: true
+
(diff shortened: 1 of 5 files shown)


app/graphql/resolvers/webhook_resolver.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    description "Query a webhook"
+
+    argument :id, ID, required: true
+
+    type Types::Webhooks::Object, null: true
+
+    def resolve(id:)
+      webhook = current_organization.webhook_endpoints.joins(:webhooks)
+        .find_by(webhooks: {id:})&.webhooks&.find(id)
```
julienbourdeau, line 18: I'm so confused by this query. The `find_by.webhooks.find` is brain fucking me.

I was expecthing this:
```suggestion
      webhook = current_organization.webhooks.find_by(id:)
```

julienbourdeau (approved):

app/graphql/resolvers/webhook_resolver.rb
mariohd: I have the same question

rsempe: good spot thanks

PR #3643, opened 2025-05-14
feat(stripe): Mark invoice as disputed for PaymentRequest

julienbourdeau:
A payment has `payable` which can be an Invoice or a `PaymentRequest`.

When a dispute is lost, Stripe sends a `charge.dispute.closed` webhook with the payment details included. When attached to an invoice, the invoice is flagged as disputed but it it was a PaymentRequest, it wasn't handled.

This PR loops over each invoices of the PaymentRequest and mark them as disputed.

The Adyen service was also updated and since the test don't mock the service call, but check the result, they don't need to be updated.

Note that the original `Invoices::LoseDisputeService` is still [used by GraphQL](https://github.com/getlago/lago-api/blob/b99e8a592fa2d736ce772ff0abff5257fc879543/app/graphql/mutations/invoices/lose_dispute.rb#L19) when marking an invoice as last.

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -49,6 +49,16 @@ class Payment < ApplicationRecord
       .where("invoices.id IS NOT NULL OR payment_requests.id IS NOT NULL")
   }
 
+  def invoices
+    if payable.is_a?(Invoice)
+      [payable]
+    elsif payable.is_a?(PaymentRequest)
+      payable.invoices.to_a
+    else
+      []
+    end
+  end
+
   def should_sync_payment?
     return false unless payable.is_a?(Invoice)
 
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -50,13 +50,7 @@ class Payment < ApplicationRecord
   }
 
   def invoices
-    if payable.is_a?(Invoice)
-      [payable]
-    elsif payable.is_a?(PaymentRequest)
-      payable.invoices.to_a
-    else
-      []
-    end
+    payable.payment_invoices
   end
 
   def should_sync_payment?
--- a/spec/factories/payment_requests.rb
+++ b/spec/factories/payment_requests.rb
@@ -11,5 +11,15 @@ FactoryBot.define do
     payment_status { "pending" }
     ready_for_payment_processing { true }
     payment_attempts { 0 }
+
+    transient do
+      invoices { [] }
+    end
+
+    after(:create) do |payment_request, evaluator|
+      evaluator.invoices.each do |invoice|
+        PaymentRequest::AppliedInvoice.create!(payment_request:, invoice:)
+      end
+    end
   end
 end
--- a/spec/factories/payment_requests.rb
+++ b/spec/factories/payment_requests.rb
@@ -11,15 +11,5 @@ FactoryBot.define do
     payment_status { "pending" }
     ready_for_payment_processing { true }
     payment_attempts { 0 }
-
-    transient do
-      invoices { [] }
-    end
-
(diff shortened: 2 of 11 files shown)


app/models/payment.rb
```diff
@@ -49,6 +49,16 @@ class Payment < ApplicationRecord
...
 
+  def invoices
+    if payable.is_a?(Invoice)
+      [payable]
+    elsif payable.is_a?(PaymentRequest)
+      payable.invoices.to_a
+    else
+      []
+    end
+  end
```
julienbourdeau, line 60: How do you feel about this?

ancorcruz: This is returning an array, and checking payable type.

What do you think about extracting the "ActAsPayable" like role into a module.

This module could define the relationship with payment and a "payment_invoices" (or whatever name works for you). It would by default raise a not implemented exception.

Then in each model including this module will implement that method ....

for invoices it could be

```
def payment_invoices
  Invoice.where(id: id)
end
```

for payment requests just an alias of `invoices` relationship
```
def payment_invoices
  invoices
end
```

Then, if it is a payable must implement this method.

This way we avoid type checking, group the behavior of payable objects and the invoices remains an active record relation allowing using scopes, and other framework nice stuff...

spec/factories/payment_requests.rb
```diff
@@ -11,5 +11,15 @@
...
+
+    transient do
+      invoices { [] }
+    end
+
+    after(:create) do |payment_request, evaluator|
+      evaluator.invoices.each do |invoice|
+        PaymentRequest::AppliedInvoice.create!(payment_request:, invoice:)
+      end
+    end
```
ancorcruz, line 23: doesn't rails automatically create the applied invoice join records when assigning invoices to the payment request record?

spec/services/payments/lose_dispute_service_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  subject(:lose_dispute_service) { described_class.new(payment:) }
+
+  describe "#call" do
+    context "when payment does not exist" do
+      let(:payment) { nil }
+
+      it "returns a failure" do
+        result = lose_dispute_service.call
+
+        aggregate_failures do
```
ancorcruz, line 15: We already aggregate_failures by default, not needed to add it explicitly.

spec/services/payments/lose_dispute_service_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+          expect(result.error).to be_a(BaseService::NotFoundFailure)
+          expect(result.error.resource).to eq("payment")
+        end
+      end
+
+      it "does not enqueue a send webhook job for the invoice" do
+        expect { lose_dispute_service.call }.not_to have_enqueued_job(SendWebhookJob)
+      end
+    end
+
```
ancorcruz, line 27: I miss one context for when the payable does not exist. Is it even possible? the guard clause is there...

ancorcruz (dismissed):
Looks great, still I would improve the `invoices` method in payment model

spec/services/payments/lose_dispute_service_spec.rb
julienbourdeau: forgot to rmeove!

app/models/payment.rb
julienbourdeau: Love it ! thank you for the suggestion 👌 

CleanShot 2025-05-23 at 10 33 55@2x

ancorcruz (approved):
LGTM! thanks

PR #3645, opened 2025-05-15
misc: avoid ActiveRecord::PreparedStatementCacheExpired errors

vincent-pochet:
## Context

In some context, when a migration is adding a new column to a table, Postgres might return a `cached plan must not change result type` in a transaction because the cached query plan is not more valid. This error is then raised as a `ActiveRecord::PreparedStatementCacheExpired` by Rails

See https://flexport.engineering/avoiding-activerecord-preparedstatementcacheexpired-errors-4499a4f961cf

## Description

This fix is forcing rails to query explicitly every columns in select instead of using `SELECT *`, so that the cached plan remains valid in the transaction. The only drawback will be with column removal, but it should never happen on a used field and would be prevented by `strong_migrations`

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -2,4 +2,8 @@
 
 class ApplicationRecord < ActiveRecord::Base
   primary_abstract_class
+
+  # Avoid raising ActiveRecord::PreparedStatementCacheExpired
+  # from transactions when a migration is adding a new column
+  self.ignored_columns = [:__fake_column__]
 end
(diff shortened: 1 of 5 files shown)


annvelents (approved):

julienbourdeau:
> The only drawback will be with column removal

Before deleting a column, it will live in `ignored_columns` for a long time, so we're good.

For future reference: https://github.com/getlago/lago-api/pull/3640

julienbourdeau (approved):

PR #3646, opened 2025-05-15
Feat(multientities): add standard algorithm of managing the taxes

annvelents:
## Context

when applying taxes on a customer or on a plan, they are applied via /update endpoint. Applying taxes on billing_entity should follow the same principle instead of managing taxes separately

## Description

- added tax_management into billing_entity/update service
- removed manage_taxes endpoint


rsempe (approved):
🙌

PR #3649, opened 2025-05-15
fix(netsuite): Use max length when creating netsuite customer

rsempe:
## Context

We currently receive deadjobs when trying to sync customers on Netsuite.
This is because `state` sent contains more characters than the maximum size allowed by Netsuite (`30`).

Deadjob:
> action_script_runtime_error: The field addr1 contained more than the maximum number ( 150 ) of characters allowed.

## Description

The goal of this PR is to only select the first 30 characters for the firstname and lastname when creating a customer on Netsuite.


julienbourdeau (approved):
beautiful

PR #3654, opened 2025-05-15
feat(webhook): email admins in case of failed webhooks

julienbourdeau:
Lago sends many webhooks to the user's system but sometimes the webhook fails.
We retry the webhook 3 times, then it dies.

This PR introduce a new email sent to all admins when a webhook fails to be delivered. Usually, it means there is something wrong on the receiving end.

This email is sent as soon as one webhook dies but is not sent again for an hour. If the receiving endpoint is down for a few minutes, you don't get flooded!

~**The email is text only, not HTML.** This is an admin email that you should very rarely receive. It's easier than HTML.~ It's HTML.

CleanShot 2025-05-21 at 17 15 07@2x


julienbourdeau:
Add test to ensure that email is set even if one admin has an invalid email address

app/services/webhooks/send_http_service.rb
```diff
@@ -28,10 +28,11 @@ def call
       EOFError => e
       mark_webhook_as_failed(e)
 
-      # NOTE: By default, Lago is retrying 3 times a webhook
-      return if webhook.retries >= ENV.fetch("LAGO_WEBHOOK_ATTEMPTS", 3).to_i
-
-      SendHttpWebhookJob.set(wait: wait_value).perform_later(webhook)
+      if webhook.retries >= ENV.fetch("LAGO_WEBHOOK_ATTEMPTS", 3).to_i
```
floganz, line 31: 🤔 Is it possible to fail webhook more than 3 times? Because condition written in a style like it's possible.
Does it mean that we'll send several emails?

app/services/webhooks/send_http_service.rb
floganz, line 1: Are you planning to add test for changes in this file?

spec/services/webhooks/notify_failure_service_spec.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Webhooks::NotifyFailureService, type: :service do
+  let(:webhook) { create(:webhook, status: :failed) }
+  let(:service) { described_class.new(webhook: webhook) }
+  let(:cache_key) { "webhook_failure_notification:#{webhook.organization_id}" }
+
+  describe "#call" do
```
floganz, line 10: If you already tests in separation `should_notify?` method I think it's totally OK to reduce context to 2 "when should notify" and "when should not notify". Of course it's up to you to decide.

app/mailers/webhook_mailer.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class WebhookMailer < ApplicationMailer
+  def failure_notification
+    @webhook = params[:webhook]
+    @org_name = @webhook.organization.name
+    @front_webhooks_url = "#{ENV["LAGO_FRONT_URL"]}/developers/webhooks"
+    @failed_webhooks_count = @webhook.organization.webhooks.failed.where(last_retried_at: 1.hour.ago..).count
+
+    #  NOTE: Members can have avec CSV string with emails, just like organization.email
```
floganz: `avec` is that French word?

spec/mailers/webhook_mailer_spec.rb
floganz, line 1: I think tests are fine. Want to bring some ideas for test structure to keep them more organized if you like it.
```ruby
describe "#failure_notification"
  describe "#to"
    subject { mail.to }
    # all test cases related to who will receive email
  describe "#body" or "#content"
    subject { mail.body.encoded }
    # related contexts or single `it` for email content expectations
  describe "delivery"
    # test under what conditions email is delivered or not
  describe "#subject"
    # to test email subject if it has it's own logic. Probably in your case easier to keep under "#body" describe".
```

floganz (approved):

app/mailers/webhook_mailer.rb
julienbourdeau: lol yes 😄 but it doesn't even make sense here.

app/services/webhooks/send_http_service.rb
julienbourdeau: It's possible because you can manually retry the webhook in the UI. The email will be sent once if it wasn't sent within the last hour.

spec/services/webhooks/notify_failure_service_spec.rb
julienbourdeau: Indeed! Thank you

spec/mailers/webhook_mailer_spec.rb
julienbourdeau: I simplified the file with your suggestion but I didn't go as far

app/services/webhooks/send_http_service.rb
julienbourdeau: I didn't realize I didn't. Done!

PR #3655, opened 2025-05-15
feat(avalara-integration): add support for multiple integration types as GQL filter

lovrocolic:
## Context

Currently integration with avalara tax provider is being implemented

## Description

This change allows FE team to easier fetch all integrations filtered by certain type

### DIFF
--- a/app/graphql/resolvers/integrations_resolver.rb
+++ b/app/graphql/resolvers/integrations_resolver.rb
@@ -11,41 +11,20 @@ module Resolvers
 
     argument :limit, Integer, required: false
     argument :page, Integer, required: false
-    argument :type, [Types::Integrations::IntegrationTypeEnum], required: false
+    argument :types, [Types::Integrations::IntegrationTypeEnum], required: false
 
     type Types::Integrations::Object.collection_type, null: true
 
-    def resolve(type: nil, page: nil, limit: nil)
+    def resolve(types: nil, page: nil, limit: nil)
       scope = current_organization.integrations.page(page).per(limit)
-      scope = scope.where(type: types(type)) if type.present?
+      scope = scope.where(type: types(types)) if types.present?
       scope
     end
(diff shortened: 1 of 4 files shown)


app/graphql/resolvers/integrations_resolver.rb
```diff
@@ -11,18 +11,22 @@ class IntegrationsResolver < Resolvers::BaseResolver
...
       scope
     end
 
     private
 
+    def types(input)
+      input.map { |type| integration_type(type) }
+    end
+
     def integration_type(type)
```
floganz, line 30: I understand that it's not part of this PR but this case statement is exact copy of one in `Integrations::BaseIntegration`.
Can we maybe replace this copy with re-using existing one?
```ruby
input.map { |type| Integrations::BaseIntegration.integration_type(type) }
```

app/graphql/resolvers/integrations_resolver.rb
```diff
@@ -11,18 +11,22 @@ class IntegrationsResolver < Resolvers::BaseResolver
 
     argument :limit, Integer, required: false
     argument :page, Integer, required: false
-    argument :type, Types::Integrations::IntegrationTypeEnum, required: false
+    argument :type, [Types::Integrations::IntegrationTypeEnum], required: false
```
floganz, line 14: Another question, since this argument now represents array of values, does it makes sense to rename it to plural `types`?

lovrocolic: synced with Alex and we changed it to types

app/graphql/resolvers/integrations_resolver.rb
lovrocolic: updated

floganz (approved):

ansmonjol (approved):

PR #3681, opened 2025-05-20
feat(pricing_group_keys): Add new `pricing_group_keys` API attribute

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/3666 and is part of an epic to add grouped_by feature on all charge models and not only on the standard one like today. 

## Description

This PR is the second step:
- It updates the GraphQL and the Rest APIs to accept and return the new `pricing_group_keys` attribute
- It makes sure that the deprecated `grouped_by` attribute is still accepted and returned
- It adapt the aggregation logic to use the `pricing_group_keys` or the `grouped_by` attributes

For now the grouped aggregation remains remains whitelisted only on the standard and dynamic charge model. This will remain like this until the logic is applied to all charge model

### DIFF
--- a/app/serializers/v1/charge_filter_serializer.rb
+++ b/app/serializers/v1/charge_filter_serializer.rb
@@ -5,9 +5,25 @@ module V1
     def serialize
       {
         invoice_display_name: model.invoice_display_name,
-        properties: model.properties,
+        properties:,
         values: model.to_h
       }
+    end
+
+    private
+
+    # TODO(pricing_group_keys): remove after deprecation of grouped_by
+    def properties
+      attributes = model.properties
+      if attributes["grouped_by"].present? && attributes["pricing_group_keys"].blank?
+        attributes["pricing_group_keys"] = attributes["grouped_by"]
+      end
+
+      if attributes["pricing_group_keys"].present? && attributes["grouped_by"].blank?
+        attributes["grouped_by"] = attributes["pricing_group_keys"]
+      end
+
+      attributes
     end
   end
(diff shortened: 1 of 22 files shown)


annvelents (approved):

app/serializers/v1/charge_filter_serializer.rb
```diff
@@ -5,9 +5,25 @@ class ChargeFilterSerializer < ModelSerializer
...
+    # TODO(pricing_group_keys): remove after deprecation of grouped_by
+    def properties
+      attributes = model.properties
+      if attributes["grouped_by"].present? && attributes["pricing_group_keys"].blank?
+        attributes["pricing_group_keys"] = attributes["grouped_by"]
+      end
+
+      if attributes["pricing_group_keys"].present? && attributes["grouped_by"].blank?
+        attributes["grouped_by"] = attributes["pricing_group_keys"]
+      end
```
julienbourdeau, line 24: I'm trying to think of a case where you want `grouped_by` to hold a different value from `princing_group_keys`. Maybe we should for `grouped_by` value as soon as `princing_group_keys` is set

```suggestion
      if attributes["pricing_group_keys"].present?
        attributes["grouped_by"] = attributes["pricing_group_keys"]
      end
```

Similarily, we could add a check in the API so only one of the 2 values can be set and reject calls with both `grouped_by` and `pricing_group_keys`. It's probably not intentional to send both.

julienbourdeau (approved):

PR #3682, opened 2025-05-20
feat(avalara-integration): small avalara improvements and enabling feature

lovrocolic:
## Context

Currently Lgao is working on integration with tax provider avalara.

## Description

This PR enables avalara feature and adds small improvements

### DIFF
--- a/app/models/integration_customers/base_customer.rb
+++ b/app/models/integration_customers/base_customer.rb
@@ -17,10 +17,14 @@ module IntegrationCustomers
     ].freeze
 
     validates :customer_id, uniqueness: {scope: :type}
-    validate :only_one_tax_integration_per_customer, if: :tax_integration?
+    validate :only_one_tax_integration_per_customer, if: :tax_kind?
 
     scope :accounting_kind, -> do
       where(type: %w[IntegrationCustomers::NetsuiteCustomer IntegrationCustomers::XeroCustomer])
+    end
+
+    scope :tax_kind, -> do
+      where(type: TAX_INTEGRATION_TYPES)
     end
 
     scope :hubspot_kind, -> do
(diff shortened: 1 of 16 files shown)


app/models/integration_customers/base_customer.rb
```diff
@@ -47,6 +53,21 @@ def self.customer_type(type)
...
+
+    private
+
+    def only_one_tax_integration_per_customer
+      conflict = IntegrationCustomers::BaseCustomer.where(customer_id:, type: TAX_INTEGRATION_TYPES)
+      conflict = conflict.where.not(id:) if persisted?
+
+      return unless conflict.exists?
+
+      errors.add(:type, "tax_integration_exists")
```
floganz, line 69: Let's split it in smaller parts?

We can extract `.where(type: TAX_INTEGRATION_TYPES)` into scope `tax_kind` similar to already existing `accounting_kind`.
`where.not(id:)` could be replaced with `.excluding(self)`
and finally we can combine `.where(customer_id:` with `exists?` -> `.exists?(customer_id:)`.

```suggestion
      if IntegrationCustomers::BaseCustomer.tax_kind.excluding(self).exists?(customer_id:)
        errors.add(:type, "tax_integration_exists")
      end
```

Also it looks like tests for this validation does not check for this `conflict = conflict.where.not(id:) if persisted?` case. If you comment that line all tests will be green.

app/services/integrations/aggregator/taxes/base_service.rb
```diff
@@ -20,7 +20,7 @@ def integration
         def integration_customer
           @integration_customer ||= begin
             int_customers = customer.integration_customers
-            int_customers.find { |ic| %w[IntegrationCustomers::AnrokCustomer IntegrationCustomers::AvalaraCustomer].include?(ic.type) }
+            int_customers.find { |ic| ic.tax_integration? }
```
floganz, line 23: That could be even shorter if you like.

```suggestion
            int_customers.find(&:tax_integration?)
```

app/models/integration_customers/base_customer.rb
lovrocolic: added new scope but left old style which is more close to me:)

app/models/integration_customers/base_customer.rb
```diff
@@ -11,12 +11,22 @@ class BaseCustomer < ApplicationRecord
...
+    ].freeze
+
     validates :customer_id, uniqueness: {scope: :type}
+    validate :only_one_tax_integration_per_customer, if: :tax_kind?
 
     scope :accounting_kind, -> do
       where(type: %w[IntegrationCustomers::NetsuiteCustomer IntegrationCustomers::XeroCustomer])
     end
 
+    scope :tax_kind, -> do
```
floganz, line 26: Are you planning to add tests for this scope?

floganz (approved):
I've also noticed that there're some linter warnings from CI. Have you checked if that something we can fix?

PR #3689, opened 2025-05-20
feat: Fallback to organization country for EU bank transfers

brunomiguelpinto:
## Context

Stripe only supports a limited list of countries when creating `eu_bank_transfer` funding instructions. If a customer's country is not in that list, the request fails with a 400 error.

This feature request aims to allow accepting bank transfers from any European country by defaulting to the organization if that country is supported.

## Description

- A fallback mechanism was added when creating Stripe funding instructions for the type `eu_bank_transfer`.
- If the customer's country is unsupported, the system checks whether the organization's country is supported.

### DIFF
--- a/app/services/payment_provider_customers/stripe/sync_funding_instructions_service.rb
+++ b/app/services/payment_provider_customers/stripe/sync_funding_instructions_service.rb
@@ -86,7 +86,7 @@ module PaymentProviderCustomers
 
       def eu_bank_transfer_payload
         customer_country = customer.country&.upcase
-        organization_country = customer.organization.country&.upcase
+        organization_country = customer.billing_entity.country&.upcase
 
         country =
           if SUPPORTED_EU_BANK_TRANSFER_COUNTRIES.include?(customer_country)
@@ -94,10 +94,10 @@ module PaymentProviderCustomers
           elsif SUPPORTED_EU_BANK_TRANSFER_COUNTRIES.include?(organization_country)
             organization_country
           else
-            return result.service_failure!(
+            result.service_failure!(
               code: "missing_country",
               message: "No country found for customer or organization supported for EU bank transfer payload"
-            )
+            ).raise_if_error!
           end
 
         {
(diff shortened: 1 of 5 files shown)


annvelents (approved):

app/services/payment_providers/stripe/payments/create_service.rb
```diff
@@ -4,6 +4,8 @@ module PaymentProviders
   module Stripe
     module Payments
       class CreateService < BaseService
+        SUPPORTED_EU_BANK_TRANSFER_COUNTRIES = %w[BE DE ES FR IE NL].freeze
```
julienbourdeau, line 7: I'd rather move this constant somewhere with other stripe configuration so it's not duplicated and scattered. Maybe  the best candidate is `PaymentProviders::StripeProvider`.

brunomiguelpinto: Yes that's a good suggestion I will do it

app/services/payment_provider_customers/stripe/sync_funding_instructions_service.rb
```diff
@@ -84,16 +85,25 @@ def funding_type_payload
       end
 
       def eu_bank_transfer_payload
-        customer_country = customer.country || customer.organization.country
+        customer_country = customer.country&.upcase
+        organization_country = customer.billing_entity.country&.upcase
```
brunomiguelpinto, line 89: change this

app/services/payment_providers/stripe/payments/create_service.rb
```diff
@@ -177,8 +179,25 @@ def bank_transfer_type
         end
 
         def handle_eu_bank_transfer
-          customer_country = payment.customer.country.upcase
-          {type: "eu_bank_transfer", eu_bank_transfer: {country: customer_country}}
+          customer_country = payment.customer.country&.upcase
+          organization_country = payment.customer.billing_entity.country&.upcase
```
brunomiguelpinto, line 183: change this

PR #3509, opened 2025-04-17
Misc past usage openstruct

vincent-pochet:
## Context

This PR is part of the global initiative to remove `OpenStruct` from the codebase

## Description

This PR replaces takes advance of the `Data` object to replace `OpenStruct` definitions in `PastUsageQuery` and `Invoices::CustomerUsageService` with proper object definitions

PR #3690, opened 2025-05-21
feat(alerting): Add LifetimeUsageAmount and BillableMetricUsageUnits alerts

julienbourdeau:
## Context

We're adding Alerting feature, see #3528, #3535, #3554, #3600

## Description

We introducing 2 new types of alerts:
* LifetimeUsageAmount: to define alert based on lifetime total usage
* BillableMetricUsageUnits: same as BillableMetricUsageAmount but with a number "used units" instead of currency amount.

Creating a new alert is essentially creating a new child class of `UsageMonitoring::Alert` and implementing `find_value`. 🙌

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -143,6 +143,10 @@ class Organization < ApplicationRecord
     end
   end
 
+  def using_lifetime_usage?
+    lifetime_usage_enabled? || progressive_billing_enabled?
+  end
+
   def admins
     users.joins(:memberships).merge!(memberships.admin)
   end
--- a/app/services/usage_monitoring/process_subscription_activity_service.rb
+++ b/app/services/usage_monitoring/process_subscription_activity_service.rb
@@ -38,11 +38,11 @@ module UsageMonitoring
           organization_id: subscription_activity.organization_id
         ).includes(:thresholds)
 
-        alerts.where(alert_type: Alert::CURRENT_USAGE_TYPES).find_each do |alert|
+        alerts.using_current_usage.find_each do |alert|
           ProcessAlertService.call(alert:, subscription:, current_metrics: current_usage)
         end
 
-        alerts.where(alert_type: "lifetime_usage_amount").find_each do |alert|
+        alerts.using_lifetime_usage.find_each do |alert|
           ProcessAlertService.call(alert:, subscription:, current_metrics: lifetime_usage)
         end
       rescue => e
--- a/app/services/usage_monitoring/process_subscription_activity_service.rb
+++ b/app/services/usage_monitoring/process_subscription_activity_service.rb
@@ -21,7 +21,7 @@ module UsageMonitoring
       #       We believe it's a good tradeoff because they should rarely raise but this can change in the future
 
       begin
-        if organization.lifetime_usage_enabled? || organization.progressive_billing_enabled?
+        if organization.using_lifetime_usage?
           LifetimeUsages::CalculateService.call!(lifetime_usage:, current_usage:)
         end
 
@@ -33,12 +33,17 @@ module UsageMonitoring
       end
 
       begin
-        Alert.where(
+        alerts = Alert.where(
           subscription_external_id: subscription.external_id,
-          organization_id: subscription_activity.organization_id,
-          alert_type: Alert::CURRENT_USAGE_TYPES
-        ).includes(:thresholds).find_each do |alert|
+          organization_id: subscription_activity.organization_id
+        ).includes(:thresholds)
+
+        alerts.where(alert_type: Alert::CURRENT_USAGE_TYPES).find_each do |alert|
           ProcessAlertService.call(alert:, subscription:, current_metrics: current_usage)
+        end
+
+        alerts.where(alert_type: "lifetime_usage_amount").find_each do |alert|
+          ProcessAlertService.call(alert:, subscription:, current_metrics: lifetime_usage)
         end
       rescue => e
         # If progressive billing already raised, we don't override the first error
--- a/app/services/usage_monitoring/update_alert_service.rb
+++ b/app/services/usage_monitoring/update_alert_service.rb
@@ -15,6 +15,10 @@ module UsageMonitoring
     def call
       return result.not_found_failure!(resource: "alert") unless alert
 
+      if params.has_key?(:thresholds) && params[:thresholds].size > AlertThreshold::SOFT_LIMIT
+        return result.single_validation_failure!(field: :thresholds, error_code: "too_many_thresholds")
+      end
+
       result.alert = alert
 
       billable_metric = find_billable_metric_from_params!
--- a/spec/models/usage_monitoring/billable_metric_usage_units_alert_spec.rb
+++ b/spec/models/usage_monitoring/billable_metric_usage_units_alert_spec.rb
@@ -3,21 +3,22 @@
 require "rails_helper"
 
 RSpec.describe UsageMonitoring::BillableMetricUsageUnitsAlert, type: :model do
-  let(:alert) { create(:billable_metric_usage_units_alert) }
-  let(:subscription) { create(:subscription) }
-  let(:charge) { create(:standard_charge, plan: subscription.plan, billable_metric: alert.billable_metric) }
+  subject { alert.find_value(current_usage) }
+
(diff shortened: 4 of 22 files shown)


spec/support/scenarios_helper.rb
```diff
@@ -207,6 +207,7 @@ def perform_billing
     Clock::SubscriptionsBillerJob.perform_later
     Clock::FreeTrialSubscriptionsBillerJob.perform_later
     perform_all_enqueued_jobs
+    perform_usage_update
```
julienbourdeau, line 210: I'm adding this perform_billing because creating invoices influence `lifetime_usage.invoiced_usage_amount_cents` so lifetime_usage should be recomputed. We should keep scenarios as close as possible to real life.

app/models/organization.rb
```diff
@@ -143,6 +143,10 @@ class Organization < ApplicationRecord
     end
   end
 
+  def using_lifetime_usage?
+    lifetime_usage_enabled? || progressive_billing_enabled?
```
julienbourdeau, line 147: Unfortunately, you can have `progressive_billing` integration without `lifetime_usage`. That's why I'm introducing this new helper. What do you think of the name?

vincent-pochet (approved):
:rocket:

spec/models/usage_monitoring/billable_metric_usage_units_alert_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    ]
+  end
+
+  describe "#find_value" do
+    it do
+      charge
+      current_usage = double(amount_cents: 100, fees:) # rubocop:disable RSpec/VerifiedDoubles
+      expect(alert.find_value(current_usage)).to eq(8)
+    end
+  end
```
floganz, line 22: Current data setup does not catch all the cases.
E.g. Test will be green if you remove this part at all `.select { |fee| matching_charge_ids.include? fee.charge_id }` because you do not assign `units` to non related charge fee.
Also you can swap `.max` with `.min` and test again will be green.

I can suggest to rewrite it like this to cover missing cases.
```suggestion
  describe "#find_value" do
    subject { alert.find_value(current_usage) }

    let(:alert) { create(:billable_metric_usage_units_alert) }
    # here was creation of subscription but it does not affect anything from what I see
    let(:charge) { create(:standard_charge, billable_metric: alert.billable_metric) }
    let(:current_usage) { instance_double(Invoice, fees:) } # We should avoid using 'double' cause it often lead to mistakes.
    # There's not such attribute for Invoice like amount_cents

    let(:fees) do
      [
        create(:charge_fee, charge:, units: 8), # amount_cents assigned were not affecting test results
        create(:charge_fee, charge:, units: 4), # will ensure that we're using max not min
        create(:charge_fee, units: 12) # ensure that we look only within correct charge fees
      ]
    end

    it "returns biggest units among fees related to the alert's billable metric" do
      expect(subject).to eq(8)
    end
  end
```

spec/services/usage_monitoring/create_alert_service_spec.rb
```diff
@@ -146,5 +147,29 @@
         expect(result.error.message).to include("too_many_thresholds")
       end
     end
+
+    context "when creating lifetime_usage alert" do
+      context "without required premium_integrations" do
```
floganz, line 152: Minor but if our predicated called `using_lifetime_usage?` I'd expect matching contexts in the tests, like "when organization using lifetime usage" / "when organization does not use lifetime usage".

spec/services/usage_monitoring/create_alert_service_spec.rb
```diff
@@ -146,5 +147,29 @@
...
+        let(:premium_integrations) { [] }
+        let(:params) { {alert_type: "lifetime_usage_amount", thresholds:, code: "first"} }
```
floganz, line 154: Also this let seems to be repeated in both contexts maybe it makes sense to put above this contexts directly after `context "when creating lifetime_usage alert" do`?

spec/services/usage_monitoring/create_alert_service_spec.rb
```diff
@@ -146,5 +147,29 @@
...
+        end
+      end
+
+      context "with required premium_integrations" do
+        let(:premium_integrations) { ["lifetime_usage"] }
+        let(:params) { {alert_type: "lifetime_usage_amount", thresholds:, code: "first"} }
+
+        around { |test| lago_premium!(&test) }
+
+        it "creates the alert" do
```
floganz, line 168: Another idea – should we check that Alert actually created and not build using `expect { subject }.to change` or at least adding check for `expect(result.alert).to be_persisted`?
If you agree would be also great to add negative counter part in another context `expect { subject }.not_to change`

app/services/usage_monitoring/process_subscription_activity_service.rb
```diff
@@ -33,13 +33,18 @@ def call
...
-        Alert.where(
+        alerts = Alert.where(
           subscription_external_id: subscription.external_id,
-          organization_id: subscription_activity.organization_id,
-          alert_type: Alert::CURRENT_USAGE_TYPES
-        ).includes(:thresholds).find_each do |alert|
+          organization_id: subscription_activity.organization_id
+        ).includes(:thresholds)
+
+        alerts.where(alert_type: Alert::CURRENT_USAGE_TYPES).find_each do |alert|
```
floganz, line 41: What we'll add scopes to the `Alert` modes: `current_usage_kind` or simply `current_usage` and similar for lifetime (like `lifetime_usage`)?
Then we don't need to expose class internals in outside services.

julienbourdeau: yes, absolutely!

app/services/usage_monitoring/update_alert_service.rb
```diff
@@ -15,6 +15,10 @@ def initialize(alert:, params:)
     def call
       return result.not_found_failure!(resource: "alert") unless alert
 
+      if params.has_key?(:thresholds) && params[:thresholds].size > AlertThreshold::SOFT_LIMIT
```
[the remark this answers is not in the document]
julienbourdeau: With GQL we're [guaranteed it's an array](https://github.com/getlago/lago-api/blob/2a2b06ef266983cc6453d99bc33dcd24c66d71d7/app/graphql/types/usage_monitoring/alerts/update_input.rb#L13) but I will double check in the [rest API PR](https://github.com/getlago/lago-api/pull/3600)

spec/models/usage_monitoring/lifetime_usage_amount_alert_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+RSpec.describe UsageMonitoring::LifetimeUsageAmountAlert, type: :model do
+  let(:alert) { create(:lifetime_usage_amount_alert) }
+  let(:lifetime_usage) { create(:lifetime_usage, invoiced_usage_amount_cents: 6, current_usage_amount_cents: 3) }
+
+  describe "#find_value" do
+    it do
+      expect(alert.find_value(lifetime_usage)).to eq(9)
+    end
+  end
```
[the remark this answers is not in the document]
julienbourdeau: `build_stubbed` worked when I passed an external_subscription_id, otherwise the factory tried to create a subscription.

I definitely agree that we rely on create way too much where `build` would be fine. I'm guilty!

spec/services/usage_monitoring/create_alert_service_spec.rb
julienbourdeau: I like `be_persisted`

spec/models/usage_monitoring/billable_metric_usage_units_alert_spec.rb
julienbourdeau: > We should avoid using 'double' cause it often lead to mistakes.
> There's not such attribute for Invoice like amount_cents

I understand by here current_usage is not an invoice, [it was an OpenStruct until recently](https://github.com/getlago/lago-api/pull/3509), but I forgot it now has it's own type.

I'll definitely update the setup tho! (and the one I copy pasted this from :P )

julienbourdeau: All good! I was just adding details 🙏

PR #3691, opened 2025-05-21
feat(audit_logs): Activity log for credit notes

rsempe:
## Roadmap Task

👉  [https://getlago.canny.io/feature-requests/p/access-audit-logs](https://getlago.canny.io/feature-requests/p/access-audit-logs)

## Context

Our users want to have visibility on what’s happening in Lago.

Examples:
- *A plan and prices have been updated: When? By whom? What?*
- *A subscription has been deleted: When? By whom? What?*

## Description

The goal of this PR is to create activity logs for credit notes.

### DIFF
--- a/app/services/credit_notes/create_from_progressive_billing_invoice.rb
+++ b/app/services/credit_notes/create_from_progressive_billing_invoice.rb
@@ -17,13 +17,13 @@ module CreditNotes
       # Important to call this method as it modifies @amount if needed
       items = calculate_items!
 
-      CreditNotes::CreateService.new(
+      CreditNotes::CreateService.call(
         invoice: progressive_billing_invoice,
         credit_amount_cents: creditable_amount_cents(amount, items),
         items:,
         reason:,
         automatic: true
-      ).call.raise_if_error!
+      ).raise_if_error!
     end
 
     private
--- a/app/services/credit_notes/create_from_progressive_billing_invoice.rb
+++ b/app/services/credit_notes/create_from_progressive_billing_invoice.rb
@@ -17,13 +17,13 @@ module CreditNotes
       # Important to call this method as it modifies @amount if needed
       items = calculate_items!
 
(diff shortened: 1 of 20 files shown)


app/services/credit_notes/create_from_progressive_billing_invoice.rb
```diff
@@ -17,13 +17,13 @@ def call
       # Important to call this method as it modifies @amount if needed
       items = calculate_items!
 
-      CreditNotes::CreateService.new(
+      CreditNotes::CreateService.call(
```
ancorcruz, line 20: you can use `call!` it will raise if error.


```suggestion
      CreditNotes::CreateService.call!(
```

app/services/utils/activity_log.rb
```diff
@@ -105,15 +106,17 @@ def resource(activity_object)
...
       end
 
       def external_subscription_id(activity_object)
-        # TODO: Implement me.
-        nil
+        return activity_object.external_id if activity_object.is_a?(Subscription)
+        return nil unless activity_object.respond_to?(:subscription)
+
+        activity_object.subscription.external_id
       end
```
ancorcruz, line 120: These methods currently rely on checking the object’s type (`is_a?`) and whether it responds to certain methods (`respond_to?`). This works, but it couples the logic to specific object structures and violates the *Tell, Don’t Ask* principle.

A more object-oriented approach would be to delegate the responsibility to the `activity_object` itself by implementing methods like `#external_customer_id` or `#external_subscription_id` directly on the relevant classes. This way, we avoid branching logic and instead rely on polymorphism.

This would make the code easier to extend and maintain—especially if we support more `activity_object` types in the future.

ancorcruz (commented):
Looks great. I have left some comments for your consideration. Thanks.

mariohd (approved):

PR #3698, opened 2025-05-21
Fix(wallet_transaction): Don't create transaction when the credits are zero or negative.  

diegocharles:
Do not create WalletTransaction if the paid_credits is zero or negative.

### DIFF
--- a/app/services/wallets/create_interval_wallet_transactions_service.rb
+++ b/app/services/wallets/create_interval_wallet_transactions_service.rb
@@ -5,13 +5,17 @@ module Wallets
     def call
       recurring_transaction_rules.each do |recurring_transaction_rule|
         wallet = recurring_transaction_rule.wallet
+        paid_credits = paid_credits(recurring_transaction_rule)
+        granted_credits = granted_credits(recurring_transaction_rule)
+
+        next if recurring_transaction_rule.target? && paid_credits.zero? && granted_credits.zero?
 
         WalletTransactions::CreateJob.perform_later(
           organization_id: wallet.organization.id,
           params: {
             wallet_id: wallet.id,
-            paid_credits: paid_credits(recurring_transaction_rule),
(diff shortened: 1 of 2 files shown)


app/services/wallets/create_interval_wallet_transactions_service.rb
```diff
@@ -5,13 +5,15 @@ class CreateIntervalWalletTransactionsService < BaseService
     def call
       recurring_transaction_rules.each do |recurring_transaction_rule|
         wallet = recurring_transaction_rule.wallet
+        paid_credits = paid_credits(recurring_transaction_rule)
+        next if paid_credits.zero?
```
brunomiguelpinto, line 9: on the case the rule is not a target rule rule.target? we need to acepet a 0 because we need to process a granted credits 

Screenshot 2025-05-22 at 9 56 29 AM

diegocharles: @brunomiguelpinto I've updated the PR:
 

-  The ` WalletTransactions::CreateJob` will be avoided to get called if the rule is `target` and all the credits are more than zero. 
That will prevent unnecessary calls and save some memory. 

Please let me know if I'm still missing anything!

app/services/wallets/create_interval_wallet_transactions_service.rb
```diff
@@ -3,15 +3,19 @@
 module Wallets
   class CreateIntervalWalletTransactionsService < BaseService
     def call
-      recurring_transaction_rules.each do |recurring_transaction_rule|
+      recurring_transaction_rules.find do |recurring_transaction_rule|
```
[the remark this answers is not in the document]
diegocharles: @brunomiguelpinto Nice catch!! Thank you 💪

brunomiguelpinto (approved):

PR #3701, opened 2025-05-22
fix(cache): temptative fix for Redis "invalid expire time"

julienbourdeau:
> Redis::CommandError: ERR invalid expire time in 'set' command

I'm not entirely sure what happens here but I think, since some services use `expires_in: 0`, down the line, we compute an invalid expiration time (in the past). As far as I understand, `expires_in: 0` is **not a NO-OP**, it still writes to the cache, yet, it's instantly invalid.

I replaced the `fetch` call by a manual read/compute/write combo. I'm open to any better idea.


ancorcruz (approved):
LGTM

PR #3703, opened 2025-05-22
Fix(wallet balance): failed transaction and voided invoice

annvelents:
## Context

we got a bug where:
1. a credit was purchased (1$)
2. invoice marked as succeeded -> wallet transaction marked as settled,
3. wallet balance is updated to 1$
4. invoice payment status updated to failed -> wallet transaction marked as failed
5. wallet balance WAS NOT updated (still 1$) - FIRST ISSUE
6. invoice is voided -> recredit wallet_transaction is created - SECOND ISSUE
7. wallet balance increases to 2$

## Description

- when fail a wallet transaction, if a transaction was settled before, it will stay settled, the credits should be voided manually
- when recredit wallet_transactions when voiding an invoice, only recredit outbound wallet transaction (exactly "applied" credits from the wallet)

### DIFF
--- a/app/services/invoices/void_service.rb
+++ b/app/services/invoices/void_service.rb
@@ -24,7 +24,7 @@ module Invoices
         Rails.logger.warn("Recrediting credit #{credit.id} failed for invoice #{invoice.id}") unless res.success?
       end
 
-      invoice.wallet_transactions.each do |wallet_transaction|
+      invoice.wallet_transactions.outbound.each do |wallet_transaction|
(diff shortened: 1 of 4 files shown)


ancorcruz (commented):
as mentioned in a call... settled wallet transactions should not be updated as failed, instead the WT should be voided and created a new WT to deduct the amount.

Another question is why a succeeded payment became failed?

brunomiguelpinto (approved):

PR #3705, opened 2025-05-22
feat(wallet-credit-limitation): add migration for precise credits values

lovrocolic:
## Context

New feature for limiting wallet credits on certain fee type is currently being made.

## Description

This PR adds necessary fields where weighted credits values will be stored for progressive billing and credit notes.

We already have similar fields that calculates weighted coupon value applied on certain fee - `precise_coupons_amount_cents`

Let's say that with new feature we want to apply prepaid credits only on subscription fees; weighted values are needed in order to calculate correctly what is the total fee amount before applying prepaid credits for certain subscription fee. It wouldn't be possibly to apply anymore prepaid credits on total amount.

### DIFF
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -208,48 +208,50 @@ end
 #
 # Table name: fees
 #
-#  id                                  :uuid             not null, primary key
-#  amount_cents                        :bigint           not null
-#  amount_currency                     :string           not null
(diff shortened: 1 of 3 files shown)


lovrocolic:
Closing this one. It will be tackle in another PR since we realised that `precise_progressive_credits_amount_cents` are already stored in the same field as coupon precise amount cents

PR #3718, opened 2025-05-26
chore(stripe): Refactor Stripe webhooks

julienbourdeau:
## Context

This PR is part of [upgrading Stripe](https://github.com/getlago/lago-api/pull/3300) (clients + API version).

## Description

This is extracted from #3300. 

**You cannot update the `api_version` of an existing WebhookEndpoint**. You must create a new WebhookEndpoint with the new version and delete the old one once you migrated. This is the best way to not have any downtime.

https://docs.stripe.com/webhooks/versioning

It means to set the version for all our customers, we'll need to recreate the endpoint for all organization. It's fine but I'd rather upgrade to a more recent version before we enforce it in the webhook.

### DIFF
--- a/app/services/payment_providers/stripe/refresh_webhook_service.rb
+++ b/app/services/payment_providers/stripe/refresh_webhook_service.rb
@@ -2,26 +2,24 @@
 
 module PaymentProviders
   module Stripe
-    class RegisterWebhookService < BaseService
+    class RefreshWebhookService < BaseService
+      Result = BaseResult
+
       def call
-        stripe_webhook = ::Stripe::WebhookEndpoint.create(
+        ::Stripe::WebhookEndpoint.update(
+          payment_provider.webhook_id,
           webhook_endpoint_shared_params,
           {api_key:}
         )
 
-        payment_provider.update!(
-          webhook_id: stripe_webhook.id,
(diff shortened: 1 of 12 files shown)


app/services/payment_providers/stripe/refresh_webhook_service.rb
```diff
@@ -0,0 +1,30 @@
...
+      def call
+        ::Stripe::WebhookEndpoint.update(
+          payment_provider.webhook_id,
+          webhook_endpoint_shared_params,
+          {api_key:}
+        )
+
+        result
+      rescue ActiveRecord::RecordInvalid => e
+        result.record_validation_failure!(record: e.record)
```
vincent-pochet, line 17: Maybe I am missing something, but it seems that we are not touching any model. So maybe these two lines are not required.

vincent-pochet (approved):
Looks great!

app/services/payment_providers/stripe/refresh_webhook_service.rb
julienbourdeau: Indeed! I'll remove in the next PR (toi avoid conflict and stuff) Thanks!

PR #3720, opened 2025-05-26
fix(wallet): Optimistic locking for in advance events

vincent-pochet:
## Context

This PR is a new attempt to approach the issue of parallel wallet refresh refresh in the case of in payed in advance events.

## Description

The PR moves fro pessimistic locking (`FOR UPDATE NOWAIT`) to an optimistic one using the `lock_version` already present on the wallets table.
An automatic retry logic is added around the wallet decrease logic with a random wait time between each attempt. It reduce the effect of the error to the decrease logic only instead of the all event processing like it is today.
Finally in case of error, the error is re-raised and the entire logic will be retried by Sidekiq

### DIFF
--- a/app/services/credits/applied_prepaid_credit_service.rb
+++ b/app/services/credits/applied_prepaid_credit_service.rb
@@ -2,6 +2,8 @@
 
 module Credits
   class AppliedPrepaidCreditService < BaseService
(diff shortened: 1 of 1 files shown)


rsempe (approved):
lgtm!

julienbourdeau (approved):

PR #3724, opened 2025-05-26
feat(multientities): applied_dunning_campaign on BE to follow org

annvelents:
## Context

We're switching dunning campaign definition to billing entity, but to do this, firstly we should support backward compatibility

## Description

- added a migration to populate billing_entity.applied_dunning_campaign_id
- added a foreign key with validation on this reference
- when creating / updating / discarding a dunning campaign, if it's applied on billing_entity, this relation also will be updated


app/models/dunning_campaign.rb
```diff
@@ -10,7 +10,8 @@ class DunningCampaign < ApplicationRecord
   belongs_to :organization
 
   has_many :thresholds, class_name: "DunningCampaignThreshold", dependent: :destroy
-  has_many :customers, foreign_key: :applied_dunning_campaign_id, dependent: :nullify
```
annvelents, line 13: We do not destroy dunning_campaigns - we discard them, so this nullify never works

ancorcruz: good catch!

app/services/dunning_campaigns/update_service.rb
```diff
@@ -103,6 +103,9 @@ def handle_applied_to_organization_update
 
       organization.reset_customers_last_dunning_campaign_attempt
 
+      new_applied_dunning_campaign = dunning_campaign.applied_to_organization ? dunning_campaign : nil
+      organization.default_billing_entity.update(applied_dunning_campaign: new_applied_dunning_campaign)
```
ancorcruz:
```suggestion
      organization.default_billing_entity.update!(applied_dunning_campaign: new_applied_dunning_campaign)
```
lets ensure the billing entity is updated.

ancorcruz (approved):
good stuff!

PR #3729, opened 2025-05-26
feat: coupon recrediting when voiding invoices

brunomiguelpinto:
## Context
When an invoice is voided, any applied coupons associated with that invoice should be properly recredited. Previously, when an invoice was voided, the credits from applied coupons were not properly handled, which could lead to customers losing coupon benefits they were entitled to. This was particularly problematic for recurring coupons and coupons with a limited number of uses.

## Description
This PR implements a comprehensive solution for handling coupon credits when invoices are voided:

Added active and voided scopes to filter credits based on invoice status
These scopes ensure that only credits from non-voided invoices are counted against the coupon's remaining amount

Added AppliedCoupons::RecreditService:
Implemented a new service that handles recrediting of applied coupons when an invoice is voided
For terminated coupons: reactivates the coupon if it was terminated due to usage, and the original coupon is still active
For recurring coupons: increments the frequency_duration_remaining to restore the usage count
Uses database locking to prevent race conditions during concurrent operations


vincent-pochet (commented):
Looks good :+1:

julienbourdeau (approved):
Awesome

diegocharles (approved):

PR #3741, opened 2025-05-27
Feat(multientities): dunning campaing - reset associated customers

annvelents:
## Context

Since source of applied dunning campaign is billing entity, when resetting dunning campaign, we should reset customers through billing_entity

### DIFF
--- a/app/models/dunning_campaign.rb
+++ b/app/models/dunning_campaign.rb
@@ -44,7 +44,7 @@ class DunningCampaign < ApplicationRecord
     )
 
     # NOTE: Reset last attempt on customers falling back to the billing_entity campaign
-    billing_entities.map(&:reset_customers_last_dunning_campaign_attempt)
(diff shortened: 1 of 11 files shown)


app/models/dunning_campaign.rb
```diff
@@ -43,8 +43,8 @@ def reset_customers_last_attempt
       last_dunning_campaign_attempt_at: nil
     )
 
-    # NOTE: Reset last attempt on customers falling back to the organization campaign
-    organization.reset_customers_last_dunning_campaign_attempt if applied_to_organization?
+    # NOTE: Reset last attempt on customers falling back to the billing_entity campaign
+    billing_entities.map(&:reset_customers_last_dunning_campaign_attempt)
```
ancorcruz:
```suggestion
    billing_entities.includes(:customers).map(&:reset_customers_last_dunning_campaign_attempt)
```

Shall we preload customers to avoid N+1 query?

ancorcruz (approved):
LGTM!

PR #3744, opened 2025-05-27
fix(invoice-preview): Support preview for subscription during upgrade or downgrade

floganz:
## Context

Current logic for invoice preview does not take into account if existing subscription is in the process of downgrade or upgrade. To generate proper invoice preview we need to slightly adjust data so that subscription state matches one of a new bill run.

## Description

Add validation for customer persistence. It improves UX by providing more precise error message for some invalid params.
Add support for proper preview of subscription in a process of being downgraded.

### DIFF
--- a/app/services/invoices/preview/find_subscriptions_service.rb
+++ b/app/services/invoices/preview/find_subscriptions_service.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+module Invoices
+  module Preview
+    class FindSubscriptionsService < BaseService
+      Result = BaseResult[:subscriptions]
+
+      def initialize(subscriptions:)
+        @subscriptions = subscriptions
+        super
+      end
+
+      def call
+        return result.not_found_failure!(resource: "subscription") if subscriptions.empty?
+
+        result.subscriptions = subscriptions.flat_map do |subscription|
+          if subscription.next_subscription
+            [
+              terminated_subscription(subscription),
+              (adjusted_next_subscription(subscription) if subscription.next_subscription.plan.pay_in_advance?)
+            ].compact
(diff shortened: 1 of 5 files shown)


app/services/invoices/preview/find_subscriptions_service.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+
+        result
+      end
+
+      private
+
+      attr_reader :subscriptions
+
+      def terminated_subscription(subscription)
```
lovrocolic, line 34: I have a feeling that those private methods are pretty similar to the private methods in `SubscriptionPlanChangeService`. Can't we find a way to have it in single place? Maybe create `Invoices::Preview::BaseService` or use some other approach?

app/services/invoices/preview/find_subscriptions_service.rb
```diff
@@ -0,0 +1,65 @@
...
+      def initialize(subscriptions:)
+        @subscriptions = subscriptions
+        super
+      end
+
+      def call
+        return result.not_found_failure!(resource: "subscription") if subscriptions.empty?
+
+        result.subscriptions = subscriptions.flat_map do |subscription|
+          if subscription.next_subscription
```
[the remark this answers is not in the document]
floganz: I've updated implementation to only modify subscriptions if it's downgrade case.

app/services/invoices/preview/find_subscriptions_service.rb
floganz: With support for only downgrade case implementation simplified and this issue no longer relevant.

lovrocolic (approved):

PR #3745, opened 2025-05-27
Feat(multientities): billing_entity source of dunning campaign

annvelents:
## Context

As default billing_entity follows organization's applied dunning campaign, we want to use billing_entity.applied_dunning_campaign to process dunning campaign


ancorcruz (approved):
LGTM!

PR #3748, opened 2025-05-27
feat: void and generate credit note

brunomiguelpinto:
## Roadmap Task

## Context

## Description

### DIFF
--- a/app/controllers/api/v1/invoices_controller.rb
+++ b/app/controllers/api/v1/invoices_controller.rb
@@ -134,7 +134,7 @@ module Api
       def void
         invoice = current_organization.invoices.visible.find_by(id: params[:id])
 
-        result = Invoices::VoidService.call(invoice:)
+        result = Invoices::VoidService.call(invoice: invoice, params: void_params)
         if result.success?
           render_invoice(result.invoice)
         else
@@ -339,6 +339,10 @@ module Api
             ]
           ]
         )
+      end
+
+      def void_params
+        params.permit(:generate_credit_note, :refund_amount, :credit_amount)
       end
 
       def sync_salesforce_id_params


app/models/invoice.rb
```diff
@@ -402,6 +402,10 @@ def should_apply_provider_tax?
     fees.any? && should_finalize_invoice
   end
 
+  def force_void!
```
diegocharles, line 405: @brunomiguelpinto Why not just `void!` ? I think we can merge those two methods on two, since the word `void_invoice!` is redundant once it already belongs to the Invoice model. 


```
  `def void_invoice!
    update!(ready_for_payment_processing: false)
  end`
```
  
  becomes: 
  
  `def unmark_ready_for_payment_process!`
  
  and our method becomes: 
  
```
  def void!
    unmark_ready_for_payment_process!
    update!(status: "voided")
  end
```
  
  What do you think?

app/services/invoices/void_service.rb
```diff
@@ -53,10 +69,129 @@ def call
...
+      total_invoice_amount = invoice.total_amount_cents
+      remaining_amount = total_invoice_amount - credit_amount - refund_amount
+      if remaining_amount.positive?
+        # Generate items for the remaining amount
+        remaining_items = distribute_amount_to_items(remaining_amount)
+
+        credit_note_to_void = CreditNotes::CreateService.call(
+          invoice: invoice,
+          reason: :other,
+          description: "Credit note created due to voided invoice",
```
diegocharles, line 185: @brunomiguelpinto What if we let this statement more explicit like "Credit note from the invoice #{invoice.number}" ?

brunomiguelpinto: That's a good ideia I will add it

app/services/invoices/void_service.rb
```diff
@@ -2,18 +2,27 @@
...
 
     def call
       return result.not_found_failure!(resource: "invoice") unless invoice
       return result.not_allowed_failure!(code: "not_voidable") if invoice.voided?
+      return result.not_allowed_failure!(code: "not_voidable") if !invoice.voidable? && !explicit_void_intent?
 
       ActiveRecord::Base.transaction do
         invoice.payment_overdue = false if invoice.payment_overdue?
-        invoice.void!
+        if invoice.may_void?
```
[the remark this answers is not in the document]
brunomiguelpinto: im using the methods that come from the state machine

app/models/invoice.rb
brunomiguelpinto: I’m using the methods provided by the state machine — void! is already part of it. I wanted the method name to reflect the intention to force the void, because this version skips the voidable validation.

brunomiguelpinto:
```
    event :void do
      transitions from: :finalized, to: :voided, guard: :voidable?, after: :void_invoice!
    end
```
im skipping this part

app/graphql/mutations/invoices/void.rb
```diff
@@ -11,14 +11,25 @@ class Void < BaseMutation
...
 
       def resolve(**args)
+        invoice = current_organization.invoices.visible.find_by(id: args[:id])
+
+        # Collect all parameters except invoice into a params hash
+        params = {
+          generate_credit_note: args[:generate_credit_note],
+          refund_amount: args[:refund_amount],
+          credit_amount: args[:credit_amount]
+        }
```
floganz: nitpick: We can achieve the same with `args.slice(:generate_credit_note, :refund_amount, :credit_amount)`

app/services/invoices/void_service.rb
```diff
@@ -2,41 +2,52 @@
...
+      @refund_amount = params[:refund_amount].to_i
+      @credit_amount = params[:credit_amount].to_i
       super
     end
 
     def call
       return result.not_found_failure!(resource: "invoice") unless invoice
       return result.not_allowed_failure!(code: "not_voidable") if invoice.voided?
+      return result.not_allowed_failure!(code: "not_voidable") if !invoice.voidable? && !explicit_void_intent?
+      unless validate_credit_note_amounts!
```
floganz: nitpick: I'd expect that method used in condition and that returns boolean value will end with "?" rather than "!".
Alternative naming could be `valid_credit_note_amount?`

app/services/credit_notes/create_service.rb
```diff
@@ -39,7 +39,7 @@ def call
           reason:,
           description:,
           credit_status: "available",
-          status: invoice.status
+          status: (invoice.status == "voided") ? "finalized" : invoice.status # credit notes dont have void state
```
floganz: Should we add unit tests for this change?

spec/graphql/mutations/invoices/void_spec.rb
```diff
@@ -44,4 +44,51 @@
...
+        params: hash_including(
+          generate_credit_note: true,
+          credit_amount: credit_amount,
+          refund_amount: refund_amount
+        )
+      )
+
+      result_data = result["data"]["voidInvoice"]
+
+      aggregate_failures do
```
floganz: nitpick: we dont need `aggregate_failures` anymore.

app/services/invoices/void_service.rb
floganz, line 1: Are you planning to add test for changes in this file?
There's not new test for this service and graphql/API rest are only stubs and expectations about service being called.

app/services/invoices/void_service.rb
```diff
@@ -53,10 +64,73 @@ def call
...
+      return true unless generate_credit_note
+
+      return false if credit_amount > invoice.creditable_amount_cents
+      return false if refund_amount > invoice.refundable_amount_cents
+      return false if (credit_amount + refund_amount) > invoice.creditable_amount_cents
+
+      true
+    end
+
+    def create_credit_notes!
```
floganz, line 97: nitpick: This method and few others adds tonn of logic to this class which might be hard to cover with tests. Maybe we should consider extracting it to new service?
However not sure if it's feasible.

floganz (commented):
Looks good but I think we miss a lot of tests for these changes.

app/services/invoices/void_service.rb
brunomiguelpinto: Yes I will add a bunch of tests on this one

app/services/invoices/void_service.rb
diegocharles: @floganz I totally agree that missing specs is not desirable. 
For this particular PR, the QA team is reviewing the flow and we need to release the future version so we can gather some feedback.

We can create another task to complete the missing tests in parallel for the next week, but this won’t block this PR.

What do you think? @brunomiguelpinto

floganz: Mainly we need tests for eliminating regression testing and not accidentally breaking features **after** it's merged.
QA team more responsible for validating that we build what expected **before** we merge it.

If won't write tests during feature development I believe we'll never do them.

P.S. this particular comment about extracting logic to separate service which can be done later or skipped.

diegocharles: @floganz I agree with your concerns. We're going to update the specs today. 

Thank you for sharing your thoughts!

app/graphql/types/invoices/void_invoice_input.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Types
+  module Invoices
+    class VoidInvoiceInput < Types::BaseInputObject
+      description "Void Invoice input arguments"
+
+      argument :credit_amount, GraphQL::Types::BigInt, required: false
+      argument :generate_credit_note, Boolean, required: false
+      argument :id, ID, required: true
```
vincent-pochet: Not a blocker, but we usually try to add the `id  argument as the first one, separated by a blank line

app/models/invoice.rb
```diff
@@ -405,6 +405,10 @@ def should_apply_provider_tax?
...
```
vincent-pochet: May I suggest `mark_as_voided`, we have this pattern in some other location.

app/services/credit_notes/create_service.rb
vincent-pochet:
```suggestion
          status: invoice.voided? ? "finalized" : invoice.status # credit notes dont have void state
```

app/services/invoices/void_service.rb
```diff
@@ -15,33 +19,40 @@ def initialize(invoice:)
...
+      end
 
       ActiveRecord::Base.transaction do
         invoice.payment_overdue = false if invoice.payment_overdue?
-        invoice.void!
+        if invoice.may_void?
+          invoice.void!
+        else
+          invoice.force_void!
+        end
```
vincent-pochet: Can you add a comment regarding this logic. I think it might raise some question for future us :D

app/models/invoice.rb
diegocharles: @vincent-pochet That's a fine suggestion, however, the intent here is follow the same action patterns set by AASM, so we have the `void!`, and we have the `force_void!`. 

One fair improvement here is that we could have something like: 

```
def  void!(force: false)
  super unless force
  mark_as_voided!
end
```

But I'm not sure if that will work as expected.

diegocharles: I've updated the code

PR #3750, opened 2025-05-27
chore(spec): better scenario helpers

julienbourdeau:
As I'm working on scenario tests, I realize there are some issues with the helpers:
* I always hated that api call helpers didn't return the http call response.
* Also, some helpers didn't empty the queue so the behavior is "incorrect".

### DIFF
--- a/spec/scenarios/delete_plan_spec.rb
+++ b/spec/scenarios/delete_plan_spec.rb
@@ -45,7 +45,7 @@ describe "Delete Plan Scenarios", :scenarios, type: :request do
 
     travel_to(jan20) do
       overridden_plan = plan.children.first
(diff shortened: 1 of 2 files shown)


spec/scenarios/delete_plan_spec.rb
```diff
@@ -45,7 +45,7 @@
 
     travel_to(jan20) do
       overridden_plan = plan.children.first
-      delete_plan(plan)
+      delete_with_token(organization, "/api/v1/plans/#{plan.code}")
```
julienbourdeau, line 48: in this case, the test ensure something before emptying the queue manually

rsempe (approved):

PR #3753, opened 2025-05-28
Feat(multientities): be manage applied dunning campaign

annvelents:
## Context

We're planning to manage applied dunning campaigns on billing entities, not using the "applied_to_organization" flag

## Description

Added service that update applied_dunning_campaign on billing_entity

### DIFF
--- a/app/services/billing_entities/update_applied_dunning_campaign_service.rb
+++ b/app/services/billing_entities/update_applied_dunning_campaign_service.rb
@@ -6,9 +6,11 @@ module BillingEntities
     def initialize(billing_entity:, dunning_campaign: nil)
       @billing_entity = billing_entity
       @dunning_campaign = dunning_campaign
+      super
     end
 
     def call
+      return result.not_found_failure!(resource: "billing_entity") if billing_entity.nil?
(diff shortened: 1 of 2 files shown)


app/services/billing_entities/update_applied_dunning_campaign_service.rb
```diff
@@ -0,0 +1,22 @@
...
+module BillingEntities
+  class UpdateAppliedDunningCampaignService < BaseService
+    Result = BaseResult[:billing_entity]
+    def initialize(billing_entity:, dunning_campaign: nil)
+      @billing_entity = billing_entity
+      @dunning_campaign = dunning_campaign
+    end
+
+    def call
+      return unless billing_entity.applied_dunning_campaign != dunning_campaign
```
ancorcruz, line 14: should it fail if billing entity is nil?

ancorcruz (approved):
LGTM!

This service will be called from a later change, right?

app/services/billing_entities/update_applied_dunning_campaign_service.rb
annvelents: Added, good spot! 🙌

PR #3765, opened 2025-05-30
fix(charge): Fix deletion cascade on large plan override

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/3747

Deleting a charge in a plan with a large number of override can cause Timeout error or lead to a memory error on Redis.

The main reasons are:
- The service will enqueue one job per child charge
- In the service, the soft deletion is done using `discard_all` witch is not scalable as it is using a DB call per record...

## Description

This PR is refactoring the logic to enqueue only one job responsible of the soft deletion of all child charges.
The filter deletion is also done by replacing `discard_all` calls with `update_all(deleted_at:)`


spec/jobs/charges/destroy_children_job_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Charges::DestroyChildrenJob, type: :job do
+  let(:charge) { create(:standard_charge) }
+
+  before do
+    allow(Charges::DestroyChildrenService).to receive(:call!).with(charge)
+      .and_return(Charges::DestroyChildrenService::Result.new)
```
floganz: If the only purpose to avoid exception because of stub we can use `.and_return_original` which elimates need to construct proper return value.
If return value is not need maybe we can drop for simplicity?

spec/jobs/charges/destroy_children_job_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Charges::DestroyChildrenJob, type: :job do
+  let(:charge) { create(:standard_charge) }
```
floganz: Another small suggestion, maybe create here discarded record to cover `.with_discarded` lookup?

spec/services/charges/destroy_children_service_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      plan_id: child_plan.id,
+      parent_id: charge_parent_id,
+      billable_metric_id: billable_metric.id,
+      properties: {amount: "300"}
+    )
+  end
+
+  before { child_charge }
+
+  describe "#call" do
```
floganz, line 28: Should we add a context to test second return? That if charge is not discarded we exit and do not change charge children?

julienbourdeau (approved):

PR #3767, opened 2025-05-30
fix(charge): Fix creation cascade on large plan override

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/3747 and https://github.com/getlago/lago-api/pull/3765

Create a new charge in a plan with a large number of override can cause Timeout error or lead to a memory error on Redis.

The main reasons is that the service will enqueue one job per child plan

## Description

This PR is refactoring the logic to enqueue only one job responsible of the creation of all child charges.

Note: The same logic will have to be applied to the update of child charges


app/services/plans/update_service.rb
```diff
@@ -173,10 +171,9 @@ def process_charges(plan, params_charges)
           next
         end
 
-        create_charge_result = Charges::CreateService.call(plan:, params: payload_charge)
-        create_charge_result.raise_if_error!
+        create_charge_result = Charges::CreateService.call!(plan:, params: payload_charge)
 
-        after_commit { cascade_charge_creation(payload_charge.merge(parent_id: create_charge_result.charge.id)) }
+        after_commit { cascade_charge_creation(created_charges_ids, payload_charge) }
         created_charges_ids.push(create_charge_result.charge.id)
```
julienbourdeau: it seems that you're passing a list of ids here but the method (and job) are expecting a proper charge object. Maybe we're missing a scenario tests or we should check the params of the enqueued job.

app/services/plans/update_service.rb
```diff
@@ -173,10 +171,9 @@ def process_charges(plan, params_charges)
           next
         end
 
-        create_charge_result = Charges::CreateService.call(plan:, params: payload_charge)
-        create_charge_result.raise_if_error!
+        create_charge_result = Charges::CreateService.call!(plan:, params: payload_charge)
 
-        after_commit { cascade_charge_creation(payload_charge.merge(parent_id: create_charge_result.charge.id)) }
+        after_commit { cascade_charge_creation(created_charges_ids, payload_charge) }
```
floganz: What if we move this `after_commit` statement outside of `each` loop? It a bit strange to see that it defined inside the loop where `created_charges_ids` variable does not have a final value yet.

app/services/plans/update_service.rb
vincent-pochet: Good catch!

app/services/plans/update_service.rb
```diff
@@ -92,13 +92,11 @@ def cascade_subscription_fee_update(old_amount_cents)
       end
     end
 
-    def cascade_charge_creation(payload_charge)
+    def cascade_charge_creation(charge, payload_charge)
```
floganz, line 95: I do not understand how this works. In this method and job and later `CreateChildrenService` we work with `charge` as instance of Charge model.

However if we look at how this method is called in `after_commit` callback below in this file we pass `charge_ids` as `charge` argument to this function. How we get single charge out if collection of ids?

I'd suggest to add these tests
```ruby
expect do
  plans_service.call
end.to have_enqueued_job(Charges::CreateJob)
end.to have_enqueued_job(Charges::CreateChildrenJob)
```
`.with` expectation to be sure that we pass correct thing to the job.

spec/jobs/charges/create_children_job_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      charge_model: "standard",
+      invoice_display_name: "charge1",
+      min_amount_cents: 100
+    }
+  end
+
+  before do
+    allow(Charges::CreateChildrenService).to receive(:call!)
+      .with(charge:, payload: params)
+      .and_return(Charges::CreateChildrenService::Result.new)
```
floganz: If result is not part of the test I'd suggest to drop that part for simplicity.

spec/jobs/charges/create_job_spec.rb
```diff
@@ -1,27 +0,0 @@
-# frozen_string_literal: true
-
-require "rails_helper"
-
-RSpec.describe Charges::CreateJob, type: :job do
```
floganz, line 5: You removed the specs but the job class itself is not deleted in this PR, is it intentional?

spec/services/charges/create_children_service_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  describe "#call" do
+    context "when charge is not found" do
+      let(:charge) { nil }
+
+      it "returns an error" do
+        result = create_service.call
+
+        aggregate_failures do
```
floganz: We don't need `aggregate_failures` since it's be default now.

app/services/plans/update_service.rb
vincent-pochet: It was a mistake with the implementation, it has been fixed

spec/services/charges/create_children_service_spec.rb
vincent-pochet: Bad copy and paste

PR #3772, opened 2025-06-02
fix(filter): Matching in advance events with ALL_FILTER_VALUES

vincent-pochet:
## Description

This PR fixes the in advance filter matching for filters defined using the `ALL_FILTER_VALUES`

Today if a filter is defined with the following structure
```json
{
  "scheme": ["visa", "mastercard"],
  "card_type": ["__ALL_FILTER_VALUES__"]
}
```

it should match an event sent with the following properties: `{"scheme": "visa", "card_type": "debit" }` but it's not

### DIFF
--- a/app/services/charge_filters/event_matching_service.rb
+++ b/app/services/charge_filters/event_matching_service.rb
@@ -13,7 +13,8 @@ module ChargeFilters
       # NOTE: Find all filters matching event properties
       matching_filters = filters.select do |filter|
         filter.to_h.all? do |key, values|
-          applicable_event_properties.key?(key) && applicable_event_properties[key].to_s.in?(values)
+          applicable_event_properties.key?(key) &&
+            (applicable_event_properties[key].to_s.in?(values) || values == [ChargeFilterValue::ALL_FILTER_VALUES])
         end
       end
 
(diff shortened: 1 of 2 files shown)


app/services/charge_filters/event_matching_service.rb
```diff
@@ -13,7 +13,8 @@ def call
       # NOTE: Find all filters matching event properties
       matching_filters = filters.select do |filter|
         filter.to_h.all? do |key, values|
-          applicable_event_properties.key?(key) && applicable_event_properties[key].to_s.in?(values)
+          applicable_event_properties.key?(key) &&
+            (applicable_event_properties[key].to_s.in?(values) || values == [ChargeFilterValue::ALL_FILTER_VALUES])
```
groyoh, line 17: Maybe you could wrap it into a method for readability:

```suggestion
          all_filter_values?(key, values) || property_match_filter_values?(key, values)
```

app/services/charge_filters/event_matching_service.rb
```diff
@@ -13,7 +13,8 @@ def call
       # NOTE: Find all filters matching event properties
       matching_filters = filters.select do |filter|
         filter.to_h.all? do |key, values|
-          applicable_event_properties.key?(key) && applicable_event_properties[key].to_s.in?(values)
+          applicable_event_properties.key?(key) &&
```
groyoh, line 16: If the property doesn't exist in the event, it won't match the `ALL_FILTER_VALUES`. Is that the behaviour we expect?

vincent-pochet: Yes, in that specific case, the event should not match.
`ALL_FILTER_VALUES` means all possible values defined at the billable metric level.

groyoh: 👍

PR #3774, opened 2025-06-02
chore(filter): Add the flat_filters database view

vincent-pochet:
## Context

Today, we are not able to use the cache system for organization using Clickhouse event store, because we are not able to easily identify if an event matches filters. 

As a result all action requiring to compute the usage (current usage, daily_usage...) have to recompute the full usage from 0 every time for all billable metric having received at least 1 event during the period. This is highly inefficient especially for charges with filters, has one filter will lead to one DB query.

With Postgres event store, we are able to use a system with charge filter granularity, we would like to use the same logic for the Clickhouse one.

## Description

This PR is a first step to add caching at Clickouse level, it adds a new `flat_filters` database view. This view will be used on event_processor side to check filters matching on all events received for organization using the Clickhouse store.

### DIFF
--- a/db/views/flat_filters_v01.sql
+++ b/db/views/flat_filters_v01.sql
@@ -1,0 +1,41 @@
+-- This view is used on events_processor side to check filters matching on events received for organization using the Clickhouse store
+-- It then allows us to expire the usage chage directly within the events events_processor service
+SELECT
+  billable_metrics.organization_id AS organization_id,
+  billable_metrics.code AS billable_metric_code,
+  charges.plan_id AS plan_id,
+  charges.id AS charge_id,
+  charges.updated_at AS charge_updated_at,
+  charge_filters.id AS charge_filter_id,
+  charge_filters.updated_at AS charge_filter_updated_at,
+  CASE WHEN charge_filters.id IS NOT NULL
+  THEN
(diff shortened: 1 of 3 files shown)


jdenquin (approved):

db/views/flat_filters_v01.sql
```diff
@@ -0,0 +1,40 @@
+SELECT
```
rsempe, line 3: what about adding a comment at the top of this file to explain why we have this view? 😅

db/views/flat_filters_v01.sql
```diff
@@ -0,0 +1,40 @@
...
+  AND charge_filters.deleted_at IS NULL
+  AND charge_filter_values.deleted_at IS NULL
+  AND billable_metric_filters.deleted_at IS NULL
+GROUP BY
+  billable_metrics.organization_id,
+  billable_metrics.code,
+  charges.plan_id,
+  charges.updated_at,
+  charges.id,
+  charges.updated_at,
```
rsempe, line 39:
```suggestion
```

db/views/flat_filters_v01.sql
vincent-pochet: Good idea :)

PR #3777, opened 2025-06-03
misc(usages): Add spec for pre-filling historical usages

ivannovosad:
## Description

Add scenario spec for backfilling historical usages.

### DIFF
--- a/app/services/daily_usages/fill_history_service.rb
+++ b/app/services/daily_usages/fill_history_service.rb
@@ -4,9 +4,10 @@ require "timecop"
 
 module DailyUsages
   class FillHistoryService < BaseService
-    def initialize(subscription:, from_datetime:)
+    def initialize(subscription:, from_datetime:, to_datetime: nil)
       @subscription = subscription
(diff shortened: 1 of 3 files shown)


app/services/daily_usages/fill_history_service.rb
```diff
@@ -99,7 +100,7 @@ def from
     end
 
     def to
-      @to ||= (subscription.terminated_at || Time.current).to_date
+      @to ||= (to_datetime || subscription.terminated_at || Time.current).to_date
```
julienbourdeau: I feel like this would be an issue with terminated subscription 🤔 You could set `to_datetime` after the subscription ending date.
One thing that come to mind would be something like this: all possible values, remove nil, use min.
But it would need some new test case.

```suggestion
      @to ||= [to_datetime, subscription.terminated_at, Time.current].compact.min.to_date
```

vincent-pochet: Good catch!

PR #3778, opened 2025-06-03
fix: use keyword arguments when calling refresh draft job

lovrocolic:
## Context

It seems that AR deserialisation in jobs does not work sometimes in combination with uniqueness constraint

## Description

After investigation, seems that preferred approach is switching to keyword arguments which are handled a bit different during deserialisation

### DIFF
--- a/spec/jobs/clock/refresh_draft_invoices_jobs_spec.rb
+++ b/spec/jobs/clock/refresh_draft_invoices_jobs_spec.rb
@@ -18,7 +18,7 @@ describe Clock::RefreshDraftInvoicesJob, job: true do
     context "when not ready to be refreshed" do
       it "does not call the refresh service" do
         described_class.perform_now
-        expect(Invoices::RefreshDraftJob).not_to have_been_enqueued.with(invoice)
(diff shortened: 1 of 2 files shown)


ancorcruz (approved):
LGTM!

app/jobs/invoices/refresh_draft_job.rb
```diff
@@ -12,7 +12,7 @@ class RefreshDraftJob < ApplicationJob
 
     unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
 
-    def perform(invoice)
+    def perform(invoice:)
```
groyoh, line 15: Isn't there a risk of introducing a minor breaking change (during the deployment) with this change ? If `clock` gets deployed before `worker` and the job runs, it would serialize the argument as keyword args and the running `worker` code would still expect a global id. This could lead to some dead jobs until `worker` finish deploying.

lovrocolic: I would say it should be fine. Even if we end up with some dead jobs (which is low probability), this job is safe to run multiple times and we can re-run it

PR #3779, opened 2025-06-03
feat(alerting): new field for alert.triggered webhook payload

julienbourdeau:
As I'm documenting this `alert.triggered` webhook I realize there was some field that could be better.
* `subscription_external_id` is how you can retrieve subscription via the API (instead of lago_id)
* `organization_id` in case you receive webhooks from multiple organizations? It should be everywhere.

I also edited the seeder to seed some TriggeredAlert.

### DIFF
--- a/db/seeds/alerting.rb
+++ b/db/seeds/alerting.rb
@@ -32,9 +32,9 @@ def create_customer_with_sub(ext_id, plan:, organization:)
   )
 end
 
-sub = create_customer_with_sub("alerting-#{SecureRandom.hex}", plan:, organization:)
+subscription = create_customer_with_sub("alerting-#{SecureRandom.hex}", plan:, organization:)
 
-UsageMonitoring::CreateAlertService.call(organization:, subscription: sub, params: {
+UsageMonitoring::CreateAlertService.call(organization:, subscription:, params: {
   alert_type: "current_usage_amount",
   code: "default",
   name: "Default Alert",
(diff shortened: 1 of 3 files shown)


brunomiguelpinto (approved):

db/seeds/alerting.rb
```diff
@@ -63,3 +63,30 @@ def create_customer_with_sub(ext_id, plan:, organization:)
     {value: 10_00, recurring: true}
   ]
 })
+
+triggered_alert = UsageMonitoring::TriggeredAlert.create!(alert:, organization:, subscription:,
+  current_value: 51,
+  previous_value: 8,
+  crossed_thresholds: [
+    {code: nil, value: 10, recurring: false}, {code: :warn, value: 25, recurring: false}, {code: :alert, value: 50, recurring: false}
+  ],
```
groyoh, line 72: Do those values make sense given that it doesn't match the actual alert thresholds ?

```
  thresholds: [
    {code: "info", value: 1000_00}
  ]
```

Or did I miss something ?

julienbourdeau: Good question! It doesn't with the current state of the alert. 

When alerts are updated, we replace (delete_all + insert) all thresholds. That why we store the threshold fully and don't "_link to the threshold_id_". We can assume this alert was set up this way 2 months ago and updated later.
The threshold make sense with the `previous_value` / `current_value` tho.

PR #3782, opened 2025-06-04
feat(coupons): Add `coupon_code` filter for "List all applied coupons" endpoint

groyoh:
## Context

Clients can review applied coupons using the relevant endpoint: https://getlago.com/docs/api-reference/coupons/get-all-applied. However, since there is no way to filter by coupon code, clients must page through all results to find redemptions for a specific coupon.

## Description

This PR adds a `coupon_code` filter to the `GET /api/v1/applied_coupons` endpoint.

Since we're joining on `coupons` and the query already includes a `WHERE` clause with `organization_id` and `coupons.deleted_at IS NULL`, we don't need any extra index as PG will rely on the existing `(organization_id,code) UNIQUE WHERE (deleted_at IS NULL)` index:

```sql
-- puts AppliedCouponsQuery.call(organization: Organization.last, filters: {coupon_code: "test"}).applied_coupons.to_sql

lago=# EXPLAIN
SELECT "applied_coupons"."id",
  "applied_coupons"."coupon_id",
  "applied_coupons"."customer_id",
  "applied_coupons"."status",
  "applied_coupons"."amount_cents",
  "applied_coupons"."amount_currency",
  "applied_coupons"."created_at",
  "applied_coupons"."updated_at",
  "applied_coupons"."terminated_at",
  "applied_coupons"."percentage_rate",
  "applied_coupons"."frequency",
  "applied_coupons"."frequency_duration",
  "applied_coupons"."frequency_duration_remaining",
  "applied_coupons"."organization_id"
FROM "applied_coupons"
  INNER JOIN "coupons" ON "applied_coupons"."coupon_id" = "coupons"."id"
  INNER JOIN "customers" ON "customers"."deleted_at" IS NULL
  AND "customers"."id" = "applied_coupons"."customer_id"
WHERE "coupons"."deleted_at" IS NULL
  AND "coupons"."organization_id" = 'xyz'
  AND "customers"."deleted_at" IS NULL
  AND "coupons"."code" = 'test'
ORDER BY "applied_coupons"."created_at" DESC,
  "applied_coupons"."id" ASC
LIMIT 25 OFFSET 0;

                                                                 QUERY PLAN                                                                 
--------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=18.08..18.09 rows=1 width=160)
   ->  Sort  (cost=18.08..18.09 rows=1 width=160)
         Sort Key: applied_coupons.created_at DESC, applied_coupons.id
         ->  Nested Loop  (cost=4.43..18.07 rows=1 width=160)
               ->  Nested Loop  (cost=4.29..17.67 rows=1 width=160)
                     ->  Index Scan using index_coupons_on_organization_id_and_code on coupons  (cost=0.12..8.14 rows=1 width=16)
                           Index Cond: ((organization_id = 'xyz'::uuid) AND ((code)::text = 'test'::text))
                     ->  Bitmap Heap Scan on applied_coupons  (cost=4.16..9.50 rows=2 width=160)
                           Recheck Cond: (coupon_id = coupons.id)
                           ->  Bitmap Index Scan on index_applied_coupons_on_coupon_id  (cost=0.00..4.16 rows=2 width=0)
                                 Index Cond: (coupon_id = coupons.id)
               ->  Index Scan using customers_pkey on customers  (cost=0.14..0.27 rows=1 width=16)
                     Index Cond: (id = applied_coupons.customer_id)
                     Filter: ((deleted_at IS NULL) AND (deleted_at IS NULL))
```

There were also some missing scenarios in the request specs such as filtering by `status` or `customer_external_id`. I added some tests for those.

Note: The SQL query above show that we may have an unnecessary `.where(customers: {deleted_at: nil})` clause as the default scope of `Customer` is `-> { kept }`. This causes `"customers"."deleted_at" IS NULL` to be both in the `JOIN` and `WHERE` clauses.

### DIFF
--- a/spec/requests/api/v1/applied_coupons_controller_spec.rb
+++ b/spec/requests/api/v1/applied_coupons_controller_spec.rb
@@ -199,7 +199,7 @@ RSpec.describe Api::V1::AppliedCouponsController, type: :request do
       end
 
       context "when no applied coupons match the coupon code" do
-        let(:params) { {coupon_code: "non_existent_code"} }
+        let(:params) { {coupon_code: ["non_existent_code"]} }
 
         it "returns an empty array" do
           subject


lovrocolic (approved):
🚀 

Have you checked if we should update api clients and openapi for the new filter?

PR #3783, opened 2025-06-04
feat(stripe): Introduce integration tests (calling Stripe API)

julienbourdeau:
As part of the stripe upgrade, I added integration tests to call the Stripe API for real. 

By default, [it still runs `2020-08-27`](https://github.com/getlago/lago-api/blob/cc5711ed8515b6c7c8d4a60b435e2f5db5cf7f9a/spec/spec_helper.rb#L57) version but locally you can run the future version by setting the version in you `api/.env`
```env
STRIPE_API_VERSION=2025-04-30.basil
```

The integration tests are not run on CI. You must run them locally.


spec/integration/stripe/simple_payment_integration_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+        ]
+      }
+    })
+
+    expect(webhooks_sent.map { it["webhook_type"] }).to match_array(%w[
+      customer.payment_provider_created customer.created customer.checkout_url_generated
+    ])
+    webhooks_sent.clear
+
+    customer = Customer.find(json[:customer][:lago_id])
```
diegocharles, line 86: @julienbourdeau Nitpick: `json.dig(:customer, :lago_id)` :strongly

julienbourdeau: I agree I like `dig` a lot, I even used it in this same file.

PR #3787, opened 2025-06-05
feat(stripe): Prepare test suite to use Stripe 2025-04-30.basil

julienbourdeau:
This PR adds fixtures for `2025-04-30.basil`. I removed the forced version for some call if the client version is set.

The fixtures were reorganize a little to make it easier for next upgrades:

1. All webhook payload are under `webhooks/`.
2. The webhook event file name is the name of the webhook event (no suffix or extra description)
3. The content is always the full event (with `evt_*` id included)
4. HTTP call response fixutres are suffixed with `_response` but can have extra description in the name.

Payloads are copy pasted AS-IS from stripe and never edited.
If you need to update the payload for the purpose of your tests, for example setting the `metadata.lago_payable_id` (very typical), use the helper to modify the payload. **Don't edit the files.**

Next time we upgrade, we only have to copy the payload to a file.

```rb
let(:event_json) do
  get_stripe_fixtures("webhooks/setup_intent_succeeded.json") do |h|
    h[:data][:object][:customer] = "cus_666" if h[:data][:object][:customer].nil?
    h[:data][:object][:metadata] = metadata
  end
end
let(:stripe_customer) do
  create(:stripe_customer, payment_provider: stripe_provider, customer:, provider_customer_id: "cus_123")
end
```


app/services/payment_providers/stripe/payments/create_service.rb
```diff
@@ -78,8 +78,9 @@ def stripe_payment_method
           end
 
           # NOTE: Retrieve list of existing payment_methods
-          payment_method = ::Stripe::PaymentMethod.list(
-            {customer: provider_customer.provider_customer_id},
+          payment_method = ::Stripe::Customer.list_payment_methods(
+            provider_customer.provider_customer_id,
```
julienbourdeau, line 82: Notice that I changed this call because it's easier to mock the http request. The response payload is the same.

spec/support/stripe_helper.rb
```diff
@@ -1,7 +1,33 @@
...
+#
+#         stub_request(:get, %r{/v1/customers/#{provider_customer_id}$}).and_return(
+#           status: 200, body: get_stripe_fixtures("customer_retrieve_response.json") do |res|
+#             res["metadata"]["lago_customer_id"] = customer.id
+#           end
+#         )
 module StripeHelper
-  def get_stripe_fixtures(file, version: ENV.fetch("STRIPE_API_VERSION", "2020-08-27"))
-    File.read(Rails.root.join("spec/fixtures/stripe/#{version}/#{file}"))
+  def get_stripe_fixtures(file, version: ENV["STRIPE_API_VERSION"])
```
annvelents, line 22: I love this new way to handle overrides! 😍

annvelents (approved):
👏  Looks good 😁

spec/support/stripe_helper.rb
julienbourdeau: Thank you! I'm *very* happy with it ^^

PR #3788, opened 2025-06-05
fix(charge): fix cascading for child charges

lovrocolic:
## Context

This PR follows:
- https://github.com/getlago/lago-api/pull/3747
- https://github.com/getlago/lago-api/pull/3765
- https://github.com/getlago/lago-api/pull/3767

Updating a charge in a plan with a large number of override can cause Timeout error or lead to a memory error on Redis.

The main reason is:
- The service will enqueue one job per child charge

## Description

This PR is refactoring the logic to enqueue only one job responsible of the updates for all child charges.


app/services/plans/update_service.rb
```diff
@@ -110,21 +110,7 @@ def cascade_charge_update(charge, payload_charge)
...
-            params: payload_charge.deep_stringify_keys,
-            cascade_options: {
-              cascade: true,
-              parent_filters: charge.filters.map(&:attributes),
-              equal_properties: charge.equal_properties?(child_charge)
-            }
-          )
-        end
-      end
+      Charges::UpdateChildrenJob.perform_later(charge:, params: payload_charge.deep_stringify_keys)
```
floganz: I'm curious cam moving this to background job create problems for us?

Previously in `#process_charges` we checked the properties of child charge and parent charge like this
```ruby
cascade_charge_update(charge, payload_charge) # schedule all children changes
Charges::UpdateService.call(charge:, params: payload_charge).raise_if_error! # change parent
```
Now we'll push it background job and by the time of it's execution parent charge might already have different properties state which could lead to other outcome from `charge.equal_properties?(child_charge)` and cascade as result.

Have you checked this case?

spec/services/charges/update_children_service_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+        update_service.call
+
+        expect(child_charge.reload).to have_attributes(
+          properties: {"amount" => "500"}
+        )
+      end
+    end
+
+    context "when charge has no children" do
+      let(:child_charge) do
```
floganz, line 121: If charge has no children (according to context name) why we need to create `child_charge` with nil parent and check that it's not changed?

`it "does not update charge properties" do` could be deleted.

spec/services/charges/update_children_service_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+      before do
+        allow(Charges::UpdateService).to receive(:call!).and_call_original
+      end
+
+      it "calls the service" do
+        update_service.call
+
+        expect(Charges::UpdateService).not_to have_received(:call!)
+      end
```
floganz: Test called "calls the service" while the expectation `.not_to have_received`.

app/services/plans/update_service.rb
lovrocolic: Good catch! Actually that was the reason why cascade is always processed before parent charge. I will think of how to rework the logic

spec/services/charges/update_children_service_spec.rb
lovrocolic: ofc, the same results would be if there is no `child_charge` at all, in this case test DB is just extended for filtering (query)

PR #3794, opened 2025-06-06
fix: Use current attributes instead of activejob_traceable

rsempe:
Unfortunately, we've seen that our current use of `activejob_traceable` is not thread safe with Sidekiq 😢 

With the addition of activity logs, we've seen that `CurrentContext` can use wrong information (`membership_id` and `source`) because we're not resetting values at the start of each job. Two jobs can sequentially run on the same thread.

The goal of this PR is to not use `activejob_traceable` anymore (and remove a dependency 🎉 ), but introduce `ActiveSupport::CurrentAttributes` in Sidekiq jobs.

This middleware allows to:
- persist values from `CurrentContext` into the Sidekiq job
- restore those values before the job starts running
- reset values after the job finishes

```ruby
module Sidekiq
  module CurrentAttributes
    class Middleware
      def call(worker, job, queue)
        # Deserialize CurrentContext attributes into thread
        ::CurrentContext.set(job["current_attributes"]) do
          yield
        end
      ensure
        ::CurrentContext.reset_all
      end
    end
  end
end
```


spec/requests/api/base_controller_spec.rb
```diff
@@ -20,13 +20,18 @@ def track_api_key_usage?
...
+  end
+
   it "sets the context source to api" do
     request.headers["Authorization"] = "Bearer #{api_key.value}"
 
     get :index
 
-    expect(CurrentContext.source).to eq "api"
-    expect(CurrentContext.api_key_id).to eq api_key.id
+    expect(CurrentContext).to have_received(:source=).with("api")
```
groyoh, line 33: Why can't we keep the previous expectation ? The context is already reset at this point ?

rsempe: Yep - It's two different threads (rspec vs job), so as you said, context is already reset at this point 👍

PR #3795, opened 2025-06-06
fix(webhooks): Fix sending of `invoice.payment_status_updated` webhook on manual payment recording

groyoh:
## Context

We didn't send the `invoice.payment_status_updated` webhook if a manual payment is recorded on an invoice and the payment status is updated.

Steps to reproduce:

1. Find an overdue invoice.
2. Select the "record a payment" option.
3. Record a payment for the amount of the overdue invoice (this changes the payment status to "succeeded").
4. No webhook is triggered.

## Description

This PR fixes sending of `invoice.payment_status_updated` webhook on manual payment recording.


groyoh:
Thanks @rsempe @julienbourdeau 🙏

PR #3798, opened 2025-06-09
update tests

annvelents:
## Context

since we're switching dunning campaigns on billing entities, we need a way to manage the dunning campaigns on billing entities. This PR adds a Graphql mutation to update applied_dunning_campaign on provided billing entity

## Description

- Added a graphql mutation to update applied_dunning_campaign on billing entity
- added small fixes in service that changes applied dunning campaign on billing entity

### DIFF
--- a/app/config/permissions/definition.yml
+++ b/app/config/permissions/definition.yml
@@ -115,6 +115,8 @@ billing_entities:
   emails:
     view:
     update:
+  dunning_campaigns:
+    manage:
 payment_requests:
   view: true
   create:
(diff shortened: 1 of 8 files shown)


app/config/permissions/definition.yml
```diff
@@ -115,6 +115,8 @@ billing_entities:
   emails:
     view:
     update:
+  dunning_campaigns:
+    manage:
```
ancorcruz: is `manage` a new kind of permission? we usually have `view`, `create`, `update` and `delete` as permissions.

annvelents: ah, you're right, it should be update 🤦 
what happens when you're making decisions tired on friday 😅

PR #3803, opened 2025-06-09
ftr(wallet-limitations): apply wallet with limitations on the invoice

lovrocolic:
## Context

Currently wallet credits are applied on all fee types. This feature adds possibility to limit wallet consumption on specific fee type.

## Description

This PR adds logic that applies wallet credits with fee limitations on certain invoice. Other pieces and pre-requirements are already implemented.

### DIFF
--- a/app/services/credits/applied_prepaid_credit_service.rb
+++ b/app/services/credits/applied_prepaid_credit_service.rb
@@ -70,7 +70,17 @@ module Credits
     end
 
     def compute_amount
-      [balance_cents, invoice.total_amount_cents].min
+      if wallet.limited_fee_types?
+        applicable_fees = invoice.fees.where(fee_type: wallet.allowed_fee_types)
+
+        total = 0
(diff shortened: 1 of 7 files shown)


app/services/credits/applied_prepaid_credit_service.rb
```diff
@@ -70,7 +70,17 @@ def already_applied?
     end
 
     def compute_amount
-      [balance_cents, invoice.total_amount_cents].min
+      if wallet.limited_fee_types?
+        applicable_fees = invoice.fees.where(fee_type: wallet.allowed_fee_types)
```
julienbourdeau: Maybe we want to use `filter` to avoid making a db query. But maybe we want to make a db queries to get the fresh data :D just sayin'

lovrocolic: good catch, just updated it 👍

PR #3807, opened 2025-06-10
chore(clickhouse): Enable charge caching

vincent-pochet:
## Description

Following the work done in https://github.com/getlago/lago/pull/542, this PR enable the caching logic for organization relying on the Clickhouse event store.

The cache expiration will be handled in the events-processor service.


app/services/invoices/customer_usage_service.rb
```diff
@@ -99,7 +99,7 @@ def charge_usage(charge)
         charge:,
         to_datetime: boundaries[:charges_to_datetime],
         # NOTE: Will be turned on for clickhouse in the future
```
groyoh:
```suggestion
```

PR #3813, opened 2025-06-10
revert(email): Remove Webhook notification in case of failed

julienbourdeau:
## Context

We recently [added this feature](https://github.com/getlago/lago-api/pull/3654), then [disabled it](https://github.com/getlago/lago-api/pull/3776). This PR removes it completely.

## Description

Many customers have organization with failing endpoints (like staging or dev env) and it's expected because they want to see the webhooks in the dashboard.

Ideally, the email should be occasional (like Stripe does). This would require more time and attention that original thought and this is very low priority.

I believe it's best to revert it completely for now.


spec/mailers/previews/webhook_mailer_preview.rb
```diff
@@ -1,9 +0,0 @@
-# frozen_string_literal: true
-
-class WebhookMailerPreview < BasePreviewMailer
```
groyoh, line 3: For my curiousity, how do we use this ?

groyoh (approved):

spec/mailers/previews/webhook_mailer_preview.rb
julienbourdeau: You can go to https://api.lago.dev/rails/mailers/ locally to get a list of all previews. Super useful when developing and email has a few variations.
https://guides.rubyonrails.org/action_mailer_basics.html#previewing-and-testing-mailers

CleanShot 2025-06-11 at 09 38 12@2x

groyoh: Neat! I didn't know about that. I guess I'm more used to working with Sendgrid, Mailchimp, etc. than Rails mailer.

PR #3815, opened 2025-06-10
Migration add voided invoice

diegocharles:
## Roadmap Task

https://www.notion.so/getlago/BE-Dive-In-Regenerate-Voided-Invoice-208ef63110d280158976c5d15a23dfd7

## Context

## Description

### DIFF
--- a/db/migrate/20250610173034_add_voided_invoice_id_to_invoices.rb
+++ b/db/migrate/20250610173034_add_voided_invoice_id_to_invoices.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddVoidedInvoiceIdToInvoices < ActiveRecord::Migration[8.0]
+  def change
+    safety_assured do
+      add_reference :invoices, :voided_invoice, type: :uuid, foreign_key: {to_table: :invoices}, index: true
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20250610173034_add_voided_invoice_id_to_invoices.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddVoidedInvoiceIdToInvoices < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :invoices, :voided_invoice, type: :uuid, index: {algorithm: :concurrently}
```
julienbourdeau, line 7: You're adding the index because you expect to find the regenerated invoices from the voided invoice, like `voided_invoice.regenerated_invoice`?
If you're not sure yet, you can wait and create the index only if needed as you implement the regeneration 👍

julienbourdeau (approved):

db/migrate/20250610173034_add_voided_invoice_id_to_invoices.rb
diegocharles: @julienbourdeau That's exactly how I'm planning to use in the Service.

vincent-pochet, line 7: You should add a foreign key as well, to ensure integrity

vincent-pochet (approved):
Could you add a bit of context in the PR description. Also, you should remove the link to the Notion page as it's not a public document :) .

brunomiguelpinto (approved):

diegocharles:
> we need to check this we use parent id

As discussed in DM that's not necessary for now.

PR #3816, opened 2025-06-11
Api endpoints api logs

mariohd:
Api Logs endpoints.


app/controllers/api/v1/api_logs_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+        )
+      end
+
+      private
+
+      def resource_name
+        "api_log"
+      end
+
+      def index_filters
```
rsempe, line 54: `api_key_ids` is missing, right?

app/queries/api_logs_query.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    api_logs = with_logged_at_range(api_logs) if filters.from_date || filters.to_date
+    api_logs = with_api_key_ids(api_logs) if filters.api_key_ids.present?
+    api_logs = with_http_statuses(api_logs) if filters.http_statuses.present?
+    api_logs = with_http_methods(api_logs) if filters.http_methods.present?
+    api_logs = with_api_version(api_logs) if filters.api_version.present?
+    api_logs = with_request_paths(api_logs) if filters.request_paths.present?
+
+    Clickhouse::ApiLog.transaction do
+      Clickhouse::ApiLog.connection.execute("SELECT 1 SETTINGS max_execution_time=5000") if filters.request_paths.present?
```
rsempe: why this value `5000`?

app/queries/api_logs_query.rb
```diff
@@ -0,0 +1,71 @@
...
+  def with_api_key_ids(scope)
+    scope.where(api_key_id: filters.api_key_ids)
+  end
+
+  def with_http_methods(scope)
+    scope.where(http_method: filters.http_methods)
+  end
+
+  def with_http_statuses(scope)
+    if (filters.http_statuses & %w[succeeded failed]).any?
```
rsempe:
```suggestion
    if filters.http_statuses.any? { |s| %w[succeeded failed].include?(s) }
```

seems better for clarity for me 😅 but happy to keep your version as well (using intersection) if you want

app/queries/api_logs_query.rb
mariohd: that's a "random" timeout, 5s. I really don't have any idea what would be the good timeout number. 

Any suggestion?

app/controllers/api/v1/api_logs_controller.rb
mariohd: that's an open question...

https://github.com/getlago/lago-api/blob/bc6d90f2d929adb752ff08cea54348d55f9366ac/app/controllers/api/v1/activity_logs_controller.rb#L54-L65

does not have api_keys as well. Are we allowing the API to search by api_keys only for api_logs?

rsempe: ohh my bad you're right, that's only possible from graphql 👍

app/queries/api_logs_query.rb
rsempe: maybe nothing for now? 😅

mariohd: sounds good!

PR #3817, opened 2025-06-11
Add PricingUnitUsage model

floganz:
## Context

Another step towards PricingUnit feature, this time adding new model PricingUnitUsage. It will store usage in pricing units for a Fee, leaving Fee model itself not changed.

## Description

Add PricingUnitUsage model.
Adjustments to the Fees::ChargeService.

### DIFF
--- a/db/migrate/20250610063400_create_pricing_unit_usages.rb
+++ b/db/migrate/20250610063400_create_pricing_unit_usages.rb
@@ -11,7 +11,7 @@ class CreatePricingUnitUsages < ActiveRecord::Migration[8.0]
       t.bigint :amount_cents, null: false
       t.decimal :precise_amount_cents, precision: 40, scale: 15, default: 0.0, null: false
       t.bigint :unit_amount_cents, default: 0, null: false
-      t.decimal :conversion_rate, precision: 40, scale: 15, default: "0.0", null: false
+      t.decimal :conversion_rate, precision: 40, scale: 15, default: 0.0, null: false
 
       t.timestamps
     end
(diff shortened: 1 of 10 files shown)


db/migrate/20250610063400_create_pricing_unit_usages.rb
```diff
@@ -0,0 +1,19 @@
...
+    create_table :pricing_unit_usages, id: :uuid do |t|
+      t.references :organization, null: false, foreign_key: true, type: :uuid
+      t.references :fee, null: false, foreign_key: true, type: :uuid
+      t.references :pricing_unit, null: false, foreign_key: true, type: :uuid
+
+      t.string :short_name, null: false
+      t.bigint :amount_cents, null: false
+      t.decimal :precise_amount_cents, precision: 40, scale: 15, default: 0.0, null: false
+      t.bigint :unit_amount_cents, default: 0, null: false
+      t.decimal :conversion_rate, precision: 40, scale: 15, default: "0.0", null: false
```
julienbourdeau: I don't know if it matters but one default is float `0.0`, the other is string `"0.0"`

spec/factories/pricing_unit_usages.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :pricing_unit_usage do
+    fee
+    pricing_unit
+    short_name { pricing_unit.short_name }
+    amount_cents { 200 }
+    precise_amount_cents { 200.0000000001 }
```
julienbourdeau: I'm always a little confused with decimal. Should this be `BigDecimal("200.0000000001")` ?

db/migrate/20250610063400_create_pricing_unit_usages.rb
floganz: That doesn't matter, in structure file it the same value but I'll change it reduce confusion for all of us.

PR #3818, opened 2025-06-11
Feat: regenerate invoice from void service

diegocharles:
## Context

## Description

### DIFF
--- a/app/controllers/api/v1/invoices_controller.rb
+++ b/app/controllers/api/v1/invoices_controller.rb
@@ -142,6 +142,18 @@ module Api
         end
       end
 
+      def regenerate
+        voided_invoice = current_organization.invoices.voided.find_by(id: params[:id])
+        return not_found_error(resource: "invoice") unless voided_invoice
+
+        result = Invoices::Voids::RegenerateFromVoidedInvoiceService.call(voided_invoice_id: voided_invoice.id)
+        if result.success?
+          render_invoice(result.invoice)
+        else
+          render_error_response(result)
+        end
+      end
+
       def lose_dispute
         invoice = current_organization.invoices.visible.find_by(id: params[:id])
 
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -593,5 +593,4 @@ end
 #  fk_rails_...  (billing_entity_id => billing_entities.id)
 #  fk_rails_...  (customer_id => customers.id)
 #  fk_rails_...  (organization_id => organizations.id)
-#  fk_rails_...  (voided_invoice_id => invoices.id)
 #
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -20,6 +20,7 @@ class Invoice < ApplicationRecord
   belongs_to :customer, -> { with_discarded }
   belongs_to :organization
   belongs_to :billing_entity, optional: true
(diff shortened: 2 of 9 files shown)


app/graphql/types/invoices/fee_input.rb
brunomiguelpinto, line 1: this should be a new mutation

app/services/invoices/create_generating_service.rb
brunomiguelpinto, line 1: This should be responsibility of our service, the regenerate one 
CreateGeneratingServic,e this is used in multiple places that don't need to care with this attribute

diegocharles: Agreed @brunomiguelpinto

app/graphql/types/invoices/fee_input.rb
diegocharles: @brunomiguelpinto I was thinking after we talked about that: All that we need is just advise the front to send :add_on_id: null in the mutation. I think the same goes for others. 
Do you think this tradeoff won't be paid when we're saving the codebase from having one duplicated FeeInput instead of reutilize the existent one? ❓

app/services/invoices/regenerate_from_voided_service.rb
```diff
@@ -0,0 +1,119 @@
...
+              allowed_attrs = %i[
+                charge_id
+                subscription_id
+                invoice_display_name
+                units
+                description
+                amount_cents
+                unit_amount_cents
+                add_on_id
+              ]
```
vincent-pochet: I guess you could turn this into a constant

app/services/invoices/regenerate_from_voided_service.rb
```diff
@@ -0,0 +1,119 @@
...
+
+        result.invoice = regenerated_invoice
+      end
+
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    rescue BaseService::FailedResult => e
+      e.result
+    rescue => e
```
vincent-pochet: I think you could let it raise so that we can easily fix it.

app/services/invoices/regenerate_from_voided_service.rb
```diff
@@ -0,0 +1,119 @@
...
+      @fees = fees
+      super
+    end
+
+    activity_loggable(
+      action: "invoice.regenerated_from_voided",
+      record: -> { voided_invoice }
+    )
+
+    def call
```
vincent-pochet, line 18: The logic looks great, but if you don't mind, I would suggest to split the logic in few private methods here to ease readability :)

diegocharles: @vincent-pochet That is definitely on the roadmap for the very last stage after the QA stage. 
I'm keeping this procedural approach until now just to avoid multiple eyes jumping while we're still implementing the logic and inspecting possible failures.

app/services/invoices/regenerate_from_voided_service.rb
diegocharles: @vincent-pochet I probably agree. I intentionally left the implementation more procedural for now but I'm performing a refactoring to be pushed after we no longer need to care about the QA stage.

PR #3821, opened 2025-06-12
fix(dates): Fix validation of impossible dates

ivannovosad:
## Context

It was possible to call `GET /api/v1/invoices` API endpoint with impossible `issuing_date_from`, `issuing_date_to` params e.g.: `2025-02-30`

## Description

Fix the validation method: `Common#valid_date?`

### DIFF
--- a/app/controllers/concerns/common.rb
+++ b/app/controllers/concerns/common.rb
@@ -8,10 +8,9 @@ module Common
   def valid_date?(date)
     return false unless date
 
-    parsed_date = Date._strptime(date)
-
-    return false unless parsed_date
-
+    Date.strptime(date)
     true
+  rescue Date::Error
+    false
   end
 end
--- a/spec/controllers/concerns/common_spec.rb
+++ b/spec/controllers/concerns/common_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.describe Common do
-  subject(:controller) { klass.new }
+  let(:controller) { klass.new }
 
   let(:klass) do
     Class.new do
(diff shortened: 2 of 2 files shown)


app/controllers/concerns/common.rb
floganz, line 1: I've noticed that we have already quite similar method to check if provided date is correct `Utils::Datetime.valid_format?(datetime)` maybe it worth adjusting it implementation and reuse it here?
Or remove this method at all and use that service directly where needed?
```ruby
def self.valid_format?(datetime)
    datetime.respond_to?(:strftime) || datetime.is_a?(String) && Time.zone.parse(datetime.to_s).present?
  rescue ArgumentError
    false
  end
```

ivannovosad: That method returns true also for a date that does not exists (even the format itself is valid).
February does not have 30 days:
```
lago-api(dev)> Utils::Datetime.valid_format? "2024-02-30"
=> true
```


Then we call `Date.strptime` which raises an error for this reason:
```
issuing_date_to: (Date.strptime(params[:issuing_date_to]) if valid_date?(params[:issuing_date_to])),
```

floganz: I understand that existing service might not catch your case but we'll most likely want to catch the same case also for places where `Utils::Datetime.valid_format?(datetime)` is used. That's why I suggested to build one validator for date instead of having two of them.

ivannovosad: I see, that one also works with date objects is used in multiple services. Lets create a separate issue that will handle unifying those 2 methods.

PR #3822, opened 2025-06-12
misc(organization): Ensure organization_id is added everywhere

vincent-pochet:
## Context

This PR is part of the epic to add `organization_id` on all tables of the application

## Description

The current one is refactoring some factories and ensures that all created records received the correct organization_id. The next step of this epic will be to mark the `organization_id` as required

### DIFF
--- a/spec/services/billing_entities/taxes/manage_taxes_service_spec.rb
+++ b/spec/services/billing_entities/taxes/manage_taxes_service_spec.rb
@@ -79,9 +79,13 @@ RSpec.describe BillingEntities::Taxes::ManageTaxesService do
         let(:tax_codes) { ["tax_code_1", "TAX_CODE_2"] }
 
         it "matches tax codes case-insensitively" do
-          service.call
+          result = service.call
 
-          expect(billing_entity.taxes).to eq([tax1, tax2])
+          expect(result).to be_success
+          expect(result.taxes).to eq([tax1, tax2])
+          expect(result.applied_taxes.count).to eq(2)
+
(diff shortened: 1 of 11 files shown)


spec/services/billing_entities/taxes/manage_taxes_service_spec.rb
```diff
@@ -81,7 +81,8 @@
         it "matches tax codes case-insensitively" do
           service.call
 
-          expect(billing_entity.taxes).to eq([tax1, tax2])
+          expect(billing_entity.reload.taxes).to eq([tax1, tax2])
```
groyoh: Could you explain why we need to reload now ?

spec/services/billing_entities/taxes/manage_taxes_service_spec.rb
```diff
@@ -81,7 +81,8 @@
         it "matches tax codes case-insensitively" do
           service.call
```
groyoh: Side note: I believe we never check the result `taxes` and `applied_taxes`.

vincent-pochet: Good catch! I will had one :+1:

spec/services/billing_entities/taxes/manage_taxes_service_spec.rb
vincent-pochet: We are now assigning, applied_taxes, but we are still testing taxes. As the taxes were assigned before, the result is cached. As it's not ideal, I will update the test to avoid it.

PR #3823, opened 2025-06-12
feat(stripe): Fix Stripe.api_version to `2025-04-30.basil`

julienbourdeau:
## Roadmap Task

## Context

Everything you need to know: https://github.com/getlago/lago-api/pull/3300

## Description

From now on, all Stripe API calls made by lago will be made using the API version `2025-04-30.basil`.

Lago users can continue to use Stripe in any versions they like on their side. They can keep their stripe account configured with any version too.

Until now, Lago would use the user's account version, which could lead to different response shape, breaking the code. It never did yet because we use very core and simple feature of Stripe, but it could in the future.

### DIFF
--- a/app/services/payment_provider_customers/stripe/retrieve_latest_payment_method_service.rb
+++ b/app/services/payment_provider_customers/stripe/retrieve_latest_payment_method_service.rb
@@ -42,8 +42,7 @@ module PaymentProviderCustomers
 
       def request_options
         {
-          api_key:,
-          stripe_version: ::Stripe.api_version || "2024-09-30.acacia" # TODO: Remove once Stripe.api_version is fully locked
+          api_key:
         }
       end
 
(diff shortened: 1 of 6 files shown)


app/services/payment_provider_customers/stripe/retrieve_latest_payment_method_service.rb
```diff
@@ -42,8 +42,7 @@ def call
 
       def request_options
         {
-          api_key:,
-          stripe_version: ::Stripe.api_version || "2024-09-30.acacia" # TODO: Remove once Stripe.api_version is fully locked
```
julienbourdeau, line 46: When I needed to add calls to stripe recently, I had forced the version to a recent one, to make the upgrade easier. This should now be removed.

vincent-pochet (approved):

PR #3826, opened 2025-06-12
fix(no-customer-on-payment):early return when payment method has no c…

mariohd:
Whenever we have a `setup_intent.succeeded` with a payment method that has no customer attached to it. 

Stripe does that when setting up payment methods that has some problems (we don't know exactly which problems are). The `setup_intent.succeeded` and `payment_method.detached` are fired almost at the same time. 

[Once `payment_method.detached` the payment method becomes unusable. 
](https://docs.stripe.com/api/payment_methods/detach)

This raises on `app/services/payment_providers/stripe/webhooks/setup_intent_succeeded_service.rb`
```
Stripe::InvalidRequestError: The customer does not have a payment method with the ID <pm_ID>. The payment method must be attached to the customer.
```

### DIFF
--- a/app/services/payment_providers/stripe/webhooks/setup_intent_succeeded_service.rb
+++ b/app/services/payment_providers/stripe/webhooks/setup_intent_succeeded_service.rb
@@ -9,6 +9,7 @@ module PaymentProviders
         def call
           return result if stripe_customer_id.nil?
           return handle_missing_customer unless stripe_customer
+          return result unless valid_payment_method?
 
           update_stripe_customer_default_payment_method
           result.payment_method_id = payment_method_id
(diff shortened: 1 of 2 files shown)


app/services/payment_providers/stripe/webhooks/setup_intent_succeeded_service.rb
```diff
@@ -40,6 +41,13 @@ def payment_method_id
           event.data.object.payment_method
         end
 
+        def valid_payment_method?
```
groyoh, line 44: Maybe leave a comment here to explain what you put in the PR description ?

groyoh (approved):

app/services/payment_providers/stripe/webhooks/setup_intent_succeeded_service.rb
mariohd: The code is also linked to the PR via blame. I don't think the comment is really necessary in the code as well.

PR #3828, opened 2025-06-13
feat: Ability to resolve customer and subscription by external id

rsempe:
- Resolve a customer by `external_id`
- Resolve a subscription by `external_id`

It will be used for audit logs for instance, to display a link to the related customer and subscription.

### DIFF
--- a/app/graphql/resolvers/subscription_resolver.rb
+++ b/app/graphql/resolvers/subscription_resolver.rb
@@ -20,7 +20,10 @@ module Resolvers
       end
 
       return current_organization.subscriptions.find(id) if id.present?
-      current_organization.subscriptions.find_by!(external_id:)
+
+      current_organization.subscriptions
+        .order("terminated_at DESC NULLS FIRST, started_at DESC")
+        .find_by!(external_id:)
     rescue ActiveRecord::RecordNotFound
       not_found_error(resource: "subscription")
     end
--- a/app/graphql/resolvers/subscription_resolver.rb
+++ b/app/graphql/resolvers/subscription_resolver.rb
@@ -9,12 +9,18 @@ module Resolvers
 
     description "Query a single subscription of an organization"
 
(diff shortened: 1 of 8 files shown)


app/graphql/resolvers/subscription_resolver.rb
```diff
@@ -9,12 +9,18 @@ class SubscriptionResolver < Resolvers::BaseResolver
...
 
-    def resolve(id: nil)
-      current_organization.subscriptions.find(id)
+    def resolve(id: nil, external_id: nil)
+      if id.nil? && external_id.nil?
+        raise GraphQL::ExecutionError, "You must provide either `id` or `external_id`."
+      end
+
+      return current_organization.subscriptions.find(id) if id.present?
+      current_organization.subscriptions.find_by!(external_id:)
```
vincent-pochet: As multiple subscription could match the external id, maybe we could use something similar to what we do in the `Event#subscription` method:

```ruby
current_organization.subscriptions
  .where(external_id:)
  .order("terminated_at DESC NULLS FIRST, started_at DESC")
  .first
```
It could make the result more predictable by always returning first the currently active subscription or the most recently terminated one.


What do you think?

vincent-pochet (approved):

app/graphql/resolvers/subscription_resolver.rb
groyoh: By the way @vincent-pochet @rsempe , I recently noticed the same behaviour on the API when updating the documentation: 

https://github.com/getlago/lago-openapi/blob/main/src/resources/subscription.yaml#L21

I guess we could update the behaviour there as well. What do you think ?

vincent-pochet: It worth a look indeed

rsempe: Good spot @vincent-pochet @groyoh 👍

PR #3830, opened 2025-06-16
feat(pricing-units): Add GQL for PricingUnit model

floganz:
## Context

Adding GQL interface for managing PricingUnit.

## Description

Add pricing unit related permissions
Add pricing unit type, resolvers, create and update mutations


spec/graphql/resolvers/pricing_units_resolver_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  let(:membership) { create(:membership) }
+  let(:required_permission) { "pricing_units:view" }
+  let!(:pricing_unit) { create(:pricing_unit, organization: membership.organization) }
+
+  it_behaves_like "requires current user"
+  it_behaves_like "requires current organization"
+  it_behaves_like "requires permission", "pricing_units:view"
+
+  it "returns a list of pricing units" do
```
groyoh, line 44: I guess we could test pagination here.

groyoh (approved):

spec/graphql/resolvers/pricing_units_resolver_spec.rb
floganz: During test noticed that pagination actually doesn't work because of use `.limit()` instead of `.per()`.
Fixed the implementation and also place where I copied that code.

PR #3835, opened 2025-06-16
chore(api-key): Add pagination and ordering tests for API Keys GraphQL endpoint

groyoh:
## Context

This PR was originally created to fix the API Keys GraphQL resolver (See previous description below). But it was also fixed and merged in https://github.com/getlago/lago-api/pull/3830. Since I added some tests, I rebased and kept the tests.

## Description

This PR adds tests for ordering and paginating the API Key GraphQL endpoint.

---

_Initial description below_

## Context

The pagination for the API Keys GraphQL resolver is not working properly.

## Description

We used the following query:

```ruby
current_organization.api_keys.order(created_at: :asc).page(page).limit(limit)
```

which resulted in the following SQL:

```sql
SELECT "api_keys"."id",
  "api_keys"."organization_id",
  "api_keys"."value",
  "api_keys"."created_at",
  "api_keys"."updated_at",
  "api_keys"."expires_at",
  "api_keys"."last_used_at",
  "api_keys"."name",
  "api_keys"."permissions"
FROM "api_keys"
WHERE (
    expires_at IS NULL
    OR expires_at > '2025-06-16 12:41:11.750724'
  )
  AND "api_keys"."organization_id" = '...'
ORDER BY "api_keys"."created_at" ASC
LIMIT 5 OFFSET 25
```

This is due to Kaminari's way of handling pagination, which requires the `page` and `per` methods to be called in a specific order. `per` will then properly override the `limit` and `offset`. If `per` is not called, Kaminari will define the `OFFSET` using the `page` and default page size.

### DIFF
--- a/spec/graphql/resolvers/api_keys_resolver_spec.rb
+++ b/spec/graphql/resolvers/api_keys_resolver_spec.rb
@@ -47,8 +47,8 @@ RSpec.describe Resolvers::ApiKeysResolver, type: :graphql do
   context "when pagination is provided" do
     let(:query) do
       <<~GQL
-        query {
-          apiKeys(limit: 2, page: 2) {
+        query($limit: Int, $page: Int) {
+          apiKeys(limit: $limit, page: $page) {
             collection { id value name createdAt }
             metadata { currentPage, totalCount totalPages}
           }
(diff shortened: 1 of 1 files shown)


vincent-pochet (approved):

spec/graphql/resolvers/api_keys_resolver_spec.rb
```diff
@@ -37,13 +35,43 @@
...
+          apiKeys(limit: 2, page: 2) {
+            collection { id value name createdAt }
+            metadata { currentPage, totalCount totalPages}
+          }
+        }
+      GQL
+    end
+
+    before do
+      3.times do |i|
```
floganz: If we want to keep this test I'd suggest to reduce amount of records created minimum. There's no benefit to make setup bigger than necessary.
We can test the order and pagination using only 2 record by checking 2nd page with per page limit 1.

groyoh: Sure.

PR #3844, opened 2025-06-17
Fix: Retry ViesCheck after rescuing from any Valvat Maintenance exception

diegocharles:
### Problem
`Valvat::MemberStateUnavailable` exceptions weren't being raised properly, preventing fallback mechanisms from triggering when VIES is unavailable.

Some countries won't accept requests during the weekend of after the business hour. 
Check the table here https://ec.europa.eu/taxation_customs/vies/#/help

### Solution

1. Added `raise_error: true` to Valvat calls in `Customers::EuAutoTaxesService` to ensure exceptions are properly raised and caught by the existing rescue block
2. Update the wait time from 30 seconds to 15.minutes, in order to avoid overheads in the Sidekiq, since the job will run until it gets the information. 

### Next steps: 
  Rethink the retry logic to make it exponential using the Sidekiq built-in helper, or create a table with the working hours for each country.

### Testing
- ✅ All existing tests pass
- ✅ VIES exceptions now properly trigger fallback mechanism
- ✅ Webhook notifications work correctly
- ✅ Retry job scheduling functions as expected

### Impact
Better reliability for German customers when VIES service is temporarily unavailable, ensuring consistent tax application through proper fallback mechanisms.

---

**Context:**
```ruby
# Before: Exception not raised
Valvat.new("DE199896394").exists?(detail: true)

# After: Exception properly raised and handled
Valvat.new("DE199896394").exists?(detail: true, raise_error: true)
```

**What happens now:**
1. When VIES is unavailable, `Valvat::MemberStateUnavailable` is properly raised
3. Existing rescue block captures the exception
4. Webhook is sent with error details
5. Retry job is scheduled for 30 seconds later


app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -39,7 +39,7 @@ def check_vies
       after_commit { SendWebhookJob.perform_later("customer.vies_check", customer, vies_check: response.presence || error_vies_check) }
 
       response
-    rescue Valvat::RateLimitError, Valvat::Timeout, Valvat::BlockedError, Valvat::InvalidRequester => e
+    rescue Valvat::RateLimitError, Valvat::Timeout, Valvat::BlockedError, Valvat::InvalidRequester, Valvat::MemberStateUnavailable => e
```
mariohd: You also need to add the param `raise_error: true` here as well
https://github.com/getlago/lago-api/blob/571cfaa2ea2eb6b3c6d36bb4100e0fa33435f45b/app/services/customers/eu_auto_taxes_service.rb#L37

Only with that you can have the exception `Valvat::MemberStateUnavailable`. 
Also, if you saying that the job succeeded (not raising any exception), you don't need any retry configured.

Either have the retry or call the job again.

app/jobs/customers/retry_vies_check_job.rb
```diff
@@ -3,6 +3,11 @@
 module Customers
   class RetryViesCheckJob < ApplicationJob
     queue_as :default
+    sidekiq_options retry: 5
+
+    sidekiq_retry_in do |count|
+      [30.seconds * (2**count), 1.hour].min
```
mariohd: there's no sense in having the `min` here.

With max 5 retrys, to have 30 seconds * 2**5, which is 16 minutes...
```
lago-api(dev)> 960.seconds/1.minute
=> 16
```

`[16.minutes, 1.hour].min` will always be `16.min`.

Either remove the `min` or increase the number 2 to 3+.

diegocharles: I think this logic is fine because we can quick tune the retry number as we get more data.

app/services/customers/eu_auto_taxes_service.rb
diegocharles: @mariohd I'm not sure if I'm following: 
You say that if there's a `Valvat::RateLimitError` for example, the rescue block will never run and the         `RetryViesCheckJob` won't be scheduled?

diegocharles: @mariohd Ok, I read the documentation and I understand now why some exceptions raises exceptions while others does not. 

https://github.com/yolk/valvat/blob/master/README.md#handling-of-maintenance-errors

mariohd: Awesome! 
https://ec.europa.eu/taxation_customs/vies/#/help

mariohd: What I say that you should have either retry OR rescue

it's because rescue DOES NOT RETRY, it pretend that job executed successfuly although scheduling other job 30secs later, even if it was raising an error and rescuing. 

You either have to raise the error to let Sidekiq retry, or handle the retry yourself using a rescue clause.

diegocharles: @mariohd Exactly! I can see it after I understood the difference from the gem docs. Thank you!

diegocharles:
@mariohd There are some overthinking from my end 💥

spec/services/customers/eu_auto_taxes_service_spec.rb
```diff
@@ -72,6 +72,35 @@
...
+        end
+
+        it "enqueues RetryViesCheckJob" do
+          spy = class_double(Customers::RetryViesCheckJob, perform_later: true).as_null_object
+          allow(Customers::RetryViesCheckJob).to receive(:set).with(wait: 30.seconds).and_return(spy)
+
+          eu_tax_service.call
+
+          expect(Customers::RetryViesCheckJob).to have_received(:set).with(wait: 30.seconds)
+          expect(spy).to have_received(:perform_later).with(customer.id)
```
groyoh:
```suggestion
          eu_tax_service.call

          expect(Customers::RetryViesCheckJob).to have_been_enqueued.at(25.seconds.from_now..35.seconds.from_now).with(customer.id).once
```

app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -34,12 +34,14 @@ def call
...
-      response = Valvat.new(customer.tax_identification_number).exists?(detail: true)
+      # Just errors extended from Valvat::Lookup are raised, while Maintenances are not.
+      # https://github.com/yolk/valvat/blob/master/README.md#handling-of-maintenance-errors
+      response = Valvat.new(customer.tax_identification_number).exists?(detail: true, raise_error: true)
 
       after_commit { SendWebhookJob.perform_later("customer.vies_check", customer, vies_check: response.presence || error_vies_check) }
 
       response
-    rescue Valvat::RateLimitError, Valvat::Timeout, Valvat::BlockedError, Valvat::InvalidRequester => e
+    rescue Valvat::RateLimitError, Valvat::Timeout, Valvat::BlockedError, Valvat::InvalidRequester, Valvat::MemberStateUnavailable => e
```
groyoh: Based on the doc, we also have:

```suggestion
    rescue Valvat::RateLimitError, Valvat::Timeout, Valvat::BlockedError, Valvat::InvalidRequester, Valvat::MemberStateUnavailable, Valvat::ServiceUnavailable => e
```

brunomiguelpinto:
We created an helper that might be useful on this  ConcurrencyThrottlable, have a look to check if it's useful

spec/services/customers/eu_auto_taxes_service_spec.rb
```diff
@@ -72,6 +72,59 @@
...
+          }).once
+        end
+
+        it "enqueues RetryViesCheckJob" do
+          spy = class_double(Customers::RetryViesCheckJob, perform_later: true).as_null_object
+          allow(Customers::RetryViesCheckJob).to receive(:set).with(wait: 15.minutes).and_return(spy)
+
+          eu_tax_service.call
+
+          expect(Customers::RetryViesCheckJob).to have_received(:set).with(wait: 15.minutes)
```
groyoh: Could you update this to use `have_been_enqueued` as well ?

app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -34,16 +34,21 @@ def call
...
 
       response
-    rescue Valvat::RateLimitError, Valvat::Timeout, Valvat::BlockedError, Valvat::InvalidRequester => e
+    rescue Valvat::RateLimitError, Valvat::Timeout, Valvat::BlockedError, Valvat::InvalidRequester,
+      Valvat::ServiceUnavailable, Valvat::MemberStateUnavailable => e
       after_commit do
         SendWebhookJob.perform_later("customer.vies_check", customer, vies_check: error_vies_check.merge(error: e.message))
         # Enqueue a job to retry the VIES check after a delay
-        RetryViesCheckJob.set(wait: 30.seconds).perform_later(customer.id)
+        RetryViesCheckJob.set(wait: 15.minutes).perform_later(customer.id)
```
groyoh: Maybe we could keep the 30.seconds in most cases:

```suggestion
        delay = case e
        when Valvat::ServiceUnavailable, Valvat::MemberStateUnavailable
          15.minutes
        else
          30.seconds
        end
        RetryViesCheckJob.set(wait: delay).perform_later(customer.id)
```

groyoh (approved):

app/services/customers/eu_auto_taxes_service.rb
diegocharles: @groyoh That's a very nice point man. 
My point here is that we don't want overheads on the queue and we have no guarantee that non Maintenance errors will be resolved in a matter of few seconds. 

Perhaps 15 minutes is too much though and we can reduce it to 5 minutes for all?

groyoh: To be fair, I have no strong opinion on this. But since I don't know how this service is meant to behave, I'd have kept the current logic for non-maintenance errors.

I guess 5 minutes should be a good compromise.

PR #3848, opened 2025-06-18
feat: Generate credit note validate License

brunomiguelpinto:
## Context
Voiding an invoice with the [generate_credit_note]

flag should be exclusive to organizations on a Premium license.  
Prior to this change, non-premium tenants could still try to generate credit notes through the void-invoice path, bypassing licensing rules this would lead to a 500 error on the server 

## Description
Added a guard clause in `Invoices::VoidService#call`:

```ruby
    def generate_credit_note_allowed?
      return true unless generate_credit_note   # nothing requested
      License.premium?                          # licence check
    end
```

### DIFF
--- a/app/services/invoices/void_service.rb
+++ b/app/services/invoices/void_service.rb
@@ -18,6 +18,7 @@ module Invoices
 
     def call
       return result.not_found_failure!(resource: "invoice") unless invoice
+      return result.forbidden_failure! if generate_credit_note && !License.premium?
       return result.not_allowed_failure!(code: "not_voidable") if invoice.voided?
       return result.not_allowed_failure!(code: "not_voidable") if !invoice.voidable? && !explicit_void_intent?
       unless valid_credit_note_amounts?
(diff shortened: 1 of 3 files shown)


app/services/invoices/void_service.rb
```diff
@@ -18,6 +18,7 @@ def initialize(invoice:, params: {})
 
     def call
       return result.not_found_failure!(resource: "invoice") unless invoice
+      return result.forbidden_failure! if generate_credit_note && !License.premium?
```
julienbourdeau: I'd use `generate_credit_note?` because at first I thought this method was generating the credit note.
Like we have `generate_pdf(credit_note) if should_generate_pdf?` (which is now on 2 lines, but you get the point)

julienbourdeau (approved):

app/services/invoices/void_service.rb
brunomiguelpinto: Sounds good I will change it

PR #3852, opened 2025-06-19
misc(sidekiq): Add configs for usage backfill

ivannovosad:
## Description

Add new queue `usages_backfill ` for `DailyUsages::FillHistoryJob` and a dedicated worker for it.

### DIFF
--- a/app/services/daily_usages/fill_history_service.rb
+++ b/app/services/daily_usages/fill_history_service.rb
@@ -83,7 +83,13 @@ module DailyUsages
           .first
 
         if invoice.present?
-          DailyUsages::FillFromInvoiceJob.perform_later(invoice:, subscriptions: [subscription])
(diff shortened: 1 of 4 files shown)


app/services/daily_usages/fill_history_service.rb
```diff
@@ -83,7 +83,13 @@ def call
           .first
 
         if invoice.present?
-          DailyUsages::FillFromInvoiceJob.perform_later(invoice:, subscriptions: [subscription])
+          queue = if ActiveModel::Type::Boolean.new.cast(ENV["SIDEKIQ_USAGES_BACKFILL"])
+            :usages_backfill
+          else
+            :long_running
```
ivannovosad: @vincent-pochet I think here the default should be `low_priority` instead of `long_running`.

jdenquin (approved):

PR #3853, opened 2025-06-19
feat(pricing-group-keys): Migrate charges/filters grouped_by to pricing_group_keys

vincent-pochet:
## Context

This PR follows the new pricing_group_keys feature that is a renaming of `grouped_by` and an extension to all the charge models.

The feature was built to allow an easy migration from one key to an other by keep accepting the `grouped_by` attribute in `charges` and `charge_filters` when no `pricing_group_key` value exists. 

## Description

The current PR adds two migrations to convert all `grouped_by` keys in `charges#properties` and `charge_filters#properties` field in `pricing_group_keys`

A next pull request will then clean the code to only rely on the new key


db/migrate/20250619143820_migrate_charge_grouped_by.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class MigrateChargeGroupedBy < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def up
+    safety_assured do
+      execute <<-SQL
+        UPDATE charges
```
groyoh, line 9: I was wondering something. If clients have not updated yet, they will send `group_by`, right ? In this case, we do not set/update the `pricing_group_keys`. https://github.com/getlago/lago-api/blob/25771435c50c7d439cd268bf0d329ffbd8367cc1/app/services/charges/filter_charge_model_properties_service.rb#L40-L41

So if after this migration, clients who still send only `group_by` will have outdated `pricing_group_keys`.

I was thinking maybe we need some live backfilling:

```ruby
 grouped_by = sliced_attributes[:grouped_by]
 pricing_group_keys = sliced_attributes[:pricing_group_keys]

 grouped_by.reject!(&:empty?) if grouped_by.present? 
 pricing_group_keys = grouped_by if grouped_by.present? && pricing_group_keys.blank?
 pricing_group_keys.reject!(&:empty?) if pricing_group_keys.present? 
```

vincent-pochet: That's indeed a good point :+1: 

It was an old branch waiting for some love and I forgot about it. I will update it.

spec/services/charges/filter_charge_model_properties_service_spec.rb
```diff
@@ -40,13 +40,16 @@
     context "with standard charge_model" do
       let(:charge_model) { "standard" }
 
-      it { expect(filter_service.call.properties.keys).to include("amount", "grouped_by") }
+      it { expect(filter_service.call.properties.keys).to include("amount", "pricing_group_keys") }
```
groyoh: Maybe check that it is set with the right value ?

PR #3854, opened 2025-06-20
feat(pricing-units): Add PricingUnitQuery with filter by name or code

floganz:
## Context

Adding ability to search pricing units by name or code for better UX, we'll implement search in the dropdown.

## Description

Add PricingUnitQuery
Adjust existing tests

### DIFF
--- a/app/queries/pricing_units_query.rb
+++ b/app/queries/pricing_units_query.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+class PricingUnitsQuery < BaseQuery
+  Result = BaseResult[:pricing_units]
+
+  def call
+    pricing_units = base_scope.result
+    pricing_units = paginate(pricing_units)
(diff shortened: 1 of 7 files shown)


app/queries/pricing_units_query.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class PricingUnitsQuery < BaseQuery
+  Result = BaseResult[:pricing_units]
+
+  def call
+    pricing_units = base_scope.result
+    pricing_units = paginate(pricing_units)
+    pricing_units = apply_consistent_ordering(pricing_units)
```
groyoh, line 9: I believe this changes the order from `asc` to `desc`. Is that okay ?

groyoh (approved):

app/queries/pricing_units_query.rb
floganz: Yes, it's expected change.

PR #3859, opened 2025-06-23
POC for Plan's Fixed Charges

ancorcruz:


app/controllers/api/v1/plans_controller.rb
```diff
@@ -122,6 +125,18 @@ def input_params
             },
             {tax_codes: []}
           ],
+          fixed_charges: [
+            :id,
+            :invoice_display_name,
+            :add_on_id,
```
ancorcruz, line 131: `add_on_code` instead of `id`

app/models/subscription.rb
```diff
@@ -142,6 +145,12 @@ def validate_external_id
     errors.add(:external_id, :value_already_exist)
   end
 
+  def units_override_for(chargeable)
+    subscriptions_units_overrides.find_by(
+      chargeable.is_a?(Charge) ? {charge: chargeable} : {fixed_charge: chargeable}
+    )&.units
+  end
```
ancorcruz, line 152: lets implement units override only for fixed charges... if needed for other kind of records like charges we will define a new relationship instead.... lets make it simple and easy.

jdenquin:
can we close this PR, since it's a PoC don't think we need PR, a branch is enough? @annvelents @ancorcruz

PR #3866, opened 2025-06-26
feat(pricing-units): Refactor invoice PDF

floganz:
## Context

It's refactoring changes needed for adjusting PDF for pricing units extracted into separate PR for ease of testing and reviewing. You can check next step changes in [this PR](https://github.com/getlago/lago-api/pull/3867).

Border and paddings between fees re-worked in order to better adapt to varying amounts of table rows (tr) representing one fee. With pricing units changes there will be one more optional row for conversion from pricing units to fiat currency. So 1 fee could be represented in 1, 3, 5 rows.

Another changes is expanding FeeDisplayHelper with more format method, for now they are just redirecting the all to MoneyHelper but in next step it will have a logic to display either values in fiat currency or in pricing units.

## Description

CSS refactoring
Expanding FeeDisplayHelper with more methods

### DIFF
--- a/app/controllers/admin/invoices_controller.rb
+++ b/app/controllers/admin/invoices_controller.rb
@@ -13,7 +13,7 @@ module Admin
     def show
       service = ::Invoices::GeneratePdfService.new(invoice:)
 
-      render(html: service.render_html.html_safe)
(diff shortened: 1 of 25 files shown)


annvelents (approved):
👏

PR #3871, opened 2025-06-26
misc(clickhouse): Remove more to_sql from aggregation queries

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/3858

## Description

It applies the same refactor to Unique Count and Weigtehd Sum aggregations queries that are built in different services


julienbourdeau (approved):
definitely not as readable but fascinating! 👏

PR #3872, opened 2025-06-26
misc: Add pt_BR translation

fgmacedo:
## Context

Adding Brazilian Portuguese (pt_BR) translation.

Related to https://github.com/getlago/lago-front/pull/2283

### DIFF
--- a/config/locales/pt_BR/email.yml
+++ b/config/locales/pt_BR/email.yml
@@ -1,0 +1,76 @@
+---
+pt_BR:
+  email:
+    api_key:
+      created:
+        access_warning: No entanto, se você não autorizou esta alteração, por favor redefina sua senha e exclua a API Key imediatamente para proteger sua conta.
+        change_notice: Se alguém da sua equipe iniciou isso, você pode começar a usar esta API Key para automatizar seu processo de faturamento.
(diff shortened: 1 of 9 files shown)


config/locales/pt_BR/email.yml
```diff
@@ -0,0 +1,76 @@
+---
+pt_BR:
+  email:
+    api_key:
+      created:
+        access_warning: No entanto, se você não autorizou esta alteração, por favor redefina sua senha e exclua a chave de API imediatamente para proteger sua conta.
+        change_notice: Se alguém da sua equipe iniciou isso, você pode começar a usar esta chave de API para automatizar seu processo de faturamento.
+        email_info: Você está recebendo este email porque uma nova chave de API foi criada na sua instância do Lago, e você tem privilégios de administrador.
+        greetings: Olá,
+        key_has_been_created: Uma nova chave de API foi criada com sucesso para sua organização, %{organization_name}.
```
diegocharles: Can we use API Key instead of Chave de API? 
Most of the developers in Brazil is already familiarized with this term from other services like Stripe, Efi..

fgmacedo: Hi @diegocharles , thanks for your review. I agree with you. Done.

PR #3873, opened 2025-06-26
feat(entitlement): Introduce Feature and Privilege models

julienbourdeau:
Introduce `Feature` and `Privilege` models, only the left part shown on this chart.

Note that everything (most things) are namespaced under `Entitlement::`.

drawSQL-image-export-2025-06-28 (2)

Also adds new endpoints to API to manage features.

```
                  Prefix Verb                      URI Pattern                                               Controller#Action
api_v1_feature_privilege DELETE                    /api/v1/features/:feature_code/privileges/:code(.:format) api/v1/features/privileges#destroy {code: /.*/, feature_code: /.*/}
         api_v1_features GET                       /api/v1/features(.:format)                                api/v1/features#index
                         POST                      /api/v1/features(.:format)                                api/v1/features#create
          api_v1_feature GET                       /api/v1/features/:code(.:format)                          api/v1/features#show {code: /.*/}
                         PATCH                     /api/v1/features/:code(.:format)                          api/v1/features#update {code: /.*/}
                         PUT                       /api/v1/features/:code(.:format)                          api/v1/features#update {code: /.*/}
                         DELETE                    /api/v1/features/:code(.:format)                          api/v1/features#destroy {code: /.*/}
```

### Next PRs

1. Add Feature to plan (entitlements)
1. Add webhooks `feature.*`
1. Add AuditLogs

### DIFF
--- a/app/controllers/api/v1/features_controller.rb
+++ b/app/controllers/api/v1/features_controller.rb
@@ -95,8 +95,6 @@ module Api
         end
       end
 
-
-
       private
 
       def feature_create_params
--- a/app/models/entitlement/feature.rb
+++ b/app/models/entitlement/feature.rb
@@ -12,7 +12,7 @@ module Entitlement
 
     validates :code, presence: true, length: {maximum: 255}
     validates :name, length: {maximum: 255}
-    validates :description, length: {maximum: 255}
+    validates :description, length: {maximum: 600}
 
     def self.ransackable_attributes(_auth_object = nil)
       %w[code name description]
--- a/spec/models/entitlement/feature_spec.rb
+++ b/spec/models/entitlement/feature_spec.rb
@@ -19,7 +19,7 @@ RSpec.describe Entitlement::Feature, type: :model do
       expect(subject).to validate_presence_of(:code)
       expect(subject).to validate_length_of(:code).is_at_most(255)
       expect(subject).to validate_length_of(:name).is_at_most(255)
-      expect(subject).to validate_length_of(:description).is_at_most(255)
+      expect(subject).to validate_length_of(:description).is_at_most(600)
     end
   end
 end
--- a/app/controllers/api/v1/features/privileges_controller.rb
+++ b/app/controllers/api/v1/features/privileges_controller.rb
@@ -37,4 +37,4 @@ module Api
       end
     end
   end
-end 
+end


app/controllers/api/v1/features_controller.rb
```diff
@@ -0,0 +1,137 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class FeaturesController < Api::BaseController
```
jdenquin, line 5: should we namespace it under `entitlements` like we do for models and services?

julienbourdeau: I was tempted but in the `controllers` folder, the subfolder usually correspond to a path segment. Everything in `analytics` folder is mapped to `/v1/analytics/...`

CleanShot 2025-06-30 at 11 42 04@2x

app/models/entitlement/privilege.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    default_scope -> { kept }
+
+    VALUE_TYPES = %w[integer string boolean select].freeze
+
+    belongs_to :organization
+    belongs_to :feature, class_name: "Entitlement::Feature", foreign_key: :entitlement_feature_id
+
+    validates :code, presence: true
+    validates :value_type, presence: true, inclusion: {in: VALUE_TYPES}
```
groyoh, line 17: I'm wondering, is there a reason to use this over `enum :value_type, VALUE_TYPES `?

app/serializers/v1/entitlement/feature_serializer.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+
+      private
+
+      def privileges
+        model.privileges.map do |privilege|
+          {
+            code: privilege.code,
+            name: privilege.name,
+            value_type: privilege.value_type
```
groyoh, line 23: We don't serialize `config` ?

app/models/entitlement/privilege.rb
groyoh, line 17: Should we validate `config` as well ?

app/models/entitlement/privilege.rb
```diff
@@ -0,0 +1,45 @@
...
+    self.discard_column = :deleted_at
+
+    default_scope -> { kept }
+
+    VALUE_TYPES = %w[integer string boolean select].freeze
+
+    belongs_to :organization
+    belongs_to :feature, class_name: "Entitlement::Feature", foreign_key: :entitlement_feature_id
+
+    validates :code, presence: true
```
groyoh: Should we validate `code` and `name` length ?

app/models/entitlement/feature.rb
```diff
@@ -0,0 +1,42 @@
...
+  class Feature < ApplicationRecord
+    include Discard::Model
+    self.discard_column = :deleted_at
+
+    default_scope -> { kept }
+
+    belongs_to :organization
+    has_many :privileges, class_name: "Entitlement::Privilege", foreign_key: "entitlement_feature_id", dependent: :destroy
+
+    validates :code, presence: true
```
groyoh: Should we validate `code`, `name` and `description` length ?

app/controllers/api/v1/features_controller.rb
```diff
@@ -0,0 +1,137 @@
...
+              root_name:
+            )
+          )
+        else
+          render_error_response(result)
+        end
+      end
+
+      def destroy_privilege
+        feature = current_organization.features.where(code: params[:code]).first
```
groyoh: Is there a reason you don't use `find_by` instead of `where().first` ?

```suggestion
        feature = current_organization.features.find_by(code: params[:code])
```

app/queries/features_query.rb
```diff
@@ -0,0 +1,30 @@
...
+    Entitlement::Feature.where(organization:).ransack(search_params)
+  end
+
+  def search_params
+    return if search_term.blank?
+
+    {
+      m: "or",
+      name_cont: search_term,
+      code_cont: search_term
```
groyoh:
```suggestion
      code_cont: search_term
      description_cont: search_term
```

app/controllers/api/v1/features_controller.rb
```diff
@@ -0,0 +1,137 @@
...
+  module V1
+    class FeaturesController < Api::BaseController
+      def index
+        result = FeaturesQuery.call(
+          organization: current_organization,
+          pagination: {
+            page: params[:page],
+            limit: params[:per_page] || PER_PAGE
+          }
+        )
```
groyoh, line 14: `search_term` is handled in the query but not called.

app/serializers/v1/entitlement/feature_serializer.rb
```diff
@@ -0,0 +1,29 @@
...
+          created_at: model.created_at.iso8601
+        }
+      end
+
+      private
+
+      def privileges
+        model.privileges.map do |privilege|
+          {
+            code: privilege.code,
```
groyoh, line 21: Does it make sense to include the `code` in the value as it's already in the key ?

config/routes.rb
```diff
@@ -51,6 +51,10 @@
         post :evaluate_expression, on: :collection
       end
 
+      resources :features, param: :code, code: /.*/, only: %i[index show create update destroy] do
+        delete ":privilege_code", to: "features#destroy_privilege", on: :member
```
groyoh: Did you consider `DELETE /api/v1/features/:code/privileges/:privilege_code` ?

app/services/entitlement/feature_update_service.rb
```diff
@@ -0,0 +1,53 @@
...
+    attr_reader :feature, :params
+
+    def update_feature_attributes
+      feature.name = params[:name] if params.key?(:name)
+      feature.description = params[:description] if params.key?(:description)
+    end
+
+    def update_privileges
+      params[:privileges].each do |code, privilege_params|
+        privilege = feature.privileges.all.find { it[:code] == code }
```
groyoh: Is the `.all` on purpose here ?

```suggestion
        privilege = feature.privileges.find { it[:code] == code }
```

db/seeds/03_entitlements.rb
```diff
@@ -0,0 +1,93 @@
...
+).find_or_create_by!(organization:, code: "seats")
+
+# FeatureEntitlementValue.where(privilege: seats.privileges).delete_all
+seats.privileges.delete_all
+
+_max = seats.privileges.create!(organization:, code: "max", name: "Maximum", value_type: "integer")
+_max_admins = seats.privileges.create!(organization:, code: "max_admins", name: "Max Admins", value_type: "integer")
+_root = seats.privileges.create!(organization:, code: "root", name: "Allow root user", value_type: "boolean")
+
+# FeatureEntitlement.where(organization:, feature: seats).delete_all
```
groyoh, line 43: Are those meant to be here ?

spec/models/entitlement/feature_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  describe "associations" do
+    it do
+      expect(subject).to belong_to(:organization)
+      expect(subject).to have_many(:privileges).class_name("Entitlement::Privilege").dependent(:destroy)
+    end
+  end
+
+  describe "validations" do
+    it { is_expected.to validate_presence_of(:code) }
```
groyoh: Should we maybe test uniqueness here ? I don't know if that's something we test in mode specs.

spec/models/entitlement/feature_spec.rb
groyoh, line 1: I believe default scope is not tested.

spec/models/entitlement/privilege_spec.rb
groyoh, line 1: I believe default scope is not tested.

spec/queries/features_query_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+        result2 = described_class.call(organization:, pagination: {page: 2, limit: 1})
+
+        expect(result1.features).not_to eq(result2.features)
+      end
+    end
+
+    context "with search term" do
+      let(:search_term) { "seats" }
+
+      it "filters results by search term" do
```
groyoh, line 98: Maybe test both name and code separately ?

spec/queries/features_query_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+
+    let(:pagination) { {page: nil, limit: nil} }
+    let(:search_term) { nil }
+    let(:filters) { {} }
+
+    it "returns features for the organization" do
+      result = subject
+
+      expect(result).to be_success
+      expect(result.features).to include(feature1, feature2, feature3)
```
groyoh: There's a risk of false positive here. If the result include another feature, the test would still pass.

```suggestion
      expect(result.features).to contain_exactly(feature1, feature2, feature3)
```

spec/queries/features_query_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+    end
+
+    context "with pagination parameters" do
+      let(:pagination) { {page: 1, limit: 1} }
+
+      it "returns paginated results" do
+        result = subject
+
+        expect(result).to be_success
+        expect(result.features.count).to eq(1)
```
groyoh, line 83:
```suggestion
        expect(result.features.count).to eq(1)
        expect(result.features.first.code).to eq(feature3.code)
```

app/models/entitlement/privilege.rb
julienbourdeau: No! I'm adding this as a cursor rule 😄

config/routes.rb
julienbourdeau: I did but it feels like `privileges` becomes "a proper resource", something you can index/show/create/update.
In the end, privileges cannot be retrieved or access outside their parent feature.

Originally, I wanted to remove privileges by passing null. Then, I though it was a little bit risky. I found that this solution was a good inbetween

db/seeds/03_entitlements.rb
julienbourdeau: They are uncommented in the next PR. I wrote the seeder in the POC when all the things (feature, entitlement and subscription overrides). After the last PR, it will be complete.

spec/models/entitlement/feature_spec.rb
julienbourdeau: I don't validate uniqueness at the ActiveModel level because I believe it makes an extra query. We rely on the unique index in PG.

app/controllers/api/v1/features_controller.rb
julienbourdeau: I didn't intend to support it until graphql but yeah, it's actually better to add it!

app/models/entitlement/feature.rb
julienbourdeau: other codes don't have a limit. Although, I agree we should enforce some limit to avoid bugs or people storing entire wikipedia page in the name or description :D 
Let me think

app/serializers/v1/entitlement/feature_serializer.rb
julienbourdeau: Yeah, I like that the object is "complete"

app/serializers/v1/entitlement/feature_serializer.rb
julienbourdeau: We should! config was added in the middle of this PR, my bad 🙏

spec/queries/features_query_spec.rb
julienbourdeau: I'll have a look

config/routes.rb
groyoh: > I did but it feels like privileges becomes "a proper resource"

I don't necessarily feel like it is the case. Especially if you only allow delete.

In any case, it's not a big deal. But I just found something like `DELETE /api/v1/features/seats/max` a bit confusing. I would have found it maybe less confusing with `DELETE /api/v1/privileges/seats/max`.

app/models/entitlement/privilege.rb
groyoh: Keep in mind that the `enum` version will raise if you pass an invalid value though 😉

app/models/entitlement/feature.rb
groyoh: > Although, I agree we should enforce some limit to avoid bugs or people storing entire wikipedia page in the name or description :D

Fun fact, I one stored the bible in a single MongoDB field back when it still had collection-level lock on write and broke the MongoDB server 😄

spec/models/entitlement/feature_spec.rb
groyoh: My question was whether we should test that DB constraint within the model spec. Especially the `WHERE delete_at IS NULL` constraint.

If that's not something we do in the code base, then you can ignore my comment 😉

app/models/entitlement/privilege.rb
julienbourdeau: Actually, I tried but enum is not so nice in this case. 😃 

1. there is already an enum at the DB level, I'd have to define `VALUE_TYPES` as a hash like `{integer: integer, string: :string...}`
1. The enum will define scopes and conflict because one value is `select`, so I also have disable scopes or prefix/suffix the values.
1. And more importantly, the "_feature_" and "_entitlements_" aren't really meant to be used in our codebase, we're basically just storing users configuration (cf Dive-in)

With all this, I think we don't need enum. ^^

app/queries/features_query.rb
julienbourdeau: Added, sorry I missed it in the batch

spec/support/matchers/be_soft_deletable_matcher.rb
groyoh, line 1: Nice one! 👏

app/models/entitlement/feature.rb
```diff
@@ -0,0 +1,44 @@
...
+    self.discard_column = :deleted_at
+
+    default_scope -> { kept }
+
+    belongs_to :organization
+    has_many :privileges, class_name: "Entitlement::Privilege", foreign_key: "entitlement_feature_id", dependent: :destroy
+
+    validates :code, presence: true, length: {maximum: 255}
+    validates :name, length: {maximum: 255}
+    validates :description, length: {maximum: 255}
```
vincent-pochet: I would suggest to allow a longer description here

app/services/entitlement/privilege_destroy_service.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def initialize(privilege:)
+      @privilege = privilege
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "privilege") unless privilege
+
+      ActiveRecord::Base.transaction do
```
vincent-pochet, line 15: Not a blocker, but you can skip the transaction block here

app/serializers/v1/entitlement/feature_serializer.rb
```diff
@@ -0,0 +1,29 @@
...
+            name: privilege.name,
+            value_type: privilege.value_type
+          }
+        end.index_by { it[:code] }
```
vincent-pochet, line 25: What the benefits of indexing this by code, compared to just returning a collection of privileges?
It seems that it's not a common pattern in the API

app/models/entitlement/feature.rb
julienbourdeau: I'll use 600, like in PricingUnit.
Description seems short next to code or name, but I think it's code and name that are veeery long :D

app/services/entitlement/privilege_destroy_service.rb
julienbourdeau: It will be reused in the next PR to delete the privileges values before, but I agree it's useless for now.

app/serializers/v1/entitlement/feature_serializer.rb
julienbourdeau: It's closer to the shape of the input, it feels weird to send a hash but receive an array. (This is only possible because privilege code is unique per feature.)

app/models/entitlement/feature.rb
julienbourdeau: I want to know more! 😂

spec/models/entitlement/feature_spec.rb
julienbourdeau: Will do in next PR

PR #3874, opened 2025-06-26
misc(pricing-units): Various adjustments

floganz:
## Context

Adjust implementation based on QA.

## Description

Limit `short_name` to 3 chars max
Change pricing units order to `name` ASC, `created_at` DESC
Add support for conversion rate override during subscription creation

### DIFF
--- a/spec/graphql/mutations/pricing_units/update_spec.rb
+++ b/spec/graphql/mutations/pricing_units/update_spec.rb
@@ -27,7 +27,7 @@ RSpec.describe Mutations::PricingUnits::Update, type: :graphql do
   let!(:membership) { create(:membership) }
   let(:input_params) { {id: pricing_unit.id, name:, shortName: short_name, description:} }
   let(:name) { "Updated Name" }
-  let(:short_name) { "Updated Short Name" }
+  let(:short_name) { "CR" }
   let(:description) { "Updated Description" }
 
   it_behaves_like "requires current user"
--- a/schema.json
+++ b/schema.json
@@ -5439,6 +5439,18 @@
               "deprecationReason": null
             },
             {
+              "name": "appliedPricingUnit",
+              "description": null,
+              "type": {
+                "kind": "INPUT_OBJECT",
+                "name": "AppliedPricingUnitOverrideInput",
+                "ofType": null
+              },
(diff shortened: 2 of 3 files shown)


spec/graphql/mutations/pricing_units/update_spec.rb
groyoh, line 1: I believe this does not test the new input parameter.

app/services/charges/override_service.rb
```diff
@@ -35,6 +35,19 @@ def call
           filters_result.raise_if_error!
         end
 
+        if charge.applied_pricing_unit
+          conversion_rate = params.dig(:applied_pricing_unit, :conversion_rate).presence
```
groyoh, line 39: So if it's present in the request but not on the charge, we do not take the parameter into consideration ?

groyoh (approved):

spec/graphql/mutations/pricing_units/update_spec.rb
floganz: I'm not sure that get it right. Can you please elaborate?

This is existing test for updating pricing unit. For pricing unit itself no new input parameters added but validation on model is added and existing test starts failing because invalid `short_name` value (more than 3 chars) in test.

app/services/charges/override_service.rb
floganz: As now we decided not to allow override of charge's currency which might be fiat one (like USD) or custom (pricing unit).
Given that, parameters that we receive should follow this logic and define/send `applied_pricing_unit` values only if original charge already has pricing unit associated. And skip this values if charge's currency is fiat currency.

However for API users (will be added later) silently ignoring this without an error might be not the best UX.

vincent-pochet (approved):

spec/graphql/mutations/pricing_units/update_spec.rb
groyoh: Oh I guess I'm talking about the wrong mutation. I just meant that a new field was added in the charge override input but I don't think it was tested as part of the mutation test. But not a big deal.

floganz: You're right. I'll add this test in one of next PRs related to the feature.

PR #3877, opened 2025-06-27
fix(warnings): Fix warnings caused by `order` and `find_each` combination

groyoh:
## Context

As we use some default scopes on models and/or associations using `order`, a warning is logged when we call `find_each` on them.

## Description

This PR fixes the warnings by using `unscope(:order)` before calling `find_each`. It also changes the `config.active_record.error_on_ignored_order` to `true` in the test environment to ensure we raise an error instead of logging.


vincent-pochet (approved):
:rocket:

julienbourdeau (approved):
🙌

PR #3878, opened 2025-06-27
feat(cursor): cursor rules and new matcher!

julienbourdeau:
As I'm working on Feature and Entitlement, I added a few new rules to cursor. I'm extracting them in this PR.

Notice the new matcher, and the rule so cursor use it

About RIPPER MODE, I'm personally not so convinced. The diff is weird but this is because I changed it from "agent request" to "manual" in the cursor UI.

### DIFF
--- a/.cursor/rules/dev-env.mdc
+++ b/.cursor/rules/dev-env.mdc
@@ -1,0 +1,17 @@
+---
+description: 
+globs: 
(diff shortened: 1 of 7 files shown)


groyoh (approved):
Looks fine to me. No opinion on RIPPER on my side.

PR #3687, opened 2025-05-20
chore(organization): Not null organization_id

vincent-pochet:
## Context

This PR is part of the epic to add `organization_id` on all tables of the application

## Description

The current one is removing the `optional` attribute on all `belongs_to :organization` relations. In a next step the presence will be enforced at database level with a not null constraint

The PR also updates the relations from customer and organization table to make sure it uses the organization_id column.

**NOTE**: Since this change will require all tables to be updated before the release, a migration guide will be added (See https://getlago.com/docs/guide/migration/migration-to-v1.29.0 as an example). The guide will state that customer will have to first run the [migrations:fill_organization_id rake task](https://github.com/getlago/lago-api/blob/main/lib/tasks/migrations/organization_id.rake)

PR #3879, opened 2025-06-27
misc(organization): Not null constraint on applied_usage_thresholds

vincent-pochet:
## Context
Now that we are done with https://github.com/getlago/lago-api/pull/3687 to add a presence validation on the organization_id field), let's add a proper NOT NULL constraint at the DB level!

## Description

This PR adds:
- The not null constraint on  `theapplied_usage_thresholds` table
- Add new `not_null_organization_id` rails generator to make it easier to create the migration files

A lot of new migrations are coming :)

### DIFF
--- a/db/migrate/20250627084852_not_null_organization_id_on_applied_usage_thresholds.rb
+++ b/db/migrate/20250627084852_not_null_organization_id_on_applied_usage_thresholds.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class NotNullOrganizationIdOnAppliedUsageThresholds < ActiveRecord::Migration[8.0]
+  def up
+    validate_check_constraint :applied_usage_thresholds, name: "applied_usage_thresholds_organization_id_null"
+    change_column_null :applied_usage_thresholds, :organization_id, false
+    remove_check_constraint :applied_usage_thresholds, name: "applied_usage_thresholds_organization_id_null"
+  end
+
+  def down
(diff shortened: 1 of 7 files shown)


db/migrate/20250627084852_not_null_organization_id_on_applied_usage_thresholds.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class NotNullOrganizationIdOnAppliedUsageThresholds < ActiveRecord::Migration[8.0]
+  def up
+    validate_check_constraint :applied_usage_thresholds, name: "applied_usage_thresholds_organization_id_null"
+    change_column_null :applied_usage_thresholds, :organization_id, false
```
groyoh, line 6: For my knowledge, couldn't we change the column to NOT NULL without validating the check first ? Wouldn't PG validate the constraints against the table anyway?

groyoh (approved):

db/migrate/20250627084852_not_null_organization_id_on_applied_usage_thresholds.rb
vincent-pochet: @groyoh It would indeed, but it with a lock on the full table, to prevent writing in it during this time, which is far from ideal on large tables :/

groyoh: Interesting. I couldn't find the PG doc related to this behaviour. So I tested it locally and indeed, it does not lock 🙌

PR #3880, opened 2025-06-27
misc(adjusted-fee): Simplify calculations and minor refactoring

floganz:
## Context

Refactoring and small changes extracted from the work on adding support for pricing units.

## Description

Simplifies calculations (remove special zero units case)
Minor code re-arrangements

### DIFF
--- a/app/services/fees/init_from_adjusted_charge_fee_service.rb
+++ b/app/services/fees/init_from_adjusted_charge_fee_service.rb
@@ -13,10 +13,11 @@ module Fees
     end
 
     def call
-      amount_result = compute_amount
-      return result.fail_with_error!(amount_result.error) unless amount_result.success?
+      if adjusted_fee.adjusted_units? && amount_result.failure?
+        return result.fail_with_error!(amount_result.error)
+      end
 
-      result.fee = init_adjusted_fee(amount_result)
+      result.fee = init_adjusted_fee
       result
     end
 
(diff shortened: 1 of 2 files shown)


app/services/fees/init_from_adjusted_charge_fee_service.rb
```diff
@@ -26,44 +27,23 @@ def call
...
       units = adjusted_fee.units
+      amount_details = adjusted_fee.adjusted_units? ? amount_result.amount_details : {}
+
       if adjusted_fee.adjusted_units?
         rounded_amount = amount_result.amount.round(currency.exponent)
         precise_amount_cents = amount_result.amount * currency.subunit_to_unit.to_d
         amount_cents = rounded_amount * currency.subunit_to_unit
         unit_amount_cents = amount_result.unit_amount * currency.subunit_to_unit
         precise_unit_amount = amount_result.unit_amount
-        amount_details = amount_result.amount_details
```
vincent-pochet, line 59: Just a question: what's the reason behind skipping the amount details here?

floganz: You can check it in next PR related to the AdjsutedFee [here](https://github.com/getlago/lago-api/pull/3881/files).

Basically because of upcoming additional `if` we need either to extract it outside of `if` or duplicate. I believe it's easier to check in next PR files rather than explain with words :confused:

PR #3885, opened 2025-06-27
Add fix migration

brunomiguelpinto:
## Context

this should not use unscoped since we are trying to update deleted webhooks


brunomiguelpinto:
Just mistake on my side copied from the rake task and didn't remove it

PR #3888, opened 2025-06-30
fix(clickhouse): Batch event end-point

vincent-pochet:
## Description

This PR makes sure that the batch create event endpoint is working as expected when an organization relies on the Clickhouse event store

### DIFF
--- a/app/services/events/create_batch_service.rb
+++ b/app/services/events/create_batch_service.rb
@@ -61,13 +61,15 @@ module Events
     end
 
     def post_validate_events
-      ActiveRecord::Base.transaction do
-        result.events.each(&:save!)
+      if organization.postgres_events_store?
+        ActiveRecord::Base.transaction do
+          result.events.each(&:save!)
+        end
       end
 
       result.events.each do |event|
         produce_kafka_event(event)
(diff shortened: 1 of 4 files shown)


app/services/events/create_batch_service.rb
```diff
@@ -61,13 +61,15 @@ def validate_events
     end
 
     def post_validate_events
-      ActiveRecord::Base.transaction do
-        result.events.each(&:save!)
+      unless organization.clickhouse_events_store?
```
groyoh: I always find `unless` harder to read. I wonder if we could also add a `organization.postgres_events_store?` method.

spec/services/events/create_batch_service_spec.rb
```diff
@@ -312,5 +312,39 @@
...
       end
     end
+
+    context "when clickhouse is enabled on the organization" do
+      let(:organization) { create(:organization, clickhouse_events_store: true) }
+
+      it "does not store the event in postgres" do
+        result = nil
+
+        expect { result = create_batch_service.call }.not_to change(Event, :count)
```
groyoh, line 322: Maybe check that `Events::PostProcessJob` is not enqueued ?

PR #3890, opened 2025-06-30
fix(event-aggregation): the query should group prorated units by day

annvelents:
## Context

when for uniq count we post several events per day like:
27th June - 10:00 - property - Add
27th June - 11:00 - property - Remove
27th June - 12:00 - property - Add
27th June - 13:00 - property - Remove

current logic will count it as two events with minimal length of proration 1 day, so the customer will be charged for 2 prorated days of usage.

Instead, we want to group events by day, because we prorate them through day, and for one day record only one event

another scenario:
27th June - 10:00 - property - Add
28th June - 10:00 - property - Remove
28th June - 11:00 - property - Add

if we "end" usage on 28th, the first Add event will "cover" usage on 28th. The event received on 28th will not do the double charge on 28th, and will "start" on 29th, but: as result anyway we'll charge for the usage on 27th, 28, 29.. so we can just ignore the remove event, and aggregate all this as as one event started on 27th

## Description

Updated prorated uniq_count_queries for postgresql and clickhouse
Algorithm:
1) prepare events table
2) ignore only REMOVE events that will be "cancelled" by the following Add event at the same day
3) using adjusted value logic nullify not needed Adds (if previous row has the same operation_type, current has adjusted value 0; otherwise it's 1 for Add; -1 for Remove)

this pr is a pure pain.
especially clickhouse - it can't do recursion, LAG and LOG are very heavy there and tests were killing the db... so I needed to write a completely different approach from postgres, thenI rewrote couple of approaches on both stores, finally found a bug in clickhouse proration, and here is the result

### DIFF
--- a/spec/services/events/stores/postgres_store_spec.rb
+++ b/spec/services/events/stores/postgres_store_spec.rb
@@ -358,11 +358,6 @@ RSpec.describe Events::Stores::PostgresStore, type: :service do
         # 4 => added on 3 day, never removed => 13/31
         # 5 => added on 4 day, never removed => 12/31
         event_store.aggregation_property = billable_metric.field_name
-
-
-        puts "=" * 80
-        puts "Event.count: #{Event.where(organization_id: organization.id, code:).count}"
-
         expect(event_store.prorated_unique_count.round(3)).to eq(1.871) # 16/31 + 3/31 + 14/31 + 13/31 + 12/31
       end
     end


vincent-pochet (commented):
It looks good (I trust you on the result :smile: )

I guess the same changes have to be applied to:
- Grouped query
- Clickhouse query

spec/services/events/stores/clickhouse_store_spec.rb
```diff
@@ -374,20 +374,65 @@
...
+      # 1 => added on 0 day, never removed => 16/31
+      # 2 => added on 0 day, removed on 0 day => 1/31
+      # 2 => added on 1 day, never removed => 15/31
+      # 3 => added on 2 day, never removed => 14/31
+      # 4 => added on 3 day, never removed => 13/31
+      # 5 => added on 4 day, never removed => 12/31
       expect(event_store.prorated_unique_count.round(3)).to eq(2.29)
     end
+
+    context "with multiple events at the same day" do
```
vincent-pochet, line 398: Do you mind adding the same test case for postgres?
I saw that you you updated the scenario for this which is great, but having the same tests on both events stores allows us to ensure that both implementation returns the exact same result

vincent-pochet (approved):
Good one! :rocket: :clap:

spec/services/events/stores/clickhouse_store_spec.rb
annvelents: I added similar scenario (actually even more 😁 ) in `billable_metrics/prorated_aggregation/uniq_count_service_spec`, because it uses postgres, but you're right, it's still better to have it separately on the store only 👍

PR #3896, opened 2025-07-01
feat(sso_enforcement): Enforce SSO login across my organization

mariohd:
**Feature Link**

https://getlago.canny.io/feature-requests/p/enforce-sso

**Context**

Today, there are 3 authentication methods for users:

- Standard email/password login
- Google OAuth
- Okta SSO, if enabled

However, there are two limitations:

- There is no mechanism to enforce SSO authentication. Even when SSO is enabled for an organization, users can continue logging in through other methods, such as email/password or Google. This exposes organizations to potential security vulnerabilities and compliance risks.
- Only Okta SSO appears in Settings/Authentication, which prevents full customization of allowed login methods.

**Description**

The goal is to be able to enforce SSO login across my organization (for security compliance) and to customize which login methods are allowed.

To do that, a new column `authentication_methods` has been added and populated at the organization level.

We've also added a new field on the `AuthToken` generated by JWT. This field, called `login_method`, holds the auth constant value defined in the Organization model. Each auth method saves this information with its own reference.


app/services/users_service.rb
```diff
@@ -81,18 +81,11 @@ def new_token(user)
...
-    JWT.encode(payload, ENV["SECRET_KEY_BASE"], "HS256")
+    Auth::TokenService.new(user: result.user, login_method: Organizations::AuthenticationMethods::EMAIL_PASSWORD).encode
   rescue => e
     result.service_failure!(code: "token_encoding_error", message: e.message)
   end
 
-  def payload
-    {
-      sub: result.user.id,
-      exp: Time.now.to_i + 8640 # 6 hours expiration
```
rsempe: we are changing 6 hours to 3 hours with that change here

app/services/users_service.rb
```diff
@@ -81,18 +81,11 @@ def new_token(user)
   private
 
   def generate_token
-    JWT.encode(payload, ENV["SECRET_KEY_BASE"], "HS256")
+    Auth::TokenService.new(user: result.user, login_method: Organizations::AuthenticationMethods::EMAIL_PASSWORD).encode
```
rsempe: maybe would be cool to have:
`Auth::TokenService.encode(user: result.user, login_method: Organizations::AuthenticationMethods::EMAIL_PASSWORD)`

app/services/users_service.rb
mariohd: Actually, this is roughly 2:24 hours.

```
lago-api(dev)> (Time.at(Time.now.to_i + 8640) - Time.now)/1.minute
=> 143.98989779976668
```
~144 minutes. 

I've increased to 3 hours.

app/controllers/concerns/authenticable_user.rb
```diff
@@ -10,37 +10,25 @@ module AuthenticableUser
...
-  end
-
   def renew_token
     return unless current_user
 
-    result = UsersService.new.new_token(current_user)
-    response.set_header("x-lago-token", result.token) if result.success?
+    renewed = Auth::TokenService.renew(token:)
+    response.set_header(Auth::TokenService::LAGO_TOKEN_HEADER, renewed) if renewed.present?
+  rescue
```
rsempe: why this? 🤔

mariohd: that's a good question...

but that's how it was made before, we don't care if the token can't be generated or not here
Previously it was calling `UsersService#new_token`, which calls `generate_token` with the _rescue all_ clause and returning failed result.

that failed result was being checked if  `result.success?` and that's it!

PR #3898, opened 2025-07-01
polymorphic associations for ransack - is another level of pain

annvelents:
## Roadmap Task

## Context

## Description


app/graphql/resolvers/customer_portal/subscriptions_resolver.rb
```diff
@@ -16,12 +16,13 @@ class SubscriptionsResolver < Resolvers::BaseResolver
 
       def resolve(page: nil, limit: nil, plan_code: nil, status: nil)
         result = SubscriptionsQuery.call(
-          organization: context[:customer_portal_user],
+          organization: nil,
```
julienbourdeau, line 19: What 😳

julienbourdeau (commented):
Most advanced usage of Ransack I've seen so far 😄 

100% agree with adding `payments.customer_id` but I think we should wait before for enforce the not null constraint. When we added `organization_id` everywhere, we had to make a "bridge version" and document that all data must be migrated before upgrading further.

I'm in favor of removing the last 3 migrations but let's get more opinions.
If we do so, maybe data migration can be moved to a job too.

annvelents:
awesome, will split it 👍

PR #3899, opened 2025-07-02
feat(entitlement): Attach features to plans

julienbourdeau:
This PR follows https://github.com/getlago/lago-api/pull/3873

It adds a way to attach features and privilges to a plan.

In serialized plans, you get

```rb
"entitlements" => [
  {
    "code"        => "seats",
    "name"        => "Seats",
    "description" => "Nb users",
    "privileges"  => {
      "max" => {
        "code"       => "max",
        "name"       => nil,
        "value_type" => "integer",
        "value"      => 100,     # <<-------
        "config"     => {}
      }
    }
  }
]
```

**Now, the core is only missing the subscription override.**

drawSQL-image-export-2025-06-28 (2)

```
                           Prefix Verb                      URI Pattern                                                                        Controller#Action
api_v1_plan_entitlement_privilege DELETE                    /api/v1/plans/:plan_code/entitlements/:entitlement_code/privileges/:code(.:format) api/v1/plans/entitlements/privileges#destroy {code: /.*/, plan_code: /.*/, entitlement_code: /.*/}
         api_v1_plan_entitlements GET                       /api/v1/plans/:plan_code/entitlements(.:format)                                    api/v1/plans/entitlements#index {plan_code: /.*/}
                                  POST                      /api/v1/plans/:plan_code/entitlements(.:format)                                    api/v1/plans/entitlements#create {plan_code: /.*/}
          api_v1_plan_entitlement GET                       /api/v1/plans/:plan_code/entitlements/:code(.:format)                              api/v1/plans/entitlements#show {code: /.*/, plan_code: /.*/}
                                  PATCH                     /api/v1/plans/:plan_code/entitlements/:code(.:format)                              api/v1/plans/entitlements#update {code: /.*/, plan_code: /.*/}
                                  PUT                       /api/v1/plans/:plan_code/entitlements/:code(.:format)                              api/v1/plans/entitlements#update {code: /.*/, plan_code: /.*/}
                                  DELETE                    /api/v1/plans/:plan_code/entitlements/:code(.:format)                              api/v1/plans/entitlements#destroy {code: /.*/, plan_code: /.*/}
```

### Next PR

* Dispatch `plan.updated` webhooks
* Create and dispatch `feature.*` webhooks


app/services/entitlement/entitlement_privilege_destroy_service.rb
```diff
@@ -0,0 +1,35 @@
...
+      result
+    end
+
+    private
+
+    attr_reader :entitlement, :privilege_code
+
+    def find_entitlement_value
+      # NOTE: use this since it's eager loaded? `entitlement.values.find { |v| v.privilege.code == privilege_code }`
+      entitlement.values.joins(:privilege).find_by(privilege: {code: privilege_code})
```
julienbourdeau, line 32: I know it's eager loaded in the controller but it's hard to guarantee that it will always be eagerloaded so maybe the dedicated query is better.

app/models/entitlement/entitlement.rb
```diff
@@ -0,0 +1,45 @@
...
+    default_scope -> { kept }
+
+    belongs_to :organization
+    belongs_to :feature, class_name: "Entitlement::Feature", foreign_key: :entitlement_feature_id
+    belongs_to :plan
+
+    has_many :values, class_name: "Entitlement::EntitlementValue", foreign_key: :entitlement_entitlement_id, dependent: :destroy
+
+    validates :entitlement_feature_id, presence: true
+    validates :plan_id, presence: true
```
groyoh: Isn't that done by `belongs_to` already ?

app/models/entitlement/privilege.rb
```diff
@@ -11,6 +11,7 @@ class Privilege < ApplicationRecord
 
     belongs_to :organization
     belongs_to :feature, class_name: "Entitlement::Feature", foreign_key: :entitlement_feature_id
+    has_many :values, class_name: "Entitlement::EntitlementValue", foreign_key: :entitlement_privilege_id, dependent: :destroy
```
groyoh, line 14: I believe there's no test for this.

spec/models/organization_spec.rb
```diff
@@ -12,44 +12,51 @@
...
+    expect(subject).to have_many(:system_generated_invoice_custom_sections).conditions(section_type: "system_generated")
+
+    expect(subject).to have_many(:features).class_name("Entitlement::Feature")
+    expect(subject).to have_many(:privileges).class_name("Entitlement::Privilege")
+    expect(subject).to have_many(:entitlements).class_name("Entitlement::Entitlement")
+    expect(subject).to have_many(:entitlement_values).class_name("Entitlement::EntitlementValue")
+
+    expect(subject).to have_one(:applied_dunning_campaign).conditions(applied_to_organization: true)
+
+    expect(subject).to validate_inclusion_of(:default_currency).in_array(described_class.currency_list)
```
groyoh: Maybe it would be better to split the associations and validation in two `describe` ?

.cursor/rules/rails-general.mdc
groyoh, line 1: Maybe add:

```
- When testing models, make sure to test all associations and group them in a `describe "associations"` block
- When testing models, make sure to test all validations and group them in a `describe "validations"` block

```

?

spec/serializers/v1/entitlement/plan_entitlement_serializer_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+require "rails_helper"
+
+RSpec.describe V1::Entitlement::PlanEntitlementSerializer do
+  subject { described_class.new(entitlement) }
+
+  let(:organization) { create(:organization) }
+  let(:plan) { create(:plan, organization:) }
+  let(:feature) { create(:feature, organization:, code: "seats") }
+  let(:entitlement) { create(:entitlement, organization:, feature:, plan:) }
+  let(:privilege) { create(:privilege, code: "max", value_type: "integer", feature:, organization:) }
```
groyoh: Maybe add multiple privilege with different types to test all type casting?

config/routes.rb
```diff
@@ -85,7 +85,13 @@
       resources :payment_receipts, only: %i[index show]
       resources :payment_requests, only: %i[create index]
       resources :payments, only: %i[create index show]
-      resources :plans, param: :code, code: /.*/
+      resources :plans, param: :code, code: /.*/ do
+        resources :entitlements, only: %i[index show create update destroy], param: :code, code: /.*/, controller: "plans/entitlements" do
+          scope do
```
groyoh: Why do we need this ?

app/services/entitlement/plan_entitlements_base_service.rb
```diff
@@ -0,0 +1,91 @@
...
+    def validate_and_stringify(value, privilege)
+      return value if value.nil?
+
+      if privilege.value_type == "select"
+        unless privilege.config.dig("select_options").include?(value)
+          raise ValidationFailure.new(result, messages: {"#{privilege.code}_privilege_value": ["value_not_in_select_options"]})
+        end
+      end
+
+      if value.is_a?(String) || value.is_a?(Integer) || [true, false].include?(value)
```
groyoh, line 60: Shouldn't we check that the `value` matches `value_type` ?

spec/requests/api/v1/plans/entitlements/privileges_controller_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    it "returns the updated entitlement in the response" do
+      delete_with_token organization, "/api/v1/plans/#{plan.code}/entitlements/#{feature.code}/privileges/#{privilege.code}"
+
+      expect(response).to have_http_status(:success)
+
+      json = JSON.parse(response.body, symbolize_names: true)
+      expect(json[:entitlement][:code]).to eq(feature.code)
+      expect(json[:entitlement][:privileges]).to have_key(privilege2.code.to_sym)
```
groyoh:
```suggestion
      expect(json[:entitlement][:privileges]).not_to have_key(privilege.code.to_sym)
      expect(json[:entitlement][:privileges]).to have_key(privilege2.code.to_sym)
```

db/seeds/03_entitlements.rb
```diff
@@ -54,17 +55,17 @@
...
 
 # Feature not in the plan but added to the subscription
 salesforce = Entitlement::Feature.create_with(
   name: "Salesforce Integration"
 ).find_or_create_by!(organization:, code: "salesforce")
 salesforce.privileges.delete_all
 
-# FeatureEntitlement.where(organization:, feature: salesforce).delete_all
-# FeatureEntitlement.create!(organization:, feature: salesforce, subscription_external_id: sub.external_id)
+# Entitlement::Entitlement.where(organization:, feature: salesforce).delete_all
```
groyoh, line 67: Still commented on purpose ?

db/seeds/03_entitlements.rb
```diff
@@ -33,19 +33,20 @@
...
 
-# FeatureEntitlement.where(organization:, feature: seats).delete_all
-# fe = FeatureEntitlement.create!(organization:, feature: seats, plan:)
-# FeatureEntitlementValue.create!(organization:, feature_entitlement: fe, privilege: max, value: 20) # Plan defaults
-# FeatureEntitlementValue.create!(organization:, feature_entitlement: fe, privilege: max_admins, value: 3) # Plan defaults
+fe = Entitlement::Entitlement.create!(organization:, feature: seats, plan:)
+Entitlement::EntitlementValue.create!(organization:, entitlement: fe, privilege: max, value: 20)
+Entitlement::EntitlementValue.create!(organization:, entitlement: fe, privilege: max_admins, value: 3)
+Entitlement::EntitlementValue.create!(organization:, entitlement: fe, privilege: root, value: true)
 # fe_sub = FeatureEntitlement.create!(organization:, feature: seats, subscription_external_id: sub.external_id)
```
groyoh, line 48: Still commented on purpose ?

app/controllers/api/v1/plans/entitlements_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+            )
+          else
+            render_error_response(result)
+          end
+        end
+
+        private
+
+        def create_params
+          params.fetch(:entitlements, {}).permit!
```
groyoh, line 93: Since we don't properly sanitize here, I believe sending invalid data will result in 500. Just FYI, it's not a big deal.

app/services/entitlement/plan_entitlements_create_service.rb
```diff
@@ -0,0 +1,46 @@
...
+    def delete_existing_entitlements
+      EntitlementValue.where(entitlement: plan.entitlements).discard_all!
+      plan.entitlements.discard_all!
+    end
+
+    def create_entitlements
+      return if entitlements_params.blank?
+
+      entitlements_params.each do |feature_code, privilege_values|
+        feature = organization.features.find_by!(code: feature_code)
```
groyoh: I think this will make 1 query for each iteration. It could be avoided by fetching all necessary features upfront.

app/services/entitlement/plan_entitlements_update_service.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+    end
+
+    private
+
+    def update_entitlements
+      return if entitlements_params.blank?
+
+      entitlements_params.each do |feature_code, privilege_values|
+        feature = organization.features.includes(:privileges).find { it.code == feature_code }
```
groyoh, line 26: Same feedback as for the create. I believe we'll have 1 extra query for each iteration.

db/seeds/03_entitlements.rb
julienbourdeau: yes, coming with "subscription override"

app/models/entitlement/entitlement.rb
julienbourdeau: Absolutely! 🤦‍♂️

.cursor/rules/rails-general.mdc
julienbourdeau: let's try!

config/routes.rb
julienbourdeau: this was `scope module: :entitlements do` and I didn't realize when I removed the module that I had left an empty scope 😬

app/services/entitlement/plan_entitlements_base_service.rb
julienbourdeau: I wanted to let the user send `"true"` or `"2"` and only let the serializer convert the data. I liked that you could change the value type and still send the same payload. (changing value_type is not allowed for now)

app/services/entitlement/plan_entitlements_create_service.rb
julienbourdeau: very good point, let me check

julienbourdeau: funnily enough, I did it for the update service (eagerload and use `.find { ... }` but totally missed it for the create service

app/services/entitlement/plan_entitlements_update_service.rb
julienbourdeau: Hmm here I expect `organization.features` to be loaded only once and `.find` should should loop over the list to find the matching element 🤔

groyoh: Ah could be indeed. I thought the usage of `includes` would retrigger a DB query.

app/services/entitlement/plan_entitlements_base_service.rb
groyoh: I guess the reasoning makes sense but what if someones sends `"two"` for a integer type ?

julienbourdeau: We'd store "two", the API returns 0, until you change the type to `string` or send working data.
I agree it's not ideal in this case but it's also fairly unlikely.

PR #3902, opened 2025-07-02
feat: all invoices can be voided now

brunomiguelpinto:
## Context

This PR addresses a limitation in the logic for voiding invoices that was causing confusion and blocking expected behaviour with some customers. Previously, certain invoice states were not allowed to be voided depending on specific internal validations, even if the user explicitly triggered a void action.

This inconsistency was particularly evident when using generate_credit_note: false, where the system applied legacy rules that no longer make sense from a user perspective.

## Description

We removed the conditional logic that prevented invoices from being voided under specific internal states.

The behaviour is now simplified:
➡️ If the user calls void, the system will proceed with the voiding process without rejecting it due to previous state-based restrictions.

Potential Downside

This is a silent breaking change for clients relying on the previous behaviour (i.e. expecting a validation error when trying to void certain invoices). However, the rationale is that if a user explicitly calls void, they intend for the invoice to be voided — so this new behaviour is more consistent and predictable.

### DIFF
--- a/app/serializers/v1/invoice_serializer.rb
+++ b/app/serializers/v1/invoice_serializer.rb
@@ -32,7 +32,7 @@ module V1
         self_billed: model.self_billed,
         created_at: model.created_at.iso8601,
         updated_at: model.updated_at.iso8601,
-        voided_at: model.updated_at&.iso8601
+        voided_at: model.voided_at&.iso8601
       }
 
       payload.merge!(customer) if include?(:customer)
--- a/app/serializers/v1/invoice_serializer.rb
+++ b/app/serializers/v1/invoice_serializer.rb
@@ -31,7 +31,8 @@ module V1
         version_number: model.version_number,
         self_billed: model.self_billed,
         created_at: model.created_at.iso8601,
-        updated_at: model.updated_at.iso8601
+        updated_at: model.updated_at.iso8601,
+        voided_at: model.updated_at&.iso8601
       }
 
       payload.merge!(customer) if include?(:customer)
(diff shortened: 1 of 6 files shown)


app/serializers/v1/invoice_serializer.rb
```diff
@@ -31,7 +31,8 @@ def serialize
         version_number: model.version_number,
         self_billed: model.self_billed,
         created_at: model.created_at.iso8601,
-        updated_at: model.updated_at.iso8601
+        updated_at: model.updated_at.iso8601,
+        voided_at: model.updated_at&.iso8601
```
julienbourdeau:
```suggestion
        voided_at: model.voided_at&.iso8601
```

app/services/invoices/void_service.rb
```diff
@@ -67,16 +66,6 @@ def call
 
     attr_reader :invoice, :params, :generate_credit_note, :credit_amount, :refund_amount
 
-    # If we're in the legacy flow (without a credit note), follow the standard voiding rules via the state machine.
-    # If we're in the new flow (with a credit note), force the invoice into a voided state, even if it's not voidable.
-    def mark_as_voided!
-      if generate_credit_note
```
julienbourdeau, line 73: I'm not sure where the credit_note is generated now 🤔

julienbourdeau (approved):

app/serializers/v1/invoice_serializer.rb
brunomiguelpinto: Ops yes

app/services/invoices/void_service.rb
brunomiguelpinto: Ah it's a bad naming this just checks if the param for generating the credit note was being passed

brunomiguelpinto: we create the credit note 

api/app/services/invoices/void_service.rb:42

PR #3905, opened 2025-07-03
feat(projected-usage): add projected usage data to the current usage endpoint

murparreira:
## Context

We offer a endpoint and a UI for the current usage of the customers of our users. They can see in real time the consumption of each feature for the current period. However, there is no way to calculate the projected usage. So we need to add this feature on our platform.
Here is the [spec document](https://www.notion.so/getlago/Spec-Projected-usage-for-current-periods-218ef63110d280df877bfcb0be4045c4).
Here is the [dive in document](https://www.notion.so/getlago/BE-Dive-In-Projected-usage-for-current-periods-WIP-21cef63110d280318540fe4310df8048).

## Description

There are many changes in this PR:

- Add a service for calculating projected usage data based on the fees and the boundary dates.
- Change the serializer for the API to use this new service to calculate the projected usage values added to the response.
- Change the graphql types to use this new service to calculate the projected usage values added to the response.
- Made several specs to test the behavior and modified existing ones, because there were changes to an existing endpoint.

### DIFF
--- a/app/graphql/types/customers/usage/charge.rb
+++ b/app/graphql/types/customers/usage/charge.rb
@@ -65,7 +65,19 @@ module Types
         private
 
         def usage_calculator
-          @usage_calculator ||= ::Customers::FeesUsageCalculationService.new(object)
+          @usage_calculator ||= begin
+            first_fee = object.first
+            from = first_fee.properties["from_datetime"]
+            to = first_fee.properties["to_datetime"]
+            duration = first_fee.properties["charges_duration"]
+
+            ::Customers::FeesUsageCalculationService.new(
+              fees: object,
+              from_datetime: from,
+              to_datetime: to,
+              charges_duration_in_days: duration
+            )
+          end
         end
       end
     end
--- a/app/graphql/types/customers/usage/charge.rb
+++ b/app/graphql/types/customers/usage/charge.rb
@@ -7,9 +7,11 @@ module Types
         graphql_name "ChargeUsage"
 
         field :amount_cents, GraphQL::Types::BigInt, null: false
+        field :projected_amount_cents, GraphQL::Types::BigInt, null: false
         field :events_count, Integer, null: false
         field :id, ID, null: false
         field :units, GraphQL::Types::Float, null: false
+        field :projected_units, GraphQL::Types::Float, null: false
 
         field :billable_metric, Types::BillableMetrics::Object, null: false
         field :charge, Types::Charges::Object, null: false
--- a/app/graphql/types/customers/usage/charge.rb
+++ b/app/graphql/types/customers/usage/charge.rb
@@ -23,7 +23,11 @@ module Types
         end
 
         def units
-          object.map { |f| BigDecimal(f.units) }.sum
+          usage_calculator.current_units
+        end
+
+        def projected_units
+          usage_calculator.projected_units
         end
 
         def events_count
@@ -31,7 +35,11 @@ module Types
         end
 
         def amount_cents
-          object.sum(&:amount_cents)
+          usage_calculator.current_amount_cents
+        end
+
+        def projected_amount_cents
+          usage_calculator.projected_amount_cents
         end
 
         def charge
@@ -52,6 +60,12 @@ module Types
           return [] unless object.any? { |f| f.grouped_by.present? }
 
           object.group_by(&:grouped_by).values
+        end
+
+        private
+
+        def usage_calculator
+          @usage_calculator ||= ::Customers::FeesUsageCalculationService.new(object)
         end
       end
     end
--- a/app/graphql/types/customers/usage/current.rb
+++ b/app/graphql/types/customers/usage/current.rb
@@ -19,6 +19,10 @@ module Types
 
         field :charges_usage, [Types::Customers::Usage::Charge], null: false
 
+        def projected_amount_cents
+          object.projected_fees_amount_cents
+        end
+
         def charges_usage
           object.fees.group_by(&:charge_id).values
         end
--- a/app/graphql/types/customers/usage/current.rb
+++ b/app/graphql/types/customers/usage/current.rb
@@ -20,7 +20,7 @@ module Types
         field :charges_usage, [Types::Customers::Usage::Charge], null: false
 
         def projected_amount_cents
-          object.projected_fees_amount_cents
+          object.projected_amount_cents
         end
 
         def charges_usage
--- a/app/graphql/types/customers/usage/current.rb
+++ b/app/graphql/types/customers/usage/current.rb
@@ -13,6 +13,7 @@ module Types
         field :issuing_date, GraphQL::Types::ISO8601Date, null: false
 
         field :amount_cents, GraphQL::Types::BigInt, null: false
+        field :projected_amount_cents, GraphQL::Types::BigInt, null: false
         field :taxes_amount_cents, GraphQL::Types::BigInt, null: false
         field :total_amount_cents, GraphQL::Types::BigInt, null: false
 
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -409,30 +409,6 @@ class Invoice < ApplicationRecord
     update!(status: "voided", ready_for_payment_processing: false, voided_at: Time.current)
   end
 
-  def ratio
-    return 0 if fees.empty?
-    fee = fees.first
-    from_date = fee.properties["from_datetime"].to_date
-    to_date = fee.properties["to_datetime"].to_date
-    current_date = Date.current
-
-    total_days = (to_date - from_date).to_i + 1
-
-    charges_duration = fee.properties["charges_duration"] || total_days
-
-    return 1.0 if current_date >= to_date
-    return 0.0 if current_date < from_date
-
-    days_passed = (current_date - from_date).to_i + 1
-
-    ratio = days_passed.to_f / charges_duration
-    ratio.clamp(0.0, 1.0)
-  end
-
-  def projected_fees_amount_cents
-    projected_amount_cents = ratio > 0 ? (fees_amount_cents / BigDecimal(ratio.to_s)).round.to_i : 0
-  end
-
   private
 
   def should_assign_sequential_id?
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -405,6 +405,34 @@ class Invoice < ApplicationRecord
     fees.any? && should_finalize_invoice
   end
 
+  def mark_as_voided!
+    update!(status: "voided", ready_for_payment_processing: false, voided_at: Time.current)
+  end
+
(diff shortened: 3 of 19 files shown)


app/services/customers/fees_usage_calculation_service.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Customers
+  class FeesUsageCalculationService
```
julienbourdeau: I believe this should be model, not a service.

Services should extend BaseService and have a `call` method that always return a `result`. Not all of them do, but they are considered _legacy to refactor one day_ 😄  

I think all this would fit veeery well in an ActiveModel model, not an ActiveRecord model. Not sure about the name but check out `SubscriptionUsage`, maybe this could be `SubscriptionUsageFee` or else.

spec/graphql/resolvers/customer_portal/customers/usage_resolver_spec.rb
```diff
@@ -3,7 +3,8 @@
 require "rails_helper"
 
 RSpec.describe Resolvers::CustomerPortal::Customers::UsageResolver, type: :graphql do
-  let(:now) { Time.zone.now }
+  let(:now) { DateTime.parse("2025-06-15").in_time_zone }
```
vincent-pochet:
```suggestion
  let(:now) { Time.zone.parse("2025-06-15").in_time_zone }
```

`DateTime` is more designed for historic time and can have some unexpected behavior with Timezones

app/models/invoice.rb
```diff
@@ -405,6 +405,10 @@ def should_apply_provider_tax?
     fees.any? && should_finalize_invoice
   end
 
+  def mark_as_voided!
```
vincent-pochet: Is it used somewhere in this pull request?

app/services/customers/fees_usage_calculation_service.rb
```diff
@@ -0,0 +1,59 @@
...
+      return current_units if recurring?
+      (time_ratio > 0) ? (current_units / BigDecimal(time_ratio.to_s)).round(2) : BigDecimal("0")
+    end
+
+    private
+
+    def time_ratio
+      from_date = from_datetime.to_date
+      to_date = to_datetime.to_date
+      current_date = Date.current
```
vincent-pochet: Simple question: do we want the ratio to be in customer timezone or in UTC?

app/services/customers/fees_usage_calculation_service.rb
```diff
@@ -0,0 +1,59 @@
...
+      total_days = (to_date - from_date).to_i + 1
+
+      charges_duration = charges_duration_in_days || total_days
+
+      return 1.0 if current_date >= to_date
+      return 0.0 if current_date < from_date
+
+      days_passed = (current_date - from_date).to_i + 1
+
+      ratio = days_passed.to_f / charges_duration
```
vincent-pochet:
```suggestion
      ratio = days_passed.fdiv(charges_duration)
```

spec/services/customers/fees_usage_calculation_service_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+        units: "5.0",
+        charge: instance_double(
+          Charge,
+          billable_metric: instance_double(BillableMetric, recurring?: recurring)
+        )
+      )
+    ]
+  end
+
+  let(:service) do
```
vincent-pochet: We are usually defining it with `subject(:service)`, it usually helps with knowing what is tested here

vincent-pochet (approved):
Looks good :+1: 

Few comments but no blockers

app/services/customers/fees_usage_calculation_service.rb
```diff
@@ -0,0 +1,59 @@
...
+      (time_ratio > 0) ? ((amount_cents || current_amount_cents) / BigDecimal(time_ratio.to_s)).round.to_i : 0
+    end
+
+    def current_units
+      fees.sum { |f| BigDecimal(f.units) }
+    end
+
+    def projected_units
+      current_units = fees.sum { |f| BigDecimal(f.units) }
+
```
julienbourdeau:
```suggestion
```

I think you ca reuse the `current_units` method here

app/services/customers/fees_usage_calculation_service.rb
```diff
@@ -0,0 +1,59 @@
...
+      days_passed = (current_date - from_date).to_i + 1
+
+      ratio = days_passed.to_f / charges_duration
+      ratio.clamp(0.0, 1.0)
+    end
+
+    def recurring?
+      fees.first&.charge&.billable_metric&.recurring?
+    end
```
julienbourdeau: I don't understand this. Fees are always all recurring or non is recurring? 🤔

app/services/invoices/refresh_draft_service.rb
```diff
@@ -74,10 +74,10 @@ def call
...
         invoice.fees.update_all(created_at: invoice.created_at) # rubocop:disable Rails/SkipsModelValidations
 
-        return result if tax_error?(calculate_result.error)
-
-        if invoice.should_update_hubspot_invoice?
-          Integrations::Aggregator::Invoices::Hubspot::UpdateJob.perform_later(invoice: invoice.reload)
+        unless tax_error?(calculate_result.error)
+          if invoice.should_update_hubspot_invoice?
+            Integrations::Aggregator::Invoices::Hubspot::UpdateJob.perform_later(invoice: invoice.reload)
+          end
```
julienbourdeau: Was this about the rubocop error but unrelated to this PR?
```
Rails/TransactionExitStatement: Exit statement return is not allowed. Use raise (rollback) or next (commit).
```
I'm not sure this is what we expect. I was wondering if we didn't expect the transaction to be aborted 🤔 🤷‍♂️

spec/graphql/resolvers/customer_portal/customers/usage_resolver_spec.rb
```diff
@@ -125,47 +131,56 @@
...
 
-    aggregate_failures do
-      expect(usage_response["fromDatetime"]).to eq(now.beginning_of_month.iso8601)
-      expect(usage_response["toDatetime"]).to eq(now.end_of_month.iso8601)
-      expect(usage_response["currency"]).to eq("EUR")
-      expect(usage_response["issuingDate"]).to eq(now.to_date.end_of_month.iso8601)
-      expect(usage_response["amountCents"]).to eq("405")
-      expect(usage_response["totalAmountCents"]).to eq("405")
-      expect(usage_response["taxesAmountCents"]).to eq("0")
+      aggregate_failures do
```
julienbourdeau, line 146: when editing tests, don't hesitate to remove the `aggregate_failures do ... end`. We have enabled it globally since.
It's fine not to do it, just sayin' 😛

app/models/invoice.rb
murparreira: Oh, I forgot to delete this. It was changed on the main branch, and in the merge it got back.

app/services/invoices/refresh_draft_service.rb
murparreira: Yeah, I fixed it because of the rubocop error.

app/services/customers/fees_usage_calculation_service.rb
murparreira: That's a great suggestion, I was in doubt where to put this class, and this make sense.

app/services/invoices/refresh_draft_service.rb
julienbourdeau: Maybe we should fix it in a separate PR so we can check what's the intended behavior (commit or rollback).

murparreira: Okay, the only problem is that the CI fails because of this linting error.
Do you think it is ok to just add a pass to this one?

app/services/customers/fees_usage_calculation_service.rb
murparreira: I think UTC is better for the results consistency. A client in New York should see the same usage as a client in Paris. I'll change it to use utc.

app/services/customers/fees_usage_calculation_service.rb
murparreira: Yeah, this is quite confusing because the fees are related to a specific charge in this scenario, and the charge can be recurring or not, so all fees from a charge have the same type.

app/services/fees/projection_service.rb
```diff
@@ -0,0 +1,173 @@
...
+      result.projected_units = charge_model_result.projected_units
+      result.projected_pricing_unit_amount_cents = calculate_projected_pricing_unit_amount_cents(charge_model_result)
+      result
+    end
+
+    private
+
+    attr_reader :fees
+
+    def first_fee
```
jdenquin: can we use it in the initialize method? so you can use the attr_reader, same for all the values that are fixed from the first_fee since it can't change after the service initialization

app/serializers/v1/customers/charge_usage_serializer.rb
```diff
@@ -32,44 +20,118 @@ def serialize
...
+
+        current_units = fees.sum { |f| BigDecimal(f.units) }
+        current_amount_cents = fees.sum(&:amount_cents)
+        events_count = fees.sum { |f| f.events_count.to_i }
+
+        projected_units = BigDecimal("0")
+        projected_amount_cents = 0
+        projected_pricing_unit_amount_cents = 0
+
+        unless is_past_usage
```
jdenquin: can we slice this method with some private sub methods? it hurts my brain 😂

app/graphql/types/customers/usage/current.rb
```diff
@@ -21,6 +22,29 @@ class Current < Types::BaseObject
...
         end
+
+        def projected_amount_cents # rubocop:disable GraphQL/ResolverMethodLength
+          fee_groups = object.fees.group_by(&:charge_id).values
+
+          fee_groups.sum do |fee_group|
+            charge = fee_group.first.charge
+            charge_has_grouping = fee_group.any? { |f| f.grouped_by.present? }
+
+            if charge.filters.any?
```
jdenquin: too many if/elsif/else, can you rework a bit this method?

app/graphql/types/customers/usage/charge.rb
```diff
@@ -58,6 +65,50 @@ def grouped_usage
...
+            grouped_fees.sum do |group_fee_list|
+              ::Fees::ProjectionService.call(fees: group_fee_list).raise_if_error!.projected_units
+            end
+          else
+            projection_result.projected_units
+          end
+        end
+
+        def projected_amount_cents # rubocop:disable GraphQL/ResolverMethodLength
+          if charge.filters.any?
```
jdenquin: same here

PR #3907, opened 2025-07-03
fix(wallet): Retry balance refresh on tax errors

vincent-pochet:
## Description

This PR adds auto-retry logic for the `Wallets::RefreshOngoingBalanceJob` for two errors happening with anrock integration:
- `Net::ReadTimeout`
- `Integrations::Aggregator::BadGatewayError`

The new `Integrations::Aggregator::BadGatewayError` directly inherit from `LagoHttpClient::HttpError` to make sure we are not breaking logic already in place around this error type (https://github.com/getlago/lago-api/pull/3234 for example)


app/services/integrations/aggregator/base_service.rb
```diff
@@ -170,7 +170,7 @@ def parse_response(response)
       rescue JSON::ParserError
         if response.body.include?(BAD_REQUEST_ERROR)
           # NOTE: Sometimes, Anrock is responding with an HTTP 200 with a payload containing a 502 error...
```
groyoh, line 172: Oh I hate it when service providers do that 🙃

app/jobs/wallets/refresh_ongoing_balance_job.rb
```diff
@@ -8,6 +8,8 @@ class RefreshOngoingBalanceJob < ApplicationJob
 
     retry_on ActiveRecord::StaleObjectError, wait: :polynomially_longer, attempts: 6
     retry_on BaseService::TooManyProviderRequestsFailure, wait: :polynomially_longer, attempts: 25
+    retry_on Net::ReadTimeout, wait: :polynomially_longer, attempts: 6
```
groyoh: For my knowledge, how do you decide the number of `attempts` ?

app/jobs/wallets/refresh_ongoing_balance_job.rb
```diff
@@ -8,6 +8,8 @@ class RefreshOngoingBalanceJob < ApplicationJob
...
+    retry_on Integrations::Aggregator::BadGatewayError, wait: :polynomially_longer, attempts: 6
```
groyoh: Should we maybe group them since they are similar HTTP-related errors ?

```suggestion
    retry_on Net::ReadTimeout, Integrations::Aggregator::BadGatewayError, wait: :polynomially_longer, attempts: 6
```

app/services/integrations/aggregator/base_service.rb
vincent-pochet: Yes... Anrock is a mess :smile:

app/jobs/wallets/refresh_ongoing_balance_job.rb
vincent-pochet: @groyoh I won't say it's random, but we have no "strong rules" around this.

Here since it's HTTP related, with `polynomially_longer`, the first 3 retry will be quite close to each other but the next ones will wait more, if we still have bad gateway or timeout after 5 minutes it's probalby best to raise as it means the service is in some trouble :smile:

groyoh: Thanks!

PR #3913, opened 2025-07-03
misc: Renew token only when expired

rsempe:
## Description

We currently renew the authorisation token for each request made by users.

This is pointless in most cases because the token expires after 3 hours (2h24 to be exact).

The aim of this PR is to renew the token only when it expires.

### DIFF
--- a/app/services/users_service.rb
+++ b/app/services/users_service.rb
@@ -89,7 +89,7 @@ class UsersService < BaseService
   def payload
     {
       sub: result.user.id,
-      exp: Time.now.to_i + 8640 # 2 hours and 24 minutes expiration
+      exp: 3.hours.from_now.to_i
     }
   end
 
--- a/app/controllers/concerns/authenticable_user.rb
+++ b/app/controllers/concerns/authenticable_user.rb
@@ -4,7 +4,7 @@ module AuthenticableUser
   extend ActiveSupport::Concern
 
   included do
-    before_action :renew_token, if: :expired_token?
+    before_action :renew_token, if: :token_near_expiration?
   end
 
   private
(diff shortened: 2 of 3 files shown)


mariohd:
TL;DR;
This will not renew the token when expired and will logout the user. 

_Let's have an example scenario_

**Consider**:
The token is expired. 
The user make the request. 

**Execution**:
`before_action` runs and token is expired, it tries to renew the token.
 To renew, it checks if `current_user` exists and to do that, it calls `decoded_token(verify_exp: true)`. 

`decoded_token(verify_exp: true)` will raise `JWT::ExpiredSignature` and front will logout the user.


```
lago-api(dev)> token = "eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOjEsImV4cCI6MTc1MTU2NTI4OX0.cYSMYKB2m6sFF-ed1StpiYKfcBhZJJ9dUYVSrIR6gt0"
=> "eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOjEsImV4cCI6MTc1MTU2NTI4OX0.cYSMYKB2m6sFF-ed1StpiYKfcBhZJJ9dUYVSrIR6gt0"
lago-api(dev)> JWT.decode(token, ENV["SECRET_KEY_BASE"], false, {algorithm: "HS256"})
=> [{"sub" => 1, "exp" => 1751565289}, {"alg" => "HS256"}]
lago-api(dev)> JWT.decode(token, ENV["SECRET_KEY_BASE"], true, {algorithm: "HS256"})
(lago-api):5:in '<main>': Signature has expired (JWT::ExpiredSignature)
```

If we want to renew the token correctly, we have decode with the attribute `exp_leeway: <time_in_int>`.

```
lago-api(dev)> JWT.decode(token, ENV["SECRET_KEY_BASE"], true, {algorithm: "HS256", exp_leeway: 30.minutes.to_i})
=> [{"sub" => 1, "exp" => 1751565289}, {"alg" => "HS256"}]
``` 
This will renew the token 30 min before it expires.

app/services/users_service.rb
```diff
@@ -89,7 +89,7 @@ def generate_token
   def payload
     {
       sub: result.user.id,
-      exp: Time.now.to_i + 8640 # 6 hours expiration
+      exp: Time.now.to_i + 8640 # 2 hours and 24 minutes expiration
```
vincent-pochet: :rofl:

vincent-pochet (approved):

mariohd:
Oh you right! 

But in that case, we are always decoding with `false` because the result will be stored at the instance variable and we never decode again with verifications.

About exp_leeway, it's an extra time we allow. 

For example, token expires in 2h24m and we have a leeway of 36min. We will consider the token valid for 3h. If the token was expired more than 3h, it would raise the exception and logout the user. 

The goal is renew the token after it passed those 2h24 min. I would suggest that we decode the token with true and leeway and then check if the exp is bigger than the current time.

If it is, token is still valid with exp only.
If not, token needs the leeway to be valid, i.e renew it!

mariohd (approved):

PR #3917, opened 2025-07-04
fix(specs): Run specs using `parallel_tests` in CI if PR is from a community member

groyoh:
## Context

Currently, if a community PR is opened, the`KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC` secret will not be set. This will cause the specs to fail with an error as we are not able to run Knapsack Pro in this case.

This is not an ideal situation as it will complexify the contribution process.

## Description

The specs will now run using `parallel_tests` (with the `parallel_rspec` CLI) in the CI workflow as a fallback if the `KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC` secret is not set.

Note that we don't use `parallel_rspec` to actually run the tests in parallel. By combining `--only-group` with `-n`, we tell `parallel_rspec` to:

1. Split the tests into `KNAPSACK_PRO_CI_NODE_INDEX` (8) groups
2. Run only the `KNAPSACK_PRO_CI_NODE_INDEX` group

So we have the same behavior as Knapsack but without a proper distribution based on timing.

### DIFF
--- a/.github/workflows/spec.yml
+++ b/.github/workflows/spec.yml
@@ -52,7 +52,7 @@ jobs:
       LAGO_CLICKHOUSE_USERNAME: ""
       LAGO_CLICKHOUSE_PASSWORD: "password"
       LAGO_KAFKA_EVENTS_CHARGED_IN_ADVANCE_TOPIC: events_charged_in_advance
-      KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC: ${{ secrets.KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC }}
+      KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC: ''
       KNAPSACK_PRO_CI_NODE_TOTAL: ${{ matrix.ci_node_total }}
       KNAPSACK_PRO_CI_NODE_INDEX: ${{ matrix.ci_node_index }}
       KNAPSACK_PRO_FIXED_QUEUE_SPLIT: true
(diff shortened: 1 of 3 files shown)


.github/workflows/spec.yml
```diff
@@ -82,7 +77,13 @@ jobs:
       - name: Set up Clickhouse database schema
         run: bin/rails db:migrate:clickhouse
       - name: Run tests
-        run: bundle exec rake knapsack_pro:queue:rspec
+        run: |
+          if [[ -z "$KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC" ]]; then
+            echo "::warning::This is a community PR. We'll default to running the test using 'parallel_tests'."
+            bundle exec parallel_rspec --only-group "${KNAPSACK_PRO_CI_NODE_INDEX}" -n "${KNAPSACK_PRO_CI_NODE_TOTAL}" --exclude-pattern "spec/integration/.*_integration_spec.rb"
```
julienbourdeau, line 83: `-n "${KNAPSACK_PRO_CI_NODE_TOTAL}"`
I'm thinking this should be the number of cores of the runner 🤔 
I haven't tried to GH actions but this should do it: `-n "${nproc}"`

vincent-pochet (approved):

.github/workflows/spec.yml
groyoh: What's happening here is that we don't actually use `parallel_rspec` to run tests in parallel. When you combine `--only-group` with `-n`, we tell `parallel_rspec` to:

1. Split the tests into `KNAPSACK_PRO_CI_NODE_INDEX` (8) groups
2. Run only the `KNAPSACK_PRO_CI_NODE_INDEX` group

So we have the same behavior as Knapsack but without a proper test distribution based on timing.

PR #3918, opened 2025-07-04
feat(entitlement): Send `plan.updated` webhooks

julienbourdeau:
When a feature is modified, we send a `plan.updated` webhook for all plan using this feature.

When a feature privilege is modified, we send a `plan.updated` webhook for all plan having a value for this privilege.

### DIFF
--- a/app/services/entitlement/feature_destroy_service.rb
+++ b/app/services/entitlement/feature_destroy_service.rb
@@ -12,10 +12,18 @@ module Entitlement
     def call
       return result.not_found_failure!(resource: "feature") unless feature
 
+      jobs = feature.entitlements.select(:plan_id).distinct.pluck(:plan_id).map do |plan_id|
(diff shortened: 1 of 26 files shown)


app/services/entitlement/feature_destroy_service.rb
```diff
@@ -12,11 +12,19 @@ def initialize(feature:)
     def call
       return result.not_found_failure!(resource: "feature") unless feature
 
+      jobs = feature.entitlements.select(:plan_id).distinct.pluck(:plan_id).map do |plan_id|
+        SendWebhookJob.new("plan.updated", Plan.new(id: plan_id))
```
groyoh, line 16: Question here: when deleting a plan, do we actually delete the related entitlements ? Cause here we might send `plan.updated` for deleted plans ?

julienbourdeau: We don't! but it's a good point, I'll check to destroy the entitlement and values

PR #3919, opened 2025-07-04
feat(usages): Add is_customer_tin_empty argument to resolvers

ivannovosad:
## Description

Add `is_customer_tin_empty` argument to GraphQL resolvers so that is can be passed to Data API.


ancorcruz (approved):
LGTM

PR #3920, opened 2025-07-04
Feat: Include AddOn object into FeeObject gql query

diegocharles:
## Roadmap Task

This pull request introduces two main changes to the GraphQL `Fee` type and its corresponding tests. The changes add new fields to the `Fee` type and ensure they are properly tested in the spec file.

## Context

For the voided_invoice_regenerate, the FE should store and send the add_on data and the fee_id
Screenshot 2025-07-04 at 14 09 47

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -4923,6 +4923,7 @@ type EventCollection {
 }
 
 type Fee implements InvoiceItem {
+  addOn: AddOn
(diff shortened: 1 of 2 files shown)


diegocharles:
Merged onto the feat-regenerate-invoice-from-voided branch

PR #3923, opened 2025-07-04
chore(webhooks): Add default `current_organization`

julienbourdeau:
We added organization_id to all tables. By default, you can assume that `object.organization` will work.

- [x] Remove all `object.organization` overrides
- [x] Update things like `object.something.organization` when object belongs to organization
- [x] Add `private` where missing (thanks claude code). Not too important but more consistant.


floganz (approved):

vincent-pochet (approved):
image
:heart:

ancorcruz (approved):
LGTM

PR #3930, opened 2025-07-07
misc(organization): Not null constraint on integrations_* models

vincent-pochet:
## Context

Let's continue with not null constraint on organization_id column after https://github.com/getlago/lago-api/pull/3879

## Description

This PR adds the not null constraint on new tables:
- `integration_collection_mappings`
- `integration_customers`
- `integration_mappings`

### DIFF
--- a/db/migrate/20250707090313_organization_id_check_constaint_on_integration_mappings.rb
+++ b/db/migrate/20250707090313_organization_id_check_constaint_on_integration_mappings.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class OrganizationIdCheckConstaintOnIntegrationMappings < ActiveRecord::Migration[8.0]
+  def change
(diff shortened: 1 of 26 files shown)


db/migrate/20250707090313_organization_id_check_constaint_on_integration_mappings.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class OrganizationIdCheckConstaintOnIntegrationMappings < ActiveRecord::Migration[8.0]
+  def change
+    add_check_constraint :integration_mappings,
+      "organization_id IS NOT NULL",
+      name: "integration_mappings_organization_id_null",
```
groyoh: I'm only realizing it now but this should probably have been:

```suggestion
      name: "integration_mappings_organization_id_not_null",
```

😄

PR #3932, opened 2025-07-07
chore(specs): Add missing tests for `Subscription` model

groyoh:
## Context

`Subscription` model was missing some tests.

## Description

This commit adds missing tests for the `Subscription` model and update the Cursor rules to match what we expect from the model tests.

### DIFF
--- a/.cursor/rules/rails-general.mdc
+++ b/.cursor/rules/rails-general.mdc
@@ -112,8 +112,7 @@ To create a webhook:
 - After making changes to the tests, always run the tests to ensure they pass.
 - When doing array comparison in tests, use `eq` or `match_array` instead of multiple `include`/`not_to include` assertions when the expected array is small enough to be readable
 - Use single-line `let` statements when they fit on one line without breaking Rubocop rules
-- Use `let!` only for objects that need to be created before the test runs; if not referencing the object in the test, use `let` instead and explicitly reference it in the test
-- For test setup objects that aren't referenced in tests, consider creating them directly in a `before` block instead of using `let`
+- Use `let!` only for objects that need to be created before the test runs; if not referencing the object in tests, consider creating them directly in a `before` block instead of using `let`
 
 ## Models
 
--- a/.cursor/rules/rails-general.mdc
+++ b/.cursor/rules/rails-general.mdc
@@ -3,7 +3,6 @@ description: General Rails rules
 globs: app/**/*.rb
 alwaysApply: true
 ---
-
 # General style
 
 - avoid `if/unless` modifier right before the last line.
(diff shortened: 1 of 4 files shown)


vincent-pochet (approved):

.cursor/rules/rails-general.mdc
```diff
@@ -113,6 +109,22 @@ To create a webhook:
   `expect(response).to be_not_found_error("alert")`
 - Prefer `expect(...).to have_received()` instead of `expect(...).to receive()`
 - never use `aggregate_failure` in new test. Do not edit existing tests to remove it.
+- After making any code changes, always run the relevant tests to ensure they pass: `lago exec api bundle exec rspec <spec_file_or_directory>`
+- When doing array comparison in tests, use `eq` or `match_array` instead of multiple `include`/`not_to include` assertions when the expected array is small enough to be readable
+- Use single-line `let` statements when they fit on one line without breaking Rubocop rules
+- Use `let!` only for objects that need to be created before the test runs; if not referencing the object in the test, use `let` instead and explicitly reference it in the test
```
julienbourdeau: I think we have a rubocop rule that says we shouldn't use `let!` (don't ask me why 😄 )

.cursor/rules/rails-general.mdc
```diff
@@ -113,6 +109,22 @@ To create a webhook:
   `expect(response).to be_not_found_error("alert")`
 - Prefer `expect(...).to have_received()` instead of `expect(...).to receive()`
 - never use `aggregate_failure` in new test. Do not edit existing tests to remove it.
+- After making any code changes, always run the relevant tests to ensure they pass: `lago exec api bundle exec rspec <spec_file_or_directory>`
```
julienbourdeau: I think with `dev-env.mdc`, cursor always figures that tests should be run this way. 🤔 
(by the way, the dev-evn file should be merged into this one)

julienbourdeau (approved):

.cursor/rules/rails-general.mdc
groyoh: I believe the Rubocop rule is actually what's describe here: don't use `let!` unless it is called explicitly in the specs. But you can actually use `let!` otherwise.

In my case, Cursor generated a test with a `let!` that wasn't called anywhere and Rubocop later failed. So I believe it's better to have this rule to ensure Cursor doesn't generate code that will break Rubocop rules.

.cursor/rules/rails-general.mdc
groyoh: Ah yeah, I didn't see that. To be honest, I wasn't totally sure about this change. I don't know if we want Cursor to always run test after changes 😅

I'll update `dev-env.mdc` to use `bundle exec` though as `lago exec api rspec` didn't work in my case.

groyoh: I've updated the rule 😉

.cursor/rules/rails-general.mdc
julienbourdeau: > But you can actually use let! otherwise.

TIL !

PR #3934, opened 2025-07-07
fix: prevent cascading when overridden plan is not linked to active sub

lovrocolic:
## Context

Currently cascade feature targets all overridden plans even if linked subscription is terminated

## Description

Described scenario can lead to performing unnecessary cascading and this PR fixes it


groyoh (commented):
Shouldn't we include `pending` as well ?

julienbourdeau (approved):
Maybe this could be a scope in the model

PR #3935, opened 2025-07-07
fix: Ensure not null organization_id on deleted payment_provider_customers

vincent-pochet:
## Context

This PR fixes the migration to add a not null constraint on the `payment_provider_customers` table.
The migration task was missing an `unscoped` scope, to take the deleted resources into account in the migration.

## Description

To make sure it will not cause any issue when running the migration, a new DB migration is added before the one changing the constraint.


julienbourdeau (approved):
Oh no 😟

PR #3937, opened 2025-07-07
feat(entitlement): GQL to create, update and delete features

julienbourdeau:
- [x] Introduce new Feature and Privilege type.
- [x] Mutation to update one feature (see note below)
- [x] Resolver to retrieve one feature
- [x] Resolver to retrieve paginated list of features

In GQL, we don't use a hash for privileges but an array a it's more inline with the GQL way (much better typing).

There is a lot more coming (plan entitlements and subscription overrides), so this might evolve a little bit before it's fully released.

### Important note

Notice how the rest API allows you to send partial object to update a feature while the graphql requires you to send the full object. Given this payload

```rb
{
  privileges: {
    max: {name: "Max."},
  }
}
```
* API: update or create privilege with code `max` to use name "Max."
* GQL: update or create privilege with code `max` **and delete all other privileges**.

I believe the API and GQL have 2 differents users (internal vs external) and they don't have the same requirements.

There are 2 services to update features: `*PartialUpdateService` and `*FullUpdateService`.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -8215,7 +8215,7 @@ type Query {
     Unique ID of the feature
     """
     id: ID!
-  ): FeatureObject
+  ): FeatureObject!
 
   """
   Query features of an organization
--- a/schema.json
+++ b/schema.json
@@ -42725,9 +42725,13 @@
               "name": "feature",
               "description": "Query a single feature",
               "type": {
-                "kind": "OBJECT",
-                "name": "FeatureObject",
-                "ofType": null
+                "kind": "NON_NULL",
+                "name": null,
+                "ofType": {
+                  "kind": "OBJECT",
+                  "name": "FeatureObject",
+                  "ofType": null
+                }
               },
               "isDeprecated": false,
               "deprecationReason": null,


app/graphql/types/entitlement/update_feature_input.rb
groyoh, line 1: Question for my knowledge: I see you didn't add test for the input. Is this something we usually skip on purpose ?

spec/graphql/mutations/entitlement/update_feature_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+        current_organization: organization,
+        permissions: required_permission,
+        query: mutation,
+        variables: {
+          input: {
+            id: feature.id,
+            name: "Updated Feature Name",
+            description: "Updated Feature Description",
+            privileges: [
+              {code: new_privilege_code, name: new_privilege_name}
```
groyoh: Maybe we should test what happens when we already have a privilege and update privileges ?

spec/graphql/mutations/entitlement/create_feature_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        }
+      )
+
+      result_data = result["data"]["createFeature"]
+
+      expect(result_data["code"]).to eq("test_feature_with_privileges")
+      expect(result_data["name"]).to eq("Test Feature With Privileges")
+      expect(result_data["description"]).to eq("Test Feature Description")
+      expect(result_data["privileges"].size).to eq(1)
+      expect(result_data["privileges"].sole["code"]).to eq(privilege_code)
```
groyoh, line 87: Damn, TIL about `sole` 💯

spec/graphql/mutations/entitlement/create_feature_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      expect(result_data["privileges"].sole["name"]).to eq(privilege_name)
```
groyoh, line 88: Maybe check `value_type` and `config` ?

spec/graphql/resolvers/entitlement/feature_resolver_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    feature_response = result["data"]["feature"]
+
+    expect(feature_response["id"]).to eq(feature.id)
+    expect(feature_response["code"]).to eq(feature.code)
+    expect(feature_response["name"]).to eq(feature.name)
+    expect(feature_response["description"]).to eq(feature.description)
+    expect(feature_response["privileges"].count).to eq(1)
+    expect(feature_response["privileges"].first["id"]).to eq(privilege.id)
+    expect(feature_response["privileges"].first["code"]).to eq(privilege.code)
+    expect(feature_response["privileges"].first["valueType"]).to eq(privilege.value_type)
```
groyoh, line 61: Maybe check `config` ?

app/graphql/resolvers/entitlement/feature_resolver.rb
```diff
@@ -0,0 +1,24 @@
...
+      include AuthenticableApiUser
+      include RequiredOrganization
+
+      REQUIRED_PERMISSION = "features:view"
+
+      description "Query a single feature"
+
+      argument :id, ID, required: true, description: "Unique ID of the feature"
+
+      type Types::Entitlement::FeatureObject, null: true
```
groyoh: Can it be null ?

```suggestion
      type Types::Entitlement::FeatureObject
```

spec/graphql/resolvers/entitlement/features_resolver_spec.rb
groyoh, line 1: I believe search and pagination are not properly tested here.

db/migrate/20250709082136_add_privilege_value_type_default.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPrivilegeValueTypeDefault < ActiveRecord::Migration[8.0]
+  def change
+    change_column_default :entitlement_privileges, :value_type, from: nil, to: "boolean"
```
groyoh: Didn't you use `string` in the model as default ? 🤔

app/services/entitlement/feature_base_update_service.rb
groyoh, line 1: Wouldn't it have been simpler to add a `partial: false/true` kwarg to the original service ?

db/migrate/20250709082136_add_privilege_value_type_default.rb
julienbourdeau: Oops! I hesitated for a long time between db default or model default. I forgot to remove the model `attribute` definition 😬 Thanks!

app/graphql/types/entitlement/update_feature_input.rb
julienbourdeau: No, there are no reasons not to test them 👍

db/migrate/20250709082136_add_privilege_value_type_default.rb
julienbourdeau: Ah, I didn't realize I used boolean here!

app/services/entitlement/feature_base_update_service.rb
julienbourdeau: Maybe, that's a good point 🤔 I'll have another look

app/graphql/resolvers/entitlement/feature_resolver.rb
julienbourdeau: Image

turns out this `null:` arg is mandatory, but you're right it should be false.

spec/graphql/mutations/entitlement/update_feature_spec.rb
julienbourdeau: Indeed!

app/services/entitlement/feature_base_update_service.rb
julienbourdeau: I went with `partial:` keyword in the end! thanks for the suggestion

PR #3944, opened 2025-07-08
fix(entitlements): Fix route for partial update of plan entitlements

julienbourdeau:
I realized that I implemented differently from what I planned in the Dive In.
It makes no sense to pass a feature_code in the URL since you can update multiple features.


spec/graphql/resolvers/payment_providers_resolver_spec.rb
```diff
@@ -219,7 +219,7 @@
         expect(stripe_provider.success_redirect_url).to be_a String
 
         payment_providers_response = result["data"]["paymentProviders"]["collection"]
-        expect(payment_providers_response.map(&:values)).to eq [[nil], [nil, nil], [nil], [nil]]
+        expect(payment_providers_response.map(&:values)).to contain_exactly([nil], [nil, nil], [nil], [nil])
```
julienbourdeau, line 222: Flacky test

Image

ancorcruz (approved):
looks good to me.

PR #3946, opened 2025-07-08
fix(redis-overload): reduce the job params size to not overload redis

mariohd:
This is part of [ISSUE 911](https://linear.app/getlago/issue/ISSUE-911/updating-a-plan-with-a-large-number-of-child-plan-is-not-working).

With the PR, me and @ivannovosad , we tried to reduce the params size that are passed to `Charges::` jobs. 

We think the params `old_parent_attrs`,`old_parent_filters_attrs`, `old_parent_applied_pricing_unit_attrs` can be initiated when needed inside the job.


lovrocolic (changes requested):
This change unfortunately does not work. We have to capture initial plan values in `plans/update service.rb` because in the jobs we need access to original plan data. With those changes, most likely `charge.attributes` will fetch charge with updated data and it is not something we want. Original data are needed so that we can compare it with child data and decide if we want to apply cascade or not (since cascade should be skipped if child data are already overridden). Let me know if we should discuss that further 😄  Also, redis overload happened when we had one job per child. Currently we have batches so probably we are more safe on that side 🤔

mariohd:
https://github.com/getlago/lago-api/pull/3946#pullrequestreview-2998353766

@lovrocolic Got it, makes total sense!
I've reduced the params to be only the ones necessary to children jobs, does it makes sense?
Do you know if it's missing any value there?

lovrocolic:
> [#3946 (review)](https://github.com/getlago/lago-api/pull/3946#pullrequestreview-2998353766)
> 
> @lovrocolic Got it, makes total sense! I've reduced the params to be only the ones necessary to children jobs, does it makes sense? Do you know if it's missing any value there?

I just checked whole stack and looks good. I hope that you also checked all the paths since those params are passed multiple times and we don't have scenario or integration test that would notify the error right away:) @floganz Please take a look as well since you have recently also added your attributes regarding pricing units

floganz (approved):

mariohd:
> where are we regarding this @lovrocolic @mariohd ?

@jdenquin I think this is not necessary anymore, the problem was solved in other PR and in a different place. This was just an attempt of improvement in case the fix wasn't enough. 
IMO, we can close this and maybe, if necessary, we can reuse the same idea or look for the problem again.

jdenquin:
Okay I close it! Can I remove the branch?

mariohd:
> Okay I close it! Can I remove the branch?

Yes!!!

PR #3952, opened 2025-07-09
feat (wallet-limitations): add wallet_targets table and model

lovrocolic:
## Context

With billable metric limitation, it would be possible to scope wallet usage only to selected billable metrics.

## Description

This PR adds new table where BM limitations would be stored


app/models/wallet_target.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class WalletTarget < ApplicationRecord
+  include PaperTrailTraceable
+
+  belongs_to :wallet
+  belongs_to :billable_metric, optional: true
```
julienbourdeau: I'm a little surprised `billable_metric` can be optional, I don't understand what the entry would mean 🤔 

By the way, in the migration the billable_metric is `null: false`.

EDIT: Is it because there are more relationship possibly coming? In this case, why not use polymorphism?

spec/factories/wallet_targets.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :wallet_billable_metric, class: "WalletTarget" do
+    wallet
+    billable_metric
+    organization
```
julienbourdeau:
```suggestion
    organization { billable_metric&.organization || wallet&.organization || association(:organization) }
```

Now that there are organization_id everywhere, we try to reuse the organization of other models so they all belong to the same orgs

spec/models/wallet_target_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+RSpec.describe WalletTarget, type: :model do
+  subject(:wallet_target) { build(:wallet_billable_metric) }
+
+  it { is_expected.to belong_to(:organization) }
```
julienbourdeau, line 6: why not tests the other relations?

lovrocolic: 👍 Added

spec/factories/wallet_targets.rb
lovrocolic: Updated. Thanks for the notice

app/models/wallet_target.rb
lovrocolic: `optional: true` was mistake, just removed it. Thanks for the catch. I used `CouponTarget` as a reference and haven't removed it:)

The goal here was to create just a join table. There is possibility that we would extend it in the future, but AFAIK it wouldn't happen that soon. If we will have more types, for sure, the option is polymorphism or STI but right now since scope is unclear it seemed like a overkill. I just used `wallet_targets` to follow naming from coupons...

spec/factories/wallet_targets.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :wallet_billable_metric, class: "WalletTarget" do
```
julienbourdeau:
```suggestion
  factory :wallet_target do
```

Since we don't expect any new type "target" soon, I'd go with the default factory name

PR #3955, opened 2025-07-09
misc(seeds): Update seeds to include tax, add-on and pay-in-advance plan

groyoh:
## Context

Current seeds do not include tax, add-on and pay-in-advance plan which makes it cumbersome to test certain cases with freshly seeded DBs.


ancorcruz (approved):

groyoh:
Thanks @ancorcruz 🙏

PR #3957, opened 2025-07-09
misc: allow to remove a logo when passing nil

ansmonjol:
## Context

We want to allow users to remove a billing entity logo.

## Description

This PR removed the billing entity (blob) if the logo param passed is `nil`

### DIFF
--- a/app/services/billing_entities/update_service.rb
+++ b/app/services/billing_entities/update_service.rb
@@ -103,7 +103,10 @@ module BillingEntities
     end
 
     def handle_base64_logo
-      return if params[:logo].blank?
+      if params[:logo].blank?
+        billing_entity.logo&.purge
+        return
+      end
 
       base64_data = params[:logo].split(",")
       data = base64_data.second
--- a/spec/services/billing_entities/update_service_spec.rb
+++ b/spec/services/billing_entities/update_service_spec.rb
@@ -174,6 +174,21 @@ RSpec.describe BillingEntities::UpdateService do
       end
     end
 
+    context "when logo is set but then removed" do
+      let(:logo) do
+        logo_file = File.read(Rails.root.join("spec/factories/images/logo.png"))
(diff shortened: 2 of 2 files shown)


app/services/billing_entities/update_service.rb
```diff
@@ -16,7 +16,11 @@ def initialize(billing_entity:, params:)
       record: -> { billing_entity }
     )
 
-    def call
+    def call(**kwargs)
+      # Allow overriding params when explicit params are passed to an already instantiated service
+      # (e.g., update_service.call(params: {logo: nil})). We avoid shadowing the `params` attr_reader
+      # by capturing the keyword args hash instead of defining a `params` local variable.
+      @params = kwargs[:params] if kwargs.key?(:params)
```
mariohd: you don't need that, you are already receiving the `params` at the `initialize`.

app/services/billing_entities/update_service.rb
```diff
@@ -103,7 +107,10 @@ def assign_premium_attributes
     end
 
     def handle_base64_logo
-      return if params[:logo].blank?
+      if params[:logo].blank?
+        billing_entity.logo.purge
+        return
+      end
```
mariohd: only this and the tests should be good for your PR.

mariohd (changes requested):
some changes...

app/services/billing_entities/update_service.rb
mariohd: actually... if there's no logo, you can't `purge` `nil`. So adding `&` as safe check when `logo` is `nil`.
```suggestion
      if params[:logo].blank?
        billing_entity.logo.purge
        return
      end
      
      billing_entity.logo&.purge

```

ancorcruz (approved):

mariohd (approved):

PR #3958, opened 2025-07-10
feat(entitlements): Require Premium license for Entitlements

julienbourdeau:
Add check to require `License.premium?` in all entitlements services and endpoint.

A few interesting things, where I'd like your opinion.

### GraphQL shared context for spec

We duplicate a looooot of things in rspec tests. I believe we can simplify them a lot by adding a shared context. You don't have to use it, but it's nice.

Notice also the new requirements:
* `it_behaves_like "requires Premium license"` for graphql
* `it_behaves_like "a premium service"` for services

### Feature-gating routes

How do you feel about this:

```
      if License.premium?
        resources :features, param: :code, only: %i[index ...]
      end
```


spec/graphql/mutations/entitlement/destroy_feature_spec.rb
```diff
@@ -3,15 +3,13 @@
 require "rails_helper"
 
 RSpec.describe Mutations::Entitlement::DestroyFeature, type: :graphql do
-  let(:required_permission) { "features:delete" }
-  let(:membership) { create(:membership) }
-  let(:organization) { membership.organization }
+  include_context "with graphql query context"
```
julienbourdeau: I considered adding this automatically when `type: :graphql` is present, via spec_helper.rb but I thought it was not explicit enough.

```rb
RSpec.configure do |config|
  # ...
  config.include_context "with graphql query context", type: :graphql
  # ...
end
```

spec/support/shared_context/graphql_context.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+RSpec.shared_context "with graphql query context" do
+  subject(:execute) do
```
vincent-pochet: Maybe not 100% convinced on this one. It hides the "core" logic of the specs files. `result = subject` is a bit cryptic :smile: 

I'm wondering if a simple helper could help here

app/graphql/mutations/entitlement/create_feature.rb
```diff
@@ -15,6 +15,8 @@ class CreateFeature < BaseMutation
       type Types::Entitlement::FeatureObject
 
       def resolve(**args)
+        raise unauthorized_error unless License.premium?
```
groyoh: Initially I wonder if this could be a hook but I couldn't find anything in the doc about GraphQL hooks. But I guess this could be a prepended module:

```ruby
class BaseMutation
  module PremiumOnly
    def resolve(**args)
      raise unauthorized_error unless License.premium?
      super(**args)
    end
  end

  def self.premium_only!
    prepend PremiumOnly
  end
end

module Mutations
  module Entitlement
    class CreateFeature < BaseMutation
      premium_only!
    end
  end
end
```

WDYT ?

spec/support/shared_context/graphql_context.rb
julienbourdeau: I didn't find a better solution. having a `subject` is super useful and in the end, you keep repeating the `execute_graphql` with the same argument all the time :( 
Notice that execute_graphql is already a helper, it doesn't use our endpoint (maybe we should change that one day).

app/graphql/mutations/entitlement/create_feature.rb
julienbourdeau: it turns out there are very little mutations/resolver that do this. It seems that we rely on services to do this in mutations (write) and allow the resolver be accessed since you cannot create stuff.

groyoh: 👍

groyoh: So are you going to update the resolvers/services to have the same behavior ?

julienbourdeau: Yes but I added checks for the resolvers 🤷‍♂️ We'll see what we can improve later.

spec/support/shared_examples/graphql_requirements.rb
```diff
@@ -20,6 +20,18 @@
   end
 end
 
+RSpec.shared_examples "requires Premium license" do
```
groyoh, line 23: Since the two shared sexamples have almost the same title, I feel like this could be very confusing. I'd suggest considering renaming them:

```suggestion
RSpec.shared_examples "a resolver which requires Premium license" do
```

groyoh (approved):

spec/support/shared_examples/graphql_requirements.rb
julienbourdeau: Yes but it breaks the "requires..." repetition. I'll try to rename the other one
CleanShot 2025-07-11 at 09 46 05@2x

PR #3960, opened 2025-07-10
feat(entitlements): Add entitlements to plan response (read-only)

julienbourdeau:
Entitlements are returned when you retrieve a plan from the API.
Notice that you cannot edit/create entitlements via the `plans/:code` endpoints, you must use `plans/:code/entitlements` for now.

### DIFF
--- a/app/controllers/api/v1/plans_controller.rb
+++ b/app/controllers/api/v1/plans_controller.rb
@@ -20,7 +20,7 @@ module Api
         result = ::Plans::UpdateService.call(plan:, params: input_params.to_h.deep_symbolize_keys)
 
         if result.success?
-          render_plan(result.plan)
+          render_plan(reload_plan(code: result.plan.code)) # Reload to eager-load relationships, like :entitlements
         else
           render_error_response(result)
         end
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/plans_controller.rb
```diff
@@ -31,19 +31,20 @@ def destroy
         result = ::Plans::PrepareDestroyService.call(plan:)
 
         if result.success?
-          render_plan(result.plan)
+          render_plan(find_plan(code: result.plan.code)) # Reload to eager-load relationships
```
groyoh: Given that the deletion is delayed, I feel like reloading here might be causing a flaky behaviors depending on whether it comes before or after the job has run..

julienbourdeau: that is a good point. because of the defaut scope, we might not find it if it was already deleted!

julienbourdeau: i renamed find_plan to `reload_plan` and added `with_discarded` scope  👌

app/controllers/api/v1/plans_controller.rb
```diff
@@ -31,19 +31,20 @@ def destroy
...
           render_error_response(result)
         end
       end
 
       def show
-        plan = current_organization.plans.parents
-          .includes(:usage_thresholds, charges: {filters: {values: :billable_metric_filter}})
-          .find_by(code: params[:code])
-        return not_found_error(resource: "plan") unless plan
+        plan = reload_plan(code: params[:code])
```
groyoh: I believe you are now including `discarded` in show.

PR #3961, opened 2025-07-10
Chore: add customer_id to payment

annvelents:
## Context

Right now payment belong to customer through payable, which sometimes complicates query when we need to find a customer, associated to a payment

## Description

added customer_id to payment
populated customer_id.
added customer to payment creation

Added a unified command to perform  data upgrades:
`rake upgrade:check_readiness`
image


db/migrate/20250710102413_populate_payments_with_customer_id.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    Payment.where(payable_type: "PaymentRequest")
+      .find_in_batches(batch_size: 1000) do |batch|
+      Payment.where(id: batch.pluck(:id))
+        .update_all("customer_id = (SELECT customer_id FROM payment_requests WHERE payment_requests.id = payments.payable_id)")
+    end
+    # rubocop:enable Rails/SkipsModelValidations
+  end
+end
```
julienbourdeau: I think this should be a job. What's one more job in this folder? 😄 
https://github.com/getlago/lago-api/tree/main/app/jobs/database_migrations

julienbourdeau (approved):
🚀

db/migrate/20250710102413_populate_payments_with_customer_id.rb
annvelents: Added it there, and also added a "universal" task inpired by @vincent-pochet:
https://github.com/getlago/lago-api/pull/3961/files#diff-41329f763f577b8b1af65e00c1b6756e05281527f0bcc73a80cb306988e54deb
(I had a discussion with one of our customers that if the command to execute required jobs was always the same, they might have added it to their deployment script)

annvelents: decided to rename it:
https://github.com/getlago/lago-api/pull/3961/files#diff-662201afda3b7005c10e71a210add51d0ad199325e8c153c89f2f251cffbd37fR10

brunomiguelpinto (approved):

db/migrate/20250710102413_populate_payments_with_customer_id.rb
julienbourdeau: awesome!

PR #3963, opened 2025-07-10
misc(true-up-fee): Improve amounts precision for true up fee

floganz:
## Context

Improves precision of amounts and minor refactoring for future support of pricing units.

## Description

Round amounts before assigning to the model
Tests adjustments

### DIFF
--- a/app/services/fees/create_true_up_service.rb
+++ b/app/services/fees/create_true_up_service.rb
@@ -2,10 +2,10 @@
 
 module Fees
   class CreateTrueUpService < BaseService
-    def initialize(fee:, amount_cents:, precise_amount_cents:)
+    def initialize(fee:, used_amount_cents:, used_precise_amount_cents:)
       @fee = fee
-      @amount_cents = amount_cents
-      @precise_amount_cents = precise_amount_cents
+      @used_amount_cents = used_amount_cents
+      @used_precise_amount_cents = used_precise_amount_cents
       @boundaries = OpenStruct.new(fee&.properties)
 
       super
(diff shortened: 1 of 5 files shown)


app/services/fees/create_true_up_service.rb
```diff
@@ -2,38 +2,44 @@
...
-        f.true_up_parent_fee = fee
-        f.unit_amount_cents = f.amount_cents
-        f.precise_unit_amount = f.unit_amount.to_f
-      end
+      return result if used_amount_cents >= prorated_min_amount_cents
+
+      amount_cents = (prorated_min_amount_cents - used_amount_cents).round
+      precise_amount_cents = prorated_min_amount_cents - used_precise_amount_cents
+      unit_amount_cents = amount_cents
+      precise_unit_amount = precise_amount_cents / charge.plan.amount.currency.subunit_to_unit.to_d
```
julienbourdeau, line 21: so the change to rounding `amount_cents` and computing `precise_unit_amount` from `precise_amount_cents`, right?

julienbourdeau (approved):
👏

app/services/fees/create_true_up_service.rb
floganz: Main change to perform this rounding `(prorated_min_amount_cents - used_amount_cents).round` because currently we ignore fractions. Like in this example
```ruby
f = Fee.new
f.amount_cents = 350.99 
f.amount_cents # => 350, no rounding applied
```

`precise_unit_amount` is just more natural to calculate from `precise_amount_cents` rather than making float out of `unit_amount_cents` integer.

PR #3969, opened 2025-07-10
fix(plans): better plan reload

julienbourdeau:
Fix not so nice implementation added here https://github.com/getlago/lago-api/pull/3960

### DIFF
--- a/app/controllers/api/v1/plans_controller.rb
+++ b/app/controllers/api/v1/plans_controller.rb
@@ -20,7 +20,14 @@ module Api
         result = ::Plans::UpdateService.call(plan:, params: input_params.to_h.deep_symbolize_keys)
 
         if result.success?
-          render_plan(reload_plan(code: result.plan.code)) # Reload to eager-load relationships, like :entitlements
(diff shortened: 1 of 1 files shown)


app/controllers/api/v1/plans_controller.rb
```diff
@@ -31,7 +38,14 @@ def destroy
         result = ::Plans::PrepareDestroyService.call(plan:)
 
         if result.success?
-          render_plan(reload_plan(code: result.plan.code)) # Reload to eager-load relationships
+          # Reload to eager-load relationships, like :entitlements
+          plan = Plan.includes(
```
groyoh: I still feel like we need to keep discarded ones here:

```suggestion
          plan = Plan.with_discarded.includes(
```

groyoh (approved):

app/controllers/api/v1/plans_controller.rb
julienbourdeau: 🙌 I tried to make a test but it wasn't so easy to simulate. I tested the query in the console.

PR #3973, opened 2025-07-11
feat(wallet-limitations): Add service updates for billable metric wallet limitation

lovrocolic:
## Context

## Description

This PR adds needed changes in create and update services for wallets. Billable metric identifiers are processed so that `codes` are used form the api side and `ids` are used from the GQL side. Each billable metric limitation per wallet results in new `WalletTarget` object

### DIFF
--- a/app/graphql/types/wallets/applies_to.rb
+++ b/app/graphql/types/wallets/applies_to.rb
@@ -5,7 +5,7 @@ module Types
     class AppliesTo < Types::BaseObject
       graphql_name "WalletAppliesTo"
 
-      field :billable_metrics, [Types::BillableMetrics::Object], null: true
+      field :billable_metrics, [Types::BillableMetrics::Object]
       field :fee_types, [Types::Fees::TypesEnum], null: true, method: :allowed_fee_types
     end
   end
--- a/app/services/wallets/create_service.rb
+++ b/app/services/wallets/create_service.rb
@@ -49,10 +49,8 @@ module Wallets
           Wallets::RecurringTransactionRules::CreateService.call(wallet:, wallet_params: params)
         end
 
-        if billable_metric_identifiers.present?
-          billable_metrics.each do |bm|
-            WalletTarget.create!(wallet:, billable_metric: bm, organization_id: wallet.organization_id)
-          end
+        billable_metrics.each do |bm|
+          WalletTarget.create!(wallet:, billable_metric: bm, organization_id: wallet.organization_id)
         end
       end
 
--- a/app/services/wallets/update_service.rb
+++ b/app/services/wallets/update_service.rb
@@ -69,7 +69,8 @@ module Wallets
     end
 
     def process_billable_metrics
-      existing_wallet_billable_metric_ids = wallet.wallet_targets.pluck(:billable_metric_id).compact
+      # In case of adding new type of limitation in wallet_targets, query from below should use compact to avoid nil values in the array
+      existing_wallet_billable_metric_ids = wallet.wallet_targets.pluck(:billable_metric_id)
 
       billable_metrics.each do |bm|
         next if existing_wallet_billable_metric_ids.include?(bm.id)


app/graphql/types/wallets/applies_to.rb
```diff
@@ -5,6 +5,7 @@ module Wallets
     class AppliesTo < Types::BaseObject
       graphql_name "WalletAppliesTo"
 
+      field :billable_metrics, [Types::BillableMetrics::Object], null: true
```
groyoh: It is nullable by default:

```suggestion
      field :billable_metrics, [Types::BillableMetrics::Object]
```

app/services/wallets/create_service.rb
```diff
@@ -45,6 +48,12 @@ def call
         if params[:recurring_transaction_rules].present?
           Wallets::RecurringTransactionRules::CreateService.call(wallet:, wallet_params: params)
         end
+
+        if billable_metric_identifiers.present?
+          billable_metrics.each do |bm|
+            WalletTarget.create!(wallet:, billable_metric: bm, organization_id: wallet.organization_id)
+          end
+        end
```
groyoh: Since you default to empty array:

```suggestion
        billable_metrics.each do |bm|
          WalletTarget.create!(wallet:, billable_metric: bm, organization_id: wallet.organization_id)
        end
```

app/services/wallets/create_service.rb
```diff
@@ -76,5 +85,26 @@ def call
...
+
+    def billable_metrics
+      return @billable_metrics if defined?(@billable_metrics)
+      return [] if billable_metric_identifiers.blank?
+
+      @billable_metrics = if api_context?
+        BillableMetric.where(code: billable_metric_identifiers, organization_id: params[:organization_id])
+      else
+        BillableMetric.where(id: billable_metric_identifiers, organization_id: params[:organization_id])
+      end
```
groyoh, line 105:
```suggestion
      identifier_key = api_context? ? :code : :id

      @billable_metrics = BillableMetric.where(organization_id: params[:organization_id], identifier_key => billable_metric_identifiers)
```

app/services/wallets/update_service.rb
```diff
@@ -61,7 +63,49 @@ def valid_expiration_at?(expiration_at:)
     end
 
     def valid_limitations?
+      result.billable_metrics = billable_metrics
+      result.billable_metric_identifiers = billable_metric_identifiers
       Wallets::ValidateLimitationsService.new(result, **params).valid?
     end
+
+    def process_billable_metrics
+      existing_wallet_billable_metric_ids = wallet.wallet_targets.pluck(:billable_metric_id).compact
```
groyoh: Why do we need compact here ?

app/graphql/types/wallets/applies_to.rb
floganz: I think that collection could not be `null`.

app/services/wallets/update_service.rb
```diff
@@ -61,7 +63,49 @@ def valid_expiration_at?(expiration_at:)
...
+        WalletTarget.create!(wallet:, billable_metric: bm, organization_id: wallet.organization_id)
+      end
+
+      sanitize_wallet_billable_metrics(existing_wallet_billable_metric_ids) if existing_wallet_billable_metric_ids.present?
+    end
+
+    def sanitize_wallet_billable_metrics(existing_wallet_billable_metric_ids)
+      not_needed_wallet_target_ids = existing_wallet_billable_metric_ids - billable_metrics.pluck(:id)
+      not_needed_wallet_target_ids.each do |wallet_billable_metric_id|
+        WalletTarget.find_by(wallet:, billable_metric_id: wallet_billable_metric_id, organization: wallet.organization)&.destroy!
```
groyoh, line 87: I believe wallet target should always be present here since you retrieve here before:

```suggestion
        WalletTarget.find_by(wallet:, billable_metric_id: wallet_billable_metric_id, organization: wallet.organization).destroy!
```

app/services/wallets/create_service.rb
```diff
@@ -76,5 +85,26 @@ def call
...
     end
+
+    def billable_metric_identifiers
+      return [] if params[:applies_to].blank?
+
+      key = api_context? ? :billable_metric_codes : :billable_metric_ids
+
+      return [] if params[:applies_to][key].blank?
+
+      params[:applies_to][key]&.compact&.uniq
```
floganz, line 94: It looks that this method called several times and do not have any memoization.

app/services/wallets/update_service.rb
floganz: Can we have `wallet_target` with null `billable_metric_id`? If not then `.compact` might not needed.

app/services/wallets/validate_service.rb
```diff
@@ -78,7 +78,11 @@ def valid_metadata?
     end
 
     def valid_limitations?
-      return true if Wallets::ValidateLimitationsService.new(BaseService::Result.new, **args).valid?
+      limitation_result = BaseService::Result.new
+      limitation_result.billable_metrics = result.billable_metrics
+      limitation_result.billable_metric_identifiers = result.billable_metric_identifiers
+
+      return true if Wallets::ValidateLimitationsService.new(limitation_result, **args).valid?
```
groyoh, line 85: Is there a reason why we don't use `result` directly ?

```suggestion
      return true if Wallets::ValidateLimitationsService.new(result, **args).valid?
```

floganz (approved):
All looks good!
I left few nit-picks but It's totally OK to skip them.

app/services/wallets/update_service.rb
lovrocolic: For coupons we have similar setup and join table is apparently used for 2 associations, but for now we don't need it here. I removed it and added a comment in case of adding new association in `wallet_targets`

lovrocolic: For coupons we have similar setup and join table is apparently used for 2 associations, but for now we don't need it here. I removed it and added a comment in case of adding new association in wallet_targets.

app/services/wallets/update_service.rb
lovrocolic: You are right, but since query is not performed in the method itself but the method just relies on the variable, I have a practice to leave it. But shouldn't be too expensive :)

app/services/wallets/validate_service.rb
lovrocolic: Hmm, just thinking and cannot remember but seems to me that there was a reason. I will try to figure it out in the meantime

lovrocolic: Seems that with passing new result we wrapped all limitations errors together and not individual errors are attached to the parent result, but right now not sure why was that needed 🤔

PR #3978, opened 2025-07-11
feat(wallet-limitations): add api support for billable metric limitation

lovrocolic:
## Context

## Description

This PR adds api support for billable metrics that should be targeted by wallet credits


groyoh (approved):
Same feedback as on #3979:

It would be nice to add a test on the update to know what happens if you already have billable metrics and update those billable metrics.

PR #3979, opened 2025-07-11
feat(wallet-limitations): add gql support for billable metric limitations

lovrocolic:
## Context

## Description

This PR adds graphQL support for billable metrics that should be targeted by wallet credits


groyoh (approved):
It would be nice to add a test on the update to know what happens if you already have billable metrics and update those billable metrics.

lovrocolic:
> It would be nice to add a test on the update to know what happens if you already have billable metrics and update those billable metrics.

On GQL and API side added basic services that test request and format of response on one example. There are bunch of tests that cover all edge cases in the PR that handles service logic: https://github.com/getlago/lago-api/pull/3973

PR #3987, opened 2025-07-14
Fix(daily-usages): Retry job fix

ivannovosad:
## Context

This job was sometimes executed before the actualy invoice existed.

## Description

Retry this job in case the invoice was not created yet.

### DIFF
--- a/app/jobs/daily_usages/fill_from_invoice_job.rb
+++ b/app/jobs/daily_usages/fill_from_invoice_job.rb
@@ -10,6 +10,8 @@ module DailyUsages
       end
     end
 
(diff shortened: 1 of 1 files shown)


ancorcruz:
@ivannovosad where is the job queued while the invoice is not existing yet? The two places in code we queue this already expect to have an invoice or returns... am I missing something?

ancorcruz (approved):

ancorcruz:
@ivannovosad then , it is not the invoice is not created but it is on generating status and it should be in a different status... but still not see the deserialization error because the invoice having generating status... anyway I have accepted the PR.

PR #3992, opened 2025-07-14
Feat Regenerate invoice from voided invoice 

brunomiguelpinto:
## Context

Currently, when voiding an invoice, it’s not possible to regenerate it with updated fee attributes. This limitation creates friction for users who want to correct or adjust values such as units, unit_amount_cents, or invoice_display_name after an invoice has already been issued and voided.

## Description

This PR introduces two new services that enable the regeneration of a voided invoice with adjusted fee values:

AdjustedFees::EstimateService

This service is used to preview how a fee would be adjusted before persisting any data. It handles all necessary backend calculations—such as tax application, amount breakdown, and unit price derivation—based on the provided input. It is designed to be used by the frontend to fetch real-time estimates while the user is making adjustments to fees.

Invoices::RegenerateFromVoidedService

This service is responsible for actually rebuilding the invoice from a voided one. It:
	•	Duplicates relevant fee records,
	•	Applies adjustments coming from the frontend (e.g., new units, updated unit_amount_cents, or invoice_display_name),
	•	Recalculates all fee amounts, taxes, and totals,
	•	Applies coupons, credit notes, and prepaid credits if applicable,
	•	Transitions the invoice to the appropriate payment status (pending or succeeded).


app/services/adjusted_fees/create_service.rb
```diff
@@ -4,16 +4,17 @@ module AdjustedFees
...
       @organization = invoice.organization
       @params = params
+      @preview = preview
 
       super
     end
 
     def call
-      return result.forbidden_failure! if !License.premium? || !invoice.draft?
+      return result.forbidden_failure! if not_authorized?
```
vincent-pochet: Definitely not a blocker but what what do you think of `forbidden?`

app/services/adjusted_fees/create_service.rb
```diff
@@ -4,16 +4,17 @@ module AdjustedFees
   class CreateService < BaseService
     Result = BaseResult[:fee, :adjusted_fee]
 
-    def initialize(invoice:, params:)
+    def initialize(invoice:, params:, preview: false)
```
vincent-pochet, line 9: Since we have more and more of these "previous", "draft"... flags, could you add a small comment to explain what the impact of passing this flag for future reader :)

app/services/adjusted_fees/estimate_service.rb
```diff
@@ -0,0 +1,197 @@
...
+      result.fee = estimated_fee
+      result
+    end
+
+    private
+
+    attr_reader :organization, :invoice, :params
+
+    def apply_taxes_and_assign_ids(fee)
+      Fees::ApplyTaxesService.call(fee:)
```
vincent-pochet: Should we test the result?

app/services/invoices/regenerate_from_voided_service.rb
```diff
@@ -0,0 +1,214 @@
...
+
+    activity_loggable(
+      action: "invoice.regenerated",
+      record: -> { voided_invoice }
+    )
+
+    def call
+      return result.not_found_failure!(resource: "invoice") unless voided_invoice
+
+      ActiveRecord::Base.transaction do
```
vincent-pochet, line 22: We are calling multiple sub-services, don't we want to test the result or at least to use `call!` to raise the failed results?

app/services/adjusted_fees/create_service.rb
brunomiguelpinto: sounds good 👍

app/services/adjusted_fees/estimate_service.rb
brunomiguelpinto: Yes

app/services/invoices/regenerate_from_voided_service.rb
brunomiguelpinto: Yes all of them should be called with !

app/services/invoices/create_one_off_service.rb
```diff
@@ -48,6 +49,7 @@ def call
         Invoices::ApplyInvoiceCustomSectionsService.call(invoice:)
         invoice.payment_status = invoice.total_amount_cents.positive? ? :pending : :succeeded
         Invoices::TransitionToFinalStatusService.call(invoice:)
+        invoice.voided_invoice_id = voided_invoice_id if voided_invoice_id.present?
```
lovrocolic, line 52: why i this needed? I cannot see any logic around this field?

brunomiguelpinto: its only to have a link to the previous invoice that was voided

brunomiguelpinto: It was a product specefication the new invoice needs to be linked to the previous one

app/services/adjusted_fees/estimate_service.rb
```diff
@@ -0,0 +1,197 @@
...
+      result.fee = estimated_fee
+      result
+    end
+
+    private
+
+    attr_reader :organization, :invoice, :params
+
+    def apply_taxes_and_assign_ids(fee)
+      Fees::ApplyTaxesService.call!(fee:)
```
lovrocolic, line 39: Should we also add support for Anrok and Avalara if customer is connected to tax provider?

brunomiguelpinto: Yes

lovrocolic (approved):

PR #3995, opened 2025-07-15
feat(wallet-limitations): apply wallet credits with BM limitation

lovrocolic:
## Context

Currently it is not possible to scope wallet consumption to target only specific BM fees

## Description

With this PR it is possible to apply wallet credits on specific charge fees. Also, it is possible to combine fee type limitation and billable metric limitation.

### Example
Fees: `subscription_fee`, `charge_fee1` (bm1_code), `charge_fee2` (bm2_code)

Type limitation: [charge, subscription]

BM limitation: [`bm1_code`]

=> Wallet credits will only target `subscription_fee` and `charge_fee1`

### DIFF
--- a/app/models/wallet.rb
+++ b/app/models/wallet.rb
@@ -52,6 +52,10 @@ class Wallet < ApplicationRecord
   def limited_fee_types?
     allowed_fee_types.present?
   end
+
+  def limited_billable_metrics?
+    billable_metrics.any?
+  end
 end
 
 # == Schema Information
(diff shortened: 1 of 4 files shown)


app/models/wallet.rb
```diff
@@ -52,6 +52,10 @@ def currency
   def limited_fee_types?
     allowed_fee_types.present?
   end
+
+  def limited_billable_metrics?
```
julienbourdeau: I think the name isn't ideal.
I'd use `limited_to_billable_metrics?` or even `billable_metrics_limited?`.

```suggestion
  def billable_metrics_limited?
```

lovrocolic: Let's go with `limited_to_billable_metrics?`

PR #4003, opened 2025-07-16
test(credit-note): Refactor `CreditNotes::CreateFromTermination` spec

groyoh:
## Context

As I was working on adding refund capabilities to the `CreditNotes::CreateFromTermination` service, I noticed that the test spec had several issues:

- The taxes were not properly defined in the setup and the test was not properly checking the resulting tax amount.
- The test setup and assertions were repetitive and not easy to update.
- The tests were using small amounts (less than 1 euros) which made it hard to ensure that proration were properly calculated.
- The test was using `aggregate_failures`.

## Description

I decided to refactor the test to make it more maintainable and easier to update.


groyoh:
Thanks @julienbourdeau 🙏

PR #4005, opened 2025-07-16
feat(logs): produce activity logs after commit by default

julienbourdeau:
## Context

`after_commit` parameter was added yesterday #3970.

## Description

This PR uses `after_commit: true` by default when using activity_loggable.

All expectations `with` call had to be updated to add the new params `with(..., ..., after_commit: true`). I took this opportunity to add a new custom matcher.

You have to appreciate the search and replace with regex in IntelliJ

```
have_received\(:produce\)\.with\(([a-z_\.]+),\s+("[a-z_\.]+")\)
have_produced($2).after_commit.with($1)
```

CleanShot 2025-07-16 at 13 55 55@2x

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -129,6 +129,7 @@ RSpec.configure do |config|
 
   config.before do |example|
     ActiveJob::Base.queue_adapter.enqueued_jobs.clear
+    allow(Utils::ActivityLog).to receive(:produce).and_call_original
 
     if example.metadata[:scenarios]
       stub_pdf_generation
--- a/spec/support/matchers/activity_log_matcher.rb
+++ b/spec/support/matchers/activity_log_matcher.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+RSpec::Matchers.define :have_produced do |activity_type|
+  match do |actual|
+    @actual = actual
+    @activity_type = activity_type
+
+    expected_params = [@object, activity_type]
+    expected_params << {after_commit: @after_commit} unless @after_commit.nil?
+
+    expect(actual).to have_received(:produce).with(*expected_params)
+  end
+
+  chain :with do |object|
+    @object = object
+  end
+
+  chain :after_commit do
+    @after_commit = true
+  end
+
+  chain :not_after_commit do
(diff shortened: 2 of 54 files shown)


spec/spec_helper.rb
```diff
@@ -129,6 +129,7 @@
 
   config.before do |example|
     ActiveJob::Base.queue_adapter.enqueued_jobs.clear
+    allow(Utils::ActivityLog).to receive(:produce).and_call_original
```
groyoh, line 132: Could this have an impact of the `Utils::ActivityLog` tests ? I guess not since we have `and_call_original` but I'm wondering 🤔

spec/support/matchers/activity_log_matcher.rb
```diff
@@ -0,0 +1,41 @@
...
+    @actual = actual
+    @activity_type = activity_type
+
+    expected_params = [@object, activity_type]
+    expected_params << {after_commit: @after_commit} unless @after_commit.nil?
+
+    expect(actual).to have_received(:produce).with(*expected_params)
+  end
+
+  chain :with do |object|
```
groyoh, line 14: Is `@object` always present? If yes, it should probably not be a chain as chains are used primarily for optional behaviors.

groyoh (approved):

spec/spec_helper.rb
julienbourdeau: It definitely shouldn't or misunderstood `and_call_original` all my life  😮 

Notice that this still works:
https://github.com/getlago/lago-api/blob/a0f82cc22edfeffd26bb478e8fcf6d4c2e429f3c/spec/services/base_service_spec.rb#L58

spec/support/matchers/activity_log_matcher.rb
julienbourdeau: It's probably required because I don't see how you can produce a log without an object, but I find this:

```rb
expect(Utils::ActivityLog).to have_produced("applied_coupon.deleted").with(applied_coupon)
```

so much nicer than this
```rb
expect(Utils::ActivityLog).to have_produced("applied_coupon.deleted", applied_coupon)
```

PR #4006, opened 2025-07-16
feat: Add regenerated_invoice_id to invoice_subscriptions and update terminating index

brunomiguelpinto:
## Context

To support regenerating invoices for terminating subscriptions, we need to allow multiple invoice_subscriptions records with the same subscription_id and invoicing_reason = 'subscription_terminating' — as long as the invoice has been regenerated.

Currently, the uniqueness constraint (index_unique_terminating_subscription_invoice) prevents this, blocking valid regeneration flows.

## Description

This PR introduces two migrations:
Add regenerated_invoice_id column to the invoice_subscriptions table.
Add a new unique partial index (index_unique_terminating_invoice_subscription) on subscription_id and  invoicing_reason, scoped to:

- invoicing_reason = 'subscription_terminating'
- regenerated_invoice_id IS NULL

The new index replaces the old one, but is added safely and non-disruptively


db/migrate/20250716132649_add_regenerated_invoice_id_and_index_to_invoice_subscriptions.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class AddRegeneratedInvoiceIdAndIndexToInvoiceSubscriptions < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_column :invoice_subscriptions, :regenerated_invoice_id, :uuid
```
julienbourdeau: I think this should be a foreign key 🤔 
I haven't tested the snippet but something like:

```rb
add_reference :invoice_subscriptions, :regenerated_invoice, foreign_key: {to_table: :invoices}, index: {algorithm: :concurrently}, type: :uuid
```

brunomiguelpinto: ah yes true let me change it

julienbourdeau (approved):

PR #4007, opened 2025-07-16
feat(entitlement): Add `Feature.subscriptions_count`

julienbourdeau:
Used to display the number of subscription in the frontend. Not returned in the API.
Notice that this is a very nice N+1 😬 but it's acceptable since it's only used in the Frontend, on the Freature list page (this is the same for plans list for instance).

Notice that this return only the active and pending subscription, unlike the plans list that shows also terminated subscriptions.

CleanShot 2025-07-16 at 15 55 49@2x

### The final query

```sql
SELECT
  COUNT(*)
FROM
  "subscriptions"
  INNER JOIN "plans" "plan" ON "plan"."id" = "subscriptions"."plan_id"
WHERE
  "subscriptions"."status" IN (1, 0)
  AND (
    "subscriptions"."plan_id" IN (
      SELECT
        "plans"."id"
      FROM
        "plans"
        INNER JOIN "entitlement_entitlements" ON "plans"."id" = "entitlement_entitlements"."plan_id"
      WHERE
        "plans"."deleted_at" IS NULL
        AND "entitlement_entitlements"."deleted_at" IS NULL
        AND "entitlement_entitlements"."entitlement_feature_id" = '753b7132-e26e-41da-9e83-b2bf49b4b0af'
    )
    OR "plan"."parent_id" IN (
      SELECT
        "plans"."id"
      FROM
        "plans"
        INNER JOIN "entitlement_entitlements" ON "plans"."id" = "entitlement_entitlements"."plan_id"
      WHERE
        "plans"."deleted_at" IS NULL
        AND "entitlement_entitlements"."deleted_at" IS NULL
        AND "entitlement_entitlements"."entitlement_feature_id" = '753b7132-e26e-41da-9e83-b2bf49b4b0af'
    )
  );
```

### DIFF
--- a/app/models/entitlement/feature.rb
+++ b/app/models/entitlement/feature.rb
@@ -11,6 +11,7 @@ module Entitlement
     has_many :privileges, class_name: "Entitlement::Privilege", foreign_key: "entitlement_feature_id", dependent: :destroy
     has_many :entitlements, class_name: "Entitlement::Entitlement", foreign_key: "entitlement_feature_id", dependent: :destroy
     has_many :entitlement_values, through: :entitlements, source: :values, class_name: "Entitlement::EntitlementValue", dependent: :destroy
+    has_many :plans, through: :entitlements
 
     validates :code, presence: true, length: {maximum: 255}
     validates :name, length: {maximum: 255}
@@ -18,6 +19,12 @@ module Entitlement
 
     def self.ransackable_attributes(_auth_object = nil)
       %w[code name description]
+    end
+
+    def subscriptions_count
+      Subscription.joins(:plan).where(plan: plans, status: [:active, :pending]).or(
+        Subscription.joins(:plan).where(plan: {parent: plans}, status: [:active, :pending])
+      ).count
     end
   end
 end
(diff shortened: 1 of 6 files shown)


app/models/entitlement/feature.rb
```diff
@@ -19,6 +20,12 @@ class Feature < ApplicationRecord
     def self.ransackable_attributes(_auth_object = nil)
       %w[code name description]
     end
+
+    def subscriptions_count
+      Subscription.joins(:plan).where(plan: plans, status: [:active, :pending]).or(
+        Subscription.joins(:plan).where(plan: {parent: plans}, status: [:active, :pending])
+      ).count
```
groyoh: I think you could have defined a base scope here

```suggestion
      base_scope = Subscription.joins(:plan).where(status: [:active, :pending])
      base_scope.where(plan: plans).or(base_scope.where(plan: {parent: plans})).count
```

?

julienbourdeau: A lot more readable! thanks!

PR #4010, opened 2025-07-16
Feat(fixed charges): fixed charge model

annvelents:
## Roadmap Task

## Context

We want to add fixed_charge model that would represent the fixed charges included in the subscription that do not depend on the received usage. This PR adds the migration and model of fixed_charges

## Description

- add migration for fixed_charges
- add model and basic tests for fixed_charges

### DIFF
--- a/db/migrate/20250716150049_create_fixed_charges.rb
+++ b/db/migrate/20250716150049_create_fixed_charges.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class CreateFixedCharges < ActiveRecord::Migration[8.0]
+  def change
+    create_table :fixed_charges, id: :uuid do |t|
+      t.belongs_to :organization, null: false, foreign_key: true, type: :uuid
+      t.belongs_to :plan, null: false, foreign_key: true, type: :uuid
+      t.belongs_to :add_on, null: false, foreign_key: true, type: :uuid
+      t.belongs_to :parent, type: :uuid, index: true
+
+      t.enum :charge_model, enum_type: "fixed_charge_charge_model", null: false, default: "standard"
+      t.jsonb :properties, null: false, default: {}
+      t.string :invoice_display_name
+      t.boolean :pay_in_advance, default: false, null: false
+      t.boolean :prorated, default: false, null: false
+      t.decimal :units, precision: 30, scale: 10, null: false, default: 0.0
+      t.datetime :deleted_at, index: true
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


spec/models/fixed_charge_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+
+require "rails_helper"
+
+RSpec.describe FixedCharge, type: :model do
+  it { is_expected.to belong_to(:organization) }
+  it { is_expected.to belong_to(:plan) }
+  it { is_expected.to belong_to(:add_on) }
+  it { is_expected.to belong_to(:parent).optional }
+
+  it { is_expected.to validate_numericality_of(:units).is_greater_than_or_equal_to(0) }
```
julienbourdeau, line 17: you can use new `expect(described_class).to be_soft_deletable` (probably works with is_expected.to but haven't tested)

db/migrate/20250716150049_create_fixed_charges.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class CreateFixedCharges < ActiveRecord::Migration[8.0]
+  def change
+    create_table :fixed_charges, id: :uuid do |t|
+      t.belongs_to :organization, null: false, foreign_key: true, type: :uuid
+      t.belongs_to :plan, null: false, foreign_key: true, type: :uuid
+      t.belongs_to :add_on, null: false, foreign_key: true, type: :uuid
+      t.belongs_to :parent, type: :uuid, index: true
+
+      t.enum :charge_model, enum_type: "fixed_charge_charge_model", null: false, default: "standard"
```
julienbourdeau, line 11: `fixed_charge_charge_model` isn't defined yet and it has only one value so far `standard`, right?

spec/models/fixed_charge_spec.rb
annvelents: owwww, niiiice! 😁

db/migrate/20250716150049_create_fixed_charges.rb
annvelents: this branch I've branched from the branch that defines this enum... this one: https://github.com/getlago/lago-api/pull/4000

spec/models/fixed_charge_spec.rb
julienbourdeau: By the way, it should actually highlight that we forgot the `default_scope { kept }`

annvelents: yep, found it at the next PR 🙈 
thank you ☺️

vincent-pochet (approved):

PR #4016, opened 2025-07-17
Feat(fixed_charges-1): migration to update fee

annvelents:
## Context

Now fee also can be related to a fixed charge, so we need to add a fixed_charge_id to fee

## Description

- migration to add fixed_charge_id to fee
- update models with relations


brunomiguelpinto (approved):

ancorcruz (approved):
LGTM!

PR #4022, opened 2025-07-18
Feat/ent sub override

julienbourdeau:
```
api_v1_subscription_entitlement_privilege DELETE                    /api/v1/subscriptions/:subscription_external_id/entitlements/:entitlement_code/privileges/:code(.:format) api/v1/subscriptions/entitlements/privileges#destroy {code: /.*/, entitlement_code: /.*/}
   remove_api_v1_subscription_entitlement POST                      /api/v1/subscriptions/:subscription_external_id/entitlements/:code/remove(.:format)                       api/v1/subscriptions/entitlements#remove {code: /.*/}
                                          DELETE                    /api/v1/subscriptions/:subscription_external_id/entitlements/:code/remove(.:format)                       api/v1/subscriptions/entitlements#unremove {code: /.*/}
         api_v1_subscription_entitlements GET                       /api/v1/subscriptions/:subscription_external_id/entitlements(.:format)                                    api/v1/subscriptions/entitlements#index
          api_v1_subscription_entitlement PATCH                     /api/v1/subscriptions/:subscription_external_id/entitlements/:code(.:format)                              api/v1/subscriptions/entitlements#update {code: /.*/}
                                          PUT                       /api/v1/subscriptions/:subscription_external_id/entitlements/:code(.:format)                              api/v1/subscriptions/entitlements#update {code: /.*/}
                                          DELETE                    /api/v1/subscriptions/:subscription_external_id/entitlements/:code(.:format)                              api/v1/subscriptions/entitlements#destroy {code: /.*/}
                                          PATCH                     /api/v1/subscriptions/:subscription_external_id/entitlements(.:format)                                    api/v1/subscriptions/entitlements#update
                                          GET|POST|PUT|DELETE|PATCH /*unmatched(.:format)                                                                                     application#not_found
```

PR #4024, opened 2025-07-18
feat(entitlements): Graphl resolver for subscription entitlements

julienbourdeau:
## How to use

This is a little different from the API endpoints because I think our apps has different need.

I expect the subscription view to do a few thing:

1. get all entitlements from the subscription plan
2. get the subscription entitlements (to be discussed if this should be part of the SubscriptionObject)
3. merge the two list by:
  1. removing features of the subscription entitlements has `removed: true`
  2. adding features from subscription of not in plan entitlements
  3. override values if feature is in both

I believe the frontend needs a way to:
- show the default plan value 
- highlight that a feature is supposed to be available but was removed specifically for this sub
  - and a way to undo the removal
- highlight that a feature  was added specifically for this sub

**This complexity is exactly why the db view exists in #4022 so we can easily retrieve the complete data.**

### Plan entitlements

```
[
    {
        "code": "seats",
        "name": "Feature Name",
        "description": "Feature Description",
        "privileges":
        [
            {
                "code": "max",
                "value": "10",
                "name": null,
                "valueType": "integer",
                "config":
                {}
            }
        ]
    },
    {
        "code": "support",
        "name": "Premium Support",
        "description": null,
        "privileges": []
    },
    {
        "code": "salesforce",
        "name": "Salesforce Integration",
        "description": null,
        "privileges": []
    }
]
```

### Subscription entitlements

```
[
    {
        "code": "seats",
        "name": "Feature Name",
        "description": "Feature Description",
        "removed": false,
        "privileges":
        [
            {
                "code": "max",
                "value": "25",
                "name": null,
                "valueType": "integer",
                "config": {}
            },
            {
                "code": "max_admin",
                "value": "2",
                "name": null,
                "valueType": "integer",
                "config": {}
            }
        ]
    },
    {
        "code": "mcp",
        "name": "MCP Server",
        "description": null,
        "privileges": []
    },
    {
        "code": "salesforce",
        "name": "Salesforce Integration",
        "description": null,
        "removed": true,
        "privileges": []
    }
]
```

### Expected merged list to display

* `seats.max` value is added overridden
* `seats.max_admin` is added
* `salesforce` is removed
* `mcp` is added

```
[
    {
        "code": "seats",
        "name": "Feature Name",
        "description": "Feature Description",
        "privileges":
        [
            {
                "code": "max",
                "value": "25",
                "name": null,
                "valueType": "integer",
                "config": {}
            },
            {
                "code": "max_admin",
                "value": "2",
                "name": null,
                "valueType": "integer",
                "config": {}
            }
        ]
    },
    {
        "code": "support",
        "name": "Premium Support",
        "description": null,
        "privileges": []
    },
    {
        "code": "mcp",
        "name": "MCP Server",
        "description": null,
        "removed": false,
        "privileges": []
    }
]
```

### DIFF
--- a/app/graphql/resolvers/entitlement/subscription_entitlements_resolver.rb
+++ b/app/graphql/resolvers/entitlement/subscription_entitlements_resolver.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module Resolvers
+  module Entitlement
+    class SubscriptionEntitlementsResolver < Resolvers::BaseResolver
+      include AuthenticableApiUser
+      include RequiredOrganization
+
+      REQUIRED_PERMISSION = "subscriptions:view"
+
+      description "Query subscription entitlements"
+
+      argument :subscription_external_id, String, required: true
+
+      type [Types::Entitlement::SubscriptionEntitlementObject], null: false
+
+      def resolve(subscription_external_id:)
+        raise forbidden_error(code: "feature_unavailable") unless License.premium?
+
+        # Get entitlements for the subscription
+        entitlements = current_organization.entitlements
+          .where(subscription_external_id: subscription_external_id)
+          .includes(:feature, :values)
+
+        # Get removed features for the subscription
+        removed_features = current_organization.subscription_feature_removals
+          .where(subscription_external_id: subscription_external_id)
+          .includes(:feature)
+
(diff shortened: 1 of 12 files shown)


julienbourdeau:
Let's keep this for future reference

PR #4027, opened 2025-07-18
Feat forecast services v2

ivannovosad:
## Context

We need to calculate prices (amounts) based on usage (units) from Data API.

## Description

This PR contains  the services involved in the calculation and also the controller for the API endpoint consumed by Data API:

```
GET /api/v1/data_api/charges/:id/forecasted_usage_amount
```


app/controllers/api/v1/data_api/charges_controller.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module DataApi
```
jdenquin: can we use a Data namespace?

jdenquin: since this is an internal endpoint, maybe it can lives in an other namespace than API? not sure we want to document/expose this one?

app/controllers/data_api/base_controller.rb
ivannovosad, line 1: What about adding a request spec to test the authentication?

app/graphql/types/data_api/usages/forecasted/object.rb
ivannovosad, line 1: Lets add a spec similar to [this one](https://github.com/getlago/lago-api/blob/feat-forecast-services-v2/spec/graphql/types/data_api/usages/invoiced/object_spec.rb).

app/controllers/data_api/v1/charges_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+              units: units,
+              charge: charge,
+              charge_filter: charge_filter
+            )
+
+            if result.success?
+              suffix = percentile_key.to_s.gsub("units_", "")
+              percentile_results["charge_amount_cents_#{suffix}"] = result.charge_amount_cents * 100
+              percentile_results["subscription_amount_cents_#{suffix}"] = result.subscription_amount_cents * 100
+              percentile_results["total_amount_cents_#{suffix}"] = result.total_amount_cents * 100
```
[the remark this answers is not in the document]
murparreira: What do you think @ivannovosad ?
I believe we need to multiply by 100, because the value is too low.

app/controllers/data_api/base_controller.rb
murparreira: Will do.

app/graphql/types/data_api/usages/forecasted/object.rb
murparreira: Will do.

app/controllers/data_api/v1/charges_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+            end
+          end
```
julienbourdeau: If we don't want to hardcode this here, we could nest it in an object and loop over it.

We'd also remove the string manipulation to get the suffix.

`charge_data` would be more like 

```rb
{
  charge_id: ...
  forecasted_units: {
    conservative: 100, 
    realistic: 120,
    optimistic: 300
  }
  ...
}
```

```suggestion
          charge_data[:forecasted_units].each do |estimate_label, units|
          
            result = Charges::CalculatePriceService.call(
              units: units,
              charge: charge,
              charge_filter: charge_filter
            )

            if result.success?
              percentile_results["charge_amount_cents_#{estimate_label}"] = result.charge_amount_cents * 100
              percentile_results["subscription_amount_cents_#{estimate_label}"] = result.subscription_amount_cents * 100
              percentile_results["total_amount_cents_#{estimate_label}"] = result.total_amount_cents * 100
            end
          end
```

app/controllers/data_api/v1/charges_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      def bulk_forecasted_usage_amount
+        charges_data = params[:charges] || []
+
+        if charges_data.empty?
+          render json: {error: "No charges provided"}, status: :bad_request
+          return
+        end
+
+        charge_ids = charges_data.map { |cd| cd[:charge_id] }.compact.uniq
```
vincent-pochet: A lor of logic is happening in this controller. Usually we tend to delegate the logic to a service class.

It helps with the single responsibility principle, and make it easier to format the controller response by relying on existing response methods defined in `render_error_response`

PR #4035, opened 2025-07-21
feat(entitlements): Add value validation

julienbourdeau:
All values are stored as string. This ensure that you send the value in the correct type.

As I'm wigint these lines, I realize that it's gonna be a nightmare for graphql.

### DIFF
--- a/spec/services/entitlement/plan_entitlements_update_service-full_spec.rb
+++ b/spec/services/entitlement/plan_entitlements_update_service-full_spec.rb
@@ -194,7 +194,7 @@ RSpec.describe Entitlement::PlanEntitlementsUpdateService, type: :service do
       it "converts boolean to string" do
         result
         entitlement_value = plan.entitlements.first.values.first
-        expect(entitlement_value.value).to eq("true")
+        expect(entitlement_value.value).to eq("t")
       end
     end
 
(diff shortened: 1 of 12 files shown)


spec/services/entitlement/plan_entitlements_update_service-full_spec.rb
```diff
@@ -194,7 +194,7 @@
       it "converts boolean to string" do
         result
         entitlement_value = plan.entitlements.first.values.first
-        expect(entitlement_value.value).to eq("true")
+        expect(entitlement_value.value).to eq("t")
```
groyoh, line 197: Why would this be `t` and not `true` ?

julienbourdeau: Because ActiveRecord will cast the value to be stored in the string column.
Before I sent `"true"` to it was stored as-is and converted when serialized.
Now I send `true` as a bool but the column type is string, so AR converts to `"t"`, which is still casted properly by `ActiveModel::Type::Boolean.new.cast(value)`

groyoh: Ok interesting.

groyoh (approved):

PR #4047, opened 2025-07-22
feat(subscription): Allow to skip invoice generation on subscription termination

groyoh:
## Context

When a subscription is terminated,whether due to an error or otherwise,Lago automatically generates a closing invoice for pay-in-arrears fees. This inflexible behavior didn't accommodate different business needs where organizations might want to skip the invoice generation.

## Description

This introduces an `on_termination_invoice` enum field to subscriptions that controls whether an invoice is generated for pay-in-arrears fees. The field supports "generate" (default) and "skip" values, allowing organizations to choose their preferred termination behavior.

The termination service now respects this setting and only generates an invoice when explicitly configured to do so. The API accepts the termination behavior as a parameter during subscription termination.

The implementation includes database schema changes, API parameter handling, service logic updates, and comprehensive test coverage for all termination scenarios. Scenario specs for termination behaviors will come later on.

### DIFF
--- a/app/services/subscriptions/terminate_service.rb
+++ b/app/services/subscriptions/terminate_service.rb
@@ -4,11 +4,12 @@ module Subscriptions
   class TerminateService < BaseService
     Result = BaseResult[:subscription]
 
-    def initialize(subscription:, async: true, upgrade: false, on_termination_credit_note: subscription&.on_termination_credit_note)
+    def initialize(subscription:, async: true, upgrade: false, on_termination_credit_note: subscription&.on_termination_credit_note, on_termination_invoice: subscription&.on_termination_invoice)
       @subscription = subscription
       @async = async
       @upgrade = upgrade
-      @on_termination_credit_note = on_termination_credit_note
+      @on_termination_credit_note = on_termination_credit_note.blank? ? :credit : on_termination_credit_note.to_sym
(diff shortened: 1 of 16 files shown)


app/services/subscriptions/terminate_service.rb
```diff
@@ -197,8 +202,25 @@ def on_termination_credit_note
...
 
-    def update_on_termination_credit_note!
-      Subscriptions::UpdateService.call!(subscription:, params: {on_termination_credit_note:})
+    def on_termination_invoice
+      return :generate if @on_termination_invoice.blank? || @on_termination_invoice.to_sym == :generate
+      return :skip if @on_termination_invoice.to_sym == :skip
+
+      # This will cause a validation error on update
+      @on_termination_invoice
+    end
```
julienbourdeau: I found this a little surprising, but it's not important. I'd have done this when assigning 🤔 
```rb
@on_termination_invoice = on_termination_invoice.to_sym || :generate
```

julienbourdeau (approved):

app/services/subscriptions/terminate_service.rb
groyoh: Nice point! Maybe some AI weirdness. I've updated it.

groyoh:
Thanks for the review @julienbourdeau 🙏

PR #4049, opened 2025-07-22
misc(credit-note): Refactor `CreditNotes::CreateFromTermination` spec

groyoh:
## Context

The current spec has quite a bit of unnecessary code repetition.

## Description

This wraps the test code into a `test_credit_note_creation_from_termination` functions that calls the service and runs assertions on the result. I used a `expectations` kwarg instead of adding the expectations as kwargs to make it more explicit that they are the expectations and not the test configuration.


julienbourdeau (approved):
I don't do this enough

groyoh:
Thanks @julienbourdeau 👍

> I don't do this enough

Yeah, I enjoy making it easier  for future me 😅

PR #4050, opened 2025-07-22
feat(pricing-units): Add API for pricing units

floganz:
## Context

Add minimal API changes to allow to work with pricing units in a system.
Ability to perform CRUD action with pricing unit model itself for now removed from scope.

## Description

Modify serializers to return information about applied pricing units for charge, pricing units amounts for fee and usage.
Allow to specify applied to a charge pricing units during plan or subscription created/update actions.


app/serializers/v1/charge_serializer.rb
```diff
@@ -43,6 +43,15 @@ def charge_filters
       ).serialize
     end
 
+    def applied_pricing_unit
+      return if model.applied_pricing_unit.nil?
+
+      {
+        conversion_rate: model.applied_pricing_unit.conversion_rate,
+        code: model.pricing_unit.code
```
julienbourdeau, line 51: Is there anywhere we'd want to eagerload this? Or maybe it's already done.

julienbourdeau (approved):

app/serializers/v1/charge_serializer.rb
floganz: You're right!
Added it to plan serializer.

PR #4051, opened 2025-07-22
feat(entitlement): append new select_options + not null db column

julienbourdeau:
Allow `config.select_options` to be updated for `select` privilege.

I don't know why the config column is nullable in the DB, this is a mistake.
Notice that I can simply change the column with "safety_assured" because **the feature is not released and there was no tagged version [since I added this model](https://github.com/getlago/lago-api/pull/3873).**

### DIFF
--- a/app/services/entitlement/feature_update_service.rb
+++ b/app/services/entitlement/feature_update_service.rb
@@ -70,6 +70,13 @@ module Entitlement
           create_privilege(code, privilege_params)
         else
           privilege.name = privilege_params[:name] if privilege_params.key?(:name)
+
+          if privilege_params.dig(:config, :select_options)
+            privilege.config ||= {}
+            privilege.config["select_options"] ||= []
+            privilege.config["select_options"] |= privilege_params[:config][:select_options]
+          end
+
           privilege.save!
         end
       end
--- a/app/services/entitlement/feature_update_service.rb
+++ b/app/services/entitlement/feature_update_service.rb
@@ -72,7 +72,6 @@ module Entitlement
           privilege.name = privilege_params[:name] if privilege_params.key?(:name)
 
           if privilege_params.dig(:config, :select_options)
-            privilege.config ||= {}
             privilege.config["select_options"] ||= []
             privilege.config["select_options"] |= privilege_params[:config][:select_options]
           end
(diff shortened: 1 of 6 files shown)


db/migrate/20250630180000_create_entitlement_features_and_privileges.rb
```diff
@@ -24,7 +24,7 @@ def change
       t.string :code, null: false
       t.string :name
       t.enum :value_type, enum_type: "entitlement_privilege_value_types", null: false
-      t.jsonb :config, default: {}
+      t.jsonb :config, default: {} # NOT NULL see: db/migrate/20250722094047_change_privilege_config_to_not_null.rb
```
jdenquin: can you add it one line before 

```ruby
# NOTE: NOT NULL see: db/migrate/20250722094047_change_privilege_config_to_not_null.rb
t.jsonb :config, default: {}
```

app/services/entitlement/feature_update_service.rb
```diff
@@ -70,6 +70,12 @@ def update_privileges
           create_privilege(code, privilege_params)
         else
           privilege.name = privilege_params[:name] if privilege_params.key?(:name)
+
+          if privilege_params.dig(:config, :select_options)
+            privilege.config["select_options"] ||= []
+            privilege.config["select_options"] |= privilege_params[:config][:select_options]
```
groyoh, line 76: This append behavior is confusing to me. Shouldn't we allow to provide a superset of the existing options ? For instance, if we have `[one two]`, you can provide `[one two three]`.

julienbourdeau: This forces you to resend all values to add one. ~but it's a good point, maybe this should depend on the `partial` args!~ actually no but we want to ensure that the values are not deleted. The frontend will send a superset and if they don't we still won't remove values

groyoh: > This forces you to resend all values to add one

TBH I've never seen an endpoint when sending a value appends to existing ones. Usually, when that happens you have an `append` key or something similar.

But if you're OK with this behavior, then it's all good 👍

groyoh (approved):

PR #4053, opened 2025-07-22
feat(subscription): Allow to skip invoice on subscription termination via GraphQL

groyoh:
## Context

When a subscription is terminated,whether due to an error or otherwise,Lago automatically generates a closing invoice for pay-in-arrears fees. This inflexible behavior didn't accommodate different business needs where organizations might want to skip the invoice generation.

The API already accepts the termination behavior as a parameter during subscription termination but the GraphQL API did not.

## Description

This update the `terminateSubscription` mutation to handle the `on_termination_invoice` behavior.


app/graphql/types/subscriptions/object.rb
```diff
@@ -17,7 +17,8 @@ class Object < Types::BaseObject
       field :billing_time, Types::Subscriptions::BillingTimeEnum
       field :canceled_at, GraphQL::Types::ISO8601DateTime
       field :ending_at, GraphQL::Types::ISO8601DateTime
-      field :on_termination_credit_note, Types::Subscriptions::OnTerminationCreditNoteEnum, null: true
+      field :on_termination_credit_note, Types::Subscriptions::OnTerminationCreditNoteEnum
+      field :on_termination_invoice, Types::Subscriptions::OnTerminationInvoiceEnum, null: false
```
julienbourdeau, line 21: I always find it confusing to know if it's nullable by default 😄

julienbourdeau (approved):

app/graphql/types/subscriptions/object.rb
groyoh: That's exactly what I was thinking when writing this. This should be a Rubocop rule to be honest.

PR #4056, opened 2025-07-23
fix(entitlements): Use Array of privileges instead of hash indexed by code

julienbourdeau:
This PR modifies the serializer for features and privileges, as well as the payload to manage features.
The payload to manage entitlements (attaching features to plans *with values* is still a hash.

The first thing that made me want to change this was GQL, which was not so easy to use with hash.
Then, I wrote the open API spec and it was even worse.

### serialized feature / Payload to manage features

```diff
      {
        feature: {
          code: "new_feature",
          name: "New Feature",
          description: "A new feature",
-          privileges: {
-            "priv1" => {value_type: "boolean"},
-            "priv2" => {name: "Maximum", value_type: "boolean"}
-          }
+          privileges: [
+            {code: "priv1", value_type: "boolean"},
+            {code: "priv2", name: "Maximum", value_type: "boolean"}
+          ]
        }
      }
```

### payload to manage entitlements

```rb
{
  "entitlements" => {
    "seats" => {
      "max" => 25
    }
  }
}
```

I think to EDIT entitlements, a hash is a lot nicer. When you receive it back, it's an array.

### DIFF
--- a/app/controllers/api/v1/features_controller.rb
+++ b/app/controllers/api/v1/features_controller.rb
@@ -101,11 +101,15 @@ module Api
       private
 
       def feature_create_params
-        params.require(:feature).permit(:code, :name, :description, privileges: {})
+        params.require(:feature).permit(:code, :name, :description, privileges: [
+          :code, :name, :value_type, :config
+        ])
       end
 
       def feature_update_params
-        params.require(:feature).permit(:name, :description, privileges: {})
+        params.require(:feature).permit(:name, :description, privileges: [
+          :code, :name, :value_type, :config
+        ])
       end
 
       def root_name
--- a/app/graphql/mutations/entitlement/create_feature.rb
+++ b/app/graphql/mutations/entitlement/create_feature.rb
@@ -21,7 +21,7 @@ module Mutations
             code: args[:code],
             name: args[:name],
             description: args[:description],
-            privileges: args[:privileges].map(&:to_h).index_by { it[:code] }
+            privileges: args[:privileges].map(&:to_h)
           }
         )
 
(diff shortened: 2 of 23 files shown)


app/controllers/api/v1/features_controller.rb
```diff
@@ -101,11 +101,15 @@ def destroy
       private
 
       def feature_create_params
-        params.require(:feature).permit(:code, :name, :description, privileges: {})
+        params.require(:feature).permit(:code, :name, :description, privileges: [
+          :code, :name, :value_type, :config
```
groyoh, line 105: Much nicer to sanitize!

app/graphql/mutations/entitlement/create_feature.rb
```diff
@@ -21,7 +21,7 @@ def resolve(**args)
             code: args[:code],
             name: args[:name],
             description: args[:description],
-            privileges: args[:privileges].map(&:to_h).index_by { it[:code] }
+            privileges: args[:privileges].map(&:to_h)
```
groyoh, line 24: Do we need `.map(&:to_h)` ? 🤔

groyoh (approved):

app/graphql/mutations/entitlement/create_feature.rb
julienbourdeau: I like that it's a proper hash and not a graphql object but it should work without it

app/controllers/api/v1/features_controller.rb
julienbourdeau: Absolutely!

app/graphql/mutations/entitlement/create_feature.rb
julienbourdeau: Actually I get this error

CleanShot 2025-07-23 at 10 36 31@2x

groyoh: Oh! I didn't realize GraphQL input were not already casted as Hashes. 👍

PR #4060, opened 2025-07-23
fix(event-aggregation): the query should group prorated units by day

annvelents:
Reverts getlago/lago-api#4059

This is a revert for  getlago/lago-api#4059, which brings back getlago/lago-api#3890 with update of clickhouse query as current version of CLickhouse at production does not support JOIN on != expression, so instead we change logic how we ignore remove event:
remove event is only not ignored, if it's the last event of the day for this property. This approach requires window function, which is less performant that postgres's version of join on next event and selecting 1, but this approach cannot be implemented yet for clickhouse

Original description of PR#3890
## Context

when for uniq count we post several events per day like:
27th June - 10:00 - property - Add
27th June - 11:00 - property - Remove
27th June - 12:00 - property - Add
27th June - 13:00 - property - Remove

current logic will count it as two events with minimal length of proration 1 day, so the customer will be charged for 2 prorated days of usage.

Instead, we want to group events by day, because we prorate them through day, and for one day record only one event

another scenario:
27th June - 10:00 - property - Add
28th June - 10:00 - property - Remove
28th June - 11:00 - property - Add

if we "end" usage on 28th, the first Add event will "cover" usage on 28th. The event received on 28th will not do the double charge on 28th, and will "start" on 29th, but: as result anyway we'll charge for the usage on 27th, 28, 29.. so we can just ignore the remove event, and aggregate all this as as one event started on 27th

## Description

Updated prorated uniq_count_queries for postgresql and clickhouse
Algorithm:
1) prepare events table
2) ignore only REMOVE events that will be "cancelled" by the following Add event at the same day
3) using adjusted value logic nullify not needed Adds (if previous row has the same operation_type, current has adjusted value 0; otherwise it's 1 for Add; -1 for Remove)

this pr is a pure pain.
especially clickhouse - it can't do recursion, LAG and LOG are very heavy there and tests were killing the db... so I needed to write a completely different approach from postgres, thenI rewrote couple of approaches on both stores, finally found a bug in clickhouse proration, and here is the result

### DIFF
--- a/app/services/events/stores/postgres/unique_count_query.rb
+++ b/app/services/events/stores/postgres/unique_count_query.rb
@@ -57,7 +57,7 @@ module Events
                   timestamp,
                   property,
                   operation_type,
-                  #{operation_value_sql} AS adjusted_value,
+                  #{operation_value_sql} AS adjusted_value
                 FROM same_day_ignored
                 WHERE is_ignored = false
                 ORDER BY timestamp ASC
@@ -410,7 +410,7 @@ module Events
                 1
               FROM events_data next_event
               WHERE next_event.property = e.property
-                AND DATE(next_event.timestamp) = DATE(e.timestamp)
+                AND DATE((next_event.timestamp)::timestamptz AT TIME ZONE :timezone) = DATE((e.timestamp)::timestamptz AT TIME ZONE :timezone)
                 AND next_event.operation_type <> e.operation_type
                 AND next_event.timestamp > e.timestamp
               LIMIT 1
@@ -426,7 +426,7 @@ module Events
               FROM events_data next_event
               WHERE next_event.property = e.property
                 AND #{group_names.map { |name| "next_event.#{name} = e.#{name}" }.join(" AND ")}
-                AND DATE(next_event.timestamp) = DATE(e.timestamp)
+                AND DATE((next_event.timestamp)::timestamptz AT TIME ZONE :timezone) = DATE((e.timestamp)::timestamptz AT TIME ZONE :timezone)
                 AND next_event.operation_type <> e.operation_type
                 AND next_event.timestamp > e.timestamp
               LIMIT 1


brunomiguelpinto (approved):

app/services/events/stores/clickhouse/unique_count_query.rb
```diff
@@ -34,9 +34,45 @@ def query
...
+                  WHEN operation_type = 'remove' AND NOT is_last_event_of_day THEN true
+                  ELSE false
+                END AS is_ignored
+              FROM (
+                SELECT
+                  timestamp,
+                  property,
+                  operation_type,
+                  -- Check if this is the last event of the day for this property
+                  timestamp = MAX(timestamp) OVER (PARTITION BY property, toDate(timestamp)) AS is_last_event_of_day
```
vincent-pochet: Since we are now working with date and not datetime, I think we should consider date in customer timezone. What do you think?

app/services/events/stores/clickhouse/unique_count_query.rb
```diff
@@ -158,6 +222,30 @@ def breakdown_query
         def prorated_breakdown_query(with_remove: false)
           <<-SQL
             #{events_cte_sql},
+            -- Only ignore remove events if they are NOT the last event of the day
```
vincent-pochet, line 212: I might be wrong but it seems that this is the same sub-query as the "regular" prorated_query, could we maybe extract it in a method to reduce duplication?

app/services/events/stores/postgres/unique_count_query.rb
```diff
@@ -330,8 +381,60 @@ def grouped_period_ratio_sql
...
+        # -- for the same day, we look at next event. if it's opposite of current, current can be ignored
+        # -- we look at previous not ignored event. if the operation type matches. we can ignore current
+        def existing_event_opposite_operation_type_sql
+          <<-SQL
+            (
+              SELECT
+                1
+              FROM events_data next_event
+              WHERE next_event.property = e.property
+                AND DATE(next_event.timestamp) = DATE(e.timestamp)
```
vincent-pochet: As for clickhouse, I guess you have to consider the day from a customer timezone point of view

vincent-pochet (commented):
Good one 👍!

Still one point regarding date and timezones. Happy to discuss about it :)

app/services/events/stores/clickhouse/unique_count_query.rb
annvelents: that's absolutely right and great suggestion - did it!

PR #4066, opened 2025-07-24
misc: add invoice payment link generation GQL mutation

ansmonjol:
## Context

API users are allowed to download a payment url for invoices, under conditions. This is not permitted through the UI

## Description

This PR adds a new GQL mutation to allow calling the GeneratePaymentUrlService and retrieve the link


app/graphql/mutations/invoices/generate_payment_url.rb
```diff
@@ -0,0 +1,26 @@
...
+      REQUIRED_PERMISSION = "invoices:update"
+
+      description "Generates a payment URL for an invoice"
+
+      argument :invoice_id, ID, required: true
+
+      field :payment_url, String, null: true
+
+      def resolve(invoice_id:)
+        invoice = current_organization.invoices.find_by(id: invoice_id)
```
groyoh: Should we keep the same logic as the API:

```suggestion
        invoice = current_organization.invoices.visible.includes(:customer).find_by(id: invoice_id)
```

?

spec/graphql/mutations/invoices/generate_payment_url_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        }
+      }
+    )
+
+    result_data = result["data"]["generatePaymentUrl"]
+
+    aggregate_failures do
+      expect(result_data["paymentUrl"]).to eq("https://payment.example.com/pay/123")
+      expect(Invoices::Payments::GeneratePaymentUrlService).to have_received(:call).with(invoice:)
+    end
```
groyoh: `aggregate_failures` is the default :

```suggestion
    expect(result_data["paymentUrl"]).to eq("https://payment.example.com/pay/123")
    expect(Invoices::Payments::GeneratePaymentUrlService).to have_received(:call).with(invoice:)
```

spec/graphql/mutations/invoices/generate_payment_url_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  before do
+    allow(Invoices::Payments::GeneratePaymentUrlService).to receive(:call).and_return(service_result)
+  end
+
+  it_behaves_like "requires current user"
+  it_behaves_like "requires current organization"
+  it_behaves_like "requires permission", "invoices:update"
+
+  it "generates a payment URL for the invoice" do
```
groyoh, line 36: You could maybe add a test for when the invoice is not found.

app/graphql/mutations/invoices/generate_payment_url.rb
ansmonjol: Oh yeah indeed. I didn't really get this query but we must follow the same pattern yep. Will update

ansmonjol: Done!

spec/graphql/mutations/invoices/generate_payment_url_spec.rb
ansmonjol: Done!

spec/graphql/mutations/invoices/generate_payment_url_spec.rb
ansmonjol: Done!

PR #4068, opened 2025-07-25
chore(gql): update plan entitlements directly in Plan update mutation

julienbourdeau:
* Allow plan entitlements to be sent as part of the PlanInput object. I initially made it into a separate mutation.

* I improved the value validation because sending valid data as string should be considered valid... and because this is how graphql send them 😄 

* Fixed the `::Entitlement::SubscriptionEntitlementsUpdateService` since it wasn't used un `partial: false` yet.

* Add ability to modify subscriptions entitlements in the subscription update mutation.

### DIFF
--- a/app/services/entitlement/subscription_entitlements_update_service.rb
+++ b/app/services/entitlement/subscription_entitlements_update_service.rb
@@ -120,9 +120,7 @@ module Entitlement
 
         # if the feature was previously removed, we restore it
         removal = subscription.entitlement_removals.find { it.entitlement_feature_id == feature.id }
-        if removal
-          removal.discard!
-        end
+        removal&.discard!
 
         entitlement = subscription.entitlements.includes(:values).find { it.entitlement_feature_id == feature.id }
 
--- a/spec/services/entitlement/subscription_entitlements_update_service-partial_spec.rb
+++ b/spec/services/entitlement/subscription_entitlements_update_service-partial_spec.rb
@@ -62,7 +62,7 @@ RSpec.describe Entitlement::SubscriptionEntitlementsUpdateService, type: :servic
     it "creates the entitlement with correct values and leave existing untouched" do
       result
 
-      expect(entitlement.reload.values.map(&:value)).to eq(["100", "t"])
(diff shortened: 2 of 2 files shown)


app/graphql/mutations/plans/update.rb
```diff
@@ -13,12 +13,29 @@ class Update < BaseMutation
...
         args[:charges].map!(&:to_h)
         plan = context[:current_user].plans.find_by(id: args[:id])
 
         result = ::Plans::UpdateService.call(plan:, params: args)
-        result.success? ? result.plan : result_error(result)
+
+        return result_error(result) unless result.success?
+
+        if entitlements.present? && License.premium?
+          result = ::Entitlement::PlanEntitlementsUpdateService.call(
```
groyoh, line 25: Shoulnd't you use a transaction ?

app/services/entitlement/subscription_entitlements_update_service.rb
```diff
@@ -56,7 +56,31 @@ def call
     attr_reader :organization, :subscription, :entitlements_params, :partial
     alias_method :partial?, :partial
 
+    def full?
```
groyoh, line 61: 👍

app/services/entitlement/subscription_entitlements_update_service.rb
```diff
@@ -56,7 +56,31 @@ def call
...
+      plan_entitlements.each do |entitlement|
+        SubscriptionFeatureRemoval.create!(
+          organization: subscription.organization,
+          feature: entitlement.feature,
+          subscription: subscription
+        )
+      end
+    end
+
     def delete_missing_entitlements
```
groyoh: I'm missing something. Is this still used ?

julienbourdeau: oh no, it should be remvoed!

app/graphql/mutations/plans/update.rb
julienbourdeau: that is a good point 🤔 I'm tempted to start without the transaction

PR #4070, opened 2025-07-25
Fix(payments): Allow manual payments only for finalized and open invoices

annvelents:
## Context

Right now we only check if invoice is draft, but if invoice doesn't have taxes, we cannot say if invoice is paid or not, so we shouldn't allow to record payments for the invoice. As well as it does not make sence to record payments for invoices that are already voided, especially because we're giving an option for voided invoices to refund a customer back

## Description

Added a check for invoice status before accepting a manual payment

### DIFF
--- a/spec/services/payments/manual_create_service_spec.rb
+++ b/spec/services/payments/manual_create_service_spec.rb
@@ -18,10 +18,8 @@ RSpec.describe Payments::ManualCreateService, type: :service do
       it "returns forbidden failure" do
         result = service.call
 
-        aggregate_failures do
-          expect(result).not_to be_success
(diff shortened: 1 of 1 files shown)


spec/services/payments/manual_create_service_spec.rb
```diff
@@ -41,6 +41,19 @@
         end
       end
 
+      context "when invoice is in status that does not allow manual payment" do
+        let(:invoice) { create(:invoice, :draft, customer:, organization:) }
+
+        it "returns forbidden failure" do
+          result = service.call
+
+          aggregate_failures do
```
groyoh: I believe there's no need for aggregate failure as it's the default.

groyoh (approved):

PR #4072, opened 2025-07-25
fix(daily-usage): Refactor DailyUsages::FillHistoryService

rsempe:
This pull request refactors the `DailyUsages::FillHistoryService` and related components to replace `datetime`-based parameters with `date`-based ones, improving consistency and simplifying the handling of date ranges.

### DIFF
--- a/app/services/daily_usages/fill_history_service.rb
+++ b/app/services/daily_usages/fill_history_service.rb
@@ -88,15 +88,22 @@ module DailyUsages
     delegate :organization, to: :subscription
 
     def from
-      @from ||= [subscription.started_at.to_date, from_date].max
+      @from ||= [
+        subscription.started_at.in_time_zone(timezone).to_date,
+        from_date
+      ].max
     end
 
     def to
       @to ||= if subscription.terminated?
-        subscription.terminated_at.to_date
+        subscription.terminated_at.in_time_zone(timezone).to_date
       else
-        to_date || Time.zone.yesterday
+        to_date || Time.zone.yesterday.in_time_zone(timezone).to_date
       end
+    end
+
+    def timezone
+      @timezone ||= subscription.customer.applicable_timezone
     end
   end
 end
--- a/app/services/daily_usages/fill_history_service.rb
+++ b/app/services/daily_usages/fill_history_service.rb
@@ -4,10 +4,10 @@ require "timecop"
 
 module DailyUsages
   class FillHistoryService < BaseService
-    def initialize(subscription:, from_datetime:, to_datetime: nil, sandbox: false)
+    def initialize(subscription:, from_date:, to_date: nil, sandbox: false)
       @subscription = subscription
-      @from_datetime = from_datetime
-      @to_datetime = to_datetime
+      @from_date = from_date
+      @to_date = to_date
       @sandbox = sandbox
 
       super
@@ -17,27 +17,20 @@ module DailyUsages
       previous_daily_usage = nil
 
       (from..to).each do |date|
+        next if !sandbox && subscription.daily_usages.where(usage_date: date).exists?
+
         datetime = date.in_time_zone(subscription.customer.applicable_timezone).beginning_of_day.utc
         datetime = date.beginning_of_day.utc if datetime < date # Handle last day for timezone with positive offset
 
-        next if !sandbox && subscription.daily_usages.where(usage_date: datetime.to_date - 1.day).exists?
-
-        ds = Subscriptions::DatesService.new_instance(subscription, date, current_usage: true)
-
-        time_to_freeze = if ds.previous_beginning_of_period.to_date == date
-          (datetime - 1.day).end_of_day
-        else
-          datetime + 1.second
-        end
-
         Timecop.thread_safe = true
+        time_to_freeze = datetime.in_time_zone(subscription.customer.applicable_timezone).end_of_day
         Timecop.freeze(time_to_freeze) do
           usage = Invoices::CustomerUsageService.call(
             customer: subscription.customer,
             subscription: subscription,
             apply_taxes: false,
             with_cache: false,
-            max_to_datetime: datetime,
+            max_to_datetime: Time.current,
             with_zero_units_filters: false
           ).raise_if_error!.usage
           next if sandbox
@@ -60,7 +53,7 @@ module DailyUsages
               to_datetime: usage.to_datetime,
               refreshed_at: datetime,
               usage_diff: {},
-              usage_date: datetime.to_date - 1.day
+              usage_date: date
             )
 
             if date != from
@@ -91,22 +84,18 @@ module DailyUsages
       result
     end
 
-    attr_reader :subscription, :from_datetime, :to_datetime, :sandbox
+    attr_reader :subscription, :from_date, :to_date, :sandbox
     delegate :organization, to: :subscription
 
     def from
-      return @from if defined?(@from)
-
-      @from = subscription.started_at.to_date
-      @from = from_datetime.to_date if @from < from_datetime
-      @from
+      @from ||= [subscription.started_at.to_date, from_date].max
     end
 
     def to
       @to ||= if subscription.terminated?
         subscription.terminated_at.to_date
       else
-        (to_datetime || Time.current).to_date
+        to_date || Time.zone.yesterday
       end
     end
   end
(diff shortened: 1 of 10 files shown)


app/services/daily_usages/fill_history_service.rb
```diff
@@ -57,9 +46,9 @@ def call
               usage: ::V1::Customers::UsageSerializer.new(usage, includes: %i[charges_usage]).serialize,
               from_datetime: usage.from_datetime,
               to_datetime: usage.to_datetime,
-              refreshed_at: datetime,
+              refreshed_at: date.end_of_day,
```
iavorbotev: What is `refreshed_at` used for? Should it be a UTC timestamp?

app/services/daily_usages/fill_history_service.rb
```diff
@@ -57,9 +46,9 @@ def call
...
               usage_diff: {},
-              usage_date: datetime.to_date - 1.day
+              usage_date: date
```
iavorbotev, line 56: `date` doesn't have a timestamp, right? How does the `CustomerUsageService` know what is the right timezone for the customer so that it knows which events to include in a given day?

iavorbotev (commented):
Added a couple of questions.

Also, FIY, I'm working on a test file in lago-api that covers multiple scenarios for the FillHistoryService.

app/services/daily_usages/fill_history_service.rb
rsempe: The `refreshed_at` is the time the daily_usage record was computed in UTC.

For the backfill task, I'm thinking of setting it to the end of the usage date (something like `2025-07-30 23:59:59`).

app/services/daily_usages/fill_history_service.rb
rsempe: I've changed it to pass the end of day as well here.

`CustomerUsageService ` is fetching the usage based on the `max_to_datetime` parameter that correspond to `date.end_of_day` here.

vincent-pochet (approved):
🚀

app/services/daily_usages/fill_history_service.rb
```diff
@@ -17,27 +17,20 @@ def call
       previous_daily_usage = nil
 
       (from..to).each do |date|
+        next if !sandbox && subscription.daily_usages.where(usage_date: date).exists?
+
         datetime = date.in_time_zone(subscription.customer.applicable_timezone).beginning_of_day.utc
         datetime = date.beginning_of_day.utc if datetime < date # Handle last day for timezone with positive offset
```
iavorbotev, line 23: I don't understand what we're trying to do in this line, could you pls explain a bit?

app/services/daily_usages/fill_history_service.rb
```diff
@@ -17,27 +17,20 @@ def call
...
         Timecop.freeze(time_to_freeze) do
           usage = Invoices::CustomerUsageService.call(
             customer: subscription.customer,
             subscription: subscription,
             apply_taxes: false,
             with_cache: false,
-            max_to_datetime: datetime
+            max_to_datetime: time_to_freeze
           ).raise_if_error!.usage
           next if sandbox
```
iavorbotev: In the line below, could we improve the if-condition? Maybe add something along the lines of 
`usage.from_datetime > previous_daily_usage.to_datetime && usage.from_datetime < (previous_daily_usage.to_datetime + 1.hour)`. I use `1.hour` as an example.

app/services/daily_usages/fill_history_service.rb
```diff
@@ -90,22 +83,18 @@ def call
       result
```
iavorbotev, line 84: In line 72, why do we check `if subscription.terminated?` after we calculate all the usages data in the for loop?

app/services/daily_usages/fill_history_service.rb
```diff
@@ -90,22 +83,18 @@ def call
...
+    attr_reader :subscription, :from_date, :to_date, :sandbox
     delegate :organization, to: :subscription
 
     def from
-      return @from if defined?(@from)
-
-      @from = subscription.started_at.to_date
-      @from = from_datetime.to_date if @from < from_datetime
-      @from
+      @from ||= [subscription.started_at.to_date, from_date].max
```
iavorbotev: I think `subscription.started_at`should be `subscription.started_at.in_time_zone(subscription.customer.applicable_timezone)`, wdyt?

app/services/daily_usages/fill_history_service.rb
```diff
@@ -90,22 +83,18 @@ def call
...
     end
 
     def to
       @to ||= if subscription.terminated?
         subscription.terminated_at.to_date
       else
-        (to_datetime || Time.current).to_date
+        to_date || Time.zone.yesterday
```
iavorbotev: Not 100%, but should we not take customer timezone into account? In other words, use something like `Time.now.in_time_zone(subscription.customer.applicable_timezone).yesterday.to_date`

app/services/daily_usages/fill_history_service.rb
```diff
@@ -90,22 +83,18 @@ def call
...
-
-      @from = subscription.started_at.to_date
-      @from = from_datetime.to_date if @from < from_datetime
-      @from
+      @from ||= [subscription.started_at.to_date, from_date].max
     end
 
     def to
       @to ||= if subscription.terminated?
         subscription.terminated_at.to_date
```
iavorbotev: also, perhaps we should convert subscription.terminated_at to the customer timezone.

app/services/daily_usages/fill_history_service.rb
rsempe: Maybe @ivannovosad you could answer as you added that ;)

app/services/daily_usages/fill_history_service.rb
rsempe: Not sure to see the changes here, can you open a PR with the proposed changes?

app/services/daily_usages/fill_history_service.rb
rsempe: if the subscription is terminated, we still need to compute and backfill all the previous daily usages

but on the termination day, we also want to have the daily usage present

app/services/daily_usages/fill_history_service.rb
rsempe: Good spot, probably an edge case but yeah timezones should be used for `started_at ` and `terminated_at`

Will add a spec for this to validate it 👍

app/services/daily_usages/fill_history_service.rb
iavorbotev: Sorry, the line I was asking about was not changed in this PR and I couldn't tag it.

I was asking about 
`if previous_daily_usage.present? && previous_daily_usage.from_datetime != usage.from_datetime`. 

If we could think of a quick and easy improvement to it, maybe we can do it? I'm not sure myself about how to improve it, given we don't have a billing period entity.

Also tagging @vincent-pochet, since he wrote the line.

rsempe: TBH I don't think we have issues due to that condition right now, and the most important for me is `previous_daily_usage.from_datetime != usage.from_datetime`. This is `true` only for the first day of the next billing period.

iavorbotev: That makes sense 👍

app/services/daily_usages/fill_history_service.rb
ivannovosad: Yes, it should be in customer's timezone.

app/services/daily_usages/fill_history_service.rb
ivannovosad: It was added in [this PR](https://github.com/getlago/lago-api/pull/3797).

iavorbotev: Sorry, I was asking about the line below, I think you wrote that one, @ivannovosad? 
`datetime = date.beginning_of_day.utc if datetime < date # Handle last day for timezone with positive offset`

iavorbotev: I just had a call with Ivan. @rsempe  just to make sure we're speaking about the same line, 😄 I am asking about the line changed in this [PR](https://github.com/getlago/lago-api/pull/3005/files) months ago by you 🙂

PR #4078, opened 2025-07-30
Change gross_revenue_share and net_revenue_share to display nil values on graphql query

murparreira:
## Context

Discussion here: https://getlago.slack.com/archives/C086GATCS3G/p1753713358811619
These values can be nil, and the graphql query was configured with `null: false`.

## Description

Just changed the values to true.


iavorbotev (approved):
LGTM! Just to restate the obvious, we shouldn't merge it until the front-end team makes the relevant changes.

PR #4081, opened 2025-07-31
fix(vies): Fail invoice if customer's VIES check is in progress

floganz:
## Context

Currently failed VIES check does not prevent invoice finalization which lead to situations when EU customers gets wrong tax amounts.

## Description

These changes will fail invoice if VIES check is not finished on a customer.

### DIFF
--- a/app/models/customer.rb
+++ b/app/models/customer.rb
@@ -263,6 +263,14 @@ class Customer < ApplicationRecord
     anrok_customer || avalara_customer
   end
 
+  def vies_check_in_progress?
+    if billing_entity.eu_tax_management?
+      taxes.where("code ILIKE ?", "lago_eu%").none?
+    else
+      false
+    end
+  end
+
   private
 
   def ensure_slug
--- a/app/services/invoices/compute_taxes_and_totals_service.rb
+++ b/app/services/invoices/compute_taxes_and_totals_service.rb
@@ -19,6 +19,12 @@ module Invoices
         after_commit { Invoices::ProviderTaxes::PullTaxesAndApplyJob.perform_later(invoice:) }
 
         return result.unknown_tax_failure!(code: "tax_error", message: "unknown taxes")
+      elsif customer.vies_check_in_progress?
+        invoice.status = "pending" if finalizing
+        invoice.tax_status = "failed"
+        invoice.save!
+
+        return result.unknown_tax_failure!(code: "tax_error", message: "vies check in progress")
       else
         Invoices::ComputeAmountsFromFees.call(invoice:)
       end
@@ -30,6 +36,8 @@ module Invoices
     private
 
     attr_reader :invoice, :finalizing
+
+    delegate :customer, to: :invoice
 
     def customer_provider_taxation?
       @customer_provider_taxation ||= invoice.customer.tax_customer
--- a/app/services/customer_portal/customer_update_service.rb
+++ b/app/services/customer_portal/customer_update_service.rb
@@ -44,7 +44,6 @@ module CustomerPortal
         customer.reload
 
         tax_codes = []
-        # This service does not return a 'result' object but a string
         eu_tax_code_result = Customers::EuAutoTaxesService.call(
           customer:,
           new_record: false,
--- a/spec/models/customer_spec.rb
+++ b/spec/models/customer_spec.rb
@@ -997,4 +997,41 @@ RSpec.describe Customer, type: :model do
       end
     end
   end
+
+  describe "#vies_check_in_progress?" do
+    subject { customer.vies_check_in_progress? }
+
+    let(:organization) { create(:organization) }
+    let(:billing_entity) { create(:billing_entity, organization:, eu_tax_management:) }
+    let(:customer) { create(:customer, organization:, billing_entity:) }
+
+    context "when billing_entity has no EU tax management" do
+      let(:eu_tax_management) { false }
+
+      it "returns false" do
+        expect(subject).to be false
+      end
+    end
+
+    context "when billing_entity has EU tax management" do
+      let(:eu_tax_management) { true }
+
+      context "when VIES tax exist" do
+        before do
+          tax = create(:tax, organization:, code: "lago_eu_vat")
+          create(:customer_applied_tax, customer:, tax:)
+        end
+
+        it "returns false" do
+          expect(subject).to be false
+        end
+      end
+
+      context "when VIES tax do not exist" do
+        it "returns true" do
+          expect(subject).to be true
+        end
+      end
+    end
+  end
 end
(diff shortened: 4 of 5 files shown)


app/services/invoices/compute_taxes_and_totals_service.rb
```diff
@@ -19,6 +19,12 @@ def call
         after_commit { Invoices::ProviderTaxes::PullTaxesAndApplyJob.perform_later(invoice:) }
 
         return result.unknown_tax_failure!(code: "tax_error", message: "unknown taxes")
+      elsif customer.vies_check_in_progress?
+        invoice.status = "pending" if finalizing
+        invoice.tax_status = "failed"
```
lovrocolic, line 24: That would enable action in frontend to re-sync tax failed invoice with tax provider (Invoices::RetryService) - not sure we want it for vies? Also, how would the invoice  switch from failed status in this case?

app/services/invoices/retry_service.rb
```diff
@@ -12,11 +12,16 @@ def call
       return result.not_found_failure!(resource: "invoice") unless invoice
       return result.not_allowed_failure!(code: "invalid_status") unless invoice.failed?
 
-      invoice.status = "pending"
-      invoice.tax_status = "pending"
-      invoice.save!
-
-      Invoices::ProviderTaxes::PullTaxesAndApplyJob.perform_later(invoice:)
+      if invoice.customer.vies_check_finished?
```
floganz: I think that might work not as expected.
Let's take a case where EU tax management enabled but VIES check not finished. This `if` will return `false` and we'll try to handle it as Anrok/Avalara which is wrong.

Probably more robust will be check if it's Anrok/Avalara first (as we do in `compute_taxes_and_totals_service.rb`) and if not then check if VIES check is finished.

floganz: I'm also not sure that we need both predicates `vies_check_finished?` and `vies_check_in_progress?`.

It seems to be that `vies_check_in_progress?` should be more safe approach because we want to delay something only if tax management enabled and not finished. `vies_check_finished?` is trickier because for non EU tax management billing entities it will return false which might be not as clear.

ivannovosad: I think that keeping `vies_check_finished?` still makes sense because `!vies_check_in_progress?` doesn't have to mean that it has finished.

ivannovosad (approved):

app/services/invoices/retry_service.rb
```diff
@@ -12,11 +12,17 @@ def call
...
-      Invoices::ProviderTaxes::PullTaxesAndApplyJob.perform_later(invoice:)
+      # Anrok or avalara
+      if invoice.customer.tax_customer && invoice.should_apply_provider_tax?
+        invoice.status = "pending"
+        invoice.tax_status = "pending"
+        invoice.save!
+
+        Invoices::ProviderTaxes::PullTaxesAndApplyJob.perform_later(invoice:)
+      elsif invoice.customer.vies_check_finished?
+        Invoices::FinalizeAfterTaxesService.call(invoice:, provider_taxes: nil)
```
lovrocolic, line 23: I am just thinking what would be the flow if VIES taxes are not finished? From UI perspective there wouldn't be any notice about the result? Can we check that case?
Also in the UI for the failed invoice we display error details, not sure what would be error message in the UI for VIEW failed invoice

julienbourdeau: With @brunomiguelpinto we arrived at this solution 

```rb
      if invoice.customer.tax_customer && invoice.should_apply_provider_tax?
        invoice.status = "pending"
        invoice.tax_status = "pending"
        invoice.save!

        Invoices::ProviderTaxes::PullTaxesAndApplyJob.perform_later(invoice:)
      elsif customer.vies_check_pending?
        invoice.status = "pending"
        invoice.tax_status = "pending"
        invoice.save!

        Customers::RetryViesCheckJob.perform_later(customer: innvoice.customer)
      else
        Invoices::FinalizeAfterTaxesService.call(invoice:, provider_taxes: nil)
      end
```

And we make the `RetryViesCheckJob` retry all the customers pending invoice.

app/models/customer.rb
```diff
@@ -264,6 +264,18 @@ def tax_customer
...
+    if billing_entity.eu_tax_management?
+      taxes.where("code ILIKE ?", "lago_eu%").none?
+    else
+      false
+    end
+  end
+
+  def vies_check_finished?
+    billing_entity.eu_tax_management? && taxes.where("code ILIKE ?", "lago_eu%").any?
+  end
```
ancorcruz, line 277: I have concerns about these db queries... and its db performance impact as these are going to be run many times within the billing run.

ancorcruz: I'm evaluating having a flag in the customer that we set on and off to when VIES is completed instead of relying on tax codes in DB.

app/services/invoices/provider_taxes/pull_taxes_and_apply_service.rb
```diff
@@ -33,37 +33,10 @@ def call
...
-          SendWebhookJob.perform_later("invoice.created", invoice)
-          Utils::ActivityLog.produce(invoice, "invoice.created")
-          GeneratePdfAndNotifyJob.perform_later(invoice:, email: should_deliver_email?)
-          Integrations::Aggregator::Invoices::CreateJob.perform_later(invoice:) if invoice.should_sync_invoice?
-          Integrations::Aggregator::Invoices::Hubspot::CreateJob.perform_later(invoice:) if invoice.should_sync_hubspot_invoice?
-          Invoices::Payments::CreateService.call_async(invoice:)
-          Utils::SegmentTrack.invoice_created(invoice)
-        end
+        finalize_result = Invoices::FinalizeAfterTaxesService.call(invoice:, provider_taxes:)
+        finalize_result.raise_if_error!
```
ancorcruz, line 37:
```suggestion
        finalize_result = Invoices::FinalizeAfterTaxesService.call!(invoice:, provider_taxes:)
```

app/models/customer.rb
```diff
@@ -264,6 +264,18 @@ def tax_customer
     anrok_customer || avalara_customer
   end
 
+  def vies_check_in_progress?
+    if billing_entity.eu_tax_management?
+      taxes.where("code ILIKE ?", "lago_eu%").none?
```
ancorcruz, line 269: this will return false negative as we assign one of these taxes to the customer if the VIES check fails... I am wondering if this is also ok...

```
def process_not_vies_tax
      return "lago_eu_#{billing_country_code.downcase}_standard" if customer.country.blank?
      return "lago_eu_#{customer.country.downcase}_standard" if eu_countries_code.include?(customer.country.upcase)

      "lago_eu_tax_exempt"
    end
```

this means a customer belonging to a Billing Entity with EU tax management enabled that the VIES check failed... is going to return false as there will be matching tax codes with the previous expression.

PR #4083, opened 2025-07-31
Feat Allow multiple subscription_starting invoice_subscriptions when regenerating invoice 

brunomiguelpinto:
## Context

To support regenerating invoices for starting subscriptions, we need to allow multiple `invoice_subscriptions` records with the same `subscription_id` and `invoicing_reason = 'subscription_starting'`, as long as the invoice has been regenerated.

Currently, the uniqueness constraint (`index_unique_starting_subscription_invoice`) prevents this scenario, blocking valid regeneration flows for subscription starts.

## Description

These changes allow invoice regeneration logic to function properly for starting subscriptions, while preserving database integrity for non-regenerated records.

### DIFF
--- a/db/migrate/20250731144632_add_scoped_index_to_starting_invoice_subscription.rb
+++ b/db/migrate/20250731144632_add_scoped_index_to_starting_invoice_subscription.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddScopedIndexToStartingInvoiceSubscription < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :invoice_subscriptions,
+      [:subscription_id, :invoicing_reason],
+      unique: true,
+      name: :index_unique_starting_invoice_subscription,
+      where: "invoicing_reason = 'subscription_starting' AND regenerated_invoice_id IS NULL",
+      algorithm: :concurrently
+  end
+end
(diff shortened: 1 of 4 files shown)


db/migrate/20250731144632_add_scoped_index_to_starting_invoice_subscription.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddScopedIndexToStartingInvoiceSubscription < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :invoice_subscriptions,
```
annvelents, line 7: it's not really needed but as a safety measure we can add
`if_not_exists: true`, so this index will only be added if it does not exist yet

db/migrate/20250731145640_remove_old_starting_invoice_subscription_index.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class RemoveOldStartingInvoiceSubscriptionIndex < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def up
+    remove_index :invoice_subscriptions, name: :index_unique_starting_subscription_invoice
```
annvelents: and `if_not_exist: true` when removing the index

annvelents (approved):

PR #4084, opened 2025-07-31
feat: change void behaviour and wallet check

brunomiguelpinto:
## Context

When voiding an invoice, the system should recredit only active wallets. If the associated wallet has been deactivated, the recredit step should be skipped.

Additionally, when regenerating invoices, we no longer need to recredit the original credit note, because if we just void the credit note, we don't want to give anything back to the customer, and if we select to create credit note, the credit value can be included on the credit note

### DIFF
--- a/spec/services/invoices/void_service_spec.rb
+++ b/spec/services/invoices/void_service_spec.rb
@@ -124,6 +124,24 @@ RSpec.describe Invoices::VoidService, type: :service do
           end
         end
 
+        context "when the invoice has applied credits from inactive wallet" do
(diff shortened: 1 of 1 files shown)


app/services/invoices/void_service.rb
```diff
@@ -34,7 +34,6 @@ def call
 
         invoice.credits.each do |credit|
           AppliedCoupons::RecreditService.call!(credit:) if credit.applied_coupon_id.present?
-          CreditNotes::RecreditService.call!(credit:) if credit.credit_note_id.present?
```
annvelents, line 37: was not sure about this, but then read the PR description 😁 👍

PR #4085, opened 2025-07-31
Feat(fixed-chages-2): extract charge_models services

annvelents:
## Context

As Charge Model will be used by both: Charges and FixedCharges, on this step we want to extract ChargeModel services into their own namespace in services directory

## Description

- moved ChargeModel out of Charges repository
- updated all usages of Charges::ChargeModel

(note: the services only change is removing Charges namespace, we updated 11 files, removing `module Charges; end` - 2 lines in each file, so 22 lines difference comes from here)


app/services/charge_models/graduated_percentage_service.rb
```diff
@@ -0,0 +1,31 @@
...
+    protected
+
+    def ranges
+      properties["graduated_percentage_ranges"]&.map(&:with_indifferent_access)
+    end
+
+    def amount_details
+      {
+        graduated_percentage_ranges: ranges.each_with_object([]) do |range, amounts|
+          amounts << Charges::AmountDetails::RangeGraduatedPercentageService.call(range:, total_units: units)
```
ancorcruz, line 14: we should avoid referencing `Charges` namespace from this `ChargeModels` namespace as we want this to be generic instead of having charge models scoped to charges.

app/services/charge_models/graduated_service.rb
```diff
@@ -0,0 +1,31 @@
...
+    protected
+
+    def ranges
+      properties["graduated_ranges"]&.map(&:with_indifferent_access)
+    end
+
+    def amount_details
+      {
+        graduated_ranges: ranges.each_with_object([]) do |range, amounts|
+          amounts << Charges::AmountDetails::RangeGraduatedService.call(range:, total_units: units)
```
ancorcruz, line 14: we should avoid referencing `Charges` namespace from this `ChargeModels` namespace as we want this to be generic instead of having charge models scoped to charges.

ancorcruz (changes requested):
could we inject the service to call instead? We can merge it as it is and fix it in an upcoming PR.

ancorcruz (approved):

PR #4088, opened 2025-08-01
Feat(fixed charges 3): apply taxes services

annvelents:
## Roadmap Task

## Context

## Description

### DIFF
--- a/app/services/fixed_charges/apply_taxes_service.rb
+++ b/app/services/fixed_charges/apply_taxes_service.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module FixedCharges
+  class ApplyTaxesService < BaseService
+    Result = BaseResult[:applied_taxes]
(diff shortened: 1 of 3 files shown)


app/services/fixed_charges/apply_taxes_service.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module FixedCharges
+  class ApplyTaxesService < BaseService
```
rsempe, line 4: can't we use `Charges::ApplyTaxesService.call(charge: fixed_charge, tax_codes:)` instead? 🤔

ancorcruz: lets keep the scope separate by now.

We could consider later extracting an AppliedTaxes context with a service to apply a tax to whatever other resource following the same pattern.

PR #4091, opened 2025-08-04
fix(invoice-preview): Ensure that plan lookup uses only parent plans

floganz:
## Context

When we look for a plan by `code` it possible to find children plan, which share the same code as parent. For invoice preview that creates an issue as it could pick up child plan which is not visible anywhere in the UI and be confusing for users when they got different values from visible parent plans.

## Description

Ensure that we look for a plan only within parent plans when using invoice preview feature.


app/services/invoices/preview/build_subscription_service.rb
```diff
@@ -47,7 +47,7 @@ def billing_time
       end
 
       def plan
-        @plan ||= organization.plans.find_by(code: params[:plan_code])
```
lovrocolic, line 50: Are you sure here? What if we need to return the invoice preview for customer with the child plan? How to differentiate those two cases? I guess we should check if subscription is linked to parent or children plan...

floganz: This class builds a new subscription for case for existing or new customer.
I think in this case it should not be possible to use child plan.

lovrocolic: Oh okay makes sense

lovrocolic (approved):

PR #4092, opened 2025-08-04
fix(dunning-campaign): send webhook when dunning campaign ends and is not paid

lovrocolic:
## Context

Request from users:

> We need webhook message when a dunning campaign ends and isn't paid so we can auto-block users who don't pay at the end of their dunning campaign. 
> 
> If possible - it would be great to include the customer's overdue balance in the message

## Description

This PR adds `dunning_campaign.finished` webhook to match described use-case

### DIFF
--- a/app/serializers/v1/dunning_campaign_finished_serializer.rb
+++ b/app/serializers/v1/dunning_campaign_finished_serializer.rb
@@ -5,7 +5,7 @@ module V1
     def serialize
       {
         customer_external_id: model.external_id,
-        dunning_campaign_id:  options[:dunning_campaign_id],
+        dunning_campaign_id: options[:dunning_campaign_id],
         overdue_balance_cents: model.overdue_balance_cents,
         overdue_balance_currency: model.currency
       }
--- a/app/serializers/v1/dunning_campaign_finished_serializer.rb
+++ b/app/serializers/v1/dunning_campaign_finished_serializer.rb
@@ -5,7 +5,7 @@ module V1
     def serialize
       {
         customer_external_id: model.external_id,
-        dunning_campaign_id: options[:dunning_campaign_id],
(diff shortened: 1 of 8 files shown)


app/serializers/v1/dunning_campaign_finished_serializer.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module V1
+  class DunningCampaignFinishedSerializer < ModelSerializer
+    def serialize
+      {
+        customer_external_id: model.external_id,
+        dunning_campaign_id: options[:dunning_campaign_id],
```
ancorcruz: wouldn't be better to include in the webhook payload the dunning campaign code instead? Users won't be able to match the dunning campaign by its ID and there are no REST endpoints for the feature (as far as I remember).

app/services/dunning_campaigns/bulk_process_service.rb
```diff
@@ -67,6 +67,18 @@ def days_between_attempts_satisfied?
...
         Time.zone.now >= next_attempt_date
       end
+
+      def send_campaign_finished_webhook
+        return result unless customer.overdue_balance_cents.positive?
+
+        SendWebhookJob.perform_later(
+          "dunning_campaign.finished",
+          customer,
+          dunning_campaign_id: dunning_campaign.id
```
ancorcruz: I would use `code` instead of lago id.

app/serializers/v1/dunning_campaign_finished_serializer.rb
lovrocolic: agree, updated!

app/services/dunning_campaigns/bulk_process_service.rb
lovrocolic: updated

PR #4102, opened 2025-08-06
fix(activity-log): Do not include fees when there are too many of them

ivannovosad:
## Context

There is an issue when we're calling `ActivityLog` with a huge payload e.g. an invoice with a lot of fees.
In that case we're hitting the Kafka payload limit (1MB by default) and getting `WaterDrop::Errors::MessageInvalidError`

## Description

In order to fix this issue we skip fees to be serialized for invoices that have more than 25 fees.
So that `ActivityLog` would still log the invoice except the fees in this case.

### DIFF
--- a/app/services/utils/activity_log.rb
+++ b/app/services/utils/activity_log.rb
@@ -120,7 +120,17 @@ module Utils
       serializer = "V1::#{object.class.name}Serializer".constantize
       root_name = object.class.name.underscore.to_sym
 
-      serializer.new(object, root_name:, includes: SERIALIZED_INCLUDED_OBJECTS[root_name] || []).serialize
+      serializer.new(object, root_name:, includes: serializer_includes(root_name)).serialize
+    end
+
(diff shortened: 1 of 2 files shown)


app/services/utils/activity_log.rb
```diff
@@ -120,7 +120,17 @@ def object_serialized
...
       root_name = object.class.name.underscore.to_sym
 
-      serializer.new(object, root_name:, includes: SERIALIZED_INCLUDED_OBJECTS[root_name] || []).serialize
+      serializer.new(object, root_name:, includes: serializer_includes(root_name)).serialize
+    end
+
+    def serializer_includes(root_name)
+      return SERIALIZED_INCLUDED_OBJECTS[root_name] || [] if root_name != :invoice
+
+      if object.fees.count > 25
```
rsempe: Maybe just extracting `25` into a constant, that can be reused for other objects

rsempe (approved):

PR #4108, opened 2025-08-07
Fix get tax regenerate invoice

brunomiguelpinto:
## Context

When regenerating an invoice from a voided one, the process to get taxes if the external provider is done in async way, so if taxes are pending, we can save the invoice because it will be finalized by another job when we have the taxes back  

We want to allow the regeneration process to proceed and persist the invoice with a tax_status of "pending", so users or automated systems can later inspect or correct the issue.

### DIFF
--- a/app/services/invoices/regenerate_from_voided_service.rb
+++ b/app/services/invoices/regenerate_from_voided_service.rb
@@ -32,10 +32,13 @@ module Invoices
         Credits::ProgressiveBillingService.call!(invoice: regenerated_invoice)
         Credits::AppliedCouponsService.call!(invoice: regenerated_invoice) if should_create_coupon_credit?
         totals_result = Invoices::ComputeTaxesAndTotalsService.call(invoice: regenerated_invoice, finalizing: true)
+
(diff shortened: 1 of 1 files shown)


app/services/invoices/regenerate_from_voided_service.rb
```diff
@@ -31,7 +31,13 @@ def call
         # apply taxes credits and coupons
         Credits::ProgressiveBillingService.call!(invoice: regenerated_invoice)
         Credits::AppliedCouponsService.call!(invoice: regenerated_invoice) if should_create_coupon_credit?
-        Invoices::ComputeTaxesAndTotalsService.call!(invoice: regenerated_invoice, finalizing: true)
+        totals_result = Invoices::ComputeTaxesAndTotalsService.call(invoice: regenerated_invoice, finalizing: true)
+        if !totals_result.success? && regenerated_invoice.tax_status == "pending"
+          regenerated_invoice.save!
```
lovrocolic: I think you don't need save! here

lovrocolic (approved):

PR #4119, opened 2025-08-08
feat(one-off-invoice): Add billing period to one-off invoices

lovrocolic:
## Context

Currently fees on one-off invoices does not contain billing period

## Description

As part of the `quotes` feature, the decision is to add billing period on one-off invoice. This information would be consumed by analytics since it would allow splitting revenue across the billing interval


floganz (approved):
Looks good but please fix tests before merging it.

PR #4121, opened 2025-08-08
misc: Add BillingPerdiodBoundaries

vincent-pochet:
## Context

In many places of the application, we are defining a `boundaries` object, sometimes as an instance of a `Hash`, sometimes as an `OpenStruct`. The fields are not always the same and since the boundaries are injected in a lot of services, this causes some uncertainty when accessing the properties.

## Description

In an attempt to improve the readability and the reliability of this key part of the application, the goal of this PR is to introduce a new `BillingPeriodBoundaries` class and to use it everywhere in replacement of `Hash` and `OpenStruct`.

It should also improve the memory consumption as OpenStructs are known to cause memory issues in Ruby when YJIT is enabled

### DIFF
--- a/app/models/billing_period_boundaries.rb
+++ b/app/models/billing_period_boundaries.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class BillingPeriodBoundaries
+  attr_reader :from_datetime, :to_datetime, :charges_from_datetime, :charges_duration, :timestamp, :issuing_date
+  attr_accessor :charges_to_datetime
+
+  def self.from_fee(fee)
+    props = fee&.properties || {}
+
+    new(
+      from_datetime: props["from_datetime"],
+      to_datetime: props["to_datetime"],
+      charges_from_datetime: props["charges_from_datetime"],
+      charges_to_datetime: props["charges_to_datetime"],
+      charges_duration: props["charges_duration"],
+      timestamp: props["timestamp"],
+      issuing_date: props["issuing_date"]
+    )
+  end
+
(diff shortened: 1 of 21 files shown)


app/models/billing_period_boundaries.rb
```diff
@@ -0,0 +1,41 @@
...
+  def self.from_fee(fee)
+    new(
+      from_datetime: fee&.properties&.[]("from_datetime"),
+      to_datetime: fee&.properties&.[]("to_datetime"),
+      charges_from_datetime: fee&.properties&.[]("charges_from_datetime"),
+      charges_to_datetime: fee&.properties&.[]("charges_to_datetime"),
+      charges_duration: fee&.properties&.[]("charges_duration"),
+      timestamp: fee&.properties&.[]("timestamp"),
+      issuing_date: fee&.properties&.[]("issuing_date")
+    )
```
rsempe:
```suggestion
    props = fee&.properties || {}
 
    new(
      from_datetime: props["from_datetime"],
      to_datetime: props["to_datetime"],
      charges_from_datetime: props["charges_from_datetime"],
      charges_to_datetime: props["charges_to_datetime"],
      charges_duration: props["charges_duration"],
      timestamp: props["timestamp"],
      issuing_date: props["issuing_date"]
    )
```

app/models/billing_period_boundaries.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def to_h
+    h = {
+      "from_datetime" => from_datetime,
+      "to_datetime" => to_datetime,
+      "charges_from_datetime" => charges_from_datetime,
+      "charges_to_datetime" => charges_to_datetime,
+      "charges_duration" => charges_duration,
+      "timestamp" => timestamp
+    }.with_indifferent_access
```
rsempe, line 39: 😍

PR #4122, opened 2025-08-08
feat(customers-filters): Add active_subscriptions_from and active_sub…

mariohd:
## Problem
List the customers that has N active subscriptions.
## Fix
Adding the `active_subscriptions_from` and `active_subscriptions_to` to `CustomersResolver`.
This will be used to filter the active subscriptions for the customers.


app/queries/customers_query.rb
```diff
@@ -2,7 +2,14 @@
...
 class CustomersQuery < BaseQuery
   Result = BaseResult[:customers]
-  Filters = BaseFilters[:organization_id, :account_type, :billing_entity_ids, :with_deleted]
+  Filters = BaseFilters[
+    :organization_id,
+    :account_type,
+    :billing_entity_ids,
+    :with_deleted,
+    :active_subscriptions_from,
+    :active_subscriptions_to
```
rsempe: oh I was expecting to have a `date` filter here... and actually it's a `count` 😅 
can we change the naming here?

mariohd: yeah, np

rsempe: What about something like this instead?

```
:active_subscriptions_min  # Minimum number of active subscriptions
:active_subscriptions_max  # Maximum number of active subscriptions
```

Naming like `active_subscriptions_from` and `active_subscriptions_to` sounds like a date range, not a numeric count filter for me 😅

mariohd: done differently, more explicit but equaly good.  

`active_subscriptions_count_from` and `active_subscriptions_count_to`

PR #4123, opened 2025-08-08
Feature E-Invoicing

mariohd:
## Feature EInvoicing

FacturX and UBL for France.


rsempe (approved):

app/jobs/invoices/generate_files_and_notify_job.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Invoices
-  class GeneratePdfAndNotifyJob < ApplicationJob
+  class GenerateFilesAndNotifyJob < ApplicationJob
```
julienbourdeau, line 4: I think you can't rename jobs because there could be `GeneratePdfAndNotifyJob` jobs enqueued and after the deployed they'll all fail because the job class is not found.

Maybe we can simply create an alias with the old name but usually, we just keep the old class and make it call the new class... _forever_.

config/routes.rb
```diff
@@ -77,6 +77,7 @@
       resources :coupons, param: :code, code: /.*/
       resources :credit_notes, only: %i[create update show index] do
         post :download, on: :member
```
julienbourdeau, line 79: I like what you did with the download_pdf and the backward compat' on invoices, I'm in favor of doing it for credit notes too.

app/services/e_invoices/ubl/base_service.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module EInvoices
+  module Ubl
+    class BaseService < EInvoices::BaseService
```
julienbourdeau, line 5: I think this entire folder to be extracted from the `services` folder. 

I think we could consider this to be a serializer so maybe `serializers/xml/ubl`.
It could also be `models/ubl` or `support/ubl`.

But I think it's should definitely be outside `services`. It doesn't return a result (and I'm sure it shouldn't!).

app/services/e_invoices/invoices/ubl/builder.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+
+module EInvoices
+  module Invoices::Ubl
+    class Builder < EInvoices::Ubl::BaseService
+      include Invoices::Common
```
julienbourdeau, line 6: The same way, I think this shouldn't be a service.

julienbourdeau (commented):
I'm not entirely done but for me the main thing to adress is extracting xml builders outside the `services` directory and definitely not make it extend the BaseService. I don't think you ever used a "service feature" but I might have missed it.

app/jobs/invoices/generate_files_and_notify_job.rb
mariohd: Good point! I didn't saw that coming! 

I will preserve the old job calling the new job and we can destroy the old job in the next pull requests, when it's safe that we don't have scheduled jobs anymore.

config/routes.rb
mariohd: Awesome catch!

app/jobs/invoices/generate_files_and_notify_job.rb
julienbourdeau: > we can destroy the old job in the next pull requests
We can destroyed the old job when we're sure there are no on upgrading from 1.34.1 to X with jobs in their queue. So we'll need a bridge version and ideally a rake task to check the queues, to greenlight the upgrade to a version without the old job.

Or we keep the old job forever 🤪

PR #4124, opened 2025-08-11
feat(projected-usage): Create a new endpoint for the projected usage calculation

murparreira:
## Context

The `current_usage` endpoint is frequently called across the application. Adding the projected usage calculation was causing significant memory spikes due to the high request volume, impacting application performance.

By moving the expensive calculation to a dedicated endpoint, we ensure that fetching basic usage remains fast and stable, while still providing the projection data when it's explicitly needed. This resolves the memory issues and improves overall stability.

## Description

Creates a new endpoint, `GET /customers/:external_customer_id/projected_usage`, to handle the computationally intensive projected usage calculation.

Reverts the existing `/customers/:external_customer_id/current_usage` endpoint to its original, lightweight state by removing the projection logic.

### DIFF
--- a/app/graphql/types/customers/usage/projected.rb
+++ b/app/graphql/types/customers/usage/projected.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+module Types
+  module Customers
+    module Usage
+      class Projected < Types::BaseObject
+        graphql_name "CustomerProjectedUsage"
+
+        field :from_datetime, GraphQL::Types::ISO8601DateTime, null: false
+        field :to_datetime, GraphQL::Types::ISO8601DateTime, null: false
+
+        field :currency, Types::CurrencyEnum, null: false
+        field :issuing_date, GraphQL::Types::ISO8601Date, null: false
+
+        field :amount_cents, GraphQL::Types::BigInt, null: false
+        field :projected_amount_cents, GraphQL::Types::BigInt, null: false
+        field :taxes_amount_cents, GraphQL::Types::BigInt, null: false
+        field :total_amount_cents, GraphQL::Types::BigInt, null: false
+
+        field :charges_usage, [Types::Customers::Usage::ProjectedCharge], null: false
+
+        def charges_usage
+          object.fees.group_by(&:charge_id).values
+        end
+
+        def projected_amount_cents
+          fee_groups_by_charge.sum { |fee_group| projected_amount_for_fee_group(fee_group) }
+        end
+
+        private
+
+        def fee_groups_by_charge
+          object.fees.group_by(&:charge_id).values
+        end
+
+        def projected_amount_for_fee_group(fee_group)
+          charge = fee_group.first.charge
+
+          if charge.filters.any?
+            projected_amount_for_filtered_fees(fee_group)
+          elsif has_grouping?(fee_group)
+            projected_amount_for_grouped_fees(fee_group)
+          else
+            projected_amount_for_simple_fees(fee_group)
+          end
+        end
+
+        def has_grouping?(fee_group)
+          fee_group.any? { |f| f.grouped_by.present? }
+        end
+
+        def projected_amount_for_filtered_fees(fee_group)
+          defined_filter_fees = fee_group.select(&:charge_filter_id)
+          defined_filter_fees.sum { |fee| project_fees([fee]) }
+        end
+
+        def projected_amount_for_grouped_fees(fee_group)
+          groups = fee_group.group_by(&:grouped_by).values
+          groups.sum { |single_group_fees| project_fees(single_group_fees) }
+        end
+
+        def projected_amount_for_simple_fees(fee_group)
+          project_fees(fee_group)
+        end
+
+        def project_fees(fees)
+          ::Fees::ProjectionService.call(fees: fees).raise_if_error!.projected_amount_cents
+        end
+      end
+    end
+  end
+end
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -4,7 +4,7 @@ class ApiKey < ApplicationRecord
   include PaperTrailTraceable
 
   RESOURCES = %w[
-    activity_log add_on analytic api_log billable_metric coupon applied_coupon credit_note customer_usage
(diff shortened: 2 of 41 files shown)


app/services/charges/charge_model_factory.rb
```diff
@@ -2,7 +2,7 @@
 
 module Charges
   class ChargeModelFactory
-    def self.new_instance(charge:, aggregation_result:, properties:, period_ratio: nil, calculate_projected_usage: false)
+    def self.new_instance(charge:, aggregation_result:, properties:, period_ratio: 1.0, calculate_projected_usage: false)
```
vincent-pochet, line 5: I'm not sure to understand this change. Is there any risk of side effect?

app/serializers/v1/customers/projected_charge_usage_serializer.rb
```diff
@@ -0,0 +1,186 @@
...
+          result = ::Fees::ProjectionService.call(fees: [fee]).raise_if_error!
+          accumulate_projection(totals, result)
+        end
+      end
+
+      def calculate_grouped_projection(fees)
+        grouped_fees = fees.group_by(&:grouped_by).values
+
+        grouped_fees.reduce(initial_projection_values) do |totals, group_fee_list|
+          result = ::Fees::ProjectionService.call(fees: group_fee_list).raise_if_error!
```
vincent-pochet:
```suggestion
          result = ::Fees::ProjectionService.call!(fees: group_fee_list)
```

app/serializers/v1/customers/projected_charge_usage_serializer.rb
```diff
@@ -0,0 +1,186 @@
...
+
+      def charge_has_grouping?(fees)
+        fees.any? { |f| f.grouped_by.present? }
+      end
+
+      def calculate_filtered_projection(fees)
+        fees_with_defined_filters = fees.select(&:charge_filter_id)
+
+        fees_with_defined_filters.reduce(initial_projection_values) do |totals, fee|
+          result = ::Fees::ProjectionService.call(fees: [fee]).raise_if_error!
```
vincent-pochet:
```suggestion
          result = ::Fees::ProjectionService.call!(fees: [fee])
```

app/serializers/v1/customers/projected_charge_usage_serializer.rb
```diff
@@ -0,0 +1,186 @@
...
+        grouped_fees = fees.group_by(&:grouped_by).values
+
+        grouped_fees.reduce(initial_projection_values) do |totals, group_fee_list|
+          result = ::Fees::ProjectionService.call(fees: group_fee_list).raise_if_error!
+          accumulate_projection(totals, result)
+        end
+      end
+
+      def calculate_simple_projection(fees)
+        result = ::Fees::ProjectionService.call(fees: fees).raise_if_error!
```
vincent-pochet:
```suggestion
        result = ::Fees::ProjectionService.call!(fees: fees)
```

app/serializers/v1/customers/projected_charge_usage_serializer.rb
```diff
@@ -0,0 +1,186 @@
+# frozen_string_literal: true
+
+module V1
+  module Customers
+    class ProjectedChargeUsageSerializer < ModelSerializer
```
vincent-pochet, line 5: It feels weird to call services from inside a serializer as serialization is supposed to be the last step of the process and should not contains "business logic". It also means the logic is duplicated between the GraphQL type and the serializer.

Do you think it's possible to extract the logic in a dedicated service that will build a custom object that then could be serialized?

Also, it seems that a lot of grouping and looping logic is made. On large payload it could probably lead to some performance issues. Do you think we could change a bit the approach to reduce the number of `select`/`map`/`group_by`, like computing multiple projection in a single pass or by using more memoization?

app/graphql/types/customers/usage/projected.rb
```diff
@@ -0,0 +1,72 @@
...
+          groups = fee_group.group_by(&:grouped_by).values
+          groups.sum { |single_group_fees| project_fees(single_group_fees) }
+        end
+
+        def projected_amount_for_simple_fees(fee_group)
+          project_fees(fee_group)
+        end
+
+        def project_fees(fees)
+          ::Fees::ProjectionService.call(fees: fees).raise_if_error!.projected_amount_cents
```
vincent-pochet:
```suggestion
          ::Fees::ProjectionService.call!(fees: fees).projected_amount_cents
```

app/services/charges/charge_model_factory.rb
murparreira: Not really. This ratio is used only when calculating the projection, this change was to make some tests simpler.

app/serializers/v1/customers/projected_charge_usage_serializer.rb
murparreira: Yes, that was one of the main concerns when doing this, because the logic was already duplicated for the current usage on both parts (API and graphql).

> Also, it seems that a lot of grouping and looping logic is made. On large payload it could probably lead to some performance issues. Do you think we could change a bit the approach to reduce the number of select/map/group_by, like computing multiple projection in a single pass or by using more memoization?

This is the main problem for this feature, it is hard to gain a little bit of performance when we are creating fees on the fly in the `charge_service`, there is no possibility of trying to do optimizations on the DB or something like that. I can try to make some changes that you mentioned to see if we gain a little bit of performance.

app/models/api_key.rb
```diff
@@ -4,7 +4,7 @@ class ApiKey < ApplicationRecord
   include PaperTrailTraceable
 
   RESOURCES = %w[
-    activity_log add_on analytic api_log billable_metric coupon applied_coupon credit_note customer_usage
+    activity_log add_on analytic api_log billable_metric coupon applied_coupon credit_note customer_usage customer_projected_usage
```
julienbourdeau: Is there a need for a new scope? I'm in favor of relying on `customer_usage` for the projected usage.

murparreira: Removed on the new commit.

PR #4125, opened 2025-08-11
fix(alerting): When BillableMetric is deleted, delete associated alerts

julienbourdeau:
Alerts can be specific to a billable metric. When a billable metric is deleted, all associated alerts should also be deleted since they cannot be triggered anymore.

I changed the relationship to ensure that serializing a triggered alert correctly retrieve the alert and the billable metric even if it's deleted. That's the point of soft deletion, right?
It's possible that the `alert.triggered` webhook is sent after the alert was deleted.

Currently, if you delete an BM and you had alerts associated with it. the `/subscriptions/:external_id/alerts` will throw a 500 because we cannot serialize it.

### DIFF
--- a/spec/serializers/v1/usage_monitoring/triggered_alert_serializer_spec.rb
+++ b/spec/serializers/v1/usage_monitoring/triggered_alert_serializer_spec.rb
@@ -48,7 +48,7 @@ RSpec.describe ::V1::UsageMonitoring::TriggeredAlertSerializer do
     end
 
     context "when billable_metric and alert are deleted" do
-      it "retrieve te alert correctly" do
+      it "retrieves the alert correctly" do
         alert.billable_metric.discard!
         alert.discard!
         triggered_alert.reload


app/models/usage_monitoring/alert.rb
```diff
@@ -21,7 +21,7 @@ class Alert < ApplicationRecord
     default_scope -> { kept }
 
     belongs_to :organization
-    belongs_to :billable_metric, optional: true
+    belongs_to :billable_metric, -> { with_discarded }, optional: true
```
julienbourdeau, line 24: What do you think of loading discarded BM only for discarded alert?
```suggestion
    belongs_to :billable_metric, -> { with_discarded if discarded? }, optional: true
```

app/services/billable_metrics/destroy_service.rb
```diff
@@ -23,6 +23,7 @@ def call
         .where(billable_metrics: {id: metric.id}).distinct.pluck(:id)
 
       ActiveRecord::Base.transaction do
+        metric.alerts.discard_all!
```
vincent-pochet: I recently discovered the hard way that `discard_all` is loading every models in memory and perform validation/callbacks.
I guess in this case we want to use `update_all(deleted_at: Time.current)` instead

For reference: https://github.com/jhawthorn/discard?tab=readme-ov-file#performance-tuning

app/models/usage_monitoring/alert.rb
vincent-pochet: It makes senses I would say 🤔

app/services/billable_metrics/destroy_service.rb
julienbourdeau: Ooooh my god! 🤯

julienbourdeau: Interestingly enough, we discovered this in the exact same transaction, 2 lines below 😅  #3747

spec/models/billable_metric_spec.rb
```diff
@@ -7,20 +7,25 @@
 
   it_behaves_like "paper_trail traceable"
 
-  it { is_expected.to belong_to(:organization) }
+  describe "associations" do
+    it do
+      expect(subject).to belong_to(:organization)
+
```
ancorcruz, line 13: why this style change... I tend to prefer the shorthand version here with one test for each association.

spec/serializers/v1/usage_monitoring/triggered_alert_serializer_spec.rb
```diff
@@ -46,5 +46,17 @@
       payload = result["triggered_alert"]
       expect(payload["billable_metric_code"]).to eq alert.billable_metric.code
     end
+
+    context "when billable_metric and alert are deleted" do
+      it "retrieve te alert correctly" do
```
ancorcruz:
```suggestion
      it "retrieves the alert correctly" do
```

spec/requests/api/v1/billable_metrics_controller_spec.rb
```diff
@@ -198,6 +198,11 @@
       expect { subject }.to change(BillableMetric, :count).by(-1)
     end
 
+    it "deletes a alerts associated with billable_metric" do
```
ancorcruz:
```suggestion
    it "deletes alerts associated with billable_metric" do
```

ancorcruz (approved):
LGTM!

spec/models/billable_metric_spec.rb
julienbourdeau: I find the `it { ... }` much nice but I think it's best to group them to avoid before/after setup. This keeps the number of example smaller.

See https://github.com/getlago/lago-api/pull/3167

PR #4126, opened 2025-08-11
Feat(fixed_charges): Add GraphQL types, update Plan resolver and create mutation

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

  Update Plans::Create mutation and Plan resolver to include fixed charges data,
  also, adds all required graphql types.

### DIFF
--- a/app/graphql/types/plans/create_input.rb
+++ b/app/graphql/types/plans/create_input.rb
@@ -8,6 +8,7 @@ module Types
       argument :amount_cents, GraphQL::Types::BigInt, required: true
       argument :amount_currency, Types::CurrencyEnum
       argument :bill_charges_monthly, Boolean, required: false
+      argument :bill_fixed_charges_monthly, Boolean, required: false
       argument :code, String, required: true
       argument :description, String, required: false
       argument :interval, Types::Plans::IntervalEnum, required: true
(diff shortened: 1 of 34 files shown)


app/graphql/types/plans/create_input.rb
```diff
@@ -18,6 +19,7 @@ class CreateInput < Types::BaseInputObject
       argument :trial_period, Float, required: false
 
       argument :charges, [Types::Charges::Input]
+      argument :fixed_charges, [Types::FixedCharges::Input]
```
annvelents: shall this also be `required: false` for now?

annvelents (approved):
LGTM!
There is only a small comment about `plans/create_input`, in case FE doesn't send us anything as `fixed_charges`

ancorcruz:
> LGTM! There is only a small comment about `plans/create_input`, in case FE doesn't send us anything as `fixed_charges`

maybe we could make both `chages` and `fixed_charges` required false... it should have at least one of them, but should we keep `charges` mandatory?

app/graphql/types/plans/create_input.rb
ancorcruz: yes, otherwise frontend will break. Thanks for spotting this. I'm going to QA this on staging before merge and deploy.

PR #4131, opened 2025-08-12
feat(activity_logs): Let the organization query unlimited activity logs

mariohd:
## Problem
An organization can't query ActivityLogs older than 30 days. 
The number 30 days is hardcoded in the code. 

## Fix
Add an attribute on `Organization` called `activity_logs_period` and use this as the period allowed to query the Activity Logs. When `nil`, it should not limit query.


db/migrate/20250812132802_add_activity_logs_period_to_organizations.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddActivityLogsPeriodToOrganizations < ActiveRecord::Migration[8.0]
+  def change
+    add_column :organizations,
+      :activity_logs_period,
```
rsempe: what about `audit_logs_period` instead?

it's the same for api logs, no?

mariohd: that's why I always request your review! 

AWESOME CATCH AND SUGGESTION!!

PR #4139, opened 2025-08-13
[POC] Enqueue scheduled jobs on a different time window for organizations

ancorcruz:

Lago relies on a single `clock.rb` schedule file. Every “clock job” (e.g. `Clock::RefreshWalletsOngoingBalanceJob`) is executed at a fixed frequency for **all** organisations.

Some organizations occasionally need tighter or looser cadences for specific jobs (e.g. wallets refreshed every **1 minute** instead of every 5 minutes). Hardcoding these exceptions would leak clients data into the codebase and require deploys.

### General design

- Introduce a **Job Schedule Override** mechanism driven by a DB table.
- Add one **generic dispatcher job** that runs every minute and consults that table.
- Each target clock job receives an ***optional*** `organization` parameter and, when running in “default” mode, excludes any organisations that have an active override.
- Default behaviour (defined in `clock.rb`) is preserved and remains the fallback for every organisation without an override.

### High-level flow:

1. Clockwork keeps the existing entry (e.g. “every 5 minutes refresh wallets”).
2. A new Clockwork entry triggers `Clock::DispatchJobOverridesJob` **every 1 minute**.
3. The dispatcher queries job_schedule_overrides and enqueues target jobs **only** when the custom frequency elapses, passing the specific `organization`.
4. Inside each target job:
- If `organization` is present → process that org exclusively.
- If absent → exclude organisations that have overrides (**ensures no double processing**).


julienbourdeau:
This is awesome @ancorcruz I think it will be very useful. I wanted something like this when I added Alerting. I didn't figure a good way to do this without hardcoding organization ids 😅 I made [a "useless" service](https://github.com/getlago/lago-api/blob/a9eb33e/app/services/usage_monitoring/process_all_subscription_activities_service.rb#L8-L10).

I think your PR will be used in the future as a reference so I'd like to improve the `RefreshWalletsOngoingBalanceJob` refactor.

I think we should target 3 differents jobs:
1. one **clock** job that list all orgs that should run said job like `Clock::RefreshAllWalletsOngoingBalanceJob` (notice `All`)
2. one that enqueues jobs per org `Wallets::RefreshOrganizationWalletsOngoingBalanceJob`
3. one that does the actual job: `Wallets::RefreshOngoingBalanceJob` (untouched, just the job we have today)

Naming can be improved I guess but I hope you get what I mean. Instead of having a job called without our without organization_id arg, we'd get 2 jobs.

Another benefit is that the clock job is much lighter, it enqueues a few jobs (one per org without override) instead of thousands of jobs (one per wallet to refresh).

This is mainly inspired by what I did for Alerting:
1. `Clock::ProcessAllSubscriptionActivitiesJob` 
  ➡️ enqueues `UsageMonitoring::ProcessOrganizationSubscriptionActivitiesJob` for each eligible org. In my case
1. `UsageMonitoring::ProcessOrganizationSubscriptionActivitiesJob` 
  ➡️ enqueues `UsageMonitoring::ProcessSubscriptionActivityJob`
1. `UsageMonitoring::ProcessSubscriptionActivityJob` 
  ➡️ processes one SubscriptionActivity

I'll definitely upgrade the SubscriptionActivity jobs to handle overrides once this is merged!

ancorcruz:
@julienbourdeau sounds good to me. Thanks.

ancorcruz:
@julienbourdeau after taking another look I'm not sure what how your suggestion about having 3 jobs will play with the schedule override.

The job schedule override is on the clock job, hence, this one is the job that should skip the organizations with an override... and also should be the one we run when running the override as we save the job class name in the override.

I'd stick with the current solution for now.

I'm looking for approval to merge it as implementation is complete (for one clock job).

In the meantime I'm working in a solution to extract the logic from the clock wallets job to not duplicate the code everywhere.

PR #4141, opened 2025-08-13
feat(fixed_charges-6): add destroy service

annvelents:
## Context

As part of fixed_charges feature, we need a service to discard fixed charges

## Description

- Added `fixed_charges/destroy_service` to discard fixed charges


ancorcruz (approved):
LGTM!

PR #4142, opened 2025-08-13
feat(revenue-streams): add new data for the revenue streams reports

murparreira:
## Context

We need to display more information on the graph for the revenue streams. So the data pipeline created and populated new columns on the Analytical DB called:
```
contra_revenue_amount_cents
credit_notes_credits_amount_cents
free_credits_amount_cents
prepaid_credits_amount_cents
progressive_billing_credit_amount_cents
```
The backend should make the graphql query send these new columns to the FE.

## Description

Just created the fields on the main object type.


iavorbotev (approved):
LGTM!

PR #4148, opened 2025-08-14
Feat(fixed_charges): Add services and jobs to manage Child Fixed Charges

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

* Add `FixedCharges::CreateChildrenService` calls `FixedCharges::CreateService` on every plan’s children.
* Add `FixedCharges::UpdateChildrenService` calls `FixedCharges::UpdateService` on every fixed charge’s children with cascade options. 
* Add `FixedCharges::DestroyChildrenService` calls `FixedCharges::DestroyService` on every fixed charge’s children. 

* Add `FixedCharges::CreateChildrenJob`
* Add `FixedCharges::UpdateChildrenJob`
* Add `FixedCharges::DestroyChildrenJob` calls `FixedCharges::DestroyChildrenService`

* Add `FixedCharges::CreateChildrenBatchJob` calls `FixedCharges::CreateChildrenService`
* Add `FixedCharges::UpdateChildrenBatchJob`  calls `FixedCharges::UpdateChildrenService`

### DIFF
--- a/spec/jobs/fixed_charges/update_children_batch_job_spec.rb
+++ b/spec/jobs/fixed_charges/update_children_batch_job_spec.rb
@@ -15,7 +15,8 @@ RSpec.describe FixedCharges::UpdateChildrenBatchJob, type: :job do
   let(:params) do
     {
       charge_model: "standard",
-      properties: {amount: "200"}
+      properties: {amount: "200"},
+      units: 1
     }
   end
 
--- a/spec/services/fixed_charges/update_children_service_spec.rb
+++ b/spec/services/fixed_charges/update_children_service_spec.rb
@@ -15,7 +15,8 @@ RSpec.describe FixedCharges::UpdateChildrenService, type: :service do
   let(:params) do
     {
       charge_model: "standard",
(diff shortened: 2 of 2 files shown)


app/services/fixed_charges/update_service.rb
```diff
@@ -19,7 +19,7 @@ def call
         # Note: when updating a fixed_charge, we can't update pay_in_advance and prorated,
         fixed_charge.charge_model = params[:charge_model] unless plan.attached_to_subscriptions?
         fixed_charge.invoice_display_name = params[:invoice_display_name] unless cascade
-        fixed_charge.units = params[:units]
+        fixed_charge.units = params[:units] if params.key?(:units)
```
annvelents: Ah, I had only one comment here:
right now for charges we do not have this check is params contain key, so we always assume that if fixed_charge is updated, the whole payload will be sent. 
If we're adding a possibility to send not the whole payload, it might make sense to do the same check for `charge_model`, `invoice_display_name`...

PR #4149, opened 2025-08-14
Feat(fixed_charges-6): update plan with fixed-charge specific values

annvelents:
## Context

There are several steps we need to do to update plans:
- update plan itself
- handle fixed_charges update
- handle fixed_charges addition
- handle fixed_charges deletion

this PR handles first part: updating plan itself + validating params for fixed_charges

## Description
- Added plan update with fixed_charges_bill_monthly
- Validate params for fixed charges

### DIFF
--- a/app/services/plans/update_service.rb
+++ b/app/services/plans/update_service.rb
@@ -36,7 +36,7 @@ module Plans
         plan.bill_fixed_charges_monthly = bill_fixed_charges_monthly?
       end
 
-       chargeables_validation_result = Plans::ChargeablesValidationService.call(
+      chargeables_validation_result = Plans::ChargeablesValidationService.call(
         organization: plan.organization,
         charges: params[:charges],
         fixed_charges: params[:fixed_charges]
--- a/spec/services/plans/update_service_spec.rb
+++ b/spec/services/plans/update_service_spec.rb
@@ -1183,7 +1183,7 @@ RSpec.describe Plans::UpdateService, type: :service do
       end
     end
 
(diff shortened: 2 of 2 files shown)


app/services/plans/update_service.rb
```diff
@@ -89,6 +91,12 @@ def bill_charges_monthly?
       params[:bill_charges_monthly] || false
     end
 
+    def bill_fixed_charges_monthly?
+      return unless params[:interval]&.to_sym == :yearly
+
+      params[:bill_fixed_charges_monthly] || false
```
ancorcruz, line 97:
```suggestion
      !!params[:bill_fixed_charges_monthly]
```

its more idiomatic perhaps.

ancorcruz (approved):
LGTM!

app/services/plans/update_service.rb
annvelents: hmm
I was checking and if someone sends us `false` as `"false"`(`{bill_fixed_charges_monthly: "false"}`), current implementation, that is the same as `bill_charges_monthly` will return `"false"`, that is then handled by ActiveRecord as falsey value, so in the record the value will be saved as `false`:
```
plan.bill_charges_monthly = "false"
plan.bill_charges_monthly
# => false
```
but if we add double negation, it will be considered true, as not-empty string it thruthy, so I'm thinking for now to keep the logic similar to `bill_charges_montly` 🙈

ancorcruz: okay...

PR #4151, opened 2025-08-14
feat(models): Introduce SoftDeletable concern

julienbourdeau:
A few weeks back, I introduced a new way to spec helper thing to ensure:
- use the `deleted_at` column name
- use default scope kept

I think it was very useful but having a proper concern to use is probably better!

The goal is to add either override `discard_all!` or create a new helper to [avoid this mistake again](https://github.com/getlago/lago-api/pull/3747). 

- [ ] Ensure all models already had `kept` default scope
- [ ] Ensure all model use spec hleper

### DIFF
--- a/app/models/charge_filter.rb
+++ b/app/models/charge_filter.rb
@@ -2,9 +2,7 @@
 
 class ChargeFilter < ApplicationRecord
   include PaperTrailTraceable
-  include Discard::Model
   include ChargePropertiesValidation
-  self.discard_column = :deleted_at
 
   belongs_to :charge, -> { with_discarded }, touch: true
   belongs_to :organization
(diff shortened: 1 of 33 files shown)


app/models/charge_filter.rb
```diff
@@ -2,9 +2,7 @@
 
 class ChargeFilter < ApplicationRecord
   include PaperTrailTraceable
-  include Discard::Model
```
julienbourdeau, line 5: oops

app/models/charge_filter_value.rb
```diff
@@ -2,8 +2,7 @@
 
 class ChargeFilterValue < ApplicationRecord
   include PaperTrailTraceable
-  include Discard::Model
-  self.discard_column = :deleted_at
+  include SoftDeletable
```
julienbourdeau, line 5: default scope

PR #4153, opened 2025-08-14
fix(entitlements): Fix views + graphql

julienbourdeau:
- [x] Better config typing
- [x] Retrieve feature by code or ID
- [x] fix select value validation 🤦‍♂️ 
- [x] Sort feature and privileges by entitlement and entitlement_value respectively
- [x] Patch plan override to return and accept **subscription** entitlements


app/graphql/types/plans/object.rb
```diff
@@ -49,6 +49,21 @@ class Object < Types::BaseObject
...
+      def entitlements
+        # NOTE: Entitlements don't rely on the overridden plan. Unfortunately, the frontend reuse the same
+        #       component for both plan edit and subscription plan override edit.
+        #       In order to keep work minimal in the frontend for now, return the subscription entitlements instead of
+        #       in the plan entitlements when resolving a plan override. A plan override is a plan with a parent_id.
+        if object.parent_id.present?
+          subscription = object.organization.subscriptions.find_by(plan: object)
+          pps subscription
+          entitlements = ::Entitlement::SubscriptionEntitlement.for_subscription(subscription)
+          ::V1::Entitlement::SubscriptionEntitlementsCollectionSerializer.new(entitlements).serialize_models
```
vincent-pochet: It feels very weird to rely on a serializer in a graphQL type 🤔 

Also, we are doing a lot of DB queries here, do you think we could preload some objects?

app/graphql/types/entitlement/plan_entitlement_object.rb
```diff
@@ -7,20 +7,27 @@ class PlanEntitlementObject < Types::BaseObject
...
-      field :privileges, [PlanEntitlementPrivilegeObject], null: false, method: :values
+      field :privileges, [PlanEntitlementPrivilegeObject], null: false
+
+      # NOTE: For regular plans, the attribute are read on the object
+      #       For plan override, we're reading the *subscrition entitlements*
+      #       See `Types::Plans::Object.entitlements` note for more details.
 
       def code
-        object.feature.code
+        object.is_a?(Hash) ? object.fetch(:code) : object.feature.code
```
jdenquin: this make me unconfortable

app/graphql/types/plans/object.rb
jdenquin: I think this is the first time we use a serializer on GraphQL, can I ask why?

julienbourdeau: Here we're making a single query to the `entitlement_subscription_entitlements_view`.
The data returned is "flat" (one feature with 2 privileges returns to rows) and the serializer is responsible for reshaping the data (grouping per feature).

I understand the surprise of seeing a serializer here, I'm happy to move this somewhere else (model or service).

julienbourdeau: The data returned is "flat" (one feature with 2 privileges returns to rows) and the serializer is responsible for reshaping the data (grouping per feature).

I understand the surprise of seeing a serializer here, I'm happy to move this somewhere else (model or service).

app/graphql/types/entitlement/plan_entitlement_object.rb
julienbourdeau: Yeah, I don't like it much either but this is the best way to keep the frontend work minimal. The subscriptions entitlements are intended to live in the subscription but to reuse frontend component and not refactor, we return subscription entitlements in the plan override.

The subscription entitlements are return from a view a turned into a hash via the serializer. The main goal is that subscription entitlements are one DB call.

app/graphql/types/plans/object.rb
vincent-pochet: Maybe it can be an intermediary model (a basic ruby Data class), what do you think?

app/graphql/types/entitlement/plan_entitlement_object.rb
julienbourdeau: We're changing the form design so I was able to remove this 🙌

julienbourdeau:
closed in favor if https://github.com/getlago/lago-api/pull/4176

PR #4155, opened 2025-08-14
misc: Remove OpenStruct from event validation errors

vincent-pochet:
## Description

This PR removes more `OpenStruct`s from the events validation error webhook


app/services/webhooks/events/validation_errors_service.rb
```diff
@@ -11,7 +11,7 @@ def current_organization
 
       def object_serializer
         ::V1::EventsValidationErrorsSerializer.new(
-          OpenStruct.new(options[:errors]),
+          (options[:errors] || {}).with_indifferent_access,
```
ancorcruz: is  `with_indifferent_access` actually needed?

PR #4157, opened 2025-08-15
feat(projected-usage): Add projected usage as premium integration

murparreira:
## Context

This feature will be hidden behind a premium integration to test load on small customers first.

## Description

Just added the `projected_usage` string to the premium integration list.

### DIFF
--- a/app/controllers/api/v1/customers/projected_usage_controller.rb
+++ b/app/controllers/api/v1/customers/projected_usage_controller.rb
@@ -32,6 +32,14 @@ module Api
         def resource_name
           "customer_usage"
         end
+
+        def authorize
+          super
+
+          return if current_organization.projected_usage_enabled?
+
+          forbidden_error(code: "projected_usage_not_enabled")
+        end
       end
     end
   end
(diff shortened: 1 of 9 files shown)


app/controllers/api/v1/customers/projected_usage_controller.rb
```diff
@@ -32,6 +32,14 @@ def current
...
           "customer_usage"
         end
+
+        def authorize
+          super
+
+          return if current_organization.projected_usage_enabled?
+
+          forbidden_error(code: "projected_usage_not_enabled")
+        end
```
mariohd, line 42: I would include this into `::Invoices::CustomerUsageService#call`. 

Why? 
because whenever someone calls this service any other place outside of controllers with variable `calculate_projected_usage = true`, this check would be triggered and prevent forbidden calls.

Would add this line [right here](https://github.com/getlago/lago-api/blob/2d683f8fa983483af582435206891dc7cf0a9af6/app/services/invoices/customer_usage_service.rb#L48)
```
return result.forbidden_error!(code: "projected_usage_not_enabled") if calculate_projected_usage && !organization.projected_usage_enabled?
```

murparreira: Marked for future refactoring :)

PR #4163, opened 2025-08-19
fix(events): Deduplicate clickhouse events in the Graphql resolver

vincent-pochet:
## Context

Idem-potency logic is different between Postgres and Clickhouse. This creates some confusion on the customer side.

The main problem is that to render the list of event in the front-end, when an organization is configured to use Clickhouse we relies on `EventRaw` which does not do any deduplication. It means that if a customer send two events with the same transaction_id, external_subsciption_id, code and timestamp, all events will appears in the event view, while only one will be considered when doing the aggregation (which relies on `EventEnriched` and that already has a deduplication logic)

## Description
This PR makes sure that the events rendered in the GraphQL API are not showing duplicated events, by applying the same deduplication rule than the `EventEnriched` table

### DIFF
--- a/app/graphql/resolvers/events_resolver.rb
+++ b/app/graphql/resolvers/events_resolver.rb
@@ -18,6 +18,7 @@ module Resolvers
       if current_organization.clickhouse_events_store?
         Clickhouse::EventsRaw.where(organization_id: current_organization.id)
           .order(ingested_at: :desc)
+          .limit_by(1, "transaction_id, timestamp, external_subscription_id, code")
           .page(page)
           .per((limit >= MAX_LIMIT) ? MAX_LIMIT : limit)
       else
--- a/spec/graphql/resolvers/events_resolver_spec.rb
+++ b/spec/graphql/resolvers/events_resolver_spec.rb
@@ -2,7 +2,7 @@
 
 require "rails_helper"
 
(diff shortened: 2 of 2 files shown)


app/graphql/resolvers/events_resolver.rb
```diff
@@ -18,6 +18,7 @@ def resolve(page: nil, limit: nil)
       if current_organization.clickhouse_events_store?
         Clickhouse::EventsRaw.where(organization_id: current_organization.id)
           .order(ingested_at: :desc)
+          .limit_by(1, "transaction_id, timestamp, external_subscription_id, code")
```
rsempe, line 21: i'm wondering if we can't take `events_enriched` instead as it should be "quick"? 🤔 

is it not a problem to not have `customer` here for instance?

rsempe: oh `delegate :customer, :customer_id, to: :subscription` actually it works with an event raw

rsempe (approved):

app/graphql/resolvers/events_resolver.rb
vincent-pochet: We could probably use events enriched but it would mean that nothing will be visible in the event list until the event is processed, and we would still need to deduplicate to handle the merging delay that can happen on clickhouse.

PR #4165, opened 2025-08-19
feat(semiannual): Prepend DateAndTime::HalfYearCalculations to DateAndTime::Calculations

ivannovosad:
## Context

A prerequisite to semiannual plans: We need to calculate start and end of half-year for some further calculations.

## Description

Prepend `DateAndTime::HalfYearCalculations` to `DateAndTime::Calculations`

### DIFF
--- a/lib/date_and_time/half_year_calculations.rb
+++ b/lib/date_and_time/half_year_calculations.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module DateAndTime
+  module HalfYearCalculations
+    def beginning_of_half_year
+      # If month is 1–6 → snap to 1 (January)
+      # If month is 7–12 → snap to 7 (July)
+      first_half_year_month = (month <= 6) ? 1 : 7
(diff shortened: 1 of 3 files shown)


lib/date_and_time/half_year_calculations.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module DateAndTime
```
lovrocolic, line 3: Have you considered using existing services/utils/ folder for that?

lovrocolic (approved):

lib/date_and_time/half_year_calculations.rb
ivannovosad: This module is different that those classes in utils module, because it is prepended to activesupport's `DateAndTime::Calculations`.

ivannovosad: You cannot call it directly. This is how it's supposed to be used: `Date.new(2025, 7, 1).beginning_of_half_year `

PR #4166, opened 2025-08-19
feat(fixed_charges-8): process fixed_charges from plans controller

annvelents:
## Context

With this PR we're allowing users to send us fixed_charges params when creating and updating plan, so we'll also create, update, delete fixed charges for the plan

## Description

- updated plans controller to receive fixed_charges specific params and pass them to plans services

### DIFF
--- a/spec/requests/api/v1/plans_controller_spec.rb
+++ b/spec/requests/api/v1/plans_controller_spec.rb
@@ -96,7 +96,7 @@ RSpec.describe Api::V1::PlansController, type: :request do
         expect(json[:plan][:invoice_display_name]).to eq(create_params[:invoice_display_name])
         expect(json[:plan][:created_at]).to be_present
         expect(json[:plan][:charges].first[:lago_id]).to be_present
-        # expect(json[:plan][:fixed_charges].first[:lago_id]).to be_present
+        expect(json[:plan][:fixed_charges].first[:lago_id]).to be_present
       end
 
       context "when license is not premium" do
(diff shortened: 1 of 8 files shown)


spec/requests/api/v1/plans_controller_spec.rb
```diff
@@ -81,6 +96,7 @@
         expect(json[:plan][:invoice_display_name]).to eq(create_params[:invoice_display_name])
         expect(json[:plan][:created_at]).to be_present
         expect(json[:plan][:charges].first[:lago_id]).to be_present
+        # expect(json[:plan][:fixed_charges].first[:lago_id]).to be_present
```
floganz: Is this expected?

spec/requests/api/v1/plans_controller_spec.rb
```diff
@@ -222,6 +238,57 @@
...
+        it "creates a plan" do
+          subject
+
+          expect(response).to have_http_status(:success)
+
+          expect(json[:plan][:lago_id]).to be_present
+          expect(json[:plan][:code]).to eq(create_params[:code])
+          expect(json[:plan][:name]).to eq(create_params[:name])
+          expect(json[:plan][:created_at]).to be_present
+          # expect(json[:plan][:fixed_charges].first[:lago_id]).to be_present
```
floganz: Can you check this and other commented expectations?
If we don't need them – let's remove them.

annvelents: I've accidentally added them in this PR to realize that fixed_charge_serializer is added at the next PR, so there I've uncommented the tests 🙈 
https://github.com/getlago/lago-api/pull/4170/files#diff-c2b30b53f949ceb6886373f744ead96986e4aca22a24511130d653ee12687ffaL99
but because I branched one on another, removing these lines will create a mess in rebase 😞  so I was thinking after approve of this PR, get approve of second PR, merge second into first, and then first into main

floganz: That works.

floganz (approved):

PR #4172, opened 2025-08-20
feat(wallet-transactions): Add priority

floganz:
## Context

Part of rework wallet calculation feature, as part of that feature we want have control over inbound transactions "consume" ordering.

## Description

Add `priority` attribute.
Adjust services and GQL to accept new column.

### DIFF
--- a/app/services/wallet_transactions/create_from_params_service.rb
+++ b/app/services/wallet_transactions/create_from_params_service.rb
@@ -25,6 +25,7 @@ module WalletTransactions
       wallet_transactions = []
       @source = params[:source] || :manual
       @metadata = params[:metadata] || []
+      @priority = params[:priority] || 50
       invoice_requires_successful_payment = if params.key?(:invoice_requires_successful_payment)
         ActiveModel::Type::Boolean.new.cast(params[:invoice_requires_successful_payment])
       else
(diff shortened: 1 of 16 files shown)


murparreira (approved):

app/services/wallet_transactions/create_from_params_service.rb
```diff
@@ -25,6 +25,7 @@ def call
       wallet_transactions = []
       @source = params[:source] || :manual
       @metadata = params[:metadata] || []
+      @priority = params[:priority] || 50
```
julienbourdeau, line 28: How do you feel about removing the default 50 here? I like relying on the db default.
I'm guessing it requires some work in other services like WalletTransactions::VoidService to "ignore" priority if nil, and that's why you added it here?

floganz: The issue with suggested change that `priority` value then passed directly to `WalletTransactions::CreateService`. When we pass value explicitly we override DB default.

We either need to keep this way of doing or change current service to use something like `params.slice`. `slice` do not return `nil` value for keys that are not in a hash compared to `params[:key]`.

I don't mind refactor it but better to do it separate branch to reduce the PR size.

PR #4175, opened 2025-08-20
Feat(fixed_charges): add fixed_charges_units_override service

annvelents:
## Context

we're implementing a mechanism to override units of a fixed charge for a subscription without creating a plan override. It's done through `SubscriptionFixedChargeUnitsOverride`. This PR creates service that'll be used to create these records

## Description

- Added `SubscriptionFixedChargeUnitsOverride`

### DIFF
--- a/app/models/subscription_fixed_charge_units_override.rb
+++ b/app/models/subscription_fixed_charge_units_override.rb
@@ -11,7 +11,7 @@ class SubscriptionFixedChargeUnitsOverride < ApplicationRecord
   belongs_to :subscription
   belongs_to :fixed_charge
 
-  validates :units, presence: true, numericality: {greater_than: 0}
+  validates :units, presence: true, numericality: {greater_than_or_equal_to: 0}
 end
 
 # == Schema Information
--- a/spec/models/subscription_fixed_charge_units_override_spec.rb
+++ b/spec/models/subscription_fixed_charge_units_override_spec.rb
@@ -16,6 +16,6 @@ RSpec.describe SubscriptionFixedChargeUnitsOverride, type: :model do
 
   describe "validations" do
     it { is_expected.to validate_presence_of(:units) }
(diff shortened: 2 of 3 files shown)


app/models/subscription_fixed_charge_units_override.rb
```diff
@@ -11,7 +11,7 @@ class SubscriptionFixedChargeUnitsOverride < ApplicationRecord
   belongs_to :subscription
   belongs_to :fixed_charge
 
-  validates :units, presence: true, numericality: {greater_than_or_equal_to: 0}
+  validates :units, presence: true, numericality: {greater_than: 0}
```
ancorcruz: why is zero not an option now.

Allowing units to be zero it was covering a customer use case where the fixed charge is defined without units, then changed to a value (for example in the first invoice) with the goal for example to charge in the first subscription invoice a one off fee.

spec/models/subscription_fixed_charge_units_override_spec.rb
```diff
@@ -16,6 +16,6 @@
 
   describe "validations" do
     it { is_expected.to validate_presence_of(:units) }
-    it { is_expected.to validate_numericality_of(:units).is_greater_than_or_equal_to(0) }
+    it { is_expected.to validate_numericality_of(:units).is_greater_than(0) }
```
ancorcruz: same applies here.

ancorcruz (changes requested):
Looks good to me.

My only concern is about the change in the units validation, not allowing zero units will block some use cases.

app/models/subscription_fixed_charge_units_override.rb
annvelents: ah, when I was writing tests, I didn't think about this case and exactly was questioning why would we create a fixed charge with 0 units🙈 
thank you, updating 🙌

ancorcruz (approved):
LGTM, thanks!

PR #4176, opened 2025-08-20
feat(entitlements): Introduce SubscriptionEntitlementQuery and remove view

julienbourdeau:
### New `SubscriptionEntitlementQuery`

This PR is a significant refactor of the SubscriptionEntitlement query. The data model did not change, it's stored the same way but I have redone the way we query the data subscriptions.
I removed the database view and replaced it with 2 queries, which hydrates ActiveModel models (not ActiveRecord models). These models are then serialized like any other model.

**100x improvement**: fixed all the bugs at once and made the thing 10x more maintainable.

### New `SubscriptionEntitlementCreateOrUpdateService` service

The frontend will allow for each subscription feature to be updated independently. The new services is tested through the GQP mutations but it doesn't have proper tests yet. This is coming in a separate PR where I will refactor services and more importantly: Add the abilitiy to remove privilege from plan in subscription.

> [!IMPORTANT]  
> Remember the feature is not released yet.

## Upcoming next

- Merge API endpoints to "remove/restore" feature in subscription to make it automatic
- Remove refactor services and add tests
- Add `entitlement_subscription_feature_removals.entitlement_privilege_id` to be able to remove privilege from subscription

### DIFF
--- a/app/models/entitlement/subscription_entitlement.rb
+++ b/app/models/entitlement/subscription_entitlement.rb
@@ -27,7 +27,5 @@ module Entitlement
         }
       )
     end
-
-    private
   end
 end
--- a/app/services/entitlement/subscription_entitlement_create_or_update_service.rb
+++ b/app/services/entitlement/subscription_entitlement_create_or_update_service.rb
@@ -139,13 +139,5 @@ module Entitlement
         [v.privilege.code, v.value]
       end.to_h.eql? privilege_params
     end
-
-    # def subscription_plan_entitlements_as_params
-    #   @subscription_plan_entitlements_as_params ||= (subscription.plan.parent || subscription.plan)
-    #     .entitlements
-    #     .includes(:feature, values: :privilege)
-    #     .map { |entitlement| [entitlement.feature.code, entitlement.values.map { |v| [v.privilege.code, v.value] }.to_h] }
-    #     .to_h
-    # end
   end
 end


app/graphql/resolvers/entitlement/subscription_entitlements_resolver.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      REQUIRED_PERMISSION = "subscriptions:view"
+
+      description "Query entitlements of a subscriptions"
+
+      argument :subscription_id, ID, required: true
+
+      type Types::Entitlement::SubscriptionEntitlementObject.collection_type, null: false
+
+      def resolve(**args)
```
vincent-pochet: Since the resolvers only has one single arguments, you should be able to turn it to:
```suggestion
      def resolve(subscription_id:)
```

app/queries/entitlement/subscription_entitlement_query.rb
```diff
@@ -0,0 +1,134 @@
...
+        SubscriptionEntitlement.new(row)
+      end
+
+      plan_entitlement_ids = features.map(&:plan_entitlement_id).compact.uniq
+      sub_entitlement_ids = features.map(&:sub_entitlement_id).compact.uniq
+
+      privileges_result = ActiveRecord::Base.connection.exec_query(
+        privilege_sql,
+        "subscription_entitlement_privileges",
+        ["{#{plan_entitlement_ids.join(",")}}", "{#{sub_entitlement_ids.join(",")}}"]
```
vincent-pochet: It will require some sanitization I guess

julienbourdeau: In this case, it's UUID coming directly from the DB (uuid columns). I quand still add quoting just in case this ever changes

app/models/entitlement/subscription_entitlement.rb
```diff
@@ -1,47 +1,33 @@
...
+      SubscriptionEntitlementQuery.call(
+        organization: subscription.organization,
+        filters: {
+          subscription_id: subscription.id,
+          plan_id: subscription.plan.parent_id || subscription.plan.id
+        }
+      )
     end
+
+    private
```
vincent-pochet: I guess you can get rid of this little one :)

app/queries/entitlement/subscription_entitlement_query.rb
```diff
@@ -0,0 +1,138 @@
...
+    def call
+      features_result = ActiveRecord::Base.connection.exec_query(
+        feature_sql,
+        "subscription_entitlement_features",
+        [filters.plan_id, filters.subscription_id]
+      )
+
+      features = features_result.map do |row|
+        SubscriptionEntitlement.new(row)
+      end
```
vincent-pochet, line 17: You might be able to use `SubscriptionEntitlement.find_by_sql(...` instead if this

app/services/entitlement/subscription_entitlement_create_or_update_service.rb
```diff
@@ -0,0 +1,151 @@
...
+    def call
+      return result.forbidden_failure! unless License.premium?
+      return result.not_found_failure!(resource: "subscription") unless subscription
+      return result.not_found_failure!(resource: "feature") unless feature
+
+      ActiveRecord::Base.transaction do
+        process_single_entitlement
+      end
+
+      result.entitlement = SubscriptionEntitlement.for_subscription(subscription).find { it.code == feature_code }
```
vincent-pochet, line 30: Do you think that the `code` could be an optional argument for the query, rather than re-loading all entitlement in memory and filtering them?

app/services/entitlement/subscription_entitlement_create_or_update_service.rb
```diff
@@ -0,0 +1,151 @@
...
+
+    def process_single_entitlement
+      plan_entitlement = plan.entitlements.includes(values: :privilege).find_by(feature: feature)
+      sub_entitlement = subscription.entitlements.includes(values: :privilege).find_by(feature: feature)
+      # TODO: add .or.where(privilege: feature.privileges)
+      removals = SubscriptionFeatureRemoval.where(subscription: subscription, feature: feature).to_a
+
+      if plan_entitlement.nil? && sub_entitlement.nil?
+        feature_removal = removals.find { it.entitlement_feature_id == feature.id }
+        feature_removal&.discard!
```
vincent-pochet, line 64: You could probably avoid the load of many resources in memory by doing:
```suggestion
removal_scope = SubscriptionFeatureRemoval.where(subscription: subscription, feature: feature)
removal_scope.where(entitlement_feature_id: feature.id).update_all!(deleted_at: Time.current)
```

The same comment applies in the next conditions

app/services/entitlement/subscription_entitlement_create_or_update_service.rb
```diff
@@ -0,0 +1,151 @@
...
+
+    def privilege_params_same_as_plan?(plan_entitlement)
+      # TODO: Fix this to ensure "9" == 9 (see )
+      #     See: SubscriptionEntitlementsUpdateService.feature_config_same_as_plan?
+      plan_entitlement.values.map do |v|
+        [v.privilege.code, v.value]
+      end.to_h.eql? privilege_params
+    end
+
+    # def subscription_plan_entitlements_as_params
```
vincent-pochet: Just to double check: are those commented blocks staying here for future usage?

app/queries/entitlement/subscription_entitlement_query.rb
julienbourdeau: `find_by_sql` is not available because SubscriptionEntitlement **is not an ActiveRecord model** but an ActiveModel model only.

I have considered making it an AR records but it must have a proper table. `find_by_sql` will still return all the dynamic attributes from the query but I'd need a "placeholder" table with some attributes like maybe just `code`. I'd also need the same for `SubscriptionEntitlementPrivilege`. 
Then, I thought this wasn't a good idea ^^

app/services/entitlement/subscription_entitlement_create_or_update_service.rb
julienbourdeau: Yes, absolutely. I'll add it in the next PR as I'm adding test the for Query object.

app/models/entitlement/subscription_entitlement.rb
julienbourdeau: 🙈

app/services/entitlement/subscription_entitlement_create_or_update_service.rb
julienbourdeau: I'm currently adding the "privilege" removal I intended for the query to become 
```
all_removal = SubscriptionFeatureRemoval.where(subscription: subscription, feature: feature).or.where(privilege: feature.privileges)
```

Also, I'm hoping to break down the service a little to pass those things as argument to the service.

app/services/entitlement/subscription_entitlement_create_or_update_service.rb
julienbourdeau: I doubt this will in this form in the end, I'll remove!

vincent-pochet (approved):

PR #4181, opened 2025-08-22
misc(wallet-transactions): Refactor void service

floganz:
## Context

Small preparation step for wallet calculation feature.
Currently this service accept each attribute as separate keyword argument creating a very long list of attributes in the `.initialize`, also we're duplicating default values set in the database.

## Description

Receive almost all params in `**transaction_params` simplifying `.initialize`.
Rely on DB defaults for non-provided attributes instead of duplicating the default values.
Rewrite specs to actually cover default DB values being assigned when not provided.

### DIFF
--- a/app/services/wallet_transactions/create_from_params_service.rb
+++ b/app/services/wallet_transactions/create_from_params_service.rb
@@ -55,13 +55,7 @@ module WalletTransactions
 
         if params[:voided_credits]
           wallet_credit = WalletCredit.new(wallet:, credit_amount: BigDecimal(params[:voided_credits]).floor(5), invoiceable: false)
-          void_result = WalletTransactions::VoidService.call(
-            wallet:,
-            wallet_credit:,
-            from_source: source,
-            metadata:,
-            priority:
-          )
+          void_result = WalletTransactions::VoidService.call(**params, wallet:, wallet_credit:)
           wallet_transactions << void_result.wallet_transaction
         end
       end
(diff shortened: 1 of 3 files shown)


annvelents (approved):

app/services/wallet_transactions/create_from_params_service.rb
```diff
@@ -55,13 +55,7 @@ def call
...
         if params[:voided_credits]
           wallet_credit = WalletCredit.new(wallet:, credit_amount: BigDecimal(params[:voided_credits]).floor(5), invoiceable: false)
-          void_result = WalletTransactions::VoidService.call(
-            wallet:,
-            wallet_credit:,
-            from_source: source,
-            metadata:,
-            priority:
-          )
+          void_result = WalletTransactions::VoidService.call(**params, wallet:, wallet_credit:)
```
julienbourdeau, line 58: I know that's the point of the named argument (vs positional) but it's the first time I see `**params` being at the beginning rather than the end.
Not a problem, just interesting.

floganz: It's was intentional, that has impact on override.
For example `params = {a: 3, b: 4}`
`call(**params, b: 7)` will be `{a: 3, b: 7}`
`call(b: 7, **params)` will be `{a: 3, b: 4}`

I've decided that we need to be extra sure that nothing will override explicitly passed values.

julienbourdeau: excellent catch 👏

PR #4188, opened 2025-08-22
feat(entitlements): Add SubscriptionEntitlementResolver

julienbourdeau:
- Rename `UpdateSubscriptionEntitlement` mutation to `CreateOrUpdateSubscriptionEntitlement` since you can also create new entitlements
- Add `SubscriptionEntitlementResolver` (singular) to retrieve only one, needed to load the edit form.


app/graphql/resolvers/entitlement/subscription_entitlement_resolver.rb
```diff
@@ -0,0 +1,33 @@
...
+      argument :subscription_id, ID, required: true
+
+      type Types::Entitlement::SubscriptionEntitlementObject, null: false
+
+      def resolve(subscription_id:, feature_code:)
+        raise forbidden_error(code: "feature_unavailable") unless License.premium?
+
+        subscription = current_organization.subscriptions.find(subscription_id)
+
+        # TODO: Replace this once we have `where` clause on SubscriptionEntitlementQuery
```
julienbourdeau, line 23: This is being added in a separate PR but I'd rather merge this one first to not block the frontned.

PR #4190, opened 2025-08-22
feat(entitlement): Allow privilege from plan to be removed from subscription

julienbourdeau:
- Add new db column so `Entitlement::SubscriptionFeatureRemoval` can belong to a privilege.
- modify the query so we ignore privileges listed in the subscription feature removals.
- Add beautiful hand-crafted test suite to the SubscriptionEntitlementQuery object.
- Add `to_h` to active model to make them much easier to work with

Next PR, will create the removals in services when needed

### DIFF
--- a/app/models/entitlement/subscription_entitlement_privilege.rb
+++ b/app/models/entitlement/subscription_entitlement_privilege.rb
@@ -22,7 +22,14 @@ module Entitlement
     attribute :sub_entitlement_value_id, :string
 
     def config
-      JSON.parse(super)
+      v = super
+      JSON.parse(v) if v.is_a?(String)
+    end
+
+    def to_h
+      h = attributes
+      h["config"] = config
+      h.with_indifferent_access
     end
   end
 end
--- a/app/models/entitlement/subscription_entitlement_privilege.rb
+++ b/app/models/entitlement/subscription_entitlement_privilege.rb
@@ -22,8 +22,7 @@ module Entitlement
     attribute :sub_entitlement_value_id, :string
 
     def config
-      v = super
-      JSON.parse(v) if v.is_a?(String)
+      super.then { it.is_a?(String) ? JSON.parse(it) : it }
     end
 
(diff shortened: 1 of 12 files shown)


app/models/entitlement/subscription_entitlement_privilege.rb
```diff
@@ -22,7 +22,14 @@ class SubscriptionEntitlementPrivilege
     attribute :sub_entitlement_value_id, :string
 
     def config
-      JSON.parse(super)
+      v = super
+      JSON.parse(v) if v.is_a?(String)
```
vincent-pochet: Not a blocker :)

```suggestion
      super.tap { JSON.parse(it) if it.is_a?(String) }
```

Also, I don't know if is is expected but if config is not a string, the method will return `nil`

spec/queries/entitlement/subscription_entitlement_query_spec.rb
```diff
@@ -0,0 +1,378 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Entitlement::SubscriptionEntitlementQuery, type: :service, transaction: false do
```
vincent-pochet: Any reason to prevent the transactional specs here?

vincent-pochet (approved):

spec/queries/entitlement/subscription_entitlement_query_spec.rb
julienbourdeau: Thank you! I was debugging and I wanted the seeder to write the data so I could check it in TablePlus. I forgot to remove 🙏

app/models/entitlement/subscription_entitlement_privilege.rb
julienbourdeau: very good point. It's unexpected to have anything else that I json string... unless you create the model manualy. I like the tap version 👌

PR #4193, opened 2025-08-25
Feat(billing-entities): migrate customers between be

annvelents:
## Context

Sometimes as en exception we need to migrate customers between billing entities. there are a lot of restrictions and this task should be executed very carefully. (we should migrate all subscriptions, wallets, but we do not care about invoices and wallet transactions)

the approach is:
- check all restrictions
- in a transaction:
1. discard old customer
2. duplicate new customer, nullifying fields, that should be auto-generated or reset on the new customer
3. duplicate subscriptions for the new customer
4. terminate active subscriptions on old customer (make sure we won't generate termination invoice)
5. generate new lifetime usage for the active subscription on the new customer
6. duplicate supported integration customers and payment_provider customers (netsuite integrations will need to be created anew)
7. duplicate wallet, wallet targets and wallet top up rules (do not create inbound transaction, simply create wallet with original balance)
8. terminate all relations for the old customer

### DIFF
--- a/lib/tasks/customers.rake
+++ b/lib/tasks/customers.rake
@@ -6,7 +6,8 @@ namespace :customers do
     Customer.unscoped.order(:created_at).find_each(&:save)
   end
 
-  desc "Migrate customer to a new billing entity"
+  # WARNING! Potentially dangerous task
+  desc "Migrate customer to a new billing entity. This version is actual on August 2025, please, check before running if anything needs to be updated"
   task migrate_to_new_entity: :environment, [:customer_external_id, :billing_entity_code] do |t, args|
     customer_external_id = args[:customer_external_id]
     billing_entity_code = args[:billing_entity_code]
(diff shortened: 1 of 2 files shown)


lib/tasks/customers.rake
```diff
@@ -5,4 +5,103 @@ namespace :customers do
...
+          rate_amount: wallet.rate_amount,
+          currency: wallet.currency,
+          expiration_at: wallet.expiration_at,
+          invoice_requires_successful_payment: wallet.invoice_requires_successful_payment,
+          applies_to: {
+            fee_types: wallet.allowed_fee_types
+          },
+          granted_credits: wallet.credits_balance.to_s
+        }
+        new_wallet = Wallets::CreateService.call!(params: wallet_params).wallet
```
julienbourdeau, line 90: I was about to ask why you didn't do `new_wallet = wallet.dup` but then I opened the Wallet::CraeteService, so now I know :D

julienbourdeau (approved):

PR #4197, opened 2025-08-26
fix(stripe): Fix `customer.updated` Stripe webhook when stripe customer not found in DB

groyoh:
## Context

When a customer is updated on Stripe, we receive a `customer.updated` webhook from Stripe. If we cannot match the payload with a `PaymentProviderCustomers::StripeCustomer`, we do the following:

- If the payload does not include the `metadata.lago_customer_id` value, it means the customer was created outside Lago and we can ignore the webhook.
- If the payload include the `metadata.lago_customer_id` but we cannot find it in the database, it means the customer was deleted or belong to another organization and we can ignore the webhook.
- Otherwise, we'll return a `stripe_customer_not_found` error.

There is an edge-case that wasn't handled here though. If someone manually updated the Stripe customer information when updating a Lago customer, the previous Stripe customer will no longer be linked to a record in our database. Not that even if the `provider_customer_id` is set to `nil`, the `PaymentProviderCustomers::StripeCustomer` will still be present but with a nil `provider_customer_id`.

## Description

This makes sure we ignore the webhook in the case mentioned above.

### DIFF
--- a/app/services/payment_providers/stripe/webhooks/base_service.rb
+++ b/app/services/payment_providers/stripe/webhooks/base_service.rb
@@ -25,14 +25,29 @@ module PaymentProviders
         end
 
         def handle_missing_customer
-          # NOTE: Stripe customer was not created from lago
-          return result unless metadata&.key?(:lago_customer_id)
+          return result if stripe_customer_created_outside_lago?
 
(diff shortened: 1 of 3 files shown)


vincent-pochet (approved):
Good one 👍

PR #4198, opened 2025-08-26
feat: change the calculations

brunomiguelpinto:
## Context

Progressive billing credits were being distributed across all invoice fees, regardless of whether a fee belonged to the same charge as the progressive billing invoice. 

## Description

This PR updates Credits::ProgressiveBillingService so that credits from a progressive billing invoice are only applied to matching fees on the target invoice (same charge_id/context), instead of being spread across all fees.

This change was needed so that when we apply the credits from the wallet, we need to know the remaining fee, removing the progressive billing part 

This issue would be more evident if we had multiple wallets, and some of them only applied to specific fees, the amount we need to take would not be correct (because the calculations would be affected by the progressive billing )

### DIFF
--- a/app/services/credits/progressive_billing_service.rb
+++ b/app/services/credits/progressive_billing_service.rb
@@ -60,7 +60,13 @@ module Credits
 
     def apply_credit_to_fees(progressive_billing_invoice)
       progressive_billing_invoice.fees.charge.each do |progressive_fee|
-        fee = invoice.fees.find_by(charge_id: progressive_fee.charge_id)
+        fee = invoice.fees.find_by(
+          charge_id: progressive_fee.charge_id,
+          charge_filter_id: progressive_fee.charge_filter_id,
+          grouped_by: progressive_fee.grouped_by,
+        )
+        next unless fee
+
         fee.precise_coupons_amount_cents += progressive_fee.amount_cents
         fee.save!
       end
--- a/app/services/credits/progressive_billing_service.rb
+++ b/app/services/credits/progressive_billing_service.rb
@@ -63,7 +63,7 @@ module Credits
         fee = invoice.fees.find_by(
           charge_id: progressive_fee.charge_id,
           charge_filter_id: progressive_fee.charge_filter_id,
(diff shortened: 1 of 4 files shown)


app/services/credits/progressive_billing_service.rb
```diff
@@ -58,17 +58,10 @@ def call
...
-        if base_amount.positive?
-          fee.precise_coupons_amount_cents += (
-            credit.amount_cents * (fee.amount_cents - fee.precise_coupons_amount_cents)
-          ).fdiv(base_amount)
-        end
-
-        fee.precise_coupons_amount_cents = fee.amount_cents if fee.amount_cents < fee.precise_coupons_amount_cents
+    def apply_credit_to_fees(progressive_billing_invoice)
+      progressive_billing_invoice.fees.charge.each do |progressive_fee|
+        fee = invoice.fees.find_by(charge_id: progressive_fee.charge_id)
```
vincent-pochet: You should also filter by `charge_filter_id` and `grouped_by` as multiple fees could share the same `charge_id`

brunomiguelpinto: Yes makes sense

app/services/credits/progressive_billing_service.rb
```diff
@@ -58,17 +58,16 @@ def call
...
 
-    def apply_credit_to_fees(credit, total_charges_amount)
-      base_amount = total_charges_amount - invoice.progressive_billing_credit_amount_cents
+    def apply_credit_to_fees(progressive_billing_invoice)
+      progressive_billing_invoice.fees.charge.each do |progressive_fee|
+        fee = invoice.fees.find_by(
+          charge_id: progressive_fee.charge_id,
+          charge_filter_id: progressive_fee.charge_filter_id,
+          grouped_by: progressive_fee.grouped_by
+        )
```
vincent-pochet: For performance purpose, maybe you should first preload the list of fees of the invoice and use `find` on the list to retrieve the matching fee

vincent-pochet (approved):

PR #4200, opened 2025-08-26
feat(fixed_charges): Add fixed charge events

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

 Add fixed charge events table and model, initial setup.

 Fixed charge events are only created internally, users won't be able to
 emit this kind of events. We only need to the fixed_charge_id and
 subscription_id as foreign keys.

### DIFF
--- a/db/migrate/20250826081205_create_fixed_charge_events.rb
+++ b/db/migrate/20250826081205_create_fixed_charge_events.rb
@@ -7,7 +7,6 @@ class CreateFixedChargeEvents < ActiveRecord::Migration[8.0]
       t.belongs_to :subscription, null: false, foreign_key: true, type: :uuid
       t.belongs_to :fixed_charge, null: false, foreign_key: true, type: :uuid
 
-      t.jsonb :properties, null: false, default: {}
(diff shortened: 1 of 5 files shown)


db/migrate/20250826081205_create_fixed_charge_events.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class CreateFixedChargeEvents < ActiveRecord::Migration[8.0]
+  def change
+    create_table :fixed_charge_events, id: :uuid do |t|
+      t.belongs_to :organization, null: false, foreign_key: true, type: :uuid
+      t.belongs_to :subscription, null: false, foreign_key: true, type: :uuid
+      t.belongs_to :fixed_charge, null: false, foreign_key: true, type: :uuid
+
+      t.jsonb :properties, null: false, default: {}
```
annvelents: I think we do not need properties here...
In normal events properties are used for billable metric's aggregation properties, but for fixed_charge_events and add_ons we have kinda "count" aggregation, so we only need units count in the event

PR #4201, opened 2025-08-26
fix(customer): Fix payment job race-condition on customer deletion

groyoh:
## Context

When a customer is deleted, we terminate all active subscriptions and generate a termination invoice for each one. Each invoice may trigger an asynchronous payment.

Currently, some payment jobs begin executing before the surrounding database transaction commits. Because the invoices are not yet persisted, the job cannot find them, causing the payment creation to fail.

## Description

This PR ensures the payment job is only scheduled once the transaction is committed.

### DIFF
--- a/app/services/invoices/payments/create_service.rb
+++ b/app/services/invoices/payments/create_service.rb
@@ -82,7 +82,7 @@ module Invoices
       def call_async
         return result unless provider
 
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):

groyoh:
Thanks @vincent-pochet ! 👍

PR #4204, opened 2025-08-27
feat(fixed_charges): FixedChargeEvents::CreateService

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

 Add a service to create FixedChargeEvent records


app/services/fixed_charge_events/create_service.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module FixedChargeEvents
+  class CreateService < BaseService
+    Result = BaseResult[:fixed_charge_event]
+
+    def initialize(organization:, subscription:, fixed_charge:, units: 0.0, timestamp: Time.current)
```
annvelents: can we only send a fixed_charge and subscription, and get organization from the subscription? (to have to pass less parameters, also we can get just organization_id form subscription to not have to load the whole organization)

annvelents (approved):
LGTM, only a small comment about not passing organization into the service

PR #4208, opened 2025-08-27
Fix(fixed_charges) - update fixed_charge serializer

annvelents:
## Context

After first round of QA, these fields needed to be updated

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -146,31 +146,6 @@ module Subscriptions
       new_subscription
     end
 
-    def after_commited_created_subscription(subscription)
-      after_commit do
-        if should_be_billed_today?(subscription)
-          # NOTE: Since job is launched from inside a db transaction
-          #       we must wait for it to be committed before processing the job.
-          #       We do not set offset anymore but instead retry jobs
-          BillSubscriptionJob.perform_later(
-            [subscription],
-            Time.zone.now.to_i,
-            invoicing_reason: :subscription_starting,
-            skip_charges: true
-          )
-        end
-
-        if subscription.active?
-          SendWebhookJob.perform_later("subscription.started", subscription)
-          Utils::ActivityLog.produce(subscription, "subscription.started")
-        end
-
-        if subscription.should_sync_hubspot_subscription?
-          Integrations::Aggregator::Subscriptions::Hubspot::CreateJob.perform_later(subscription:)
-        end
-      end
-    end
-
     def upgrade_subscription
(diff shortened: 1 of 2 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -116,6 +116,7 @@ def create_subscription
       else
         new_subscription.mark_as_active!
       end
+      create_fixed_charge_units_override(new_subscription) if overrides_only_fixed_charge_units?
```
ancorcruz, line 119: we should create an override instead of units override. This line is not required.

app/services/subscriptions/create_service.rb
```diff
@@ -162,17 +163,18 @@ def downgrade_subscription
...
-        plan: params.key?(:plan_overrides) ? override_plan(plan) : plan,
+        plan: params.key?(:plan_overrides) ? plan_for_override_params(plan) : plan,
         name:,
         external_id: current_subscription.external_id,
         subscription_at: current_subscription.subscription_at,
         status: :pending,
         billing_time: current_subscription.billing_time,
         ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at
       )
+      create_fixed_charge_units_override(new_subscription) if overrides_only_fixed_charge_units?
```
ancorcruz, line 177: we should create an override instead of units override. This line is not required.

app/services/subscriptions/create_service.rb
```diff
@@ -223,8 +225,27 @@ def editable_subscriptions
         .order(started_at: :desc)
     end
 
-    def override_plan(plan)
+    def plan_for_override_params(plan)
+      if overrides_only_fixed_charge_units?
+        return plan
+      end
```
ancorcruz, line 231: lastest product decision; it should create a plan override also when only fixed charge units are changed. Hence this condition is not needed anymore.

app/services/subscriptions/create_service.rb
```diff
@@ -223,8 +225,27 @@ def editable_subscriptions
...
+
+    def create_fixed_charge_units_override(subscription)
+      params[:plan_overrides][:fixed_charges].each do |fixed_charge|
+        FixedChargeUnitOverrideService.call!(
+          subscription:,
+          fixed_charge: subscription.plan.fixed_charges.find(fixed_charge[:id]),
+          units: fixed_charge[:units]
+        )
+      end
+    end
```
ancorcruz, line 249:
```suggestion
```

not needed anymore.

ancorcruz (changes requested):
fixed charge subscription unit override should be removed as we won't implement it for now.

PR #4209, opened 2025-08-27
fix(credit-notes): Prevent `ArgumentError` masking with proper `reason` validation

groyoh:
## Context

The `CreditNote::Create` service was catching all `ArgumentError` exceptions and converting them to `reason` validation errors, which could mask legitimate errors from other parts of the code. This made debugging difficult and could hide real bugs.

## Description

This fixes the issue by adding an explicit validation.

### DIFF
--- a/app/services/credit_notes/create_service.rb
+++ b/app/services/credit_notes/create_service.rb
@@ -23,6 +23,7 @@ module CreditNotes
       return result.not_found_failure!(resource: "invoice") unless invoice
       return result.forbidden_failure! unless should_create_credit_note?
       return result.not_allowed_failure!(code: "invalid_type_or_status") unless valid_type_or_status?
+      return result.single_validation_failure!(field: :reason, error_code: "value_is_invalid") if invalid_reason?
 
       ActiveRecord::Base.transaction do
         result.credit_note = CreditNote.new(
@@ -93,8 +94,6 @@ module CreditNotes
       result
     rescue ActiveRecord::RecordInvalid => e
       result.record_validation_failure!(record: e.record)
-    rescue ArgumentError
-      result.single_validation_failure!(field: :reason, error_code: "value_is_invalid")
     rescue BaseService::FailedResult => e
       e.result
     end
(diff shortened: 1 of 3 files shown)


app/services/credit_notes/create_service.rb
```diff
@@ -23,6 +23,7 @@ def call
       return result.not_found_failure!(resource: "invoice") unless invoice
       return result.forbidden_failure! unless should_create_credit_note?
       return result.not_allowed_failure!(code: "invalid_type_or_status") unless valid_type_or_status?
+      return result.single_validation_failure!(field: :reason, error_code: "value_is_invalid") if invalid_reason?
```
julienbourdeau: This is to continue to return the same error as before? instead of a "record_validation" error?

app/services/credit_notes/create_service.rb
```diff
@@ -113,6 +112,10 @@ def call
     delegate :credit_note, to: :result
     delegate :customer, to: :invoice
 
+    def invalid_reason?
+      CreditNote.reasons.keys.exclude?(reason.to_s)
```
julienbourdeau, line 115: TIL! I didn't know `enum` creates this plural method.

julienbourdeau (approved):

app/services/credit_notes/create_service.rb
groyoh: What happens is that setting an invalid enum will raise `ArgumentError` instead of an adding a validation error when calling `.valid?` :

```ruby
lago-api(dev)> CreditNote.new(reason:'draft')
(lago-api):2:in '<main>': 'draft' is not a valid reason (ArgumentError)

          raise ArgumentError, "'#{value}' is not a valid #{name}"
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

My guess is that the initial code was introduced to catch invalid `reason` being provided. So what I did was to keep the same behavior for `reason` but allow other `ArgumentError` to bubble up.

Does that make sense ?

vincent-pochet (approved):
Good catch, It's a good fix!

Have you also considered adding the `validate: true` to the `reason` enum?
It was introduce after the creation of the Credit Note feature and it has the advantage on relying on the standard rails validation that is already in place in this service.
The downside is that it could have some unexpected side-effects but it should be very limited.

app/services/credit_notes/create_service.rb
groyoh: I've updated it based on Vincent's comment.

PR #4210, opened 2025-08-27
fix(entitlements): Convert boolean string for frontend

julienbourdeau:
With ActiveRecord if you do `EntitlementValue.update value: true`, it will convert the value to `"t"` because the db column is string.
This is intentional and the serializer will use `ActiveModel::Type::Boolean.new.cast("t")` to convert it to `true.

In GQL, we rely on `"true"` and `"false"` string to display the select box. So we need to convert on read.
On write it's fine because the value will always be casted when returned via the API.


groyoh (approved):
It doesn't impact the tests ?

PR #4212, opened 2025-08-27
fix(invoice-pdf): Fix the address on the invoice PDF based on the country

murparreira:
## Context

There is a need to format the address in the invoice PDF based on the billing entity and the customer country.

## Description

Created a helper function to format the address and use it on all versions of the invoice PDFs.

### DIFF
--- a/app/views/helpers/slim_helper.rb
+++ b/app/views/helpers/slim_helper.rb
@@ -11,4 +11,46 @@ class SlimHelper
       )
     end.render(context, **locals)
   end
+
+  def self.format_address(address_line1, address_line2, city, state, zipcode, country_code)
+    country = ISO3166::Country.new(country_code)&.common_name
+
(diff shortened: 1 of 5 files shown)


app/views/helpers/slim_helper.rb
```diff
@@ -11,4 +11,46 @@ def self.render(path, context, **locals)
...
+
+    case country_code&.upcase
+    when "DE", "FR", "IT"
+      [
+        address_line1,
+        address_line2,
+        [zipcode, city].compact.join(" "),
+        country
+      ].compact.reject(&:empty?)
+    when "US", "CA", "GB", "AU"
```
groyoh: I believe `GB` and `AU` have different formats, like no comma and uppercase.

groyoh (changes requested):

PR #4214, opened 2025-08-28
misc(open-struct): Fake in advance invoices

vincent-pochet:
## Roadmap Task

## Context

## Description


vincent-pochet:
> Would `freeze` on the invoice work instead or do we expect to modify some attributes ?

It could work, we are not not modifying anything. Still I prefer to use a dedicated structure, it could save some unexpected behavior.

PR #4216, opened 2025-08-28
fix(stripe): Skip Checkout URL generation for `crypto` and `customer_balance`

groyoh:
## Context

Stripe checkout URLs were being generated for all payment methods, including `crypto` and `customer_balance` which don't require setup flows. This created unnecessary processing and errors down the line in the `PaymentProviderCustomers::StripeService#generate_checkout_url` method.

This is a follow-up to #3247.

## Description

This skips the checkout url generation altogether unless there's at least one payment method that requires it.


app/services/payment_provider_customers/stripe_service.rb
```diff
@@ -107,6 +109,10 @@ def generate_checkout_url(send_webhook: true)
 
     delegate :customer, to: :stripe_customer
 
+    def payment_methods_do_not_require_setup?
```
annvelents: what do you think if we change the "sign" of this method to `payment_methods_require_setup?`, and will also reuse on the line 26?

annvelents (approved):
LGTM!

app/services/payment_provider_customers/stripe_service.rb
julienbourdeau: I'm not a big fan of the name either, I was about to propose `no_payment_methods_require_setup?` but I think yours is better.

julienbourdeau (approved):

app/services/payment_provider_customers/stripe_service.rb
groyoh: Sure, no problem!

groyoh:
Thanks for the review @annvelents @julienbourdeau 🙏

PR #4217, opened 2025-08-28
Feat(fixed_charges-11): allow fixed_charges overrides in subscription through API

annvelents:
## Context

Allow fixed charges params in plan_overrides for subscriptions controller,
include fixed_charges when serializing a plan for a subscription

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -20,7 +20,7 @@ class Subscription < ApplicationRecord
   has_many :usage_thresholds, through: :plan
   has_many :entitlements, class_name: "Entitlement::Entitlement"
   has_many :entitlement_removals, class_name: "Entitlement::SubscriptionFeatureRemoval"
-  has_many :fixed_charges, through: :plan
+  has_many :fixed_charges, -> { kept }, through: :plan
   has_many :add_ons, through: :fixed_charges
   has_many :activity_logs,
     -> { order(logged_at: :desc) },


app/serializers/v1/plan_serializer.rb
```diff
@@ -49,7 +49,8 @@ def fixed_charges
       ::CollectionSerializer.new(
         model.fixed_charges,
         ::V1::FixedChargeSerializer,
-        collection_name: "fixed_charges"
+        collection_name: "fixed_charges",
+        includes: include?(:taxes) ? %i[taxes] : []
```
groyoh, line 53: I don't think this behavior is tested.

app/controllers/api/v1/subscriptions_controller.rb
```diff
@@ -232,6 +232,15 @@ def plan_overrides
...
               }
             ],
+            fixed_charges: [
+              :id,
+              :add_on_id,
+              :invoice_display_name,
+              :charge_model,
+              {properties: {}},
+              :units,
+              {tax_codes: []}
```
groyoh: I guess it would be nice to includes all those fields in the tests.

app/serializers/v1/subscription_serializer.rb
```diff
@@ -53,7 +53,7 @@ def entitlements
     def plan
       ::V1::PlanSerializer.new(
         model.plan,
-        includes: %i[charges usage_thresholds taxes minimum_commitment]
+        includes: %i[charges fixed_charges usage_thresholds taxes minimum_commitment]
```
jdenquin: do we really need it on the sub? We add a new query on the sub serializer and subscriptions endpoints are really slow because of loading everything

app/serializers/v1/plan_serializer.rb
annvelents: that's true!
Added 😁  thank you for reminder!

app/controllers/api/v1/subscriptions_controller.rb
annvelents: yep, you're right, updating

app/controllers/api/v1/subscriptions_controller.rb
```diff
@@ -232,6 +232,15 @@ def plan_overrides
                 ]
               }
             ],
+            fixed_charges: [
```
ancorcruz: please add `apply_units_immediately` as permitted param for fixed charges.

app/controllers/api/v1/subscriptions_controller.rb
```diff
@@ -232,6 +232,15 @@ def plan_overrides
...
+              :id,
+              :add_on_id,
+              :invoice_display_name,
+              :charge_model,
```
ancorcruz: charge model is not allowed to be overridden

app/serializers/v1/subscription_serializer.rb
ancorcruz: @annvelents lets remove the `fixed_charges` nested collection here and define a new endpoint to only get the fixed charges.... it won't be consistent with the `charges` but we can not keep adding more nested stuff into subscriptions serializer for performance reasons...

`GET /v1/subscription/:id/fixed_charges` should return the fixed charges for the subscription, getting them from the subscription's plan and overrides.

spec/requests/api/v1/subscriptions_controller_spec.rb
```diff
@@ -373,6 +373,50 @@
...
+          }
+        }
+      end
+
+      it "creates a subscription with overridden plan with fixed_charges" do
+        subject
+
+        expect(response).to have_http_status(:success)
+        expect(json[:subscription][:plan][:fixed_charges].first).to include(
+          lago_add_on_id: fixed_charge.add_on.id,
```
ancorcruz: shall we also test it have the fixed charge lago id?

app/controllers/api/v1/subscriptions_controller.rb
```diff
@@ -232,6 +232,15 @@ def plan_overrides
                 ]
               }
             ],
+            fixed_charges: [
+              :id,
+              :add_on_id,
```
ancorcruz: `add_on_id` is not allowed to be changed, we don't need it here as well...

app/models/subscription.rb
```diff
@@ -21,7 +21,7 @@ class Subscription < ApplicationRecord
   has_many :entitlements, class_name: "Entitlement::Entitlement"
   has_many :entitlement_removals, class_name: "Entitlement::SubscriptionFeatureRemoval"
   has_many :subscription_fixed_charge_units_overrides, dependent: :destroy
-  has_many :fixed_charges, through: :plan
+  has_many :fixed_charges, -> { kept }, through: :plan
```
julienbourdeau, line 23: I'm not sure why this is needed 🤔 it seems that fixed charges have `kept` default scope. Is it because of the `through:`

app/controllers/api/v1/subscriptions/base_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+        private
+
+        attr_reader :subscription
+
+        def find_subscription
+          @subscription = current_organization.subscriptions
+            .order("terminated_at DESC NULLS FIRST, started_at DESC") # TODO: Confirm
+            .find_by!(
+              external_id: params[:subscription_external_id]
```
julienbourdeau, line 17:
```suggestion
              external_id: params[:subscription_external_id]
              status: params[:status] || :active
```

I'm thinking we should add this status param same as the main controller
https://github.com/getlago/lago-api/blob/a2c7c7135c65ab2b67e2901c6de7183636c9e922/app/controllers/api/v1/subscriptions_controller.rb#L138

Otherwise, if the subscription is terminated we'll get:
```
GET /v1/subscriptions/:id 
 ===> not found
GET /v1/subscriptions/:id/fixed_charges
 ===> works
```
 being able to pass the status to all endpoint would be clearer I think.

For alerts and entitlements it would be better too and we can delete the `find_susbscription` in entitlement controller. **We can do it in a separte PR if you prefer!**

app/models/subscription.rb
annvelents: yes. It was a very interesting "what I learned today" lesson 😅  apparently when we have has_many :through, rails ignores the default scope on the model 🤯 
while calling `subscription.plan.fixed_charges` would still respect it

app/controllers/api/v1/subscriptions/base_controller.rb
annvelents: ow, I like the idea of adding status in the params! because I exactly wanted being able show these for pending subscriptions or terminated... so having status in params - I like it...
but yes, alerts and entitlements have different behaviour and I wasn't sure what to do.

let's do it in a separate PR, as it changes something else, not related to this one

app/models/subscription.rb
julienbourdeau: TIL !

PR #4225, opened 2025-08-29
fix(fees): Avoid duplicated in advance - Phase 1

vincent-pochet:
## Context

We recently identified a bug in the processing of events for pay in advance charges that led to some duplicated fees in the database. See https://github.com/getlago/lago-api/pull/4220 for the fix of the initial bug.

On the fees table, an index named `idx_on_pay_in_advance_event_transaction_id_charge_i_16302ca167` was supposed to prevent duplicated in advance fees:

```sql
CREATE UNIQUE INDEX idx_on_pay_in_advance_event_transaction_id_charge_i_16302ca167 ON public.fees 
USING btree (pay_in_advance_event_transaction_id, charge_id, charge_filter_id) 
WHERE ((created_at > '2025-01-21 00:00:00'::timestamp without time zone) 
  AND (pay_in_advance_event_transaction_id IS NOT NULL) 
  AND (pay_in_advance = true));
```

The issue reside in the way Postgres handles null values in indexes. Each null value is considered a distinct value meaning the index is useless in its current form as since `charge_filter_id` is null most of the time...

## Goal

The main goal of the fix is to add a new set of index to really prevent duplication in the database.
The difficulty will reside in the fact that we already have duplicated values in DB and we cannot remove them as they have already impacted the customers.

The complete fix will follow these steps:
- Phase 1:
  - Add a new `duplicated_in_advance` flag on the fees table with default value to false
  - Set this field to true for all existing duplicated fees
- Phase 2: 
  - Add new indexes making preventing the duplication of fees making sure that null values are not considered distinct and ignoring pre-existing duplication
  - Remove the existing index 

## Description

This PR is the phase 1.
It adds two migrations to:
- Add the new `duplicated_in_advance` field to the fees table
- Fill the field for all existing duplicated fees

### DIFF
--- a/db/migrate/20250828144553_populate_duplicated_in_advance_fees.rb
+++ b/db/migrate/20250828144553_populate_duplicated_in_advance_fees.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class PopulateDuplicatedInAdvanceFees < ActiveRecord::Migration[8.0]
+  def up
+    query = <<~SQL
+      WITH duplicated_fees AS (
+        SELECT organization_id, pay_in_advance_event_transaction_id, charge_id, charge_filter_id
+        FROM fees
+        WHERE fees.deleted_at IS NULL
+          AND fees.pay_in_advance = TRUE
+          AND fees.pay_in_advance_event_transaction_id IS NOT NULL
+          AND (created_at > '2025-01-21 00:00:00')  -- TODO: check what happened before...
+        GROUP BY pay_in_advance_event_transaction_id, charge_id, charge_filter_id, organization_id
+        HAVING COUNT(*) > 1
+      )
+      UPDATE fees f
+      SET duplicated_in_advance = TRUE
+      FROM duplicated_fees df
+      WHERE f.pay_in_advance_event_transaction_id = df.pay_in_advance_event_transaction_id
+        AND f.charge_id = df.charge_id
+        AND f.charge_filter_id = df.charge_filter_id
+        AND f.organization_id = df.organization_id
+        AND f.pay_in_advance = TRUE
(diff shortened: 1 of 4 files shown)


db/migrate/20250828144553_populate_duplicated_in_advance_fees.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class PopulateDuplicatedInAdvanceFees < ActiveRecord::Migration[8.0]
+  def up
+    query = <<~SQL
+      WITH duplicated_fees AS (
+        SELECT organization_id, pay_in_advance_event_transaction_id, charge_id, charge_filter_id
+        FROM fees
+        WHERE fees.deleted_at IS NULL
```
groyoh, line 9: Is the deleted_at check necessary ? Shouldn't it be  present in the query below or in the index ?

groyoh (approved):
Looks good to me but it would be worht having another pair of 👀 to double check.

db/migrate/20250828144553_populate_duplicated_in_advance_fees.rb
vincent-pochet: Good catch! I guess it should be in the update query as well, and latter in the index.

I will update this query

julienbourdeau (approved):
Beautiful 🥇

julienbourdeau:
Since the data migration is in a db migration, we won't event need a bridge version, right?

vincent-pochet:
> Since the data migration is in a db migration, we won't event need a bridge version, right?

No indeed, the migration will be enough. Plus it should not be an issue for other instances as the root cause of the duplication is related to the event-processor pipeline

PR #4228, opened 2025-08-29
fix(applied-coupons): Fix `GET /api/v1/applied_coupons` when coupon is deleted

groyoh:
## Context

When fetching `GET /api/v1/applied_coupons`, if the related coupon was deleted, we would fail to retrieve and serialize the coupon data due to the coupon's default scope. This would result in:

```ruby
NoMethodError
undefined method 'id' for nil (NoMethodError)

        lago_coupon_id: model.coupon.id,
                                    ^^^
```

## Description

This fixes it by preloading the coupon using the `Coupon.with_discarded` scope.

I noticed while fixing that, when filtering by coupon code, we didn't take into account deleted coupons. So we could have ended up in the following situation:

```
# Context

Coupon 1 (Deleted)
Coupon 2
Applied coupon 1 (Terminated)
Applied coupon 2

# GET /api/v1/applied_coupons
[applied_coupon_2, applied_coupon_1]

# GET /api/v1/applied_coupons?coupon_code[]=applied_coupon_1&coupon_code[]=applied_coupon_2
[applied_coupon_2]
```

So I've updated the `AppliedCouponsQuery` to fix that as well.

Note that another solution would be to change `belongs_to :coupon` into `belongs_to :coupon, -> { with_discarded}` but that could have had a bigger impact.

### DIFF
--- a/app/controllers/api/v1/applied_coupons_controller.rb
+++ b/app/controllers/api/v1/applied_coupons_controller.rb
@@ -41,7 +41,7 @@ module Api
         if result.success?
           render(
             json: ::CollectionSerializer.new(
-              result.applied_coupons.includes(:credits),
+              result.applied_coupons.includes(:credits, :coupon, :customer),
               ::V1::AppliedCouponSerializer,
               collection_name: "applied_coupons",
               meta: pagination_metadata(result.applied_coupons),
--- a/app/models/applied_coupon.rb
+++ b/app/models/applied_coupon.rb
@@ -4,7 +4,7 @@ class AppliedCoupon < ApplicationRecord
   include PaperTrailTraceable
   include Currencies
 
-  belongs_to :coupon
+  belongs_to :coupon, -> { with_discarded }
   belongs_to :customer
   belongs_to :organization
 
--- a/spec/factories/applied_coupons.rb
+++ b/spec/factories/applied_coupons.rb
@@ -10,5 +10,9 @@ FactoryBot.define do
     amount_currency { "EUR" }
     status { "active" }
     frequency { "once" }
+
+    trait :terminated do
+      terminated_at { 1.day.ago }
+    end
   end
 end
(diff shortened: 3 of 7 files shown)


app/controllers/api/v1/applied_coupons_controller.rb
```diff
@@ -74,6 +76,15 @@ def index_filters
       def resource_name
         "applied_coupon"
       end
+
+      def preload_applied_coupons_associations(applied_coupons)
+        [
+          :credits,
+          [:coupon, Coupon.with_discarded]
```
julienbourdeau: Maybe I'm getting old but it took me a minute to understand. Maybe this, is a little clearer.

```suggestion
          [:credits, nil],
          [:coupon, Coupon.with_discarded]
```

app/queries/applied_coupons_query.rb
```diff
@@ -22,7 +22,10 @@ def base_scope
   end
 
   def with_coupon_code(scope)
-    scope.joins(:coupon).where(coupons: {code: filters.coupon_code})
+    # We don't use `.joins(:coupon)` because we still want to retrieve applied coupons
+    # even if the coupon is deleted
+    scope.joins("INNER JOIN coupons ON coupons.id = applied_coupons.coupon_id")
+      .where(coupons: {code: filters.coupon_code})
```
julienbourdeau: As we discussed, updating the relationship is better.

```rb
belongs_to :coupon, -> { with_discarded }
```

groyoh: Yeah that makes sense.

app/controllers/api/v1/applied_coupons_controller.rb
groyoh: Even when writing the code I wasn't sure it would work but it did 😂

app/queries/applied_coupons_query.rb
groyoh: I updated the PR to match this.

PR #4229, opened 2025-08-29
fix(clickhouse): Avoid nil connections when performing queries

vincent-pochet:
## Context

This PR is an attempt to fix the `undefined method 'select_all' for nil (NoMethodError)` happening from time to time with clickhouse connection

## Description

The PR refactors the way we relies on Clikchouse connections to execute raw SQL queries to make sure that we always have an active connection.

### DIFF
--- a/app/services/events/stores/clickhouse_store.rb
+++ b/app/services/events/stores/clickhouse_store.rb
@@ -71,7 +71,7 @@ module Events
       def grouped_last_event
         groups, group_names = grouped_arel_columns
 
-        with_retry do
+        connection_with_retry do |connection|
           cte_sql = events_sql(
             ordered: true,
             select: groups + [arel_table[:decimal_value].as("property"), arel_table[:timestamp]]
@@ -88,7 +88,7 @@ module Events
             ORDER BY #{group_names}, events.timestamp DESC
           SQL
 
-          prepare_grouped_result(::Clickhouse::EventsEnriched.connection.select_all(sql).rows, timestamp: true)
+          prepare_grouped_result(connection.select_all(sql).rows, timestamp: true)
         end
       end
 
@@ -99,7 +99,7 @@ module Events
       end
 
       def count
-        with_retry do
+        connection_with_retry do |connection|
           sql = <<-SQL
           WITH events AS (#{events_sql})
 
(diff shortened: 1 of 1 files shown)


app/services/events/stores/clickhouse_store.rb
```diff
@@ -708,6 +708,31 @@ def with_retry(&)
...
+
+            sleep(0.05)
+            retry
+          end
+
+          raise
+        ensure
+          ::Clickhouse::BaseRecord.connection_handler.clear_active_connections!
+        end
+      end
```
groyoh: Maybe using `with_connection` would be best as it is the default behavior of ActiveRecord query methods:

```suggestion
      def connection_with_retry(&)
        attempts = 0
        
        begin
          attempts += 1

          ::Clickhouse::BaseRecord.with_connection(&)
        rescue Errno::ECONNRESET, ActiveRecord::ActiveRecordError, NoMethodError
          if attempts < MAX_RETRIES
            sleep(0.05)
            retry
          end

          raise
        end
      end
```

I tested using the following:

```ruby
loop do 
  begin
    Clickhouse::BaseRecord.with_connection do |connection|
      connection.select_all("SELECT 1 FROM events_enriched LIMIT 1").to_a
    end

    puts "Query successful"
  
  rescue => e
    puts "Failed to get connection: #{e.message}"
  end
  sleep 1
end
```

And it seems to work when shutting down Clickhouse:

```
Failed to get connection: Failed to open TCP connection to clickhouse:8123 (getaddrinfo(3): Name or service not known)
Failed to get connection: Failed to open TCP connection to clickhouse:8123 (getaddrinfo(3): Name or service not known)
Failed to get connection: Failed to open TCP connection to clickhouse:8123 (getaddrinfo(3): Name or service not known)
Failed to get connection: Failed to open TCP connection to clickhouse:8123 (getaddrinfo(3): Name or service not known)
Failed to get connection: Failed to open TCP connection to clickhouse:8123 (getaddrinfo(3): Name or service not known)

Query successful
Query successful
Query successful
Query successful
Query successful
```

groyoh (approved):

app/services/events/stores/clickhouse_store.rb
vincent-pochet: Good catch, I'll give it a try!

PR #4233, opened 2025-09-01
bug: rename portuguese-brasil locale code

ansmonjol:
## Context

A new `pt_BR` locale have been added recently.
We can save this informations in the customer, the billing entity and the organization
It sets the locale for translations displayed in the transactional emails and the customer portal.

However this locale does not follow the BCP 47 language syntax that expect the language and the extended language tag to be separated by a hyphen `-` instead of an underscore `_`.

## Description

This PR renames the locale to be `pr-BR` and updated the code accordingly.
A migration is present to make sure all data saved follows this new notation.

Fixes https://github.com/getlago/lago-front/issues/2420 and ISSUE-1048
Related Frontend PR: https://github.com/getlago/lago-front/pull/2431

### DIFF
--- a/db/migrate/20250901141844_update_document_locale_for_pt_br.rb
+++ b/db/migrate/20250901141844_update_document_locale_for_pt_br.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class UpdateDocumentLocaleForPtBr < ActiveRecord::Migration[8.0]
+  def up
+    Organization.where(document_locale: "pt_BR").update(document_locale: "pt-BR")
+    BillingEntity.where(document_locale: "pt_BR").update(document_locale: "pt-BR")
+    Customer.where(document_locale: "pt_BR").update(document_locale: "pt-BR")
+  end
+
+  def down
+    Organization.where(document_locale: "pt-BR").update(document_locale: "pt_BR")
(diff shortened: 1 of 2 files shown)


db/migrate/20250901141844_update_document_locale_for_pt_br.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class UpdateDocumentLocaleForPtBr < ActiveRecord::Migration[8.0]
+  def change
```
groyoh:
```suggestion
  def up
```

db/migrate/20250901141844_update_document_locale_for_pt_br.rb
```diff
@@ -0,0 +1,9 @@
...
+    Organization.where(document_locale: "pt_BR").update(document_locale: "pt-BR")
```
groyoh, line 5: You may want to use `find_each` here depending on the affected numbers.

mariohd:
```
lago-api(prod)> Organization.where(document_locale: "pt_BR").count
=> 0
lago-api(prod)> BillingEntity.where(document_locale: "pt_BR").count
=> 1
lago-api(prod)> Customer.where(document_locale: "pt_BR").count
=> 0
```

PR #4234, opened 2025-09-01
fix(date): Fix ISO8601 date and datetime parsing

groyoh:
## Context

When creating a one-off invoice, sending certain date format such as `2025-06-21 10:00:00Z` will result in a 500 due to the fact `Utils::Datetime.valid_format?` returns `true` but we later try and parse it into an ISO8601 date time:

```ruby
lago-api(dev)> Utils::Datetime.valid_format?("2025-06-21 10:00:00Z")
=> true
lago-api(dev)> DateTime.strptime("2025-06-21 10:00:00Z")
(lago-api):2:in 'DateTime.strptime': invalid date (Date::Error)
        from (lago-api):2:in '<main>'
lago-api(dev)> DateTime.strptime("2025-06-21T10:00:00Z")
=> Sat, 21 Jun 2025 10:00:00 +0000
```

## Description

This update does two things:

1. Replaces all calls to `Date.strptime`/`DateTime.strptime` with the `"%Y-%m-%dT%H:%M:%S.%LZ"` format or the default format by `Date.iso8601`/`DateTime.iso8601`. This make the date/datetime parsing more explicit and allows to skip the format parameter altogether.
2. Allows to provide the `:format` keyword parameter to the `Utils::Datetime.valid_format?` which defaults to `:iso8601`. This ensures that all future code depending on it will expect datetimes in ISO8601 format. All existing code expecting any datetime format was updated with `format: :any`.


groyoh:
Thanks @vincent-pochet @julienbourdeau for the review 🙏

PR #4236, opened 2025-09-01
feat(pre-aggregation): Implement count aggregation

vincent-pochet:
## Context

This PR is part of the Pre-aggregation epic.

The main goal of this feature is to setup a complete aggregation pipeline that will be leveraged to compute the customer usage without re-querying the full set of events.

## Description

This PR
- The basic skeleton of the new `Events::Store::AggregatedClickhouseStore` event store
- Implements the `count` and `grouped_count` aggregations
- Fixes the migration for the `events_aggregated_mv` materialized view

PR #4237, opened 2025-09-02
feat(pre-aggregation): Implement sum aggregation

vincent-pochet:
## Context

## Description

This PR implements the `sum` and `grouped_sum` aggregations


app/services/events/stores/aggregated_clickhouse_store.rb
```diff
@@ -127,11 +127,30 @@ def grouped_sum_precise_total_amount_cents
...
+        connection_with_retry do |connection|
+          sql = aggregated_events_sql(select: [
+            to_decimal128(Arel::Nodes::NamedFunction.new(
+              "sumMerge",
+              [aggregated_arel_table[:sum_state]]
+            )).as("sum_value")
+          ])
+
+          connection.select_value(sql)
+        end
```
groyoh: I don't know about the other PR but I kinda feel like we could have a generic:

```ruby
def merge_aggregation(aggregation_type, column, as: nil)
  connection_with_retry do |connection|
    sql = aggregated_events_sql(select: [
      to_decimal128(Arel::Nodes::NamedFunction.new(
        "#{aggregation_type}Merge",
        [aggregated_arel_table[column]]
      )).as(as || column)
    ])
  
    connection.select_value(sql)
end

def sum
  merge_aggregation(:sum, :sum_state, as: :sum_value)
end

**Review by groyoh — APPROVED** [review]

File: app/services/events/stores/aggregated_clickhouse_store.rb [hunk 2]
```diff
@@ -205,6 +186,33 @@ def to_decimal128(value)
...
         Arel::Nodes::SqlLiteral.new("#{attribute.relation.name}.#{attribute.name}::JSON")
       end
+
+      def merge_aggregation(aggregation_type, column, as)
+        connection_with_retry do |connection|
+          sql = aggregated_events_sql(select: [
+            to_decimal128(Arel::Nodes::NamedFunction.new(
+              aggregation_type,
+              [aggregated_arel_table[column]]
+            )).as(as)
```
groyoh, line 196: Also just realized we might not even need to specify `.as` since we retrive by index and not by column name/alias.

vincent-pochet: it's not mandatory but it helps with readability. I prefere to keep it for now as I know I will have to extend the select in future queries (for proration especially)

PR #4240, opened 2025-09-02
fix(sequence-error): Add wait, attempts and jitter to retry on Sequenced::SequenceError

ivannovosad:
## Context

On billing day there are sometimes invoices stuck in `generating` status. 
The reason is that they raise Sequenced::SequenceError and the jobs are retried very shortly after they fail and they end up in the `Dead` queue.
These invoices usually belong to an organization with `per_organization` document numbering.

## Description

This PR mitigates the issue by changing the `retry_on` from:
```
retry_on Sequenced::SequenceError
```
to:
```
retry_on Sequenced::SequenceError, wait: :polynomially_longer, attempts: 15, jitter: 0.75
```

### DIFF
--- a/app/jobs/bill_subscription_job.rb
+++ b/app/jobs/bill_subscription_job.rb
@@ -14,6 +14,8 @@ class BillSubscriptionJob < ApplicationJob
   retry_on Sequenced::SequenceError, ActiveJob::DeserializationError, wait: :polynomially_longer, attempts: 15, jitter: 0.75
 
   def perform(subscriptions, timestamp, invoicing_reason:, invoice: nil, skip_charges: false)
+    Rails.logger.info("BillSubscriptionJob[Invoice ID: #{invoice&.id}] - Started")
+
     result = Invoices::SubscriptionService.call(
       subscriptions:,
       timestamp:,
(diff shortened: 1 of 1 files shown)


julienbourdeau:
For reference, this is the delay after first try. There is some randomness so this is just about the order of magnitude

```
Try 1 ==> 1 seconds (after the first try)
Try 2 ==> 17 seconds (after the first try)
Try 3 ==> 2 minutes (after the first try)
Try 4 ==> 7 minutes (after the first try)
Try 5 ==> 21 minutes (after the first try)
Try 6 ==> 51 minutes (after the first try)
Try 7 ==> 1 hours (after the first try)
Try 8 ==> 2 hours (after the first try)
Try 9 ==> 3 hours (after the first try)
Try 10 ==> 4 hours (after the first try)
Try 11 ==> 7 hours (after the first try)
Try 12 ==> 13 hours (after the first try)
Try 13 ==> 19 hours (after the first try)
Try 14 ==> 1 days (after the first try)
Try 15 ==> 1 days (after the first try)
```

```rb
func = ->(executions, jitter = 0.75) { ((executions**4) + (Kernel.rand * (executions**4) * jitter)) }

prev = 0
(1..15).each do |loop|
  sec = func.call(loop).to_i
  sec_since_first_try = prev + sec

  delay = if sec_since_first_try < 60
      "#{sec_since_first_try} seconds"
    elsif sec_since_first_try < 3600
      "#{sec_since_first_try / 60} minutes"
    elsif sec_since_first_try < 86400
      "#{sec_since_first_try / 3600} hours"
    else
      "#{sec_since_first_try / 86400} days"
  end

  puts "Try #{loop} ==> #{delay} (after the first try)"

  prev = sec
end

```

app/jobs/bill_subscription_job.rb
```diff
@@ -11,27 +11,42 @@ class BillSubscriptionJob < ApplicationJob
...
     )
-    return if result.success?
+
+    if result.success?
+      Rails.logger.info("BillSubscriptionJob[Invoice ID: #{invoice&.id}] - Finished [SUCCESS]")
+      return
+    end
+
+    Rails.logger.info("BillSubscriptionJob[Invoice ID: #{invoice&.id}] - Before reload [#{result.invoice.inspect}]")
+    result.invoice.reload
```
floganz: I'd suggest keeping `&.` here as in if just in case if invoice is not set by service above.

floganz (approved):

PR #4242, opened 2025-09-02
feat(pre-aggregation): Implement events_values

vincent-pochet:
## Context

## Description

This PR implements:
- the base "events" method, relying on the Clickhouse events_enriched_expanded allowing use to filter events using `charge_id`, `charge_filter_id` and `grouped_by`
- the `events_values` and `prorated_events_values` logic
- a new `distinct_charge_filter_ids` that will later allow a pre-filtering the aggregation queries to speedup the computation

### DIFF
--- a/app/services/events/stores/aggregated_clickhouse_store.rb
+++ b/app/services/events/stores/aggregated_clickhouse_store.rb
@@ -14,11 +14,11 @@ module Events
             .where(charge_filter_id: charge_filter_id || "")
 
           scope = scope.order(timestamp: :asc) if ordered
-          scope.where(timestamp: from_datetime..) if force_from || use_from_boundary
-          scope.where(timestamp: ..to_datetime) if to_datetime
+          scope = scope.where(timestamp: from_datetime..) if force_from || use_from_boundary
+          scope = scope.where(timestamp: ..to_datetime) if to_datetime
 
           scope = if grouped_by_values?
-            scope.where(sorted_grouped_by: formated_grouped_by_values)
+            scope.where("toJSONString(sorted_grouped_by) = ?", formated_grouped_by_values)
           else
             # TODO: take grouped by into account when no grouped_by_values
             scope.where(sorted_grouped_by: "{}")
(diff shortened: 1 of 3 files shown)


app/services/events/stores/aggregated_clickhouse_store.rb
```diff
@@ -29,12 +48,33 @@ def aggregated_events_sql(force_from: false, select: aggregated_arel_table[Arel.
...
 
-      def events_values
-        # TODO(pre-aggregation): Implement
+      def distinct_charge_filter_ids
+        ::Clickhouse::EventsEnrichedExpanded
+          .where(subscription_id: subscription.id)
+          .where(organization_id: subscription.organization_id)
+          .where(timestamp: from_datetime..to_datetime)
+          .pluck("DISTINCT(charge_filter_id)")
+          .reject(&:blank?)
```
groyoh: I guess this could be done in SQL directly.

app/services/events/stores/aggregated_clickhouse_store.rb
```diff
@@ -151,7 +201,7 @@ def formated_grouped_by_values
           .transform_values { |value| value || NIL_GROUP_VALUE }
           .sort_by { |key, _| key }
           .to_h
-          .to_json
+          .to_json(escape_html_entities: false)
```
groyoh: Could you explain this pleae ? Maybe leave a comment ?

spec/services/events/stores/aggregated_clickhouse_store_spec.rb
```diff
@@ -123,6 +128,117 @@
...
+
+  describe ".distinct_charge_filter_ids" do
+    it "returns an empty array of when no charge filters are present" do
+      expect(event_store.distinct_charge_filter_ids).to be_empty
+    end
+
+    context "when charge filters are present" do
+      let(:charge_filter) { create(:charge_filter, charge:) }
+
+      it "returns an array of distinct charge filter ids" do
```
groyoh, line 194: It might be interesting to have at least two charges.

spec/services/events/stores/aggregated_clickhouse_store_spec.rb
```diff
@@ -387,4 +503,102 @@
...
     end
   end
+
+  describe ".active_unique_property?" do
+    before { event_store.aggregation_property = billable_metric.field_name }
+
+    it "returns false when no previous events exist" do
+      bm_value = SecureRandom.uuid
+
+      event = ::Clickhouse::EventsEnriched.create!(
```
groyoh: Shouldn't this be `EventsEnrichedExpanded` ? 🤔 Same below.

groyoh (approved):
I'm not sure to fully understand how those functions are used but it seems to match the other clickhouse store. There's a few duplication between the two Clickhouse store though.

Also, as mentioned in the previous PRs, it would be nice to make the tests for robust. For instance, `events` is not fully tested.

PR #4244, opened 2025-09-03
chore(result): extract result behavior to module

julienbourdeau:
When BaseResult was introduced, the behavior was copy pasted from LegacyResult since this class was supposed to go away. Many months later, less than half of results have been migrated from LegacyResult to BaseResult.

The behavior evolves ([like here](https://github.com/getlago/lago-api/pull/3653)) and as we add method or arguments, we have to add them twice.

This PR extract the Result behavior into a module, which is included in both.

It's done now because I have a few change on validation failure coming up.


groyoh (approved):
Maybe you could add a shared example and test the legacy result as well ?

PR #4246, opened 2025-09-03
fix(tests): Fix flaky test on usage

jdenquin:


groyoh (approved):

groyoh:
@jdenquin which test is this meant to fix BTW ? If it's `spec/graphql/resolvers/customer_portal/customers/projected_usage_resolver_spec.rb`, I believe the issue comes from the test setup. The billable metrics don't have fixed names (randomly generated from the factory) and therefore are not ordered in a consistent way.

jdenquin:
let me add fixed name!

groyoh (approved):
I guess you can revert the service change (or not 😄). I'm not even sure if there's a reason to sort by update_at/created_at.

PR #4248, opened 2025-09-04
fix(clickhouse): Ensure presence of AR internal tables

vincent-pochet:
## Context

Creating the initial database structure fails with clickhouse raising the following error:

```
lago-migrate_dev  | ActiveRecord::ActiveRecordError: Response code: 404: (ActiveRecord::ActiveRecordError)
lago-migrate_dev  | Code: 60. DB::Exception: Table default.schema_migrations does not exist. (UNKNOWN_TABLE) (version 25.4.13.22 (official build))
lago-migrate_dev  |
```

## Description

The root cause of the issue is that the `clickhouse-activerecord` gem is not exporting these tables when the `config.active_record.schema_format` app config is set to `:sql`

See https://github.com/PNixx/clickhouse-activerecord/issues/138 for more details

This PR fixes the migration for Clickhouse database by setting `schema_dump` to false for clickhouse database. It's not mandatory as the setup in local, test and CI environment all uses the migrations to create the DB structure rather the structure file to manage the database environment.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -26,6 +26,7 @@ development:
     migrations_paths: db/clickhouse_migrate
     debug: true
     database_tasks: <% if ENV['LAGO_CLICKHOUSE_MIGRATIONS_ENABLED'].present? %> true <% else %> false <% end %>
+    schema_dump: false
 
 test:
   primary:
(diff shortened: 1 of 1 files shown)


groyoh:
I've actually struggled with this one quite a while. One way I managed to resolved this is by setting `LAGO_DISABLE_SCHEMA_DUMP="true"`, when creating/migrate the clickhouse database. If you're trying to fix the dev env, I believe we could have a `LAGO_DISABLE_CLICKHOUSE_SCHEMA_DUMP` and set it to true by default.

If we don't go with the above solution, I'd suggest re-using AR method:

```ruby
Clickhouse::BaseRecord.connection_pool.schema_migration.create_table
Clickhouse::BaseRecord.connection_pool.internal_metadata.create_table
```

PR #4256, opened 2025-09-05
feat(wallet-transaction): Add a `name` to wallet transactions

groyoh:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/define-the-name-of-a-wallet-transaction-at-top-up

## Context

When creating a wallet transaction, the name displayed on the invoice defaults to the wallet name. However, for organizations managing multiple types of top-ups, it would be beneficial to display specific naming on invoices that corresponds to the particular transaction being made.

## Description 

This change adds a `name` attribute to the wallet transaction models, params and serializers.

### DIFF
--- a/app/services/wallet_transactions/void_service.rb
+++ b/app/services/wallet_transactions/void_service.rb
@@ -9,7 +9,8 @@ module WalletTransactions
         :source,
         :metadata,
         :priority,
-        :credit_note_id
+        :credit_note_id,
+        :name
       )
 
       super
--- a/spec/requests/api/v1/wallet_transactions_controller_spec.rb
+++ b/spec/requests/api/v1/wallet_transactions_controller_spec.rb
@@ -27,7 +27,8 @@ RSpec.describe Api::V1::WalletTransactionsController, type: :request do
       {
         wallet_id:,
         paid_credits: "10",
-        granted_credits: "10"
+        granted_credits: "10",
+        name: "Custom Top-up Name"
       }
     end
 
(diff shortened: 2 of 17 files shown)


app/services/wallet_transactions/void_service.rb
```diff
@@ -21,7 +21,8 @@ def call
           transaction_type: :outbound,
           status: :settled,
           settled_at: Time.current,
-          transaction_status: :voided
+          transaction_status: :voided,
+          name:
```
floganz: What's the case where `**transaction_params` wont succeed with passing `name` to `CreateService`?
For what case we need this method?
```ruby
def name
  transaction_params[:name].presence
end
```

spec/requests/api/v1/wallet_transactions_controller_spec.rb
```diff
@@ -47,6 +49,19 @@
       expect(json[:wallet_transactions].second[:lago_wallet_id]).to eq(wallet.id)
     end
 
+    context "with name parameter" do
```
floganz: I guess we can simplify tests here. We already have all cases tested on a service level.
So it should be enough to only test, let's say that data assigned case in existing test `it "creates a wallet transactions" do`. Without introducing new context and test.

app/services/wallet_transactions/void_service.rb
groyoh: You're right. It's not needed. I completely overlooked the `transaction_params` as I usually expect splat kwargs to be the last parameter and not first.

spec/requests/api/v1/wallet_transactions_controller_spec.rb
groyoh: Sure. To be honest, those were generated by Cursor 😄

app/services/wallet_transactions/void_service.rb
floganz: When I was adding it my concern was that these params shouldn't override the value we hardcode after them. That's why in several I've decided to put it first but maybe it's not so common pattern to do.

groyoh: @floganz I guess it depends on what you expect in the params. I usually prefer to sanitize it beforehand:

```ruby
sliced_params = params.slice(:arg_1, :arg_2)
my_function(arg_3:, **sliced_params)
```

This way, you're sure that it is not overwritten.

But having it before the other params works just as well. I'm just not used to it. 😉

groyoh:
Thanks for the review @floganz 🙏

PR #4257, opened 2025-09-05
feat(fixed_charges): Emit fixed charge events on subscription creation

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

This change creates Fixed Charge Events on subscription creation and activation, these events will be aggregated on the billing run.

### DIFF
--- a/app/graphql/types/fixed_charges/input.rb
+++ b/app/graphql/types/fixed_charges/input.rb
@@ -5,6 +5,8 @@ module Types
     class Input < Types::BaseInputObject
       graphql_name "FixedChargeInput"
 
+      argument :id, ID, required: false
+
       argument :add_on_id, ID, required: true
       argument :apply_units_immediately, Boolean, required: false
       argument :charge_model, Types::FixedCharges::ChargeModelEnum, required: true
--- a/app/graphql/types/subscriptions/fixed_charge_overrides_input.rb
+++ b/app/graphql/types/subscriptions/fixed_charge_overrides_input.rb
@@ -3,13 +3,14 @@
 module Types
   module Subscriptions
     class FixedChargeOverridesInput < Types::BaseInputObject
-      argument :add_on_id, ID, required: true
+      argument :add_on_id, ID, required: false
       argument :id, ID, required: false
 
+      argument :apply_units_immediately, Boolean, required: false
       argument :invoice_display_name, String, required: false
       argument :properties, Types::FixedCharges::PropertiesInput, required: false
       argument :tax_codes, [String], required: false
-      argument :units, String, required: true
+      argument :units, String, required: false
     end
   end
 end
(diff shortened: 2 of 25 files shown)


app/services/subscriptions/organization_emit_fixed_charge_events_service.rb
```diff
@@ -0,0 +1,341 @@
...
+      @organization = organization
+      @timestamp = timestamp
+      super
+    end
+
+    def call
+      eligible_subscriptions_by_customer.each do |_customer_id, customer_subscriptions|
+        emitting_subscriptions = []
+        customer_subscriptions.each do |subscription|
+          next if subscription.next_subscription&.pending?
```
ancorcruz: this condition does not apply for emitting fixed charges events.

ancorcruz: @annvelents after checking this again, it makes sense we skip the subscription (not emit new fixed charge events for it) because the subscription won't continue for the billing period as it will be terminated and the pending one started on this billing run.

This will happen in `Subscriptions::TerminateService`.

However, we need to ensure the events for the next subscription that will become active are emitted...

Following our previous example:

```
Given a plan, monthly - calendar,
And plan has fixed charges (paid in arrears)

And customer starts a subscription (subs_1) at 22nd Sept
Then we emit Fixed Charge Events on subscription creation for the plan's fixed charge

When subscription subs_1 is downgraded on 25th Sept
Then we create a new pending NEXT subscription (subs_2)
And we emit events for the Fixed Charge (shall we? or should this happen when the subscription subs_2 becomes active?)

When 1st Oct ticks
Then we TERMINATE subscription subs_1 billing the subscription of September
And we bill emitted fixed charge events of subs_1 in September (creation time + others 22nd Sept - 30 Sept)
And we DO NOT emit new fixed charge events for the old subscription subs_1
And we activate subscription subs_2
And we emit fixed charge events for subs_2
```

Then this condition will skip emitting fixed charge events for a subscription that is going to be terminated on the next billing run as we emit FCE at the beginning of the subscription period (billing happens at the end).

Emitting fixed charge events on subscription activation makes more sense than on creation
also on update we should only emit events for active subscriptions.

does it make sense for you?

app/jobs/subscriptions/organization_emit_fixed_charge_events_job.rb
```diff
@@ -0,0 +1,22 @@
...
+        :clock
+      end
+    end
+
+    unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
+
+    def perform(organization:, timestamp: Time.current.to_i)
+      Subscriptions::OrganizationEmitFixedChargeEventsService.call!(
+        organization:,
+        timestamp: Time.zone.at(timestamp || Time.current.to_i)
```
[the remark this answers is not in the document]
ancorcruz: this is not true.

PR #4265, opened 2025-09-08
feat(wallets): Add Min and Max paid top up limits

julienbourdeau:
- [x] Wallets has 2 new attributes to set a upper and lower limit to wallet transactions.
- [x] Recurring rules have a boolean attribute to bypass this limitation.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -39,7 +39,7 @@ en:
             attributes:
               permissions:
                 forbidden_key: 'contains forbidden keys: %{keys}'
-        must_be_greater_than_min: must_be_greater_than_min
+        must_be_greater_than_or_equal_min: must_be_greater_than_or_equal_min
         not_compatible_with_aggregation_type: not_compatible_with_aggregation_type
         not_compatible_with_pay_in_advance: not_compatible_with_pay_in_advance
         one_of_plan_or_subscription_required: one_of_plan_or_subscription_required
--- a/spec/models/wallet_spec.rb
+++ b/spec/models/wallet_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe Wallet, type: :model do
       subject.paid_top_up_max_amount_cents = 1
 
       expect(subject).not_to be_valid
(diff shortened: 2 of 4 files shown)


spec/models/wallet_spec.rb
```diff
@@ -13,6 +13,20 @@
...
     it { is_expected.to validate_numericality_of(:rate_amount).is_greater_than(0) }
+    it { is_expected.to validate_numericality_of(:paid_top_up_min_amount_cents).is_greater_than(0).allow_nil }
+    it { is_expected.to validate_numericality_of(:paid_top_up_max_amount_cents).is_greater_than(0).allow_nil }
+
+    it "validates than max is greater than min" do
+      subject.paid_top_up_min_amount_cents = 100
+      subject.paid_top_up_max_amount_cents = 1
+
+      expect(subject).not_to be_valid
+      expect(subject.errors["paid_top_up_max_amount_cents"]).to eq ["must_be_greater_than_min"]
```
floganz: Validation allows max value to be equal to the min. Should we also reflect that behaviour in error code?
Like `greater_than_or_equal` or something like that. If you agree then also do not forget to update locale file.

app/services/wallets/recurring_transaction_rules/create_service.rb
```diff
@@ -32,6 +32,10 @@ def call
           transaction_metadata: rule_params[:transaction_metadata] || []
         }
 
+        if rule_params.key? :ignore_paid_top_up_limits
```
floganz, line 35: These changes are not covered with tests?

floganz (approved):
It looks like there're no tests for changes in GQL input files.

Otherwise all looks good

config/locales/en.yml
```diff
@@ -39,6 +39,7 @@ en:
             attributes:
               permissions:
                 forbidden_key: 'contains forbidden keys: %{keys}'
+        must_be_greater_than_or_equal_min: must_be_greater_than_or_equal_min
```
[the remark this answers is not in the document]
julienbourdeau: It's intended

PR #4267, opened 2025-09-08
fix(coupon-with-plans): Allow a coupon to be created when a plan has …

mariohd:
## Scenario
Whenever the user tries to create a coupon for some plans, it fails if one of the plans has child plans.

## Problem
Right here https://github.com/getlago/lago-api/blob/main/app/services/coupons/create_service.rb#L41, we check by the plans count. 
But if any of the plans has a child with the same code, it will fail. 
```
"applies_to": {"plan_codes": ["plan_1"]}
```

```
Plan.where(code: ["plan_1"] , organization_id: 'some-org').count
=> 2
```

## Fix
Instead of counting, have the groups keys reduced and expect it to be empty. 
```
      if plan_identifiers.present? && (plan_identifiers - plans.pluck(plan_key)).present?
        return result.not_found_failure!(resource: "plans")
      end
```

### DIFF
--- a/app/services/coupons/create_service.rb
+++ b/app/services/coupons/create_service.rb
@@ -38,7 +38,7 @@ module Coupons
         reusable:
       )
 
-      if plan_identifiers.present? && plans.count != plan_identifiers.count
+      if plan_identifiers.present? && (plan_identifiers - plans.pluck(plan_key)).present?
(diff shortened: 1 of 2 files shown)


mariohd:
@groyoh 

Yes, it will allow the coupon to be applied to child plans as well, after all, it has the same code.

PR #4268, opened 2025-09-09
fix(wallet-transaction): Set `invoice_requires_successful_payment` to `false` for granted credits

groyoh:
## Context

Granted credit never triggers payment not generates invoices. So their `invoice_requires_successful_payment` attribute should always be set to `false`.

## Description

This ensures that it is the case.


julienbourdeau (approved):

groyoh:
Thanks for the review 🙏

PR #4274, opened 2025-09-09
feat(wallets): Add Min and Max validation on wallet creation

julienbourdeau:
Introduce validation on wallet creation

### DIFF
--- a/app/services/wallets/create_service.rb
+++ b/app/services/wallets/create_service.rb
@@ -39,6 +39,8 @@ module Wallets
 
       wallet = Wallet.new(attributes)
 
+      return result unless wallet_amount_valid?(wallet)
+
       ActiveRecord::Base.transaction do
         if params[:currency].present?
           Customers::UpdateCurrencyService.call!(customer: result.current_customer, currency: params[:currency])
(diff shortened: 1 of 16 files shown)


spec/requests/api/v1/wallets_controller_spec.rb
```diff
@@ -26,14 +26,16 @@
         expiration_at:,
         invoice_requires_successful_payment: true,
         paid_top_up_min_amount_cents: 5_00,
-        paid_top_up_max_amount_cents: 100_00
+        paid_top_up_max_amount_cents: 100_00,
+        ignore_paid_top_up_limits_on_creation: "false"
```
floganz: I'd suggest to pass here "true" and have `paid_credits` above or below limits.
Then we can check that wallet transactions are saved (behaviour) instead of testing allow/expect (implementation) now.

floganz (dismissed):

app/services/wallets/create_service.rb
```diff
@@ -86,6 +88,15 @@ def valid?
       Wallets::ValidateService.new(result, **params).valid?
     end
 
+    def wallet_amount_valid?(wallet)
+      Validators::WalletTransactionAmountLimitsValidator.new(
+        result,
+        wallet:,
+        credits_amount: ActiveModel::Type::Integer.new.cast(params[:paid_credits]),
```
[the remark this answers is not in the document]
julienbourdeau: This is a good point, the paid amout can be decimal!

app/services/wallets/create_service.rb
```diff
@@ -47,6 +47,8 @@ def call
         wallet.currency = wallet.customer.currency
         wallet.save!
 
+        validate_wallet_amount! wallet
```
julienbourdeau, line 50: I moved this back to the transaction because right above, we update the currency if it's part of the params. Our first version (before the transaction) could be using the "old" currency.

PR #4275, opened 2025-09-09
fix(activity-log+api-log): Fix date filters in API Logs and Activity Logs resolvers

groyoh:
## Context

Offset pagination is fragile: if new record arrive between fetching page 1 and page 2, rows shift and clients either miss or re‑read items. This is what happened with the Activity Logs and API Logs feature for which we create new records more often that other features.

In order to mitigate this issue, the frontend is now always sending the `to_date` filter with the following rules:

1. If the filter is not set, we default to the timestamp at which we loaded the data for the first time
2. If the filter is set, we make sure that the value cannot be later than the timestamp at which we loaded the data for the first time

This ensures that no matter the requested page, we will never receive newer records.

The issue is that the backend GraphQL resolver did not accept `to_date` as a `ISO8601DateTime` but as a `ISO8601Date`, causing the backend to wrongly filter out all of today's data.

## Description

This fixes it by accepting both date and datetime as input for `to_date/from_date`.

It also adds tests for the resolvers filtering.


app/graphql/resolvers/api_logs_resolver.rb
```diff
@@ -12,8 +12,8 @@ class ApiLogsResolver < Resolvers::BaseResolver
     argument :limit, Integer, required: false
     argument :page, Integer, required: false
 
-    argument :from_date, GraphQL::Types::ISO8601Date, required: false
-    argument :to_date, GraphQL::Types::ISO8601Date, required: false
+    argument :from_date, GraphQL::Types::ISO8601DateTime, required: false
+    argument :to_date, GraphQL::Types::ISO8601DateTime, required: false
```
rsempe: can we rename them to from_datetime and to_datetime in that case?

PR #4276, opened 2025-09-09
test(subscriptions): Add more scenario tests for subscription termination

groyoh:
## Context

18ae6454b7eb28291b021ec797d512843e0f3216 introduce a way to skip the generation of invoice on subscription termination but was not tested via scenario tests. 

## Description

This adds missing scenario tests for subscription termination.

This also adds a few things to the test setup and helpers:

- `api_call` will now raise an error on 4xx and 5xx status code:

     ```ruby
    RuntimeError:
      API call failed:
      - Method: POST
      - Path: /api/v1/payments
      - Request body: {"payment":{"invoice_id":"0e83db9c-5e98-4e65-9469-e8626f8ecec6","amount_cents":7500,"reference":"05c4d208-fede-486d-a91b-fc25e5a40c19"}}
      - HTTP status: 403
      - Response body: {"status":403,"error":"Forbidden","code":"feature_unavailable"}
    ```
 - `api_call` and all scenario helpers now accept a `perform_jobs` and `raise_on_error` keyword arguments to provide more flexibility and reusability
 - a `:premium` spec metadata was added to avoid the repetition of `around { |test| lago_premium!(&test) }`

### DIFF
--- a/spec/services/invoices/regenerate_from_voided_service_spec.rb
+++ b/spec/services/invoices/regenerate_from_voided_service_spec.rb
@@ -11,9 +11,9 @@ describe "Regenerate From Voided Invoice Scenarios", :scenarios, type: :request 
   let(:subscription) do
     travel_to(DateTime.new(2023, 1, 1)) do
       create_subscription(
-        external_customer_id: customer.external_id,
-        external_id: "sub_#{customer.external_id}",
-        plan_code: plan.code
+        {external_customer_id: customer.external_id,
+         external_id: "sub_#{customer.external_id}",
+         plan_code: plan.code}
       )
     end
 


spec/spec_helper.rb
```diff
@@ -199,4 +199,12 @@
...
     end
   end
+
+  config.around do |example|
+    if example.metadata[:premium]
+      lago_premium!(&example)
+    else
+      example.run
+    end
+  end
```
julienbourdeau, line 209: awesome. I hate `around { |test| lago_premium!(&test) }` :D

spec/support/scenarios_helper.rb
```diff
@@ -92,9 +105,9 @@ def create_subscription(params, authorization = nil)
     end
   end
 
-  def terminate_subscription(subscription)
-    api_call do
-      delete_with_token(organization, "/api/v1/subscriptions/#{subscription.external_id}")
+  def terminate_subscription(subscription, params: {}, perform_jobs: true, raise_on_error: true)
+    api_call(perform_jobs:, raise_on_error:) do
```
julienbourdeau: I'm thinking that it maybe the last arg of all method should be a hash forwarded to api_call. I agree with adding the flags but I think it's annoying to repeat them on all method.

Typically here, only this `terminate_subscription` can receive `perform_jobs` until we add it (and duplicate default value) to all methods

something like this. There is probably a more elegant way but you get the idea.

```ruby
def terminate_subscription(subscription, params: {}, **args)
  api_call(args) do
    delete_with_token(organization, "/...")
  end
end

def api_call(args)
  param = defaults.merge(args)
  ...
end
```

groyoh: Sure, I can do that.

groyoh:
Thanks for the review @julienbourdeau !

spec/support/scenarios_helper.rb
groyoh: Done!

PR #4277, opened 2025-09-09
feat(pre-aggregation): Implement weighted_sum

vincent-pochet:
## Context

## Description

This PR implements:
- The `unique_count` aggregation
- The `weighted_sum` aggregation
- The `weighted_sum_breakdown` method

### DIFF
--- a/app/services/events/stores/aggregated_clickhouse/weighted_sum_query.rb
+++ b/app/services/events/stores/aggregated_clickhouse/weighted_sum_query.rb
@@ -25,7 +25,7 @@ module Events
             #{grouped_events_cte_sql(initial_values)}
 
             SELECT
-              grouped_by,
+              grouped_by::JSON,
               SUM(period_ratio) as aggregation
             FROM (
               SELECT
@@ -142,46 +142,26 @@ module Events
 
         def grouped_initial_value_sql(initial_values)
           values = initial_values.map do |initial_value|
-            groups = store.grouped_by.map do |g|
-              "'#{ActiveRecord::Base.sanitize_sql_for_conditions(initial_value[:groups][g])}'"
-            end
-
             [
-              groups,
+              "'#{formated_groupes_values(initial_value)}'",
               "toDateTime64(:from_datetime, 5, 'UTC')",
               "toDecimal128(#{initial_value[:value]}, :decimal_scale)"
-            ].flatten.join(", ")
+            ].join(", ")
           end
 
-          <<-SQL
-            SELECT
-              #{store.grouped_by.map.with_index { |_, index| "tuple.#{index + 1} AS g_#{index}" }.join(", ")},
-              tuple.#{store.grouped_by.count + 1} AS timestamp,
-              tuple.#{store.grouped_by.count + 2} AS difference
-            FROM ( SELECT arrayJoin([#{values.map { "tuple(#{it})" }.join(", ")}]) AS tuple )
-          SQL
+          tuple_select_sql(values)
         end
 
         def grouped_end_of_period_value_sql(initial_values)
           values = initial_values.map do |initial_value|
-            groups = store.grouped_by.map do |g|
-              "'#{ActiveRecord::Base.sanitize_sql_for_conditions(initial_value[:groups][g])}'"
-            end
-
             [
-              groups,
+              "'#{formated_groupes_values(initial_value)}'",
               "toDateTime64(:to_datetime, 5, 'UTC')",
               "toDecimal128(0, :decimal_scale)"
-            ].flatten.join(", ")
+            ].join(", ")
           end
 
-          <<-SQL
-            SELECT
-              #{store.grouped_by.map.with_index { |_, index| "tuple.#{index + 1} AS g_#{index}" }.join(", ")},
-              tuple.#{store.grouped_by.count + 1} AS timestamp,
-              tuple.#{store.grouped_by.count + 2} AS difference
-            FROM ( SELECT arrayJoin([#{values.map { "tuple(#{it})" }.join(", ")}]) AS tuple )
-          SQL
+          tuple_select_sql(values)
         end
 
         def grouped_period_ratio_sql
@@ -202,6 +182,24 @@ module Events
               -- NOTE: duration was null so usage is null
               toDecimal128(0, :decimal_scale)
             )
+          SQL
+        end
+
+        def formated_groupes_values(initial_value)
+          store.grouped_by
+            .index_with { initial_value[:groups][it] || store.class::NIL_GROUP_VALUE }
+            .sort_by { |key, _| key }
+            .to_h
+            .to_json(escape_html_entities: false)
+        end
+
+        def tuple_select_sql(values)
+          <<-SQL
+            SELECT
+              tuple.1 AS grouped_by,
+              tuple.2 AS timestamp,
+              tuple.3 AS difference
+            FROM ( SELECT arrayJoin([#{values.map { "tuple(#{it})" }.join(", ")}]) AS tuple )
           SQL
         end
       end
--- a/app/services/events/stores/aggregated_clickhouse/weighted_sum_query.rb
+++ b/app/services/events/stores/aggregated_clickhouse/weighted_sum_query.rb
@@ -1,0 +1,210 @@
+# frozen_string_literal: true
+
+module Events
+  module Stores
+    module AggregatedClickhouse
+      class WeightedSumQuery
+        def initialize(store:)
+          @store = store
+        end
+
+        def query
+          <<-SQL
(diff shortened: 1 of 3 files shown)


app/services/events/stores/concerns/clickhouse_sql_helpers.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Events
+  module Stores
+    module Concerns
+      module ClickhouseSqlHelpers
+        # NOTE: Compute pro-rata of the duration in days between the datetimes over the duration of the billing period
+        #       Dates are in customer timezone to make sure the duration is good
+        def duration_ratio_sql(from, to, duration)
```
groyoh: I would suggest making these module methods:

```suggestion
        def self.duration_ratio_sql(from, to, duration)
```


That way we can test it in isolation easily. WDYT ?

app/services/events/stores/concerns/clickhouse_connection.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module Events
+  module Stores
+    module Concerns
+      module ClickhouseConnection
+        MAX_RETRIES = 3
+
+        def with_retry(&)
```
groyoh: Same feedback as the other module. We can use module method instead of instance methods to make it easier to test in isolation. Also, it makes the usage for explicit:

```ruby
def my_function
  Events::Stores::ClickhouseConnection.connection_with_retry do |connection|
    # ...
  end
end
```

app/services/events/stores/concerns/clickhouse_sql_helpers.rb
```diff
@@ -0,0 +1,30 @@
...
+          from_in_timezone = date_in_customer_timezone_sql(from)
+          to_in_timezone = date_in_customer_timezone_sql(to)
+
+          "(date_diff('days', #{from_in_timezone}, #{to_in_timezone}) + 1) / #{duration}"
+        end
+
+        def date_in_customer_timezone_sql(date)
+          sql = if date.is_a?(String)
+            "toTimezone(#{date}, :timezone)"
```
groyoh: I know this code already exists but what is the `date` value you'd be expecting in this case ?

Because I tested the following:

```ruby
lago-api(dev)> date = 'toDateTime64(\'2020-01-01T01:01:01\', 5)'
=> "toDateTime64('2020-01-01T01:01:01', 5)"
lago-api(dev)> Clickhouse::BaseRecord.connection.select_one("SELECT toTimezone(#{date}, 'EST') as test")
=> {"test" => "2019-12-31 20:01:01.00000"}

lago-api(dev)> date = '2020-01-01T01:01:01Z'
=> "2020-01-01T01:01:01Z"
lago-api(dev)> Clickhouse::BaseRecord.connection.select_one("SELECT toTimezone(#{date}, 'EST') as test")
(lago-api):18:in '<main>': Response code: 400: (ActiveRecord::ActiveRecordError)
Code: 62. DB::Exception: Syntax error: failed at position 35 (:): :01Z, 'EST') as test FORMAT JSONCompactEachRowWithNamesAndTypes. Expected one of: token, DoubleColon, Comma, ClosingRoundBracket. (SYNTAX_ERROR) (version 25.4.13.22 (official build))
```

app/services/events/stores/aggregated_clickhouse/weighted_sum_query.rb
```diff
@@ -0,0 +1,210 @@
...
+              -- NOTE: full duration of the period
+              #{charges_duration.days.to_i},
+
+              -- NOTE: duration was null so usage is null
+              toDecimal128(0, :decimal_scale)
+            )
+          SQL
+        end
+
+        def formated_groupes_values(initial_value)
```
groyoh:
```suggestion
        def formated_groups_values(initial_value)
```

app/services/events/stores/aggregated_clickhouse/weighted_sum_query.rb
```diff
@@ -0,0 +1,210 @@
...
+            ].join(", ")
+          end
+
+          tuple_select_sql(values)
+        end
+
+        def grouped_end_of_period_value_sql(initial_values)
+          values = initial_values.map do |initial_value|
+            [
+              "'#{formated_groupes_values(initial_value)}'",
```
groyoh: I feel like there's a risk of SQL injection here. Having a `'` in the group might break this query.

app/services/events/stores/aggregated_clickhouse_store.rb
```diff
@@ -390,16 +394,82 @@ def sum_date_breakdown
...
       def grouped_weighted_sum(initial_values: [])
-        # TODO(pre-aggregation): Implement
+        connection_with_retry do |connection|
+          query = AggregatedClickhouse::WeightedSumQuery.new(store: self)
+
+          # NOTE: build the list of initial values for each groups
+          #       from the events in the period
+          formated_initial_values = grouped_count.map do |group|
+            value = 0
+            previous_group = initial_values.find { |g| g[:groups] == group[:groups] }
```
groyoh: Why is this called "previous" ? I'm a bit confused by the naming.

app/services/events/stores/aggregated_clickhouse_store.rb
```diff
@@ -390,16 +394,82 @@ def sum_date_breakdown
...
       end
 
       def grouped_weighted_sum(initial_values: [])
-        # TODO(pre-aggregation): Implement
+        connection_with_retry do |connection|
+          query = AggregatedClickhouse::WeightedSumQuery.new(store: self)
+
+          # NOTE: build the list of initial values for each groups
+          #       from the events in the period
+          formated_initial_values = grouped_count.map do |group|
```
groyoh: I feel like there might be a simpler way to handle this. Something like:

```ruby
groups_in_period = grouped_count.pluck(:groups)
groups_in_initial_values = initial_values.pluck(:groups)
formatted_initial_values = initial_values.dup
(groups_in_period - groups_in_initial_values).each do |groups|
  formatted_initial_values << { groups: , value: 0 }
end
```

But I might be missing something

app/services/events/stores/aggregated_clickhouse_store.rb
```diff
@@ -390,16 +394,82 @@ def sum_date_breakdown
...
+            value = 0
+            previous_group = initial_values.find { |g| g[:groups] == group[:groups] }
+            value = previous_group[:value] if previous_group
+            {groups: group[:groups], value:}
+          end
+
+          # NOTE: add the initial values for groups that are not in the events
```
groyoh: Does it make sense to add the initial values if there's no event for those groups in the period ? 🤔

app/services/events/stores/aggregated_clickhouse_store.rb
```diff
@@ -390,16 +394,82 @@ def sum_date_breakdown
...
+              }
+            ]
+          )
+
+          prepare_grouped_result(connection.select_all(sql).rows)
+        end
       end
 
       # NOTE: not used in production, only for debug purpose to check the computed values before aggregation
       def weighted_sum_breakdown(initial_value: 0)
```
groyoh, line 463: I believe this one is not tested.

spec/services/events/stores/aggregated_clickhouse_store_spec.rb
```diff
@@ -1186,4 +1186,255 @@
...
+        result = event_store.grouped_weighted_sum
+
+        expect(result.count).to eq(0)
+      end
+    end
+
+    context "with initial values" do
+      let(:initial_values) do
+        [
+          {groups: {"agent_name" => "frodo", "other" => nil}, value: 1000},
```
groyoh: It would be nice to use different initial values and different groups values to properly test the results. Here if we return the same result, in the query for all groups, the test would still pass.

groyoh (approved):

app/services/events/stores/aggregated_clickhouse_store.rb
vincent-pochet: In case of a `recurring`  billable metric, the units from the previous period will be kept and considered as consumed for the entire period.

app/services/events/stores/aggregated_clickhouse_store.rb
vincent-pochet: It's never used for "production" purpose, it's more a helper for debug. I'll still add a spec just to make sure it's "working"

app/services/events/stores/aggregated_clickhouse_store.rb
groyoh: Ah yeah, right!

PR #4278, opened 2025-09-09
feat(fixed_charges): Emit fixed charge events on updates

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

Trigger emission of fixed charge events on plan updates and subscription overrides when units are changed and user selects to apply the change immediately. Events will be emitted for all active subscriptions belonging to the fixed charge's plan.

Also, this change fixes a few bugs like;
- not able to update an existing fixed charge through graphql (UI).
-

### DIFF
--- a/app/services/fixed_charges/emit_events_for_active_subscriptions_service.rb
+++ b/app/services/fixed_charges/emit_events_for_active_subscriptions_service.rb
@@ -2,17 +2,28 @@
 
 module FixedCharges
   class EmitEventsForActiveSubscriptionsService < BaseService
-    def initialize(fixed_charge:)
+    def initialize(fixed_charge:, subscription: nil)
       @fixed_charge = fixed_charge
+      @subscription = subscription
       super
     end
 
     def call
-      fixed_charge.plan.subscriptions.active.find_each do |subscription|
+      if subscription
+        # When a specific subscription is provided, emit event for that subscription only
+        # This handles cases like plan overrides where the subscription hasn't been updated yet
         FixedCharges::EmitFixedChargeEventService.call!(
           subscription:,
           fixed_charge:
         )
+      else
+        # Default behavior: emit events for all active subscriptions on the plan
+        fixed_charge.plan.subscriptions.active.find_each do |subscription|
+          FixedCharges::EmitFixedChargeEventService.call!(
+            subscription:,
+            fixed_charge:
+          )
+        end
       end
 
       result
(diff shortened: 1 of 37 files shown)


app/services/fixed_charges/override_service.rb
```diff
@@ -31,6 +32,10 @@ def call
         end
         new_fixed_charge.save!
 
+        if params[:apply_units_immediately] && new_fixed_charge.units != fixed_charge.units
+          FixedCharges::EmitEventsForActiveSubscriptionsService.call!(fixed_charge: new_fixed_charge, subscription:)
+        end
```
ancorcruz, line 37: same here...

app/services/subscriptions/update_service.rb
```diff
@@ -105,7 +107,8 @@ def handle_plan_override
       else
         Plans::OverrideService.call(
           plan: current_plan,
-          params: params[:plan_overrides].to_h.with_indifferent_access
+          params: params[:plan_overrides].to_h.with_indifferent_access,
+          subscription:
```
annvelents, line 111: I think we do not need to pass the subscription, because overridden plan is exactly created exclusively for this subscription, and then we'll keep less context passed down

annvelents: (so from the override we can get what this override is issued for)

app/services/fixed_charges/emit_events_for_active_subscriptions_service.rb
```diff
@@ -0,0 +1,44 @@
...
+    attr_reader :fixed_charge, :subscription, :apply_units_immediately, :timestamp
+
+    def subscriptions
+      # When a specific subscription is provided, emit event for that subscription only
+      # This handles cases like plan overrides where the subscription hasn't been updated yet
+      # otherwise, emit events for all active subscriptions on the plan
+      if subscription
+        [subscription]
+      else
+        fixed_charge.plan.subscriptions.active
```
annvelents, line 36: and since for override we anyway have only one subscription associated to the plan, this will work as well for override, and we do not need to send a subscription here

ancorcruz: @annvelents we need to pass the subscription as the override happens right before the subscription is persisted with the new plan id. Hence, `fixed_charge.plan.subscriptions` will be empty.

PR #4280, opened 2025-09-09
feat(webhooks): Add organization id to webhooks payload

mariohd:
This PR adds the organization_id to all webhooks. 

## OpenAPI changes
https://github.com/getlago/lago-openapi/pull/447

### DIFF
--- a/app/services/webhooks/base_service.rb
+++ b/app/services/webhooks/base_service.rb
@@ -16,6 +16,7 @@ module Webhooks
       payload = {
         :webhook_type => webhook_type,
         :object_type => object_type,
+        :organization_id => current_organization.id,
         object_type => object_serializer.serialize
       }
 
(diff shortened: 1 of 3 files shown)


app/services/webhooks/base_service.rb
```diff
@@ -16,6 +16,7 @@ def call
       payload = {
         :webhook_type => webhook_type,
         :object_type => object_type,
+        :organization_id => current_organization&.id,
```
groyoh: I don't think `current_organization` can be `nil` given we have `return if current_organization.webhook_endpoints.none?` above:

```suggestion
        :organization_id => current_organization.id,
```

mariohd: This is here because some models might now have the organization.

i've did the same as we have here
https://github.com/getlago/lago-api/blob/c64ff0186bc5bc08303163e15459c9de4ec7ae63/app/services/webhooks/base_service.rb#L53

groyoh: But if they don't have an organization wouldn't this line raise a `NoMethodError` error beforehand: https://github.com/getlago/lago-api/blob/c64ff0186bc5bc08303163e15459c9de4ec7ae63/app/services/webhooks/base_service.rb#L14 ?

mariohd: @groyoh excellent point, i've changed that!

PR #4283, opened 2025-09-10
feat(wallets): Validate top up min/max limit for new transactions

julienbourdeau:
New wallet transaction are validated against the wallet min/max limits
- [x] API
- [x] GraphQL

### DIFF
--- a/app/services/validators/wallet_transaction_amount_limits_validator.rb
+++ b/app/services/validators/wallet_transaction_amount_limits_validator.rb
@@ -13,8 +13,7 @@ module Validators
 
     def raise_if_invalid!
       return if valid?
-      result.raise_if_error!
-      raise "invalid but no error set in result" # TODO: find generic error to raise instead
+      result.raise_if_error! # NOTE: if you didn't set the error in result, this won't raise if `valid?` is false
     end
 
     def valid?
--- a/app/services/validators/wallet_transaction_amount_limits_validator.rb
+++ b/app/services/validators/wallet_transaction_amount_limits_validator.rb
@@ -5,9 +5,16 @@ module Validators
     def initialize(result, wallet:, credits_amount:, ignore_validation: false, field_name: :paid_credits)
       @result = result
       @wallet = wallet
+      # NOTE: credits_amount must be a string to be able to use ::Validators::DecimalAmountService
       @credits_amount = credits_amount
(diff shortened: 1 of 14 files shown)


spec/graphql/mutations/wallet_transactions/create_spec.rb
```diff
@@ -101,4 +102,32 @@
...
     )
   end
+
+  context "when wallet has a minimum amount" do
+    before do
+      wallet.update!(paid_top_up_min_amount_cents: 10_00)
+    end
+
+    it "returns an error" do
+      # TODO: check error content when we return metadata
```
floganz, line 112: When you think we'll be able to fix this todo?
Is that dependent on error pr?

spec/graphql/mutations/wallet_transactions/create_spec.rb
```diff
@@ -101,4 +102,32 @@
...
+      let(:input) do
+        {
+          walletId: wallet.id,
+          paidCredits: "5.00",
+          ignorePaidTopUpLimits: true
+        }
+      end
+
+      it "creates the transaction" do
+        expect { result }.to change(organization.wallet_transactions, :count).by(1)
```
floganz: nit: we can use `expect { subject }` here, while referring to same result as `result` below.

app/services/validators/wallet_transaction_amount_limits_validator.rb
```diff
@@ -5,11 +5,18 @@ class WalletTransactionAmountLimitsValidator
...
       @result = result
       @wallet = wallet
+      # NOTE: credits_amount must be a string to be able to use ::Validators::DecimalAmountService
       @credits_amount = credits_amount
       @ignore_validation = ActiveModel::Type::Boolean.new.cast(ignore_validation)
       @field_name = field_name
     end
 
+    def raise_if_invalid!
+      return if valid?
```
floganz, line 15: Let's also test this case in validatior.

spec/graphql/mutations/wallet_transactions/create_spec.rb
julienbourdeau: Yes, when we return errors with metadata either in https://github.com/getlago/lago-api/pull/4251 or in a smaller and hackier version 😬

PR #4285, opened 2025-09-10
feat(wallets): follow Min and Max for transaction created by `target` rules

julienbourdeau:
When a target rule is processed we apply the minimum transaction amount on the wallet, but not the maximum.

The `gap` is the different between the current balance and the target balance.

**Examples**
 
```
- current balance: 12
- target balance: 30
- gap: 18
- wallet minimum: 25
```
In this case, we'll create a transaction of 25 credits putting the **balance to 37, instead of 30**.

```
- current balance: 12
- target balance: 30
- gap: 18
- wallet maximum: 10
```
In this case, we'll ignore the maximum and create a transaction of 18 credits to put the balance to target. Reaching target is more important than maximum.


app/models/recurring_transaction_rule.rb
```diff
@@ -44,6 +44,14 @@ def mark_as_terminated!(timestamp = Time.zone.now)
       .where("expiration_at IS NOT NULL AND expiration_at <= ?", Time.current)
   }
   scope :expired, -> { where("recurring_transaction_rules.expiration_at::timestamp(0) <= ?", Time.current) }
+
+  def ignore_paid_top_up_limits?
+    ignore_paid_top_up_limits
```
jdenquin: its a boolean, you already have the `ignore_paid_top_up_limits?`  method with rails...

app/models/recurring_transaction_rule.rb
```diff
@@ -44,6 +44,14 @@ def mark_as_terminated!(timestamp = Time.zone.now)
...
+  end
+
+  def follow_paid_top_up_limits?
+    !ignore_paid_top_up_limits?
```
jdenquin: no need for an extra method here, you can just comment if you have the feeling that its hard to read

app/models/wallet.rb
```diff
@@ -43,6 +43,18 @@ def mark_as_terminated!(timestamp = Time.zone.now)
     terminated!
   end
 
+  def has_paid_top_up_limits?
+    paid_top_up_min_amount_cents.present? || paid_top_up_max_amount_cents.present?
+  end
+
+  def paid_top_up_min_credits
+    WalletCredit.from_amount_cents(wallet: self, amount_cents: paid_top_up_min_amount_cents).credit_amount
```
jdenquin, line 49: why using a method here if you do not memoize it?

app/services/wallets/create_interval_wallet_transactions_service.rb
```diff
@@ -31,9 +32,27 @@ def today
     end
 
     def paid_credits(rule)
-      return [(rule.target_ongoing_balance - rule.wallet.credits_ongoing_balance), 0.0].max if rule.target?
+      if rule.target?
```
jdenquin: ruby way please

```ruby
return target_paid_credits(rule) if rule.target?

rule.paid_credits
```

app/services/wallets/create_interval_wallet_transactions_service.rb
```diff
@@ -31,9 +32,27 @@ def today
...
 
-      rule.paid_credits
+    def target_paid_credits(rule)
+      if rule.wallet.credits_ongoing_balance > rule.target_ongoing_balance
+        return 0.0
+      end
+
+      gap = rule.target_ongoing_balance - rule.wallet.credits_ongoing_balance
+
+      if rule.wallet.paid_top_up_min_amount_cents.present? && rule.follow_paid_top_up_limits?
```
jdenquin: can you please simplify this?

julienbourdeau: We did, with using clamp.

app/models/wallet.rb
julienbourdeau: We don't reuse it so we don't see a reason to memoize it

app/models/recurring_transaction_rule.rb
julienbourdeau: Not needed anymore, it's removed. I'm not sure a comment is better than a method tho.

app/models/recurring_transaction_rule.rb
julienbourdeau: yeah, I don't know what I made this but its gone, we refactored the implementation.

app/models/wallet.rb
jdenquin: hard to read in the review but, do you call it a lot?

app/models/recurring_transaction_rule.rb
jdenquin: for readability should be better, especially for this kind of method imo

app/models/recurring_transaction_rule.rb
```diff
@@ -44,6 +44,43 @@ def mark_as_terminated!(timestamp = Time.zone.now)
       .where("expiration_at IS NOT NULL AND expiration_at <= ?", Time.current)
   }
   scope :expired, -> { where("recurring_transaction_rules.expiration_at::timestamp(0) <= ?", Time.current) }
+
+  def apply_min_top_up_limits(credit_amount:)
+    if ignore_paid_top_up_limits?
+      credit_amount
+    else
+      credit_amount.clamp(wallet.paid_top_up_min_credits, nil)
```
groyoh, line 52: TIL about `#clamp` 😄

app/models/recurring_transaction_rule.rb
```diff
@@ -44,6 +44,43 @@ def mark_as_terminated!(timestamp = Time.zone.now)
       .where("expiration_at IS NOT NULL AND expiration_at <= ?", Time.current)
   }
   scope :expired, -> { where("recurring_transaction_rules.expiration_at::timestamp(0) <= ?", Time.current) }
+
+  def apply_min_top_up_limits(credit_amount:)
```
groyoh, line 48: Does this need to be public ?

app/services/wallets/create_interval_wallet_transactions_service.rb
```diff
@@ -2,26 +2,29 @@
...
-      recurring_transaction_rules.each do |recurring_transaction_rule|
-        wallet = recurring_transaction_rule.wallet
-        paid_credits = paid_credits(recurring_transaction_rule)
-        granted_credits = granted_credits(recurring_transaction_rule)
+      recurring_transaction_rules.each do |rule|
+        paid_credits = rule.compute_paid_credits(ongoing_balance: rule.wallet.credits_ongoing_balance)
+        granted_credits = rule.compute_granted_credits
 
-        next if recurring_transaction_rule.target? && paid_credits.zero? && granted_credits.zero?
+        next if rule.target? && paid_credits.zero? && granted_credits.zero?
```
groyoh, line 12: This is existing code but shouldn't we have:

```suggestion
        next if paid_credits.zero? && granted_credits.zero?
```

I fail to see why we'd create a transaction if both are zero 🤔

spec/models/wallet_spec.rb
```diff
@@ -79,4 +79,24 @@
       end
     end
   end
+
+  describe "#paid_top_up_min_credits" do
+    it "converts min amount cents to credits using the wallet rate" do
+      wallet.paid_top_up_min_amount_cents = 1_00
```
groyoh, line 86: Maybe we should use a rate different than 1 in tests ?

spec/models/recurring_transaction_rule_spec.rb
```diff
@@ -41,4 +41,120 @@
...
+
+      context "when ongoing balance equals to target balance" do
+        let(:target_ongoing_balance) { 100.0 }
+
+        it "returns zero" do
+          expect(subject).to eq 0.0
+        end
+      end
+
+      context "when ongoing balance is smaller than target balance" do
```
groyoh, line 151: I think we're missing context: when min is set. When min isn't set. when ignore_paid_top_up_limits is true. 

Or do you assume it's tested above with `#apply_min_top_up_limits` ?

app/models/recurring_transaction_rule.rb
julienbourdeau: me too! All credits go to @floganz

app/models/recurring_transaction_rule.rb
floganz: It's not required to be public but it's easier to test it in all cases in isolation.

app/services/wallets/create_interval_wallet_transactions_service.rb
floganz: Rule could have two modes `fixed` and `target`. For `fixed` we use `paid_credits` and `granted_credits` as amounts that will be top up at certain events.
For `target` amount of top up defined by the gap between set balance on rule and actual ongoing balance.

spec/models/recurring_transaction_rule_spec.rb
floganz: Yes, it was intent to test methods in isolation and do not repeat all the cases everywhere.

spec/models/wallet_spec.rb
floganz: I don't mind. Will adjust it few tests.

app/services/wallets/create_interval_wallet_transactions_service.rb
groyoh: @floganz that's clear, thanks. Maybe I'm missing something but my point was that, whether we're `fixed` or `target`, if both paid and granted are 0, it doesn't make sense to schedule the job. Right now, if both are 0, we only skip the job scheduling for `target`. That means we'll schedule a 0 credit + 0 granted transaction. Or am I misunderstanding the code here ? 🤔

floganz: Indeed it looks like nothing prevents us to schedule a job with zero amounts.

I'd say that in first place it should not be possible to created `fixed` rule with zero amounts. If we'll have such limitation on rule then current condition can exist as is. Because it's either `target` and we skip if amounts are zeros or it's `fixed` and by validation we sure that it has at least one non-zero amount.

wdyt @julienbourdeau ?

groyoh: Indeed. That should not be possible in the first place but I don't think we have such validation right now, right ? I believe we only validate that it's a decimal: https://github.com/getlago/lago-api/blob/170f68bc1a3a8520b5ce8f07c041dc7e64bdf890/app/services/wallets/recurring_transaction_rules/validate_service.rb#L59

julienbourdeau: Let's check this separately. It probably shouldn't be possible but I'd rather not change this behavior at the same time.

PR #4291, opened 2025-09-11
feat(fixed_charges): Calculate minimum commitment w. fixed charges

ancorcruz:
## Roadmap Task

 ## Context

 **Option 1:**
 User has to create a one off invoice alongside the subscription, it will create 2 different invoices.

 **Option 2:**
 User can add a recurring billable metric and use event to have this fee invoice on subscription renewal, but it won’t appear on the first billing subscription.

 At subscription creation or afterward, there is no clear way to invoice a fixed fee that is not tied to events, aside from the subscription fee itself. This fee could be either a one-time charge or a recurring one.

 ## Description

 Compute minimum commitment including fixed charges fees

### DIFF
--- a/app/services/commitments/minimum/in_advance/calculate_true_up_fee_service.rb
+++ b/app/services/commitments/minimum/in_advance/calculate_true_up_fee_service.rb
@@ -37,14 +37,12 @@ module Commitments
         end
 
         def charge_fees
-          invoices_result = FetchInvoicesService.call(commitment: minimum_commitment, invoice_subscription:)
-
           Fee
             .charge
             .joins(:charge)
             .where(
               subscription_id: subscription.id,
-              invoice_id: invoices_result.invoices.ids,
+              invoice_id: invoices_ids,
               charge: {pay_in_advance: false}
             )
         end
@@ -92,14 +90,12 @@ module Commitments
         end
 
         def fixed_charge_fees
-          invoices_result = FetchInvoicesService.call(commitment: minimum_commitment, invoice_subscription:)
-
           Fee
             .fixed_charge
             .joins(:fixed_charge)
             .where(
               subscription_id: subscription.id,
-              invoice_id: invoices_result.invoices.ids,
+              invoice_id: invoices_ids,
               fixed_charge: {pay_in_advance: false}
             )
         end
@@ -128,6 +124,13 @@ module Commitments
             commitment: minimum_commitment,
             invoice_subscription: invoice_subscription.previous_invoice_subscription
           ).call
+        end
+
+        def invoices_ids
+          @invoices_ids ||= FetchInvoicesService.call(
+            commitment: minimum_commitment,
+            invoice_subscription:
+          ).invoices.ids
         end
       end
     end


app/services/commitments/minimum/in_advance/calculate_true_up_fee_service.rb
```diff
@@ -105,6 +105,43 @@ def charge_in_advance_recurring_fees
               dates_service.end_of_period&.iso8601(3)
             )
         end
+
+        def fixed_charge_fees
+          invoices_result = FetchInvoicesService.call(commitment: minimum_commitment, invoice_subscription:)
```
annvelents: I think we can extract this to a separate method and memoize, because it's called from several methods, so we'd make the calculation several times when calculating different fees

app/services/commitments/minimum/in_advance/calculate_true_up_fee_service.rb
```diff
@@ -105,6 +105,43 @@ def charge_in_advance_recurring_fees
...
+            .joins(:fixed_charge)
+            .where(
+              subscription_id: subscription.id,
+              invoice_id: invoices_result.invoices.ids,
+              fixed_charge: {pay_in_advance: false}
+            )
+        end
+
+        def fixed_charge_in_advance_fees
+          dates_service = Commitments::DatesService.new_instance(
```
annvelents: this one also is initialized for each `*_in_advance_fees` with the same params, so we might benefit from memoization

app/services/commitments/minimum/in_arrears/calculate_true_up_fee_service.rb
```diff
@@ -115,6 +115,53 @@ def charge_in_advance_recurring_fees
 
           scope
         end
+
+        def fixed_charge_fees
+          dates_service = Commitments::DatesService.new_instance(
```
annvelents: also here

app/services/commitments/minimum/in_arrears/calculate_true_up_fee_service.rb
```diff
@@ -115,6 +115,53 @@ def charge_in_advance_recurring_fees
...
+              dates_service.previous_beginning_of_period
+            )
+            .where(
+              "(fees.properties->>'fixed_charges_to_datetime') <= ?",
+              dates_service.end_of_period&.iso8601(3)
+            )
+        end
+
+        def fixed_charge_in_advance_fees
+          dates_service = Commitments::DatesService.new_instance(
```
annvelents: and here

spec/factories/fixed_charges.rb
```diff
@@ -9,6 +9,11 @@
     units { 1 }
     properties { {amount: Faker::Number.between(from: 100, to: 500).to_s} }
     invoice_display_name { Faker::Fantasy::Tolkien.location }
+    pay_in_advance { false }
```
ivannovosad: We can skip this, the default is `false`

app/services/commitments/minimum/in_arrears/calculate_true_up_fee_service.rb
```diff
@@ -115,6 +115,53 @@ def charge_in_advance_recurring_fees
...
+            )
+            .where(
+              "(fees.properties->>'fixed_charges_from_datetime') >= ?",
+              dates_service.previous_beginning_of_period
+            )
+            .where(
+              "(fees.properties->>'fixed_charges_to_datetime') <= ?",
+              dates_service.end_of_period&.iso8601(3)
+            )
+        end
```
ivannovosad, line 125: Would it make sense to extract a method that contains most of the logic from this method and `charge_fees` and just have:

```
def fixed_charge_fees()
   fees(fixed: true)
end

def charge_fees()
   fees(fixed: false)
end

```

app/services/commitments/minimum/in_arrears/calculate_true_up_fee_service.rb
```diff
@@ -115,6 +115,53 @@ def charge_in_advance_recurring_fees
...
+
+        def fixed_charge_in_advance_fees
```
ivannovosad, line 127: Same as the above.

ivannovosad (changes requested):

app/services/commitments/minimum/in_advance/calculate_true_up_fee_service.rb
```diff
@@ -105,6 +105,43 @@ def charge_in_advance_recurring_fees
...
+            commitment: minimum_commitment,
+            invoice_subscription: invoice_subscription.previous_invoice_subscription
+          ).call
```
ancorcruz: extract this

annvelents (commented):
small things that can be extracted
(but Ivan's suggestion I think is even better)

app/services/commitments/minimum/in_arrears/calculate_true_up_fee_service.rb
ancorcruz: I don't have a good name for this `fees` is confusing as there are also fees belonging to the subscription `subscription_fees`. Also, this is a generalization that could not work in the long run. I would extract the `invoices_result` into a memoized variable and leave this `charge_fees` and `fixed_charge_fees` with the active record query.

PR #4308, opened 2025-09-12
feat(semiannual): Add semiannual interval to services

ivannovosad:
## Context

When creating a plan, a new interval option should be available: `semiannual`.
In order to enable that, we’ll need to add it to the model, services and GQL.

## Description

This PR adds the new interval `semiannual` to all the services where necessary.

### DIFF
--- a/spec/scenarios/subscriptions/billing_spec.rb
+++ b/spec/scenarios/subscriptions/billing_spec.rb
@@ -853,6 +853,141 @@ describe "Billing Subscriptions Scenario", :scenarios, type: :request do
     end
   end
 
+  context "with semiannual plan and monthly charge" do
+    let(:plan_interval) { "semiannual" }
+    let(:plan_monthly_charges) { true }
+
+    context "with calendar billing" do
+      let(:billing_time) { "calendar" }
+
+      let(:subscription_time) { DateTime.new(2022, 2, 4) }
+
+      let(:before_billing_times) { [DateTime.new(2021, 12, 31)] }
+      let(:billing_times) { [DateTime.new(2022, 7, 1, 1), DateTime.new(2022, 7, 1, 2)] }
+      let(:after_billing_times) { [DateTime.new(2022, 7, 2)] }
+
+      it_behaves_like "a subscription billing without duplicated invoices"
+
+      context "with UTC+ timezone" do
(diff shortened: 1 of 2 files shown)


app/services/invoices/calculate_fees_service.rb
```diff
@@ -230,6 +232,24 @@ def should_create_subscription_fee?(subscription, boundaries)
         (subscription.terminated? && subscription.terminated_at > invoice.created_at)
     end
 
+    def should_create_semiannual_subscription_fee?(subscription)
```
vincent-pochet, line 235: Since the logic is quite complex in here, do you mind adding some comments on what are the different conditions we are checking? It will definitely help future us understanding what's going on in here :D

app/services/plans/update_service.rb
```diff
@@ -86,13 +86,13 @@ def call
     delegate :organization, to: :plan
 
     def bill_charges_monthly?
-      return unless params[:interval]&.to_sym == :yearly
+      return if params[:interval]&.to_sym != :yearly && params[:interval]&.to_sym != :semiannual
```
vincent-pochet:
```suggestion
      return unless %[yearly semiannual].include?(params[:interval]&.to_sym)
```

You can even extract it in a method with memoization as it used also in `bill_fixed_charges_monthly?`

spec/services/subscriptions/dates/semiannual_service_spec.rb
```diff
@@ -0,0 +1,887 @@
...
+    )
+  end
+
+  let(:customer) { create(:customer, timezone:) }
+  let(:plan) { create(:plan, interval: :semiannual, pay_in_advance:) }
+  let(:pay_in_advance) { false }
+  let(:current_usage) { false }
+
+  let(:subscription_at) { Time.zone.parse("02 Feb 2021") }
+  # let(:billing_at) { Time.zone.parse("07 Mar 2022") }
```
vincent-pochet: Maybe this one should be removed :)

app/services/invoices/calculate_fees_service.rb
ivannovosad: It is the same as in this case: `should_create_yearly_subscription_fee?` just for semiannual interval here. Adding it.

PR #4317, opened 2025-09-15
fix(invoices): Add invoice_subscriptions timestamp index

jdenquin:
- Using `COALESCE` in an ORDER has a very high impact on CPU with PostgreSQL, this should be covered in an index if used.


db/migrate/20250915100607_add_subscription_with_timestamp_index_to_invoice_subscriptions.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class AddSubscriptionWithTimestampIndexToInvoiceSubscriptions < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  INDEX_NAME = "idx_invoice_subscriptions_on_subscription_with_timestamps".freeze
```
julienbourdeau: all string literals are frozen

PR #4319, opened 2025-09-15
fix(invoice): Make invoice overdue marking asynchronous

groyoh:
## Context

The `Clock::MarkInvoicesAsPaymentOverdueJob` currently mark the invoices as overdue synchronously which can take a while if there are a lot of invoices.

This could in turn block other clock jobs.

## Description

This issue was fixed by marking the invoice as overdue in an asynchronous job. This also ensures we have a uniqueness on the job to avoid running it multiple times in parallel.

### DIFF
--- a/app/jobs/clock/mark_invoices_as_payment_overdue_job.rb
+++ b/app/jobs/clock/mark_invoices_as_payment_overdue_job.rb
@@ -2,6 +2,8 @@
 
 module Clock
   class MarkInvoicesAsPaymentOverdueJob < ClockJob
+    unique :until_executed, on_conflict: :log
+
     def perform
       Invoice
         .finalized
@@ -9,8 +11,11 @@ module Clock
         .where(payment_overdue: false)
         .where(payment_dispute_lost_at: nil)
         .where(payment_due_date: ...Time.current)
-        .find_each do |invoice|
-          Invoices::Payments::MarkOverdueService.call(invoice:)
+        .find_in_batches do |invoices|
+          jobs = invoices.map do |invoice|
+            Invoices::Payments::MarkOverdueJob.new(invoice:)
+          end
+          ActiveJob.perform_all_later(jobs)
         end
     end
   end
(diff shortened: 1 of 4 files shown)


app/jobs/clock/mark_invoices_as_payment_overdue_job.rb
```diff
@@ -2,15 +2,20 @@
...
         .where(payment_overdue: false)
         .where(payment_dispute_lost_at: nil)
         .where(payment_due_date: ...Time.current)
-        .find_each do |invoice|
-          Invoices::Payments::MarkOverdueService.call(invoice:)
+        .find_in_batches do |invoices|
+          jobs = invoices.map do |invoice|
+            Invoices::Payments::MarkOverdueJob.new(invoice:)
+          end
+          ActiveJob.perform_all_later(jobs)
```
julienbourdeau, line 18: Always happy to see jobs being added in batch

app/jobs/clock/mark_invoices_as_payment_overdue_job.rb
```diff
@@ -2,15 +2,20 @@
...
     def perform
       Invoice
         .finalized
         .not_payment_succeeded
         .where(payment_overdue: false)
         .where(payment_dispute_lost_at: nil)
         .where(payment_due_date: ...Time.current)
-        .find_each do |invoice|
-          Invoices::Payments::MarkOverdueService.call(invoice:)
+        .find_in_batches do |invoices|
```
julienbourdeau, line 14: If we want to take this one step closer, I like when the clock jobs dispatches a PerOrganization job, than then makes a batch.

(I detailled it here when discussing override https://github.com/getlago/lago-api/pull/4139#issuecomment-3187471913)

See `Clock::SubscriptionsBillerJob` dispatching `Subscriptions::OrganizationBillingJob` for isntance. Or `Clock::ProcessAllSubscriptionActivitiesJob` dispatching `ProcessOrganizationSubscriptionActivitiesJob`.

--

In this specific case, it will probably enqueue a tiny bit more jobs because yours will get it on one batch while mine needs on extra job per org but I think it's negligeable.

groyoh: Sounds great. We can tackle one issue at a time though.

PR #4320, opened 2025-09-15
Feat(fixed_charges): add events aggregators

annvelents:
## Context

For every subscription, when a subscription starts or when units for a fixed charge are updated, we'll be issuing the FixedChargeEvent. Later, when calculating the fixedChargesFees, we need to aggregate these events (prorated and not) in order to correctly calculate the fees. This PR introduces the aggregation services, that return units amount, onto which later we'll be applying fixed_charges charge_models to get the resulting fee amount

## Description

- Introduced FixedChargeEvent aggregation services

### DIFF
--- a/app/services/fixed_charge_events/aggregations/base_service.rb
+++ b/app/services/fixed_charge_events/aggregations/base_service.rb
@@ -28,7 +28,7 @@ module FixedChargeEvents
         @events_in_range ||= begin
           events_in_period_ids = base_events.where(timestamp: from_datetime..to_datetime).ids
           last_event_before_range_id = base_events.where("timestamp < ?", from_datetime).order(timestamp: :desc).limit(1).ids
-          
+
           # Combine using UNION
           FixedChargeEvent.where(id: events_in_period_ids + last_event_before_range_id)
             .order(timestamp: :asc)
--- a/app/services/fixed_charge_events/aggregations/base_service.rb
+++ b/app/services/fixed_charge_events/aggregations/base_service.rb
@@ -10,9 +10,9 @@ module FixedChargeEvents
         @subscription = subscription
         @customer = subscription.customer
         # TODO: switch to fixed_charges_boundaries
-        @from_datetime = boundaries.charges_from_datetime
-        @to_datetime = boundaries.charges_to_datetime
-        @charges_duration = boundaries.charges_duration
+        @from_datetime = boundaries.fixed_charges_from_datetime
+        @to_datetime = boundaries.fixed_charges_to_datetime
+        @charges_duration = boundaries.fixed_charges_duration
 
         super(nil)
       end
--- a/app/services/fixed_charge_events/aggregations/base_service.rb
+++ b/app/services/fixed_charge_events/aggregations/base_service.rb
@@ -7,7 +7,7 @@ module FixedChargeEvents
         @fixed_charge = fixed_charge
         @subscription = subscription
         @from_datetime = charges_from_datetime
-        # NOTE: we add 1 day to the duration to include the last day of the period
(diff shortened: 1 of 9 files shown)


ancorcruz (approved):
LGTM!

app/services/fixed_charge_events/aggregations/base_service.rb
```diff
@@ -0,0 +1,43 @@
...
+    class BaseService < BaseService
+      Result = BaseResult[:count, :aggregation, :current_usage_units, :full_units_number, :total_aggregated_units]
+
+      def initialize(fixed_charge:, subscription:, boundaries:)
+        @fixed_charge = fixed_charge
+        @subscription = subscription
+        @customer = subscription.customer
+        @from_datetime = boundaries.charges_from_datetime.in_time_zone(customer.applicable_timezone).to_date
+        # NOTE: we add 1 day to the duration to include the last day of the period.
+        @to_datetime = boundaries.charges_to_datetime.in_time_zone(customer.applicable_timezone).to_date + 1.day
```
vincent-pochet: Out of curiosity, is there any reason for relying on `datetime.in_timezone.to_date` rather than using absolute datetime? 

Using datetime would allow you to avoid the timezone conversion (and sometimes forget about it 🙂) and are supposed to take care of subscription start/upgrade/termination.

app/services/fixed_charge_events/aggregations/base_service.rb
```diff
@@ -0,0 +1,43 @@
...
+        @charges_duration = to_datetime - from_datetime
```
vincent-pochet: This could be done via the boundaries and filled via the date service, what do you think?

app/services/fixed_charge_events/aggregations/prorated_aggregation_service.rb
```diff
@@ -0,0 +1,132 @@
...
+
+          SELECT COALESCE(SUM(weighted_units), 0) AS aggregation
+          FROM (
+            SELECT CASE WHEN (#{period_ratio_sql} * units) < 0 THEN 0 ELSE (#{period_ratio_sql} * units) END AS weighted_units
+            FROM fixed_charge_events_ignored
+          ) cumulated_ratios
+        SQL
+      end
+
+      def debug_query
```
vincent-pochet, line 56: The name speaks for itself, but maybe you could add a comment to explain what it is supposed to do

app/services/fixed_charge_events/aggregations/prorated_aggregation_service.rb
```diff
@@ -0,0 +1,132 @@
...
+          ]
+        )
+        sql_result = ActiveRecord::Base.connection.select_one(sql)
+        result.aggregation = sql_result["aggregation"]
+        result
+      end
+
+      private
+
+      def prorated_query
```
vincent-pochet, line 34: If you don't mind, since SQL is not always the most readable language (😅 ), could you add some lines explaining the purpose of this query?

app/services/fixed_charge_events/aggregations/base_service.rb
annvelents: totally make sense! 😆 
honestly, why: because originally I was sending to and from as separate variables, and in my tests I was using some random dates, which was resulting in some weird calculation 🙈 
then I changes to boundaries, but since I already had this `in_timezone.to_datetime`, I forgot to change it 🙈

app/services/fixed_charge_events/aggregations/base_service.rb
annvelents: 🙌

PR #4324, opened 2025-09-15
feat(fixed-charges): Fixed charges on subscriptions terminated dates service

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

consider fixed charges boundaries in subscriptions terminated dates service as we currently do for charges boundaries.


ancorcruz:
@vincent-pochet @annvelents I have rebased main branch and added some specs... Do you mind to take another look please?

PR #4328, opened 2025-09-16
fix(billable_metric_filter): Fix batch remove

vincent-pochet:
## Context

This PR fixes the `Discard::RecordNotDiscarded - A discarded record cannot be discarded (Discard::RecordNotDiscarded)` that is raised in `BillableMetricFilters::CreateOrUpdateBatchService` when try to remove all filters on a billable metric.

The issue was raised because the same billable_metric_filter was used multiple time on the same charge filter.

## Description

This PR makes sure that the filter value that we are trying to soft delete is not already soft deleted, to avoid raising the exception

### DIFF
--- a/app/services/billable_metric_filters/create_or_update_batch_service.rb
+++ b/app/services/billable_metric_filters/create_or_update_batch_service.rb
@@ -86,7 +86,7 @@ module BillableMetricFilters
       filter_value.discard!
       return if filter_value.charge_filter.values.where.not(id: filter_value.id).exists?
 
-      filter_value.charge_filter.discard!
+      filter_value.charge_filter.discard! unless filter_value.charge_filter.discarded?
     end
 
     def refresh_draft_invoices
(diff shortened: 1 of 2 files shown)


app/services/billable_metric_filters/create_or_update_batch_service.rb
```diff
@@ -86,7 +86,7 @@ def discard_filter_value(filter_value, new_values: [])
       filter_value.discard!
       return if filter_value.charge_filter.values.where.not(id: filter_value.id).exists?
 
-      filter_value.charge_filter.discard!
+      filter_value.charge_filter.discard! unless filter_value.charge_filter.discarded?
```
rsempe, line 89: how can `filter_value.charge_filter` not be `nil` if it's discarded? 🤔 

I can't see `with_discarded` in the `BillableMetricFilter#has_many`

vincent-pochet: @rsempe It is removed in the same service instance, so it's in memory

rsempe: ok, can't we just reload `filter_value.charge_filter` in that case?

`filter_value.reload.charge_filter&.discard!`

rsempe (approved):

app/services/billable_metric_filters/create_or_update_batch_service.rb
vincent-pochet: We could, but it would cost us a lot more DB queries when checking the discarded_at is free

PR #4331, opened 2025-09-16
fix(plan): Add validation on interval enum

vincent-pochet:
## Description
This PR adds a new validation on the `interval` enum of the `Plan` model

The goal is to avoid an `ArgumentError` when an invalid value is provided via the API as it results in an `HTTP 500` instead of a `422`


julienbourdeau (approved):

spec/models/plan_spec.rb
```diff
@@ -28,7 +28,7 @@
     expect(subject).to have_many(:entitlement_values).through(:entitlements).source(:values).class_name("Entitlement::EntitlementValue").dependent(:destroy)
 
     expect(subject).to validate_presence_of(:interval)
-    expect(subject).to define_enum_for(:interval).with_values(Plan::INTERVALS)
+    expect(subject).to define_enum_for(:interval).with_values(Plan::INTERVALS).validating
```
groyoh, line 30: Nice!

ancorcruz (approved):
Looks great. :shipit:

PR #4336, opened 2025-09-17
feat(stripe): Add 3DS support

julienbourdeau:
With this PR, when a Stripe payment requires a 3DS authentication.

I added integration tests to test SEPA, ACH and bank transfer (`customer_balance`). 
They are not all tested (boleto, crypto and BACS credit are not tested yet.

A unified Payment model, used for the `payment.requires_action` webhook was introduced here: https://github.com/getlago/lago-api/pull/4350

> [!IMPORTANT]
> Previously, when an invoice requires 3DS the invoice payment is **marked as failed**.
> Now, the payment is **marked as pending**.

All the work to support 3DSecure (invoice status, sending `next_action`...) was already done to support customer in India, this is why you don't see anything in the diff.

It turned out that  `off_session` and `error_on_requires_action` were basically set to prevent 3DS, so I removed them entirely.

As far as I understand, this was only used for `card` payment and I have not seen any effect on other payment method.


spec/integration/stripe/simple_payment_integration_spec.rb
```diff
@@ -140,6 +134,34 @@
     })
   end
 
+  context "when invoice payment requires 3DS" do
+    it "create a pending payment and send webhook with `next_action" do
+      customer = create_customer
+      stripe_customer = customer.payment_provider_customers.sole
+      webhooks_sent.clear
+
+      threeds_pm_id = attach_card!(stripe_customer, :pm_card_authenticationRequired)
```
groyoh: For readability:

```suggestion
      three_ds_pm_id = attach_card!(stripe_customer, :pm_card_authenticationRequired)
```

or 
```suggestion
      3ds_pm_id = attach_card!(stripe_customer, :pm_card_authenticationRequired)
```

groyoh: Why using bang on those methods ?

spec/integration/stripe/simple_payment_integration_spec.rb
```diff
@@ -164,23 +186,85 @@
...
+      stripe_customer = customer.payment_provider_customers.sole
+
+      create_one_off_invoice(customer, [addon], units: 22)
+
+      # Stripe should return a 400 because of `confirm: true` and no payment method
+      invoice = customer.invoices.sole
+      expect(invoice.ready_for_payment_processing).to eq true
+      expect(invoice.payment_status).to eq("failed")
+      expect(invoice.payments.count).to eq(1)
+      p = invoice.payments.sole
```
groyoh: For readability:

```suggestion
      failed_payment = invoice.payments.sole
      expect(failed_payment.status).to eq 'failed'
```

groyoh (approved):
A few nitpicks about naming with could be more readable but looks fine otherwise ✅

spec/integration/stripe/simple_payment_integration_spec.rb
julienbourdeau: I would have used `3ds_pm_id` but variable name cannot start with an number.
I agree `threeds_pm_id` sucked, I added a `_`.

About bang, no reason. I like the `!` too much 🤣 I remove it from all methods!

spec/integration/stripe/simple_payment_integration_spec.rb
julienbourdeau: renamed!

PR #4338, opened 2025-09-17
Feat(fixed_charges-16): move charge model factory out of charges namespace

annvelents:
## Context

Since now ChargeModel can be applied on both: charges and FixedCharges, the factory that returns charge_model that should be applied, should be in it's own namespace.

## Description

- Move `ChargeModelFactory` out of `Charges` namespace
- Allow `ChargeModel::Factory` to accept chargeable that will be Charge of FixedCharge


ancorcruz (approved):
LGTM!

PR #4340, opened 2025-09-17
chore(admin): Add premium_integrations for org admin endpoint

jdenquin:
- Add `premium_integrations` for `/admin/organization` endpoints

### DIFF
--- a/app/controllers/admin/organizations_controller.rb
+++ b/app/controllers/admin/organizations_controller.rb
@@ -22,7 +22,7 @@ module Admin
       result = ::Organizations::CreateService
         .call(
           name: create_params[:name],
-          document_numbering: "per_organization",
+          document_numbering: "per_customer",
           premium_integrations: create_params[:premium_integrations]
         )
 


app/controllers/admin/organizations_controller.rb
```diff
@@ -20,7 +20,11 @@ def update
 
     def create
       result = ::Organizations::CreateService
-        .call(name: create_params[:name], document_numbering: "per_organization")
+        .call(
+          name: create_params[:name],
+          document_numbering: "per_organization",
```
julienbourdeau: How do you feel about defaulting to `per_customer` instead. It's the recommended way and it performs better. Users can change it to `per_organization` if that what they need.

jdenquin:
> Maybe add a test to ensure you cannot specify an invalid integration (there's validation for that) ?

There is already a validation on the model for this

app/controllers/admin/organizations_controller.rb
jdenquin: I don't know tbh, its was already `per_organization`

jdenquin: done!

PR #4342, opened 2025-09-17
fix(invoice): add missing actions for regenerate voided invoice

lovrocolic:
## Context

Regenerated invoice (after voiding) missed some core actions like processing payment or sync with 3rd party providers

## Description

This PR adds missing actions

### DIFF
--- a/app/services/invoices/regenerate_from_voided_service.rb
+++ b/app/services/invoices/regenerate_from_voided_service.rb
@@ -49,6 +49,9 @@ module Invoices
       end
 
       result.invoice = regenerated_invoice
+
+      call_invoice_finalization_jobs(regenerated_invoice)
+
       result
     end
 
@@ -56,6 +59,8 @@ module Invoices
 
     attr_accessor :regenerated_invoice
     attr_reader :voided_invoice, :fees_params
+
+    delegate :customer, to: :voided_invoice
 
     def should_create_credit_note_credit?
       return false unless regenerated_invoice.total_amount_cents&.positive?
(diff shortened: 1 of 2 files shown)


app/services/invoices/regenerate_from_voided_service.rb
```diff
@@ -217,5 +222,21 @@ def create_regenerated_invoice
         invoice.update!(voided_invoice_id: voided_invoice.id)
       end
     end
+
+    def call_invoice_finalization_jobs(invoice)
+      return if invoice.closed?
+
+      Utils::SegmentTrack.invoice_created(invoice)
```
ancorcruz, line 229: shall we also expect this is called in the specs?

app/services/invoices/regenerate_from_voided_service.rb
```diff
@@ -217,5 +222,21 @@ def create_regenerated_invoice
...
+      SendWebhookJob.perform_later("invoice.created", invoice)
+      Utils::ActivityLog.produce(invoice, "invoice.created")
+      GeneratePdfAndNotifyJob.perform_later(invoice:, email: should_deliver_email?)
+      Integrations::Aggregator::Invoices::CreateJob.perform_later(invoice:) if invoice.should_sync_invoice?
+      Integrations::Aggregator::Invoices::Hubspot::CreateJob.perform_later(invoice:) if invoice.should_sync_hubspot_invoice?
```
ancorcruz, line 234: it would be great to cover these with the specs

ancorcruz (approved):
Looks good.

PR #4344, opened 2025-09-17
feat(recurring-transaction-rules): Apply wallet limits to paid_credits on update

floganz:
## Description

Apply wallet limits validation on a recurring rule during update. Recurring transaction rule's `paid_credits` should not exceeds wallet limits (unless ignore option is passed) but only if rule's method is `fixed`.

Zero values for `paid_credits` does not trigger an error even if zero is exceeds wallet limitations. That expected since `fixed` rule we have two amounts: `paid_credits` or `granted_credits`. One of them can be zero but not both at the same time. Currently they can be both zero but that will be changed in next steps.

### DIFF
--- a/app/services/wallets/recurring_transaction_rules/update_service.rb
+++ b/app/services/wallets/recurring_transaction_rules/update_service.rb
@@ -20,7 +20,7 @@ module Wallets
           lago_id = payload_rule[:lago_id]
           rule_attributes = payload_rule.except(:lago_id)
           # Normalize transaction_name to nil if empty
-          rule_attributes[:transaction_name] = rule[:transaction_name].presence if rule.key?(:transaction_name)
+          rule_attributes[:transaction_name] = rule_attributes[:transaction_name].presence if rule_attributes.key?(:transaction_name)
 
           recurring_rule = wallet.recurring_transaction_rules.active.find_by(id: lago_id)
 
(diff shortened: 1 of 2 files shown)


app/services/wallets/recurring_transaction_rules/update_service.rb
```diff
@@ -54,6 +60,31 @@ def sanitize_recurring_rules(args_recurring_rules, created_recurring_rules_ids)
...
+          ignore_validation: attributes[:ignore_paid_top_up_limits]
+        )
+
+        unless validator.valid?
+          result.single_validation_failure!(field: :recurring_transaction_rules, error_code: "invalid_recurring_rule")
+          result.raise_if_error!
+        end
+      end
+
+      def parsed_credits_amount(value)
```
groyoh, line 86: Since we validate the aount in the wallet update service, when would this occur ?

spec/services/wallets/recurring_transaction_rules/update_service_spec.rb
groyoh, line 1: Since the behavior is quite clear ("error vs no error"), I don't think we need to mock `allow(Validators::WalletTransactionAmountLimitsValidator).to receive(:new)`. Simply checking the result is sufficient IMO.

groyoh (approved):

app/services/wallets/recurring_transaction_rules/update_service.rb
floganz: It's a problem right now since existing validator for recurring rules will fail only if both `paid_credits` and `granted_credits` are invalid. If `granted_credits` has proper value it will allow code to execute, even if `paid_credits` equals to `asd`. The last will cause 500.

Until we changed that validator it's needed.

PR #4347, opened 2025-09-17
chore(applied_coupons): DRY applied_coupons index in controller and test

groyoh:
## Context

https://github.com/getlago/lago-api/commit/b2befc225a914ef65009f96232aa38e80cebc58c introduce a `GET /api/v1/customer/:external_customer_id/applied_couponss` endpoint. This commit was mainly a copy-paste from the `GET /api/v1/applied_couponss` endpoints without the `external_customer_id` filter.

The main issue here is that when updating one endpoint (adding a filter for instance), we might forgot to update the other one.

## Description

This refactors the controller and test to reuse as much code as possible and keep both controllers in sync.


vincent-pochet (approved):
🚀

PR #4349, opened 2025-09-17
feat(ai-agent): Ability to fetch ai conversations

rsempe:
## Context

We want the AI agent to assist users in running common actions (e.g., API queries) or accessing product knowledge without requiring engineering support.

## Description

The goal of this PR is to be able to retrieve conversations for a specific membership.


config/environments/staging.rb
```diff
@@ -37,6 +37,10 @@
   config.i18n.fallbacks = true
   config.active_support.report_deprecations = false
 
+  # Allow requests from any origin
+  config.action_cable.disable_request_forgery_protection = true
+  config.action_cable.allowed_request_origins = ["https://agent-app.staging.getlago.com"]
```
julienbourdeau: Is there a dedicated pod always called `agent-app` for all staging? Otherwise, I think this would work: `["https://*.staging.getlago.com"]`

app/graphql/resolvers/ai_conversation_resolver.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    argument :id, ID, required: true, description: "Uniq ID of the ai conversation"
+
+    type Types::AiConversations::Object, null: true
+
+    def resolve(id: nil)
+      current_organization.ai_conversations.find(id)
+    rescue ActiveRecord::RecordNotFound
+      not_found_error(resource: "ai_conversation")
+    end
```
julienbourdeau: I'm not sure how this will behave if `::AiConversations::FetchMessagesService.call(ai_conversation: object)` returns an error (like a timeout or a 404).

Adding the `messages` to the main object means that the frontend can request messages when querying the collection.

I'm wondering if we shouldn't have a dedicated `AiConversationMessageResolver` instead of this one. It would be easier to manage the "network error" 
I feel like the frontend have all the info needed in the collection but call this one specifically to get the messages.

What do you think?


```rb
module Resolvers
  class AiConversationResolver < Resolvers::BaseResolver
    include AuthenticableApiUser
    include RequiredOrganization

    REQUIRED_PERMISSION = "ai_conversations:view"

    description "Query messages from an AI conversation"

    argument :id, ID, required: true, description: "Uniq ID of the ai conversation"

    type Types::AiConversationsMessage::Object, null: true

    def resolve(id: nil)
      ai_conversation = current_organization.ai_conversations.find(id)
      result = ::AiConversations::FetchMessagesService.call(ai_conversation:)

      result.success? ? result.messages : do_magic_here(result)
    rescue ActiveRecord::RecordNotFound
      not_found_error(resource: "ai_conversation")
    end
  end
end
```

config/environments/staging.rb
jdenquin: it should use same behavior as what we have for CORS etc, is it possible with action cable @rsempe

rsempe: yep forgot to push the acceptance of all namespaces

app/services/ai_conversations/fetch_messages_service.rb
```diff
@@ -0,0 +1,32 @@
...
+    MISTRAL_CONVERSATIONS_API_URL = "https://api.mistral.ai/v1/conversations"
+
+    def initialize(ai_conversation:)
+      @ai_conversation = ai_conversation
+      @http = LagoHttpClient::Client.new(api_url)
+    end
+
+    def call
+      result = @http.get(headers:)
+      result["messages"].map { |h| h.slice("content", "created_at", "type") }
```
groyoh: I believe this service doesn't match what we do with other service and doesn't actually return a `Result`:

```suggestion
      http_result = @http.get(headers:)
      messages = http_result["messages"].map { |h| h.slice("content", "created_at", "type") }
      result.messages = messages
      result
```

app/services/ai_conversations/fetch_messages_service.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module AiConversations
+  class FetchMessagesService < BaseService
+    Result = BaseResult[:ai_conversation]
```
groyoh:
```suggestion
    Result = BaseResult[:messages]
```

app/graphql/types/ai_conversations/message.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Types
+  module AiConversations
+    class Message < Types::BaseObject
+      graphql_name "AiConversationMessage"
+
+      field :content, String, null: false
```
groyoh, line 8: If I'm not mistaken, the messages are hashes. Does GraphQL `field` work with hashes ?

Maybe we can also define a non-persisted model for those messages.

app/graphql/resolvers/ai_conversation_resolver.rb
groyoh: > I'm wondering if we shouldn't have a dedicated AiConversationMessageResolver instead of this one. It would be easier to manage the "network error"

I guess we can do this or skip rendering the messages for the collection. But I totally agree that we should try and avoid fetching messages on the collection.

app/services/ai_conversations/stream_service.rb
```diff
@@ -26,7 +26,7 @@ def call
           )
         end
 
-        Thread.pass
+        sleep 0.01
       end
```
julienbourdeau, line 30: I have no idea what's going on here 😆

rsempe: that allows to act as a "stream" when we're displaying chunks of responses

PR #4352, opened 2025-09-18
feat(wallet): Adjust serializers to return limits related attributes

floganz:
## Description

Add min and max paid top up limits to wallet serializer.
Add ignore limits flag to recurring transaction rule serializer.


spec/services/wallets/create_service_spec.rb
```diff
@@ -86,6 +86,20 @@
       end
     end
 
+    context "when paid_credits is above the maximum and ignore validation flag passed" do
+      let(:paid_credits) { "1002.0" }
+      let(:ignore_paid_top_up_limits_on_creation) { "true" }
+
+      it "returns an error" do
+        perform_enqueued_jobs(only: WalletTransactions::CreateJob) do
```
groyoh, line 94: I am not sure what to think of this one.

On one side I like such "integration" tests.

On the other side, it would have been sufficient to test the job was called with the right param. IMO this could lead to "abuse" in the sense that we may start to `perform_enqueued_jobs` in all services. I'm not necessarily saying we shouldn't but I could definitely see the impact on the test if we start to run all jobs in service tests.

app/serializers/v1/wallet_serializer.rb
```diff
@@ -23,7 +23,9 @@ def serialize
         last_balance_sync_at: model.last_balance_sync_at&.iso8601,
         last_consumed_credit_at: model.last_consumed_credit_at&.iso8601,
         terminated_at: model.terminated_at,
-        invoice_requires_successful_payment: model.invoice_requires_successful_payment?
+        invoice_requires_successful_payment: model.invoice_requires_successful_payment?,
+        paid_top_up_min_amount_cents: model.paid_top_up_min_amount_cents,
```
groyoh, line 27: No need for `min_credit` and `max_credit` on API ?

spec/services/wallets/create_service_spec.rb
floganz: I really hope that we'll drop using sidekiq job for creating single transactions during wallet creation and there will be no need for that perform.

Agree with you that it shouldn't become a practice with service tests.

PR #4357, opened 2025-09-18
test(model): Remove redundant `type: :model` metadata from model tests

groyoh:
## Context

All model tests have the `type: :model` RSpec metadata but this metadata is already infered due to the `config.infer_spec_type_from_file_location!` setting.

## Description

This removes the redundant metadata.

### DIFF
--- a/spec/models/add_on_spec.rb
+++ b/spec/models/add_on_spec.rb
@@ -2,7 +2,7 @@
 
 require "rails_helper"
 
-RSpec.describe AddOn, type: :model do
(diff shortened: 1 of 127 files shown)


groyoh:
@julienbourdeau you can find the list here: https://rspec.info/features/8-0/rspec-rails/directory-structure/. It's basically one-to-one with Rails convention `spec/models` -> `type: :model`, `spec/jobs` -> `type: :job`, etc.

julienbourdeau (approved):

PR #4358, opened 2025-09-18
chore(stripe): rename `setup_future_usage`

julienbourdeau:
## Context

I'm adding 3DS support for all Stripe customers. This is extracted from #4336 to keep last PR more manageable.

## Description

`error_on_requires_action?` was introduced in #2690, but #2959 refactored the payment, this function was left here unused.

`setup_future_usage` was added between the 2 PR mentioned above so I reused the `off_session?` method. Since the method is extracted, it's best to rename it. Soon, `off_session?` will be for all customers but `setup_future_usage: "off_session"` should still be only for customer in India.

### DIFF
--- a/app/services/invoices/payments/stripe_service.rb
+++ b/app/services/invoices/payments/stripe_service.rb
@@ -134,7 +134,7 @@ module Invoices
           expires_at: payment_intent.expires_at.to_i,
           payment_intent_data: {
             description:,
-            setup_future_usage: off_session? ? "off_session" : nil, # save payment method for future use
+            setup_future_usage: setup_future_usage? ? "off_session" : nil,
             metadata: {
               lago_customer_id: customer.id,
               lago_invoice_id: invoice.id,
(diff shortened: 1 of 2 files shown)


app/services/invoices/payments/stripe_service.rb
```diff
@@ -202,21 +202,15 @@ def handle_missing_payment(organization_id, stripe_payment)
         result
       end
 
-      # NOTE: Due to RBI limitation, all indians payment should be off_session
-      # to permit 3D secure authentication
-      # https://docs.stripe.com/india-recurring-payments
-      def off_session?
+      # NOTE: Due to RBI limitation, all indians payment should be "on session". See: https://docs.stripe.com/india-recurring-payments
+      # crypto payments don't support 'off_session'
+      def setup_future_usage?
```
groyoh, line 207: Don't know if it would be useful for others but I guess a comment with a link to https://docs.stripe.com/api/payment_intents/create#create_payment_intent-setup_future_usage might be useful.

groyoh (approved):

PR #4361, opened 2025-09-18
fix(progressive_billing): Avoid parallel processing

vincent-pochet:
## Context

Sometimes `UsageMonitoring::ProcessSubscriptionActivityService` and `LifetimeUsages::RecalculateAndCheckJob` are performed at the same time.

This leads to a lot of `Idempotency::IdempotencyError` as a logic to avoid duplicating invoices is in place to prevent the same usage threshold from being billed twice.
In some rare situtation (when we are out of luck) it could lead to a double invoicing of the same thresold, with the amount of the first invoice immediately re-credited on the second one. This situation is related to a race condition appearing when the `LifetimeUsages::CalculateService` start to process just before the commit of the Idempotent transaction.

It appears that the `Idempotent.transaction` is not enough to mitigate this situation as it does not cover the full computation leading to a Progressive Billing invoice.

## Description

This PR re-use the existing `LifetimeUsages::RecalculateAndCheckJob` in the `UsageMonitoring::ProcessSubscriptionActivityService` in order to take advantage of the `active_job_uniqueness`  making sure that we cannot process multiple check for the same lifetime usage record in parallel.

The job is enqueued using `perform_now` because it takes the current_usage as an input parameter and we don't want to add new load with serialization/deserialization of huge JSON objects to this pipeline.

### DIFF
--- a/app/jobs/lifetime_usages/recalculate_and_check_job.rb
+++ b/app/jobs/lifetime_usages/recalculate_and_check_job.rb
@@ -12,9 +12,17 @@ module LifetimeUsages
 
     unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
 
-    def perform(lifetime_usage)
-      LifetimeUsages::CalculateService.call!(lifetime_usage:)
-      LifetimeUsages::CheckThresholdsService.call(lifetime_usage:)
+    # NOTE: do not pass current usage with perform_later as it will be a huge JSON
+    def perform(lifetime_usage, current_usage: nil)
+      LifetimeUsages::CalculateService.call!(lifetime_usage:, current_usage:)
+
+      if lifetime_usage.organization.progressive_billing_enabled?
+        LifetimeUsages::CheckThresholdsService.call!(lifetime_usage:)
+      end
+    end
+
+    def lock_key_arguments
+      [arguments.first]
     end
   end
 end
--- a/app/services/usage_monitoring/process_subscription_activity_service.rb
+++ b/app/services/usage_monitoring/process_subscription_activity_service.rb
@@ -21,12 +21,9 @@ module UsageMonitoring
       #       We believe it's a good tradeoff because they should rarely raise but this can change in the future
 
       begin
+        # Note we rely on the jobs rather than on the services to take advantage of the job's uniqueness strategy
         if organization.using_lifetime_usage?
-          LifetimeUsages::CalculateService.call!(lifetime_usage:, current_usage:)
-        end
-
-        if organization.progressive_billing_enabled?
-          LifetimeUsages::CheckThresholdsService.call(lifetime_usage:)
+          LifetimeUsages::RecalculateAndCheckJob.perform_now(lifetime_usage, current_usage:)
         end
       rescue => e
         exception_to_raise = e
--- a/app/services/lifetime_usages/check_thresholds_service.rb
+++ b/app/services/lifetime_usages/check_thresholds_service.rb
@@ -25,6 +25,7 @@ module LifetimeUsages
           SendWebhookJob.perform_later("subscription.usage_threshold_reached", subscription, usage_threshold:)
         end
       end
+
       result
     end
 
--- a/spec/jobs/lifetime_usages/recalculate_and_check_job_spec.rb
+++ b/spec/jobs/lifetime_usages/recalculate_and_check_job_spec.rb
@@ -3,13 +3,38 @@
 require "rails_helper"
 
 RSpec.describe LifetimeUsages::RecalculateAndCheckJob do
-  let(:lifetime_usage) { create(:lifetime_usage) }
+  let(:organization) { create(:organization, :premium, premium_integrations:) }
+  let(:lifetime_usage) { create(:lifetime_usage, organization:) }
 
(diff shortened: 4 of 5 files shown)


app/services/usage_monitoring/process_subscription_activity_service.rb
```diff
@@ -21,12 +21,9 @@ def call
...
 
       begin
+        # Note we rely on the jobs rather than on the services to take advantage of the job's uniqueness strategy
         if organization.using_lifetime_usage?
-          LifetimeUsages::CalculateService.call!(lifetime_usage:, current_usage:)
-        end
-
-        if organization.progressive_billing_enabled?
-          LifetimeUsages::CheckThresholdsService.call(lifetime_usage:)
+          LifetimeUsages::RecalculateAndCheckJob.perform_now(lifetime_usage, current_usage:)
```
groyoh, line 26: I believe you changed the condition here. Before `LifetimeUsages::CheckThresholdsService.call!(lifetime_usage:)` was called if `organization.progressive_billing_enabled?`. Now it's called if `organization.using_lifetime_usage? && organization.progressive_billing_enabled?`. Is this change wanted ?

app/jobs/lifetime_usages/recalculate_and_check_job.rb
```diff
@@ -12,9 +12,17 @@ class RecalculateAndCheckJob < ApplicationJob
 
     unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
 
-    def perform(lifetime_usage)
-      LifetimeUsages::CalculateService.call!(lifetime_usage:)
-      LifetimeUsages::CheckThresholdsService.call(lifetime_usage:)
+    # NOTE: do not pass current usage with perform_later as it will be a huge JSON
```
groyoh, line 15: Isn't it serialized as a GID though ?

```ruby
◦ lago exec api bundle exec rails console
lago-api(dev)> class TestJob < ApplicationJob
lago-api(dev)*     queue_as :default
lago-api(dev)* 
lago-api(dev)*     def perform(customer:, subscription: nil)
lago-api(dev)*     end
lago-api(dev)>   end
lago-api(dev)> 
lago-api(dev)> TestJob.perform_later(customer: Customer.last, subscrption: Subscription.last)

◦ lago exec redis redis-cli
127.0.0.1:6379> lrange queue:default 0 -1
1) "{\"retry\":0,\"queue\":\"default\",\"wrapped\":\"TestJob\",\"args\":[{\"job_class\":\"TestJob\",\"job_id\":\"b1a1e5a6-b46d-4103-a45c-08c48dc63888\",\"provider_job_id\":null,\"queue_name\":\"default\",\"priority\":null,\"arguments\":[{\"customer\":{\"_aj_globalid\":\"gid://lago-api/Customer/fed93a9b-f530-4285-a149-0226c10e7325\"},\"subscrption\":{\"_aj_globalid\":\"gid://lago-api/Subscription/d95ba2f9-7fa5-414f-a082-823d11b6283c\"},\"_aj_ruby2_keywords\":[\"customer\",\"subscrption\"]}],\"executions\":0,\"exception_executions\":{},\"locale\":\"en\",\"timezone\":\"UTC\",\"enqueued_at\":\"2025-09-18T14:08:10.393239836Z\",\"scheduled_at\":null}],\"class\":\"ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper\",\"jid\":\"0bcb6dec7953c3a804866ebc\",\"created_at\":1758204490.3939157,\"enqueued_at\":1758204490.3951573}"
```

vincent-pochet: `lifetime_usage` is, but `current_usage` is not a persisted object, it's an in-memory struct, so sadly, Global IDs will not really help here

app/services/usage_monitoring/process_subscription_activity_service.rb
vincent-pochet: The `using_lifetime_usage` remains for `LifetimeUsages::CalculateService` (and for the call of the `LifetimeUsages::RecalculateAndCheckJob`), bu the check for `progressive_billing_enabled` as been moved to the job.

Note also that using_life_time_usage is defined as:
```ruby  
def using_lifetime_usage?
  lifetime_usage_enabled? || progressive_billing_enabled?
end
```
So it covers both cases, the initial condition was a bit redundant.

app/jobs/lifetime_usages/recalculate_and_check_job.rb
groyoh: Makes sense! Thanks for the explanation.

I guess in that case, it would be nice in the future to extract this into a service or move the `LifetimeUsages::CheckThresholdsService.call(lifetime_usage:)` as part of the `LifetimeUsages::CalculateService` with a `check_threshold: true` kwarg. This would avoid people mistakenly calling the job with the param. I don't trust anyone (myself included) read comments 100% of the time 😄

groyoh (approved):

app/services/usage_monitoring/process_subscription_activity_service.rb
groyoh: Makes sense!

PR #4366, opened 2025-09-18
fix(netsuite): handle range issue

lovrocolic:
## Context

Netsuite does not allow any value to be larger than 9 999 999.

## Description

For AI companies, when large amount of tokens is consumed, `units` on fee can be really large. We will avoid such issue by setting quantity to `1` and setting unit price to be the same as fee amount, only in such edge cases.


app/services/integrations/aggregator/invoices/payloads/netsuite.rb
```diff
@@ -116,12 +116,22 @@ def item(fee)
             from_property = fee.charge? ? "charges_from_datetime" : "from_datetime"
             to_property = fee.charge? ? "charges_to_datetime" : "to_datetime"
 
+            ns_quantity_limit = 10_000_000_000 # NetSuite quantity max magnitude
```
ancorcruz: Could you define the max magnitude as a class constant?

```
  NS_QUANTITY_LIMIT = 10_000_000_000 # NetSuite quantity max magnitude
```

app/services/integrations/aggregator/invoices/payloads/netsuite.rb
```diff
@@ -116,12 +116,22 @@ def item(fee)
...
+
             {
               "item" => mapped_item.external_id,
               "account" => mapped_item.external_account_code,
-              "quantity" => limited_rate(fee.units),
-              "rate" => limited_rate(fee.precise_unit_amount),
-              "amount" => limited_rate(amount(fee.amount_cents, resource: invoice)),
+              "quantity" => quantity_value,
+              "rate" => unit_rate_value,
+              "amount" => line_amount_value,
```
ancorcruz, line 126: just in case, is there a limit on the amount value?

lovrocolic: If quantity is big, the unit price is small in that case, so fee amount is usually number in expected range. For amount itself, we should not apply any hack. I would say it is better if we end up with dead job in that case, but this case hasn't been spotted yet

app/services/integrations/aggregator/invoices/payloads/netsuite.rb
ivannovosad: Shouldn't this be `10_000_000`?

lovrocolic: Nope, the 10 billion is limitation

PR #4368, opened 2025-09-19
test(invoice): Refactor invoice tests with `match_html_snapshot` matcher

groyoh:
## Context

Today, we have some tests that rely on snapshot testing to test the rendering of the invoice. It was an ad-hoc solution that was not made re-usable or flexible.

## Description

In order to be able to re-use the same testing strategy in other tests, I have extracted the snapshot logic into a `match_html_snapshot` matcher.


julienbourdeau (approved):

groyoh:
Thanks @julienbourdeau 🙏

PR #4369, opened 2025-09-19
test(wallet): Test invoice rendering in wallet top-up scenarios tests

groyoh:
## Context

We currently have some scenario tests related to wallet top-up but they don't actually test the impact on the generated PDF.

## Description

This adds an assertion on the wallet top-up scenario test to render and test the invoice via snapshot testing.

In order to do that, I had to modify the `stub_pdf_generation` to use the rendered HTML instead of a blank PDF in the mocked response. This way, the invoice file will contain the rendered HTML which we can test using `expect(invoice.file.download).to match_html_snapshot`. Note that the mocked response is currently not used and has no impact the test suite.


vincent-pochet (approved):

groyoh:
Thanks @vincent-pochet 🙏

PR #4370, opened 2025-09-19
feat(taxes): Soft-delete Tax model

julienbourdeau:


app/services/taxes/destroy_service.rb
```diff
@@ -11,16 +13,35 @@ def initialize(tax:)
...
+
+        tax.applied_taxes.delete_all
+        tax.draft_fee_taxes.delete_all
+        tax.draft_invoice_taxes.delete_all
+        tax.credit_notes_taxes.delete_all
+        tax.add_ons_taxes.delete_all
+        tax.plans_taxes.delete_all
+        tax.charges_taxes.delete_all
+        tax.commitments_taxes.delete_all
+        tax.fixed_charges_taxes.delete_all
```
vincent-pochet, line 37: Don't we want to keep some of those applied taxes, especially on finalized resources?

PR #4372, opened 2025-09-22
fix(wallet): Fix behavior when `invoice_requires_successful_payment` is nil

groyoh:
## Context

Because `ActiveModel::Type::Boolean.new.cast(nil)` returns `nil` and the `invoice_requires_successful_payment` has a `NOT NULL` db constraint, we fail to create a wallet or wallet transaction when `invoice_requires_successful_payment` is set to `nil`.

## Description

This fixes this behavior by defaulting to `false` when creating a wallet and defaulting to the wallet's value when creating a transaction. Validations were also added on the model to ensure saving such value will fail.

### DIFF
--- a/app/services/wallets/create_service.rb
+++ b/app/services/wallets/create_service.rb
@@ -30,7 +30,7 @@ module Wallets
       }
 
       if params.key?(:invoice_requires_successful_payment)
-        attributes[:invoice_requires_successful_payment] = ActiveModel::Type::Boolean.new.cast(params[:invoice_requires_successful_payment])
+        attributes[:invoice_requires_successful_payment] = ActiveModel::Type::Boolean.new.cast(params[:invoice_requires_successful_payment]) || false
       end
 
       if params.key?(:applies_to)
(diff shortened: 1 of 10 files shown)


app/services/wallets/create_service.rb
```diff
@@ -30,7 +30,7 @@ def call
       }
 
       if params.key?(:invoice_requires_successful_payment)
-        attributes[:invoice_requires_successful_payment] = ActiveModel::Type::Boolean.new.cast(params[:invoice_requires_successful_payment])
+        attributes[:invoice_requires_successful_payment] = ActiveModel::Type::Boolean.new.cast(params[:invoice_requires_successful_payment]) || false
```
julienbourdeau, line 33: I'm starting to regret using this cast. It's great for query params but I think it was a mistake to use it for json payload. 

This shouldn't be accepted for json payload 🙈 
```json
 "invoice_requires_successful_payment": "off"
```

groyoh: Maybe we can add a custom casting:

```ruby
module Cast
  def self.to_boolean(value, strict: true)
    if strict
      return case value
      when true
        true
      when false, nil
        false
      else
        raise ArgumentError, "non castable value"
      end
    end
    
    ActiveModel::Type::Boolean.new.cast(value) || false
  end
end
```

?

PR #4373, opened 2025-09-22
Fix tax rates credit notes

vincent-pochet:
## Roadmap Task

## Context

## Description


spec/scenarios/credit_notes/credit_note_spec.rb
```diff
@@ -816,17 +825,17 @@
...
         estimate = json[:estimated_credit_note]
         expect(estimate).to include(
-          taxes_amount_cents: 1319,
+          taxes_amount_cents: 13_19,
           precise_taxes_amount_cents: "1319.25547",
-          sub_total_excluding_taxes_amount_cents: 6596,
-          max_creditable_amount_cents: 7916,
-          coupons_adjustment_amount_cents: 237,
+          sub_total_excluding_taxes_amount_cents: 65_96,
+          max_creditable_amount_cents: 79_15,
```
annvelents, line 831: I am trying to understand why these values changed 
aaaaah
aaah!!!!
Because in the previous approach, that we have at `credit_notes/create_service#adjust_credit_note_tax_rounding`, we were always adjusting the taxes, so it was not matematically correct, while now we're also adjusting total!
so definitely now we're using the `AdjustAmountsWithRoundingService`, so we can remove that method 🙌

spec/scenarios/credit_notes/credit_note_spec.rb
```diff
@@ -883,14 +893,15 @@
...
-          total_amount_cents: 23207,
-          taxes_amount_cents: 3868,
+          credit_amount_cents: 232_07,
+          total_amount_cents: 232_07,
+          taxes_amount_cents: 38_68,
           precise_taxes_amount_cents: 3868.00001,
-          precise_coupons_adjustment_amount_cents: 694.0239
+          precise_coupons_adjustment_amount_cents: 694.0239,
+          precise_total: 23206.97611,
+          taxes_rounding_adjustment: -0.00001
```
annvelents, line 902: and it means we can also get rid of this... (`taxes_rounding_adjustment`)

annvelents (approved):
Awesome!
I'm so glad we finally have a service to deal with this issue!

spec/scenarios/credit_notes/credit_note_spec.rb
vincent-pochet: Actually, I still need to rely on the `precise_taxes_amount_cents` with the rounding adjustment.
When I get rid of it I end-up having a lot of validation issues when trying to create the credit notes

PR #4375, opened 2025-09-22
feat(fixed_charges-16): call fixed charge fee creation from invoice service (3)

annvelents:
## Context

Finally when we create an invoice we'll also create a fixed charge fee for this subscription

## Description

- updated invoices/calculate_fee_service to create fixed_charge_fee with all the services created in previous PRs

### DIFF
--- a/app/services/invoices/calculate_fees_service.rb
+++ b/app/services/invoices/calculate_fees_service.rb
@@ -105,6 +105,8 @@ module Invoices
     end
 
     def fixed_charge_boundaries_valid?(boundaries)
+      return false if boundaries.fixed_charges_from_datetime.blank? || boundaries.fixed_charges_to_datetime.blank?
+
       boundaries.fixed_charges_from_datetime <= boundaries.fixed_charges_to_datetime
     end
 
--- a/spec/services/invoices/calculate_fees_service_spec.rb
+++ b/spec/services/invoices/calculate_fees_service_spec.rb
@@ -2438,6 +2438,32 @@ RSpec.describe Invoices::CalculateFeesService do
         end
       end
     end
+
+    # this case might happen because we only started populating fixed_charge boundaries in the last PR,
+    # and we can have some invoices that we're recalculating with old invoice_subscriptions
+    context "when invoice has an invoice_subscription with empty fixed_charge boundaries" do
+      let(:invoice_subscription) do
+        create(
+          :invoice_subscription,
+          subscription:,
+          invoice:,
+          timestamp:,
+          from_datetime: date_service.from_datetime,
+          to_datetime: date_service.to_datetime,
+          charges_from_datetime: date_service.charges_from_datetime,
+          charges_to_datetime: date_service.charges_to_datetime,
+          fixed_charges_from_datetime: nil,
+          fixed_charges_to_datetime: nil
+        )
+      end
+
+      it "does not fail and does not create fixed charge fees" do
+        result = invoice_service.call
+
+        expect(result).to be_success
+        expect(result.invoice.fees.fixed_charge.count).to eq(0)
+      end
+    end
   end
 
   describe "#should_create_yearly_subscription_fee?" do


app/services/subscriptions/free_trial_billing_service.rb
```diff
@@ -66,6 +68,11 @@ def ending_trial_subscriptions
           INNER JOIN customers ON subscriptions.customer_id = customers.id
           INNER JOIN billing_entities ON customers.billing_entity_id = billing_entities.id
           LEFT JOIN already_billed_today ON already_billed_today.subscription_id = subscriptions.id
+          LEFT JOIN (
+            SELECT DISTINCT plan_id, id as fixed_charge_id
+            FROM fixed_charges
+            WHERE pay_in_advance = true AND deleted_at IS NULL
+          ) pay_in_advance_fixed_charges ON pay_in_advance_fixed_charges.plan_id = plans.id
```
annvelents: check that all indexes exist

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -1774,15 +2350,16 @@
...
-        aggregate_failures do
-          expect(result).to be_success
-          expect(result.invoice.fees_amount_cents).to eq(20_000)
-          expect(result.invoice.taxes_amount_cents).to eq(1_550)
-          expect(result.invoice.sub_total_excluding_taxes_amount_cents).to eq(15_500)
-          expect(result.invoice.sub_total_including_taxes_amount_cents).to eq(17_050)
-          expect(result.invoice.progressive_billing_credit_amount_cents).to eq(3_000)
-          expect(result.invoice.total_amount_cents).to eq(9_727) # 17_050 - 1_000 (credit note) - 6_323 (wallet)
-        end
+        expect(result).to be_success
```
annvelents, line 2334: what is this?

spec/services/subscriptions/create_service_spec.rb
```diff
@@ -822,14 +836,19 @@
               fixed_charge_2
             end
 
-            it "does not create fixed charge events for the new subscription" do
+            it "creates fixed charge events for the subscription" do
```
annvelents: WHe did I change to create the event? 🤔

app/views/templates/invoices/v4/_charge.slim
```diff
@@ -6,6 +6,8 @@ table.invoice-resume-table width="100%"
     td.body-2 = I18n.t('invoice.tax_rate')
     td.body-2 = I18n.t('invoice.amount')
   - fees.order(:succeeded_at, :created_at).each do |fee|
+    - if fee.fixed_charge?
```
annvelents, line 9: Need to delete this!

app/services/fees/charge_service.rb
```diff
@@ -378,5 +378,13 @@ def calculate_period_ratio
...
       ratio.clamp(0.0, 1.0)
     end
+
+    def filtered_for_charge_boundaries(boundaries)
+      properties = boundaries.to_h
+      properties["fixed_charges_from_datetime"] = nil
+      properties["fixed_charges_to_datetime"] = nil
+      properties["fixed_charges_duration"] = nil
+      properties
+    end
```
ancorcruz, line 388: question... Why do we set fixed charge boundaries in the charge fee as nil (same question applies for the fixed charge fee and the charge boundaries?

Aside of the boundary itself does not make sense at all... we are not doing the same for the subscription fee boundary for charges nor fixed charges fees.

app/services/invoices/calculate_fees_service.rb
```diff
@@ -293,7 +332,29 @@ def should_create_charge_fees?(subscription)
       true
     end
 
+    def should_create_fixed_charge_fees?(subscription, boundaries)
+      fee_count = subscription.fees
+        .fixed_charge
+        .includes(:invoice)
```
ancorcruz: we don't need to eager load `invoices` as the query is only checking the `invoice_id` that is already fees table column.

app/services/invoices/calculate_fees_service.rb
```diff
@@ -293,7 +332,29 @@ def should_create_charge_fees?(subscription)
...
+      #       fee if the plan is in pay in arrears, otherwise this fee will never
+      #       be created.
+      subscription.active? ||
+        (subscription.terminated? && subscription.plan.fixed_charges.pay_in_arrears.any?) ||
+        (subscription.terminated? && subscription.terminated_at > invoice.created_at)
+    end
+
     def wallets
+      return @wallets if @wallets
+
```
ancorcruz: Do we need this? `@wallets` is already memoized and returns its value if not nil. What the is the goal of this line.

app/services/invoices/calculate_fees_service.rb
```diff
@@ -138,6 +146,37 @@ def should_not_create_charge_fee?(charge, subscription)
...
+          context:
+        )
+      end
+    end
+
+    # In current PR we just always create the fixed charges. In the upcoming we'll handle upgrade/downgrade/termination scenarios
+    def should_create_fixed_charge_fee?(fixed_charge, subscription)
+      # when "starting" invoice - it's only for pay_in_advance fees
+      if !fixed_charge.pay_in_advance? && subscription.invoice_subscriptions.count == 1 &&
+          subscription.invoice_subscriptions.first.subscription_starting?
```
ancorcruz: these complex conditions would be great to have them encapsulated in a method with a self describing naming. It is kind of hard to follow.

app/services/subscriptions/free_trial_billing_service.rb
ancorcruz: have you checked?

app/views/templates/invoices/v4/_charge.slim
ancorcruz: this partial is only used within the `v4/charge` template, this is to invoice a standalone charge... fixed charge does not make sense within this context. I think this can be safely removed.

spec/factories/fixed_charges.rb
```diff
@@ -9,6 +9,8 @@
     units { 1 }
     properties { {amount: Faker::Number.between(from: 100, to: 500).to_s} }
     invoice_display_name { Faker::Fantasy::Tolkien.location }
+    pay_in_advance { false }
+    prorated { false }
```
ancorcruz: we won't need to define the already default values for the columns...

spec/services/subscriptions/free_trial_billing_service_spec.rb
```diff
@@ -73,5 +73,23 @@
           .to have_received(:perform_later).with(subscription: subscription)
       end
     end
+
+    context "with plan pay in arrears" do
+      context "when plan has fixed charges" do
+        context "when fixed_charges are pay in advance" do
```
ancorcruz, line 81: should we also include the scenario for fixed charges pay in arrears here?

spec/services/subscriptions/create_service_spec.rb
```diff
@@ -338,37 +352,37 @@
...
-    context "when plan is pay_in_advance and subscription_at is current date" do
-      let(:plan) { create(:plan, amount_cents: 100, organization:, pay_in_advance: true) }
+        it "does not enqueue a job to bill the subscription" do
+          expect { create_service.call }.not_to have_enqueued_job(BillSubscriptionJob)
+        end
+      end
 
-      it "enqueues a job to bill the subscription" do
-        expect { create_service.call }.to have_enqueued_job(BillSubscriptionJob)
+      context "when ubscription_at is current date" do
```
ancorcruz:
```suggestion
      context "when subscription_at is current date" do
```

spec/services/fees/fixed_charge_service_spec.rb
```diff
@@ -420,5 +420,38 @@
...
+      end
+
+      it "creates a fee with current boundaries" do
+        result = fixed_charge_service.call
+        expect(result).to be_success
+        # subscription started at 2022-03-17, so all charges only start from 17th
+        expect(result.fee.properties).to include(
+          "fixed_charges_from_datetime" => Time.parse("2022-03-17T00:00:00.000Z"),
+          "fixed_charges_to_datetime" => Time.parse("2022-03-31T23:59:59.999Z"),
+          "fixed_charges_duration" => 31
```
ancorcruz, line 572: why is duration 31? shouldn't it be 31 - 17 as the from date is March 17th instead of March 1st?

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
           expect(Fee.charge.where(invoice_id: nil).count).to eq(1)
         end
       end
+
+      context "when plan has fixed charges" do
```
ancorcruz: plan already had in the top context.... isn't this context redundant?

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+
+          it "creates a fixed charge fee with period of the next month" do
+            result = invoice_service.call
+
+            expect(result).to be_success
+            expect(invoice.fees.fixed_charge.count).to eq(1)
+            fee_properties = invoice.fees.fixed_charge.first.properties
+            expect(fee_properties["fixed_charges_from_datetime"]).to match_datetime(Time.parse("2022-03-06T00:00:00.000Z"))
+            expect(fee_properties["fixed_charges_to_datetime"]).to match_datetime(Time.parse("2022-04-05T23:59:59.999Z"))
+          end
```
ancorcruz: first test is duplicated in the second, lets put this test on a diet.

```suggestion
          it "creates a fixed charge fee with period of the next month" do
            result = invoice_service.call

            expect(result).to be_success
            expect(invoice.fees.fixed_charge.count).to eq(1)
            fee_properties = invoice.fees.fixed_charge.first.properties
            expect(fee_properties["fixed_charges_from_datetime"]).to match_datetime(Time.parse("2022-03-06T00:00:00.000Z"))
            expect(fee_properties["fixed_charges_to_datetime"]).to match_datetime(Time.parse("2022-04-05T23:59:59.999Z"))
          end
```

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+              create(:fixed_charge_event, fixed_charge: new_fixed_charge1, subscription: new_subscription, timestamp: timestamp, units: 1)
+            end
+
+            let(:new_fixed_charge2_event) do
+              create(:fixed_charge_event, fixed_charge: new_fixed_charge2, subscription: new_subscription, timestamp: timestamp, units: 1)
+            end
+
+            before do
+              old_fixed_charge
+              fixed_charge.update(pay_in_advance: false)
```
ancorcruz: why not redefine the fixed charge with the config we would like to have?

Here `update` will silently fail and continue with the test hiding real behaviour, if you actually want to update use `update!` instead but I would highly recommend redefining the variable with the right config.

```
let(:fixed_charge) do
  create(:fixed_charge, .... pay_in_advance: false)
end
```

even `pay_in_advance` is already false by default... hence no need to be explicit here.

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+            context "when subscription is upgraded" do
+              context "when passing old subscription in the service" do
+                let(:subscription) { old_subscription }
+
+                it "does not create a fixed charge fee for the previous subscription" do
+                  result = invoice_service.call
+
+                  expect(result).to be_success
+                  expect(subscription.fixed_charges.count).to eq(2)
+                  expect(invoice.fees.fixed_charge.count).to eq(1)
```
ancorcruz: to which subs this fee belongs to?

I would like to understand what is going on her reading the expectations.

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+                end
+              end
+
+              # issuing a new fee for pay_in_advance charge, terminated subscription
+              context "when passing new subscription in the service" do
+                let(:subscription) { new_subscription }
+
+                before do
+                  invoice_subscription[:invoicing_reason] = :subscription_starting
+                  invoice_subscription.save
```
ancorcruz: always use `!` bang methods to ensure the test fail if the save/update fails on the setup.

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+                end
+
+                it "creates a fixed charge fee for the new subscription only for pay in advance charges" do
+                  result = invoice_service.call
+
+                  expect(result).to be_success
+                  expect(subscription.fixed_charges.count).to eq(3)
+                  expect(invoice.fees.fixed_charge.count).to eq(2)
```
ancorcruz: not sure what is going on here... it creates a fixed charge for the new subs only for pay in advance charge but the expected results are 3 fixed charges in the subscription and 2 fixed charges fees in the invoice... I'm not sure what does this means.... where is data come from, etc.

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+              let(:new_plan) { create(:plan, organization:, interval:, pay_in_advance:, trial_period:, amount_cents: 2000) }
+
+              context "when passing old subscription in the service" do
+                let(:subscription) { old_subscription }
+
+                it "does create a fixed charge fee for the previous subscription" do
+                  result = invoice_service.call
+
+                  expect(result).to be_success
+                  expect(invoice.fees.fixed_charge.count).to eq(1)
```
ancorcruz: shall we check the fee created also belongs to the old subscription? Same question applies to the upgrade scenario above.

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+                let(:fixed_charge) do
+                  create(
+                    :fixed_charge,
+                    plan: new_plan,
+                    charge_model: "standard",
+                    pay_in_advance: false,
+                    properties: {amount: "10"},
+                    units: 1
+                  )
+                end
```
ancorcruz: does this has an event?

also, the context is about pay in advance fixed charges and this fixed charge is set a pay in arrears.

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+
+          it "creates a fixed charge fee with period of this month" do
+            result = invoice_service.call
+
+            expect(result).to be_success
+            expect(invoice.fees.fixed_charge.count).to eq(1)
+            fee_properties = invoice.fees.fixed_charge.first.properties
+            expect(fee_properties["fixed_charges_from_datetime"]).to match_datetime(Time.parse("2022-02-06T00:00:00.000Z"))
+            expect(fee_properties["fixed_charges_to_datetime"]).to match_datetime(Time.parse("2022-03-05T23:59:59.999Z"))
+          end
```
ancorcruz: no need for duplicated tests... this is already long enough


```suggestion
          it "creates a fixed charge fee with period of this month" do
            result = invoice_service.call

            expect(result).to be_success
            expect(invoice.fees.fixed_charge.count).to eq(1)
            fee_properties = invoice.fees.fixed_charge.first.properties
            expect(fee_properties["fixed_charges_from_datetime"]).to match_datetime(Time.parse("2022-02-06T00:00:00.000Z"))
            expect(fee_properties["fixed_charges_to_datetime"]).to match_datetime(Time.parse("2022-03-05T23:59:59.999Z"))
          end
```

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+                  result = invoice_service.call
+
+                  expect(result).to be_success
+                  expect(invoice.fees.fixed_charge.count).to eq(0)
+                end
+              end
+            end
+          end
+        end
+      end
```
ancorcruz, line 617: same question for pay in advance and pay in arrears tests, it is hard to easily understand what is going on when reading the tests, it would be great to improve the assertions to ensure behavior can be understood at first glance.

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+                  expect(invoice.fees.fixed_charge.count).to eq(1)
+                end
+              end
+
+              context "when passing new subscription in the service" do
+                let(:subscription) { new_subscription }
+
+                it "does not create a fixed charge fee pay in arrears fees when billing for :subscription_starting" do
+                  invoice_subscription[:invoicing_reason] = :subscription_starting
+                  invoice_subscription.save
```
ancorcruz:
```suggestion
                  invoice_subscription.save!
```

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
```
ancorcruz:
```suggestion
                  invoice_subscription.save!
```

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -356,49 +365,427 @@
...
+              )
+            end
+
+            let(:new_fixed_charge1_event) do
+              create(:fixed_charge_event, fixed_charge: new_fixed_charge1, subscription: new_subscription, timestamp: timestamp, units: 1)
+            end
+
+            before do
+              old_fixed_charge
+              fixed_charge.update(pay_in_advance: false)
```
ancorcruz:
```suggestion
              fixed_charge.update!(pay_in_advance: false)
```

even better redefine the fixed_charge variable in a let block with the desired config.

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -425,6 +812,114 @@
...
+
+          expect(invoice.subscriptions.first).to eq(subscription)
+          expect(invoice.fees.subscription.count).to eq(1)
+          expect(invoice.fees.fixed_charge.count).to eq(1)
+          expect(invoice.fees.commitment.count).to eq(0)
+
+          fixed_charge_fee = invoice.fees.fixed_charge.first
+          expect(fixed_charge_fee.fixed_charge).to eq(fixed_charge)
+          expect(fixed_charge_fee.units).to eq(3)
+          expect(fixed_charge_fee.amount_cents).to eq(3000)
```
ancorcruz, line 796: it would be great to see the math behind this magic amount number...

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -853,11 +1376,120 @@
...
+            properties: {amount: "10"}
+          )
+        end
+
+        let(:fixed_charge_event) do
+          create(:fixed_charge_event, fixed_charge:, subscription:, timestamp: event_timestamp, units: 5, created_at: event_timestamp)
+        end
+
+        context "when plan has no minimum commitment" do
+          it "creates a fixed charge fee but no minimum commitment fee", transaction: false do
```
ancorcruz, line 1302: why is `transaction: false` needed here?

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -853,11 +1376,120 @@
...
+          )
+        end
+
+        let(:fixed_charge_event) do
+          create(:fixed_charge_event, fixed_charge:, subscription:, timestamp: event_timestamp, units: 5, created_at: event_timestamp)
+        end
+
+        before do
+          invoice_subscription[:invoicing_reason] = :subscription_starting
+          invoice_subscription.save
```
ancorcruz:
```suggestion
          invoice_subscription.save!
```

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -853,11 +1376,120 @@
...
+              pay_in_advance: true,
+              properties: {amount: "10"}
+            )
+          end
+
+          it "does not create a fixed charge fee for the old subscription" do
+            result = invoice_service.call
+
+            expect(result).to be_success
+            expect(invoice.fees.fixed_charge.count).to eq(0)
```
ancorcruz, line 1348: shouldn't the invoice include one fixed charge for the new subscription upgrade as it is pay in advance fixed charge?

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -853,11 +1376,120 @@
...
+              properties: {amount: "10"}
+            )
+          end
+
+          it "does create a fixed charge fee for the old subscription pay in arrears charge" do
+            result = invoice_service.call
+
+            expect(result).to be_success
+            expect(invoice.fees.fixed_charge.count).to eq(0)
           end
```
ancorcruz: shouldn't the pay in arrears fixed charge  be billed in the terminating subscription on the upgrade?

spec/services/invoices/calculate_fees_service_spec.rb
```diff
@@ -1297,6 +1995,11 @@
               expect(invoice.subscriptions.first).to eq(subscription)
               expect(invoice.fees.subscription.count).to eq(0)
               expect(invoice).to have_empty_charge_fees # Because we didn't fake usage events
+
+              # plan is billed anniversary on 21 of November (started_at, 2024). At the timestamp, 1st Jan 2025,
+              # previous billing period is "empty" - 21Nov 2024 - 21 Nov 2024, because it's not current usage, so we get an "empty"
+              # boundaries and cannot fetch any usage
+              expect(invoice.fees.fixed_charge.count).to eq(0)
```
ancorcruz, line 1878: is this the expected behavior or work in pogress?

spec/scenarios/billing/billing_monthly_spec.rb
```diff
@@ -141,20 +161,48 @@
       last_invoice = subscription.invoices.order(:created_at).last
 
       # yet fixed_charge is 0, will be 4...
```
ancorcruz: remove the outdated comments please.

spec/scenarios/billing/billing_monthly_spec.rb
```diff
@@ -243,7 +335,9 @@
       expect(subscription.reload.invoices.count).to eq(10)
       last_invoice = subscription.invoices.order(:created_at).last
       # 0 is only for now!
```
ancorcruz: remove this please.

app/services/fees/charge_service.rb
annvelents: I think it exactly does not make sense to have some random boundaries that have no relation to the fee itself.
What I mean by random:
a plan can have only pay_in_advance fixed_charges, but when creating a fee for standard charge, we'll use "default" boundaries, inherited from the subscription, which are boundaries for pay_in_arrears charges (so they have nothing to do with actual pay_in_advance charges). Since we have just introduced these fields in fee (fixed_charge_from, fixed_charge_to, fixed_charge_duration), I think it's the right time to make the call and not include an information that has no relation to this model (fee related to a normal charge) and makes no sense as well

app/services/invoices/calculate_fees_service.rb
annvelents: ah, I got it due to merge conflicts, I guess, I accidentally left a line that was deleted later

app/services/subscriptions/free_trial_billing_service.rb
annvelents: right, we're missing pay_in_advance index

spec/services/fees/fixed_charge_service_spec.rb
annvelents: because this duration is used in calculation of the proration, so should be the whole month

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: at this point in the subscription that is passed to the service, that matches old_subscription, we have pay_in_advance and pay_in_arrears fixed_charge fees, and only pay_in_arrears will be charges

spec/services/subscriptions/free_trial_billing_service_spec.rb
annvelents: completely removing this logic as it should be done with FixedChargesPayInAdvance service, added later

app/services/subscriptions/free_trial_billing_service.rb
annvelents: removed this logic, as it will be rewritten completely and if we merge in the current state we need to add an index

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: yes, event is created at the beginning of the spec...
about the context - it got mixed... as this is upgrade / downgrade scenario, there are actually multiple fixed_charges of different types created on all subscriptions... I'll think how to get it untangled...

spec/scenarios/billing/billing_monthly_spec.rb
```diff
@@ -115,16 +115,36 @@
...
-      # pay_in_advance_fixed_charges_invoice = subscription.invoices.first
-      # this will be a part of another PR
-      # expect(pay_in_advance_fixed_charges_invoice.fees.fixed_charge.count).to eq(2)
+      expect(subscription.reload.invoices.count).to eq(1)
+      pay_in_advance_fixed_charges_invoice = subscription.invoices.first
+      expect(pay_in_advance_fixed_charges_invoice.fees.count).to eq(2)
+      # check fixed_charge fees
+      expect(pay_in_advance_fixed_charges_invoice.fees.fixed_charge.count).to eq(2)
+      # fixed_charge_not_prorated_in_advance: 200 * 100 * 10, fixed_charge_prorated_in_advance: 2000 * 100 * 10 * 26/29
+      expect(pay_in_advance_fixed_charges_invoice.fees.fixed_charge.map(&:amount_cents).sort).to match_array([179_310, 200_000])
```
ancorcruz, line 125: this is okay but I prefer more precise specs, matching the amount with the right fixed charge fee.

ancorcruz (approved):
LGTM, still a few comments to be considered.

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: updated these test to compare with named fixed_charges (old/new in_advance/in_arrears)

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: for both: upgrade and downgrade we have a test for passing old subscription and new subscription

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: fixed the contexts around these scenarios 🙌

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: reworked them, I think now it's more readable

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: added a comments explaining the math

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: because without this the event will not be saved, and the math doesn't match

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: it will be created for the new subscription, but here we're passing the old subscription only

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: you're right, it should be created, checking what's going on. Hm, the title also says that it shold be created, when I run the test - it was created... not sure how this 0 got here and why it was passing successfully 🤷

spec/services/invoices/calculate_fees_service_spec.rb
annvelents: yes, that's expected behaviour for these boundaries, as we're trying to find events with timestamp before the interval end, but our timestamp matches interval beginning and also matches interval end, so we can't return anything

ancorcruz:
:shipit:

PR #4383, opened 2025-09-23
feat(customer): Add `currencies[]` filter to `GET /api/v1/customers`

groyoh:
## Context

The filtering on the `GET /api/v1/customers` endpoint is too limited.

## Description

This adds a `currencies[]` filter to this endpoint.

Also, some of the existing filters were not properly validated. They are now validated.


groyoh:
Thanks @ancorcruz ! 🙏

PR #4384, opened 2025-09-23
fix(billing): snapshot customer data on invoice finalization

ancorcruz:
## Context

Finalized invoices were displaying live customer data instead of data that
existed at finalization time, causing inconsistency between tax calculations
and invoice display.

When customers updated their address after an invoice was finalized, the PDF
(if regenerated) and dashboard would show the new address while taxes were
calculated using the original address.

 ## Description

Add customer data snapshotting to preserve customer information at invoice
finalization time.

When an invoice transitions to finalized status, all relevant customer data
(name, email, billing address, legal information, etc.) is captured and stored
in dedicated invoice columns.

This ensures tax calculation consistency and provides accurate historical
customer information on invoices regardless of subsequent customer data
changes.

Existing data will be migrated in a later PR, the same applies to pdf invoice generation.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -4267,6 +4267,26 @@ type CustomerProjectedUsage {
   totalAmountCents: BigInt!
 }
 
+type CustomerSnapshot {
+  addressLine1: String
+  addressLine2: String
+  applicableTimezone: String
+  city: String
+  country: String
+  displayName: String
+  email: String
+  firstname: String
+  lastname: String
+  legalName: String
+  legalNumber: String
+  phone: String
+  shippingAddress: CustomerAddress
+  state: String
+  taxIdentificationNumber: String
+  url: String
+  zipcode: String
+}
+
 enum CustomerTypeEnum {
   company
   individual
@@ -5746,6 +5766,7 @@ type Invoice {
   creditableAmountCents: BigInt!
   currency: CurrencyEnum
   customer: Customer!
+  customerSnapshot: CustomerSnapshot
   errorDetails: [ErrorDetail!]
   externalHubspotIntegrationId: String
   externalIntegrationId: String
--- a/schema.json
+++ b/schema.json
@@ -19058,6 +19058,221 @@
           "enumValues": null
         },
         {
+          "kind": "OBJECT",
+          "name": "CustomerSnapshot",
+          "description": null,
+          "interfaces": [],
+          "possibleTypes": null,
+          "fields": [
+            {
+              "name": "addressLine1",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "addressLine2",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "applicableTimezone",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "city",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "country",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "displayName",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "email",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "firstname",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "lastname",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "legalName",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "legalNumber",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "phone",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "shippingAddress",
+              "description": null,
+              "type": {
+                "kind": "OBJECT",
+                "name": "CustomerAddress",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "state",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
+              "isDeprecated": false,
+              "deprecationReason": null,
+              "args": []
+            },
+            {
+              "name": "taxIdentificationNumber",
+              "description": null,
+              "type": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              },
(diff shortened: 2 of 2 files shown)


app/models/invoice.rb
```diff
@@ -530,6 +595,30 @@ def set_finalized_at
 
     self.finalized_at ||= Time.current
   end
+
+  def snapshot_customer_data
+    return unless status_changed_to_finalized?
```
groyoh: Maybe this could be useful with a manual action:

```suggestion
  def snapshot_customer_data(force: false)
    return unless status_changed_to_finalized? || force
```

app/models/invoice.rb
```diff
@@ -411,6 +412,70 @@ def allow_manual_payment?
     MANUALLY_PAYABLE_INVOICE_STATUS.include?(status.to_sym)
   end
 
+  def customer_display_name
```
groyoh: I feel like this could be simplified:

```ruby
[
  :display_name,
  :address_line1,
  #...
].each do |customer_attribute|
  define_method("customer_#{customer_attribute}") { customer_data_snapshotted_at? ? super : customer.public_send(customer_attribute) }
end
```

db/migrate/20250922113901_add_customer_snapshot_to_invoices.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class AddCustomerSnapshotToInvoices < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_column :invoices, :customer_data_snapshotted_at, :datetime
```
groyoh: Is there a reason you don't use `change_table :invoices, bulk: true`  ?

app/graphql/types/invoices/object.rb
```diff
@@ -83,6 +83,29 @@ def payable_type
...
+          c.address_line1 = object.customer_address_line1
+          c.address_line2 = object.customer_address_line2
+          c.city = object.customer_city
+          c.state = object.customer_state
+          c.zipcode = object.customer_zipcode
+          c.country = object.customer_country
+          c.legal_name = object.customer_legal_name
+          c.legal_number = object.customer_legal_number
+        end
+      end
```
groyoh: What about 

```suggestion
      class CustomerDecorator
        [
          :display_name,
          :firstname,
          # ..
          :legal_number
        ].each { |attr| define_method(attr) { public_send("customer_#{attr}") }
      end
      def customer
        CustomerDecorator.new(object.customer)
      end
```

?

That way we don't need a trick on the `Invoice` model to please GraphQL 😉

app/serializers/v1/invoice_serializer.rb
```diff
@@ -32,7 +32,24 @@ def serialize
         self_billed: model.self_billed,
         created_at: model.created_at.iso8601,
         updated_at: model.updated_at.iso8601,
-        voided_at: model.voided_at&.iso8601
+        voided_at: model.voided_at&.iso8601,
+        customer_data_snapshotted_at: model.customer_data_snapshotted_at&.iso8601,
+        customer_display_name: model.customer_display_name,
```
groyoh: I wonder if it will be clear for customers that these are the snapshotted attributes. Have you considered having a nested `snapshotted_customer` or  `snapshotted_customer_data` since you already use this naming ?

db/migrate/20250922113901_add_customer_snapshot_to_invoices.rb
julienbourdeau: Last time I used `bulk: true`, strong_migration complained but I don't remember why. It seemed legit.

app/models/invoice.rb
ancorcruz: great idea, I'll add it. thanks.

app/serializers/v1/invoice_serializer.rb
ancorcruz: good point however customer_* attributes are within the invoice body. It is supposed to be part of the invoice itself, "snapshotted_customer_data" sound like insider info (we could even remove the timestamp flag... but could be useful to match with the finalized_at (if we have it).

db/migrate/20250922113901_add_customer_snapshot_to_invoices.rb
```diff
@@ -0,0 +1,25 @@
...
+    add_column :invoices, :customer_city, :string
+    add_column :invoices, :customer_zipcode, :string
+    add_column :invoices, :customer_state, :string
+    add_column :invoices, :customer_country, :string
+    add_column :invoices, :customer_phone, :string
+    add_column :invoices, :customer_url, :string
+    add_column :invoices, :customer_tax_identification_number, :string
+    add_column :invoices, :customer_applicable_timezone, :string
+    add_column :invoices, :customer_firstname, :string
+    add_column :invoices, :customer_lastname, :string
```
julienbourdeau: I understand it might be a little late and it's annoying to refactor but I'm thinking this could be stored as a json object 🤔 
We never use this for query, but we might have to add more details.

```rb
add_column :invoices, :customer_data, :jsonb
```

In the invoice model we return this raw json as a proper Data object

```rb
InvoiceCustomer = Data.define(:display_name, :city, ...)

class Invoice < ApplicationRecord

  def customer_data
    customer_data_snapshotted_at ? Data.new(**read_attribute(:customer_data)) : do_it_from_relation(..)
  end
end
```

**EDIT: rename from customer to customer_data (customer is already the belongs_to relationship name**

julienbourdeau (approved):
👏 👏 👏

app/graphql/types/invoices/object.rb
julienbourdeau: I think here we need a new object. Other graphql would consider this to be the object relationship and can ask things from the customers like `customer.organization.id`, which would break here

```rb
field :customer, Types::Invoices::CustomerObject, null: false
```

app/serializers/v1/invoice_serializer.rb
julienbourdeau: I think of we have the info as dedicated column for each attribute, I'd keep it at the root level be "closer to the DB".
But I still think it would be even better to have `customer_data: model.customer_data` 😄

db/migrate/20250922113901_add_customer_snapshot_to_invoices.rb
groyoh: I guess it might just say it cannot know whether it's truly safe in a `change_table` and we are responsible.

app/graphql/types/invoices/object.rb
groyoh: @julienbourdeau I forgot to add `< SimpleDelegator` in my suggestion 🤦 

```ruby
      class CustomerDecorator < SimpleDelegator
        [
          :display_name,
          :firstname,
          # ..
          :legal_number
        ].each { |attr| define_method(attr) { public_send("customer_#{attr}") }
      end
      def customer
        CustomerDecorator.new(object.customer)
      end
```

So no issue with other methods.

app/serializers/v1/invoice_serializer.rb
groyoh: My main concern is that we'll both serialize `customer_*` attributes and a `customer` relationship/objet as well, which could lead to confusion as in "which customer data am I supposed to use ?". That's why I suggested the `snapshotted_` prefix.

app/graphql/types/invoices/object.rb
ancorcruz: sure thing, I have added a delegator with simple delegator and remove the attribute writer for display_name in customer... I did not like it tbh.

db/migrate/20250922113901_add_customer_snapshot_to_invoices.rb
ancorcruz: yes, @julienbourdeau that is because strong migration can not evaluate what happen within the change_table block and requires it to be included within a safety_assured block.

app/models/invoice.rb
ancorcruz: I have moved this into a concern.

db/migrate/20250922113901_add_customer_snapshot_to_invoices.rb
ancorcruz: @julienbourdeau I did consider this approach but as I could not anticipate for sure that we are not going to filter invoices by the customer name (or whatever other info) existing at invoice finalization point in time. I would add the columns instead.

app/serializers/v1/invoice_serializer.rb
ancorcruz: @groyoh My reasoning is: customer_* data within the invoice body is the customer data included in the invoice, the customer object is a related object but not the data included in the invoice. It makes sense for me at least. I'll ask others to hear more opinions.

ancorcruz: @vincent-pochet @rsempe @annvelents What do you think?

groyoh: > @groyoh My reasoning is: customer_* data within the invoice body is the customer data included in the invoice, the customer object is a related object but not the data included in the invoice. It makes sense for me at least. I'll ask others to hear more opinions.

Yeah that's fine with me. Just wanted to make sure we're all aligned on the potential confusion. 😉 If we're all okay with this I wonder if we shouldn't keep the same logic on the GraphQL side then rather than overriding `customer` ?

ancorcruz: @groyoh that is a good point... as overriding the customer present a glitch in UI of a few milliseconds where you can see the actual customer info then change to the snapshotted info... I'll check with FE team.

annvelents: I am actually thinking if we'll need similar thing for fees... because we have these pay_in_advance non-invoiceable fees, for which we do apply taxes, but we do not create an invoice for them - `Events::PayInAdvanceService` -> `Fees::CreatePayInAdvanceJob`

and in this case, would it make sense to create a separate table of customer_snapshot to not have similar confusion there? though, there are no PDFs, and I guess, it's harder to spot the misalignment...

annvelents: as for the exact naming, i might have a little more preference to `snapshotted_customer_*`, but I'm good with `customer_*` too (especially because there is `customer_data_snapshotted_at` which helps to identify that this is the customer data that was used for the invoice).
And a question: since we send all customer data, related to the invoice within the invoice, shall we remove serialization of the full customer in the invoice?
I'm talking about:`payload.merge!(customer)`

rsempe: I'm personally totally fine with merging this and using `customer_*` without any prefix. That's a good improvement 👏 - Having the table `customer_snapshots` could be the ideal solution for later 👍

PR #4385, opened 2025-09-24
chore(seeders): Rework all seeders

julienbourdeau:
## Context

We use seeders to seed dev environment.
We use `signup:seed_organization` to seed staging.

## Description

The hooli organization ID is now forced to `11111111-2222-3333-4444-555555555555`.

The API key is hardcoded so re-seeding your env will not require you to updated you script or Postman config.

The main goal is to:
- Re-order seeders
- Create expected stuff so other seeders can rely on it (like plans and John Doe subscription)
- Use services when possible to create correct data (using factories isn't always correct)

Other details:
- I have removed webhooks because we use services, it already creates proper webhooks
- Credit notes were creating **very weird** data, now there is only one for John Doe created with the service

Expected data:

key | value
-- | --
hooli org id | 11111111-2222-3333-4444-555555555555
hooli api key | lago_key-hooli-1234567890
customer external id | cust_john-doe
subscription external id | sub_john-doe-main

## Next steps

- Use `01_base.rb` instead of `signup:seed_organization` and maybe reuse env variable to make base seeder customizable.
- Add Stripe payment provider seeder
- Update openapi to use these values in examples?

CleanShot 2025-09-24 at 10 20 00@2x

CleanShot 2025-09-24 at 10 19 29@2x

CleanShot 2025-09-24 at 10 19 11@2x


db/seeds/03_john_doe.rb
```diff
@@ -0,0 +1,105 @@
...
+    paid_top_up_min_amount_cents: 12_00,
+    recurring_transaction_rules: [
+      {
+        granted_credits: "10",
+        interval: :weekly,
+        method: :fixed,
+        expiration_at: 1.year.from_now,
+        trigger: :interval,
+        transaction_metadata: [{key: :origin, value: :seeder}],
+        transaction_name: "10 credits for free 🎁"
```
groyoh: To trigger a transaction today:

```suggestion
        transaction_name: "10 credits for free 🎁",
        started_at: 1.week.ago
```

db/seeds/01_base.rb
```diff
@@ -55,246 +59,55 @@
...
-  balance: BigDecimal("86.00"),
-  ongoing_balance: BigDecimal("86.00"),
-  ongoing_usage_balance: BigDecimal("86.00"),
-  credits_balance: BigDecimal("86.00"),
-  credits_ongoing_balance: BigDecimal("86.00"),
-  credits_ongoing_usage_balance: BigDecimal("86.00"),
-  consumed_credits: BigDecimal("114.00"),
-  currency: "EUR"
-)
+# == Coupons
```
groyoh, line 88: Why did you move coupons here ?

groyoh (approved):
Nice work! A bit hard to review given the number of changes though 😅

db/seeds/01_base.rb
julienbourdeau: There was only 2 and they feel like "base setup" just like addons or taxes for me. They are not used or assocated with customers/invoices

groyoh: >There was only 2 and they feel like "base setup" just like addons or taxes for me. They are not used or assocated with customers/invoices

If we use this logic, plan should be here as well 🙈

julienbourdeau: Actually yeah, they should! I'll move them 😄

PR #4387, opened 2025-09-24
chore(stripe): Remove "beta" for Stripe pre-auth feature

julienbourdeau:
## Context

This feature was introduced in March in #3262 as a _beta_ feature.

## Description

The _beta_ aspect was handled by a premium integration. I'm removing the premium integration to release the feature in _GA_.

Note that "removing" integration isn't something too common, all apps with this integration would be considered invalid. 
I propose to introduce a list of previous integrations. 

Another way is a db migration to remove the data but it creates "down time" where the feature is removed from DB but the code still rely on this flag. I think this can be done later, if the list grows.

### DIFF
--- a/spec/requests/api/v1/subscriptions_controller_spec.rb
+++ b/spec/requests/api/v1/subscriptions_controller_spec.rb
@@ -266,110 +266,97 @@ RSpec.describe Api::V1::SubscriptionsController, type: :request do
     end
 
     context "with payment pre-authorization" do
-      context "when the feature isn't enabled" do
-        let(:body) { {authorization: {}, subscription: params} }
-
-        it "returns a forbidden error" do
-          subject
-
-          expect(response).to have_http_status(:forbidden)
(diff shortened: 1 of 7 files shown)


spec/requests/api/v1/subscriptions_controller_spec.rb
```diff
@@ -266,112 +266,99 @@
...
-          expect(json[:message]).to match(/beta_payment_authorization/)
-        end
+      it "returns a success" do
+        allow(PaymentProviders::CancelPaymentAuthorizationJob).to receive(:perform_later)
+
+        subject
+        expect(json[:authorization]).to include(stripe_pi)
+        expect(json[:subscription]).to include(status: "active")
+
+        expect(PaymentProviders::CancelPaymentAuthorizationJob).to have_received(:perform_later).with(
```
groyoh, line 299: I believe we could use `to have_enqueued_job()`.

groyoh (approved):

PR #4389, opened 2025-09-24
fix(integrations): Fix create credit note service

ivannovosad:
## Context

There was an issue when creating credit note via integrations aggregator:

```
NoMethodError
undefined method 'raise_if_error!' for an instance of SendWebhookJob (NoMethodError)
```

Instead of returning a result object an instance of `SendWebhookJob` was returned.

## Description

Return result object in case the payload is not correct.

### DIFF
--- a/app/services/integrations/aggregator/credit_notes/create_service.rb
+++ b/app/services/integrations/aggregator/credit_notes/create_service.rb
@@ -56,6 +56,7 @@ module Integrations
           raise e
         rescue Integrations::Aggregator::BasePayload::Failure => e
           deliver_error_webhook(customer:, code: e.code, message: e.code.humanize)
+          result
         end
 
         def call_async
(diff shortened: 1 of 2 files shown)


groyoh (commented):
Should we maybe return or raise the error so that we can debug ?

app/services/integrations/aggregator/credit_notes/create_service.rb
```diff
@@ -56,6 +56,7 @@ def call
           raise e
         rescue Integrations::Aggregator::BasePayload::Failure => e
           deliver_error_webhook(customer:, code: e.code, message: e.code.humanize)
+          result
```
lovrocolic, line 59: maybe `result.record_validation_failure!(record: e.record)`? wdyt?

ivannovosad: Then it's different than in case of creating an invoice: https://github.com/getlago/lago-api/blob/fix-integrations-credit-note/app/services/integrations/aggregator/invoices/create_service.rb#L63

Either we leave it like this or we raise an error in the invoice service as well.

@lovrocolic @groyoh

lovrocolic: Let's keep it as for the invoice. Thanks! 🙌

groyoh (approved):

PR #4391, opened 2025-09-24
feat: added migrations

brunomiguelpinto:
## Roadmap Task

## Context

## Description


app/models/wallet.rb
```diff
@@ -27,6 +27,7 @@ class Wallet < ApplicationRecord
   validates :invoice_requires_successful_payment, exclusion: [nil]
   validates :paid_top_up_min_amount_cents, numericality: {greater_than: 0}, allow_nil: true
   validates :paid_top_up_max_amount_cents, numericality: {greater_than: 0}, allow_nil: true
+  validates :priority, inclusion: {in: 1..50}
```
vincent-pochet: Maybe you could turn this `50` into a constant, what do you think?

vincent-pochet (approved):

app/models/wallet.rb
brunomiguelpinto: sounds good

spec/services/wallets/create_service_spec.rb
```diff
@@ -100,6 +102,25 @@
       end
     end
 
+    context "when priority is not set" do
```
annvelents, line 135: can we add a test where priority is out of the allowed range please?

PR #4392, opened 2025-09-25
feat(customer): Add `metadata[key]` filter to `GET /api/v1/customers`

groyoh:
## Context

The filtering on the `GET /api/v1/customers` endpoint is too limited.

## Description

This adds a `metadata[key]` filter to this endpoint.


ancorcruz (approved):
LGTM!

PR #4396, opened 2025-09-25
chore(stripe): Remove "beta" for Stripe pre-auth feature [2]

julienbourdeau:
Re-apply https://github.com/getlago/lago-api/pull/4387

I believe this is the first time we remove an integration. Unfortunately, this created a bug where data in the database was being serialized in the GraphQL but this was invalid beacuse the value wasn't found in the enum.

The second commit is the fix of the first PR. Basically, whatever is in the DB, we return the intersection of db value and valid integrations.

CleanShot 2025-09-25 at 12 08 48@2x

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -163,6 +163,10 @@ class Organization < ApplicationRecord
     end
   end
 
+  def premium_integrations
+    super & INTEGRATIONS
+  end
+
   def using_lifetime_usage?
     lifetime_usage_enabled? || progressive_billing_enabled?
   end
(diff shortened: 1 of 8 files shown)


jdenquin:
should the fix be on FE side instead of monkey patching it?

app/models/organization.rb
```diff
@@ -160,6 +163,10 @@ class Organization < ApplicationRecord
     end
   end
 
+  def premium_integrations
+    super & INTEGRATIONS
```
groyoh: I don't think `anrok` should be in there:

```suggestion
    super & PREMIUM_INTEGRATIONS
```

groyoh:
> should we not remove it in the data before so? Sorry I'm not sure to get it

You're right but the data needs to be removed **afterward**. If you add a migration to remove it, the migration will come **before** the code is updated and the running API instance will start to fail until the deployment is complete. If deploy goes well it might not be a big deal. But if something goes wrong during deployment and only the migration was successful, you'll potentially end up with plenty of failed requests.

app/models/organization.rb
julienbourdeau: yes!

groyoh (approved):

PR #4401, opened 2025-09-25
feat: Apply multiple wallet

brunomiguelpinto:
## Context

We’re introducing the priority field to wallets to support custom execution or display ordering. This PR builds on the base work from [lago-api#4391](https://github.com/getlago/lago-api/pull/4391), which added the field to the Wallet model and exposed it via the API.

The goal of this PR is to implement the logic that uses the priority value when handling wallets — for example, when determining which wallets to apply or process first.

## Description

This PR includes:
	•	Logic to order wallets by priority (and created_at as a secondary key).
	•	Ensures active_wallets_in_application_order returns wallets in the correct priority sequence.
	•	Updates any internal wallet processing to respect the defined order.

This builds directly on the foundation laid in PR #4391 and assumes the priority field and its validations are already in place.

### DIFF
--- a/app/services/wallets/balance/decrease_service.rb
+++ b/app/services/wallets/balance/decrease_service.rb
@@ -3,9 +3,10 @@
 module Wallets
   module Balance
     class DecreaseService < BaseService
-      def initialize(wallet:, wallet_transaction:)
+      def initialize(wallet:, wallet_transaction:, skip_refresh: true)
         @wallet = wallet
         @wallet_transaction = wallet_transaction
+        @skip_refresh = skip_refresh
 
         super
       end
@@ -23,7 +24,12 @@ module Wallets
           last_consumed_credit_at: Time.current
         )
 
-        Wallets::Balance::RefreshOngoingService.call(wallet:, include_generating_invoices: true)
+        unless skip_refresh
+          Wallets::Balance::RefreshOngoingService.call(
+            wallet:,
+            include_generating_invoices: true
+          )
+        end
 
         after_commit { SendWebhookJob.perform_later("wallet.updated", wallet) }
 
@@ -33,7 +39,7 @@ module Wallets
 
       private
 
-      attr_reader :wallet, :wallet_transaction
+      attr_reader :wallet, :wallet_transaction, :skip_refresh
     end
   end
 end


app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+  class AppliedPrepaidCreditsService < BaseService
+    MAX_WALLET_DECREASE_ATTEMPTS = 5
+
+    def initialize(invoice:, wallets:)
+      @invoice = invoice
+      @wallets = wallets
+
+      super(nil)
+    end
+
```
lovrocolic, line 13: Should we add `activity_loggable` in the same way we had it before?

lovrocolic: I found it below:)

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+    def call
+      result.prepaid_credit_amount_cents ||= 0
+      result.wallet_transactions ||= []
+
```
annvelents: shall we have an early return if invoice is not found?

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+      result
+    end
+
+    private
+
+    attr_accessor :invoice, :wallets
+
+    def wallets_already_applied?
+      return false unless invoice
+      wallet_ids = wallets.map { |w| w.id }
```
annvelents: `wallet_ids = wallets.pluck(:id)`
(check that it's an AR, not an Array)

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+      applied
+    end
+
+    def eligible_fee_keys_for_wallet(wallet)
+      all_fees = invoice.fees
+
+      chosen = if wallet.limited_to_billable_metrics?
+        fees_for_wallet_limited_to_bm(wallet, all_fees)
+      elsif wallet.limited_fee_types?
+        all_fees.select { |f| wallet.allowed_fee_types.include?(f.fee_type) }
```
annvelents: is it possible to have this instead?`all_fees.where(type: wallet.allowed_fee_types)`

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+      else
+        all_fees
+      end
+
+      chosen.map { |f| [f.fee_type, f.charge&.billable_metric_id] }.uniq
+    end
+
+    def fees_for_wallet_limited_to_bm(wallet, all_fees)
+      limited_to_bm_ids = wallet.wallet_targets.pluck(:billable_metric_id)
+      fees_limited_by_bm = all_fees.select do |f|
```
annvelents: most likely we also can do where to not initialize an array and use PG search

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+        f.charge&.billable_metric_id && limited_to_bm_ids.include?(f.charge.billable_metric_id)
+      end
+      # a wallet can be limited only to BM or to BM AND some fee_types
+      return fees_limited_by_bm unless wallet.limited_fee_types?
+
+      remaining_fees = all_fees - fees_limited_by_bm
```
annvelents:
```
fees_limited_by_bm_ids = fees_limited_by_bm.ids
remaining_fees = all_fees.where.not(id: fees_limited_by_bm_ids)
```

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+      # when wallet is limited to BM, it can't be applied to charge fees
+      remaining_fees.reject! { |f| f.fee_type == "charge" }
```
annvelents: if we're using ActiveRecord::Relation, here we also can do `where.not(fee_type: "charge")`

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+
+      fees_limited_by_type = remaining_fees.select { |f| wallet.allowed_fee_types.include?(f.fee_type) }
```
annvelents: fees_limited_by_type = remaining_fees.where(fee_type: wallet.allowed_fee_types)

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+      result.wallet_transactions ||= []
+
+      if wallets_already_applied?
+        return result.service_failure!(code: "already_applied", message: "Prepaid credits already applied")
+      end
+
+      # build a hash of remaining_amounts for each (fee_type and billable_metric_id) pair
+      remaining_amounts = calculate_amounts_for_fees_by_type_and_bm
+
+      ApplicationRecord.transaction do
```
annvelents: shall we have an explicit way to lock the wallets?

annvelents (commented):
I'd suggest perform a test, using ruby-client:
1) create a customer
2) create a plan with pay in advance charge
3) create multiple wallets with different priorities and targets for the customer
4) using ruby-client, send 1000 events for this charge  for this customer

app/services/wallets/balance/refresh_ongoing_service.rb
```diff
@@ -93,18 +93,62 @@ def credits_ongoing_balance
...
+        return fee_wallet if fees.blank?
+
+        wallets = customer.active_wallets_in_application_order.includes(:wallet_targets)
+
+        wallets_data = wallets.map do |w|
+          {
+            id: w.id,
+            bm_limited: w.limited_to_billable_metrics?,
+            type_limited: w.limited_fee_types?,
+            bm_ids: (w.wallet_targets.map(&:billable_metric_id).to_set if w.limited_to_billable_metrics?),
```
annvelents: use :uniq instead of :to_set
use :pluck instead of :map
remove _limited fields

app/services/wallets/balance/refresh_ongoing_service.rb
```diff
@@ -93,18 +93,62 @@ def credits_ongoing_balance
...
+        fee_wallet
+      end
+
+      def calculate_total_usage_with_limitation(usage_amount_cents)
         all_fees = usage_amount_cents.flat_map { |usage| usage[:invoice].fees }
-        charge_ids = Charge.where(id: all_fees.map(&:charge_id)).where(billable_metric_id: wallet.wallet_targets.pluck(:billable_metric_id)).pluck(:id)
+        return 0 if all_fees.empty?
 
-        return usage_amount_cents.sum { |e| e[:total_usage_amount_cents] } if charge_ids.empty?
+        owners = assign_wallet_per_fee(all_fees) # { fee_key => wallet_id }
```
annvelents: owners => walletes_applicable_on_fees

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -0,0 +1,151 @@
...
+        status: :settled,
+        settled_at: Time.current,
+        transaction_status: :invoiced
+      ).wallet_transaction
+
+      # Decrease balance with retry (optimistic locking)
+      decrease_attempt = 0
+      begin
+        decrease_attempt += 1
+        Wallets::Balance::DecreaseService.new(wallet:, wallet_transaction: wallet_transaction).call
```
annvelents, line 128: can we while creating the set of transactions only decrease the balance, but to not refresh the wallets?
and do one final refresh of all wallets at the end?
do we need it?

app/services/wallets/balance/refresh_ongoing_service.rb
```diff
@@ -93,18 +93,62 @@ def credits_ongoing_balance
         ongoing_balance_cents.to_f.fdiv(currency.subunit_to_unit).fdiv(wallet.rate_amount)
       end
 
-      def calculate_total_usage_with_limitation(usage_amount_cents)
-        return usage_amount_cents.sum { |e| e[:total_usage_amount_cents] } unless wallet.limited_to_billable_metrics?
+      # Assigns each fee to a wallet based on the wallet's limitations
+      # the ongoing balance should only affect the wallet
+      # with the highest priority with the corresponding limitations ( can go negative )
+      def assign_wallet_per_fee(fees)
+        fee_wallet = {}
```
brunomiguelpinto, line 97: if only one wallet this should not  iterate all the fees

app/services/invoices/calculate_fees_service.rb
```diff
@@ -310,10 +310,9 @@ def should_create_coupon_credit?
 
     def should_create_applied_prepaid_credit?
       return false if not_in_finalizing_process?
-      return false unless wallet&.active?
```
lovrocolic, line 313: Should we keep this condition? wdyt?

app/services/invoices/calculate_fees_service.rb
```diff
@@ -324,9 +323,7 @@ def create_credit_note_credit
     end
 
     def create_applied_prepaid_credit
-      prepaid_credit_result = Credits::AppliedPrepaidCreditService.call(invoice:, wallet:)
-      prepaid_credit_result.raise_if_error!
```
lovrocolic, line 328: I am not sure why don't we return failed result anymore?

floganz: Miguel changed `call` to `call!` which raises if result is not success.

PR #4411, opened 2025-09-29
fix(credit_notes): Handle rounding issues with taxes

julienbourdeau:


spec/scenarios/credit_notes/rounding_issues_spec.rb
```diff
@@ -19,40 +19,93 @@
...
-    end
+  context "when the thing is greater" do
+    let(:amount_cents) { 20000 }
 
-    subscription = customer.subscriptions.last
-    invoice = customer.invoices.last
-    expect(invoice.fees_amount_cents).to eq(20000)
-    expect(invoice.taxes_amount_cents).to eq(5000)
-    expect(invoice.total_amount_cents).to eq(25000)
+    it "handles the rounding issues" do
```
jdenquin, line 25: can we rename this file and test as `rounding` instead of `rounding_issue` ?

PR #4412, opened 2025-09-29
feat(plan-cascade): add possibility to sync all not existing charges from parent plan to child plan

annvelents:
## Context

When updating plan and not sending charges with IDs, the plan's charges will be recreated. If while doing this, cascade_updates was false, we won't affect children plans. As result, charges of children plan will have a link to deleted parent_id, and there is no way to update / delete them. 

## Description

Added a service that gets all charges of parent plan and one by one, updates children plans with them.
firstly it checks if child plan already has a charge with parent_id equal current -> no action needed
then it checks if child plan has a charge, that has the same billable_metric_id and activity_log, and it's parent is deleted -> in this case we assume that the "recreated parent" case happened and restore link to the parent
finally, if there are no corresponding charges at children plan, we'll create a child charge for the children plan

### DIFF
--- a/app/services/plans/sync_new_charges_with_children_service.rb
+++ b/app/services/plans/sync_new_charges_with_children_service.rb
@@ -9,7 +9,7 @@ module Plans
     end
 
     def call
-      # Note: this check is only needed if child charges "lost" their parents: 
+      # Note: this check is only needed if child charges "lost" their parents:
       # when a parent plan was updated without sending charges_ids, so the "original" parents were deleted
       return result.forbidden_failure!(code: "plan_has_undistinguishable_charges") if plan_has_undistinguishable_charges?
 
@@ -32,8 +32,8 @@ module Plans
     end
 
     def plan_has_undistinguishable_charges?
-       # we can also add a check by filters as well
-      all_charges_summary = plan.charges.map {|ch| {metric: ch.billable_metric_id, charge_model: ch.charge_model}}
+      # we can also add a check by filters as well
+      all_charges_summary = plan.charges.map { |ch| {metric: ch.billable_metric_id, charge_model: ch.charge_model} }
       all_charges_summary.uniq.count != all_charges_summary.count
     end
   end
--- a/spec/scenarios/plans/sync_new_charges_with_children_spec.rb
+++ b/spec/scenarios/plans/sync_new_charges_with_children_spec.rb
@@ -13,14 +13,14 @@ describe "Sync New Charges With Children Scenario" do
     20.times do |i|
       aggregation_types = %w[count_agg sum_agg latest_agg max_agg unique_count_agg]
       aggregation_type = aggregation_types[i % aggregation_types.length]
-      
+
       billable_metrics << create(
         :billable_metric,
         organization:,
         name: "Metric #{i + 1}",
         code: "metric_#{i + 1}",
         aggregation_type:,
-        field_name: aggregation_type == "count_agg" ? nil : "field_#{i + 1}"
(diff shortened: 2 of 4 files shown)


app/services/plans/sync_new_charges_with_children_service.rb
```diff
@@ -0,0 +1,40 @@
...
+    def sync_charge_for_children(charge)
+      plan.children.joins(:subscriptions).where(subscriptions: {status: %w[active pending]}).distinct.pluck(:id).each_slice(20) do |child_ids|
+        Charges::SyncChildrenBatchJob.perform_later(
+          children_plans_ids: child_ids,
+          charge:
+        )
+      end
+    end
+
+    def plan_has_undistinguishable_charges?
```
lovrocolic, line 34: So if we cannot match exactly children charges, since there are multiple ones with the same BM, the feature is not applicable? Is this correct?
That was the reason why we need to establish association between parent and child charges...

annvelents: yes. Because the goal is to find possible parents by billable metric and charge_model, to not have duplicates
what happened in the situation this service solves:
- org created plan
- they created children with overrides
- they were sending plan updates, without populating charges ids and not sending the updates
- in the parent plan charges were deleted and recreated
- children charges lost their parents, while the "spirit" of them is still in the plan

if we cannot distinguish this relations, we'll need to have a manual intervention, so we throw an error

lovrocolic: Just to understand, the main issue is in the third bullet: children exist -> parent is updated without charge ids and without cascading -> children lost relationship to parent?

lovrocolic: Yep it seems reasonable that such usage could lead to these issues:/

lovrocolic (approved):

app/services/plans/sync_new_charges_with_children_service.rb
annvelents: yes, they will work on not sending us updates without ids, but for the billing run they want to have all charges in children plans

app/services/plans/sync_new_charges_with_children_service.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Plans
+  class SyncNewChargesWithChildrenService < BaseService
```
julienbourdeau, line 4: This service is triggered from the console, as a maintenance on-demand operation, right?

app/services/plans/sync_new_charges_with_children_service.rb
```diff
@@ -0,0 +1,40 @@
...
+    private
+
+    attr_reader :plan
+
+    def sync_charge_for_children(charge)
+      plan.children.joins(:subscriptions).where(subscriptions: {status: %w[active pending]}).distinct.pluck(:id).each_slice(20) do |child_ids|
+        Charges::SyncChildrenBatchJob.perform_later(
+          children_plans_ids: child_ids,
+          charge:
+        )
```
julienbourdeau, line 30: I like that the work is done in small batches but I'm not sure what work is "grouped". Is it to avoid creating to many jobs?
I feel like this could be one job per child_plan_id.

app/services/plans/sync_new_charges_with_children_service.rb
julienbourdeau: > they were sending plan updates, without populating charges ids and not sending the updates
Am I correct to think that this is done _"by mistake"? _ even if technically there is nothing wrong with this call, we'd expect them to cascade in this case?

PR #4414, opened 2025-09-29
chore(integration): Remove unused code related to integration mapping update

groyoh:
## Context

When updating a integration mapping, it doesn't really make sense to update the `mappable_id`, `mappable_type` and `integration_id` as this mapping is intrinsically bound to the `integration` and `mappable`. For instance, one would never "move" a mapping from one integration to the other.

Also when updating an integration mapping, [the frontend is currently resending the same data as it received](https://github.com/getlago/lago-front/blob/2c4967927a0c7bb34af7c48784f1f32fd457113d/src/components/settings/integrations/NetsuiteIntegrationMapItemDialog.tsx#L304-L306). 

Therefore the `integration_id`, `mappable_id`, `mappable_type` arguments of the `update_integration_mapping` GraphQL query are currently unnecessary. 

## Description

This PR removes the code related these field update for the sake of simplifying this integration mapping code. We still needed to keep the GraphQL fields in the schema (although unused) to avoid a breaking change.

### DIFF
--- a/spec/graphql/mutations/integration_mappings/update_spec.rb
+++ b/spec/graphql/mutations/integration_mappings/update_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe Mutations::IntegrationMappings::Update do
   let(:required_permission) { "organization:integrations:update" }
   let(:integration_mapping) { create(:netsuite_mapping, integration:) }
   let(:integration) { create(:netsuite_integration, organization:) }
-  let(:mappable) { create(:add_on, organization:) }
+  let(:mappable) { integration_mapping.mappable }
   let(:organization) { membership.organization }
   let(:membership) { create(:membership) }
   let(:external_account_code) { Faker::Barcode.ean }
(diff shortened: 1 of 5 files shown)


spec/graphql/mutations/integration_mappings/update_spec.rb
```diff
@@ -42,9 +42,6 @@
       variables: {
         input: {
           id: integration_mapping.id,
-          integrationId: integration.id,
-          mappableId: mappable.id,
-          mappableType: "AddOn",
```
ivannovosad, line 47: We can remove these also from [here](https://github.com/getlago/lago-api/pull/4414/files#diff-7b88130d44e881dba4eae5574a210c582ff69d65ba9636e12bf64ddc3c35613cL21-L23)

groyoh: I think it make sense to keep it in response to make sure we render it properly.

ivannovosad (approved):

groyoh:
@ivannovosad I had to update the PR to keep the field in the GraphQL schema. Otherwise it breaks the GraphQL API since the frontend still sends the data.

PR #4423, opened 2025-09-30
feat(multiple-payment-methods): add table and model for payment methods

lovrocolic:
## Context

Currently in Lago, there can be only one payment provider customer and also only one payment method.

## Description

With this feature, it will be possible to add multiple provider customers and also attach many payment methods to it.

This is the first PR for this feature, that adds `payment_methods` table and model


db/migrate/20250930111323_create_payment_methods.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class CreatePaymentMethods < ActiveRecord::Migration[8.0]
+  def change
+    create_table :payment_methods, id: :uuid do |t|
+      t.references :organization, type: :uuid, null: false, foreign_key: true
+      t.references :customer, type: :uuid, null: false, foreign_key: true
+      t.references :payment_provider_customer, type: :uuid, null: false, foreign_key: true
+
+      t.string :provider_method_id
+      t.string :method_type, index: true
```
ivannovosad: @lovrocolic 
When looking at it now, I'm thinking if not to unify the naming here:

`provider_method_id, provider_method_type`
or
`method_id, method_type`

lovrocolic:
Closing because of - https://github.com/getlago/lago-api/pull/4455

PR #4427, opened 2025-10-01
test: Fix failing `spec/services/daily_usages/compute_service_spec.rb` test

groyoh:
## Context

As the subscription `started_at/subscription_at` were set to 1 year ago relative to the current date, the charges `from_datetime` would be set the current time instead of beginning of the month.

## Description

This fix the spec.


ancorcruz (approved):

groyoh:
Thanks folks 👍

PR #4429, opened 2025-10-02
fix: add guard clause when applying prog billing credits on fee

lovrocolic:
## Context

We need to ensure that fee amount_cents are never less than credits applied on that fee

## Description

This PR fixes described case


ancorcruz (approved):
LGTM!

PR #4430, opened 2025-10-02
fix(billing): Backfill CustomerSnapshot for all finished invoices

ancorcruz:
finished and voided invoices...


db/migrate/20250930113902_backfill_invoice_customer_snapshots.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class BackfillInvoiceCustomerSnapshots < ActiveRecord::Migration[8.0]
+  def change
+    Invoice.where(status: ["finalized", "voided"])
```
vincent-pochet, line 5: Any idea of how many invoices we are talking about?

I'm just wondering if we should use a "Migration task" approach here as this might take a while to perform.

PR #4433, opened 2025-10-03
fix(invoice-pdf): Display prepaid credits for progressive billing invoices

floganz:
## Description

We enter this template only with invoices having type `progressive_billing`. Progressive invoice can never be subscription type leading to never showing prepaid credits line in the pdf.


groyoh:
@floganz now that there's an easy way to test invoice rendering, I would suggest adding a unit or scenario test for this.

PR #4434, opened 2025-10-03
fix(usage): Use Parallel for better performance in GQL context

floganz:
## Description

For unclear reasons computing all usage in 1 thread (without Parallel) as part of GQL request triggers "Stack too deep" exception on cold cache.
Reverting to version where we use Parallel to decrease impact on customers.

### DIFF
--- a/app/services/invoices/customer_usage_service.rb
+++ b/app/services/invoices/customer_usage_service.rb
@@ -93,13 +93,16 @@ module Invoices
         .includes(:taxes, billable_metric: :organization, filters: {values: :billable_metric_filter})
         .order(Arel.sql("lower(unaccent(billable_metrics.name)) ASC"))
 
+      received_event_codes = distinct_event_codes(subscription, boundaries)
+
       # we're capturing the context here so we can re-use inside the threads. This will correctly propagate spans to this current span
       context = OpenTelemetry::Context.current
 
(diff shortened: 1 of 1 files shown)


app/services/invoices/customer_usage_service.rb
```diff
@@ -89,18 +89,22 @@ def compute_usage
...
-        .includes(:taxes, billable_metric: :organization, filters: {values: :billable_metric_filter})
-        .find_each do |charge|
-        bypass_aggregation = !received_event_codes.include?(charge.billable_metric.code)
-        fees += charge_usage(charge, bypass_aggregation)
+      # we're capturing the context here so we can re-use inside the threads. This will correctly propagate spans to this current span
+      context = OpenTelemetry::Context.current
+
+      fees = Parallel.flat_map(query.all, in_threads: ENV["LAGO_PARALLEL_THREADS_COUNT"]&.to_i || 0) do |charge|
+        OpenTelemetry::Context.with_current(context) do
+          ActiveRecord::Base.connection_pool.with_connection do
```
julienbourdeau, line 103: Is this to make sure you keep the same connection in all subsequent calls and no other thread touches this connection?

floganz: To be honest, I'm reverting to how it was before recent changes. Since we spawn new threads I think it's safer to explicitly establish AR connection.

You mean something else?

jdenquin: its to ensure we use the same connection pool because of the threading yes!

PR #4437, opened 2025-10-03
misc(clickhouse): Add specs for deduplication

vincent-pochet:
## Description
This PR adds 2 specs to validate the event deduplication logic in order to avoid issues with clickhouse queries

- the first case validates that events with same transaction_id are kept
- the second one validates that events sharing the same transaction_id and timestamp are merged


groyoh (approved):
You mentioned that the events are "merged". What does that mean exactly ? Does the second event only gets dropped ? Or does something else happen during this "merge" ?

Because right now, the test only ensure that the 6th event does not exist.

PR #4442, opened 2025-10-06
[WIP] fix(event): Fix `Events::Stores::ClickhouseStore` store

groyoh:
## Description

This is a WIP PR.


vincent-pochet (approved):
Amazing one! 👍 
Thank you for diging into it.

It bring one new topic I guess: the clickhouse specs were aligned with the postgres ones, in order to be able to compare the results between the two implementation.
I guess we should either: update the postgres specs as well, or add a new layer of specs making a clear comparison of the results of both clickhouse and postgres. Maybe it could be some new scenarios

groyoh:
> It bring one new topic I guess: the clickhouse specs were aligned with the postgres ones, in order to be able to compare the results between the two implementation.
I guess we should either: update the postgres specs as well, or add a new layer of specs making a clear comparison of the results of both clickhouse and postgres. Maybe it could be some new scenarios

That's a very valid point that I didn't take into consideration. I'll try and figure something out.

julienbourdeau (approved):

PR #4444, opened 2025-10-07
chore(integration): Remove unused integration collection mapping update code

groyoh:
## Context

When updating a integration collection mapping, it doesn't really make sense to update the `mapping_type ` and `integration_id` as this collection mapping is intrinsically bound to the `integration` and `mapping_type`. For instance, one would never "move" a collection mapping from one integration to the other.

Also when updating an integration collection mapping, [the frontend is currently resending the same data as it received](https://github.com/getlago/lago-front/blob/2c4967927a0c7bb34af7c48784f1f32fd457113d/src/components/settings/integrations/NetsuiteIntegrationMapItemDialog.tsx#L289-L290). 

Therefore the `integration_id` and `mapping_type` arguments of the `update_integration_collection_mapping` GraphQL mutation are currently unnecessary. 

## Description

This removes the code related to these fields update for the sake of simplifying this integration collection mapping code. We still needed to keep the GraphQL fields in the schema (although unused) to avoid a breaking change.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -56687,30 +56687,6 @@
               "deprecationReason": null
             },
             {
-              "name": "integrationId",
-              "description": null,
-              "type": {
-                "kind": "SCALAR",
-                "name": "ID",
-                "ofType": null
-              },
-              "defaultValue": null,
-              "isDeprecated": false,
-              "deprecationReason": null
-            },
-            {
-              "name": "mappingType",
-              "description": null,
-              "type": {
-                "kind": "ENUM",
-                "name": "MappingTypeEnum",
-                "ofType": null
-              },
-              "defaultValue": null,
-              "isDeprecated": false,
-              "deprecationReason": null
-            },
-            {
               "name": "taxCode",
               "description": null,
               "type": {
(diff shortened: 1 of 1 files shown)


app/graphql/types/integration_collection_mappings/update_input.rb
julienbourdeau: Not important but I'd move them down


```diff
-      # @deprecated This field is deprecated and will be ignored. Integration ID cannot be updated.
-      argument :integration_id, ID, required: false
-      # @deprecated This field is deprecated and will be ignored. Mapping type cannot be updated.
-      argument :mapping_type, Types::IntegrationCollectionMappings::MappingTypeEnum, required: false
      argument :tax_code, String, required: false
      argument :tax_nexus, String, required: false
      argument :tax_type, String, required: false


+      # @deprecated This field is deprecated and will be ignored. Integration ID cannot be updated.
+      argument :integration_id, ID, required: false
+      # @deprecated This field is deprecated and will be ignored. Mapping type cannot be updated.
+      argument :mapping_type, Types::IntegrationCollectionMappings::MappingTypeEnum, required: false
```

EDIT: Like in the file below

julienbourdeau (approved):

groyoh:
Thanks for the review @julienbourdeau 🙏

PR #4445, opened 2025-10-07
feat(integration): Add `billing_entity_id` to integration mapping collection

groyoh:
## Context

TBD

## Description

TBD

PR #4446, opened 2025-10-07
feat(integration): Add `billing_entity_id` to integration mapping

groyoh:
## Context

Currently, our customers cannot sync invoices to multiple accounting entities (subsidiaries) as our mapping is made at the organization level. However, in NetSuite, for instance, the tax and item mapping are made at the entity level. We need to allow our mapping to be defined at the entity level.

## Description

This is a follow-up PR to https://github.com/getlago/lago-api/pull/4445.

This PR adds a `billing_entity_id` field to the `integration_mappings` along with the proper indexes and validations.

There will be a follow-up PR which updates the integration payload services to define a lookup and fallback strategy based on the available mappings.


app/models/integration_mappings/base_mapping.rb
```diff
@@ -10,35 +10,53 @@ class BaseMapping < ApplicationRecord
     belongs_to :integration, class_name: "Integrations::BaseIntegration"
     belongs_to :mappable, polymorphic: true
     belongs_to :organization
+    belongs_to :billing_entity, optional: true
 
     MAPPABLE_TYPES = %i[AddOn BillableMetric].freeze
 
+    validates :mappable_type, inclusion: {in: MAPPABLE_TYPES.map(&:to_s)}
```
julienbourdeau, line 17: this map(&:to_s) makes me thing that the MAPPABLE_TYPES should be an array of string, instead of symbol 🤔

julienbourdeau (approved):

app/models/integration_mappings/base_mapping.rb
groyoh: I double checked and everywhere we use symbol to define enums. Here it's an enum in the sense that it can only have two values but don't need the `enum` hence the explicit validation. It's still defined as enum in GraphQL though.

PR #4447, opened 2025-10-07
misc: Avoid using decimal128 in Clickhouse when not needed

rsempe:
## Context

This PR updates how decimal types and scales are handled in ClickHouse queries for `count`, `unique_count` and `weighted_sum` aggregations.

The objective is to switch from `toDecimal128` to lower-precision types like `toDecimal32` or `toDecimal64` where appropriate.


annvelents (approved):
Wow, we had so many of them for just 0s and 1s 🙈

PR #4450, opened 2025-10-07
feat(integration): Update billing entity mappings in integration payload

groyoh:
## Context

Currently, our customers cannot sync invoices to multiple accounting entities (subsidiaries) as our mapping is made at the organization level. However, in NetSuite, for instance, the tax and item mapping are made at the entity level. We need to allow our mapping to be defined at the entity level.

## Description

This is a follow-up PR to https://github.com/getlago/lago-api/pull/4445 and https://github.com/getlago/lago-api/pull/4446.


app/services/integrations/aggregator/base_payload.rb
```diff
@@ -67,13 +60,46 @@ def amount(amount_cents, resource:)
...
 
-      def collection_mapping(type)
-        integration.integration_collection_mappings.where(mapping_type: type)&.first
+      private
+
+      attr_reader :integration, :billing_entity
+
+      def fallback_item(scope)
+        mappings = integration.integration_collection_mappings
+        fallback_items = mappings.filter { |mapping| mapping.mapping_type.to_sym == :fallback_item }
```
ivannovosad, line 69: What about performing this filtering on DB level?

app/services/integrations/aggregator/base_payload.rb
```diff
@@ -67,13 +60,46 @@ def amount(amount_cents, resource:)
...
+        if scope == :billing_entity && billing_entity
+          return fallback_items.find { |mapping| mapping.billing_entity_id == billing_entity.id }
+        end
+
+        fallback_items.find { |mapping| mapping.billing_entity_id.nil? }
```
ivannovosad, line 74: Same as above:

_What about performing this filtering on DB level?_

app/services/integrations/aggregator/base_payload.rb
```diff
@@ -67,13 +60,46 @@ def amount(amount_cents, resource:)
...
       end
 
-      private
+      def lookup_collection_mapping(mapping_type, with_fallback_item: true)
+        mappings = integration.integration_collection_mappings
+        matching_mappings = mappings.filter { |mapping| mapping.mapping_type.to_sym == mapping_type.to_sym }
```
ivannovosad, line 79: Here as well: 

_What about performing this filtering on DB level?_

app/services/integrations/aggregator/base_payload.rb
```diff
@@ -67,13 +60,46 @@ def amount(amount_cents, resource:)
...
+        billing_entity_mapping = matching_mappings.find { |mapping| mapping.billing_entity_id == billing_entity.id }
+        organization_mapping = matching_mappings.find { |mapping| mapping.billing_entity_id.nil? }
```
ivannovosad, line 81: Also here:

_What about performing this filtering on DB level?_

app/services/integrations/aggregator/base_payload.rb
```diff
@@ -67,13 +60,46 @@ def amount(amount_cents, resource:)
...
+        end
 
-      attr_reader :integration
+        billing_entity_mapping ||
+          organization_mapping
+      end
+
+      def lookup_mapping(mappable_type, mappable_id)
+        mappings = integration.integration_mappings
+        matching_mappings = mappings.filter { |mapping| mapping.mappable_type == mappable_type && mapping.mappable_id == mappable_id }
```
ivannovosad, line 95: Also here:

_What about performing this filtering on DB level?_

app/services/integrations/aggregator/base_payload.rb
```diff
@@ -67,13 +60,46 @@ def amount(amount_cents, resource:)
...
+        billing_entity_mapping = matching_mappings.find { |mapping| mapping.billing_entity_id == billing_entity.id }
+        organization_mapping = matching_mappings.find { |mapping| mapping.billing_entity_id.nil? }
```
ivannovosad, line 97: I think this is the last one :)

_What about performing this filtering on DB level?_

app/services/integrations/aggregator/base_payload.rb
groyoh: I don't think it makes sense to be honest. I'll give you a scenario to showcase that:

Let's say there's no item defined on the billing entity and only one fallback item on the org. We'll make 4 DB calls for every single fees/item. Now let's say there's 5 fees on the invoices. We'll make 20 DB calls.

With the current code, we'll load the integration mappings once and that's it.

I guess we could optimize by loading only the integration mapping matching the current billing entity or scoped to the organization and then memoizing this value. Let me know what you think.

ivannovosad: But we don't have to filter in memory (ruby) by pushing the filtering to sql:

```
fallback_items = integration.integration_collection_mappings.where(mapping_type: :fallback_item)
```

groyoh: @ivannovosad but that still means we're doing one DB call per `mapping_type`, right ? So we end up with multiple DB calls to avoid doing the filtering in-memory.

TBH I don't mind but I usually prefer to limit the number of DB whenever possible.

ivannovosad: Lets leave it as it is - and we can implement some memoization later if needed.

PR #4454, opened 2025-10-07
fix: Order events by timestamp and property for unique count

rsempe:
## Context

The goal of this PR is to correctly handle several events with the same timestamp for the unique count aggregation.

In addition to the `timestamp`, we decided to add the ordering on the `property` as well.

### DIFF
--- a/app/services/events/stores/aggregated_clickhouse/unique_count_query.rb
+++ b/app/services/events/stores/aggregated_clickhouse/unique_count_query.rb
@@ -25,7 +25,7 @@ module Events
                   operation_type,
                   #{operation_value_sql(partition_by: %w[property])} AS adjusted_value
                 FROM events_data
-                ORDER BY timestamp ASC
+                ORDER BY timestamp ASC, property ASC
               ) adjusted_event_values
               GROUP BY property
             )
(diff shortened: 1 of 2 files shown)


spec/services/events/stores/clickhouse_store_spec.rb
```diff
@@ -686,6 +686,19 @@
...
       )
 
+      Clickhouse::EventsEnriched.create!(
+        transaction_id: SecureRandom.uuid,
+        organization_id: organization.id,
+        external_subscription_id: subscription.external_id,
+        code:,
+        timestamp: boundaries[:from_datetime] + 1.day,
+        properties: {
+          billable_metric.field_name => 3
```
julienbourdeau, line 696: If I understand correctly what we're trying to test, should we create an event with units lower than the previous events.
In this case, I think `, property ASC` is the same as nothing because it's the order it's been inserted.

I don't have all the context so I might be completely off

rsempe: For unique count there is no such thing as `units`, properties are used to count distinct values. What is important here is the consistent ordering. The issue was that clickhouse is not ensuring ordering.

PR #4455, opened 2025-10-07
feat(multiple-payment-methods): add payment method model and table

lovrocolic:
## Context

Currently in Lago, there can be only one payment provider customer and also only one payment method.

## Description

With this feature, it will be possible to add multiple provider customers and also attach many payment methods to it.

This is the first PR for this feature, that adds `payment_methods` table and model

### DIFF
--- a/app/models/payment_method.rb
+++ b/app/models/payment_method.rb
@@ -8,8 +8,9 @@ class PaymentMethod < ApplicationRecord
   default_scope -> { kept }
 
   belongs_to :organization
-  belongs_to :customer
-  belongs_to :payment_provider_customer, class_name: "PaymentProviderCustomers::BaseCustomer"
+  belongs_to :customer, -> { with_discarded }
+  belongs_to :payment_provider, optional: true, class_name: "PaymentProviders::BaseProvider"
+  belongs_to :payment_provider_customer, optional: true, class_name: "PaymentProviderCustomers::BaseCustomer"
 
   validates :provider_method_id, presence: true
   validates :is_default, inclusion: {in: [true, false]}
@@ -28,7 +29,8 @@ end
 #  updated_at                   :datetime         not null
 #  customer_id                  :uuid             not null
 #  organization_id              :uuid             not null
-#  payment_provider_customer_id :uuid             not null
+#  payment_provider_customer_id :uuid
+#  payment_provider_id          :uuid
 #  provider_method_id           :string           not null
 #
 # Indexes
@@ -36,6 +38,7 @@ end
 #  index_payment_methods_on_customer_id                   (customer_id)
 #  index_payment_methods_on_organization_id               (organization_id)
 #  index_payment_methods_on_payment_provider_customer_id  (payment_provider_customer_id)
+#  index_payment_methods_on_payment_provider_id           (payment_provider_id)
 #  index_payment_methods_on_provider_method_type          (provider_method_type)
 #  unique_default_payment_method_per_customer             (customer_id) UNIQUE WHERE ((is_default = true) AND (deleted_at IS NULL))
 #
@@ -44,4 +47,5 @@ end
 #  fk_rails_...  (customer_id => customers.id)
 #  fk_rails_...  (organization_id => organizations.id)
 #  fk_rails_...  (payment_provider_customer_id => payment_provider_customers.id)
+#  fk_rails_...  (payment_provider_id => payment_providers.id)
 #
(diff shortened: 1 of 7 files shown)


db/migrate/20251007160309_create_payment_methods.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      t.string :provider_method_id, null: false
+      t.string :provider_method_type, null: true, index: true
+      t.boolean :is_default, default: false, null: false
+      t.datetime :deleted_at
+      t.jsonb :details, null: false, default: {}
+
+      t.timestamps
+
+      t.index [:customer_id], where: "is_default = TRUE", name: "unique_default_payment_method_per_customer", unique: true
```
julienbourdeau: I think we're missing the `AND deleted_at IS NULL` in the where condition. 

Even if there will probably be something to ensure that you don't delete the default payment method, I'm sure it could happen in an `update_all` for some reason.

app/models/payment_method.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class PaymentMethod < ApplicationRecord
+  include PaperTrailTraceable
+  include Discard::Model
+  self.discard_column = :deleted_at
+
+  belongs_to :organization
+  belongs_to :customer
+  belongs_to :payment_provider_customer, class_name: "PaymentProviderCustomers::BaseCustomer"
```
julienbourdeau: I'm still convinced this model (and similar) should be attached to `payment_provider` instead of `payment_provider_customer`. Even if we can retrieve the payment provider through the customer, I feel like it would be better to retrieve the payment_provider_customer through the payment_provider 🤔

Maybe we need more opinions on this, I could definitely be wrong.

Basically, I prefer A over B

Image

---

Image

db/migrate/20251007160309_create_payment_methods.rb
julienbourdeau: I realize I would have written this index this way (multiple columns instead of condition). I don't know if one is better than the other, just sharing.

```rb
t.index [:customer_id, :is_default], where: "deleted_at IS NULL", unique: true , name: "..."
```

spec/models/payment_method_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe PaymentMethod do
```
julienbourdeau, line 5: There is a shared example to test soft deletion now

```rb
it { expect(described_class).to be_soft_deletable }
```

app/models/payment_method.rb
lovrocolic: Since `payment_method` has direct association to the `customer` and `customer` can have only one payment provider per connection, both approaches could work and from payment method we can fetch all needed data in both approaches (each customer could have adyen1, stripe1, stripe2 for example).

From `belongs_to` side, suggestion A makes sense.

However, mostly we will need to find payment method based on provider customer. Currently this is how we get payment method id when creating payment intent: `stripe_pm = provider_customer.payment_method_id`. Also, we would need to list payment methods per customer and per each connection. That's why we chose B approach, mostly because of the `has_many` side. With approach B we can find default payment method by `provider_customer.payment_methods.default` or we can list all payment methods with `provider_customer.payment_methods`. Based how we see it, payment methods naturally belong to each customer in external provider. `payment_provider.payment_methods` would return payment methods across all customers and we believe that payment methods only make sense in the context of customer in external provider.

Does it makes sense? 🤔

julienbourdeau: Yeah, I get what you mean. I think we're debating personal preference so this is not an issue to approve and merge for me.

I believe the `provider_customer` is "not a real model" but almost a cache of something we don't control. I feel like this being the link to the payment method is giving too much responsibility to this model.

I understand that looking at the code if feels natural compared to how it exists today but `provider.payment_methods.where(customer:)` is the same amount of queries and like the data modeling better.

Again, not an issue to approve if you prefer this way.

spec/models/payment_method_spec.rb
lovrocolic: added 👍

db/migrate/20251007160309_create_payment_methods.rb
lovrocolic: The real flow is that if default payment method is detached (marked as inactive), we wouldn't have any other marked as default. So I guess initial index also works but changed it to your first suggestion. Second suggestion doesn't allow multiple payment methods that are not default per customer (but we need this case).

julienbourdeau: > Second suggestion doesn't allow multiple payment methods that are not default per customer 

Absolutely! 🤦‍♂️

PR #4461, opened 2025-10-09
fix(charges): Fix validation of invoiceable for in arrears charges

ivannovosad:
## Context

Currently it is possible to create pay in advance = `false` and invoiceable = `false` charge from the UI

## Description

Add validation for presence of invoiceable in case of in-arrears charges.
This PR also contains a data migration that fixes the invalid data.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -40,9 +40,9 @@ en:
               permissions:
                 forbidden_key: 'contains forbidden keys: %{keys}'
         must_be_greater_than_or_equal_min: must_be_greater_than_or_equal_min
+        must_be_true_unless_pay_in_advance: must_be_true_unless_pay_in_advance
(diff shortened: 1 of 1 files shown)


app/models/charge.rb
```diff
@@ -45,6 +45,7 @@ class Charge < ApplicationRecord
   validate :validate_dynamic, if: -> { dynamic? }
   validates :min_amount_cents, numericality: {greater_than_or_equal_to: 0}, allow_nil: true
   validates :charge_model, presence: true
+  validates :invoiceable, presence: true, unless: -> { pay_in_advance? }
```
julienbourdeau: I find this validation not very clear. The presence with a boolean is a little confusing

brunomiguelpinto: true i wil change it

julienbourdeau (approved):

PR #4462, opened 2025-10-09
feat: Introduce middleware pattern to base service

vincent-pochet:
## Context

This PR is part of the live-aggregation epic.

Now that the aggregation pipeline is in place, we need a new system to refresh the data when a plan, a charge or a filter gets updated.

This has to be done by monitoring the `Plans::CreateService` and `Plans::UpdateService`. Since we need to compare results before and after update, it has to be done "around" the existing `call` method and as such requires a logic to be put in place.

As the approach is similar to the ActivityLog monitoring that is already defined in the `BaseService`, the idea is to avoid "polluting" this service with new logic. Here comes the middleware pattern!

## Description

This PR introduces a new `BaseMiddleware` class that can be used to create "generic" logic that can be plugged in any class inheriting from the `BaseService`.

For example, let's imagine that you want to add a logic to measure the execution time of a service:
```ruby
class DurationMiddleware < BaseMiddleWare
  def before_call
     @start_time = Time.current
  end

  def after_call
     duration = Time.current - @start_time
     Rails.logger.info("#{kwargs[:label]} - #{service_instance.class} execution took #{duration} seconds")
  end
end
```

Alternatively you could also re-define the `call` method:
```ruby
class DurationMiddleware < BaseMiddleWare
  def call
     start_time = Time.current
    result = super
    
     duration = Time.current - start_time
     Rails.logger.info("#{service_instance.class} execution took #{duration} seconds")

    result
  end
end
```

To use in a service just use the `use` instruction:
```ruby
class MyDummyService < BaseService
  Result = BaseResult[:some_value]

  use DurationMiddleware, label: "[Test]" # example with arguments passed to the middleware

  def call
    result
  end
end
```

This new pattern has been used to refactor a bit the code base.
- The open telemetry tracer that was done in `BaseService#call` has been extracted in the `Utils::LogTracerMiddleware` included by default in the `BaseService`
- The ActivityLog monitoring has been moved into the `UtilsActivityLogMiddleware` and is included when calling `activity_loggable`

As a next step, a new `Plans::NotifyRefreshMiddleware` will be added to track updates that need to be propagated in the pre-aggregated records

A new RSpec matcher has also been added to ease the test of middleware inclusion:
```ruby
it { is_expected.to use_middleware(Utils::LogTracerMiddleware) }
# OR
it { expect(service_class).to use_middleware(Utils::LogTracerMiddleware) }
```

### DIFF
--- a/app/services/base_service.rb
+++ b/app/services/base_service.rb
@@ -155,8 +155,22 @@ class BaseService
   end
 
   # Register a new middleware
-  def self.use(middleware_class, *args, **kwargs)
-    self.middlewares += [[middleware_class, args, kwargs]]
+  def self.use(middleware_class, *args, on_conflict: :raise, **kwargs)
+    existing_middleware = middlewares.map(&:first)
+
+    if !existing_middleware.include?(middleware_class) || on_conflict == :append
+      return self.middlewares += [[middleware_class, args, kwargs]]
+    end
+
+    # Middleware already exists
+    case on_conflict
+    when :raise
+      raise Middlewares::AlreadyAddedError.new(middleware_class, self)
+    when :replace
+      self.middlewares[existing_middleware.index(middleware_class)] = [middleware_class, args, kwargs]
+    when :ignore
+      # Do nothing
+    end
   end
 
   use(Middlewares::LogTracerMiddleware)
--- a/app/services/middlewares/activity_log_middleware.rb
+++ b/app/services/middlewares/activity_log_middleware.rb
@@ -2,21 +2,21 @@
 
 module Middlewares
   class ActivityLogMiddleware < BaseMiddleware
-    def call
+    def call(&block)
       if produce_activity_log?
         log_kwargs = {after_commit:}.compact
 
         case action
         when /updated/
-          Utils::ActivityLog.produce(record, action, **log_kwargs) { super }
+          Utils::ActivityLog.produce(record, action, **log_kwargs) { call_next(&block) }
 
         else
-          super.tap do |result|
+          call_next(&block).tap do |result|
             Utils::ActivityLog.produce(record, action, **log_kwargs) { result }
           end
         end
       else
-        super
+        call_next(&block)
       end
     end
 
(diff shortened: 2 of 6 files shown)


app/services/middlewares/log_tracer_middleware.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Utils
+  class LogTracerMiddleware < BaseMiddleware
+    def call
+      LagoTracer.in_span("#{service_instance.class.name}#call") do
+        super
```
groyoh: I feel like it would be less confusing to call the block explicitely rather than calling `super`:

```suggestion
    def call(&block)
      LagoTracer.in_span("#{service_instance.class.name}#call") do
        block.call
```

groyoh: The same goes for activity log IMO.

app/services/base_service.rb
```diff
@@ -150,19 +151,18 @@ class LegacyResult < OpenStruct
   Result = LegacyResult
 
   def self.activity_loggable(action:, record:, condition: -> { true }, after_commit: true)
-    self.activity_log_config = {action:, record:, condition:, after_commit:}
+    use(Utils::ActivityLogMiddleware, action:, record:, condition:, after_commit:)
```
groyoh: I don't know if it's the case but it might happen that a service and its parent both call `activity_loggable` in which case we'll end with two activity logs.

app/services/middlewares/activity_log_middleware.rb
groyoh, line 1: Could be in `middlewares` folder rather than `utils` IMO.

app/services/base_service.rb
vincent-pochet: Yep indeed, that's the case already. It can make sense anyway as it would track different things

groyoh: I fee like this could be dangerous as we're changing the behaviour we have today that only one log can be produced per service. Maybe `use` should raise an error when adding the middleware twice and have a behaviour on conflict:

```
def use(middleware, *args, on_conflict: :raise **opts)
  if middlewares.map(&:first).include?(middleware)
    case on_conflict
    when :raise
      raise "..."
    when :replace
      ...
    when :append
      ...
    end
  end
  
  ...
end

PR #4464, opened 2025-10-09
feat(multiple-payment-methods): add gql support for listing payment methods

lovrocolic:
## Roadmap Task

Currently in Lago, there can be only one payment method per customer

## Context

With this feature, each customer could have multiple payment provider connections and multiple payment methods per connection.

## Description

This PR adds GQL support for listing payment methods per customer.

### DIFF
--- a/app/graphql/resolvers/payment_methods_resolver.rb
+++ b/app/graphql/resolvers/payment_methods_resolver.rb
@@ -27,7 +27,7 @@ module Resolvers
         }
       )
 
-      result.payment_methods
+      result.payment_methods.includes(:payment_provider)
     end
   end
 end


app/graphql/resolvers/payment_methods_resolver.rb
```diff
@@ -0,0 +1,33 @@
...
+        filters: {
+          external_customer_id:
+        },
+        pagination: {
+          page:,
+          limit:
+        }
+      )
+
+      result.payment_methods
```
julienbourdeau: The payment provider type and code will likely be always requested so I'm in favor of eager loading.

```suggestion
      result.payment_methods.includes(:payment_providers)
```

PR #4466, opened 2025-10-10
Feat(fixed_charges): update fixed_charge and subscription models (1)

annvelents:
## Context

As part of deconstruction of a big PR responsible for creating fixed charges fees, we need these missing updates in the models

## Description

- added new scope to a fixed_charge
- fix validations to not throw warnings during tests
- added included_in_next_subscription to a fixed charge

- added fixed_charge boundaries to subscription's adjusted_boundaries

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -246,8 +246,7 @@ class Subscription < ApplicationRecord
       fixed_charges_duration: dates_service.fixed_charges_duration_in_days
     )
 
-    # TODO: remove except when fixed charges are fully life
-    InvoiceSubscription.matching?(self, previous_period_boundaries.except(:fixed_charges_from_datetime, :fixed_charges_to_datetime, :fixed_charges_duration)) ? boundaries : previous_period_boundaries
+    InvoiceSubscription.matching?(self, previous_period_boundaries) ? boundaries : previous_period_boundaries
   end
 end
 


app/models/fixed_charge.rb
```diff
@@ -32,13 +33,22 @@ class FixedCharge < ApplicationRecord
...
     charge_model == fixed_charge.charge_model && properties == fixed_charge.properties
   end
+
+  def included_in_next_subscription?(subscription)
+    return false if subscription.next_subscription.nil?
+
+    next_subscription_charges = subscription.next_subscription.plan.fixed_charges
+    return false if next_subscription_charges.blank?
+
+    next_subscription_charges.pluck(:add_on_id).include?(add_on_id)
```
groyoh:
```suggestion
    subscription.next_subscription.plan.fixed_charges.exists?(add_on_id:)
```

app/models/subscription.rb
```diff
@@ -238,8 +238,11 @@ def adjusted_boundaries(datetime, boundaries)
...
       charges_to_datetime: dates_service.charges_to_datetime,
+      fixed_charges_from_datetime: dates_service.fixed_charges_from_datetime,
+      fixed_charges_to_datetime: dates_service.fixed_charges_to_datetime,
       timestamp: datetime,
-      charges_duration: dates_service.charges_duration_in_days
+      charges_duration: dates_service.charges_duration_in_days,
+      fixed_charges_duration: dates_service.fixed_charges_duration_in_days
     )
 
     InvoiceSubscription.matching?(self, previous_period_boundaries) ? boundaries : previous_period_boundaries
```
annvelents, line 249: if we start populating this values, will it break? (comparing new invoice with stored in the db)

annvelents: firstly we want to make sure that subscriptions have these values! then we'll do some population of this values, and only after we can rely on this logic of adding fixed_charges into the comparison :(

annvelents: cancel all the worries, it only applies if "fixed_charges_billed_in_monthly_split_intervals" - plan is yearly, fixed_charges are applied monthly 😌

PR #4469, opened 2025-10-10
feat(multiple-payment-methods): add API support for listing customer payment methods

lovrocolic:
## Context

Currently in Lago, there can only be one payment provider customer with only one attached payment method.

## Description

With this feature, it will be possible to add multiple payment methods per provider customer

This PR adds API support for listing payment methods per customer

### DIFF
--- a/app/serializers/v1/payment_method_serializer.rb
+++ b/app/serializers/v1/payment_method_serializer.rb
@@ -3,24 +3,13 @@
 module V1
   class PaymentMethodSerializer < ModelSerializer
     def serialize
-      payload = {
+      {
         lago_id: model.id,
         is_default: model.is_default,
         payment_provider_code: model.payment_provider&.code,
         payment_provider_type: model.payment_provider_type,
+        provider_method_id: model.provider_method_id,
         created_at: model.created_at.iso8601
-      }
-
-      payload.merge!(customer) if include?(:customer)
-
-      payload
-    end
-
-    private
-
-    def customer
-      {
-        customer: ::V1::CustomerSerializer.new(model.customer).serialize
       }
     end
   end
--- a/app/controllers/concerns/payment_method_index.rb
+++ b/app/controllers/concerns/payment_method_index.rb
@@ -22,8 +22,7 @@ module PaymentMethodIndex
           result.payment_methods.preload(:customer, :payment_provider),
           ::V1::PaymentMethodSerializer,
           collection_name: "payment_methods",
-          meta: pagination_metadata(result.payment_methods),
-          includes: %i[customer]
+          meta: pagination_metadata(result.payment_methods)
         )
       )
     else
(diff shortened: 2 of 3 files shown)


app/serializers/v1/payment_method_serializer.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module V1
+  class PaymentMethodSerializer < ModelSerializer
+    def serialize
+      payload = {
+        lago_id: model.id,
```
julienbourdeau, line 7: Do we plan to ever use this id? 

I'd prefer to use a combination of `payment_provider_code` and `provider_payment_method_id` and avoid lago_ids unless really necessary.
Similar to `external_id` and `code` whenever we can.

app/serializers/v1/payment_method_serializer.rb
```diff
@@ -0,0 +1,27 @@
...
+        is_default: model.is_default,
+        payment_provider_code: model.payment_provider&.code,
+        payment_provider_type: model.payment_provider_type,
+        created_at: model.created_at.iso8601
+      }
+
+      payload.merge!(customer) if include?(:customer)
```
julienbourdeau: I'd prefer to return only the `customer.external_id` and let people make a dedicated call if they need the details 🤔 
At least, if we want to keep sending the entire customer, I think we need to eager load `metadata` and `billing_entity` in the controller.

app/serializers/v1/payment_method_serializer.rb
```diff
@@ -0,0 +1,27 @@
...
+
+module V1
+  class PaymentMethodSerializer < ModelSerializer
+    def serialize
+      payload = {
+        lago_id: model.id,
+        is_default: model.is_default,
+        payment_provider_code: model.payment_provider&.code,
+        payment_provider_type: model.payment_provider_type,
+        created_at: model.created_at.iso8601
```
julienbourdeau, line 12: I think the provider payment method id is missing. It's very important if you want to get all the details directly from the provider.
I'd also add the details which are necessary to build a UI for instance.

lovrocolic: Sounds good, included it

app/serializers/v1/payment_method_serializer.rb
lovrocolic: Agree, not including customer by default. External id is also probably unnecessary since it is mandatory param

app/serializers/v1/payment_method_serializer.rb
lovrocolic: Hmmm, not sure here TBH.
Usually `external_id` is used where applicable as identifier in Lago's client system. However it seems strange to me to use identifier in 3rd party provider to manage Lago API. If there is no `external_id` or if `code` is not applicable, it makes more sense to me to use `lago_id` (as we do for charges for example)

lovrocolic: But anyway, this does not affect this PR and we can talk more about it when we will really need identifier to find certain payment method and perform dedicated action

PR #4476, opened 2025-10-13
feat(fixed charges): Adjusted fee for fixed charges service

ancorcruz:
## Roadmap Task

 ## Context

 ## Description

 Foundation to create adjusted fees for fixed charges

### DIFF
--- a/app/services/adjusted_fees/create_service.rb
+++ b/app/services/adjusted_fees/create_service.rb
@@ -126,14 +126,6 @@ module AdjustedFees
     def create_fee(subscription, chargeable, fee_type)
       invoice_subscription = invoice.invoice_subscriptions.find_by(subscription_id: subscription.id)
 
-      boundaries = {
-        timestamp: invoice_subscription.timestamp,
-        charges_from_datetime: invoice_subscription.charges_from_datetime,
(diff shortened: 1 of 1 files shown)


app/services/adjusted_fees/create_service.rb
```diff
@@ -75,63 +85,66 @@ def find_or_create_fee
...
+    def create_fee(subscription, chargeable, fee_type)
       invoice_subscription = invoice.invoice_subscriptions.find_by(subscription_id: subscription.id)
 
       boundaries = {
         timestamp: invoice_subscription.timestamp,
         charges_from_datetime: invoice_subscription.charges_from_datetime,
-        charges_to_datetime: invoice_subscription.charges_to_datetime
+        charges_to_datetime: invoice_subscription.charges_to_datetime,
+        fixed_charges_from_datetime: invoice_subscription.fixed_charges_from_datetime,
+        fixed_charges_to_datetime: invoice_subscription.fixed_charges_to_datetime
```
ancorcruz: shall it only include the boundaries related to the charge type?

only charge_from/to_date for charges, and fixed_charges_from/to_date for fixed charges?

annvelents (approved):
Nice rework of adjusted_fee/create_service! 👏

PR #4478, opened 2025-10-13
fix(credit-note): Regenerate number after finalizing

ivannovosad:
## Context

When a draft credit note is finalized, there was a bug that prevented to regenerate the number, so it remained as e.g.: `LAG-C123-DRAFT-CN001`

## Description

This PR fixes `CreditNote#ensure_number` so that the number is regenerated.


julienbourdeau (approved):
I see we do the same in invoice 👌

PR #4481, opened 2025-10-13
feat(multiple-payment-methods): add gql support for setting payment method as default

lovrocolic:
## Context

Currently in Lago, there can only be one payment provider customer with only one attached payment method.

## Description

This PR adds logic that enables user to select certain payment method as the default one. Also, previous default payment method is marked as `is_default = false`

### DIFF
--- a/spec/models/organization_spec.rb
+++ b/spec/models/organization_spec.rb
@@ -43,7 +43,7 @@ RSpec.describe Organization do
       expect(subject).to have_many(:invoice_custom_sections)
       expect(subject).to have_many(:ai_conversations)
       expect(subject).to have_many(:manual_invoice_custom_sections).conditions(section_type: "manual")
-      expect(subject).to have_many(:payment_methods) }
+      expect(subject).to have_many(:payment_methods)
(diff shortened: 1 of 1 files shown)


app/services/payment_methods/set_as_default_service.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def initialize(payment_method:)
+      @payment_method = payment_method
+
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "payment_method") unless payment_method
+      return result if payment_method.is_default?
```
vincent-pochet: should we attach the payment_method to the result as well here. Since it will be a success, we might want to use it to build responses. What do you think?

app/models/customer.rb
```diff
@@ -54,6 +54,7 @@ class Customer < ApplicationRecord
   has_many :payment_provider_customers,
     class_name: "PaymentProviderCustomers::BaseCustomer",
     dependent: :destroy
+  has_many :payment_methods, dependent: :destroy
```
vincent-pochet, line 57: I guess you could add the tests for this new relation and the new one at organization level

lovrocolic: added

PR #4482, opened 2025-10-13
misc(sidekiq): Add `sidekiq-pro` as an optional dependency

groyoh:
## Context

Today, we are using Sidekiq to handle our background jobs. While it works well for most of our needs, we are reaching some limitations with the open-source version, particularly around job reliability during high load.

Sidekiq uses `BRPOP` to fetch a job from the queue in Redis. This is very efficient and simple but it has one drawback: the job is now removed from Redis. If Sidekiq crashes while processing that job, it is lost forever.

## Description

This adds `sidekiq-pro` as an optional dependency which can be install when building the Docker image.

This also updated the `Dockerfile` to:

1. Accept a `BUNDLE_WITH` which can be used to conditionally enabled Sidekiq Pro when the value is `sidekiq-pro`.
2. Mount the `BUNDLE_GEMS__CONTRIBSYS__COM` variable as a Docker secret as it's only required during the build and is a sensitive information.
3. Rely on `BUNDLE_WITHOUT` env variable instead of the `--without` command line option to be consistent and because it is deprecated:

    ```
    [DEPRECATED] The `--without` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set without 'test'`, and stop using this flag
    ```

Note that we also added a `LAGO_SIDEKIQ_PRO_REQUIRED` env variable which will cause the app to not boot if Sidekiq Pro is not properly installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM ruby:3.4.7-slim AS build
 
+ARG BUNDLE_WITH
+
 WORKDIR /app
 
 RUN apt update && apt upgrade -y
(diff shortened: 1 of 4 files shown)


julienbourdeau:
Am I correct to understand that oss release will ship **without** sidekiq pro in the image and you **must rebuild the image** with your own `BUNDLE_GEMS__CONTRIBSYS__COM` to use sidekiq pro?

Dockerfile
```diff
@@ -28,6 +36,9 @@ ENV SEGMENT_WRITE_KEY=$SEGMENT_WRITE_KEY
 ENV GOCARDLESS_CLIENT_ID=$GOCARDLESS_CLIENT_ID
 ENV GOCARDLESS_CLIENT_SECRET=$GOCARDLESS_CLIENT_SECRET
 
+ENV BUNDLE_WITH=${BUNDLE_WITH:-}
+ENV BUNDLE_WITHOUT="development test"
```
IxDay, line 45: Did we try without this here? I have the intuition that it is needed for the build process, not really for the actual run. WDYT?

groyoh: Yes I did test (I actually forgot to add it initially 😄). And yes, it's needed. When we start the container, we run `bundle exec rails s`, without the env variable `bundle` will try to load `development` and `test` gems and fail.

IxDay: 👍

PR #4483, opened 2025-10-14
fix(mailer): Use lago from email if org.email is not set

julienbourdeau:
If an organization has the premium integration `from_email`, we'll use `organization.email` as a from address. If the address is not set in the organization, we should default to the lago from address.

The Admin emails should always use the lago address. **Ideally, the "public" and "internal" emails should be separated!**

This was spotted from a dead job `ArgumentError: SMTP From address may not be blank: nil` on a staging organization.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -202,7 +202,7 @@ class Organization < ApplicationRecord
   end
 
   def from_email_address
-    return email if from_email_enabled? && email
+    return email if from_email_enabled?
 
     ENV["LAGO_FROM_EMAIL"]
   end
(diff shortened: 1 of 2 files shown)


app/mailers/organization_mailer.rb
groyoh, line 1: Maybe we could add a `when organization from_email is set, it sends with Lago from email` test ?

julienbourdeau: I'm not sure. I understand the idea of adding a non-regression tests but it should not have been there in the first place. The test already checks that the email is the lago one.

julienbourdeau:
I have reverted the other change.

As I was writing writing to product team, I realize that it's very unlikely that it happens to production because we need some manual setup in aws to configure the from email.
Maybe a better fix is to add validation on organization and adding `from_email` can only be done when a `organization.email` is set. It's probably best to fail and fix manually for now.

PR #4484, opened 2025-10-14
feat(multiple-payment-methods): add api support for setting payment method as default

lovrocolic:
## Context

## Description

This PR adds API support for selecting certain payment method as the default one. Also, previous default payment method is marked as `is_default = false`

### DIFF
--- a/spec/requests/api/v1/customers/payment_methods_controller_spec.rb
+++ b/spec/requests/api/v1/customers/payment_methods_controller_spec.rb
@@ -72,6 +72,17 @@ RSpec.describe Api::V1::Customers::PaymentMethodsController, type: :request do
       end
     end
 
+    context "with unknown payment method" do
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -43,10 +43,12 @@
           resources :credit_notes, only: %i[index]
           resources :invoices, only: %i[index]
           resources :payments, only: %i[index]
-          resources :payment_methods, only: %i[index]
           resources :payment_requests, only: %i[index]
           resources :subscriptions, only: %i[index]
           resources :wallets, only: %i[index]
+          resources :payment_methods, only: %i[index] do
+            put :set_as_default, on: :member
```
lovrocolic, line 50: @julienbourdeau Here is my answer about identifier: https://github.com/getlago/lago-api/pull/4469/files#r2425785169

Somehow using `:id` makes sense to me here. Many other resources that are created 'under the hood' also use internal lago id as identifier (like invoice).

Let me know if you strongly disagree 😄

PR #4490, opened 2025-10-15
fix(manual-payments): Fix manual payments with no amount_cents

ivannovosad:
## Context

When calling the endpoint for creating a manual payment: `POST /api/v1/payments`
It failed if `amount_cents` was missing or misspelled e.g.: `amount_in_cents`.

## Description

This PR addresses the issue with returning `422` error.

```
error_code: "value_is_mandatory", field: "amount_cents"
```

### DIFF
--- a/app/services/payments/manual_create_service.rb
+++ b/app/services/payments/manual_create_service.rb
@@ -73,7 +73,7 @@ module Payments
       return result.forbidden_failure! if !License.premium?
       return result.forbidden_failure! unless invoice.allow_manual_payment?
       result.single_validation_failure!(error_code: "invalid_date", field: "paid_at") unless valid_paid_at?
-      result.single_validation_failure!(error_code: "value_is_mandatory", field: "amount_cents") unless valid_amount_cents?
+      result.single_validation_failure!(error_code: "invalid_value", field: "amount_cents") unless valid_amount_cents?
     end
 
     def valid_paid_at?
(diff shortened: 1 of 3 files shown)


app/services/payments/manual_create_service.rb
```diff
@@ -73,10 +73,15 @@ def check_preconditions
       return result.forbidden_failure! if !License.premium?
       return result.forbidden_failure! unless invoice.allow_manual_payment?
       result.single_validation_failure!(error_code: "invalid_date", field: "paid_at") unless valid_paid_at?
+      result.single_validation_failure!(error_code: "value_is_mandatory", field: "amount_cents") unless valid_amount_cents?
```
groyoh: Since users could provide a string or float, maybe it should be:

```suggestion
      result.single_validation_failure!(error_code: "invalid_value", field: "amount_cents") unless valid_amount_cents?
```

spec/requests/api/v1/payments_controller_spec.rb
```diff
@@ -26,22 +26,47 @@
...
+    context "when all parameters are valid" do
+      before do
+        allow(Payments::ManualCreateService).to receive(:call).and_return(
+          BaseService::Result.new.tap { |r| r.payment = payment }
+        )
+      end
+
+      include_examples "requires API permission", "payment", "write"
+
+      it "delegates to Payments::ManualCreateService", :aggregate_failures do
```
groyoh:
```suggestion
      it "delegates to Payments::ManualCreateService" do
```

app/services/payments/manual_create_service.rb
ivannovosad: Good point - changing it.

PR #4496, opened 2025-10-15
feat(fixed-charges): Add invoice#different_boundaries_for_subscription_and_fixed_charges

ancorcruz:
`different_boundaries_for_subscription_and_fixed_charges` is required to decide how we render the invoice fixed charge fees within the invoice document.

also cover with tests invoice#different_boundaries_for_subscription_and_charge

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -356,20 +356,22 @@ class Invoice < ApplicationRecord
     end
   end
 
-  def different_boundaries_for_subscription_and_charges(subscription)
-    subscription_from = invoice_subscription(subscription.id).from_datetime_in_customer_timezone&.to_date
-    subscription_to = invoice_subscription(subscription.id).to_datetime_in_customer_timezone&.to_date
-    charges_from = invoice_subscription(subscription.id).charges_from_datetime_in_customer_timezone&.to_date
-    charges_to = invoice_subscription(subscription.id).charges_to_datetime_in_customer_timezone&.to_date
+  def has_different_boundaries_for_subscription_and_charges?(subscription)
+    invoice_subscription = invoice_subscription(subscription.id)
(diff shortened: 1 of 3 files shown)


app/models/invoice.rb
```diff
@@ -365,6 +365,15 @@ def different_boundaries_for_subscription_and_charges(subscription)
     subscription_from != charges_from && subscription_to != charges_to
   end
 
+  def different_boundaries_for_subscription_and_fixed_charges(subscription)
```
julienbourdeau: I think the name should at least end with a `?`
Maybe it should start with `has_` 🤷‍♂️

app/models/invoice.rb
```diff
@@ -365,6 +365,15 @@ def different_boundaries_for_subscription_and_charges(subscription)
...
+    subscription_from = invoice_subscription(subscription.id).from_datetime_in_customer_timezone&.to_date
+    subscription_to = invoice_subscription(subscription.id).to_datetime_in_customer_timezone&.to_date
+    fixed_charges_from = invoice_subscription(subscription.id).fixed_charges_from_datetime_in_customer_timezone&.to_date
+    fixed_charges_to = invoice_subscription(subscription.id).fixed_charges_to_datetime_in_customer_timezone&.to_date
```
julienbourdeau: invoice_subscription will use `find_by` and it's not memoized, so I think you should extract it


```suggestion
    is = invoice_subscription(subscription.id)
    
    subscription_from = is.from_datetime_in_customer_timezone&.to_date
    subscription_to = is.to_datetime_in_customer_timezone&.to_date
    fixed_charges_from = is.fixed_charges_from_datetime_in_customer_timezone&.to_date
    fixed_charges_to = is.fixed_charges_to_datetime_in_customer_timezone&.to_date
```

julienbourdeau (approved):

PR #4498, opened 2025-10-16
Misc optimize estimate fees

ivannovosad:
## Description

This PR contains some minor optimisations of `Fees::EstimatePayInAdvanceService` that is called by `Fees::EstimatePayInAdvanceService`.

### DIFF
--- a/app/services/fees/create_pay_in_advance_service.rb
+++ b/app/services/fees/create_pay_in_advance_service.rb
@@ -6,13 +6,7 @@ module Fees
 
     def initialize(charge:, event:, billing_at: nil, estimate: false)
       @charge = charge
-      ActiveRecord::Associations::Preloader.new(
-        records: [@charge],
-        associations: [:billable_metric, :applied_pricing_unit, :pricing_unit]
-      ).call
-
       @event = Events::CommonFactory.new_instance(source: event)
-
       @billing_at = billing_at || @event.timestamp
       @estimate = estimate
 


app/services/fees/create_pay_in_advance_service.rb
```diff
@@ -6,7 +6,13 @@ class CreatePayInAdvanceService < BaseService
 
     def initialize(charge:, event:, billing_at: nil, estimate: false)
       @charge = charge
+      ActiveRecord::Associations::Preloader.new(
+        records: [@charge],
+        associations: [:billable_metric, :applied_pricing_unit, :pricing_unit]
+      ).call
```
julienbourdeau: I'm not entirely convinced this is useful because there is only one charge here and all these relations re has_one or belongs too 🤔 
My understanding is that it would result in the same amount of queries so doing it lazily may even be better.
Maybe I'm missing something

app/services/fees/create_pay_in_advance_service.rb
```diff
@@ -55,6 +61,10 @@ def call
     delegate :billable_metric, to: :charge
     delegate :subscription, to: :event
 
+    def filter
+      @filter ||= ChargeFilters::EventMatchingService.call(charge:, event:).charge_filter
+    end
```
julienbourdeau, line 60: 💯 💯

app/services/fees/create_pay_in_advance_service.rb
ivannovosad: Lets remove that for now.

vincent-pochet (approved):

PR #4500, opened 2025-10-17
test(event): Move more event store tests to `shared_examples`

groyoh:
## Context

https://github.com/getlago/lago-api/commit/b3131f433bfd38831a0a32c5deb1376659028902 introduced a `shared_examples` to improve consistency between the event stores tests. The following methods were not tested in the `shared_examples` though:

- `prorated_events_values`
- `prorated_sum`
- `grouped_prorated_sum`
- `weighted_sum`
- `grouped_weighted_sum`
- `prorated_unique_count`
- `grouped_prorated_unique_count`
- `prorated_unique_count_breakdown`

Furthermore, the `weighted_sum_breakdown` method was not tested.

## Description

This moves the following methods tests to the `shared_examples`:

- `prorated_events_values`
- `prorated_sum`
- `grouped_prorated_sum`
- `weighted_sum`
- `grouped_weighted_sum`
 
Tests were also added for the `weighted_sum_breakdown` method.

It looks like the behavior of the prorated unique count methods actually differs from the Postgres and Clickhouse stores so they were not moved to the `shared_examples`.

This also fixes an issue with the `weighted_sum_breakdown` method of the Clickhouse store which returned the `second_duration` as string rather than int. That's because `date_diff` actually returns an `Int64` and ActiveRecord transform that into a `String` so we call `#to_i` to ensure consistency between implementations:

```ruby
lago-api(staging)> Clickhouse::BaseRecord.connection.select_one("SELECT 1::Int64")
=> {"CAST('1', 'Int64')" => "1"}
lago-api(staging)> Clickhouse::BaseRecord.connection.select_one("SELECT 1::Int32")
=> {"CAST('1', 'Int32')" => 1}
```

Note that the `weighted_sum_breakdown` method of the Clickhouse stores returns timestamps with a 5 decimal precision rather than 3 as in the other methods.


app/services/events/stores/clickhouse_store.rb
```diff
@@ -566,6 +566,9 @@ def weighted_sum_breakdown(initial_value: 0)
               ]
             )
           ).rows
+          rows.map do |row|
+            [row[0], row[1], row[2], row[3].to_i, row[4]]
+          end
```
ancorcruz: could we avoid `map` returning the expected values from the SQL query? we could use clickhouse cast functions to achieve the `to_i` call.

groyoh: We could and that's what I tried first. But `date_diff` actually returns an `Int64` and ActiveRecord transform that into a `String`. If we cast the result in a `Int32`, then we get the result as `Integer`:

```ruby
lago-api(staging)> Clickhouse::BaseRecord.connection.select_one("SELECT 1::Int64")
=> {"CAST('1', 'Int64')" => "1"}
lago-api(staging)> Clickhouse::BaseRecord.connection.select_one("SELECT 1::Int32")
=> {"CAST('1', 'Int32')" => 1}
```

I'm not sure if it's a good thing to cast into `Int32` so I kept the logic in Ruby 😅 I'll add a comment.

PR #4501, opened 2025-10-17
fix(event): Fix Postgres event store connection

groyoh:
## Context

Some of the Postgres store methods used `ActiveRecord::Base.connection` which caused the tests to fail when running with `transaction` database cleaner strategy. This is due to the fact that we would create the event in a transaction of the `primary` connection and check the results in the `events` connection which could not retrieve the uncommitted data.

Note that it works properly outside of tests because both connections use the same database and we never actually try to fetch events from the store within the same transaction that create events.

## Description

This fixes it by relying on `Event.connection`. It also slightly refactors the code to limit mistakes in the future.

### DIFF
--- a/app/services/events/stores/postgres_store.rb
+++ b/app/services/events/stores/postgres_store.rb
@@ -59,7 +59,7 @@ module Events
           )
           .to_sql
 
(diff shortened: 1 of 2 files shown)


ancorcruz (approved):
looks good.

PR #4504, opened 2025-10-19
chore(events): Improve events post processing

jdenquin:


app/services/events/post_process_service.rb
```diff
@@ -109,7 +106,9 @@ def handle_pay_in_advance
       return unless billable_metric
       return unless charges.any?
 
-      Events::PayInAdvanceJob.perform_later(Events::CommonFactory.new_instance(source: event).as_json)
+      Events::PayInAdvanceJob
+        .set(wait: rand(10..60).seconds)
```
julienbourdeau, line 110: I think before we merge this we want to complete this one: https://github.com/getlago/lago-api/pull/4136

Happy to discuss the details but today we count all events present in DB, it works because queues are super fast and pay in advance are _rare-ish_ by design.
If we introduce delay, we increase the chance of new events being received and being counted in the aggregation.

With the `rand(10..60)` is likely that event 2 received after event 1 is processed first. When processing event 1, we'll count event 2 in the aggregation.

PR #4507, opened 2025-10-20
feat(multiple-payment-methods): add api support for removing payment method

lovrocolic:
## Context

Currently each customer can be attached to only one payment provider and also have only one payment method.

## Description

With this feature it will be possible for Lago customers to have multiple payment methods.

This PR adds API support for removing payment method.

### DIFF
--- a/spec/requests/api/v1/customers/payment_methods_controller_spec.rb
+++ b/spec/requests/api/v1/customers/payment_methods_controller_spec.rb
@@ -131,11 +131,11 @@ RSpec.describe Api::V1::Customers::PaymentMethodsController, type: :request do
   end
 
   describe "DELETE /api/v1/customers/:external_id/payment_methods/:id" do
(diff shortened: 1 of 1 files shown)


spec/requests/api/v1/customers/payment_methods_controller_spec.rb
```diff
@@ -125,8 +125,65 @@
...
+
+      it "soft deletes the payment method" do
+        freeze_time do
+          expect { subject }.to change(PaymentMethod, :count).by(-1)
+            .and change { payment_method.reload.deleted_at }.from(nil).to(Time.current)
+        end
+      end
+    end
+
+    it "returns deleted payment method" do
```
groyoh: This should be in the context above I guess.

lovrocolic: nice catch!

PR #4508, opened 2025-10-20
fix(cached_aggregation): Refact index for better performances

vincent-pochet:
## Description

This PR follows https://github.com/getlago/lago-api/pull/4505

The previously introduced index was not performant enough in some situation and it was especially missing the issue with the ordering performance (on `timestamp` and `created_at`)

A new index is now added to take care of this ordering and of the `charge_filter_id` field as well.

### DIFF
--- a/db/migrate/20251020142629_add_index_on_cached_aggregation_created_at.rb
+++ b/db/migrate/20251020142629_add_index_on_cached_aggregation_created_at.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+class AddIndexOnCachedAggregationCreatedAt < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def up
+    remove_index :cached_aggregations, name: :idx_aggregation_lookup_with_transaction_id, if_exists: true
+
+    safety_assured do
+      add_index(
+        :cached_aggregations,
+        [:organization_id, :external_subscription_id, :charge_id, :timestamp, :created_at],
+        order: {timestamp: :desc, created_at: :desc},
+        include: [:grouped_by, :charge_filter_id, :event_transaction_id],
+        name: :idx_cached_aggregation_filtered_lookup,
+        algorithm: :concurrently
+      )
+    end
+  end
+
+  def down
+    safety_assured do
+      add_index(
+        :cached_aggregations,
+        [:event_transaction_id, :external_subscription_id, :charge_id, :timestamp],
+        include: [:organization_id, :grouped_by],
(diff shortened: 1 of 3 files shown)


db/migrate/20251020142629_add_index_on_cached_aggregation_created_at.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class AddIndexOnCachedAggregationCreatedAt < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def up
+    remove_index :cached_aggregations, name: :idx_aggregation_lookup_with_transaction_id, if_exists: true
```
groyoh, line 7: Should this index be removed afterward to avoid performance issues in between the removal and the index ?

db/migrate/20251020142629_add_index_on_cached_aggregation_created_at.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    safety_assured do
+      add_index(
+        :cached_aggregations,
+        [:organization_id, :external_subscription_id, :charge_id, :timestamp, :created_at],
+        order: {timestamp: :desc, created_at: :desc},
+        include: [:grouped_by, :charge_filter_id, :event_transaction_id],
```
groyoh, line 14: Based on the queries I could find, I'm not sure to see why you would use `include` here. Isn't `include` is only useful when you select those specific fields through `select` or `pluck` ?

brunomiguelpinto (approved):

db/migrate/20251020142629_add_index_on_cached_aggregation_created_at.rb
vincent-pochet: We have an other one actually, this one was never used in prod (added this morning and not yet deployed)

db/migrate/20251020142629_add_index_on_cached_aggregation_created_at.rb
vincent-pochet: PG is also able to use those field for conditional filtering without enlarging the index itself. It will be some sort of post filtering.
We made some test in production and it appears to be working as expected even if we filter with a `event_transaction_id`

An other approach would be to add 3 indexes on this table for different filtering scenarios. It would bring more complexity on it so we decided to give a try with a "simpler" approach

groyoh: Interesting 👍

PR #4519, opened 2025-10-22
feat(netsuite): Introduce `currencies` mapping

julienbourdeau:
## Context

With Netsuite integration, we need to map currencies to their netsuite objects.

## Description

This PR introduces a new mapping type `currencies`. It's only used for Netsuite.

We save the mapping as a hashmap because it's best way to store mapping 😄  The mapping is meant to be very rarely edited but used a lot.

Usage:
```rb
payload[:currency] = mapping_model.currencies[invoice.currency]
```

Next PR will start using this mapping to update integration payload.

### Validation

The model validates the format of currencies hashmap. Create and Update services already handles `ActiveRecord::InvalidRecord`

GraphQL isn't really meant to use free form hashmaps so we use an array of hashmap. This PR introduces a custom validator  to ensure that a "key" is only used once. I wanted this in the past for threshold for instance so I made it configurable.

In this case, we cannot return an error code like we usually do. Notice that the frontend should already make this duplication impossible, but just in case, the backend will also throw an error.

```rb
# GQL
[ {currency_code: "EUR", external_currency_code: "2"} ]

# Model
{"EUR" => "2"}
```

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1042,8 +1042,9 @@ type ChargeUsage {
 
 type CollectionMapping {
   billingEntityId: ID
+  currencies: [CurrencyMappingItem!]
   externalAccountCode: String
-  externalId: String!
+  externalId: String
   externalName: String
   id: ID!
   integrationId: ID!
@@ -2849,8 +2850,9 @@ input CreateIntegrationCollectionMappingInput {
   A unique identifier for the client performing the mutation.
   """
   clientMutationId: String
+  currencies: [CurrencyMappingItemInput!]
   externalAccountCode: String
-  externalId: String!
+  externalId: String
   externalName: String
   integrationId: ID!
   mappingType: MappingTypeEnum!
@@ -3983,6 +3985,16 @@ enum CurrencyEnum {
   Zambian Kwacha
   """
   ZMW
+}
+
+type CurrencyMappingItem {
+  currencyCode: CurrencyEnum!
+  currencyExternalCode: String!
+}
+
+input CurrencyMappingItemInput {
+  currencyCode: CurrencyEnum!
+  currencyExternalCode: String!
 }
 
 """
@@ -6241,6 +6253,7 @@ enum MappingTypeEnum {
   account
   coupon
   credit_note
+  currencies
   fallback_item
   minimum_commitment
   prepaid_credit
@@ -11255,6 +11268,7 @@ input UpdateIntegrationCollectionMappingInput {
   A unique identifier for the client performing the mutation.
   """
   clientMutationId: String
+  currencies: [CurrencyMappingItemInput!]
   externalAccountCode: String
   externalId: String
   externalName: String
--- a/schema.json
+++ b/schema.json
@@ -6624,6 +6624,26 @@
               "deprecationReason": null
             },
             {
+              "name": "currencies",
+              "description": null,
+              "args": [],
+              "type": {
+                "kind": "LIST",
+                "name": null,
+                "ofType": {
+                  "kind": "NON_NULL",
+                  "name": null,
+                  "ofType": {
+                    "kind": "OBJECT",
+                    "name": "CurrencyMappingItem",
+                    "ofType": null
+                  }
+                }
+              },
+              "isDeprecated": false,
+              "deprecationReason": null
+            },
+            {
               "name": "externalAccountCode",
               "description": null,
               "args": [],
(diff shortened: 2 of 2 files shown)


app/graphql/validators/unique_code_in_array_validator.rb
```diff
@@ -0,0 +1,20 @@
...
+    def initialize(code_key: :code, **default_options)
+      @code_key = code_key
+      super(**default_options)
+    end
+
+    def validate(object, context, value)
+      duplicates = value.map { it[code_key] }.tally.select { |_, count| count > 1 }.keys
+
+      if duplicates.any?
+        "duplicate_currency_code"
```
groyoh: This should be generic.

app/graphql/validators/unique_code_in_array_validator.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Validators
+  class UniqueCodeInArrayValidator < GraphQL::Schema::Validator
```
groyoh: This could be "unique by field" rather than code to be even more generic.

app/models/integration_collection_mappings/netsuite_collection_mapping.rb
```diff
@@ -3,6 +3,35 @@
...
   class NetsuiteCollectionMapping < BaseCollectionMapping
     settings_accessors :tax_nexus, :tax_type, :tax_code
+    settings_accessors :currencies
+
+    validate :currency_mapping_format
+
+    private
+
+    def currency_mapping_format
+      # Other mapping_types shouldn't have currencies, but if they do, we validate the format
```
groyoh, line 20: Shouldn't we ensure abscence instead ?

app/models/integration_collection_mappings/netsuite_collection_mapping.rb
```diff
@@ -3,6 +3,35 @@
 module IntegrationCollectionMappings
   class NetsuiteCollectionMapping < BaseCollectionMapping
     settings_accessors :tax_nexus, :tax_type, :tax_code
+    settings_accessors :currencies
+
+    validate :currency_mapping_format
```
groyoh, line 8: We should validate that currencies are on org-level.

app/graphql/types/integration_collection_mappings/base_input.rb
```diff
@@ -0,0 +1,25 @@
...
+      argument :external_id, String, required: false
+      argument :external_name, String, required: false
+
+      argument :tax_code, String, required: false
+      argument :tax_nexus, String, required: false
+      argument :tax_type, String, required: false
+
+      argument :currencies, [CurrencyMappingItemInput], required: false,
+        prepare: :convert_currencies_to_hash,
+        validates: {::Validators::UniqueCodeInArrayValidator => {code_key: :currency_code}}
```
groyoh: 💯

app/graphql/types/integration_collection_mappings/currency_mapping_item_input.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Types
+  module IntegrationCollectionMappings
+    class CurrencyMappingItemInput < Types::BaseInputObject
```
groyoh, line 5: I believe this is not tested.

app/graphql/validators/unique_code_in_array_validator.rb
groyoh: Do you think you could add a unit test for this ?

app/models/integration_collection_mappings/netsuite_collection_mapping.rb
```diff
@@ -3,6 +3,35 @@
...
+        errors.add(:currencies, "cannot_be_empty")
+      elsif !currencies_hash_valid?
+        errors.add(:currencies, "invalid_format")
+      end
+    end
+
+    def currencies_hash_valid?
+      valid_currencies = Currencies::ACCEPTED_CURRENCIES.keys
+      currencies.keys.all? { it.is_a?(String) && valid_currencies.include?(it.to_sym) } &&
+        currencies.values.all? { it.is_a?(String) && it != "" }
```
groyoh:
```suggestion
        currencies.values.all? { it.is_a?(String) && it.present? }
```

because:

```ruby
lago-api(dev)> "   ".present?
=> false
```

app/graphql/validators/unique_code_in_array_validator.rb
julienbourdeau: Absolutely! 🤦‍♂️

app/graphql/validators/unique_code_in_array_validator.rb
julienbourdeau: Yeah, I'll rename.
For the unit test, I'm not too sure how to instanciate and stuff

app/models/integration_collection_mappings/netsuite_collection_mapping.rb
julienbourdeau: I remember why I did this.
It made testing easier (in the short term) in gql mutaitons ^^

PR #4521, opened 2025-10-22
feat(subscription): Expose subscription_at in milliseconds

julienbourdeau:
When retrieving events to be billed, we'll only consider events sent after the `subscription.started_at`.

On Subscription model, we store `started_at` with nanoseconds precision: `2024-04-23T10:01:09.123456Z`
On Event model in CH, we store the `timestamp` at the millisecond precision: `2024-04-23T10:01:09.123Z`
On Event model in PG, we store the `timestamp` at the nanoseconds precision: `2024-04-23T10:01:09.123456Z`, but in the service, converting the timestamp to a ruby datetime will lose some precision.

This means for a subscription started at 10:01:09.666, an event sent at 10:01:09.000 will not be counted. This is not a bug, nor recent, but our API only expose the `started_at` at the second precision so this is impossible to know if an event at the same second will be counted or not.

**Adding precision is not a breaking change because this is still `iso8601` format**, but with more precision.

This PR adds precision down to the microseconds (3 places).

### DIFF
--- a/app/serializers/v1/subscription_serializer.rb
+++ b/app/serializers/v1/subscription_serializer.rb
@@ -13,7 +13,7 @@ module V1
         status: model.status,
         billing_time: model.billing_time,
         subscription_at: model.subscription_at&.iso8601,
-        started_at: model.started_at&.iso8601,
+        started_at: model.started_at&.iso8601(3),
         trial_ended_at: model.trial_ended_at&.iso8601,
         ending_at: model.ending_at&.iso8601,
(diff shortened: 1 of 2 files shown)


groyoh (approved):

spec/scenarios/create_event_spec.rb
groyoh, line 1: Maybe we should add a event within the same second but that's out of range when it comes to milliseconds.

vincent-pochet (approved):

PR #4527, opened 2025-10-23
feat(e-invoices/tags): FacturX and UBL tags for France

mariohd:
This PR contains only the EInvoices tags for FacturX and UBL, along with the tests for each tag.

It's the initial PR of a series of PRs that are going to be merged for EInvoices.

1. Add attachment to PDFs
3. Invoices
3. Credit Notes
4. Payments

### DIFF
--- a/app/serializers/e_invoices/factur_x/applicable_trade_tax.rb
+++ b/app/serializers/e_invoices/factur_x/applicable_trade_tax.rb
@@ -12,7 +12,7 @@ module EInvoices
         @tax_amount = tax_amount
       end
 
-      def call
+      def serialize
         xml.comment "Tax Information #{percent(tax_rate)} #{VAT}"
         xml["ram"].ApplicableTradeTax do
           xml["ram"].CalculatedAmount format_number(tax_amount)
--- a/spec/serializers/e_invoices/factur_x/header_spec.rb
+++ b/spec/serializers/e_invoices/factur_x/header_spec.rb
@@ -5,7 +5,7 @@ require "rails_helper"
 RSpec.describe EInvoices::FacturX::Header do
   subject do
     xml_document(:factur_x) do |xml|
-      described_class.call(xml:, resource:, type_code:, notes:)
+      described_class.serialize(xml:, resource:, type_code:, notes:)
     end
   end
 
@@ -19,7 +19,7 @@ RSpec.describe EInvoices::FacturX::Header do
 
   before { invoice }
 
-  describe ".call" do
+  describe ".serialize" do
     it { is_expected.not_to be_nil }
 
     it "contains section name as comment" do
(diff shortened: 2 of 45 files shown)


spec/serializers/e_invoices/factur_x/header_spec.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe EInvoices::FacturX::Header, type: :service do
```
julienbourdeau: I know this pre-dates the refactor from Yohan but you can get rid of all the `, type: ...`

app/serializers/e_invoices/ubl/base_serializer.rb
```diff
@@ -0,0 +1,35 @@
...
+        @resource = resource
+      end
+
+      def self.call(*, **, &)
+        new(*, **).call(&)
+      end
+
+      private
+
+      attr_accessor :xml, :resource
```
julienbourdeau: I think this is the same for ubl and for factur_x so I'd move it to the BaseSerializer.

spec/support/xml_helper.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module XMLHelper
+  NAMESPACES = {
+    factur_x: EInvoices::FacturX::BaseSerializer::ROOT_NAMESPACES,
+    ubl: EInvoices::Ubl::BaseSerializer::COMMON_NAMESPACES
```
julienbourdeau, line 6: You absolutely don't have to change anything here but it's surprising that one is ROOT and the other is COMMON.

maybe we could have ROOT_NAMESPACES for Ubl and make it a dup of common. Maybe rename common to root, maybe just leave it 😄 I'll let you judge!

app/serializers/e_invoices/factur_x/applicable_trade_tax.rb
```diff
@@ -0,0 +1,39 @@
...
+      def initialize(xml:, tax_category:, tax_rate:, basis_amount:, tax_amount:)
+        super(xml:)
+
+        @tax_category = tax_category
+        @tax_rate = tax_rate
+        @basis_amount = basis_amount
+        @tax_amount = tax_amount
+      end
+
+      def call
```
julienbourdeau: Because they are serializers now, I'm kinda tempted to rename `call` to `serialize` everywhere 😬 

What do you think?

spec/support/xml_helper.rb
mariohd: It's strange but it's because the UBL has different namespaces depending the record. 

it has a different namespaces for invoices, credit_notes and payments, so it has a `COMMON_NAMESPACE`, `INVOICE_NAMESPACES`, `CREDIT_NOTE_NAMESPACES` and will have `PAYMENT_NAMESPACES`.

app/serializers/e_invoices/factur_x/applicable_trade_tax.rb
mariohd: Great! I can change that =)

PR #4528, opened 2025-10-23
feat(multiple-payment-methods): add payment method reference to some collections

lovrocolic:
## Context

## Description

This PR adds migration that would allow attaching payment method to `wallet`, `wallet recurring rule` and `subscription`.

For example, customer could have default payment method, but this features allows to set different payment method on certain subscription.

Also, in order not to create breaking change, `payment_method_id` is optional. That also means that all existing subscriptions would have `payment_method_id` set to `nil` (and new ones if this attribute is not directly set). However, we also need to provide a way for customer to select manual payment, so having `payment_method_id` as `nil` is not enough to catch this case. That's why we need `payment_provider_type` enum:

1. **provider** - for invoicing use `payment_method_id` or inherit default payment method from customer if `payment_method_id` is `nil`
2. **manual** - usually in combination with `payment_method_id = nil`. Payment method is not inherited from customer, but manual payment will be used.


db/migrate/20251023121031_add_payment_method_to_core_tables.rb
```diff
@@ -0,0 +1,33 @@
...
+    safety_assured do
+      change_table :subscriptions, bulk: true do |t|
+        t.references :payment_method, type: :uuid, null: true, index: {algorithm: :concurrently}
+        t.enum :payment_method_type, enum_type: "payment_method_types", default: "provider", null: false
+      end
+    end
+    add_foreign_key :subscriptions, :payment_methods, validate: false
+
+    safety_assured do
+      change_table :wallets, bulk: true do |t|
```
vincent-pochet: I think we should avoid migrating all table in the same migration file. Specially for wallet as this table is already known for beeing "touchy" 😅

lovrocolic: Separated:)

PR #4529, opened 2025-10-23
(feat/e-invoices-facturx-attachments): Add pdfcpu binary and pdf atta…

mariohd:
**This PR is part of a bigger feature for EInvoices.**

The intent of this PR is to add the [pdfcpu](https://github.com/pdfcpu/pdfcpu) binary to the API Docker image.

It also adds the service to add an attachment to a PDF. 

### Usage
Adds the attachment to the PDF file.
```
PdfAttachmentService.call(file:, attachment:)
# OR
PdfAttachmentService.new(file:, attachment:).call
```

### DIFF
--- a/app/services/utils/pdf_attachment_service.rb
+++ b/app/services/utils/pdf_attachment_service.rb
@@ -24,7 +24,6 @@ module Utils
 
       result
     end
-    alias_method :attach, :call
 
     private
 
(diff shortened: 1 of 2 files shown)


app/services/utils/pdf_attachment_service.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      if success
+        result.file = file
+      else
+        result.third_party_failure!(third_party: "pdfcpu", error_code: "failed", error_message: "")
+      end
+
+      result
+    end
+    alias_method :attach, :call
```
julienbourdeau: Any reason to add the `attach` alias? is it needed somewhere else in following PRs?

julienbourdeau (approved):
I think we can start with `pdfcpu` in the docker image while it's still under development. We'll probably need some "load testing" before General Availability.

app/services/utils/pdf_attachment_service.rb
mariohd: no reason at all, it just feels me more idiomatic to me...but it's not really used anywhere in the code

PR #4531, opened 2025-10-23
feat(api): Introduce `timestamp_from_started_at` filters for Event List endpoint

julienbourdeau:
When billing a customer, we only count events where timestamps is after the `started_at`.

In the API, to retrieve only valid events we introduce this new boolean filter instead of requesting the precise timestamp.

In a following PR, I'll add validation on the filters as well. This filter can only be passed if `timestamp_from` is not set. I think the cast to bool will be done in the validation as well.

In term of performance, there is nothing new, we can already filter on timestamps so the indexes already exists. The one thing added is retrieving a subscription by external id.

### DIFF
--- a/app/queries/events_query.rb
+++ b/app/queries/events_query.rb
@@ -51,7 +51,11 @@ class EventsQuery < BaseQuery
   end
 
   def subscription
-    @subscription ||= organization.subscriptions.find_by(external_id: filters.external_subscription_id)
+    @subscription ||= organization.subscriptions
+      .order("terminated_at DESC NULLS FIRST, started_at DESC")
+      .find_by(
+        external_id: filters.external_subscription_id
+      )
   end
 
   def timestamp_from


app/queries/events_query.rb
```diff
@@ -38,16 +39,34 @@ def with_external_subscription_id(scope)
...
+
     scope
   end
 
+  def subscription
+    @subscription ||= organization.subscriptions
+      .order("terminated_at DESC NULLS FIRST, started_at DESC")
+      .find_by!(
+        external_id: filters.external_subscription_id
+      )
```
julienbourdeau, line 58: This is how we retrieve the correct subscription via external_id on the `/subscriptions/:external_id/*` endpoints.
See [here](https://github.com/getlago/lago-api/blob/c674e09fe46ff4023cd1187481d347341af0af00/app/controllers/api/v1/subscriptions_controller.rb#L135), [here](https://github.com/getlago/lago-api/blob/c674e09fe46ff4023cd1187481d347341af0af00/app/graphql/resolvers/subscription_resolver.rb#L25), [here](https://github.com/getlago/lago-api/blob/c674e09fe46ff4023cd1187481d347341af0af00/app/controllers/api/v1/subscriptions/entitlements_controller.rb#L70) or [here](https://github.com/getlago/lago-api/blob/c674e09fe46ff4023cd1187481d347341af0af00/app/controllers/api/v1/subscriptions/base_controller.rb#L15). And yes, it needs some refactoring.

vincent-pochet (approved):

PR #4533, opened 2025-10-23
Feat/e-invoices changes for Billing Entities

mariohd:
DB Changes necessary for EInvoicing initial release.

It adds 1 column `einvoicing#Boolean!` on Billing Entities table.

This also have all necessary changes for BillingEntities because it comes from front/api.


app/services/billing_entities/update_service.rb
```diff
@@ -74,6 +75,7 @@ def call
         end
 
         handle_invoice_custom_sections if params.key?(:invoice_custom_section_ids) || params.key?(:invoice_custom_section_codes)
+        handle_einvoincing_setting if params.key?(:country) && billing_entity.einvoicing
```
julienbourdeau: I'm not a big how this is handled. First, we assign something and 50+ lines later we override the value. I'd prefer if the correct value was set in the assignment.

**More importantly,** since we have validation for this, shouldn't we let the service handle the validation via the ActiveRecord::InvalidRecord exception instead of assigning another value silently? Like the create service?

db/migrate/20251024130659_add_e_invoincing_to_billing_entities.rb
julienbourdeau, line 1: Not too important but maybe we can rename this file to today's date. (Make sure to rollback your env before)

julienbourdeau (approved):

PR #4535, opened 2025-10-24
chore(queries): Refactor query filter

julienbourdeau:
As I'm [adding a filter to the EventQuery](https://github.com/getlago/lago-api/pull/4531), I wanted to describe a new filter validation. I realized there was a few things I didn't like.

Filters are nested under a `filter` key, to accomodate `search_terms` validation.

We don't do anything with search terms validation and its treated a little differently. It's an argument to BaseQuery, just like filters.

This means we also returns errors nested under `filters` in the API response, which is not standard. Filters are an internal thing and should be exposed in the API. This will help with the [new error format](https://github.com/getlago/lago-api/pull/4251) (coming on day, but not urgent).

Example:

```http
GET /v1/customers?countries[]=INVALID
```

Before
```rb
{
    status: 422,
    error: "Unprocessable Entity",
    code: "validation_errors",
    error_details: {
        filters: {
            countries: {
                "0": [
                    "must be one of: AD, ..."
                ]
            }
        }
    }
}
```

After
```rb
{
    status: 422,
    error: "Unprocessable Entity",
    code: "validation_errors",
    error_details: {
        countries: {
            "0": [
                "must be one of: AD, ..."
            ]
        }
    }
}

```

### DIFF
--- a/spec/requests/api/v1/customers_controller_spec.rb
+++ b/spec/requests/api/v1/customers_controller_spec.rb
@@ -469,7 +469,7 @@ RSpec.describe Api::V1::CustomersController, type: :request do
           expect(json).to match({
             code: "validation_errors",
             error: "Unprocessable Entity",
-            error_details: {filters: {customer_type: ["must be nil when has_customer_type is false"]}},
+            error_details: {customer_type: ["must be nil when has_customer_type is false"]},
             status: 422
           })
         end
@@ -549,7 +549,7 @@ RSpec.describe Api::V1::CustomersController, type: :request do
           expect(json).to match({
             code: "validation_errors",
             error: "Unprocessable Entity",
-            error_details: {filters: {countries: {"0": [/must be one of: AD, .*XK$/]}}},
+            error_details: {countries: {"0": [/must be one of: AD, .*XK$/]}},
             status: 422
           })
         end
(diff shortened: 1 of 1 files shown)


app/contracts/queries/billable_metrics_query_filters_contract.rb
```diff
@@ -3,12 +3,8 @@
 module Queries
   class BillableMetricsQueryFiltersContract < Dry::Validation::Contract
     params do
-      required(:filters).hash do
-        optional(:recurring).filled(:bool)
-        optional(:aggregation_types).array(:string, included_in?: %w[max_agg count_agg])
-      end
-
-      optional(:search_term).maybe(:string)
```
julienbourdeau, line 11: On GQL, it's typed as a string.
On controller, it's not even exposed

Image

app/contracts/queries/customers_query_filters_contract.rb
```diff
@@ -3,34 +3,30 @@
...
-        optional(:states).array(:string)
-        optional(:zipcodes).array(:string)
-        optional(:currencies).array(:string, included_in?: Customer.currency_list)
-        optional(:has_tax_identification_number).value(:"coercible.string", included_in?: %w[true false])
-        optional(:metadata).value(:hash)
-        optional(:has_customer_type).value(:"coercible.string", included_in?: %w[true false])
-        optional(:customer_type).value(:string, included_in?: Customer::CUSTOMER_TYPES.values)
-      end
-
-      optional(:search_term).maybe(:string)
```
julienbourdeau, line 19: On GQL, it's typed as a string.
On controller it's a GET endpoint so strong params will only return a string

PR #4536, opened 2025-10-24
feat(estimate): Allow event estimate with Clickhouse

vincent-pochet:
## Context

## Description

This PR is the last step of the feature. It removes the need for a "fake" persisted event (and so the transaction opened) on the events table to estimate the event.
For estimate, the logic is turned from a 
`result = amount_including_fake_persisted_event - amount_without_event_value`
to
`result = amount_from_already_persisted_events + amount_plus_event_value`


app/services/billable_metrics/aggregations/base_service.rb
```diff
@@ -88,10 +88,10 @@ def compute_grouped_by_precise_total_amount_cents(options: {})
         raise NotImplementedError
       end
 
-      def per_event_aggregation(exclude_event: false, grouped_by_values: nil)
+      def per_event_aggregation(exclude_event: false, include_event_value: false, grouped_by_values: nil)
```
rsempe, line 96:
```suggestion
      # NOTE: When include_event_value: true, the current event (not yet persisted) will be included in aggregation. Used only for estimate_fees.
      def per_event_aggregation(exclude_event: false, include_event_value: false, grouped_by_values: nil)
```

rsempe: In what situation can we have `exclude_event: true` and `include_event_value: true`?
Maybe putting a note as well as the interface contract may not seem very explicit

app/services/billable_metrics/aggregations/count_service.rb
```diff
@@ -52,8 +52,10 @@ def running_total(options, aggregation:)
         (1..aggregation).to_a
       end
 
-      def compute_per_event_aggregation(exclude_event:)
-        (0...event_store.count).map { |_| 1 }
+      def compute_per_event_aggregation(exclude_event:, include_event_value:)
+        values = (0...event_store.count).map { |_| 1 }
+        values += [1] if include_event_value
```
groyoh: There's no reason to create a new array here:

```suggestion
        values << 1 if include_event_value
```

app/services/billable_metrics/aggregations/unique_count_service.rb
```diff
@@ -120,8 +120,10 @@ def running_total(options, aggregation:)
         (1..aggregation).to_a
       end
 
-      def compute_per_event_aggregation(exclude_event:)
-        (0...event_store.events_values(force_from: true).count).map { |_| 1 }
+      def compute_per_event_aggregation(exclude_event:, include_event_value:)
+        count = event_store.events_values(force_from: true).count
```
groyoh, line 124: Shouldn't we check values uniqueness here ?

app/services/fees/create_pay_in_advance_service.rb
```diff
@@ -8,7 +8,12 @@ def initialize(charge:, event:, billing_at: nil, estimate: false)
       @charge = charge
       @event = Events::CommonFactory.new_instance(source: event)
       @billing_at = billing_at || @event.timestamp
-      @estimate = estimate
+
+      @estimate = if @event.persisted
+        estimate
+      else
+        true
+      end
```
groyoh: I feel like it would be must more explicit and easier to understand if we had:

```suggestion
      raise ArgumentError, "estimate must be true if event if not persisted" if !event.persisted? && !estimate
```

And then pass `estimate: true` in the caller.

app/services/fees/estimate_pay_in_advance_service.rb
```diff
@@ -43,16 +42,22 @@ def call
...
         external_subscription_id: subscriptions.first&.external_id,
         properties: event_params[:properties] || {},
         transaction_id: SecureRandom.uuid,
         timestamp: Time.current,
-        precise_total_amount_cents: event_params[:precise_total_amount_cents] || 0
+        precise_total_amount_cents: event_params[:precise_total_amount_cents] || 0,
+        persisted: false
       )
+
+      expression_result = Events::CalculateExpressionService.call(organization:, event: @event)
```
groyoh, line 57: Why is this needed now ?

app/services/billable_metrics/aggregations/base_service.rb
vincent-pochet: @rsempe There is no scenario were both attributes are set to true. I'm adding some extra comments to clarify what the fields are doing

groyoh (approved):
That's a tough one 😅

app/services/fees/estimate_pay_in_advance_service.rb
vincent-pochet: It's not directly part of the feature, but it was missing on this endpoint. I'll open a new PR to add it separately.

app/services/billable_metrics/aggregations/unique_count_service.rb
vincent-pochet: @groyoh In theory yes we should but the current logic is already false and I was thinking at fixing it outside of the current scope.

That said when billed in advance, this logic only applies for percentage charge model with a min/max amount per transaction. Clearly this kind of setup does not make real sense as percentage charge model is usually applied on sum aggregation.

PR #4538, opened 2025-10-24
fix(job): Fix `activejob-uniqueness` retry on error

groyoh:
## Context

https://github.com/getlago/lago-api/commit/91cdb1742a18ff1c69683da2835150032966a723 introduced an issue. It increased the `retry_count` to `3`. Redlock retry logic actually behaves the same whether there's an error (network, Redis error, etc.) or whether it fails to lock because the resource is already locked. So whenever we try to enqueue a job for which there's already a job locked (enqueued or running), we’ll try 3 times to enqueue the job causing a ~0.5s delay. 

## Description

This fixes it by introducing a patch to the `Redlock::Client`.

I'll try to push a fix upstream to https://github.com/leandromoreira/redlock-rb/ when I have some time.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ ruby "3.4.7"
 # Core
 gem "aasm"
 gem "activejob-uniqueness", require: "active_job/uniqueness/sidekiq_patch"
+gem "redlock", "~> 2.0.6" # Used through `activejob-uniqueness`. It's pinned to 2.0.x because we patched the library to fix a bug.
 gem "active_storage_validations"
 gem "bootsnap", require: false
 gem "clockwork", require: false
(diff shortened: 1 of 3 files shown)


vincent-pochet (approved):

lib/redlock/client_patch.rb
```diff
@@ -0,0 +1,40 @@
...
+        super(resource, ttl, options, &block)
+      end
+    end
+
+    private
+
+    def with_retry_on_error(&block)
+      tries = Redlock::Client::DEFAULT_RETRY_COUNT + 1
+      error = nil
+      tries.times do |attempt_number|
```
julienbourdeau, line 25: Is `DEFAULT_RETRY_COUNT` just a constant int? I'm don't get why we're adding 1 here 🤔

groyoh: Attempts vs retry. If we retry 3 times, we want to attempt 4 times.

groyoh: I guess it could be rewritten as `(1..tries).each` and avoid the `- 1` below though.

PR #4540, opened 2025-10-24
feat(fixed-charges): Add pay in advance and prorated with charge model validations to fixed charges

ancorcruz:
## Roadmap Task

 ## Context

## Description

Adds proration and pay in advance validation to fixed charge model.

These are the allowed combinations:
**Standard:**
  pay in arrears + full 
  pay in arrears + prorated 
  pay in advance + full 
  pay in advance + prorated 

**Volume:**
  pay in arrears + full 
  pay in arrears + prorated 

**Graduated:**
  pay in arrears + full 
  pay in arrears + prorated 
  pay in advance + full

### DIFF
--- a/app/models/fixed_charge.rb
+++ b/app/models/fixed_charge.rb
@@ -68,7 +68,7 @@ class FixedCharge < ApplicationRecord
       errors.add(:prorated, :invalid_charge_model)
     end
   end
-  
+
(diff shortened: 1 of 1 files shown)


annvelents (approved):
🥇 Nice!

PR #4542, opened 2025-10-27
feat(estimate): Prepare Clickhouse event estimate

vincent-pochet:
## Context

## Description

Since the change requires a lot of file update, it will be split in multiple parts.
This first part:
- Adds a new `persisted` attribute to the `Events::Common` structure, with a default value to true. It will allow us to mark define an event as "temporary"
- Define the `==` method on the `BaseResult` class, to allow comparison of result states and values in specs
- Extract the charge model class initialization for in advance billing from `Charges::ApplyPayInAdvanceChargeModelService` to `ChargeModels::Factory`

### DIFF
--- a/app/services/base_result.rb
+++ b/app/services/base_result.rb
@@ -3,7 +3,23 @@
 class BaseResult
   include Result
 
+  class_attribute :attributes, default: [] # rubocop:disable ThreadSafety/ClassAndModuleAttributes
+
   def self.[](*attributes)
-    Class.new(BaseResult) { attr_accessor(*attributes) }
+    Class.new(BaseResult) do
+      attr_accessor(*attributes)
+
+      self.attributes = attributes
+    end
+  end
+
+  def ==(other)
+    return false unless other.class == self.class
+    return false unless failure? == other.failure?
+    return false unless other.error == error
+
+    self.class.attributes.all? do |attribute|
+      send(attribute) == other.send(attribute)
+    end
   end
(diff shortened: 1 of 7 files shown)


app/services/base_result.rb
```diff
@@ -3,7 +3,22 @@
...
   include Result
 
+  class_attribute :exposed_attributes, default: [] # rubocop:disable ThreadSafety/ClassAndModuleAttributes
+
   def self.[](*attributes)
-    Class.new(BaseResult) { attr_accessor(*attributes) }
+    Class.new(BaseResult) do
+      attr_accessor(*attributes)
+
+      self.exposed_attributes = attributes
```
julienbourdeau: `public_attributes` maybe? Any reason to not use `attributes` like on a model? I'm not a big fan of `exposed_` but this is no big deal.

app/services/base_result.rb
```diff
@@ -3,7 +3,22 @@
...
+  end
+
+  def ==(other)
+    return false unless other.class == self.class
+    return false unless failure? == other.failure?
+
+    self.class.attributes.all? do |attribute|
+      send(attribute) == other.send(attribute)
+    end
   end
```
julienbourdeau, line 24: Actually, I think we're not comparing that both failed results failed with the same error.

vincent-pochet: Hooo, good catch!

PR #4544, opened 2025-10-27
chore(services): Make BaseQuery independant from BaseService

julienbourdeau:
Other than the call class method helper and initializing a result, there is no reason to extend BaseService.
I think this will help make Query object evolve on their own.

### DIFF
--- a/app/queries/base_query.rb
+++ b/app/queries/base_query.rb
@@ -1,12 +1,20 @@
 # frozen_string_literal: true
 
-class BaseQuery < BaseService
+class BaseQuery
   # nil values force Kaminari to apply its default values for page and limit.
   DEFAULT_PAGINATION_PARAMS = {page: nil, limit: nil}
   DEFAULT_ORDER = {created_at: :desc}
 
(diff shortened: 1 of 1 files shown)


app/queries/base_query.rb
```diff
@@ -1,26 +1,43 @@
...
   def initialize(organization:, pagination: DEFAULT_PAGINATION_PARAMS, filters: {}, search_term: nil, order: nil)
     @organization = organization
     @pagination_params = pagination
     @filters = self.class::Filters.new(**(filters || {}))
     @search_term = search_term.to_s.strip
     @order = order
 
-    super
+    @result = self.class::Result.new
+    @source = CurrentContext&.source
```
julienbourdeau, line 27: I kept the source but I think it's useless.

vincent-pochet (commented):
I'm not against this change, but I think it goes a bit deeper that just removing the inheritance.
For example, `BaseService` class also implements the middleware pattern and uses it to plugs the `Middlewares::LogTracerMiddleware`

PR #4545, opened 2025-10-27
feat(filters): Add filter validation for EventsQuery

julienbourdeau:
I have added a new filter on the `GET /v1/events` endpoint. This PR adds the filter validation.

See https://github.com/getlago/lago-api/pull/4531

### DIFF
--- a/app/contracts/queries/events_query_filters_contract.rb
+++ b/app/contracts/queries/events_query_filters_contract.rb
@@ -8,6 +8,7 @@ module Queries
       optional(:timestamp_from_started_at).value(:"coercible.string", included_in?: %w[true false])
       optional(:timestamp_from)
       optional(:timestamp_to)
+      optional(:enriched).value(:bool)
     end
 
     rule("timestamp_from_started_at", "timestamp_from") do


Gemfile.lock
```diff
@@ -1025,6 +1025,7 @@ PLATFORMS
   arm64-darwin-22
   arm64-darwin-23
   arm64-darwin-24
+  arm64-darwin-25
```
julienbourdeau, line 1034: This is because I started using node 25 locally

app/queries/events_query.rb
```diff
@@ -30,6 +32,10 @@ def call
 
   private
 
+  def filters_contract
+    @filters_contract ||= Queries::EventsQueryFiltersContract.new
+  end
```
julienbourdeau, line 44: I think we don't need the memoization here but I have done it like in others query objects. I'd love to refactor by moving this to a constant just like Filters or Result.

app/contracts/queries/events_query_filters_contract.rb
```diff
@@ -0,0 +1,25 @@
...
+      optional(:code).maybe(:string)
+      optional(:external_subscription_id).maybe(:string)
+      optional(:timestamp_from_started_at).value(:"coercible.string", included_in?: %w[true false])
+      optional(:timestamp_from)
+      optional(:timestamp_to)
+    end
+
+    rule("timestamp_from_started_at", "timestamp_from") do
+      if ActiveModel::Type::Boolean.new.cast(values["timestamp_from_started_at"]) && values["timestamp_from"].present?
+        key(:timestamp_from).failure("cannot be used with timestamp_from_started_at")
```
julienbourdeau, line 16: Notice that contracts are the only place where we don't use error codes. It's already like this in others. I think we should address this later.

PR #4546, opened 2025-10-27
feat(multiple-payment-methods): add support for payment method on wallets

lovrocolic:
## Context

## Description

This PR adds GQL support for attaching payment method to the wallet. With this setup, it will be possible to have paid credits invoice connected to different payment method than subscription invoice.

### DIFF
--- a/app/services/wallets/create_service.rb
+++ b/app/services/wallets/create_service.rb
@@ -42,7 +42,7 @@ module Wallets
 
       if params.key?(:payment_method)
         attributes[:payment_method_type] = params[:payment_method][:payment_method_type] if params[:payment_method].key?(:payment_method_type)
-        attributes[:payment_method_id] = result.payment_method.id if params[:payment_method].key?(:payment_method_id)
+        attributes[:payment_method_id] = params[:payment_method][:payment_method_id] if params[:payment_method].key?(:payment_method_id)
       end
 
       wallet = Wallet.new(attributes)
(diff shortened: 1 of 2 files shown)


app/services/wallets/create_service.rb
```diff
@@ -127,5 +133,12 @@ def billable_metrics
         BillableMetric.where(id: billable_metric_identifiers, organization_id: params[:organization_id])
       end
     end
+
+    def payment_method
+      return @payment_method if defined? @payment_method
+      return nil if params[:payment_method].blank? || params[:payment_method][:payment_method_id].blank?
+
+      @payment_method = PaymentMethod.find_by(id: params[:payment_method][:payment_method_id])
```
vincent-pochet: This lookup should be scoped to the organization or any id belonging to it

app/services/wallets/update_service.rb
```diff
@@ -134,5 +145,12 @@ def billable_metrics
         BillableMetric.where(id: billable_metric_identifiers, organization_id: wallet.organization_id)
       end
     end
+
+    def payment_method
+      return @payment_method if defined? @payment_method
+      return nil if params[:payment_method].blank? || params[:payment_method][:payment_method_id].blank?
+
+      @payment_method = PaymentMethod.find_by(id: params[:payment_method][:payment_method_id])
```
vincent-pochet: Same comment regarding the scope

app/graphql/types/wallets/create_input.rb
```diff
@@ -22,6 +22,8 @@ class CreateInput < Types::BaseInputObject
       argument :transaction_name, String, required: false
 
       argument :applies_to, Types::Wallets::AppliesToInput, required: false
+
+      argument :payment_method, Types::PaymentMethods::ReferenceInput, required: false
```
vincent-pochet, line 27: Could you add the specs for the two GraphQL types as well? 🙏

PR #4549, opened 2025-10-27
feat(event_store): Use event.timestamp as upper bound when aggregating for Pay In Advance charge

julienbourdeau:
With Pay in advance events, we currently count all events in the period to determine the price. We should only count events stored PRIOR to this specific event timestamp.

This is relying on an existing feature of the store and simply pass `max_timetamp` to the query builder.

I added an extra events in the scenario tests. Most of them fail without the fix but in some cases it's still green, probably due to the aggregation type.

These tests don't run in Clickhouse because our scenario setup isn't running the real processor and I did not manage to make it work.

### DIFF
--- a/spec/scenarios/pay_in_advance_charges_spec.rb
+++ b/spec/scenarios/pay_in_advance_charges_spec.rb
@@ -126,7 +126,7 @@ describe "Pay in advance charges Scenarios", transaction: false do
 
       # Create events with timestamp AFTER the current event to invoice, to ensure `event.timestamp` is used as upper bound.
       # NOTE: This does not seem to matter
-      create(:event, code: billable_metric.code, organization:, subscription:, timestamp: feb15, properties: {unique_id: "id_1"})
+      create(:event, code: billable_metric.code, organization:, subscription:, timestamp: feb15 + 10.days, properties: {unique_id: "id_1"})
 
       travel_to(feb15) do
         expect do


spec/scenarios/pay_in_advance_charges_spec.rb
```diff
@@ -37,9 +37,13 @@
 
       subscription = customer.subscriptions.first
 
-      ### 15 february: Send an event.
+      ### 15 February: Send an event.
       feb15 = DateTime.new(2023, 2, 15)
 
+      # Create events with timestamp AFTER the current event to invoice, to ensure `event.timestamp` is used as upper bound.
+      # NOTE: This does not seem to matter
+      create(:event, code: billable_metric.code, organization:, subscription:, timestamp: feb15 + 4.days)
```
groyoh, line 45: Is there a reason you create from a factory and not the helper ?

spec/scenarios/pay_in_advance_charges_spec.rb
```diff
@@ -120,6 +124,10 @@
       ### 15 february: Send an event.
       feb15 = DateTime.new(2023, 2, 15)
 
+      # Create events with timestamp AFTER the current event to invoice, to ensure `event.timestamp` is used as upper bound.
+      # NOTE: This does not seem to matter
+      create(:event, code: billable_metric.code, organization:, subscription:, timestamp: feb15, properties: {unique_id: "id_1"})
```
groyoh: Don't you need timestamp to be later ?

spec/scenarios/pay_in_advance_charges_spec.rb
julienbourdeau: Yes, to not trigger the jobs that bill for the events otherwise it's messing with the amount for some aggregation. Like if you have 1 event free and 1 paid, suddenly, the event free was the event in the future.

This is how it happens in production but I didn't want to tests to change, just to ensure that this future events was not taken into account.

spec/scenarios/pay_in_advance_charges_spec.rb
julienbourdeau: I updated for predictability but it makes no difference with this aggregation I think 🙏

PR #4550, opened 2025-10-27
feat(fixed-charges): Fix fixed charge event creation on plan override

ancorcruz:
Emit fixed charge events for the next billing period when a subscription plan is overridden (the first time) by default, when not applied units immediately or when params are missing.

### DIFF
--- a/app/services/fixed_charges/emit_events_for_active_subscriptions_service.rb
+++ b/app/services/fixed_charges/emit_events_for_active_subscriptions_service.rb
@@ -31,7 +31,9 @@ module FixedCharges
       # This handles cases like plan overrides where the subscription hasn't been updated yet
       # otherwise, emit events for all active subscriptions on the plan
       if subscription
-        [subscription]
+        # Only emit events for active subscriptions, even when explicitly provided
+        # Pending subscriptions will have events created when they activate
+        subscription.active? ? [subscription] : []
       else
         fixed_charge.plan.subscriptions.active
       end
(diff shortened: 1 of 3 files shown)


vincent-pochet (approved):

app/services/fixed_charges/override_service.rb
```diff
@@ -35,7 +35,7 @@ def call
         FixedCharges::EmitEventsForActiveSubscriptionsService.call!(
           fixed_charge: new_fixed_charge,
           subscription:,
-          apply_units_immediately: true
+          apply_units_immediately: !!params[:apply_units_immediately]
```
annvelents, line 38: I am thinking:
will it then emit events when creating a subscription with overrides from the beginning and not sending :apply_units_imediately at all?

ancorcruz: good point, I'll ensure both scenarios (new subs and update sub) are covered by tests. Thanks for raising it.

ancorcruz: @annvelents I think all cases are covered and by the tests all looks good. So far we ensure the following.

if subscription is not active it does not create events, this applies for each scenario.

if subscription is active:

- On subscription creation with plan override,
emit events for each override fixed charge with timestamp subs started_at

- On subscription update with plan override,
emit events with timestamp now for each override fixed charge if apply immediately.
emit events with next billing period for each override fixed charge if not apply immediately.

- On subscription with an already overridden plan, updates the plan and only emit events on fixed charge units changes.

PR #4562, opened 2025-10-29
feat(api): Introduce /1/events_enriched endpoint

julienbourdeau:
The Event list endpoint was designed to help debug sending events. On PG it doesn't really make a difference, but on Clickhouse, this explains why we return events_raw.
It means you can see events in the API but they might not be taken into account in billing.

This new filter let's you retrieve "billing events", ie events in `enriched_raw`. 

**On PG**
* `/v1/events` returns `events` from the table.

**On ClickHouse**
* `/v1/events` returns the events in `events_raw` table.
* `/v1/events_enriched` returns the events in `events_enriched` table.

> [!TIP]
> This PR also introduce a new custom `X-Lago-Endpoint-Status` header to signal if an endpoint is `beta`. I'll add depecated too in another PR.

There should be some filter validation on this query object, right? [Yes, it's coming here in this PR](https://github.com/getlago/lago-api/pull/4545).

### DIFF
--- a/app/controllers/api/v1/events_controller.rb
+++ b/app/controllers/api/v1/events_controller.rb
@@ -5,6 +5,8 @@ module Api
     class EventsController < Api::BaseController
       skip_audit_logs!
 
+      before_action :ensure_organization_uses_clickhouse, only: [:index_enriched]
+
       ACTIONS_WITH_CACHED_API_KEY = %i[create batch estimate_instant_fees batch_estimate_instant_fees].freeze
 
       def create
@@ -91,10 +93,6 @@ module Api
 
       def index_enriched
         set_beta_header!
-
-        unless current_organization.clickhouse_events_store?
-          return forbidden_error(code: "endpoint_not_available")
-        end
 
         result = EventsQuery.call(
           organization: current_organization,
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -19,6 +19,10 @@ module Api
 
     attr_reader :current_api_key, :current_organization
 
+    def ensure_organization_uses_clickhouse
+      forbidden_error(code: "endpoint_not_available") unless current_organization.clickhouse_events_store?
+    end
+
     def authenticate
       return unauthorized_error unless auth_token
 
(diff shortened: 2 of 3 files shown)


Gemfile.lock
```diff
@@ -1031,6 +1031,7 @@ PLATFORMS
   arm64-darwin-22
   arm64-darwin-23
   arm64-darwin-24
+  arm64-darwin-25
```
julienbourdeau, line 1034: This is because I started using Node 25 locally

app/queries/events_query.rb
```diff
@@ -1,27 +1,34 @@
...
+      events.order(ingested_at: :desc)
+    elsif ch_event_enriched?
+      events.order(enriched_at: :desc)
+    end
 
     events = with_code(events) if filters.code
     events = with_external_subscription_id(events) if filters.external_subscription_id
     events = with_timestamp_range(events)
 
+    result.event_model = event_model.to_s
```
lovrocolic, line 31: Do we need to attach it?

app/controllers/api/v1/events_controller.rb
```diff
@@ -89,6 +89,36 @@ def index
         end
       end
 
+      def index_enriched
+        set_beta_header!
+
+        unless current_organization.clickhouse_events_store?
```
lovrocolic: I would personally put this condition first, before setting a header

app/queries/events_query.rb
lovrocolic: As I can see, attached for tests 😃

app/controllers/api/v1/events_controller.rb
julienbourdeau: I'd prefer to keep the beta header first for us. I hoping that this makes it a little clearer than if it's in the middle of the method function.
But this makes me think that this piece should be a method in BaseController like `ensure_organization_uses_clickhouse!` 🤔

app/queries/events_query.rb
julienbourdeau: Yes and also, I'm afraid that without this, people start doing `result.events.first.class.name` to know what it is, which would trigger a query.

In a next PR, I'll dig into events raw because I believe there is an N+1 🙈

app/controllers/api/v1/events_controller.rb
julienbourdeau: moved it to a before_action

PR #4563, opened 2025-10-30
test(event): Test aggregation scenarios on both Postgres and Clickhouse

groyoh:
## Context

We have a few scenario tests ensuring the behavior of the different aggregations but:

1. We duplicate the code for `unique_count` aggregation to test Postgres and Clickhouse
2. We don't test the other aggregations with Clickhouse

## Description

This updates the `create_event` scenario helper to check whether the organization has the clickhouse store enabled and, if it is the case, create the `event_enriched` record and schedule the pay-in-advance jobs.


vincent-pochet (approved):
Great one! It will help a lot improving the specs and the trust in clickhouse store!!

spec/support/scenarios_helper.rb
```diff
@@ -255,27 +255,17 @@ def ingest_event(subscription, billable_metric, amount)
   end
 
   def create_event(params, **kwargs)
+    params[:transaction_id] ||= SecureRandom.uuid
+    subscription = fetch_subscription(params[:external_subscription_id])
```
julienbourdeau: I don't understand what this is for? to ensure that the sub exists?

groyoh: To retrieve the `organization` to know whether it's a clickhouse-enabled organization. We might be able to retrieve it from the response though, I'll check.

groyoh: I feel dumb, we already have access to the organization here 🤦

PR #4568, opened 2025-10-30
Feat/e invoices payments

mariohd:
This PR is part of the e-invoicing set of pull requests.

It has the necessary code to generate the files for Payments.

Please review it looking commit by commit, it's going to be easier to follow (trust me, I promise).

### DIFF
--- a/spec/graphql/mutations/payment_receipts/download_xml_spec.rb
+++ b/spec/graphql/mutations/payment_receipts/download_xml_spec.rb
@@ -30,23 +30,21 @@ RSpec.describe Mutations::PaymentReceipts::DownloadXml do
   it_behaves_like "requires permission", "invoices:view"
 
   it "generates the XML for the given payment receipt" do
-    freeze_time do
-      result = execute_graphql(
-        current_user: membership.user,
-        current_organization: organization,
-        permissions: required_permission,
-        query: mutation,
-        variables: {
-          input: {id: payment_receipt.id}
-        }
-      )
+    result = execute_graphql(
+      current_user: membership.user,
+      current_organization: organization,
+      permissions: required_permission,
+      query: mutation,
+      variables: {
+        input: {id: payment_receipt.id}
+      }
+    )
 
-      result_data = result["data"]["downloadXmlPaymentReceipt"]
+    result_data = result["data"]["downloadXmlPaymentReceipt"]
 
-      aggregate_failures do
-        expect(result_data["id"]).to be_present
-        expect(result_data["xmlUrl"]).to be_present
-      end
+    aggregate_failures do
+      expect(result_data["id"]).to be_present
+      expect(result_data["xmlUrl"]).to be_present
     end
   end
 end
--- a/spec/jobs/payment_receipts/notify_job_spec.rb
+++ b/spec/jobs/payment_receipts/notify_job_spec.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe PaymentReceipts::NotifyJob do
+  let(:payment_receipt) { create(:payment_receipt) }
+
+  it "sends the email" do
+    expect { described_class.perform_now(payment_receipt:) }
+      .to have_enqueued_mail(PaymentReceiptMailer, :created).with(params: {payment_receipt:}, args: [])
+  end
+end
--- a/spec/models/payment_receipt_spec.rb
+++ b/spec/models/payment_receipt_spec.rb
@@ -10,6 +10,7 @@ RSpec.describe PaymentReceipt do
     expect(subject).to belong_to(:organization)
     expect(subject).to belong_to(:billing_entity)
     expect(subject).to have_one_attached(:file)
+    expect(subject).to have_one_attached(:xml_file)
   end
 
   describe "#file_url" do
--- a/spec/models/payment_spec.rb
+++ b/spec/models/payment_spec.rb
@@ -29,6 +29,24 @@ RSpec.describe Payment do
       .backed_by_column_of_type(:enum)
   end
 
+  describe "attributes" do
+    describe "amount_currency" do
+      before { payment.update(amount_currency: "BRL") }
+
+      it "aliases to currency" do
+        expect(payment.currency).to eq("BRL")
+      end
+    end
+  end
+
+  describe "delegations" do
+    describe "billing_entity" do
+      it "delegates to customer" do
+        expect(payment.billing_entity).to be(payment.customer.billing_entity)
+      end
+    end
+  end
+
   describe "validations" do
     let(:errors) { payment.errors }
 
@@ -411,6 +429,36 @@ RSpec.describe Payment do
       it "returns the payment method type" do
         expect(payment_method).to eq("Bank Transfer")
       end
+    end
+  end
+
+  describe "#card_last_digits" do
+    subject { payment.card_last_digits }
+
+    context "when provider_payment_method_data is empty" do
+      let(:payment) { build(:payment, provider_payment_method_data: {}) }
+
+      it { is_expected.to be_nil }
+    end
+
+    context "when provider_payment_method_data type is not card" do
+      let(:payment) do
+        build(:payment, provider_payment_method_data: {"type" => "link"})
+      end
+
+      it { is_expected.to be_nil }
+    end
+
+    context "when provider_payment_method_data type is card" do
+      let(:payment) do
+        build(:payment, provider_payment_method_data: {
+          "type" => "card",
+          "brand" => "visa",
+          "last4" => "1234"
+        })
+      end
+
+      it { is_expected.to eq("**** 1234") }
     end
   end
 


app/models/payment.rb
groyoh, line 1: Any chance you could add test for the changes in this model ?

app/serializers/e_invoices/payments/common.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module EInvoices
+  module Payments
+    module Common
+      def credits_and_payments(&block)
```
groyoh, line 12: Does this really need to be in the module ? It seems to me this is only used in one class ?

spec/serializers/e_invoices/payments/factur_x/builder_spec.rb
groyoh, line 1: This file is quite hard to review to be honest. Plus I don't think it makes sense to split the tests that much as each `it` will have to instantiate all the factories.

I feel like this might benefit from a single test with snapshot testing. WDYT ?

spec/services/e_invoices/payments/factur_x/create_service_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let(:payment) { create(:payment, organization:) }
+  let(:xml_builder_double) { instance_double(Nokogiri::XML::Builder, to_xml: xml_content) }
+  let(:xml_content) { "<xml>content</xml>" }
+
+  around { |test| lago_premium!(&test) }
+
+  describe "#call" do
+    context "when payment exists" do
+      it "builds the XML" do
+        allow(Nokogiri::XML::Builder).to receive(:new).with(encoding: "UTF-8")
```
groyoh, line 16: Do we need to mock here ? I believe we can call the build and simply assess some of the expected values.

app/serializers/e_invoices/base_serializer.rb
```diff
@@ -38,6 +38,13 @@ class BaseSerializer
     # https://docs.peppol.eu/pracc/catalogue/1.0/codelist/UNECERec20/
     UNIT_CODE = "C62"
 
+    # Response codes for Payments
+    # There are no strict codes for this
+    ACKNOWLEDGEMENT = "AC"
+    PENDING = "PE"
+    REJECTED = "RE"
+    PAID = "PD"
```
groyoh, line 46: Where are these used ?

spec/serializers/e_invoices/ubl/receiver_party_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+
+      it "expects to have zipcode" do
+        expect(subject).to contains_xml_node("#{xpath}/cbc:PostalZone").with_value(customer.zipcode)
+      end
+
+      it "expects to have country" do
+        expect(subject).to contains_xml_node("#{xpath}/cac:Country/cbc:IdentificationCode")
+          .with_value(customer.country)
+      end
```
groyoh, line 59: May I suggest to rewrite those tests like this :

```suggestion
    it "serializes the receiver party" do
      expect(subject.to_s).to eq <<~XML
        <?xml version="1.0" encoding="UTF-8"?>
        <TestRoot xmlns:cac="urn:oasis:names:specification:ubl:schema:xsd:CommonAggregateComponents-2" xmlns:cbc="urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2">
          <!--Receiver Party-->
          <cac:ReceiverParty>
            <cac:PartyName>
              <cbc:Name>Customer Jr</cbc:Name>
            </cac:PartyName>
            <cac:PostalAddress>
              <cbc:StreetName>Somewhere</cbc:StreetName>
              <cbc:AdditionalStreetName>Turn around, wrong way</cbc:AdditionalStreetName>
              <cbc:CityName>Some city</cbc:CityName>
              <cbc:PostalZone>09876 CE</cbc:PostalZone>
              <cac:Country>
                <cbc:IdentificationCode>BR</cbc:IdentificationCode>
              </cac:Country>
            </cac:PostalAddress>
          </cac:ReceiverParty>
        </TestRoot>
      XML
    end
```

?

I would find it much easier to review the test based on the actual serializer code.

spec/services/e_invoices/payments/ubl/create_service_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let(:payment) { create(:payment, organization:) }
+  let(:xml_builder_double) { instance_double(Nokogiri::XML::Builder, to_xml: xml_content) }
+  let(:xml_content) { "<xml>content</xml>" }
+
+  around { |test| lago_premium!(&test) }
+
+  describe "#call" do
+    context "when payment exists" do
+      it "builds the XML" do
+        allow(Nokogiri::XML::Builder).to receive(:new).with(encoding: "UTF-8")
```
groyoh, line 16: Same feedback as before, do we actually need to mock here ?

app/models/payment_receipt.rb
```diff
@@ -8,12 +8,19 @@ class PaymentReceipt < ApplicationRecord
...
   has_one_attached :file
+  has_one_attached :xml_file
 
   def file_url
     return if file.blank?
 
     Rails.application.routes.url_helpers.rails_blob_url(file, host: ENV["LAGO_API_URL"])
   end
+
+  def xml_url
```
groyoh, line 19:
```suggestion
  def xml_file_url
```

Maybe there's even a more descriptive name ? `e_invoicing_xml_file` or `ubl_file` ?

app/models/payment_receipt.rb
```diff
@@ -8,12 +8,19 @@ class PaymentReceipt < ApplicationRecord
   delegate :customer, to: :payment
 
   has_one_attached :file
+  has_one_attached :xml_file
```
groyoh, line 11: Could you have a test for this ?

```ruby
    expect(subject).to have_one_attached(:xml_file)
```

spec/services/payment_receipts/generate_xml_service_spec.rb
groyoh, line 1: I would suggest not mocking here but that's mainly a personal preference.

app/jobs/payment_receipts/generate_documents_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      if ActiveModel::Type::Boolean.new.cast(ENV["SIDEKIQ_PDFS"])
+        :pdfs
+      else
+        :low_priority
+      end
+    end
+
+    retry_on LagoHttpClient::HttpError, Errno::ECONNREFUSED, EOFError, wait: :polynomially_longer, attempts: 6
+
+    def perform(payment_receipt:, notify: false)
```
groyoh, line 15: Side-note but I guess this could be a good candidate for https://guides.rubyonrails.org/active_job_basics.html#job-continuations.

app/jobs/payment_receipts/generate_documents_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      PaymentReceipts::GenerateXmlService.call(payment_receipt:).raise_if_error!
+      PaymentReceipts::GeneratePdfService.call(payment_receipt:).raise_if_error!
+
+      PaymentReceipts::NotifyJob.perform_later(payment_receipt:) if notify
```
groyoh, line 19: Do we need a job here ? Can't we schedule the email directly like previously ?

app/graphql/types/mutation_type.rb
```diff
@@ -119,6 +119,7 @@ class MutationType < Types::BaseObject
     field :void_invoice, mutation: Mutations::Invoices::Void
 
     field :download_payment_receipt, mutation: Mutations::PaymentReceipts::Download
+    field :download_xml_payment_receipt, mutation: Mutations::PaymentReceipts::DownloadXml
```
groyoh, line 123: Shouldn't this be

```suggestion
    field :generate_payment_receipt_xml, mutation: Mutations::PaymentReceipts::DownloadXml
```

?

app/graphql/types/payment_receipts/object.rb
```diff
@@ -12,6 +12,7 @@ class Object < Types::BaseObject
       field :number, String, null: false
       field :organization, Types::Organizations::OrganizationType, null: false
       field :payment, Types::Payments::Object, null: false
+      field :xml_url, String, null: true
```
groyoh, line 15:
```suggestion
      field :xml_file_url, String, null: true
```

spec/graphql/mutations/payment_receipts/download_xml_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  before { payment_receipt }
+
+  it_behaves_like "requires current user"
+  it_behaves_like "requires current organization"
+  it_behaves_like "requires permission", "invoices:view"
+
+  it "generates the XML for the given payment receipt" do
+    freeze_time do
```
groyoh: Why do we need this ?

spec/jobs/payment_receipts/notify_job_spec.rb
groyoh, line 1: This is empty ⚠️

app/serializers/v1/payment_receipt_serializer.rb
```diff
@@ -6,6 +6,8 @@ def serialize
       {
         lago_id: model.id,
         number: model.reload.number,
+        file_url: model.file_url,
+        xml_url: model.xml_url,
```
groyoh, line 10:
```suggestion
        xml_file_url: model.xml_url,
```

app/services/payment_receipts/generate_xml_service.rb
```diff
@@ -0,0 +1,67 @@
...
+        xml_file = build_xml_file
+        attach_xml_to_payment_receipt(xml_file)
+        payment_receipt.save!
+      ensure
+        cleanup_tempfiles(xml_file)
+      end
+    end
+
+    def build_xml_file
+      xml_file = Tempfile.new([payment_receipt.number, ".xml"])
```
groyoh, line 38: I think it should be possible use a `StringIO` instead of Tempfile. You wouldn't need to clean file afterward.

app/services/payment_receipts/generate_pdf_service.rb
```diff
@@ -32,27 +32,60 @@ def render_html
...
+      pdf_file.binmode
+      pdf_file.write(pdf_content)
+      pdf_file.flush
+
+      pdf_file
+    end
+
+    def attach_facturx(pdf_file)
+      xml_file = Tempfile.new([payment_receipt.number, ".xml"])
+      xml_file.write(EInvoices::Payments::FacturX::CreateService.call(payment: payment_receipt.payment).xml)
```
groyoh, line 57: We silently ignore errors here ?

spec/serializers/e_invoices/payments/factur_x/builder_spec.rb
mariohd: Yes, it's hard because it has a lot of scenarios to test. 

I added in a way that it tests all the required tag in the XML and it's variations depending on the data.

I believe this format is better than have the snapshot because it tests the different scenarios with tags and data being validated in a single file instead of multiple snapshots.

spec/services/e_invoices/payments/factur_x/create_service_spec.rb
mariohd: I've mocked here because the idea of this test is to be sure that the `CreateService` calls the builder, doesn't matter what produces and the data used. 

If I let the Builder to build a proper XML, I would have to add at least the simplest XML data to build something without breaking the serialize method and this would make this test more complex without a proper reason.

app/serializers/e_invoices/base_serializer.rb
mariohd: The all are not being used right now, but they are possible values to be added as Payment information. They can be used for other formats/countries other FR. eg. DE (I think we also will have Saudi Arabia soon too)

spec/serializers/e_invoices/ubl/receiver_party_spec.rb
mariohd: This is way is also good, but IMO it is more "fragile" because it can break the test just because an extra char (or a missing one). 

And some tags have data variations (I know that can be also tested with string matching), but I feel testing each tag is safer. In the end, it's just a matter of choice.

spec/services/e_invoices/payments/ubl/create_service_spec.rb
mariohd: same idea. The idea here is to test if the CreateService is calling the builder, not to build the XML itself.

app/models/payment_receipt.rb
mariohd: That's a tricky one. 

I don't like the name either, but the `file_url` (the PDF file of the record) is spread all over the system and I wanted to keep simpler for now, keep the same pattern. 

Maybe in the future we can rename this to better names, I totally agree with you.

app/models/payment_receipt.rb
mariohd: Great! I will add this!

app/jobs/payment_receipts/generate_documents_job.rb
mariohd: I totally agree, we could try to implement this in the future.

app/jobs/payment_receipts/generate_documents_job.rb
mariohd: Yes, we could, but if the email fail (for any reason), we would have to call the PDF/XML generate again (they have safe guards to not generate again, but they will be called anyway) to just send the email. 

So in case only the email is broken, have the email job run, not the entire job. That was my idea.

app/graphql/types/mutation_type.rb
mariohd: no, this does not generate the XML (not intended to). The XML is generated by another background process and this only download the XML file. 

I've kept the same pattern as the PDF. We could see this in the future if we can change those names to proper ones, but FE needs to be in the loop as well.

app/graphql/types/payment_receipts/object.rb
mariohd: same idea, keep the same pattern `anything_url`. 

IMO having `XML` is implicit that is a file. I don't like this format too.

I think the best format is to have a single method that takes the file format as a param, but that can be in a future refactor.

spec/jobs/payment_receipts/notify_job_spec.rb
mariohd: WOW, super good notice. 

I will add these tests. Thanks!!

spec/graphql/mutations/payment_receipts/download_xml_spec.rb
mariohd: Good question, I believe it was from the test I've got inspiration. 

I will check if we can remove this.

app/serializers/v1/payment_receipt_serializer.rb
mariohd: keeping the same pattern.

spec/serializers/e_invoices/ubl/receiver_party_spec.rb
groyoh: > This is way is also good, but IMO it is more "fragile" because it can break the test just because an extra char (or a missing one).

But wouldn't we actually want the test to break in such case (whitespace excluded of course) ?

app/services/payment_receipts/generate_pdf_service.rb
mariohd: if the XML fails to build, it will raise the error and stop the PDF creation.

app/services/payment_receipts/generate_xml_service.rb
mariohd: Yes, we could `StringIO` here, I just used the same pattern as the PDF generation.

app/models/payment.rb
mariohd: Sure, I can add.

app/serializers/e_invoices/payments/common.rb
mariohd: The idea to have this in this module it's because other formats/countries may rely on this soon. We will have `Rechnung` (DE) format soon and it will be used there (_if I remember right_)

app/jobs/payment_receipts/generate_documents_job.rb
groyoh: That makes sense but I feel a bit overkill given the job PDF/XML services are idempotents.

spec/serializers/e_invoices/ubl/receiver_party_spec.rb
mariohd: yes, it would break the same way as I did. 
I feel it's easier to test tag by tag and it's variations than having the full XML tag file copied over and over. 

Like I said, In the end it's just a matter of choice.

app/jobs/payment_receipts/generate_documents_job.rb
mariohd: Also other reason that I don't like that format it's because the 2 different "actions" done by the same job. 

Generate the files AND send the email. 
At least to me, if feels better to have each job have it's action well defined. 

A job to generates the files.
A job to send the email.

PR #4569, opened 2025-10-31
perf(cache): Optimize cache by removing null attributes

groyoh:
## Context

We heavily rely on caching for improving the current usage performance. We do so by caching the fees as aggregating the events for fee calculating can be quite costly and impactful on the database resource.

While investigating, we noticed that the cached fees are stored with null attributes which can actually be removed before storing them.

## Description

This removes the null attributes before storing the fees into cache. We explicitly define which attributes are compactable and which are not to avoid losing information. For instance, the `grouped_by` should keep the `nil` values.

We also reduced the `compress_threshold` to ensure that compacted cache entries will still be compressed.

### DIFF
--- a/app/services/subscriptions/charge_cache_middleware.rb
+++ b/app/services/subscriptions/charge_cache_middleware.rb
@@ -31,10 +31,14 @@ module Subscriptions
         fees.map do |fee|
           fee_attributes = fee.attributes
           if (pricing_unit_usage = fee.pricing_unit_usage).present?
-            pricing_unit_usage_attributes = compact_hash(pricing_unit_usage.attributes, COMPACTABLE_PRICING_UNIT_USAGE_ATTRIBUTES)
+            pricing_unit_usage_attributes = if compact_cache?
+              compact_hash(pricing_unit_usage.attributes, COMPACTABLE_PRICING_UNIT_USAGE_ATTRIBUTES)
+            else
+              pricing_unit_usage.attributes
+            end
             fee_attributes["pricing_unit_usage"] = pricing_unit_usage_attributes
           end
-          compact_fee(fee_attributes)
+          compact_cache? ? compact_fee(fee_attributes) : fee_attributes
         end.to_json
       end
 
@@ -115,6 +119,10 @@ module Subscriptions
 
     def compact_hash(object, compactable_keys)
       object.reject { |key, value| compactable_keys.include?(key) && value.nil? }
+    end
+
+    def compact_cache?
+      @compact_cache ||= subscription.organization.feature_flag_enabled?(:null_attributes_charge_cache_optimization)
     end
 
     def cache_expiration
--- a/app/config/feature_flags.yaml
+++ b/app/config/feature_flags.yaml
@@ -2,6 +2,11 @@ multiple_payment_methods:
   description: "Allow multiple payment methods per customer"
   owners: ["@lovrocolic", "@mariohd"]
 
+null_attributes_charge_cache_optimization:
+  description: "Optimize current usage cache by stripping nil attributes from cached fees"
+  owners: ["@groyoh"]
+  backend_only: true
+
 postgres_enriched_events:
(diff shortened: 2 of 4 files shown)


app/services/subscriptions/charge_cache_middleware.rb
```diff
@@ -12,28 +12,90 @@ def initialize(subscription:, charge:, to_datetime:, cache: true)
...
 
-      json = Subscriptions::ChargeCacheService.call(subscription:, charge:, charge_filter:, expires_in: cache_expiration) do
-        yield
-          .map { |fee| fee.attributes.merge("pricing_unit_usage" => fee.pricing_unit_usage&.attributes) }
-          .to_json
+      fees = nil
+      cached_fees = Subscriptions::ChargeCacheService.call(subscription:, charge:, charge_filter:, expires_in: cache_expiration) do
+        fees = yield
+        fees.map do |fee|
+          fee_attributes = fee.attributes
```
jdenquin, line 32: can you add a small comment just to explain what we do here? its not very clear after a first read (missing the business logic context)

app/services/subscriptions/charge_cache_middleware.rb
```diff
@@ -12,28 +12,90 @@ def initialize(subscription:, charge:, to_datetime:, cache: true)
...
+
+    COMPACTABLE_PRICING_UNIT_USAGE_ATTRIBUTES = Set.new([
+      "id",
+      "fee_id",
+      "created_at",
+      "updated_at"
+    ]).freeze
+
+    def parse_cached_fees(cached_fees)
+      JSON.parse(cached_fees).map do |fee_attributes|
```
jdenquin, line 99: any potential CPU usage?

jdenquin: impact*

config/environments/production.rb
```diff
@@ -64,7 +64,8 @@
         Rails.logger.error(exception.backtrace.join("\n"))
 
         Sentry.capture_exception(exception)
-      }
+      },
+      compress_threshold: 512 # 512 bytes
```
jdenquin: what is this threshold for?

app/services/subscriptions/charge_cache_middleware.rb
```diff
@@ -12,28 +12,90 @@ def initialize(subscription:, charge:, to_datetime:, cache: true)
...
     attr_reader :subscription, :charge, :to_datetime, :cache
 
+    COMPACTABLE_PROPERTIES = Set.new([
+      "fixed_charges_duration",
+      "fixed_charges_from_datetime",
+      "fixed_charges_to_datetime"
+    ]).freeze
+
+    COMPACTABLE_ATTRIBUTES = Set.new([
+      "add_on_id",
```
jdenquin, line 69: if we add/delete fields, we have to maintain this list right?

config/environments/production.rb
groyoh: Current threshold is 1024 bytes. If the payload to cache is bigger is than this threshold, Rails will compress before caching. If it's below the threshold it will cached as is. With the null fields being removed, some charges with 1 fee could end up below the threshold causing the cached payload size to stay as big as it currently is.

app/services/subscriptions/charge_cache_middleware.rb
groyoh: > if we add/delete fields, we have to maintain this list right?

Yes. But I made the list explicit to avoid nullifying new fields which shouldn't be nullified. I'll see if I can add a test to catch any issue with new/deleted fields.

app/services/subscriptions/charge_cache_middleware.rb
groyoh: I don't think so. In most cases, we'll have one single fee.

app/services/subscriptions/charge_cache_middleware.rb
groyoh: Sure.

PR #4570, opened 2025-10-31
fix(fees): Return negative total aggregated units

vincent-pochet:
## Context

We recently exposed the `total_aggregated_units` field on both fees and charge_usage objects. It was made to help people knowing how many units are active on weighted sum billable metrics.
However it appears that this value is floured to 0 when the number of units is negative, making it hard for user to rely on this value.

## Description

This PR fixes the situation by always returning the real total_aggregated units even when the value is negative


groyoh (approved):
Could you add a test for this behavior please ?

PR #4571, opened 2025-10-31
fix(billing): add pending_vies_checks table for EU tax validation tracking

ancorcruz:
## Context

When customers use EU tax management, VAT numbers must be validated with the VIES API before invoices can be finalized. The validation process can fail temporarily (timeouts, rate limits, service unavailable) requiring retries.

Currently, there's no persistent tracking of pending VIES checks, making it difficult to monitor validation status, retry attempts, or debug failures.

 ## Description

Add `pending_vies_checks` table to track customers awaiting VIES validation:

- Presence-based status: record exists = validation pending/in-progress
- Tracks retry attempts and error information for debugging
- Unique constraint on customer_id ensures one pending check per customer
- Stores denormalized tax_identification_number for quick access

Table schema:
- organization_id, billing_entity_id, customer_id (with indexes)
- attempts_count: number of retry attempts
- last_attempt_at: timestamp of last validation attempt
- tax_identification_number: VAT number being validated
- last_error_type: normalized error category (timeout, rate_limit, etc.)
- last_error_message: raw error message for debugging

This table will be used in subsequent commits to:
- Block invoice finalization during pending VIES checks
- Implement smart retry logic with exponential backoff
- Provide visibility into validation failures

No behavioral changes in this commit - only database structure.

### DIFF
--- a/app/models/pending_vies_check.rb
+++ b/app/models/pending_vies_check.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+class PendingViesCheck < ApplicationRecord
+  belongs_to :organization
+  belongs_to :billing_entity
+  belongs_to :customer
+end
+
+# == Schema Information
+#
+# Table name: pending_vies_checks
+# Database name: primary
+#
+#  id                        :uuid             not null, primary key
+#  attempts_count            :integer          default(0), not null
+#  last_attempt_at           :datetime
+#  last_error_message        :text
+#  last_error_type           :string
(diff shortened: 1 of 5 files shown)


ancorcruz:
@julienbourdeau we only create the record when it first fails, first VIES check attempt success does not create a record on this table, hence, I see no value on soft-delete to keep uncomplete historical data of VIES checks as its sole purpose is keep track of VIES check process until it is completed and avoid invoice finalization.

PR #4572, opened 2025-10-31
misc(estimate): Add persentage scenario

vincent-pochet:
## Context

## Description

This PR follows https://github.com/getlago/lago-api/pull/4555
It adds one more scenario to cover the sum aggregation with percentage charge model including free units per transaction.
It also fixes one case with clickhouse when using dynamic charge model if no events are found in the period

### DIFF
--- a/app/services/events/stores/clickhouse_store.rb
+++ b/app/services/events/stores/clickhouse_store.rb
@@ -334,7 +334,7 @@ module Events
           sql = <<-SQL
             WITH events AS (#{events_sql})
 
-            SELECT SUM(events.precise_total_amount_cents)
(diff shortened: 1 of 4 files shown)


app/services/events/stores/clickhouse_store.rb
```diff
@@ -338,7 +338,7 @@ def sum_precise_total_amount_cents
             FROM events
           SQL
 
-          connection.select_value(sql)
+          connection.select_value(sql) || 0
```
groyoh: I guess we could also `COALESCE` here.

groyoh (approved):

PR #4573, opened 2025-10-31
chore(cache): Reduce API key cache TTL in development environment

groyoh:
## Context

To optimize performance, we cache API key along with the linked organization for 1 hour. In development environment though, we might often update the organization with premium integrations or enable a specific event store. This can lead the organization to be cached with a different value and therefore to API calls behaving differently from what we might expect.

## Description

This reduces the API key TTL to 10 seconds in development environment to ensure that:

- We still rely on the cache and test it properly locally
- Ensure that any organization update will be reflected at most 10 seconds later

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -46,6 +46,8 @@ module LagoApi
     end
 
     config.active_support.cache_format_version = 7.1
+
+    config.api_key_cache_ttl = 1.hour
   end
 end
 
--- a/app/services/api_keys/cache_service.rb
+++ b/app/services/api_keys/cache_service.rb
@@ -3,12 +3,11 @@
 module ApiKeys
   class CacheService < ::CacheService
     CACHE_KEY_VERSION = "1"
-    CACHE_DURATION = 1.hour
 
     def initialize(auth_token, with_cache: false)
       @auth_token = auth_token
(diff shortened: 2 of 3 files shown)


config/application.rb
```diff
@@ -46,6 +46,8 @@ class Application < Rails::Application
     end
 
     config.active_support.cache_format_version = 7.1
+
+    config.api_key_cache_ttl = 1.hour
```
julienbourdeau, line 50: I'm 100% convinced 1.hour is waaaayyy too long. It's a great improvement to cache the key when under heavy load but it's cheap enough to read it from the source of truth every few minutes at least.

And don't tell me "expiring cache" is easy 😝 In case of emergency, you'd typically connect to the console and expire (or hard delete) the key. If it work for another minute, it's "bad" but another hour it's a disaster. Yes, you should use the service and all, but it's typically, under pressure, when it matters the most that you do the non-optimal thing.

julienbourdeau (approved):
Approved because it's already 1 hour. Nice refactor 👌

vincent-pochet (approved):

config/application.rb
groyoh: I agree with you on that 👍

PR #4582, opened 2025-11-04
chore(console): Add new console helpers for support

julienbourdeau:
Some improvements collected from my recent support shifts.

* Using `find` now prints the organization name directly. It's often the first thing I do
* `retry_generating_invoice(i)` does not raise of error in purposed, to use it in a loop
* `current_usage(sub)` gets you the current usage without cache and witout taxes but let's you override all params

All other change is just disabling `Rails/Output` for the entire file. By design, we want to print stuff with these helpers.

### DIFF
--- a/config/initializers/console.rb
+++ b/config/initializers/console.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# rubocop:disable Rails/Output
 Rails.application.console do
   if Rails.env.development?
     def gavin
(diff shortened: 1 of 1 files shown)


config/initializers/console.rb
```diff
@@ -40,6 +53,16 @@ def enable_premium_integration!(org_id, integration_name)
     org.reload.premium_integrations
   end
 
+  def current_usage(subscription, apply_taxes: false, with_cache: false, **kwargs)
```
vincent-pochet, line 56: ❤️

PR #4586, opened 2025-11-05
feat(api): Add Events `transaction_id` filter

julienbourdeau:
Transaction Id is unique in PG **but not in clickhouse**.

The endpoint [`/events/{transaction_id}`](https://swagger.getlago.com/#/events/findEvent) should be deprecated and replace by this one, a filter on the list of events.

> [!WARNING]
> I made `transaction_id` filter work only with `external_transaction_id` for now (because of the DB index

```
GET /events?transaction_id=abc
``` 

With PG, you always get a list of 1 element.
With CH, you get all the events with this transaction_id.

CleanShot 2026-01-12 at 12 07 07@2x

### DIFF
--- a/spec/requests/api/v1/events_controller_spec.rb
+++ b/spec/requests/api/v1/events_controller_spec.rb
@@ -315,6 +315,62 @@ RSpec.describe Api::V1::EventsController do
         expect(json[:events].count).to eq(1)
         expect(json[:events].first[:lago_id]).to eq(event.id)
       end
+
+      context "with external_subscription_id" do
+        let(:params) { {transaction_id: event.transaction_id, external_subscription_id: event.external_subscription_id} }
+
+        it "returns events" do
+          subject
+
+          expect(response).to have_http_status(:ok)
+          expect(json[:events].count).to eq(1)
+          expect(json[:events].first[:lago_id]).to eq(event.id)
+        end
+      end
+
+      context "with code" do
+        let(:params) { {transaction_id: event.transaction_id, code: event.code} }
+
+        it "returns a validation error" do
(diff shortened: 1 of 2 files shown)


spec/queries/events_query_spec.rb
```diff
@@ -62,6 +62,20 @@
...
+      let(:event2) { create(:event, organization:) }
+      let(:filters) { {transaction_id: event.transaction_id} }
+
+      before { event2 }
+
+      it "applies the filter" do
+        result = events_query.call
+
+        expect(result).to be_success
+        expect(result.events.count).to eq(1)
```
groyoh, line 75: Maybe check that it's `event` we got in the result.

spec/requests/api/v1/events_controller_spec.rb
```diff
@@ -315,6 +315,62 @@
...
+      end
+
+      context "with code" do
+        let(:params) { {transaction_id: event.transaction_id, code: event.code} }
+
+        it "returns a validation error" do
+          subject
+
+          expect(response).to have_http_status(:unprocessable_content)
+          expect(json[:error_details]).to include(transaction_id: ["cannot be combined with code, timestamp_from_started_at, timestamp_from, or timestamp_to"])
```
julienbourdeau, line 338: I know we typically set codes instead of sentences but this is already the case for all filters error. this is how DRY validation gem works. To be fixed later, all together, if necessary.

julienbourdeau:
Closing for now even if I think we should do it. I added a warning in the docs in the meantime. https://github.com/getlago/lago-openapi/pull/474

PR #4588, opened 2025-11-05
chore(queries): Remove constructor override to use Filters with validation

julienbourdeau:
I was looking at the pagination and realize that 2 Query object (only 2 out of ~30) have an override of the constructor.

I think it's better to rely on filters of this and let Dry::Validation gem handle the required/optional filters.

The main benefits are:
- remove duplications of the the defaults value in optional arguments (fitler, search_terms...)
- better validation
- more consistency

### DIFF
--- a/spec/requests/api/v1/wallet_transactions_controller_spec.rb
+++ b/spec/requests/api/v1/wallet_transactions_controller_spec.rb
@@ -315,11 +315,11 @@ RSpec.describe Api::V1::WalletTransactionsController do
       context "with invalid transaction_status value" do
         let(:params) { {transaction_status: "invalid"} }
 
-        it "ignores invalid transaction_status values" do
+        it "returns an error" do
           subject
 
-          expect(response).to have_http_status(:success)
-          expect(json[:wallet_transactions].count).to eq(4)
+          expect(response).to have_http_status(:unprocessable_content)
+          expect(json[:error_details][:transaction_status].size).to eq(1)
         end
       end
     end


app/queries/wallet_transactions_query.rb
```diff
@@ -2,33 +2,34 @@
...
   private
 
-  attr_reader :wallet
+  def filters_contract
+    @filters_contract ||= Queries::WalletTransactionsQueryFiltersContract.new
+  end
+
+  def wallet
+    @wallet ||= organization.wallets.find_by(id: filters.wallet_id)
+  end
```
julienbourdeau, line 33: Maybe I should remove this.
then,  use `organization.where(id: ...).exists?` to return not_found_failure
and add `.where(wallet_id: ...)` to the base scope 🤔

app/queries/wallet_transactions_query.rb
```diff
@@ -2,33 +2,34 @@
...
-    end
-
-    wallet_transactions = with_status(wallet_transactions) if valid_status?(filters.status)
+    wallet_transactions = with_transaction_type(wallet_transactions) if filters.transaction_type.present?
+    wallet_transactions = with_status(wallet_transactions) if filters.status.present?
 
     result.wallet_transactions = wallet_transactions
     result
+  rescue BaseService::FailedResult
+    result
```
ancorcruz, line 22: what is being rescued here?

app/queries/wallet_transactions_query.rb
ancorcruz: it is clear as it is.

however if we can drop the not_found_failure on wallet and use a base scope like we do for the webhooks query is also good. All in all the wallet is a filter, if wallet is not found no wallet transaction would be returned. However this could be a breaking change on public APIs

app/queries/webhooks_query.rb
```diff
@@ -21,14 +18,18 @@ def call
 
     result.webhooks = webhooks
     result
+  rescue BaseService::FailedResult
+    result
```
ancorcruz, line 19: same question, what is being rescued here... I haven't seen this rescue in other query objects.

ancorcruz (approved):
really nice refactor. looks good to me.

app/queries/wallet_transactions_query.rb
julienbourdeau: I think it's because the validate_filters raise a validation failure. So we could argue the `return result unless validate.success?` is the "useless" part but it's done this way in other query objects so I want to follow the pattern for now.

I think we should definitely improve the validation part. Like make the BaseQuery ensure it has validation. We'll see

app/queries/webhooks_query.rb
julienbourdeau: the majority of objects don't have contracts validation

app/queries/wallet_transactions_query.rb
julienbourdeau: Yeah, I'd prefer to do it this way but I definitely don't want to introduce a breaking change

PR #4589, opened 2025-11-05
chore(sidekiq): Allow to configure Sidekiq concurrency in dev env

groyoh:
## Context

When testing concurrency problematics locally, we would usually enable the different workers (`events`, `billing`, etc.). The current setting make it so that each worker would spawn 10 threads. This would result in 60+ Sidekiq threads which can overwhelm the Docker resources and be counter productive.

It also makes sense to configure those independently. For instance, when testing invoice concurrency, we wouldn't need to run 10 `pdf` or `webhook` workers.

## Description

This allows to define the concurrency dynamically. A later change to https://github.com/getlago/lago will then allow to define the concurrency for each worker independently.

### DIFF
--- a/config/sidekiq/sidekiq_analytics.yml
+++ b/config/sidekiq/sidekiq_analytics.yml
@@ -1,4 +1,4 @@
-concurrency: <%= ENV.fetch('SIDEKIQ_CONCURRENCY', 5) %>
+concurrency: <%= ENV.fetch('SIDEKIQ_CONCURRENCY', 10) %>
 timeout: 25
 retry: 1
 queues:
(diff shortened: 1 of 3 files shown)


julienbourdeau:
I think you need to merge the other one first, to not change production data 🤔

groyoh:
@julienbourdeau production is defined separately so we're fine:

```
concurrency: 10
concurrency: <%= ENV.fetch('SIDEKIQ_CONCURRENCY', 5) %>
timeout: 25
retry: 1
queues:
  - analytics
production:
  concurrency: <%= ENV.fetch('SIDEKIQ_CONCURRENCY', 10) %>
staging:
  concurrency: 10
```

PR #4595, opened 2025-11-07
feat(issuing_date-1): Add issuing date preference settings to BillingEntity and Customer

vankiru:
## Roadmap Task

👉  [Spec](https://www.notion.so/getlago/Spec-Define-the-issuing-date-preferences-of-subscription-invoices-2a0ef63110d2807d860cc22ced334bfb)
👉  [Dive In](https://www.notion.so/getlago/BE-Dive-In-Define-the-issuing-date-preferences-of-subscription-invoices-2a3ef63110d2800e8620fb5687f0edf9?d=2a3ef63110d2808f96fd001cbc71298b#2a3ef63110d2807abf14f498b984ac57)

## Context

Lago, since the beginning, creates the invoices with an issuing date at the moment the invoice is issued. We got multiple feedback from customers, where, in that case, they would like the invoice `issuing_date` to be on the last day of the billing period

## Description

This PR is the first part – it adds issuing date settings to both `BillingEntity` and `Customer`.

### DIFF
--- a/db/migrate/20250325145324_assign_customers_to_billing_entities.rb
+++ b/db/migrate/20250325145324_assign_customers_to_billing_entities.rb
@@ -1,8 +1,16 @@
 # frozen_string_literal: true
 
 class AssignCustomersToBillingEntities < ActiveRecord::Migration[7.2]
+  class BillingEntity < ApplicationRecord
+    attribute :subscription_invoice_issuing_date_anchor, :string, default: "next_period_start"
+    attribute :subscription_invoice_issuing_date_adjustment, :string, default: "keep_anchor"
+  end
+
   class Customer < ApplicationRecord
     self.ignored_columns = []
+
+    attribute :subscription_invoice_issuing_date_anchor, :string, default: "next_period_start"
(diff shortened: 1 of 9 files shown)


app/models/customer.rb
```diff
@@ -279,55 +292,57 @@ def ensure_slug
...
+#  shipping_address_line2                       :string
+#  shipping_city                                :string
+#  shipping_country                             :string
+#  shipping_state                               :string
+#  shipping_zipcode                             :string
+#  skip_invoice_custom_sections                 :boolean          default(FALSE), not null
+#  slug                                         :string
+#  state                                        :string
+#  subscription_invoice_issuing_date_adjustment :enum             default("align_with_finalization_date"), not null
+#  subscription_invoice_issuing_date_anchor     :enum             default("next_period_start"), not null
```
ansmonjol: Shouldn't they be nullable on the customer? As we can "fallback" on the billing entity

vankiru: Yep, you're right, totally missed that. Thanks!

julienbourdeau (approved):

julienbourdeau:
I did not expect so many migrations to edit ^^ we should plan a new squash soon!

PR #4598, opened 2025-11-10
chore(database): Allow to easily customize DB pool in dev env

groyoh:
## Context

When debugging performance issue in dev environment, you may be limited by the size of the database connection pool.

## Description

This allows to easily change the pool size using the `DATABASE_POOL` env variable.


groyoh:
Thanks @vincent-pochet!

PR #4601, opened 2025-11-10
feat(multiple-payment-methods): add payment method support for new subscription

lovrocolic:
## Context

## Description

This PR adds GQL support for attaching payment method to the subscription - in create flow. It also covers upgrade and downgrade cases.

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -22,7 +22,7 @@ module Subscriptions
     end
 
     def call
-      return result unless valid?(customer:, plan:, subscription_at:, ending_at: params[:ending_at])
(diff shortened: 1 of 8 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -231,5 +244,12 @@ def editable_subscriptions
     def override_plan(plan)
       Plans::OverrideService.call(plan:, params: params[:plan_overrides].to_h.with_indifferent_access).plan
     end
+
+    def payment_method
+      return @payment_method if defined? @payment_method
+      return nil if params[:payment_method].blank? || params[:payment_method][:payment_method_id].blank?
+
+      @payment_method = PaymentMethod.find_by(id: params[:payment_method][:payment_method_id], organization_id: customer.organization_id)
```
mariohd, line 252: same suggestion here

https://github.com/getlago/lago-api/pull/4605#discussion_r2514044411

PR #4605, opened 2025-11-11
feat(multiple-payment-methods): add GQL support for payment method for one-off invoices

lovrocolic:
## Context

## Description

This PR adds GQL support for attaching payment method to the one-off invoice. Currently, only new attributes are exposed, without changing any logic under the hood.

### DIFF
--- a/app/services/invoices/create_one_off_service.rb
+++ b/app/services/invoices/create_one_off_service.rb
@@ -123,7 +123,7 @@ module Invoices
     def valid_payment_method?
       result.payment_method = payment_method
 
-      PaymentMethods::ValidateService.new(result, **{payment_method: payment_method_params}).valid?
+      PaymentMethods::ValidateService.new(result, payment_method: payment_method_params).valid?
     end
 
     def payment_method


app/services/invoices/create_one_off_service.rb
```diff
@@ -117,5 +119,18 @@ def add_on_identifiers
...
+      result.payment_method = payment_method
+
+      PaymentMethods::ValidateService.new(result, payment_method: payment_method_params).valid?
+    end
+
+    def payment_method
+      return @payment_method if defined? @payment_method
+      return nil if payment_method_params.blank? || payment_method_params[:payment_method_id].blank?
+
+      @payment_method = PaymentMethod.find_by(id: payment_method_params[:payment_method_id], organization_id: customer.organization_id)
```
mariohd, line 133:
```suggestion
      return if payment_method_params.blank? || payment_method_params[:payment_method_id].blank?

      @payment_method ||= PaymentMethod.find_by(id: payment_method_params[:payment_method_id], organization_id: customer.organization_id)
```

lovrocolic: @mariohd It seems to me that memoization is not working in suggested code. If `PaymentMethod.find_by(id: payment_method_params[:payment_method_id], organization_id: customer.organization_id)` returns `nil`, on second method call, the query would be still executed. That's why I added `return @payment_method if defined? @payment_method` - to also cache `nil` values that are result of the query.
Or I am missing something?:)

mariohd: oh, if the `find_by` can return `nil` (I mean, if it is an expected scenario), your code is totally correct!

PR #4607, opened 2025-11-11
feat(multiple-payment-methods): add support for attaching payment method on wallet transaction

lovrocolic:
## Context

## Description

This PR adds GQL support for attaching payment method to the wallet transaction. Currently, only new attributes are exposed, without changing any logic under the hood.

### DIFF
--- a/db/migrate/20251112112544_add_payment_method_to_wallet_transactions.rb
+++ b/db/migrate/20251112112544_add_payment_method_to_wallet_transactions.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddPaymentMethodToWalletTransactions < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    safety_assured do
+      change_table :wallet_transactions, bulk: true do |t|
+        t.references :payment_method, type: :uuid, null: true, index: {algorithm: :concurrently}
+        t.enum :payment_method_type, enum_type: "payment_method_types", default: "provider", null: false
+      end
+    end
+    add_foreign_key :wallet_transactions, :payment_methods, validate: false
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrate/20251112112544_add_payment_method_to_wallet_transactions.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddPaymentMethodToWalletTransactions < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    safety_assured do
+      change_table :wallet_transactions, bulk: true do |t|
+        t.references :payment_method, type: :uuid, null: true, index: {algorithm: :concurrently}
+        t.enum :payment_method_type, enum_type: "payment_method_types", default: "provider", null: false
```
mariohd, line 10: why the default is `provider`?

lovrocolic: It is per design, the same as for subscription or wallets

Example:

```
resource = subscription

payment_method_id = nil, payment_method_type = provider => DEFAULT FLOW - fetches payment method from customer if any

payment_method_id = 123, payment_method_type = provider => attaches dedicated payment method on sub - OVERRIDES ONE FROM CUSTOMER

payment_method_id = nil, payment_method_type = manual => use manual payment method even if there are some payment methods attached on customer - OVERRIDE ONE FROM CUSTOMER WITH MANUAL ONE
```

mariohd: We just have to remember this when fetching `payment_method_type = 'provider'` that we also have the use `payment_method_id IS NOT NULL` for the real provider payments. 

I would prefer to have `payment_method_type` as `nil` in the default flow, but that also works.

PR #4611, opened 2025-11-12
Fixes for pb invoices

annvelents:
## Context

There are two fixes:
1) when calculating creditable amount on progressive billing invoice, we're taking into account issued credit notes, but only "available". We should also be concidering "consumed"
2) Added an additional check when issuing a credit note, that checks is the issued credit note has covered the full amount that should have been issued


ancorcruz (approved):
LGTM!

vincent-pochet (approved):

PR #4614, opened 2025-11-13
perf(webhook): Skip webhook job if no webhook endpoint defined

groyoh:
## Context

When delivering webhooks, we only check whether there's any webhook endpoint defined for the organization once we're processing the `SendWebhookJob` job itself.

This means that when we're generating a lot of invoices (on billing day for instance), we're scheduling thousands of `SendWebhookJob` only to return early. This adds unnecessary pressure on Redis, Sidekiq and DB and might actually delay delivery for organization which have webhook endpoints defined.

## Description

This overrides the `SendWebhookJob.perform_later` method to skip the job scheduling if there's not webhook endpoint defined for the organization. This allows for an easy fix without impacting all the services depending on this job.

Given that I relied on `object.organization`, I added a `Organization#organization` method to have a consistent interface between all organization models.

This also broke a lot of tests that I had to update by adding the `options` and `webhook_id` default values to the job expectations :

```diff
-    expect(SendWebhookJob).to have_been_enqueued.with("wallet.terminated", Wallet)
+    expect(SendWebhookJob).to have_been_enqueued.with("wallet.terminated", Wallet, {}, nil)
```

I also need to update the `perform_all_enqueued_jobs` helper with an `except` option in order to fix the `spec/requests/api/v1/wallets_controller_spec.rb` spec.

### DIFF
--- a/app/jobs/send_webhook_job.rb
+++ b/app/jobs/send_webhook_job.rb
@@ -75,13 +75,18 @@ class SendWebhookJob < ApplicationJob
     "wallet_transaction.payment_failure" => Webhooks::PaymentProviders::WalletTransactionPaymentFailureService
   }.freeze
 
+  # This is a placeholder object to know which arguments were provided.
+  UNDEFINED = Object.new.freeze
+  private_constant :UNDEFINED
+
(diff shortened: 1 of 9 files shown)


groyoh:
Thanks @vincent-pochet !

PR #4615, opened 2025-11-13
fix(clickhouse): Fix cloud migration for events dead letter

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/4602

## Description

The goal here is to fix the migration scripts used to setup the events_dead_letter pipeline in Clickhouse production instances

NOTE: These scripts are not run automatically, they are only here to keep track of the setup and allow us to setup a new instance if required

### DIFF
--- a/db/clickhouse_migrate/cloud/10_events_dead_letter_mv.sql
+++ b/db/clickhouse_migrate/cloud/10_events_dead_letter_mv.sql
@@ -13,13 +13,13 @@ CREATE MATERIALIZED VIEW events_dead_letter_mv TO events_dead_letter
     `error_message` String
 )
 AS SELECT
-  JSONExtractString(event, 'organization_id') AS organization_id,
-  JSONExtractString(event, 'external_subscription_id') AS external_subscription_id,
-  JSONExtractString(event, 'code') AS code,
-  JSONExtractString(event, 'transaction_id') AS transaction_id,
-  toDateTime64(JSONExtractString(event, 'timestamp'), 3) AS timestamp,
-  toDateTime64(JSONExtractString(event, 'ingested_at'), 3) AS ingested_at,
-  toDateTime64(parseDateTime64BestEffort(failed_at), 3) as failed_at,
+  event.organization_id AS organization_id,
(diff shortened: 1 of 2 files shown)


db/clickhouse_migrate/cloud/10_events_dead_letter_mv.sql
```diff
@@ -13,13 +13,13 @@ CREATE MATERIALIZED VIEW events_dead_letter_mv TO events_dead_letter
...
-  JSONExtractString(event, 'code') AS code,
-  JSONExtractString(event, 'transaction_id') AS transaction_id,
-  toDateTime64(JSONExtractString(event, 'timestamp'), 3) AS timestamp,
-  toDateTime64(JSONExtractString(event, 'ingested_at'), 3) AS ingested_at,
-  toDateTime64(parseDateTime64BestEffort(failed_at), 3) as failed_at,
+  event.organization_id AS organization_id,
+  event.external_subscription_id AS external_subscription_id,
+  event.code AS code,
+  event.transaction_id AS transaction_id,
+  toDateTime64OrNull(toString(event.timestamp), 3) AS timestamp,
```
groyoh, line 20: In which case would we have an invalid date here ? Also what's the type of `event.timestamp` ?

groyoh (approved):

db/clickhouse_migrate/cloud/10_events_dead_letter_mv.sql
vincent-pochet: @groyoh the topic contains history data and the format changed a bit with different improvements that were made.
We have, DateTime in ISO8601 strings, timestamp in integer format, and timestamp in decimal with milliseconds precision... We also have some invalid values... :D
To make sure that the events are correctly consumed from the `events_dead_letter_queue`, we have to make sure that all events (even for months ago) can be processed no matter the format.

groyoh: Thanks for the explanation!

PR #4616, opened 2025-11-14
chore(add-on): Clean-up applied add-on dead code

groyoh:
## Context

https://github.com/getlago/lago-api/commit/d45cff61492ad0d15a7803766476040f85efa764 removed code related to applied add-on but the code related to billing those applied add on was never removed.

## Description

This removes this dead code.


groyoh:
@vincent-pochet FYI I also removed `app/services/webhooks/invoices/add_on_created_service.rb` and updated `app/jobs/send_webhook_job.rb` to remove the webhook as it is no longer used as well.

PR #4618, opened 2025-11-17
chore(credit): Refactor `Credits::AppliedPrepaidCreditService` code

groyoh:
## Context

`Credits::AppliedPrepaidCreditService` optimistic retry is currently not configurable which is preventing us from disabling the retry in certain cases where we would like to rely on ActiveJob/Sidekiq retry mechanisms.

## Description

This make `Credits::AppliedPrepaidCreditService` optimistic retry configurable and refactor it slightly while adding tests

### DIFF
--- a/app/services/credits/applied_prepaid_credits_service.rb
+++ b/app/services/credits/applied_prepaid_credits_service.rb
@@ -2,13 +2,13 @@
 
 module Credits
   class AppliedPrepaidCreditsService < BaseService
-    MAX_WALLET_DECREASE_ATTEMPTS = 6
+    DEFAULT_MAX_WALLET_DECREASE_ATTEMPTS = 6
 
-    def initialize(invoice:, wallets:, max_wallet_decrease_attempts: MAX_WALLET_DECREASE_ATTEMPTS)
+    def initialize(invoice:, wallets:, max_wallet_decrease_attempts: DEFAULT_MAX_WALLET_DECREASE_ATTEMPTS)
       @invoice = invoice
       @wallet = wallets.first
       @max_wallet_decrease_attempts = max_wallet_decrease_attempts
-      raise ArgumentError, "max_wallet_decrease_attempts must be greater than or equal to 1" if max_wallet_decrease_attempts < 1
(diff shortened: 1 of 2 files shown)


app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -2,11 +2,13 @@
...
   class AppliedPrepaidCreditsService < BaseService
-    MAX_WALLET_DECREASE_ATTEMPTS = 5
+    MAX_WALLET_DECREASE_ATTEMPTS = 6
 
-    def initialize(invoice:, wallets:)
+    def initialize(invoice:, wallets:, max_wallet_decrease_attempts: MAX_WALLET_DECREASE_ATTEMPTS)
       @invoice = invoice
       @wallet = wallets.first
+      @max_wallet_decrease_attempts = max_wallet_decrease_attempts
+      raise ArgumentError, "max_wallet_decrease_attempts must be greater than or equal to 1" if max_wallet_decrease_attempts < 1
```
floganz: Just an idea – should we also limit max attempts to maybe const that we have?
Thinking how we can make it harder to do stupid things occasionally, like passing 60 attempts.

spec/services/credits/applied_prepaid_credits_service_spec.rb
```diff
@@ -235,5 +235,55 @@
...
+
+      context "when it fails after the max attempts" do
+        before do
+          mock_wallet_balance_decrease_service(succeed_on_attempt: 7)
+        end
+
+        it "raises an error and rolls back the transaction" do
+          expect { subject }.to raise_error(ActiveRecord::StaleObjectError)
+
+          expect(wallet.wallet_transactions.count).to eq(0)
```
floganz, line 290: Do you think it make sense to add transaction count expectation also to success cases?

app/services/credits/applied_prepaid_credits_service.rb
groyoh: I wouldn't have thought of that but yes, we can do that 😄

spec/services/credits/applied_prepaid_credits_service_spec.rb
groyoh: I added this because we didn't have the transaction before, I wanted the test to ensure we rollback everything. It's working today because all the services which call this service actually wrap everything into a transaction. But this service didn't include a transaction in itself. This test would fail on `main` right now.

app/services/credits/applied_prepaid_credits_service.rb
groyoh: Updated.

PR #4619, opened 2025-11-17
feat(multiple-payment-methods): expose payment method in subscription GQL type

lovrocolic:
## Context

## Description

This PR attaches payment method to subscription in GQL context

### DIFF
--- a/app/graphql/types/subscriptions/object.rb
+++ b/app/graphql/types/subscriptions/object.rb
@@ -40,6 +40,9 @@ module Types
 
       field :lifetime_usage, Types::Subscriptions::LifetimeUsageObject, null: true
 
+      field :payment_method, Types::PaymentMethods::Object
+      field :payment_method_type, Types::PaymentMethods::MethodTypeEnum
+
       def next_plan
         object.next_subscription&.plan
       end
--- a/schema.graphql
+++ b/schema.graphql
@@ -9826,6 +9826,8 @@ type Subscription {
   nextSubscriptionType: NextSubscriptionTypeEnum
   onTerminationCreditNote: OnTerminationCreditNoteEnum
   onTerminationInvoice: OnTerminationInvoiceEnum!
+  paymentMethod: PaymentMethod
+  paymentMethodType: PaymentMethodTypeEnum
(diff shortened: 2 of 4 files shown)


app/graphql/types/subscriptions/object.rb
```diff
@@ -40,6 +40,8 @@ class Object < Types::BaseObject
 
       field :lifetime_usage, Types::Subscriptions::LifetimeUsageObject, null: true
 
+      field :payment_method, Types::PaymentMethods::ReferenceObject, method: :itself
```
julienbourdeau: I'm confused why there is `method: :itself` here? 🤔 
I think `object` (the itself here) refers to a subscription object but we want `object.payment_method`

lovrocolic: Those attributes exist in subscription table (just a reference and a type) but I wanted to create nested object, so that's why I used `itself`

However, FE asked for a bit different shape and I changed it.

julienbourdeau: Oooh yeah I get it now 👌 

I prefer the old version then haha but of course, FE should decide.

lovrocolic: actually old version was sth like:
```
paymentMethod {
    paymentMethodId
    paymentMethodType
}
```
But FE realised that there is full object needed + manual / provider distinguisher in the parent level so we chose existing shape. Hope it works for you:)

app/graphql/types/subscriptions/object.rb
```diff
@@ -40,6 +40,9 @@ class Object < Types::BaseObject
 
       field :lifetime_usage, Types::Subscriptions::LifetimeUsageObject, null: true
 
+      field :payment_method, Types::PaymentMethods::Object
```
julienbourdeau, line 43: I'm not sure if `null:` is true or false, here we want to make sure it's nullable (for manual type), right?

lovrocolic: exactly, when manual is selected (but payment methods exist) or when payment methods does not exist at all on customer

PR #4620, opened 2025-11-17
Feat customer okta host

mariohd:
This PR allows add a custom host to access Okta. 

It adds the `host` to `settings` JSON of `OktaIntegration` record. 

When the `host` is not available, it constructs a default host using the `organization_name`.

### DIFF
--- a/app/services/integrations/okta/create_service.rb
+++ b/app/services/integrations/okta/create_service.rb
@@ -17,7 +17,8 @@ module Integrations
           client_id: args[:client_id],
           client_secret: args[:client_secret],
           domain: args[:domain],
-          organization_name: args[:organization_name]
+          organization_name: args[:organization_name],
+          host: args[:host]
         )
 
         integration.save!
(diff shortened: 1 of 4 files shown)


app/models/integrations/okta_integration.rb
```diff
@@ -5,9 +5,13 @@ class OktaIntegration < BaseIntegration
     validates :client_secret, :client_id, :domain, :organization_name, presence: true
     validate :domain_uniqueness
 
-    settings_accessors :client_id, :domain, :organization_name
+    settings_accessors :client_id, :domain, :organization_name, :host
     secrets_accessors :client_secret
 
+    def okta_host
+      host || "#{organization_name.downcase}.okta.com"
+    end
```
groyoh: What about 
```suggestion
    def host
      get_from_settings(:host) || "#{organization_name.downcase}.okta.com"
    end
```

mariohd: Updated! Thanks!

groyoh (approved):
Not a blocker here but just wanted to mentioned that the create/update service tests are a bit "light". They should test that it was created/updated with the right attributes.

jdenquin (approved):

PR #4623, opened 2025-11-17
feat(stripe): Retry with 3DS support if needed

julienbourdeau:
New way of handling 3DS with Lago.

In the first version, I removed `off_session: true` and unfortunaltey, Stripe started triggering a lot of 3ds requests.
In this version, we continue to make the payment with `offsession: true` to reduce the chances of having a 3ds. If there is a 3DS request, the payments fails, then we retry it without `offsession: true` to handle the `requires_action` properly.

CleanShot 2025-11-19 at 10 15 52@2x
CleanShot 2025-11-19 at 10 17 48@2x

## Implementation details

We catch the `authentication_required` error just like we catch `amount_too_small` and:
- mark payment as failed
- saved the error code 🆕  
- enqueue a new job to retry

On the second try, we'll see that there was already a payment with `error_code=authentication_required` for this payable so we don't try again.

## v1: enqueue job manually

I initially re-enqueued the job manually, using the `call_async`.

```ruby
def retry_payment_with_3ds_support
  if payment.payable.is_a? Invoice
    Invoices::Payments::CreateService.call_async(invoice: payment.payable)
  elsif payment.payable.is_a? PaymentRequest
    PaymentRequests::Payments::CreateService.call_async(payable: payment.payable)
  end
end
```

Unfortunately, the new job would fail to be re-enqueued because the uniqueness contraint. It's being enqueued by itself and the current job must be finished before the new one is enqueued.

CleanShot 2025-11-19 at 12 33 02@2x

I figured a good way would be to use the regular job retry mechanism with `retry_on`.

## v2: Introduce `RetriableJob`

What if you could trigger the retry of any jobs from anywhere with a dedicated error.

Notice that the stripe-specific work is done inside another generic service to create payment. I'm adding `should_retry` the same way we already do `reraise`.

CleanShot 2025-11-20 at 10 40 50@2x

## Notes / todo

- [x] confirm the webhook to mess with the **payable** status, especially if they are "late"
- [x] ensure the `invoice.payment_failure` is not sent if payment is retried

### DIFF
--- a/app/services/payment_providers/stripe/payments/create_service.rb
+++ b/app/services/payment_providers/stripe/payments/create_service.rb
@@ -39,7 +39,7 @@ module PaymentProviders
             #       For now we keep it as pending, the user can still update it manually
             prepare_failed_result(e, payable_payment_status: :pending)
           when StripeProvider::NEED_3DS_ERROR_CODE
-            prepare_failed_result(e, should_retry: true)
+            prepare_failed_result(e, should_retry: payment_provider.supports_3ds)
           else
             prepare_failed_result(e)
           end
--- a/spec/integration/stripe/simple_payment_integration_spec.rb
+++ b/spec/integration/stripe/simple_payment_integration_spec.rb
@@ -136,6 +136,8 @@ describe "Stripe Payment Integration Test", :with_pdf_generation_stub, type: :re
 
   context "when invoice payment requires 3DS" do
     it "create a pending payment and send webhook with `next_action`" do
+      provider.update! supports_3ds: true
+
       customer = create_customer
       stripe_customer = customer.payment_provider_customers.sole
       webhooks_sent.clear
--- a/app/models/payment_providers/stripe_provider.rb
+++ b/app/models/payment_providers/stripe_provider.rb
@@ -36,6 +36,7 @@ module PaymentProviders
 
     settings_accessors :webhook_id
     secrets_accessors :secret_key
+    settings_accessors :supports_3ds
 
     def payment_type
       "stripe"
--- a/spec/services/payment_providers/stripe/payments/create_service_spec.rb
+++ b/spec/services/payment_providers/stripe/payments/create_service_spec.rb
@@ -281,32 +281,68 @@ RSpec.describe PaymentProviders::Stripe::Payments::CreateService do
           end
 
           context "when it's the first try" do
-            it "enqueued a new payment creation job" do
-              WebMock.stub_request(:post, "https://api.stripe.com/v1/payment_intents")
-                .with(body: ->(request) {
-                  params = Rack::Utils.parse_query(request)
-                  expect(params["confirm"]).to eq "true"
-                  expect(params["off_session"]).to eq "true"
-                  expect(params["error_on_requires_action"]).to eq "true"
-                })
-                .to_return(
-                  status: 400,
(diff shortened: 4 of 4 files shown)


spec/integration/stripe/simple_payment_integration_spec.rb
julienbourdeau, line 1: Remember this file does not run on CI

app/services/payment_providers/stripe/payments/create_service.rb
```diff
@@ -216,13 +229,14 @@ def error_on_requires_action?
...
-        def prepare_failed_result(error, reraise: false, payable_payment_status: :failed)
+        def prepare_failed_result(error, reraise: false, payment_status: :failed, payable_payment_status: :failed, should_retry: false)
           result.error_message = error.message
           result.error_code = error.code
           result.reraise = reraise
+          result.should_retry = should_retry
 
           # stripe may return us a Stripe::CardError error if payment_intent was created, but it's processing failed, in this case error would contain payment_intent id
-          payment.update!(status: :failed, payable_payment_status:, provider_payment_id: error.error&.payment_intent&.id)
+          payment.update!(status: payment_status, payable_payment_status:, provider_payment_id: error.error&.payment_intent&.id, error_code: error.code)
```
julienbourdeau: Notice that we save the error code for all errors, not just the error I'm interested in 👌

app/support/retry_job_error.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+class RetryJobError < StandardError; end
```
groyoh:
```suggestion
class RetriableError < StandardError; end
```

app/services/invoices/payments/create_service.rb
```diff
@@ -134,6 +133,15 @@ def update_invoice_payment_status(payment_status:)
         )
       end
 
+      def skip_error_webhook(e)
```
groyoh:
```suggestion
      def skip_error_webhook?(e)
```

spec/services/invoices/payments/create_service_spec.rb
```diff
@@ -315,6 +315,30 @@
...
+        context "when error_code is is pending" do
+          let(:result) do
+            BaseService::Result.new.tap do |r|
+              r.payment = instance_double(Payment, status: "failed", payable_payment_status: "failed")
+              r.error_message = "stripe_error"
+              r.error_code = error_code
+            end
+          end
+
+          it "updates the invoice payment status and does not delivers an error webhook" do
```
groyoh, line 332: >updates the invoice payment status

This is not checked in this test I believe.

app/jobs/application_job.rb
```diff
@@ -3,9 +3,13 @@
 class ApplicationJob < ActiveJob::Base
   sidekiq_options retry: 0
 
+  # This is a generic error to trigger the retry of any job
+  # The max attempt is set to avoid infinite loops
+  retry_on RetryJobError, attempts: 20
```
groyoh: Maybe have a polynomially longer retry here ?

julienbourdeau: I hesitated but now I can't thing of a reason not to do it 👍

spec/services/invoices/payments/create_service_spec.rb
julienbourdeau: indeed!

PR #4624, opened 2025-11-18
feat(issuing-date-4): Update issuing_date logic to use subscription_invoice_issuing_date settings

vankiru:
## Roadmap Task

## Context

## Description

This PR implements the logic for `subscription_invoice_issuing_date_anchor` and `subscription_invoice_issuing_date_adjustment`. These settings only apply to **recurring** invoices – meaning invoices with type = "subscription" and invoicing_reason = "subscription_periodic". They don't affect invoices created when a subscription starts or terminated, or any other invoice types.

### Basic idea
`subscription_invoice_issuing_date_anchor` can be either `current_period_end` or `next_period_start`
`subscription_invoice_issuing_date_adjustment` can be either `align_with_finalization` or `keep_anchor`

Billing period: from – to

1. current_period_end + align_with_finalization
```
issuing_date = to + grace_period + 1.day
```

2. current_period_end + keep_anchor
```
issuing_date = to
```

3. next_period_start + align_with_finalization
```
issuing_date = to + grace_period + 1.day
```

4. next_period_start + keep_anchor
```
issuing_date = to + 1.day
```

The extra + 1.day in cases 1 and 3 matches how grace periods currently behave.

### Example

Billing period: Oct 1 – Oct 31, 2025
Grace period: 3 days

1. current_period_end + align_with_finalization → Nov 4, 2025
2. current_period_end + keep_anchor → Oct 31, 2025
3. next_period_start + align_with_finalization → Nov 4, 2025
4. next_period_start + keep_anchor → Nov 1, 2025

### Notes
Right now, invoices are actually generated on the first day of the next billing period, so the code is based on `to + 1.day` instead of just `to`. The implementation is adjusted to match this behavior.

### `Invoice::IssuingDateService`
`Invoices::IssuingDateService` returns an "adjustment" for the issuing date, which it calculates using all three issuing date settings (anchor, adjustment, and grace period). It can take both customer and billing entity as objects or hashes. We use the object version when calculating adjustments for new invoices, and the hash version when updating settings on the customer or billing entity, so we can still calculate adjustments using the old values.

For the `current_period_end + align_with_finalization_date `setup: if the grace period is 0, we need to fall back to the actual end date of the current period – that’s why we do -1 in that case.

### Other changes
Renamed and updated all the `UpdateInvoiceGracePeriodFromBillingEntity`-style services and jobs to `UpdateInvoiceIssuingDateFromBillingEntity`, so they now update the issuing date based on all the settings, not just the grace period.

### DIFF
--- a/app/services/invoices/create_generating_service.rb
+++ b/app/services/invoices/create_generating_service.rb
@@ -52,7 +52,7 @@ module Invoices
     # NOTE: accounting date must be in customer timezone
     def issuing_date
       date = datetime.in_time_zone(customer.applicable_timezone).to_date
-      return date if !grace_period? || charge_in_advance
+      return date if charge_in_advance
 
       issuing_date_service = Invoices::IssuingDateService.new(customer_settings: customer, recurring:)
       date + issuing_date_service.issuing_date_adjustment.days
--- a/spec/services/invoices/create_generating_service_spec.rb
+++ b/spec/services/invoices/create_generating_service_spec.rb
@@ -159,12 +159,13 @@ RSpec.describe Invoices::CreateGeneratingService do
           :customer,
           subscription_invoice_issuing_date_anchor:,
           subscription_invoice_issuing_date_adjustment:,
-          invoice_grace_period: 3
+          invoice_grace_period:
         )
       end
 
       let(:invoice_type) { :subscription }
       let(:invoicing_reason) { "subscription_periodic" }
+      let(:invoice_grace_period) { 3 }
 
       context "with current_period_end + keep_anchor" do
         let(:subscription_invoice_issuing_date_anchor) { "current_period_end" }
@@ -175,6 +176,17 @@ RSpec.describe Invoices::CreateGeneratingService do
 
           expect(result.invoice.issuing_date).to eq(datetime.to_date - 1.day)
           expect(result.invoice.expected_finalization_date).to eq(datetime.to_date + 3.days)
+        end
+
+        context "with no invoice_grace_period" do
+          let(:invoice_grace_period) { 0 }
+
+          it "sets issuing_date to the current billing period end date" do
+            result = create_service.call
+
+            expect(result.invoice.issuing_date).to eq(datetime.to_date - 1.day)
+            expect(result.invoice.expected_finalization_date).to eq(datetime.to_date)
+          end
         end
       end
 
@@ -188,6 +200,17 @@ RSpec.describe Invoices::CreateGeneratingService do
           expect(result.invoice.issuing_date).to eq(datetime.to_date + 3.days)
           expect(result.invoice.expected_finalization_date).to eq(datetime.to_date + 3.days)
         end
+
+        context "with no invoice_grace_period" do
+          let(:invoice_grace_period) { 0 }
+
+          it "sets issuing_date to the current billing period end date" do
+            result = create_service.call
+
+            expect(result.invoice.issuing_date).to eq(datetime.to_date - 1.day)
+            expect(result.invoice.expected_finalization_date).to eq(datetime.to_date)
+          end
+        end
       end
 
       context "with next_period_start + keep_anchor" do
@@ -200,6 +223,17 @@ RSpec.describe Invoices::CreateGeneratingService do
           expect(result.invoice.issuing_date).to eq(datetime.to_date)
           expect(result.invoice.expected_finalization_date).to eq(datetime.to_date + 3.days)
         end
+
+        context "with no invoice_grace_period" do
+          let(:invoice_grace_period) { 0 }
+
+          it "sets issuing_date to the next billing period start date" do
+            result = create_service.call
+
+            expect(result.invoice.issuing_date).to eq(datetime.to_date)
+            expect(result.invoice.expected_finalization_date).to eq(datetime.to_date)
+          end
+        end
       end
 
       context "with next_period_start + align_with_finalization_date" do
@@ -211,6 +245,17 @@ RSpec.describe Invoices::CreateGeneratingService do
 
           expect(result.invoice.issuing_date).to eq(datetime.to_date + 3.days)
           expect(result.invoice.expected_finalization_date).to eq(datetime.to_date + 3.days)
+        end
+
+        context "with no invoice_grace_period" do
+          let(:invoice_grace_period) { 0 }
+
+          it "sets issuing_date to the next billing period start date" do
+            result = create_service.call
+
+            expect(result.invoice.issuing_date).to eq(datetime.to_date)
+            expect(result.invoice.expected_finalization_date).to eq(datetime.to_date)
+          end
         end
       end
 


app/models/customer.rb
```diff
@@ -171,6 +171,18 @@ def applicable_invoice_grace_period
     billing_entity.invoice_grace_period || 0
   end
 
+  def applicable_subscription_invoice_issuing_date_anchor
+    return subscription_invoice_issuing_date_anchor if subscription_invoice_issuing_date_anchor.present?
+
+    billing_entity.subscription_invoice_issuing_date_anchor
```
groyoh: More of a personal preference but since we can't have a `false` value for `subscription_invoice_issuing_date_anchor`, this could be written:

```suggestion
    subscription_invoice_issuing_date_anchor || billing_entity.subscription_invoice_issuing_date_anchor
```

app/services/invoices/issuing_date_service.rb
```diff
@@ -0,0 +1,51 @@
...
+      date.in_time_zone(customer.applicable_timezone).to_date
+    end
+
+    def grace_period
+      period = customer.applicable_invoice_grace_period
+
+      return period unless recurring
+      return 0 if keep_anchor?
+
+      current_period_end? ? period + 1 : period
```
groyoh: I believe this logic might be incorrect. `current_period_end` should have no impact here.

You can actually see the impact in the `spec/services/invoices/create_generating_service_spec.rb` where `subscription_invoice_issuing_date_anchor` has no impact on the issuing date whether `subscription_invoice_issuing_date_anchor` is `current_period_end` or `next_period_start`.

app/services/invoices/issuing_date_service.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module Invoices
+  class IssuingDateService < BaseService
```
groyoh: I don't think this should inherit from `BaseService` as it does not define the `call` method:

```suggestion
  class IssuingDateService
```

app/services/invoices/issuing_date_service.rb
```diff
@@ -0,0 +1,51 @@
...
+      return 0 if keep_anchor?
+
+      current_period_end? ? period + 1 : period
+    end
+
+    def grace_period_diff(old_grace_period)
+      diff = customer.applicable_invoice_grace_period - old_grace_period
+
+      return diff unless recurring
+      return 0 if keep_anchor?
```
groyoh: I believe this might be incorrect as well. If we have `keep_anchor`, it's the new grace period which should be `0`, not the diff. I think this should be:

```suggestion
      return -old_grace_period if keep_anchor?
```

spec/services/invoices/update_issuing_date_from_billing_entity_service_spec.rb
```diff
@@ -88,4 +89,46 @@
...
+      it "does not change the issuing_date" do
+        expect { subject.call }.not_to change { invoice.reload.issuing_date }
+      end
+    end
+
+    context "with align_with_finalization_date" do
+      let(:subscription_invoice_issuing_date_adjustment) { "align_with_finalization_date" }
+
+      it "updates the issuing_date" do
+        expect { subject.call }.to change(invoice, :issuing_date).by(-12)
```
groyoh, line 144: I might be missing something here. I would expect this to change by `11` not `12`.

app/models/customer.rb
vankiru: Yeah, I prefer this style too, but I left it that way since the other methods follow the same pattern.

groyoh: I don't think it's an issue to break the pattern 😉 Plus, I'm pretty sure there might be a Rubocop rule that would automatically do this.

app/services/invoices/issuing_date_service.rb
vankiru: Updated and simplified the logic here – since we only have four cases.

app/services/invoices/issuing_date_service.rb
vankiru: Updated.

app/services/invoices/issuing_date_service.rb
vankiru: Yeah, it was incorrect and didn't account for changes to the anchor or adjustment values – it only handled the grace period. I updated the whole logic so it's easier to follow now.

spec/services/invoices/update_issuing_date_from_billing_entity_service_spec.rb
vankiru: Updated this as well – it should be more clear now )

app/jobs/invoices/update_all_invoice_grace_period_from_billing_entity_job.rb
groyoh, line 1: We shouldn't remove the job in this PR as it might cause dead jobs during deployment. This should be removed in the following OSS version.

app/jobs/invoices/update_grace_period_from_billing_entity_job.rb
groyoh, line 1: We shouldn't remove the job in this PR as it might cause dead jobs during deployment. This should be removed in the following OSS version.

app/jobs/invoices/update_issuing_date_from_billing_entity_job.rb
```diff
@@ -0,0 +1,17 @@
...
+  class UpdateIssuingDateFromBillingEntityJob < ApplicationJob
+    queue_as do
+      if ActiveModel::Type::Boolean.new.cast(ENV["SIDEKIQ_BILLING"])
+        :billing
+      else
+        :invoices
+      end
+    end
+
+    def perform(invoice, old_issuing_date_settings)
```
groyoh: I'm nitpicking here but I believe `previous` would be more suited than `old`

```suggestion
    def perform(invoice, previous_issuing_date_settings)
```

spec/services/billing_entities/update_invoice_issuing_date_settings_service_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+        customer:,
+        billing_entity:,
+        status: :draft,
+        issuing_date: DateTime.parse("19 Jun 2022").to_date,
+        applied_grace_period: 9
+      )
+    end
+
+    context "with premium feature" do
+      around { |test| lago_premium!(&test) }
```
groyoh:
```suggestion
    context "with premium feature", :premium do
```

spec/services/billing_entities/update_invoice_issuing_date_settings_service_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        expect(billing_entity.invoice_grace_period).to eq(15)
+        expect(billing_entity.subscription_invoice_issuing_date_anchor).to eq("current_period_end")
+        expect(billing_entity.subscription_invoice_issuing_date_adjustment).to eq("align_with_finalization_date")
+      end
+
+      it "updates issuing_date and payment_due_date on draft invoices" do
+        expect { perform_enqueued_jobs { update_service.call } }
+          .to change { invoice_draft.reload.issuing_date }.to(DateTime.parse("25 Jun 2022"))
+          .and change { invoice_draft.reload.payment_due_date }.to(DateTime.parse("30 Jun 2022"))
```
groyoh: We rarely perform jobs in the services as far as I know. It should be enough to ensure we schedule the job here.

app/services/billing_entities/update_invoice_issuing_date_settings_service.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def set_issuing_date_settings
+      billing_configuration = params[:billing_configuration]&.to_h || {}
+
+      if billing_configuration.key?(:subscription_invoice_issuing_date_anchor)
+        billing_entity.subscription_invoice_issuing_date_anchor = billing_configuration[:subscription_invoice_issuing_date_anchor]
+      end
+
+      if billing_configuration.key?(:subscription_invoice_issuing_date_adjustment)
+        billing_entity.subscription_invoice_issuing_date_adjustment = billing_configuration[:subscription_invoice_issuing_date_adjustment]
```
groyoh: Since this only applies to grace period, shouldn't it be a premium feature as well ?

spec/jobs/invoices/update_all_invoice_issuing_date_from_billing_entity_job_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      .with(billing_entity:, old_issuing_date_settings:)
+      .and_call_original
+  end
+
+  it "calls the service" do
+    subject
+
+    expect(Invoices::UpdateAllInvoiceIssuingDateFromBillingEntityService)
+      .to have_received(:call)
+  end
```
groyoh, line 27: Couldn't this be :
```suggestion
  it "calls the service" do
    expect(Invoices::UpdateAllInvoiceIssuingDateFromBillingEntityService)
      .to receive(:call)
      .with(billing_entity:, old_issuing_date_settings:)
      .and_call_original
  
    subject
  end
```

spec/jobs/invoices/update_issuing_date_from_billing_entity_job_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  let(:old_issuing_date_settings) do
+    {
+      subscription_invoice_issuing_date_anchor: "current_period_end",
+      subscription_invoice_issuing_date_adjustment: "keep_anchor",
+      invoice_grace_period: 3
+    }
+  end
+
+  before do
+    allow(Invoices::UpdateIssuingDateFromBillingEntityService)
```
groyoh: Same comment as above.

spec/services/billing_entities/update_service_spec.rb
```diff
@@ -126,15 +126,20 @@
...
+          allow(Invoices::UpdateAllInvoiceIssuingDateFromBillingEntityJob).to receive(:perform_later)
         end
 
         it "triggers async updates grace_period of invoices" do
           old_invoice_grace_period = billing_entity.invoice_grace_period
           result = update_service.call
 
           expect(result.billing_entity.invoice_grace_period).to eq(2)
-          expect(Invoices::UpdateAllInvoiceGracePeriodFromBillingEntityJob).to have_received(:perform_later).with(billing_entity, old_invoice_grace_period)
+          expect(Invoices::UpdateAllInvoiceIssuingDateFromBillingEntityJob).to have_received(:perform_later).with(
```
groyoh:
```suggestion
          expect(Invoices::UpdateAllInvoiceIssuingDateFromBillingEntityJob).to have_been_enqueued.with(
```

spec/services/billing_entities/update_service_spec.rb
```diff
@@ -126,15 +126,20 @@
         let(:invoice_grace_period) { 2 }
 
         before do
-          allow(Invoices::UpdateAllInvoiceGracePeriodFromBillingEntityJob).to receive(:perform_later)
+          allow(Invoices::UpdateAllInvoiceIssuingDateFromBillingEntityJob).to receive(:perform_later)
```
groyoh:
```suggestion
```

app/services/customers/update_service.rb
```diff
@@ -78,29 +78,15 @@ def call
...
-          customer.subscription_invoice_issuing_date_anchor = billing_configuration[:subscription_invoice_issuing_date_anchor]
-        end
-
-        if billing_configuration.key?(:subscription_invoice_issuing_date_adjustment)
-          customer.subscription_invoice_issuing_date_adjustment = billing_configuration[:subscription_invoice_issuing_date_adjustment]
-        end
-      end
-
-      if License.premium? && args.key?(:invoice_grace_period)
-        Customers::UpdateInvoiceGracePeriodService.call(customer:, grace_period: args[:invoice_grace_period])
```
groyoh, line 92: This got me really confused. Seems like this one is used by the GraphQL API. We should eventually update the GraphQL endpoint to send the grace period in the `billing_configuration` input.

app/services/customers/update_invoice_issuing_date_settings_service.rb
```diff
@@ -0,0 +1,82 @@
...
+      if customer.changed? && customer.save!
+        # NOTE: Update issuing_date on draft invoices.
+        customer.invoices.draft.find_each do |invoice|
+          invoice.issuing_date = invoice.issuing_date + issuing_date_adjustment(invoice)
+          invoice.payment_due_date = grace_period_payment_due_date(invoice)
+          invoice.save!
+        end
+
+        customer.invoices.ready_to_be_finalized.find_each do |invoice|
+          Invoices::FinalizeJob.perform_later(invoice)
```
groyoh:
```suggestion
          Invoices::FinalizeJob.perform_after_commit(invoice)
```

app/services/customers/update_invoice_issuing_date_settings_service.rb
```diff
@@ -0,0 +1,82 @@
...
+      @params = params
+      @old_issuing_date_settings = {
+        invoice_grace_period: customer.invoice_grace_period,
+        subscription_invoice_issuing_date_anchor: customer.subscription_invoice_issuing_date_anchor,
+        subscription_invoice_issuing_date_adjustment: customer.subscription_invoice_issuing_date_adjustment
+      }
+      super
+    end
+
+    def call
```
groyoh, line 16: I'd suggest wrapping the service into a transaction.

spec/services/customers/update_invoice_issuing_date_settings_service_spec.rb
```diff
@@ -0,0 +1,296 @@
...
+      end
+
+      it "finalizes corresponding draft invoices" do
+        current_date = DateTime.parse("22 Jun 2022")
+
+        travel_to(current_date) do
+          result = update_service.call
+
+          expect(result.customer.invoice_grace_period).to eq(2)
+          expect(Invoices::FinalizeJob).not_to have_received(:perform_later).with(invoice_to_not_be_finalized)
```
groyoh:
```suggestion
          expect(Invoices::FinalizeJob).not_to have_been_enqueued.with(invoice_to_not_be_finalized)
```

spec/services/customers/upsert_from_api_service_spec.rb
```diff
@@ -740,13 +740,13 @@
       end
 
       before do
-        allow(Customers::UpdateInvoiceGracePeriodService).to receive(:call)
+        allow(Customers::UpdateInvoiceIssuingDateSettingsService).to receive(:call)
```
groyoh:
```suggestion
        allow(Customers::UpdateInvoiceIssuingDateSettingsService).to receive(:call).and_call_original
```

app/services/invoices/issuing_date_service.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module Invoices
+  class IssuingDateService
+    def initialize(customer:, billing_entity: nil, recurring: false)
```
groyoh: This could be a bit confusing as it may imply that we expect the model while we expect an object responding to `#[]`:
```suggestion
    def initialize(customer:, billing_entity_issuing_date_settings: nil, recurring: false)
```

spec/services/organizations/update_service_spec.rb
```diff
@@ -159,9 +159,14 @@
 
             expect(result.organization.invoice_grace_period).to eq(2)
             expect(result.organization.default_billing_entity.invoice_grace_period).to eq(2)
-            expect(Invoices::UpdateAllInvoiceGracePeriodFromBillingEntityJob)
+            expect(Invoices::UpdateAllInvoiceIssuingDateFromBillingEntityJob)
               .to have_received(:perform_later)
```
groyoh:
```suggestion
              .to have_been_enqueued
```

app/services/invoices/issuing_date_service.rb
```diff
@@ -0,0 +1,57 @@
...
+    def current_period_end_keep_anchor
+      -1
+    end
+
+    def current_period_end_align_with_finalization_date
+      if align_with_finalization_date? && grace_period.zero?
+        -1
+      else
+        grace_period
+      end
```
groyoh: Shouldn't this be 
```suggestion
      grace_period - 1
```

?

app/jobs/invoices/update_all_invoice_grace_period_from_billing_entity_job.rb
vankiru: Returned it back.

app/jobs/invoices/update_grace_period_from_billing_entity_job.rb
vankiru: Returned it back.

app/services/billing_entities/update_invoice_issuing_date_settings_service.rb
vankiru: We decided to make the feature available for everyone, since we basically want users to be able to generate invoices with issuing dates at the end of the current period vs start of the next period (the anchor setting), and that doesn't depend on the grace period.

app/services/customers/update_service.rb
vankiru: Yeah, I noticed that too – for customer, `invoice_grace_period` isn't inside `billing_configuration` in GraphQL. I added the new settings under billing_configuration. Decided not to change `invoice_grace_period` for now since that would require extra frontend work.

app/services/invoices/issuing_date_service.rb
vankiru: According to the spec, for `align_with_finalization_date`, both `current_period_end` and `next_period_start` should give the same date if a grace period is set. That's why we always add the grace period here – except when it's 0.

app/jobs/invoices/update_grace_period_from_billing_entity_job.rb
```diff
@@ -11,7 +11,14 @@ class UpdateGracePeriodFromBillingEntityJob < ApplicationJob
     end
 
     def perform(invoice, old_grace_period)
-      Invoices::UpdateGracePeriodFromBillingEntityService.call!(invoice:, old_grace_period:)
```
groyoh, line 14: Nice!

app/services/billing_entities/update_invoice_issuing_date_settings_service.rb
groyoh: I agree for the anchor but I was talking about the adjustment one.

app/services/billing_entities/update_service.rb
```diff
@@ -55,12 +55,7 @@ def call
 
         handle_eu_tax_management if params.key?(:eu_tax_management)
 
-        if License.premium? && billing.key?(:invoice_grace_period)
-          BillingEntities::UpdateInvoiceGracePeriodService.call(
-            billing_entity:,
-            grace_period: billing[:invoice_grace_period]
-          )
-        end
+        BillingEntities::UpdateInvoiceIssuingDateSettingsService.call(billing_entity:, params:)
```
julienbourdeau: Not really important but maybe we could pass only the `params[:billing_configuration]`, but you don't have to change it.

app/services/invoices/issuing_date_service.rb
```diff
@@ -0,0 +1,54 @@
...
+    def initialize(customer_settings:, billing_entity_settings: nil, recurring: false)
+      @customer_settings = customer_settings
+      @billing_entity_settings = billing_entity_settings || customer_settings.try(:billing_entity) || {}
+      @recurring = recurring
+    end
+
+    def issuing_date_adjustment
+      return grace_period unless recurring
+
+      send("#{anchor}_#{adjustment}")
```
julienbourdeau, line 14: 👌

julienbourdeau (approved):

app/services/billing_entities/update_invoice_issuing_date_settings_service.rb
vankiru: We decided to make both of them free to keep things simpler on the frontend, and the adjustment won't actually do anything unless a grace period is defined.

PR #4629, opened 2025-11-20
feat(ai-agent): MCP server connection

rsempe:
## Context

We want the AI agent to assist users in running common actions (e.g., API queries) or accessing product knowledge without requiring engineering support.

## Description

The goal of this PR is to setup the connection between Lago API and our MCP server.

### DIFF
--- a/app/services/ai_conversations/stream_service.rb
+++ b/app/services/ai_conversations/stream_service.rb
@@ -19,6 +19,8 @@ module AiConversations
       stream_response
       save_conversation_id!
       notify_completion
+      result.ai_conversation = ai_conversation
+      result
     rescue => e
       handle_error(e)
     end
--- a/lib/lago_mcp_client/lago_mcp_client/client.rb
+++ b/lib/lago_mcp_client/lago_mcp_client/client.rb
@@ -2,6 +2,8 @@
 
 module LagoMcpClient
   class Client
+    include SseParser
+
     PROTOCOL_VERSION = "2024-11-05"
     CLIENT_NAME = "lago-mcp-client"
     CLIENT_VERSION = "0.1"
@@ -57,8 +59,8 @@ module LagoMcpClient
       {
         status: response.code.to_i,
         headers: response.each_header.to_h,
-        body: parse_sse_body(response.body),
-        sse_id: extract_sse_id(response.body)
+        body: parse_sse_data(find_sse_data_line(response.body)),
+        sse_id: extract_sse_id(find_sse_id_line(response.body))
       }
     rescue => e
       {error: e.message}
@@ -91,20 +93,6 @@ module LagoMcpClient
 
       @session_id ||= response[:headers]["mcp-session-id"]
       make_request(method: "notifications/initialized")
-    end
-
-    def parse_sse_body(body)
-      return unless body
-
-      line = body.split("\n").find { |l| l.start_with?("data: ") }
-      JSON.parse(line&.delete_prefix("data: "))
-    rescue JSON::ParserError
-    end
-
-    def extract_sse_id(body)
-      return unless body
-
-      body.split("\n").find { |l| l.start_with?("id: ") }&.delete_prefix("id: ")
     end
   end
 end
--- a/lib/lago_mcp_client/lago_mcp_client/mistral/client.rb
+++ b/lib/lago_mcp_client/lago_mcp_client/mistral/client.rb
@@ -2,6 +2,16 @@
 
 module LagoMcpClient
   module Mistral
+    class ApiError < StandardError
+      attr_reader :error_code, :error_body
+
+      def initialize(error_code, error_body)
+        @error_code = error_code
+        @error_body = error_body
+        super("Mistral API Error (#{error_code}): #{error_body}")
+      end
+    end
+
     class Client
       MISTRAL_CONVERSATIONS_URL = "https://api.mistral.ai/v1/conversations"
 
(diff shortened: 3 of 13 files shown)


lib/lago_mcp_client/lago_mcp_client/run_context.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    private
+
+    def get_tool(name)
+      @tools.find { |tool| tool.name == name }
+    end
+
+    def call_tool(name, arguments = {})
+      tool = get_tool(name)
+      raise "Tool '#{name}' not found" unless tool
```
vincent-pochet: Not a blocker, but I'm just wondering if we should return typed errors for this kind of things. Maybe a `ToolNotFoundError`

lib/lago_mcp_client/lago_mcp_client/sse_client.rb
```diff
@@ -0,0 +1,59 @@
...
+      @running = false
+      @mutex = Mutex.new
+      @thread = nil
+      @callbacks = []
+    end
+
+    def start(&block)
+      @mutex.synchronize { @callbacks << block if block }
+      @running = true
+      @thread ||= Thread.new { run } # rubocop:disable ThreadSafety/NewThread
```
vincent-pochet, line 19: As discussed, it might be a good candidate for Fiber or Ractor, rather than "heavy" blocking threads

lib/lago_mcp_client/lago_mcp_client/sse_client.rb
```diff
@@ -0,0 +1,59 @@
...
+          response.read_body do |chunk|
+            break unless @running
+            chunk.split("\n").each do |line|
+              next if line.strip.empty?
+              if line.start_with?("data: ")
+                event_data = begin
+                  JSON.parse(line[6..])
+                rescue JSON::ParserError
+                  line[6..]
+                end
```
vincent-pochet: Maybe this should be extracted in a `parse_line` method. What do you think?

lib/lago_mcp_client/lago_mcp_client/sse_client.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module LagoMcpClient
+  class SseClient
```
vincent-pochet, line 4: It seems that this client is not covered with specs

app/services/ai_conversations/stream_service.rb
```diff
@@ -3,63 +3,94 @@
 module AiConversations
   class StreamService < BaseService
     Result = BaseResult[:ai_conversation]
```
vincent-pochet, line 5: It seems that the call service is not returning a Result.
Maybe you could return the provided `ai_conversation` so that it matches the signature

spec/services/ai_conversations/stream_service_spec.rb
```diff
@@ -2,85 +2,170 @@
 
 require "rails_helper"
 
-RSpec.describe AiConversations::StreamService do
+describe AiConversations::StreamService, type: :service do
```
vincent-pochet: I think this is no more mandatory, the type is inferred automatically from the file path

lib/lago_mcp_client/lago_mcp_client/client.rb
```diff
@@ -0,0 +1,110 @@
...
+      @session_id ||= response[:headers]["mcp-session-id"]
+      make_request(method: "notifications/initialized")
+    end
+
+    def parse_sse_body(body)
+      return unless body
+
+      line = body.split("\n").find { |l| l.start_with?("data: ") }
+      JSON.parse(line&.delete_prefix("data: "))
+    rescue JSON::ParserError
```
vincent-pochet: Should we log, at least for debug purpose?

lib/lago_mcp_client/lago_mcp_client/mistral/client.rb
```diff
@@ -0,0 +1,114 @@
...
+          end
+        end
+
+        {
+          "conversation_id" => conversation_id,
+          "outputs" => outputs,
+          "tool_calls" => tool_calls.empty? ? nil : tool_calls
+        }
+      rescue LagoHttpClient::HttpError => e
+        raise "Mistral Conversations API Error (#{e.error_code}): #{e.error_body}"
```
vincent-pochet: Should we type the errors here as well? (Maybe we have no need for it 😅 )

lib/lago_mcp_client/lago_mcp_client/mistral/client.rb
```diff
@@ -0,0 +1,114 @@
...
+          "Authorization" => "Bearer #{ENV["MISTRAL_API_KEY"]}",
+          "Accept" => "text/event-stream"
+        }
+
+        conversation_id = nil
+        outputs = []
+        tool_calls = []
+
+        http_client.post_with_stream(payload, headers) do |_type, data, _id, _reconnection_time|
+          next if data == "[DONE]"
```
vincent-pochet: Just wondering if this block could be extracted into a proper "ConversationHandler" class that would be responsible from parsing the messages and acting accordingly.
It could be done in similar way of what we are doing with Webhook handler while keeping the conversation context (outputs, tool_calls...)

vincent-pochet (commented):
Looking good!
Happy to discuss the few comments I added 🙂

vincent-pochet (approved):
🚀 🤖

PR #4632, opened 2025-11-20
perf(invoice): Improve pay-in-advance charge invoice performance

groyoh:
## Context

**TLDR; Wallet currently have performance issues due to long running transactions, pessimistic locks and concurrent updates.** 

<details>
  <summary><i>Click here for the full context</i></summary><br/>

  Today, wallet updates can have multiple performance/scalability impact due to :
  
  1. Optimistic locking which can cause errors when receiving multiple wallet updated at the same time.
  2. Pessimistic locking when updating wallets as we may receive plenty of events within a short period of time causing either wallet balance or `ready_to_be_refreshed` updates to be slow (up to a few seconds)
  
  We noticed that two queries were especially impacted by this issue:
  
  1. Updating wallet's `ready_to_be_refreshed` which took more than 50ms on average:
  
      ```sql
      UPDATE "wallets"
      SET "ready_to_be_refreshed" = true,
        "lock_version" = COALESCE("lock_version", 0) + 1
      WHERE "wallets"."customer_id" = '8d3d0c18-8dfd-4b3c-b356-bfc4ac983111'
        AND "wallets"."status" = 0;
      ```
  
  2. Updating wallet's balance which could sometime takes a few seconds:
  
      ```sql
      UPDATE "wallets"
      SET "updated_at" = "2025-11-08 08:24:08.329614",
        "ongoing_balance_cents" = 99999999000,
        "credits_ongoing_balance" = "999999990.0",
        "ready_to_be_refreshed" = false,
        "last_ongoing_balance_sync_at" = "2025-11-08 08:24:08.328737",
        "lock_version" = 6
      WHERE "wallets"."id" = "ffea9fdc-6491-43fa-8184-69fe25fb89d9"
        AND "wallets"."lock_version" = 5;
      ```
  
  ### `EXPLAIN`’ing the queries
  
  There was not clear reason as to why we have such a high average time on wallet update. Updating a single wallet locally takes less than a few ms:
  
  ```sql
  EXPLAIN (ANALYZE true, VERBOSE true) UPDATE "wallets"
  SET "ready_to_be_refreshed" = true, "lock_version" = COALESCE("lock_version", 0) + 1
  WHERE "wallets"."customer_id" = 'ffffd865-fb67-4492-bf95-1760dc256415' AND "wallets"."status" = 0;
                                                                       QUERY PLAN                                                                     
  ----------------------------------------------------------------------------------------------------------------------------------------------------
   Update on public.wallets  (cost=0.28..8.30 rows=0 width=0) (actual time=0.316..0.317 rows=0 loops=1)
     ->  Index Scan using index_wallets_on_customer_id on public.wallets  (cost=0.28..8.30 rows=1 width=11) (actual time=0.065..0.067 rows=1 loops=1)
           Output: (COALESCE(lock_version, 0) + 1), true, ctid
           Index Cond: (wallets.customer_id = 'ffffd865-fb67-4492-bf95-1760dc256415'::uuid)
           Filter: (wallets.status = 0)
   Planning Time: 1.056 ms
   Execution Time: 0.582 ms
  (7 rows)
  
  EXPLAIN (ANALYZE TRUE, VERBOSE TRUE)
  UPDATE "wallets"
  SET "updated_at" = '2025-11-08 08:24:08.329614',
    "ongoing_balance_cents" = 99999997000,
    "credits_ongoing_balance" = '999999990.0',
    "ready_to_be_refreshed" = false,
    "last_ongoing_balance_sync_at" = '2025-11-08 08:24:08.328737',
    "lock_version" = 8
  WHERE "wallets"."id" = 'ffea9fdc-6491-43fa-8184-69fe25fb89d9'
    AND "wallets"."lock_version" = 7
                                                                                                     QUERY PLAN                                                                                                   
  ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
   Update on public.wallets  (cost=0.28..8.30 rows=0 width=0) (actual time=0.252..0.252 rows=0 loops=1)
     ->  Index Scan using wallets_pkey on public.wallets  (cost=0.28..8.30 rows=1 width=61) (actual time=0.131..0.132 rows=1 loops=1)
           Output: '2025-11-08 08:24:08.329614'::timestamp(6) without time zone, '99999997000'::bigint, 999999990.00000::numeric(30,5), 8, false, '2025-11-08 08:24:08.328737'::timestamp without time zone, ctid
           Index Cond: (wallets.id = 'ffea9fdc-6491-43fa-8184-69fe25fb89d9'::uuid)
           Filter: (wallets.lock_version = 7)
   Query Identifier: -3248043656870226248
   Planning Time: 0.346 ms
   Execution Time: 0.356 ms
  (8 rows)
  ```
  
  ### Actual concurrency
  
  Naturally, the next step was to understand the impact of high concurrency on those queries performance.
  
  I wanted to confirm that such case could happen on production so I tried to retrieve a list of events which occurred within the same second over the last few days. So I ran the following query on organizations with lots of wallets:
  
  ```sql
  WITH
    external_ids AS (
      SELECT
        subscriptions.external_id
      FROM
        wallets
        JOIN subscriptions ON wallets.customer_id = subscriptions.customer_id
      WHERE
        wallets.organization_id = 'xyz'
      AND wallets.updated_at >= NOW() - INTERVAL '1 day'
	  LIMIT 4000
    )
  SELECT
    external_customer_id,
    DATE_TRUNC('second', timestamp) AS second_bucket,
    COUNT(*) AS events_count
  FROM
    events
  WHERE
    deleted_at IS NULL
    AND external_subscription_id IN (
      SELECT
        external_id
      FROM
        external_ids
    )
    AND organization_id = 'xyz'
    AND timestamp >= NOW() - INTERVAL '1 day'
    AND external_customer_id IS NOT NULL
  GROUP BY
    external_customer_id,
    DATE_TRUNC('second', timestamp)
  HAVING
    COUNT(*) > 10
  ORDER BY
    events_count DESC,
    external_customer_id,
    second_bucket;
  ```
  
  I was able to confirm have plenty of occurrences where we receive hundreds of events within the same seconds and therefore causing hundreds of wallet updates. These events were both for pay-in-advance and pay-in-arrear charges so it wasn't clear at this point was the root cause was.
  
  ### Reproducing concurrency on `ready_to_be_refreshed`
  
  I added a small script to try and reproducing high concurrency on `ready_to_be_refreshed` update and check the actual query performance there using `ActiveSupport` subscription and PG stats:
  
  ```ruby
  puts "Connection pool size: #{ActiveRecord::Base.connection_pool.size}" # => Connection pool size: 200
  
  SQL_TIMES = []
  
  module ActiveRecord
    class StatsSubscriber < ActiveSupport::Subscriber
      attach_to :active_record
  
      def sql(event)
        SQL_TIMES << event.duration if event.payload[:sql].start_with?("UPDATE \"wallets\"")
      end
    end
  end
  
  def bench_update(count = 10)
    customer= Wallet.active.order(:created_at).last.customer
    SQL_TIMES.clear
    count.times.map do
      Thread.new do
        customer.flag_wallets_for_refresh
      end
    end.each(&:join)
  
    if SQL_TIMES.size != count
      raise "SQL_TIMES.size != count"
    end
    puts "Average time: #{((SQL_TIMES.sum / SQL_TIMES.size)).round(2)}ms"
  end
  
  bench_update(1) # => Average time: 1ms - PG hero time: 0ms
  bench_update(5) # => Average time: 3ms - PG hero time: 0ms
  bench_update(10) # => Average time: 6ms - PG hero time: 0ms
  bench_update(50) # => Average time: 15ms - PG hero time: 1ms
  ```
  
  We can see that there might be a slight impact on the application time, which could be due to either the high number of threads or network congestion between Rails and PG, but nothing significant when checking PG stats.
  
  This shows that high concurrency on this query does not justify the values we have on production.
  
  ### Finding the root cause
  
  To find the root cause, we relied on RDS Performance Insight to give us a better understand of the actual wait time.
  
  We were able to see that `ready_to_be_refreshed` updates actually gets slow when there are concurrent balance update.
  
  Checking the wait times, we noticed two things:
  1. `ready_to_be_refreshed` update is mostly waiting due to `Lock::tuple wait` ([Amazon Aurora Lock:tuple](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/apg-waits.locktuple.html)).
  2. balance update is mostly waiting due to Lock::transactionid wait ([Amazon Aurora Lock:transactionid](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/apg-waits.locktransactionid.html))
  
  To understand better those waits, we had to know which part of the code it was related to. We assumed them to be related to pay-in-advance charges on a single wallet and were able to confirm that by checking the logs during those high wait times.
  
  ### Understanding the root cause
  
  The key element here is to understand what happens when we receive an event for a `pay-in-advance` charge:
  
  1. Upon receiving the event, we’ll schedule the pay-in-advance invoice generation and mark all of the customer’s wallets as `ready_to_be_refreshed`
  2. Then, when generating the invoice for pay-in-advance charges, we’ll try and update the wallet balance within the same transaction as the invoice generation. We rely on optimistic locking to avoid concurrently updating the wallet balance. If optimistic locking fails, we’ll retry to update the wallet up to 5 times causing a wait within a transaction of up to 2.5 seconds.
  
  What we can interpret from the previous information is that:
  
  1. Updating the wallet balance will cause PG to lock the wallet record until the invoice DB transaction completes.
  2. Any concurrent `ready_to_be_refreshed` update will have to wait for this lock to be released in order to complete. One important thing here is that this update does not have a `lock_version` condition in the `WHERE` so it cannot cause an `ActiveRecord::StaleObjectError` (`UPDATE "wallets" SET "ready_to_be_refreshed" = false, "lock_version" = COALESCE("lock_version", 0) + 1 WHERE "wallets"."customer_id" = '8d3d0c18-8dfd-4b3c-b356-bfc4ac983111' AND "wallets"."status" = 0`).
  3. Any concurrent balance update will have to wait for the lock to be released BUT will most likely fail with an `ActiveRecord::StaleObjectError` as the first transaction would have updated the `lock_version` upon committing the transaction. This would in turn cause a retry and therefore make the DB transaction even longer.
  
  ## Reproducing the root cause
  
  In order to debug this issue, we setup a plan with pay-in-advance plan and ran the following scripts:
  
  ```ruby
  PARALLEL_CUSTOMER_COUNT = 1
  
  organization = Organization.find_by(name: "Hooli")
  subscription_ids = PARALLEL_CUSTOMER_COUNT.times.map do
    customer_external_id = SecureRandom.uuid
    subscription_external_id = SecureRandom.uuid
    customer = Customers::CreateService.call!(
      organization_id: organization.id,
      external_id: customer_external_id,
      name: "f_Customer #{customer_external_id}",
      email: "customer-#{customer_external_id}@example.com",
      currency: "USD",
      ).customer
  
    wallet = Wallets::CreateService.call!(
      params: {
        organization_id: organization.id,
        customer: customer,
        name: "Wallet #{customer_external_id}",
        rate_amount: 1,
        granted_credits: '0',
        currency: "USD",
      }
    ).wallet
  
    WalletTransactions::CreateFromParamsService.call!(
      organization:,
      params: {
        wallet_id: wallet.id,
        organization_id: organization.id,
        customer: customer,
        name: "Wallet #{customer_external_id}",
        rate_amount: 1,
        granted_credits: '1000000000.0',
        currency: "USD",
      }
    )
  
    subscription = Subscriptions::CreateService.call!(
      customer: customer,
      plan: Plan.find_by(code: "pay_in_advance"),
      params: {
        subscription_at: 1.day.ago,
        external_id: subscription_external_id,
      }
    )
  
    subscription_external_id
  end
  
  1000.times do |i|
    puts "Iteration #{i}" if i % 1000 == 0
    Parallel.flat_map(subscription_ids, in_threads: 4) do |subscription_id|
    Events::CreateService.call!(
      organization: organization,
      params: {
        transaction_id: SecureRandom.uuid,
        external_subscription_id: subscription_id,
        code: 'count_bm'
      },
      timestamp: Time.now.to_f,
      metadata: {}
    )
    end
  end
  ```
  
  This allowed us to get similar results as on production.
</details>

## Description

The fact that we update the wallet within a long running transaction and retry on lock failure will cause longer transactions and longer lock wait times for other concurrent jobs.

Given we are already retrying the job in case of `ActiveRecord::StateObjectError`, we could skip the optimistic locking retry and let the job be retried.

One issue with this is that the failed jobs will be retried with a polynomially longer retry but Sidekiq scheduler will actually retry jobs on average every few seconds (see https://github.com/sidekiq/sidekiq/blob/d09147898edcc64879432764a6bd355666eeb8ba/lib/sidekiq/scheduled.rb#L129). That means that every few seconds we’ll retry a whole set of scheduled jobs, only a few will acquire the lock and be processed, while the others will be scheduled for retry.

To alleviate all these issues, we can:

1. Reduce the `average_scheduled_poll_interval` value to a lower value (Done in https://github.com/getlago/lago-api/commit/7b98fe7e15b156a3f3bd0280094afe6e446aa08b).
2. Rely on a distributed lock around the whole job processing. This way, we will fail fast and retry the job rather than increasing the duration of the transaction. Note that we decided to got for advisory lock instead of redlock because it's already used in the codebase, allows.
3. Changing the existing retry strategy (`polynomially_longer`) to be randomly distributed within a range.

### DIFF
--- a/spec/jobs/invoices/create_pay_in_advance_charge_job_spec.rb
+++ b/spec/jobs/invoices/create_pay_in_advance_charge_job_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe Invoices::CreatePayInAdvanceChargeJob do
     it "returns a random delay between 0 and 8 seconds" do
       values = Array.new(1000) { described_class.retry_delay }
 
-      expect(values).to all(be_between(0, 8))
+      expect(values).to all(be_between(0, 16))
     end
   end
 
@@ -73,8 +73,8 @@ RSpec.describe Invoices::CreatePayInAdvanceChargeJob do
 
     [
       [Sequenced::SequenceError.new("Sequenced::SequenceError"), 15],
-      [WithAdvisoryLock::FailedToAcquireLock.new("customer-1"), 15],
-      [ActiveRecord::StaleObjectError.new("Attempted to update a stale object: Wallet."), 15],
+      [WithAdvisoryLock::FailedToAcquireLock.new("customer-1"), 25],
+      [ActiveRecord::StaleObjectError.new("Attempted to update a stale object: Wallet."), 25],
       [BaseService::ThrottlingError.new(provider_name: "Stripe"), 25]
     ].each do |error, attempts|
       error_class = error.class


vincent-pochet (approved):
Nice one!

app/services/invoices/create_pay_in_advance_charge_service.rb
```diff
@@ -74,6 +78,10 @@ def call
     delegate :subscription, to: :event
     delegate :customer, to: :subscription
 
+    def customer_lock_key
+      "customer-#{customer.id}"
```
floganz, line 82: Do you think that we might need more specific name? Currently we lock on customer only for wallets.
Can there be another customer specific lock which in fact not about using wallets?

app/services/invoices/create_pay_in_advance_charge_service.rb
```diff
@@ -17,34 +20,35 @@ def call
...
+          invoice.fees_amount_cents = invoice.fees.sum(:amount_cents)
+          invoice.sub_total_excluding_taxes_amount_cents = invoice.fees_amount_cents
+          Credits::AppliedCouponsService.call(invoice:) if invoice.fees_amount_cents&.positive?
+
+          if tax_error?(fee_result)
+            invoice.failed!
+            invoice.fees.each { |f| SendWebhookJob.perform_later("fee.created", f) }
+            create_error_detail(fee_result.error.messages.dig(:tax_error)&.first)
+            Utils::ActivityLog.produce(invoice, "invoice.failed")
+            next
```
floganz, line 38: I'm a bit surprised to see `next` outside of loop. How this suppose to work?

groyoh: We're within a block here so next returns from the block. See https://gemdocs.org/gems/rubocop-rails/2.22.1/RuboCop/Cop/Rails/TransactionExitStatement.html.

app/services/invoices/create_pay_in_advance_charge_service.rb
groyoh: That's a question I asked my self. Technically, the lock is not only for wallet but also other things like credit note which currently has no lock for instance.

In the end, I'm actually wondering whether we need:

1. One single lock per customer
2. One lock per customer/linked resource pair (customer+walllet, customer+credit note, etc.)

I'm not sure I got the answer.

PR #4634, opened 2025-11-21
feat(payment): Add error_code

julienbourdeau:
Extracted from https://github.com/getlago/lago-api/pull/4623

This introduce a new columns on payment to save the provider error_code.
This is needed for the 3ds retry but I think it's generally useful, even if we go a different route with 3ds.

Now that I write these line, I'm thinking `provider_error_code` is also a good name.

### DIFF
--- a/app/services/invoices/payments/stripe_service.rb
+++ b/app/services/invoices/payments/stripe_service.rb
@@ -36,6 +36,7 @@ module Invoices
 
         payable_payment_status = payment.payment_provider&.determine_payment_status(payment.status)
         payment.payable_payment_status = payable_payment_status
+        payment.error_code = stripe_payment.error_code if stripe_payment.error_code
         payment.save!
 
         update_invoice_payment_status(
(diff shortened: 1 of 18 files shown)


app/services/invoices/payments/stripe_service.rb
```diff
@@ -36,6 +36,7 @@ def update_payment_status(organization_id:, status:, stripe_payment:)
 
         payable_payment_status = payment.payment_provider&.determine_payment_status(payment.status)
         payment.payable_payment_status = payable_payment_status
+        payment.error_code = stripe_payment.error_code if stripe_payment.error_code
```
vincent-pochet, line 39: What about something like:
```suggestion
        payment.error_code = stripe_payment.error_code.presence
```

julienbourdeau: I think this would do the same as `payment.error_code = stripe_payment.error_code` 🤔 

I initially wanted to not assign the field if the value is nil to avoid overriding error_code, but maybe it should be overridden if there is no error_code in the webhook content.

vincent-pochet: Hmm, ok. Make sense. Not a blocker for me, feel free to ignore :)

PR #4635, opened 2025-11-21
feat(webhooks): Sends invoice.payment_failed when retry to pay withou…

mariohd:
This PR sends webhook "invoice.payment_failure" when we retry the payment without a payment provider.

### Attention
only retries, 1st attempt does not send webhook because the missing payment provider.

### DIFF
--- a/app/services/invoices/payments/retry_service.rb
+++ b/app/services/invoices/payments/retry_service.rb
@@ -20,6 +20,12 @@ module Invoices
           return result.not_allowed_failure!(code: "payment_processor_is_currently_handling_payment")
         end
 
+        if invoice.customer.payment_provider.nil?
+          Utils::ActivityLog.produce(invoice, "invoice.payment_failure")
+          SendWebhookJob.perform_later("invoice.payment_failure", invoice, {provider_error: {error_message: "no linked_payment provider", error_code: "no_linked_payment_provider"}})
+          return result.single_validation_failure!(error_code: "no_linked_payment_provider")
+        end
+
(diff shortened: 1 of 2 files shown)


app/services/invoices/payments/retry_service.rb
```diff
@@ -20,6 +20,12 @@ def call
           return result.not_allowed_failure!(code: "payment_processor_is_currently_handling_payment")
         end
 
+        if invoice.customer.payment_provider.nil?
+          Utils::ActivityLog.produce(invoice, "invoice.payment_failure")
+          SendWebhookJob.perform_later("invoice.payment_failure", invoice, {provider_error: {error_message: "no linked_payment provider", error_code: "no_linked_payment_provider"}})
+          return result.single_validation_failure!(error_code: "no_linked_payment_provider")
```
vincent-pochet, line 26: I'm wondering if we should return a `result.not_found_failure!` rather than a `validation_failure` 🤔

mariohd: that's a good question!

i've used `validation_failure` to be consistent with other "payment" endpoint. 

https://github.com/getlago/lago-api/blob/main/app/services/customers/generate_checkout_url_service.rb#L16-L18

Having `not_found_failure!` IMO is more related to "I have an ID, use it" and it can't found the record; which is not the case. Here we have a payment  retry but it fails to perform because it don't have the payment method, so it feels to me more like a validation failure.

vincent-pochet: Alright! I'm fine with it :)

PR #4636, opened 2025-11-24
misc(datadog): Add service middleware

vincent-pochet:
## Description

This PR follows https://github.com/getlago/lago-api/pull/4621 that was adding an optional config to enable datadog monitoring.

This part is adding a new service middleware name `DatadogMiddleware` to monitor the service execution.

### DIFF
--- a/app/services/base_service.rb
+++ b/app/services/base_service.rb
@@ -174,6 +174,7 @@ class BaseService
   end
 
   use(Middlewares::LogTracerMiddleware)
+  use(Middlewares::DatadogMiddleware) if ENV["DD_AGENT_HOST"].present?
 
   def self.call(*, **, &)
     new(*, **).call_with_middlewares(&)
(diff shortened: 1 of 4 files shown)


app/services/base_service.rb
```diff
@@ -174,6 +174,7 @@ def self.use(middleware_class, *args, on_conflict: :raise, **kwargs)
   end
 
   use(Middlewares::LogTracerMiddleware)
+  use(Middlewares::DatadogMiddleware)
```
julienbourdeau: Is there a way we can only check DD_AGENT_HOST once when the app is eager loaded during boot rather than every service call 🤔 

Or maybe we could only call use if DD is enabled here.

vincent-pochet: That's a fair point. I will try to add it only if DD is enabled

PR #4637, opened 2025-11-24
chore(deps): Allow to install dependencies in `vendor/bundle`

groyoh:
## Context

While debugging, it is sometimes helpful to edit a dependency source code to add breakpoints, test a fix, etc.

The current setup forces us to either:

1. Edit files directly from within the container.
2. Set the `BUNDLE_PATH` env to be within the `/app` folder.

The first strategy is not efficient as you have to connect to the container and install an editor to do so.

The second one is a more efficient but has multiple issues right now:

1. IT makes app boot time longer as the gems are within a mounted volume
2. The folder appears in `git`.
3. The `scripts/start.dev.sh` relies on `rails` CLI which is taken from the gem installed during `docker build` and may drift from the current version until we re-build the image again.

## Description

This ensures we rely on `bundle exec rails` in the `scripts/start.dev.sh` and gitignore the `vendor/bundle` folder.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -2,7 +2,6 @@ FROM ruby:3.4.7-slim
 
 WORKDIR /app
 ENV POSTGRES_VERSION=14
-ENV BUNDLE_PATH=/app/vendor/bundle
 
 RUN apt update -qq && \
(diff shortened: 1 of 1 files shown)


julienbourdeau:
I thnik you still need to restart the container, changes in the vendor dir a not "reload", are they?

PR #4638, opened 2025-11-24
feat(metadata): Introduce Metadata model

nepalez:
# Roadmap Task

## Context

The PR implements the necessary database and model changes to support metadata in credit notes.

This is the first part of the feature addressing a customer request who needs to flag credit notes synced to its internal system to prevent re-syncing.

## Description

This PR introduces a new unified metadata system using a single item_metadata table with JSONB storage.

The design addresses the limitations of existing metadata implementations (separate tables per entity, row-per-key-value granularity, complex merge semantics) and enables future migration of existing metadata (`Customer`, `Invoice`, `WalletTransaction`), as well as adding metadata to new models with a minimal boilerplate.

### Part 1: Add owner-agnostic ItemMetadata model

- New `Metadata::ItemMetadata` model with polymorphic owner association
- Single item_metadata table stores flat key-value objects as JSONB
- Database constraints enforce limits (50 keys max, 40 char keys, 255 char values)
- PostgreSQL trigger validates structure at database layer
- Comprehensive specs covering app validations and database constraints

### Part 2: Add metadata to credit notes

- `CreditNote` gains optional metadata association
- Composite foreign key ensures referential integrity (metadata belongs to correct credit note and organization)
- Application-level validation with user-friendly error messages
- Database constraints prevent invalid data even when validations bypassed
- Specs verify both app validation and constraint enforcement

### DIFF
--- a/app/models/credit_note.rb
+++ b/app/models/credit_note.rb
@@ -14,6 +14,10 @@ class CreditNote < ApplicationRecord
   belongs_to :organization
 
   has_one :billing_entity, through: :invoice
+  has_one :metadata,
+    class_name: "Metadata::ItemMetadata",
+    as: :owner,
+    dependent: :destroy
 
   has_many :items, class_name: "CreditNoteItem", dependent: :destroy
   has_many :fees, through: :items
--- a/spec/factories/credit_notes.rb
+++ b/spec/factories/credit_notes.rb
@@ -42,5 +42,14 @@ FactoryBot.define do
     trait :with_items do
       items { create_pair(:credit_note_item) }
     end
+
+    trait :with_metadata do
+      after(:create) do |credit_note|
+        credit_note.create_metadata!(
+          organization_id: credit_note.organization_id,
+          value: {"key" => "value"}
+        )
+      end
+    end
   end
 end
--- a/spec/models/credit_note_spec.rb
+++ b/spec/models/credit_note_spec.rb
@@ -13,6 +13,7 @@ RSpec.describe CreditNote do
   it_behaves_like "paper_trail traceable"
 
   it { is_expected.to belong_to(:organization) }
+  it { is_expected.to have_one(:metadata).class_name("Metadata::ItemMetadata").dependent(:destroy) }
   it { is_expected.to have_many(:integration_resources) }
   it { is_expected.to have_many(:error_details) }
 


app/models/metadata/item_metadata.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+module Metadata
+  class ItemMetadata < ApplicationRecord
+    belongs_to :organization
+    belongs_to :owner, polymorphic: true
+
+    validates :owner_id, uniqueness: {scope: :owner_type}
```
julienbourdeau: Personally, I'm not a big fan of validating this in rails because I think it adds a query every time you update the metadata. There are very very few cases where you could create 2 ItemMetadata for the same model and I think it's best to rely on the unique index error raised.
What's your take on this?

app/models/metadata/item_metadata.rb
```diff
@@ -0,0 +1,70 @@
...
+          errors.add(:value, "value for key '#{key}' must be empty or a String up to 255 characters")
+        end
+      end
+    end
+
+    def owner_consistency
+      return if owner.blank?
+
+      if owner.organization_id != organization_id
+        errors.add(:owner, "must belong to the same organization as the metadata")
```
julienbourdeau: about the error message, we don't use proper english string in the backend but only error code. 
You can check `en.yml` to see how to replaced translation with error codes.
Here I'd use something like `inconsistent_organization` maybe. 


```suggestion
        errors.add(:owner, "inconsistent_organization")
```

app/models/credit_note.rb
```diff
@@ -181,53 +186,64 @@ def ensure_number
...
     self.number = "#{invoice.number}-CN#{formatted_sequential_id}"
   end
+
+  def ensure_metadata_consistency
+    return unless metadata
+
+    if metadata.organization_id != organization_id
+      errors.add(:metadata, "must belong to the same organization as the credit note")
+    end
+  end
```
julienbourdeau: I'm not sure about this kind of verification. 
Now, if you change the credit note status, the app will load the metadata to ensure it's organization_id is correct.
Maybe we could do this kind of check on creation only. This reduce the risk of a mistake but keep things light.

Today, we don't have this kind of validation anywhere in the app. I feel like if we start adding this all over the place it's going to trigger a lot of extra requests.

Please share your thoughts!

db/migrate/20251125114057_create_item_metadata.rb
```diff
@@ -0,0 +1,85 @@
...
+
+            IF val_length > 255 THEN
+              RAISE EXCEPTION 'metadata value length cannot exceed 255 characters (key "%" has value with % characters)', key_name, val_length;
+            END IF;
+          END LOOP;
+
+          RETURN NEW;
+        END;
+        $$ LANGUAGE plpgsql;
+      SQL
```
julienbourdeau: I think we should choose if we want this in the app or in the DB but I don't like having the validation twice.

app/models/credit_note.rb
julienbourdeau: I gave this more thoughts and I think it should be done in the services instead. You'd usually have all the model and ids necessary for this type of checks.

app/models/metadata/item_metadata.rb
vincent-pochet: I tend to agree as well. I would suggest to only validate the presence, and leave the uniqueness part to the database

app/models/metadata/item_metadata.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def value_correctness
+      return if value.blank?
+
+      unless value.is_a?(Hash)
+        errors.add(:value, "must be a Hash")
+        return
+      end
+
+      if value.size > 50
```
vincent-pochet: We have a lot of "magic" constant here, can you turn them into proper constants?

db/migrate/20251125114057_create_item_metadata.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+class CreateItemMetadata < ActiveRecord::Migration[8.0]
+  def up
+    create_table :item_metadata, id: :uuid do |t|
+      t.references :organization,
+        null: false,
+        type: :uuid,
+        foreign_key: {on_delete: :cascade},
+        comment: "Reference to the organization"
```
vincent-pochet: Comments feels a bit overkill here to be honest (not a blocker ;) )

db/migrate/20251125114057_create_item_metadata.rb
```diff
@@ -0,0 +1,85 @@
...
+      t.timestamps
+
+      t.index [:owner_type, :owner_id], unique: true
+      t.index [:id, :owner_id, :organization_id], unique: true, name: "index_item_metadata_for_fk"
+      t.index :value, name: "index_item_metadata_on_value", using: :gin
+    end
+
+    safety_assured do
+      execute <<-SQL.squish
+        CREATE OR REPLACE FUNCTION validate_item_metadata_value()
```
vincent-pochet: Enforcing this kind of validation at database level is not a good idea in my opinion. Business logic should stay in the application layer, not the database

app/models/metadata/item_metadata.rb
groyoh: FYI I have a WIP branch for adding a custom validator to validate this everywhere:

```ruby
class OrganizationValidator
  def validate(record)
    record.class.reflect_on_all_associations(:belongs_to).map do |reflection|
      next if reflection.name == :organization
      association = association.public_send(reflection.name)
      next if association.nil? || association.organization_id = record.organization_id
      
      errors.add(reflection.name, "inconsistent_organization")
    end
  end
end

class Customer < ApplicationRecord
  validates_with OrganizationValidator
end
```

app/models/metadata/item_metadata.rb
nepalez: fixed

db/migrate/20251125114057_create_item_metadata.rb
nepalez: I simplified it by leaving the object-type-check only.

And because such an invariant can be ensured in a check constrant, I dropped the trigger along with its function and switched from the `up/down` to the `change` syntax as well.

app/models/credit_note.rb
nepalez: fixed

app/models/metadata/item_metadata.rb
nepalez: Thank you, I'll remeber it for the future, but now I just cancelled the particular check

app/models/metadata/item_metadata.rb
nepalez: Fixed

db/migrate/20251125114057_create_item_metadata.rb
nepalez: TLTR: fixed

I looked at it as not at a buisiness rule, but the data invariants, that should be enforced at the database layer. But during our morning conversation with @julienbourdeau we agreed that we can afford big data in the database.

The reason is that (1) as the data are protected at the model level, the passing through that protection should be rare, (2) even if later some big metadata will be created by whatever means, this will break the webhooks at first, then the serialization, and only then will threat the app/database as a whole. So we could just rely on Sentry.

db/migrate/20251201094058_add_metadata_to_credit_notes.rb
```diff
@@ -0,0 +1,32 @@
...
+  def up
+    safety_assured do
+      change_table :credit_notes, bulk: true do |t|
+        t.uuid :metadata_id, comment: "Reference to the credit note metadata"
+
+        t.index [:metadata_id, :id, :organization_id],
+          where: "metadata_id IS NOT NULL",
+          unique: true,
+          name: "index_credit_notes_metadata_fk"
+      end
```
julienbourdeau: I believe the index creation will create a lock on the table. Nothing will match `where: "metadata_id IS NOT NULL"` so I'm sure it will be very fast but usually we disable the transaction and add the index concurrently.

What do you think?

```suggestion
class AddMetadataToCreditNotes < ActiveRecord::Migration[8.0]
 disable_ddl_transaction!
 
  def up
    safety_assured do
      change_table :credit_notes, bulk: true do |t|
        t.uuid :metadata_id, comment: "Reference to the credit note metadata"
      end
      
      
      add_index [:metadata_id, :id, :organization_id],
        where: "metadata_id IS NOT NULL",
        unique: true,
        name: "index_credit_notes_metadata_fk"
```

nepalez: Thank you! Will fix

julienbourdeau: I think it can't be done in the `bulk:` block

julienbourdeau (approved):

PR #4650, opened 2025-11-26
Service layer for credit notes metadata

nepalez:
## Roadmap Task

## Context

This PR adds the service layer for credit note metadata management.

This is the second part of the feature addressing a customer request from Mistral AI who needs to
flag credit notes synced to SAP to prevent re-syncing.

### Service architecture

The implementation introduces a reusable `Metadata::UpdateItemService` that handles all metadata operations. This owner-agnostic design allows the same service to be reused for future metadata implementations (Invoice, Customer, WalletTransaction).

`CreditNotes::CreateService` and `CreditNotes::UpdateService` both delegate metadata handling to `Metadata::UpdateItemService`.

### Metadata::UpdateItemService

Core service that manages metadata lifecycle with the following parameters:

| Parameter | Type | Description |
|-----------|------|-------------|
| `owner` | ActiveRecord | The entity owning metadata (e.g., CreditNote) |
| `value` | Hash/nil | New metadata content |
| `partial` | Boolean | `true` for merge, otherwise replace |
| `preview` | Boolean | Build in memory without persisting |

Behavior matrix:

| old value | new value | partial | action |
|-----------|-----------|---------|--------|
| nil | nil | any | no-op |
| nil | non-nil | any | create metadata |
| non-nil | nil | true | no-op |
| non-nil | nil | false | delete metadata |
| non-nil | non-nil | true | merge new keys |
| non-nil | non-nil | false | replace entirely |

Empty hash `{}` is treated as a valid non-nil value (replaces metadata with empty object unless the `partial` is set to `true`).

### Metadata::DeleteItemKeyService

Service for removing individual keys from existing metadata. Returns `result.deleted_value` and `result.changed` flag.

### Integration with CreditNote services

**CreateService** accepts `metadata` parameter, always replaces (new credit note has no existing metadata).

**UpdateService** supports merge (default) and replace modes:
- `metadata: {key: value}, partial_metadata: true` — merges new keys with existing
- `metadata: {key: value}` — replaces entirely
- `metadata: nil` — deletes all metadata

### Preview mode

The CreateService supports the `preview` mode for credit note estimation. When `context: :preview` is passed, both the credit note and its metadata are only built in memory with generated UUIDs but not persisted to the database.

### DIFF
--- a/app/services/metadata/delete_item_key_service.rb
+++ b/app/services/metadata/delete_item_key_service.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Metadata
+  # Remove a key from an existing metadata.
+  # Return an error result if the metadata has already been deleted.
+  class DeleteItemKeyService < BaseService
+    Result = BaseResult[:item, :metadata_changed]
+
+    # @option [Metadata::MetadataItem] :item The metadata item to modify
+    # @option [#to_s] :key The key of the metadata item to delete
+    def initialize(item:, key:)
+      @item = item
+      @key = key.to_s
+
+      super()
+    end
+
+    def call
+      item.update!(value: item.value.to_h.except(key))
+
+      result.item = item
+      result.metadata_changed = item.previous_changes.any?
+      result
+    rescue ActiveRecord::RecordNotFound
+      result.not_found_failure!(resource: "metadata_item")
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    end
+
+    private
+
+    attr_reader :item, :key
+  end
+end
--- a/spec/services/metadata/delete_item_key_service_spec.rb
+++ b/spec/services/metadata/delete_item_key_service_spec.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Metadata::DeleteItemKeyService do
+  subject(:service) { described_class.new(item:, key:) }
+
+  let(:organization) { create(:organization) }
+  let(:owner) { create(:credit_note, organization:) }
+  let(:item) { create(:item_metadata, owner:, organization:, value:) }
+  let(:value) { {"foo" => "bar", "baz" => "qux"} }
+  let(:key) { "foo" }
+
+  describe "#call" do
+    context "when key exists" do
+      it "removes the key from metadata" do
+        result = service.call
+
+        expect(result).to be_success
+        expect(result.metadata_changed).to be(true)
+        expect(item.reload.value).to eq({"baz" => "qux"})
+      end
+    end
+
+    context "when key does not exist" do
+      let(:key) { "nonexistent" }
+
+      it "does not modify metadata" do
+        result = service.call
+
+        expect(result).to be_success
+        expect(result.metadata_changed).to be(false)
+        expect(item.reload.value).to eq({"foo" => "bar", "baz" => "qux"})
+      end
+    end
+
+    context "when key is a symbol" do
+      let(:key) { :foo }
+
+      it "converts key to string and removes it" do
+        result = service.call
+
+        expect(result).to be_success
+        expect(result.metadata_changed).to be(true)
+        expect(item.reload.value).to eq({"baz" => "qux"})
+      end
+    end
+
+    context "when removing the last key" do
+      let(:value) { {"foo" => "bar"} }
+
+      it "leaves empty hash" do
+        result = service.call
+
+        expect(result).to be_success
+        expect(result.metadata_changed).to be(true)
+        expect(item.reload.value).to eq({})
+      end
+    end
+
+    context "when value contains nil" do
+      let(:value) { {"foo" => nil, "baz" => "qux"} }
+
+      it "removes key with nil value" do
+        result = service.call
+
+        expect(result).to be_success
+        expect(result.metadata_changed).to be(true)
+        expect(item.reload.value).to eq({"baz" => "qux"})
+      end
+    end
+
+    context "when value contains empty string" do
+      let(:value) { {"foo" => "", "baz" => "qux"} }
+
+      it "removes key with empty string" do
+        result = service.call
+
+        expect(result).to be_success
+        expect(result.metadata_changed).to be(true)
+        expect(item.reload.value).to eq({"baz" => "qux"})
+      end
+    end
+  end
+end


app/services/metadata/delete_item_key_service.rb
```diff
@@ -0,0 +1,38 @@
...
+      old_value = item.value.to_h
+      key = result.key.to_s
+
+      result.changed = old_value.key?(key)
+      result.deleted_value = old_value[key]
+      item.update!(value: old_value.except(key))
+
+      result
+    rescue ActiveRecord::RecordNotFound
+      result.not_found_failure!(resource: "metadata_item")
```
julienbourdeau, line 25: I'm not sure what part would trigger a not found exception.

app/services/credit_notes/update_service.rb
```diff
@@ -3,32 +3,53 @@
...
+      ActiveRecord::Base.transaction do
+        if update_refund_status?
+          credit_note.refund_status = refund_status
+          credit_note.refunded_at = Time.current if credit_note.succeeded?
+        end
+
+        update_metadata!
+
+        # Added for visibility of what's going on
+        # (it is expected, though, that the `update_metadata!` to save the credit not by itself)
```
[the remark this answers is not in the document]
nepalez: fixed

app/services/metadata/update_item_service.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    delegate :owner, :value, :replace, :preview, to: :result
+    delegate :metadata, :metadata_id, :organization_id, to: :owner
+
+    # the `metadata` is a `has_one` association,
+    # while the `metadata_id` is the other way fk-protected reference.
+    # both must be present to ensure the owner supports metadata.
+    def metadata_supported?
+      owner.respond_to?(:organization_id) && owner.respond_to?(:metadata) && owner.respond_to?(:metadata_id)
+    end
```
julienbourdeau: I think it's faire to assume that if this service is called, the owner supports metadata. If it doesn't the code will break.

app/services/credit_notes/update_service.rb
```diff
@@ -3,32 +3,53 @@
...
 
-    attr_reader :credit_note, :params
+    attr_reader :credit_note, :params, :refund_status
+
+    def update_refund_status?
+      params.key?(:refund_status)
+    end
+
+    def update_metadata!
+      value = params[:metadata]&.then { |m| m.respond_to?(:to_unsafe_h) ? m.to_unsafe_h : m.to_h }
```
julienbourdeau, line 49: I like that this is passing a proper hash. to often we pass params from controller or graphql input objects! 👍 👍 👍

app/services/metadata/update_item_service.rb
```diff
@@ -0,0 +1,118 @@
...
+      end
+
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    end
+
+    private
+
+    delegate :owner, :value, :replace, :preview, to: :result
```
julienbourdeau: I understand the idea and I kinda like it but we have so mny services and they all use `@something = val` and an `attr_reader` that I'd prefer to keep it the way way.
(same for the other service)

app/services/metadata/update_item_service.rb
```diff
@@ -0,0 +1,118 @@
...
+  # without persisting them to the database. The owner's `metadata` and `metadata_id`
+  # attributes are modified in memory, but nothing is saved.
+  class UpdateItemService < BaseService
+    Result = BaseResult[:owner, :value, :replace, :preview]
+
+    # @param [ActiveRecord::Base] owner The record whose metadata is to be updated
+    # @option [#to_h] :value The new content of the metadata item
+    # @option [Boolean] :replace Whether to replace the existing content or merge with it
+    # @option [Boolean] :preview When true, builds objects in memory without persisting
+    def initialize(owner, value:, replace: false, preview: false)
```
julienbourdeau: Ideally, I'd rather use the name `partial:` rather than `replace`. It's the same (reversed) but it's consistant with Entitlements.
I think there should not be a default value to make sure it's always called with explicite value.

app/services/metadata/update_item_service.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+    def change_metadata!
+      if delete_metadata?
+        owner.metadata_id = nil
+      elsif create_metadata?
+        owner.id ||= SecureRandom.uuid
+        new_metadata = owner.build_metadata(id: SecureRandom.uuid, organization_id:, value:)
+        new_metadata.save! unless preview
+        owner.metadata_id = new_metadata.id
```
julienbourdeau: I'm surprised that we need to generate IDs here. Is it because we have the `owner.metadata_id`?


With a regular has_many I'd typically do this
```suggestion
        new_metadata = owner.metadata.build(organization_id:, value:)
        new_metadata.save # Outside the service here ideally
```

app/services/metadata/update_item_service.rb
```diff
@@ -0,0 +1,118 @@
...
+      #   (metadata_id, id, organization_id) -> (id, owner_id, organization_id)
+      #   In Postgres 14 and earlier, ON DELETE CASCADE does NOT have a SET NULL option for a particular column
+      #   of the composite foreign key, that's why we cannot delete the metadata item before
+      #   saving the owner record with a `metadata_id: nil`.
+      #   After migration to Postgres 15+, the foreign key could receive the clause
+      #   `ON DELETE SET NULL (metadata_id)` for the `metadata_id` column,
+      #   in which case `change_metadata!` could delete the metadata item by itself.
+      ActiveRecord::Base.transaction do
+        change_metadata!
+        next result if preview
```
julienbourdeau: Ideally, I'd like to remove the `preview` argument from this service? 🤔

app/services/metadata/delete_item_key_service.rb
nepalez: I'm thinking about the case when the service itself is called from outside of the transaction context, so the `item` could be removed by another transaction between the time it was taken from the db and the beginning of the implicit transaction in the LOC:25

Won't fix ❓

app/services/metadata/update_item_service.rb
nepalez: Will fix

app/services/metadata/update_item_service.rb
nepalez: Well, actually I just looked at how this was done, for example, here https://github.com/getlago/lago-api/blob/main/app/services/credit_notes/create_service.rb#L112-L113

Do we really want those variables ❓

app/services/metadata/update_item_service.rb
nepalez: It will. I use it only to provide a specific error message in LOC:47 above

Won't fix ❓

app/services/metadata/update_item_service.rb
nepalez: Yes, because we protect the metadata from not being created without an `owner_id`, but at the same time the owner should refer back to its own metadata. That's why both ids should be known in advance.

At least, this logics is incapsulated in this service, and doesn't leak out

Won't fix :question:

app/services/metadata/update_item_service.rb
nepalez: But this won't let the service be reused inside the credit node creation in a preview mode.
This is because this service saves metadata along with its owner atomically.

Won't fix ❓

app/services/metadata/update_item_service.rb
nepalez: The interesting consequence of this decision is that `partial: false` became the default instead of the `replace: false`. Is that desired?

From the API/GraphQL point of view this doesn't matter, my question is just about the "ergonomics" of the service itself

nepalez: Fixed :white_check_mark:

PR #4655, opened 2025-11-27
feat(exports): Exports usage thresholds

murparreira:
## Context

We need a view for the usage_thresholds table

## Description

Create the view

### DIFF
--- a/db/views/exports_usage_thresholds_v01.sql
+++ b/db/views/exports_usage_thresholds_v01.sql
@@ -1,6 +1,6 @@
 SELECT
     organization_id,
-    plan_id,
+    plan_id AS lago_plan_id,
     id AS lago_id,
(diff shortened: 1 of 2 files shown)


db/views/exports_usage_thresholds_v01.sql
```diff
@@ -0,0 +1,11 @@
+SELECT
+    organization_id,
+    plan_id,
```
jdenquin: should be lago_plan_id

jdenquin: ```plan_id as lago_plan_id,```

murparreira: Okay, will do it

jdenquin (approved):

PR #4663, opened 2025-11-28
feat(metadata): Services, GraphQL and API layers

nepalez:
## Roadmap Task

  ## Context

  This PR adds Services, GraphQL and REST API support for credit note metadata. These are parts 2-4 of the feature

  ## Description

  ### Read operations

  - **GET /v1/credit_notes/:id** — returns metadata in the response
  - **GET /v1/credit_notes** — includes metadata with eager loading (also fixes an existing N+1 query)

  ### Write operations via credit note endpoints

  - **POST /v1/credit_notes** — accepts a `metadata` hash on creation
  - **PUT /v1/credit_notes/:id** — supports metadata merge and replace

  ### Dedicated metadata endpoints

  | Method | Endpoint | Description |
  |--------|----------|-------------|
  | POST | `/v1/credit_notes/:id/metadata` | Replaces all metadata |
  | PATCH | `/v1/credit_notes/:id/metadata` | Merges new keys into existing metadata |
  | DELETE | `/v1/credit_notes/:id/metadata` | Removes all metadata |
  | DELETE | `/v1/credit_notes/:id/metadata/:key` | Removes a single key |

### DIFF
--- a/app/services/credit_notes/update_service.rb
+++ b/app/services/credit_notes/update_service.rb
@@ -47,8 +47,7 @@ module CreditNotes
     # @return [Boolean] if the metadata was changed in any way
     def update_metadata!
       value = params[:metadata]&.then { |m| m.respond_to?(:to_unsafe_h) ? m.to_unsafe_h : m.to_h }
-      service = Metadata::UpdateItemService.new(owner: credit_note, value:, partial: partial_metadata.present?)
-      result = service.call!
+      Metadata::UpdateItemService.call!(owner: credit_note, value:, partial: partial_metadata.present?)
       @metadata_changed = result.metadata_changed
     end
 


app/services/credit_notes/update_service.rb
```diff
@@ -2,33 +2,62 @@
...
+
+    def update_refund_status?
+      params.key?(:refund_status)
+    end
+
+    # @return [Boolean] if the metadata was changed in any way
+    def update_metadata!
+      value = params[:metadata]&.then { |m| m.respond_to?(:to_unsafe_h) ? m.to_unsafe_h : m.to_h }
+      service = Metadata::UpdateItemService.new(owner: credit_note, value:, partial: partial_metadata.present?)
+      result = service.call!
```
julienbourdeau:
```suggestion
      Metadata::UpdateItemService.call!(owner: credit_note, value:, partial: partial_metadata.present?)
```
Not important but service class are typically called this way. I know some are used with `new().call` but it's not common way. Older services had multiple method instead of one `call` too.

This is FYI for future PR, no need to change it here if you prefer keeping it this way.

app/controllers/concerns/credit_note_index.rb
```diff
@@ -34,7 +34,15 @@ def credit_note_index(external_customer_id:)
...
-          result.credit_notes.includes(:items, :applied_taxes, :invoice),
+          result.credit_notes.includes(
+            :items,
+            :applied_taxes,
+            :file_attachment,
+            :xml_file_attachment,
+            :error_details,
+            :metadata,
+            invoice: :billing_entity
+          ),
```
julienbourdeau, line 45: great improvement 🙆‍♂️

PR #4665, opened 2025-12-01
fix(validation): Fixes sanitization and validation issues

groyoh:
## Context

We are currently lacking some sanitizations and validations sometimes causing 500 errors:

|Description| Error|
|---|---|
|`BillableMetricController` doesn't properly sanitize the `billable_metric` parameter and fail with an exception when passing a invalid value | `undefined method 'permit' for an instance of String (NoMethodError)`|
|`CustomerController` doesn't properly sanitize the `customer` parameter and fail with an exception when passing a invalid value | `undefined method 'permit' for an instance of String (NoMethodError)`|
|`Coupon`'s `reusable` attribute cannot be nil but is not validated in the model causing a PG error when passing `nil` | `PG::NotNullViolation: ERROR:  null value in column "reusable" of relation "coupons" violates not-null constraint (ActiveRecord::NotNullViolation)`|
|`Customer`'s `account_type` is not validated causing `ArgumentError` when passing invalid value | `'partner_customer' is not a valid account_type (ArgumentError)`|
|`Customer`'s attributes fail when receiving strings containing null byte (`\u0000`) | `string contains null byte (ArgumentError)`|
| When we receive an blank Okta `state`, we fail to retrieve the cache |`key cannot be blank (ArgumentError)`|
|When we receive an blank Okta `code`, we fail to retrieve to authenticate the user|`Error: {"error":"invalid_grant","error_description":"The authorization code is invalid or has expired."} (LagoHttpClient::HttpError)` |
|Login fails when either `email` or `password` contains null byte | `string contains null byte (ArgumentError)`|
|Refreshing draft invoices does not work with `one_off` invoices | |
|Wallet transactions amounts cannot be greater than `10^25` |`PG::NumericValueOutOfRange: ERROR:  numeric field overflow (ActiveRecord::RangeError)`|
|Wallet transactions amounts cannot be negative ||

## Description

This fixes those sanitization/validation issues.

### DIFF
--- a/app/services/invoices/refresh_draft_and_finalize_service.rb
+++ b/app/services/invoices/refresh_draft_and_finalize_service.rb
@@ -9,7 +9,7 @@ module Invoices
 
     def call
       return result.not_found_failure!(resource: "invoice") if invoice.nil?
-      return result.forbidden_failure! if invoice.one_off?
+      return result.forbidden_failure! unless invoice.subscription?
       return result unless invoice.draft?
       drafted_issuing_date = invoice.issuing_date
 
--- a/spec/services/wallet_transactions/validate_service_spec.rb
+++ b/spec/services/wallet_transactions/validate_service_spec.rb
@@ -66,27 +66,27 @@ RSpec.describe WalletTransactions::ValidateService do
       :voided_credits,
       :granted_credits,
       :paid_credits
-    ].each do |amount|
-      context "with #{amount} >= 10^25" do
-        let(amount) { 10**25 }
+    ].each do |attr|
+      context "with #{attr} >= 10^25" do
+        let(attr) { 10**25 }
 
         it "returns false and result has errors" do
           expect(validate_service).not_to be_valid
-          expect(result.error.messages[amount]).to eq(["invalid_#{amount}", "invalid_amount"])
+          expect(result.error.messages[attr]).to eq(["invalid_#{attr}", "invalid_amount"])
         end
       end
 
-      context "with #{amount} < 0" do
-        let(amount) { "-1.00" }
+      context "with #{attr} < 0" do
+        let(attr) { "-1.00" }
 
         it "returns false and result has errors" do
           expect(validate_service).not_to be_valid
-          expect(result.error.messages[amount]).to eq(["invalid_#{amount}", "invalid_amount"])
+          expect(result.error.messages[attr]).to eq(["invalid_#{attr}", "invalid_amount"])
         end
       end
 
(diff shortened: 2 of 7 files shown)


app/services/invoices/refresh_draft_and_finalize_service.rb
```diff
@@ -9,6 +9,7 @@ def initialize(invoice:)
 
     def call
       return result.not_found_failure!(resource: "invoice") if invoice.nil?
+      return result.forbidden_failure! if invoice.one_off?
```
vincent-pochet: Should we whitelist rather than blacklist here?
Refreshing credit or advance charge is not possible either

spec/services/wallet_transactions/validate_service_spec.rb
```diff
@@ -62,6 +62,49 @@
       end
     end
 
+    [
+      :voided_credits,
+      :granted_credits,
+      :paid_credits
+    ].each do |amount|
+      context "with #{amount} >= 10^25" do
+        let(amount) { 10**25 }
```
vincent-pochet: It's maybe a bit confusing to have both the iterator variable and the tested value sharing the same name. WDYT?

groyoh: Not really cause it's not a Symbol here  but a variable to a Symbol which can be `granted_credits`, `voided_credits` and `paid_credits`. So the `let` variable will be one of those three.

But I agree that `amount` is not actually the best naming here. I'll update it.

app/services/invoices/refresh_draft_and_finalize_service.rb
groyoh: We probably should. Which ones should be allowed then ?

app/models/coupon.rb
```diff
@@ -58,6 +58,8 @@ class Coupon < ApplicationRecord
 
   validates :percentage_rate, presence: true, if: :percentage?
 
+  validates :reusable, exclusion: [nil]
```
floganz, line 61: Isn't it the same as `presence: true` validation?
Or you want to achieve certain error message?

groyoh: `presence: true` considers `false` as invalid unfortunately. See https://guides.rubyonrails.org/active_record_validations.html#presence. Using `validates :boolean_field_name, inclusion: [true, false]` will cause some warnings: https://github.com/thoughtbot/shoulda-matchers/blob/main/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb#L281-L292.

PR #4667, opened 2025-12-02
feat(issuing-date-5) Finalize invoices based on expected_finalization_date

vankiru:
## Roadmap Task

## Context

With the new issuing date adjustment setting, `issuing_date` might skip using `invoice_grace_period` (when the adjustment set to "keep_anchor"). But we still need to factor in `invoice_grace_period` when finalizing invoices.

## Description

This PR adds a new field, `expected_finalization_date`, which stores the date when a draft invoice should be auto-finalized. It's basically:
```
expected_finalization_date = invoice creation date + invoice_grace_period.days
```

The `expected_finalization_date` field is mostly for internal use. We only expose it in GraphQL so we can show the right message about when a draft invoice will be finalized. It's not available in the REST API or in the SDK clients.

### DIFF
--- a/db/migrate/20251201084648_add_expected_finalization_date_to_invoices.rb
+++ b/db/migrate/20251201084648_add_expected_finalization_date_to_invoices.rb
@@ -1,18 +1,7 @@
 # frozen_string_literal: true
 
 class AddExpectedFinalizationDateToInvoices < ActiveRecord::Migration[8.0]
-  def up
+  def change
     add_column :invoices, :expected_finalization_date, :date
-
-    safety_assured do
-      execute <<-SQL
-        UPDATE invoices
-        SET expected_finalization_date = issuing_date
-      SQL
-    end
-  end
-
-  def down
-    remove_column :invoices, :expected_finalization_date
   end
 end
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -121,7 +121,7 @@ class Invoice < ApplicationRecord
   scope :invisible, -> { where(status: INVISIBLE_STATUS.keys) }
   scope :with_generated_number, -> { where(status: %w[finalized voided]) }
   scope :ready_to_be_refreshed, -> { draft.where(ready_to_be_refreshed: true) }
-  scope :ready_to_be_finalized, -> { draft.where("expected_finalization_date <= ?", Time.current.to_date) }
+  scope :ready_to_be_finalized, -> { draft.where("COALESCE(expected_finalization_date, issuing_date) <= ?", Time.current.to_date) }
 
   scope :created_before,
     lambda { |invoice|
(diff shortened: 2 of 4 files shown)


db/migrate/20251201084648_add_expected_finalization_date_to_invoices.rb
```diff
@@ -0,0 +1,18 @@
...
+class AddExpectedFinalizationDateToInvoices < ActiveRecord::Migration[8.0]
+  def up
+    add_column :invoices, :expected_finalization_date, :date
+
+    safety_assured do
+      execute <<-SQL
+        UPDATE invoices
+        SET expected_finalization_date = issuing_date
+      SQL
+    end
```
julienbourdeau: I could be wrong but I feel like this would create a lock on the `invoices` table 🤔 

This is a pretty sensitive table but since the update is simple, maybe it goes fast enough.

In the past we did this kind of migration via a job (like [here](https://github.com/getlago/lago-api/pull/3633/files#diff-e71c9b23a67f49a4aaa0a895d1eff23070ef999c8c561d46ec2940e0167e41b8)) but the job can't be triggered in the db migration because the code with the job is usually not deployed yet 😬 
We did it with a rake task and manual intervention.

I'm happy to get other opinion on this.

vankiru: As far as I remember, postgres uses row-level locks for updates, so there shouldn't be a table lock on `invoices` 🤔 

But if the table is pretty big, the update might still take a bit. I also would love another opinion on this too.

groyoh: I may have the answer to this. AFAIk Postgres does use row-level locks BUT by default it will execute your query in a transaction (this is called `autocommit`). Any row-level lock acquired during that transaction will be held until the end of the transaction. So if you're iteration on all invoices, it will acquire a lock on one invoice at a time but hold all the locks until the command complete.

Since invoices are rarely updated, it could be fine as there shouldn't be any concurrent update. But I'm sure @jdenquin would tell us to err on the safe-side and run in batch. 😄

nepalez: There IS a problem here.

It is true that PostgreSQL would take row-level locks for updates and won't escalate it to page- or table-level. But all these locks will be kept until the end of the transaction, isn't it? In which case if the `invoices` table has, let's say, 1M rows, the database will create and finally hold 1M row-level locks, as @groyoh noticed above :man_shrugging:

It can be even more painful if there is a (hopefully rare) case when some code wanted to update just 2 invoices but in the order different from what the query uses. Well, here we'll bump into the deadlock, and lucky us if the shorter transaction will be dropped by the database. Otherwise the update will start from the beginning.

vankiru: Thanks everyone for the answers! Let's play it safe – I'll rewrite the migration into a rake task that updates everything in batches 😄

vankiru: Moved the data migration to a rake task and added a few safeguards to the code so nothing breaks while the task is running.

PR #4672, opened 2025-12-02
Multi payments attach invoice sections

mariohd:
This PR allows InvoiceCustomSections to be attached to Subscription and Wallets. 

[X] API
[X] GraphQL

### DIFF
--- a/app/graphql/types/wallets/create_input.rb
+++ b/app/graphql/types/wallets/create_input.rb
@@ -22,6 +22,8 @@ module Types
       argument :recurring_transaction_rules, [Types::Wallets::RecurringTransactionRules::CreateInput], required: false
       argument :transaction_name, String, required: false
 
+      argument :invoice_custom_section, Types::InvoiceCustomSections::ReferenceInput, required: false
+
       argument :applies_to, Types::Wallets::AppliesToInput, required: false
 
       argument :payment_method, Types::PaymentMethods::ReferenceInput, required: false
--- a/app/graphql/types/wallets/object.rb
+++ b/app/graphql/types/wallets/object.rb
@@ -41,6 +41,9 @@ module Types
       field :paid_top_up_min_amount_cents, GraphQL::Types::BigInt, null: true
       field :paid_top_up_min_credits, GraphQL::Types::BigInt, null: true
 
+      field :selected_invoice_custom_sections, [Types::InvoiceCustomSections::Object], null: true
+      field :skip_invoice_custom_sections, Boolean
+
       field :applies_to, Types::Wallets::AppliesTo, null: true, method: :itself
 
       field :created_at, GraphQL::Types::ISO8601DateTime, null: false
--- a/app/graphql/types/wallets/update_input.rb
+++ b/app/graphql/types/wallets/update_input.rb
@@ -14,6 +14,7 @@ module Types
       argument :paid_top_up_max_amount_cents, GraphQL::Types::BigInt, required: false
       argument :paid_top_up_min_amount_cents, GraphQL::Types::BigInt, required: false
 
+      argument :invoice_custom_section, Types::InvoiceCustomSections::ReferenceInput, required: false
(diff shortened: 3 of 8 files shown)


app/graphql/types/subscriptions/object.rb
```diff
@@ -26,6 +26,9 @@ class Object < Types::BaseObject
       field :current_billing_period_ending_at, GraphQL::Types::ISO8601DateTime
       field :current_billing_period_started_at, GraphQL::Types::ISO8601DateTime
 
+      field :selected_invoice_custom_sections, [Types::InvoiceCustomSections::Object], null: true
+      field :skip_invoice_custom_sections, Boolean, null: true
```
lovrocolic: I guess it can only be true or false

app/services/invoice_custom_sections/attach_to_resource_service.rb
```diff
@@ -13,10 +13,10 @@ def call
       return result unless params.key?(:invoice_custom_section)
 
       ActiveRecord::Base.transaction do
-        if skip_flag.present?
```
lovrocolic, line 16: Why is this change needed?

mariohd: Here it can be 3 states, `true`, `false` and `nil`. 

```
lago-api(dev)> true.present?
=> true
lago-api(dev)> false.present?
=> false
lago-api(dev)> nil.present?
=> false
```

the 2nd scenario, `false` value IS PRESENT (`skip_invoice_custom_sections: false`), but it considers as if the flag was not sent, calling `handle_implicit_skip_flag` wrongly.  

```
lago-api(dev)> true.nil?
=> false
lago-api(dev)> false.nil?
=> false
lago-api(dev)> nil.nil?
=> true
```

changing to `nil?`, will only call `handle_implicit_skip_flag` when the flag is `nil` (`skip_invoice_custom_sections: null` or not sent)

app/services/subscriptions/create_service.rb
```diff
@@ -43,7 +43,11 @@ def call
             .or(editable_subscriptions.where(external_id:))
             .first
 
-          result.subscription = handle_subscription
+          subscription = handle_subscription
```
lovrocolic, line 46: `downgrade_subscription` returns old subscription but we actually need to attach custom sections on the new subscription

app/services/invoice_custom_sections/attach_to_resource_service.rb
lovrocolic: hmm okay, but from the code it seems the end result will be the same, no? I mean if we send `skip_invoice_custom_sections: false`, end result will be the same

mariohd: if the previously stored is `true`, it would not work. 

e.g
record `skip_invoice_custom_sections: true`
param `skip_invoice_custom_sections: false`

`false.present?` will be `false`, we will call `handle_implicit_skip_flag` and [in the first line it will early return](https://github.com/getlago/lago-api/blob/main/app/services/invoice_custom_sections/attach_to_resource_service.rb#L51), doing nothing.

lovrocolic: Oh I see, nice catch 🙌

app/graphql/types/wallets/create_input.rb
```diff
@@ -22,6 +22,8 @@ class CreateInput < Types::BaseInputObject
       argument :recurring_transaction_rules, [Types::Wallets::RecurringTransactionRules::CreateInput], required: false
```
lovrocolic, line 22: It seems we miss `argument :invoice_custom_section, Types::InvoiceCustomSections::ReferenceInput, required: false` in wallet update input

mariohd: Totally right!! 

Excellent catch!

lovrocolic (approved):

PR #4673, opened 2025-12-03
feat(multiple-payment-methods): update attach cs to invoice logic

lovrocolic:
## Context

With multiple payment methods feature, it will be possible to attach payment method to different resources like subscription, wallet, wallet transaction, or one-off invoice

## Description

This PR prepares logic that attaches custom sections on the invoice.

Before custom sections were copied from customer or billing entity to the invoice. To support new feature, the logic needs to  be extended so that it support custom sections from other resources as well

### DIFF
--- a/app/services/invoices/apply_invoice_custom_sections_service.rb
+++ b/app/services/invoices/apply_invoice_custom_sections_service.rb
@@ -56,7 +56,7 @@ module Invoices
 
     def resource_has_custom_sections?
       return false unless resource
-      return false unless resource&.respond_to?(:selected_invoice_custom_sections)
+      return false unless resource.respond_to?(:selected_invoice_custom_sections)
       return false if resource.skip_invoice_custom_sections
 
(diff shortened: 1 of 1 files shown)


app/services/invoices/apply_invoice_custom_sections_service.rb
```diff
@@ -30,6 +32,38 @@ def call
...
+      else
+        customer.configurable_invoice_custom_sections
+      end
+
+      manual_sections | customer.system_generated_invoice_custom_sections
+    end
+
+    def resource_has_custom_sections?
+      return false unless resource
+      return false unless resource&.respond_to?(:selected_invoice_custom_sections)
```
mariohd: You already checked for `resource` on line before this

```suggestion
      return false unless resource.respond_to?(:selected_invoice_custom_sections)
```

lovrocolic: updated

PR #4675, opened 2025-12-03
feat(usage_threshold): Attach UsageThreshold to subscription model

julienbourdeau:
## Context

Overriding UsageThreshold on a subscription, will no longer create a plan override.

## Description

This PR allows usage_thresholds to be attached directly to a subscription, not just a plan. This is similar to `Entitlement::Entitlement`.

This is only the ground work to update the model and database. Nothing is attached to subscription yet.

### DIFF
--- a/app/services/plans/update_usage_thresholds_service.rb
+++ b/app/services/plans/update_usage_thresholds_service.rb
@@ -15,7 +15,7 @@ module Plans
       return result unless plan.organization.progressive_billing_enabled?
 
       if usage_thresholds_params.empty?
-        plan.usage_thresholds.discard_all
+        plan.usage_thresholds.update_all(deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
       else
         process_usage_thresholds
         LifetimeUsages::FlagRefreshFromPlanUpdateJob.perform_later(plan)


app/services/usage_thresholds/override_service.rb
```diff
@@ -28,7 +28,6 @@ def call
       result
     rescue ActiveRecord::RecordInvalid => e
       result.record_validation_failure!(record: e.record)
-      result.fail_with_error!
```
julienbourdeau, line 31: Notice that this had never worked because it raised a argument error.

app/models/usage_threshold.rb
```diff
@@ -7,16 +7,18 @@ class UsageThreshold < ApplicationRecord
...
   has_many :invoices, through: :applied_usage_thresholds
 
-  monetize :amount_cents, with_currency: ->(threshold) { threshold.plan.amount_currency }
+  monetize :amount_cents, with_currency: ->(threshold) { threshold.currency }
 
   validates :amount_cents, numericality: {greater_than: 0}
-  validates :amount_cents, uniqueness: {scope: %i[plan_id recurring deleted_at]}, if: -> { deleted_at.nil? }
-  validates :recurring, uniqueness: {scope: %i[plan_id deleted_at]}, if: -> { recurring? && deleted_at.nil? }
+  validates :amount_cents, uniqueness: {scope: %i[plan_id recurring deleted_at]}, if: -> { deleted_at.nil? && subscription_id.nil? }
+  validates :recurring, uniqueness: {scope: %i[plan_id deleted_at]}, if: -> { recurring? && deleted_at.nil? && subscription_id.nil? }
```
julienbourdeau, line 20: I intend to get rid of these uniqueness rails validation (which triggers query) when I rework the service.

app/services/usage_thresholds/override_service.rb
vincent-pochet: I guess the intended behavior was use `raise_if_error!`. Not sure why to be honnest

db/migrate/20251210133246_change_indexes_for_usage_thresholds.rb
```diff
@@ -0,0 +1,69 @@
...
+  def change
+    safety_assured do
+      add_reference :usage_thresholds, :subscription,
+        foreign_key: true,
+        index: {algorithm: :concurrently},
+        type: :uuid,
+        if_not_exists: true
+    end
+
+    remove_index :usage_thresholds,
```
vincent-pochet, line 7: Looks good but since this migration is doing a lot, I would just suggest first add the new column, and then redefine the indexes in a second migration

julienbourdeau: splitted in 2

PR #4678, opened 2025-12-04
feat(wallets): Support multiple wallets

floganz:
## Context

We’re introducing the priority field to wallets to support custom execution or display ordering. This PR builds on the base work from https://github.com/getlago/lago-api/pull/4391, which added the field to the Wallet model and exposed it via the API.

The goal of this PR is to implement the logic that uses the priority value when handling wallets — for example, when determining which wallets to apply or process first.

## Description

This PR includes:
• Logic to order wallets by priority (and created_at as a secondary key).
• Ensures active_wallets_in_application_order returns wallets in the correct priority sequence.
• Updates any internal wallet processing to respect the defined order.

This builds directly on the foundation laid in PR https://github.com/getlago/lago-api/pull/4391 and assumes the priority field and its validations are already in place.

### DIFF
--- a/app/services/credits/applied_prepaid_credits_service.rb
+++ b/app/services/credits/applied_prepaid_credits_service.rb
@@ -35,12 +35,8 @@ module Credits
 
           ordered_remaining_amounts.each do |fee_key, remaining_amount|
             next if remaining_amount <= 0
-            target_match = wallet_targets_array.include?(fee_key)
-            type_match = wallet_types_array.include?(fee_key.first)
-            unrestricted_wallet = wallet_targets_array.empty? && wallet_types_array.empty?
-            should_apply_wallet_on_this_fee = target_match || type_match || unrestricted_wallet
 
-            next unless should_apply_wallet_on_this_fee
+            next unless applicable_fee?(fee_key:, targets: wallet_targets_array, types: wallet_types_array)
 
             used_amount = wallet_fee_transactions.sum { |t| t[:amount_cents] }
             remaining_wallet_balance = wallet.balance_cents - used_amount
@@ -145,6 +141,14 @@ module Credits
 
         raise
       end
+    end
+
+    def applicable_fee?(fee_key:, targets:, types:)
+      target_match = targets.include?(fee_key)
+      type_match = types.include?(fee_key.first)
+      unrestricted_wallet = targets.empty? && types.empty?
+
+      target_match || type_match || unrestricted_wallet
     end
   end
 end
--- a/app/services/wallets/build_allocation_rules_service.rb
+++ b/app/services/wallets/build_allocation_rules_service.rb
@@ -3,6 +3,8 @@
 # app/services/wallets/build_allocation_rules_service.rb
 module Wallets
   class BuildAllocationRulesService < BaseService
+    Result = BaseResult[:allocation_rules]
+
     def initialize(customer:)
       @customer = customer
       super
--- a/app/services/wallets/find_applicable_on_fees_service.rb
+++ b/app/services/wallets/find_applicable_on_fees_service.rb
@@ -2,6 +2,8 @@
 
 module Wallets
   class FindApplicableOnFeesService < BaseService
+    Result = BaseResult[:top_priority_wallet]
+
     def initialize(allocation_rules:, fee:)
       @allocation_rules = allocation_rules
       @fee = fee
@@ -28,7 +30,7 @@ module Wallets
     attr_reader :allocation_rules, :fee
 
     def result_with(wallets)
-      result.applicable_wallets = wallets.first
+      result.top_priority_wallet = wallets.first
       result
     end
   end
--- a/app/services/wallets/balance/refresh_ongoing_service.rb
+++ b/app/services/wallets/balance/refresh_ongoing_service.rb
@@ -103,7 +103,7 @@ module Wallets
 
           applicable_wallets = Wallets::FindApplicableOnFeesService
             .call!(allocation_rules:, fee:)
-            .applicable_wallets
+            .top_priority_wallet
 
           fee_wallet[key] = applicable_wallets.presence
         end
--- a/app/services/wallets/balance/refresh_ongoing_usage_service.rb
+++ b/app/services/wallets/balance/refresh_ongoing_usage_service.rb
@@ -80,7 +80,7 @@ module Wallets
 
           applicable_wallets = Wallets::FindApplicableOnFeesService
             .call!(allocation_rules:, fee:)
-            .applicable_wallets
+            .top_priority_wallet
 
           fee_wallet[key] = applicable_wallets.presence
         end
--- a/spec/services/wallets/find_applicable_on_fees_service_spec.rb
+++ b/spec/services/wallets/find_applicable_on_fees_service_spec.rb
@@ -31,7 +31,7 @@ RSpec.describe Wallets::FindApplicableOnFeesService do
 
         it "returns matching by billable metric wallet" do
           expect(result).to be_success
-          expect(result.applicable_wallets).to eq matching_wallet_id
+          expect(result.top_priority_wallet).to eq matching_wallet_id
         end
       end
 
@@ -41,7 +41,7 @@ RSpec.describe Wallets::FindApplicableOnFeesService do
 
         it "returns matching by fee type wallet" do
           expect(result).to be_success
-          expect(result.applicable_wallets).to eq matching_wallet_id
+          expect(result.top_priority_wallet).to eq matching_wallet_id
         end
       end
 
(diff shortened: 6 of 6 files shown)


app/services/invoices/progressive_billing_service.rb
```diff
@@ -146,13 +146,16 @@ def create_credit_note_credit
...
     def create_applied_prepaid_credit
-      wallets = subscription.customer.wallets.active.with_positive_balance
+      wallets = subscription.customer.wallets.active.includes(:wallet_targets)
+        .with_positive_balance.in_application_order
 
       return if wallets.none?
       return unless invoice.total_amount_cents.positive?
 
-      prepaid_credit_result = Credits::AppliedPrepaidCreditsService.call!(invoice:, wallets:)
+      wallets = subscription.customer.active_wallets_in_application_order
```
brunomiguelpinto: we dont need to grab wallets again

app/services/invoices/progressive_billing_service.rb
```diff
@@ -146,13 +146,16 @@ def create_credit_note_credit
...
+      return unless wallets.any? { |w| w.balance.positive? }
```
brunomiguelpinto: This should be covered by the scope on the frist call

app/services/wallets/validate_service.rb
```diff
@@ -2,7 +2,7 @@
 
 module Wallets
   class ValidateService < BaseValidator
-    MAXIMUM_WALLETS_PER_CUSTOMER = 1
+    MAXIMUM_WALLETS_PER_CUSTOMER = 5
```
brunomiguelpinto: If we merge, this still needs to be 1

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -6,52 +6,117 @@ class AppliedPrepaidCreditsService < BaseService
 
     def initialize(invoice:, wallets:, max_wallet_decrease_attempts: DEFAULT_MAX_WALLET_DECREASE_ATTEMPTS)
       @invoice = invoice
-      @wallet = wallets.first
+      @wallets = wallets
```
annvelents, line 9: do we believe that wallets will always be sorted by the priority or do we want to reorder them here?

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -6,52 +6,117 @@ class AppliedPrepaidCreditsService < BaseService
...
+          end
+
+          total_amount_cents = wallet_fee_transactions.sum { |t| t[:amount_cents] }
+          next if total_amount_cents <= 0
+
+          wallet_transaction = create_wallet_transaction(wallet, total_amount_cents)
+          amount_cents = wallet_transaction.amount_cents
+
+          with_optimistic_lock_retry(wallet) do
+            Wallets::Balance::DecreaseService.call(wallet:, wallet_transaction:, skip_refresh: true)
```
annvelents, line 62: skip_refresh - noice!

app/services/wallets/build_allocation_rules_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    #   unrestricted  = [wallet1, wallet4]
+    def call
+      bm_map = Hash.new { |h, k| h[k] = [] }
+      type_map = Hash.new { |h, k| h[k] = [] }
+      unrestricted = []
+
+      wallets.each do |wallet|
+        if wallet.wallet_targets.any?
+          handle_billable_metric_wallet(wallet, bm_map, type_map, unrestricted)
+        elsif wallet.allowed_fee_types.present?
```
annvelents, line 44: is it my imagination or was it possible that a wallet is restricted to billable_metric AND a fee_type?

annvelents: just checked: yes, a wallet can have a limit of a specific billable metric and a fee_type

app/services/wallets/find_applicable_on_fees_service.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Wallets
+  class FindApplicableOnFeesService < BaseService
```
annvelents, line 4: Love this service!

app/services/wallets/find_applicable_on_fees_service.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      result_with([])
+    end
+
+    private
+
+    attr_reader :allocation_rules, :fee
+
+    def result_with(wallets)
+      result.applicable_wallets = wallets.first
```
annvelents: ah, one question here: do you want it to be a "top_priority_wallet" or something, because now I would expect to have all applicable wallets to be returned

app/services/credits/applied_prepaid_credits_service.rb
floganz: Exactly! We'll do some improvements after we merge this.
It's very fragile approach to expect that we always receive that array in right order. Instead we're going to remove that attribute at all, passing only `invoice` to this service. And service will fetch wallets inside with right order.

app/services/wallets/build_allocation_rules_service.rb
floganz: Will forward this question to product team and we'll address it in next PR, thanks for point at it.

app/services/wallets/find_applicable_on_fees_service.rb
floganz: All credits for this going to @brunomiguelpinto!

app/services/wallets/build_allocation_rules_service.rb
floganz: @brunomiguelpinto Maybe you have context for this question?

brunomiguelpinto: For the ongoing balance we only consider charges 

And if we put a specific charge we don't apply to any other charge even if we have type charge on the wallet 


So we can simplify this code a little bit but let's do it in another pr this one is quite big already

app/services/wallets/find_applicable_on_fees_service.rb
brunomiguelpinto: Yes just to top priority one we can change the name @floganz

app/services/wallets/build_allocation_rules_service.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+# app/services/wallets/build_allocation_rules_service.rb
+module Wallets
+  class BuildAllocationRulesService < BaseService
+    def initialize(customer:)
```
vincent-pochet, line 8: Could you define the `Result` type 🙏

app/services/wallets/build_allocation_rules_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    private
+
+    attr_reader :customer
+
+    def add_unique(array, id)
+      array << id unless array.include?(id)
+    end
+
+    # Add multiple items uniquely to target array
+    def add_all_unique(target_array, source_array)
```
vincent-pochet, line 64: I'm wondering if it can be a bit more efficient to just call `uniq` on the arrays when assigning to the result object, rather than looking for the presence in the value each time

app/services/wallets/find_applicable_on_fees_service.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Wallets
+  class FindApplicableOnFeesService < BaseService
+    def initialize(allocation_rules:, fee:)
```
vincent-pochet, line 7: Same comment regarding the `Result` :)

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -6,52 +6,117 @@ class AppliedPrepaidCreditsService < BaseService
...
+            end
+          end
+          wallet_types_array = wallet.allowed_fee_types
+
+          ordered_remaining_amounts.each do |fee_key, remaining_amount|
+            next if remaining_amount <= 0
+            target_match = wallet_targets_array.include?(fee_key)
+            type_match = wallet_types_array.include?(fee_key.first)
+            unrestricted_wallet = wallet_targets_array.empty? && wallet_types_array.empty?
+            should_apply_wallet_on_this_fee = target_match || type_match || unrestricted_wallet
```
vincent-pochet: Since this call method is pretty massive, I'm wondering if we could extract some blocks like this one is sub-method, just to ease with readability

app/services/wallets/build_allocation_rules_service.rb
floganz: Yes, it would be.
Planning to make that refactor in following PR together with [this improvement](https://github.com/getlago/lago-api/pull/4678#discussion_r2592438698).

app/services/credits/applied_prepaid_credits_service.rb
floganz: Extracted it for now in a method but imo I think we should refactor this into a dedicated service at some point where it does not have time pressure.

PR #4684, opened 2025-12-05
feat(issuing-date-6): Use issuing date settings in Invoices::ProviderTaxes::PullTaxesAndApplyService

vankiru:
## Roadmap Task

## Context

## Description

The PR updates the `Invoices::ProviderTaxes::PullTaxesAndApplyService` service to use the issuing date anchor setting. It's more in line with the service original logic of ignoring the grace period.

### DIFF
--- a/app/services/invoices/provider_taxes/pull_taxes_and_apply_service.rb
+++ b/app/services/invoices/provider_taxes/pull_taxes_and_apply_service.rb
@@ -115,7 +115,14 @@ module Invoices
       end
 
       def issuing_date
-        @issuing_date ||= Time.current.in_time_zone(customer.applicable_timezone).to_date
+        return @issuing_date if defined?(@issuing_date)
+
+        date = Time.current.in_time_zone(customer.applicable_timezone).to_date
+        recurring = invoice.invoice_subscriptions.first&.recurring?
+
+        issuing_date_service = Invoices::IssuingDateService.new(customer_settings: customer, recurring:)
+
+        @issuing_date = date + issuing_date_service.tax_issuing_date_adjustment.days
       end
 
       def payment_due_date
--- a/app/services/invoices/issuing_date_service.rb
+++ b/app/services/invoices/issuing_date_service.rb
@@ -14,6 +14,12 @@ module Invoices
       send("#{anchor}_#{adjustment}")
     end
 
+    def tax_issuing_date_adjustment
+      return 0 unless recurring
+
+      (anchor == "current_period_end") ? -1 : 0
+    end
+
     def grace_period
       customer_settings[:invoice_grace_period] || billing_entity_settings[:invoice_grace_period] || 0
     end
--- a/spec/services/invoices/issuing_date_service_spec.rb
+++ b/spec/services/invoices/issuing_date_service_spec.rb
@@ -131,6 +131,107 @@ RSpec.describe Invoices::IssuingDateService do
     end
   end
 
+  describe "#tax_issuing_date_adjustment" do
+    let(:recurring) { true }
+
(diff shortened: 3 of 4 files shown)


app/services/invoices/provider_taxes/pull_taxes_and_apply_service.rb
```diff
@@ -115,7 +115,14 @@ def create_applied_prepaid_credit
...
       def issuing_date
-        @issuing_date ||= Time.current.in_time_zone(customer.applicable_timezone).to_date
+        return @issuing_date if defined?(@issuing_date)
+
+        date = Time.current.in_time_zone(customer.applicable_timezone).to_date
+        recurring = invoice.invoice_subscriptions.first&.recurring?
+
+        issuing_date_service = Invoices::IssuingDateService.new(customer_settings: customer, recurring:)
+
+        @issuing_date = date + issuing_date_service.tax_issuing_date_adjustment.days
```
julienbourdeau, line 125: I was wondering if the issue date could be left untouched here 🤔 

At this point, unlike the refresh, it's an invoice that has been created, the fees are already set but for some reason we couldn't get the taxes from the external service.
I think in this case, the invoice was created correctly and the issuing date is already set.

Am I missing something?

vankiru: I'm not totally sure about removing this – it was added a year ago in [this commit ](https://github.com/getlago/lago-api/commit/cad0daf35c3b3572b4e6bb0203879a20c36406f1) by @lovrocolic. I think we'll need his input to figure out what to do with it.

lovrocolic: I think we need to set issuing date in this service.

It is true for recurring invoices that issuing date is set before reaching this service (e.g. in generating service) and this could be unnecessary.
However, this service is also called when we finalize draft invoice. In this case, if invoice is finalized manually before grece period ends + tax error is retried, this is the final code that is executed on such invoice. I think that issuing date + due date are not set on correct date before this service got called. Maybe we can refactor it and check all use cases, but for now I think we should keep setting issuing date here (and due date).

vankiru: If this mainly matters when we manually finalize invoices, then I don't think we need the changes in this PR. Issuing date settings don't apply when an invoice is manually finalized, and we just set the date to the time it gets finalized. So I believe the logic here should stay the same.

@julienbourdeau @lovrocolic wdyt? 🤔

lovrocolic: Make sense. I am just not sure if manual finalizing is the only use-case why we added issuing date here :/ It was long time ago when the change was made on my side

vankiru: After looking through the code and the invoice workflow again, I realized we need to set the issuing date in `PullTaxesAndApplyService` and `RefreshDraftAndFinalizeService` when finalizing an invoice..

When we have to fetch taxes, the invoice isn't finalized until all taxes are pulled in `PullTaxesAndApplyService` – and that can take a while if we hit API limits. So during finalization we can't show the final issuing/payment due dates yet, which means we shouldn't set them in `RefreshDraftAndFinalizeService` for this case.

If we don't need to pull taxes, then we can just set issuing/payment due dates in `RefreshDraftAndFinalizeService`.

Closing this PR in favor of this change: https://github.com/getlago/lago-api/pull/4624/commits/917afaf5969d5c210ad0610a360e84b7c991b0ab

PR #4685, opened 2025-12-05
Multi payments attach sections 2

mariohd:
This PR adds the endpoints for RecurringTransactionRules and WalletTransactions. 

Review commit by commit so you can have a more defined context for each part of the feature.

### DIFF
--- a/app/graphql/types/wallet_transactions/create_input.rb
+++ b/app/graphql/types/wallet_transactions/create_input.rb
@@ -9,6 +9,7 @@ module Types
 
       argument :granted_credits, String, required: false
       argument :ignore_paid_top_up_limits, Boolean, required: false
+      argument :invoice_custom_section, Types::InvoiceCustomSections::ReferenceInput, required: false
       argument :invoice_requires_successful_payment, Boolean, required: false
       argument :metadata, [Types::WalletTransactions::MetadataInput], required: false
       argument :name, String, required: false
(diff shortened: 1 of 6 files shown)


app/services/invoices/apply_invoice_custom_sections_service.rb
```diff
@@ -2,18 +2,28 @@
...
+            code: custom_section.code,
+            details: custom_section.details,
+            display_name: custom_section.display_name,
+            name: custom_section.name
+          )
+        end
+      end
 
-      customer.applicable_invoice_custom_sections.each do |custom_section|
+      extra_sections.each do |custom_section|
```
lovrocolic: AFAIK, custom sections defined in the one-off form, override customer selection. We don't join them

app/services/wallets/recurring_transaction_rules/update_service.rb
```diff
@@ -32,6 +32,11 @@ def call
           recurring_rule = wallet.recurring_transaction_rules.active.find_by(id: lago_id)
 
           if recurring_rule
+            if rule_attributes.key?(:invoice_custom_section)
+              InvoiceCustomSections::AttachToResourceService.call(resource: recurring_rule, params: rule_attributes)
+              rule_attributes.delete(:invoice_custom_section)
```
lovrocolic, line 37: Why is this needed?

mariohd: it's needed because the next line uses the `rule_attributes` variable to update some attributes of recurring rule. 


```
recurring_rule.update!(rule_attributes)
```

PR #4691, opened 2025-12-08
misc(webhook): Sent invoice.payment_failure when retry without payment provider

vincent-pochet:
## Context

The `invoice.created` webhook was recently removed from the payment retry flow, because the invoice was not really created.

However, some customer have built a flow around this webhook message. They are collecting payments outside of Lago, but they are relying on the `invoice.created` webhook that was sent when an invoice was resent for collection. 

As in this case, no payment method is attached to the customers, the payment retry on an an invoice does not trigger any webhook message. 

## Description

This PR make sure we send the existing `invoice.payment_failure` webhook in this case

### DIFF
--- a/app/services/invoices/payments/retry_service.rb
+++ b/app/services/invoices/payments/retry_service.rb
@@ -20,7 +20,8 @@ module Invoices
           return result.not_allowed_failure!(code: "payment_processor_is_currently_handling_payment")
         end
 
-        Invoices::Payments::CreateService.call_async(invoice:)
+        create_result = Invoices::Payments::CreateService.call_async(invoice:)
+        deliver_webhook if create_result.payment_provider.nil?
 
         result.invoice = invoice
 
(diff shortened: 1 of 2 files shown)


app/services/invoices/payments/retry_service.rb
```diff
@@ -32,6 +33,12 @@ def call
       attr_reader :invoice
 
       delegate :customer, to: :invoice
+
+      def deliver_webhook
+        SendWebhookJob.perform_later(
+          "invoice.payment_failure", invoice, error_details: "No payment provider"
```
groyoh: I believe `error_details` is a hash in other services. Also maybe, we should use error codes:

```suggestion
          "invoice.payment_failure", invoice, error_details: { code: "customer_must_have_payment_provider" }
```

Or something like that.

vincent-pochet: You are right, updating

PR #4692, opened 2025-12-09
fix(forecasted-usage): Fallback to GraduatedService when we are forecasting amounts for ProratedGraduatedService

murparreira:
## Context

Prorated calculations considers event data that we don't have when forecasting units on the data side, so we fall back to a default approach of using `GraduatedService` for that.

## Description

Added a default param to the charge model class that checks if the factory has an `aggregator`. When we are forecasting, we do not have this `aggregator`, so it fallback to the `GraduatedService`.

### DIFF
--- a/app/services/charge_models/factory.rb
+++ b/app/services/charge_models/factory.rb
@@ -28,6 +28,10 @@ module ChargeModels
       end
     end
 
+    # The has_aggregator param determines whether to use prorated charge models.
+    # When forecasting (no aggregator available), prorated graduated charges fall back to
+    # the non-prorated GraduatedService since per-event aggregation data is not available.
+    # This allows forecasting to work for all charge types without failing on nil aggregator.
     def self.charge_model_class(chargeable:, has_aggregator: true)
       case chargeable.charge_model.to_sym
       when :standard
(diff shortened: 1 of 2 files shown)


app/services/charge_models/factory.rb
```diff
@@ -24,12 +28,12 @@ def self.new_instance(chargeable:, aggregation_result:, properties:, period_rati
       end
     end
 
-    def self.charge_model_class(chargeable:)
+    def self.charge_model_class(chargeable:, has_aggregator: true)
```
vincent-pochet, line 35: It looks good to me. Maybe could you just add a small comment explaining the why of this new argument

spec/services/charge_models/factory_spec.rb
```diff
@@ -40,6 +40,7 @@
 
         context "when charge is prorated" do
           let(:charge) { build(:graduated_charge, prorated: true) }
+          let(:aggregation_result) { BaseService::Result.new.tap { |r| r.aggregator = [BaseService::Result.new] } }
```
vincent-pochet, line 43: Maybe you can also add a new test case for the new condition

vincent-pochet (approved):

app/services/charge_models/factory.rb
murparreira: Sure, it is an edge case only for forecasting amounts.

spec/services/charge_models/factory_spec.rb
murparreira: Nice catch, forgot that!

PR #4693, opened 2025-12-09
fix(webhooks): race condition when deleting end-points

vincent-pochet:
## Description
In some rare situation, a `SendWebhookJob` is processed while a Webhook endpoint is removed from the database, this leads to a `ActiveRecord::InvalidForeignKey` error when we try to persist the webhook.

This PR simply rescues the error, logs it and continue the process of delivering the webhooks to other end-points


groyoh (approved):
Maybe you can add a test for that ?

I think should be possible by loading the organization and webhook endpoints and deleting it via DB in the test setup.

PR #4694, opened 2025-12-09
fix(integration): Fix customer integration creation concurrency

groyoh:
## Context

It may happen that a customer was updated in a short time after it was created. As creating integration customers is a long operation, by the time we receive the update we still haven't created the integration customer. Therefore we will schedule a second `IntegrationCustomers::CreateJob`.

## Description

This fixes it by adding job uniqueness until executed. Note that we kept the `integration_customer_params` in the lock key arguments to ensure we still raise an error if the customer update changed the integration customer settings.

I also used this change to rename `IntegrationCustomers::CreateOrUpdateService` to `IntegrationCustomers::CreateOrUpdateBatchService` for consistency.

### DIFF
--- a/app/jobs/integration_customers/create_job.rb
+++ b/app/jobs/integration_customers/create_job.rb
@@ -3,14 +3,23 @@
 module IntegrationCustomers
   class CreateJob < ApplicationJob
     include ConcurrencyThrottlable
+
     queue_as "integrations"
 
     retry_on LagoHttpClient::HttpError, wait: :polynomially_longer, attempts: 3
     retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
 
+    # It may happen that a customer was updated in a short time after it was created. As creating integration customers is a
+    # long operation, by the time we receive the update we still haven't created the integration customer. Therefore we will
+    # schedule a second `IntegrationCustomers::CreateJob`. This second job should be ignored if the first one is still
+    # running.
+    #
+    # Note that we kept the `integration_customer_params` in the lock key arguments to ensure we still raise an error if the
+    # update changes the integration customer settings.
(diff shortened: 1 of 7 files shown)


Wallet rounding - Issue 1.md
```diff
@@ -0,0 +1,92 @@
+# Wallet Ongoing Balance Rounding Bug - QA Reproduction Steps
+
```
ancorcruz: this one shouldn't be here.

app/jobs/integration_customers/create_job.rb
```diff
@@ -3,14 +3,23 @@
...
     retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
 
+    # It may happen that a customer was updated in a short time after it was created. As creating integration customers is a
+    # long operation, by the time we receive the update we still haven't created the integration customer. Therefore we will
+    # schedule a second `IntegrationCustomers::CreateJob`. This second job should be ignored if the first one is still
+    # running.
+    #
+    # Note that we kept the `integration_customer_params` in the lock key arguments to ensure we still raise an error if the
+    # update changes the integration customer settings.
+    unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
```
ancorcruz, line 19: does it mean if an update is done with different params the job will be queued. right? Isn't that still an issue?

Wallet rounding - Issue 1.md
groyoh: Ah yeah, right. 🙈

app/jobs/integration_customers/create_job.rb
groyoh: Yes, and it will be logged so we can investigate and take further actions. It's better than silently ignore IMO.

PR #4697, opened 2025-12-09
Fix wallet refresh for multiple wallets

brunomiguelpinto:
## Context
When wallets have restrictions (targeting specific billable metrics via wallet_targets or limiting to specific fee types via allowed_fee_types), the ongoing balance calculation was incorrectly including all customer fees rather than only the fees that would actually be deducted from that specific wallet.

This resulted in inaccurate ongoing balance projections for restricted wallets, as they were accounting for usage that would never actually consume their credits.

Description
Refactored wallet balance refresh to respect wallet allocation rules
Added scenario tests and unit tests validating correct balance calculations for wallets with billable metric targets, fee type restrictions, and unrestricted wallets

### DIFF
--- a/spec/scenarios/wallets/customer_wallets_balance_refresh_spec.rb
+++ b/spec/scenarios/wallets/customer_wallets_balance_refresh_spec.rb
@@ -41,7 +41,7 @@ describe "Use wallet's credits and recalculate balances", transaction: false do
 
   let(:wallet) { create(:wallet, wallet_attrs) }
   let(:wallet2) { create(:wallet, wallet_attrs) }
-  let(:wallet3) { create(:wallet, wallet_attrs.merge({ name:"wallet 3" })) }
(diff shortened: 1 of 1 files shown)


floganz (commented):
I've added small improvements to guardfile and minor cleanup in service.
Tests looks good, except 2 cases that are not currently covered:

If [here](https://github.com/getlago/lago-api/pull/4697/files#diff-47c91aa006d35233faa90544da3d78ceb254fbb6eada049782799a0342d96679R80) we remove `- fee.precise_coupons_amount_cents` all tests remains to be green. I think to test this we need have scenario when billing period changing like we did during QA session.

[Here](https://github.com/getlago/lago-api/pull/4697/files#diff-47c91aa006d35233faa90544da3d78ceb254fbb6eada049782799a0342d96679R63) we can add back `&& f.charge.invoiceable?` to condition and all will be green again.

I've also noticed that we have some commented code in tests like `# ingest_event(subscription, 10, another_billable_metric.code)` should we maybe remove them if we're not using it?

PR #4698, opened 2025-12-09
fix(plan): Fix plan creation and update bugs

groyoh:
## Context

When creating a plan with charges, two bugs are currently present:

- Sending a `charge` without `billable_metric_id` will result in an `undefined method 'custom_agg?' for nil (NoMethodError)` error.
- Sending a null `values` for a charge filter will result in an `undefined method 'each' for nil:NilClass (NoMethodError)` error. This also occurred for plan update.
- Sending a empty `values` for a charge filter will result in a filter without values. This doesn't actually match the behavior expected by our OpenAPI which marks the `values` as required. This also occurred for plan update.

## Description

The first bug occured because `Plans::CreateService` had a private `create_charge` method that didn't properly check that the `charge` was valid (billable metric present) before proceeding to create it. The charge creation logic was already present in `Charges::CreateService` with proper validations. This replaces the inline charge creation logic in `Plans::CreateService` with a call to `Charges::CreateService`.

The second bug occurred because the `ChargeFilters::CreateOrUpdateBatchService` service assumed that `filter_params[:values]` would always be present and iterable.

The last bug occurred because we didn't validate that `values` had at least one element.

### DIFF
--- a/app/services/charge_filters/create_or_update_batch_service.rb
+++ b/app/services/charge_filters/create_or_update_batch_service.rb
@@ -25,6 +25,8 @@ module ChargeFilters
         return result
       end
 
+      return result.single_validation_failure!(field: :values, error_code: "value_is_mandatory") if empty_filter_values?
+
       # We only care about order when you have less than 100 filters.
       touch = filters_params.size < 100
 
@@ -163,6 +165,10 @@ module ChargeFilters
         filter.properties.delete("pricing_group_keys")
         filter.properties.delete("grouped_by")
       end
+    end
+
+    def empty_filter_values?
+      filters_params.any? { |filter_param| filter_param[:values].blank? }
     end
   end
 end
--- a/app/services/plans/create_service.rb
+++ b/app/services/plans/create_service.rb
@@ -52,13 +52,8 @@ module Plans
         end
 
         if args[:charges].present?
-          args[:charges].each do |charge|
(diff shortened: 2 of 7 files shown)


app/services/charge_filters/create_or_update_batch_service.rb
```diff
@@ -25,6 +25,8 @@ def call
         return result
       end
 
+      return result.validation_failure!(errors: {values: ["value_is_mandatory"]}) if empty_filter_values?
```
vincent-pochet: single_validation_failure!(
```suggestion
      return result.single_validation_failure!!(field: :values, error_code: value_is_mandatory") if empty_filter_values?
```

app/services/plans/create_service.rb
```diff
@@ -52,13 +52,8 @@ def call
...
         if args[:charges].present?
-          args[:charges].each do |charge|
-            new_charge = create_charge(plan, charge)
-
-            if charge[:tax_codes].present?
-              taxes_result = Charges::ApplyTaxesService.call(charge: new_charge, tax_codes: charge[:tax_codes])
-              taxes_result.raise_if_error!
-            end
+          args[:charges].each do |charge_params|
+            Charges::CreateService.call!(plan:, params: charge_params)
```
ancorcruz, line 56: good catch removing duplicated code already on a service that should be the way to create the charges.

PR #4700, opened 2025-12-10
fix(invoice-preview):cover pending subscription that starts in the future

lovrocolic:
## Context

Invoice preview feature has not covered the case of pending subscription that starts in the future.

## Description

This PR covers described case


spec/services/invoices/preview/find_subscriptions_service_spec.rb
```diff
@@ -34,6 +34,18 @@
         end
       end
 
+      context "when subscription is pending" do
```
floganz, line 37: I'd suggest to add to this context another`it` that ensures original subscription not being changed.
You can copy one right above this context.

lovrocolic: Done

PR #4720, opened 2025-12-15
misc(prepaid-credits): Remove wallets parameter in apply prepaid credits service

floganz:
## Description

Adjust various services to eliminate reliance on passing `wallets` explicitly, instead leveraging internal wallet retrieval with correct order.
Update corresponding tests and specs for the new logic.

### DIFF
--- a/app/services/invoices/calculate_fees_service.rb
+++ b/app/services/invoices/calculate_fees_service.rb
@@ -345,7 +345,6 @@ module Invoices
         (subscription.terminated? && subscription.terminated_at > invoice.created_at)
     end
 
-
     def should_create_credit_note_credit?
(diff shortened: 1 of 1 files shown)


app/services/invoices/calculate_fees_service.rb
```diff
@@ -294,8 +294,7 @@ def should_create_charge_fees?(subscription)
     end
 
     def wallets
```
groyoh: I'm guessing we could get rid of the `wallets` from the calling services as well by having the guard within the `AppliedPrepaidCreditService`.

brunomiguelpinto: Yes

PR #4723, opened 2025-12-15
chore(wallet): Skip top-up scheduling on wallet creation if unnecessary

groyoh:
## Context

When creating a wallet, it is possible to skip the `paid_credits` and `granted_credits` fields. In such case, we'll still schedule the `WalletTransactions::CreateJob` even though it is not necessary.

## Description

This change will force the service to skip the job when unnecessary.

### DIFF
--- a/app/services/wallets/create_service.rb
+++ b/app/services/wallets/create_service.rb
@@ -72,20 +72,9 @@ module Wallets
 
       result.wallet = wallet
 
-      SendWebhookJob.perform_later("wallet.created", wallet)
+      SendWebhookJob.perform_after_commit("wallet.created", wallet)
 
-      WalletTransactions::CreateJob.perform_later(
-        organization_id:,
-        params: {
-          wallet_id: wallet.id,
-          paid_credits: params[:paid_credits],
-          granted_credits: params[:granted_credits],
-          source: :manual,
-          metadata: params[:transaction_metadata],
-          name: params[:transaction_name],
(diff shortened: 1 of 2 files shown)


app/services/wallets/create_service.rb
```diff
@@ -72,31 +72,49 @@ def call
...
         }
       )
+    end
 
-      result
-    rescue ActiveRecord::RecordInvalid => e
-      result.record_validation_failure!(record: e.record)
-    rescue BaseService::FailedResult => e
-      e.result
+    def positive_amount?(amount)
```
floganz, line 107: Do we need to rescue any exceptions here or all data already validated by this step?

spec/services/wallets/create_service_spec.rb
```diff
@@ -66,7 +66,59 @@
...
+            expect { result = create_service.call }.to have_enqueued_job(WalletTransactions::CreateJob).with(
+              organization_id: organization.id,
+              params: hash_including(
+                paid_credits: paid_credits,
+                granted_credits: granted_credits
+              )
+            )
+          else
+            expect { result = create_service.call }.not_to have_enqueued_job(WalletTransactions::CreateJob)
+
```
floganz: Do we need this empty line?

app/services/wallets/create_service.rb
groyoh: It is already validated by validator.

spec/services/wallets/create_service_spec.rb
groyoh: No, indeed.

PR #4725, opened 2025-12-16
misc(clickhouse): Add service to remove duplicated events

vincent-pochet:
## Context

An issue has been raised with the Clickhouse implementation, leading to a risk of duplicated events when the same transaction_id/timestamp are sent multiple time.

## Description

A fix to allow a full deduplication is ongoing but it's not ready yet. In the meantime, this PR is adding a small utility service to allow a manual cleanup of all duplicated events for a given subscription on a given billing period.

```ruby
Events::Stores::Clickhouse::CleanDuplicatedService.call!(subscription:)
```

It will remove all duplicated events except the last received one (ordered by `enriched_at`)

### DIFF
--- a/app/services/events/stores/clickhouse/clean_duplicated_service.rb
+++ b/app/services/events/stores/clickhouse/clean_duplicated_service.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+module Events
+  module Stores
+    module Clickhouse
+      class CleanDuplicatedService < BaseService
+        Result = BaseResult
+
+        def initialize(subscription:, timestamp: Time.current)
+          @subscription = subscription
+          @timestamp = timestamp
+          super
+        end
+
+        def call
+          remove_duplicated_events
+          Subscriptions::ChargeCacheService.expire_for_subscription(subscription)
+
+          result
+        end
+
+        private
+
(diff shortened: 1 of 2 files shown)


rsempe (approved):

app/services/events/stores/clickhouse/clean_duplicated_service.rb
```diff
@@ -0,0 +1,86 @@
...
+        Result = BaseResult
+
+        def initialize(subscription:, timestamp: Time.current)
+          @subscription = subscription
+          @timestamp = timestamp
+          super
+        end
+
+        def call
+          return result.not_allowed_failure!(code: "no_clickhouse_event_store") unless organization.clickhouse_events_store?
```
julienbourdeau: I think we still send all events (including subscriptions on PG) to clickhouse so maybe we should allow this service to run 🤔

vincent-pochet: @julienbourdeau In case of duplication for PG organizations, the `Events::Create#call` will failed to store the event in PG and so, nothing will be sent to Clickhouse either.
I could still remove this condition as in the end, it will do nothing

PR #4729, opened 2025-12-16
feat(thresholds): expose subscription.applicable_usage_thresholds

julienbourdeau:
## Context 

Usage_thresholds are being extracted from plan override and attached to the subscription

## Description

**I'm describing what's going on below so someone can check that my logic is not flowed.** 

The goal is to introduce the new way of getting usage_thresholds with no breaking change between now and final release

This PR exposes a new `applicable_usage_thresholds` which represents the final state of what's used for this subscription.
Users will (soon) be able to start using this field and never look at `plan.usage_thresholds` anymore, which will deprecated and ultimately removed.

At the end, thesholds will be attached to the subscription or the parent plan, so this attribute will show:
- empty array of the (upcoming) `subscription.disable_progressive_billing` flag is set to true
- `subscription.usage_thresholds` if any (subscription-specific overrides)
- `(subscription.plan.parent || subscription.plan).usage_thresholds` for the thresholds from the parent plan

For now:
- we read subscription thresholds but they are not set yet so always empty
- then read the subscription plan (never reach for the parent) because they are the applicable thresholds
  - if plan is parent: great
  - if plan is child, it will contain the override thresholds OR a copy of the parent thresholds

This field is only added to `show` and `create` endpoint for now. Maybe update should be added as well 🤔

### DIFF
--- a/spec/requests/api/v1/subscriptions_controller_spec.rb
+++ b/spec/requests/api/v1/subscriptions_controller_spec.rb
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 RSpec.describe Api::V1::SubscriptionsController do
-  let(:organization) { create(:organization) }
+  let(:organization) { create(:organization, premium_integrations: %w[progressive_billing]) }
   let(:customer) { create(:customer, organization:) }
   let(:plan) { create(:plan, organization:, amount_cents: 500, description: "desc") }
   let(:commitment_invoice_display_name) { "Overriden minimum commitment name" }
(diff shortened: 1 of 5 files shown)


spec/requests/api/v1/subscriptions_controller_spec.rb
```diff
@@ -1246,12 +1261,19 @@ def test_termination(expected_on_termination_credit_note: nil, expected_on_termi
         external_id: subscription.external_id
       )
       expect(json[:subscription][:entitlements]).to contain_exactly({
-        code: "feature_1",
+        code: start_with("feature_"),
```
julienbourdeau, line 1264: this started returning `feature_2` and I have no idea why :/

app/models/subscription.rb
```diff
@@ -262,6 +262,14 @@ def billed_on_activation?
   def has_progressive_billing?
     plan.usage_thresholds.any?
```
groyoh: What about this ?

```suggestion
    applicable_usage_thresholds.any?
```

julienbourdeau: Yeah good point. It's probably best to do it now

PR #4730, opened 2025-12-16
feat(thresholds): Introduce `subscription.disable_progressive_billing` flag

julienbourdeau:
Notice that this is stacked on top of https://github.com/getlago/lago-api/pull/4729

The flag is set to false and there is no way to updated it yet (coming in main PR).

### DIFF
--- a/db/migrate/20251216100247_add_disable_progressive_billing_on_subscription.rb
+++ b/db/migrate/20251216100247_add_disable_progressive_billing_on_subscription.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDisableProgressiveBillingOnSubscription < ActiveRecord::Migration[8.0]
+  def change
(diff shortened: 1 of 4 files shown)


db/migrate/20251216100247_add_disable_progressive_billing_on_subscription.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDisableProgressiveBillingOnSubscription < ActiveRecord::Migration[8.0]
+  def change
+    add_column :subscriptions, :disable_progressive_billing, :boolean, default: false, null: false
```
groyoh: To stay consisten with the rest of the code base.

```suggestion
    add_column :subscriptions, :progressive_billing_disabled, :boolean, default: false, null: false
```

julienbourdeau: thank you very good point!

PR #4733, opened 2025-12-16
feat(fixed-charge): add support to invoice preview service

ancorcruz:
## Context

Invoice preview did not include fixed charges, making it impossible for users to see accurate invoice estimates when subscriptions included fixed charges. This affected preview accuracy for REST API consumers.

  ## Description

Added comprehensive fixed charge support to invoice preview functionality:

- Calculate fixed charge boundaries from DatesService with proper handling of fixed_charges_from_datetime and fixed_charges_to_datetime
- Include fixed charge fees in preview calculations using existing FixedChargeService with invoice_preview context
- Handle pay_in_advance vs pay_in_arrears billing timing correctly, excluding pay_in_arrears charges on first invoice and pay_in_advance charges on terminated subscriptions
- Respect subscription lifecycle states including starting subscriptions, terminated subscriptions, and upgrade/downgrade scenarios
- Fixed FeeSerializer to include date boundaries for fixed_charge fee type and properly serialize lago_fixed_charge_id
- Updated pay_in_advance serialization logic to handle both charge and fixed_charge fee types

The implementation mirrors CalculateFeesService patterns to ensure consistency between preview and actual invoice generation. Fixed charge events are created at subscription start and aggregation picks up the latest event for each billing period, following the recurring charge pattern.

Includes 10 comprehensive preview service tests covering all charge models (standard, graduated, volume) and subscription scenarios, plus 2 serializer tests for REST API v1 compatibility.

### DIFF
--- a/spec/services/invoices/preview_service_spec.rb
+++ b/spec/services/invoices/preview_service_spec.rb
@@ -388,7 +388,8 @@ RSpec.describe Invoices::PreviewService, cache: :memory do
             end
 
             context "with multiple fixed charges" do
-              let(:add_on2) { create(:add_on, organization:, amount_cents: 500) }
+              let(:add_on2) { create(:add_on, organization:) }
+              let(:add_on3) { create(:add_on, organization:) }
 
               let(:fixed_charge_advance) do
                 create(
@@ -412,10 +413,22 @@ RSpec.describe Invoices::PreviewService, cache: :memory do
                   properties: {amount: "5"}
                 )
               end
+              let(:fixed_charge_in_arrears) do
+                create(
+                  :fixed_charge,
+                  plan:,
+                  add_on: add_on3,
+                  charge_model: "standard",
+                  pay_in_advance: false,
+                  units: 1,
+                  properties: {amount: "25"}
+                )
+              end
 
               before do
                 fixed_charge_advance
                 fixed_charge_advance2
+                fixed_charge_in_arrears
 
                 event_timestamp = subscription.started_at + 1.second
                 create(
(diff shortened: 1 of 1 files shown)


app/services/invoices/preview_service.rb
```diff
@@ -313,5 +339,29 @@ def should_not_create_charge_fee?(charge, subscription)
         subscription.upgraded? &&
         charge.included_in_next_subscription?(subscription)
     end
+
+    def fixed_charge_boundaries_valid?(boundaries)
+      return false if boundaries.fixed_charges_from_datetime.nil?
```
annvelents, line 344: can we have `return false if boundaries.fixed_charges_from_datetime.nil? || boundaries.fixed_charges_to_datetime.nil?`
to have consistency with calculate fees service

spec/services/invoices/preview_service_spec.rb
```diff
@@ -222,6 +222,357 @@
...
+                  expect(fixed_charge_fees.size).to eq(1)
+
+                  fixed_charge_fee = fixed_charge_fees.first
+                  expect(fixed_charge_fee.fixed_charge).to eq(fixed_charge)
+                  expect(fixed_charge_fee.amount_cents).to eq(500) # $5 * 1 unit * 100
+                end
+              end
+            end
+
+            context "with multiple fixed charges" do
```
annvelents, line 390: if it's possible I'd ask also add a test for fixed charges with different payment schedule: pay in advance and pay in arrears

app/services/invoices/preview_service.rb
ancorcruz: okay, but it was already covered but split in two lines... one line for each date boundary, the result is the same 🤷

ancorcruz: I actually prefer this way as lines are shorter... what do you think?

PR #4740, opened 2025-12-17
Feat(fixed_charges): move fee dates logic into model and reuse it for graphql

annvelents:
## Context

On the API we already send actual boundaries of the fee, while in the graphql in fee we send subscription boundaries, but in the subscription we send boundaries of the current period and paid in advance period. Instead, we can always send correct boundaries on the fee

### DIFF
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -226,6 +226,57 @@ class Fee < ApplicationRecord
   def has_charge_filters?
     charge&.filters&.any?
   end
+
+  def date_boundaries
+    if charge? && !pay_in_advance? && charge.pay_in_advance?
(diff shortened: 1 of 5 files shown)


app/models/fee.rb
```diff
@@ -225,6 +225,55 @@ def add_on
...
+  def to_date
+    property = if charge?
+      "charges_to_datetime"
+    elsif fixed_charge?
+      "fixed_charges_to_datetime"
+    else
+      "to_datetime"
+    end
+    properties[property]&.to_datetime&.iso8601
+  end
```
ancorcruz, line 279: Lets make these two methods private as should not be called from outside, from/to_date depends on the logic defined in date_boundaries method.

PR #4741, opened 2025-12-17
feat(Custom roles & permissions) Data layer

nepalez:
# Roadmap Task

## Context

Currently, Lago supports a fixed set of membership roles (admin, manager, finance) with hardcoded permissions. Organizations cannot create custom roles tailored to their specific access control needs.

This is the first of three PRs implementing custom roles using the **expand-migrate-contract** pattern:

1. **This PR (Expand)**: Add new data structures and APIs without changing existing behavior
2. **Next PR (Migrate)**: Switch internal logic to use new system while keeping API backward-compatible
3. **Final PR (Contract)**: Remove legacy code and finalize the new API

This PR implements the Work Breakdown | Part 1.

## What This PR Does

Introduces the data layer foundation for custom roles without activating it:

- Creates new database tables (`roles`, `membership_roles`)
- Adds new GraphQL queries and mutations for role management
- Marks existing `role` field as deprecated (but fully functional)
- Seeds predefined roles (Admin, Manager, Finance) in the database

**Important**: The existing permission system continues to work unchanged. The new tables exist but are not yet used for permission checking.

## New Models

* `Role` - Represents both predefined system roles (Admin, Finance, Manager) and custom organization-specific roles

* `MembershipRole` - Join model connecting memberships to roles (many-to-many relationship)

### `roles` table

| Column | Type | Description |
|--------|------|-------------|
| `id` | uuid | Primary key |
| `organization_id` | uuid | NULL for predefined roles, organization ID for custom roles |
| `name` | string | Role name (unique per organization, case-insensitive) |
| `description` | string | Optional description |
| `permissions` | string[] | Array of permission keys |
| `admin` | boolean | Whether this is an admin role with full permissions |
| `deleted_at` | datetime | Soft delete timestamp |

* Check constraints ensuring data integrity (non-empty names, valid permission arrays)
* Unique index on role names per organization (case-insensitive)
* Database trigger preventing modification of predefined/deleted roles

### `membership_roles` table

| Column | Type | Description |
|--------|------|-------------|
| `id` | uuid | Primary key |
| `organization_id` | uuid | Organization reference |
| `membership_id` | uuid | Membership reference |
| `role_id` | uuid | Role reference |
| `deleted_at` | datetime | Soft delete timestamp |

* Composite foreign keys ensuring organizational consistency
* Unique constraint preventing duplicate role assignments

### `invites` table changes

* Adds `roles` string array column for future use

### New Query: `roles`

Returns all roles available for the organization (predefined + custom):

### New Mutations: Role CRUD

```graphql
mutation CreateRole {
  createRole(input: { name: "Reviewer", permissions: ["customers:view"] }) {
    id name permissions
  }
}

mutation UpdateRole {
  updateRole(input: { id: "...", name: "Updated Name" }) {
    id name
  }
}

mutation DestroyRole {
  destroyRole(input: { id: "..." }) {
    id
  }
}
```

### Deprecations (informational only)

* `Membership.role` field marked deprecated in favor of future `roles` field
* These fields continue to work exactly as before

## Backward Compatibility

**Full backward compatibility guaranteed:**

- All existing API endpoints work unchanged
- Legacy `role` enum field remains the source of truth
- Permission checking mechanism unchanged
- No migration of existing data required
- No UI changes required to deploy this PR

### DIFF
--- a/spec/models/membership_spec.rb
+++ b/spec/models/membership_spec.rb
@@ -59,7 +59,7 @@ RSpec.describe Membership do
       let(:membership) { create(:membership, roles: %i[admin]) }
 
       it "includes all existing permissions" do
-        expect(permissions_hash.keys).to contain_exactly(*Permission.permissions_hash.keys)
+        expect(permissions_hash.keys).to match_array(Permission.permissions_hash.keys)
       end
 
       it "returns all permissions as true" do
@@ -71,7 +71,7 @@ RSpec.describe Membership do
       let(:membership) { create(:membership, roles: %i[finance]) }
 
       it "includes all existing permissions" do
-        expect(permissions_hash.keys).to contain_exactly(*Permission.permissions_hash.keys)
+        expect(permissions_hash.keys).to match_array(Permission.permissions_hash.keys)
       end
 
       it "returns true for finance-specific permissions" do
@@ -87,7 +87,7 @@ RSpec.describe Membership do
       let(:membership) { create(:membership, roles: %i[manager]) }
 
       it "includes all existing permissions" do
-        expect(permissions_hash.keys).to contain_exactly(*Permission.permissions_hash.keys)
+        expect(permissions_hash.keys).to match_array(Permission.permissions_hash.keys)
       end
 
       it "returns true for manager-specific permissions" do
@@ -105,7 +105,7 @@ RSpec.describe Membership do
       let(:membership) { create(:membership, organization:, role:) }
 
       it "includes all existing permissions" do
-        expect(permissions_hash.keys).to contain_exactly(*Permission.permissions_hash.keys)
+        expect(permissions_hash.keys).to match_array(Permission.permissions_hash.keys)
       end
 
       it "excludes non-existing permissions" do
@@ -127,7 +127,7 @@ RSpec.describe Membership do
       let(:membership) { create(:membership, organization:, role:, roles: %i[finance]) }
 
       it "includes all existing permissions" do
-        expect(permissions_hash.keys).to contain_exactly(*Permission.permissions_hash.keys)
+        expect(permissions_hash.keys).to match_array(Permission.permissions_hash.keys)
       end
 
       it "excludes non-existing permissions" do


db/migrate/20251215174251_create_roles.rb
```diff
@@ -0,0 +1,149 @@
...
+
+class CreateRoles < ActiveRecord::Migration[8.0]
+  def up
+    create_table :roles, id: :uuid do |t|
+      t.string :name, null: false
+      t.boolean :admin, null: false, default: false
+      t.text :permissions, array: true, null: false, default: []
+      t.references :organization, type: :uuid
+      t.text :description
+      t.datetime :deleted_at, comment: "Soft deletion timestamp"
```
julienbourdeau: I'm in favor of removing this kind of comments

config/locales/en.yml
```diff
@@ -43,6 +43,11 @@ en:
             attributes:
               permissions:
                 forbidden_key: 'contains forbidden keys: %{keys}'
+          member_role:
+            attributes:
+              base:
+                last_admin_role: the last admin of the organization cannot be removed
+                last_member_role: the last role of the organization member cannot be removed
```
julienbourdeau:
```suggestion
                last_admin_role: last_admin_role
                last_member_role: last_member_role
```
We don't use english here, only code. (I know the one just above is a bad example)

app/models/role.rb
```diff
@@ -0,0 +1,48 @@
...
+    presence: true,
+    uniqueness: {case_sensitive: false, conditions: -> { where(deleted_at: nil) }, scope: :organization_id},
+    if: -> { organization_id && deleted_at.blank? }
+  validates :permissions, presence: true, if: :organization_id
+
+  private
+
+  RESERVED_NAMES = %w[admin finance manager].freeze
+
+  def name_is_available
```
julienbourdeau:
```suggestion
  def name_is_not_reserved
```
Nitpik but I'd like to see "reserved" in the name here

app/models/membership.rb
```diff
@@ -65,6 +66,7 @@ def permissions_hash
 #  index_memberships_on_organization_id              (organization_id)
 #  index_memberships_on_user_id                      (user_id)
 #  index_memberships_on_user_id_and_organization_id  (user_id,organization_id) UNIQUE WHERE (revoked_at IS NULL)
+#  membership_fk_idx                                 (id,organization_id) UNIQUE
```
julienbourdeau: Notice that all other index use `index_` or `idx_` as prefix, never as suffix. Same for Role model indexes.
I guess you're probably used to the opposite ^^

app/models/member_role.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+class MemberRole < ApplicationRecord
```
julienbourdeau: Thinking out loud here but I'm wondering if we should go with `MembershipRole` 🤔

app/models/member_role.rb
```diff
@@ -0,0 +1,61 @@
...
+  include Discard::Model
+  self.discard_column = :revoked_at
```
julienbourdeau: I understand the intent and I agree that `revoke` is a much better naming to communicate what is actually happening but ALL models (34 so far) are using `deleted_at` so I'm afraid it's too unexpected 🤔 

I'm happy to get more opinion on this.

app/models/member_role.rb
```diff
@@ -0,0 +1,61 @@
...
+  before_discard :forbid_last_member_role_discard
+
+  private
+
+  def forbid_last_admin_role_discard
+    return unless admin
+    return if organization.member_roles.admins.where.not(id:).exists?
+
+    errors.add(:base, :last_admin_role)
+    throw(:abort)
```
julienbourdeau: We don't do this anywhere else. adding an error should prevent saving already

db/migrate/20251221174733_create_membership_roles.rb
```diff
@@ -0,0 +1,79 @@
...
+            
+            RETURN NEW;
+          END;
+        $$ LANGUAGE plpgsql;
+
+        CREATE TRIGGER ensure_member_role_consistency
+        BEFORE INSERT OR UPDATE OR DELETE ON member_roles
+        FOR EACH ROW EXECUTE FUNCTION ensure_member_role_consistency();
+      SQL
+    end
```
julienbourdeau: I'm in favor of removing the entire function (and other `ensure_*_consistency` functions) and keeping the app responsible for this.

spec/models/role_spec.rb
```diff
@@ -0,0 +1,220 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Role do
```
julienbourdeau, line 5: To test soft deletion of new models, you can use `it { expect(described_class).to be_soft_deletable }`

db/migrate/20251215174251_create_roles.rb
```diff
@@ -0,0 +1,149 @@
...
+
+      t.index [:id, :admin], unique: true, name: "role_fk_idx"
+      t.index :admin,
+        unique: true,
+        where: "admin AND deleted_at IS NULL",
+        name: "unique_admin_role_idx"
+      t.index "organization_id, lower(name)",
+        unique: true,
+        where: "deleted_at IS NULL",
+        name: "unique_role_names_idx"
```
julienbourdeau:
```suggestion
        name: "index_unique_role_names_per_organization"
```

  We typically use `_per_organisation` or `_per_*` in the name

db/migrate/20251221174251_create_roles.rb
```diff
@@ -0,0 +1,149 @@
...
+                'billing_entities:invoices:update',
+                'billing_entities:taxes:view',
+                'billing_entities:taxes:update',
+                'billing_entities:emails:view',
+                'billing_entities:emails:update',
+                'billing_entities:dunning_campaigns:update',
+                'payment_requests:view',
+                'payment_requests:create',
+                'payments:view',
+                'payments:create'
```
julienbourdeau: I feel like this should live somewhere in the code, like `Permissions::FINANCE_PERMISSIONS_LIST`.

Adding a permission to this profile would be adding an entry to this array and we can have a function to reuse in db migration like below. I'd rather avoid updating arrays in SQL each time we modify permissions.

```rb
class Something < ActiveRecord::Migration[8.0]
  def change
    Permission.refresh_reserved_role_permissions!
  end
end
```

app/models/role.rb
nepalez: fixed

app/models/membership.rb
nepalez: D'oh! Once again, I should have rechecked.
(yeah, it is just the thing I was accustomed to)

app/models/member_role.rb
nepalez: it shouldn't because this is not a validation, but a callback
`throw(:abort)` is a standard way to stop execution inside callbacks

surely, I could turn this to validation with a special guard clauses catching discarding, but this would lose expressiveness of the syntax, isn't it?

```ruby
return unless deleted_at_was.blank?
return unless deleted_at.present?
```

app/models/member_role.rb
nepalez: fixed

Well, have to admit that you're right, I missed the fact it it so widespread. Initially I looked at the `revoke_at` for `Invite` and `Membership`, but after rechecking I recognized they aren't soft deletion-s by themselves.

In which case it is less surprising (no need to check the name for future devs)

app/models/member_role.rb
nepalez: fixed

config/locales/en.yml
nepalez: fixed

db/migrate/20251215174251_create_roles.rb
nepalez: Ahh, undercleaned!
fixed

db/migrate/20251215174251_create_roles.rb
nepalez: fixed

db/migrate/20251221174251_create_roles.rb
nepalez: I'm not sure this would be for better.

I see the reason for deployment, but I also have two reasons against this design:

* the first (strong) reason is that it makes any operation's logic more complex. well, we have one exclusions from the check of the role against the role.permissions -- admin, but the specifics of admin is to check literally nothing, so it is like an extreme, that doesn't add any complexity. But with a special lists it adds the complexity. Every time we should look at the different sources, which opens doors to bugs.

* the second (and much weaker) reason is that we already have a "natural" requirement to add permissions "pessimistically" -- new permissions should not be added to roles. Surely, it makes more sense for custom roles, but I suggest that adding new permissions to finances and management is not that often operation?

db/migrate/20251221174733_create_membership_roles.rb
nepalez: I somehow expected this comment ))

But the reason why I still went this way is that roles and permissions are about the security. IMO we should be very careful about the data itself here, not relying on the app's interfaces to prevent data leakage. That's why this trigger was created to care about the organization_id consistency first, and only then about the rest of invariants because why not?

spec/models/role_spec.rb
nepalez: fixed

db/migrate/20251221174251_create_roles.rb
```diff
@@ -0,0 +1,151 @@
...
+      t.check_constraint "NOT ('' = ANY(permissions))", name: "permissions_must_not_be_empty"
+      t.check_constraint "NOT (permissions::text ~ '([\\{,]:|::|:[,\\}])')", name: "permissions_has_no_empty_parts"
+
+      t.index [:id, :admin],
+        unique: true,
+        name: "index_roles_by_id_and_admin"
+      t.index :admin,
+        unique: true,
+        where: "admin AND deleted_at IS NULL",
+        name: "index_roles_by_unique_admin"
```
groyoh, line 25: So we want only one global role with `admin = true` ?

app/models/role.rb
groyoh, line 1: We may want to add length validation to name and description.

db/migrate/20251221174733_create_membership_roles.rb
```diff
@@ -0,0 +1,85 @@
...
+    unless index_exists?(:memberships, [:id, :organization_id], name: "index_memberships_by_id_and_organization")
+      add_index :memberships, [:id, :organization_id], unique: true, name: "index_memberships_by_id_and_organization", algorithm: :concurrently
+    end
+
+    unless table_exists?(:membership_roles)
+      create_table :membership_roles, id: :uuid do |t|
+        t.uuid :organization_id, null: false
+        t.uuid :membership_id, null: false
+        t.references :role, type: :uuid, null: false
+        t.boolean :admin, default: false, null: false
```
groyoh: Is there any reason to duplicate the `admin` field ?

app/models/member_role.rb
julienbourdeau: Good point. I totally missed that it wasn't validation!
Will we get a `ActiveRecord::InvalidRecord` exception here with errors.add + throw?
If so, it's all good 👍 

Otherwise, maybe we can introduce a new helper in `discarding?` if needed. I agree the expressiveness is lost.

db/migrate/20251221174251_create_roles.rb
julienbourdeau: > I suggest that adding new permissions to finances and management is not that often operation?
It's gonna be pretty often I think. A lot of features we add are for finance/manager users so we give them permissions to new featues.

In any case, I thikn this is fine, we can improve later when we need to upgrade if needed 👍 👍 👍

db/migrate/20251221174251_create_roles.rb
nepalez: yes, it is required by the spec

db/migrate/20251221174733_create_membership_roles.rb
nepalez: the reason why I added it here is because we should check if the deleted admin is the last one in the organization
with this denormalization we could avoid joining the `roles` table for that check

app/models/role.rb
nepalez: Yeah, I suggested this in the dive-in, but we haven't got any decision yet; probably this should be dictated by the front-end.

In the database I'd prefer to add this right now (to reasonably protect the data itself) -- to avoid problems in the future migrations; probably later the stricter validation should be added to the model

db/migrate/20251221174733_create_membership_roles.rb
groyoh: I get the idea but I feel like this complexifies the architecture (especially when reviewing/reading the code). I guess joining in this case would be reasonable.

julienbourdeau: Deleting users is a fairly rare operation, few queries a day, probably not even every day, so I agree we should join instead of denomalizing.

nepalez: ok, will fix

nepalez: fixed in this and the following PR-s

db/migrate/20251221174733_create_membership_roles.rb
julienbourdeau: I understand what you mean but thihnk today we don't want to add all these constraint in the BD

nepalez: fixed

app/graphql/resolvers/roles_resolver.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    description "Query roles available for the organization"
+
+    type [Types::RoleType], null: false
+
+    def resolve
+      Role
+        .where(organization_id: [nil, current_organization.id])
+        .order(Role.arel_table[:organization_id].asc.nulls_first)
+        .order(Arel::Nodes::NamedFunction.new("LOWER", [Role.arel_table[:name]]).asc)
```
julienbourdeau: as mentionned before, I'm in favor of using the SQL version here but no big deal

app/graphql/resolvers/role_resolver.rb
```diff
@@ -0,0 +1,22 @@
...
+    REQUIRED_PERMISSION = "roles:view"
+
+    description "Query a single role"
+
+    argument :id, ID, required: true, description: "Uniq ID of the role"
+
+    type Types::RoleType, null: true
+
+    def resolve(id:)
+      Role.with_organization(current_organization.id).find(id)
```
julienbourdeau, line 17: I see multiple classes doing the query like this, I think we would typically write it as `current_organization.roles.find(id:)`

app/graphql/sources/memberships_for_role.rb
```diff
@@ -0,0 +1,32 @@
...
+        .includes(:membership)
+        .where(role_id: role_ids, organization: @organization)
+
+      memberships_by_role = membership_roles.group_by(&:role_id).transform_values do |mrs|
+        mrs.map(&:membership)
+      end
+
+      role_ids.map { |role_id| memberships_by_role[role_id].to_a }
+    end
+  end
```
julienbourdeau, line 33: Beautiful 🌟

app/controllers/graphql_controller.rb
```diff
@@ -29,7 +29,7 @@ def execute
       current_membership:,
       customer_portal_user:,
       request:,
-      permissions: current_membership&.permissions_hash || Permission::EMPTY_PERMISSIONS_HASH
+      permissions: (current_membership || Permission)&.permissions_hash
```
groyoh, line 32:
```suggestion
      permissions: (current_membership || Permission).permissions_hash
```

app/graphql/mutations/roles/update.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      graphql_name "UpdateRole"
+      description "Updates an existing custom role"
+
+      input_object_class Types::Roles::UpdateInput
+
+      type Types::RoleType
+
+      def resolve(id:, **args)
+        role = Role.with_organization(current_organization.id).find_by(id:)
```
groyoh, line 19: Wouldn't this allow to update "public" roles ?

app/graphql/mutations/roles/destroy.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      graphql_name "DestroyRole"
+      description "Deletes a custom role"
+
+      argument :id, ID, required: true
+
+      type Types::RoleType
+
+      def resolve(id:)
+        role = Role.with_organization(current_organization.id).find_by(id:)
```
groyoh, line 19: Wouldn't this allow to delete "public" roles ?

app/graphql/resolvers/roles_resolver.rb
```diff
@@ -0,0 +1,21 @@
...
+    include RequiredOrganization
+
+    REQUIRED_PERMISSION = "roles:view"
+
+    description "Query roles available for the organization"
+
+    type [Types::RoleType], null: false
+
+    def resolve
+      Role
```
groyoh, line 15: Is the number of role per org limited ? Since we don't have paginated result, it could be an issue.

app/graphql/mutations/roles/update.rb
nepalez: It would on this layer. But the result will be error from [validation](https://github.com/getlago/lago-api/blob/custom-roles/data/app/models/role.rb#L19) (it is also prevented at the [database layer](https://github.com/getlago/lago-api/blob/760de8f80136a326b42b19f83966eb3bb9db5979/db/structure.sql#L1275))

app/graphql/mutations/roles/destroy.rb
nepalez: The same reason. Prevention of deletion is not the responcibility of the interface, but of the model and data

app/graphql/resolvers/roles_resolver.rb
nepalez: We started discussion about pagination and server-side filtering, but decided to postpone this until later, because for now it looks a bit premature.

In principle, I agree, that both features should be added at some time

db/migrate/20251221174251_create_roles.rb
```diff
@@ -0,0 +1,96 @@
...
+              RAISE EXCEPTION 'Deleted role cannot be modified';
+            ELSIF OLD.organization_id IS NULL THEN
+              RAISE EXCEPTION 'Predefined role cannot be modified';
+            ELSIF OLD.code IS DISTINCT FROM NEW.code THEN
+              RAISE EXCEPTION 'The code of the role cannot be changed';
+            ELSIF OLD.organization_id IS DISTINCT FROM NEW.organization_id THEN
+              RAISE EXCEPTION 'Custom role cannot be moved to another organization';
+            ELSIF NEW.permissions != OLD.permissions THEN
+              NEW.permissions := ARRAY(SELECT DISTINCT unnest(NEW.permissions) ORDER BY 1);
+            END IF;
```
julienbourdeau: I'm still against some of the things done here. A lot of models have soft deletion and we don't prevent hard deletion at the DB layer.
Some checks are interesting tho

I like those 2 but, I'm still not sure if it's a good idea to have it in the DB rather than Rails.
```suggestion
            IF OLD.organization_id IS NULL THEN
              RAISE EXCEPTION 'Predefined role cannot be modified';
            ELSIF NEW.permissions != OLD.permissions THEN
              NEW.permissions := ARRAY(SELECT DISTINCT unnest(NEW.permissions) ORDER BY 1);
            END IF;
```

app/graphql/resolvers/roles_resolver.rb
julienbourdeau: The only question is for the `membership` attribute.
If I understand correctly there is no N+1 if you request one roles that have multiple membership. But what if you request multiple roles with their memberships 🤔 

to be honest I think this is good enough for now because realistically very few people are going to visit at the same time.

app/models/membership_role.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    errors.add(:base, :last_membership_role)
+    throw(:abort)
+  end
+
+  def role_belongs_to_organization
+    return if role.nil?
+    return if role.organization_id.nil? || role.organization_id == organization_id
+
+    errors.add(:role, :invalid_organization)
```
julienbourdeau: Here I'd just go with Invalid value. It shouldn't be an issue people see anyway.

julienbourdeau:
@groyoh To be fair, the PR was much smaller before when it was reviewed initially. Then, I said we could merge them together to handle the small change needed from frontend or QA.
It's QA validated and should be released soon so I think we can keep the big one.
I agree it's tough to review tho 🙏

db/migrate/20251221174251_create_roles.rb
nepalez: Ok, I would agree that protection against the modification of the deleted role is excessive (who cares!).

But changing of the code or organization should be protected IMO. I'd also like to protect from deletion, at least from the hard one. Maybe more protection/checks should be added for the soft deletion as well.

The reason is that we keep "references" to the role in the invitations by keys (so the key is a kind of index, even though not a primary one), and also disallowing creating new memberships without any roles. This opens the edge case when the role existed at the moment the invitation (having this role as the only one) was sent, but evaded at the moment the user accepted it. In this case the membership won't be created. This would be quite strange from the user's pov.

I see, that the protection is not absolute (soft deletion would open the same case), but the surface of the problem is reduced.

And I cannot even imagine what should mean moving the role to the other organization :open_mouth:

app/graphql/resolvers/roles_resolver.rb
nepalez: fixed

db/migrate/20251221174251_create_roles.rb
nepalez: As for the rails vs db, I'd always prefer keeping invariants as close to the data as it is possible (except some cases when the cost for the db-layer consistency is not too high) -- just to make it sure they are guaranteed whatever the app would do (like skipping validations etc.).

app/graphql/resolvers/role_resolver.rb
nepalez: The `current_organization.roles` wouldn't return the predefined ones because they don't belong to any organization

nepalez: Probably, I could redefine the method, but I'd prefer not to, because it not the developers would expect

app/models/membership_role.rb
nepalez: fixed

db/migrate/20251221174251_create_roles.rb
nepalez: I removed invariants on which I don't rely in the code, and left those whose breaking would require additional checks on the application side here and there

app/graphql/resolvers/role_resolver.rb
julienbourdeau: Oh yeah!! totally missed this. Don't change it 👌

db/migrate/20251221174251_create_roles.rb
julienbourdeau: Alright let's try it! This can be modified later if needed 👌

PR #4742, opened 2025-12-17
feat(Custom roles & permissions) API for roles and permissions

nepalez:
# Roadmap Task

## Context

At this point, the database contains roles and permissions data from the previous PR. However, there is no dedicated API endpoint for listing available roles - the UI uses hardcoded values or GraphQL enum introspection.

This PR switches to using the list of predefined roles from the database instead of hardcoded values, and adds API for retrieving permissions with their associated roles.

This implements the [Work Breakdown | Part 2 & 3](https://www.notion.so/getlago/BE-Dive-In-Custom-roles-permissions-version-2-2c7ef63110d2805d9a22d76a2f9e8c1a?source=copy_link#2c7ef63110d281e38eefea65e928cfc5)

## GraphQL Query `roles`

New resolver returning all roles available for the organization:
* Predefined system roles (Admin, Finance, Manager) with `organization_id = NULL`
* Custom organization-specific roles (for now they are absent in the database)

```graphql
query {
  roles {
    id
    name
    description
    admin
    permissions
  }
}
```

## GraphQL Query `permissions`

New resolver returning all available permissions with their associated roles:

```graphql
query {
  permissions {
    key
    name
    roles { id name }
  }
}
```

## Deprecation of `role` field

The `role` enum field on `MembershipType` is deprecated in favor of the new `roleDetails` field:
* Returns full role objects instead of enum values
* Prepares the UI for transition to the new role system where a membership can have multiple roles

```graphql
# Before (deprecated)
membership {
  role  # Returns: "admin" | "manager" | "finance"
}

# After
membership {
  roleDetails {
    id
    name
    permissions
  }
}
```

## Backward Compatibility

* New endpoints are added without changing existing ones
* The existing permission checking mechanism remains unchanged
* It is expected the UI to switch to new API endpoints to retrieve roles, but still use the existing endpoints for managing members, their roles and permissions

### DIFF
--- a/app/graphql/resolvers/roles_resolver.rb
+++ b/app/graphql/resolvers/roles_resolver.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Resolvers
+  class RolesResolver < Resolvers::BaseResolver
+    include AuthenticableApiUser
+    include RequiredOrganization
+
+    description "Query roles available for the organization"
+
+    type [Types::RoleType], null: false
+
+    def resolve
+      Role
+        .where(organization_id: [nil, current_organization.id])
+        .order(Role.arel_table[:organization_id].asc.nulls_first)
+        .order(Arel::Nodes::NamedFunction.new("LOWER", [Role.arel_table[:name]]).asc)
+    end
+  end
+end
--- a/app/graphql/types/membership_type.rb
+++ b/app/graphql/types/membership_type.rb
@@ -7,8 +7,11 @@ module Types
     field :organization, Types::Organizations::OrganizationType, null: false
     field :user, Types::UserType, null: false
 
-    field :permissions, Types::PermissionsType, null: false
-    field :role, Types::Memberships::RoleEnum, null: false
+    field :permissions, Types::PermissionsType, null: false,
+      deprecation_reason: "Use permissions enum instead"
+    field :role, Types::Memberships::RoleEnum, null: false,
+      deprecation_reason: "Use roleDetails instead"
+    field :roles, [Types::RoleType], null: false
     field :status, Types::Memberships::StatusEnum, null: false
 
     field :created_at, GraphQL::Types::ISO8601DateTime, null: false
--- a/app/graphql/types/role_type.rb
+++ b/app/graphql/types/role_type.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  class RoleType < Types::BaseObject
+    field :admin, Boolean, null: false
+    field :description, String, null: true
+    field :id, ID, null: false
+    field :name, String, null: false
+    field :permissions, [PermissionEnum], null: false
+  end
+end
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -85,6 +85,7 @@ module Types
     field :plans, resolver: Resolvers::PlansResolver
     field :pricing_unit, resolver: Resolvers::PricingUnitResolver
     field :pricing_units, resolver: Resolvers::PricingUnitsResolver
+    field :roles, resolver: Resolvers::RolesResolver
     field :subscription, resolver: Resolvers::SubscriptionResolver
     field :subscription_entitlement, resolver: Resolvers::Entitlement::SubscriptionEntitlementResolver
     field :subscription_entitlements, resolver: Resolvers::Entitlement::SubscriptionEntitlementsResolver
--- a/schema.graphql
+++ b/schema.graphql
@@ -6289,9 +6289,10 @@ type Membership {
   createdAt: ISO8601DateTime!
   id: ID!
   organization: Organization!
-  permissions: Permissions!
+  permissions: Permissions! @deprecated(reason: "Use permissions enum instead")
   revokedAt: ISO8601DateTime!
-  role: MembershipRole!
+  role: MembershipRole! @deprecated(reason: "Use roleDetails instead")
+  roles: [Role!]!
   status: MembershipStatus!
   updatedAt: ISO8601DateTime!
   user: User!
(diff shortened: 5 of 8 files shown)


app/graphql/types/membership_type.rb
```diff
@@ -8,7 +8,9 @@ class MembershipType < Types::BaseObject
     field :user, Types::UserType, null: false
 
     field :permissions, Types::PermissionsType, null: false
-    field :role, Types::Memberships::RoleEnum, null: false
+    field :role, Types::Memberships::RoleEnum, null: false,
+      deprecation_reason: "Use roleDetails instead"
+    field :role_details, [Types::RoleType], null: false, method: :roles
```
julienbourdeau: How do you feel about `roles` name? I think at first you went for `_details` because we didn't want to do multiple roles so we couldn't differenciate with the existing field. Now that it's plural, I think it's better to use just roles.

```suggestion
    field :roles, [Types::RoleType], null: false, method: :roles
```

app/graphql/types/membership_type.rb
```diff
@@ -8,7 +8,9 @@ class MembershipType < Types::BaseObject
     field :user, Types::UserType, null: false
 
     field :permissions, Types::PermissionsType, null: false
-    field :role, Types::Memberships::RoleEnum, null: false
+    field :role, Types::Memberships::RoleEnum, null: false,
+      deprecation_reason: "Use roleDetails instead"
```
julienbourdeau, line 13: TIL `deprecation_reason` 👍

app/graphql/resolvers/roles_resolver.rb
```diff
@@ -0,0 +1,16 @@
...
+  class RolesResolver < Resolvers::BaseResolver
+    include AuthenticableApiUser
+    include RequiredOrganization
+
+    description "Query roles available for the organization"
+
+    type [Types::RoleType], null: false
+
+    def resolve
+      Role.where(organization_id: [nil, current_organization.id]).order(:name)
```
julienbourdeau: Thinking out loud, how about ordering by organization_id null first (so reserved roles are always first), then by name? 🤔

app/graphql/types/permission_type.rb
```diff
@@ -0,0 +1,21 @@
...
+    field :name, String, null: false
+    field :roles, [Types::RoleType], null: false
+
+    def key
+      object
+    end
+
+    def name
+      I18n.t("permissions.#{object.tr(":", ".")}", default: object)
+    end
```
julienbourdeau: All translation stuff are done by the frontend so it should be removed

app/graphql/types/permission_type.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def roles
+      Role.where("? = ANY(permissions)", object)
+    end
```
julienbourdeau: Is this really necessary?

I think without this, we can get rid of the entire new type and resolver and keep using the existing [`PermissionsType`](https://github.com/getlago/lago-api/blob/main/app/graphql/types/permissions_type.rb)

If it's really necessary, we might want to find a way to not make a db query everytime. With about 50 permissions and a frontend that is probably going to request them all, it's going to be a lot of queries.

julienbourdeau (commented):
pretty much all good here. I mostly want to confirm if `PermissionsResolver` and the new `PermissionType` are needed.

app/graphql/types/role_type.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  class RoleType < Types::BaseObject
+    field :admin, Boolean, null: false
+    field :description, String, null: true
+    field :id, ID, null: false
+    field :name, String, null: false
+    field :permissions, [String], null: false
```
julienbourdeau: ~I think this should be something like~
```suggestion
    field :permissions, [Types::PermissionsType], null: false
```
EDIT: it should be an enum
```suggestion
    field :permissions, [Types::PermissionsEnum], null: false
```

app/graphql/types/permission_type.rb
nepalez: I'm not sure. Probably you're right and front-end can do this on its side having the total list of roles with their permissions (@AllanMichay wdyt?)

app/graphql/types/membership_type.rb
nepalez: fixed

nepalez: fixed
(this would make sense when roles were limited by 1 per membership, but without it it doesn't any more)

app/graphql/resolvers/roles_resolver.rb
nepalez: fixed (also modified the corresponding index)

app/graphql/types/permission_type.rb
nepalez: fixed

app/graphql/types/permission_type.rb
nepalez: dropped the whole type in favor of `PermissionEnum`

app/graphql/types/role_type.rb
nepalez: fixed

app/graphql/resolvers/roles_resolver.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    description "Query roles available for the organization"
+
+    type [Types::RoleType], null: false
+
+    def resolve
+      Role
+        .where(organization_id: [nil, current_organization.id])
+        .order(Role.arel_table[:organization_id].asc.nulls_first)
+        .order(Arel::Nodes::NamedFunction.new("LOWER", [Role.arel_table[:name]]).asc)
```
vincent-pochet, line 16: Not a blocker, but it seems that using pure SQL makes it easier to read
```suggestion
        .order("organization_id ASC NULLS FIRST, LOWER(name) ASC")
```

app/graphql/types/role_type.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  class RoleType < Types::BaseObject
```
julienbourdeau, line 4: I just realized we sometime used *Type and sometimes *Object. Object is miuch more common, which make sense since it's extending `BaseObject`. It seems to be the oldest model that are named with Type.
Not a big deal but I'm in favor of renaming.

julienbourdeau (approved):

nepalez:
Closed in favor of updated https://github.com/getlago/lago-api/pull/4741

PR #4745, opened 2025-12-18
feat(thresholds): Subscription create and update

julienbourdeau:
**This pull request introduces a new way to manage usage thresholds override on subscriptions, without creating any plan override.**

See also  https://github.com/getlago/lago-api/pull/4729, https://github.com/getlago/lago-api/pull/4743, https://github.com/getlago/lago-api/pull/4735

This is the **last backward compatible PR**.

### New attributes

2 new attributes are added:

1. `subscriptions.applicable_usage_thresholds` which are exactly what the system will use in progressive billing. It's (in that order): 
  * `subscription.usage_thresholds`
  * `subscription.plan.usage_thresholds`
  * `subscription.plan.parent.usage_thresholds`
2. `subscription.plan.applicable_usage_thresholds` which are always the (parent) plan thresholds, whether there is an override or not

> [!TIP]
> Users should start reading these values.
> If they **write** values using the new attributes, the values will be ignored and not shown in the frontend yet

> [!CAUTION]
> In January, once all users have migrated, we'll remove the old way (breaking change) once the frontend is merged.
> - the `plan_overrides.usage_thresholds` param 🧹
> - the `subscription.plan.usage_thresholds` attribute  🧹
> - migrate all existing thresholds overrides

### Old way (is still working the same way)

`subscription.plan.usage_thresholds` returns the exact same thing: `subscription.plan.usage_thresholds` regardless of plan being a parent or an override.

Passing the thresholds under `plan_overrides` will continue to attach them to the plan override. If plan is not overridden yet, it's created.

In this case, `subscription.usage_thresholds` EQUALS `subscription.plan.usage_thresholds`.

## Examples

Given a subscription and parent plan with 2 thresholds.

```rb
[
  {
    amount_cents: 100_00,
    threshold_display_name: "custom",
    recurring: false
  }, {
    amount_cents: 999_00,
    threshold_display_name: nil,
    recurring: true
  }
]
```

### New Subscription override example

**request**
```rb
{
  subscriptions: {
    name: "subscription name new",
    usage_thresholds: [
      {
        amount_cents: 100_00,
        threshold_display_name: "custom"
      }
    ],
    plan_overrides: {
      name: "rename plan to create override"
    }
  }
}
```

**response**
```rb
{
  subscriptions: {
    name: "subscription name new",
    applicable_usage_thresholds: [
      {
        amount_cents: 100_00,
        threshold_display_name: "custom",
        recurring: false
      }
    ],
    plan: {
      name: "rename plan to create override",
      usage_thresholds: [], # ⬅️ ⬅️ ⬅️  Legacy thresholds from plan or plan override. 
                                       Existing thresholds attached to the plan override are DELETED 
                                       when attaching thresholds to sub
      applicable_usage_thresholds: [] # ⬅️ ⬅️ ⬅️  parent thresholds, cannot be changed when updating sub
    }
  }
}
```

### Using old way

**request**
```rb
{
  subscriptions: {
    name: "subscription name new",
    plan_overrides: {
      name: "rename plan to create override",
      usage_thresholds: [
        {
          amount_cents: 100_00,
          threshold_display_name: "custom"
        }
      ],
    }
  }
}
```

**response**
```rb
{
  subscriptions: {
    name: "subscription name new",
    applicable_usage_thresholds: [ # ⬅️ ⬅️ ⬅️  No thresholds on sub so return the plan override thresholds
      {
        amount_cents: 100_00,
        threshold_display_name: "custom",
        recurring: false
      }
    ],
    plan: {
      name: "rename plan to create override",
      usage_thresholds: [ # ⬅️ ⬅️ ⬅️  thresholds were attached to plan override
        {
          amount_cents: 100_00,
          threshold_display_name: "custom"
        }
      ], 
      applicable_usage_thresholds: [] # ⬅️ ⬅️ ⬅️  parent thresholds, cannot be changed when updating sub
    }
  }
}
```

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -63,6 +63,14 @@ class Plan < ApplicationRecord
     %w[name code]
   end
 
+  def is_parent?
+    !is_child?
+  end
+
+  def is_child?
+    parent_id.present?
+  end
+
   def applicable_usage_thresholds
     (parent || self).usage_thresholds
   end
--- a/app/services/usage_thresholds/update_service.rb
+++ b/app/services/usage_thresholds/update_service.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+module UsageThresholds
+  class UpdateService < BaseService
+    Result = BaseResult
+
+    def initialize(model:, usage_thresholds_params:, partial:)
+      @model = model
+      @usage_thresholds_params = sanitize_params(usage_thresholds_params)
+      @partial = partial
+
+      super
+    end
+
+    def call
+      return result if usage_thresholds_params.empty? && partial?
+
+      return result.single_validation_failure!(error_code: "missing_amount_cents", field: :usage_thresholds) if missing_amount_cents?
+      return result.single_validation_failure!(error_code: "duplicated_values", field: :usage_thresholds) if duplicated_amount_cents?
+      return result.single_validation_failure!(error_code: "multiple_recurring_thresholds", field: :usage_thresholds) if multiple_recurring_thresholds?
+
+      ActiveRecord::Base.transaction do
+        delete_all_thresholds if full?
+
+        update_recurring_threshold
+        update_or_create_thresholds
+      end
+
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    end
+
+    private
+
+    attr_reader :model, :usage_thresholds_params, :partial
+    alias_method :partial?, :partial
+
+    def full?
+      !partial
+    end
+
+    def sanitize_params(usage_thresholds_params)
+      usage_thresholds_params.map do |p|
+        h = p.to_h.deep_symbolize_keys.slice(:threshold_display_name, :amount_cents, :recurring)
+        h[:recurring] ||= false
+        h
+      end
+    end
+
+    def missing_amount_cents?
+      usage_thresholds_params.any? { |p| p[:amount_cents].blank? }
+    end
+
+    def duplicated_amount_cents?
+      grouped = usage_thresholds_params.group_by { |p| [p[:amount_cents], p[:recurring]] }
+      grouped.any? { |_, v| v.size > 1 }
+    end
+
+    def multiple_recurring_thresholds?
+      usage_thresholds_params.count { |p| p[:recurring] } > 1
+    end
+
+    def delete_all_thresholds
+      model.usage_thresholds.update_all(deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
+    end
+
+    def update_recurring_threshold
+      recurring_params = usage_thresholds_params.find { |p| p[:recurring] }
+      return unless recurring_params
+
+      existing_threshold = model.usage_thresholds.find { |t| t.recurring }
+
+      if existing_threshold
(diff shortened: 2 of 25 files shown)


app/models/plan.rb
```diff
@@ -58,6 +58,14 @@ def self.ransackable_attributes(_auth_object = nil)
     %w[name code]
   end
 
+  def parent?
```
vincent-pochet: I would suggest to name this one `has_parent?`.
I think `parent?` in itself is a bit confusing, it could means both "it has a parent plan" and "it is a parent plan"

app/services/usage_thresholds/update_service.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+module UsageThresholds
+  class UpdateService < BaseService
+    Result = BaseResult
+
+    def initialize(model:, usage_thresholds_params:, partial:)
+      @model = model
+      @usage_thresholds_params = sanitize_params(usage_thresholds_params)
+      @partial = partial
```
vincent-pochet, line 10: Could you add a comment on what this `partial` argument is doing?

spec/services/usage_thresholds/update_service_table_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+  end
+
+  # Helper to build threshold attributes for comparison
+  def threshold_attrs(thresholds)
+    thresholds.map { |t| {amount_cents: t.amount_cents, threshold_display_name: t.threshold_display_name, recurring: t.recurring} }
+  end
+
+  describe "threshold updates" do
+    # rubocop:disable Layout/LineLength
+    test_cases = [
```
vincent-pochet, line 22: Is there any reason to use the "go" approach of test cases? It's a bit different from the "rspec" way 🤔 

I think you can achieve something similar by using a local shared example, while keeping the "context" design of rspec

julienbourdeau: I found this to be a lot clearer than shared example to be honest

app/services/usage_thresholds/update_service.rb
julienbourdeau: `partial` is now used for entitlements, metadata and usage thresholds, I'll check what's the best way to document it 🤔 Maybe even a PartialConcern to avoid duplication for `alias :partial?` and `def full?`

It means that when updating a "list" of things, what's not part of the params is left untouched. If full (partial false), everything missing from params is deleted.

app/models/plan.rb
julienbourdeau: > I would suggest to name this one `has_parent?`.
Interesting because in this case the method is `is_parent?`

app/services/usage_thresholds/update_service.rb
vincent-pochet: I think that a comment before the method definition might be enough. It's just to help "future us" understanding what it is when opening the file :smile:

spec/services/usage_thresholds/update_service_table_spec.rb
vincent-pochet: It's just that it's not the common pattern we usually have in specs. That said, it's not a blocker for me, what's important is that the test coverage is here

PR #4755, opened 2025-12-19
chore(rubocop): Update some `rubocop*` gems version

groyoh:
## Description

This updates the following rubocop gems:

- `rubocop`: `1.75.5` -> `1.81.7`
- `rubocop-graphql`: `1.5.4` -> `1.5.6`
- `rubocop-performance`: `1.25.0` -> `1.26.1`
- `rubocop-rails`: `2.25.1` -> `2.34.2`
- `rubocop-thread_safety`: `0.5.1` -> `0.7.3`

This also fixes the following cop reports:

- `GraphQL/OrderedFields`: Fields should be sorted in an alphabetical order within their section.
- `Lint/UselessOr`: Checks for useless OR (|| and or) expressions. Some methods always return a truthy value, even when called on nil (e.g. nil.to_i evaluates to 0). Therefore, OR expressions appended after these methods will never evaluate
- `Lint/DuplicateMethods`: Checks for duplicated instance (or singleton) method definitions.
- `Rails/Delegate`: Use `delegate` to define delegations.
- `Style/EmptyStringInsideInterpolation`: Do not return empty strings in string interpolation. (https://rubystyle.guide#empty-strings-in-interpolation)

### DIFF
--- a/spec/services/billable_metrics/evaluate_expression_service_spec.rb
+++ b/spec/services/billable_metrics/evaluate_expression_service_spec.rb
@@ -49,6 +49,25 @@ RSpec.describe BillableMetrics::EvaluateExpressionService do
       end
     end
 
+    context "when timestamp is missing" do
+      let(:expression) { "event.timestamp" }
+      let(:event) do
+        {
+          "code" => "test_code",
+          "properties" => {}
+        }
+      end
+
+      it "uses current time as fallback" do
+        freeze_time do
+          result = evaluate_service.call
+
+          expect(result).to be_success
+          expect(result.evaluation_result).to eq(Time.current.to_i)
+        end
+      end
+    end
+
     context "when the event failed to evaluate" do
       let(:event) do
         {


config/initializers/sentry.rb
```diff
@@ -6,7 +6,7 @@
     config.release = LagoUtils::Version.call(default: Rails.env).number
     config.breadcrumbs_logger = %i[active_support_logger http_logger]
     config.traces_sample_rate = 0
-    config.traces_sample_rate = ENV["SENTRY_TRACES_SAMPLE_RATE"].to_f || 0
+    config.traces_sample_rate = (ENV["SENTRY_TRACES_SAMPLE_RATE"] || 0).to_f
```
floganz: Do even need `|| 0` part here?
If it's `nil` then `nil.to_f => 0.0`.

groyoh: Indeed. Nice catch!

PR #4766, opened 2025-12-22
feat(multiple-payment-methods): attach payment method on payments

lovrocolic:
## Context

With `multiple payment methods` feature, it will be possible to attach multiple payment methods on customer, but also attach payment method on resources like subscription, wallet, recurring rules or wallet transaction.

## Description

This PR attaches payment method on payment object

### DIFF
--- a/db/migrate/20251222163416_add_payment_method_to_payments.rb
+++ b/db/migrate/20251222163416_add_payment_method_to_payments.rb
@@ -4,9 +4,7 @@ class AddPaymentMethodToPayments < ActiveRecord::Migration[8.0]
   disable_ddl_transaction!
 
   def change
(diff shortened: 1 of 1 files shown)


db/migrate/20251222163416_add_payment_method_to_payments.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddPaymentMethodToPayments < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    safety_assured do
```
vincent-pochet: Not sure that the `safety_assured` is required here

PR #4767, opened 2025-12-22
feat(feature-flag): Introduce Flipper FeatureFlag

julienbourdeau:
**Revival of https://github.com/getlago/lago-api/pull/1872**

It adds the [Flipper gem](https://www.flippercloud.io/docs/introduction) to the project with the Redis driver. It doesn't rely on any external services.

A model FeatureFlag is used as a wrapper around Flipper to ensure flag exists and are always used the way we want.

## How to use

In your code, you can check if a flag is enabled for an organzation.  Feature flag MUST be checked against an organization, not globally and for no other models.

You can use the console to enable it.

```ruby
FeatureFlag.enable :feature1, organization: hooli

FeatureFlag.enabled? :feature1, organization: hooli
# => true

FeatureFlag.enabled? :feature1, organization: another_org
# => false
```

CleanShot 2025-12-22 at 17 27 04@2x

### Using a hash to define all the feature flags.

Flipper can create feature flags on the fly but I didn't really like it workin with it before, that's why I used a defined hardcoded list and `ensure_feature_flag_exists!` in dev mode.
This allows us to easily type the object in GraphQL and can be added to the documentation.

### Frontend

Let's start without anything synced in the frontend but I think we'll add a `feature_flags` array on Organization object.

### `FeatureFlag.cleanup!`

This method removed or add the flags from Redis based on the definition hash. It doesn't have to be run automatically, just sometimes to keep things clean.

### Going Further: Add the UI (not part of the POC)

There is a UI where you can add flag to an actor or enable/disable. See https://github.com/getlago/lago-api/pull/1872

## Limitations

Notice that the flag per actor is not designed to allow for thousands of actors. This must be used as a temporary feature flag, to enable about 100 orgs max. It's fine because it's exactly how we intend ot use it, but it's important to be aware of it.

### DIFF
--- a/config/initializers/flipper.rb
+++ b/config/initializers/flipper.rb
@@ -4,11 +4,14 @@ require "flipper/adapters/redis"
 
 Flipper.configure do |config|
   config.adapter do
+    redis_config = {
+      url: ENV["LAGO_REDIS_FEATURE_FLAGS_URL"] || ENV["REDIS_URL"]
+    }
+    password = ENV["LAGO_REDIS_FEATURE_FLAGS_PASSWORD"].presence || ENV["REDIS_PASSWORD"].presence
+    redis_config[:password] = password if password
+
     Flipper::Adapters::Redis.new(
-      Redis.new({
-        url: ENV["LAGO_REDIS_FEATURE_FLAGS_URL"] || ENV["REDIS_URL"],
-        password: ENV["LAGO_REDIS_FEATURE_FLAGS_PASSWORD"] || ENV["REDIS_PASSWORD"]
-      }),
+      Redis.new(redis_config),
       key_prefix: "flipper:"
     )
   end


app/models/feature_flag.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class FeatureFlag
+  def self.yaml_to_hash(filename)
+    yaml = YAML.parse_file(Rails.root.join(filename))
+    hash = yaml.presence&.to_ruby || {} # Handle empty yaml file
+    hash.with_indifferent_access.freeze
+  end
+
+  DEFINITION = yaml_to_hash("app/config/feature_flags.yaml") # rubocop:disable Layout/ClassStructure
```
vincent-pochet: I guess it's okay to move the constant in the first list of this class to avoid the rubocop comment. 🙂

config/initializers/flipper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require "flipper/adapters/redis"
+
+Flipper.configure do |config|
+  config.adapter { Flipper::Adapters::Redis.new(Redis.new(db: 8)) }
```
vincent-pochet: We should use an explicit Redis config (as we do with sidekiq)

Gemfile
```diff
@@ -57,6 +57,10 @@ gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby]
 gem "graphql"
 gem "graphql-pagination"
 
+# Feature flags
+gem "flipper", "1.3.0"
+gem "flipper-redis"
```
vincent-pochet, line 62: I'm curious about the choice of Redis, over a plain active record model?

app/models/feature_flag.rb
julienbourdeau: When I did it for Permission model it didn't work because the `yaml_to_hash` method is not defined yet. The method had to be appear first.

https://github.com/getlago/lago-api/blob/a22fb4b81d31838d8e5cba59909b7d37981f6d35/app/models/permission.rb#L4-L12

config/initializers/flipper.rb
julienbourdeau: Good point! thanks

app/models/feature_flag.rb
vincent-pochet: I see :facepalm:... forget my comment :smile:

Gemfile
julienbourdeau: I have always used it with Redis 😄 There is an [ActiveRecord adapter](https://www.flippercloud.io/docs/adapters/active-record) if we prefer to put the flags in PG

app/models/feature_flag.rb
mariohd: I guess something like this would work, so you don't have to keep the method in the class object


```suggestion
  DEFINITION = begin
   yaml = YAML.parse_file(Rails.root.join("config/feature_flags.yaml"))
   hash = yaml.presence&.to_ruby || {}
   hash.with_indifferent_access
  end.freeze
```

app/models/feature_flag.rb
```diff
@@ -0,0 +1,49 @@
...
+      ensure_feature_flag_exists!(feature_name)
+      Flipper.enable(feature_name, organization)
+    end
+
+    def disable(feature_name, organization:)
+      ensure_feature_flag_exists!(feature_name)
+      Flipper.disable(feature_name, organization)
+    end
+
+    def cleanup!
```
mariohd: `cleanup!` is a good name, but I feel that `sanitize!` is more common method name for this.

Gemfile
lovrocolic: I am just wondering if it could be handy for product team to review such flags in posgres since they don't use redis 🤔  For me personally both works

julienbourdeau: I had looked into the PG version and I honestly didn't like it. It's really using PG as KV store so i didn't bring anything interesting like eager loading flags or better querying.
Without Redis, you don't have sidekiq so it's already mandatory to have redis I think.

app/models/feature_flag.rb
julienbourdeau: Yeah I agree 👍

julienbourdeau:
@jnunemaker Thank a lot for checking this out 🙏 

Indeed I'll enable strict mode on dev. For some reason I thought it was only wit the PG driver. Maybe because the setup created the config file automatically 🤷‍♂️ 

The main intent of the definition file was definitly to mimic strict mode. But it should help maintain what flag are still active or done and it removing all flags from DB. We're not going to add the UI yet I think.

config/initializers/flipper.rb
julienbourdeau: So I was tempted to add more env variables but I'm having second thoughts. It's already kinda hard to follow all the env variables. I see `LAGO_REDIS_CACHE_DB` used in `development.rb` but not `production.rb` for instance.

I think we don't want this in the cache, because it should be "persisted" (cache could virtually be dropped and rebuilt). In this case, it will use the standard REDIS_URL (so same as sidekiq).

Maybe the DB name should be in a var at least. 🤔

app/config/feature_flags.yaml
```diff
@@ -0,0 +1,3 @@
+mutiple_payment_methods:
+  description: "Allow multiple payment methods per customer"
+  owner: "Lovro & Mario"
```
groyoh: I'd suggest an array with Github handles:

```suggestion
  owners:
    - @lovrocolic 
    - @mariohd 
```

config/initializers/flipper.rb
groyoh: `LAGO_REDIS_CACHE_DB` should actually not exist. It should be part of the `LAGO_REDIS_CACHE_URL` directly. 

But I kind believe that we can use a different URL for feature cache (while re-using the same Redis instance but different DB).

app/config/feature_flags.yaml
julienbourdeau: very good point, I like the github handles

mariohd (approved):

config/initializers/flipper.rb
julienbourdeau: I went for full URL and password 👍 💯

PR #4768, opened 2025-12-22
feat(Custom roles & permissions) Backward-compatible migration to new roles

nepalez:
# Roadmap Task

## Context

This is the second of three PRs implementing custom roles using the **expand-migrate-contract** pattern:

1. **Previous PR (Expand)**: Added new data structures and APIs without changing existing behavior
2. **This PR (Migrate)**: Switch internal logic to use new system while keeping API backward-compatible
3. **Next PR (Contract)**: Remove legacy code and finalize the new API

At this point, the database contains empty `roles` and `membership_roles` tables from the previous PR. The legacy `role` enum on memberships and invites remains the source of truth.

This PR:
- Backfills `membership_roles` from existing `role` values
- Switches all services to write to both old and new systems (dual-write)
- Adds new `roles` arguments alongside deprecated `role` arguments
- Keeps the legacy `role` field working for existing clients

**Why this intermediate step?** It allows the UI team to migrate at their own pace. Old API calls continue to work, while new API calls are available for gradual adoption.

This PR implements the Work Breakdown | Part 2.

### Data Migration

Backfills `membership_roles` table from existing memberships:

### Internal Logic Switch

All services now use `membership_roles` as the source of truth:

| Service | Change |
|---------|--------|
| `Invites::CreateService` | Creates invite with `roles` array |
| `Invites::UpdateService` | Validates roles via `Role` model |
| `Invites::ValidateService` | Checks roles exist in organization |
| `Memberships::UpdateService` | Creates/discards `MembershipRole` records |
| `Memberships::RevokeService` | Checks admin count via `admin_membership_roles` |
| `UsersService` | Creates `MembershipRole` on registration/invite accept |

### API Changes (Backward-Compatible)

All mutations accept both old and new arguments:

```graphql
# Old way (still works)
mutation {
  createInvite(input: { email: "user@example.com", role: admin }) { ... }
}

# New way (preferred)
mutation {
  createInvite(input: { email: "user@example.com", roles: ["Admin"] }) { ... }
}
```

When both `role` and `roles` are provided, `roles` takes precedence.

| Mutation | Old Argument | New Argument |
|----------|--------------|--------------|
| `createInvite` | `role: MembershipRole` (deprecated) | `roles: [String]` |
| `updateInvite` | `role: MembershipRole` (deprecated) | `roles: [String]` |
| `updateMembership` | `role: MembershipRole` (deprecated) | `roles: [String]` |

### Type Changes

| Type | Old Field | New Field |
|------|-----------|-----------|
| `Invite` | `role` (deprecated, returns first) | `roles: [String!]!` |
| `Membership` | `role` (deprecated) | `roles: [String!]!` |

## Backward Compatibility

**Full backward compatibility maintained:**

- Legacy `role` argument works in all mutations
- Legacy `role` field returns first role from array
- Existing API clients continue to work without changes
- UI can migrate to `roles` API incrementally
- Safe to deploy without UI coordination

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -35,8 +35,8 @@ class Organization < ApplicationRecord
   has_many :admins_memberships, -> { active.admins }, class_name: "Membership"
   has_many :admins, through: :admins_memberships, source: :user
   # New way to access admin users
-  has_many :membership_roles, through: :memberships
-  has_many :admin_membership_roles, -> { admins }, class_name: "MembershipRole"
+  has_many :membership_roles, through: :active_memberships
+  has_many :admin_membership_roles, -> { admins }, through: :active_memberships, source: :membership_roles
   has_many :admin_memberships, through: :admin_membership_roles, source: :membership
   has_many :admin_users, through: :admin_memberships, source: :user
 
--- a/app/services/roles/destroy_service.rb
+++ b/app/services/roles/destroy_service.rb
@@ -12,7 +12,7 @@ module Roles
     def call
       return result.not_found_failure!(resource: "role") unless role
       return result.forbidden_failure!(code: "predefined_role") if predefined_role?
-      return result.forbidden_failure!(code: "role_assigned_to_members") if role.membership_roles.kept.exists?
+      return result.forbidden_failure!(code: "role_assigned_to_members") if role.active_memberships.exists?
 
       role.discard!
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -12,6 +12,7 @@ en:
         greater_than: value_is_out_of_range
         greater_than_or_equal_to: value_is_out_of_range
         inclusion: value_is_invalid
+        invalid: value_is_invalid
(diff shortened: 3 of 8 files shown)


app/models/permission.rb
```diff
@@ -11,7 +11,7 @@ def permissions_hash(role = nil)
   private
 
   def yaml_to_hash(filename)
-    h = YAML.parse_file(Rails.root.join("app/config", filename)).to_ruby
+    h = YAML.parse_file(Rails.root.join("config", filename)).to_ruby
```
julienbourdeau, line 14: Any reason you wanted to move this in the main `config` file?
Maybe the feature flag yaml should be moved too.

nepalez: Well, in a standard-ish Rails approach, it is the `config` where all the app's configuration is expected to be placed to. It was a bit surprising for me to find two configs :man_shrugging: 

I recognized that the reasons for this separation was historical, and probably the desire to keep settings closer to the code which made sense while there were different files for finances, management etc. But with one global setting for permissions I think it is less surprising to keep it as a part of the configuration.

PR #4777, opened 2025-12-24
chore(seeds): Add a organization with Clickhouse store in seeds

groyoh:
## Context

We currently allow two event stores: Postgres and Clickhouse. Switching an organization between the two is currently handled via Rails console which is quite cumbersome when testing both scenario locally or on staging.

## Description

This updates the seeds to create a second test organization with Clickhouse enabled as a store.


db/seeds/01_base.rb
```diff
@@ -7,176 +7,202 @@
...
+  end
+
+  unless PricingUnit.exists?(organization:, code: "xyz")
+    PricingUnits::CreateService.call!(
+      organization:,
+      name: "xyz",
+      code: "xyz",
+      short_name: "XYZ"
+    )
+  end
```
julienbourdeau, line 207: I guess this was moved because it's what you needed?
One day we should have a way to run seeder "per organization". Otherwise, everything will be eventually moved inside this loop :D

groyoh: Indeed. The thing is pricing unit is one of those things that cannot be created from the API (UI or seed only). So, if it is needed to test on the Clickhouse org, we'd have to create it from the UI. It's the only thing I moved in here for practical purposes.

PR #4779, opened 2025-12-26
feat(Custom roles & permissions) Code cleaning

nepalez:
# Roadmap Task

## Context

This is the final of three PRs implementing custom roles. It removes use of the single `role` in the code, but keeps it in the database (as an optional nullified column) and in the API for the backward-compatibility

At this point:
- All data has been migrated to `membership_roles` table
- All services use the new role system internally
- UI has migrated to use `roles` instead of deprecated `role` field

This PR removes the legacy implementation:
- Makes `role` columns in `memberships` and `invites` tables optional
- Makes `roles` argument required

**Why separate this as a final PR?** Breaking changes require coordination with UI. This PR should only be merged after confirming all clients have migrated to the new API.

## Breaking Changes

**This PR contains breaking changes.** All clients must update before deployment:

### Mutations

```graphql
# Before (no longer works)
createInvite(input: { email: "...", role: "admin" })

# After (required)
createInvite(input: { email: "...", roles: ["admin"] })
```

## Pre-merge Checklist

- [ ] UI has migrated all `role` usages to `roles`
- [ ] No API clients use `MembershipRole` enum
- [ ] Coordinated release with frontend team

### DIFF
--- a/app/graphql/mutations/billing_entities/update_applied_dunning_campaign.rb
+++ b/app/graphql/mutations/billing_entities/update_applied_dunning_campaign.rb
@@ -6,7 +6,7 @@ module Mutations
       include AuthenticableApiUser
       include RequiredOrganization
 
-      REQUIRED_PERMISSION = "billing_entities:dunning_campaigns:update"
+      REQUIRED_PERMISSION = "billing_entities:update"
 
       graphql_name "BillingEntityUpdateAppliedDunningCampaign"
       description "Updates the applied dunning campaign for a billing entity"
--- a/app/graphql/mutations/charge_filters/create.rb
+++ b/app/graphql/mutations/charge_filters/create.rb
@@ -6,7 +6,7 @@ module Mutations
       include AuthenticableApiUser
       include RequiredOrganization
 
-      REQUIRED_PERMISSION = "charge_filters:create"
+      REQUIRED_PERMISSION = "charges:create"
 
       graphql_name "CreateChargeFilter"
       description "Creates a new Charge Filter"
--- a/app/graphql/mutations/charge_filters/destroy.rb
+++ b/app/graphql/mutations/charge_filters/destroy.rb
@@ -6,7 +6,7 @@ module Mutations
       include AuthenticableApiUser
       include RequiredOrganization
 
-      REQUIRED_PERMISSION = "charge_filters:delete"
+      REQUIRED_PERMISSION = "charges:delete"
 
       graphql_name "DestroyChargeFilter"
       description "Deletes a Charge Filter"
--- a/app/graphql/mutations/charge_filters/update.rb
+++ b/app/graphql/mutations/charge_filters/update.rb
@@ -6,7 +6,7 @@ module Mutations
       include AuthenticableApiUser
       include RequiredOrganization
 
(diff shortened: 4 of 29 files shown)


app/graphql/mutations/invites/update.rb
```diff
@@ -11,13 +11,14 @@ class Update < BaseMutation
       graphql_name "UpdateInvite"
       description "Update an invite"
 
-      input_object_class Types::Invites::UpdateInput
+      argument :id, ID, required: true
+      argument :roles, [String], required: true
```
julienbourdeau: I think this happened because we have a stupid rule saying after 3 arguments you must extract, and one rule that says with less than 3 you must inline 😄 
Not your fault but very annoying.

db/migrate/20251222151015_remove_role_from_memberships.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class RemoveRoleFromMemberships < ActiveRecord::Migration[8.0]
+  def change
+    safety_assured do
+      remove_column :memberships, :role, :integer, default: 0, null: false
```
julienbourdeau: To remove column, you must make sure the code running un production does not use these columns. For cloud, it just means doing 2 deploys but for OSS. it means you must make a "bridge version" where you ask users to upgrade to version 1.x, run it un prod and then upgrade to 1.y.

app/graphql/types/memberships/update_input.rb
```diff
@@ -6,7 +6,6 @@ class UpdateInput < Types::BaseInputObject
       graphql_name "UpdateMembershipInput"
 
       argument :id, ID, required: true
-      argument :role, Types::Memberships::RoleEnum, required: false, deprecation_reason: "Use `roles` instead"
```
julienbourdeau, line 9: We must be sure that frontend does not use this anymore, so I'd prefer to put this entire PR in the next bridge version 👍

julienbourdeau (commented):
All good with the code but make sure to not release this at the same time as the others.

db/migrate/20251222151015_remove_role_from_memberships.rb
nepalez: Ahh, good point, thank you! Will split this one to two. Probably, I'll have to split the migration as well, to enable NULLs in the first part, and remove in the second one.

Will fix

app/graphql/mutations/invites/update.rb
nepalez: :man_shrugging:

db/migrate/20251222151015_remove_role_from_memberships.rb
nepalez: Fixed. In this PR I only loose the requirements making `role` columns optional (and updated the specs to ensure the `role` is not needed anymore). The next PR drops the columns along with the deprectated fields in the API.

app/graphql/types/memberships/update_input.rb
nepalez: I moved the GraphQL API as well as a serialization cleaning into the [next PR](https://github.com/getlago/lago-api/pull/4864)

PR #4780, opened 2025-12-26
fix(minimum-commitment): Persist minimum commitment fee date boundaries

ancorcruz:
## Context

The Invoice v4 PDF template needs to group fees by their billing period boundaries. Minimum commitment fees reconcile a specific billing period, but this information was not being persisted, making it impossible to correctly group commitment fees in the invoice PDF.

## Description

Store the billing period boundaries (from_datetime, to_datetime) in the fee's properties JSONB field. The boundaries come from the invoice subscription that represents the period being reconciled:
- Pay in arrears: the current invoice_subscription
- Pay in advance: the previous invoice_subscription

Also skip commitment fee creation for pay-in-advance plans on their first invoice, since there is no previous period to reconcile.


annvelents (approved):
Nice!

PR #4781, opened 2025-12-29
Feat(metadata): add metadata on plans

annvelents:
## Context

Some customers want to have some additional data on the plan that is not currently supported in Lago. To help them keep lago as a source of truth, we need to give them a possibility to store all related tada on the plan

## Description

Added possibility to store metadata on plan

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -13,8 +13,7 @@ class Plan < ApplicationRecord
   has_one :minimum_commitment, -> { where(commitment_type: :minimum_commitment) }, class_name: "Commitment"
   has_one :metadata,
     class_name: "Metadata::ItemMetadata",
-    as: :owner,
-    dependent: :destroy
+    as: :owner
 
   has_many :commitments
   has_many :charges, dependent: :destroy
--- a/app/serializers/v1/metadata_serializer.rb
+++ b/app/serializers/v1/metadata_serializer.rb
@@ -3,9 +3,7 @@
 module V1
   class MetadataSerializer < ModelSerializer
     def serialize
-      return nil unless model
-
-      model.value
+      model&.value
     end
   end
 end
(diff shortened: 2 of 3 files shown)


app/controllers/api/v1/plans/metadata_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+        end
+
+        private
+
+        def metadata_params
+          {metadata: params.fetch(:metadata, {}).permit!.to_h}
+        end
+
+        def render_metadata
+          render(json: {metadata: plan.reload.metadata&.value})
```
vincent-pochet, line 56: Should we use a dedicated serializer? 🤔 
Maybe a `V1::MetadataSerializer` that could be shared between resources. What do you think?

annvelents: yes, that does sounds good!

app/models/plan.rb
```diff
@@ -11,6 +11,10 @@ class Plan < ApplicationRecord
   belongs_to :parent, class_name: "Plan", optional: true
 
   has_one :minimum_commitment, -> { where(commitment_type: :minimum_commitment) }, class_name: "Commitment"
+  has_one :metadata,
+    class_name: "Metadata::ItemMetadata",
+    as: :owner,
+    dependent: :destroy
```
nepalez: nit: probably, it would be better not to destroy here because plans are soft-deleted (see LOC:8 above)

app/serializers/v1/metadata_serializer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module V1
+  class MetadataSerializer < ModelSerializer
+    def serialize
+      return nil unless model
+
+      model.value
```
nepalez: nitpick:

```suggestion
      model&.value
```

PR #4784, opened 2025-12-29
feat(charge): Add code support to charge and fixed charge services and serializers

rsempe:
## Context

We want to make charges and charge filters independent from the plan/subscription object, enabling
add/update of a single charge or charge filter without sending the entire plan's payload.

## Description

The Charge model now has a `code` attribute that needs to be exposed through the API and handled in create/update operations.

This adds the following:
- Add `code` to ChargeSerializer/FixedChargeSerializer output along with `parent_id`
- Add `lago_id` and `charge_code` to ChargeFilterSerializer/FixedChargeSerializer
- Handle `code` param in Charges::CreateService and FixedCharges::CreateService
- Handle `code` param in Charges::UpdateService and FixedCharges::UpdateService for both cascade and non-cascade modes

### DIFF
--- a/app/serializers/v1/charge_serializer.rb
+++ b/app/serializers/v1/charge_serializer.rb
@@ -18,7 +18,7 @@ module V1
         min_amount_cents: model.min_amount_cents,
         properties:,
         applied_pricing_unit:,
-        parent_id: model.parent_id
+        lago_parent_id: model.parent_id
       }
 
       payload.merge!(charge_filters)
--- a/app/serializers/v1/fixed_charge_serializer.rb
+++ b/app/serializers/v1/fixed_charge_serializer.rb
@@ -15,7 +15,7 @@ module V1
         prorated: model.prorated,
         properties: model.properties,
         units: model.units,
-        parent_id: model.parent_id
+        lago_parent_id: model.parent_id
       }
 
       payload.merge!(taxes) if include?(:taxes)
--- a/app/controllers/api/v1/plans_controller.rb
+++ b/app/controllers/api/v1/plans_controller.rb
@@ -124,6 +124,7 @@ module Api
           ],
           charges: [
             :id,
+            :code,
             :invoice_display_name,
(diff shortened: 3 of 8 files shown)


annvelents (approved):

app/serializers/v1/charge_serializer.rb
```diff
@@ -16,7 +17,8 @@ def serialize
         prorated: model.prorated,
         min_amount_cents: model.min_amount_cents,
         properties:,
-        applied_pricing_unit:
+        applied_pricing_unit:,
+        parent_id: model.parent_id
```
vincent-pochet: I'm wondering if we use `parent_lago_id` 🤔

app/serializers/v1/fixed_charge_serializer.rb
```diff
@@ -6,14 +6,16 @@ def serialize
...
         invoice_display_name: model.invoice_display_name,
         add_on_code: model.add_on.code,
         created_at: model.created_at.iso8601,
         charge_model: model.charge_model,
         pay_in_advance: model.pay_in_advance,
         prorated: model.prorated,
         properties: model.properties,
-        units: model.units
+        units: model.units,
+        parent_id: model.parent_id
```
vincent-pochet: Same question 🙂

vincent-pochet (approved):

app/serializers/v1/charge_serializer.rb
rsempe: We are using `parent_id` at the `Plan` level: https://github.com/getlago/lago-api/blob/main/app/serializers/v1/plan_serializer.rb#L23.

But it would be nice to change it to `lago_parent_id` I agree.
Doing the change here just for charges... wdyt?

vincent-pochet: Arf of course :sweat_smile: 
Let's update this one for now. We will update the other one later

PR #4786, opened 2025-12-30
feat(multiple-payment-methods): split periodic invoices per payment method

lovrocolic:
## Context

With the multiple payment methods feature, it will be possible to attach multiple payment methods to a customer, but also attach a payment method to resources like subscriptions, wallets, recurring rules, or wallet transactions.

## Description

If a customer has multiple subscriptions where each subscription has a different payment method, we cannot group such subscriptions in a single invoice, like it is done today. Instead, these subscriptions should be grouped per payment method and there should be as many invoices as there are subscription groups.

### DIFF
--- a/app/services/subscriptions/organization_billing_service.rb
+++ b/app/services/subscriptions/organization_billing_service.rb
@@ -530,6 +530,7 @@ module Subscriptions
     #   - [pm_1, provider] + [nil, provider]  → single group (both resolve to pm_1)
     #   - [pm_1, provider] + [pm_2, provider] → two groups (different resolved id)
     def group_by_payment_method(subscriptions)
+      return [subscriptions] unless organization.feature_flag_enabled?(:multiple_payment_methods)
       return [subscriptions] if subscriptions.size <= 1
 
       customer = subscriptions.first.customer
--- a/spec/services/subscriptions/organization_billing_service_spec.rb
+++ b/spec/services/subscriptions/organization_billing_service_spec.rb
@@ -443,6 +443,7 @@ RSpec.describe Subscriptions::OrganizationBillingService do
     end
 
     context "when grouping subscriptions by payment method" do
+      let(:organization) { create(:organization, feature_flags: ["multiple_payment_methods"]) }
       let(:interval) { :monthly }
       let(:billing_time) { :anniversary }
       let(:current_date) { subscription_at.next_month }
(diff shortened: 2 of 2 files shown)


app/services/subscriptions/organization_billing_service.rb
```diff
@@ -22,11 +22,15 @@ def call
           end
         end
 
-        BillSubscriptionJob.perform_later(
-          billing_subscriptions,
-          today.to_i,
-          invoicing_reason: :subscription_periodic
-        )
+        subscription_groups = group_by_payment_method(billing_subscriptions)
```
lovrocolic, line 27: For current subscriptions in the DB, without any migration, this method should return only one group and there should be no change in the logic.
@mariohd Can you double check the logic? Do you think we should still add additional feature flag guard?

app/services/subscriptions/organization_billing_service.rb
```diff
@@ -420,5 +424,45 @@ def already_billed_today
...
+    #       - If nil, inherit from customer's default payment method
+    def resolve_payment_method_key(subscription, default_payment_method)
+      if subscription.payment_method_id.present?
+        [subscription.payment_method_id, subscription.payment_method_type]
+      elsif subscription.payment_method_type == "manual"
+        [nil, "manual"]
+      elsif default_payment_method.present?
+        [default_payment_method.id, "provider"]
+      else
+        [nil, subscription.payment_method_type]
```
groyoh, line 559: I might be wrong but I believe this branch is not tested.

groyoh (approved):

app/services/subscriptions/organization_billing_service.rb
mariohd: that's the default scenario (tested 2 times) and it's being tested at least once here as well. 

https://github.com/getlago/lago-api/pull/4786/changes#diff-89768692814ee284bd75d04ed037c9bf0b09a899611aea86ab31163af21c7c9eR695-R711

app/services/subscriptions/organization_billing_service.rb
```diff
@@ -420,5 +424,45 @@ def already_billed_today
...
+      default_payment_method = customer.default_payment_method
+
+      resolved_keys = subscriptions.map { |s| resolve_payment_method_key(s, default_payment_method) }.uniq
+
+      if resolved_keys.size == 1
+        return [subscriptions]
+      end
+
+      subscriptions.group_by { |s| resolve_payment_method_key(s, default_payment_method) }.values
+    end
```
mariohd, line 546:
```suggestion
      payment_method_keys = subscriptions.map { |s| resolve_payment_method_key(s, default_payment_method) }

      payment_method_keys.uniq.count == 1 ?  [subscriptions] : payment_method_keys.values
    end
```

lovrocolic: New approach miss `group_by` so results are different

spec/services/subscriptions/organization_billing_service_spec.rb
```diff
@@ -648,5 +648,313 @@
         end
       end
     end
+
+    context "when grouping subscriptions by payment method" do
+      let(:organization) { create(:organization, feature_flags: ["multiple_payment_methods"]) }
```
julienbourdeau, line 446: 🚀

julienbourdeau (approved):
I havent reviewed the rest, but the use of the feature flag is correct! 🚀

PR #4787, opened 2025-12-30
feat(wallet): Implement prepaid credit consumption tracking for wallet traceability

groyoh:
## Context

## Description

This implements step #1 of the above breakdown: Implement core traceability infrastructure to track which inbound transactions fund each outbound transaction when prepaid credits are applied to an invoice.

### Join-table approach

The implementation uses a join table approach with a new `wallet_transaction_consumptions` table that links outbound transactions to their source inbound transactions. Each record captures which inbound transaction funded (partially or fully) an outbound transaction and how much was consumed in cents. This approach is non-breaking, doesn't increase the number of wallet transactions, and provides flexibility for complex scenarios like partial consumption and multiple sources.

Each inbound transaction tracks its `remaining_amount_cents` to know how much is still available for consumption. A partial index on `wallet_transactions` enables efficient lookup of available inbound transactions. Database-level constraints (including a trigger) ensure data integrity by validating that inbound transactions are actually inbound and outbound transactions are actually outbound.

*Example 1: Simple Consumption (Single Source)*

Customer tops up $100, then consumes $40:

```
INBOUND                                   OUTBOUND
┌─────────────────────────┐              ┌─────────────────────────┐
│ TX1 (purchased)         │              │ TX2 (invoiced)          │
│ amount: $100            │────$40─────▶ │ amount: $40             │
│ remaining: $60          │              │                         │
└─────────────────────────┘              └─────────────────────────┘

Join table:
│ TX1 │ TX2 │ $40 │
```

*Example 2: Consumption Spanning Multiple Inbounds*

Customer has two top-ups ($30 granted, $50 purchased), then consumes $60:

```
INBOUND                                 OUTBOUND
┌─────────────────────────┐
│ TX1 (granted)           │             ┌─────────────────────────┐
│ amount: $30             │────$30─────▶│ TX3 (invoiced)          │
│ remaining: $0           │         ┌──▶│ amount: $60             │
└─────────────────────────┘         │   └─────────────────────────┘
┌─────────────────────────┐         │
│ TX2 (purchased)         │         │
│ amount: $50             │───$30───┘
│ remaining: $20          │
└─────────────────────────┘

Join table:
│ TX1 │ TX3 │ $30 │  (granted consumed first)
│ TX2 │ TX3 │ $30 │  (then purchased)
```

*Example 3: Multiple Outbounds from Same Inbound*

Customer tops up $100, then has two invoices ($25 and $35):

```
INBOUND                              OUTBOUND
┌─────────────────────────┐              ┌─────────────────────────┐
│ TX1 (purchased)         │────$25──────▶│ TX2 (invoiced)          │
│ amount: $100            │              │ amount: $25             │
│ remaining: $40          │              └─────────────────────────┘
│                         │              ┌─────────────────────────┐
│                         │────$35──────▶│ TX3 (invoiced)          │
│                         │              │ amount: $35             │
└─────────────────────────┘              └─────────────────────────┘

Join table:
│ TX1 │ TX2 │ $25 │
│ TX1 │ TX3 │ $35 │
```

Note that such scenario is bound to create a drift between the sum of the derived consumed credits and the inbound transaction credits:

```
Wallet rate: 0.66666

INBOUND                                  OUTBOUND
┌─────────────────────────┐              ┌─────────────────────────┐
│ TX1 (purchased)         │────$25──────▶│ TX2 (invoiced)          │
│ amount: $100            │              │ amount: $65             │
│ credits: 150.00150      │              │ credits: 97.50099       │
│ remaining: $0           │              │                         │ 
│ remaining_credits: 0    │              └─────────────────────────┘
│                         │              ┌─────────────────────────┐
│                         │────$35──────▶│ TX3 (invoiced)          │
│                         │              │ amount: $35             │
│                         │              │ credits: 52.50053       │
└─────────────────────────┘              └─────────────────────────┘

Join table:
│ TX1 │ TX2 │ $25 │
│ TX1 │ TX3 │ $35 │

Credits of inbound TX = 150.00150
Sum of credits from outbounds = 150.00152 != 150.00150
Sum of credits from consumption table = 150.00152 != 150.00150
```

*Example 4: Complex Scenario with Priority*

Customer has: $20 granted (priority 1), $50 purchased, $30 granted (priority 2).

Then consumes $80:

```
INBOUND (ordered by priority)        OUTBOUND
┌─────────────────────────┐
│ TX1 (granted, prio 1)   │            ┌─────────────────────────┐
│ amount: $20             │────$20────▶│ TX4 (invoiced)          │
│ remaining: $0           │         ┌─▶│ amount: $80             │
└─────────────────────────┘         │  │                         │
┌─────────────────────────┐         │  │                         │
│ TX3 (granted, prio 2)   │         │  │                         │
│ amount: $30             │───$30───┘  │                         │
│ remaining: $0           │         ┌─▶│                         │
└─────────────────────────┘         │  └─────────────────────────┘
┌─────────────────────────┐         │
│ TX2 (purchased)         │         │
│ amount: $50             │───$30───┘
│ remaining: $20          │
└─────────────────────────┘

Consumption order: TX1 (granted, prio 1) → TX3 (granted, prio 2) → TX2 (purchased)

Join table:
│ TX1 │ TX4 │ $20 │  (granted, priority 1)
│ TX3 │ TX4 │ $30 │  (granted, priority 2)
│ TX2 │ TX4 │ $30 │  (purchased, last)
```

### Wallet-Level Traceability Flag

Traceability is controlled per-wallet via a traceable boolean flag, allowing gradual rollout. When credits are applied to an invoice for a traceable wallet, the system calls `WalletTransactions::TrackConsumptionService` which implements a FIFO consumption strategy: granted credits are consumed first (ordered by priority, then created_at), followed by purchased credits in the same order. This ensures free credits are used before paid ones.

Once feature is ready: new wallets will be created with `traceable = true`.

### DIFF
--- a/app/models/wallet_transaction.rb
+++ b/app/models/wallet_transaction.rb
@@ -10,6 +10,18 @@ class WalletTransaction < ApplicationRecord
   belongs_to :invoice, optional: true
   belongs_to :credit_note, optional: true
 
+  has_many :consumptions,
+    class_name: "WalletTransactionConsumption",
+    foreign_key: :inbound_wallet_transaction_id,
+    inverse_of: :inbound_wallet_transaction,
+    dependent: :destroy
+
+  has_many :fundings,
+    class_name: "WalletTransactionConsumption",
+    foreign_key: :outbound_wallet_transaction_id,
+    inverse_of: :outbound_wallet_transaction,
+    dependent: :destroy
+
   has_many :applied_invoice_custom_sections,
     class_name: "WalletTransaction::AppliedInvoiceCustomSection",
     dependent: :destroy
@@ -50,10 +62,24 @@ class WalletTransaction < ApplicationRecord
   validates :priority, presence: true, inclusion: {in: 1..50}
   validates :name, length: {minimum: 1, maximum: 255}, allow_nil: true
   validates :invoice_requires_successful_payment, exclusion: [nil]
+  validates :remaining_amount_cents,
+    numericality: {greater_than_or_equal_to: 0},
+    allow_nil: true,
+    if: :inbound?
+  validates :remaining_amount_cents, absence: true, if: :outbound?
 
   delegate :customer, to: :wallet
 
   scope :pending, -> { where(status: :pending) }
+  scope :available_inbound, -> { inbound.settled.where("remaining_amount_cents > 0") }
+  scope :in_consumption_order, -> {
+    granted_status = transaction_statuses[:granted]
+    order(
+      :priority,
+      Arel.sql("CASE WHEN transaction_status = #{granted_status} THEN 0 ELSE 1 END") => :asc,
+      :created_at => :asc
+    )
+  }
 
   def self.order_by_priority
     order(:priority)
@@ -91,6 +117,7 @@ end
 #  name                                :string(255)
 #  payment_method_type                 :enum             default("provider"), not null
 #  priority                            :integer          default(50), not null
+#  remaining_amount_cents              :bigint
 #  settled_at                          :datetime
 #  skip_invoice_custom_sections        :boolean          default(FALSE), not null
 #  source                              :integer          default("manual"), not null
@@ -107,6 +134,7 @@ end
 #
 # Indexes
 #
+#  idx_wallet_transactions_available_inbound       (wallet_id, priority, (\nCASE\n    WHEN (transaction_status = 1) THEN 0\n    ELSE 1\nEND), created_at) WHERE ((remaining_amount_cents > 0) AND (transaction_type = 0) AND (status = 1))
 #  index_wallet_transactions_on_credit_note_id     (credit_note_id)
 #  index_wallet_transactions_on_invoice_id         (invoice_id)
 #  index_wallet_transactions_on_organization_id    (organization_id)
--- a/app/services/wallet_transactions/track_consumption_service.rb
+++ b/app/services/wallet_transactions/track_consumption_service.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module WalletTransactions
+  class TrackConsumptionService < BaseService
+    Result = BaseResult
+
+    def initialize(outbound_wallet_transaction:)
+      @outbound_wallet_transaction = outbound_wallet_transaction
+
+      super
+    end
+
+    def call
+      ActiveRecord::Base.transaction do
+        consume_by_priority
+      end
+
+      result
+    end
+
+    private
+
+    attr_reader :outbound_wallet_transaction
+
+    delegate :wallet, to: :outbound_wallet_transaction
+
+    def consume_by_priority
+      amount_cents = outbound_wallet_transaction.amount_cents
+      inbounds = available_inbounds.to_a
+      available_amount = inbounds.sum(&:remaining_amount_cents)
+
+      if amount_cents > available_amount
+        return result.single_validation_failure!(
+          field: :amount_cents,
+          error_code: "exceeds_available_amount"
+        )
+      end
+
+      amount_left = amount_cents
+
+      inbounds.each do |inbound|
+        break if amount_left <= 0
+
+        consume_amount = [inbound.remaining_amount_cents, amount_left].min
+
+        create_consumption(inbound, consume_amount)
+        amount_left -= consume_amount
+      end
+    end
+
+    def create_consumption(inbound, amount_cents)
+      WalletTransactionConsumption.create!(
+        organization: wallet.organization,
+        inbound_wallet_transaction: inbound,
+        outbound_wallet_transaction: outbound_wallet_transaction,
+        consumed_amount_cents: amount_cents
+      )
+
+      # this raises a DB error if the remaining_amount_cents goes below zero
+      inbound.decrement!(:remaining_amount_cents, amount_cents) # rubocop:disable Rails/SkipsModelValidations
+    end
+
+    def available_inbounds
+      wallet.wallet_transactions.available_inbound.in_consumption_order
+    end
+  end
+end
--- a/db/migrate/20251224152733_add_remaining_amount_cents_to_wallet_transactions.rb
+++ b/db/migrate/20251224152733_add_remaining_amount_cents_to_wallet_transactions.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddRemainingAmountCentsToWalletTransactions < ActiveRecord::Migration[8.0]
+  def change
+    add_column :wallet_transactions, :remaining_amount_cents, :bigint
+
+    add_check_constraint :wallet_transactions,
+      "remaining_amount_cents >= 0 OR remaining_amount_cents IS NULL",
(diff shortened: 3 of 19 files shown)


db/migrate/20251224152733_add_remaining_amount_cents_to_wallet_transactions.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddRemainingAmountCentsToWalletTransactions < ActiveRecord::Migration[8.0]
+  def change
+    add_column :wallet_transactions, :remaining_amount_cents, :bigint
+
+    add_check_constraint :wallet_transactions,
+      "remaining_amount_cents >= 0 OR remaining_amount_cents IS NULL",
+      name: "remaining_amount_cents_non_negative",
+      validate: false
```
floganz, line 10: Am I right that `remaining_amount_cents IS NULL` are for outbound transactions?

app/models/wallet_transaction.rb
```diff
@@ -74,6 +85,15 @@ def mark_as_failed!(timestamp = Time.zone.now)
...
     update!(status: :failed, failed_at: timestamp)
   end
+
+  private
+
+  def remaining_amount_cents_only_for_inbound
+    return if remaining_amount_cents.nil?
+    return if inbound?
+
+    errors.add(:remaining_amount_cents, :invalid)
```
floganz: Should we maybe have more specific error code? Like `present`, to indicate that it should be at all.

spec/models/wallet_transaction_spec.rb
```diff
@@ -12,13 +12,35 @@
     it { is_expected.to validate_presence_of(:transaction_type) }
     it { is_expected.to validate_presence_of(:source) }
     it { is_expected.to validate_presence_of(:transaction_status) }
+
+    describe "remaining_amount_cents validation" do
+      let(:wallet) { create(:wallet) }
```
floganz: Do we really need `wallet` variable for these test? It looks like we can remove it.

spec/models/wallet_transaction_consumption_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+RSpec.describe WalletTransactionConsumption do
+  describe "associations" do
+    it do
+      expect(subject).to belong_to(:organization)
+      expect(subject).to belong_to(:inbound_wallet_transaction).class_name("WalletTransaction")
+      expect(subject).to belong_to(:outbound_wallet_transaction).class_name("WalletTransaction")
+    end
+  end
+
+  describe "validations" do
```
floganz, line 14: I've noticed that some of these tests using FactoryBot to build consumption object, another using `described_class.new`. Should we align to one of them?
Factory seems to have benefits removing necessity for us to specify `organization` and `consumed_amount_cents`

app/services/wallet_transactions/track_consumption_service.rb
```diff
@@ -0,0 +1,95 @@
...
+    def create_consumption(inbound, amount_cents)
+      WalletTransactionConsumption.create!(
+        organization: wallet.organization,
+        inbound_wallet_transaction: inbound,
+        outbound_wallet_transaction: outbound_wallet_transaction,
+        consumed_amount_cents: amount_cents
+      )
+
+      # this raises a DB error if the remaining_amount_cents goes below zero
+      inbound.decrement!(:remaining_amount_cents, amount_cents) # rubocop:disable Rails/SkipsModelValidations
```
floganz, line 60: Could we have any rounding issues or differences compared to ruby version we use in many other places?

app/services/wallet_transactions/track_consumption_service.rb
```diff
@@ -0,0 +1,95 @@
...
+          error_code: "exceeds_available_amount"
+        )
+      end
+
+      create_consumption(inbound, amount_cents)
+    end
+
+    def consume_by_priority
+      amount_cents = outbound_wallet_transaction.amount_cents
+      available_amount = available_inbounds.sum(:remaining_amount_cents) || 0
```
floganz: Won't `sum` already return zero for empty collection?
For what case we need `|| 0`?

app/services/wallet_transactions/track_consumption_service.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      wallet.wallet_transactions
+        .inbound
+        .settled
+        .where("remaining_amount_cents > 0")
+        .order(
+          Arel.sql("CASE WHEN transaction_status = ? THEN 0 ELSE 1 END", granted_status),
+          :priority,
+          created_at: :asc
+        )
```
floganz: Do you think there's any sense to extract all of these or partially to scopes in wallet transaction model?
E.g. `available_inbound` and `in_application_order`?

floganz (approved):
Great work!
Left few minor comments if you like them.

db/migrate/20251224152733_add_remaining_amount_cents_to_wallet_transactions.rb
groyoh: No necessarily. Until we migrate wallets and enable traceability, the inbound wallet transaction won't have `remaining_amount_cents` set. The proper check would be:

```
wallets.traceable = FALSE OR transaction_type = 1 OR remaining_amount_cents >= 0
```

But that's not possible with check. Only with triggers.

Once the data migration is complete, we can update the check to be:

```
transaction_type = 1 OR remaining_amount_cents >= 0
```

app/models/wallet_transaction.rb
groyoh: Makes sense. I guess we can use `absence` validation.

spec/models/wallet_transaction_spec.rb
groyoh: You're right, the wallet transaction factory should create one.

spec/models/wallet_transaction_consumption_spec.rb
groyoh: Nice catch! Claude code being drunk I guess 😄

app/services/wallet_transactions/track_consumption_service.rb
groyoh: This does a DB call which returns `nil` if there's no matching row:

```sql
lago=# SELECT SUM(1) FROM wallet_transactions WHERE 1 = 0;
 sum 
-----
    
(1 row)
```

app/services/wallet_transactions/track_consumption_service.rb
groyoh: Nice point! 👍

app/services/wallet_transactions/track_consumption_service.rb
groyoh: There shouldn't be a different between rounding in Ruby and rounding in DB. But theorically, there could be an issue as the wallet transaction amount is a decimal but there's actually no way to reproduce this today as we always round the wallet transaction amount with 2 decimals.

PR #4789, opened 2025-12-31
fetat(multiple-payment-methods): Update parent payment service with core logic

lovrocolic:
## Context

With the multiple payment methods feature, it will be possible to attach multiple payment methods to a customer, but also attach a payment method to resources like subscriptions, wallets, recurring rules, or wallet transactions.

## Description

In this PR there is main logic that determines payment method when creating an invoice. Payment method is attached to payment object and is always accessible from the invoice. E.g. `invoice.payments.first.payment_method`

We need to attach `payment method` to `payment` so that payment providers which process payment intent in the job have access to determined payment method.

Also this PR adds some optional params to send payment method in one-off flow and retry flow

### DIFF
--- a/app/services/invoices/payments/create_service.rb
+++ b/app/services/invoices/payments/create_service.rb
@@ -184,71 +184,9 @@ module Invoices
       #       nil means: skip automatic payment (manual type or no payment method configured)
       #       payment_method_params takes precedence (used for retry with override)
       def determine_payment_method
-        @determine_payment_method ||= if payment_method_params.present?
-          determine_override_payment_method
-        else
-          determine_invoice_payment_method
-        end
-      end
-
-      def determine_override_payment_method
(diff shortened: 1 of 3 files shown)


app/services/invoices/payments/create_service.rb
```diff
@@ -165,6 +179,77 @@ def processing_payment
...
+
+        wallet = wallet_transaction.wallet
+        return nil if wallet.payment_method_type == "manual"
+
+        if wallet.payment_method_id.present?
+          return customer.payment_methods.find_by(id: wallet.payment_method_id)
+        end
+
+        customer.default_payment_method
+      end
```
vincent-pochet, line 190: This logic is not directly related to the creation of the payment itself. Maybe it could be moved to a dedicated service, what do you think?

PR #4796, opened 2026-01-05
Feat(fixed charges): fixed charges billed monthly

annvelents:
## Context

Fixed Charges should be chargeable monthly when the plan is yearly. However, in this case for each month we'll create an invoice subscription with date boundaries:
```
{
  from: beginning of year,
  to: end_of_year,
  charges_from: beginning_of_year,
  charges_to: end_of_year,
  fixed_charges_from: beginning_of_current_month,
  fixed_charges_to: beginning_of_current_month
}
```
however, on the invoice_subscription we have a uniqueness index of `subscription_id + charges_from + charges_to`, which is essential to not have duplicates due to race conditions, but having current set up makes it impossible to bill fixed_charges monthly, because charges boundaries will be repeated

That's why we decided to change the logic of filling these values, as `monthly fixed charge invoice` is a new type of invoice, so there is no expectation on it, the same, fixed_charges from / to is a new field - customers also do not have any logic built around it, and now:
- charges_from/to are only filled if current invoice includes charges
- fixed_charges from/to are only filled if current_invoice includes fixed charges

## Description

Added responsibility to DateService of checking if specific boundaries should be filled

### DIFF
--- a/spec/services/subscriptions/dates/semiannual_service_spec.rb
+++ b/spec/services/subscriptions/dates/semiannual_service_spec.rb
@@ -400,6 +400,50 @@ RSpec.describe Subscriptions::Dates::SemiannualService do
           end
         end
       end
+
+      context "when plan has fixed_charges monthly" do
+        let(:billing_time) { :calendar }
+        let(:subscription_at) { Time.zone.parse("02 Feb 2020") }
+        let(:plan) { create(:plan, interval: :semiannual, pay_in_advance:, bill_fixed_charges_monthly: true) }
+
+        context "when charges should be billed" do
+          let(:billing_at) { Time.zone.parse("01 Jul 2022") }
+
+          it "returns charges_from_datetime" do
+            expect(result).to eq("2022-01-01 00:00:00 UTC")
+          end
+        end
+
+        context "when charges should not be billed" do
+          let(:billing_at) { Time.zone.parse("01 Feb 2023") }
+
+          it "does not return charges_from_datetime" do
+            expect(result).to eq("")
+          end
+        end
+      end
+
+      context "when plan has charges monthly" do
+        let(:billing_time) { :calendar }
+        let(:subscription_at) { Time.zone.parse("02 Feb 2020") }
+        let(:plan) { create(:plan, interval: :semiannual, pay_in_advance:, bill_charges_monthly: true) }
+
+        context "when charges should be billed" do
+          let(:billing_at) { Time.zone.parse("01 Jul 2022") }
+
+          it "returns charges_from_datetime" do
+            expect(result).to eq("2022-06-01 00:00:00 UTC")
+          end
+        end
+
+        context "when charges should billed as monthly" do
+          let(:billing_at) { Time.zone.parse("01 Feb 2023") }
+
+          it "does return charges_from_datetime" do
+            expect(result).to eq("2023-01-01 00:00:00 UTC")
+          end
+        end
+      end
     end
 
     context "when billing_time is anniversary" do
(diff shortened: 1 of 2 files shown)


app/services/invoices/create_invoice_subscription_service.rb
```diff
@@ -81,17 +81,17 @@ def subscriptions_boundaries
...
-        fixed_charges_to_datetime: date_service.fixed_charges_to_datetime,
-        fixed_charges_duration: date_service.fixed_charges_duration_in_days,
+        from_datetime: ds.from_datetime,
+        to_datetime: ds.to_datetime,
+        charges_from_datetime: ds.should_fill_charges_boundaries? ? ds.charges_from_datetime : nil,
+        charges_to_datetime: ds.should_fill_charges_boundaries? ? ds.charges_to_datetime : nil,
+        charges_duration: ds.should_fill_charges_boundaries? ? ds.charges_duration_in_days : nil,
+        fixed_charges_from_datetime: ds.should_fill_fixed_charges_boundaries? ? ds.fixed_charges_from_datetime : nil,
+        fixed_charges_to_datetime: ds.should_fill_fixed_charges_boundaries? ? ds.fixed_charges_to_datetime : nil,
+        fixed_charges_duration: ds.should_fill_fixed_charges_boundaries? ? ds.fixed_charges_duration_in_days : nil,
```
ancorcruz: these `should_fill_charges_boundaries?` and `should_fill_fixed_charges_boundaries?` methods are called on the same object and sounds like internal knowledge of the object.

Could we move this to the actual method following the tell don't ask principle?
`charges_from_datetime`, `charges_to_datetime`, `charges_duration_in_days`, `fixed_charges_from_datetime`... will return a datetime or `nil` based on its own status.


Would it break something else?

app/services/invoices/create_invoice_subscription_service.rb
```diff
@@ -124,18 +124,18 @@ def termination_boundaries(subscription, boundaries)
...
+        to_datetime: ds.to_datetime,
+        charges_from_datetime: ds.should_fill_charges_boundaries? ? ds.charges_from_datetime : nil,
+        charges_to_datetime: ds.should_fill_charges_boundaries? ? ds.charges_to_datetime : nil,
+        fixed_charges_from_datetime: ds.should_fill_fixed_charges_boundaries? ? ds.fixed_charges_from_datetime : nil,
+        fixed_charges_to_datetime: ds.should_fill_fixed_charges_boundaries? ? ds.fixed_charges_to_datetime : nil,
         timestamp: datetime,
-        charges_duration: dates_service.charges_duration_in_days,
-        fixed_charges_duration: dates_service.fixed_charges_duration_in_days
+        charges_duration: ds.should_fill_charges_boundaries? ? ds.charges_duration_in_days : nil,
+        fixed_charges_duration: ds.should_fill_fixed_charges_boundaries? ? ds.fixed_charges_duration_in_days : nil
```
ancorcruz: same applies here.

app/services/subscriptions/dates/yearly_service.rb
```diff
@@ -16,6 +16,30 @@ def first_month_in_first_yearly_period?
...
+
+      # if bill_fixed_charges_monthly=true, fixed charge boundaries should be filled
+      # if bill_charges_monthly=true, fixed charge boundaries should be filled only for the first month of the period
+      # For yearly plans with bill_charges_monthly=false, and bill_fixed_charges_monthly=false,
+      # boundaries are always filled
+      def should_fill_fixed_charges_boundaries?
+        return true if plan.bill_fixed_charges_monthly?
+
+        return first_month_in_yearly_period? if plan.bill_charges_monthly?
+
```
ancorcruz, line 41: wondering if we should also consider here if plan has charges/fixed charges. It would made sense the invoice does not have date boundaries for charges/fixedcharges if the plan does not have them.

What do you think?

ancorcruz (approved):
Looks great.

I have left a few questions for you to reconsider. I'm available to discuss them.

PR #4798, opened 2026-01-05
feat(rate_limits): Implement rate limits for organizations

jdenquin:
Implements per-organization rate limiting on the /api/v1/events/batch endpoint using Rack Attack. Organizations are limited to 10 requests per second by default, with the ability to configure custom limits per organization.

- New column: api_rate_limits (JSONB) on organizations table for configurable per-endpoint limits
- Use `rate_limiting` from Rails, waiting for an improvement on their side to fully use it out of the box
- Kubernetes-ready: Rate limits are shared across all pods via Redis/Memcache (Rails.cache)

## Configuration

 Default: 10 requests/second per organization

 Custom limit example:
 ```ruby
 # Period is in seconds
 organization.update!(api_rate_limits: {events_batch: { limit: 10, period: 1 })
 ```

 Response when rate limited
With the header `Retry-After` in seconds
```json
  {
    "status": 429,
    "error": "Rate limit exceeded",
    "code": "rate_limit_exceeded"
  }
```

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2600,8 +2600,8 @@ CREATE TABLE public.organizations (
     hmac_key character varying NOT NULL,
     authentication_methods character varying[] DEFAULT '{email_password,google_oauth}'::character varying[] NOT NULL,
     audit_logs_period integer DEFAULT 30,
-    api_rate_limits jsonb DEFAULT '{}'::jsonb NOT NULL,
     pre_filter_events boolean DEFAULT false NOT NULL,
+    api_rate_limits jsonb DEFAULT '{}'::jsonb NOT NULL,
     clickhouse_deduplication_enabled boolean DEFAULT false NOT NULL,
     CONSTRAINT check_organizations_on_invoice_grace_period CHECK ((invoice_grace_period >= 0)),
     CONSTRAINT check_organizations_on_net_payment_term CHECK ((net_payment_term >= 0))
@@ -11672,3 +11672,4 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20220530091046'),
 ('20220526101535'),
 ('20220525122759');
+


app/services/api_keys/cache_service.rb
```diff
@@ -27,15 +27,15 @@ def call
         # Avoid returning an expired API key
         unless api_key.expired?
           organization = Organization.instantiate(cache_json["organization"].slice(*Organization.column_names))
-          return api_key, organization
+          return api_key, organization, cache_json["api_rate_limits"]
```
groyoh: I'm not sure to understand why we need to return a third value here since the `api_rate_limits` are already part of the organization attributes 🤔

app/services/api_keys/cache_service.rb
groyoh, line 1: I believe you'll need to update the cache version as order versions of the cache won't include the `api_rate_limits`.

groyoh (commented):
I'd suggest something a bit more flexible like:

```ruby
api_rate_limits: [ { "action": "events_batch", "limit": 10, "period": 1 } ]
```

This would allow us to also specify rate limit per minute/hour if necessary without creating a new attribute or busting the cache again.

app/services/api_keys/cache_service.rb
jdenquin: yes, it'll use default values, we can clear the cache for org that want different limits

app/controllers/concerns/rate_limit.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module RateLimit
+  extend ActiveSupport::Concern
+
+  # Note: Support for lambda with rails `rate_limit` is not yet released, so we have to support it by ourself
+  # To remove when its fully supported by rails : https://github.com/rails/rails/commit/ac29e7d61cc93b832289120625633ac72635b552
+
+  private
+
+  def rate_limit(store: cache_store, name: nil)
```
groyoh: `store` parameter is unused.

app/controllers/concerns/rate_limit.rb
```diff
@@ -0,0 +1,26 @@
...
+    limit_name = "#{controller_name}_#{action_name}"
```
julienbourdeau: I'd use `#` which I think is standard way to separate controller and action, controller_name can contain `_`.

app/controllers/api/v1/events_controller.rb
```diff
@@ -3,11 +3,17 @@
...
+
       skip_audit_logs!
 
       before_action :ensure_organization_uses_clickhouse, only: [:index_enriched]
+      before_action :rate_limit, only: %i[batch]
 
       ACTIONS_WITH_CACHED_API_KEY = %i[create batch estimate_instant_fees batch_estimate_instant_fees].freeze
+      DEFAULT_RATE_LIMITS = {
+        "events_batch" => { "limit" => 10, "period" => 1 }
+      }.freeze
```
groyoh: What about a class method instead for simplicity and explicitness:

```suggestion
class Api::BaseController < ApplicationController
  def self.rate_limit(action, limit:, period:)
    before_action(-> do
       limit_name = "#{controller_name}_#{action_name}"
      applicable_limit = api_rate_limits&.dig(limit_name) || { "limit" => limit, "period" => period }
      to = api_rate_limits.fetch("limit")
      within = applicable_limit.fetch("period")
      rate_limiting(to: to, within: within, by: by, with: with, store: store, name: name, scope: scope || controller_path),
    end, only: action)
  end
end

class Api::V1::EventsController < Api::BaseController
  rate_limit(:batch, limit: 10, period: 1)
end
```

app/controllers/concerns/rate_limit.rb
```diff
@@ -0,0 +1,26 @@
...
+    applicable_limit = api_rate_limits&.dig(limit_name) || self.class::DEFAULT_RATE_LIMITS.dig(limit_name)
+    to = applicable_limit.dig("limit")
+    within = applicable_limit.dig("period")
```
julienbourdeau: In this case, period is a number of seconds, right?

config/environments/test.rb
```diff
@@ -46,15 +46,6 @@
 
   Dotenv.load
 
-  if ENV["LAGO_REDIS_CACHE_URL"].present?
-    redis_store_config = {
-      url: ENV["LAGO_REDIS_CACHE_URL"],
-      ssl_params: {verify_mode: OpenSSL::SSL::VERIFY_NONE}
-    }
-    config.cache_store = :redis_cache_store, redis_store_config
-  end
```
groyoh, line 55: I believe this was done on purpose to configure the redis cache store in tests.

app/controllers/api/v1/events_controller.rb
```diff
@@ -3,11 +3,17 @@
...
+      include RateLimit
+
       skip_audit_logs!
 
       before_action :ensure_organization_uses_clickhouse, only: [:index_enriched]
+      before_action :rate_limit, only: %i[batch]
 
       ACTIONS_WITH_CACHED_API_KEY = %i[create batch estimate_instant_fees batch_estimate_instant_fees].freeze
+      DEFAULT_RATE_LIMITS = {
+        "events_batch" => { "limit" => 10, "period" => 1 }
```
julienbourdeau: I think for the naming, we should stick with rails naming

```suggestion
        "events_batch" => { "to" => 10, "within" => 1 }
```

app/controllers/concerns/rate_limit.rb
julienbourdeau: Yeah because I thnk store should be passed to `rate_limiting` instead of Rails.cache

app/controllers/api/v1/events_controller.rb
julienbourdeau: In this case, only `only:` is supported. 

`rate_limit` is the official rails name so in this case, I'd use the same signature and use **options to pass to before_action

Image

config/environments/test.rb
jdenquin: it was not used because right after we add a config_cache_store = null_store again, so idk!

app/controllers/api/v1/events_controller.rb
jdenquin: Like julien said, Rails use a class method, its better for now to use a private one, waiting for the imp release on their side

app/controllers/concerns/rate_limit.rb
jdenquin: yes we use the default one in here, I will remove it!

app/controllers/concerns/rate_limit.rb
jdenquin: yes!

app/controllers/api/v1/events_controller.rb
jdenquin: its hard to understand for eng and documentation, wdyt?

app/controllers/api/v1/events_controller.rb
groyoh: > In this case, only only: is supported.

Well yes but that's how it is used today. Also I guess this is how it will be used in the future unless we plan to have a generic rate-limit across multiple APIs.

> rate_limit is the official rails name so in this case, I'd use the same signature and use **options to pass to before_action

Well we can rename it if that's what you're worried about: `rate_limit_action(action, limit:, period:).

---

Anyway, no strong opinion here. I simply found it more explicit and less error prone that expecting people to define a `before_action` and expecting a `DEFAULT_RATE_LIMITS` constant to be defined on the controller.

config/environments/test.rb
groyoh: Actually you're right. I was expecting the spec helper to rely on this (https://github.com/getlago/lago-api/blob/041d395378833913eb0041adfd7c5cc16bd84992/spec/spec_helper.rb#L175). But I did some tests and it does not. So we can safely remove it indee.

app/controllers/api/v1/events_controller.rb
groyoh: TBH I don't think Rails has the best naming convention 😄 Also the method parameter naming is different from configuration naming. I think this is much more comprehensible. Only thing I could suggest is to use `window` instead of `period` but they have the same meaning.

julienbourdeau: I like window better as well.

I think the rails naming is designed to use with the class method and does not feels as good in a hash.

`rate_limit to: 100, within: 1.minute` is awesome 👌

app/controllers/api/v1/events_controller.rb
jdenquin: I'll test what you proposed, I used a concern to avoid having context issues

app/controllers/api/v1/events_controller.rb
jdenquin: done!

PR #4802, opened 2026-01-06
perf(daily-usage): Improve daily usage compute scheduling

groyoh:
## Context

Once a day, we compute the daily usage for the previous day. To do this, we schedule thousands of `DailyUsages::ComputeJob`. Metrics show that each job scheduling takes around 1ms causing the whole process to sometime last dozens of minutes.

## Description

This improves the performance by doing a bulk scheduling by `ActiveJob.perform_all_later`.

### DIFF
--- a/app/services/daily_usages/compute_all_service.rb
+++ b/app/services/daily_usages/compute_all_service.rb
@@ -9,8 +9,11 @@ module DailyUsages
     end
 
     def call
-      subscriptions.find_each do |subscription|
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):
Nice! I'm pretty sure the same logic could apply to other clock jobs

groyoh:
> I'm pretty sure the same logic could apply to other clock jobs

Definitely but I only noticed such a huge duration (> 10m) for this job and `DunningCampaigns::BulkProcessService`.

PR #4805, opened 2026-01-06
feat: fix refundable amount on credit notes

brunomiguelpinto:
## Context

This PR fixes credit note functionality by enforcing that refunds can only be issued for amounts that have actually been paid on an invoice, rather than just the invoice total amount. This prevents issuing refunds for unpaid invoices.

## Description
- Modified refund validation to check against total_paid_amount_cents instead of total_amount_cents
- Updated refundable_amount_cents calculation to properly handle partially paid invoices and credit invoices with wallets

### DIFF
--- a/app/services/credit_notes/validate_service.rb
+++ b/app/services/credit_notes/validate_service.rb
@@ -80,7 +80,7 @@ module CreditNotes
       add_error(field: :base, error_code: "does_not_match_item_amounts")
     end
 
-    # NOTE: Check if refunded amount is less than or equal to invoice total amount
+    # NOTE: Check if refunded amount is less than or equal to the invoice's paid amount
     def valid_refund_amount?
       return true if credit_note.refund_amount_cents.zero?
 


app/models/invoice.rb
```diff
@@ -327,11 +327,19 @@ def refundable_amount_cents
     return 0 if version_number < CREDIT_NOTES_MIN_VERSION || draft?
     return 0 if !payment_succeeded? && total_paid_amount_cents == total_amount_cents
 
-    amount = total_paid_amount_cents - credit_notes.sum("refund_amount_cents + credit_amount_cents")
-    amount = amount.negative? ? 0 : amount
+    already_refunded_cents = credit_notes.sum("refund_amount_cents")
```
vincent-pochet, line 330: Just wondering if we should check the status of the credit notes here 🤔
(Not directly related to this PR)

brunomiguelpinto: That's a good question, let me check

PR #4806, opened 2026-01-06
feat(charge-filters): Add CRUD endpoints for charge filters

rsempe:
## Context

We want to make charges and charge filters independent from the plan/subscription object, enabling
add/update of a single charge or charge filter without sending the entire plan's payload.

## Description

This adds the ability to manage charge filters individually through dedicated API endpoints, rather than only through the batch update mechanism on charges.

Add new REST API endpoints for charge filters nested under charges:
- `GET /api/v1/plans/:plan_code/charges/:charge_code/filters`
- `GET /api/v1/plans/:plan_code/charges/:charge_code/filters/:id`
- `POST /api/v1/plans/:plan_code/charges/:charge_code/filters`
- `PUT /api/v1/plans/:plan_code/charges/:charge_code/filters/:id`
- `DELETE /api/v1/plans/:plan_code/charges/:charge_code/filters/:id`

Create dedicated services for filter management:
- `ChargeFilters::CreateService` for creating individual filters
- `ChargeFilters::UpdateService` for updating filter properties
- `ChargeFilters::DestroyService` for soft-deleting filters

### DIFF
--- a/app/controllers/api/v1/plans/charges/filters_controller.rb
+++ b/app/controllers/api/v1/plans/charges/filters_controller.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Plans
+      module Charges
+        class FiltersController < Api::BaseController
+          before_action :find_plan
+          before_action :find_charge
+          before_action :find_charge_filter, only: %i[show update destroy]
+
+          def index
+            charge_filters = charge.filters
+              .includes(:charge, :billable_metric_filters)
+              .page(params[:page])
+              .per(params[:per_page] || PER_PAGE)
+
+            render(
+              json: ::CollectionSerializer.new(
+                charge_filters,
+                ::V1::ChargeFilterSerializer,
+                collection_name: "filters",
+                meta: pagination_metadata(charge_filters)
+              )
+            )
+          end
+
+          def show
+            render(json: ::V1::ChargeFilterSerializer.new(charge_filter, root_name: "filter"))
+          end
+
+          def create
+            result = ChargeFilters::CreateService.call(charge:, params: input_params.to_h.deep_symbolize_keys)
+
+            if result.success?
+              render(
+                json: ::V1::ChargeFilterSerializer.new(
+                  result.charge_filter,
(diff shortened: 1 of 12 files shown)


app/controllers/api/v1/plans/charges/filters_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+    module Plans
+      module Charges
+        class FiltersController < Api::BaseController
+          before_action :find_plan
+          before_action :find_charge
+          before_action :find_charge_filter, only: %i[show update destroy]
+
+          def index
+            charge_filters = charge.filters
+              .includes(:billable_metric_filters)
```
julienbourdeau: The serializer uses `model.charge.code` so I think we need to eagerload the charge here as well

app/services/charge_filters/destroy_service.rb
```diff
@@ -0,0 +1,32 @@
...
+      @charge_filter = charge_filter
+
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "charge_filter") unless charge_filter
+
+      ActiveRecord::Base.transaction do
+        charge_filter.values.each(&:discard!)
```
julienbourdeau: I have no idea how many values to expect here, but I'm generally in favor of not looping

```suggestion
        charge_filter.values.update_all deteted_at: Time.now
```

app/controllers/api/v1/plans/charges/filters_controller.rb
rsempe: Thanks - Will do it on [ChargeSerializer](https://github.com/getlago/lago-api/blob/main/app/serializers/v1/charge_serializer.rb#L40) as well 👍

app/services/charge_filters/destroy_service.rb
rsempe: Good point!
Will update [`ChargeFilters::CreateOrUpdateBatchService`](https://github.com/getlago/lago-api/blob/main/app/services/charge_filters/create_or_update_batch_service.rb#L134) as well 👍

PR #4807, opened 2026-01-06
feat(exports): Exports entitlement tables

murparreira:
## Context

We need the `entitlement_entitlements` and `entitlement_features` views.

## Description

Create both views with scenic.


julienbourdeau (approved):
If you need this, you might need `entitlement_subscription_feature_removals` **unless you're only interested in the plan's entitlements**.

PR #4810, opened 2026-01-07
feat(wallets): events with wallet-target

annvelents:


app/services/charges/pay_in_advance_aggregation_service.rb
```diff
@@ -46,10 +46,15 @@ def aggregation_filters
       filters = {event:, charge_id: charge.id}
 
       model = charge_filter.presence || charge
-      if model.pricing_group_keys.present?
-        filters[:grouped_by_values] = model.pricing_group_keys.index_with { event.properties[it] }
+      grouped_by_values = model.pricing_group_keys&.index_with { event.properties[it] } || {}
+
+      # Add wallet_id grouping if enabled on charge
+      if charge.group_by_wallet && event.properties["wallet_id"].present?
```
julienbourdeau, line 52: I strongly, **very strongly** believe, we should introduce `code` on wallet first, and require a code here.

```suggestion
      if charge.group_by_wallet && event.properties["wallet_code"].present?
```

PR #4811, opened 2026-01-07
feat(wallet): add prepaid credit breakdown to invoice PDF

groyoh:
## Context

## Description

This implements step #7 of the above breakdown: Display free vs prepaid credits breakdown on invoice PDFs.

To do this, we added the `prepaid_granted_credit_amount_cents` and `prepaid_purchased_credit_amount_cents` to the `invoices` table and now calculate the breakdown if the customer's wallets are traceable.

We then use this information when generating the invoice PDF.

To keep backward compatibility, these fields are `nil` by default and we still calculate and display the overall `prepaid_credit_amount_cents` as a fallback.

When reviewing the invoice snapshots, the following script helps:

```sh
snapshot_folder=spec/views/app/views/templates/invoices/__snapshots__/templates_invoices_v4.slim/when_invoice_type_is_subscription_with_prepaid_credits
base_snapshot_name="without_breakdown_(legacy_behavior)/renders_correctly.html.snap"
base_snapshot="$snapshot_folder/$base_snapshot_name"
for f in $(ls $snapshot_folder/*/*); do
  if [ "$f" == "$base_snapshot" ]; then
		continue
	fi
	snapshot_name=$(basename $(dirname $f))/$(basename $f)
	echo -e "Comparing $base_snapshot_name to $snapshot_name:\n"
	set -o pipefail
	diff -u "$base_snapshot" $f | tail -n +3 && echo "No difference."
	set +o pipefail
	echo ""
done

```

which gives the following result:

```sh
Comparing without_breakdown_(legacy_behavior)/renders_correctly.html.snap to with_both_granted_and_purchased_credits/renders_correctly.html.snap:

@@ -143,8 +143,13 @@
             </tr>
             <tr>
               <td class="body-2"></td>
+              <td class="body-2">Free credits</td>
+              <td class="body-2">-$3.00</td>
+            </tr>
+            <tr>
+              <td class="body-2"></td>
               <td class="body-2">Prepaid credits</td>
-              <td class="body-2">-$5.00</td>
+              <td class="body-2">-$2.00</td>
             </tr>
             <tr>
               <td class="body-2"></td>

Comparing without_breakdown_(legacy_behavior)/renders_correctly.html.snap to with_only_granted_credits/renders_correctly.html.snap:

@@ -143,7 +143,7 @@
             </tr>
             <tr>
               <td class="body-2"></td>
-              <td class="body-2">Prepaid credits</td>
+              <td class="body-2">Free credits</td>
               <td class="body-2">-$5.00</td>
             </tr>
             <tr>

Comparing without_breakdown_(legacy_behavior)/renders_correctly.html.snap to with_only_purchased_credits/renders_correctly.html.snap:

No difference.
```


app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -57,7 +57,7 @@ def call
           next if total_amount_cents <= 0
 
           wallet_transaction = create_wallet_transaction(wallet, total_amount_cents)
-          amount_cents = wallet_transaction.amount_cents
+          wallet_transaction.amount_cents
```
floganz: Why we need this call? It looks that we never use this value.

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -97,6 +96,36 @@ def schedule_webhook_notifications(wallet_transactions)
...
+
+      total_amount = wallet_transactions.sum(&:amount_cents)
+      result.prepaid_credit_amount_cents += total_amount
+      invoice.prepaid_credit_amount_cents += total_amount
+
+      calculate_prepaid_credit_breakdown(wallet_transactions)
+    end
+
+    def calculate_prepaid_credit_breakdown(wallet_transactions)
+      return if wallet_transactions.any? { |wt| !wt.wallet.traceable? }
```
floganz: Am I right that we'll skip breakdown if any of wallet transactions' wallet is not traceable?
Won't it break traceability for traceable wallets?

app/services/credits/applied_prepaid_credits_service.rb
groyoh: It was used before but moved to `update_prepaid_credit_amounts`. It's not used anymore. I guess what happened is I moved the code, ran Rubocop which found the variable unused and removed the variable so I didn't notice it 😅

app/services/credits/applied_prepaid_credits_service.rb
groyoh: The idea that I got was that it should be all-or-nothing per customer. Either all the customer's wallets are traceable or none are. This is how I planned the migration as well. I guess I forgot to add this logic to https://github.com/getlago/lago-api/pull/4787 as well though. I'll add this to my TODO.

spec/services/credits/applied_prepaid_credits_service_spec.rb
```diff
@@ -452,6 +452,77 @@
...
       end
+
+      it "sets prepaid_granted_credit_amount_cents on invoice" do
+        result
+
+        expect(invoice.prepaid_granted_credit_amount_cents).to eq(100)
+        expect(invoice.prepaid_purchased_credit_amount_cents).to be_nil
+      end
+
+      context "when inbound transaction is purchased" do
```
floganz, line 463: Should we add a case for non-traceable wallet and ensure that nil values are set for breakdown, while ensuring that total is set?

groyoh: I believe we do have that below.

PR #4818, opened 2026-01-08
Feat/fixed charges/main delete fc when addon deleted

annvelents:
## Context

When discarding an addon, related fixed charges should be discarded as well

## Description

During the process of discarding the add_on, also discard all related fixed_charges

### DIFF
--- a/app/services/add_ons/destroy_service.rb
+++ b/app/services/add_ons/destroy_service.rb
@@ -14,7 +14,7 @@ module AddOns
 
       ActiveRecord::Base.transaction do
         add_on.discard!
-        add_on.fixed_charges.discard_all!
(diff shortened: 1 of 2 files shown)


app/services/add_ons/destroy_service.rb
```diff
@@ -12,10 +12,17 @@ def initialize(add_on:)
     def call
       return result.not_found_failure!(resource: "add_on") unless add_on
 
-      add_on.discard!
+      ActiveRecord::Base.transaction do
+        add_on.discard!
+        add_on.fixed_charges.discard_all!
```
vincent-pochet: `discard_all!` is actually processing records 1 by 1. Not sure if it's an issue here, but we should consider using
```suggestion
        add_on.fixed_charges.update_all(deleted_at: Time.current)
```

PR #4828, opened 2026-01-12
feat(feature-flags): Introduce feature flags without Flipper

julienbourdeau:
I originally wanted to [add feature flags using Flipper gem](https://github.com/getlago/lago-api/pull/4767).

In the end, we'd rather store the flags in DB and not in Redis. Looking at the active_record adapter, I didn't like it too much because it stores data as key value thing, so you cannot eagerload flag for an organization.

I didn't want to make a custom implem'... but in the end, I think it's the best.

I added restriction that feature flag should only be handled by organization (as the flipper actor, so in the end, why not put it in the organizations table.

The org attributes is already cached in redis for "heavy" endpoints, so no need to add extra cache here.

### Premium integration copycat?

If you're familliar with premium integrations, it's basically the same. Premium integrations were even used as feature flags in the past.

Except Premium Integrations means "_features available for this org, as part of their contract_". 
They are not meant to be removed ever.

**The `feature_flags` attributes are meant to be TEMPORARY and added/removed constantly.**

### Pros

* No other gem needed
* Simple and extensible

### Cons

* We can't use the flipper UI, although we didn't plan to use it
* we cant enable flags per other actors (like billing entity / users), although we didn't plan to use it

## Usage

```rb
if hooli.feature_flag_enabled? :mutiple_payment_methods
  # ...
end
```

```rb
# Production: will return false
# Other env: raise an ArgumentError in any other env
hooli.feature_flag_enabled? :yolo 
```

In the console, you can enable/disable for an org

```rb
hooli.enable_feature_flag! :mutiple_payment_methods
hooli.disable_feature_flag! :mutiple_payment_methods
```


app/models/feature_flag.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def sanitize!
+      valid_keys = DEFINITION.keys
+
+      Organization.where.not(feature_flags: []).find_each do |organization|
+        valid_flags = organization.feature_flags & valid_keys
+
+        if valid_flags != organization.feature_flags
+          organization.update!(feature_flags: valid_flags)
+        end
```
mariohd: I believe if you call directly `update!` with the exact same flags the organization already has, it does not execute the update statement i.e saving unnecessary processing. 

So IMO it's safe to call this way
```suggestion
        organization.update!(feature_flags: valid_flags)
```

app/models/organization.rb
```diff
@@ -239,6 +239,39 @@ def organization
     self
   end
 
+  # Feature flag methods
```
mariohd: Instead of having this in the Organization class.
Why not having this in a `Concern`? 

So it's easily usable for reusable for other (if we decide to expand this). 

You can see the example as `app/models/concerns/organizations/authentication_methods.rb`

app/models/feature_flag.rb
julienbourdeau: absolutly!

app/models/organization.rb
julienbourdeau: Done! I agree

PR #4829, opened 2026-01-12
feat(fixed-charges): Bill pay-in-advance fixed charges correctly across subscription lifecycle

ancorcruz:
## Summary

  - Fix pay-in-advance fixed charge billing across subscription lifecycle events
  - Fix delta calculation when plan overrides create child fixed charges
  - Remove `billed_on_activation?` method that incorrectly combined billing logic

## Context

  Pay-in-advance fixed charges on pay-in-arrears plans were not being billed correctly during various subscription lifecycle events. The root cause was the `billed_on_activation?` method which combined two different billing paths (plan pay-in-advance and fixed charge pay-in-advance) into a single check, causing the wrong billing job to be called or billing to be skipped entirely.

  Additionally, when a subscription is updated with `plan_overrides`, it creates a child plan with new fixed charges (different IDs from the parent). The delta billing calculation couldn't find the previous fee because it only searched by the new child fixed charge ID, not the parent.

## Description

  **Separate billing paths for pay-in-advance plans vs fixed charges:**

  The fix separates the billing logic into two distinct paths:
  - Pay-in-advance **plans**: Use `BillSubscriptionJob` (bills subscription fee + charges)
  - Pay-in-advance **fixed charges** (on pay-in-arrears plans): Use `Invoices::CreatePayInAdvanceFixedChargesJob`

  This separation is applied consistently across:
  - `Subscriptions::CreateService` - subscription creation
  - `Subscriptions::ActivateService` - subscription activation
  - `Subscriptions::UpdateService` - subscription update with plan overrides
  - `Subscriptions::TerminateService` - termination with next subscription
  - `Subscriptions::FreeTrialBillingService` - trial ending
  - `Subscriptions::PlanUpgradeService` - plan upgrades

  **Fix delta calculation for plan overrides:**

  When looking for previously billed fees to calculate the delta, the service now searches for fees from both the current fixed charge AND its parent:

  ```ruby
  fixed_charge: [fixed_charge, fixed_charge.parent]

  This ensures correct delta billing when plan overrides create child fixed charges with different IDs.

  Test plan

  - Add regression test for fixed charge pay-in-advance with plan override scenarios
  - Test subscription creation with pay-in-advance fixed charges
  - Test subscription activation billing paths
  - Test subscription update with plan_overrides creating child fixed charges
  - Test delta calculation: increase (10 → 15), decrease (10 → 3), and multi-step (10 → 3 → 15)
  - Verify existing billing scenarios still pass

### DIFF
--- a/app/services/fees/build_pay_in_advance_fixed_charge_service.rb
+++ b/app/services/fees/build_pay_in_advance_fixed_charge_service.rb
@@ -67,7 +67,7 @@ module Fees
       existing_fees = Fee.where(
         organization:,
         subscription:,
-        fixed_charge:,
+        fixed_charge: [fixed_charge, fixed_charge.parent],
         fee_type: :fixed_charge
       ).where(
         "properties->>'fixed_charges_from_datetime' = ?",
--- a/app/services/subscriptions/update_service.rb
+++ b/app/services/subscriptions/update_service.rb
@@ -34,29 +34,24 @@ module Subscriptions
 
       return result.forbidden_failure! if !License.premium? && params.key?(:plan_overrides)
 
-      subscription.name = params[:name] if params.key?(:name)
-      subscription.ending_at = params[:ending_at] if params.key?(:ending_at)
-
-      if pay_in_advance? && params.key?(:on_termination_credit_note)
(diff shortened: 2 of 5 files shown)


annvelents (approved):
🙌

PR #4831, opened 2026-01-12
Multi Payments for Stripe webhooks write path

mariohd:
This PR updates the Stripe webhooks to handle PaymentMethods.

**Webhooks**:

1. `Stripe::Webhooks::SetupIntentSucceededService`
2. `Stripe::Webhooks::CustomerUpdatedService`
3. `Stripe::Webhooks::PaymentIntentSucceededService` *
      a. This one has no edits, it was done through `Payments::SetPaymentMethodAndCreateReceiptJob`
4. `Stripe::Payments::AuthorizeService`

**Services**:
`PaymentProviderCustomers::Stripe::UpdatePaymentMethodService`

**Jobs**:
`Payments::SetPaymentMethodAndCreateReceiptJob`

### Feature Flag

- multiple_payment_methods

### DIFF
--- a/app/services/payment_methods/create_from_provider_service.rb
+++ b/app/services/payment_methods/create_from_provider_service.rb
@@ -21,10 +21,9 @@ module PaymentMethods
         payment_method.provider_method_type = provider_method_type
         payment_method.provider_method_id = provider_method_id
         payment_method.payment_provider_id = payment_provider_id
-        payment_method.is_default = !customer.payment_methods.exists?(is_default: true)
       end
-
       payment_method.save!
+      PaymentMethods::SetAsDefaultService.call(payment_method:)
 
       result.payment_method = payment_method
       result
--- a/app/services/payment_provider_customers/stripe/update_payment_method_service.rb
+++ b/app/services/payment_provider_customers/stripe/update_payment_method_service.rb
@@ -18,10 +18,23 @@ module PaymentProviderCustomers
         stripe_customer.save!
 
         if stripe_customer.organization.feature_flag_enabled?(:multiple_payment_methods)
-          payment_method = PaymentMethod.find_by(payment_provider_customer: stripe_customer)
+          if payment_method_id
+            payment_method = PaymentMethod.find_by(
+              customer:,
+              payment_provider_customer: stripe_customer,
+              provider_method_id: payment_method_id
+            )
 
-          if payment_method && payment_method_id
-            payment_method.update!(provider_method_id: payment_method_id)
+            payment_method ||= PaymentMethods::CreateFromProviderService.call(
+              customer:,
+              params: {provider_payment_methods: stripe_customer.provider_payment_methods},
+              provider_method_id: payment_method_id,
+              payment_provider_id: stripe_customer.payment_provider_id,
+              payment_provider_customer: stripe_customer
+            ).payment_method
+
+            PaymentMethods::SetAsDefaultService.call(payment_method:)
+
             result.payment_method = payment_method
           end
         end
--- a/app/services/payment_providers/stripe/webhooks/setup_intent_succeeded_service.rb
+++ b/app/services/payment_providers/stripe/webhooks/setup_intent_succeeded_service.rb
@@ -19,18 +19,6 @@ module PaymentProviders
             payment_method_id: payment_method_id
           ).raise_if_error!
 
-          if stripe_customer.organization.feature_flag_enabled?(:multiple_payment_methods)
-            pm_result = PaymentMethods::CreateFromProviderService.call(
-              customer:,
-              params: {provider_payment_methods: stripe_customer.provider_payment_methods},
-              provider_method_id: payment_method_id,
-              payment_provider_id: stripe_customer.payment_provider_id,
-              payment_provider_customer: stripe_customer
-            )
-
-            result.payment_method = pm_result.payment_method
-          end
-
           result.stripe_customer = stripe_customer
           result
         rescue ::Stripe::PermissionError => e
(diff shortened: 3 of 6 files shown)


app/services/payment_provider_customers/stripe/update_payment_method_service.rb
```diff
@@ -17,6 +17,15 @@ def call
         stripe_customer.payment_method_id = payment_method_id
         stripe_customer.save!
 
+        if stripe_customer.organization.feature_flag_enabled?(:mutiple_payment_methods)
+          payment_method = PaymentMethod.find_by(payment_provider_customer: stripe_customer)
+
```
lovrocolic, line 37: If there are multiple payment methods attached on the customer, I guess this can return invalid PM when it is called from `PaymentProviders::Stripe::Webhooks::SetupIntentSucceededService` or `CustomerUpdatedService`

mariohd: Why this would be invalid? (in this scenario)

I believe there should be one PaymentProviderCustomer to one PaymentMethod, no?

lovrocolic: I think that payment_provider_customer_id can be present in multiple payment methods, that are linked to the same customer

mariohd: I'm not able to visualize this scenario, we are doing basically the same while fetching 
```
PaymentProviderCustomers::StripeCustomer
            .by_provider_id_from_organization(organization.id, stripe_customer_id)
            .first
```
Checking the data, only 1 customer has more than one associated `PaymentProviderCustomers` and it's a test customer in our own test org, probably done manually. 

can you give me some example of this scenario that this would break?

mariohd: Got it what you said, i've did a wrote a new logic. Can you please check again?

app/services/payment_methods/create_from_provider_service.rb
```diff
@@ -21,9 +21,10 @@ def call
         payment_method.provider_method_type = provider_method_type
         payment_method.provider_method_id = provider_method_id
         payment_method.payment_provider_id = payment_provider_id
-        payment_method.is_default = !customer.payment_methods.exists?(is_default: true)
       end
 
+      PaymentMethods::SetAsDefaultService.call(payment_method:)
+
       payment_method.save!
```
aquinofb: nit: if this is a new record, we might want to save first and then set as default. The reason is that `SetAsDefaultService` also saves the record (via `update!`).

That's a super minor thing because it technically works as expected, but given the service already performs db operations (i.e., `where.not(id: payment_method.id)`) and can evolve in the future, passing a persisted record could make it slightly more "future-proof"?

app/services/payment_provider_customers/stripe/update_payment_method_service.rb
```diff
@@ -17,6 +17,28 @@ def call
...
+              provider_method_id: payment_method_id
+            )
+
+            payment_method ||= PaymentMethods::CreateFromProviderService.call(
+              customer:,
+              params: {provider_payment_methods: stripe_customer.provider_payment_methods},
+              provider_method_id: payment_method_id,
+              payment_provider_id: stripe_customer.payment_provider_id,
+              payment_provider_customer: stripe_customer
+            ).payment_method
```
aquinofb, line 34: nit: maybe extract this lad to a `find_or_create_payment_method(psp_customer, pm_id)`?

mariohd: IMO that would be necessary just in case this would be called somewhere else.

app/services/payment_methods/create_from_provider_service.rb
mariohd: Excellent catch!

app/jobs/payments/set_payment_method_and_create_receipt_job.rb
```diff
@@ -13,6 +13,12 @@ class SetPaymentMethodAndCreateReceiptJob < ApplicationJob
     def perform(payment:, provider_payment_method_id:)
       ::Payments::SetPaymentMethodDataService.call!(payment:, provider_payment_method_id:)
 
+      payment_method = payment.customer.payment_methods.find_by(provider_method_id: provider_payment_method_id)
+      if payment_method.present?
+        payment.payment_method = payment_method
+        payment.save!
+      end
```
aquinofb, line 31: nit: shall we "name" this action to something like `define_local_payment_method!`? The prefix could be `define_`, `assign_`, or maybe `set_`? WDYT?

aquinofb: Also, the `SetPaymentMethodDataService` assigns a bunch of details from the remote pm to the local payment, shall we move this assignment to within that service? 🤔

mariohd: I think this should not belong to `SetPaymentMethodDataService` for one reason; the `Data` in the name. 

`PaymentMethod` already have the `details` attribute that stores this data.
What we can do is set the payment_method to payment before calling the service and also updates the `details` there. 

I believe this service was edited by @lovrocolic in other PR, right?

lovrocolic: yep I touched details service in another PR.

Btw, TBH I am not 100% sure if we need to assign payment method to payment on payment intent success webhook. This relation is established when we create payment intent. However, it does not do any harm, so I guess we can keep it. Also, maybe I am missing sth

PR #4833, opened 2026-01-13
Feat(metadata): add metadata on wallet

annvelents:
## Context

Sometimes on the wallet we need to store some additional values and external ids. So we're providing a mechanism to store it, adding metadata

## Description

Added ItemMetadata relation to wallets

### DIFF
--- a/app/controllers/api/v1/wallets/metadata_controller.rb
+++ b/app/controllers/api/v1/wallets/metadata_controller.rb
@@ -49,7 +49,7 @@ module Api
         private
 
         def metadata_params
-          {metadata: params.fetch(:metadata, {}).permit!.to_h}
+          params.permit(metadata: {}).to_h
         end
 
         def render_metadata
--- a/app/controllers/api/v1/plans/metadata_controller.rb
+++ b/app/controllers/api/v1/plans/metadata_controller.rb
@@ -49,7 +49,7 @@ module Api
         private
 
         def metadata_params
-          {metadata: params.fetch(:metadata, {}).permit!.to_h}
+          params.permit(metadata: {}).to_h
         end
 
         def render_metadata
--- a/spec/requests/api/v1/plans/metadata_controller_spec.rb
+++ b/spec/requests/api/v1/plans/metadata_controller_spec.rb
@@ -74,12 +74,12 @@ RSpec.describe Api::V1::Plans::MetadataController do
     context "when metadata param is not provided" do
       subject { post_with_token(organization, "/api/v1/plans/#{plan_code}/metadata", {}) }
 
-      it "creates metadata with empty hash" do
+      it "does not create metadata" do
         subject
 
         expect(response).to have_http_status(:success)
-        expect(json[:metadata]).to eq({})
-        expect(plan.reload.metadata.value).to eq({})
+        expect(json[:metadata]).to eq(nil)
+        expect(plan.reload.metadata).to eq(nil)
       end
     end
   end
--- a/spec/requests/api/v1/wallets/metadata_controller_spec.rb
+++ b/spec/requests/api/v1/wallets/metadata_controller_spec.rb
@@ -75,12 +75,12 @@ RSpec.describe Api::V1::Wallets::MetadataController do
     context "when metadata param is not provided" do
       subject { post_with_token(organization, "/api/v1/wallets/#{wallet_id}/metadata", {}) }
 
-      it "creates metadata with empty hash" do
+      it "does not create empty metadata" do
         subject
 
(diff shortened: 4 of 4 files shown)


app/services/wallets/create_service.rb
```diff
@@ -172,5 +174,12 @@ def payment_method
 
       @payment_method = PaymentMethod.find_by(id: params[:payment_method][:payment_method_id], organization_id:)
     end
+
+    def create_metadata(wallet, metadata_value)
+      wallet.create_metadata!(
```
groyoh: I'm guessing you could use `Metadata::UpdateItemService` here as well.

app/graphql/types/wallets/object.rb
```diff
@@ -49,6 +49,8 @@ class Object < Types::BaseObject
 
       field :applies_to, Types::Wallets::AppliesTo, null: true, method: :itself
 
+      field :metadata, [Types::Metadata::Object], null: true
```
groyoh, line 52: I wonder if we couldn't have a `Types::Metadata::Array` type to avoid the metadata remapping below. Might be a good refactor for later.

app/services/wallets/update_service.rb
```diff
@@ -154,5 +157,12 @@ def payment_method
 
       @payment_method = PaymentMethod.find_by(id: params[:payment_method][:payment_method_id], organization_id: wallet.organization_id)
     end
+
+    def update_metadata!
+      return unless params.key?(:metadata)
+
+      value = params[:metadata]&.then { |m| m.respond_to?(:to_unsafe_h) ? m.to_unsafe_h : m.to_h }
```
groyoh: Why would we use `to_unsafe_h` ?

config/routes.rb
```diff
@@ -142,7 +142,13 @@
         post :payment_url, on: :member
       end
       get "/wallets/:id/wallet_transactions", to: "wallet_transactions#index"
-      resources :wallets, only: %i[create update show index]
+      resources :wallets, only: %i[create update show index] do
+        scope module: :wallets do
+          resource :metadata, only: %i[create update destroy] do
```
groyoh, line 147: I might be missing something but where are those controller ?

groyoh (commented):
A few minor feedback. Also I believe the GraphQL type tests were not updated.

config/routes.rb
annvelents: on my local machine 🤦‍♀️ 
problem when you add files with `git add -p` 🤦‍♀️

app/services/wallets/update_service.rb
annvelents: right, it's not needed. Was not sure as we're sending params and not permitting metadata keys, but was wrong 🤦

app/graphql/types/wallets/object.rb
annvelents: I reused the existing type changing the way we're getting the values + updated all the places where it was used

app/controllers/api/v1/wallets_controller.rb
```diff
@@ -133,6 +134,7 @@ def update_params
           :invoice_requires_successful_payment,
           :paid_top_up_min_amount_cents,
           :paid_top_up_max_amount_cents,
+          metadata: {},
```
groyoh, line 137: I wonder if we should add metadata in create/update given we have specific endpoints.

app/controllers/api/v1/wallets/metadata_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+            render_metadata
+          else
+            render_error_response(result)
+          end
+        end
+
+        private
+
+        def metadata_params
+          {metadata: params.fetch(:metadata, {}).permit!.to_h}
```
groyoh: `permit!` seems dangerous here. Does this work ?

```suggestion
          {metadata: params.expect(metadata: {})}
```

groyoh (approved):
We're still missing tests for the GraphQL objects.

app/controllers/api/v1/wallets_controller.rb
annvelents: For now it just follows pattern of plans and credit_notes
Also, I think it does make sense to allow to create object with related metadata. For wallet especially we do have a possibility to create it even with transactions and transaction_metadata... so that's odd to have a possibility to create metadata for recurring transaction while creating the wallet and not have it when creating it for the wallet itself

app/controllers/api/v1/wallets/metadata_controller.rb
annvelents: yes, let me check

PR #4838, opened 2026-01-13
feat(wallet): Add API endpoints for consumption/funding traceability

groyoh:
## Context

## Description

This implements step #5 of the above breakdown: Retrieving Consumptions and Fundings via API.

Add two new REST API endpoints on wallet transactions:
- `GET /wallet_transactions/:id/consumptions` - returns consumption records for inbound transactions showing which outbound transactions consumed from them
- `GET /wallet_transactions/:id/fundings` - returns funding records for outbound transactions showing which inbound transactions funded them

Both endpoints include pagination support and nested transaction data. Validation ensures the wallet is traceable and the transaction type matches the endpoint (inbound for consumptions, outbound for fundings).

This also adds fields to existing APIs:

- Handle wallet transaction `priority` in request and responses
- Adds `remaining_amount_cents` to WalletTransactionSerializer

### DIFF
--- a/app/controllers/api/v1/wallets_controller.rb
+++ b/app/controllers/api/v1/wallets_controller.rb
@@ -77,6 +77,7 @@ module Api
           :paid_top_up_max_amount_cents,
           :ignore_paid_top_up_limits_on_creation,
           :transaction_name,
+          :transaction_priority,
           metadata: {},
           transaction_metadata: [
             :key,
--- a/app/services/wallets/create_service.rb
+++ b/app/services/wallets/create_service.rb
@@ -111,6 +111,7 @@ module Wallets
           source: :manual,
           metadata: params[:transaction_metadata],
           name: params[:transaction_name],
+          priority: params[:transaction_priority],
           ignore_paid_top_up_limits: params[:ignore_paid_top_up_limits_on_creation]
         }
       )
--- a/spec/graphql/mutations/wallets/create_spec.rb
+++ b/spec/graphql/mutations/wallets/create_spec.rb
@@ -136,7 +136,16 @@ RSpec.describe Mutations::Wallets::Create do
 
     expect(WalletTransactions::CreateJob).to have_been_enqueued.with(
       organization_id: membership.organization.id,
-      params: {wallet_id: Regex::UUID, paid_credits: "10.00", granted_credits: "0.00", source: :manual, metadata: nil, name: "Initial Credits Purchase", ignore_paid_top_up_limits: nil}
+      params: {
+        wallet_id: Regex::UUID,
+        paid_credits: "10.00",
+        granted_credits: "0.00",
+        source: :manual,
+        metadata: nil,
+        priority: nil,
+        name: "Initial Credits Purchase",
+        ignore_paid_top_up_limits: nil
+      }
     )
     expect(SendWebhookJob).to have_been_enqueued.with("wallet.created", Wallet)
   end
@@ -268,7 +277,16 @@ RSpec.describe Mutations::Wallets::Create do
 
       expect(WalletTransactions::CreateJob).to have_been_enqueued.with(
         organization_id: membership.organization.id,
-        params: {wallet_id: Regex::UUID, paid_credits: "10.00", granted_credits: "0.00", source: :manual, metadata: nil, name: nil, ignore_paid_top_up_limits: nil}
+        params: {
+          wallet_id: Regex::UUID,
+          paid_credits: "10.00",
+          granted_credits: "0.00",
+          source: :manual,
+          metadata: nil,
+          priority: nil,
+          name: nil,
+          ignore_paid_top_up_limits: nil
+        }
       )
     end
   end
--- a/spec/requests/api/v1/wallets_controller_spec.rb
+++ b/spec/requests/api/v1/wallets_controller_spec.rb
@@ -197,6 +197,32 @@ RSpec.describe Api::V1::WalletsController do
       end
     end
 
+    context "when transaction_priority is provided" do
+      let(:create_params) do
+        {
+          external_customer_id: customer.external_id,
+          rate_amount: "1",
+          name: "Wallet1",
+          currency: "EUR",
+          paid_credits: "10",
+          granted_credits: "10",
+          expiration_at:,
+          transaction_priority: 5
+        }
+      end
+
+      before { subject }
+
+      it "schedules a WalletTransactions::CreateJob with the transaction priority" do
+        expect(WalletTransactions::CreateJob).to have_been_enqueued.with(
+          organization_id: organization.id,
+          params: hash_including(
+            priority: 5
+          )
+        )
+      end
+    end
+
     context "with recurring transaction rules" do
       around { |test| lago_premium!(&test) }
 
--- a/spec/scenarios/wallets/topup_with_priority_spec.rb
+++ b/spec/scenarios/wallets/topup_with_priority_spec.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "Wallet Transaction with priority" do
+  let(:organization) { create(:organization, webhook_url: nil) }
+  let(:customer) { create(:customer, organization:) }
+
+  context "when creating a wallet with transaction_priority" do
+    it "sets the priority on the initial wallet transactions" do
+      wallet = create_wallet({
+        external_customer_id: customer.external_id,
+        rate_amount: "1",
+        currency: "EUR",
+        paid_credits: "100",
+        granted_credits: "50",
+        transaction_priority: 5
+      }, as: :model)
+
+      expect(wallet.wallet_transactions.count).to eq(2)
+      expect(wallet.wallet_transactions.pluck(:priority)).to eq([5, 5])
+    end
+  end
+
+  context "when creating a wallet without transaction_priority" do
+    it "uses the default priority on the initial wallet transactions" do
+      wallet = create_wallet({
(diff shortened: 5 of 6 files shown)


app/controllers/api/v1/wallet_transactions_controller.rb
```diff
@@ -80,8 +80,39 @@ def payment_url
...
+
+        includes = (direction == :consumptions) ? %i[outbound_wallet_transaction] : %i[inbound_wallet_transaction]
+
+        render(
+          json: ::CollectionSerializer.new(
+            result.wallet_transaction_consumptions.includes(includes),
+            ::V1::WalletTransactionConsumptionSerializer,
+            collection_name: "wallet_transaction_consumptions",
+            meta: pagination_metadata(result.wallet_transaction_consumptions),
+            includes: includes
```
ancorcruz:
```suggestion
            includes:
```

app/queries/wallet_transaction_consumptions_query.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class WalletTransactionConsumptionsQuery < BaseQuery
```
ancorcruz, line 3: Would you write a contract for this Query object? This should enforce direction is one of the two accepted values for example, otherwise I miss query object tests with unexpected `direction` value.

spec/queries/wallet_transaction_consumptions_query_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+      let(:outbound_transaction1) { create(:wallet_transaction, wallet:, transaction_type: "outbound") }
+      let(:outbound_transaction2) { create(:wallet_transaction, wallet:, transaction_type: "outbound") }
+      let!(:consumption1) do
+        create(:wallet_transaction_consumption,
+          organization:,
+          inbound_wallet_transaction: inbound_transaction,
+          outbound_wallet_transaction: outbound_transaction1,
+          consumed_amount_cents: 500)
+      end
+      let!(:consumption2) do
```
ancorcruz: we prefer `let` over `let!` and then create the records within a before block for example.

spec/queries/wallet_transaction_consumptions_query_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+      let(:inbound_transaction1) { create(:wallet_transaction, wallet:, transaction_type: "inbound", remaining_amount_cents: 10000) }
+      let(:inbound_transaction2) { create(:wallet_transaction, wallet:, transaction_type: "inbound", remaining_amount_cents: 10000) }
+      let!(:funding1) do
+        create(:wallet_transaction_consumption,
+          organization:,
+          inbound_wallet_transaction: inbound_transaction1,
+          outbound_wallet_transaction: outbound_transaction,
+          consumed_amount_cents: 500)
+      end
+      let!(:funding2) do
```
ancorcruz: same here.

app/serializers/v1/wallet_transaction_consumption_serializer.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module V1
+  class WalletTransactionConsumptionSerializer < ModelSerializer
+    def serialize
+      payload = {
+        lago_id: model.id,
+        lago_inbound_wallet_transaction_id: model.inbound_wallet_transaction_id,
+        lago_outbound_wallet_transaction_id: model.outbound_wallet_transaction_id,
```
ancorcruz: what is the point of having these ids if the actual records could be included?

spec/requests/api/v1/invoices_controller_spec.rb
```diff
@@ -196,6 +196,9 @@
           lago_id: invoice.id,
           payment_status: invoice.payment_status,
           status: invoice.status,
+          prepaid_credit_amount_cents: 0,
+          prepaid_granted_credit_amount_cents: nil,
+          prepaid_purchased_credit_amount_cents: nil,
```
ancorcruz, line 205: shouldn't be all zero instead of nil values? for consistency.

app/controllers/api/v1/wallet_transactions_controller.rb
```diff
@@ -80,8 +80,39 @@ def payment_url
...
+        wallet_transaction_consumptions(direction: :consumptions)
+      end
+
+      def fundings
+        wallet_transaction_consumptions(direction: :fundings)
+      end
+
       private
 
+      def wallet_transaction_consumptions(direction:)
```
ancorcruz, line 94: we could define a single controller action with two routes and default values here instead... both options are good anyway.

```
# config/routes.rb

        get :consumptions, on: :member, to: 'wallet_transactions#consumptions', defaults: { direction: :consumptions)
        get :fundings, on: :member, to: 'wallet_transactions#consumptions', defaults: { direction: :fundings)

```

then in the controller a single action will do it;

```
      def consumptions
         # then use: params[:direction]
         ...

```

app/serializers/v1/wallet_transaction_consumption_serializer.rb
groyoh: That's a good point. I'll check whether we have this "double id/include" pattern else where and cleanup if necessary.

app/queries/wallet_transaction_consumptions_query.rb
groyoh: Indeed I checked the compatibility between direction and transaction_status but not an invalid direction. I'll actually add an `ArgumentError` instead here as wrong direction would mean broken code.

app/controllers/api/v1/wallet_transactions_controller.rb
groyoh: I wonder why Rubocop doesn't fix/complain on its own 🤔

app/controllers/api/v1/wallet_transactions_controller.rb
groyoh: That's a very clever way to solve this but I feel like it adds some unnecessary indirection/magic and I believe the explicitness of the current solution is clearer to understand.

spec/requests/api/v1/invoices_controller_spec.rb
groyoh: Not really. These values are set from invoices on which we applied traceable prepaid credit. For historical data, they will be kept to `nil` to express "we don't know how much free/paid credit were applied". Does that make sense ?

spec/queries/wallet_transaction_consumptions_query_spec.rb
groyoh: I'm okay with that but based on the code base, I feel like this is more a personal choice 😅 If it really is the case, we should probably enable a Rubocop rule for this. But I don't think it adds more clarity personally.

app/serializers/v1/wallet_transaction_consumption_serializer.rb
groyoh: One thing though, for `fundings` we'll want to includes the inbound transactions but we won't include the outbound transaction as it is unique and would be repeated multiple times. And it goes the other way around for `consumptions`. So I guess it still makes sense to keep the id and include.

Also I wouldn't remove the includes since there's no way to retrieve the wallet transactions in bulk via ids. Maybe I should implement such feature instead and remove the includes altogether ? 🤔

ancorcruz: if the API consumer is not able to retrieve the transactions by id, what is the point of sending the ID in the response?

groyoh: Well, you could retrieve via `GET /api/v1/wallet_transactions/:id` individually. I get your point and I'm not sure what's best here to be honest.

floganz: For `outbound` there's no much sense into `lago_outbound_wallet_transaction_id` field because it will have the value of transaction we made request with. `lago_inbound_wallet_transaction_id` in this case will contain id of related inbound transaction. But here I see two issues: data duplication and list of related transaction can have more that 1 record. Last one adds ambiguity what `id` we should display in `lago_inbound_wallet_transaction_id`.
Similar will be for `inbound` transactions and `consumptions` endpoint.
Considering this I'd sugest to remove both `lago_inbound_wallet_transaction_id` and `lago_outbound_wallet_transaction_id` fields.

floganz: After some testing I've noticed that for any particular transaction only one endpoint is working.
Meaning that `outbound` transaction will only capable to use `funnding` endpoint, for `inbound` type only `consumptions` will work. Maybe it will make sense to have two serializers?

groyoh: > Considering this I'd sugest to remove both lago_inbound_wallet_transaction_id and lago_outbound_wallet_transaction_id fields.

I agree that we already have the ID in the path but the idea was to provide enough context within the payload itself so that it's self-describing. That said, I can remove it if needed.

What I wonder though is whether I should call it `funding_wallet_transactions` and `consuming_wallet_transactions` instead of `inbound_wallet_transactions` and `outbound_wallet_transactions` in the payload.

> After some testing I've noticed that for any particular transaction only one endpoint is working.
Meaning that outbound transaction will only capable to use funnding endpoint, for inbound type only consumptions will work. Maybe it will make sense to have two serializers?

That is indeed on purpose. There's no possibility of "fundings" for an inbound and vice/versa. I don't think it warrants two serializer for now.

PR #4839, opened 2026-01-13
feat(Adding Email event to Activity logs)

nepalez:
## Roadmap Task

[Spec](https://www.notion.so/getlago/Spec-Improving-the-Email-experience-2d5ef63110d2801ea42bc51689bb4c1d)
[Dive-In](https://www.notion.so/getlago/BE-Dive-In-Adding-Email-event-to-Activity-logs-2e2ef63110d280dd986ad8916cc58400)

## Context

When invoices, credit notes, or payment receipts are sent by email, these events should be tracked in ClickHouse with information about who triggered the send and whether it succeeded or failed.

## Description

Add email event logging to the Activity Logs system in ClickHouse.

**New activity type:** `email.sent`

The activity source indicates how the email was triggered:
- `system` — automatic sends (e.g., invoice finalization)
- `front` — user-initiated resends from UI (passes `user_id`)
- `api` — API-triggered sends (passes `api_key_id`)

**Changes:**

- Create `Utils::EmailActivityLog` service that produces email events to Kafka
- Extend `SendEmailJob` to log activity after successful delivery or after all retries exhausted
- Normalize mailer interface with `loggable?`, `document`, and `log` methods in `ApplicationMailer`
- Update `InvoiceMailer`, `CreditNoteMailer`, and `PaymentReceiptMailer` to support logging
- Add `email.sent` activity type and `PaymentReceipt` resource type to ClickHouse model
- Rename `InvoiceMailer#finalized` to `#created` for consistency with other mailers

**Activity object structure:**

```json
{
  "status": "sent|resent|failed",
  "email": {
    "subject": "...",
    "to": ["..."],
    "cc": [],
    "bcc": [],
    "body_preview": "..."
  },
  "document": {
    "type": "Invoice|CreditNote|PaymentReceipt",
    "number": "...",
    "lago_id": "..."
  },
  "error": { "class": "...", "message": "..." }
}
```

> Note: the `error` field is only present when status is `failed`

### DIFF
--- a/spec/graphql/resolvers/activity_log_resolver_spec.rb
+++ b/spec/graphql/resolvers/activity_log_resolver_spec.rb
@@ -51,9 +51,7 @@ RSpec.describe Resolvers::ActivityLogResolver, clickhouse: true do
     it_behaves_like "blocked feature", "feature_unavailable"
   end
 
-  context "with premium feature" do
-    around { |test| lago_premium!(&test) }
-
+  context "with premium feature", :premium do
     it "returns a single activity log with parsed object changes" do
       result = execute_graphql(
         current_user: membership.user,


app/jobs/send_email_job.rb
```diff
@@ -1,10 +1,67 @@
...
+#
+#      CreditNoteMailer.with(
+#        credit_note:,
+#        api_key_id: CurrentContext.api_key_id
+#      ).created.deliver_later
+#
 class SendEmailJob < ActionMailer::MailDeliveryJob
   queue_as "mailers"
 
+  after_perform :log
```
julienbourdeau, line 34: 👌

app/mailers/application_mailer.rb
```diff
@@ -12,4 +12,19 @@ def set_shared_variables
...
+
+  def document
+  end
+
+  def created
+  end
+
+  def log(**options)
+    Utils::EmailActivityLog.produce(document:, message:, **options) if loggable? && message.to.present?
+  end
```
julienbourdeau, line 25: I'm not a big fan of `document` and `created` here. I feel like there is no guarantee that all emails to log will have a document. The same way, the `created` name works great here for document but not for other thing.

Another email that I have in mind is the alerting email. **Note that it doesn't exist yet** because we only handle webhook but we could send emails to end customers when they are getting close to there quota.

Maybe a good way is to introduce a DocumentMailer class with this code and make Invoice/CreditNote/ Receipt mailer extend it.

What do you think?


```suggestion
  # Shared interface for logging purposes
  def loggable?
    false
  end

  def log(**options)
  end
```

```rb
class DocumentMailer < ApplicationMailer
  def loggable?
    true
  end

  def document
  end

  def created
  end

  def log(**options)
    Utils::EmailActivityLog.produce(document:, message:, **options) if loggable? && message.to.present?
  end
end
```

julienbourdeau (approved):

app/mailers/application_mailer.rb
nepalez: Yeah, I like this. Will fix

nepalez: Fixed, and also supported the very possibility to log emails without document

julienbourdeau (approved):

PR #4849, opened 2026-01-14
feat(alerting): Better error management

julienbourdeau:
The original loop was design so "if an error happens with processive billing, alerting still runs". It's good but we also need "if any alert fails, the other alert still runs.

Review without whitespace


groyoh (approved):
Should we maybe have an `AggregatedError` which wraps all the errors ?

julienbourdeau:
Good idea, that would be awesome

PR #4855, opened 2026-01-15
fix(redis): Config for subscriptions refresh store

vincent-pochet:
## Description

This PR
- Addresses https://github.com/getlago/lago/issues/650. 
- Fixes the redis config for the staging environment where the `LAGO_REDIS_STORE_URL` is passed with the `redis://` prefix, leading to an invalid `redis_client` instance
- Check for presence of `LAGO_REDIS_STORE_PASSWORD` and `LAGO_REDIS_STORE_DB` before using them to init the redis client instance
- Adds an optional `LAGO_REDIS_STORE_SSL` to force SSL connection when no `rediss://` prefix is passed in `LAGO_REDIS_STORE_URL`

### DIFF
--- a/app/services/subscriptions/consume_subscription_refreshed_queue_service.rb
+++ b/app/services/subscriptions/consume_subscription_refreshed_queue_service.rb
@@ -41,9 +41,6 @@ module Subscriptions
 
       config = {
         url:,
-        ssl_params: {
-          verify_mode: OpenSSL::SSL::VERIFY_NONE
-        },
         timeout: 5.0,
         reconnect_attempts: 3
       }
@@ -53,6 +50,10 @@ module Subscriptions
 
       if ENV["LAGO_REDIS_STORE_SSL"].present? || ENV["LAGO_REDIS_STORE_URL"].start_with?(/rediss?:/)
         config[:ssl] = true
+      end
+
+      if ENV["LAGO_REDIS_STORE_DISABLE_SSL_VERIFY"].present?
+        config[:ssl_params] = {verify_mode: OpenSSL::SSL::VERIFY_NONE}
       end
 
       @redis_client ||= Redis.new(config)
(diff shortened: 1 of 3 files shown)


app/services/subscriptions/consume_subscription_refreshed_queue_service.rb
```diff
@@ -33,20 +33,29 @@ def call
...
+        ENV["LAGO_REDIS_STORE_URL"]
+      else
+        "redis://#{ENV["LAGO_REDIS_STORE_URL"]}"
+      end
+
+      config = {
+        url:,
         ssl_params: {
-          verify_mode: OpenSSL::SSL::VERIFY_PEER
+          verify_mode: OpenSSL::SSL::VERIFY_NONE
```
groyoh: I believe this should be configurable. The library default (and in general) is set to `VERIFY_PEER` for good reasons.

vincent-pochet: @groyoh 
It is set with VERIFY_NONE in every other redis config of the app. 
The issue with `VERIFY_PEER` is that it does not plays well with self-signed certificates... 

So sure we could make it configurable, but it's a broader topic :)

vincent-pochet: You can find more details on https://github.com/getlago/lago/issues/614 and https://github.com/getlago/lago/issues/650

groyoh: > The issue with VERIFY_PEER is that it does not plays well with self-signed certificates...

Couldn't they add their self-signed certificates to the trust store (ex: https://documentation.ubuntu.com/server/how-to/security/install-a-root-ca-certificate-in-the-trust-store/) ? 🤔

vincent-pochet: They probably could.
In the meantime (mainly because I'm stuck with it for some tests on staging :smile: ), I added a new `LAGO_REDIS_STORE_DISABLE_SSL_VERIFY` to allow turning off the verify mode from the config

PR #4860, opened 2026-01-15
fix: change method name to not mix it with association

lovrocolic:
## Context

In payment model, payment_method association share the same name with existing method

## Description

This PR changes the method name so that we have clear distinction

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -80,7 +80,7 @@ class Payment < ApplicationRecord
     payment_provider&.payment_type
   end
 
-  def payment_method_display
+  def method_display_name
     return nil if provider_payment_method_data.blank?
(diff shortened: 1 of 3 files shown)


mariohd (approved):

app/models/payment.rb
```diff
@@ -80,7 +80,7 @@ def payment_provider_type
     payment_provider&.payment_type
   end
 
-  def payment_method
+  def payment_method_display
```
julienbourdeau: I'd rather `*_display_name` than just `*_display.
Since we're already in Payment model, I'd even go with `method_display_name`.

lovrocolic: sounds like a plan

PR #4862, opened 2026-01-15
fix(billing): Exclude removed items from tier assignment

rsempe:
## Description

Items removed before a billing period were still "occupying" tier slots in the prorated graduated charge calculation, causing active items to be pushed into higher tiers.

In `ProratedGraduatedService.compute_amount`, `full_sum` determines tier boundaries. It was incrementing for ALL add events, including those with `prorated_value = 0` (items removed before the period). This fix skips ghost add events and their corresponding orphan remove events to ensure only active items affect tier assignment.


annvelents (approved):
NIce one!

PR #4865, opened 2026-01-16
feat(offset-invoice-on-pay-in-advance-termination) Add `offset` value to Subscription `on_termination_credit_note` enum

vankiru:
## Context

This is the first step toward supporting invoice offset when terminating a pay-in-advance subscription. 

## Description

The PR just adds a new `offset` value to the db, subscription model, and GraphQL API.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -8300,6 +8300,7 @@ input OktaLoginInput {
 
 enum OnTerminationCreditNoteEnum {
   credit
+  offset
   refund
   skip
 }
--- a/schema.json
+++ b/schema.json
@@ -40169,6 +40169,12 @@
               "description": null,
               "isDeprecated": false,
               "deprecationReason": null
+            },
+            {
+              "name": "offset",
+              "description": null,
+              "isDeprecated": false,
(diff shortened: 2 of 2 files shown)


db/migrate/20260115164124_add_offset_value_to_subscription_on_termination_credit_note_enum.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddOffsetValueToSubscriptionOnTerminationCreditNoteEnum < ActiveRecord::Migration[8.0]
+  def up
+    add_enum_value :subscription_on_termination_credit_note, "offset", if_not_exists: true
+  end
+
+  def down
+  end
```
ancorcruz, line 9: should down remove the enum value?

vankiru: Postgres doesn't support removing enum values, the only way to do it is to recreate the whole enum without that value. So, no rollback here.

PR #4869, opened 2026-01-16
feat(credit_notes): Add new business logic for applying credit note to parent invoice

D1353L:
## Context
PR 2/3

Currently, when a credit note is created against an invoice, the credit amount is stored in a "credit wallet" that can only be consumed against future invoices. This doesn't allow customers to reduce the amount due on the original invoice directly.
We want to allow users, when creating a credit note, to explicitly choose how the credit note total is allocated, including the portion that should directly reduce the parent invoice’s outstanding balance.

## Description

This pull request adds business logic for applying credit notes to parent invoices through an "offset" mechanism. The feature allows credit notes to directly reduce the amount due on an invoice rather than creating wallet credits or refunds. This is implemented through a new offset_amount_cents field on credit notes and a new InvoiceSettlement tracking system.

Changes:

- Added offset_amount_cents and offset_amount_currency columns to the credit_notes table
- Introduced InvoiceSettlements::CreateService to track how credit notes and payments settle invoices
- Updated payment and invoice logic to account for offset amounts when calculating due amounts and payment status
- Enhanced validation logic to handle offset amounts for both regular and credit invoices (prepaid credits)

Changes to GraphQL, PDF and Email templates will be in PR #4877

### DIFF
--- a/app/services/credit_notes/validate_item_service.rb
+++ b/app/services/credit_notes/validate_item_service.rb
@@ -67,6 +67,7 @@ module CreditNotes
     # NOTE: Check if item amount is less than or equal to fee remaining creditable amount
     def valid_individual_amount?
       return true if item.amount_cents <= fee.creditable_amount_cents
+      return true if invoice.credit? && invoice.payment_pending?
 
       if invoice.credit? && item.amount_cents > invoice.associated_active_wallet&.balance_cents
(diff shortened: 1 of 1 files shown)


vankiru (approved):
Lgtm 👍

PR #4872, opened 2026-01-19
feat(fixed-charges): display billing period in pay-in-advance PDF

ancorcruz:
## Context

The fixed charge pay-in-advance invoice PDF template was displaying a static "Item" header instead of showing the billing period dates like the regular v4 invoice template does.

## Description

Update the `_fixed_charge.slim` template to use
`FeeBoundariesHelper.group_fees_by_billing_period()` for grouping fees by their billing period, consistent with the regular v4 invoice template. The header now shows "Fees from X to Y" instead of "Item", and fees are rendered using the `_fixed_charge_fee.slim` partial for consistency with how fixed charges appear in regular invoices.

Add snapshot tests for the fixed charge PDF invoice template covering:
- Single fixed charge fee with billing period header
- Multiple fixed charge fees in different billing periods (grouped)
- Multiple fixed charge fees in the same billing period
- Graduated charge model with breakdown
- Prorated fixed charges
- Zero amount fees
- Coupon discounts

### DIFF
--- a/spec/views/app/views/templates/invoices/v4/__snapshots__/templates_invoices_v4_fixed_charge.slim/with_a_single_fixed_charge_fee/renders_with_billing_period_header.html.snap
+++ b/spec/views/app/views/templates/invoices/v4/__snapshots__/templates_invoices_v4_fixed_charge.slim/with_a_single_fixed_charge_fee/renders_with_billing_period_header.html.snap
@@ -1,0 +1,160 @@
+<!DOCTYPE html>
+<html>
+  <body>
+    <div class="wrapper">
+      <div class="mb-24">
+        <h1 class="invoice-title">
+          Invoice
+        </h1>
+      </div>
+      <div class="mb-24 overflow-auto">
+        <div class="invoice-information-column">
+          <table class="invoice-information-table">
+            <tr>
+              <td class="body-1">
+                Invoice Number
+              </td>
+              <td class="body-2">
+                LAGO-202509-FC-001
+              </td>
+            </tr>
+            <tr>
+              <td class="body-1">
+                Issue Date
+              </td>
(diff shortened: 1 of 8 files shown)


app/views/templates/invoices/v4/_fixed_charge.slim
```diff
@@ -1,20 +1,21 @@
...
-    tr.fee
-      td.body-1
-        = fee.invoice_name + FeeDisplayHelper.grouped_by_display(fee)
-      td.body-2 = RoundingHelper.round_decimal_part(fee.units)
-      td.body-2 = FeeDisplayHelper.format_precise_unit_amount(fee)
-      td.body-2 == TaxHelper.applied_taxes(fee)
-      td.body-2 = FeeDisplayHelper.format_amount(fee)
+- invoice_subscription = invoice_subscriptions.first
+- fixed_charge_fees = fees.fixed_charge.order(:succeeded_at, :created_at)
+- grouped_fees = FeeBoundariesHelper.group_fees_by_billing_period(fixed_charge_fees, invoice_subscription:)
```
vincent-pochet, line 3: For fixed charges, is it possible to have different periods for invoice_subscriptions if multiple?
Like for charges with some billed monthly on a yearly plan and some billed quarterly? 
If so, would this approach work?

ancorcruz: @vincent-pochet currently we don't support that scenario of fixed charges with different billing periods but as agreed with @getlago/product this could be implemented soon as we are revamping plans and charges.

This implementation supports this future scenario by grouping fees by their date boundaries and rendering them correctly ordered as we do for the standard invoice subscription template.

So, we are future proof 😉

PR #4879, opened 2026-01-20
feat(multi-payments): Adyen changes

mariohd:
## Services
`PaymentProviders::Adyen::Payments::CreateService`
- added the methods to use the `payment_method.provider_method_id`

`Invoices::Payments::AdyenService` 
- removed left over code that was not used anymore and was replaced by the service above.

`PaymentProviderCustomers::AdyenService`
- `#preauthorise` creates the PaymentMethod based on `payment_method_id` supplied by the webhook call.

### DIFF
--- a/app/services/payment_provider_customers/adyen_service.rb
+++ b/app/services/payment_provider_customers/adyen_service.rb
@@ -71,12 +71,7 @@ module PaymentProviderCustomers
         adyen_customer.update!(payment_method_id:, provider_customer_id: shopper_reference)
 
         if organization.feature_flag_enabled?(:multiple_payment_methods)
-          PaymentMethods::FindOrCreateFromProviderService.call(
-            customer:,
-            payment_provider_customer: adyen_customer,
-            provider_method_id: payment_method_id,
-            set_as_default: true
-          )
+          handle_payment_methods(payment_method_id)
         end
 
         SendWebhookJob.perform_later("customer.payment_provider_created", customer)
@@ -156,6 +151,17 @@ module PaymentProviderCustomers
       return result if Customer.find_by(external_id: shopper_reference).nil?
 
       result.not_found_failure!(resource: "adyen_customer")
+    end
+
+    def handle_payment_methods(payment_method_id)
+      PaymentMethods::FindOrCreateFromProviderService.call(
+        customer:,
+        payment_provider_customer: adyen_customer,
+        provider_method_id: payment_method_id,
+        set_as_default: true
+      )
+    # race condition for multiple calls while creating the PM
+    rescue ActiveRecord::RecordNotUnique
     end
   end
 end
--- a/spec/services/payment_provider_customers/adyen_service_spec.rb
+++ b/spec/services/payment_provider_customers/adyen_service_spec.rb
@@ -253,6 +253,16 @@ RSpec.describe PaymentProviderCustomers::AdyenService do
 
             expect(existing_payment_method.reload.is_default).to be(true)
           end
+
+          context "when payment method lookup raises RecordNotUnique" do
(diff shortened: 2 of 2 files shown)


app/services/payment_provider_customers/stripe/update_payment_method_service.rb
```diff
@@ -18,25 +18,14 @@ def call
         stripe_customer.save!
 
         if stripe_customer.organization.feature_flag_enabled?(:multiple_payment_methods)
-          if payment_method_id
```
lovrocolic, line 21: I already updated this logic with new service - `PaymentMethods::FindOrCreateFromProviderService` in my stripe branch, so I guess we can skip it here

app/services/invoices/payments/adyen_service.rb
```diff
@@ -98,56 +98,6 @@ def adyen_payment_provider
         @adyen_payment_provider ||= payment_provider(customer)
       end
 
-      def update_payment_method_id
```
lovrocolic, line 101: How is this flow triggered? I am not sure to understand

lovrocolic (approved):

app/services/invoices/payments/adyen_service.rb
mariohd: this logic was removed in favor of `PaymentProviders::Adyen::Payments::CreateService`.

db/migrate/20260204130807_add_unique_index_to_payment_methods.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class AddUniqueIndexToPaymentMethods < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :payment_methods,
+      [:payment_provider_customer_id, :provider_method_id],
+      unique: true,
+      name: "index_payment_methods_on_provider_customer_and_provider_method",
+      algorithm: :concurrently
```
osmarluz, line 11: @mariohd since we have the discarded records and we currently don't discard the `PaymentProviderCustomer` objects (so the id is always the same for the same payment provider), this needs to be a partial index:

```suggestion
    add_index :payment_methods,
      [:payment_provider_customer_id, :provider_method_id],
      unique: true,
      where: "deleted_at IS NULL",
      name: "index_payment_methods_on_provider_customer_and_provider_method",
      algorithm: :concurrently
```

lovrocolic: @mariohd Can you please double check it? I am not sure to understand why we need new `where` clause

mariohd: @osmarluz idea is that when we delete a `PaymentProviderCustomer` and then create the same record, we would not be able to recreate the payment methods because the `PaymentMethod` table has this unique index, so he added the where clause. 

Being honest, IDK if this is a real world scenario that could eventually happen, but it's a valid point. 
About testing, I can perform the same tests I did for Adyen and see if the original solution still works after this gets into qa3 test branch.

lovrocolic: @osmarluz Please check with Mario if latest commit is needed since I am not sure TBH. Also, if yes, I am not sure why structure.sql is not updated or payment method model with index description.

This branch is already merged in QA one as it is, so not sure if we need to rollback migration there and run it again

@mariohd wdyt?

mariohd: Just tested this on QA3 and it works!

osmarluz (approved):
🚀

PR #4883, opened 2026-01-20
feat(invoices): display subscription name on all v4 PDF invoices

ancorcruz:
## Context

Previously, the subscription name was only displayed on PDF invoices when a customer had multiple subscriptions. The subscription name was visible in the UI but hidden from the PDF for single-subscription customers.

Also, the v4/charge.slim template is used for pay-in-advance charge invoices but lacked dedicated view tests to ensure rendering correctness.

## Description

Always display the subscription name header on v4 PDF invoices regardless of the number of subscriptions. This applies to regular subscription invoices, pay-in-advance charge invoices, and pay-in-advance fixed charge invoices. The page-break behavior for multi-subscription invoices is preserved.

Add comprehensive view tests for the v4/charge.slim invoice template following the pattern established by the v4/fixed_charge.slim tests. The tests use rspec-snapshot to validate HTML output across 9 scenarios: single charge, multiple charges, percentage charges (basic and detailed), prorated charges, non-invoiceable charges, filtered charges, coupons, and tax breakdowns.

### DIFF
--- a/spec/views/app/views/templates/invoices/v4/charge.slim_spec.rb
+++ b/spec/views/app/views/templates/invoices/v4/charge.slim_spec.rb
@@ -25,7 +25,15 @@ RSpec.describe "templates/invoices/v4/charge.slim" do
   end
 
   let(:subscription) do
-    create(:subscription, customer:, plan:, status: "active")
+    create(
+      :subscription,
+      customer:,
+      plan:,
+      status: "active",
+      started_at: Time.zone.parse("2025-09-01 00:00:00"),
+      subscription_at: Time.zone.parse("2025-09-01 00:00:00"),
+      billing_time: :calendar
+    )
   end
 
   let(:invoice) do
@@ -250,6 +258,17 @@ RSpec.describe "templates/invoices/v4/charge.slim" do
       create(:standard_charge, :pay_in_advance, plan:, billable_metric: recurring_billable_metric, prorated: true)
     end
 
+    let(:pay_in_advance_event) do
+      create(
+        :event,
+        organization:,
+        subscription:,
+        customer:,
+        code: recurring_billable_metric.code,
+        timestamp: Time.zone.parse("2025-09-15 10:00:00")
+      )
+    end
+
     let(:prorated_fee) do
       create(
         :charge_fee,
@@ -257,6 +276,7 @@ RSpec.describe "templates/invoices/v4/charge.slim" do
         charge: prorated_charge,
         subscription:,
         pay_in_advance: true,
+        pay_in_advance_event_id: pay_in_advance_event.id,
         amount_cents: 2500,
         amount_currency: "USD",
         units: 5,
--- a/spec/views/app/views/templates/invoices/v4/__snapshots__/templates_invoices_v4_charge.slim/with_prorated_charge/renders_correctly.html.snap
+++ b/spec/views/app/views/templates/invoices/v4/__snapshots__/templates_invoices_v4_charge.slim/with_prorated_charge/renders_correctly.html.snap
@@ -116,7 +116,7 @@
           <tr class="fee">
             <td>
               <div class="body-1">Prorated Seats</div>
-              <div class="body-3">Used  out of  days</div>
+              <div class="body-3">Used 16 out of 30 days</div>
             </td>
             <td class="body-2">5</td>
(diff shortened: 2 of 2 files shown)


spec/views/app/views/templates/invoices/v4/charge.slim_spec.rb
```diff
@@ -0,0 +1,515 @@
...
+          "charges_from_datetime" => "2025-09-01 00:00:00",
+          "charges_to_datetime" => "2025-09-30 23:59:59"
+        }
+      )
+    end
+
+    before { filtered_fee }
+
+    it "renders charge with filter" do
+      expect(rendered_template).to include("Filtered Charge")
```
groyoh: Are those `include` checks necessary since we have the snapshot ?

app/views/templates/invoices/v4/_subscription_details.slim
```diff
@@ -1,8 +1,7 @@
 - if subscription?
   - subscriptions.each do |subscription|
     - invoice_subscription = invoice_subscription(subscription.id)
-    - if subscriptions.count > 1
-      h2.title-2.mb-24 class="#{'invoice-details-title' if subscriptions.count > 1}" = I18n.t('invoice.details', resource: subscription.invoice_name)
+    h2.title-2.mb-24 class="#{'invoice-details-title' if subscriptions.count > 1}" = I18n.t('invoice.details', resource: subscription.invoice_name)
```
groyoh: I guess it's a bit out of scope, but the case where we have more than 1 subscription is not tested.

app/views/templates/invoices/v4/_fixed_charge.slim
```diff
@@ -2,6 +2,8 @@
 - fixed_charge_fees = fees.fixed_charge.order(:succeeded_at, :created_at)
 - grouped_fees = FeeBoundariesHelper.group_fees_by_billing_period(fixed_charge_fees, invoice_subscription:)
 
+h2.title-2.mb-24 = I18n.t('invoice.details', resource: invoice_subscription.subscription.invoice_name)
```
groyoh: It might be overkill, but should we maybe have a `_subscription_name` partial to avoid repetition (and potential mistake) across templates ? I had the issue this week where I updated the prepaid credit rendering for subscription but didn't realise the code was duplicated for pay-in-advance.

spec/views/app/views/templates/invoices/v4/__snapshots__/templates_invoices_v4_charge.slim/with_prorated_charge/renders_prorated_charge_with_proration_details.html.snap
```diff
@@ -0,0 +1,162 @@
...
+            <td class="body-2">Item</td>
+            <td class="body-2">Units</td>
+            <td class="body-2">Unit price</td>
+            <td class="body-2">Tax rate</td>
+            <td class="body-2">Amount</td>
+          </tr>
+          <tr class="fee">
+            <td>
+              <div class="body-1">Prorated Seats</div>
+              <div class="body-3">Used  out of  days</div>
```
groyoh: I believe we're missing some variable from the `breakdown_for_days` translations.

groyoh (approved):

spec/views/app/views/templates/invoices/v4/charge.slim_spec.rb
ancorcruz: It is included in the snapshoot indeed, but just for visibility. As you raised the point, I will remove them as this is not adding value but just making the tests slower.

Thanks.

app/views/templates/invoices/v4/_subscription_details.slim
ancorcruz: Thanks. I'll add them.

spec/views/app/views/templates/invoices/v4/__snapshots__/templates_invoices_v4_charge.slim/with_prorated_charge/renders_prorated_charge_with_proration_details.html.snap
ancorcruz: it was a test setup issue. fixed, thanks for spotting.

PR #4887, opened 2026-01-21
feat(invoices): order subscriptions alphabetically by invoice_name

ancorcruz:
When an invoice has multiple subscriptions, the order in which they appeared in the PDF, GraphQL, and REST API responses was undefined (database insertion order). This made it difficult to find specific subscriptions on invoices.

Add consistent alphabetical ordering by subscription invoice_name across all invoice rendering contexts. The ordering uses the same fallback chain as the invoice_name method: subscription.name, then plan.invoice_display_name, then plan.name.

Changes:
- Add order_by_subscription_invoice_name scope to InvoiceSubscription
- Add sorted_invoice_subscriptions and sorted_subscriptions to Invoice
- Apply ordering in V4 PDF template, GraphQL resolvers, and REST API
- Add tests for ordering in all three contexts


spec/views/app/views/templates/invoices/__snapshots__/templates_invoices_v4.slim/when_invoice_has_multiple_subscriptions/renders_correctly.html.snap
```diff
@@ -110,12 +110,12 @@
             <td class="body-2">Amount</td>
           </tr>
           <tr>
-            <td class="body-1">Basic Plan</td>
-            <td class="body-2">€30.00</td>
+            <td class="body-1">Zebra Plan</td>
```
groyoh, line 113: Is it expected that it's not in alphabetical order here ?

ancorcruz: yes, for some reason the snapshot was not okay, it was not even the right filename but tests still pass... I would take a look to the trustability of these view tests...

ancorcruz: sorted now. thanks for spotting this.

PR #4888, opened 2026-01-21
Feat(wallets): add code to wallets

annvelents:
Added codes to wallets.
Codes are uniq within customer.
Codes are formed from wallet.name, otherwise it's "default". If customer has more than one wallet with the same name (or without), timestamp is added

Updated services that create and update wallets (we need possibility to update code, so clients will be able to update auto-generated code to the one they need)
Note: on create we generate code:
- get it if provided, parameterize name if exist, otherwise "default"
- if code was generated (not provided) and is taken, we add a timestamp to the generated code
- if code was provided and is taken, we return an error

### DIFF
--- a/app/models/wallet.rb
+++ b/app/models/wallet.rb
@@ -107,7 +107,7 @@ class Wallet < ApplicationRecord
   end
 
   def unique_code_per_customer
-    if Wallet.where(customer_id: customer_id, code: code).where.not(id: id).exists?
+    if code && Wallet.where(customer_id: customer_id, code: code).where.not(id: id).exists?
       errors.add(:code, :taken)
     end
   end
--- a/app/graphql/types/customer_portal/wallets/object.rb
+++ b/app/graphql/types/customer_portal/wallets/object.rb
@@ -9,7 +9,7 @@ module Types
 
         field :id, ID, null: false
 
-        field :code, String, null: false
+        field :code, String, null: true
         field :currency, Types::CurrencyEnum, null: false
         field :expiration_at, GraphQL::Types::ISO8601DateTime, null: true
         field :name, String, null: true
--- a/app/graphql/types/wallets/object.rb
+++ b/app/graphql/types/wallets/object.rb
@@ -10,7 +10,7 @@ module Types
 
       field :customer, Types::Customers::Object
 
-      field :code, String, null: false
+      field :code, String, null: true
       field :currency, Types::CurrencyEnum, null: false
       field :name, String, null: true
       field :priority, Integer, null: false
--- a/schema.graphql
+++ b/schema.graphql
@@ -4402,7 +4402,7 @@ CustomerPortalWallet
 """
 type CustomerPortalWallet {
   balanceCents: BigInt!
-  code: String!
+  code: String
   consumedAmountCents: BigInt!
   consumedCredits: Float!
   creditsBalance: Float!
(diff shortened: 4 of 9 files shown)


app/models/wallet.rb
```diff
@@ -44,6 +44,7 @@ class Wallet < ApplicationRecord
   validates :paid_top_up_max_amount_cents, numericality: {greater_than: 0}, allow_nil: true
   validates :priority, inclusion: {in: 1..LOWEST_PRIORITY}
   validate :paid_top_up_max_greater_than_or_equal_min
+  validate :unique_code_per_customer
```
julienbourdeau: I think this will trigger a new db request everytime you update wallet. so no updating balance will add an extra query 😄 

Personally, I'm in favor of not using model validation for this and rely on the DB index and catch the `AR::RecordNotUnique` in the services.

If we want to use the model validation, at least add a condition to only trigger if code was changed.

app/services/wallets/create_service.rb
```diff
@@ -19,10 +19,16 @@ def call
 
       return result unless valid?
 
+      code = params[:code] || params[:name].to_s.parameterize(separator: "_").presence || "default"
+      existing_wallet = Wallet.where(organization_id:, customer_id: customer.id, code: code).exists?
+      # if code is provided but is already taken, we won't modify it, just raise validation error later
+      code = "#{code}_#{Time.current.to_i}" if params[:code].nil? && existing_wallet
```
julienbourdeau: I'm thinking this could be made clearer, maybe something like this:


```suggestion
      code = params[:code]
      
      if code.nil?
        code =  params[:name]&.to_s&.parameterize(separator: "_").presence || "default"
        code_taken = Wallet.where(organization_id:, customer_id: customer.id, code: code).exists?
        code += "_#{Time.current.to_i}" if code_taken
      end
```

db/migrate/20260121111431_add_indexes_to_wallets_code.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddIndexesToWalletsCode < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def up
+    safety_assured do
+      add_index :wallets, [:customer_id, :code], unique: true, algorithm: :concurrently
```
julienbourdeau: I like naming the index but not mandatory

db/migrate/20260120200341_fill_wallet_codes.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+class FillWalletCodes < ActiveRecord::Migration[8.0]
```
julienbourdeau: I'm not sure if I prefer an sql version or a job 🤔 More opinions welcome

app/models/wallet.rb
annvelents: you are right... I'd still like to have the validation as it makes model more explicit about the limits it has... yes, I'll update it to run the validation only when code is changed

db/migrate/20260120200341_fill_wallet_codes.rb
annvelents: my thoughts were: 
1) we do not have millions of wallets, so it will not be running for too long
2) if migration fails, running it again will finish the update

but I see we were already filling organization_id for wallets with job, so maybe it worth continuing it here as well...
will rewrite it now 😁

PR #4894, opened 2026-01-22
feat(offset-invoice-on-pay-in-advance-termination) Add offset option for pay-in-advance subscription termination

vankiru:
## Context

This is the second step toward supporting invoice offset when terminating a pay-in-advance subscription.

## Description

It adds the logic for the `offset` option when terminating a pay-in-advance subscription. When `offset` is selected, the unpaid unused amount is applied to the invoice, and the paid used amount is refunded.

### DIFF
--- a/app/services/credit_notes/create_from_termination.rb
+++ b/app/services/credit_notes/create_from_termination.rb
@@ -4,6 +4,10 @@ module CreditNotes
   class CreateFromTermination < BaseService
     Result = CreditNotes::CreateService::Result
 
+    # on_termination controls what to do with the unused subscription amount:
+    #   :credit - credits all unused amount back to the customer
+    #   :refund - refunds the unused paid amount, credits any updaid unused amount back to the customer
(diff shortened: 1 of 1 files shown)


app/services/credit_notes/create_from_termination.rb
```diff
@@ -4,31 +4,33 @@ module CreditNotes
   class CreateFromTermination < BaseService
     Result = CreditNotes::CreateService::Result
 
-    def initialize(subscription:, reason: "order_change", upgrade: false, context: nil, refund: false)
+    def initialize(subscription:, reason: "order_change", upgrade: false, context: nil, on_termination: :credit)
```
vincent-pochet, line 11: Could you maybe document the possible values for the new attribute?

vincent-pochet (approved):
Looks good 👍

app/services/credit_notes/create_from_termination.rb
vankiru: Yep, sure!

PR #4861, opened 2026-01-15
feat(multiple-payment-methods): Adapt stripe intent with correct payment method

lovrocolic:
## Context

With this feature, it will be possible to attach multiple payment methods to single customer.

## Description

This PR adapts stripe flow, where payment intent is created with the new logic that consumes payment method resource.

PR #4896, opened 2026-01-22
feat(multi-payment-methods): add FindOrCreateFromProviderService

aquinofb:
## Context
A new `PaymentMethods::FindOrCreateFromProviderService` to find existing or create new payment methods from provider data, to be (re)used by customer providers such as Stripe, Adyen, etc...

### DIFF
--- a/app/services/payment_provider_customers/stripe/update_payment_method_service.rb
+++ b/app/services/payment_provider_customers/stripe/update_payment_method_service.rb
@@ -18,14 +18,25 @@ module PaymentProviderCustomers
         stripe_customer.save!
 
         if stripe_customer.organization.feature_flag_enabled?(:multiple_payment_methods)
-          find_or_create_result = PaymentMethods::FindOrCreateFromProviderService.call(
-            customer:,
-            payment_provider_customer: stripe_customer,
-            provider_method_id: payment_method_id,
-            params: {provider_payment_methods: stripe_customer.provider_payment_methods},
-            set_as_default: true
-          )
-          result.payment_method = find_or_create_result.payment_method
+          if payment_method_id
+            payment_method = PaymentMethod.find_by(
+              customer:,
+              payment_provider_customer: stripe_customer,
+              provider_method_id: payment_method_id
+            )
+
+            payment_method ||= PaymentMethods::CreateFromProviderService.call(
+              customer:,
+              params: {provider_payment_methods: stripe_customer.provider_payment_methods},
+              provider_method_id: payment_method_id,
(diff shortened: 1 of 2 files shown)


app/services/payment_methods/find_or_create_from_provider_service.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    def create_from_provider
+      CreateFromProviderService.call(
+        customer:,
+        params:,
+        provider_method_id:,
+        payment_provider_id: payment_provider_customer.payment_provider_id,
+        payment_provider_customer:
+      ).payment_method
```
aquinofb, line 48: Hey @mariohd, @lovrocolic, I understand this service will soon require a new arg for `details`. Since it's a simple change and the other pull request is still a _draft_, what do you all think about we going ahead with this service as-is, and include the details here as part of the other pull request (#4861)?

Alternatively, which would also not block this PR, we could pass the `details` within the existing `params`, to the `PaymentMethods::CreateFromProviderService` would access it as `params[:details]`. Does that make sense?

mariohd: If i'm understanding right, `params` will be available to receive any extra attribute, `details` being one of them. 

so for @lovrocolic's PR, he can pass the `details` something like this 

```
CreateFromProviderService.call(
  customer:,
  params:,
  provider_method_id:,
  payment_provider_id: payment_provider_customer.payment_provider_id,
  payment_provider_customer:,
  details: params[:details]
).payment_method
```

is this right @aquinofb ?

aquinofb: Yes, that's correct. Or, as the alternative I mentioned, leave `PaymentMethods::CreateFromProviderService` as-is, and change it to access the details via `params`.

The caller (e.g., PSP Customer) would do something like following:
```ruby
# payment_provider_customers/{any}.rb
PaymentMethods::FindOrCreateFromProviderService.call(
  customer:,
  payment_provider_customer: provider_customer,
  provider_method_id: payment_method_id,
  set_as_default: true,
  params: {
    details: # Any extra `details` metadata
  }
)
```

Then, on `PaymentMethods::FindOrCreateFromProviderService`, there will be no changes to how it calls the `PaymentMethods::CreateFromProviderService`. Finally, on the `PaymentMethods::CreateFromProviderService`, we could just access the details via `params[:details]`, instead of receiving the details as its own argument.

WDYT?

lovrocolic: All good

lovrocolic (approved):

PR #4897, opened 2026-01-22
feat: support payment_failed status on offset with credit note

brunomiguelpinto:
## Context

Credit notes for credit invoices previously only supported `payment_pending` status for offsetting. This change extends that support to include `payment_failed` status, allowing customers to cancel prepaid credits even when the payment has failed.

## Description

- Extend [CreditNotes::CreateService](cci:2://file:///Users/brunomiguelpinto/Desktop/getlago/lago/api/app/services/credit_notes/create_service.rb:3:2-303:5) to allow credit note creation for credit invoices with `payment_failed` status (offset only, no credit/refund)
- Update [CreditNotes::ValidateItemService](cci:2://file:///Users/brunomiguelpinto/Desktop/getlago/lago/api/app/services/credit_notes/validate_item_service.rb:3:2-84:5) to validate items for `payment_failed` credit invoices
- Update `Fee#offsettable_amount_cents` to return full amount for `payment_failed` credit invoices
- Update `Invoice#offsettable_amount_cents` to return full amount for `payment_failed` credit invoices

### DIFF
--- a/spec/models/invoice_spec.rb
+++ b/spec/models/invoice_spec.rb
@@ -1988,8 +1988,8 @@ RSpec.describe Invoice do
       it "returns 0 when failed but fully offsetted by credit note" do
         invoice = create(:invoice, invoice_type: :credit, payment_status: :failed,
           total_amount_cents: 1000, total_paid_amount_cents: 0)
-        allow(invoice).to receive(:total_due_amount_cents).and_return(0) # simulate credit note with offset_amount_cents
-        expect(invoice.offsettable_amount_cents).to eq(0) # min(0, 800) since due is 0
+        create(:credit_note, invoice:, status: :finalized, offset_amount_cents: 1000)
+        expect(invoice.offsettable_amount_cents).to eq(0)
       end
     end
 


spec/models/invoice_spec.rb
```diff
@@ -1973,12 +1973,24 @@
...
         invoice = create(:invoice, invoice_type: :credit, payment_status: :succeeded,
           total_amount_cents: 1000, total_paid_amount_cents: 1000)
         allow(invoice).to receive(:creditable_amount_cents).and_return(800)
         expect(invoice.offsettable_amount_cents).to eq(0) # total_due is 0
       end
+
+      it "returns 0 when failed but fully offsetted by credit note" do
+        invoice = create(:invoice, invoice_type: :credit, payment_status: :failed,
+          total_amount_cents: 1000, total_paid_amount_cents: 0)
+        allow(invoice).to receive(:total_due_amount_cents).and_return(0) # simulate credit note with offset_amount_cents
```
groyoh: Do we need to mock here ? Can't we set the state properly ?

app/models/invoice.rb
```diff
@@ -330,12 +330,16 @@ def creditable_amount_cents
...
+  # amount cents onto which we can issue a credit note as offset
+  # when invoice type is credit theres no partial payments/refund/offset only full amount
   def offsettable_amount_cents
-    if credit? && payment_pending?
-      return total_amount_cents
-    end
+    due_amount_cents = total_due_amount_cents
+
+    return total_amount_cents if credit? &&
+      due_amount_cents.positive? &&
```
groyoh, line 339: Is there a test for this ?

spec/models/invoice_spec.rb
brunomiguelpinto: yup i will change it

app/models/invoice.rb
brunomiguelpinto: api/spec/models/invoice_spec.rb:1988 yes this case  due_amount_cents is 0 so we dont return full amount

PR #4901, opened 2026-01-23
feat(multi-payments): PaymentProviders::Gocardless changes

mariohd:
## Services
`PaymentProviders::Gocardless::Payments::CreateService`
- Add the method `set_provider_method_id` which sets the `GoCardlessCustomer#provider_mandate_id` and set  as default new `PaymentMethod` with the `mandate_id` returned by their API.

### DIFF
--- a/app/services/payment_provider_customers/gocardless_service.rb
+++ b/app/services/payment_provider_customers/gocardless_service.rb
@@ -106,7 +106,10 @@ module PaymentProviderCustomers
       client.billing_requests.create(
         params: {
           mandate_request: {
-            scheme: "bacs"
+            scheme: "bacs",
+            metadata: {
+              provider_customer_id: gocardless_customer_id
+            }
           },
           links: {
             customer: gocardless_customer_id
--- a/app/services/payment_providers/gocardless/handle_event_service.rb
+++ b/app/services/payment_providers/gocardless/handle_event_service.rb
@@ -5,6 +5,7 @@ module PaymentProviders
     class HandleEventService < BaseService
       PAYMENT_ACTIONS = %w[paid_out failed cancelled customer_approval_denied charged_back].freeze
       REFUND_ACTIONS = %w[created funds_returned paid refund_settled failed].freeze
+      MANDATES_ACTIONS = %w[created cancelled active].freeze
 
       PAYMENT_SERVICE_CLASS_MAP = {
         "Invoice" => Invoices::Payments::GocardlessService,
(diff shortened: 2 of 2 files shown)


app/services/payment_providers/gocardless/payments/create_service.rb
lovrocolic: Hmm it seems to me that gocardless experience would not work. We would always fetch latest mandate and therefore we would not be able to have multiple payment methods.

What was your idea, how to have multiple payment methods for gocardless?

app/services/payment_providers/gocardless/payments/create_service.rb
```diff
@@ -100,6 +98,21 @@ def prepare_failed_result(error, reraise: false)
...
+        def set_provider_method_id(provider_method_id:)
+          provider_customer.provider_mandate_id = provider_method_id
+          provider_customer.save!
+
+          if customer.organization.feature_flag_enabled?(:multiple_payment_methods)
+            PaymentMethods::FindOrCreateFromProviderService.call(
+              customer:,
+              payment_provider_customer: provider_customer,
+              provider_method_id:,
+              params: {provider_payment_methods: "mandate"},
```
lovrocolic, line 111: I haven't seen mandate as payment method in `PaymentProviderCustomers::GocardlessCustomer`

app/services/payment_providers/gocardless/payments/create_service.rb
```diff
@@ -68,9 +68,7 @@ def mandate_id
 
           raise MandateNotFoundError unless mandate
 
-          provider_customer.provider_mandate_id = mandate.id
-          provider_customer.save!
-
+          set_provider_method_id(provider_method_id: mandate.id)
           mandate.id
```
lovrocolic, line 72: Should we return here: `payment&.payment_method&.provider_method_id`?

lovrocolic: I have a feeling that gocardless experience does not work. In `Invoices::Payments::CreateService` we would not even trigger a payment if PM does not exist. But for Gocardless PM is created one step later, with sending payment request 🤔

Also, if we always pull latest mandate, we would not take into consideration selected PM (e.g. on subscription), so I guess we should include somehow `payment&.payment_method&.provider_method_id `

Do you have any idea?

app/services/payment_providers/gocardless/payments/create_service.rb
lovrocolic: Should we use `card`?

lovrocolic:
@mariohd Should we close this PR? Gocardless changes are already in main and if I am not mistaken we could close this one...

PR #4903, opened 2026-01-23
feat(multiple-payment-methods): fetch default payment method when existing provider customer connects to Lago

lovrocolic:
## Context

With this feature, it will be possible to attach multiple payment methods to single customer.

## Description

This PR covers flow when existing stripe customer is connected to Lago. In that case, we want to import default PM for that customer so that we don't end up with multiple payment failures.

new logic is hidden behind feature flag and performed in async job

### DIFF
--- a/app/services/payment_providers/stripe/customers/fetch_default_payment_method_service.rb
+++ b/app/services/payment_providers/stripe/customers/fetch_default_payment_method_service.rb
@@ -47,7 +47,7 @@ module PaymentProviders
 
           details = {type: pm.type}
 
-          if pm&.card
+          if pm.card
             details[:last4] = pm.card.last4
             details[:brand] = pm.card.display_brand
             details[:expiration_month] = pm.card.exp_month


app/services/payment_providers/stripe/customers/fetch_default_payment_method_service.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        def payment_method_details(payment_method_id:)
+          pm = ::Stripe::PaymentMethod.retrieve(
+            payment_method_id,
+            {api_key: provider_customer.payment_provider.secret_key}
+          )
+
+          details = {type: pm.type}
+
+          if pm&.card
```
aquinofb: Do we need the safe navigation (`&`) here? If the `pm` is `nil`, it will raise a NoMethodError on line 48.

lovrocolic: Also I have guard earlier in the code that ensure pm existance 👍 

Removed `&`

Thanks

PR #4917, opened 2026-01-27
fix(multi-payment-methods): Discard provider customer and associated payment methods when payment provider is removed/replaced

osmarluz:
## Context

A Lago customer can be associated with a customer in a payment provider (e.g., Stripe). In Lago, if a customer is disconnected from a payment provider (removed when it's performed through the UI and replaced when it's performed through the API), we want to discard the old provider customer and all payment methods associated with it.

## Description

- On the flow done through the UI (handled through the `app/services/customers/update_service.rb`) we handle the discarding of the old provider customer and it's associated payment methods whenever the payment provider is **removed** for the customer.
- On the flow done through the API (handled through the `app/services/customers/upsert_from_api_service.rb`) we handle the discarding of the old provider customer and it's associated payment methods whenever the payment provider is **replaced** for the customer. It's not possible to remove a payment provider through the API.

### DIFF
--- a/spec/services/customers/update_service_spec.rb
+++ b/spec/services/customers/update_service_spec.rb
@@ -408,7 +408,7 @@ RSpec.describe Customers::UpdateService do
     end
 
     context "when removing payment provider" do
-      let!(:customer) do
+      let(:customer) do
         create(
           :customer,
           organization:,
@@ -417,8 +417,8 @@ RSpec.describe Customers::UpdateService do
           payment_provider_code:
         )
       end
-      let!(:stripe_customer) { create(:stripe_customer, customer:) }
-      let!(:payment_method) { create(:payment_method, customer:, payment_provider_customer: stripe_customer) }
+      let(:stripe_customer) { create(:stripe_customer, customer:) }
+      let(:payment_method) { create(:payment_method, customer:, payment_provider_customer: stripe_customer) }
 
       let(:update_args) do
         {
@@ -430,22 +430,21 @@ RSpec.describe Customers::UpdateService do
       end
 
       before do
+        payment_method
         create(:stripe_provider, organization:, code: payment_provider_code)
       end
 
       it "discards the provider customer and its payment methods" do
         result = customers_service.call
 
-        aggregate_failures do
-          expect(result).to be_success
+        expect(result).to be_success
 
-          customer = result.customer
-          expect(customer.payment_provider).to be_nil
-          expect(customer.payment_provider_code).to be_nil
+        customer = result.customer
+        expect(customer.payment_provider).to be_nil
+        expect(customer.payment_provider_code).to be_nil
 
-          expect(stripe_customer.reload).to be_discarded
-          expect(payment_method.reload).to be_discarded
-        end
+        expect(stripe_customer.reload).to be_discarded
+        expect(payment_method.reload).to be_discarded
       end
     end
 
--- a/spec/services/customers/upsert_from_api_service_spec.rb
+++ b/spec/services/customers/upsert_from_api_service_spec.rb
@@ -921,7 +921,7 @@ RSpec.describe Customers::UpsertFromApiService do
           )
         end
         let(:stripe_customer) { create(:stripe_customer, customer:, payment_provider: stripe_provider) }
-        let!(:payment_method) { create(:payment_method, customer:, payment_provider_customer: stripe_customer) }
+        let(:payment_method) { create(:payment_method, customer:, payment_provider_customer: stripe_customer) }
 
         let(:create_args) do
           {
@@ -935,6 +935,7 @@ RSpec.describe Customers::UpsertFromApiService do
         end
 
         before do
+          payment_method
           create(:gocardless_provider, organization:, code: "gocardless_1")
         end
 


app/services/customers/upsert_from_api_service.rb
```diff
@@ -243,12 +246,17 @@ def handle_api_billing_configuration(customer, new_customer)
       return if customer.payment_provider.nil?
 
       update_provider_customer = (billing || {})[:provider_customer_id].present?
-      update_provider_customer ||= customer.provider_customer&.provider_customer_id.present?
+      update_provider_customer ||= customer.provider_customer&.provider_customer_id.present? || old_provider_customer.present?
```
osmarluz, line 249: We need the `|| old_provider_customer.present?` check here to make sure the flow proceeds when we switch from one payment provider to another (for which the `provider_customer` does not exist yet) and do not pass the `provider_customer_id` (relying on the `sync_with_provider` flag to populate this)

app/services/customers/update_service.rb
```diff
@@ -124,8 +124,12 @@ def call
       end
 
       ActiveRecord::Base.transaction do
-        if old_provider_customer && args[:payment_provider].nil? && args[:payment_provider_code].present?
+        if old_provider_customer &&
+            args.key?(:provider_customer) && args[:provider_customer].nil? &&
+            args.key?(:payment_provider) && args[:payment_provider].nil?
+          old_provider_customer.payment_methods.find_each(&:discard!)
```
lovrocolic: I would rather call here `PaymentMethods::DestroyService`

That way we would keep centralised logic for destroying PM and we can add extra stuff inside without changing all places

app/services/customers/upsert_from_api_service.rb
```diff
@@ -243,12 +246,17 @@ def handle_api_billing_configuration(customer, new_customer)
...
 
       return unless update_provider_customer
 
       create_or_update_provider_customer(customer, billing)
 
+      if old_provider_customer && old_payment_provider != billing[:payment_provider]
+        old_provider_customer.payment_methods.find_each(&:discard!)
```
lovrocolic: same comment about destroy service

app/services/customers/update_service.rb
osmarluz: @lovrocolic done!

app/services/customers/upsert_from_api_service.rb
osmarluz: @lovrocolic done!

app/services/customers/upsert_from_api_service.rb
```diff
@@ -243,12 +247,24 @@ def handle_api_billing_configuration(customer, new_customer)
...
 
       create_or_update_provider_customer(customer, billing)
 
+      if old_provider_customer
+        provider_changed = old_payment_provider != billing[:payment_provider]
+        provider_code_changed = old_payment_provider_code != billing[:payment_provider_code]
+
+        if provider_changed || provider_code_changed
+          old_provider_customer.payment_methods.find_each do |payment_method|
+            PaymentMethods::DestroyService.call(payment_method: payment_method)
```
lovrocolic: Usually we would prefer following pattern: `PaymentMethods::DestroyService.call(payment_method:)`

osmarluz: @lovrocolic ah, yes, of course! I'll change it, thanks!

spec/services/customers/update_service_spec.rb
```diff
@@ -407,6 +407,48 @@
...
+        create(
+          :customer,
+          organization:,
+          billing_entity:,
+          payment_provider: "stripe",
+          payment_provider_code:
+        )
+      end
+      let!(:stripe_customer) { create(:stripe_customer, customer:) }
+      let!(:payment_method) { create(:payment_method, customer:, payment_provider_customer: stripe_customer) }
```
ancorcruz: we use `let` and `before` block to create the objects instead of `let!`

spec/services/customers/update_service_spec.rb
```diff
@@ -407,6 +407,48 @@
...
+        aggregate_failures do
+          expect(result).to be_success
+
+          customer = result.customer
+          expect(customer.payment_provider).to be_nil
+          expect(customer.payment_provider_code).to be_nil
+
+          expect(stripe_customer.reload).to be_discarded
+          expect(payment_method.reload).to be_discarded
+        end
```
ancorcruz: failures are aggregated by default, remove `aggregate_failures` block please.

spec/services/customers/upsert_from_api_service_spec.rb
```diff
@@ -907,6 +907,46 @@
...
+            :customer,
+            organization:,
+            external_id:,
+            payment_provider: "stripe",
+            payment_provider_code: "stripe_1"
+          )
+        end
+        let(:stripe_customer) { create(:stripe_customer, customer:, payment_provider: stripe_provider) }
+        let!(:payment_method) { create(:payment_method, customer:, payment_provider_customer: stripe_customer) }
+
```
ancorcruz, line 925: same here.

ancorcruz (approved):
Looks great @osmarluz , just some style adjustment in the tests.

spec/services/customers/update_service_spec.rb
osmarluz: @ancorcruz got it, I'll update it here!

PR #4918, opened 2026-01-27
feat(pre-enrich): Create enriched_events from events

vincent-pochet:
## Description

This PR follows https://github.com/getlago/lago-api/pull/4916.
Its goal is to create `enriched_events` records from an event in the `Events::PostProcessService`.
A new service names `Events::ReEnrichAllService` is also added to allow re-enrichment of all the events received for a subscription on the current (or past) billing period.

### DIFF
--- a/app/services/events/enrich_service.rb
+++ b/app/services/events/enrich_service.rb
@@ -48,10 +48,13 @@ module Events
       enriched_event.subscription_id = subscription.id
       enriched_event.plan_id = subscription.plan_id
 
-      enriched_event.properties = event.properties
       enriched_event.enriched_at = Time.current
       enriched_event.value = (event.properties || {})[billable_metric.field_name] || 0
       enriched_event.value = 1 if billable_metric.count_agg?
+
+      # NOTE: We might not be able to parse the value as a decimal, it will then fall back to 0
+      #       The behavior is aligned with the Clickhouse implementation but differs
+      #       a bit from the current PG one where we explicitly filter events with invalid values
       enriched_event.decimal_value = decimal_value(enriched_event.value)
       enriched_event
     end
--- a/app/services/events/re_enrich_all_service.rb
+++ b/app/services/events/re_enrich_all_service.rb
@@ -12,6 +12,8 @@ module Events
     end
 
     def call
+      return result if organization.clickhouse_events_store?
+
       events.find_in_batches do |events|
         EnrichedEvent.transaction do
           drop_enriched_events(events)
@@ -31,6 +33,8 @@ module Events
     private
 
     attr_reader :subscription, :timestamp
+
+    delegate :organization, to: :subscription
 
     def boundaries
       return @boundaries if @boundaries
@@ -59,11 +63,7 @@ module Events
     end
 
     def drop_enriched_events(events)
-      EnrichedEvent
-        .where(organization_id: subscription.organization_id, subscription_id: subscription.id)
-        .where(timestamp: boundaries.charges_from_datetime...boundaries.charges_to_datetime)
-        .where(event_id: events.map(&:id))
-        .delete_all
+      EnrichedEvent.where(event_id: events.map(&:id)).delete_all
     end
 
     def enriched_events(events)
--- a/spec/services/events/enrich_service_spec.rb
+++ b/spec/services/events/enrich_service_spec.rb
@@ -49,7 +49,6 @@ RSpec.describe Events::EnrichService do
         event_id: event.id,
         external_subscription_id: subscription.external_id,
         plan_id: plan.id,
-        properties: event.properties,
         grouped_by: {}
       )
     end
@@ -145,10 +144,7 @@ RSpec.describe Events::EnrichService do
         expect(result.enriched_events.count).to eq(1)
 
         enriched_event = result.enriched_events.first
-        expect(enriched_event).to have_attributes(
-          value: "foo_bar",
-          decimal_value: 0.0
-        )
+        expect(enriched_event.value).to eq("foo_bar")
       end
     end
 
@@ -236,8 +232,9 @@ RSpec.describe Events::EnrichService do
         expect(enriched_event).to have_attributes(
           charge_id: charge.id,
           charge_filter_id: charge_filter.id,
-          properties: event.properties,
-          grouped_by: {}
+          grouped_by: {},
+          value: "12",
+          decimal_value: 12.0
         )
       end
 
@@ -269,9 +266,7 @@ RSpec.describe Events::EnrichService do
           expect(enriched_event).to have_attributes(
             charge_id: charge.id,
             charge_filter_id: charge_filter.id,
-            properties: {
-              billable_metric.field_name => 12,
-              "region" => "eu",
+            grouped_by: {
               "cloud" => "aws",
               "provider" => "visa"
             }
--- a/spec/services/events/re_enrich_all_service_spec.rb
+++ b/spec/services/events/re_enrich_all_service_spec.rb
@@ -45,6 +45,27 @@ RSpec.describe Events::ReEnrichAllService do
 
         expect { result = re_enrich_service.call }.to change(EnrichedEvent, :count).by(-2)
         expect(result).to be_success
+
+        # Make sure the enriched event matched the event
+        enriched_event = EnrichedEvent.find_by(event_id: event.id)
+        expect(enriched_event).to have_attributes(
+          charge_id: charge.id,
+          charge_filter_id: nil,
+          grouped_by: {},
+          value: "12",
+          decimal_value: 12.0
+        )
+      end
+    end
+
+    context "with clickhouse events store" do
+      let(:organization) { create(:organization, clickhouse_events_store: true) }
+
+      it "returns success" do
+        result = nil
+
+        expect { result = re_enrich_service.call }.not_to change(EnrichedEvent, :count)
+        expect(result).to be_success
       end
     end
   end


app/services/events/re_enrich_all_service.rb
```diff
@@ -0,0 +1,91 @@
...
+        charges_duration: date_service.charges_duration_in_days,
+        timestamp:
+      )
+    end
+
+    def drop_enriched_events
+      EnrichedEvent
+        .where(organization_id: subscription.organization_id, subscription_id: subscription.id)
+        .where(timestamp: boundaries.charges_from_datetime...boundaries.charges_to_datetime)
+        .delete_all
```
groyoh: To avoid killing the DB:

```suggestion
        .in_batches.delete_all
```

app/services/events/re_enrich_all_service.rb
```diff
@@ -0,0 +1,91 @@
...
+    def initialize(subscription:, timestamp: Time.current)
+      @subscription = subscription
+      @timestamp = timestamp
+
+      super
+    end
+
+    def call
+      EnrichedEvent.transaction do
+        drop_enriched_events
```
groyoh: This might be a lot of writes happening in a single transaction. I wonder if we shouldn't commit in batches:

```
events.find_in_batches do |events|
  transaction do
    delete_enriched_events(events)
    enriched_events(events)
  end
end
```

groyoh (commented):
I'm not able to review everything right now so I'm submitted my partial feedback.

app/services/events/re_enrich_all_service.rb
vincent-pochet: You are right! :+1:

app/services/events/enrich_service.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      enriched_event.external_subscription_id = subscription.external_id
+      enriched_event.subscription_id = subscription.id
+      enriched_event.plan_id = subscription.plan_id
+
+      enriched_event.properties = event.properties
+      enriched_event.enriched_at = Time.current
+      enriched_event.value = (event.properties || {})[billable_metric.field_name] || 0
+      enriched_event.value = 1 if billable_metric.count_agg?
+      enriched_event.decimal_value = decimal_value(enriched_event.value)
```
groyoh, line 58: Should we skip the event if we cannot parse ? This could otherwise create issues with `max`, `unique_count`, `latest_value`, etc.

app/services/events/enrich_service.rb
```diff
@@ -0,0 +1,71 @@
...
+    def init_enriched_event
+      enriched_event = EnrichedEvent.new
+      enriched_event.event_id = event.id
+      enriched_event.organization_id = event.organization_id
+      enriched_event.code = event.code
+      enriched_event.transaction_id = event.transaction_id
+      enriched_event.timestamp = event.timestamp
+
+      enriched_event.external_subscription_id = subscription.external_id
+      enriched_event.subscription_id = subscription.id
```
groyoh, line 48: Do we need both ?

app/services/events/post_process_service.rb
```diff
@@ -50,26 +51,32 @@ def subscriptions
...
+    def expire_cached_charges
+      return if active_subscriptions.blank?
       return unless billable_metric
 
-      charges = billable_metric.charges
-        .joins(:plan)
-        .where(plans: {id: active_subscription.map(&:plan_id)})
-        .includes(filters: {values: :billable_metric_filter})
+      charges_and_filters.each do |charge, filter|
+        active_subscriptions.each do |subscription|
```
groyoh: I believe I might be missing something here. When does `active_subscriptions` actually have more than 1 sub ? Don't we have a unique active one per org/external_id tuple ?

If so, then we should update the code to make it more explicit:

```ruby
    def active_subscription
      @active_subscription ||= begin
        subs = subscriptions.select(&:active?)
        raise "should not happen" if sub.length > 1
        subs.first
        end
    end
```

WDYT ?

app/services/events/re_enrich_all_service.rb
```diff
@@ -0,0 +1,94 @@
...
+    def drop_enriched_events(events)
+      EnrichedEvent
+        .where(organization_id: subscription.organization_id, subscription_id: subscription.id)
+        .where(timestamp: boundaries.charges_from_datetime...boundaries.charges_to_datetime)
+        .where(event_id: events.map(&:id))
+        .delete_all
+    end
+
+    def enriched_events(events)
+      events.map do |event|
```
groyoh: Instead of flattening above:
```suggestion
      events.flat_map do |event|
```

app/services/events/enrich_service.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+module Events
+  class EnrichService < BaseService
+    Result = BaseResult[:enriched_events]
+
+    def initialize(event:, subscription:, billable_metric:, charges_and_filters:, persist: true)
```
groyoh, line 7: Is there a real use case where we'd have multiple charges matching the event for a single subscription ? If so, could you add a scenario test to reproduce this ? I'm confused by this use case.

groyoh (commented):
I did a first pass but the behaviour is a bit confusing for now. I'll do another review once I get the answers.

app/services/events/enrich_service.rb
vincent-pochet: Unique count will use the `value` field, as it is intended for unique distinct "identifier" values and as such are strings. For the other it will fallback to 0 and so will have no effect on the aggregation result, except for the `latest`. But in this case, it's the responsibility of the user to send a valid value.
Note that this behavior is the one applied in the clickhouse pipeline.

app/services/events/enrich_service.rb
vincent-pochet: A billable metric can be used on different charges in the same plan, and as such the event will match multiple charges. It's not common, but it's possible. I will add a test case for that

app/services/events/enrich_service.rb
groyoh: > Note that this behavior is the one applied in the clickhouse pipeline.

Isn't this behaviour slightly different from the PG one though ? I believe in PG, we skip events for which we cannot parse the property as decimal (using regexp).

app/services/events/enrich_service.rb
groyoh: Makes sense!

app/services/events/enrich_service.rb
vincent-pochet: Not necessarily for now. But since the cost to add a new column on this table will be very high, I prefer to have both.
We will still be able to remove an unused one in the future if we find out it's not usefull

app/services/events/enrich_service.rb
vincent-pochet: Yes we do for now, to avoid failure when performing the aggregation.
But I definitely think we should tend to align the behavior of both stores. And I prefer not to add a condition on "should I still persist this event since it's for a unique_count agg?". 

Also, keep in mind two aspects of this feature:
- At first, this new table will only be  used `optionally`, to "pre-flag" the charges and filters to aggregate on a given period, the aggregation itself will still be performed on the `events` table.
- On the long run, since the pre-enrichement is a breaking change in itself, due to the implication it has with the plan update, we can definitely use it has an opportunity to improve the current approach and reduce its complexity

Happy to discuss about it if you want to

groyoh: Thanks for the explanation. I'm fine with that. I just wanted to ensure we're aware of the behaviour here.

Maybe you can add a small comment to say it differentiate from the PG implementation ?

Also I'm wondering if we should maybe reject the event request if we cannot parse as decimal 😅

app/services/events/re_enrich_all_service.rb
```diff
@@ -0,0 +1,94 @@
...
+    def events
+      Event.where(organization_id: subscription.organization_id)
+        .where(external_subscription_id: subscription.external_id)
+        .where(timestamp: boundaries.charges_from_datetime...boundaries.charges_to_datetime)
+    end
+
+    def drop_enriched_events(events)
+      EnrichedEvent
+        .where(organization_id: subscription.organization_id, subscription_id: subscription.id)
+        .where(timestamp: boundaries.charges_from_datetime...boundaries.charges_to_datetime)
```
groyoh: I believe `event_id` should be sufficient, no ?
```suggestion
      EnrichedEvent
```

app/services/events/re_enrich_all_service.rb
```diff
@@ -0,0 +1,94 @@
...
+    Result = BaseResult[]
+
+    def initialize(subscription:, timestamp: Time.current)
+      @subscription = subscription
+      @timestamp = timestamp
+
+      super
+    end
+
+    def call
```
groyoh, line 14: Should we skip if clickhouse store enabled ?

spec/services/events/enrich_service_spec.rb
```diff
@@ -0,0 +1,311 @@
...
+      it "creates an enriched event with the right value" do
+        result = enrich_service.call
+
+        expect(result).to be_success
+        expect(result.enriched_events.count).to eq(1)
+
+        enriched_event = result.enriched_events.first
+        expect(enriched_event).to have_attributes(
+          value: "foo_bar",
+          decimal_value: 0.0
```
groyoh: If the property is a number, this would not be `0`. Should we enforce zero for `unique_count` ?

spec/services/events/enrich_service_spec.rb
```diff
@@ -0,0 +1,311 @@
...
+
+        expect(result).to be_success
+        expect(result.enriched_events.count).to eq(1)
+
+        enriched_event = result.enriched_events.first
+        expect(enriched_event).to have_attributes(
+          charge_id: charge.id,
+          charge_filter_id: charge_filter.id,
+          properties: event.properties,
+          grouped_by: {}
```
groyoh: Maybe check the decimal value ?

spec/services/events/enrich_service_spec.rb
```diff
@@ -0,0 +1,311 @@
...
+          result = enrich_service.call
+
+          expect(result).to be_success
+          expect(result.enriched_events.count).to eq(1)
+
+          enriched_event = result.enriched_events.first
+          expect(enriched_event).to have_attributes(
+            charge_id: charge.id,
+            charge_filter_id: charge_filter.id,
+            properties: {
```
groyoh: Didn't you want to check the `group_by` instead ?

app/services/events/enrich_service.rb
```diff
@@ -0,0 +1,71 @@
...
+      enriched_event.organization_id = event.organization_id
+      enriched_event.code = event.code
+      enriched_event.transaction_id = event.transaction_id
+      enriched_event.timestamp = event.timestamp
+
+      enriched_event.external_subscription_id = subscription.external_id
+      enriched_event.subscription_id = subscription.id
+      enriched_event.plan_id = subscription.plan_id
+
+      enriched_event.properties = event.properties
```
groyoh: Do we need to store all properties ? Since we're already pre-aggregating, we only need the properties for value, grouped_by and filters, no ?

spec/services/events/re_enrich_all_service_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  describe "#call" do
+    it "creates the relevant enriched_events" do
+      result = nil
+
+      expect { result = re_enrich_service.call }.to change(EnrichedEvent, :count).by(1)
+      expect(result).to be_success
+    end
+
+    context "with pre-existing enriched events" do
```
groyoh, line 40: It would be nice to add a more thorough test. For instance, this one doesn't ensure that we recompute all the values. If the code simply deleted 2 enriched events, the test would still pass.

app/services/events/re_enrich_all_service.rb
vincent-pochet: Good catch indeed :)

spec/services/events/enrich_service_spec.rb
vincent-pochet: Since we do not care for the decimal_value in this case, I will remove the check on it to avoid confusion

app/services/events/enrich_service.rb
vincent-pochet: That's a fair point indeed. I'll do a migration in a separate PR to remove this field :+1:

groyoh: Sure. I guess it also depends how we want to use this table in the future. Might be worth keeping if we're not sure.

PR #4920, opened 2026-01-27
Restore migrations that removed role columns

nepalez:
## Context

This is the continuation of [the commit that fixed the OSS releases](https://github.com/getlago/lago-api/pull/4919).
It restores the migration after ignoring the corresponding columns.

## Description

The PR should be applied
* to the production code along with the previous one (so that the migration remained the same for the production server)
* to the OSS releases AFTER releasing the previous one

### DIFF
--- a/app/models/invite.rb
+++ b/app/models/invite.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class Invite < ApplicationRecord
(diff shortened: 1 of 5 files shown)


nepalez:
:bangbang: Converted this to draft to prevent its occasional merging before the next OS release

PR #4921, opened 2026-01-27
Feat(wallet-targets): add accepts_wallet_target to charge

annvelents:
in order to provide a mechanism to target wallets from an event, we need to indicate on a charge that fee for this charge, when calculating, will be grouped by target_wallet_code

### DIFF
--- a/app/services/charges/update_service.rb
+++ b/app/services/charges/update_service.rb
@@ -31,7 +31,7 @@ module Charges
         end
 
         accepts_target_wallet = params.delete(:accepts_target_wallet)
-        if License.premium? && plan.organization.events_targeting_wallets_enabled?
+        if plan.organization.events_targeting_wallets_enabled?
           charge.accepts_target_wallet = accepts_target_wallet
         end
 


app/services/charges/update_service.rb
```diff
@@ -30,6 +30,11 @@ def call
           charge.properties = ChargeModels::FilterPropertiesService.call(chargeable: charge, properties:).properties
         end
 
+        accepts_target_wallet = params.delete(:accepts_target_wallet)
+        if License.premium? && plan.organization.events_targeting_wallets_enabled?
```
mariohd: I believe the `License.premium?` is not necessary because the `events_targeting_wallets_enabled?` already does this check, no?

https://github.com/getlago/lago-api/blob/main/app/models/organization.rb#L176-L178

mariohd: also, 
Isn't the [PREMIUM_INTEGRATIONS](https://github.com/getlago/lago-api/blob/main/app/models/organization.rb:114) `events_targeting_wallets` missing?

annvelents: lecense.premium - you're right, was thinking overcautious thoughts 😅 

Premium integrations - merged on main

PR #4925, opened 2026-01-27
feat(alerting): Introducing Wallet Alerts 

julienbourdeau:
## General Design

Alerts already exists and are linked to subscription via the `external_id`. This was done on purpose to ensure that alerts are kept across downgrade/upgrade.

### Alertable

This PR introduce a new way to **attach alerts to subscriptions XOR wallets.**

Many files were modified due to this but it's small changes. Like services now accept `alertable:` instead of `subscriptions:`.

### Direction

Currently, Alert are based on usage, so it's always trending up 📈 
The alert is triggered only when the previous value is below the thresholds and the current value is above the thresholds.
When usage go down (end of the month reset for instance) there are no alert triggered.

With Wallet, we're tracking "balance" which goes to the opposite direction of usage.
In this case, the alert are triggered when the thresholds is crossed from higher value to lower value.

The `direction` is inferred from the `alert_type` so you cannot set it explicitely. This is done on purpose and could be change one day if needed.

## API

Wallet and alerts are targeted via `code`, no `lago_id` 🫶 

```
  The controller at app/controllers/api/v1/customers/wallets/alerts_controller.rb exposes a RESTful API for managing wallet alerts:
  ┌───────────┬──────────────────────────────────────────────────────────────────┬─────────┬──────────────────────────────────────────┐
  │  Method   │                             Endpoint                             │ Action  │               Description                │
  ├───────────┼──────────────────────────────────────────────────────────────────┼─────────┼──────────────────────────────────────────┤
  │ GET       │ /customers/:customer_external_id/wallets/:code/alerts            │ index   │ List all alerts for a wallet (paginated) │
  ├───────────┼──────────────────────────────────────────────────────────────────┼─────────┼──────────────────────────────────────────┤
  │ GET       │ /customers/:customer_external_id/wallets/:code/alerts/:code      │ show    │ Get a specific alert by code             │
  ├───────────┼──────────────────────────────────────────────────────────────────┼─────────┼──────────────────────────────────────────┤
  │ POST      │ /customers/:customer_external_id/wallets/:code/alerts            │ create  │ Create a new alert                       │
  ├───────────┼──────────────────────────────────────────────────────────────────┼─────────┼──────────────────────────────────────────┤
  │ PUT/PATCH │ /customers/:customer_external_id/wallets/:code/alerts/:code      │ update  │ Update an existing alert                 │
  ├───────────┼──────────────────────────────────────────────────────────────────┼─────────┼──────────────────────────────────────────┤
  │ DELETE    │ /customers/:customer_external_id/wallets/:code/alerts/:code      │ destroy │ Delete an alert                          │
  └───────────┴──────────────────────────────────────────────────────────────────┴─────────┴──────────────────────────────────────────┘
```

## Frontend

No graphql and frontend yet, coming later.

## Attention needed

Things that I think need to actually be reviewed are:
* db migrations
* Scenario test

### DIFF
--- a/spec/scenarios/wallets/alerting/api_error_handling_spec.rb
+++ b/spec/scenarios/wallets/alerting/api_error_handling_spec.rb
@@ -65,7 +65,7 @@ describe "Wallet Alert API Error Handling", :premium, transaction: false do
 
       expect(response[:status]).to eq(422)
       expect(response[:code]).to eq("validation_errors")
-      expect(response[:error_details]).to include("alert_type" => ["value_is_invalid"])
+      expect(response[:error_details]).to include("alert_type" => ["invalid_type"])
     end
   end
 
@@ -124,7 +124,7 @@ describe "Wallet Alert API Error Handling", :premium, transaction: false do
 
       expect(response[:status]).to eq(422)
       expect(response[:code]).to eq("validation_errors")
-      expect(response[:error_details]).to include("alert_type" => ["value_is_mandatory"])
+      expect(response[:error_details]).to include("alert_type" => ["value_is_mandatory", "value_is_invalid"])
     end
   end
 
@@ -172,7 +172,7 @@ describe "Wallet Alert API Error Handling", :premium, transaction: false do
 
       expect(response[:status]).to eq(422)
       expect(response[:code]).to eq("validation_errors")
-      expect(response[:error_details]).to include("alert_type" => ["value_is_invalid"])
+      expect(response[:error_details]).to include("alert_type" => ["invalid_type"])
     end
   end
 
--- a/spec/services/usage_monitoring/alerts/create_batch_service_spec.rb
+++ b/spec/services/usage_monitoring/alerts/create_batch_service_spec.rb
@@ -123,7 +123,7 @@ RSpec.describe UsageMonitoring::Alerts::CreateBatchService do
         expect(result.alerts).to be_empty
         expect(result.error.messages).to have_key(1)
         expect(result.error.messages[1][:params]).to eq(alerts_params[1])
-        expect(result.error.messages[1][:errors]).to include("value_is_invalid")
+        expect(result.error.messages[1][:errors]).to include("invalid_type")
       end
     end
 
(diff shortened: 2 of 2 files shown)


app/models/invoice_subscription.rb
```diff
@@ -153,7 +153,7 @@ def total_amount_currency
 #  index_invoice_subscriptions_on_regenerated_invoice_id          (regenerated_invoice_id)
 #  index_invoice_subscriptions_on_subscription_id                 (subscription_id)
 #  index_uniq_invoice_subscriptions_on_charges_from_to_datetime   (subscription_id,charges_from_datetime,charges_to_datetime) UNIQUE WHERE ((created_at >= '2023-06-09 00:00:00'::timestamp without time zone) AND (recurring IS TRUE) AND (regenerated_invoice_id IS NULL))
-#  index_uniq_invoice_subscriptions_on_fixed_charges_boundaries   (subscription_id,fixed_charges_from_datetime,fixed_charges_to_datetime) UNIQUE WHERE ((fixed_charges_from_datetime IS NOT NULL) AND (recurring = true) AND (regenerated_invoice_id IS NULL))
+#  index_uniq_invoice_subscriptions_on_fixed_charges_boundaries   (subscription_id,fixed_charges_from_datetime,fixed_charges_to_datetime) UNIQUE WHERE ((fixed_charges_from_datetime IS NOT NULL) AND (recurring IS TRUE) AND (regenerated_invoice_id IS NULL))
 #  index_unique_starting_invoice_subscription                     (subscription_id,invoicing_reason) UNIQUE WHERE ((invoicing_reason = 'subscription_starting'::subscription_invoicing_reason) AND (regenerated_invoice_id IS NULL))
```
julienbourdeau, line 157: this is killing me

app/models/usage_monitoring/alert.rb
```diff
@@ -11,17 +11,22 @@ class Alert < ApplicationRecord
...
     BILLABLE_METRIC_TYPES = %w[billable_metric_current_usage_amount billable_metric_current_usage_units]
+    WALLET_TYPES = %w[wallet_balance_amount wallet_credits_balance]
+
+    DIRECTIONS = {increasing: "increasing", decreasing: "decreasing"}.freeze
 
     default_scope -> { kept }
 
     belongs_to :organization
     belongs_to :billable_metric, -> { with_discarded }, optional: true
+    belongs_to :wallet, optional: true
```
vankiru, line 33: nitpick: With `wallet` association defined, it feels a bit odd that there's no association for subscriptions (even though it seems we don't really use it) 🤔

julienbourdeau: I agree! I wondered why it wasn't there when I got into it.

The subscription is linked via an external_subscription_id, not a primary key, so I didn't add it back then. I also never needed it but yeah it feels imbalanced!

app/models/subscription.rb
```diff
@@ -39,6 +39,8 @@ class Subscription < ApplicationRecord
   has_one :lifetime_usage, autosave: true
   has_one :subscription_activity, class_name: "UsageMonitoring::SubscriptionActivity"
 
+  has_many :alerts, ->(s) { where(organization_id: s.organization_id) }, class_name: "UsageMonitoring::Alert", foreign_key: :subscription_external_id, primary_key: :external_id
```
julienbourdeau, line 42: 👏

PR #4927, opened 2026-01-27
feat(billing): add service to finalize VIES-pending invoices

ancorcruz:
## Context

When an invoice is blocked due to VIES validation pending, it remains in pending status until the VIES check completes. Once VIES validation succeeds, these invoices need to be finalized with the correct taxes.

## Description

Add FinalizePendingViesInvoiceService that handles finalization of invoices that were blocked due to pending VIES validation. The service applies local taxes, sets issuing/payment dates, applies credits, and triggers all finalization jobs (webhooks, document generation, payments, integrations). Customers with tax providers are skipped as their invoices are handled by PullTaxesAndApplyService instead.


groyoh (commented):
Can we re-use some existing code ? It feels like we're copy-pasting a lot of code here which will make it harder and harder to keep everything in sync.

PR #4929, opened 2026-01-28
fix: when priority is null should not be considered

brunomiguelpinto:
## Context

Wallet creation previously assigned `priority` whenever the key existed in params.
After switching to a truthiness check, `priority: nil` now behaves like “not provided”.

## Description

When `priority` is `nil`, the wallet is created successfully, and the priority defaults to `50`.


annvelents (approved):
🙌

PR #4931, opened 2026-01-28
feat(billing): enqueue invoice finalization after VIES success

ancorcruz:
## Context

When VIES validation succeeds after being temporarily unavailable, invoices that were blocked due to the pending VIES check need to be finalized.

## Description

Enhance RetryViesCheckJob to find all pending invoices with pending tax_status for the customer and enqueue FinalizePendingViesInvoiceJob for each one after the VIES check succeeds and taxes are applied.

### DIFF
--- a/spec/jobs/customers/retry_vies_check_job_spec.rb
+++ b/spec/jobs/customers/retry_vies_check_job_spec.rb
@@ -107,16 +107,5 @@ RSpec.describe Customers::RetryViesCheckJob do
     it "enqueues another retry job" do
       expect { described_class.perform_now(customer.id) }.to have_enqueued_job(described_class)
     end
-
-    xit "does not apply taxes" do
-      described_class.perform_now(customer.id)
-
-      expect(Customers::ApplyTaxesService).not_to have_received(:call)
-    end
-
-    xit "does not enqueue invoice finalization" do
-      expect { described_class.perform_now(customer.id) }
-        .not_to have_enqueued_job(Invoices::FinalizePendingViesInvoiceJob)
-    end
   end
 end


app/jobs/customers/retry_vies_check_job.rb
```diff
@@ -22,6 +22,17 @@ def perform(customer_id)
           tax_codes: [tax_code]
         )
       end
+
+      # Finalize any invoices that were blocked by VIES
+      enqueue_pending_invoice_finalization(customer)
```
groyoh, line 26: There's something I don't necessarily understand here. From the code below, we silence certain errors from VIES: https://github.com/getlago/lago-api/blob/2e1df2cb48f0e26939cc9ebf4ffc4a90320f37b2/app/services/customers/eu_auto_taxes_service.rb#L46-L47

In such case, we would return a code based on the following code: https://github.com/getlago/lago-api/blob/2e1df2cb48f0e26939cc9ebf4ffc4a90320f37b2/app/services/customers/eu_auto_taxes_service.rb#L78-L83

Shouldn't we raise in such case ? 🤔 

Also should we return early here instead of schedluing jobs:

```suggestion
      enqueue_pending_invoice_finalization(customer) if customer.pending_vies_check.none?
```

app/jobs/customers/retry_vies_check_job.rb
```diff
@@ -22,6 +22,17 @@ def perform(customer_id)
...
+    end
+
+    private
+
+    def enqueue_pending_invoice_finalization(customer)
+      customer.invoices.pending.where(tax_status: "pending").find_each do |invoice|
```
lovrocolic, line 32: Just confirming, there is no case where we could by accidence fetch tax provider pending invoices?

ancorcruz: those will be skipped later in the `Invoices::FinalizePendingViesInvoiceService` run.

app/jobs/customers/retry_vies_check_job.rb
ancorcruz: @groyoh that service will be handled in an upcoming PR (https://github.com/getlago/lago-api/pull/4932)

PR #4932, opened 2026-01-28
feat(billing): prevent invoice finalization during VIES validation

ancorcruz:
# Context

When a customer's VAT number validation against the EU VIES system fails due to transient errors (rate limits, timeouts, service unavailability), the system was falling back to a default tax rate and proceeding with invoice creation. This could result  in incorrect tax calculations if the VAT number is actually valid but VIES was temporarily unavailable.

# Description

This PR changes the VIES failure behavior to block invoice finalization until validation succeeds:

- EuAutoTaxesService: When VIES check fails due to transient errors, the service now returns an error instead of a default tax rate. It creates a PendingViesCheck record to track failures and implements exponential backoff for retries (5min → 10min → 20min → 40min → 1hr cap). The pending check is automatically deleted when VIES validation eventually succeeds.
- EnsureCompletedViesCheckService: New service that checks if VIES validation is pending for a customer. When called during invoice finalization, it sets the invoice to pending status with pending tax_status and returns an unknown_tax_failure to block finalization.
- Pay-in-advance fixed charges: Integrated VIES check into CreatePayInAdvanceFixedChargesService. When VIES is pending, invoices are created with pending status instead of being finalized. The job handles tax errors gracefully to avoid creating dead jobs in Sidekiq.
- Invoice finalization: ComputeTaxesAndTotalsService now calls the VIES check service, ensuring all invoice types respect pending VIES validation.

Once the VIES retry succeeds, the FinalizePendingViesInvoiceJob automatically finalizes any blocked invoices with the correct tax rate.

### DIFF
--- a/app/services/invoices/finalize_pending_vies_invoice_service.rb
+++ b/app/services/invoices/finalize_pending_vies_invoice_service.rb
@@ -21,6 +21,7 @@ module Invoices
         invoice.payment_due_date = payment_due_date
 
         Invoices::ComputeAmountsFromFees.call(invoice:)
+        Invoices::ApplyInvoiceCustomSectionsService.call(invoice:)
 
         create_credit_note_credit if should_create_credit_note_credit?
         create_applied_prepaid_credit if should_create_applied_prepaid_credit?
--- a/spec/services/invoices/finalize_pending_vies_invoice_service_spec.rb
+++ b/spec/services/invoices/finalize_pending_vies_invoice_service_spec.rb
@@ -149,6 +149,14 @@ RSpec.describe Invoices::FinalizePendingViesInvoiceService do
         expect(invoice.total_amount_cents).to be_positive
       end
 
+      it "applies invoice custom sections" do
+        allow(Invoices::ApplyInvoiceCustomSectionsService).to receive(:call)
+
+        finalize_service.call
+
+        expect(Invoices::ApplyInvoiceCustomSectionsService).to have_received(:call).with(invoice:)
+      end
+
       it "sets payment_status to pending when total is positive" do
         finalize_service.call
 


app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
```diff
@@ -42,6 +42,9 @@ def call
             Utils::ActivityLog.produce(invoice, "invoice.failed")
             next
           end
+        else
+          @vies_check_result = Invoices::EnsureCompletedViesCheckService.call(invoice:)
```
groyoh: I'd suggest not to rely on instance variable here. We can simply define a `tax_error = false` before the transaction and set it to `true` if there's a failure here.

groyoh: Above we handle activity log and webhook for fees. Shouldn ht we handle it here as well ?

app/services/invoices/ensure_completed_vies_check_service.rb
```diff
@@ -0,0 +1,40 @@
...
+      @finalizing = finalizing
+
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "invoice") unless invoice
+
+      # VIES does not apply if EU tax management is disabled
+      return result unless customer.billing_entity.eu_tax_management?
```
groyoh: Already checked in `customer.vies_check_in_progress?`

app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -104,5 +107,47 @@ def should_apply_eu_taxes?
...
+      case attempts
+      when 0..1 then 5.minutes
+      when 2 then 10.minutes
+      when 3 then 20.minutes
+      when 4 then 40.minutes
+      else 1.hour
+      end
+    end
+
+    def error_type_for(exception)
```
groyoh: Use `PendingViesCheck::error_type_for` instead

app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -24,7 +24,22 @@ def call
...
+      end
+
+      delete_pending_vies_check_if_exists
       result
+    rescue Valvat::RateLimitError, Valvat::Timeout, Valvat::BlockedError, Valvat::InvalidRequester,
+      Valvat::ServiceUnavailable, Valvat::MemberStateUnavailable => e
+      create_or_update_pending_vies_check(e)
+      after_commit do
+        SendWebhookJob.perform_later("customer.vies_check", customer, vies_check: error_vies_check.merge(error: e.message))
+        RetryViesCheckJob.set(wait: retry_delay).perform_later(customer.id)
```
groyoh: To avoid reloading the customer:

```suggestion
      check = create_or_update_pending_vies_check(e)
      after_commit do
        SendWebhookJob.perform_later("customer.vies_check", customer, vies_check: error_vies_check.merge(error: e.message))
        RetryViesCheckJob.set(wait: retry_delay(check)).perform_later(customer.id)
```

app/jobs/invoices/create_pay_in_advance_fixed_charges_job.rb
```diff
@@ -11,10 +11,26 @@ class CreatePayInAdvanceFixedChargesJob < ApplicationJob
...
+
+      result.raise_if_error!
+    end
+
+    private
+
+    def tax_error?(result)
+      return false unless result.error.is_a?(BaseService::ValidationFailure)
+
+      result.error&.messages&.dig(:tax_error).present?
```
groyoh:
```suggestion
      result.error.messages&.dig(:tax_error).present?
```

app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -104,5 +107,47 @@ def should_apply_eu_taxes?
...
+      attempts = customer.reload.pending_vies_check&.attempts_count.to_i
+
+      case attempts
+      when 0..1 then 5.minutes
+      when 2 then 10.minutes
+      when 3 then 20.minutes
+      when 4 then 40.minutes
+      else 1.hour
+      end
+    end
```
groyoh:
```suggestion
RETRY_DELAYS = [5.minutes, 5.minutes, 10.minutes, 20.minutes, 40.minutes].freeze
MAX_RETRY_DELAY = 1.hour

def retry_delay
  attempts = customer.reload.pending_vies_check&.attempts_count.to_i
  RETRY_DELAYS[attempts] || MAX_RETRY_DELAY
end
```

app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
ancorcruz: Good catch on the asymmetry. I looked into how both error branches behave here and in the subscription flow:

In this service, the provider tax error branch sends `fee.created` webhooks and produces an `invoice.failed` activity log, but the VIES pending branch does neither.

In the subscription flow (`SubscriptionService` → `CalculateFeesService` → `ComputeTaxesAndTotalsService`), neither provider tax errors nor VIES pending send fee webhooks or activity logs (only `invoice.drafted` if there's a grace period).

So the subscription flow is internally consistent (no webhooks/activity log for either case), but this service is not — provider tax errors get webhooks + activity log, VIES pending doesn't.

I'll add the fee webhooks and activity log to the VIES pending branch to make it consistent within this service.

app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
```diff
@@ -34,14 +37,22 @@ def call
...
-            create_error_detail(@taxes_result.error.code)
+            create_error_detail(tax_error)
             Utils::ActivityLog.produce(invoice, "invoice.failed")
             next
           end
+        else
+          vies_check_failed = Invoices::EnsureCompletedViesCheckService.call(invoice:).failure?
+          if vies_check_failed
+            invoice.fees.each { |f| SendWebhookJob.perform_later("fee.created", f) }
+            Utils::ActivityLog.produce(invoice, "invoice.pending")
```
groyoh: Should be wrapped in a method to avoid any mistake in the future.

app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
```diff
@@ -34,14 +37,22 @@ def call
...
+            next
```
groyoh, line 54: Here we'll skip `ApplyInvoiceCustomSectionsService` which won't be called in `FinalizePendingViesInvoiceService`. Is that expected ?

app/services/customers/eu_auto_taxes_service.rb
```diff
@@ -104,5 +110,28 @@ def should_apply_eu_taxes?
...
+      pending_check.save!
+      pending_check
+    end
+
+    def delete_pending_vies_check_if_exists
+      customer.pending_vies_check&.destroy!
+    end
+
+    def retry_delay(pending_vies_check)
+      RETRY_DELAYS[pending_vies_check.attempts_count.to_i] || MAX_RETRY_DELAY
```
groyoh, line 134: Since we update attempt beforehand:

```suggestion
      RETRY_DELAYS[pending_vies_check.attempts_count - 1] || MAX_RETRY_DELAY
```

PR #4933, opened 2026-01-28
feat(multiple payments): Expose payment_provider_name and id fields

domenicofalco:
## Context

  When customers have multiple payment providers configured, API consumers need
  to identify which provider connection a payment method belongs to. The existing
  payment_provider_code is available but the human-readable connection name was
  not exposed.

  ## Description

  Add payment_provider_name field to the PaymentMethod resource in both the
  GraphQL API and REST API serializer. This field returns the name of the
  associated payment provider connection, allowing consumers to display a
  user-friendly identifier for the payment method's provider.

### DIFF
--- a/app/graphql/types/payment_methods/object.rb
+++ b/app/graphql/types/payment_methods/object.rb
@@ -14,6 +14,7 @@ module Types
       field :payment_provider_customer_id, ID, null: true
       field :payment_provider_name, String, null: true
       field :payment_provider_type, Types::PaymentProviders::ProviderTypeEnum, null: true
+      field :provider_method_id, String, null: false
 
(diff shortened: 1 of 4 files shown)


app/graphql/types/payment_methods/object.rb
```diff
@@ -12,6 +12,7 @@ class Object < Types::BaseObject
       field :is_default, Boolean, null: false
       field :payment_provider_code, String, null: true
       field :payment_provider_customer_id, ID, null: true
+      field :payment_provider_name, String, null: true
```
lovrocolic: I guess you only need `provider_method_id` here

mariohd (approved):
Looks correct

PR #4934, opened 2026-01-28
feat: Add credit note types filter

brunomiguelpinto:
## Context

Users need to filter credit notes by their “type” at a high level to distinguish between credit issuance, refunds, and invoice offsets.

## Description

- Add a `types` filter to credit note listing that supports multi-select.
- Supported values are `credit`, `refund`, and `offset` (centralized in `CreditNote::TYPES`).

- The filter uses OR semantics when multiple values are provided (match any selected type).
- A credit note “type” is inferred from amounts:
  - `credit` => `credit_amount_cents > 0`
  - `refund` => `refund_amount_cents > 0`
  - `offset` => `offset_amount_cents > 0`


D1353L (approved):

annvelents (approved):
LGTM! I'd only suggest to add a test with no results :)

PR #4940, opened 2026-01-29
feat(invoice): Add settlements filter to invoices index

brunomiguelpinto:
## Context
We need to be able to filter invoices based on how they were settled, to support workflows around credit note settlements and payment settlements.

## Description
This change adds support for a settlements filter on invoice listing endpoints and query interfaces, allowing consumers to retrieve invoices that have settlements matching one or more settlement types. The filter accepts a single settlement type or multiple settlement types (including comma-separated values) and applies the corresponding filtering on invoice settlements.

### Note: 
for now we only have credit note on settlements

### DIFF
--- a/app/queries/invoices_query.rb
+++ b/app/queries/invoices_query.rb
@@ -199,7 +199,9 @@ class InvoicesQuery < BaseQuery
 
   def with_settlements(scope)
     scope.where(
-      id: InvoiceSettlement.where(settlement_type: valid_settlements).distinct.select(:target_invoice_id)
+      id: scope.joins(:invoice_settlements)
+        .where(invoice_settlements: {settlement_type: valid_settlements})
+        .select(:id)
     )
   end
 


app/queries/invoices_query.rb
```diff
@@ -195,11 +197,29 @@ def with_self_billed(scope)
     scope.where(self_billed: ActiveModel::Type::Boolean.new.cast(filters.self_billed))
   end
 
+  def with_settlements(scope)
+    subquery = scope
+      .joins(:invoice_settlements)
+      .where(invoice_settlements: {settlement_type: valid_settlements}).limit(nil)
+      .offset(nil)
+
+    scope.where(id: subquery.select(:id))
```
julienbourdeau: This feels weird to have a subquery built from scope, they use this subquery to extend the scope 🤔 

Can you post examples of the SQL generated? Did you explore any alternative, like doing it in multiple query or with CTE maybe?

Note that I could be totally wrong and this could be the best way.

brunomiguelpinto: You are rigth changed to a simpler version

spec/queries/invoices_query_spec.rb
```diff
@@ -829,4 +829,62 @@
       expect(returned_ids).to eq([invoice_with_subscription_1.id])
     end
   end
+
+  context "when filtering by settlements" do
```
annvelents, line 833: might also worth adding tests:
1) when there are no matches
2) when trying to filter by not-existing settlement

app/queries/invoices_query.rb
```diff
@@ -195,11 +197,22 @@ def with_self_billed(scope)
     scope.where(self_billed: ActiveModel::Type::Boolean.new.cast(filters.self_billed))
   end
 
+  def with_settlements(scope)
+    scope.where(
+      id: InvoiceSettlement.where(settlement_type: valid_settlements).distinct.select(:target_invoice_id)
+    )
```
julienbourdeau, line 205: I'm afraid this will include ALL invoice_ids that ever had this settlement types.

Definitely needs to be tested but what do you think of this 🤔 

```suggestion
    scope
      .joins(:settlements)
      .where(settlements: { settlement_type: valid_settlements })
      .distinct
```

brunomiguelpinto: Let me do some tests and check que queries

brunomiguelpinto:
```
  def with_settlements(scope)
    scope.where(
      id: scope.joins(:invoice_settlements)
        .where(invoice_settlements: {settlement_type: valid_settlements})
        .select(:id)
    )
  end
```

  
  This is the safest one since we keep all the filters from the invoice organization id etc

PR #4946, opened 2026-01-30
feat(alerting): Better errors for thresholds

julienbourdeau:
As I'm doing some [QA on wallet alerts](https://github.com/getlago/lago-api/pull/4925), I realized the error for wrong thresholds params was not handled.

I'm doing it separately because I don't know when wallet alerts would be merged, even if it's going to be annoying to rebase.

## BEFORE

CleanShot 2026-01-30 at 10 41 34@2x

## AFTER

CleanShot 2026-01-30 at 10 56 24@2x

CleanShot 2026-01-30 at 10 56 03@2x


ancorcruz (approved):
LGTM

PR #4949, opened 2026-01-30
feat(security logs): implementation

nepalez:
## Context

Enterprise customers require visibility into security-critical operations for compliance and incident investigation. Security Logs track user management, role changes, API key rotations, and webhook configuration.

## Description

**Service Interface** (`Utils::SecurityLog`) unblocks parallel work on event integrations — business services can start calling `produce` immediately, before the storage is implemented.

**GraphQL API Interface** unblocks frontend integration — UI can use real API contracts while receiving empty results until storage is ready.

Both are stubs. The actual implementation (ClickHouse storage, Kafka producer) will follow in a separate PR.

### DIFF
--- a/app/graphql/types/security_logs/log_event_enum.rb
+++ b/app/graphql/types/security_logs/log_event_enum.rb
@@ -5,8 +5,9 @@ module Types
     class LogEventEnum < Types::BaseEnum
       description "Security Log event"
 
-      # More events will be added as event integrations are implemented
-      value "user_signed_up", value: "user.signed_up", description: "User signed up"
+      Clickhouse::SecurityLog::LOG_EVENTS.each do |event|
+        value event.tr(".", "_"), value: event, description: event
+      end
     end
   end
 end
--- a/app/services/utils/security_log.rb
+++ b/app/services/utils/security_log.rb
@@ -18,19 +18,12 @@ module Utils
     # @param log_type [String] event category (e.g. "user", or "api_key")
     # @param log_event [String] specific event (e.g. "user.invited")
     # @param user [User, nil] the user who performed the action (nil for API key operations)
+    # @param api_key [ApiKey, nil] the API key used for the action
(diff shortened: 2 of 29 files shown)


app/services/utils/security_log.rb
```diff
@@ -0,0 +1,45 @@
...
+    # Produces a security log event to Kafka.
+    #
+    # @param organization [Organization] the organization context
+    # @param log_type [String] event category (e.g. "user", or "api_key")
+    # @param log_event [String] specific event (e.g. "user.invited")
+    # @param user [User, nil] the user who performed the action (nil for API key operations)
+    # @param resources [Hash] additional context (e.g., {invitee_email: "..."})
+    # @param device_info [Hash] device metadata for login events
+    # @return [Boolean] true if log was produced, false otherwise
+    def self.produce(
```
murparreira: So, we need to use `SecurityLog.produce` on every place we will log an event?

app/graphql/types/security_logs/log_event_enum.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Types
+  module SecurityLogs
+    class LogEventEnum < Types::BaseEnum
+      description "Security Log event"
+
+      # More events will be added as event integrations are implemented
+      value "user_signed_up", value: "user.signed_up", description: "User signed up"
```
murparreira: We already have that list no? Of all events.

app/services/utils/security_log.rb
nepalez: yes

app/graphql/types/security_logs/log_event_enum.rb
nepalez: We have, but I found it safer to register a new event at the moment we implement its logging. Here I've only added one value to provide the framework/enum by itself

PR #4952, opened 2026-01-30
feat(wallet-targets): events-post-process

annvelents:
When accepting wallet_targets is turned on on the charge, and we receive an event with target_wallet_code, we need to ensure that this wallet exist on the customer. If it is not, we should send a webhook that this wallet does not exist (otherwise standard application logic will be applied)

### DIFF
--- a/app/services/events/post_process_service.rb
+++ b/app/services/events/post_process_service.rb
@@ -15,7 +15,8 @@ module Events
       create_enriched_events
       track_subscription_activity
       customer&.flag_wallets_for_refresh
-      check_targeted_wallets if organization.events_targeting_wallets_enabled? && event.properties["target_wallet_code"].present?
+      # TODO: update also event-processor to process targeted wallets
+      check_targeted_wallets
 
       handle_pay_in_advance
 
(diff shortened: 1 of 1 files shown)


app/services/events/post_process_service.rb
```diff
@@ -88,6 +89,19 @@ def track_subscription_activity
       end
     end
 
+    def check_targeted_wallets
+      return unless subscriptions
+      return unless Charge.where(organization_id: event.organization_id, plan_id: subscriptions.map(&:plan_id),
+                                 billable_metric:, accepts_target_wallet: true).exists?
+      return if customer.wallets.where(code: event.properties["target_wallet_code"]).exists?
```
annvelents: .active

app/services/events/post_process_service.rb
```diff
@@ -15,6 +15,7 @@ def call
       create_enriched_events
       track_subscription_activity
       customer&.flag_wallets_for_refresh
+      check_targeted_wallets if organization.events_targeting_wallets_enabled? && event.properties["target_wallet_code"].present?
```
julienbourdeau: Just for style, I'd put `check_target_wallets` here and move the condition to the method `return unless ...`
I feel like we're breaking the rhythm here 😄

annvelents: hah, I wanted to highlight that we need it in a very rare case comparing to the other events 😅 
happy to change though 😁

PR #4956, opened 2026-02-02
fix(multi-payment-methods): Discard old provider customer's associated payment methods when payment provider is removed/replaced

osmarluz:
## Context

A Lago customer can be associated with a customer in a payment provider (e.g., Stripe). In Lago, if a customer is disconnected from a payment provider (removed when it's performed through the UI and replaced when it's performed through the API), we want to discard the old provider customer's associated payment methods.

## Description

- On the flow done through the UI (handled through the `app/services/customers/update_service.rb`) we added the code for discarding the old provider customer's associated payment methods whenever the payment provider is **removed** for the customer. There is a bug with the discarding logic of the provider customer that will be fixed in a separate ticket/branch.
- On the flow done through the API (handled through the `app/services/customers/upsert_from_api_service.rb`) we handle the discarding of the old provider customer's associated payment methods whenever the payment provider is **replaced** for the customer and we pass the `provider_customer_id` parameter. When we rely on the `sync_with_provider`, there's also a bug on the existing flow that's documented on the `api/spec/services/customers/upsert_from_api_service_spec.rb` spec


app/services/customers/upsert_from_api_service.rb
```diff
@@ -243,12 +246,19 @@ def handle_api_billing_configuration(customer, new_customer)
...
 
       update_provider_customer = (billing || {})[:provider_customer_id].present?
-      update_provider_customer ||= customer.provider_customer&.provider_customer_id.present?
+      update_provider_customer ||= customer.provider_customer&.provider_customer_id.present? || old_provider_customer.present?
 
       return unless update_provider_customer
 
       create_or_update_provider_customer(customer, billing)
 
+      if old_provider_customer && old_payment_provider != billing[:payment_provider]
```
lovrocolic: This is the condition that introduced issue on production and it is still here

billing[:payment_provider] can be nil

lovrocolic (changes requested):
Condition that introduced issue is still here as I can see

jdenquin:
can we make sure that we are extra cautious here? I want to see a detailed QA plan/success and better test coverage 🙏

lovrocolic:
> can we make sure that we are extra cautious here? I want to see a detailed QA plan/success and better test coverage 🙏

Yep, we just synced, and the only changes added here would be related to discarding payment methods for de-attached payment provider. It will be safe to merge it since there are no payment methods objects in prod and that would be unreachable code. That was the original goal of the PR, but this logic overall is messy and Osmar wanted to handle few other bugs along the way, which unfortunately had some unwanted impact.

To sum up, changes here would be related only to code that is not affecting prod customers and that would be tested during QA (and is safe in main). All other changes would be treated as a bug in separate PR.

app/services/customers/upsert_from_api_service.rb
lovrocolic, line 256: if `customer.payment_provider.nil?` is `true`, I think we would exit here and we would not reach your logic?

spec/services/customers/upsert_from_api_service_spec.rb
```diff
@@ -907,6 +907,83 @@
...
+          end
+        end
+
+        context "when provider_customer_id is not sent" do
+          let(:create_args) do
+            {
+              external_id:,
+              billing_configuration: {
+                sync_with_provider: true,
+                payment_provider: "gocardless",
```
lovrocolic, line 1010: you miss the test case where `payment_provider: nil`

app/services/customers/upsert_from_api_service.rb
osmarluz: @lovrocolic yes, that's correct. But I think we should only discard the payment methods in this scenario when the `payment_provider` is explicitly `nil`. The `customer.payment_provider.nil?` also applies when an invalid value is passed, which can be mistake. What do you think?

lovrocolic: But we would set it to nil in the line 237, if nil is passed

lovrocolic: What would happen if someone sends 
```
billing: {
  payment_provider: nil
}
```

app/services/customers/upsert_from_api_service.rb
```diff
@@ -264,6 +287,18 @@ def create_or_update_provider_customer(customer, billing_configuration = {})
...
+    def discard_all_payment_methods(payment_methods)
+      payment_methods.find_each do |payment_method|
+        PaymentMethods::DestroyService.call(payment_method:)
+      end
+    end
+
+    def discard_payment_methods_tied_to_old_payment_provider(payment_methods, old_payment_provider_id)
+      payment_methods.where(payment_provider_id: old_payment_provider_id).find_each do |payment_method|
+        PaymentMethods::DestroyService.call(payment_method:)
+      end
```
mariohd:
```suggestion
      payment_methods = payment_methods.where(payment_provider_id: old_payment_provider_id)
      discard_all_payment_methods(payment_methods)
```

osmarluz: @mariohd done, thanks!

app/services/customers/upsert_from_api_service.rb
```diff
@@ -252,6 +260,21 @@ def handle_api_billing_configuration(customer, new_customer)
...
+
+      if old_provider_customer
+        new_payment_provider_id = payment_provider(customer)&.id
+
+        if old_payment_provider != customer.payment_provider
+          discard_payment_methods(old_provider_customer.payment_methods)
+        elsif old_payment_provider_id.present? &&
+            new_payment_provider_id.present? &&
+            old_payment_provider_id != new_payment_provider_id
+          discard_payment_methods_tied_to_old_payment_provider(
```
lovrocolic: I am not sure this method is needed. Payment methods are related to provider customer and we can also use something like: `discard_payment_methods(old_provider_customer.payment_methods)`

app/services/customers/upsert_from_api_service.rb
```diff
@@ -230,6 +230,10 @@ def handle_api_billing_configuration(customer, new_customer)
         return
       end
 
+      old_provider_customer = customer.provider_customer
+      old_payment_provider = customer.payment_provider
+      old_payment_provider_id = payment_provider(customer)&.id
```
lovrocolic: It seems to me that using this method is not safe as it is. Please take a look at the definition. There are cases when this method can return error and not nil

lovrocolic: Can you please cover those cases with the specs

osmarluz: @lovrocolic you're right, there's a scenario where this might fail when attempting to retrieve the id for the old_payment_provider:

- The `customer.payment_provider_code` gets updated to `nil` while there's just one payment provider, so the payment_provider(customer) returns the one option, even when the payment_provider_code is nil
- We then add a second payment provider of the same type
- We run the update flow once again. It will try to get the payment provider that was used before without a payment_provider_code, but this time it will raise the error because there are 2 payment providers available

I'll check if there's a safe way to retrieve this, but probably not without saving the payment_provider_code like we do on line 211

lovrocolic: Please do not change existing logic, you can also rescue error or handle service response and return nil

app/services/customers/upsert_from_api_service.rb
osmarluz: @lovrocolic I'm double checking the logic on `api/app/services/payment_providers/stripe/customers/create_service.rb` to be sure, specially for the `should_create_provider_customer?` and `should_fetch_payment_method?`. We might not need this indeed

lovrocolic: What I wanted to say is, since payment method already have association to provider customer, it seems to me that filtering by payment providers is not necessary

osmarluz: @lovrocolic it's safe to discard everything with `discard_payment_methods(old_provider_customer.payment_methods)` because `should_fetch_payment_method?` never evaluates to `true` when just the connected account is switched, so we don't have a payment method getting created while we discard others.

I was additionally doing `payment_methods.where(payment_provider_id: old_payment_provider_id)` to prevent the newly created default payment method from getting discarded along with the ones that should, since they're all associated to the same object

app/services/customers/upsert_from_api_service.rb
osmarluz: @lovrocolic done!

spec/services/customers/update_service_spec.rb
```diff
@@ -407,6 +407,62 @@
...
+      let(:customer) do
+        create(
+          :customer,
+          organization:,
+          billing_entity:,
+          payment_provider: "stripe",
+          payment_provider_code:
+        )
+      end
+      let(:stripe_customer) { create(:stripe_customer, customer:) }
```
lovrocolic: Can you please link this to payment provider. The same comment applies to payment method

spec/services/customers/update_service_spec.rb
```diff
@@ -407,6 +407,62 @@
...
+      end
+
+      it "sets the customer parameters to nil" do
+        result = customers_service.call
+
+        expect(result).to be_success
+
+        customer = result.customer
+        expect(customer.payment_provider).to be_nil
+        expect(customer.provider_customer).to be_nil
```
lovrocolic, line 429: Why is this nil. My understanding is that customer.provider_customer = customer.stripe_customer

Based on the test below, stripe customer is not nil

spec/services/customers/upsert_from_api_service_spec.rb
```diff
@@ -907,6 +907,258 @@
...
+          # NOTE: This describes a scenario with incorrect behavior that currently exists.
+          #       The new provider customer does not get created and the previous one is not discarded
+          it "does not create the gocardless provider customer" do
+            expect(result).to be_success
+
+            expect(result.customer.payment_provider).to eq("gocardless")
+            expect(result.customer.provider_customer).to be_nil
+          end
+
+          it "does not discard the old provider customer's payment methods" do
```
lovrocolic: why dont we discard PMs in that case

spec/services/customers/upsert_from_api_service_spec.rb
```diff
@@ -907,6 +907,258 @@
...
+            expect(result).to be_success
+
+            expect(payment_method.reload).to be_discarded
+          end
+        end
+
+        # NOTE: This a scenario with incorrect behavior that currently exists.
+        #       The old customer ID doesn't exist on the new Stripe account, causing an error
+        #       when trying to update the customer on Stripe.
+        context "when provider_customer_id is not sent" do
```
lovrocolic, line 1110: TBH I am not sure if we should have those tests. Can you please share the line that is failing?

spec/services/customers/update_service_spec.rb
osmarluz: @lovrocolic because of line 74 of `api/app/services/customers/update_service.rb`:

```ruby
customer.payment_provider = args[:payment_provider] if args.key?(:payment_provider)
```

The stripe_customer still exists and is not discarded, but the provider_customer method will return nil because we're setting the payment_provider to nil

spec/services/customers/upsert_from_api_service_spec.rb
osmarluz: @lovrocolic because of this check that we perform on the:

```ruby
update_provider_customer = (billing || {})[:provider_customer_id].present?
update_provider_customer ||= customer.provider_customer&.provider_customer_id.present?

```

On this scenario here:

- `billing || {})[:provider_customer_id].present? = false` because the provider_customer_id is not present
- `customer.provider_customer&.provider_customer_id.present? = false` because we don't have a provider_customer that matches the "gocardless" payment_provider yet.

So this leaves the customer in an inconsistent state where:

- `customer.payment_provider = "gocardless"`
- `customer.provider_customer = nil`
- `customer.payment_methods = not discarded`

spec/services/customers/upsert_from_api_service_spec.rb
osmarluz: @lovrocolic this is not testing anything indeed. I just wanted to have it here for documentation purposes. The problem is that this scenario fails when we call `PaymentProviderCustomers::UpdateService.call(customer)` and we hide this behavior on the tests because we always stub the `Stripe::Customer` using `allow(Stripe::Customer).to receive(:update).and_return(BaseService::Result.new)`. In a real scenario this will raise:

```ruby
lago_api_dev        | D, [2026-02-05T16:46:14.856571 #77] DEBUG -- : [httplog] Sending: POST http://api.stripe.com:443/v1/customers/test
lago_api_dev        | D, [2026-02-05T16:46:14.856667 #77] DEBUG -- : [httplog] Data: name=
lago_api_dev        | D, [2026-02-05T16:46:14.856748 #77] DEBUG -- : [httplog] Status: 404
lago_api_dev        | D, [2026-02-05T16:46:14.856815 #77] DEBUG -- : [httplog] Benchmark: 0.28176 seconds
lago_api_dev        | D, [2026-02-05T16:46:14.856869 #77] DEBUG -- : [httplog] Response:
lago_api_dev        | {
lago_api_dev        |   "error": {
lago_api_dev        |     "code": "resource_missing",
lago_api_dev        |     "doc_url": "https://stripe.com/docs/error-codes/resource-missing",
lago_api_dev        |     "message": "No such customer: 'test'",
lago_api_dev        |     "param": "id",
lago_api_dev        |     "request_log_url": "https://dashboard.stripe.com/acct_1Ss2QyPPWqFajAz1/test/workbench/logs?object=req_WLiaYg7fRRExFX",
lago_api_dev        |     "type": "invalid_request_error"
lago_api_dev        |   }
lago_api_dev        | }
```

spec/services/customers/update_service_spec.rb
osmarluz: @lovrocolic done!

PR #4970, opened 2026-02-03
fix(taxes): improve error details tracking with provider

vincent-pochet:
## Context

The debug of taxes provider related errors is hard as we are not correctly storing the error received from the provider, letting us with some cryptic `action_script_failure` without additional context.

This happen when the provider returns this type of message:
```json
{
  "error": {
    "message": "The action script failed with an error: {}",
    "code": "action_script_failure",
    "payload": {
      "error": "Error starting integration 'netsuite-customer-create': {\n  \"name\": \"TRPCClientError\",\n  \"message\": \"fetch failed\"\n}"
    }
  }
}
```

## Description

This PR tries to improves this situation in two distinct path:
- When applied the provider taxes to an invoice, by making sure that the `error_details` object created in case of failures contains the full error message from the provider.

- When computing the current usage with applied taxes for wallet refresh, in case of error, ensure that the dead jobs contains the full error message, to make it easier to understand the root cause

### DIFF
--- a/spec/services/customers/refresh_wallets_service_spec.rb
+++ b/spec/services/customers/refresh_wallets_service_spec.rb
@@ -171,7 +171,7 @@ RSpec.describe Customers::RefreshWalletsService do
       it "fails with an error" do
         expect(result).to be_failure
         expect(result.error).to be_a(BaseService::ValidationFailure)
-        expect(result.error.messages[:tax_error]).to eq(["customerAddressCouldNotResolve"])
+        expect(result.error.messages[:tax_error]).to eq(["customerAddressCouldNotResolve: Customer address could not resolve"])
       end
     end
   end


spec/services/integrations/aggregator/taxes/invoices/create_draft_service_spec.rb
```diff
@@ -291,6 +291,25 @@
...
+        let(:response_status) { Faker::Number.between(from: 500, to: 599) }
+        let(:body) do
+          path = Rails.root.join("spec/fixtures/integration_aggregator/error_script_response.json")
+          File.read(path)
+        end
+
+        it "returns an error" do
+          result = service_call
+
+          aggregate_failures do
```
groyoh: Not needed.

app/services/invoices/customer_usage_service.rb
```diff
@@ -175,7 +175,7 @@ def compute_amounts_with_provider_taxes
 
       taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateDraftService.call(invoice:, fees: invoice.fees)
 
-      return result.validation_failure!(errors: {tax_error: [taxes_result.error.code]}) unless taxes_result.success?
+      return result.validation_failure!(errors: {tax_error: [taxes_result.error.message]}) unless taxes_result.success?
```
groyoh, line 178: This feels like it could be a breaking change. Julien's PR about error with metadata could be useful for this purpose.

spec/services/integrations/aggregator/taxes/invoices/create_draft_service_spec.rb
vincent-pochet: How indeed, bat copy and paste 🙂

app/services/invoices/customer_usage_service.rb
vincent-pochet: Today, we are only checking for the presence of the `tax_error` key in the payload. The only difference will be the format of the reported error, it will keep containing the error code, with the format `error_code: error_message`.
So not sure it's breaking (for now). What's true is that the error handling of the provider in general must be reworked so that they all work in a predictable way that can then be used for different purpose (webhooks, api responses, logs...)

PR #4973, opened 2026-02-03
feat(multiple-payment-methods): add initial logic for GCL with multiple payments

lovrocolic:
## Context

Currently customers can be associated with only one payment method

## Description

This PR introduces minimal changes that are needed to enable `multiple payment method` feature to gocardless customers.

### DIFF
--- a/app/services/payment_providers/gocardless/webhooks/mandate_cancelled_service.rb
+++ b/app/services/payment_providers/gocardless/webhooks/mandate_cancelled_service.rb
@@ -17,6 +17,7 @@ module PaymentProviders
           payment_method = find_payment_method_by_mandate
 
           return result unless payment_method
+          return result unless payment_provider.organization.feature_flag_enabled?(:multiple_payment_methods)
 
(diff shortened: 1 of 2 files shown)


app/jobs/payment_providers/gocardless/handle_event_job.rb
```diff
@@ -11,17 +11,17 @@ class HandleEventJob < ApplicationJob
         end
       end
 
-      def perform(organization: nil, events_json: nil, event_json: nil)
+      def perform(payment_provider: nil, events_json: nil, event_json: nil)
```
aquinofb: I think this change would need to be done in two steps: first add the `payment_provider` param, then another PR to remove the organization.

The reason is that right after deployment we still have jobs enqueued using the old param. When those are picked up to processing, it then will break.

lovrocolic: makes sense, will keep organization param

aquinofb (approved):
LGTM!

PR #4974, opened 2026-02-03
feat(customer): Add lock service

floganz:
## Context

We're going to add advisory lock on customer level to more and more places. To keep it unified and avoid duplication adding service that wraps given block in advisory lock.
Also adding new custom exception class (`Customers:: FailedToAcquireLock`) to distinguish from other lock exceptions.

## Description

Add lock service with custom exception if lock is not acquired.
Adjusted current places for relay on a new exception.

### DIFF
--- a/spec/services/customers/lock_service_spec.rb
+++ b/spec/services/customers/lock_service_spec.rb
@@ -23,35 +23,12 @@ RSpec.describe Customers::LockService do
     end
 
     context "when lock cannot be acquired", transaction: false do
-      let(:lock_released_after) { 2.seconds }
       let(:timeout_seconds) { 0.seconds }
 
       around do |test|
-        customer_id = customer.id
-        queue = Queue.new
-        thread = start_lock_thread(queue, customer_id)
-        sleep 0.5
-        test.run
-      ensure
-        stop_thread(thread, queue) if thread
-      end
-
-      def start_lock_thread(queue, customer_id)
-        Thread.start do
-          start_time = Time.zone.now
-          ApplicationRecord.transaction do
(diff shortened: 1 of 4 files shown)


app/services/customers/lock_service.rb
```diff
@@ -0,0 +1,35 @@
...
+      @customer = customer
+      @timeout_seconds = timeout_seconds
+      @transaction = transaction
+
+      super
+    end
+
+    def call
+      Customer.with_advisory_lock!(lock_key, timeout_seconds:, transaction:) do
+        yield
```
groyoh, line 17: Maybe allow block-less call ?

```suggestion
        if block_given?
          yield
        elsif !transaction
          raise ArgumentError, "you must pass a block when transaction is `false`"
        end

File: spec/services/customers/lock_service_spec.rb [hunk 2]
```diff
@@ -0,0 +1,82 @@
...
+        customer_id = customer.id
+        queue = Queue.new
+        thread = start_lock_thread(queue, customer_id)
+        sleep 0.5
+        test.run
+      ensure
+        stop_thread(thread, queue) if thread
+      end
+
+      def start_lock_thread(queue, customer_id)
```
groyoh: Would be nice to make this re-usable rather than copy-paste.

floganz: I've made some changes to it, do you think it's good enough for now? @groyoh

app/services/customers/lock_service.rb
floganz: I've spent a some time and didn't find a way to properly test this change.
I'll skip it for now.

PR #4975, opened 2026-02-03
feat(security logs implementation)

nepalez:
## Context

This PR completes the Security Logs infrastructure implementation (Step 3 from release strategy). It follows the interface-only PR that unblocked frontend and service integrations.

Security Logs track security-critical operations: user management, role changes, API key rotations, and webhook configuration. The implementation follows the Activity Logs architecture (Service → Kafka → ClickHouse queue → MV → table).

## Description

**ClickHouse Storage**

Three migrations create the data pipeline:
- `security_logs` table (ReplacingMergeTree) for permanent storage
- `security_logs_queue` (Kafka engine) consuming from `LAGO_KAFKA_SECURITY_LOGS_TOPIC`
- `security_logs_mv` (Materialized View) inserting from queue to table

**Model & Query**

`Clickhouse::SecurityLog` model with `LOG_TYPES` and `LOG_EVENTS` constants. GraphQL enums are generated from these constants. `SecurityLogsQuery` filters by date range, user_ids, log_types, and log_events. Retention period (90 days) is enforced via `default_scope`.

**Service Implementation**

`Utils::SecurityLog.produce` sends events to Kafka asynchronously. It checks infrastructure availability (`LAGO_CLICKHOUSE_ENABLED`, `LAGO_KAFKA_BOOTSTRAP_SERVERS`, `LAGO_KAFKA_SECURITY_LOGS_TOPIC`) and organization's `security_logs_enabled?` before producing.

**REST API**

- `GET /api/v1/security_logs` — paginated list with filters
- `GET /api/v1/security_logs/:log_id` — single log

Controller uses `resource_name "security_log"` for API key scope authorization and `skip_audit_logs!` to avoid noise.

**GraphQL API**

Resolvers now return actual data from ClickHouse instead of empty results. The `toDatetime` argument anchors pagination for consistency.

**CI**

`LAGO_KAFKA_SECURITY_LOGS_TOPIC` environment variable added to spec.yml.

### DIFF
--- a/db/seeds/01_base.rb
+++ b/db/seeds/01_base.rb
@@ -23,9 +23,12 @@ Role.find_or_create_by!(code: "manager", organization_id: nil) do |role|
   role.description = "The predefined manager role"
 end
 
-# NOTE: create a user and an organization
+# NOTE: create users and an organization
 user = User.create_with(password: "ILoveLago")
   .find_or_create_by(email: "gavin@hooli.com")
+
+dinesh = User.create_with(password: "ILoveLago")
+  .find_or_create_by(email: "dinesh@hooli.com")
 
 organizations_data = [
   {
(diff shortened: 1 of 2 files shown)


db/clickhouse_migrate/20260202135506_create_security_logs.rb
```diff
@@ -0,0 +1,27 @@
...
+    create_table :security_logs, id: false, options: do |t|
+      t.string :organization_id, null: false
+      t.string :user_id
+      t.string :api_key_id
+
+      t.string :log_id, null: false
+      t.string :log_type, null: false
+      t.string :log_event, null: false
+
+      t.string :device_info, map: true
```
vincent-pochet, line 20: Depending on what you want to store, Clickhouse now has a `json` type.

db/clickhouse_migrate/20260202135507_create_security_logs_queue.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.string :organization_id, null: false
+      t.string :user_id
+      t.string :api_key_id
+
+      t.string :log_id, null: false
+      t.string :log_type, null: false
+      t.string :log_event, null: false
+
+      t.string :device_info, map: true
+      t.string :resources, map: true
```
vincent-pochet, line 24: Not sure how clickhouse behave with this. Usualy, the queue should just consume "string", and the materialized view should be responsible from converting it into a proper map or json

app/services/utils/security_log.rb
```diff
@@ -39,7 +32,71 @@ def self.produce(
...
+      log_type:,
+      log_event:,
+      user: nil,
+      api_key: nil,
+      resources: nil,
+      device_info: nil,
+      skip_organization_check: false
+    )
+      @organization = organization
+      @skip_organization_check = skip_organization_check
```
vincent-pochet, line 53: What's the purpose of this flag?

nepalez: All security logs are available to premium subscribers only, EXCEPT for the unique case of `users.sign-up` because during signing up the organization is created, it cannot have a premium subscription yet. The flag is used to skip the organization primary subscription for that case (or probably for some others where we'd better record security logs unconditionally, but display to premium users only)

db/clickhouse_migrate/20260202135506_create_security_logs.rb
nepalez: Yes, but a part of the specification required the same mechanics for security logs as for the others, that's why I just followed the same approach

db/clickhouse_migrate/20260202135507_create_security_logs_queue.rb
nepalez: I'd answer in 2 parts
* as for the clickhouse behavior, this approach of mimicking the view in a queue definitely should work well (and works for the other models we use)
* but the fair answer is why I'd prefer going that way -- because I'd better fail fast during the queue insertion in case the string cannot be converted to a proper structure, than keep it as is and then fail during the materialization

PR #4977, opened 2026-02-03
feat(wallet): link re-credited wallet transaction to voided invoice

groyoh:
## Context

| --- | ---------------------------------------------- | ------------------------------------------------------------------------- |

## Description

This implements step #3 of the above breakdown: Voiding invoice.

This adds a `voided_invoice_id` column to `wallet_transactions` to explicitly link inbound transactions created when an invoice is voided. When an invoice is voided, the `RecreditService` creates a new inbound wallet transaction to restore credits. This field links back to the voided invoice for full traceability.

This approach avoids breaking API changes while providing explicit semantics. The field is exposed in both REST API (`lago_voided_invoice_id`) and GraphQL (`voidedInvoice`).

### DIFF
--- a/app/graphql/types/wallet_transactions/object.rb
+++ b/app/graphql/types/wallet_transactions/object.rb
@@ -27,6 +27,7 @@ module Types
       field :remaining_credit_amount, String, null: true
       field :settled_at, GraphQL::Types::ISO8601DateTime, null: true
       field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
+      field :voided_invoice, Types::Invoices::Object, null: true
 
       field :selected_invoice_custom_sections, [Types::InvoiceCustomSections::Object], null: true
       field :skip_invoice_custom_sections, Boolean
@@ -42,6 +43,8 @@ module Types
         currency = wallet.currency_for_balance
         object.remaining_amount_cents.fdiv(currency.subunit_to_unit).fdiv(wallet.rate_amount).to_s
       end
+
+      delegate :voided_invoice, to: :object
 
       def wallet_name
         object.wallet.name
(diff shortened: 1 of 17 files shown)


app/graphql/types/wallet_transactions/object.rb
```diff
@@ -33,6 +34,10 @@ def invoice
         object.invoice&.visible? ? object.invoice : nil
       end
 
+      def voided_invoice
+        object.voided_invoice&.visible? ? object.voided_invoice : nil
```
groyoh: I'm not sure whether the visible check actually makes sense here since we shouldn't be able to change the status once voided.

spec/serializers/v1/wallet_transaction_serializer_spec.rb
```diff
@@ -19,6 +19,7 @@
         "lago_wallet_id" => wallet_transaction.wallet_id,
         "lago_invoice_id" => nil,
         "lago_credit_note_id" => nil,
+        "lago_voided_invoice_id" => nil,
```
floganz, line 22: Currently this test checks that we have such fields but we do not check if it ever filled with right data.
Should we maybe add another test to ensure that it has proper data?

spec/graphql/types/wallet_transactions/object_spec.rb
```diff
@@ -22,6 +22,7 @@
     expect(subject).to have_field(:created_at).of_type("ISO8601DateTime!")
     expect(subject).to have_field(:failed_at).of_type("ISO8601DateTime")
     expect(subject).to have_field(:invoice).of_type("Invoice")
+    expect(subject).to have_field(:voided_invoice).of_type("Invoice")
```
floganz, line 27: This fields in GQL has some logic related to invoice visibility which could be tested.

spec/serializers/v1/wallet_transaction_serializer_spec.rb
groyoh: I added proper test 👍

spec/graphql/types/wallet_transactions/object_spec.rb
groyoh: I removed the visibility logic. See https://github.com/getlago/lago-api/pull/4977#discussion_r2763151920.

brunomiguelpinto (approved):
Lgtm

PR #4978, opened 2026-02-03
feat(wallet): Limit credit notes on prepaid credit invoices to transaction remaining amount

groyoh:
## Context

| --- | ---------------------------------------------- | ------------------------------------------------------------------------- |

## Description

This implements step #4 of the above breakdown: Credit Notes on Prepaid Credit invoices.

When issuing a credit note refund on a prepaid credit invoice with a traceable wallet, the system now validates against the specific inbound transaction's remaining amount (not the total wallet balance) and consumes from that specific transaction.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -5559,6 +5559,7 @@ Organization Feature Flag Values
 enum FeatureFlagEnum {
   multiple_payment_methods
   postgres_enriched_events
+  wallet_traceability
 }
 
 type FeatureObject {
--- a/schema.json
+++ b/schema.json
@@ -26750,6 +26750,12 @@
               "description": null,
               "isDeprecated": false,
               "deprecationReason": null
+            },
+            {
+              "name": "wallet_traceability",
+              "description": null,
+              "isDeprecated": false,
+              "deprecationReason": null
             }
           ],
           "possibleTypes": null


app/models/invoice.rb
```diff
@@ -380,7 +380,7 @@ def refundable_amount_cents
   def associated_active_wallet
     return if !credit? || customer.wallets.active.empty?
 
-    wallet = fees.credit.first&.invoiceable&.wallet
+    wallet = fees.first&.invoiceable&.wallet
```
aquinofb: 💭 I see we’re using `invoice.fees.first` in quite a few places, including when fetching wallet transactions (`invoice.fees.first.invoiceable`).

Would it make sense to encapsulate this inside the Invoice class? For example, exposing something like `invoice.prepaid_credit_fee` instead of relying on `invoice.fees.first`.

It would improve semantics and avoid leaking assumptions about ordering. That would also protect us from, for example, if fees.first is assumed to always be the prepaid credit fee, that’s a hidden invariant. If that invariant changes, subtle bugs will appear everywhere.

Another benefit would be removing the need for safe navigation when accessing the wallet, e.g.:
```ruby
invoice.fees.first&.invoiceable
```
to something like:
```ruby
invoice.prepaid_wallet
#=> Wallet or nil
```
changes required:
```ruby
# fee.rb
delegate :wallet, to: :invoiceable, allow_nil: true

# invoice.rb
delegate :wallet, to: :prepaid_credit_fee, prefix: :prepaid, allow_nil: true
```

app/services/credit_notes/create_service.rb
```diff
@@ -292,14 +286,23 @@ def all_rounding_tax_adjustments
...
-      @associated_wallet ||= invoice.associated_active_wallet
+    def prepaid_credit_wallet
+      @prepaid_credit_wallet ||= invoice.associated_active_wallet
     end
 
-    def wallet_credit
-      return nil unless invoice.credit? && associated_wallet.present?
+    def void_prepaid_credit
+      wallet_credit = WalletCredit.from_amount_cents(wallet: prepaid_credit_wallet, amount_cents: credit_note.refund_amount_cents)
+      wallet_transaction = prepaid_credit_wallet.traceable? ? invoice.fees.first.invoiceable : nil
```
D1353L: If the `fees` array is empty
```suggestion
      wallet_transaction = prepaid_credit_wallet.traceable? ? invoice.fees.first&.invoiceable : nil
```

app/models/fee.rb
```diff
@@ -216,10 +216,27 @@ def offsettable_amount_cents
...
+
     remaining_amount
   end
 
+  def creditable_from_wallet_amount_cents
+    return 0 unless credit?
+
+    wallet_transaction = invoiceable
+    wallet = wallet_transaction.wallet
+    return 0 unless wallet.active?
```
brunomiguelpinto:
```suggestion
  return 0 unless wallet&.active?
```

app/models/invoice.rb
```diff
@@ -369,8 +369,8 @@ def refundable_amount_cents
     # so creditable_amount_cents is always 0 but on that case we should allow to issue a credit note
     # as refund only if the wallet balance is greater or equal than the remaining paid amount
     if credit?
-      wallet_balance = associated_active_wallet&.balance_cents || 0
-      return [wallet_balance, remaining_paid_cents].min
+      prepaid_credit_fee = fees.first
```
brunomiguelpinto: add a coment to be easy to understand why we do this 
Credit invoices have a single credit-type fee

groyoh: Done.

app/models/invoice.rb
groyoh: Done.

PR #4979, opened 2026-02-03
Feat(wallet-targets): application

annvelents:
Final PR of  wallet-targets feature introduces 
- finding the applicable wallet on the fee by grouped_by
- adds some refactoring about wallet refresh
- and some optimisation to reduce n+1 queries

### DIFF
--- a/spec/services/wallets/find_applicable_on_fees_service_spec.rb
+++ b/spec/services/wallets/find_applicable_on_fees_service_spec.rb
@@ -4,9 +4,10 @@ require "rails_helper"
 
 RSpec.describe Wallets::FindApplicableOnFeesService do
   describe ".call" do
-    subject(:result) { described_class.call(allocation_rules:, fee:, fee_targeting_wallets_enabled:) }
+    subject(:result) { described_class.call(allocation_rules:, fee:, customer_id:, fee_targeting_wallets_enabled:) }
 
     let(:fee_targeting_wallets_enabled) { nil }
+    let(:customer_id) { nil }
 
     context "when there are applicable wallets for billable metrics, fee types and unrestricted" do
       let(:allocation_rules) do
@@ -151,6 +152,7 @@ RSpec.describe Wallets::FindApplicableOnFeesService do
       let(:wallet) { create(:wallet, customer:, code: "target_wallet") }
       let(:charge) { create(:standard_charge, organization:, accepts_target_wallet:) }
       let(:accepts_target_wallet) { nil }
+      let(:customer_id) { customer.id }
       let(:fee) do
         create(:charge_fee, invoice:, subscription:, charge:,
           grouped_by: {"target_wallet_code" => "target_wallet"})


app/services/wallets/find_applicable_on_fees_service.rb
```diff
@@ -11,6 +11,15 @@ def initialize(allocation_rules:, fee:)
     end
 
     def call
+      # Priority 1: Check for target_wallet_code in fee.grouped_by
+      if fee.organization.events_targeting_wallets_enabled? && fee.charge&.accepts_target_wallet
```
annvelents: we can send inside organization and charge to have less queries when we're speaking of thousands of fees

app/services/wallets/find_applicable_on_fees_service.rb
```diff
@@ -29,6 +38,13 @@ def call
 
     attr_reader :allocation_rules, :fee
 
+    def find_wallet_by_code(code)
+      wallet_customer_id = fee.subscription&.customer_id || fee.invoice&.customer_id
```
annvelents: can we send customer_id in this service?

app/services/wallets/find_applicable_on_fees_service.rb
```diff
@@ -11,6 +11,15 @@ def initialize(allocation_rules:, fee:)
     end
 
     def call
+      # Priority 1: Check for target_wallet_code in fee.grouped_by
+      if fee.organization.events_targeting_wallets_enabled? && fee.charge&.accepts_target_wallet
+        target_wallet_code = fee.grouped_by&.dig("target_wallet_code")
+        if target_wallet_code.present?
+          targeted_wallet = find_wallet_by_code(target_wallet_code)
+          return result_with([targeted_wallet.id]) if targeted_wallet
```
annvelents, line 21: `.present?`, because if not found, it's an empty relation that is "true"

app/services/fees/apply_taxes_service.rb
```diff
@@ -68,7 +70,8 @@ def applicable_taxes
       return fee.charge.taxes if fee.charge? && fee.charge.taxes.any?
       return fee.fixed_charge.taxes if fee.fixed_charge? && fee.fixed_charge.taxes.any?
       return fee.invoiceable.taxes if fee.commitment? && fee.invoiceable.taxes.any?
-      if (fee.charge? || fee.subscription? || fee.commitment? || fee.fixed_charge?) && fee.subscription.plan.taxes.any?
+      plan = @plan || fee.subscription&.plan
```
floganz: Can we move this to initialize and then define attr_reader to it and customer.
```@plan = @plan || fee.subscription&.plan```

File: app/services/fees/apply_taxes_service.rb [hunk 5]
```diff
@@ -2,9 +2,11 @@
 
 module Fees
   class ApplyTaxesService < BaseService
-    def initialize(fee:, tax_codes: nil)
+    def initialize(fee:, tax_codes: nil, customer: nil, plan: nil)
       @fee = fee
       @tax_codes = tax_codes
+      @customer = customer
```
floganz: If you want to define customer in the initialize I guess you need to remove lazy loaded method below in this file. And also define attr_reader for customer.
```
def customer
   @customer ||= fee.invoice&.customer || fee.subscription.customer
end
```

app/services/invoices/customer_usage_service.rb
```diff
@@ -151,9 +151,10 @@ def boundaries
 
     def compute_amounts
       invoice.fees_amount_cents = invoice.fees.sum(&:amount_cents)
+      plan = subscription.plan
 
       invoice.fees.each do |fee|
-        taxes_result = Fees::ApplyTaxesService.call(fee:)
+        taxes_result = Fees::ApplyTaxesService.call(fee:, customer: customer, plan: plan)
```
floganz: Minor but you can omit passing variable if the name matcher keyword argument name.
```customer:, plan:)```

File: app/services/wallets/balance/update_ongoing_service.rb [hunk 7]
```diff
@@ -3,11 +3,11 @@
 module Wallets
   module Balance
     class UpdateOngoingService < BaseService
-      def initialize(wallet:, update_params:)
+      def initialize(wallet:, update_params:, skip_single_wallet_update: false)
         super
 
         @wallet = wallet
-        update_params[:last_ongoing_balance_sync_at] = Time.current
+        update_params[:last_ongoing_balance_sync_at] = Time.current unless skip_single_wallet_update
```
floganz, line 10: Is there any test for this change?

app/services/wallets/find_applicable_on_fees_service.rb
annvelents: nope, all good here, because we do `.where(...).find_by(code:)`, which returns nil if not found

app/services/customers/refresh_wallets_service.rb
```diff
@@ -26,15 +26,39 @@ def call
...
+            usage_amount_cents:,
+            skip_single_wallet_update: true,
+            current_usage_fees: applicable_fees(usage_fees, wallets_applicable_on_usage_fees, wallet),
+            draft_invoices_fees: applicable_fees(draft_invoice_fees, wallets_applicable_on_draft_fees, wallet),
+            progressive_billing_fees: applicable_fees(progressive_billing_fees, wallets_applicable_on_pb_fees, wallet),
+            pay_in_advance_fees: applicable_fees(pay_in_advance_fees, wallets_applicable_on_adv_fees, wallet)
+          )
+        end
       end
+      wallets_to_process.presence&.update_all(last_ongoing_balance_sync_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
```
floganz: Do we really need `.presence&`? It looks like in worth case it would be empty array.

app/services/wallets/find_applicable_on_fees_service.rb
```diff
@@ -27,7 +38,14 @@ def call
...
     private
 
-    attr_reader :allocation_rules, :fee
+    attr_reader :allocation_rules, :fee, :fee_targeting_wallets_enabled
+
+    def find_wallet_by_code(code)
+      wallet_customer_id = @customer_id || fee.subscription&.customer_id || fee.invoice&.customer_id
+      return nil unless wallet_customer_id
+
+      Wallet.active.where(organization_id: fee.organization_id, customer_id: wallet_customer_id).find_by(code:)
```
floganz: We can combine `where` and `find_by` into `find_by`.

spec/scenarios/events_targeting_wallets_spec.rb
```diff
@@ -68,6 +68,15 @@
           external_subscription_id: subscription.external_id,
           properties: {value: "20", target_wallet_code: "wallet_2"}
         })
+
+        # Refresh wallets to update ongoing balance
+        Customers::RefreshWalletsService.call(customer:)
```
floganz: Better to use `recalculate_wallet_balances` helper method rather than direct call of refresh service. Because in real app use customers should be marked to be refreshed and picked up by job. Calling service directly does not cover those cases and could lead to some wrong behaviour.
I see that it used several times in new tests.

app/services/wallets/balance/update_ongoing_service.rb
annvelents: added, thank you! 🙌

app/services/customers/refresh_wallets_service.rb
```diff
@@ -48,6 +72,32 @@ def call
...
+    attr_reader :customer, :include_generating_invoices, :allocation_rules
+
+    def assign_wallet_per_fee(fees)
+      fee_wallet = {}
+      fee_targeting_wallets_enabled = customer.organization.events_targeting_wallets_enabled?
+
+      fees.each do |fee|
+        key = fee.id || fee.object_id
+
+        if fee_targeting_wallets_enabled && fee.charge.accepts_target_wallet && fee.grouped_by&.dig("target_wallet_code").present?
```
brunomiguelpinto: fee.charge&.accepts_target_wallet ?

app/services/wallets/balance/refresh_ongoing_usage_service.rb
```diff
@@ -21,58 +25,32 @@ def call
...
+        billed_progressive_invoices_amount_cents +
+          billed_pay_in_advance_amount_cents
       end
 
-      def billed_progressive_invoices_amount_cents(invoice_subscriptions)
-        fees = progressive_billing_fees(invoice_subscriptions)
+      def billed_progressive_invoices_amount_cents
+        fees = progressive_billing_fees
 
         fees.sum do |fee|
```
vincent-pochet:
```suggestion
        progressive_billing_fees.sum do |fee|
```

(Not that it changes anything to the logic itself 😛)

app/services/wallets/balance/refresh_ongoing_usage_service.rb
```diff
@@ -81,13 +59,17 @@ def draft_invoices_total_amount_cents
         end
       end
 
-      def billed_pay_in_advance_amount_cents(invoice)
-        paid_in_advance_fees = invoice_pay_in_advance_fees(invoice)
+      def billed_pay_in_advance_amount_cents
+        paid_in_advance_fees = pay_in_advance_fees
         # Invoice that is returned from CustomerUsageService includes the taxes in total_usage
         # so if the fees ae already paid, we should exclude fees AND their taxes
         paid_in_advance_fees.sum { |fee| fee.amount_cents + fee.taxes_amount_cents }
```
vincent-pochet:
```suggestion
        # Invoice that is returned from CustomerUsageService includes the taxes in total_usage
        # so if the fees ae already paid, we should exclude fees AND their taxes
        pay_in_advance_fees.sum { |fee| fee.amount_cents + fee.taxes_amount_cents }
```

app/services/wallets/find_applicable_on_fees_service.rb
```diff
@@ -27,7 +38,14 @@ def call
 
     private
 
-    attr_reader :allocation_rules, :fee
+    attr_reader :allocation_rules, :fee, :fee_targeting_wallets_enabled
+
+    def find_wallet_by_code(code)
+      wallet_customer_id = @customer_id || fee.subscription&.customer_id || fee.invoice&.customer_id
```
vincent-pochet: Just a though, should we force the `customer_id` or even `customer` argument to be passed to the service to completely remove this "let's guess where I can find a customer_id" operation?

app/services/customers/refresh_wallets_service.rb
```diff
@@ -48,6 +68,32 @@ def call
...
 
-    attr_reader :customer, :include_generating_invoices
+    attr_reader :customer, :include_generating_invoices, :allocation_rules
+
+    def assign_wallet_per_fee(fees)
+      fee_wallet = {}
+      fee_targeting_wallets_enabled = customer.organization.events_targeting_wallets_enabled?
+
+      fees.each do |fee|
+        key = fee.id || fee.object_id
```
vincent-pochet: It appears (I just discovered it) that we have a `fee#item_key` that is doing exactly this. Maybe we could reuse it (not at all a blocker)

app/services/customers/refresh_wallets_service.rb
```diff
@@ -26,15 +26,35 @@ def call
...
+      draft_invoice_fees = customer.invoices.draft.where.not(total_amount_cents: 0).includes(fees: :charge).flat_map(&:fees)
+      wallets_applicable_on_draft_fees = assign_wallet_per_fee(draft_invoice_fees)
+
+      progressive_billing_fees = usage_amount_cents.flat_map { |usage| usage[:billed_progressive_invoice_subscriptions].flat_map { it.invoice.fees } }
+      wallets_applicable_on_pb_fees = assign_wallet_per_fee(progressive_billing_fees)
+
+      pay_in_advance_fees = usage_amount_cents.flat_map { |usage| usage[:invoice].fees.select { |f| f.charge.pay_in_advance? } }
+      wallets_applicable_on_adv_fees = assign_wallet_per_fee(pay_in_advance_fees)
+
+      wallets_to_process = customer.wallets.active.includes(:recurring_transaction_rules)
```
vincent-pochet, line 45: A lot is happening here, can you maybe extract it in sub methods with clear name, I feel that it would ease the understanding of what is happening.

app/services/wallets/balance/refresh_ongoing_usage_service.rb
annvelents: it was kinda respect to the previous logic 😅

app/services/wallets/find_applicable_on_fees_service.rb
annvelents: hah, actually IT IS being called from only one place 🤦‍♀️

app/services/customers/refresh_wallets_service.rb
annvelents: TIL! 😍

app/services/customers/refresh_wallets_service.rb
annvelents: I think I found even better way to refactor it!

annvelents: no, I was wrong. Refactored approach makes us go through all fees per each wallet.
In case of 1K wallets and 6K fees it's not survivable.
So, unfortunately we need to define all the fees, and then wallet_application_map
So later when going through wallets, we would be able to select only applicable fees without going through ALL of them again

PR #4981, opened 2026-02-03
feat(webhook): flexible webhook events

rinasergeeva:
## Context

This PR introduces the ability for users to selectively subscribe to webhook endpoints for specific event types. Currently, all webhook endpoints receive all events fired by Lago. With this feature, users gain granular control over which events trigger notifications to each endpoint, reducing noise and enabling more efficient integrations.

## Description

The new `event_types` attribute allows users to specify the event types the endpoint should receive. If left undefined, all events will be fired to the endpoint.

#5083 was also merged into this PR.

### DB migrations

Two new columns were added to the `webhook_endpoints` table.

| Column | Type | Nullable | Default value
|-------|---------|---------|---------|
| `event_types` | string[] | true | null |
| `name` | string | true | null |

### REST API changes

Two new attributes (`event_types` and `name` ) were added to the `WebhookEndpoint` object.

Affected endpoints:
* GET `/v1//webhook_endpoints`
* POST `/v1/webhook_endpoints`
* GET `/v1/webhook_endpoints/{lago_id}`
* PUT `/v1/webhook_endpoints/{lago_id}`
* DELETE `/v1/webhook_endpoints/{lago_id}`

### GraphQL changes

* Two new attributes (`event_types` and `name` ) were added to the `WebhookEndpoint` object.
* A new query `{ eventTypes }` has been added to provide a complete list of supported event types.

### DIFF
--- a/app/models/webhook_endpoint.rb
+++ b/app/models/webhook_endpoint.rb
@@ -45,7 +45,7 @@ class WebhookEndpoint < ApplicationRecord
 
     # since AR casts non-array values to [] we need to check the raw value
     if event_types.is_a?(Array) && event_types.blank? && !event_types_before_type_cast.is_a?(Array)
-      errors.add(:event_types, :invalid_format)
+      errors.add(:event_types, :must_be_array)
     end
 
     invalid_types = event_types - WEBHOOK_EVENT_TYPES
--- a/config/locales/de/webhook_endpoint.yml
+++ b/config/locales/de/webhook_endpoint.yml
@@ -7,6 +7,3 @@ de:
           attributes:
             base:
               exceeded_limit: Maximum number of webhook endpoints was reached
-            event_types:
-              invalid_format: Must be an array
-              invalid_types: 'Contains invalid types: %{invalid_types}'
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -41,12 +41,8 @@ en:
         less_than_or_equal_to: value_is_out_of_range
         missing_graduated_ranges: missing_graduated_ranges
         missing_volume_ranges: missing_volume_ranges
-        models:
-          api_key:
-            attributes:
-              permissions:
-                forbidden_key: 'contains forbidden keys: %{keys}'
         modification_not_allowed: modification_not_allowed
+        must_be_array: must_be_array
         must_be_greater_than_or_equal_min: must_be_greater_than_or_equal_min
         must_be_true_unless_pay_in_advance: must_be_true_unless_pay_in_advance
         not_compatible_with_aggregation_type: not_compatible_with_aggregation_type
@@ -61,6 +57,15 @@ en:
         unsupported_value: unsupported_value
         url_invalid: url_is_invalid
         value_already_exist: value_already_exist
+      models:
+        api_key:
+          attributes:
+            permissions:
+              forbidden_keys: 'contains forbidden keys: %{keys}'
+        webhook_endpoint:
+          attributes:
+            event_types:
+              invalid_types: 'contains invalid types: %{invalid_types}'
   date:
     formats:
       default: "%b %d, %Y"
--- a/config/locales/en/webhook_endpoint.yml
+++ b/config/locales/en/webhook_endpoint.yml
@@ -7,6 +7,3 @@ en:
           attributes:
             base:
               exceeded_limit: Maximum number of webhook endpoints was reached
-            event_types:
-              invalid_format: Must be an array
-              invalid_types: 'Contains invalid types: %{invalid_types}'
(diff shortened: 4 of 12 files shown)


app/services/webhooks/base_service.rb
```diff
@@ -23,6 +23,9 @@ def call
       # TODO: Wrap in transaction so we create all webhook models or none
       #       Ensure the http jobs are dispatched after the transaction is committed
       current_organization.webhook_endpoints.each do |webhook_endpoint|
+        if webhook_endpoint.event_types.present?
```
murparreira: Good 👍🏽

app/services/webhooks/base_service.rb
```diff
@@ -23,6 +23,9 @@ def call
       # TODO: Wrap in transaction so we create all webhook models or none
       #       Ensure the http jobs are dispatched after the transaction is committed
       current_organization.webhook_endpoints.each do |webhook_endpoint|
+        unless webhook_endpoint.event_types.nil?
+          next unless webhook_endpoint.event_types.include?(webhook_type)
+        end
```
julienbourdeau: Not too important but I'd extract this to a method for clarity
```suggestion
        next unless webhook_endpoint.subscribed_to?(webhook_type)
```

julienbourdeau (approved):

app/services/webhooks/base_service.rb
rinasergeeva: Thought about this too, but wasn't sure since it's literally just one line. Let's do this :)

rinasergeeva: Done!

config/initializers/webhook_endpoints.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+WEBHOOK_EVENT_TYPES = YAML.load_file(Rails.root.join("config/webhook_event_types.yml")).deep_symbolize_keys
```
vincent-pochet: Should we define this in the Webhook or WebhookEnpoint? It would scope the constant to the right namespace

config/locales/de/webhook_endpoint.yml
```diff
@@ -7,3 +7,6 @@ de:
           attributes:
             base:
               exceeded_limit: Maximum number of webhook endpoints was reached
+            event_types:
```
vincent-pochet: WE are usually returning error codes rather than plain text message. The reason is to avoid managing multiple translation for this, I get the need for `invalid_types` but I'm not sure we should change the current behavior :/

app/graphql/types/webhook_endpoints/create_input.rb
```diff
@@ -5,6 +5,8 @@ module WebhookEndpoints
     class CreateInput < BaseInputObject
       graphql_name "WebhookEndpointCreateInput"
 
+      argument :event_types, [String], required: false
```
vincent-pochet: Could we define an enum type for the `event_type`? It would auto-validate the input

app/models/webhook_endpoint.rb
```diff
@@ -14,19 +14,56 @@ class WebhookEndpoint < ApplicationRecord
...
+  before_validation :normalize_event_types, if: :event_types_changed?
 
   enum :signature_algo, SIGNATURE_ALGOS
 
   def self.ransackable_attributes(_auth_object = nil)
     %w[webhook_url]
   end
 
+  def self.event_types
+    SendWebhookJob::WEBHOOK_SERVICES.keys.map(&:to_s)
```
vincent-pochet: I'm wondering if we should turn it in a constant. What do you think?

config/initializers/webhook_endpoints.rb
rinasergeeva: Moved to `WebhookEnpoint`.

app/models/webhook_endpoint.rb
rinasergeeva: Done!

app/graphql/types/webhook_endpoints/create_input.rb
rinasergeeva: We could do that, but it will need an inelegant hack to support `"*"` value along with the enum values.

rinasergeeva: Switched to the enum!

config/locales/de/webhook_endpoint.yml
rinasergeeva: Not sure we have other options if we want to provide a meaningful message pointing to incorrect types.

rinasergeeva: @vincent-pochet If you want to avoid expanding translations, I can hardcode the error message into the model. Would it be better?

rinasergeeva: UPD:
* I've moved translations to the main `en.yml` file; all dedicated `webhook_endpoint.yml` files remain unchanged
* One error was converted into code
* The other I keep as text to preserve the list of invalid event types in the error message

PR #4983, opened 2026-02-04
fix(subscriptions): Add amount_cents and amount_currency to subscriptions list

D1353L:
## Context
Customer requests the `amount_cents` field for the subscriptions list endpoint. Currently, the only other way they can get custom plan amounts is by retrieving each subscription, which takes a while.

## Description

This PR adds `plan_amount_cents` and `plan_amount_currency` to the subscription payload. There is no performance impact as `model.plan` is already preloaded and used in other fields.


vincent-pochet (approved):
Could you update the specs as well 🙏 
All good otherwise!

PR #4985, opened 2026-02-04
fix(wallet): Add validation to prevent negative balance and credits balance

floganz:
## Context

Surprisingly we didn't had validation on a wallet that ensures that `balance_cents` or `credits_balance` cannot be negative. In certain cases that could lead to consuming amounts bigger than wallet balances.

## Description

Add rails validation and DB constrain to validate that those amounts are bigger or equal to zero but only for traceable wallets.
Linking that validation to traceable wallets will allow us to rollout this checks now and as main topic wallet traceability goes  live all traceable wallets will have it in place. Leaving some non traceable wallet behind but as part of wallet traceability those wallets should be gone.

### DIFF
--- a/app/models/wallet.rb
+++ b/app/models/wallet.rb
@@ -43,6 +43,8 @@ class Wallet < ApplicationRecord
   validates :paid_top_up_min_amount_cents, numericality: {greater_than: 0}, allow_nil: true
   validates :paid_top_up_max_amount_cents, numericality: {greater_than: 0}, allow_nil: true
   validates :priority, inclusion: {in: 1..LOWEST_PRIORITY}
+  validates :balance_cents, numericality: {greater_than_or_equal_to: 0}, if: :traceable?
+  validates :credits_balance, numericality: {greater_than_or_equal_to: 0}, if: :traceable?
   validate :paid_top_up_max_greater_than_or_equal_min
   validate :unique_code_per_customer, if: :code_changed?
 
--- a/spec/models/wallet_spec.rb
+++ b/spec/models/wallet_spec.rb
@@ -90,6 +90,132 @@ RSpec.describe Wallet do
       expect(subject).to be_valid
       expect(subject.errors).to be_empty
     end
+
+    describe "of balance cents" do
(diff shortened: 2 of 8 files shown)


app/models/wallet.rb
```diff
@@ -43,6 +43,8 @@ class Wallet < ApplicationRecord
   validates :paid_top_up_min_amount_cents, numericality: {greater_than: 0}, allow_nil: true
   validates :paid_top_up_max_amount_cents, numericality: {greater_than: 0}, allow_nil: true
   validates :priority, inclusion: {in: 1..LOWEST_PRIORITY}
+  validates :balance_cents, numericality: {greater_than_or_equal_to: 0}, if: :traceable?
+  validates :credits_balance, numericality: {greater_than_or_equal_to: 0}, if: :traceable?
```
groyoh, line 47: Given how credits are calculated today, there could be a case where balance is positive but credit balance is negative. So this could cause some issues. We should maybe cap the credit balance in the void and apply prepaid credit services so that it never goes below zero.

Then we'll need to fix the calculation later on.

spec/models/wallet_spec.rb
```diff
@@ -90,6 +90,132 @@
...
+          it "adds an error" do
+            expect(error).to be_present
+          end
+        end
+
+        context "when balance_cents is zero" do
+          let(:balance_cents) { 0 }
+
+          it "does not add an error" do
+            expect(error).not_to be_present
```
groyoh, line 115: I would suggest to test `expect(subject) to be_valid` to avoid false positive.

floganz: It should not be false positive since `error` is defined using errors API as `wallet.errors.where(:balance_cents, :greater_than_or_equal_to)`. This way we can be sure that model produced right error. `be_valid` or rather `not_to be_valid` has ambiguity since it will be true from any error which be not from the validation we're testing.

groyoh:
Closing as it was handled in https://github.com/getlago/lago-api/issues/5070.

PR #4987, opened 2026-02-05
feat(wallet): Replace retry logic with advisory lock

floganz:
## Context

Current retry logic for wallets can lead to issues when balance becomes negative. It's not safe to repeat withdrawal amounts from wallet when we apply credits. Available amount on a wallet might less because of concurrent updates and we need also to recalculate wallet transaction and other related records.
It'll be safer in cases when cannot safe wallet to rollback everything and try again whole generation.
That in it's turn might hurt our API/GQL customers, since current approach of optimistic locking allows to start concurrent updates quickly, also quickly returning error to one of them which won't be ideal behaviour.

Using advisory lock should work better to API/GQL customers, since it will lock early and wait for lock release, rather quickly failing. In worth scenarios we'll increase our response time for certain action but should avoid returning a lot of errors "try later".

## Description

Add advisory lock for all wallet calculation during applying prepaid credits.
Remove retry logic and and proper retry on for affected sidekiq jobs.

### DIFF
--- a/spec/jobs/plans/destroy_job_spec.rb
+++ b/spec/jobs/plans/destroy_job_spec.rb
@@ -65,27 +65,4 @@ RSpec.describe Plans::DestroyJob do
       end
     end
   end
-
-  describe "retry_on" do
-    [
-      [Customers::FailedToAcquireLock.new("customer-1-prepaid_credit"), 25],
-      [ActiveRecord::StaleObjectError.new("Attempted to update a stale object: Wallet."), 25]
-    ].each do |error, attempts|
-      error_class = error.class
-
-      context "when a #{error_class} error is raised" do
-        before do
-          allow(Plans::DestroyService).to receive(:call).and_raise(error)
-        end
-
-        it "raises a #{error_class.class.name} error and retries" do
-          assert_performed_jobs(attempts, only: [described_class]) do
-            expect do
-              described_class.perform_later(plan)
-            end.to raise_error(error_class)
-          end
-        end
-      end
-    end
-  end
 end


app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
```diff
@@ -26,43 +26,46 @@ def call
...
-          taxes_result = apply_provider_taxes
-          unless taxes_result.success?
-            tax_error = taxes_result.error.code
-            invoice.failed!
-            deliver_fee_webhooks
-            create_error_detail(tax_error)
-            Utils::ActivityLog.produce(invoice, "invoice.failed")
-            next
+        # We acquire a lock on the customer to prevent concurrent pay-in-advance invoice creation.
+        Customers::LockService.call(customer:) do
```
lovrocolic: Why is lock added on the invoice level? Don't we have it here?app/services/credits/applied_prepaid_credits_service.rb?

app/jobs/bill_subscription_job.rb
```diff
@@ -11,6 +11,7 @@ class BillSubscriptionJob < ApplicationJob
 
   unique :until_executed, on_conflict: :log, lock_ttl: 12.hours
 
+  retry_on Customers::FailedToAcquireLock, attempts: 25, wait: ->(_) { rand(0...16) }
```
groyoh: `Invoices::SubscriptionService` swallows errors:

```
# app/services/invoices/subscription_service.rb:98
rescue => e
  result.fail_with_error!(e)
```

We need to update it to re-raise the error.

app/jobs/lifetime_usages/recalculate_and_check_job.rb
```diff
@@ -10,6 +10,8 @@ class RecalculateAndCheckJob < ApplicationJob
       end
     end
 
+    retry_on Customers::FailedToAcquireLock, attempts: 25, wait: ->(_) { rand(0...16) }
```
groyoh: `Invoices::ProgressiveBillingService` swallows errors:
```
rescue => e
  result.fail_with_error!(e)
```

We need to update it to re-raise the error.

app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
groyoh: This is a left-over from our previous work on the pay-in-advance. What we did was adding a global customer lock on the invoice to fail early for pay-in-advance.

This is not needed anymore. @D1353L we also need to remove the lock on the pay-in-advance service.

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -23,57 +19,57 @@ def call
...
-        ordered_remaining_amounts = calculate_amounts_for_fees_by_type_and_bm
-        remaining_invoice_amount = invoice.total_amount_cents
-
-        wallets.each do |wallet|
-          wallet.reload
-          wallet_fee_transactions = []
-          wallet_targets_array = wallet.wallet_targets.map do |wt|
-            if wt&.billable_metric_id
-              ["charge", wt.billable_metric_id]
+        Customers::LockService.call(customer:) do
```
groyoh: Let's update the `Customers::LockService` to add a scope:

```suggestion
        Customers::LockService.call(customer:, scope: :prepaid_credit) do
```

Let's make sure any invalid scope raises `ArgumentError` to avoid any typo.

groyoh (commented):
We need to add tests to the different service to ensure that they do not swallow the `FailedToAcquireLock`.

app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
```diff
@@ -10,6 +10,7 @@ class PullTaxesAndApplyJob < ApplicationJob
       retry_on OpenSSL::SSL::SSLError, wait: :polynomially_longer, attempts: 6
       retry_on Net::ReadTimeout, wait: :polynomially_longer, attempts: 6
       retry_on Net::OpenTimeout, wait: :polynomially_longer, attempts: 6
+      retry_on Customers::FailedToAcquireLock, attempts: 25, wait: ->(_) { rand(0...16) }
```
groyoh: I believe we can use `:polynomially_longer` for non pay-in-advance jobs.

app/jobs/invoices/create_pay_in_advance_charge_job.rb
```diff
@@ -19,7 +19,7 @@ def self.retry_delay
 
     # We acquire a lock on the customer to prevent concurrent pay-in-advance invoice creation.
     # When it fails, it raises a Customers::FailedToAcquireLock error.
-    # It the lock succeeds but another job/request updates the wallet concurrently, it will raise a ActiveRecord::StaleObjectError error.
+    # If the lock succeeds but another job/request updates the wallet concurrently, it will raise a ActiveRecord::StaleObjectError error.
     retry_on Customers::FailedToAcquireLock, ActiveRecord::StaleObjectError, attempts: 25, wait: ->(_) { CreatePayInAdvanceChargeJob.retry_delay }
```
groyoh: Let's define constant for the `attempts` since it's shared between multiple jobs. For the delay random retry, I believe we can define a re-usable proc/module from on the `ApplicationJob` class.

app/jobs/plans/destroy_job.rb
```diff
@@ -4,6 +4,8 @@ module Plans
   class DestroyJob < ApplicationJob
     queue_as "default"
 
+    retry_on Customers::FailedToAcquireLock, attempts: 25, wait: ->(_) { rand(0...16) }
```
D1353L: @groyoh As the `DestroyJob` has a call to `BillSubscriptionJob` under the hood, there is a possibility of multiplied retries.
Both have `retry_on Customers::FailedToAcquireLock`. 
So if BillSubscriptionJob exhausts 25 retries, the DestroyJob starts executing its 25 retries. In the worst case 25*25= 625 retries total.
Do you think it's something we should be concerned about?

app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
D1353L: As far as I understand, the `polynomially_longer` makes sense when job calls external APIs like here in pull_taxes_and_apply_job (e.g. retries with exponential backoff to prevent reaching the request rate limit).
For other cases, a short retry delay would be better as the job gets to success/failure faster. What do you think?

app/jobs/plans/destroy_job.rb
groyoh: Given we rely on `perform_now` in this case, I believe there's no retry involved. Worth testing though.

app/jobs/invoices/provider_taxes/pull_taxes_and_apply_job.rb
groyoh: To be honest, I'm not sure which one makes more sense here. My main concern here is getting slammed by pay-in-advance events which prevent a recurring billing job to lock. I guess it should be fine either way.

app/jobs/plans/destroy_job.rb
D1353L: I did some tests and found there are no cascading retries with the `perform_now`. The parent job executed 1 time, and the child job retried 25 times.
The `perform_now` retries are enqueued to Sidekiq, and no exception bubbles up (also discussed in [this](https://github.com/rails/rails/issues/48281) thread)

spec/services/invoices/create_pay_in_advance_charge_service_spec.rb
```diff
@@ -311,61 +310,14 @@
...
-      context "when it fails to acquire the lock" do
-        let(:lock_released_after) { 2.seconds }
-
-        it "raises a Customers::FailedToAcquireLock error" do
-          expect { invoice_service.call }.to raise_error(Customers::FailedToAcquireLock)
-
-          expect(customer.invoices.count).to eq(0)
-        end
+        allow(described_class).to receive(:call)
+          .and_raise(Customers::FailedToAcquireLock.new("customer-123-prepaid_credit"))
```
groyoh: I believe this test is not properly testing the behaviour compared to what we had behaviour. Now the code is saying "raise this error and expect to have raised this error" which is not testing the behaviour.

app/jobs/plans/destroy_job.rb
groyoh: Is this still needed then if it doesn't bubble up ?

app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
groyoh: @D1353L I believe you forgot to remove this one.

groyoh (approved):

app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
D1353L: Removed in [this](https://github.com/getlago/lago-api/pull/4987/changes/cb6e7395e3fcf84ad34df957a68300c9f9926230) commit

app/jobs/plans/destroy_job.rb
D1353L: It's not reachable, so it's not needed. I've removed this

PR #4989, opened 2026-02-06
feat(filter): Add charge filter mutations

rsempe:
Add three new GraphQL mutations for managing charge filters at the subscription level:

- CreateSubscriptionChargeFilter: Creates a new filter on a subscription's charge override
- UpdateSubscriptionChargeFilter: Updates an existing filter's properties (invoice_display_name, properties)
- DestroySubscriptionChargeFilter: Soft deletes a filter from a subscription's charge override


app/services/subscriptions/concerns/charge_override_concern.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  module Concerns
+    module ChargeOverrideConcern
+      extend ActiveSupport::Concern
+      include PlanOverrideConcern
```
ancorcruz: do we need to include PlanOverride here as a depenency? shouldn't the base service include both instead?

app/graphql/mutations/subscriptions/create_charge_filter.rb
```diff
@@ -0,0 +1,34 @@
...
+      type Types::ChargeFilters::Object
+
+      def resolve(**args)
+        subscription = context[:current_user].subscriptions.find_by(id: args[:subscription_id])
+        charge = subscription&.plan&.charges&.find_by(code: args[:charge_code])
+
+        params = args.except(:subscription_id, :charge_code).to_h.deep_symbolize_keys
+        params[:properties] = params[:properties].to_h if params[:properties]
+
+        result = ::Subscriptions::CreateChargeFilterService.call(
```
ancorcruz: would you consider organize this code within its own namespace instead? `Subscriptions::ChargeFilters::CreateService`

julienbourdeau (approved):

app/services/subscriptions/concerns/charge_override_concern.rb
rsempe: good spot, thanks!

app/graphql/mutations/subscriptions/create_charge_filter.rb
rsempe: i agree, that's cleaner, doing the changes

PR #4992, opened 2026-02-06
fix(integrations): handle fixed_charge fees in accounting and tax sync

ancorcruz:
## Context

When an invoice or credit note contains a fixed_charge fee, syncing to accounting providers (NetSuite, Xero) fails with an "Invalid mapping" error because the item dispatch logic in the payload classes only handles `charge`, `add_on`, `credit`, `commitment`, and `subscription` fee types. The `fixed_charge` fee type was added to the `Fee` model but the integration payload logic was never updated to handle it. Tax provider payloads (Anrok, Avalara) also silently produce nil item codes for fixed_charge fees.

## Description

Add a `fixed_charge_item` method to the base payload that maps fixed_charge fees to their underlying add-on's integration mapping via `fee.fixed_charge_add_on`. Add a `fixed_charge?` branch to all nine fee type dispatch sites across accounting payloads (invoices and credit notes for NetSuite, Xero, Anrok) and tax payloads (invoices and credit notes for Anrok, Avalara). Since every fixed charge belongs to an add-on, the mapping reuses the existing AddOn lookup mechanism.

### DIFF
--- a/app/services/integrations/aggregator/taxes/credit_notes/payloads/anrok.rb
+++ b/app/services/integrations/aggregator/taxes/credit_notes/payloads/anrok.rb
@@ -43,6 +43,8 @@ module Integrations
                 billable_metric_item(fee)
               elsif fee.add_on_id.present?
                 add_on_item(fee)
+              elsif fee.fixed_charge?
+                fixed_charge_item(fee)
               elsif fee.commitment?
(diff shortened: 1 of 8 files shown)


annvelents (approved):
Super!

PR #4993, opened 2026-02-06
Feat(alerts): add batch actions on alerts

annvelents:
Right now alerts can only be created 1 by 1
if alerts depend on plan values (entitlements), when plan is updated, all subscriptions should get alerts updated
if there are a lot of subscriptions and several alerts, having a separate request per each alert per each subscription will take a lot of time and a lot of requests. 
to help our customers and our API to process it faster, we can have batch actions on alerts

### DIFF
--- a/app/services/usage_monitoring/alerts/create_batch_service.rb
+++ b/app/services/usage_monitoring/alerts/create_batch_service.rb
@@ -37,11 +37,7 @@ module UsageMonitoring
               else
                 error_details = {}
                 error_details[:params] = alert_params
-                error_details[:errors] = if create_result.error.respond_to?(:messages)
-                  create_result.error.messages
-                else
-                  create_result.error&.message
-                end
+                error_details[:errors] = create_result.error&.message
                 result.errors[index] = error_details
                 raise ActiveRecord::Rollback
               end
--- a/spec/requests/api/v1/subscriptions/alerts_controller_spec.rb
+++ b/spec/requests/api/v1/subscriptions/alerts_controller_spec.rb
@@ -424,14 +424,14 @@ RSpec.describe Api::V1::Subscriptions::AlertsController do
         errors = json[:error_details]
         alert_params = params[:alerts]
         expect(errors[:"1"][:params]).to eq(alert_params[1])
-        expect(errors[:"1"][:errors]).to eq(alert_type: ["invalid_type"])
+        expect(errors[:"1"][:errors]).to include("invalid_type")
 
         expect(errors[:"2"][:params]).to eq(alert_params[2])
-        expect(errors[:"2"][:errors]).to eq("billable_metric_not_found")
+        expect(errors[:"2"][:errors]).to include("billable_metric_not_found")
 
         expect(errors[:"3"][:params]).to eq(alert_params[3])
         # type is already taken
-        expect(errors[:"3"][:errors]).to eq({code: ["value_already_exist"]})
+        expect(errors[:"3"][:errors]).to include("value_already_exist")
       end
     end
   end
--- a/spec/services/usage_monitoring/alerts/create_batch_service_spec.rb
+++ b/spec/services/usage_monitoring/alerts/create_batch_service_spec.rb
@@ -93,7 +93,7 @@ RSpec.describe UsageMonitoring::Alerts::CreateBatchService do
         expect(result.alerts).to be_empty
         expect(result.error.messages).to have_key(1)
         expect(result.error.messages[1][:params]).to eq(alerts_params[1])
-        expect(result.error.messages[1][:errors][:alert_type]).to include("invalid_type")
+        expect(result.error.messages[1][:errors]).to include("invalid_type")
       end
     end
 
@@ -120,7 +120,7 @@ RSpec.describe UsageMonitoring::Alerts::CreateBatchService do
         expect(result).to be_failure
         expect(result.error.messages).to have_key(1)
         expect(result.error.messages[1][:params]).to eq(alerts_params[1])
-        expect(result.error.messages[1][:errors][:code]).to include("value_already_exist")
+        expect(result.error.messages[1][:errors]).to include("value_already_exist")
       end
     end
 
@@ -147,7 +147,7 @@ RSpec.describe UsageMonitoring::Alerts::CreateBatchService do
         expect(result).to be_failure
         expect(result.error.messages).to have_key(1)
         expect(result.error.messages[1][:params]).to eq(alerts_params[1])
-        expect(result.error.messages[1][:errors][:base]).to include("alert_already_exists")
+        expect(result.error.messages[1][:errors]).to include("alert_already_exists")
       end
     end
 
@@ -190,13 +190,13 @@ RSpec.describe UsageMonitoring::Alerts::CreateBatchService do
         expect(result).to be_failure
         expect(result.error.messages.size).to eq(4)
         expect(result.error.messages[0][:params]).to eq(alerts_params[0])
-        expect(result.error.messages[0][:errors][:alert_type]).to include("invalid_type")
+        expect(result.error.messages[0][:errors]).to include("invalid_type")
         expect(result.error.messages[1][:params]).to eq(alerts_params[1])
-        expect(result.error.messages[1][:errors][:thresholds]).to include("value_is_mandatory")
+        expect(result.error.messages[1][:errors]).to include("value_is_mandatory")
         expect(result.error.messages[3][:params]).to eq(alerts_params[3])
-        expect(result.error.messages[3][:errors][:base]).to include("alert_already_exists")
+        expect(result.error.messages[3][:errors]).to include("alert_already_exists")
         expect(result.error.messages[4][:params]).to eq(alerts_params[4])
-        expect(result.error.messages[4][:errors][:code]).to include("value_already_exist")
+        expect(result.error.messages[4][:errors]).to include("value_already_exist")
       end
     end
 


app/services/usage_monitoring/create_batch_alerts_service.rb
```diff
@@ -0,0 +1,62 @@
...
+      if alerts_params.blank?
+        return result.single_validation_failure!(error_code: "no_alerts", field: :alerts)
+      end
+
+      result.alerts = []
+      result.errors = {}
+
+      ActiveRecord::Base.transaction do
+        alerts_params.each_with_index do |alert_params, index|
+          ActiveRecord::Base.transaction(requires_new: true) do
```
annvelents: this is needed to collect all the errors.
Right now uniqueness is only checked by index, so to get the error we need to try to save, but without an addition transaction, the top transaction fails, and all the other attempts to save return PG error

config/routes.rb
```diff
@@ -67,7 +67,10 @@
 
       resources :subscriptions, only: %i[create update show index], param: :external_id do
         resource :lifetime_usage, only: %i[show update], controller: "subscriptions/lifetime_usages"
-        resources :alerts, only: %i[create index update show destroy], param: :code, controller: "subscriptions/alerts"
+        resources :alerts, only: %i[create index update show destroy], param: :code, controller: "subscriptions/alerts" do
+          post :batch_create, on: :collection
+          delete :destroy_all, on: :collection
```
ancorcruz: The `destroy_all` suffix in the path is redundant. In REST conventions, `DELETE` on a collection endpoint already implies removing all items in that collection. This should be simplified to:

`DELETE /api/v1/subscriptions/:external_id/alerts`

The HTTP method (`DELETE`) conveys the action, and the URI (`/alerts`) identifies the resource (the entire collection). Adding `destroy_all` mixes RPC-style naming into what should be a resource-oriented design.

app/services/usage_monitoring/create_batch_alerts_service.rb
```diff
@@ -0,0 +1,62 @@
...
+              else
+                create_result.error&.message
+              end
+              result.errors[index] = error_details
+              raise ActiveRecord::Rollback
+            end
+          end
+        end
+
+        raise ActiveRecord::Rollback if result.errors.any?
```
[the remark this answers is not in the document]
annvelents: the outer transaction will be rolled back due to `raise ActiveRecord::Rollback if result.errors.any?` on line 47

app/services/usage_monitoring/create_batch_alerts_service.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module UsageMonitoring
+  class CreateBatchAlertsService < BaseService
```
ancorcruz: would you consider the namespace `UsageMonitoring::Alerts::CreateBatchService` instead? does it make sense for you?

app/services/usage_monitoring/create_batch_alerts_service.rb
```diff
@@ -0,0 +1,62 @@
...
+    Result = BaseResult[:alerts, :errors]
+
+    def initialize(organization:, subscription:, alerts_params:)
+      @organization = organization
+      @subscription = subscription
+      @alerts_params = alerts_params
+      super
+    end
+
+    def call
```
ancorcruz: shall it return if no subscription or no organization?

app/services/usage_monitoring/create_batch_alerts_service.rb
ancorcruz: good one, try them all then rollback instead of stop and rollback on first failure

app/services/usage_monitoring/create_batch_alerts_service.rb
```diff
@@ -0,0 +1,62 @@
...
+              subscription:,
+              params: alert_params.to_h
+            )
+
+            if create_result.success?
+              result.alerts << create_result.alert
+            else
+              error_details = {}
+              error_details[:params] = alert_params
+              error_details[:errors] = if create_result.error.respond_to?(:messages)
```
ancorcruz: not a fan of `respond_to?` usage... can we rely in the error type/code instead of if having or not messages method.... looks like our error result api should be consistent but...

app/services/usage_monitoring/destroy_all_alerts_service.rb
```diff
@@ -0,0 +1,31 @@
...
+  class DestroyAllAlertsService < BaseService
+    Result = BaseResult[:alerts]
+
+    def initialize(organization:, subscription:)
+      @organization = organization
+      @subscription = subscription
+      super
+    end
+
+    def call
```
ancorcruz: this should return failure if organzation or subscription is not present.

config/routes.rb
```diff
@@ -67,7 +67,10 @@
 
       resources :subscriptions, only: %i[create update show index], param: :external_id do
         resource :lifetime_usage, only: %i[show update], controller: "subscriptions/lifetime_usages"
-        resources :alerts, only: %i[create index update show destroy], param: :code, controller: "subscriptions/alerts"
+        resources :alerts, only: %i[create index update show destroy], param: :code, controller: "subscriptions/alerts" do
+          post :batch_create, on: :collection
```
ancorcruz: The `/batch_create` suffix isn't necessary here. In REST conventions, `POST` on a collection endpoint already means "create items in this collection," and it can accept an array in the request body for bulk creation. I'd suggest simplifying this to:

`POST /api/v1/subscriptions/:external_id/alerts`

The HTTP method (`POST`) already conveys the creation intent, and the URI (`/alerts`) identifies the target collection. Adding `batch_create` leans toward RPC-style naming rather than resource-oriented design.

ancorcruz (approved):
a few comments to consider, mostly about ReST API design.

app/services/usage_monitoring/create_batch_alerts_service.rb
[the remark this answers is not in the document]
ancorcruz: > the outer transaction will be rolled back due to `raise ActiveRecord::Rollback if result.errors.any?` on line 47

copilot is myopic

app/services/usage_monitoring/destroy_all_alerts_service.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      ActiveRecord::Base.transaction do
+        alerts.each do |alert|
+          alert.thresholds.delete_all
+          alert.discard!
+        end
+      end
+
+      result.alerts = alerts
+      result
```
julienbourdeau: I think I'd do it without a loop but maybe there is a reason I'm missing.

In any case, the alert returned by the service won't have the thresholds because there weren't loaded before being hard deleted.

I think we should return just a `head :ok` from the controller. No need to return the alert deleted. Just my sentiment, maybe I'm wrong ^^

```suggestion
      alert_ids = organization.alerts.where(subscription_external_id: subscription.external_id).ids

      ActiveRecord::Base.transaction do
        AlertThreshold.where(id: alert_ids).delete_all
        Alert.where(id: alert_ids).update_all deleted_at: Time.current
      end
      
      result
```

config/routes.rb
julienbourdeau: I agree

app/controllers/api/v1/subscriptions/alerts_controller.rb
```diff
@@ -6,6 +6,47 @@ module Subscriptions
...
+          result = UsageMonitoring::CreateBatchAlertsService.call(
+            organization: current_organization,
+            subscription:,
+            alerts_params: batch_create_params[:alerts]
+          )
+
+          if result.success?
+            render(
+              json: ::CollectionSerializer.new(
+                result.alerts.map { |alert| alert.reload },
```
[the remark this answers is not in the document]
julienbourdeau: i'm also confused why the reload is necessary here 🤔

app/services/usage_monitoring/create_batch_alerts_service.rb
julienbourdeau: Is this designed after batch events?

config/routes.rb
annvelents: oh, shoot, right

app/controllers/api/v1/subscriptions/alerts_controller.rb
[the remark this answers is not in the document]
annvelents: yep, agree

app/services/usage_monitoring/create_batch_alerts_service.rb
annvelents: yes, totally

app/services/usage_monitoring/destroy_all_alerts_service.rb
annvelents: yes, 100% it's better without loop

app/services/usage_monitoring/create_batch_alerts_service.rb
annvelents: not exactly because on events we firstly validate params and return validation errors in a separate step, while in Alerts the validations are on the indexes level

app/services/usage_monitoring/create_batch_alerts_service.rb
annvelents: @ancorcruz , changed logic to always use "message" that is defined in the base error result

app/controllers/api/v1/subscriptions/alerts_controller.rb
```diff
@@ -38,16 +51,10 @@ def show
...
         def create
-          result = UsageMonitoring::CreateAlertService.call(
-            organization: current_organization,
-            subscription:,
-            params: create_params.to_h
-          )
-
-          if result.success?
-            render_alert(result.alert)
+          if params[:alerts].present?
```
julienbourdeau, line 54: I initially had in mind to create 2 endpoints but I like this a lot more 👏 👏 👏 
I think we should always handle batch this way now.

annvelents: yeah, with batch I was initially inspired by events... but it does make sense to be able handle them this way

PR #4994, opened 2026-02-06
fix(activity_log): Handle error with kafka error message

vincent-pochet:
## Context

Activity log tracking is sometime failing with the following error:

```
WaterDrop::Errors::MessageInvalidError
{payload: "is more than `max_payload_size` config value"} (WaterDrop::Errors::MessageInvalidError)
```

It happens with plan or subscription tracking when the plan has a large number of charges and or charge filters.

## Description

This PR is updating the way we are building the payload for plans and subscriptions by checking the number of charges. When it has more than 50 charges, it just skip the charges from the payload. The diff will still contains all the changes made by the user.

In parallel, the `WaterDrop::Errors::MessageInvalidError` is also catched to avoid raising it to the end-user. It is then pushed to Sentry for later analysis.

To allow the `SubscriptionSerializer` to specify which relation to include in the plan field, an update has been made to the base serializer to handle nested relation.

It can be used this way:
```ruby
V1::SubscriptionSerializer.new(
  object, root_name: :subscriptions, includes: [{plans: [:charge]}]
).serialize
```
It will serialize the subscription, with the plan and the plan's charges. It will replaces the default includes defined in the serializer (https://github.com/getlago/lago-api/blob/main/app/serializers/v1/subscription_serializer.rb#L61)

### DIFF
--- a/app/serializers/model_serializer.rb
+++ b/app/serializers/model_serializer.rb
@@ -6,7 +6,11 @@ class ModelSerializer
   # The possible values for the options are:
   # - includes: Specify the relation to include in the payload
   #   Expected format: [:customer, {plan: [:charges]}]
-  #   Hash values will be passed to the relation serializer
+  #
+  #   Hash values can be passed to the relation serializer
+  #   when used with `included_relations`.
+  #   Example:
+  #     included_relations(:plan, default: [:charges])`
   def initialize(model, options = {})
     @model = model
     @options = options
@@ -51,7 +55,7 @@ class ModelSerializer
       next include.key?(value) if include.is_a?(Hash)
     end
 
-    return default if include.is_a?(Symbol)
+    return default if include.is_a?(Symbol) || include.nil?
 
     include[value]
   end
--- a/spec/serializers/model_serializer_spec.rb
+++ b/spec/serializers/model_serializer_spec.rb
@@ -94,6 +94,14 @@ RSpec.describe ModelSerializer do
           expect(serializer.included_relations(:name)).to eq([:first, :last])
         end
       end
+
+      context "when include is not found" do
+        let(:options) { {includes: [:id]} }
+
+        it "returns an empty array" do
+          expect(serializer.included_relations(:name)).to eq([])
+        end
+      end
     end
   end
 end


app/services/utils/activity_log.rb
```diff
@@ -125,12 +132,27 @@ def object_serialized
...
+          SERIALIZED_INCLUDED_OBJECTS[:invoice] - [:fees]
+        else
+          SERIALIZED_INCLUDED_OBJECTS[:invoice]
+        end
+      when :plan
+        if object.charges.count > MAX_SERIALIZED_CHARGES
+          SERIALIZED_INCLUDED_OBJECTS[:plan] - [:charges]
+        else
+          SERIALIZED_INCLUDED_OBJECTS[:plan]
+        end
```
julienbourdeau, line 147: If I understand correctly, the diff is always in the payload anyway, so maybe it's worth always removing the charges here 🤔

app/serializers/model_serializer.rb
```diff
@@ -3,6 +3,10 @@
 class ModelSerializer
   attr_reader :model, :options
 
+  # The possible values for the options are:
+  # - includes: Specify the relation to include in the payload
+  #   Expected format: [:customer, {plan: [:charges]}]
+  #   Hash values will be passed to the relation serializer
```
julienbourdeau: There is no guarantee it's passed.
```suggestion
  #   Hash values can be passed to the relation serializer using `includes: included_relations(:plan)`
  #   Defaults can be used too `includes: included_relations(:plan, default: [:charges])`
```

app/serializers/model_serializer.rb
```diff
@@ -22,9 +26,33 @@ def root_name
...
+    return default if includes.blank?
+
+    include = includes.find do |include|
+      next value == include if include.is_a?(Symbol)
+      next include.key?(value) if include.is_a?(Hash)
+    end
+
+    return default if include.is_a?(Symbol)
+
+    include[value]
```
D1353L: Consider checking if `include` is not nil. As if nothing was found, NoMethodError will be raised

app/serializers/model_serializer.rb
vincent-pochet: That's true, it would require updating all existing includes (a task for another day 😅 )

app/services/utils/activity_log.rb
vincent-pochet: Yes the diff is always stored (at least if it's not long as well... 😅).
The objective here is to be "best effort" when trying to create the log. I don't want to re-challenge the initial feature

PR #4996, opened 2026-02-06
fix(integrations): split credit note into pre-tax and tax portions for NetSuite

ancorcruz:
## Context

When an invoice has a credit note that fully covers the amount (fees + taxes), the NetSuite sync fails with "The total can not be negative". This happens because the credit note discount was sent as a single post-tax line item, while fees are sent pre-tax with taxes in a separate taxdetails section.

## Description

Split the credit note discount into its pre-tax (line item) and tax (tax detail) portions. A new `credit_notes_taxes_amount_cents` method computes the proportional tax from each applied credit note. The line item now subtracts only the pre-tax amount, and the corresponding tax detail carries the negative tax amount instead of zero.

### DIFF
--- a/app/services/integrations/aggregator/invoices/payloads/netsuite.rb
+++ b/app/services/integrations/aggregator/invoices/payloads/netsuite.rb
@@ -191,7 +191,7 @@ module Integrations
                 "item" => credit_note_item.external_id,
                 "account" => credit_note_item.external_account_code,
                 "quantity" => 1,
-                "rate" => -amount(invoice.credit_notes_amount_cents, resource: invoice),
+                "rate" => -amount(invoice.credit_notes_amount_cents - credit_notes_taxes_amount_cents, resource: invoice),
                 "taxdetailsreference" => "credit_note_item",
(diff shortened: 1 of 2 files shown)


app/services/integrations/aggregator/invoices/payloads/netsuite.rb
```diff
@@ -253,6 +253,18 @@ def discount_taxes
...
           end
 
+          def credit_notes_taxes_amount_cents
+            @credit_notes_taxes_amount_cents ||= invoice.credits.credit_note_kind.includes(:credit_note).sum do |credit|
+              cn = credit.credit_note
+              subtotal = cn.sub_total_including_taxes_amount_cents
+              if subtotal.zero?
+                0
+              else
+                credit.amount_cents * cn.taxes_amount_cents / subtotal
```
vincent-pochet: Should it be (maybe with some rounding/flooring logic)?
```suggestion
                (credit.amount_cents * cn.taxes_amount_cents).fdiv(subtotal)
```

vincent-pochet (approved):

ancorcruz:
Won't merge, managed in Netsuite side.

PR #5000, opened 2026-02-09
Fix(indexes): add indexes on wallets and on customers

annvelents:
To help DB finding customers and wallets, we need these indexes

### DIFF
--- a/app/queries/wallets_query.rb
+++ b/app/queries/wallets_query.rb
@@ -35,6 +35,6 @@ class WalletsQuery < BaseQuery
   end
 
   def customer
-    organization.customers.where(external_id: filters.external_customer_id, deleted_at: nil)
+    organization.customers.where(external_id: filters.external_customer_id)
   end
 end


app/models/customer.rb
```diff
@@ -394,6 +394,7 @@ def ensure_slug
 #  index_customers_on_awaiting_wallet_refresh            (awaiting_wallet_refresh)
 #  index_customers_on_billing_entity_id                  (billing_entity_id)
 #  index_customers_on_deleted_at                         (deleted_at)
+#  index_customers_on_external_id                        (organization_id,external_id)
 #  index_customers_on_external_id_and_organization_id    (external_id,organization_id) UNIQUE WHERE (deleted_at IS NULL)
```
ancorcruz, line 398: isn't this new index redundant, we already have one for orga_id and external_id or do we need to include soft deleted customers on the results?

app/models/wallet.rb
```diff
@@ -157,11 +157,12 @@ def unique_code_per_customer
...
 #
-#  index_uniq_wallet_code_per_customer     (customer_id,code) UNIQUE
-#  index_wallets_on_customer_id            (customer_id)
-#  index_wallets_on_organization_id        (organization_id)
-#  index_wallets_on_payment_method_id      (payment_method_id)
-#  index_wallets_on_ready_to_be_refreshed  (ready_to_be_refreshed) WHERE ready_to_be_refreshed
+#  index_uniq_wallet_code_per_customer               (customer_id,code) UNIQUE
+#  index_wallets_on_customer_id                      (customer_id)
+#  index_wallets_on_organization_id                  (organization_id)
+#  index_wallets_on_organization_id_and_customer_id  (organization_id,customer_id)
```
ancorcruz, line 163: on this new index... do we really need it? we already have indices for organization_id and customer_id individually, Postgres will combine them via BitmapAnd, is this a bottleneck currently? The index adds write overhead (on every write), maybe without a meaningful read benefit.

app/models/customer.rb
annvelents: weirdly, that index is not being used by PG when running the query, but this one was used when Jeremy manually created it in prod

app/models/wallet.rb
annvelents: same here: PG was not picking them up, adding it allowed to lower DB load significantly

app/models/customer.rb
vincent-pochet: We could probably clean the `organization_id` after the merge as it will be indeed redundant (PG hero was suggesting this change)

ancorcruz (approved):
LGTM

PR #5003, opened 2026-02-09
fix(alerts): Allow thresholds to be used for recurring and non-recurring

julienbourdeau:
Due to a validation error, we weren't allowing the recurring threshold to be set to the same value as the normal thresholds.

### BEFORE
CleanShot 2026-02-09 at 16 17 58@2x

### AFTER
CleanShot 2026-02-09 at 16 18 28@2x

### DIFF
--- a/app/services/usage_monitoring/create_alert_service.rb
+++ b/app/services/usage_monitoring/create_alert_service.rb
@@ -26,8 +26,7 @@ module UsageMonitoring
         return result.single_validation_failure!(field: :thresholds, error_code: "too_many_thresholds")
       end
 
-      threshold_keys = params[:thresholds].map { |t| [t[:value], ActiveModel::Type::Boolean.new.cast(t[:recurring]) || false] }
-      if threshold_keys.size != threshold_keys.uniq.size
+      if duplicate_threshold_values?(params[:thresholds])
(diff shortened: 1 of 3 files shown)


app/services/usage_monitoring/create_alert_service.rb
```diff
@@ -26,8 +26,8 @@ def call
         return result.single_validation_failure!(field: :thresholds, error_code: "too_many_thresholds")
       end
 
-      threshold_values = params[:thresholds].map { |t| t[:value] }.compact
-      if threshold_values.size != threshold_values.uniq.size
+      threshold_keys = params[:thresholds].map { |t| [t[:value], t[:recurring] || false] }
```
groyoh: Do we check beforehand that `t[:recurring]` is only `true` or `false` ? Maybe cast to boolean to make sure ?

julienbourdeau: I'll double check

vincent-pochet (approved):

app/services/usage_monitoring/create_alert_service.rb
julienbourdeau: Oohh excellent catch

PR #5006, opened 2026-02-09
Feat/wallet targets/further optimisation

annvelents:
## Context
 
The charge fee calculation was issuing a number of redundant and N+1 queries. This is a follow-up optimisation pass on the wallet targets work (when we might have thousands of fees) to bring down query volume on these hot paths, with no change to invoicing or usage behaviour.
 
## Description
 
**Skip adjusted-fee lookups when none apply**
 
- `CalculateFeesService` and `CustomerUsageService` now run a single existence check (`adjusted_fee_exists`) per subscription, matching adjusted charge fees on the boundary datetimes (`charges_from_datetime` / `charges_to_datetime`).
- The result is threaded down to `Fees::ChargeService` through a new `skip_adjusted_fees` option. When no adjusted fee exists, `ChargeService#adjusted_fee` short-circuits and returns early instead of running a lookup per charge/filter.

**Preload and reuse already-loaded records**
 
- `Fees::ChargeService` now preserves the `charge` and `charge_filter` associations (in addition to `billable_metric`) on every fee it builds, including cached ones, so downstream code doesn't re-query them.
- `init_fee` uses `subscription.organization_id` instead of loading the full `organization` record just to read its id.
- `plan` and `customer` are now passed into `ChargeService` and forwarded to `Fees::ApplyTaxesService`, allowing taxes to be applied without reloading them.
- `Fees::ApplyTaxesService` fetches billing-entity default taxes via `BillingEntity::AppliedTax.where(billing_entity_id: ...)` rather than traversing the `customer.billing_entity` association.
- `:applied_pricing_unit` is added to the charge `includes(...)` in both services to eager-load it and avoid an N+1.
**Dependencies:** none.

### DIFF
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -21,11 +21,10 @@ module Fees
       apply_taxes: false,
       calculate_projected_usage: false,
       with_zero_units_filters: true,
-      usage_filters: UsageFilters::NONE
-      with_zero_units_filters: true,
+      usage_filters: UsageFilters::NONE,
       skip_adjusted_fees: false,
       plan: nil,
-      customer: nil,
+      customer: nil
     )
       @invoice = invoice
       @charge = charge
@@ -289,7 +288,7 @@ module Fees
 
       new_fee = Fee.new(
         invoice:,
-        organization_id: organization.id,
+        organization_id: subscription.organization_id,
         billing_entity_id: subscription.applicable_billing_entity_id,
         subscription:,
         charge:,
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -152,7 +152,6 @@ module Fees
           fee.association(:charge_filter).target = charge_filter
         end
       end
-      end
 
       result.fees.concat(fees.compact)
     end
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -146,11 +146,7 @@ module Fees
       fees.each do |fee|
         fee.association(:billable_metric).target = billable_metric
         fee.association(:charge_filter).target = charge_filter if charge_filter&.id
-        # TODO: check it!
         fee.association(:charge).target = charge
-        if charge_filter && fee.charge_filter_id
-          fee.association(:charge_filter).target = charge_filter
-        end
       end
 
       result.fees.concat(fees.compact)
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -22,6 +22,8 @@ module Fees
       calculate_projected_usage: false,
       with_zero_units_filters: true,
       usage_filters: UsageFilters::NONE
+      with_zero_units_filters: true,
+      skip_adjusted_fees: false
     )
       @invoice = invoice
       @charge = charge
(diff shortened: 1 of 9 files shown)


app/services/fees/charge_service.rb
```diff
@@ -309,7 +317,7 @@ def init_fee(amount_result, properties:, charge_filter:, adjusted:)
       end
 
       if apply_taxes
-        taxes_result = Fees::ApplyTaxesService.call(fee: new_fee)
+        taxes_result = Fees::ApplyTaxesService.call(fee: new_fee, plan: @plan, customer: @customer)
```
annvelents, line 320: plan and customer are passed in `fee::apply_taxes_service` to be used in 
```
      @customer = customer || fee.invoice&.customer || fee.subscription.customer
      @plan = plan || fee.subscription&.plan
```
to get taxes from the customer or from the plan

brunomiguelpinto (approved):

app/services/invoices/calculate_fees_service.rb
```diff
@@ -116,11 +116,17 @@ def create_charges_fees(subscription, boundaries)
       return unless charge_boundaries_valid?(boundaries)
 
       filters = event_filters(subscription, boundaries).charges
+      plan = subscription.plan
+      customer = subscription.customer
+      adjusted_fee_exists = AdjustedFee
+        .where(invoice:, subscription:, fee_type: :charge)
+        .where("(properties->>'charges_from_datetime')::timestamptz = ?", boundaries.charges_from_datetime&.iso8601(3))
+        .where("(properties->>'charges_to_datetime')::timestamptz = ?", boundaries.charges_to_datetime&.iso8601(3)).exists?
```
rsempe: what about extracting it into a scope?

```
scope :matching_charge_boundaries, ->(boundaries) {
  where(fee_type: :charge)
    .where("(properties->>'charges_from_datetime')::timestamptz = ?", boundaries.charges_from_datetime&.iso8601(3))
    .where("(properties->>'charges_to_datetime')::timestamptz = ?", boundaries.charges_to_datetime&.iso8601(3))
}
```

app/services/fees/charge_service.rb
```diff
@@ -21,7 +21,10 @@ def initialize(
       apply_taxes: false,
       calculate_projected_usage: false,
       with_zero_units_filters: true,
-      usage_filters: UsageFilters::NONE
+      usage_filters: UsageFilters::NONE,
+      skip_adjusted_fees: false,
```
rsempe, line 25: what about using the naming `with_adjusted_fees` to be  consistent with `with_zero_units_filters`?

app/services/invoices/calculate_fees_service.rb
rsempe, line 239: should we pass `customer` and `plan` here? 🤔

app/services/fees/charge_service.rb
annvelents: I wanted default value to be false, so when not defined it's more naturally perceived as "false"

app/services/invoices/calculate_fees_service.rb
annvelents: ha! nice catch!
yes

app/services/fees/charge_service.rb
rsempe: ok!

rsempe (approved):

brunomiguelpinto (approved):

PR #5015, opened 2026-02-10
fix: attempt to fix race condition

lovrocolic:
## Context

This PR tries to fix race condition on credit note consumption where same CN could be applied multiple time on the invoice

### DIFF
--- a/app/services/credits/credit_note_service.rb
+++ b/app/services/credits/credit_note_service.rb
@@ -17,8 +17,10 @@ module Credits
       if context == :preview
         apply_credits
       else
-        CreditNotes::LockService.new(customer:).call do
-          apply_credits
+        ActiveRecord::Base.transaction do
+          CreditNotes::LockService.new(customer:).call do
+            apply_credits
+          end
         end
       end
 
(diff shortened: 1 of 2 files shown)


app/services/credits/credit_note_service.rb
```diff
@@ -36,47 +38,44 @@ def call
...
     end
 
     def credit_notes
-      @credit_notes ||= customer.credit_notes
+      customer.credit_notes
         .finalized
         .available
         .where.not(invoice_id: invoice.id)
         .order(created_at: :asc)
+        .to_a
```
osmarluz, line 78: @lovrocolic just a small question here: why do we need to convert the collection to an array?

lovrocolic: It is a bit optional, but for this use-case to avoid any lazy loading and to be sure that we get fresh results whenever `credit_notes` is called

osmarluz: @lovrocolic got it, thanks!

osmarluz (approved):
🚀

PR #5017, opened 2026-02-10
rake(entitlement): Remove duplicated subscription entitlements

julienbourdeau:
Due to a bug recently fixed, we created subscrition entitlements when there was no override.
In this case, the entitlements are inherited from the plan and no entitlement on the sub should exists.

This is not an issue until you delete the entitlement on the plan... and you realize all subs still have the feature.

This take task (soft-)delete the extra entitlement that should not have been created.

[The fix for the bug is here](https://github.com/getlago/lago-api/pull/5002) 🙈

### DIFF
--- a/lib/tasks/entitlements.rake
+++ b/lib/tasks/entitlements.rake
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+namespace :entitlements do
+  desc "Soft-delete duplicate subscription entitlements that have no values and whose feature is already on the parent plan"
+  task cleanup_duplicate_subscription_entitlements: :environment do
+    deleted_at = Time.current.beginning_of_hour
+    batch_size = 5_000
+    total_deleted = 0
+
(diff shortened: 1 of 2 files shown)


lib/tasks/entitlements.rake
```diff
@@ -0,0 +1,44 @@
...
+              SELECT 1 FROM entitlement_entitlement_values v
+              WHERE v.entitlement_entitlement_id = sub_ent.id
+                AND v.deleted_at IS NULL
+            )
+          LIMIT $2
+        )
+      SQL
+
+      total_deleted += result
+      puts "  Progress: #{total_deleted} entitlements deleted..." if (total_deleted % 25_000) < batch_size
```
mariohd:
```suggestion
      puts "  Progress: #{total_deleted} entitlements deleted..." if (total_deleted % 25_000).zero?
```

julienbourdeau: Actually I think it would skip the last puts in this case. It's not too important tho.

PR #5020, opened 2026-02-11
fix: coupons and credit notes race condition fix

lovrocolic:
## Context

Currently if multiple invoices for the same customer are created simultaneously, credits are not spent correctly

## Description

This PR fixes described case with using advisory lock.


osmarluz:
@lovrocolic don't we need to add the `ActiveRecord::Base.transaction` to the `api/app/services/credits/applied_coupons_service.rb` the same way we did for the `api/app/services/credits/credit_note_service.rb`?

PR #5022, opened 2026-02-11
fix(stripe-error): When Customer does not exist, handle with third_pa…

mariohd:
## Summary                                                                                                                                                                                                          
                                         
  - Change `StripeService#update` to return `third_party_failure!` instead of `service_failure!` when Stripe raises `InvalidRequestError` or `PermissionError`, so the GraphQL layer returns a proper error instead of
   a 500                                                    
  - Rescue `BaseService::ThirdPartyFailure` in `StripeCheckoutUrlJob` so the job doesn't fail when the Stripe customer doesn't exist                                                                                  
                                                                                                                                                                                                                      
  ## Context

  When a `provider_customer_id` references a customer that no longer exists in Stripe:

  1. **GraphQL (`updateCustomer`)**: `StripeService#update` returned `service_failure!`, which is not handled by `ExecutionErrorResponder#result_error` — falls through to the `else` branch returning a 500. Changing
   to `third_party_failure!` matches the existing handler.

  2. **Job (`StripeCheckoutUrlJob`)**: `generate_checkout_url` already returned `third_party_failure!`, but the job only rescued `UnauthorizedFailure`. Adding `ThirdPartyFailure` to the rescue lets the job complete
   gracefully.

  ## Test plan

  - [x] Updated `stripe_service_spec.rb` assertions for `#update` error cases
  - [x] Verify `updateCustomer` mutation returns a proper error when `provider_customer_id` is invalid

### DIFF
--- a/app/services/payment_provider_customers/stripe_service.rb
+++ b/app/services/payment_provider_customers/stripe_service.rb
@@ -42,7 +42,7 @@ module PaymentProviderCustomers
     rescue ::Stripe::InvalidRequestError, ::Stripe::PermissionError => e
       deliver_error_webhook(e)
 
-      result.service_failure!(code: "stripe_error", message: e.message)
+      result.third_party_failure!(third_party: "Stripe", error_code: e.code, error_message: e.message)
     rescue ::Stripe::AuthenticationError => e
       deliver_error_webhook(e)
 
--- a/spec/graphql/mutations/customers/update_spec.rb
+++ b/spec/graphql/mutations/customers/update_spec.rb
@@ -141,6 +141,31 @@ RSpec.describe Mutations::Customers::Update do
     expect(result_data["billingEntity"]["code"]).to eq(billing_entity.code)
   end
 
+  context "when stripe customer does not exist" do
+    before do
+      stripe_provider
+
+      allow(Stripe::Customer).to receive(:update)
+        .and_raise(Stripe::InvalidRequestError.new("No such customer: 'cus_invalid'", nil, code: "resource_missing"))
+    end
+
+    it "returns a third party error" do
+      result = execute_graphql(
+        current_user: membership.user,
+        permissions: required_permissions,
+        query: mutation,
+        variables: {
+          input: input.merge(
+            providerCustomer: {providerCustomerId: "cus_invalid"}
+          )
+        }
+      )
+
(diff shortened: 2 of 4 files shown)


spec/graphql/mutations/customers/update_spec.rb
```diff
@@ -141,6 +141,34 @@
     expect(result_data["billingEntity"]["code"]).to eq(billing_entity.code)
   end
 
+  context "when stripe customer does not exist" do
+    let(:stripe_customer) { create(:stripe_customer, customer:) }
+
+    before do
+      stripe_provider
+      stripe_customer
```
osmarluz: @mariohd it's safe to remove the `stripe_customer` here, since it's not being used:


```suggestion
    before do
      stripe_provider
```

app/services/payment_provider_customers/stripe_service.rb
```diff
@@ -42,7 +42,7 @@ def update
     rescue ::Stripe::InvalidRequestError, ::Stripe::PermissionError => e
       deliver_error_webhook(e)
 
-      result.service_failure!(code: "stripe_error", message: e.message)
+      result.third_party_failure!(third_party: "Stripe", error_code: e.code, error_message: e.message)
```
osmarluz, line 45: @mariohd this service is also used by `api/app/services/customers/upsert_from_api_service.rb`. Would it be safe to change the type of error we return via the API? These are the responses:


```json
// result.service_failure!(code: "stripe_error", message: e.message)
{
    "status": 422,
    "error": "Unprocessable Entity",
    "code": "validation_errors",
    "error_details": {
        "base": [
            "stripe_error"
        ]
    }
}
```


```json
// result.third_party_failure!(third_party: "Stripe", error_code: e.code, error_message: e.message)
{
    "status": 422,
    "error": "Unprocessable Entity",
    "code": "third_party_error",
    "error_details": {
        "third_party": "Stripe",
        "thirdparty_error": "No such customer: 'whatever'"
    }
}
```

I'm asking because we don't have the `third_party_error` documented on https://swagger.getlago.com/#/customers/createCustomer

mariohd: Just the details are different, the status and error are the same, it's safe. 
We can change this in the open api once we got this merged.

osmarluz (approved):
🚀

PR #5023, opened 2026-02-11
Chore: webhook improvements

annvelents:


app/jobs/send_http_webhook_job.rb
```diff
@@ -15,6 +15,20 @@ class SendHttpWebhookJob < ApplicationJob
...
     Webhooks::SendHttpService.call(webhook:)
   end
+
+  private
+
+  def pending_non_slow_webhooks?
+    Webhook.pending
+      .joins(:webhook_endpoint)
+      .where(webhook_endpoints: {slow_response: false})
+      .exists?
```
annvelents: will this check be faster if we 
1) select distinct webhook_endpoint_ids
2) run exist on webhook endpoints table separately

is it possible that we have a pending webhook that is not in the queue?

annvelents: 1) index on webhooks is:
webhook_endpoint_id, status, that's why it's beeter to start with webhook_endpoint
2) no, after webhook is processed, it's marked either as succeeded or failed

app/jobs/send_slow_http_webhook_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class SendSlowHttpWebhookJob < ApplicationJob
+  queue_as :low_priority
```
annvelents: or should it be `low_priority_webhooks`?

brunomiguelpinto (approved):

app/jobs/send_slow_http_webhook_job.rb
brunomiguelpinto: Since they go to the end of the queue i would keep the same

jdenquin: This queue is not used on the webhook worker, we should have an other queue for only slow webhooks, so the webhook worker can take care of it (with a lower priority)

PR #5025, opened 2026-02-12
feat(partman): Add hourly cron for pg_partman

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/4916 and https://github.com/getlago/lago-api/pull/4918

## Description

The local config of pg_partman uses the built-in `pg_partman_bgw` postgres extension to schedule the maintenance tasks of the `pg_partman` extension.
This task is responsible from moving the enriched_events stored in the `enriched_events_default` partition to the timestamped partition allowing better performances when reading the enriched_events.

The `pg_partman_bgw` is part of `pg_partman` but is not available on amazon RDS. However, RDS exposes the `pg_cron` extension to handle automatic and recurring tasks.

This PR activate the extension when present on the server and creates a scheduled task that will execute the pg partman maintenance procedure every hours

### DIFF
--- a/db/migrate/20260212091141_setup_pg_cron.rb
+++ b/db/migrate/20260212091141_setup_pg_cron.rb
@@ -1,25 +1,19 @@
 # frozen_string_literal: true
 
 class SetupPgCron < ActiveRecord::Migration[8.0]
+  include Migrations::ExtensionHelper
+
   def up
     safety_assured do      # Check if pg_partman is available on the server
-      partman_result = execute <<~SQL
-        SELECT 1 FROM pg_available_extensions WHERE name = 'pg_partman';
-      SQL
       # No partitioning was configured in previous migrations, we can skip this migration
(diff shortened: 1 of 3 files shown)


db/migrate/20260212091141_setup_pg_cron.rb
```diff
@@ -0,0 +1,26 @@
...
+      result = execute <<~SQL
+        SELECT 1 FROM pg_available_extensions WHERE name = 'pg_cron'
+      SQL
+
+      if result.ntuples.zero?
+        Rails.logger.debug "pg_cron extension is not available on this PostgreSQL server, skipping..."
+      else
+        execute <<~SQL
+          CREATE EXTENSION IF NOT EXISTS pg_cron;
+        SQL
```
groyoh: We should define a helper for checking whether an extension is available.

PR #5028, opened 2026-02-12
misc(database): Add pg extension helper

vincent-pochet:
## Context

This PR is related to https://github.com/getlago/lago-api/pull/5025#discussion_r2798059134

## Description

It adds a new helper for migrations with postgres to check the presence on an extension library.

Usage:
```ruby
require Rails.root.join("lib/migrations/extension_helper")

class SomeMigration < ActiveRecord::Migration[8.0]
  include Migrations::ExtensionHelper

  def up
    return unless pg_extension_present?("some_pg_extension")
    
    # ... some migration logic
  end
end
```

### DIFF
--- a/db/migrate/20260109092932_setup_partman.rb
+++ b/db/migrate/20260109092932_setup_partman.rb
@@ -3,19 +3,15 @@
 class SetupPartman < ActiveRecord::Migration[8.0]
   def up
     safety_assured do
-      # Check if pg_partman is available on the server
-      result = execute <<~SQL
-        SELECT 1 FROM pg_available_extensions WHERE name = 'pg_partman';
-      SQL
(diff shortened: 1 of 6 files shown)


db/migrate/20260109092932_setup_partman.rb
```diff
@@ -1,21 +1,21 @@
 # frozen_string_literal: true
 
+require Rails.root.join("lib/migrations/extension_helper")
```
julienbourdeau: I'd personnaly try to load this file in application.rb or an initializer because don't like `require`s in the app but it is just a personnal preference.

julienbourdeau (approved):
Awesome 👌

PR #5032, opened 2026-02-12
chore(usage): usage service can accept charge, filter, full sub usage

annvelents:
## Context

in case when a subscription has pricing_group_keys to monitor usages, current usage will return all possible fees.
In case when there is a plan with 17 BM and 23K groups due to pricing group keys, we might end up with 390K fees of usage. We'll struggle to keep them in the memory + it won't be possible to send via API.

This PR prepares service to be able to:
- return usage for a specific charge
- return usage for a specific group
- skip pricing_group_keys completely (will be needed for wallet calculation)
- return full_usage instead of only current usage to be able to see granular usage for subscriptions with length 17months, for example

### DIFF
--- a/app/models/usage_filters.rb
+++ b/app/models/usage_filters.rb
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 
 class UsageFilters
+  # filter_by_charge - when set, allows to calculate usage only for a specific charge and not all charges of a plan
+  #          example: Charge.last
+  #
+  # filter_by_group  - when set, calculates usage only for specific {pricing_group_key: value}.
+  #                    Note that if sent, will ignore grouping by this pricing_group_key
+  #          example: {user_id: '123'}
+  #
+  # skip_grouping    - when set, will ignore grouping by pricing_group_keys
+  # full_usage       - when set, will ignore boundaries and will return usage since subscription.started_at
+
   attr_reader :filter_by_charge, :filter_by_group, :skip_grouping, :full_usage
 
   def initialize(filter_by_charge: nil, filter_by_group: nil, skip_grouping: false, full_usage: false)
--- a/app/services/invoices/customer_usage_service.rb
+++ b/app/services/invoices/customer_usage_service.rb
@@ -47,8 +47,7 @@ module Invoices
     def call
       return result.not_found_failure!(resource: "customer") unless customer
       return result.not_allowed_failure!(code: "no_active_subscription") if subscription.blank?
-      full_usage_allowed = (usage_filters.filter_by_charge.present? || usage_filters.filter_by_group.present?) && subscription.plan.charges.where(prorated: true).none?
-      return result.not_allowed_failure!(code: "full_usage_not_allowed") if usage_filters.full_usage && !full_usage_allowed
+      return result.not_allowed_failure!(code: "full_usage_not_allowed") if usage_filters.full_usage && !querying_full_usage_allowed
 
       result.usage = compute_usage
       result.invoice = invoice
@@ -103,7 +102,7 @@ module Invoices
         .charges
         .joins(:billable_metric)
         .includes(:taxes, billable_metric: :organization, filters: {values: :billable_metric_filter})
-      if usage_filters.filter_by_charge
+      if usage_filters.filter_by_charge.present?
         charges = charges.where(id: usage_filters.filter_by_charge.id)
       end
 
@@ -124,7 +123,7 @@ module Invoices
 
       applied_boundaries = boundaries
       applied_boundaries = boundaries.dup.tap { it.max_timestamp = max_timestamp } if max_timestamp
-      if usage_filters.filter_by_group
+      if usage_filters.filter_by_group.present?
         cache_middleware = nil
       end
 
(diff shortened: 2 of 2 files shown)


app/services/events/stores/clickhouse_store.rb
```diff
@@ -54,8 +54,8 @@ def events_cte_queries_without_deduplication(force_from: false, ordered: false,
           applicable_to_datetime
         )
 
-        query = apply_arel_grouped_by_values(query) if grouped_by_values?
         query = arel_filters_scope(query)
+        query = apply_arel_grouped_by_values(query) if grouped_by_values?
```
annvelents, line 58: hope it won't break anything, but in our case it's better to 
1) limit the scope, applying filters
2) apply grouping

without this change I was getting into the method of filters per each group (so still, 20K fees in our case)

app/services/fees/charge_service.rb
```diff
@@ -12,7 +12,8 @@ def initialize(
       filtered_aggregations: nil,
       apply_taxes: false,
       calculate_projected_usage: false,
-      with_zero_units_filters: true
+      with_zero_units_filters: true,
+      for_pricing_group_pairs: nil
```
annvelents: naming 100% can be better

app/services/fees/charge_service.rb
```diff
@@ -368,6 +370,12 @@ def aggregation_filters(charge_filter: nil)
         filters[:ignored_filters] = result.ignored_filters
       end
 
+      if @for_pricing_group_pairs.present?
+        filters[:grouped_by] = nil
+        filters[:matching_filters] ||= {}
+        filters[:matching_filters].merge!(@for_pricing_group_pairs)
+      end
```
annvelents, line 388: this is too complex
but works 😅 
but needs to be refactored with a fresh head

app/services/invoices/customer_usage_service.rb
```diff
@@ -10,7 +10,9 @@ def initialize(
       with_cache: true,
       max_timestamp: nil,
       calculate_projected_usage: false,
-      with_zero_units_filters: true
+      with_zero_units_filters: true,
+      for_charge: nil,
+      for_pricing_group_pairs: nil
```
annvelents: better naming!

annvelents: filter_by_group

annvelents: filter_by_charge

app/services/fees/charge_service.rb
```diff
@@ -368,6 +370,12 @@ def aggregation_filters(charge_filter: nil)
         filters[:ignored_filters] = result.ignored_filters
       end
 
+      if @for_pricing_group_pairs.present?
+        filters[:grouped_by] = nil
```
annvelents: should not be nil:
if grouping_keys are:
workspace and user_id,
and now we're filtering by workspace,
inside workspace we should be able to see groups of users

annvelents: so we need only remove group_key that matches for_pricing_group_pairs.keys

app/services/fees/charge_service.rb
```diff
@@ -368,6 +379,15 @@ def aggregation_filters(charge_filter: nil)
...
       end
 
+      if filter_by_group.present?
+        # when pricing group keys on a charge are "workspace" and "user", and filter_by_group is {"workspace" => "A"},
+        # we want to remove the grouping keys "workspace", but keep the grouping key "user", so the usage will still be granular within the workspace
+        filter_by_group.keys.each { |key| filters[:grouped_by]&.delete(key) }
+        filters[:matching_filters] ||= {}
+        # expected matching_filters format is { "workspace" => ["A", "B"], "user" => ["U1", "U2"] }
+        filters[:matching_filters].merge!(filter_by_group)
+      end
```
annvelents, line 388: Can it be refactored to be more readable or are these comments enough?

app/services/events/stores/clickhouse_store.rb
vincent-pochet: Not sure to get this one. Both method are just adding to where conditions to the query. Does the order of the condition really matter? 🤔

app/models/usage_filters.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class UsageFilters
+  attr_reader :filter_by_charge, :filter_by_group, :skip_grouping, :full_usage
```
vincent-pochet, line 14: Looks great! Could you maybe add a small documentation for each attribute to explain what it does?

app/services/fees/charge_service.rb
```diff
@@ -12,7 +20,8 @@ def initialize(
       filtered_aggregations: nil,
       apply_taxes: false,
       calculate_projected_usage: false,
-      with_zero_units_filters: true
+      with_zero_units_filters: true,
+      usage_filters: UsageFilters::NONE
```
vincent-pochet, line 24: Smart! 👍

app/services/invoices/customer_usage_service.rb
```diff
@@ -45,6 +47,8 @@ def self.with_ids(organization_id:, customer_id:, subscription_id:, apply_taxes:
     def call
       return result.not_found_failure!(resource: "customer") unless customer
       return result.not_allowed_failure!(code: "no_active_subscription") if subscription.blank?
+      full_usage_allowed = (usage_filters.filter_by_charge.present? || usage_filters.filter_by_group.present?) && subscription.plan.charges.where(prorated: true).none?
```
vincent-pochet: I would suggest to move this condition in a dedicated method, and maybe add a small comment explaining its purpose

app/models/usage_filters.rb
annvelents: yes, great idea!

PR #5033, opened 2026-02-12
fix: do not try to create zero value CreditNotes

mariohd:
Whenever we try to create a CreditNote with `credit_amount_cents: 0`, skip the credit note creation.


spec/services/credit_notes/create_from_progressive_billing_invoice_spec.rb
```diff
@@ -74,6 +74,26 @@
...
+            result.taxes_amount_cents = 0.0
+            result.precise_taxes_amount_cents = 0.0
+          end
+        end
+
+        before do
+          allow(CreditNotes::ApplyTaxesService).to receive(:call).once.and_return(cn_ats_result)
+        end
+
+        it "does nothing" do
```
aquinofb: "does not create a credit note"?

PR #5036, opened 2026-02-13
refacto(queries): Avoid overriding Query constructor and use filter instead

julienbourdeau:
I think we should strive to avoid overriding constructor of query object. The param added should usually be done via filter.
The end goal is to refacto the pagination too and avoid so much repetition.

See https://github.com/getlago/lago-api/pull/4588 for the rest (it's coming)

### DIFF
--- a/spec/queries/wallet_transaction_consumptions_query_spec.rb
+++ b/spec/queries/wallet_transaction_consumptions_query_spec.rb
@@ -6,9 +6,11 @@ RSpec.describe WalletTransactionConsumptionsQuery do
   subject(:result) do
     described_class.call(
       organization:,
-      wallet_transaction_id:,
-      direction:,
-      pagination:
+      pagination:,
+      filters: {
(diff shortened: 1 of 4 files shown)


spec/queries/wallet_transaction_consumptions_query_spec.rb
```diff
@@ -106,7 +110,7 @@
   end
 
   describe "fundings direction" do
-    let(:direction) { :fundings }
```
groyoh, line 109: Symbol do not work anymore ?

vincent-pochet (approved):

spec/queries/wallet_transaction_consumptions_query_spec.rb
julienbourdeau: No because the dry contract requires it to be a string
```
required(:wallet_transaction_id).filled(:string)
```

PR #5052, opened 2026-02-16
Chore(current usage): add params to current_usage endpoint

annvelents:
## Context

We have a case where full usage for a subscription might include ~400K fees. To balance the load, we can allow to query usage per charge or per group. Another addition - allow to query usage for the full lifetime of a subscription (from `subscription.started_at`, but only if any filter is provided)

## Description

allow to accept in the controller:
- full_usage flag
- filter_by_charge_id
- filter_by_charge_code
- filter_by_group

### DIFF
--- a/app/models/usage_filters.rb
+++ b/app/models/usage_filters.rb
@@ -15,6 +15,7 @@ class UsageFilters
 
   def self.init_from_params(params)
     group = params[:filter_by_group]
+    group = JSON.parse(group) if group.is_a?(String)
     group = group.to_unsafe_h if group.respond_to?(:to_unsafe_h)
 
     new(


app/controllers/api/v1/customers/usage_controller.rb
```diff
@@ -67,6 +72,14 @@ def past_usage_filters
         def resource_name
           "customer_usage"
         end
+
+        def find_charge_for_filter(params)
+          if params[:filter_by_charge_code].present?
+            current_organization.charges.find_by!(code: params[:filter_by_charge_code])
```
vincent-pochet: Maybe this might not work if the plan was overridden right?

annvelents: 😖😖😖 yes, it will not. Thank you!

app/services/invoices/customer_usage_service.rb
```diff
@@ -102,13 +103,15 @@ def compute_charge_fees
         .charges
         .joins(:billable_metric)
         .includes(:taxes, billable_metric: :organization, filters: {values: :billable_metric_filter})
-      if usage_filters.filter_by_charge.present?
-        charges = charges.where(id: usage_filters.filter_by_charge.id)
+      if usage_filters.filter_by_charge_id.present?
+        charges = charges.where(id: usage_filters.filter_by_charge_id)
+      elsif usage_filters.filter_by_charge_code.present?
+        charges = charges.where(code: usage_filters.filter_by_charge_code)
```
annvelents: is it a good idea to have where or should we fail with 404?

PR #5070, opened 2026-02-18
feat(wallet): Add advisory lock for WalletTransactions::VoidService

D1353L:
## Context
Related to https://github.com/getlago/lago-api/pull/4987
Current retry logic for wallets can lead to issues when balance becomes negative. It's not safe to repeat withdrawal amounts from wallet when we apply credits. Available amount on a wallet might less because of concurrent updates and we need also to recalculate wallet transaction and other related records.
It'll be safer in cases when cannot safe wallet to rollback everything and try again whole generation.
That in it's turn might hurt our API/GQL customers, since current approach of optimistic locking allows to start concurrent updates quickly, also quickly returning error to one of them which won't be ideal behaviour.

Using advisory lock should work better to API/GQL customers, since it will lock early and wait for lock release, rather quickly failing. In worst scenarios we'll increase our response time for certain action but should avoid returning a lot of errors "try later".

The `WalletTransactions::VoidService` is reachable only in sync (API/GraphQL) so there is no need to add `retry_on Customers::FailedToAcquireLock`.

## Description

Add advisory lock to `WalletTransactions::VoidService`.

### DIFF
--- a/app/services/wallet_transactions/void_service.rb
+++ b/app/services/wallet_transactions/void_service.rb
@@ -42,6 +42,9 @@ module WalletTransactions
               inbound_wallet_transaction_id: inbound_wallet_transaction&.id
             )
           end
+
+          Wallets::Balance::DecreaseService.new(wallet:, wallet_transaction:).call
+          result.wallet_transaction = wallet_transaction
         end
       end
 


app/jobs/invoices/prepaid_credit_job.rb
groyoh, line 25: Not related to this PR but this should be wrapped in a service with a transaction.

app/jobs/clock/terminate_ended_subscriptions_job.rb
groyoh: We may want to rescue here instead and let the clock job retry it later or retry manually here. This would avoid blocking other subscription termination.

app/jobs/invoices/refresh_draft_job.rb
```diff
@@ -10,6 +10,8 @@ class RefreshDraftJob < ApplicationJob
       end
     end
 
+    retry_on Customers::FailedToAcquireLock, ActiveRecord::StaleObjectError, attempts: MAX_LOCK_RETRY_ATTEMPTS, wait: random_lock_retry_delay
```
groyoh: Why is this needed ?

app/jobs/subscriptions/terminate_job.rb
```diff
@@ -10,6 +10,8 @@ class TerminateJob < ApplicationJob
       end
     end
 
+    retry_on Customers::FailedToAcquireLock, ActiveRecord::StaleObjectError, attempts: MAX_LOCK_RETRY_ATTEMPTS, wait: random_lock_retry_delay
```
groyoh: I believe it might be worth having a `retry_on_concurrency_error` method (name TBD).

app/services/wallet_transactions/void_service.rb
```diff
@@ -22,22 +22,24 @@ def call
...
-          transaction_type: :outbound,
-          status: :settled,
-          settled_at: Time.current,
-          transaction_status: :voided,
-          **transaction_params
-        ).wallet_transaction
-
-        if wallet.traceable?
-          TrackConsumptionService.call!(outbound_wallet_transaction: wallet_transaction)
+        Customers::LockService.call(customer:, scope: :prepaid_credit) do
```
groyoh, line 27: I believe we should reload the wallet once the lock is acquired to have the latest wallet value.

spec/jobs/clock/terminate_ended_subscriptions_job_spec.rb
```diff
@@ -54,4 +51,31 @@
...
+      [ActiveRecord::StaleObjectError.new("Attempted to update a stale object: Wallet."), 25]
+    ].each do |error, attempts|
+      error_class = error.class
+
+      context "when a #{error_class} error is raised" do
+        before do
+          allow(Subscriptions::TerminateService).to receive(:call).and_raise(error)
+        end
+
+        it "raises a #{error_class.class.name} error and retries" do
```
groyoh:
```suggestion
        it "raises a #{error_class.name} error and retries" do
```

app/jobs/invoices/refresh_draft_job.rb
D1353L: There is a call chain: `Invoices::RefreshDraftJob` -> `Invoices::RefreshDraftService` -> `Invoices::CalculateFeesService` -> `Credits::ProgressiveBillingService` -> `CreditNotes::CreateFromProgressiveBillingInvoice` -> `CreditNotes::CreateService` -> `WalletTransactions::VoidService`

D1353L: I did double-check: seems the VoidService is not reachable from the RefreshDraftJob due to a different invoice type. So this retry isn't needed

app/jobs/clock/terminate_ended_subscriptions_job.rb
groyoh: Let's actually move this to a `TerminateSubscriptionJob`.

groyoh: Let's schedule the job directly instead of calling service. There's not reason to run the first try synchronously 😉

app/jobs/subscriptions/terminate_ended_subscription_job.rb
```diff
@@ -0,0 +1,22 @@
...
+        :billing
+      else
+        :default
+      end
+    end
+
+    retry_on Customers::FailedToAcquireLock, ActiveRecord::StaleObjectError, attempts: MAX_LOCK_RETRY_ATTEMPTS, wait: random_lock_retry_delay
+
+    def perform(subscription:)
+      Subscriptions::TerminateService.call(subscription:)
```
groyoh:
```suggestion
      Subscriptions::TerminateService.call!(subscription:)
```

app/jobs/subscriptions/terminate_ended_subscription_job.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+#
+# This job is used for async retries in Clock:TerminateEndedSubscriptionsJob
+
+module Subscriptions
+  class TerminateEndedSubscriptionJob < ApplicationJob
+    queue_as do
```
groyoh: There's no reason to have this in billing queue IMO. It will schedule the billing job in the billing queue after terminating the subscription.

groyoh: Let's also add job uniqueness to prevent running it twice.

app/jobs/clock/terminate_ended_subscriptions_job.rb
D1353L: Just one concern about possible queue congestion. With the new approach there will be an additional bunch of jobs in the `default` queue every hour (not sure if the amount is big). Do you think it's the case?

app/jobs/subscriptions/terminate_job.rb
D1353L: There is a suggestion to keep retry_on explicitly in each job so it's always in front of the eyes

spec/services/wallet_transactions/void_service_spec.rb
```diff
@@ -663,5 +663,33 @@
...
+        end
+      end
+
+      context "when the lock is acquired" do
+        let(:lock_released_after) { 0.1.seconds }
+
+        it "voids the wallet transaction successfully" do
+          expect { subject }.not_to raise_error
+        end
+      end
```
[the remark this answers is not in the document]
D1353L: Increased values for lock_released_after and ACQUIRE_LOCK_TIMEOUT

app/jobs/subscriptions/terminate_ended_subscription_job.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Used for async termination in Clock::TerminateEndedSubscriptionsJob
+
+module Subscriptions
+  class TerminateEndedSubscriptionJob < ApplicationJob
```
groyoh, line 8:
```suggestion
  # Handles async termination of ended subscriptions from `Clock::TerminateEndedSubscriptionsJob`.
  # Intentionally on the `default` queue: this job only triggers termination which schedules
  # billing separately — it doesn't perform billing itself, so it shouldn't compete
  # with billing jobs on the :billing queue.
  class TerminateEndedSubscriptionJob < ApplicationJob
```

app/jobs/subscriptions/terminate_ended_subscription_job.rb
```diff
@@ -0,0 +1,13 @@
...
+    retry_on Customers::FailedToAcquireLock, ActiveRecord::StaleObjectError, attempts: MAX_LOCK_RETRY_ATTEMPTS, wait: random_lock_retry_delay
+
+    unique :until_executed, on_conflict: :log
+
+    def perform(subscription)
+      Subscriptions::TerminateService.call!(subscription:)
+    end
```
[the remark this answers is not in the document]
D1353L: The `default` queue is intentional

PR #5071, opened 2026-02-18
feat(security logs): add device info

nepalez:
Populate `device_info` in security logs from GraphQL requests.                                                                                     
                                                        
## Device Info Parsing                                                                                                                             
                                                                         
`Utils::DeviceInfo.parse(request)` extracts device metadata from the `User-Agent` header                                                           
using the `device_detector` gem: browser, OS, device type, and IP address.                                                                         

## Context Propagation

`CurrentContext.device_info` is set in `GraphqlController#execute` and used as a fallback
in `Utils::SecurityLog` when `device_info` is not passed explicitly.
This populates `device_info` in all security logs from GraphQL context without changing any existing callers.

### DIFF
--- a/db/seeds/07_security_logs.rb
+++ b/db/seeds/07_security_logs.rb
@@ -13,12 +13,20 @@ end
 
 organization = Organization.find_by!(name: "Hooli")
 user = organization.memberships.first!.user
+device_info = {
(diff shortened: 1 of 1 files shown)


app/controllers/graphql_controller.rb
```diff
@@ -32,6 +32,8 @@ def execute
       permissions: (current_membership || Permission)&.permissions_hash
     }
 
+    CurrentContext.device_info = Utils::DeviceInfo.parse(request)
```
murparreira, line 35: Do you know the speed of this?

nepalez: ~0.2ms (200μs) on my local machine

PR #5074, opened 2026-02-18
Feat(invoicing): add setting to skip invoice PDF

annvelents:
## Context

As not all organizations need invoice PDFs relying on usage calculation or credits consumption, it makes sense to skip option of generating the PDF itself

## Description

- added skip_invoice_pdf to billing_configuration on billing_entity
- act on skip_invoice_pdf when generating an invocie

## discussed with product:
- skip_inovice_pdf -> skip_automatic_pdf_generation: ['invoices'] (later credit notes and payments will be added)

### DIFF
--- a/app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
+++ b/app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
@@ -75,7 +75,7 @@ module Invoices
         Utils::SegmentTrack.invoice_created(invoice)
         deliver_webhooks
         Utils::ActivityLog.produce(invoice, "invoice.created")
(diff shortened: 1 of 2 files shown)


app/jobs/invoices/generate_pdf_and_notify_job.rb
```diff
@@ -2,6 +2,7 @@
 
 module Invoices
   class GeneratePdfAndNotifyJob < ApplicationJob
+    # TODO: I didn't find where this service is being called from, can we delete it?
```
vincent-pochet, line 5: It looks like we can indeed. Maybe it was kept to avoid breaking during a release phase 🤔

PR #5076, opened 2026-02-19
fix(active-job): Prevent calling `ActiveJob.perform_all_later` on unique jobs

groyoh:
## Context

`ActiveJob.perform_all_later` bypasses the `before_enqueue` callbacks used by `activejob-uniqueness` to acquire uniqueness locks. As a result, unique jobs scheduled through `perform_all_later` silently skip their uniqueness checks, which can lead to duplicate job execution.

Nothing in the codebase prevented this footgun: a developer could either call `ActiveJob.perform_all_later` directly or stub the wrapper in specs and the uniqueness guarantee would silently disappear.

## Description

This adds a runtime guard and two RuboCop cops to make bulk-enqueueing safe for unique jobs.

A new `ApplicationJob.perform_all_later` wraps `ActiveJob.perform_all_later` and raises `ArgumentError` when any job in the batch has uniqueness enabled (i.e. defines a `lock_strategy_class`). All existing call sites were updated to go through `ApplicationJob.perform_all_later`.

Two custom cops back the runtime check:
- `Lago/ActiveJobPerformAllLater` flags direct calls to `ActiveJob.perform_all_later` and points to `ApplicationJob.perform_all_later`.
- `Lago/StubPerformAllLater` flags stubbing `ApplicationJob.perform_all_later` in specs, since stubbing would silence the runtime guard. Specs should stub `ActiveJob.perform_all_later` or use `have_enqueued_job` matchers instead.

## Test plan

- [x] `spec/jobs/application_job_spec.rb` covers the runtime guard for unique and non-unique jobs.
- [x] `spec/lib/cops/active_job_perform_all_later_cop_spec.rb` covers the cop flagging direct `ActiveJob.perform_all_later` calls.
- [x] `spec/lib/cops/stub_perform_all_later_cop_spec.rb` covers the cop flagging `ApplicationJob.perform_all_later` stubs.
- [x] Existing specs for updated call sites still pass.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -27,6 +27,19 @@ class ApplicationJob < ActiveJob::Base
     end
   end
 
+  # This method wraps ActiveJob.perform_all_later with a runtime check to ensure
+  # that none of the jobs have uniqueness enabled, as perform_all_later bypasses
+  # the before_enqueue callbacks used by activejob-uniqueness.
+  #
+  def self.perform_all_later(jobs)
+    unique_jobs = jobs.select { |job| job.class.lock_strategy_class }
+    if unique_jobs.any?
+      raise ArgumentError, "perform_all_later is not compatible with unique jobs: #{unique_jobs.map { |j| j.class.name }.uniq.join(", ")}"
+    end
+
+    ActiveJob.perform_all_later(jobs) # rubocop:disable Lago/ActiveJobPerformAllLater
+  end
+
   # This method is a generic proc for specifying random wait between retries
   # Usage:
   # retry_on ExceptionClass, attempts: 5, wait: random_delay(16)
--- a/spec/services/usage_monitoring/process_organization_subscription_activities_service_spec.rb
+++ b/spec/services/usage_monitoring/process_organization_subscription_activities_service_spec.rb
@@ -11,10 +11,12 @@ RSpec.describe UsageMonitoring::ProcessOrganizationSubscriptionActivitiesService
       create_list(:subscription_activity, 3, organization:, enqueued: false)
       create_list(:subscription_activity, 2, organization:, enqueued: true)
 
-      expect { service.call }
+      result = nil
+      expect { result = service.call }
         .to have_enqueued_job(UsageMonitoring::ProcessSubscriptionActivityJob).exactly(3).times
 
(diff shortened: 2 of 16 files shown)


app/jobs/application_job.rb
```diff
@@ -27,6 +27,19 @@ def self.perform_after_commit(...)
...
+  # the before_enqueue callbacks used by activejob-uniqueness.
+  #
+  def self.perform_all_later(jobs)
+    unique_jobs = jobs.select { |job| job.class.lock_strategy_class }
+    if unique_jobs.any?
+      raise ArgumentError, "perform_all_later is not compatible with unique jobs: #{unique_jobs.map { |j| j.class.name }.uniq.join(", ")}"
+    end
+
+    ActiveJob.perform_all_later(jobs) # rubocop:disable Lago/ActiveJobPerformAllLater
+  end
```
julienbourdeau, line 41: Maybe this could be done only non production env


```suggestion
  unless Rails.env.production?
    #...
  end
  
  ActiveJob.perform_all_later(jobs) # rubocop:disable Lago/ActiveJobPerformAllLater
```

groyoh: That's a good point. To be honest, I'd rather it breaks in prod as well. That way, 1) we know there are tests missing 2) we will prevent unwanted duplicates.

WDYT ?

vincent-pochet (approved):

spec/services/usage_monitoring/process_organization_subscription_activities_service_spec.rb
```diff
@@ -7,23 +7,14 @@
...
-      end
-
-      expect(result).to be_success
-      expect(result.nb_jobs_enqueued).to eq(3)
+      expect { service.call }
+        .to have_enqueued_job(UsageMonitoring::ProcessSubscriptionActivityJob).exactly(3).times
 
+      expect(service.call).to be_success
       expect(organization.subscription_activities.where(enqueued: true).count).to eq(5)
       expect(organization.subscription_activities.where(enqueued: false).count).to eq(0)
```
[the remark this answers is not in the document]
groyoh: Fixed in 980232d63 — capturing the result from the single invocation inside the `have_enqueued_job` block and asserting on that.

PR #5079, opened 2026-02-19
feat(multi-payments): add stripe migration scripts for multiple payment methods

lovrocolic:
## Context

Currently customers can have attached only one payment method.

## Description

This PR is the last step for the `multiple-payment-methods` feature. It adds migration for populating payment methods on existing customers.

### DIFF
--- a/app/jobs/database_migrations/backfill_stripe_payment_method_card_details_job.rb
+++ b/app/jobs/database_migrations/backfill_stripe_payment_method_card_details_job.rb
@@ -27,8 +27,7 @@ module DatabaseMigrations
               SELECT 1 FROM payments p
               WHERE p.payment_provider_customer_id = pm.payment_provider_customer_id
                 AND p.provider_payment_method_id = pm.provider_method_id
-                AND p.provider_payment_method_data IS NOT NULL
-                AND p.provider_payment_method_data != '{}'::jsonb
+                AND p.provider_payment_method_data->>'last4' IS NOT NULL
             )
           LIMIT #{BATCH_SIZE}
         ),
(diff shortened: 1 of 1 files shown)


lib/tasks/migrations/payment_methods.rake
```diff
@@ -0,0 +1,57 @@
...
+    else
+      pp "  -> Nothing to do ✅"
+    end
+
+    while count > 0
+      sleep 5
+      puts "\n#### Checking status ####"
+
+      count = ActiveRecord::Base.connection.select_value(<<~SQL).to_i
+        SELECT COUNT(*) FROM payment_provider_customers ppc
```
vincent-pochet: If I'm not wrong, this condition is the same as before. Maybe you could define it as a reusable constant. What do you think?

lovrocolic: Updated

lib/tasks/migrations/payment_methods.rake
```diff
@@ -0,0 +1,49 @@
...
+    count_sql = <<~SQL
+      SELECT COUNT(*) FROM payment_provider_customers ppc
+      WHERE ppc.type = 'PaymentProviderCustomers::StripeCustomer'
+        AND ppc.settings->>'payment_method_id' IS NOT NULL
+        AND ppc.deleted_at IS NULL
+        AND NOT EXISTS (
+          SELECT 1 FROM payment_methods pm
+          WHERE pm.payment_provider_customer_id = ppc.id
+            AND pm.provider_method_id = ppc.settings->>'payment_method_id'
+        )
```
osmarluz, line 17: @lovrocolic one thing that would be needed here is the following check on the customer. Since we don't discard the payment_provider_customers (and populate the `deleted_at`) when a payment provider is detached, this would guarantee that we wouldn't create payment_methods for detached payment providers:

```suggestion
      SELECT COUNT(*) FROM payment_provider_customers ppc
      WHERE ppc.type = 'PaymentProviderCustomers::StripeCustomer'
        AND ppc.settings->>'payment_method_id' IS NOT NULL
        AND ppc.deleted_at IS NULL
        AND EXISTS (
          SELECT 1 FROM customers c
          WHERE c.id = ppc.customer_id
            AND c.payment_provider IS NOT NULL
        )
        AND NOT EXISTS (
          SELECT 1 FROM payment_methods pm
          WHERE pm.payment_provider_customer_id = ppc.id
            AND pm.provider_method_id = ppc.settings->>'payment_method_id'
        )
```

PR #5084, opened 2026-02-19
feat(payment-methods): add multiple payment methods migration for non-stripe providers

lovrocolic:
## Context

Currently customers can have attached only one payment method.

## Description

This PR is the last step for the `multiple-payment-methods` feature. It adds migration for populating payment methods on existing customers for non-stripe payment providers. Stripe related jobs are already done

### DIFF
--- a/spec/jobs/database_migrations/backfill_gocardless_payment_methods_job_spec.rb
+++ b/spec/jobs/database_migrations/backfill_gocardless_payment_methods_job_spec.rb
@@ -27,7 +27,7 @@ RSpec.describe DatabaseMigrations::BackfillGocardlessPaymentMethodsJob do
 
       payment_method = customer.payment_methods.first
       expect(payment_method.provider_method_id).to eq("mandate_123")
-      expect(payment_method.provider_method_type).to eq("mandate")
+      expect(payment_method.provider_method_type).to eq("card")
       expect(payment_method.payment_provider_customer).to eq(gocardless_customer)
       expect(payment_method.payment_provider).to eq(gocardless_provider)
       expect(payment_method.organization).to eq(organization)


lib/tasks/migrations/payment_methods.rake
```diff
@@ -46,4 +46,142 @@ namespace :migrations do
...
+    count_sql = <<~SQL
+      SELECT COUNT(*) FROM payment_provider_customers ppc
+      WHERE ppc.type = 'PaymentProviderCustomers::AdyenCustomer'
+        AND ppc.settings->>'payment_method_id' IS NOT NULL
+        AND ppc.deleted_at IS NULL
+        AND NOT EXISTS (
+          SELECT 1 FROM payment_methods pm
+          WHERE pm.payment_provider_customer_id = ppc.id
+            AND pm.provider_method_id = ppc.settings->>'payment_method_id'
+        )
```
osmarluz, line 63: @lovrocolic same idea from the [comment](https://github.com/getlago/lago-api/pull/5079#discussion_r2828275642) on #5079:


```suggestion
      SELECT COUNT(*) FROM payment_provider_customers ppc
      WHERE ppc.type = 'PaymentProviderCustomers::AdyenCustomer'
        AND ppc.settings->>'payment_method_id' IS NOT NULL
        AND ppc.deleted_at IS NULL
        AND EXISTS (
          SELECT 1 FROM customers c
          WHERE c.id = ppc.customer_id
            AND c.payment_provider IS NOT NULL
        )
        AND NOT EXISTS (
          SELECT 1 FROM payment_methods pm
          WHERE pm.payment_provider_customer_id = ppc.id
            AND pm.provider_method_id = ppc.settings->>'payment_method_id'
        )
```

app/jobs/database_migrations/backfill_adyen_payment_methods_job.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+module DatabaseMigrations
+  class BackfillAdyenPaymentMethodsJob < ApplicationJob
+    queue_as :low_priority
+    unique :until_executed
+
+    BATCH_SIZE = 1_000
+
+    def perform(organization_id = nil, batch_number = 1)
```
mariohd, line 10: the `batch_number` is not being used anywhere, right?

lovrocolic: Actually it is used on the line 67

mariohd: Yes, but it's just passing to a new job call for no reason.

I mean, its just for internal control so we know how many succeeded?

lovrocolic: I guess it is just for `unique :until_executed` not to raise error if we call the same job second time

mariohd: makes sense now! thanks!

PR #5087, opened 2026-02-20
fix(alerts): handle alert with only recurring thresholds

julienbourdeau:
## Context

Alerts can have one time thresholds and recurring thresholds.
Currently, **the UI won't let you set only a recurring** thresholds. This is not handled by the alerting system so it breaks when processing alerts with only a recurring thresholds.

Unfortunately, it's possible to set an alert with no thresholds via the API.

## Description

This PR allows alert to have only recurring thresholds.
Alternatively, we could add validation to disallow alert to have only a reucrring threshold. I'm can't find any reason to prevent it tho.

Note that there are currently 0 alerts in production without only recurring thresholds.

## Screesnhots

CleanShot 2026-02-20 at 08 42 08@2x
CleanShot 2026-02-20 at 08 44 42@2x

### DIFF
--- a/app/models/usage_monitoring/alert.rb
+++ b/app/models/usage_monitoring/alert.rb
@@ -76,10 +76,13 @@ module UsageMonitoring
     def find_thresholds_crossed_increasing(current)
       crossed = []
       return crossed if current <= previous_value
-      return crossed if current < one_time_thresholds_values.first
 
-      if previous_value < one_time_thresholds_values.last
-        crossed += one_time_thresholds_values.filter { it > previous_value && it <= current }
+      if one_time_thresholds_values.present?
+        return crossed if current < one_time_thresholds_values.first
+
+        if previous_value < one_time_thresholds_values.last
+          crossed += one_time_thresholds_values.filter { it > previous_value && it <= current }
+        end
       end
 
       crossed += find_recurring_thresholds_crossed_increasing(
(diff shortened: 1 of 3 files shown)


app/models/usage_monitoring/alert.rb
```diff
@@ -51,10 +51,13 @@ def self.sti_name
     def find_thresholds_crossed(current)
       crossed = []
       return crossed if current <= previous_value
-      return crossed if current < one_time_thresholds_values.first
 
-      if previous_value < one_time_thresholds_values.last
-        crossed += one_time_thresholds_values.filter { it > previous_value && it <= current }
+      if one_time_thresholds_values.present?
+        return crossed if current < one_time_thresholds_values.first
```
groyoh, line 81: I'm not too familiar with threshold but here we'll skip the `find_recurring_thresholds_crossed` as well. Is that expected ?

app/models/usage_monitoring/alert.rb
groyoh: I'd rather make the `initial` and `step` fallback explicit:

```suggestion
highest_threshold = one_time_thresholds_values.last || 0
if recurring_threshold
crossed += find_recurring_thresholds_crossed(
previous_value, current, recurring_threshold.value, highest_threshold
) 
end
```

Or can the `value` actually be `nil` ? 🤔

app/models/usage_monitoring/alert.rb
julienbourdeau: yes, because if there are one=tome AND recurring thresholds, the recurring only start after the last one-time.
When there is no onetime, recurring start from 0.

app/models/usage_monitoring/alert.rb
julienbourdeau: no, the value shouldn't be nil.
I defintely agree with adding `if recurring_thresholds` block 👍 I'll have a look.

vankiru: Agree with Yohan here about wrapping the call in a `if recurring_thresholds`. Otherwise lgtm 👍

PR #5089, opened 2026-02-20
feat(wallets): Add wallet traceability feature flag

groyoh:
## Context

The wallet traceability feature uses a `traceable` boolean on wallets which defines whether the feature is enabled on not per wallet (default `false` in DB). We want to be able to gradually rollout this feature per organization. 

## Description

This adds a `wallet_traceability` feature flag that, when enabled on an organization, sets `traceable: true` on wallets created for that organization. Without the flag, wallets remain non-traceable by default.


groyoh:
Thanks @D1353L 👍

PR #5090, opened 2026-02-20
Fix(entitlements): return entitlements for pending subscriptions

annvelents:
## Context

Since subscriptions can share external_id, but there might be only one active subscription with defined external_id, we should provide mechanisms to query related records for subscriptions in different statuses, not only :active

## Description

- updated entitlements controller
- updated subscriptions/base_controller

### DIFF
--- a/app/controllers/api/v1/subscriptions/lifetime_usages_controller.rb
+++ b/app/controllers/api/v1/subscriptions/lifetime_usages_controller.rb
@@ -5,7 +5,7 @@ module Api
     module Subscriptions
       class LifetimeUsagesController < Api::BaseController
         def show
-          # TODO: should we consider statuses here?
+          # Note: lifetime usage is counted against all sub upgrades-downgrades
           lifetime_usage = current_organization.subscriptions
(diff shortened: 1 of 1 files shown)


app/controllers/api/v1/subscriptions/lifetime_usages_controller.rb
```diff
@@ -5,6 +5,7 @@ module V1
     module Subscriptions
       class LifetimeUsagesController < Api::BaseController
         def show
+          # TODO: should we consider statuses here?
```
julienbourdeau: For consistancy I'd say yes.

lessentiersdupaty (approved):

app/controllers/api/v1/subscriptions/lifetime_usages_controller.rb
annvelents: checked the code: lifetime usage is shared between ALL the subscriptions sharing external_id, so we should not do this check here 🙌

vincent-pochet (approved):

PR #5095, opened 2026-02-20
Feat/wallets/not use grouping when not needed

annvelents:
## Context

When on a charge we have a pricing_group_key that is used only to track different usage, there is no need to respect this grouping when recalculating wallet balance.

## Description

- added a condition on a subscription if we can skip grouping
- respect skip grouping to skip pricing_group_keys


app/models/charge.rb
```diff
@@ -90,6 +90,10 @@ def included_in_next_subscription?(subscription)
     next_subscription_charges.pluck(:billable_metric_id).include?(billable_metric_id)
   end
 
+  def pricing_group_agnostic?
+    charge_model == "standard" && filters.empty?
```
annvelents, line 94: TODO: this was my assumption based on checking the code + playing around. Ask product if it's correct

PR #5096, opened 2026-02-20
perf(cache): Excluding zero units/amounts/events fees from cache

groyoh:
## Context

We heavily rely on caching for improving the current usage performance. We do so by caching the fees as aggregating the events for fee calculating can be quite costly and impactful on the database resource.

While investigating, we noticed that the fees are cached even though they wouldn't be persisted for invoicing (zero unit/amount/event fees).

## Description

This pull request introduces a new feature flag to optimize the caching of non-persistable (zero-value) fees. The main focus is to prevent storing unnecessary zero-value fees in the cache, when the new feature flag is enabled, while still rendering those fees to avoid any breaking behavior on the API.

This adds a whole set of tests to ensure non-regression.

### DIFF
--- a/spec/services/billable_metrics/aggregations/base_service_spec.rb
+++ b/spec/services/billable_metrics/aggregations/base_service_spec.rb
@@ -53,5 +53,32 @@ RSpec.describe BillableMetrics::Aggregations::BaseService do
         expect(null_result.count).to eq(0)
       end
     end
+
+    context "with apply_aggregation and grouped_by_keys" do
+      let(:args) { {grouped_by_keys: %w[region], apply_aggregation: true} }
+
+      it "wraps a null result inside aggregations" do
+        expect(null_result.aggregations.size).to eq(1)
+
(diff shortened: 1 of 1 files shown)


app/services/billable_metrics/aggregations/base_service.rb
```diff
@@ -33,6 +33,16 @@ class BaseService < ::BaseService
       ]
       PerEventAggregationResult = BaseResult[:event_aggregation]
 
+      def self.empty_result(result = BaseService::Result.new, grouped_by_keys: nil)
+        empty_result = result
```
vincent-pochet: Should we duplicate the result to avoid unexpected side effect if for some reason we decide to pass the service result instance?

groyoh: The current `empty_result` below actually modifies the service result so I would wager it's okay like this as we want to modify the provided result. We'd need to refactor that logic otherwise.

PR #5143, opened 2026-03-03
remove not needed join

annvelents:
## Context

we can get rid of joining customers when searching by customer_external_id, because joining whole table of subscriptions with whole table of customers is a heavy operation. Having one separate request for it would make more sense

### DIFF
--- a/app/queries/subscriptions_query.rb
+++ b/app/queries/subscriptions_query.rb
@@ -74,7 +74,7 @@ class SubscriptionsQuery < BaseQuery
 
   def with_external_customer(scope)
     customers = Customer.where(external_id: filters.external_customer_id)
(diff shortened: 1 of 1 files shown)


app/queries/subscriptions_query.rb
```diff
@@ -59,7 +62,8 @@ def search_params
   end
 
   def with_external_customer(scope)
-    scope.joins(:customer).where(customers: {external_id: filters.external_customer_id})
+    customers = Customer.where(external_id: filters.external_customer_id)
+    scope.where(customer_id: customers.ids)
```
rsempe:
```suggestion
    scope.where(customer_id: customers.select(:id))
```

rsempe (approved):

PR #5165, opened 2026-03-06
perf(optimize customers lookup)

nepalez:
> This PR is provided as is for the review.
> Its preparation part was also separated into the https://github.com/getlago/lago-api/pull/5196 that **SHOULD BE RELEASED IN ADVANCE**

## Context

The `InvoicesQuery` searches customers by `ILIKE '%term%'` across 5 columns (`name`, `firstname`, `lastname`, `external_id`, `email`). Leading-wildcard `ILIKE` cannot use B-tree indexes, so PostgreSQL falls back to a sequential scan of the customers within the organization.

Invoice-specific search (`id`, `number`) does not need a GIN index — the query is already narrowed by `organization_id`, `status`, and other B-tree-indexed filters, making the `ILIKE` on two short fields cheap enough.

The goal is to add a `pg_trgm` GIN-indexed `search_text` column to the `customers` table, making substring search on customer fields indexable.

The `search_term` parameter is used by three API surfaces:
- REST API (`InvoiceIndex` concern) via `params[:search_term]`
- GraphQL (`Resolvers::InvoicesResolver`, `Resolvers::Customers::InvoicesResolver`, `Resolvers::CustomerPortal::InvoicesResolver`) via `search_term` argument

No changes to the external API contract are required — `search_term` remains a single string parameter. The change is internal: how the query is executed.

## General design

- Add a `search_text` column to the `customers` table containing a concatenation of customer-specific search fields (`name`, `firstname`, `lastname`, `external_id`, `email`).
- Create a composite GIN index `(organization_id, search_text gin_trgm_ops)` using the `btree_gin` extension, so that both the organization scope and the trigram search are handled in a single index scan.
- Replace ransack `_cont` predicates for customer fields in the `with_customers_filter` method with a `WHERE` clause using `ILIKE` against the `search_text` column.
- Invoice-specific search (`id_cont`, `number_cont`) remains unchanged — ransack handles it efficiently with existing B-tree indexes.
- Populate `search_text` via a database trigger to ensure consistency regardless of the write path (Rails, background jobs, direct SQL).

### Table changes

| Table | Column | Type | Required | Default value |
|-------|--------|------|----------|---------------|
| `customers` | `search_text` | `text` | No | — |

### Index changes

- Enable `pg_trgm` and `btree_gin` extensions
- Add `CREATE INDEX ... ON customers USING gin (organization_id, search_text gin_trgm_ops)`

### Data migration

- Backfill `customers.search_text` from `CONCAT_WS(' ', name, firstname, lastname, external_id, email)`
- Create a database trigger to keep `search_text` updated on `INSERT` and `UPDATE`

### API, Permissions, UI, and other changes

No changes.

## Future optimizations

Once `customers.search_text` is indexed, the same column and index can be reused to optimize customer search in `CreditNotesQuery`, `PaymentsQuery`, and `SubscriptionsQuery`, which currently use the same ransack `_cont` pattern with a JOIN on `customers`.

# QA

The search behavior must remain identical across all surfaces that use `search_term`. No UI changes are expected — only the underlying query execution changes.

## Invoices list page (app)

- [x] Search by invoice number (full and partial) — verify correct results
- [x] Search by customer name — verify invoices of matching customers appear
- [x] Search by customer email — same
- [x] Search by customer external ID — same
- [x] Search by customer firstname / lastname — same
- [x] Search by partial UUID (invoice id) — verify the matching invoice appears
- [x] Empty search field — verify all invoices are returned (no filtering)
- [x] Search combined with filters (status, payment status, date range, currency, amount, billing entity) — verify both search and filters apply correctly
- [x] Search with pagination — verify results remain consistent across pages
- [x] Export invoices with active search term — verify the exported data matches the filtered list

## Customer invoices tab (app)

- [x] Search by invoice number within a specific customer's invoices tab — verify results
- [x] Verify that customer-field search is not applied (only invoice fields: `id`, `number`)


db/migrate/20260306180435_backfill_customer_search_text.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class BackfillCustomerSearchText < ActiveRecord::Migration[8.0]
+  def up
+    safety_assured do
+      execute <<-SQL
+        UPDATE customers
```
vankiru: I believe we have around ~4.5M customers on the EU servers and ~1M on the US servers. Do you think updating all the records would take a long time (especially for EU, I think US should be fine), or should we move it to a background job?

nepalez: You're right. I was about to shoot myself into the foot, thank you!

I've fixed this in 2 directions

* at first, I separated the preparation step (it adds the column/trigger along with the rake task/job to backfill the data which won't be used yet, but be kept actual thanks to the trigger)
* then I kept the migration for a feature protection (to ensure the data are prepared), but split it into batches and run it outside the DDL transaction. It would be a long-running migration (for customers self-hosting their apps and who haven't run the job at the previous step), but not blocking the whole table for too long. Finally the preparation code is removed in the same commit as it became dead.

annvelents:
@vankiru , this is included in your improvements PR, right?

vankiru:
> @vankiru , this is included in your improvements PR, right?

@annvelents There is some overlap with my PR – we both use `pg_trgm` and GIN indexes. The main difference is Andrew went with a single combined field for all values, while my PR adds multiple GIN indexes.

It's a pretty interesting approach – fewer indexes, less overhead on inserts/updates, and just one search term without OR conditions. I'm gonna benchmark this idea too and see which one gives us better optimization 👍

PR #5173, opened 2026-03-10
feat(order_forms): Add Quotes

rinasergeeva:
## Context

This PR lays the foundation for the upcoming _Quote_ feature, a premium offering that lets sales and finance teams formalize commercial intent with their customers before it becomes actual billing.

A quote represents one of three commercial moves:

- creating a new subscription
- amending an existing one
- selling a one-off

It goes through a review/approval workflow before any subscription or invoice is created on the customer's account.

The feature is hidden behind `License.premium?` and a new `order_forms` organization feature flag, and there is no public surface (API, GraphQL, webhooks) in this PR. Follow-up PRs will expose it.

### Domain model

```mermaid
classDiagram
    class Quote {
      +uuid id
      +uuid customer
      +uuid subscription
      +string number
      +enum order_type
    }
    class QuoteVersion {
      +uuid id
      +uuid quote_id
      +bigint version
      +enum status
      +jsonb billing_items
      +jsonb content
    }
    class QuoteOwner {
      +bigint id
      +uuid quote_id
      +uuid user_id
    }
    class User {
      +uuid id
      +string email
    }

    Quote "1" -- "0..*" QuoteVersion : has_many (versions)
    QuoteVersion --> Quote : belongs_to (quote_id)
    Quote "1" o-- "0..*" QuoteOwner : quote_owners
    User "1" o-- "0..*" QuoteOwner : quote_owners
    QuoteOwner --> User : belongs_to (user_id)
    QuoteOwner --> Quote : belongs_to (quote_id)
    Quote "1" -- "1" QuoteVersion : has_one (current_version)
    User "0..*" -- "0..*" Quote : owners (through QuoteOwner)
```

### Quotes

A quote is the commercial container shared with the customer. It is identified by a human-readable number of the form `QT-YYYY-####`, sequenced per organization so each tenant gets its own clean numbering. The same quote can be co-owned by multiple internal users (the sales reps, AEs, finance contacts working the deal), so collaboration on a single deal does not require duplicate records.

When a quote is created, the originating commercial intent (`subscription_creation`, `subscription_amendment`, `one_off`) is captured up front. Amendments must point at an existing subscription on the same customer; the service layer enforces that the customer/subscription actually belong to the organization driving the quote.

### Versions and lifecycle

Customers iterate. A quote therefore has a stack of versions, each carrying its own `billing_items` payload and free-form `content`, and only one of them can be _active_ (`draft` or `approved`) at any time. That invariant is enforced at the database level so two parallel edits cannot both end up active.

A version moves through a small lifecycle:

- `draft` - the working copy, freely editable, shareable with the customer via a unique `share_token`.
- `approved` - the customer has accepted the offer; the version becomes immutable and timestamped.
- `voided` - the version is no longer in play, with an explicit reason (`manual`, `superseded`, `cascade_of_expired`, `cascade_of_voided`) so the team always knows _why_ a quote went away.

Cloning is the supersession path: starting from the current draft, the service voids it with reason `superseded` and produces a fresh draft pre-filled with its content, so reps can iterate on a proposal without losing history. Cloning is refused once a version has been approved, to prevent accidentally overwriting a deal the customer already accepted.

### Premium gating and access control

Every quote-related action runs through a shared `OrderForms::Premium` gate, so the feature only lights up when both the premium license and the per-organization `order_forms` flag are on. A new `quotes` permission group covers `view`, `create`, `update`, `approve`, `clone`, and `void`, scoped to the appropriate roles (managers can act, finance can view).

### Out of scope

This PR is foundation only. No controllers, GraphQL, serializers, or webhooks. `SendWebhookJob` calls and the materialization step (`OrderForms::CreateService`, which will turn an approved version into a real subscription/invoice) are stubbed as `TODO`s and will land in follow-up PRs.

### DIFF
--- a/app/config/feature_flags.yaml
+++ b/app/config/feature_flags.yaml
@@ -29,3 +29,7 @@ payment_gated_subscriptions:
 multi_entity_billing:
   description: "Allow customers to have billing objects across multiple billing entities"
   owners: ["@lovrocolic", "@feliperodrigues", "@mariohd"]
+
+order_forms:
+  description: "Enable quotes, order forms and orders"
+  owners: ["@murparreira", "@toommz", "@rinasergeeva"]
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -75,6 +75,8 @@ class Organization < ApplicationRecord
   has_many :error_details
   has_many :dunning_campaigns
   has_many :roles
+  has_many :quotes
+  has_many :quote_versions
   has_many :activity_logs, class_name: "Clickhouse::ActivityLog"
   has_many :features, class_name: "Entitlement::Feature"
   has_many :privileges, class_name: "Entitlement::Privilege"
@@ -144,6 +146,7 @@ class Organization < ApplicationRecord
     events_targeting_wallets
     security_logs
     granular_lifetime_usage
+    order_forms
   ].freeze
 
   SECURITY_LOGS_RETENTION_DAYS = 90
--- a/app/models/quote.rb
+++ b/app/models/quote.rb
@@ -19,7 +19,10 @@ class Quote < ApplicationRecord
   has_many :owners, through: :quote_owners, source: :user, class_name: "User"
 
   has_many :versions, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
-  has_one :current_version, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
+  has_one :current_version,
+    -> { where(status: %w[draft approved]).order(sequential_id: :desc) },
+    class_name: "QuoteVersion"
+  has_one :latest_version, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
 
   enum :order_type, ORDER_TYPES,
     instance_methods: false,
--- a/app/models/quote.rb
+++ b/app/models/quote.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+class Quote < ApplicationRecord
+  include Sequenced
+
+  ORDER_TYPES = {
+    subscription_creation: "subscription_creation",
+    subscription_amendment: "subscription_amendment",
+    one_off: "one_off"
+  }.freeze
+
+  before_save :ensure_number
+
+  belongs_to :organization
+  belongs_to :customer
+  belongs_to :subscription, optional: true
+
+  has_many :quote_owners, dependent: :destroy
+  has_many :owners, through: :quote_owners, source: :user, class_name: "User"
+
+  has_many :versions, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
+  has_one :current_version, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
+
+  enum :order_type, ORDER_TYPES,
+    instance_methods: false,
+    validate: true
+
+  validates :subscription_id,
+    presence: true,
+    if: -> { order_type == "subscription_amendment" }
+
+  sequenced(
+    scope: ->(quote) { quote.organization.quotes },
+    lock_key: ->(quote) { quote.organization_id }
+  )
+
+  private
+
+  def ensure_number
+    return if number.present?
+    return if sequential_id.blank?
+
+    time = created_at || Time.current
+    formatted_sequential_id = format("%04d", sequential_id)
+    self.number = "QT-#{time.strftime("%Y")}-#{formatted_sequential_id}"
+  end
+end
+
+# == Schema Information
+#
+# Table name: quotes
+# Database name: primary
+#
+#  id              :uuid             not null, primary key
+#  number          :string           not null
+#  order_type      :enum             not null
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  customer_id     :uuid             not null
+#  organization_id :uuid             not null
+#  sequential_id   :integer          not null
+#  subscription_id :uuid
+#
+# Indexes
+#
+#  index_quotes_on_customer_id                        (customer_id)
+#  index_quotes_on_subscription_id                    (subscription_id)
+#  index_unique_quotes_on_organization_number         (organization_id,number) UNIQUE
+#  index_unique_quotes_on_organization_sequential_id  (organization_id,sequential_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (customer_id => customers.id)
+#  fk_rails_...  (organization_id => organizations.id)
+#  fk_rails_...  (subscription_id => subscriptions.id)
+#
--- a/app/models/quote.rb
+++ b/app/models/quote.rb
@@ -19,10 +19,7 @@ class Quote < ApplicationRecord
   has_many :owners, through: :quote_owners, source: :user, class_name: "User"
 
   has_many :versions, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
-  has_one :current_version,
-    -> { where(status: %w[draft approved]).order(sequential_id: :desc) },
-    class_name: "QuoteVersion"
-  has_one :latest_version, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
+  has_one :current_version, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
 
   enum :order_type, ORDER_TYPES,
     instance_methods: false,
--- a/app/models/quote_owner.rb
+++ b/app/models/quote_owner.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+class QuoteOwner < ApplicationRecord
+  belongs_to :organization
+  belongs_to :quote
+  belongs_to :user
+end
+
+# == Schema Information
+#
+# Table name: quote_owners
+# Database name: primary
+#
+#  id              :bigint           not null, primary key
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  organization_id :uuid             not null
+#  quote_id        :uuid             not null
+#  user_id         :uuid             not null
+#
+# Indexes
+#
+#  index_quote_owners_on_organization_id    (organization_id)
+#  index_quote_owners_on_user_id            (user_id)
+#  index_unique_quote_owners_on_quote_user  (quote_id,user_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (organization_id => organizations.id)
+#  fk_rails_...  (quote_id => quotes.id)
+#  fk_rails_...  (user_id => users.id)
+#
--- a/app/models/quote_version.rb
+++ b/app/models/quote_version.rb
@@ -21,9 +21,6 @@ class QuoteVersion < ApplicationRecord
   belongs_to :organization
   belongs_to :quote
 
-  has_one :order_form
-  has_one :order, through: :order_form
-
   enum :status, STATUSES,
     default: :draft,
     validate: true
--- a/app/models/quote_version.rb
+++ b/app/models/quote_version.rb
@@ -1,0 +1,94 @@
+# frozen_string_literal: true
+
+class QuoteVersion < ApplicationRecord
+  include Sequenced
+
+  STATUSES = {
+    draft: "draft",
+    approved: "approved",
+    voided: "voided"
+  }.freeze
+
+  VOID_REASONS = {
+    manual: "manual",
+    superseded: "superseded",
+    cascade_of_expired: "cascade_of_expired",
+    cascade_of_voided: "cascade_of_voided"
+  }.freeze
+
+  before_save :ensure_share_token
+
+  belongs_to :organization
+  belongs_to :quote
+
+  has_one :order_form
+  has_one :order, through: :order_form
+
+  enum :status, STATUSES,
+    default: :draft,
+    validate: true
+  enum :void_reason, VOID_REASONS,
+    instance_methods: false,
+    validate: {allow_nil: true}
+
+  validates :share_token,
+    on: :update,
+    presence: true,
+    if: -> { draft? || approved? }
+
+  validates :void_reason, :voided_at,
+    presence: true,
+    if: -> { voided? }
+
+  validates :approved_at,
+    presence: true,
+    if: -> { approved? }
+
+  sequenced(
+    scope: ->(quote_version) { quote_version.quote.versions },
+    lock_key: ->(quote_version) { quote_version.quote_id }
+  )
+
+  def version = sequential_id
+
+  private
+
+  def ensure_share_token
+    return if voided?
+
+    self.share_token ||= SecureRandom.uuid
+  end
+end
+
+# == Schema Information
+#
+# Table name: quote_versions
+# Database name: primary
+#
+#  id              :uuid             not null, primary key
+#  approved_at     :datetime
+#  billing_items   :jsonb
+#  content         :text
+#  share_token     :string
+#  status          :enum             default("draft"), not null
+#  void_reason     :enum
+#  voided_at       :datetime
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  organization_id :uuid             not null
+#  quote_id        :uuid             not null
+#  sequential_id   :integer          not null
+#
+# Indexes
+#
+#  index_quote_versions_on_organization_id             (organization_id)
+#  index_quote_versions_on_quote_id                    (quote_id)
+#  index_unique_quote_versions_on_quote_active_status  (quote_id) UNIQUE WHERE (status = ANY (ARRAY['draft'::quote_status, 'approved'::quote_status]))
+#  index_unique_quote_versions_on_quote_sequential_id  (quote_id,sequential_id) UNIQUE
+#  index_unique_quote_versions_on_share_token          (share_token) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (organization_id => organizations.id)
+#  fk_rails_...  (quote_id => quotes.id)
+#
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -14,6 +14,9 @@ class User < ApplicationRecord
   has_many :active_memberships, -> { where(status: "active") }, class_name: "Membership"
   has_many :active_organizations, through: :active_memberships, source: :organization
 
+  has_many :quote_owners, dependent: :destroy
+  has_many :quotes, through: :quote_owners
+
   validates :email, presence: true
   validates :password, presence: true
 
--- a/app/services/quote_versions/clone_service.rb
+++ b/app/services/quote_versions/clone_service.rb
@@ -25,7 +25,7 @@ module QuoteVersions
     def call
       return result.not_found_failure!(resource: "quote_version") unless quote_version
       return result.forbidden_failure! unless order_forms_enabled?(quote_version.organization)
-      return result.not_allowed_failure!(code: "inappropriate_state") unless clonable?
+      return result.forbidden_failure!(code: "inappropriate_state") unless clonable?
 
       cloned = QuoteVersion.transaction do
         void!(quote_version:)
@@ -38,6 +38,8 @@ module QuoteVersions
       result
     rescue ActiveRecord::RecordInvalid => e
       result.record_validation_failure!(record: e.record)
+    rescue ActiveRecord::RecordNotUnique
+      result.forbidden_failure!(code: "active_version_exists")
     rescue CloneError => e
       result.service_failure!(code: "clone_failed", message: e.message, error: e)
     end
@@ -47,7 +49,8 @@ module QuoteVersions
     def clonable?
       return false if quote_version.quote.versions.where(status: :approved).exists?
 
-      true
+      active_draft = quote_version.quote.versions.where(status: :draft).first
+      active_draft.nil? || active_draft.id == quote_version.id
     end
 
     def create_next_version(quote_version:)
--- a/app/services/quote_versions/clone_service.rb
+++ b/app/services/quote_versions/clone_service.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+module QuoteVersions
+  class CloneService < BaseService
+    class CloneError < StandardError
+      attr_reader :cause, :error
+
+      def initialize(cause: nil, error: nil)
+        @cause = cause
+        @error = error
+        super("QuoteVersion clone failed due to #{cause.class}")
+      end
+    end
+
+    attr_reader :quote_version
+
+    Result = BaseResult[:quote_version]
+
+    def initialize(quote_version:)
+      @quote_version = quote_version
+      super
+    end
+
+    def call
+      return result.forbidden_failure! unless License.premium?
+      return result.not_found_failure!(resource: "quote_version") unless quote_version
+      return result.forbidden_failure! unless quote_version.organization.feature_flag_enabled?(:order_forms)
+      return result.not_allowed_failure!(code: "inappropriate_state") unless clonable?
+
+      cloned = QuoteVersion.transaction do
+        void!(quote_version:)
+        create_next_version(quote_version:)
+      end
+
+      # TODO: SendWebhookJob.perform_after_commit("quote_version.cloned", cloned)
+
+      result.quote_version = cloned
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    rescue CloneError, ActiveRecord::ActiveRecordError => e
+      result.service_failure!(code: "clone_failed", message: e.message, error: e)
+    end
+
+    private
+
+    def clonable?
+      return false if quote_version.approved?
+
+      true
+    end
+
+    def create_next_version(quote_version:)
+      quote_version.dup.tap do |cloned|
+        cloned.status = :draft
+        cloned.sequential_id = nil
+        cloned.share_token = nil # will be generated on saving
+        cloned.void_reason = nil
+        cloned.voided_at = nil
+        cloned.approved_at = nil
+        cloned.save!
+      end
+    end
+
+    def void!(quote_version:)
+      return if quote_version.voided?
+
+      void_result = QuoteVersions::VoidService.new(
+        quote_version: quote_version,
+        reason: :superseded
+      ).call
+
+      raise CloneError.new(error: void_result.error, cause: void_result) unless void_result&.success?
+    end
+  end
+end
--- a/app/services/quote_versions/clone_service.rb
+++ b/app/services/quote_versions/clone_service.rb
@@ -2,6 +2,8 @@
 
 module QuoteVersions
   class CloneService < BaseService
+    include OrderForms::Premium
+
     class CloneError < StandardError
       attr_reader :source_result
 
@@ -21,9 +23,8 @@ module QuoteVersions
     end
 
     def call
-      return result.forbidden_failure! unless License.premium?
       return result.not_found_failure!(resource: "quote_version") unless quote_version
-      return result.forbidden_failure! unless quote_version.organization.feature_flag_enabled?(:order_forms)
+      return result.forbidden_failure! unless order_forms_enabled?(quote_version.organization)
       return result.not_allowed_failure!(code: "inappropriate_state") unless clonable?
 
       cloned = QuoteVersion.transaction do
--- a/app/services/quote_versions/clone_service.rb
+++ b/app/services/quote_versions/clone_service.rb
@@ -3,12 +3,11 @@
 module QuoteVersions
   class CloneService < BaseService
     class CloneError < StandardError
-      attr_reader :cause, :error
+      attr_reader :source_result
 
-      def initialize(cause: nil, error: nil)
-        @cause = cause
-        @error = error
-        super("QuoteVersion clone failed due to #{cause.class}")
+      def initialize(source_result:)
+        @source_result = source_result
+        super("QuoteVersion clone failed: #{source_result&.error&.message}")
       end
     end
 
@@ -38,14 +37,14 @@ module QuoteVersions
       result
     rescue ActiveRecord::RecordInvalid => e
       result.record_validation_failure!(record: e.record)
-    rescue CloneError, ActiveRecord::ActiveRecordError => e
+    rescue CloneError => e
       result.service_failure!(code: "clone_failed", message: e.message, error: e)
     end
 
     private
 
     def clonable?
-      return false if quote_version.approved?
+      return false if quote_version.quote.versions.where(status: :approved).exists?
 
       true
     end
@@ -54,7 +53,7 @@ module QuoteVersions
       quote_version.dup.tap do |cloned|
         cloned.status = :draft
         cloned.sequential_id = nil
-        cloned.share_token = nil # will be generated on saving
+        cloned.share_token = nil # regenerated on save
         cloned.void_reason = nil
         cloned.voided_at = nil
         cloned.approved_at = nil
@@ -70,7 +69,7 @@ module QuoteVersions
         reason: :superseded
       ).call
 
-      raise CloneError.new(error: void_result.error, cause: void_result) unless void_result&.success?
+      raise CloneError.new(source_result: void_result) unless void_result&.success?
     end
   end
 end
--- a/app/services/quote_versions/create_service.rb
+++ b/app/services/quote_versions/create_service.rb
@@ -2,6 +2,8 @@
 
 module QuoteVersions
   class CreateService < BaseService
+    include OrderForms::Premium
+
     attr_reader :quote, :params
 
     Result = BaseResult[:quote_version]
@@ -13,9 +15,8 @@ module QuoteVersions
     end
 
     def call
-      return result.forbidden_failure! unless License.premium?
       return result.not_found_failure!(resource: "quote") unless quote
-      return result.forbidden_failure! unless quote.organization.feature_flag_enabled?(:order_forms)
+      return result.forbidden_failure! unless order_forms_enabled?(quote.organization)
 
       quote_version = quote.versions.create!(
         organization: quote.organization,
--- a/app/services/quote_versions/create_service.rb
+++ b/app/services/quote_versions/create_service.rb
@@ -2,32 +2,24 @@
 
 module QuoteVersions
   class CreateService < BaseService
-    attr_reader :organization, :quote, :params
+    attr_reader :quote, :params
 
     Result = BaseResult[:quote_version]
 
-    def initialize(organization:, quote:, params: {})
-      @organization = organization
+    def initialize(quote:, params: {})
       @quote = quote
       @params = params
-
       super
     end
 
     def call
       return result.forbidden_failure! unless License.premium?
-      return result.not_found_failure!(resource: "organization") unless organization
       return result.not_found_failure!(resource: "quote") unless quote
-      return result.forbidden_failure! unless organization.feature_flag_enabled?(:order_forms)
-
-      create_params = params.slice(
-        :billing_items,
-        :content
-      )
+      return result.forbidden_failure! unless quote.organization.feature_flag_enabled?(:order_forms)
 
       quote_version = quote.versions.create!(
-        organization:,
-        **create_params
+        organization: quote.organization,
+        **params.slice(:billing_items, :content)
       )
 
       result.quote_version = quote_version
@@ -37,8 +29,6 @@ module QuoteVersions
       result
     rescue ActiveRecord::RecordInvalid => e
       result.record_validation_failure!(record: e.record)
-    rescue ActiveRecord::ActiveRecordError => e
-      result.service_failure!(code: "create_failed", message: e.message, error: e)
     end
   end
 end
--- a/app/services/quote_versions/create_service.rb
+++ b/app/services/quote_versions/create_service.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module QuoteVersions
+  class CreateService < BaseService
+    attr_reader :organization, :quote, :params
+
+    Result = BaseResult[:quote_version]
+
+    def initialize(organization:, quote:, params: {})
+      @organization = organization
+      @quote = quote
+      @params = params
+
+      super
+    end
+
+    def call
+      return result.forbidden_failure! unless License.premium?
+      return result.not_found_failure!(resource: "organization") unless organization
+      return result.not_found_failure!(resource: "quote") unless quote
+      return result.forbidden_failure! unless organization.feature_flag_enabled?(:order_forms)
+
+      create_params = params.slice(
+        :billing_items,
+        :content
+      )
+
+      quote_version = quote.versions.create!(
+        organization:,
+        **create_params
+      )
+
+      result.quote_version = quote_version
+
+      # TODO: SendWebhookJob.perform_after_commit("quote_version.created", quote_version)
+
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    rescue ActiveRecord::ActiveRecordError => e
+      result.service_failure!(code: "create_failed", message: e.message, error: e)
+    end
+  end
+end
--- a/app/services/quote_versions/create_service.rb
+++ b/app/services/quote_versions/create_service.rb
@@ -17,6 +17,7 @@ module QuoteVersions
     def call
       return result.not_found_failure!(resource: "quote") unless quote
       return result.forbidden_failure! unless order_forms_enabled?(quote.organization)
+      return result.forbidden_failure!(code: "active_version_exists") if active_version_exists?
 
       quote_version = quote.versions.create!(
         organization: quote.organization,
@@ -30,6 +31,14 @@ module QuoteVersions
       result
     rescue ActiveRecord::RecordInvalid => e
       result.record_validation_failure!(record: e.record)
+    rescue ActiveRecord::RecordNotUnique
+      result.forbidden_failure!(code: "active_version_exists")
+    end
+
+    private
+
+    def active_version_exists?
+      quote.versions.where(status: %w[draft approved]).exists?
     end
   end
 end
--- a/app/services/quote_versions/update_service.rb
+++ b/app/services/quote_versions/update_service.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module QuoteVersions
+  class UpdateService < BaseService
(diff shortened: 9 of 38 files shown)


db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+      t.boolean :order_only, null: false, default: false
+      t.boolean :backdated_billing, null: false, default: false
+      t.jsonb :execution_result
+
+      t.timestamps
+    end
+
+    create_table :order_form_catalog_references, id: false, primary_key: %i[order_form_id referenced_type referenced_id] do |t|
+      t.references :organization, null: false, foreign_key: true, type: :uuid
+      t.references :order_form, null: false, foreign_key: true, type: :uuid
```
nepalez: I think this should be composite foreign keys here `(organization_id, order_form_id) -> order_forms.(organization_id, id)` to ensure referentional integrity

db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+
+class AddOrderForms < ActiveRecord::Migration[8.0]
+  def change
+    create_enum :order_form_status, %w[draft published signed executed voided]
+    create_enum :order_form_void_reason, %w[manual expired superseded invalid]
+
+    create_table :order_forms, id: :uuid do |t|
+      # identity
+      t.references :organization, null: false, foreign_key: true, type: :uuid
+      t.references :customer, null: false, foreign_key: true, type: :uuid
```
nepalez: And here the composite fk would keep the integrity better `(organization_id, customer_id)` -> `customers.(organization_id, id)`

db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+      t.datetime :voided_at
+      # content
+      t.jsonb :billing_payload, null: false, default: {}
+      t.boolean :auto_execute, null: false, default: false
+      t.boolean :order_only, null: false, default: false
+      t.boolean :backdated_billing, null: false, default: false
+      t.jsonb :execution_result
+
+      t.timestamps
+    end
```
nepalez: We need some uniqueness constraints here, at least for the `(organization_id, sequential_id, version)` and the `(organization_id, number, version)`.

Then probably the unique index by `(organization_id, id)` instead of the plain index by `organization_id` -- it could be reused for fk constraints.

I also thought about incrementing the sequential_id and version on the database side (to simplify the app's race conditions prevention), but not sure the additional complexity on the db side worth it.

db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+      t.references :organization, null: false, foreign_key: true, type: :uuid
+      t.references :customer, null: false, foreign_key: true, type: :uuid
+      t.string :number, null: false
+      t.integer :version, null: false, default: 1
+      t.integer :sequential_id, null: false
+      # lifecycle
+      t.enum :status, enum_type: :order_form_status, null: false, default: "draft"
+      t.enum :void_reason, enum_type: :order_form_void_reason
+      t.uuid :signed_by_user_id
+      t.string :share_token
```
nepalez: Can this guy be null? Maybe in the void state only, maybe it should never be null -- even void order form could be shareable, otherwise how it could be cloned?

db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+      # identity
+      t.references :organization, null: false, foreign_key: true, type: :uuid
+      t.references :customer, null: false, foreign_key: true, type: :uuid
+      t.string :number, null: false
+      t.integer :version, null: false, default: 1
+      t.integer :sequential_id, null: false
+      # lifecycle
+      t.enum :status, enum_type: :order_form_status, null: false, default: "draft"
+      t.enum :void_reason, enum_type: :order_form_void_reason
+      t.uuid :signed_by_user_id
```
nepalez: I'd suggest the constraint requiring `signed_at` and `signed_by_user_id` to be null/non-null at the same time: `(signed_by_user_id IS NULL) = (signed_at IS NULL)`

db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+      t.string :share_token
+      t.jsonb :validation_errors
+      t.datetime :validated_at
+      t.datetime :published_at
+      t.datetime :signed_at
+      t.datetime :executed_at
+      t.datetime :expires_at
+      t.datetime :voided_at
```
nepalez: I am NOT sure we really need all constraints for the lifecycle at the database side, but probably some of them? 

For example, can the OF be executed without being signed? I'd say this particular case is a security breach because we won't know who exactly enabled the execution (no reference to the user), and this one should be prevernted via `(executed_at IS NULL or signed_at IS NOT NULL)`

db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+      t.timestamps
+    end
+
+    create_table :order_form_catalog_references, id: false, primary_key: %i[order_form_id referenced_type referenced_id] do |t|
+      t.references :organization, null: false, foreign_key: true, type: :uuid
+      t.references :order_form, null: false, foreign_key: true, type: :uuid
+      t.string :referenced_type, null: false
+      t.uuid :referenced_id, null: false
+
+      t.timestamps
```
nepalez: IMO this could be dropped because who cares? This table is just a search optimisaiton, isn't it?

db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    add_index :order_form_catalog_references, [:referenced_type, :referenced_id],
+      name: :index_order_form_catalog_references_on_referenced_type_and_id
+
+    create_table :order_form_attachments, id: :uuid do |t|
+      t.references :organization, null: false, foreign_key: true, type: :uuid
+      t.references :order_form, null: false, foreign_key: true, type: :uuid
```
nepalez: And here the composite fk would be better than 2 independent ones

db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+  def change
+    create_enum :order_form_status, %w[draft published signed executed voided]
+    create_enum :order_form_void_reason, %w[manual expired superseded invalid]
+
+    create_table :order_forms, id: :uuid do |t|
+      # identity
+      t.references :organization, null: false, foreign_key: true, type: :uuid
+      t.references :customer, null: false, foreign_key: true, type: :uuid
+      t.string :number, null: false
+      t.integer :version, null: false, default: 1
```
nepalez: It must be positive, isn't it?

db/migrate/20260304074159_add_order_forms.rb
```diff
@@ -0,0 +1,61 @@
...
+      t.integer :sequential_id, null: false
```
nepalez: And this one have to be positive (`t.check_constraint "must be positive", "sequential_id > 0"`)

app/models/order_form/attachment.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    validates :file_name, presence: true
+    validates :file_type, presence: true
+    validates :file_url, presence: true
+    validates :file_size,
+      presence: true,
+      numericality: {only_integer: true, greater_than_or_equal_to: 0}
+    validates :position,
+      presence: true,
+      numericality: {only_integer: true, greater_than_or_equal_to: 0}
```
nepalez: I'm not sure these checks should be here. Model validation are for communication with a user, not for the data protection, especially when it comes to the "technical" terms. Probably the db constraints are enough here.

app/models/order_form/catalog_reference.rb
```diff
@@ -0,0 +1,37 @@
...
+
+class OrderForm
+  class CatalogReference < ApplicationRecord
+    self.primary_key = %i[order_form_id referenced_type referenced_id]
+
+    belongs_to :order_form
+    belongs_to :organization
+
+    validates :referenced_type, presence: true
+    validates :referenced_id, presence: true
```
nepalez: And here is the same comment: if these 2 were absent, it is not the problem of the user, there is nobody to communicate it to. The database would protect the data.

app/models/order_form.rb
```diff
@@ -0,0 +1,101 @@
...
+  validates :sequential_id,
+    presence: true,
+    uniqueness: {scope: :organization_id},
+    numericality: {only_integer: true, greater_than: 0},
+    if: :persisted?
+  validates :version,
+    presence: true,
+    uniqueness: {scope: %i[organization_id sequential_id]},
+    numericality: {only_integer: true, greater_than: 0},
+    if: :persisted?
```
nepalez: There are not a user-provided values, so we don't need to report the error of this kind to any customer. The errors of this sort are just bugs in the application.

As for the uniqueness, the app-level protection is not enough, race conditions cannot be prevented at this layer, so it is useless (it could be useful if and only if it is the user who provides the data so it would be necessary to tell him they aren't unique, but here the case is different)

app/models/order_form.rb
```diff
@@ -0,0 +1,101 @@
...
+  enum :void_reason, VOID_REASONS.index_with(&:to_s), instance_methods: false, validation: true
+
+  sequenced(
+    scope: ->(order_form) { order_form.organization.order_forms },
+    lock_key: ->(order_form) { order_form.organization_id }
+  )
+
+  validates :number,
+    presence: true,
+    if: :persisted?
```
nepalez: This can be checked at the db layer to protect the data. All the rest is up to us, not a user's problem

app/services/order_forms/create_service.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module OrderForms
+  class CreateService < BaseService
+    attr_reader :organization, :customer, :params
+
+    def initialize(organization:, customer:, params:)
```
nepalez: I'd just like to clarify my previous comments here (what I meant commenting validation/constraints).

Look, this is what the client provides, and these are only parameters whose errors it make sense to communicate back. All the rest are our own problems, so we could check them as close to the data/transaction as it possible (taking the complexity in mind) and not care about those human-readable errors.

app/services/order_forms/create_service.rb
```diff
@@ -0,0 +1,32 @@
...
+        :auto_execute,
+        :backdated_billing,
+        :order_only
+      )
+      order_form = organization.order_forms.new(
+        customer: @customer,
+        **create_params
+      )
+      order_form.save!
+      result.order_form = order_form
```
nepalez: nitpick: why not just `result.order_form = organization.order_forms.create!(...)` ?

nepalez (commented):
Maybe I'm doing this a bit early (for the draft, i know 🤷‍♂️), but I also suggest some additional specs for models to ensure the invalid states are impossible by any means (either by railsing the db error or whatever).

The reason is the state machine here is complex enough at it would be better to cover its states as early as it possible

app/services/order_forms/create_service.rb
rinasergeeva: Done!

app/services/quotes/clone_service.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      true
+    end
+
+    def create_next_version(quote:)
+      quote.dup.tap do |cloned|
+        cloned.status = :draft
+        cloned.version += 1
+        cloned.share_token = nil # will be generated on saving
+        cloned.save!
```
groyoh: ⚠️ `create_next_version` only resets `status`, `version`, and `share_token`. Because `clonable?` allows voided quotes (as long as no higher version exists), a voided quote can be cloned — but `voided_at`, `void_reason`, and `approved_at` are copied over by `dup` and persisted on the new draft.

Reachable path: draft → clone (v2, original voided `superseded`) → user voids v2 manually → clone v2 again. The new v3 is saved as `status=draft` with `voided_at`/`void_reason` still set from v2. If the user later approves v3, the record ends up with both `approved_at` and `voided_at` set, and no validation catches it (all the related validations are `on: :update` and gated by the new `status`, so they never fire on create).

Explicitly reset these fields when creating the next version:

```ruby
def create_next_version(quote:)
  quote.dup.tap do |cloned|
    cloned.status = :draft
    cloned.version += 1
    cloned.share_token = nil
    cloned.approved_at = nil
    cloned.voided_at = nil
    cloned.void_reason = nil
    cloned.save!
  end
end
```

Also add a spec covering the `clone from voided` path — it's currently untested.

app/services/quotes/clone_service.rb
```diff
@@ -0,0 +1,87 @@
...
+    private
+
+    def clonable?
+      return false if quote.approved?
+      return false if quote.organization.quotes.where(
+        sequential_id: quote.sequential_id,
+        version: (quote.version + 1)..
+      ).exists?
+
+      true
```
groyoh: `quote.organization.quotes.where(...)` will miss pre-existing higher versions if `Quote` ever gains a `default_scope` (e.g. `kept`, or `where.not(status: :voided)`). Be explicit with `.unscoped` or `Quote.where(organization_id: quote.organization_id, sequential_id: quote.sequential_id, version: (quote.version + 1)..)` so adding a default scope later doesn't silently allow concurrent/duplicate versions to be created.

app/services/quotes/void_service.rb
```diff
@@ -0,0 +1,49 @@
...
+    private
+
+    def voidable?
+      quote.approved? || quote.draft?
+    end
+
+    def is_valid?(reason:)
+      return false if reason.blank?
+
+      Quote::VOID_REASONS.has_key?(reason.to_sym)
```
groyoh: Two issues here:

1. `reason.to_sym` raises `NoMethodError` if `reason` is neither a String nor a Symbol (e.g. an integer from a client). Use `reason.to_s.to_sym` to be safe.
2. Ruby convention is `valid?`, not `is_valid?` — rubocop's `Naming/PredicatePrefix` cop will flag this. Rename to `valid_reason?` (and take `reason` positionally) to keep it idiomatic.

app/services/quotes/create_service.rb
```diff
@@ -0,0 +1,66 @@
...
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    end
+
+    private
+
+    def add_owners!(quote:, owners:)
+      return if owners.blank?
+
+      owners.uniq.each do |user_id|
```
groyoh: `add_owners!` accepts any `user_id` and creates `quote_owners` rows without checking the user belongs to the `organization`. Since the service is the unit that enforces business rules (controllers will call it directly), it should filter/validate the list:

```ruby
organization_user_ids = organization.memberships.active.pluck(:user_id)
owners.uniq.each do |user_id|
  next unless organization_user_ids.include?(user_id)
  quote.quote_owners.create!(organization:, user_id:)
end
```

Or validate up-front and return a `validation_failure!` on unknown IDs. Same concern applies to `Quotes::UpdateService#sync_owners!`.

app/services/quotes/update_service.rb
```diff
@@ -0,0 +1,71 @@
...
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    end
+
+    private
+
+    def editable?
+      quote.draft?
+    end
```
groyoh, line 47: If `params[:owners]` is passed as `nil` (different from absent), `owners.uniq` here raises `NoMethodError`. Either guard with `owners.blank?` at the top of `sync_owners!` (mirroring `CreateService#add_owners!`), or tighten the caller to `if params[:owners].present?`.

spec/models/quote_spec.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Quote do
+  subject(:quote) { create(:quote) }
+
+  it do
```
groyoh: Per `AGENTS.md`, model specs must cover:

- All enums inside `describe "enums"` with a single `it` block — currently only `status` and `void_reason` are tested; `order_type`, `execution_mode`, and `backdated_billing` are missing.
- All associations inside `describe "associations"` (`belongs_to :organization`, `:customer`, `has_one :order_form`, `has_one :order, through: :order_form`, `has_many :quote_owners, dependent: :destroy`, `has_many :owners, through: :quote_owners`).
- All validations inside `describe "validations"` — the conditional `share_token`, `void_reason`/`voided_at`, `approved_at` presence validations are non-trivial and need coverage.
- Behaviour of `ensure_number` and `ensure_share_token` callbacks.

The current single top-level `it do ... end` block doesn't match the convention.

app/models/order.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
```
groyoh: No spec file for `Order`, `OrderForm`, or `QuoteOwner` — every new model must have an associated spec covering enums, associations, and validations. These are currently untested (the only factory added is `:quote`, so `Order`/`OrderForm` don't even have factories).

app/models/quote.rb
```diff
@@ -0,0 +1,126 @@
...
+    cascade_of_expired: 2,
+    cascade_of_voided: 3
+  }.freeze
+
+  before_save :ensure_number
+  before_save :ensure_share_token
+
+  belongs_to :organization
+  belongs_to :customer
+  has_one :order_form
```
groyoh: `has_one :order_form` / `has_one :order, through: :order_form` are not backed by a DB uniqueness constraint on `order_forms.quote_id`. If a future flow produces a second order form for the same quote (e.g. after voiding the first one), `has_one` will silently return an arbitrary row and the data model will desync. Either add a unique index on `order_forms(quote_id)` (possibly partial on active statuses) or document that multiple order forms per quote are allowed and switch to `has_many`.

db/migrate/20260304074158_add_order_forms_foundations.rb
```diff
@@ -0,0 +1,180 @@
...
+        index: false, # covered by the composite unique index below
+        type: :uuid
+      t.references :user,
+        null: false,
+        foreign_key: true,
+        type: :uuid
+      t.timestamps
+
+      t.index [:quote_id, :user_id],
+        unique: true,
```
groyoh: `orders.execution_record` is declared as `t.json` while every other structured column in this migration is `jsonb`. Given this field is likely to be queried/indexed later (contains references to created subscriptions/invoices/etc.), use `t.jsonb` for consistency and to enable GIN indexes.

app/config/feature_flags.yaml
```diff
@@ -25,3 +25,7 @@ multi_currency:
 payment_gated_subscriptions:
   description: "Enable payment-gated subscription activation rules"
   owners: ["@osmarluz", "@ancorcruz"]
+
+order_forms:
```
groyoh, line 33: I'd suggest to either keep some consistency. The API key resource is `quote` so let's keep the feature flag as `quote`.

app/models/organization.rb
```diff
@@ -143,6 +146,7 @@ class Organization < ApplicationRecord
     events_targeting_wallets
     security_logs
     granular_lifetime_usage
+    order_forms
```
groyoh, line 149: Let's keep some consistency here as well using the `quote` naming.

app/services/quotes/approve_service.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    Result = BaseResult[:quote]
+
+    def initialize(quote:)
+      @quote = quote
+      super
+    end
+
+    def call
+      return result.forbidden_failure! unless License.premium?
```
groyoh: The organization premium integration is never checked.

db/migrate/20260304074158_add_order_forms_foundations.rb
```diff
@@ -0,0 +1,180 @@
...
+      # content
+      t.jsonb :billing_items
+      t.jsonb :commercial_terms
+      t.text :content
+      t.text :legal_text
+      t.text :internal_notes
+      t.jsonb :contacts
+      t.jsonb :metadata
+      t.boolean :auto_execute, null: false, default: false
+      t.integer :backdated_billing, comment: "Rails enum"
```
groyoh: Why are we using Rails enums here and PG enums for other columns like `status` ?

db/migrate/20260304074158_add_order_forms_foundations.rb
```diff
@@ -0,0 +1,180 @@
+# frozen_string_literal: true
+
+class AddOrderFormsFoundations < ActiveRecord::Migration[8.0]
+  def change
+    create_enum :quote_status, %w[draft approved voided]
+
+    create_table :quotes, id: :uuid do |t|
```
groyoh: I would suggest to split the data structure change into multiple migrations.

app/models/order_form.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+class OrderForm < ApplicationRecord
```
groyoh: I believe we should remove order and order form from this PR and only introduce them when needed.

app/services/quotes/clone_service.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    private
+
+    def clonable?
+      return false if quote.approved?
+      return false if quote.organization.quotes.where(
+        sequential_id: quote.sequential_id,
+        version: (quote.version + 1)..
+      ).exists?
```
groyoh: We currently allow to duplicate an older version on the frontend.

db/migrate/20260304074158_add_order_forms_foundations.rb
```diff
@@ -0,0 +1,180 @@
...
+      t.timestamps
+      # content
+      t.jsonb :billing_items
+      t.jsonb :commercial_terms
+      t.text :content
+      t.text :legal_text
+      t.text :internal_notes
+      t.jsonb :contacts
+      t.jsonb :metadata
+      t.boolean :auto_execute, null: false, default: false
```
groyoh: Everything related to auto execution should be moved to order.

db/migrate/20260304074158_add_order_forms_foundations.rb
```diff
@@ -0,0 +1,180 @@
...
+      t.integer :void_reason, comment: "Rails enum"
+      t.timestamps
+      # content
+      t.jsonb :billing_items
+      t.jsonb :commercial_terms
+      t.text :content
+      t.text :legal_text
+      t.text :internal_notes
+      t.jsonb :contacts
+      t.jsonb :metadata
```
groyoh: Could you please re-review those fields ? I believe some of them are not actually defined in the specs anymore.

db/migrate/20260304074158_add_order_forms_foundations.rb
```diff
@@ -0,0 +1,180 @@
...
+      # identity
+      t.references :organization,
+        null: false,
+        foreign_key: true,
+        index: false, # covered by the composite unique index below
+        type: :uuid
+      t.references :customer,
+        null: false,
+        foreign_key: true,
+        type: :uuid
```
groyoh: Let's add an optional FK  to the `subscription` as it will be neede by the frontend.

db/migrate/20260304074158_add_order_forms_foundations.rb
groyoh: Let's just keep `content` and `billing_items` for now.

db/migrate/20260304074158_add_order_forms_foundations.rb
```diff
@@ -0,0 +1,180 @@
...
+      t.string :number, null: false
+      t.integer :version, null: false, default: 1
+      t.integer :sequential_id, null: false
+      t.integer :order_type, null: false, comment: "Rails enum"
+      t.string :currency
+      t.text :description
```
groyoh: I'm not sure if this is needed either.

db/migrate/20260304074158_add_order_forms_foundations.rb
```diff
@@ -0,0 +1,180 @@
...
+        type: :uuid
+      t.uuid :contract_uploaded_by_user
+      t.datetime :contract_uploaded_at
+      t.datetime :expires_at
+      t.datetime :signed_at
+      t.datetime :voided_at
+      t.timestamps
+      # content
+      t.jsonb :billing_snapshot, null: false
+      t.text :content
```
groyoh: Isn't this already in the quote itself ?

app/config/feature_flags.yaml
rinasergeeva: We hide the entire feature behind this feature flag, not just quotes

app/services/quotes/clone_service.rb
rinasergeeva: That's against the spec and needs to be clarified.

db/migrate/20260304074158_add_order_forms_foundations.rb
rinasergeeva: The result of the discussion: let's make all of them PG enums.

app/models/order.rb
rinasergeeva: Yes, that's the result of the epic branch's evolution. The Quote part was just completed and merged ahead of the other parts. Since we decided to develop in the main branch, the OrderForm and Orders will be deleted here.

db/migrate/20260304074158_add_order_forms_foundations.rb
rinasergeeva: The orders table was removed.

db/migrate/20260304074158_add_order_forms_foundations.rb
rinasergeeva: Deleted!

db/migrate/20260304074158_add_order_forms_foundations.rb
rinasergeeva: Deleted!

db/migrate/20260304074158_add_order_forms_foundations.rb
rinasergeeva: Done!

db/migrate/20260304074158_add_order_forms_foundations.rb
rinasergeeva: Done!

db/migrate/20260304074158_add_order_forms_foundations.rb
rinasergeeva: Done!

app/models/order_form.rb
rinasergeeva: Done!

db/migrate/20260304074158_add_order_forms_foundations.rb
rinasergeeva: Only Quote-related tables remain.

app/services/quotes/create_service.rb
rinasergeeva: Done!

app/services/quotes/clone_service.rb
rinasergeeva: Done!

app/services/quotes/update_service.rb
rinasergeeva: Normalization is added!

app/services/quotes/clone_service.rb
rinasergeeva: Update: any version allowed!

app/models/quote.rb
rinasergeeva: Totally agree! Let's add constraints to the `order_forms` table when it's ready.

app/services/quotes/void_service.rb
rinasergeeva: Done!

app/services/quotes/clone_service.rb
rinasergeeva: Since the cloning rules changed, the service no longer performs this check.

spec/models/quote_spec.rb
rinasergeeva: I've expanded both specs

db/migrate/20260304074159_add_order_forms.rb
rinasergeeva: Some of the lifecycle constraints were added.

db/migrate/20260304074159_add_order_forms.rb
rinasergeeva: Since this part was moved to OrderForms, it is no longer relevant

db/migrate/20260304074159_add_order_forms.rb
rinasergeeva: We just assign a new token when cloning, I believe. Currently, only draft and approved quotes are shareable.

db/migrate/20260304074159_add_order_forms.rb
rinasergeeva: Added!

app/models/organization.rb
rinasergeeva: The same here: the generalized term for the whole feature is `order_forms`, if I'm not mistaken.

app/config/feature_flags.yaml
groyoh: Indeed, but the feature begins with quote (which gates the other features as well) so I guess it does makes sense to keep consistency with `quotes` here. Also the generic term for this is CPQ (Configure, Price, Quote), so I guess it would be less confusing than `order_forms`.

toommz (approved):
It's IMO a sane basis for the upcoming work 👍

app/services/quotes/create_service.rb
```diff
@@ -0,0 +1,108 @@
...
+      @params = params
+      @owners = normalize_owners(owners: params[:owners])
+      super
+    end
+
+    def call
+      return result.forbidden_failure! unless License.premium?
+      return result.not_found_failure!(resource: "organization") unless organization
+      return result.not_found_failure!(resource: "customer") unless customer
+      return result.not_found_failure!(resource: "subscription") if subscription_required? && subscription.blank?
```
groyoh, line 24: ⚠️ The `subscription` is only checked for presence here — its `organization_id` and `customer_id` aren't compared with the quote's. If a caller passes a subscription belonging to a different customer or organization, the quote would still be created and linked to it across tenants.

Tightening the guard would close that gap, e.g. before the feature-flag check:

```ruby
if subscription && (subscription.customer_id != customer.id || subscription.organization_id != organization.id)
  return result.not_found_failure!(resource: "subscription")
end
```

A spec context like `"when subscription belongs to another customer"` would lock the behavior in.

db/migrate/20260304074158_add_quotes.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+
+class AddQuotes < ActiveRecord::Migration[8.0]
+  def change
+    create_enum :quote_status, %w[draft approved voided]
+    create_enum :quote_void_reason, %w[manual superseded cascade_of_expired cascade_of_voided]
+    create_enum :order_type, %w[subscription_creation subscription_amendment one_off]
```
groyoh: Other PostgreSQL enums in this codebase prefix the type name with the model/table context (`subscription_on_termination_credit_note`, `subscription_activation_rule_types`, `usage_monitoring_alert_types`), and `AGENTS.md` calls that pattern out. Once the upcoming `orders` table arrives, `order_type` reads as if it belonged to that table rather than to `quotes`. Renaming to `quote_order_type` lines it up with the rest, and it's still cheap to do here since no production rows exist yet.

app/models/quote.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  belongs_to :organization
+  belongs_to :customer
+  belongs_to :subscription, optional: true
+
+  has_many :quote_owners, dependent: :destroy
+  has_many :owners, through: :quote_owners, source: :user, class_name: "User"
+
+  has_many :versions, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
+  has_one :current_version, -> { order(sequential_id: :desc) }, class_name: "QuoteVersion"
```
groyoh, line 22: `current_version` returns the latest version by `sequential_id` regardless of status, so on a quote whose versions have all been voided, `quote.current_version` returns a voided record — which could be confusing. Two ways to align name and behavior:

Scope it to active statuses:

```ruby
has_one :current_version,
  -> { where(status: %w[draft approved]).order(sequential_id: :desc) },
  class_name: "QuoteVersion"
```

Or rename it to `latest_version` so callers don't read "current" as "active". The first option also reads more naturally alongside the partial unique index on `(quote_id) WHERE status IN ('draft','approved')`.

app/services/quote_versions/clone_service.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module QuoteVersions
+  class CloneService < BaseService
+    class CloneError < StandardError
+      attr_reader :cause, :error
```
groyoh: `attr_reader :cause` overlaps with Ruby's built-in `Exception#cause`, and the value passed to `cause:` here is actually a service `Result` rather than an exception — so a later `rescue ... => e; e.cause` reader could be surprised by what they find. A domain-specific name (e.g. `:source_result`) would avoid the overlap; alternatively, if the goal is to preserve the original error chain, `raise CloneError.new(...), cause: original_exception` lets Ruby's built-in chaining do the work and the manual accessor can go away.

spec/services/quotes/create_service_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      let(:customer) { nil }
+
+      it "returns a not found error" do
+        expect(result).not_to be_success
+        expect(result.error).to be_a(BaseService::NotFoundFailure)
+        expect(result.error.message).to eq("customer_not_found")
+      end
+    end
+
+    context "when subscription is required but not provided", :premium do
```
groyoh, line 132: The `subscription_amendment` order type only has a failure-case spec (subscription missing). A positive-path context that builds a `subscription` belonging to the same `customer` and asserts the quote ends up with `subscription_id` set would round out the coverage — and pairs naturally with the cross-customer/cross-org check noted in `create_service.rb`.

spec/models/quote_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      subscription = create(:subscription)
+      quote = build(:quote, order_type: :subscription_amendment, subscription: nil, organization: subscription.organization, customer: create(:customer, organization: subscription.organization))
+      expect(quote).not_to be_valid
+      quote.subscription = subscription
+      expect(quote).to be_valid
+    end
+  end
+
+  describe "callbacks" do
+    describe "ensure_number" do
```
groyoh: We're lacking some scenario (blank created_at, number already set).

app/models/quote_owner.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class QuoteOwner < ApplicationRecord
```
groyoh, line 3: We're lacking tests for this model.

app/models/quote_version.rb
```diff
@@ -0,0 +1,94 @@
...
+    cascade_of_voided: "cascade_of_voided"
+  }.freeze
+
+  before_save :ensure_share_token
+
+  belongs_to :organization
+  belongs_to :quote
+
+  has_one :order_form
+  has_one :order, through: :order_form
```
groyoh:
```suggestion
```

spec/models/quote_version_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      expect(subject).to belong_to(:organization)
+      expect(subject).to belong_to(:quote)
+
+      # TODO: Uncomment when the order form and order associations are implemented
+      # expect(subject).to have_one(:order_form)
+      # expect(subject).to have_one(:order).through(:order_form)
+    end
+  end
+
+  describe "validations" do
```
groyoh, line 36: There should be a test ensuring it is valid by default.

app/models/quote_version.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  validates :void_reason, :voided_at,
+    presence: true,
+    if: -> { voided? }
+
+  validates :approved_at,
+    presence: true,
+    if: -> { approved? }
+
+  sequenced(
```
groyoh, line 44: This is not tested.

app/models/quote.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  enum :order_type, ORDER_TYPES,
+    instance_methods: false,
+    validate: true
+
+  validates :subscription_id,
+    presence: true,
+    if: -> { order_type == "subscription_amendment" }
+
+  sequenced(
```
groyoh, line 32: This is not tested.

app/models/user.rb
```diff
@@ -14,6 +14,9 @@ class User < ApplicationRecord
   has_many :active_memberships, -> { where(status: "active") }, class_name: "Membership"
   has_many :active_organizations, through: :active_memberships, source: :organization
 
+  has_many :quote_owners, dependent: :destroy
+  has_many :quotes, through: :quote_owners
```
groyoh, line 18: This is not tested.

app/services/quote_versions/clone_service.rb
```diff
@@ -0,0 +1,76 @@
...
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    rescue CloneError, ActiveRecord::ActiveRecordError => e
+      result.service_failure!(code: "clone_failed", message: e.message, error: e)
+    end
+
+    private
+
+    def clonable?
```
groyoh, line 49: I believe we should check whether any quote version is already approved.

app/services/quote_versions/create_service.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module QuoteVersions
+  class CreateService < BaseService
+    attr_reader :organization, :quote, :params
+
+    Result = BaseResult[:quote_version]
+
+    def initialize(organization:, quote:, params: {})
```
groyoh: We don't need to pass `organization` here since we can fetch it from the quote.

app/services/quote_versions/create_service.rb
```diff
@@ -0,0 +1,44 @@
...
+      )
+
+      result.quote_version = quote_version
+
+      # TODO: SendWebhookJob.perform_after_commit("quote_version.created", quote_version)
+
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    rescue ActiveRecord::ActiveRecordError => e
```
groyoh: Not sure which error we're catching here but we should be a bit more explicit and it should be tested.

app/services/quote_versions/update_service.rb
```diff
@@ -0,0 +1,43 @@
...
+    Result = BaseResult[:quote_version]
+
+    def initialize(quote_version:, params:)
+      @quote_version = quote_version
+      @params = params
+      super
+    end
+
+    def call
+      return result.forbidden_failure! unless License.premium?
```
groyoh: We should probably define a helper for `License.premium? && quote_version.organization.feature_flag_enabled?(:order_forms)`

app/services/quote_versions/void_service.rb
```diff
@@ -0,0 +1,52 @@
...
+      @quote_version = quote_version
+      @reason = reason
+      super
+    end
+
+    def call
+      return result.forbidden_failure! unless License.premium?
+      return result.not_found_failure!(resource: "quote_version") unless quote_version
+      return result.forbidden_failure! unless quote_version.organization.feature_flag_enabled?(:order_forms)
+      return result.validation_failure!(errors: {quote_version: ["invalid_void_reason"]}) unless valid_reason?(reason:)
```
groyoh: No need for custom error:

```suggestion
      return result.single_validation_failure!(field: :void_reason, error_code: :invalid}) unless valid_reason?(reason:)
```

app/services/quote_versions/void_service.rb
```diff
@@ -0,0 +1,52 @@
...
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    rescue ActiveRecord::ActiveRecordError => e
+      result.service_failure!(code: "void_failed", message: e.message, error: e)
+    end
+
+    private
+
+    def voidable?
+      quote_version.approved? || quote_version.draft?
```
groyoh: Can we really void an approved version ?

app/services/quotes/create_service.rb
```diff
@@ -0,0 +1,108 @@
...
+      return result.forbidden_failure! unless organization.feature_flag_enabled?(:order_forms)
+      return result.validation_failure!(errors: {quotes: ["invalid_owner"]}) unless check_owners(owners:)
```
groyoh: No need for customer error here:

```suggestion
      return result.validation_failure!(errors: {owners: ["invalid"]}) unless check_owners(owners:)
```

app/services/quotes/create_service.rb
```diff
@@ -0,0 +1,108 @@
...
+    private
+
+    def subscription_required?
+      params[:order_type].to_s == "subscription_amendment"
+    end
+
+    def check_owners(owners:)
+      return true if owners.blank?
+
+      valid_owners = organization.memberships.active.pluck(:user_id)
```
groyoh: `organization.memberships.active.pluck(:user_id)` reads every active membership of the org into memory just to check (typically) 1–3 owner ids, so the cost scales with the org size rather than with the input. Restricting the query to the candidate ids keeps it proportional:

```ruby
valid = organization.memberships.active.where(user_id: owners).pluck(:user_id)
(owners - valid).empty?
```

app/services/quotes/create_service.rb
```diff
@@ -0,0 +1,108 @@
...
+
+      raise CreateError.new(error: create_result.error, cause: create_result) unless create_result&.success?
+    end
+
+    def add_owners!(quote:, owners:)
+      return if owners.blank?
+
+      owners.each do |user_id|
+        quote.quote_owners.create!(organization:, user_id:)
+      end
```
groyoh: `add_owners!` issues one INSERT per owner, which adds up when a quote starts with several. A single `insert_all` keeps it to one round trip:

```ruby
now = Time.current
quote.quote_owners.insert_all(
  owners.map { |uid| { organization_id: organization.id, user_id: uid, created_at: now, updated_at: now } }
)
```

The same shape would apply to `sync_owners!` in `update_service.rb`.

PR #5228, opened 2026-03-24
feat(multi-currency): split multiple subscriptions invoice per currency

lovrocolic:
## Context

Lago is working on adding more flexibility into customer relations. The end-goal is to support parent/children associations between customers.

First step is to support multiple currencies per customer.  It will be possible that customer handles billing objects like wallet or subscription in different currencies.

## Description

This PR updates billing logic. If a customer has multiple subscriptions, we need to group them by currency (in addition to payment method).

### DIFF
--- a/spec/services/subscriptions/organization_billing_service_spec.rb
+++ b/spec/services/subscriptions/organization_billing_service_spec.rb
@@ -442,6 +442,176 @@ RSpec.describe Subscriptions::OrganizationBillingService do
       end
     end
 
+    context "when grouping subscriptions by currency" do
+      let(:organization) { create(:organization, feature_flags: ["multi_currency"]) }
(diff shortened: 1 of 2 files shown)


spec/services/subscriptions/organization_billing_service_spec.rb
```diff
@@ -442,6 +442,176 @@ def expect_to_bill_together(subscriptions, date)
...
+            started_at: current_date - 10.days,
+            billing_time:,
+            created_at:
+          )
+        end
+
+        before do
+          usd_subscription
+          eur_subscription
+        end
```
toommz, line 483: Could we have used `let!` instead of having a `before` block?

toommz (approved):

PR #5234, opened 2026-03-25
feat(multi-currency): adjust credit note logic for multi-currency flow

lovrocolic:
## Context

Lago is working on adding more flexibility into customer relations. The end-goal is to support parent/children associations between customers.

First step is to support multiple currencies per customer.  It will be possible that customer handles billing objects like wallet or subscription in different currencies.

## Description

This PR updates credit note logic. Filtering credit notes per currency is safe and backward-compatible (since currently we support only one currency per customer) and this PR can be merged without any feature flag.

### DIFF
--- a/app/graphql/types/customers/object.rb
+++ b/app/graphql/types/customers/object.rb
@@ -82,6 +82,10 @@ module Types
         GraphQL::Types::BigInt,
         null: false,
         description: "Credit notes credits balance available per customer"
+      field :credit_notes_balances,
+        [Types::Customers::CreditNotesBalance],
+        null: false,
+        description: "Credit notes credits balance available per customer per currency"
       field :credit_notes_credits_available_count,
         Integer,
         null: false,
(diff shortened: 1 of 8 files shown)


app/graphql/types/customers/object.rb
```diff
@@ -153,6 +157,15 @@ def credit_notes_balance_amount_cents
         object.credit_notes.finalized.sum("credit_notes.balance_amount_cents")
       end
 
+      def credit_notes_balances
+        object.credit_notes
+          .finalized
```
toommz, line 162: We are accessing quite a few times `object.credit_notes.finalized` (4x it seems). Should we memoize those once for all?

toommz (approved):

app/graphql/types/customers/object.rb
lovrocolic: Actually memoizing wouldn't help here. Each of the four methods builds a different query with different WHERE, GROUP BY, and aggregate functions, so we would not reduce DB calls.

PR #5237, opened 2026-03-25
feat(invoices): Add preview state to be used with regeneration 

tiagolupepic:
This commit introduces a new GraphQL query for building preview invoices for regeneration.

We have a current problem related to taxes applied in Invoice, Fee and its relations, applied_taxes. 
These taxes are generated during the invoice generation and they keep persisted, they're based
on customer data, when the customer data changes, the previous applied taxes are not affected.

However, we have a scenario when customer changes their taxes and they want to regenerate invoices.
During the process to display the invoice information, frontend ask the invoice details using the invoice GraphQL query and as expected, we're still showing the applied taxes based on previous customer information which for this context is wrong.

In order to correctly show the information to the users and consider the current customer taxes, we're introducing the `preview` state via the service `Invoice::BuildRegenerationPreviewService`. This service uses the invoice fees to build new preview state considering the current customer taxes and correctly generate the applied taxes. These data is generated on the fly and we use the same `id` for the preview fees and invoices.

Once the user submits the regenerate action, the persisted fees will be updated containing the correct taxes. 

Some minor changes
- Changed from #count to #length, we're prioritizing the records in memory instead of touching the DB. It also solves the problem when fees are not persisted.
- Changed the `.sum` with `&` to use ruby methods instead of calling the AR#sum which touches the DB.
- ComputeAmountsFromFees will not persist fee if the invoice is not persisted


spec/services/invoices/compute_amounts_from_fees_spec.rb
```diff
@@ -61,6 +61,27 @@
     expect { compute_amounts.call }.to change(invoice, :total_amount_cents).from(0).to(559)
   end
 
+  context "when invoice is not persisted and have not persisted fees" do
```
toommz:
```suggestion
  context "when invoice is not persisted and has no persisted fees" do
```

spec/services/invoices/compute_amounts_from_fees_spec.rb
```diff
@@ -61,6 +61,27 @@
...
+    subject { described_class.new(invoice: draft_invoice) }
+
+    let(:draft_invoice) { build(:invoice, organization:, customer:, fees: fees) }
+    let(:fees) do
+      [build(:fee, amount_cents: 100), build(:fee, amount_cents: 150)]
+    end
+
+    it "avoids persists fees" do
```
toommz:
```suggestion
    it "avoids persisting fees" do
```

spec/services/invoices/compute_amounts_from_fees_spec.rb
```diff
@@ -61,6 +61,27 @@
...
+      result = subject.call
+
+      expect(result.invoice.fees.first.id).to be_nil
```
toommz: Could be more resilient to check that absolutely no new `Fee` has been persisted.

```suggestion
      expect { subject.call }.not_to change(Fee, :count)
```

spec/services/invoices/draft_service_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+            unit_precise_amount: fee.amount.currency.subunit_to_unit,
+            charge_id: nil,
+            charge_filter_id: nil,
+            fixed_charge_id: nil,
+            invoice_display_name: "Subscription Fee"
+          }
+        )
+
+        expect(result.invoice.fees.first.taxes_rate).to eq(0)
+      end
```
toommz: The example name does not seem to totally reflect the test itself. We say "for each fee" but as there's only one fee, it does not prove that it will call it accordingly with multiple fees, isn't it?

spec/services/invoices/draft_service_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+
+      context "when exists taxes" do
```
toommz:
```suggestion
      context "with taxes" do
```

Seems to be more in sync with other context names

spec/services/invoices/draft_service_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+          amount_cents: 1000,
+          taxes_rate: 10,
+          amount_currency: "EUR",
+          invoice_display_name: "Subscription Fee"
+        )
+      end
+
+      before { fee }
+
+      it "creates a draft invoice with estimated fees" do
```
toommz:
```suggestion
      it "builds a draft invoice with estimated fees" do
```

I might be wrong but it does not "create" anything in the db, right?

app/services/invoices/draft_service.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Invoices
+  class DraftService < BaseService
```
toommz: Might be because of my bad understanding, but I can't really figure what the service is supposed to do with the current naming. After looking at the implementation I would say that it "builds a draft invoice from a persisted invoice".

toommz (commented):
I have the feeling that fixing it this way is hiding the true problem (being that our tax and fee calculations rely on activerecord objects' state).

One thing that made me think that is that it leaks into the services and already existing types (e.g. by needing to call `persisted?` now). I fear that it can create more problems later 🤔

spec/services/invoices/compute_amounts_from_fees_spec.rb
tiagolupepic: it has the same effect, no?

app/services/invoices/draft_service.rb
tiagolupepic: The idea is to create a draft state and present to the user during the regenerate invoice. I didn't put `preview` because we're already using such name in a premium feature. I'm open to suggestions for the name.

spec/services/invoices/draft_service_spec.rb
tiagolupepic: I'll change this to make it clear

spec/services/invoices/compute_amounts_from_fees_spec.rb
toommz: Not exactly, your version asserts that the first fee linked to the invoice is not persisted, mine asserts that we haven't persisted any fee, linked to this invoice or not. Maybe at least we can assert with your version that there is only one fee linked to the invoice?

app/services/invoices/draft_service.rb
toommz: Sure if preview is already "taken" it makes sense to lean into the draft naming.

app/graphql/resolvers/invoice_draft_resolver.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    description "Build a draft of a single Invoice of an organization."
+
+    argument :id, ID, required: true, description: "Uniq ID of the invoice"
+
+    type Types::Invoices::Object, null: true
+
+    def resolve(id:)
+      invoice = current_organization.invoices.visible.find(id)
+      result = Invoices::DraftService.call(invoice: invoice)
```
lovrocolic: `draft` naming is misleading. Draft is AASM state and enum value for the invoice, so I would suggest using `RegenerationPreviewService` or something similar

app/services/invoices/regeneration_preview_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    def call
+      draft_invoice = invoice.dup
+      invoice.fees.each do |fee|
+        result = ::AdjustedFees::EstimateService.call(
+          invoice: invoice,
+          params: {
+            invoice_subscription_id: fee.subscription_id,
+            fee_type: fee.fee_type,
+            units: fee.units,
+            unit_precise_amount: fee.amount.currency.subunit_to_unit,
```
lovrocolic: We have a bug here - this returns constant

app/services/invoices/draft_service.rb
lovrocolic: we are missing result definition:
`Result = BaseResult[:invoice]`

app/services/invoices/regeneration_preview_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    def initialize(invoice:)
+      @invoice = invoice
+
+      super
+    end
+
+    def call
+      draft_invoice = invoice.dup
+      invoice.fees.each do |fee|
+        result = ::AdjustedFees::EstimateService.call(
```
lovrocolic: It seems we miss error handling here

app/graphql/resolvers/invoice_draft_resolver.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Resolvers
+  class InvoiceDraftResolver < Resolvers::BaseResolver
```
lovrocolic: Can we please add specs for this file

app/services/invoices/regeneration_preview_service.rb
```diff
@@ -0,0 +1,47 @@
...
+            fixed_charge_id: fee.fixed_charge_id,
+            invoice_display_name: fee.invoice_display_name
+          }
+        )
+        result.fee.id = fee.id
+        result.fee.adjusted_fee = nil
+        draft_invoice.fees << result.fee
+      end
+
+      result = Invoices::ComputeAmountsFromFees.call(invoice: draft_invoice, provider_taxes: nil)
```
lovrocolic: What happens if provider taxes are used like Anrok on Avalara? I guess we miss this case

app/services/invoices/regeneration_preview_service.rb
lovrocolic: I am not sure TBH why we need to call `AdjustedFees::EstimateService`? Can't we just use something like this:


```
preview = invoice.dup
preview.applied_taxes = []
preview.fees = invoice.fees.map do |fee|
  dup_fee = fee.dup
  dup_fee.applied_taxes = []
  dup_fee.taxes_amount_cents = 0
  dup_fee.taxes_rate = 0
  dup_fee.id = fee.id
  dup_fee
end
```

After that we just need to assign new taxes and compute invoice totals

lovrocolic (commented):
Hey @tiagolupepic .

Thanks for taking care of this case!

Added some comments:)

app/services/invoices/regeneration_preview_service.rb
tiagolupepic: currency yes, returns the Money. subunit_to_unit returns the integer in cents.

Screenshot 2026-03-30 at 14 54 46

app/graphql/resolvers/invoice_draft_resolver.rb
tiagolupepic: I agree. `RegenerationPreviewService` it's much better and avoids misleading with state machine names.

app/services/invoices/regeneration_preview_service.rb
tiagolupepic: The idea is to have this service agnostic of taxes attributes. From this class, I don't know how the taxes are calculated and how the Fee attributes are generated, it's basically delegating to each dependency.

app/services/invoices/regeneration_preview_service.rb
tiagolupepic: good point! going to research these cases and make sure we're covering, thanks for pointing them out!

app/services/invoices/regeneration_preview_service.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Invoices
+  class RegenerationPreviewService < BaseService
```
rsempe: We're usually using a verb for the name of our services.

And, we have an existing `preview` namespace inside `app/services/invoices`, does it make sense to use it?

Something like `Invoices::Preview::RegenerateService` or `Invoices::RegeneratePreviewService`?

app/services/invoices/compute_amounts_from_fees.rb
```diff
@@ -18,12 +18,12 @@ def call
...
           end
 
-          fee.save!
+          fee.save! if invoice.persisted?
         end
       end
 
-      invoice.fees_amount_cents = invoice.fees.sum(:amount_cents)
-      invoice.coupons_amount_cents = invoice.credits.coupon_kind.sum(:amount_cents)
+      invoice.fees_amount_cents = invoice.fees.sum(&:amount_cents)
```
rsempe, line 25: is it a perf issue when an invoice has more than 1000 fees?

we're now loading every fee instead of doing a single sql sum

app/services/invoices/build_regeneration_preview_service.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      super
+    end
+
+    def call
+      preview_invoice = invoice.dup
+
+      invoice.fees.each do |fee|
+        dup_fee = fee.dup
+        result = Fees::ApplyTaxesService.call!(fee:)
```
rsempe: do we need to do this? 🤔

spec/services/invoices/regeneration_preview_service_spec.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Invoices::RegenerationPreviewService do
+  subject(:draft_service) { described_class.new(invoice:) }
```
rsempe:
```suggestion
  subject(:preview_service) { described_class.new(invoice:) }
```

rsempe:
can you add a scenario test that can verify the actual fix?

app/graphql/types/invoices/object.rb
```diff
@@ -97,7 +97,11 @@ def regenerated_invoice_id
       end
 
       def applied_taxes
-        object.applied_taxes.order(tax_rate: :desc)
+        if object.applied_taxes.any? { |applied_tax| !applied_tax.persisted? }
```
rsempe, line 100: can't we just check `object.persisted?` instead?

rsempe (changes requested):

app/services/invoices/build_regeneration_preview_service.rb
tiagolupepic: yes, this is the logic that applies the current taxes from the customer. It's the same used in AdjustedFee service.

app/services/invoices/regeneration_preview_service.rb
tiagolupepic: it was a suggestion by Lovro's https://github.com/getlago/lago-api/pull/5237#discussion_r3000180969. I can rename it again. no problem.

tiagolupepic: `Invoices::RegenerationPreviewService` gets closer to the functionality. Regenerate sounds the service will persist something, no?

app/services/invoices/compute_amounts_from_fees.rb
tiagolupepic: actually, we're saving a trip to the DB. We're loading the fees already some lines above. https://github.com/getlago/lago-api/blob/main/app/services/invoices/compute_amounts_from_fees.rb#L13-L23. I checked the production database and based on the invoices generated we skipped the guard clause `should_apply_fee_taxes?`  only for ~1,6% of the invoices (`advance_charges`).

app/services/invoices/regeneration_preview_service.rb
rsempe: `Invoices::Build{Regeneration}PreviewService`? 😅

app/services/invoices/compute_amounts_from_fees.rb
rsempe: good spot 👍

app/graphql/types/invoices/object.rb
tiagolupepic: it's also possible. I put in this way to keep consistency with the same logic used in fee object https://github.com/getlago/lago-api/blob/main/app/graphql/types/fees/object.rb#L54-L60

app/services/invoices/regeneration_preview_service.rb
tiagolupepic: done!

rsempe (approved):

PR #5248, opened 2026-03-27
feat(presentation group keys): POC with 2 requests

annvelents:
the goal of this POC is to be able to compare approaches of having a separate request to get the presentation_breakdown (but having 2 requests per each charge filter), or have everything in one request and then retransform the data
current POC implements 2 requests approach

Note:
20k fees (20k events) were generated in 1.5min - test that was done 2 months ago
with this POC I had 5 charges with 10 pricing group keys in each (50 fees) with 200 presentation breakdowns inside - response was received in 6s

### DIFF
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -366,7 +366,15 @@ module Fees
     end
 
     def compute_presentation_breakdowns(fees, charge_filter:)
-      presentation_keys = charge.presentation_group_keys_values
+      all_presentation_keys = charge.presentation_group_keys_values
+
+      presentation_keys = if usage_filters.filter_by_presentation.present?
+        all_presentation_keys & usage_filters.filter_by_presentation
+      else
+        all_presentation_keys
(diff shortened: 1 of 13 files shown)


app/services/fees/charge_service.rb
```diff
@@ -361,6 +365,53 @@ def persist_recurring_value(aggregation_results, charge_filter)
...
+        all_presentation_keys
+      end
+
+      return if presentation_keys.blank?
+
+      combined_filters = aggregation_filters(charge_filter:)
+      combined_keys = ((combined_filters[:grouped_by] || []) + presentation_keys).uniq
+      combined_filters[:grouped_by] = combined_keys
+
+      combined_aggregator = BillableMetrics::AggregationFactory.new_instance(
```
annvelents, line 383: SUM aggregation makes 2 queries: sum itself and count to get the number of events affecting this fee. For presentation breakdown we don't need count. Can we not make it?

PR #5259, opened 2026-03-30
feat(customer-multi-currencies): Adjust dunning campaigns and payment requests

lovrocolic:
## Context

Lago is adding multi-currency support per customer. The existing dunning campaign logic uses a single global counter (`last_dunning_campaign_attempt`) on `Customer`, which breaks when multiple currencies are dunned in the same round — the counter increments per currency, exhausting `max_attempts` prematurely, and concurrent `ProcessAttemptJobs` race on the same counter.

## Description

Introduces per-currency dunning attempt tracking via a JSONB column (`dunning_currency_attempts`) on `Customer`.                                                                                 
   
  Key changes:                                                                                                                                                                                 
  - Counter moved from `ProcessAttemptService` to `BulkProcessService` — incremented once per evaluation round (atomically, before enqueuing jobs), eliminating the race condition between
  concurrent per-currency jobs                                                                                                                                                                 
  - Per-currency max_attempts — each currency independently tracks its attempt count (e.g. {"USD" => 2, "EUR" => 1}), so one currency can't consume another's attempts
  - `dunning_campaign.finished` webhook — now fires from `BulkProcessService` exactly once, when all currencies are exhausted                                                                      
  - N+1 fix in `UpdateService#reset_customers_if_no_threshold_match` — uses single `overdue_balances` GROUP BY query instead of per-threshold queries                                              
  - Migration seeds existing `last_dunning_campaign_attempt` into JSONB for backward compatibility                                                 
  - Timing fix — changed >= to > in `days_between_attempts_satisfied?` to match the strict `.past?` semantics that `ProcessAttemptService` previously enforced


mariohd (approved):

aquinofb (approved):
LGTM!

app/models/customer.rb
```diff
@@ -310,17 +310,33 @@ def empty_billing_and_shipping_address?
...
       last_dunning_campaign_attempt_at: nil
     )
   end
 
+  def reset_dunning_campaign_for_currency!(currency)
+    attempts = dunning_currency_attempts.dup
+    attempts.delete(currency.to_s)
+    update!(
+      dunning_currency_attempts: attempts,
+      last_dunning_campaign_attempt: 0,
```
ancorcruz, line 336: is this reseting the `last_dunning_campaign_attempt` to zero as the field is not used anymore? In that case why not just ignore it and flag it as ignored / deprecated column in the customer model?

spec/models/customer_spec.rb
```diff
@@ -1106,21 +1106,121 @@
...
+        create(:invoice, customer: customer, payment_overdue: true, currency: "USD", total_amount_cents: 3_00)
+        create(:invoice, customer: customer, payment_overdue: true, currency: "EUR", total_amount_cents: 10_00)
+      end
+
+      it "returns per-currency breakdown" do
+        expect(customer.overdue_balances).to eq("USD" => 5_00, "EUR" => 10_00)
+      end
+    end
+
+    context "when there are non-overdue invoices" do
```
ancorcruz, line 1144: this one reads weird... I understand this is when customer has overdue and not overdue invoices (mixed) but the context naming is misleading.

ancorcruz (approved):
LGTM

db/migrate/20260401143315_add_dunning_currency_attempts_to_customers.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class AddDunningCurrencyAttemptsToCustomers < ActiveRecord::Migration[8.0]
+  def up
+    add_column :customers, :dunning_currency_attempts, :jsonb, default: {}, null: false
+
+    # Seed JSONB from existing single-currency counter to prevent
+    # re-dunning customers who already reached max_attempts
+    safety_assured do
```
vincent-pochet: I would suggest to use two distinct migrations, one to add the new field and a second one to fill the values

app/services/dunning_campaigns/bulk_process_service.rb
```diff
@@ -30,46 +30,78 @@ def initialize(customer)
...
+            .thresholds
+            .where(currency:)
+            .find_by("amount_cents <= ?", amount_cents)
+        end
       end
 
-      def max_attempts_reached?
-        customer.last_dunning_campaign_attempt >= dunning_campaign.max_attempts
+      def increment_per_currency_attempts(thresholds)
+        attempts = customer.dunning_currency_attempts.dup
```
vincent-pochet, line 81: With the previous implementation, the increment was made in the DB transaction, meaning that in case of error, it was just rolled-back. It's not the case with this change. Not sure if the risk is high or if it can cause a problem, but I think it should be checked.

Also, I seems that the `DunningCampaigns::BulkProcessJob` does not define any `uniqueness` handling. What happen if some concurrent call are playing at the same time? Should we enforce this update?

vincent-pochet (approved):

app/services/dunning_campaigns/bulk_process_service.rb
lovrocolic: In the old code the increment could live inside the `PaymentRequests::CreateService` transaction because there was one `ProcessAttemptJob` per customer per round (single threshold → single counter). With per-currency tracking a single evaluation can dispatch N jobs against the same customer, each incrementing the same `dunning_currency_attempts` hash. Keeping the write in `ProcessAttemptService` would reintroduce the exact race the refactor was designed to eliminate: two concurrent jobs both read {"USD" => 1}, both write {"USD" => 2}, counter under-counts. Moving the write into `BulkProcessService` before enqueueing is what makes per-currency counting safe.

You are right about the downside, we can maybe fix the race condition problem with the advisory locks, but not 100% sure if needed since in worst case scenario, there would just be less attempts for certain customer. wdyt? In old behavior we also had similar issue: counter was increased and we scheduled payment job which could fail as well...

Also, concurrent invocations are already prevented upstream: `Clock::ProcessDunningCampaignsJob` (the only caller) has `unique :until_executed, on_conflict: :log`

app/models/customer.rb
lovrocolic: Not used, but I would like to keep old logic so that we don't miss any current behaviour and that we preserve current field value. After we consider new upgrade as stable we can deprecate the old field. wdyt?

PR #5308, opened 2026-04-09
perf(invoices): Preload offset_amount_cents for invoices collections

D1353L:
## Context

The /api/v1/invoices endpoint and GraphQL invoice list queries produce N+1 queries on the credit_notes table. For each invoice in the response, offset_amount_cents executes `SELECT SUM(offset_amount_cents) FROM credit_notes WHERE invoice_id = ? AND status = ?` individually.

A batch preloading pattern already existed in the CSV data export service but was never applied to the API or GraphQL layers.

## Description

Add Invoice.preload_offset_amounts class method that loads all offset amounts for a collection of invoices in a single `SUM ... GROUP BY invoice_id` query, caching the result on each instance via an instance variable. This replaces 2N queries with 1.

  Apply preloading to all invoice collection endpoints:
  - REST API: InvoiceIndex concern (serves /api/v1/invoices and /api/v1/customers/:id/invoices)
  - GraphQL: InvoicesResolver, Customers::InvoicesResolver, CustomerPortal::InvoicesResolver

### DIFF
--- a/app/graphql/resolvers/invoices_resolver.rb
+++ b/app/graphql/resolvers/invoices_resolver.rb
@@ -82,7 +82,9 @@ module Resolvers
         }
       )
 
-      result.invoices
+      invoices = result.invoices
+      Invoice.preload_offset_amounts(invoices)
+      invoices
     end
   end
 end
--- a/app/graphql/resolvers/invoices_resolver.rb
+++ b/app/graphql/resolvers/invoices_resolver.rb
@@ -82,6 +82,8 @@ module Resolvers
         }
       )
 
+      return result_error(result) unless result.success?
+
       invoices = result.invoices
       Invoice.preload_offset_amounts(invoices)
       invoices
--- a/app/graphql/resolvers/invoices_resolver.rb
+++ b/app/graphql/resolvers/invoices_resolver.rb
@@ -84,9 +84,7 @@ module Resolvers
 
       return result_error(result) unless result.success?
 
-      invoices = result.invoices
-      Invoice.preload_offset_amounts(invoices)
-      invoices
+      Invoice.preload_offset_amounts(result.invoices)
     end
   end
 end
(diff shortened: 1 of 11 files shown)


app/models/invoice.rb
```diff
@@ -294,10 +296,22 @@ def charge_pay_in_advance_proration_range(fee, timestamp)
     }
   end
 
-  # Caches offset_amount_cents in the invoice instance to avoid N+1 queries when exporting invoices.
-  # Allows batch calculation of offset amounts for many invoices in a single aggregated query.
-  def save_precalculated_offset_amount_cents(offset_amount_cents)
-    @precalculated_offset_amount_cents = offset_amount_cents
+  # Batch-loads offset_amount_cents for a collection of invoices in a single query,
+  # caching the result on each instance to avoid N+1 queries during serialization.
+  def self.preload_offset_amounts(invoices)
```
ancorcruz: move this class method to the top of the class along with the other class methods.

app/graphql/resolvers/invoices_resolver.rb
```diff
@@ -82,7 +82,11 @@ def resolve( # rubocop:disable Metrics/ParameterLists
         }
       )
 
-      result.invoices
+      return result_error(result) unless result.success?
+
+      invoices = result.invoices
+      Invoice.preload_offset_amounts(invoices)
```
ancorcruz: could we return the invoices instead of modifying the objects as side effect of the class method when calling this?

ancorcruz (approved):
LGTM

app/graphql/resolvers/invoices_resolver.rb
D1353L: Thanks Ancor! Modifying the passed objects seems unavoidable here as this is intentional preloading. I've just slightly refactored it to return the modified collection.

PR #5309, opened 2026-04-09
feat(clickhouse): Add models layer for event store migrations

vincent-pochet:
## Context

## Description

This PR adds the model layer for monitoring the event store migration process.
- `EnrichedStoreMigration`: Handle the migration at organization level
- `EnrichedStoreSubscriptionMigration`: Handle to migration at subscription level

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -56,6 +56,8 @@ Lago/NoDropColumnOrTable:
     - "db/migrate/20260116121019_remove_role_from_invites.rb"
     - "db/migrate/20260119162712_drop_subscription_fixed_charge_units_overrides.rb"
     - "db/migrate/20260129145352_drop_properties_from_enriched_events.rb"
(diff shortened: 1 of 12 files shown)


toommz (approved):

toommz (approved):
Just one question about `aasm`, we use it mainly to check state transitions right? As we keep logic in services I'm wondering what we can put into those events.

Asking as we have a lot of states in the Quotes feature

vincent-pochet:
> Just one question about `aasm`, we use it mainly to check state transitions right? As we keep logic in services I'm wondering what we can put into those events.
> 
> Asking as we have a lot of states in the Quotes feature

@toommz My idea was only to manage the transition and nothing more.
With the current pattern the model should only ""validate" the transitions and the logic itself should remain in the service layer.
Happy to talk about it :)

PR #5311, opened 2026-04-10
misc(clickhouse): Extract enriched_events:compare core logic into a service

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/5309

## Description

This PR extract the core logic of `enriched_events:compare` into a new `Events::Stores::Clickhouse::EnrichedStoreMigration::ComparisonService` with testing coverage.

### DIFF
--- a/app/services/events/stores/clickhouse/enriched_store_migration/comparison_service.rb
+++ b/app/services/events/stores/clickhouse/enriched_store_migration/comparison_service.rb
@@ -1,0 +1,157 @@
+# frozen_string_literal: true
+
+require "benchmark"
+
+module Events
+  module Stores
+    module Clickhouse
+      module EnrichedStoreMigration
+        class ComparisonService < BaseService
+          Result = BaseResult[:diff_count, :fee_details, :legacy_elapsed, :enriched_elapsed]
+
+          FieldDiff = Data.define(:legacy, :enriched)
+          FeeValues = Data.define(:units, :amount_cents, :events_count, :total_aggregated_units)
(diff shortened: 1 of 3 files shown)


app/services/events/stores/clickhouse/enriched_store_migration/comparison_service.rb
```diff
@@ -0,0 +1,149 @@
...
+
+          attr_reader :subscription, :organization, :deduplicate
+
+          def compute_usage(enriched:)
+            usage_result = nil
+
+            ActiveRecord::Base.transaction do
+              if enriched
+                organization.enable_feature_flag!(:enriched_events_aggregation)
+                organization.update!(clickhouse_deduplication_enabled: deduplicate, pre_filter_events: true)
```
toommz, line 61: Would it leave the feature flag enabled after the service has been run?

toommz: Oh we roll back the transaction at the end of the block 🤦

vincent-pochet: Exactly :)

toommz (approved):
If the diff part of fees becomes more complex to compute we might want to extract a (value) object for that, but at the moment it does the job

vincent-pochet:
> If the diff part of fees becomes more complex to compute we might want to extract a (value) object for that, but at the moment it does the job

@toommz thanks for the review!
The value object is a great idea, I will have a look

PR #5314, opened 2026-04-13
feat(payment-gated-subs): add webhooks for incomplete and canceled subscriptions

ancorcruz:
## Context

Payment-gated subscription activation requires notifying API consumers when a subscription enters the `incomplete` state (waiting for payment confirmation) and when it transitions to `canceled` (payment failed or timed out).

## Description

Add two new outbound webhook services following the existing subscription webhook pattern: `subscription.incomplete` and `subscription.canceled`. Both serialize the subscription with plan and customer includes via `V1::SubscriptionSerializer`. Register the new webhook types in the `SendWebhookJob::WEBHOOK_SERVICES` mapping so they can be dispatched via `SendWebhookJob.perform_later`.


osmarluz (approved):
🚀

PR #5315, opened 2026-04-13
feat(payment-gated-subs): add event guards for incomplete subscriptions

ancorcruz:
## Context

Incomplete subscriptions have a `started_at` in the past, so they pass time-boundary queries unlike pending subscriptions. Without explicit guards, customer usage events would be processed for billing and pay-in-advance charges during the incomplete window.

Conversely, internal fixed charge events must still be emitted for incomplete subscriptions to produce correct activation invoices.

## Description

Exclude incomplete subscriptions from `Events::PostProcessService` query so usage events are ignored for billing, fee generation, and activity tracking while a subscription awaits payment confirmation.

Include incomplete subscriptions in
`FixedCharges::EmitEventsForActiveSubscriptionsService` so internal billing events needed for gated invoice creation are emitted correctly.

### DIFF
--- a/app/services/fixed_charges/emit_events_for_active_subscriptions_service.rb
+++ b/app/services/fixed_charges/emit_events_for_active_subscriptions_service.rb
@@ -31,11 +31,11 @@ module FixedCharges
       # This handles cases like plan overrides where the subscription hasn't been updated yet
       # otherwise, emit events for all active subscriptions on the plan
       if subscription
-        # Only emit events for active subscriptions, even when explicitly provided
+        # Emit events for active and incomplete subscriptions
         # Pending subscriptions will have events created when they activate
-        subscription.active? ? [subscription] : []
+        (subscription.active? || subscription.incomplete?) ? [subscription] : []
       else
-        fixed_charge.plan.subscriptions.active
+        fixed_charge.plan.subscriptions.where(status: %i[active incomplete])
       end
     end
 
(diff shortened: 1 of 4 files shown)


app/services/fixed_charges/emit_events_for_active_subscriptions_service.rb
osmarluz, line 1: @ancorcruz it would be good to have a separate PR to rename this service, since it does not do what it claims  considering the new changes

spec/services/fixed_charges/emit_events_for_active_subscriptions_service_spec.rb
```diff
@@ -77,6 +77,43 @@
...
+    context "when a provided subscription is incomplete" do
+      let(:subscription) do
+        create(:subscription, :incomplete, :anniversary, plan:, started_at: 1.day.ago, subscription_at: 1.day.ago)
+      end
+
+      it "creates fixed charge event for the incomplete subscription" do
+        expect { result }.to change(FixedChargeEvent, :count).by(1)
+
+        event = FixedChargeEvent.find_by(subscription:, fixed_charge:)
+        expect(event).to be_present
```
osmarluz, line 113: @ancorcruz we should also consider including the events on the service result so we don't have to explicitly find the create events like this

osmarluz (approved):
LGTM! Just left a couple of comments that can be addressed separately 🚀

app/services/fixed_charges/emit_events_for_active_subscriptions_service.rb
ancorcruz: yep, separate PR for that, thanks.

PR #5316, opened 2026-04-13
feat(payment-gated-subs): guard manual operations on incomplete subscriptions

ancorcruz:
## Context

Incomplete subscriptions must only be resolved by the system through payment success, payment failure, or timeout. Manual operations like terminate or update would bypass the activation flow and leave the subscription in an inconsistent state.

## Description

Reject terminate and update requests on incomplete subscriptions with a not_allowed error (code: subscription_incomplete). Silence the invoice.payment_failure webhook for open subscription invoices since payment failures on gated invoices are handled through the subscription.canceled webhook instead.

### DIFF
--- a/app/services/invoices/payments/deliver_error_webhook_service.rb
+++ b/app/services/invoices/payments/deliver_error_webhook_service.rb
@@ -9,6 +9,8 @@ module Invoices
       end
 
       def call_async
+        return result if invoice.subscription? && invoice.open?
+
         if invoice.credit? && (invoice.open? || invoice.visible?)
           wallet_transaction = invoice.fees.credit.first.invoiceable
           SendWebhookJob.perform_later("wallet_transaction.payment_failure", wallet_transaction, params)
--- a/app/services/invoices/payments/deliver_error_webhook_service.rb
+++ b/app/services/invoices/payments/deliver_error_webhook_service.rb
@@ -9,8 +9,6 @@ module Invoices
       end
 
       def call_async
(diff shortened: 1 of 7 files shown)


app/services/subscriptions/update_service.rb
```diff
@@ -19,6 +19,7 @@ def initialize(subscription:, params:)
 
     def call
       return result.not_found_failure!(resource: "subscription") unless subscription
+      return result.not_allowed_failure!(code: "subscription_incomplete") if subscription.incomplete?
```
osmarluz, line 22: @ancorcruz I modified `app/services/subscriptions/validate_service.rb` to include this check on the validation, but it'll be unnecessary since we're already returning a `not_allowed_failure!` here. I'll remove this from PR #5281 then

app/services/invoices/payments/deliver_error_webhook_service.rb
```diff
@@ -9,6 +9,8 @@ def initialize(invoice, params)
       end
 
       def call_async
+        return result if invoice.subscription? && invoice.open?
```
osmarluz, line 12: @ancorcruz since the checks below already perform `invoice.credit?` and `invoice.visible?` to send the webhooks, we don't need any changes here

app/services/subscriptions/update_service.rb
osmarluz: @ancorcruz done!

osmarluz (approved):
🚀

PR #5317, opened 2026-04-13
feat(payment-gated-subs): add activation rule evaluate services

ancorcruz:
## Context

Payment-gated subscriptions need a mechanism to evaluate activation rules and transition them through their lifecycle. Rules start as inactive and must be evaluated to determine if they apply, then resolved when the payment outcome is known.

## Description

Add EvaluateService as a delegator that iterates a subscription's activation rules and dispatches to the type-specific evaluate service. Add Payment::EvaluateService that handles the payment rule state machine: inactive rules transition to pending (applicable) or not_applicable based on plan billing configuration; pending rules transition to satisfied, failed, or expired based on payment outcome. Add factory alias payment_subscription_activation_rule for clarity.

### DIFF
--- a/app/services/subscriptions/activation_rules/evaluate_service.rb
+++ b/app/services/subscriptions/activation_rules/evaluate_service.rb
@@ -3,17 +3,22 @@
 module Subscriptions
   module ActivationRules
     class EvaluateService < BaseService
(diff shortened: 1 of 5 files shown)


osmarluz (approved):
🚀

PR #5318, opened 2026-04-13
feat(payment-gated-subs): allow fee creation for incomplete subscriptions

ancorcruz:
## Context

When a subscription enters the incomplete state for payment gating, the billing flow still needs to create subscription and fixed charge fees to produce the activation invoice that will be sent to the PSP.

## Description

Add incomplete status to the conditions in should_create_subscription_fee and should_create_fixed_charge_fees in CalculateFeesService so fees are generated for incomplete subscriptions alongside active ones.


osmarluz (approved):
🚀

PR #5319, opened 2026-04-13
feat(payment-gated-subs): keep gated invoices open in TransitionToFinalStatusService

ancorcruz:
## Context

Payment-gated invoices must remain in open status until payment is resolved. The transition service should not finalize or close them.

## Description

Add a guard in TransitionToFinalStatusService that preserves the open status for subscription-gated invoices with a positive total amount. Zero-amount gated invoices continue through the existing finalize/close logic unchanged.

Refactor the spec to use subject(:result) pattern.

### DIFF
--- a/app/services/invoices/transition_to_final_status_service.rb
+++ b/app/services/invoices/transition_to_final_status_service.rb
@@ -10,7 +10,9 @@ module Invoices
     end
 
     def call
-      if should_finalize_invoice?
+      if invoice.subscription_gated? && invoice.total_amount_cents.positive?
+        # Keep open for payment-gated invoices awaiting payment resolution
+      elsif should_finalize_invoice?
         Invoices::FinalizeService.call!(invoice: invoice)
       else
         invoice.status = :closed
--- a/app/services/invoices/transition_to_final_status_service.rb
+++ b/app/services/invoices/transition_to_final_status_service.rb
@@ -10,14 +10,17 @@ module Invoices
     end
 
     def call
-      if invoice.subscription_gated? && invoice.total_amount_cents.positive?
-        # Keep open for payment-gated invoices awaiting payment resolution
-      elsif should_finalize_invoice?
+      result.invoice = invoice
+
+      # Keep open for payment-gated invoices awaiting payment resolution
+      return result if invoice.subscription_gated? && invoice.total_amount_cents.positive?
+
+      if should_finalize_invoice?
         Invoices::FinalizeService.call!(invoice: invoice)
       else
         invoice.status = :closed
       end
-      result.invoice = invoice
+
       result
     end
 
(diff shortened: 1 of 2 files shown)


app/services/invoices/transition_to_final_status_service.rb
```diff
@@ -10,7 +10,9 @@ def initialize(invoice:)
     end
 
     def call
-      if should_finalize_invoice?
+      if invoice.subscription_gated? && invoice.total_amount_cents.positive?
+        # Keep open for payment-gated invoices awaiting payment resolution
+      elsif should_finalize_invoice?
```
osmarluz, line 15: @ancorcruz it would be better to have an early return here instead of an empty block:


```suggestion
      # Keep open for payment-gated invoices awaiting payment resolution
      if invoice.subscription_gated? && invoice.total_amount_cents.positive?
        result.invoice = invoice
        return result
      end
    
      if should_finalize_invoice?
```

spec/services/invoices/transition_to_final_status_service_spec.rb
osmarluz, line 1: @ancorcruz we're missing the test scenario `"when invoice is subscription_gated with zero amount"` here

spec/services/invoices/transition_to_final_status_service_spec.rb
```diff
@@ -18,66 +18,103 @@
...
+    end
+  end
 
-      context "with billing entity and customer settings defined to not finalize" do
-        let(:organization_setting) { "false" }
-        let(:customer_setting) { "skip" }
+  context "when invoice is subscription_gated with positive amount" do
+    let(:fees_amount_cents) { 100 }
+    let(:plan) { create(:plan, organization:, pay_in_advance: true) }
+    let(:subscription) { create(:subscription, :incomplete, organization:, customer:, plan:) }
```
osmarluz, line 47: @ancorcruz just a suggestion, but on `main` we have a trait for activation rules in case you wanna use it here:


```suggestion
    let(:subscription) { create(:subscription, :incomplete, :with_activation_rules, organization:, customer:, plan:) }
```

app/services/invoices/transition_to_final_status_service.rb
ancorcruz: done it. thanks

spec/services/invoices/transition_to_final_status_service_spec.rb
ancorcruz: added

spec/services/invoices/transition_to_final_status_service_spec.rb
ancorcruz: good catch, thanks.

osmarluz (approved):
🚀

PR #5320, opened 2026-04-13
feat(payment-gated-subs): skip grace period for gated invoices

ancorcruz:
## Context

Payment-gated invoices must attempt payment immediately rather than waiting for the grace period to expire. The invoice should go directly to open status, not draft.

## Description

Add a subscription_gated parameter to CreateGeneratingService. When true the grace period is skipped, setting the issuing date and expected finalization date to the current date regardless of customer grace period configuration.


osmarluz (approved):
🚀

PR #5321, opened 2026-04-13
feat(organizations): Add slug field to Organization model

domenicofalco:
## Context

Organizations have no human-readable identifier. Frontend URLs are flat
(/customers, /plans) with no organization context, making them impossible
to bookmark, share, or use reliably in multi-tab workflows. This is the
backend foundation for the organization slug URL architecture that will
prefix all authenticated routes with the org slug (e.g. /acme-corp/customers).

The backend releases independently before the frontend so every org
silently gets a slug before URLs change.

## Description

Introduce a `slug` column on organizations with auto-generation,
validation, GraphQL exposure, REST API support, and slug editing via
the existing updateOrganization mutation.

Phase 1 — Model + Migration + Backfill:
- Extract slug logic into Organizations::Sluggable concern with
  SLUG_FORMAT regex, RESERVED_SLUGS list, validations (presence,
  uniqueness, length 3-40, format, exclusion), and before_validation
  callback on create
- Slug auto-generated from org name via transliterate + parameterize,
  with fallback to org-{random} for non-transliterable scripts,
  numeric-only, too-short, or reserved names
- Collision handling appends random 3-char alphanumeric suffix to avoid
  leaking org count
- Single migration adds column, backfills all existing orgs with
  self-contained generation logic, and sets NOT NULL constraint
- Separate migration for concurrent unique index

Phase 2 — GraphQL API:
- Add slug field (String!) to OrganizationType and
  CurrentOrganizationType
- Add optional slug argument to UpdateOrganizationInput

Phase 3 — Services + REST API:
- Accept slug param in Organizations::UpdateService
- Whitelist slug in V1 OrganizationsController input_params
- Include slug in V1::OrganizationSerializer response
- Add locale mappings for too_short (value_is_too_short) and exclusion
  (value_is_reserved) to ensure consistent error codes across all APIs

Fixes LAGO-1343


app/models/concerns/organizations/sluggable.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    RESERVED_SLUGS = %w[
+      auth login sign-up forgot-password reset-password invitation
+      customer-portal 404 forbidden api admin graphql webhooks google okta
+      settings new design-system devtool
+      customers customer plans plan invoices invoice subscriptions
+      coupons coupon add-ons add-on billable-metrics billable-metric
+      credit-notes analytics analytics-v2 forecasts payments payment
+      features feature tax webhook api-keys create update duplicate
+    ].freeze
```
ancorcruz, line 17: why this list?

app/models/credit_note.rb
```diff
@@ -205,39 +205,41 @@ def status_changed_to_finalized?
...
-#  offset_amount_cents                     :bigint           default(0), not null
-#  offset_amount_currency                  :string
-#  precise_coupons_adjustment_amount_cents :decimal(30, 5)   default(0.0), not null
-#  precise_taxes_amount_cents              :decimal(30, 5)   default(0.0), not null
-#  reason                                  :integer          not null
-#  refund_amount_cents                     :bigint           default(0), not null
-#  refund_amount_currency                  :string
-#  refund_status                           :integer
-#  refunded_at                             :datetime
-#  status                                  :integer          default("finalized"), not null
```
ancorcruz, line 228: these changes are unrelated to this PR, aren't they?

app/models/credit.rb
```diff
@@ -82,6 +82,7 @@ def invoice_coupon_display_name
 #  id                             :uuid             not null, primary key
 #  amount_cents                   :bigint           not null
 #  amount_currency                :string           not null
+#  apply_after_finalization       :boolean          default(FALSE), not null
```
ancorcruz: these changes are unrelated to this PR, aren't they?

app/services/organizations/update_service.rb
```diff
@@ -27,6 +27,7 @@ def call
       organization.country = params[:country]&.upcase if params.key?(:country)
       organization.default_currency = params[:default_currency]&.upcase if params.key?(:default_currency)
       organization.document_number_prefix = params[:document_number_prefix] if params.key?(:document_number_prefix)
+      organization.slug = params[:slug] if params.key?(:slug)
```
ancorcruz: shall we normalize the slug before assign it? `params[:slug]&.strip&.downcase` would prevent avoidable validation errors from trailing spaces or mixed case

ancorcruz (approved):
LGTM

spec/factories/organizations.rb
```diff
@@ -7,6 +7,7 @@
     end
 
     name { Faker::Company.name }
+    sequence(:slug) { |n| "#{Faker::Internet.slug(glue: "-")}-#{n}" }
```
ancorcruz: I would use a more deterministic option instead...


```suggestion
    sequence(:slug) { |n| "test-org-#{n}" }
```

app/models/concerns/organizations/sluggable.rb
domenicofalco: These are reserved frontend routes (auth flows, app pages, system paths like /api, /graphql). Since the slug will be used as a URL path segment (/:slug/customers), we need to block collisions server-side too, not just on the frontend. It is fundamentally a safety net to prevent URL routing conflicts.

Details here: https://www.notion.so/getlago/Solution-Dive-in-Organization-Slug-BE-1d0766c2a18080a5a19dc58e218720a2

app/services/organizations/update_service.rb
domenicofalco: great point, fixed!

app/models/credit.rb
domenicofalco: fixed

spec/factories/organizations.rb
domenicofalco: fixed

app/models/credit_note.rb
domenicofalco: right, these were auto-generated by the annotate gem when running migrations locally. reverted.

db/migrate/20260416111923_backfill_organization_slugs.rb
```diff
@@ -0,0 +1,72 @@
...
+    safety_assured do
+      add_column :organizations, :slug, :string
+
+      Organization.unscoped.find_each do |org|
+        candidate = generate_slug_for(org.name)
+        candidate = resolve_collision(candidate)
+        org.update_column(:slug, candidate) # rubocop:disable Rails/SkipsModelValidations
+      end
+
+      change_column_null :organizations, :slug, false
```
groyoh: Can you split the DDL and DML into multiple independent migrations? 1 for column, 1 for migration. Also setting `NOT NULL` here is quite dangerous as there could be a new organization created in the middle of the migration and breaking the migration altogether.

If you really want the `NOT NULL`, you'll need to either:

1. Define a default random value before migrating the data
2. Make 2 OSS releases: one with the first migration with `NULL` and a second one which applies a `NOT NULL`.

groyoh (changes requested):

db/migrate/20260416111923_backfill_organization_slugs.rb
groyoh: Also it would be better to try and reuse the existing code rather than copy it as it increases the risk of errors.

domenicofalco: @groyoh thanks for the insights really! Addressed both points:

**Migration split:** Split into 3 independent migrations: 
- DDL (add column with a random default as safety net)
- DML (backfill + NOT NULL + remove default)
- unique index

This way even if a new org is created between migration 1 and 2, the slug column is never NULL. The default is removed after backfill since the model concern handles it from there. WDYT?

**Code reuse:** I assumed migrations should be self-contained and not reference application code imported here, if the concern changes in the future, anyone running the migration later would get different behavior. I kept the logic inlined to ensure the migration is reproducible and immutable. Let me know if there's a better way to handle this

db/migrate/20260416111922_add_slug_to_organizations.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddSlugToOrganizations < ActiveRecord::Migration[8.0]
+  def up
+    safety_assured do
+      add_column :organizations, :slug, :string, default: -> { "'org-' || substr(md5(random()::text), 1, 8)" }
```
groyoh: To avoid blocking the table, set default to `nil` and then update the default for new rows. This way new rows created during the migration will have the volatile default:

```suggestion
      add_column :organizations, :slug, :string, default: nil
      change_column_default from: nil, to: -> { "'org-' || substr(md5(random()::text), 1, 8)" }
```

db/migrate/20260416111923_backfill_organization_slugs.rb
groyoh: I believe you're still mixing data migration and structure update in the same migration. Could you move the `change_column_null` and `change_column_default` to the next migration.

domenicofalco: done in a separate migration

db/migrate/20260416111922_add_slug_to_organizations.rb
domenicofalco: done

groyoh (approved):

app/models/concerns/organizations/sluggable.rb
```diff
@@ -0,0 +1,59 @@
...
+      return if slug.present?
+
+      candidate = ActiveSupport::Inflector.transliterate(name.to_s)
+        .parameterize
+        .tr("_", "-")
+        .gsub(/-{2,}/, "-")
+        .gsub(/\A-|-\z/, "")
+        .truncate(40, omission: "")
+
+      if candidate.length < 3 || candidate.match?(/\A\d+\z/) || RESERVED_SLUGS.include?(candidate)
```
groyoh, line 43: ⚠️ `.gsub(/\A-|-\z/, "")` runs **before** `.truncate(40, omission: "")`, so truncation can reintroduce a trailing hyphen that fails `SLUG_FORMAT`.

Example: an org named `"Alpha Beta Gamma Delta Epsilon Zeta Eta T"` parameterizes to `"alpha-beta-gamma-delta-epsilon-zeta-eta-t"` (41 chars). Truncating to 40 yields `"alpha-beta-gamma-delta-epsilon-zeta-eta-"` — which fails the format validation and prevents org creation.

The fix is to move the cleanup after truncation:

```ruby
candidate = ActiveSupport::Inflector.transliterate(name.to_s)
  .parameterize
  .tr("_", "-")
  .gsub(/-{2,}/, "-")
  .truncate(40, omission: "")
  .gsub(/\A-|-\z/, "")
```

The same issue exists in the backfill migration's `generate_slug_for` method — there it silently persists invalid slugs via `update_column`.

groyoh (commented):
Two issues found — one bug that can prevent org creation, one cosmetic issue with collision-suffixed slugs. The overall design (phased migrations, concern extraction, reserved slug list, collision handling) is solid.

app/models/concerns/organizations/sluggable.rb
```diff
@@ -0,0 +1,59 @@
...
+        loop do
+          candidate = "org-#{SecureRandom.alphanumeric(5).downcase}"
+          break unless self.class.exists?(slug: candidate)
+        end
+      else
+        base = candidate.truncate(36, omission: "")
+        while self.class.exists?(slug: candidate)
+          suffix = SecureRandom.alphanumeric(3).downcase
```
groyoh, line 51: `base = candidate.truncate(36, omission: "")` can also end with a trailing hyphen, producing slugs like `some-base--abc` (double hyphen). This passes `SLUG_FORMAT` so it's not a validation failure, but it looks wrong in URLs. Applying the same trailing-hyphen cleanup to `base` would prevent this:

```ruby
base = candidate.truncate(36, omission: "").gsub(/-\z/, "")
```

domenicofalco: fixed

app/models/concerns/organizations/sluggable.rb
domenicofalco: good catch! fixed

groyoh (approved):

PR #5322, opened 2026-04-13
feat(payment-gated-subs): use empty payment reference for gated invoices

ancorcruz:
## Context

Open invoices for payment-gated subscriptions do not have a real invoice number assigned yet. The number is only set when the invoice transitions from open to finalized on payment success.

## Description

Extract payment reference into a method and return an empty string when the invoice is subscription_gated. This prevents sending a DRAFT placeholder number to the payment provider.

### DIFF
--- a/app/services/invoices/payments/create_service.rb
+++ b/app/services/invoices/payments/create_service.rb
@@ -54,7 +54,7 @@ module Invoices
         payment_result = ::PaymentProviders::CreatePaymentFactory.new_instance(
           provider:,
           payment:,
-          reference: "#{invoice.billing_entity.name} - Invoice #{invoice.number}",
+          reference: payment_reference,
           metadata: {
             lago_invoice_id: invoice.id,
             lago_customer_id: invoice.customer_id,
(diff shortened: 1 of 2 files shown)


app/services/invoices/payments/create_service.rb
```diff
@@ -169,6 +169,12 @@ def deliver_error_webhook(e)
         })
       end
 
+      def payment_reference
+        return "" if invoice.subscription_gated?
```
osmarluz, line 173: @ancorcruz this approach works for Stripe but is risky for Adyen, since the `reference` is mandatory for [their API](https://docs.adyen.com/api-explorer/Checkout/70/post/payments#request-reference) and they recommend it's unique:

> The reference to uniquely identify a payment. This reference is used in all communication with you about the payment status. To provide multiple references for one transaction, separate the reference values with the hyphen (-) character. We strongly recommend that you use a unique value for each transaction. Maximum length: 80 characters.

We could have `"#{invoice.billing_entity.name} - Invoice #{invoice.id}"` instead of `""`, but it's better to confirm with product about this behavior.

ancorcruz: legit, lets us that for all of them, it should be fine that way. Thanks.

osmarluz (approved):
🚀

PR #5323, opened 2026-04-13
fix(clickhouse): events_enriched cloud migration

vincent-pochet:
## Description

This PR fixes the cloud migration for the `events_enriched` table.

NOTE: this migration is executed **manually** when creating a new cluster, so it is safe to update it directly.


groyoh (approved):
Would it be possible to have a process to ensure these migrations execute properly ?

vincent-pochet:
> Would it be possible to have a process to ensure these migrations execute properly ?

@groyoh it was performed manually bye Jeremy yesterday. Appart from that, it's not part of any automated process for now. It might change in the future.

groyoh:
@vincent-pochet sorry my point wasn't clear. I was thinking of a kind of automated test in the codebase which runs those migrations.

PR #5326, opened 2026-04-13
fix(charge): Add advisory lock to avoid updating same charge children…

jdenquin:
## Summary                                                                                                                                                                                
                                                                                                                                                                                            
  - Add PostgreSQL advisory lock in `Charges::UpdateChildrenService` keyed on the parent charge ID to prevent concurrent cascades from overlapping                                          
  - Add `retry_on WithAdvisoryLock::FailedToAcquireLock` to `Charges::UpdateChildrenBatchJob` so blocked jobs retry later                                                                   
                                                                                                                                                                                            
  ## Context                                                                                                                                                                                
                                                                                                                                                                                            
  When the same parent charge is updated multiple times in quick succession, each update triggers a separate `UpdateChildrenJob`, which spawns overlapping `UpdateChildrenBatchJob`s for the
   same child charges. These concurrent jobs update the same `charge_filters` rows simultaneously, causing prolonged row-level locking (queries lasting over 1 hour) and                    
  `PG::TRDeadlockDetected` errors.                                                                                                                                                          
                                                                                                                                                                                            
  The advisory lock ensures only one cascade runs at a time per parent charge. With `timeout_seconds: 0`, a batch job that can't acquire the lock fails immediately instead of waiting, and 
  the `retry_on` with `polynomially_longer` backoff re-enqueues it for later — avoiding both the lock contention and redundant concurrent work.

### DIFF
--- a/app/jobs/charges/update_children_batch_job.rb
+++ b/app/jobs/charges/update_children_batch_job.rb
@@ -3,6 +3,7 @@
 module Charges
   class UpdateChildrenBatchJob < ApplicationJob
     queue_as :low_priority
+    retry_on WithAdvisoryLock::FailedToAcquireLock, wait: :polynomially_longer, attempts: 5
 
     def perform(child_ids:, params:, old_parent_attrs:, old_parent_filters_attrs:, old_parent_applied_pricing_unit_attrs:)
       Rails.logger.info("Charges::UpdateChildrenBatchJob - Started the execution for parent charge with ID: #{old_parent_attrs["id"]}")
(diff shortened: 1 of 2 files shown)


app/jobs/charges/update_children_batch_job.rb
```diff
@@ -3,6 +3,7 @@
 module Charges
   class UpdateChildrenBatchJob < ApplicationJob
     queue_as :low_priority
+    retry_on WithAdvisoryLock::FailedToAcquireLock, wait: :polynomially_longer, attempts: 5
```
groyoh, line 6: You may want to increase the retry number here.

groyoh (approved):

PR #5330, opened 2026-04-14
fix(invoices): Unblock regenerating voided advance_charges invoice

D1353L:
## Context
Related to #5387

Regenerating a voided `advance_charges` invoice fails with a 500 Internal Server Error. Three chained bugs were identified:

**Bug 1 — `duplicate_fee` copies `pay_in_advance_event_transaction_id` (UniqueViolation)**

`app/services/invoices/regenerate_from_voided_service.rb:205-216`

Primary production failure — the UI always sends a fee id, so duplicate_fee is always called.

duplicate_fee uses `voided_fee.dup` which preserves `pay_in_advance_event_transaction_id`. Since the original fee still exists on the voided invoice, both fees share the same (`pay_in_advance_event_transaction_id`, `charge_id`) pair, violating the unique index `idx_pay_in_advance_duplication_guard_charge`.

---
**Bug 2 — Missing invoice_subscriptions for advance_charges**

`app/services/invoices/regenerate_from_voided_service.rb:24`

`create_invoice_subscriptions` is only called when `invoice_type == "subscription"`, but advance_charges invoices also have `invoice_subscriptions`. Without them, `AdjustedFees::CreateService` cannot find the subscription via `invoice.subscriptions`. The error is silently swallowed by `process_fees` (which uses `.call` not `.call!`), producing a regenerated invoice with 0 fees and 0 amounts.

---
**Bug 3 — RefreshDraftService rejects non-subscription invoices**

`app/services/invoices/refresh_draft_service.rb:29` / `app/services/adjusted_fees/create_service.rb:54-55`

Even with Bugs 1 and 2 fixed, `AdjustedFees::CreateService` calls `RefreshDraftService` which returns `forbidden_failure! unless invoice.subscription?`. For advance_charges invoices this check fails, and `raise_if_error!` propagates the error as a 500.

## Description

- Fix 1 — Allow duplication of `pay_in_advance_event_transaction_id` on duplicated fees. Added `original_fee_id` to link regenerated fees to the root fee in a void/regenerate chain, so `GET /fees?event_transaction_id=xxx` returns all related fees.
- Fix 2 — Remove the `invoice_type` guard on `create_invoice_subscriptions`. The method iterates with each, so an empty collection simply skips the block. If the original voided invoice has invoice_subscriptions, they will be duplicated regardless of the invoice type.
- Fix 3 — Guard `RefreshDraftService` call with `unless preview`, skipping this part for RegenerateFromVoidedService.


app/services/invoices/regenerate_from_voided_service.rb
```diff
@@ -21,7 +21,7 @@ def call
 
       ActiveRecord::Base.transaction do
         create_regenerated_invoice
-        create_invoice_subscriptions if regenerated_invoice.invoice_type == "subscription"
+        create_invoice_subscriptions
```
groyoh, line 24: I believe we might be missing a guard (and tests) on this service for certain invoice types like `one_off`.

app/services/invoices/regenerate_from_voided_service.rb
```diff
@@ -211,6 +211,7 @@ def duplicate_fee(voided_fee)
       dup_fee.precise_coupons_amount_cents = 0
       dup_fee.taxes_base_rate = 0
       dup_fee.taxes_rate = 0
+      dup_fee.pay_in_advance_event_transaction_id = nil
```
groyoh, line 214: I'm not sure this is the correct way to handle this to be fair. This fee would no longer be available in the `Fees::Query`.

app/services/adjusted_fees/create_service.rb
```diff
@@ -51,8 +51,10 @@ def call
       fixed_charge_id = fee.fixed_charge_id
       charge_filter_id = fee.charge_filter_id
 
-      refresh_result = Invoices::RefreshDraftService.call(invoice: invoice)
-      refresh_result.raise_if_error!
+      if invoice.subscription?
+        refresh_result = Invoices::RefreshDraftService.call(invoice: invoice)
```
groyoh, line 55: I'm wondering what is the purpose of the `RefreshDraft` here. I'm wondering whether the guard might be skipping an important step that would be required.

app/services/invoices/regenerate_from_voided_service.rb
D1353L: The idea of this fix is that if the original voided invoice has invoice_subscriptions, they are copied to the new regenerated invoice. Otherwise - no. And I see one-off invoices don't have invoice_subscriptions, so there is nothing to copy. I'll add tests for different types

app/services/invoices/regenerate_from_voided_service.rb
D1353L: Nice catch! I'm looking for other options to fix this

app/services/adjusted_fees/create_service.rb
annvelents: It is possible that we need a slight refactor of this service at all, because it's only called from `RegenerateFromVoidedService`, where it's called in the cycle of `fees_params.each`. So for each adjusted fee we'd refresh invoice
also, later we run finalization that should have refresh under the hood....
@brunomiguelpinto , please, take a look 🙏

app/services/adjusted_fees/create_service.rb
```diff
@@ -51,8 +51,10 @@ def call
       fixed_charge_id = fee.fixed_charge_id
       charge_filter_id = fee.charge_filter_id
 
-      refresh_result = Invoices::RefreshDraftService.call(invoice: invoice)
-      refresh_result.raise_if_error!
+      unless preview
```
annvelents: Let's rename it to `regenerating_voided` or something, because actually this service has never been used for `preview` 🙈

annvelents (approved):
Good for me! 😁

PR #5335, opened 2026-04-14
feat(presentation_group_keys): Add `presentation_group_keys` to Charge#properties and ChargeFilter#properties

tiagolupepic:
This commit introduces `presentation_group_keys` to Charge#properties and ChargeFilter#properties

This new option will be used in different services (aggregations) in order to compute the breakdowns. In order to do that, we're introducing the validation for each charge and adding the option to receive the `presentation_group_keys` in properties for Charge and ChargeFilter. 

We're making both models to have the same format of `presentation_group_keys`, an array of `{ value: 'name', options: { display_in_invoice: true/false}` and, we're limiting the number of presentation_group_keys to be max 2. The main reason is to avoid generating excessive data that would not be useful. If in the near future we see the option to have more keys, we can simple make the validation more flexible.


app/models/charge_filter.rb
```diff
@@ -54,6 +54,16 @@ def pricing_group_keys
...
   end
 
+  def presentation_group_keys
+    properties["presentation_group_keys"].presence
+  end
+
+  def presentation_group_keys_values
+    return [] if properties["presentation_group_keys"].blank?
+
+    properties["presentation_group_keys"].flat_map { |e| e.fetch("value", nil) }.compact
```
ancorcruz:
```suggestion
    return [] if presentation_group_keys.blank?

    presentation_group_keys.map { |e| e.fetch("value", nil) }.compact
```

we can use the already defined `presentation_group_keys` method to access the keys, also, why are you using `flat_map` instead of `map`. I would say key value is not a collection but a single value. right? map communicates better the intent here.

app/models/charge.rb
```diff
@@ -68,6 +68,16 @@ def pricing_group_keys
...
   end
 
+  def presentation_group_keys
+    properties["presentation_group_keys"].presence
+  end
+
+  def presentation_group_keys_values
+    return [] if properties["presentation_group_keys"].blank?
+
+    properties["presentation_group_keys"].flat_map { |e| e.fetch("value", nil) }.compact
```
ancorcruz:
```suggestion
    return [] if presentation_group_keys.blank?

    presentation_group_keys.map { |e| e.fetch("value", nil) }.compact
```

we can use the already defined `presentation_group_keys` method to access the keys, also, why are you using `flat_map` instead of `map`. I would say key value is not a collection but a single value. right? map communicates better the intent here.

app/services/charges/validators/base_service.rb
```diff
@@ -50,6 +51,37 @@ def validate_pricing_group_keys
...
+      def validate_presentation_group_keys
+        raw_keys = properties["presentation_group_keys"]
+        return if raw_keys.blank?
+
+        valid_presentation_group_keys = raw_keys.is_a?(Array) && raw_keys.all? do |key|
+          next false unless key.is_a?(Hash)
+
+          # NOTE: Support hashes with strings and symbols as keys to avoid issues with different formats
+          value_key_present = key.key?("value") || key.key?(:value)
+          options_key_valid = !key.key?("options") && !key.key?(:options)
```
ancorcruz: could you use hash_with_indiferent_access instead?

app/services/charges/validators/base_service.rb
```diff
@@ -50,6 +51,37 @@ def validate_pricing_group_keys
 
         add_error(field: grouped_key, error_code: "invalid_type")
       end
+
+      def validate_presentation_group_keys
```
ancorcruz, line 58: it does not validate the "value" content and will accept nil, empty string, numbers and empty array as values for value key. Is this desired or a validation is missing?

nil is silently dropped by compact in presentation_group_keys_values, but empty strings and non-string types will propagate to downstream aggregation services. Consider adding a type + presence check:

`value_key_present = value.is_a?(String) && value.present?`

pricning_group_keys are validated this way.

app/services/charges/validators/base_service.rb
```diff
@@ -50,6 +51,37 @@ def validate_pricing_group_keys
...
+
+        unless valid_presentation_group_keys
+          return add_error(
+            field: "presentation_group_keys",
+            error_code: "presentation_group_keys must be an array of hashes with a 'value' key"
+          )
+        end
+
+        if raw_keys.size > 2
+          add_error(field: "presentation_group_keys", error_code: "presentation_group_keys have a maximum of 2 elements")
```
ancorcruz: we use error codes instead of human readable messages. We can use `invalid_type` and `too_many_keys` here.

spec/support/shared_examples/charges/presentation_group_keys_validation.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples "presentation_group_keys property validation" do
+  let(:grouping_properties) { {"presentation_group_keys" => presentation_group_keys} }
+  let(:presentation_group_keys) { {} }
```
ancorcruz: why a Hash? isn't expected to be nil or an array?

ancorcruz (commented):
looks great, I have left you a few comments.

app/models/charge_filter.rb
tiagolupepic: changed to use your suggestion and using `presentation_group_keys`. I was using the flat_map because the collection was different and when I simplified, I missed to use map here instead.

app/services/charges/validators/base_service.rb
tiagolupepic: I changed to use `deep_symbolize_keys` which is used in other parts of the services as well

app/services/charges/validators/base_service.rb
tiagolupepic: good point! I reviewed the validations and I updated to avoid wrong values as you suggested.

app/services/charges/validators/base_service.rb
tiagolupepic: changed! thanks!

spec/support/shared_examples/charges/presentation_group_keys_validation.rb
tiagolupepic: it's a leftover, the initial/draft implementation was using a hash instead of an array. I removed this.

ancorcruz (approved):
LGTM

PR #5336, opened 2026-04-14
fix(charges): Skip stale cascade updates

rsempe:
When a charge's filters are updated rapidly (e.g. a customer updating 160+ filters one by one via the API), each filter update triggers a full cascade to all child charges.
These cascades are redundant because each one reads the current DB state which already includes all prior filter changes. The result is dozens of overlapping cascade jobs contending on the advisory lock, causing dead jobs and wasted work.

Add a `cascaded_at` timestamp to detect and skip stale cascades.


rsempe:
Closed in favor of https://github.com/getlago/lago-api/pull/5357.

PR #5338, opened 2026-04-14
feat(payment-gated-subs): reject subscription creation when external_id is incomplete

ancorcruz:
## Context

When a subscription is incomplete (awaiting payment confirmation), no new subscription should be created with the same external_id. This prevents upgrades, downgrades, or duplicate subscriptions while an activation is in flight.

## Description

Add an early guard in CreateService that checks for incomplete subscriptions matching the requested external_id before looking up editable subscriptions. Returns a subscription_incomplete validation error. This replaces the previous next_subscription.incomplete? guard which is now subsumed by the broader check since next subscriptions share the same external_id.

### DIFF
--- a/spec/services/subscriptions/create_service_spec.rb
+++ b/spec/services/subscriptions/create_service_spec.rb
@@ -1134,12 +1134,12 @@ RSpec.describe Subscriptions::CreateService do
               )
             end
 
-            it "returns pending_plan_change error" do
+            it "returns subscription_incomplete error" do
               result = create_service.call
 
               expect(result).not_to be_success
               expect(result.error).to be_a(BaseService::ValidationFailure)
-              expect(result.error.messages[:subscription]).to eq(["pending_plan_change"])
+              expect(result.error.messages[:subscription]).to eq(["subscription_incomplete"])
             end
           end
         end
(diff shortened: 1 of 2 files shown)


spec/services/subscriptions/create_service_spec.rb
```diff
@@ -1348,18 +1348,34 @@
...
               expect(result.error).to be_a(BaseService::ValidationFailure)
-              expect(result.error.messages[:subscription]).to eq(["pending_plan_change"])
+              expect(result.error.messages[:subscription]).to eq(["subscription_incomplete"])
             end
           end
         end
       end
     end
 
+    context "when existing subscription with same external_id is incomplete" do
```
osmarluz, line 1363: @ancorcruz we should also make it clear that's for the same customer here:

```suggestion
    context "when existing subscription with same external_id is incomplete for the same customer" do
```

and maybe add the scenario `"when existing subscription with same external_id is incomplete for a different customer"`. What do you think?

ancorcruz: kind of pointless as the subscriptions are always scoped to the customer.

osmarluz (approved):
🚀

PR #5339, opened 2026-04-15
feat(payment-gated-subs): extract ActivateService

ancorcruz:
## Context

Subscription activation logic is duplicated across `CreateService`, `ActivateAllPendingService`, and `TerminateService` with slightly different side effects. A single entry point makes activation consistent and provides a reusable service for gated activation flows.

## Description

Extract `ActivateService` that handles the common activation concerns: `mark_as_active`, emit fixed charge events, bill subscription or pay-in-advance fixed charges, send subscription.started webhook and activity log, and trigger HubSpot sync. Accepts a Time object as timestamp parameter. Returns early if subscription is already active.

Refactor `ActivateAllPendingService` to delegate to the new service, removing inline activation logic.

### DIFF
--- a/app/services/subscriptions/activate_all_pending_service.rb
+++ b/app/services/subscriptions/activate_all_pending_service.rb
@@ -21,26 +21,7 @@ module Subscriptions
           Time.zone.at(timestamp)
         )
         .find_each do |subscription|
-          subscription.mark_as_active!(Time.zone.at(timestamp))
-          fixed_charge_timestamp = subscription.started_at + 1.second
-
-          EmitFixedChargeEventsService.call!(
-            subscriptions: [subscription],
-            timestamp: fixed_charge_timestamp
-          )
-
-          if subscription.should_sync_hubspot_subscription?
-            Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob.perform_later(subscription:)
-          end
-
-          SendWebhookJob.perform_later("subscription.started", subscription)
-          Utils::ActivityLog.produce(subscription, "subscription.started")
-
-          if subscription.plan.pay_in_advance? && !subscription.in_trial_period?
-            BillSubscriptionJob.perform_later([subscription], timestamp, invoicing_reason: :subscription_starting)
-          elsif subscription.fixed_charges.pay_in_advance.any?
-            Invoices::CreatePayInAdvanceFixedChargesJob.perform_later(subscription, fixed_charge_timestamp)
-          end
+          ActivateService.call!(subscription:, timestamp: Time.zone.at(timestamp))
         end
 
       result
--- a/app/services/subscriptions/activate_all_pending_service.rb
+++ b/app/services/subscriptions/activate_all_pending_service.rb
@@ -5,7 +5,7 @@ module Subscriptions
     Result = BaseResult
 
     def initialize(timestamp:)
-      @timestamp = timestamp
+      @timestamp = Time.zone.at(timestamp)
 
       super
     end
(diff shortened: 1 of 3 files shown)


app/services/subscriptions/activate_all_pending_service.rb
```diff
@@ -21,26 +21,7 @@ def call
...
-
-          SendWebhookJob.perform_later("subscription.started", subscription)
-          Utils::ActivityLog.produce(subscription, "subscription.started")
-
-          if subscription.plan.pay_in_advance? && !subscription.in_trial_period?
-            BillSubscriptionJob.perform_later([subscription], timestamp, invoicing_reason: :subscription_starting)
-          elsif subscription.fixed_charges.pay_in_advance.any?
-            Invoices::CreatePayInAdvanceFixedChargesJob.perform_later(subscription, fixed_charge_timestamp)
-          end
+          ActivateService.call!(subscription:, timestamp: Time.zone.at(timestamp))
```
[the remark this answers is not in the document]
osmarluz: That's a valid point

app/services/subscriptions/activate_service.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    attr_reader :subscription, :timestamp
+
+    def bill_subscription
+      if subscription.plan.pay_in_advance? && !subscription.in_trial_period?
+        BillSubscriptionJob.perform_later(
+          [subscription],
+          timestamp.to_i,
+          invoicing_reason: :subscription_starting,
+          skip_charges: true
```
osmarluz: @ancorcruz just a heads up that we're using `skip_charges: true` here but the previous version of `app/services/subscriptions/activate_all_pending_service.rb` was using the default job value (`skip_charges: false`). We'll probably need to receive this as a parameter

app/services/subscriptions/activate_all_pending_service.rb
osmarluz, line 1: @ancorcruz we can also update the `spec/services/subscriptions/activate_all_pending_service_spec.rb` to just verify it delegates to `ActivateService`

spec/services/subscriptions/activate_service_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      expect(BillSubscriptionJob).not_to have_been_enqueued
+    end
+  end
+
+  context "when subscription is already active" do
+    let(:subscription) { create(:subscription, organization:, customer:, plan:) }
+
+    it "returns the subscription without changes" do
+      expect(result.subscription).to be_active
+      expect(SendWebhookJob).not_to have_been_enqueued
```
osmarluz, line 160: @ancorcruz it would be good to check that the other services/methods/jobs are not called here to be more thorough. What do you think?

app/services/subscriptions/activate_service.rb
```diff
@@ -0,0 +1,56 @@
...
+        timestamp: subscription.started_at + 1.second
+      )
+
+      bill_subscription
+
+      SendWebhookJob.perform_later("subscription.started", subscription)
+      Utils::ActivityLog.produce(subscription, "subscription.started")
+
+      if subscription.should_sync_hubspot_subscription?
+        Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob.perform_later(subscription:)
```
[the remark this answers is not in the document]
osmarluz: @ancorcruz I'm assuming we're including the flows we have on `app/services/subscriptions/create_service.rb` later on, right? So for now we won't support `after_commit`.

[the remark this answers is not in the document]
ancorcruz: yes, this is the first step, then other service will implement this service instead of custom code for activation.

app/services/subscriptions/activate_all_pending_service.rb
ancorcruz: I would leave as it is... tests are already there testing the integration, it will ensure it does not break when we introduce changes...

app/services/subscriptions/activate_service.rb
ancorcruz: good catch, I'm going to remove the param for now as this is changing the current behaviour. It will be added later if needed.

osmarluz (approved):
🚀

PR #5341, opened 2026-04-15
refactor(fixed-charges): return created events from EmitEventsForActiveSubscriptionsService

ancorcruz:
## Context

The service creates fixed charge events but does not expose them in the result, forcing callers and tests to query the database to access them.

## Description

Collect the fixed charge events created by FixedChargeEvents::CreateService and return them as result.fixed_charge_events. Update tests to assert on the result instead of querying FixedChargeEvent directly.

### DIFF
--- a/spec/services/fixed_charges/emit_events_service_spec.rb
+++ b/spec/services/fixed_charges/emit_events_service_spec.rb
@@ -57,15 +57,16 @@ RSpec.describe FixedCharges::EmitEventsService do
     it "creates fixed charge events for all active subscriptions" do
       expect { result }.to change(FixedChargeEvent, :count).by(2)
 
-      event_1 = FixedChargeEvent.find_by(subscription: active_subscription_1, fixed_charge:)
-      event_2 = FixedChargeEvent.find_by(subscription: active_subscription_2, fixed_charge:)
+      events = result.fixed_charge_events
+      expect(events.size).to eq(2)
 
-      expect(event_1).to be_present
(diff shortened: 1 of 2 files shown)


tiagolupepic (approved):
:shipit:

spec/services/fixed_charges/emit_events_service_spec.rb
```diff
@@ -109,8 +108,8 @@
       it "creates fixed charge event for the incomplete subscription" do
         expect { result }.to change(FixedChargeEvent, :count).by(1)
 
-        event = FixedChargeEvent.find_by(subscription:, fixed_charge:)
-        expect(event).to be_present
+        expect(result.fixed_charge_events.size).to eq(1)
```
osmarluz, line 111: @ancorcruz nit: we could have just the expectation above since they are basically the same

osmarluz (approved):
LGTM! Just left a small comment

spec/services/fixed_charges/emit_events_service_spec.rb
ancorcruz: not the same, the first one expects fixed charge events to have been created on DB the second one expect the results returned by the service contains the fixed charge events created.

osmarluz: @ancorcruz the point here was: if `FixedChargeEvents::CreateService.call!` fails to persist, an exception is raised and we never reach the assignment, so they are redundant in that sense

PR #5342, opened 2026-04-15
feat(order_forms): add graphql layer for quotes

rinasergeeva:
## Context

The previous PR ([#5173](https://github.com/getlago/lago-api/pull/5173)) introduced the data and service foundation for _Quotes_. This PR opens that foundation up to the dashboard by exposing the GraphQL surface that the front-end will use to let sales and finance teams browse, build and negotiate quotes with their customers.

Like the foundation, the whole surface stays gated behind the premium license and the `order_forms` organization feature flag, and is permission-scoped (`quotes:view`, `quotes:create`, `quotes:update`, `quotes:approve`, `quotes:clone`, `quotes:void`).

### Browsing quotes

Two queries power the list and detail views:

- `quotes` — paginated list of an organization's quotes, filterable by customer, quote number, status, owner, order type, and creation date range. The status filter is intentionally evaluated against the **latest version** of each quote, so the list reflects what the user actually sees on screen rather than the historical content of older versions.
- `quote` — single quote by id, returning a 404-style not-found error when the quote does not belong to the current organization.

The `Quote` type exposes everything the UI needs in one place: the human-readable `number`, the customer and (optional) subscription, the order type, the list of internal `owners`, the full `versions` history, and the `current_version` highlight. The `QuoteVersion` type carries the lifecycle (`status`, `approved_at`, `voided_at`, `void_reason`), the content payload (`billing_items`, `content`), the auto-generated `share_token` used in customer-facing links, and the version number.

### Acting on quotes

Six mutations cover the actions the team performs on a deal:

- `createQuote` / `updateQuote` — start a new quote (subscription creation, amendment or one-off) and update its owners.
- `updateQuoteVersion` — keep iterating on the working draft.
- `approveQuoteVersion` — lock the deal once the customer has accepted.
- `voidQuoteVersion` — drop a version with an explicit reason so the team always knows _why_ a quote went away.
- `cloneQuoteVersion` — supersede the current draft with a fresh editable copy, preserving history.

Each mutation maps directly to its service-layer counterpart from #5173, which already enforces the lifecycle rules (only drafts are editable / voidable / approvable, only one active version per quote, etc.).

### Performance

Quote pages typically render a quote together with its customer, subscription, owners and full version history. To keep that cheap, a small reusable `dataload_association` macro is added on `Types::BaseObject` and used on `Quote`. It batches association fetches through a `Sources::ActiveRecordAssociation` dataloader source so listing N quotes with their nested associations stays an O(1) number of queries per association instead of O(N).

### Developer experience

A `db/seeds/70_order_forms.rb` seed is added to make the feature easy to demo and exercise locally: a multi-version quote chain on the canonical Hooli/John Doe customer, and one draft quote per seeded customer.


app/contracts/queries/quote_versions_query_filters_contract.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Queries
+  class QuoteVersionsQueryFiltersContract < Dry::Validation::Contract
```
murparreira: Both this file and `QuotesQueryFiltersContract` are identical. Maybe we can extract a shared base contract and make them inherit to remove the duplication?

app/contracts/queries/quote_versions_query_filters_contract.rb
```diff
@@ -0,0 +1,22 @@
...
+    params do
+      optional(:customer).maybe do
+        array(:string, format?: Regex::UUID)
+      end
+      optional(:status).maybe do
+        array(:string, included_in?: QuoteVersion::STATUSES.values)
+      end
+      optional(:number).maybe do
+        array(:string, format?: /\AQT-\d{4}-\d{4,}\z/i)
```
murparreira: If the above is not possible, we can use a constant on quote for the regex, as this is duplicated between files, and if we need to change this, we have 2 places to change. `Quote::NUMBER_FORMAT` or something like that.

app/graphql/types/quotes/object.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+      field :current_version, Types::QuoteVersions::Object, null: false
+      field :customer, Types::Customers::Object, null: false
+      field :id, ID, null: false
+      field :number, String, null: false
+      field :order_type, Types::Quotes::OrderTypeEnum, null: false
+      field :organization, Types::Organizations::OrganizationType, null: false
+      field :owners, [Types::UserType], null: true
+      field :subscription, Types::Subscriptions::Object, null: true
```
murparreira, line 16: Should this have a dataloader too?

app/graphql/types/quote_versions/void_reason_enum.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  module QuoteVersions
+    class VoidReasonEnum < Types::BaseEnum
+      QuoteVersion::VOID_REASONS.keys.each do |status|
```
murparreira: Minor nitpick: Should this be reason instead of status?

app/graphql/resolvers/quote_versions_resolver.rb
```diff
@@ -0,0 +1,43 @@
...
+module Resolvers
+  class QuoteVersionsResolver < Resolvers::BaseResolver
+    include AuthenticableApiUser
+    include RequiredOrganization
+
+    REQUIRED_PERMISSION = "quotes:view"
+
+    description "Query quote versions of an organization"
+
+    argument :customer, [ID], required: false
```
murparreira: Should this be `customers`? As we accept an array.

app/graphql/resolvers/quote_versions_resolver.rb
```diff
@@ -0,0 +1,43 @@
...
+    argument :from_date, GraphQL::Types::ISO8601Date, required: false
+    argument :limit, Integer, required: false
+    argument :number, [String], required: false
```
murparreira: Same here, should this be `numbers`? As we accept an array.

app/queries/quotes_query.rb
```diff
@@ -0,0 +1,56 @@
...
+  def with_status(scope)
+    scope.joins(:current_version).where(quote_versions: {status: filters.status})
+  end
+
+  def with_date(scope)
+    scope.where(created_at: filters.from_date..filters.to_date)
+  end
+
+  def with_owners(scope)
+    scope.joins(:quote_owners).where(quote_owners: {user_id: filters.owners})
```
murparreira: Should we add a `.distinct` here? This could produce duplicate rows (one per matched owner).

app/queries/quote_versions_query.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    quote_versions = base_scope
+    quote_versions = with_customer(quote_versions) if filters.customer.present?
+    quote_versions = with_number(quote_versions) if filters.number.present?
+    quote_versions = with_status(quote_versions) if filters.status.present?
+    quote_versions = with_date(quote_versions) if filters.from_date.present? || filters.to_date.present?
+    quote_versions = with_owners(quote_versions) if filters.owners.present?
+
+    # final ordering and pagination
+    quote_versions = quote_versions.order(created_at: :desc)
```
murparreira: Minor niptick: Ordering quote versions by `created_at` instead of the `version` integer field is semantically odd. The `version` field exists specifically to represent ordering within a quote's lifecycle.

app/contracts/queries/quote_versions_query_filters_contract.rb
rinasergeeva: I don't think we win much by doing this. Moreover, it's a draft, so we'll soon adjust the filters to the frontend's needs, and these contracts will very likely be different.

app/contracts/queries/quote_versions_query_filters_contract.rb
rinasergeeva: Good point! Done.

app/graphql/types/quotes/object.rb
rinasergeeva: True! Forgot to add it.
Fixed.

app/graphql/types/quote_versions/void_reason_enum.rb
rinasergeeva: Yep, makes sense. Done.

app/graphql/resolvers/quote_versions_resolver.rb
rinasergeeva: Yep. Done.

app/graphql/resolvers/quote_versions_resolver.rb
rinasergeeva: Done.

app/queries/quotes_query.rb
rinasergeeva: Ouch. 
Thanks! Fixed.

app/queries/quote_versions_query.rb
rinasergeeva: Is there any benefit to ordering this list by version? Since the list contains _versions_ of multiple quotes, the result might be quite strange, tbh:
- Version 12 (of Quote 89), created 5 years ago
- Version 12 (of Quote 9325), created 6 months ago
- Version 11 (of Quote 11893), created a week ago
- Version 11 (of Quote 89), created 5 years ago
- Version 11 (of Quote 9325), created 6 months ago
- Version 10 (of Quote 11893), created 8 days ago
- etc

Having the most recent data on the top makes much more sense to me.

app/contracts/queries/quote_versions_query_filters_contract.rb
murparreira: Got it!

app/queries/quote_versions_query.rb
murparreira: Yeah, not a problem at all!

toommz (approved):

app/graphql/resolvers/quotes_resolver.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    REQUIRED_PERMISSION = "quotes:view"
+
+    description "Query quotes of an organization"
+
+    argument :customers, [ID], required: false
+    argument :from_date, GraphQL::Types::ISO8601Date, required: false
+    argument :limit, Integer, required: false
+    argument :numbers, [String], required: false
+    argument :order_type, [Types::Quotes::OrderTypeEnum], required: false
```
groyoh: The argument is plural in shape (`[OrderTypeEnum]`) but named in the singular, while every other multi-value filter on this resolver is plural: `customers`, `numbers`, `statuses`, `owners`. Renaming this to `order_types` (and the matching `order_type` key in the filters hash + `Filters` struct in `QuotesQuery` + the contract) keeps the public GraphQL surface consistent and avoids confusion about whether it accepts a single value.

app/contracts/queries/quotes_query_filters_contract.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+      optional(:numbers).maybe do
+        array(:string, format?: Quote::QUOTE_NUMBER_REGEX)
+      end
+      optional(:from_date).maybe(:date)
+      optional(:to_date).maybe(:date)
+      optional(:owners).maybe do
+        array(:string, format?: Regex::UUID)
+      end
+      optional(:order_type).maybe do
```
groyoh: Same naming concern as on the resolver: this key takes an array but is singular, while the surrounding keys (`customers`, `statuses`, `numbers`, `owners`) are all plural. Renaming to `order_types` here keeps the contract consistent with the rest of the filters.

spec/contracts/queries/quotes_query_filters_contract_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+      context "when filter is an array with invalid values" do
+        let(:filters) { {owners: ["wrong"]} }
+
+        it "is invalid" do
+          expect(result.success?).to be(false)
+          expect(result.errors.to_h).to include({owners: {0 => ["is in invalid format"]}})
+        end
+      end
+    end
+  end
```
groyoh, line 154: Every other filter declared in the contract is exercised here, but `order_type` (lines 20–22 of the contract) has no coverage. A `context "when filtering by order_type"` block mirroring the others — valid (`["one_off"]`), wrong type (string), unknown value (`["wrong"]`) — would lock the validation behavior in.

spec/graphql/mutations/quote_versions/approve_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  before do
+    membership.organization.enable_feature_flag!(:order_forms)
+    quote_version
+  end
+
+  it_behaves_like "requires current user"
+  it_behaves_like "requires current organization"
+  it_behaves_like "requires permission", "quotes:approve"
+
+  context "with valid input", :premium do
```
groyoh, line 39: The mutation specs (here and in `clone_spec.rb` / `update_spec.rb` / `void_spec.rb` / `quotes/create_spec.rb` / `quotes/update_spec.rb`) only cover the happy path. The underlying services have well-defined failure branches — `not_found_failure!` when the id resolves to nil, `forbidden_failure!` when the `order_forms` flag is off or the license is not premium, and `not_allowed_failure!(code: "inappropriate_state")` when approving/updating/voiding a non-draft version — and a regression in any of those paths would currently slip through. Adding one example per branch (asserting the `errors` payload returned by the mutation) would close that gap.

app/queries/quotes_query.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  def with_status(scope)
+    # check status of the current (latest) version
+    quote_ids = QuoteVersion
+      .where(
+        organization:,
+        status: filters.statuses
+      )
+      .where("sequential_id = (SELECT MAX(sequential_id) FROM quote_versions qv WHERE qv.quote_id = quote_versions.quote_id)")
```
groyoh, line 53: This raw `WHERE sequential_id = (SELECT MAX(sequential_id) ... WHERE qv.quote_id = quote_versions.quote_id)` is a correlated subquery — Postgres re-runs it for every candidate row, which scales poorly as the `quote_versions` table grows. A pattern that might work better is to rely on `DISTINCT ON`:

```sql
SELECT quote_id
FROM (
  SELECT DISTINCT ON (quote_id) quote_id, status FROM quote_versions WHERE organization_id = ?
)
WHERE status IN (?)
```

within a subquery/CTE

spec/graphql/resolvers/quote_resolver_spec.rb
groyoh, line 1: We're lacking test case for not found error.

spec/graphql/resolvers/quotes_resolver_spec.rb
groyoh, line 1: We're lacking pagination tests.

spec/graphql/resolvers/quote_resolver_spec.rb
rinasergeeva: Looks like it has been added

spec/graphql/resolvers/quotes_resolver_spec.rb
rinasergeeva: Looks like it has been added

PR #5343, opened 2026-04-15
feat(db): Avoid RDS Proxy connection pinning

aquinofb:
## Summary

Adds an opt-in patch to ActiveRecord's PostgreSQL adapter so that new connections no longer emit session-level `SET` commands, which is a prerequisite for AWS RDS Proxy to multiplex connections rather than pin them 1:1.

## Context

Lago is rolling out AWS RDS Proxy in front of PostgreSQL on the EU staging environment, with the aim of enabling connection multiplexing and seamless blue/green upgrade cutovers.

Out of the box though, the proxy wasn't actually multiplexing anything. Any time a client sends a session-level `SET` command, RDS Proxy pins the client to a specific database connection for the rest of that session because it can't guarantee session state is safe to share. Rails' PostgreSQL adapter sends five `SET` commands on every new connection via `configure_connection`, plus a sixth `SET SESSION timezone` that's emitted indirectly through `add_pg_decoders → update_typemap_for_default_timezone → reconfigure_connection_timezone`. With those pins in place, the proxy degraded to a pure passthrough: just an extra network hop, no connection count reduction.

AWS does not offer a session-pinning filter for PostgreSQL. The one filter that exists (`EXCLUDE_VARIABLE_SETS`) is MySQL-only, [confirmed in the AWS docs](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy-pinning.html). For PostgreSQL, the only path to unpinned connections is to stop sending session `SET`s from the client and configure the equivalent state via the proxy's `initQuery` parameter instead.

## What changed

Three new files:

- `lib/lago/rds_proxy/connection_patch.rb`: the `Lago::RdsProxy::ConnectionPatch` module, defining a replacement `configure_connection` (keeps `check_version`, the notice receiver, user-defined `variables:`, and the type map setup; drops the SETs) plus a no-op override for `reconfigure_connection_timezone` so `add_pg_decoders` doesn't re-introduce `SET SESSION timezone`.
- `config/initializers/rds_proxy.rb`: gated on `DATABASE_VIA_RDS_PROXY`; when set, it `require`s the patch module and prepends it onto `ActiveRecord::ConnectionAdapters::PostgreSQLAdapter`. Also resets `Encoding.default_internal = nil` to stop the `pg` gem from calling `set_client_encoding` independently of `configure_connection`.
- `spec/lib/lago/rds_proxy/connection_patch_spec.rb`: nine specs covering the `configure_connection` override (check_version called, `@schema_search_path` cached, no SETs emitted, user-defined variables still applied, nil values skipped, `schema_order` fallback) and the `reconfigure_connection_timezone` no-op.

Without `DATABASE_VIA_RDS_PROXY` set, nothing loads and the adapter behaves exactly as before. Local development, CI, and non-proxy deployments are unaffected.

## Infra-side prerequisite

The proxy's `initQuery` must set the six equivalent values on the database side, otherwise sessions will start with PostgreSQL defaults rather than what Rails expects:

```sql
SET client_encoding = 'UTF8';
SET client_min_messages TO 'warning';
SET search_path TO 'public';
SET standard_conforming_strings = on;
SET intervalstyle = iso_8601;
SET timezone TO 'UTC'
```

## Local verification

Tested locally with PostgreSQL query logging enabled (`ALTER SYSTEM SET log_statement = 'all'`).

Baseline run (env var unset), one new connection:
```
SET client_min_messages TO 'warning'
SET standard_conforming_strings = on
SET intervalstyle = iso_8601
SET SESSION timezone TO 'UTC'
SHOW search_path
SELECT 1 AS baseline_test
```
Five pre-query statements (4 SETs + 1 SHOW). Every one of the SETs would pin the connection on RDS Proxy.

Patched run (`DATABASE_VIA_RDS_PROXY=true`), one new connection:
```
SELECT 99 AS fresh_verify_test
```
Zero pre-query statements. Just the application query.

## How to try locally

You won't have an RDS Proxy locally, but you can reproduce the before/after connection behavior against plain PostgreSQL to confirm the patch is working as designed:

1. Enable query logging on the dev database:
   ```bash
   docker exec lago_db_dev psql -U lago -d lago -c "ALTER SYSTEM SET log_statement = 'all';"
   docker exec lago_db_dev psql -U lago -d lago -c "SELECT pg_reload_conf();"
   ```
2. Run a baseline query without the patch:
   ```bash
   lago exec api bin/rails runner 'ActiveRecord::Base.connection.execute("SELECT 1 AS baseline")'
   ```
3. Run the same query with the patch active:
   ```bash
   lago exec api env DATABASE_VIA_RDS_PROXY=true bin/rails runner 'ActiveRecord::Base.connection.execute("SELECT 2 AS patched")'
   ```
4. Compare what was sent to PostgreSQL:
   ```bash
   docker logs --since 30s lago_db_dev 2>&1 | grep -E "(SET|SHOW|baseline|patched)"
   ```
5. Restore the logging setting afterwards:
   ```bash
   docker exec lago_db_dev psql -U lago -d lago -c "ALTER SYSTEM RESET log_statement;"
   docker exec lago_db_dev psql -U lago -d lago -c "SELECT pg_reload_conf();"
   ```

You should see the baseline process emit 4 SETs + 1 SHOW before the query, and the patched process emit only the query itself.

## Test plan

- [x] `spec/lib/lago/rds_proxy/connection_patch_spec.rb` passes (9 examples, 0 failures)
- [x] Locally verified zero session-level SETs on new connections when `DATABASE_VIA_RDS_PROXY=true`
- [x] Locally verified default behavior unchanged when the env var is unset
- [ ] Stage on `eu-west-1` alongside the proxy's updated `initQuery` and confirm `DatabaseConnectionsCurrentlySessionPinned` drops to ~0 in CloudWatch
- [ ] Stage with `DATABASE_PREPARED_STATEMENTS=false` to also eliminate extended-query-protocol pinning
- [ ] Measure p99 query latency on billing/event-ingestion paths before/after to confirm no regression from the added proxy hop

### DIFF
--- a/config/initializers/rds_proxy.rb
+++ b/config/initializers/rds_proxy.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+# Activates the RDS Proxy connection patch. See lib/lago/rds_proxy/connection_patch.rb
+# for what the patch does and why.
+
+if ENV["DATABASE_VIA_RDS_PROXY"].present?
+  require "lago/rds_proxy/connection_patch"
+
+  # The pg gem auto-sends SET client_encoding when default_internal is set,
+  # bypassing configure_connection. Reset to nil to suppress.
+  Encoding.default_internal = nil
+
+  ActiveSupport.on_load(:active_record) do
+    ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.prepend(
+      Lago::RdsProxy::ConnectionPatch
+    )
+  end
+end
--- a/lib/lago/rds_proxy/connection_patch.rb
+++ b/lib/lago/rds_proxy/connection_patch.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+# Patches PostgreSQLAdapter#configure_connection to skip session-level SET
+# commands that cause RDS Proxy to pin the connection. The skipped settings
+# must be configured via the proxy's initQuery instead:
+#
+#   SET client_encoding = 'UTF8';
+#   SET client_min_messages TO 'warning';
+#   SET search_path TO 'public';
+#   SET standard_conforming_strings = on;
+#   SET intervalstyle = iso_8601;
+#   SET timezone TO 'UTC'
+#
+# `@schema_search_path` is cached locally to avoid an extra `SHOW search_path`
+# query on first access. This assumes the proxy's initQuery sets a `search_path`
+# matching `@config[:schema_search_path]` (both default to 'public').
+#
+# `reconfigure_connection_timezone` is overridden to a no-op because it is
+# invoked by `add_pg_decoders` during type map setup and would otherwise emit
+# `SET SESSION timezone TO 'UTC'`.
+#
+# See: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy-pinning.html
+module Lago
+  module RdsProxy
+    module ConnectionPatch
+      def configure_connection
+        check_version
+
+        @schema_search_path = @config[:schema_search_path] || @config[:schema_order]
+
+        unless ActiveRecord.db_warnings_action.nil?
+          @raw_connection.set_notice_receiver do |result|
+            message = result.error_field(PG::Result::PG_DIAG_MESSAGE_PRIMARY)
+            code = result.error_field(PG::Result::PG_DIAG_SQLSTATE)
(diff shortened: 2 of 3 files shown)


groyoh:
@aquinofb let's close this for now.

PR #5345, opened 2026-04-15
perf(invoices): add index on payment_due_date

brunomiguelpinto:
## Context 

Clock::MarkInvoicesAsPaymentOverdueJob scans invoices by payment_due_date to flag overdue ones, but the column has no supporting index. In production the first find_in_batches page falls back to a PK scan with inline filtering and a LIMIT 1 probe already takes ~25 seconds, making the clock job unusable

## Description
Add a partial index on invoices.payment_due_date restricted to the predicate used by the overdue job: finalized invoices that are not yet paid, not yet flagged overdue, and  not disputed. The partial clause keeps the index minimal and self-maintaining, since rows leave it once they are paid, voided, marked overdue, or lose a dispute.

PR #5348, opened 2026-04-15
perf(invoices): align overdue job with new index

brunomiguelpinto:
## Context

Follow-up to #5345 (closes [ISSUE-1812](https://linear.app/getlago/issue/ISSUE-1812/add-index-for-invoice-payment-due-date)).

The partial index on `invoices.payment_due_date` shipped in #5345 was
not being picked by the planner for `Clock::MarkInvoicesAsPaymentOverdueJob`.
`find_in_batches` forces `ORDER BY id LIMIT`, and combined with stale
multi-column selectivity estimates the planner preferred a PK scan
over the new index.

## Description

Replace `find_in_batches` with `in_batches(of: 1000, cursor: [:payment_due_date, :id])`.
The generated query orders by `(payment_due_date, id)`, whose prefix
matches the partial index, so the planner picks it consistently.

The index-usage spec is updated to assert the plan for the new query
order.

### DIFF
--- a/app/jobs/clock/mark_invoices_as_payment_overdue_job.rb
+++ b/app/jobs/clock/mark_invoices_as_payment_overdue_job.rb
@@ -11,8 +11,8 @@ module Clock
         .where(payment_overdue: false)
         .where(payment_dispute_lost_at: nil)
         .where(payment_due_date: ...Time.current)
-        .find_in_batches do |invoices|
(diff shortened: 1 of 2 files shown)


annvelents (approved):
My personal preference: if .order_by allows us to also rely on this index, having it without cursor would be better, otherwise, this is good :)

PR #5349, opened 2026-04-16
misc(clickhouse): Add re-enrich subscription orchestrator

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/5311

## Description

This PR adds the subscription state machine to monitor the full migration process at subscription level

### DIFF
--- a/app/services/events/stores/clickhouse/enriched_store_migration/subscription_orchestrator_service.rb
+++ b/app/services/events/stores/clickhouse/enriched_store_migration/subscription_orchestrator_service.rb
@@ -1,0 +1,161 @@
+# frozen_string_literal: true
+
+module Events
+  module Stores
+    module Clickhouse
+      module EnrichedStoreMigration
+        # Drives a single subscription through the enriched store migration pipeline.
+        # Called repeatedly — it reads the current state and executes the next step:
+        #
+        #   pending     → run initial comparison
+        #   comparing   → if no diffs: completed (fast path)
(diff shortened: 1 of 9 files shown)


app/services/events/stores/clickhouse/enriched_store_migration/subscription_orchestrator_service.rb
```diff
@@ -0,0 +1,156 @@
...
+            when "pending"
+              handle_pending
+            when "comparing"
+              handle_comparing
+            when "deduplicating"
+              handle_deduplicating
+            when "validating"
+              handle_validating
+            else
+              Rails.logger.error("Unknown status: #{subscription_migration.status}")
```
toommz: Should we return an error in this case?

app/services/events/stores/clickhouse/enriched_store_migration/subscription_orchestrator_service.rb
```diff
@@ -0,0 +1,156 @@
...
+            elsif subscription_migration.billable_metric_codes.present?
+              subscription_migration.start_reprocessing!
+              reprocess_events
+            else
+              fail_migration!("Diffs found but no billable metric codes to reprocess")
+            end
+          end
+
+          def handle_deduplicating
+            dedup_result = ::Events::Stores::Clickhouse::CleanDuplicatedEnrichedExpandedService.call(
```
toommz, line 76: Could this service fail? If so, should we handle those failures?

toommz (approved):

PR #5353, opened 2026-04-16
fix(superset): handle service failure in dashboards resolver

toommz:
## Context

A customer hits an error page when accessing the Superset dashboard.
The GraphQL query returns "Cannot return null for non-nullable field
Query.supersetDashboards".

## Description

The resolver was missing the standard `result.success?` check after
calling `Auth::SupersetService`. When the service fails (missing config,
auth error, timeout, etc.), `result.dashboards` is never set, so `nil`
is returned on a non-nullable field, crashing the query.

Add the error guard and a test for the failure path.


murparreira (approved):
LGTM!

PR #5354, opened 2026-04-16
feat(payment-gated-subs): add payment resolution flow for gated subscriptions

ancorcruz:
## Context

When a payment succeeds or fails for a gated subscription, the system needs to resolve the activation rule, transition the invoice and subscription to their final states, and fire the appropriate side effects. This is the core payment outcome handling for Milestone 1.

## Description

Add Payment::ResolveService that handles payment outcomes for incomplete subscriptions within a row-level lock for idempotency.

* On success: marks the payment rule as satisfied via Payment::EvaluateService, finalizes the invoice, and delegates to ResolveSubscriptionStatusService for subscription activation.

* On failure: marks the rule as failed, closes the invoice, delegates to ResolveSubscriptionStatusService for subscription cancellation, and explicitly saves the cancelation_reason.

Extract ResolveSubscriptionStatusService from EvaluateService to separate rule evaluation (inactive → pending/not_applicable) from subscription status resolution (activate when all rules satisfied, cancel when any rule failed/expired/declined). EvaluateService now calls both in sequence. ResolveService calls only the status resolution after transitioning the rule, avoiding redundant re-evaluation.

Add polymorphic evaluate! method to ActivationRule for type-specific dispatch without case statements. Add guard in Payment::EvaluateService#transition_pending_rule against nil status.

Add Payment::ResolveJob as an async wrapper, wired into Invoices::UpdateService via perform_after_commit following the same pattern as PrepaidCreditJob.

### DIFF
--- a/app/services/subscriptions/activation_rules/payment/resolve_service.rb
+++ b/app/services/subscriptions/activation_rules/payment/resolve_service.rb
@@ -44,7 +44,7 @@ module Subscriptions
           EvaluateService.call!(rule: payment_rule, status: :failed)
           invoice.closed!
           ActivationRules::ResolveSubscriptionStatusService.call!(subscription:)
-          subscription.update!(cancelation_reason: :payment_failed) if subscription.canceled?
+          subscription.update!(cancelation_reason: :payment_failed)
         end
 
         def payment_rule
--- a/app/services/subscriptions/activation_rules/payment/resolve_service.rb
+++ b/app/services/subscriptions/activation_rules/payment/resolve_service.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  module ActivationRules
+    module Payment
+      class ResolveService < BaseService
+        Result = BaseResult
+
+        def initialize(subscription:, invoice:, payment_status:)
+          @subscription = subscription
+          @invoice = invoice
+          @payment_status = payment_status.to_sym
+          super
+        end
+
+        def call
+          subscription.with_lock do
+            case payment_status
+            when :succeeded
+              handle_success
+            when :failed
+              handle_failure
+            end
+          end
+
+          result
+        end
+
+        private
+
+        attr_reader :subscription, :invoice, :payment_status
+
+        def handle_success
+          # TODO: implement in next TDD cycle
+        end
+
+        def handle_failure
+          return unless subscription.incomplete? && invoice.open? && invoice.subscription?
+
+          payment_rule.failed!
+          invoice.closed!
+          subscription.cancelation_reason = :payment_failed
+          subscription.mark_as_canceled!
+
+          SendWebhookJob.perform_later("subscription.canceled", subscription)
+        end
+
+        def payment_rule
+          @payment_rule ||= subscription.activation_rules.find_by(type: "payment")
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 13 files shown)


app/services/subscriptions/activation_rules/resolve_subscription_status_service.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      attr_reader :subscription
+
+      def all_rules_satisfied?
+        subscription.activation_rules.all? { |rule| rule.satisfied? || rule.not_applicable? }
+      end
+
+      def any_rule_failed?
+        subscription.activation_rules.any? { |rule| rule.failed? || rule.expired? || rule.declined? }
+      end
```
osmarluz, line 37: @ancorcruz not a big deal because the `activation_rules` probably won't be large, but we could convert to just queries here:


```suggestion
      def all_rules_satisfied?
        statuses = Subscription::ActivationRule::STATUSES.values_at(:satisfied, :not_applicable)
        subscription.activation_rules.where.not(status: statuses).none?
      end

      def any_rule_failed?
        statuses = Subscription::ActivationRule::STATUSES.values_at(:failed, :expired, :declined)
        subscription.activation_rules.where(status: statuses).exists?
      end
```

app/services/subscriptions/activation_rules/payment/resolve_service.rb
```diff
@@ -0,0 +1,56 @@
...
+          ActivationRules::ResolveSubscriptionStatusService.call!(subscription:)
+        end
+
+        def handle_failure
+          return unless subscription.incomplete? && invoice.open? && invoice.subscription?
+
+          EvaluateService.call!(rule: payment_rule, status: :failed)
+          invoice.closed!
+          ActivationRules::ResolveSubscriptionStatusService.call!(subscription:)
+          subscription.update!(cancelation_reason: :payment_failed) if subscription.canceled?
```
osmarluz: @ancorcruz here we could remove the check `subscription.canceled?`, because the subscription will for sure be `canceled` after we run `EvaluateService.call!(rule: payment_rule, status: :failed)` and `ActivationRules::ResolveSubscriptionStatusService.call!(subscription:)`


```suggestion
          subscription.update!(cancelation_reason: :payment_failed)
```

app/services/subscriptions/activation_rules/resolve_subscription_status_service.rb
ancorcruz: ok, thanks, makes sense.

app/services/subscriptions/activation_rules/payment/resolve_service.rb
ancorcruz: you are right, this is just noise.

spec/services/invoices/update_service_spec.rb
```diff
@@ -244,6 +244,38 @@
...
     end
 
+    context "when invoice is subscription_gated and payment_status changes" do
+      let(:plan) { create(:plan, organization: invoice.organization, pay_in_advance: true) }
+      let(:subscription) { create(:subscription, :incomplete, organization: invoice.organization, customer: invoice.customer, plan:) }
+      let(:rule) { create(:subscription_activation_rule, subscription:, status: "pending") }
+      let(:invoice) { create(:invoice, status: :open, invoice_type: :subscription, payment_overdue: false) }
+
+      before do
+        rule
```
osmarluz: @ancorcruz just a small suggestion here:


```suggestion
      let(:subscription) do
        create(:subscription, :incomplete, :with_activation_rules,
          organization: invoice.organization, customer: invoice.customer, plan:,
          activation_rules_config: [{type: "payment", timeout_hours: 48, status: "pending"}])
      end
      let(:invoice) { create(:invoice, status: :open, invoice_type: :subscription, payment_overdue: false) }

      before do
```

osmarluz (approved):
LGTM! Just left a couple of small comments 🚀

PR #5332, opened 2026-04-14
feat(presentation_group_keys): Add presentation_breakdowns table and its relations and dependencies

tiagolupepic:
This commit introduces a new `PresentationBreakdown` model to the application, along with its database table, associations, and tests. The changes allow each Fee to have a set of presentation breakdowns, explaining the units based on each presentation key defined and the aggregation choosen.

example defining two presentation keys, `"region"` and `"city"`.

```
Storage
  Instance A                  28 units    $28.00
    └ EU + Barcelona          10
    └ EU + Paris              3         
    └ US + Miami              15          
  Instance B                   7 units     $7.00
    └ EU + Barcelona           4         
    └ US + New York            3  
``` 

These changes are the initial step for generating and tracking the fees units breakdowns

PR #5355, opened 2026-04-16
feat(presentation_group_keys): Add breakdowns to sum aggregation and persist them on fee

tiagolupepic:
This commit introduces the breakdowns query in sum service for Postgres store.

Continuing the development of the new functionality of presentation group keys, we're introducing in 
this commit with some changes to make such option available in Fee#presentation_breakdowns relation.

First, we're changing the aggregation base service to return the `breakdowns`, each service that inherits
the BaseService will be able to compute and return the breakdowns. In the same service, we're changing the 
initializer to receive `presentation_by` to be used later to perform queries for breakdowns.

For the store perspective, we're changing the existing method of grouped_sum to receive the columns we're interested to use in the group in the sql. By default, the columns is using the grouped_by array. For the breakdown, we're sending the `uniq_grouped_by_and_presentation_by` to compute the breakdowns.

After the breakdowns are queried in the database, we can associate each corresponding breakdown with the fee,
this responsibility is part of `Fees::ChargeService`. Another change we're doing in this service is to identify the filters provided by the union of `filter_by_presentation` and `charge#presentation_group_keys_values` to the aggregator service. 

One detail regarding the `grouped_by` and `presentation_by` keys, we're making the difference to avoiding sending a duplicated key to the query, this responsibility is part of store base service. This is done in a method called `uniq_grouped_by_and_presentation_by`.

Another service involved in this commit is the `ChargeCacheMiddleware`, the service serialize and deserialize the fee attributes and its relations, to keep compatible with presentation breakdowns, we're including the presentation_breakdowns relation in this service.

Finally, we're disabling the breakdowns computation from the Refresh Wallet services, since wallet only depends on pricing and the breakdowns are irrelevant, we're saving some resources there sending an empty array for the`filter_by_presentation` 

We're going to add the same changes to the other aggregation services, such as Count, Latest and etc in another PR.

This commit is part of these other commits and belongs to presentation group keys functionality:
- https://github.com/getlago/lago-api/pull/5332
- https://github.com/getlago/lago-api/pull/5335
- https://github.com/getlago/lago-api/pull/5351


app/services/fees/charge_service.rb
```diff
@@ -139,15 +155,26 @@ def hydrate_non_persistable_fees(properties:, charge_filter:)
...
         # TODO: check if this is still needed as we now skip certain zero units fees
         next if current_usage && charge_filter && amount_result.units.zero? && !with_zero_units_filters
 
-        init_fee(amount_result, properties:, charge_filter:)
+        fee = init_fee(amount_result, properties:, charge_filter:)
+        next if fee.nil?
+
+        if breakdowns_indexed_by_group.present?
+          breakdowns_indexed_by_group.fetch(fee.grouped_by, EMPTY_BREAKDOWN)[:breakdowns].each do |breakdown|
+            fee.presentation_breakdowns.build(breakdown.merge(organization_id: charge.organization_id))
```
lovrocolic: Based on the code it seems that `fee` could have many `presentation_breakdowns`?
However this logic does not work with the migration we have today:
`index_presentation_breakdowns_on_fee_id (fee_id) UNIQUE`

lovrocolic:
Can we please add two more tests:
- persisted fee with multiple breakdowns                                                                                                                   
- combined pricing + presentation

app/services/fees/charge_service.rb
tiagolupepic: yes, you're right. Added specs to cover this, thanks!

app/services/fees/charge_service.rb
```diff
@@ -139,15 +154,33 @@ def hydrate_non_persistable_fees(properties:, charge_filter:)
...
+    def fees_from_charge_model_result(charge_model_result, properties:, charge_filter:, breakdowns:)
       charge_model_result.grouped_results.map do |amount_result|
         # TODO: check if this is still needed as we now skip certain zero units fees
         next if current_usage && charge_filter && amount_result.units.zero? && !with_zero_units_filters
 
-        init_fee(amount_result, properties:, charge_filter:)
+        fee = init_fee(amount_result, properties:, charge_filter:)
+        next if fee.nil?
+
+        Array(breakdowns).each do |breakdown|
```
annvelents: please, move it into `build_breakdowns_for_fee`

tiagolupepic: done!

annvelents (approved):
LGTM 🚀 , I'd suggest to get also a review from @vincent-pochet or @groyoh as it's touching usage and events aggregation and it's a PR putting the foundation for breakdowns aggregations

app/services/events/stores/base_store.rb
```diff
@@ -12,6 +12,7 @@ def initialize(subscription:, boundaries:, code: nil, filters: {}, deduplicate:
 
         @grouped_by = filters[:grouped_by]
         @grouped_by_values = filters[:grouped_by_values]
+        @presentation_by = filters[:presentation_by]
```
vincent-pochet, line 15: It seems that we are not relying on this value, right?

app/services/fees/charge_service.rb
```diff
@@ -411,6 +441,11 @@ def aggregation_filters(charge_filter: nil)
       grouped_by_keys = grouped_by_keys(charge_filter:)
       filters[:grouped_by] = grouped_by_keys if grouped_by_keys.present?
 
+      presentation_group_keys_values = charge.presentation_group_keys_values
```
vincent-pochet, line 444: Is the `presentation_group_keys` only at charge level or are we planing to add it also at the filter one?

app/services/subscriptions/charge_cache_middleware.rb
```diff
@@ -23,10 +30,18 @@ def call(charge_filter:)
           PricingUnitUsage.new(j["pricing_unit_usage"].slice(*PricingUnitUsage.column_names))
         end
 
-        Fee.new(
+        fee = Fee.new(
           **j.slice(*Fee.column_names),
           pricing_unit_usage:
         )
+
+        j.fetch("presentation_breakdowns", EMPTY_ARRAY).each do |breakdown|
```
vincent-pochet, line 38: Not sure of the need for a constant here. The optimization will be very minimal (compared to the number of instances initialized in this path)

vincent-pochet (commented):
It does look good 👍 
Only few small comments

My only "fear" is that its adding more heavy requests in the critical path, but I don't think we can avoid it. With good caching and future optimization on the other events pipeline it should not be an issue

vincent-pochet (approved):

app/services/fees/charge_service.rb
tiagolupepic: yes! I'll add it in another PR, I didn't want to include several changes in this PR and make it bigger

app/services/events/stores/base_store.rb
tiagolupepic: it's true, it's a leftover, I'll remove it in another PR that is touching the stores and aggregators

app/services/subscriptions/charge_cache_middleware.rb
tiagolupepic: there are some scenarios for fees that we might have ~100 presentation breakdowns. It can be a bit helpful

PR #5357, opened 2026-04-17
refactor(charge-filters): Cascade at filter level instead of charge level

rsempe:
## Context

When a customer rapidly updates multiple charge filters via the API (e.g. 160+ PUT requests), each filter update triggered a full charge-level cascade that re-applied the entire charge state to every child. This caused ordering issues, advisory lock contention, and stale `old_parent_filters_attrs` snapshots that broke customization detection.

## Description

Introduce a filter-level cascade for individual filter endpoints (`create`, `update`, `destroy`). Each filter update now enqueues a lightweight `ChargeFilters::CascadeJob` that cascades only that specific filter to child charges — no ordering dependency, no lock contention, no stale snapshots.

The cascade matches child filters by their values hash (same mechanism as `CreateOrUpdateBatchService`) and preserves customer overrides by comparing the child's properties against the old parent properties for that specific filter.

### DIFF
--- a/app/services/charge_filters/cascade_service.rb
+++ b/app/services/charge_filters/cascade_service.rb
@@ -59,24 +59,26 @@ module ChargeFilters
     def create_child_filter(child_charge)
       return if find_child_filter(child_charge)
 
-      child_filter = child_charge.filters.new(
-        organization_id: child_charge.organization_id,
-        invoice_display_name:,
-        properties: ChargeModels::FilterPropertiesService.call(
-          chargeable: child_charge,
-          properties: new_properties
-        ).properties
-      )
-      child_filter.save!
-
-      filter_values.each do |key, values|
-        billable_metric_filter = child_charge.billable_metric.filters.find_by(key:)
-
-        child_filter.values.create!(
-          billable_metric_filter_id: billable_metric_filter&.id,
+      ActiveRecord::Base.transaction do
+        child_filter = child_charge.filters.new(
           organization_id: child_charge.organization_id,
-          values:
+          invoice_display_name:,
+          properties: ChargeModels::FilterPropertiesService.call(
+            chargeable: child_charge,
+            properties: new_properties
+          ).properties
         )
+        child_filter.save!
+
+        filter_values.each do |key, values|
+          billable_metric_filter = child_charge.billable_metric.filters.find_by(key:)
+
+          child_filter.values.create!(
+            billable_metric_filter_id: billable_metric_filter&.id,
+            organization_id: child_charge.organization_id,
+            values:
+          )
+        end
       end
     end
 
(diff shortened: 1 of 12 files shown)


app/services/charge_filters/cascade_service.rb
```diff
@@ -0,0 +1,133 @@
...
+      child_filter = find_child_filter(child_charge)
+      return unless child_filter
+
+      child_filter.values.update_all(deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
+      child_filter.discard!
+    end
+
+    def find_child_filter(child_charge)
+      child_charge.filters.includes(values: :billable_metric_filter).find do |f|
+        f.to_h.sort == filter_values.sort
```
vincent-pochet: It seems that the `sort` can be avoided as hash comparison does not take order into account

app/services/charge_filters/destroy_service.rb
```diff
@@ -16,7 +16,7 @@ def initialize(charge_filter:, cascade_updates: false)
     def call
       return result.not_found_failure!(resource: "charge_filter") unless charge_filter
 
-      old_filters_attrs = capture_old_filters_attrs
+      filter_values = charge_filter.to_h_with_discarded
```
vincent-pochet, line 21: Not sure to get the need for `_with_discarded` here. Can you maybe add a comment?

spec/scenarios/plans/cascade_filter_updates_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    child_charge = ctx[:child_charge]
+    filter_eu = parent_charge.filters.find_by(invoice_display_name: "EU region")
+
+    expect(child_charge.filters.count).to eq(2)
+
+    delete_plan_charge_filter(parent_plan, parent_charge.code, filter_eu.id)
+
+    # Destroy via API doesn't pass cascade_updates through the helper,
+    # so cascade manually to test the destroy path
+    ChargeFilters::CascadeService.call!(
```
vincent-pochet, line 145: Does it means that today, we are not when cascading when deleting a filter on API?

vincent-pochet (commented):
Looks great!

I just want to confirm one thing: Are we 100% confident that we are not changing the override behavior regarding charges defined at a parent level but not at child one?

vincent-pochet (approved):

app/services/charge_filters/cascade_service.rb
rsempe: thanks 👍

spec/scenarios/plans/cascade_filter_updates_spec.rb
rsempe: It works today, but only if the client sends a body with the DELETE request:             
                                                                                         
```
  DELETE /api/v1/plans/.../charges/.../filters/123                                         
  Content-Type: application/json              
                                              
  {"filter": {"cascade_updates": true}}
```

The helper `delete_plan_charge_filter` does not accept it right now

annvelents (approved):
Awesome work! the only-filter cascade is so much better than the whole charge cascade! 🤩 
I might only advise to add a test on cascading pricing_group_keys change to customised filters

PR #5359, opened 2026-04-17
Revert "misc(clickhouse): Speedup deduplication (#5334)"

vincent-pochet:
This reverts commit 8b43396c14aee39ff0ab7c52dd6388c9957054c5.

## Description

The changes introduced with https://github.com/getlago/lago-api/pull/5334 can lead to invalid results when the same `timestamp`/`transaction_id` tuple is used with different filters properties. The same "duplicated event" could be counted multiple times in different buckets.


groyoh (approved):
Just to make sure we don't reproduce that issue, let's add a tests which breaks which breaks with the "fix" and passes now ?

rsempe (approved):

PR #5360, opened 2026-04-17
feat(order-forms): add OrderForm REST API

toommz:
## Roadmap Task

👉 [Quotes & Order Forms](https://getlago.canny.io/feature-requests/p/create-quotes-order-forms)

## Context

This is PR 1 of 2 in a stacked split of the original combined read-only branch. REST is separated from GraphQL so it can be QA'd via curl/Postman and shipped independently of frontend availability.

Stack:
- **#this** — OrderForm REST (merges first)
- OrderForm GraphQL (stacked) #5361

## Description

Add OrderForm read-only REST API plus the shared infrastructure used by all the stacked branches:

- `OrderFormsController` with inlined `index` and `show` actions
- `V1::OrderFormSerializer` for REST API serialization
- `OrderFormsQuery` with filters: status, customer_id, number, quote_number, owner_id, created_at range, expires_at range
- `Queries::OrderFormsQueryFiltersContract` for filter validation
- `OrderForm` factory with traits (signed, expired, voided, expiring_tomorrow, expired_yesterday)
- Route `resources :order_forms, only: %i[show index]`
- `ApiKey::RESOURCES` updated with `order_form`
- Full specs for serializer, query, controller, and model

### DIFF
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -4,6 +4,8 @@ module Api
   module V1
     class OrderFormsController < Api::BaseController
       def index
+        return forbidden_error(code: "feature_not_available") unless current_organization.feature_flag_enabled?(:order_forms)
+
         result = OrderFormsQuery.call(
           organization: current_organization,
           pagination: {
@@ -29,6 +31,8 @@ module Api
       end
 
       def show
+        return forbidden_error(code: "feature_not_available") unless current_organization.feature_flag_enabled?(:order_forms)
+
         order_form = current_organization.order_forms.find_by(id: params[:id])
         return not_found_error(resource: "order_form") unless order_form
 
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class OrderFormsController < Api::BaseController
+      def index
+        result = OrderFormsQuery.call(
+          organization: current_organization,
+          pagination: {
+            page: params[:page],
+            limit: params[:per_page] || PER_PAGE
+          },
+          filters: index_filters,
+          search_term: params[:search_term]
+        )
+
+        if result.success?
+          render(
+            json: ::CollectionSerializer.new(
+              result.order_forms,
+              ::V1::OrderFormSerializer,
+              collection_name: "order_forms",
+              meta: pagination_metadata(result.order_forms)
+            )
+          )
+        else
+          render_error_response(result)
+        end
+      end
+
+      def show
+        order_form = current_organization.order_forms.find_by(id: params[:id])
+        return not_found_error(resource: "order_form") unless order_form
+
+        render_order_form(order_form)
+      end
+
+      private
+
+      def index_filters
+        {
+          status: params[:status],
+          customer_id: params[:customer_id],
+          number: params[:number],
+          quote_number: params[:quote_number],
+          owner_id: params[:owner_id],
+          created_at_from: params[:created_at_from],
+          created_at_to: params[:created_at_to],
+          expires_at_from: params[:expires_at_from],
+          expires_at_to: params[:expires_at_to]
+        }
+      end
+
+      def render_order_form(order_form)
+        render(
+          json: ::V1::OrderFormSerializer.new(order_form, root_name: "order_form")
+        )
+      end
+
+      def resource_name
+        "order_form"
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -3,9 +3,9 @@
 module Api
   module V1
     class OrderFormsController < Api::BaseController
-      def index
-        return forbidden_error(code: "feature_not_available") unless current_organization.feature_flag_enabled?(:order_forms)
+      before_action :ensure_feature_flag!
 
+      def index
         result = OrderFormsQuery.call(
           organization: current_organization,
           pagination: {
@@ -31,8 +31,6 @@ module Api
       end
 
       def show
-        return forbidden_error(code: "feature_not_available") unless current_organization.feature_flag_enabled?(:order_forms)
-
         order_form = current_organization.order_forms.find_by(id: params[:id])
         return not_found_error(resource: "order_form") unless order_form
 
@@ -40,6 +38,10 @@ module Api
       end
 
       private
+
+      def ensure_feature_flag!
+        forbidden_error(code: "feature_not_available") unless current_organization.feature_flag_enabled?(:order_forms)
+      end
 
       def index_filters
         {
--- a/app/models/order_form.rb
+++ b/app/models/order_form.rb
@@ -32,6 +32,10 @@ class OrderForm < ApplicationRecord
 
   validates :billing_snapshot, presence: true
 
+  def self.ransackable_attributes(_ = nil)
+    %w[number]
+  end
+
   sequenced(
     scope: ->(order_form) { order_form.organization.order_forms },
     lock_key: ->(order_form) { order_form.organization_id }
--- a/app/models/order_form.rb
+++ b/app/models/order_form.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class OrderForm < ApplicationRecord
-  include Sequenced
-
   STATUSES = {
     generated: "generated",
     signed: "signed",
@@ -16,11 +14,12 @@ class OrderForm < ApplicationRecord
     invalid: "invalid"
   }.freeze
 
-  before_save :ensure_number
+  before_validation :ensure_number
 
   belongs_to :organization
   belongs_to :customer
-  belongs_to :quote
+  belongs_to :quote_version
+  has_one :quote, through: :quote_version
 
   enum :status, STATUSES,
     default: :generated,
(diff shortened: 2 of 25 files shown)


app/controllers/api/v1/order_forms_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+              meta: pagination_metadata(result.order_forms)
+            )
+          )
+        else
+          render_error_response(result)
+        end
+      end
+
+      def show
+        return forbidden_error(code: "feature_not_available") unless current_organization.feature_flag_enabled?(:order_forms)
```
vincent-pochet: Since the same guard is shared between index and show action, maybe you could turn it as a `before_action`

vincent-pochet (approved):

app/controllers/api/v1/order_forms_controller.rb
toommz: b4939f6d026883874ae30f0acb6ff62da74cfc8a

app/models/order_form.rb
```diff
@@ -0,0 +1,83 @@
...
+#  signed_at         :datetime
+#  status            :enum             default("generated"), not null
+#  void_reason       :enum
+#  voided_at         :datetime
+#  created_at        :datetime         not null
+#  updated_at        :datetime         not null
+#  customer_id       :uuid             not null
+#  organization_id   :uuid             not null
+#  quote_version_id  :uuid             not null
+#  signed_by_user_id :uuid
```
vincent-pochet: The relation to the user is not define in this model. Is it expected?

vincent-pochet (approved):

app/models/order_form.rb
toommz: Nope, you're right. I'll add it :)

toommz: 530debe0a3d79a69656860d6f301d5656762da8a

and added another one missing:

e799b092de0a50dc50b61e84a96b997ac879ca21

app/models/order_form.rb
```diff
@@ -0,0 +1,91 @@
...
+    expired: "expired",
+    invalid: "invalid"
+  }.freeze
+
+  before_save :ensure_number
+
+  belongs_to :organization
+  belongs_to :customer
+  belongs_to :quote_version
+  belongs_to :signed_by_user, class_name: "User", optional: true
```
groyoh: Should the order form be signed by a "User" ?

app/models/order_form.rb
```diff
@@ -0,0 +1,91 @@
...
+  has_one :quote, through: :quote_version
+
+  enum :status, STATUSES,
+    default: :generated,
+    validate: true
+  enum :void_reason, VOID_REASONS,
```
groyoh, line 30: Validate presence if voided.

app/models/order_form.rb
```diff
@@ -0,0 +1,91 @@
...
+    instance_methods: false,
+    validate: {allow_nil: true}
+
+  validates :billing_snapshot, presence: true
```
groyoh: I'm still not convinced by having a snapshot on the order form TBH. Could you delay adding it to the schema/model until it's actually used ?

app/queries/order_forms_query.rb
```diff
@@ -0,0 +1,110 @@
...
+    :created_at_to,
+    :expires_at_from,
+    :expires_at_to
+  ]
+
+  def call
+    return result unless validate_filters.success?
+
+    order_forms = base_scope.result
+    order_forms = with_status(order_forms) if filters.status.present?
```
groyoh, line 21: We might want to add indexes for those attributes eventually.

app/queries/order_forms_query.rb
```diff
@@ -0,0 +1,110 @@
...
+    order_forms = with_quote_number(order_forms) if filters.quote_number.present?
+    order_forms = with_owner_id(order_forms) if filters.owner_id.present?
+    order_forms = with_created_at_range(order_forms) if created_at_range?
+    order_forms = with_expires_at_range(order_forms) if expires_at_range?
+    order_forms = paginate(order_forms)
+    order_forms = apply_consistent_ordering(order_forms)
+
+    result.order_forms = order_forms
+    result
+  end
```
groyoh, line 35: `parse_datetime_filter` (used by `with_created_at_range`/`with_expires_at_range`) calls `raise_if_error!`, which raises `BaseService::FailedResult` on an unparseable date. The filters contract coerces `created_at_from` etc. with the lenient `params` `:time` type, but the query re-parses the raw string with the stricter `DateTime.iso8601`, so a value the contract accepts (e.g. a space-separated `2026-01-01 10:00:00`) can still raise here. Other queries that use `parse_datetime_filter` (`invoices_query.rb`, `credit_notes_query.rb`, `fees_query.rb`) wrap `call` with `rescue BaseService::FailedResult` -> `result` so it surfaces as a validation failure instead of an unhandled exception. Adding the same rescue here would keep the behaviour consistent.

app/queries/order_forms_query.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  private
+
+  def filters_contract
+    @filters_contract ||= Queries::OrderFormsQueryFiltersContract.new
+  end
+
+  def base_scope
+    organization.order_forms.includes(:customer, :quote_version).ransack(search_params)
```
groyoh: Do we need the customer `includes` ?

db/migrate/20260518152858_create_order_forms.rb
```diff
@@ -0,0 +1,55 @@
...
+        foreign_key: true,
+        type: :uuid
+      t.references :quote_version,
+        null: false,
+        foreign_key: true,
+        index: {unique: true},
+        type: :uuid
+      t.references :signed_by_user,
+        foreign_key: {to_table: :users},
+        index: false,
```
groyoh: Why `index: false` ?

groyoh (changes requested):

app/models/order_form.rb
toommz: `signed_by_user` records the internal Lago user who marks the order form as signed in-app (it's set by `OrderForms::MarkAsSignedService` on the sign-lifecycle branch, and by the factory `:signed` trait). It isn't the customer's signature, just who triggered the action, so it stays a `User`. Keeping as is.

app/models/order_form.rb
toommz: Added in 4cf54b6f3: `validates :void_reason, presence: true, if: :voided?`.

app/models/order_form.rb
toommz: Removed `billing_snapshot` from the migration, model, serializer, factory and seeds in 4cf54b6f3. We'll bring it back when `Order` creation actually consumes it.

app/queries/order_forms_query.rb
toommz: Added in 4cf54b6f3. Indexed the filterable columns that weren't already covered: `status`, `created_at`, `expires_at` (plus `signed_by_user_id`). `customer_id` and `number` were already indexed, and `quote_number` / `owner_id` filter through the already-indexed `quotes(organization_id, number)` and `quote_owners.user_id`.

app/queries/order_forms_query.rb
toommz: Good catch. Added `rescue BaseService::FailedResult` to `#call` in 4cf54b6f3, matching `invoices_query` / `credit_notes_query` / `fees_query`, with a spec covering a space-separated datetime that the contract accepts but `DateTime.iso8601` rejects.

app/queries/order_forms_query.rb
toommz: Correct, the serializer only reads `customer_id`, never the `customer` association, so I dropped `:customer` from the `includes` in 4cf54b6f3 and kept `:quote_version` (used for `quote_id`).

db/migrate/20260518152858_create_order_forms.rb
toommz: No good reason, that was an oversight. Removed `index: false` in 4cf54b6f3 so `signed_by_user_id` gets indexed.

app/models/order_form.rb
groyoh: We should probably rename it as `marked_as_signed_by_user` then.

db/seeds/70_order_forms.rb
```diff
@@ -4,6 +4,8 @@
 @organization = Organization.find_by!(name: "Hooli")
 @customer = Customer.find_by!(external_id: "cust_john-doe")
 
+@organization.enable_feature_flag!(:order_forms)
```
groyoh: I'm wondering whether we should enable this feature flag in the seeds 🤔 This means that anyone who checkouts the repo or runs the seed will have access to this unreleased feature. This could be confusing for customers as they may think it's ready.

db/migrate/20260518152858_create_order_forms.rb
```diff
@@ -0,0 +1,55 @@
...
+        name: "order_forms_constraint_sequential_id_positive"
+      t.index [:organization_id, :sequential_id],
+        unique: true,
+        name: "index_unique_order_forms_on_organization_sequential_id"
+      t.index [:organization_id, :number],
+        unique: true,
+        name: "index_unique_order_forms_on_organization_number"
+      t.index :status
+      t.index :created_at
+      t.index :expires_at
```
groyoh: I'd suggest scoping the indexes to org to increase index cardinality. Otherwise, there's a high chance PG will be a sequence scan rather than an index scan.

```suggestion
      t.index [:organization,:status]
      t.index [:organization,:created_at]
      t.index [:organization, :expires_at]
```

groyoh (approved):

app/models/order_form.rb
toommz: 7500d96cf3095d710a8431ce9dda9e5bb48ebeec

db/seeds/70_order_forms.rb
toommz: Removed it

db/migrate/20260518152858_create_order_forms.rb
toommz: Done 7500d96cf3095d710a8431ce9dda9e5bb48ebeec

PR #5367, opened 2026-04-20
fix(presentation_breakdowns): Change presentation_breakdowns units to remove precision

tiagolupepic:
This commit changes the unused column of presentation_breakdowns units to avoid rounding. Having an value computed from WeightedSumService, like `0.99998341049382716049e0` and associated to a presentation breakdown, it causes a rounding.

The change is quite simple, change the units column to match the fees units. Since the table is not used yet, it will not be an issue remove it in production.

### DIFF
--- a/db/migrate/20260420114717_change_presentation_breakdowns_units_decimal.rb
+++ b/db/migrate/20260420114717_change_presentation_breakdowns_units_decimal.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class ChangePresentationBreakdownsUnitsDecimal < ActiveRecord::Migration[8.0]
+  def up
+    safety_assured do
+      remove_column :presentation_breakdowns, :units # rubocop:disable Lago/NoDropColumnOrTable
+    end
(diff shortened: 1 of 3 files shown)


db/migrate/20260420114717_change_presentation_breakdowns_units_decimal.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class ChangePresentationBreakdownsUnitsDecimal < ActiveRecord::Migration[8.0]
+  def up
+    safety_assured do
+      remove_column :presentation_breakdowns, :units # rubocop:disable Lago/NoDropColumnOrTable
```
D1353L, line 6: Is there a reason not to use `change_column` instead?

D1353L (approved):

db/migrate/20260420114717_change_presentation_breakdowns_units_decimal.rb
tiagolupepic: It's the same actually, for this case (no data and the table is not used), there is no difference

PR #5366, opened 2026-04-20
misc(usage): Reduce subscription flagging delay and speedup clock

vincent-pochet:
## Context

This PR is related to https://github.com/getlago/lago/pull/729

## Description

Following the changes in the events_processor, it reduces the subscription refresh bucket duration from 15 seconds to 10 and it consumes the queue every 10 seconds instead of every 1 minute.
The main goal is to get closer to "realtime" wallet refresh and alerting.

The `CLICKHOUSE_MERGE_DELAY` constant is renamed into `SUBSCRIPTION_BUCKET_DURATION` to better reflect its purpose.

PR #5368, opened 2026-04-20
misc(wallet): Dedicated refresh balance worker

vincent-pochet:
## Description

This PR adds the ability to configure a dedicated wallet refresh worker to move some organization wallets closer to "realtime" wallet refresh. 

This changes follow the improvement made on the subscription usage flagging front (https://github.com/getlago/lago-api/pull/5366)

The new dedicated clock will be performed every 5 seconds, enqueuing wallet refresh jobs in a new dedicated `dedicated_wallet_refresh` queue.


clock.rb
```diff
@@ -47,12 +47,21 @@ module Clockwork
...
+    unless ENV["LAGO_DISABLE_WALLET_REFRESH"] == "true"
+      every(5.minutes, "schedule:refresh_wallets_ongoing_balance") do
         Clock::RefreshWalletsOngoingBalanceJob
           .set(sentry: {"slug" => "lago_refresh_wallets_ongoing_balance", "cron" => "*/5 * * * *"})
           .perform_later
       end
+
+      if Utils::DedicatedWalletRefreshConfig.any?
+        dedicated_wallet_refresh_interval = ENV["LAGO_DEDICATED_WALLET_REFRESH_INTERVAL_SECONDS"].presence.to_i
+        dedicated_wallet_refresh_interval = dedicated_wallet_refresh_interval.positive? ? dedicated_wallet_refresh_interval : 5
```
toommz: This logic could have been in `Utils::DedicatedWalletRefreshConfig`:

```ruby
module Utils
  module DedicatedWalletRefreshConfig
    def self.refresh_interval
      # Implementation
    end
  end
end
```

Not a blocker but as we have a dedicated module for the config it would have removed some logic from the `clock.rb` file 🤔

toommz (approved):

clock.rb
vincent-pochet: Good idea indeed!

PR #5369, opened 2026-04-20
perf(invoices): Preload customer associations on invoices

D1353L:
## Context

The invoice serializer accesses `customer` fields for every invoice. Without eager loading, this has an N+1 queries issue on all invoice list endpoints. This causes some `GET /invoices` requests to be slow (up to 11s)

## Description

Include customer fields in the InvoicesQuery, which is used in all related serializers/resolvers.

### DIFF
--- a/app/controllers/concerns/invoice_index.rb
+++ b/app/controllers/concerns/invoice_index.rb
@@ -72,7 +72,6 @@ module InvoiceIndex
             :gocardless_customer,
             :cashfree_customer,
             :adyen_customer,
-            :flutterwave_customer,
             :moneyhash_customer,
             {integration_customers: :integration}
           ]
--- a/app/controllers/concerns/invoice_index.rb
+++ b/app/controllers/concerns/invoice_index.rb
@@ -60,7 +60,23 @@ module InvoiceIndex
 
     if result.success?
       invoices = Invoice.preload_offset_amounts(
-        result.invoices.includes(:metadata, :applied_taxes, :billing_entity, :applied_usage_thresholds)
+        result.invoices.includes(
+          :metadata,
+          :applied_taxes,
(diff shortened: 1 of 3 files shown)


vincent-pochet (approved):

spec/requests/api/v1/invoices_controller_spec.rb
```diff
@@ -332,6 +332,52 @@
       end
     end
 
+    context "with N+1 query detection on customer and tax associations", :with_bullet, bullet: {n_plus_one_query: true, unused_eager_loading: false} do
```
rsempe: do we really want to keep this test?

app/controllers/concerns/invoice_index.rb
```diff
@@ -60,7 +60,23 @@ def invoice_index(customer_external_id: nil)
...
+          :billing_entity,
+          :applied_usage_thresholds,
+          customer: [
+            :billing_entity,
+            :metadata,
+            :stripe_customer,
+            :gocardless_customer,
+            :cashfree_customer,
+            :adyen_customer,
+            :flutterwave_customer,
```
rsempe: not sure `flutterwave_customer` is needed, right?

rsempe (approved):

spec/requests/api/v1/invoices_controller_spec.rb
D1353L: I think it's worth keeping this for regression coverage. I also cleaned up duplicated attachments tests

app/controllers/concerns/invoice_index.rb
D1353L: Yeah, it isn't used in this flow

PR #5370, opened 2026-04-20
feat(payment-gated-subs): add gated activation path to ActivateService

ancorcruz:
## Context

When a subscription has pending activation rules, it should enter an incomplete state instead of activating immediately. The billing flow creates an invoice and attempts payment before the subscription becomes active.

## Description

Extend ActivateService to handle the full subscription activation lifecycle with three distinct paths based on subscription status:

- **pending**: evaluates rules via EvaluateService, then either gates (marks as incomplete, emits events, bills, sends incomplete webhook) or activates normally if rules are not applicable
- **incomplete (resolved)**: marks as active + sends started webhook + hubspot sync. Skips billing and events since they were already done during gating. Bills only if no payment rules exist (for future non-payment rule types)
- **active/gated**: returns early (guard clauses)

Make EvaluateService pure — remove ResolveSubscriptionStatusService call. EvaluateService only evaluates individual rules with no subscription status changes.

Add payment_gated? predicate to Subscription model for distinguishing payment-gated from non-payment-gated incomplete subscriptions.

Register subscription.incomplete and subscription.canceled in Clickhouse activity log.

## Changes since last review

- **Redesigned ActivateService** with status-based branching to avoid double billing (the previous version applied the same side effects regardless of entry point, causing duplicate InvoiceSubscription on the post-payment activation path)
- **EvaluateService made pure** — removed ResolveSubscriptionStatusService call to eliminate circular dependency risk (ActivateService now calls EvaluateService internally for pending subscriptions)
- **Added payment_gated?** to Subscription model for explicit billing decisions in ActivateService
- **Adjusted activation scenario test dates** for timezone-aware date comparison in CreateService
- **Comprehensive ActivateService spec rewrite** covering all paths: pending without rules, pending with gated rules, pending with not_applicable rules, incomplete post-payment (no re-billing), incomplete without payment rules (bills), guard clauses


app/services/subscriptions/activate_service.rb
```diff
@@ -24,27 +27,38 @@ def call
       after_commit do
         bill_subscription
 
-        SendWebhookJob.perform_later("subscription.started", subscription)
-        Utils::ActivityLog.produce(subscription, "subscription.started")
+        if gated?
+          SendWebhookJob.perform_later("subscription.incomplete", subscription)
+          Utils::ActivityLog.produce(subscription, "subscription.incomplete")
```
mariohd, line 32: `subscription.incomplete` this needs to be listed here https://github.com/getlago/lago-api/blob/main/app/models/clickhouse/activity_log.rb#L42

app/services/subscriptions/activate_service.rb
```diff
@@ -24,27 +27,38 @@ def call
       after_commit do
         bill_subscription
```
mariohd, line 28: are we going to `bill_subscription` even if `subscription.mark_as_incomplete!`? 

I know it skips some charges, but it still charge the subscription fees, should we charge this even for incomplete ones?

osmarluz (dismissed):
LGTM! 🚀

app/services/subscriptions/activate_service.rb
ancorcruz: yes, that is exactly the point of `incomplete`. It is becuase of to be finalized need first the payment (or other rules to be satisfied) perhaps we could also ensure it is gated with payment (the only kind of gating right now). let me improve that detail

app/services/subscriptions/activate_service.rb
```diff
@@ -24,27 +27,44 @@ def call
...
+      subscription.incomplete?
+    end
+
+    def payment_gated?
+      gated? && subscription.activation_rules.payment.pending.any?
+    end
+
     def bill_subscription
+      return if gated? && !payment_gated?
+
```
ancorcruz: @osmarluz I have added this to ensure we only bill payment gated subs instead of all kind of gating.

osmarluz: @ancorcruz looks good!

app/services/subscriptions/activate_service.rb
ancorcruz: `bill_subscription` now returns when the subscription is gated and not payment_gated, that means, we run billing only for:
1. active subs
2. payment gated subs.

app/services/subscriptions/activate_service.rb
```diff
@@ -11,10 +11,14 @@ def initialize(subscription:, timestamp: Time.current)
     end
 
     def call
-      result.subscription = subscription
       return result if subscription.active?
+      return result if subscription.incomplete? && subscription.pending_rules?
```
ancorcruz: we could use `if subscription.gated?` instead.

app/models/clickhouse/activity_log.rb
```diff
@@ -68,6 +68,8 @@ class ActivityLog < BaseRecord
       billing_entities_created: "billing_entities.created",
       billing_entities_updated: "billing_entities.updated",
       billing_entities_deleted: "billing_entities.deleted",
+      subscription_canceled: "subscription.canceled",
```
osmarluz, line 71: @ancorcruz I noticed we're not calling `Utils::ActivityLog.produce(subscription, "subscription.canceled")` on `ResolveSubscriptionStatusService`. We should probably have it there alongside the webhook call

app/services/subscriptions/activation_rules/evaluate_service.rb
```diff
@@ -18,8 +18,6 @@ def call
           result.rules << rule
         end
 
-        ResolveSubscriptionStatusService.call!(subscription:)
```
osmarluz, line 21: @ancorcruz it's much better not to have this here. Good job!

osmarluz (approved):
🚀

PR #5372, opened 2026-04-21
misc(subscription_activity): Dedicated processing queue

vincent-pochet:
## Context

This PR follow https://github.com/getlago/lago-api/pull/5368 its goal is to make alert and progressive billing almost "realtime" (via the `SubscriptionActivity`) for certain organizations by relying on a dedicated worker,

### DIFF
--- a/app/jobs/clock/process_dedicated_orgs_subscription_activities_job.rb
+++ b/app/jobs/clock/process_dedicated_orgs_subscription_activities_job.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Clock
+  class ProcessDedicatedOrgsSubscriptionActivitiesJob < ClockJob
+    unique :until_executed, on_conflict: :log
+
+    def perform
+      return unless License.premium?
+
+      organization_ids = Utils::DedicatedWorkerConfig.organization_ids
+      return if organization_ids.empty?
+
+      UsageMonitoring::SubscriptionActivity
+        .where(organization_id: organization_ids, enqueued: false)
+        .distinct
+        .pluck(:organization_id)
+        .each do |organization_id|
+          UsageMonitoring::ProcessOrganizationSubscriptionActivitiesJob.perform_later(organization_id)
+        end
+    end
+  end
+end
--- a/app/services/usage_monitoring/process_all_subscription_activities_service.rb
+++ b/app/services/usage_monitoring/process_all_subscription_activities_service.rb
@@ -9,7 +9,12 @@ module UsageMonitoring
       #     This is also where we should report metrics
       #     That's why it's a dedicated service and not just done in the job
 
-      SubscriptionActivity.group(:organization_id).count.each do |organization_id, count|
+      scope = SubscriptionActivity.where(enqueued: false)
+
+      dedicated_org_ids = Utils::DedicatedWorkerConfig.organization_ids
+      scope = scope.where.not(organization_id: dedicated_org_ids) if dedicated_org_ids.any?
+
+      scope.distinct.pluck(:organization_id).each do |organization_id|
         ProcessOrganizationSubscriptionActivitiesJob.perform_later(organization_id)
       end
 
--- a/app/services/usage_monitoring/process_organization_subscription_activities_service.rb
+++ b/app/services/usage_monitoring/process_organization_subscription_activities_service.rb
@@ -16,10 +16,15 @@ module UsageMonitoring
 
       # NOTE: If we need to handle different delays per organization, this would be done here.
 
+      queue_name = if dedicated?
+        Utils::DedicatedWorkerConfig::DEDICATED_ALERTS_QUEUE.to_s
+      else
+        ProcessSubscriptionActivityJob.queue_name
+      end
+
       organization.subscription_activities.where(enqueued: false).select(:id).in_batches(of: BATCH_SIZE) do |batch|
-        jobs = []
-        batch.each do |subscription_activity|
-          jobs << ProcessSubscriptionActivityJob.new(subscription_activity.id)
+        jobs = batch.map do |sa|
+          ProcessSubscriptionActivityJob.new(sa.id).tap { |j| j.queue_name = queue_name }
         end
 
         ActiveRecord::Base.transaction do
@@ -37,5 +42,9 @@ module UsageMonitoring
     private
 
     attr_reader :organization
+
+    def dedicated?
+      Utils::DedicatedWorkerConfig.enabled_for?(organization.id)
+    end
   end
 end
--- a/config/sidekiq/sidekiq_dedicated.yml
+++ b/config/sidekiq/sidekiq_dedicated.yml
@@ -2,7 +2,8 @@ concurrency: <%= ENV.fetch('SIDEKIQ_CONCURRENCY', 10) %>
 timeout: 25
 retry: 1
 queues:
-  - dedicated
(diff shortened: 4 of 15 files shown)


config/sidekiq/sidekiq_dedicated.yml
```diff
@@ -2,7 +2,8 @@ concurrency: <%= ENV.fetch('SIDEKIQ_CONCURRENCY', 10) %>
 timeout: 25
 retry: 1
 queues:
-  - dedicated
+  - ["dedicated_alerts", 1]
+  - ["dedicated_wallets", 1]
```
vincent-pochet, line 6: This makes sure that both queues have the same priority

app/jobs/clock/process_dedicated_orgs_subscription_activities_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      return unless License.premium?
+
+      organization_ids = Utils::DedicatedWorkerConfig.organization_ids
+      return if organization_ids.empty?
+
+      UsageMonitoring::SubscriptionActivity
+        .where(organization_id: organization_ids, enqueued: false)
+        .distinct
+        .pluck(:organization_id)
+        .each do |organization_id|
```
toommz, line 17: I know those are only ids but should we do that `in_batches`?

toommz (approved):

app/jobs/clock/process_dedicated_orgs_subscription_activities_job.rb
```diff
@@ -0,0 +1,22 @@
...
+    unique :until_executed, on_conflict: :log
+
+    def perform
+      return unless License.premium?
+
+      organization_ids = Utils::DedicatedWorkerConfig.organization_ids
+      return if organization_ids.empty?
+
+      UsageMonitoring::SubscriptionActivity
+        .where(organization_id: organization_ids, enqueued: false)
```
groyoh, line 14: Should we exclude these orgs from the `Clock::ProcessAllSubscriptionActivitiesJob` ?

vincent-pochet: @groyoh 
The job is just calling `UsageMonitoring::ProcessAllSubscriptionActivitiesService` with no arguments. This last service is excluding these orgs so I think we are covered

app/jobs/clock/process_dedicated_orgs_subscription_activities_job.rb
vincent-pochet: @toommz 
The main purpose of this setup is to add a temporary setup to make some organizations more "realtime" compared to all others.
Since as it is, it will be really heavy for the infrastructure, we are not expecting more than a few (1 or 2) organizations with this setup. Otherwise it would just make it pointless and would probably cause huge scalability issues.

The end-goal is to remove it completely and finally rely on the pre-enriched (and one day pre-aggregated) event pipeline to speedup the processing for everybody.

toommz: Thanks for the clarification!

app/services/usage_monitoring/process_organization_subscription_activities_service.rb
```diff
@@ -37,5 +39,13 @@ def call
...
 
     attr_reader :organization
+
+    def dedicated_queue_name
+      return @dedicated_queue_name if defined?(@dedicated_queue_name)
+
+      @dedicated_queue_name = if Utils::DedicatedWorkerConfig.enabled_for?(organization.id)
+        Utils::DedicatedWorkerConfig::DEDICATED_ALERTS_QUEUE.to_s
+      end
+    end
```
groyoh, line 48: The queue name doesn't change between batches or between jobs, so we can compute it once at the top of `#call` and always assign to `job.queue_name`, which also removes the conditional in the loop:

```ruby
def call
  nb_jobs_enqueued = 0
  queue_name = dedicated? ? Utils::DedicatedWorkerConfig::DEDICATED_ALERTS_QUEUE.to_s : ProcessSubscriptionActivityJob.queue_name

  organization.subscription_activities.where(enqueued: false).select(:id).in_batches(of: BATCH_SIZE) do |batch|
    jobs = batch.map do |sa|
      ProcessSubscriptionActivityJob.new(sa.id).tap { |j| j.queue_name = queue_name }
    end
    ...
  end
end

def dedicated?
  Utils::DedicatedWorkerConfig.enabled_for?(organization.id)
end
```

app/jobs/clock/process_dedicated_orgs_subscription_activities_job.rb
```diff
@@ -0,0 +1,22 @@
...
+        .distinct
+        .pluck(:organization_id)
+        .each do |organization_id|
+          UsageMonitoring::ProcessOrganizationSubscriptionActivitiesJob.perform_later(organization_id)
+        end
+    end
+  end
+end
```
groyoh, line 22: With a 5s default refresh interval, we need to make sure there's a supporting index on `(organization_id, enqueued)` or that existing index performance is sufficiently fast to avoid overlaps in jobs.

app/services/usage_monitoring/process_all_subscription_activities_service.rb
```diff
@@ -9,7 +9,12 @@ def call
       #     This is also where we should report metrics
       #     That's why it's a dedicated service and not just done in the job
 
-      SubscriptionActivity.group(:organization_id).count.each do |organization_id, count|
+      scope = SubscriptionActivity.all
+
+      dedicated_org_ids = Utils::DedicatedWorkerConfig.organization_ids
+      scope = scope.where.not(organization_id: dedicated_org_ids) if dedicated_org_ids.any?
+
+      scope.group(:organization_id).count.each do |organization_id, count|
```
groyoh:
```suggestion
      scope.distinct.pluck(:organization_id).each do |organization_id|
```

app/services/usage_monitoring/process_all_subscription_activities_service.rb
```diff
@@ -9,7 +9,12 @@ def call
       #     This is also where we should report metrics
       #     That's why it's a dedicated service and not just done in the job
 
-      SubscriptionActivity.group(:organization_id).count.each do |organization_id, count|
+      scope = SubscriptionActivity.all
+
+      dedicated_org_ids = Utils::DedicatedWorkerConfig.organization_ids
+      scope = scope.where.not(organization_id: dedicated_org_ids) if dedicated_org_ids.any?
```
groyoh, line 15: Ain't we missing the `enqueued: false` condition here as well ?

groyoh (approved):

app/services/usage_monitoring/process_all_subscription_activities_service.rb
vincent-pochet: We are indeed. I'm adding it :+1:

app/jobs/clock/process_dedicated_orgs_subscription_activities_job.rb
vincent-pochet: Both the clock job and the `ProcessOrganizationSubscriptionActivitiesJob` are configured to be `uniq`  so it should not be an issue. Plus, this last job is flagging `SubscriptionActivities` as `enqueued` and (with the current fix) we are not re-enqueing already processing activities.

PR #5373, opened 2026-04-21
feat(payment-gated-subs): handle gated subscriptions in Invoices::SubscriptionService

ancorcruz:
## Context

When a gated subscription is billed, the invoice must be created as open (not finalized or draft), the grace period must be skipped, and side effects like webhooks and document generation must be deferred until payment succeeds. Only the payment attempt should be triggered.

## Description

Add gated subscription handling to Invoices::SubscriptionService. After creating the generating invoice, set status to open if any subscription is gated. Skip grace period for gated subscriptions. In the post-billing side effects, route gated invoices to payment only, skipping webhooks, documents, integrations, and segment tracking.

### DIFF
--- a/app/services/invoices/subscription_service.rb
+++ b/app/services/invoices/subscription_service.rb
@@ -24,6 +24,7 @@ module Invoices
       return result if active_subscriptions.empty? && recurring
 
       create_generating_invoice unless invoice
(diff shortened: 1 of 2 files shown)


osmarluz (approved):
🚀

PR #5374, opened 2026-04-21
feat(payment-gated-subs): handle gated subscriptions in CreatePayInAdvanceFixedChargesService

ancorcruz:
## Context

When a gated subscription has pay-in-advance fixed charges on a pay-in-arrears plan, the fixed charge invoice must be created as open and only trigger payment, deferring all other side effects until payment succeeds.

## Description

Allow the service to process gated subscriptions alongside active ones. After creating the generating invoice, set status to open if the subscription is gated. Route post-billing side effects to payment only for gated invoices, skipping webhooks, documents, integrations, and segment tracking.

### DIFF
--- a/app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
+++ b/app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
@@ -14,7 +14,7 @@ module Invoices
     end
 
     def call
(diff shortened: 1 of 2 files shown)


osmarluz (approved):
🚀

PR #5375, opened 2026-04-21
feat(payment-gated-subs): handle gated invoices in tax edge case services

ancorcruz:
## Context

When a gated invoice encounters tax provider errors or pending VIES validation, the invoice status must remain open (not transition to pending) and post-tax-success side effects must route to payment only, deferring webhooks, documents, and integrations until the subscription activates.

## Description

Apply consistent gated invoice handling across six tax-related services and jobs:

- ComputeTaxesAndTotalsService: preserve open status when setting pending tax status for gated invoices
- EnsureCompletedViesCheckService: preserve open status during VIES pending state for gated invoices
- RetryService: restore to open (not pending) when retrying failed tax on gated invoices, checking subscription gated status directly since the invoice is in failed state at retry time
- PullTaxesAndApplyService: expand guard to allow gated invoices and route post-success to payment only
- FinalizePendingViesInvoiceService: expand guard to allow gated invoices and route post-success to payment only
- ViesCheckJob: expand pending invoice query to include open invoices so gated invoices blocked by VIES are picked up for finalization

### DIFF
--- a/app/jobs/customers/vies_check_job.rb
+++ b/app/jobs/customers/vies_check_job.rb
@@ -33,7 +33,7 @@ module Customers
 
     def enqueue_pending_invoice_finalization(customer)
       customer.invoices
-        .where(status: %i[pending open], tax_status: "pending")
+        .where(status: %i[pending open], tax_status: :pending)
         .find_each do |invoice|
           Invoices::FinalizePendingViesInvoiceJob.perform_later(invoice)
         end
--- a/app/jobs/customers/vies_check_job.rb
+++ b/app/jobs/customers/vies_check_job.rb
@@ -32,9 +32,11 @@ module Customers
     end
 
     def enqueue_pending_invoice_finalization(customer)
-      customer.invoices.pending.where(tax_status: "pending").find_each do |invoice|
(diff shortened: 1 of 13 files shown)


app/jobs/customers/vies_check_job.rb
```diff
@@ -32,9 +32,11 @@ def schedule_retry(customer, vies_check_result)
...
 
     def enqueue_pending_invoice_finalization(customer)
-      customer.invoices.pending.where(tax_status: "pending").find_each do |invoice|
-        Invoices::FinalizePendingViesInvoiceJob.perform_later(invoice)
-      end
+      customer.invoices
+        .where(status: %i[pending open], tax_status: :pending)
+        .find_each do |invoice|
+          Invoices::FinalizePendingViesInvoiceJob.perform_later(invoice)
+        end
```
osmarluz, line 44: @ancorcruz I think it would be good to add a note here explaining that the combination of `"status:open + tax_status:pending"` is only possible for gated invoices, that's why we're not checking that explicitly on the query and just adding the `open` status to it. What do you think?

osmarluz (approved):
@ancorcruz just left a suggestion for `app/jobs/customers/vies_check_job.rb`

PR #5377, opened 2026-04-21
fix(applied_coupons): Validate frequency_duration for recurring coupons

D1353L:
## Context

Recurring coupons can be created through the API with `frequency: "recurring"` and `frequency_duration: nil`. This leads to applied coupons with `frequency_duration_remaining: nil`, and invoice finalization fails when `Credits::AppliedCouponService` decrements the remaining duration. The same applies to coupons as well.

## Description

Add validation for `frequency_duration` and `frequency_duration_remaining` when the coupon/applied coupon is `recurring`.

### DIFF
--- a/app/models/applied_coupon.rb
+++ b/app/models/applied_coupon.rb
@@ -28,6 +28,8 @@ class AppliedCoupon < ApplicationRecord
 
   validates :amount_cents, numericality: {greater_than_or_equal_to: 0}, allow_nil: true
   validates :amount_currency, inclusion: {in: currency_list}, allow_nil: true
+  validates :frequency_duration, presence: true, numericality: {greater_than: 0}, if: :recurring?
+  validates :frequency_duration_remaining, presence: true, numericality: {greater_than_or_equal_to: 0}, if: :recurring?
 
   def mark_as_terminated!(timestamp = Time.zone.now)
     self.terminated_at ||= timestamp
--- a/app/models/coupon.rb
+++ b/app/models/coupon.rb
@@ -59,6 +59,8 @@ class Coupon < ApplicationRecord
 
   validates :percentage_rate, presence: true, if: :percentage?
 
+  validates :frequency_duration, presence: true, numericality: {greater_than: 0}, if: :recurring?
+
   validates :reusable, exclusion: [nil]
 
   default_scope -> { kept }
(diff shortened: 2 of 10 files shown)


brunomiguelpinto (approved):

app/services/credits/applied_coupon_service.rb
```diff
@@ -39,7 +39,10 @@ def call
         fee.save!
       end
 
-      applied_coupon.frequency_duration_remaining -= 1 if applied_coupon.recurring?
+      if applied_coupon.recurring?
+        applied_coupon.frequency_duration_remaining = [applied_coupon.frequency_duration_remaining.to_i - 1, 0].max
+      end
```
tiagolupepic: all this logic belongs to `applied_coupon` object, do we should move to a method instead? something like `set_frequency_duration_remaining!`

app/models/applied_coupon.rb
```diff
@@ -28,6 +28,8 @@ class AppliedCoupon < ApplicationRecord
 
   validates :amount_cents, numericality: {greater_than_or_equal_to: 0}, allow_nil: true
   validates :amount_currency, inclusion: {in: currency_list}, allow_nil: true
+  validates :frequency_duration, presence: true, numericality: {greater_than: 0}, if: :recurring?
+  validates :frequency_duration_remaining, presence: true, numericality: {greater_than_or_equal_to: 0}, if: :recurring?
```
tiagolupepic, line 32: is fine to have this validation for the existing persisted applied coupons? The scenario I have in mind: We currently have some applied coupon persisted and they're failing because frequency_duration_remaining is nil, but if another service tries to update it and doesn't assign the `frequency_duration_remaining` correctly, the validation will fail

app/models/coupon.rb
```diff
@@ -59,6 +59,8 @@ class Coupon < ApplicationRecord
 
   validates :percentage_rate, presence: true, if: :percentage?
 
+  validates :frequency_duration, presence: true, numericality: {greater_than: 0}, if: :recurring?
```
tiagolupepic, line 62: same scenario here just to double check, existing coupons persisted might fail when another service tries to update them

app/models/applied_coupon.rb
D1353L: Yeah, I'll add a backfill migration for OSS users. Our DBs are fine - the data was corrected.

app/services/credits/applied_coupon_service.rb
D1353L: Good idea, added the `decrement_frequency_duration_remaining` method

PR #5386, opened 2026-04-22
misc(invoices): Backfill updated_at from metadata

aquinofb:
## Context

Invoice metadata created before `belongs_to :invoice, touch: true` was added to `Metadata::InvoiceMetadata` does not bump the parent invoice's `updated_at`. Downstream data sync pipelines that watch that column to detect changes therefore skip those invoices and their metadata never propagates, even though it is perfectly visible inside Lago.

New writes are now correct, but the historical drift is still there for organizations that accumulated metadata before the `touch: true` fix landed. This PR adds the tool to converge those rows back to a consistent state on demand, without a migration (this is a targeted, operator-run fix rather than something every Lago deployment needs to apply).

## Description

Adds a rake task:

```
rake invoices:backfill_metadata_updated_at[organization_id]
```

For each visible-status invoice belonging to the given organization whose `updated_at` lags behind the latest metadata row, the task updates `updated_at` to `MAX(invoice_metadata.updated_at)`. It runs in batches, respects a total cap, and is idempotent.

Key design choices:

- **Batched with a total cap** (`BATCH_SIZE` default 1_000, `TOTAL_LIMIT` default 10_000). Keeps each transaction short-lived and lets an operator observe the first few batches before running at full volume.
- **Idempotent**. The outer ``WHERE invoices.updated_at < stale.max_metadata_updated_at`` guard means rows already bumped by an earlier invocation (or by a concurrent touch) are not rewritten.
- **Scoped to visible statuses only** (`Invoice::VISIBLE_STATUS`), matching the export view that sync consumers read from. `generating`, `open`, and `closed` invoices are intentionally left alone.
- **Functional-dependency GROUP BY**. Postgres allows the bare `invoices.updated_at` reference in `HAVING` because the GROUP BY is on `invoices.id` (the primary key), so no aggregate wrapper is needed and the stale filter runs inside the LIMITed subquery.

## How to try locally

1. In a Rails console, pick an org and attach metadata to one of its visible invoices, then backdate the invoice so it looks historically stale:

   ```ruby
   org = Organization.first
   invoice = org.invoices.visible.last
   invoice.metadata.create!(organization: org, key: "k", value: "v")  # touches invoice
   invoice.update_column(:updated_at, 1.day.ago)
   ```

2. Run the task with small limits so you can eyeball the output:

   ```
   lago exec api bundle exec rake 'invoices:backfill_metadata_updated_at[<org_id>]' BATCH_SIZE=100 TOTAL_LIMIT=100
   ```

   Expected output:

   ```
   Starting backfill for organization <org_id> (batch_size: 100, total_limit: 100)...
     Batch updated: 1 rows (total: 1)
   Done. Total updated: 1. Remaining stale invoices: 0.
   ```

3. Re-run the same command to confirm idempotency: the second run reports `Batch updated: 0 rows` and does not touch the invoice again.

4. Run the spec:

   ```
   lago exec api bundle exec rspec spec/lib/tasks/invoices_rake_spec.rb
   ```

   7 examples covering the empty case, a single stale invoice, invisible-status invoices, cross-org scoping, idempotency, the `TOTAL_LIMIT` cap, and the missing-argument abort.

### DIFF
--- a/lib/tasks/invoices.rake
+++ b/lib/tasks/invoices.rake
@@ -70,13 +70,15 @@ namespace :invoices do
         UPDATE invoices
         SET updated_at = stale.max_metadata_updated_at
         FROM (
-          SELECT invoices.id, MAX(im.updated_at) AS max_metadata_updated_at
-          FROM invoices
-          JOIN invoice_metadata im ON im.invoice_id = invoices.id
-          WHERE invoices.organization_id = $1
+          SELECT im.invoice_id AS id, MAX(im.updated_at) AS max_metadata_updated_at
+          FROM invoice_metadata im
+          JOIN invoices
+            ON invoices.id = im.invoice_id
+           AND im.updated_at > invoices.updated_at
+          WHERE im.organization_id = $1
             AND invoices.status IN (#{visible_statuses_sql})
-          GROUP BY invoices.id
-          HAVING MAX(im.updated_at) > invoices.updated_at
+          GROUP BY im.invoice_id
+          ORDER BY im.invoice_id
           LIMIT #{effective_limit}
         ) AS stale
         WHERE invoices.id = stale.id
--- a/lib/tasks/invoices.rake
+++ b/lib/tasks/invoices.rake
@@ -43,4 +43,59 @@ namespace :invoices do
   task fill_expected_finalization_date: :environment do
     Invoice.in_batches(of: 10_000).update_all("expected_finalization_date = COALESCE(expected_finalization_date, issuing_date)") # rubocop:disable Rails/SkipsModelValidations
   end
+
+  desc "Touch invoices.updated_at to MAX(invoice_metadata.updated_at) so Prequel re-syncs stale metadata"
+  task :backfill_metadata_updated_at, [:organization_id] => :environment do |_task, args|
+    organization_id = args[:organization_id]
+    abort "Missing organization_id argument\n\nUsage: rake invoices:backfill_metadata_updated_at[organization_id]" if organization_id.blank?
+
+    batch_size = (ENV["BATCH_SIZE"] || 1_000).to_i
+    total_limit = (ENV["TOTAL_LIMIT"] || 10_000).to_i
+    abort "BATCH_SIZE must be positive" if batch_size <= 0
+    abort "TOTAL_LIMIT must be positive" if total_limit <= 0
+
+    # Interpolated directly since these are trusted integer constants, not user input.
+    visible_statuses_sql = Invoice::VISIBLE_STATUS.values.join(",")
+    total_updated = 0
+
+    puts "Starting backfill for organization #{organization_id} (batch_size: #{batch_size}, total_limit: #{total_limit})..."
+
+    loop do
+      remaining_budget = total_limit - total_updated
+      break if remaining_budget <= 0
+
+      effective_limit = [batch_size, remaining_budget].min
+
+      update_sql = <<~SQL.squish
+        UPDATE invoices
+        SET updated_at = stale.max_metadata_updated_at
+        FROM (
+          SELECT invoices.id, MAX(im.updated_at) AS max_metadata_updated_at
+          FROM invoices
+          JOIN invoice_metadata im ON im.invoice_id = invoices.id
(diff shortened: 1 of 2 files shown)


spec/lib/tasks/invoices_rake_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      expect { task.invoke(organization.id) }.to output(/Remaining stale invoices: 1/).to_stdout
+    end
+  end
+
+  context "without an organization_id argument" do
+    it "aborts with a usage message" do
+      expect { task.invoke }.to raise_error(SystemExit).and output(/Missing organization_id argument/).to_stderr
+    end
+  end
+end
```
groyoh, line 133: No test exercises the multi-iteration path: every happy-path context has at most one stale invoice, and the `TOTAL_LIMIT` test caps at one. The core loop — process a batch, decrement the budget, continue — is therefore untested end-to-end. Add a context with, say, 3 stale invoices and `BATCH_SIZE=1, TOTAL_LIMIT=10` that asserts all 3 are bumped and `Remaining stale invoices: 0` is reported, so a regression in the loop termination logic (e.g., an off-by-one on `remaining_budget` or `break if rows < effective_limit`) would be caught.

groyoh (approved):

lib/tasks/invoices.rake
```diff
@@ -43,4 +43,59 @@ namespace :invoices do
...
+
+    loop do
+      remaining_budget = total_limit - total_updated
+      break if remaining_budget <= 0
+
+      effective_limit = [batch_size, remaining_budget].min
+
+      update_sql = <<~SQL.squish
+        UPDATE invoices
+        SET updated_at = stale.max_metadata_updated_at
```
groyoh, line 71: The query is correct and safe, but a few design choices make each batch cost O(invoices-in-org) instead of O(stale-rows). At backfill scale (the realistic case here) that adds up. Three cheap wins, in order of payoff:

**1. Push the stale predicate into the join, not `HAVING`.**
Today every `invoice × metadata` row for the org is materialized and aggregated before `HAVING` discards fresh ones. Filtering at join time is strictly more selective and lets the planner skip most rows:

```sql
FROM invoices
JOIN invoice_metadata im
  ON im.invoice_id = invoices.id
 AND im.updated_at > invoices.updated_at   -- row-level early filter
WHERE invoices.organization_id = $1
  AND invoices.status IN (#{visible_statuses_sql})
GROUP BY invoices.id
LIMIT #{effective_limit}
```

`MAX(im.updated_at)` in the SELECT still works, and `HAVING` becomes redundant.

**2. Drive the query from `invoice_metadata`, not `invoices`.**
`invoice_metadata` is capped at 5 rows per invoice (`COUNT_PER_INVOICE`) and has its own `index_invoice_metadata_on_organization_id`. For orgs with many invoices but relatively few with metadata, pivoting the query around metadata means only invoices that *have* metadata are ever considered — a tiny fraction of the invoice table for most orgs.

**3. Add `ORDER BY invoices.id` to the `LIMIT` subquery.**
Without an order, PG may revisit the same-ish rows across iterations, and the `break if rows < effective_limit` heuristic only holds probabilistically. An ordered subquery is deterministic at zero cost, and it opens the door to keyset pagination (`AND invoices.id > :last_seen_id`) if per-iteration re-scan cost turns out to matter in production.

Before shipping: `EXPLAIN (ANALYZE, BUFFERS)` against a prod-sized org would confirm whether PG streams the aggregate under the `LIMIT` or materializes it first — the plan dictates which of the three is most worth doing.

spec/lib/tasks/invoices_rake_spec.rb
aquinofb: Addressed by https://github.com/getlago/lago-api/pull/5386/changes/4be3114cbcafa8aee4691d0ebd15d2ed2f86f527.

lib/tasks/invoices.rake
aquinofb: Addressed by https://github.com/getlago/lago-api/pull/5386/changes/4be3114cbcafa8aee4691d0ebd15d2ed2f86f527.

PR #5387, opened 2026-04-22
fix(fees): Add original_fee_id with improved uniqueness index

D1353L:
## Context

This PR should be deployed before #5330.

In the regenerate-voided-invoice flow, fees are duplicated onto a new invoice. For pay-in-advance charges, fees have a `pay_in_advance_event_transaction_id`, and duplicate fees should keep the same `id` since they refer to the same event. The existing uniqueness index on `(pay_in_advance_event_transaction_id, charge_id)` prevented that, so it needs to be scoped to exclude duplicates.

## Description

- Added `original_fee_id` on fees, pointing to the root fee in a void/regenerate chain.
- Narrowed the pay-in-advance duplication-guard indexes with `original_fee_id IS NULL` so only root fees enforce uniqueness and duplicated fees are exempt.

### DIFF
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -360,6 +360,7 @@ end
 #  invoice_id                          :uuid
 #  invoiceable_id                      :uuid
 #  organization_id                     :uuid             not null
+#  original_fee_id                     :uuid
 #  pay_in_advance_event_id             :uuid
 #  pay_in_advance_event_transaction_id :string
 #  subscription_id                     :uuid
@@ -367,8 +368,8 @@ end
 #
 # Indexes
 #
-#  idx_pay_in_advance_duplication_guard_charge         (pay_in_advance_event_transaction_id,charge_id) UNIQUE WHERE ((deleted_at IS NULL) AND (charge_filter_id IS NULL) AND (pay_in_advance_event_transaction_id IS NOT NULL) AND (pay_in_advance = true) AND (duplicated_in_advance = false))
-#  idx_pay_in_advance_duplication_guard_charge_filter  (pay_in_advance_event_transaction_id,charge_id,charge_filter_id) UNIQUE WHERE ((deleted_at IS NULL) AND (charge_filter_id IS NOT NULL) AND (pay_in_advance_event_transaction_id IS NOT NULL) AND (pay_in_advance = true) AND (duplicated_in_advance = false))
+#  idx_pay_in_advance_duplication_guard_charge         (pay_in_advance_event_transaction_id,charge_id) UNIQUE WHERE ((deleted_at IS NULL) AND (charge_filter_id IS NULL) AND (pay_in_advance_event_transaction_id IS NOT NULL) AND (pay_in_advance = true) AND (duplicated_in_advance = false) AND (original_fee_id IS NULL))
+#  idx_pay_in_advance_duplication_guard_charge_filter  (pay_in_advance_event_transaction_id,charge_id,charge_filter_id) UNIQUE WHERE ((deleted_at IS NULL) AND (charge_filter_id IS NOT NULL) AND (pay_in_advance_event_transaction_id IS NOT NULL) AND (pay_in_advance = true) AND (duplicated_in_advance = false) AND (original_fee_id IS NULL))
 #  index_fees_on_add_on_id                             (add_on_id)
 #  index_fees_on_applied_add_on_id                     (applied_add_on_id)
 #  index_fees_on_billing_entity_id                     (billing_entity_id)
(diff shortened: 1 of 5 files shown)


annvelents (approved):
LGTM!

app/models/fee.rb
```diff
@@ -360,15 +360,16 @@ def to_date
 #  invoice_id                          :uuid
 #  invoiceable_id                      :uuid
 #  organization_id                     :uuid             not null
+#  original_fee_id                     :uuid
```
rsempe, line 363: Not sure about the naming `original_fee_id`. It's vague for me.
What about `regenerated_from_fee_id` to be explicit?

db/migrate/20260421103557_update_pay_in_advance_duplication_guard_indexes.rb
```diff
@@ -0,0 +1,67 @@
...
+    add_index :fees,
+      [:pay_in_advance_event_transaction_id, :charge_id, :charge_filter_id],
+      unique: true,
+      name: CHARGE_FILTER_INDEX_NEW,
+      where: "deleted_at IS NULL AND charge_filter_id IS NOT NULL AND pay_in_advance_event_transaction_id IS NOT NULL AND pay_in_advance = true AND duplicated_in_advance = false AND original_fee_id IS NULL",
+      algorithm: :concurrently,
+      if_not_exists: true
+
+    remove_index :fees, name: CHARGE_INDEX, if_exists: true
+    remove_index :fees, name: CHARGE_FILTER_INDEX, if_exists: true
```
rsempe: why not using `algorithm: :concurrently` here as well to avoid locking the table?

db/migrate/20260421103557_update_pay_in_advance_duplication_guard_indexes.rb
```diff
@@ -0,0 +1,67 @@
...
+    add_index :fees,
+      [:pay_in_advance_event_transaction_id, :charge_id, :charge_filter_id],
+      unique: true,
+      name: CHARGE_FILTER_INDEX_NEW,
+      where: "deleted_at IS NULL AND charge_filter_id IS NOT NULL AND pay_in_advance_event_transaction_id IS NOT NULL AND pay_in_advance = true AND duplicated_in_advance = false",
+      algorithm: :concurrently,
+      if_not_exists: true
+
+    remove_index :fees, name: CHARGE_INDEX, if_exists: true
+    remove_index :fees, name: CHARGE_FILTER_INDEX, if_exists: true
```
rsempe: same here

db/migrate/20260421013319_add_original_fee_id_to_fees.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddOriginalFeeIdToFees < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :fees, :original_fee, type: :uuid, index: {algorithm: :concurrently}
+    add_foreign_key :fees, :fees, column: :original_fee_id, validate: false
```
rsempe, line 8: We probably need to add another migration to put the validation here, no?

`validate_foreign_key :fees, column: :original_fee_id`

rsempe (changes requested):

app/models/fee.rb
rsempe: maybe adding the `belongs_to` as well?

D1353L: The `original_fee_id` is the root fee in the regenerate flow with multiple void-regenerate steps. `regenerated_from_fee_id` would mean the direct parent fee.

> maybe adding the belongs_to as well?

This is done in #5330 with a descriptive comment
```
belongs_to :original_fee, class_name: "Fee", optional: true # Points to the root fee in a void/regenerate chain
```

db/migrate/20260421103557_update_pay_in_advance_duplication_guard_indexes.rb
D1353L: Good point, I'll add this

db/migrate/20260421013319_add_original_fee_id_to_fees.rb
D1353L: Definitely true, will add this as well

app/models/fee.rb
rsempe: > The original_fee_id is the root fee in the regenerate flow with multiple void-regenerate steps. regenerated_from_fee_id would mean the direct parent fee.

oooh ok! maybe `regeneration_root_fee_id` or just `original_fee_id`? 😅

D1353L: Let's keep the `original_fee_id`😉 I think the comment on the relation should be enough

rsempe (approved):

PR #5388, opened 2026-04-22
fix(events): Rescue ArgumentError in datetime filter parsing

toommz:
## Description

Extend the rescue clause to catch `ArgumentError` (which is a superset of `Date::Error`) so those inputs surface as a proper `invalid_date` validation failure. The helper is shared, so this auto-fixes every query using it: events, fees, invoices, credit notes, activity logs, security logs.

Regression test added on `EventsQuery` using a 200-char input that specifically exercises the previously unhandled `ArgumentError` path (verified it fails on the old rescue and passes on the new one).

### DIFF
--- a/app/queries/base_query.rb
+++ b/app/queries/base_query.rb
@@ -53,6 +53,9 @@ class BaseQuery < BaseService
     return value if [Time, ActiveSupport::TimeWithZone, Date, DateTime].include?(value.class)
 
     DateTime.iso8601(value)
+  # Date::Error inherits from ArgumentError so it is caught here too.
+  # A bare ArgumentError is raised e.g. for strings longer than 128 chars
+  # ("string length exceeds the limit 128").
   rescue ArgumentError
     result.single_validation_failure!(field: field_name.to_sym, error_code: "invalid_date")
       .raise_if_error!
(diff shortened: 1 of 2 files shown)


app/queries/base_query.rb
```diff
@@ -53,7 +53,7 @@ def parse_datetime_filter(field_name)
     return value if [Time, ActiveSupport::TimeWithZone, Date, DateTime].include?(value.class)
 
     DateTime.iso8601(value)
-  rescue Date::Error
```
groyoh, line 56: Please add a comment to mention that `Date::Error` inherits from `ArgumentError` and is also caught here, and in which case a bare `ArgumentError` is raised.

spec/queries/events_query_spec.rb
```diff
@@ -76,6 +76,20 @@
...
         expect(result.events.count).to eq(1)
       end
+
+      context "when a timestamp value raises ArgumentError during iso8601 parsing" do
+        let(:filters) { {timestamp_from: "1" * 200} }
+
+        it "returns a validation failure instead of raising" do
+          expect { events_query.call }.not_to raise_error
+
+          result = events_query.call
```
groyoh:
```suggestion
          result = nil
          expect { result = events_query.call }.not_to raise_error
```

groyoh (approved):

spec/queries/events_query_spec.rb
toommz: ffb7074ff060790660be82edbf148e81e7d0dd32

app/queries/base_query.rb
toommz: ffb7074ff060790660be82edbf148e81e7d0dd32

PR #5389, opened 2026-04-22
fix(coupons): Backfill invalid coupon frequency duration

D1353L:
## Context
Related to #5377
As the new validation was added, it is necessary to backfill existing records with correct values.

## Description

Added backfill migration for `frequency_duration` and `frequency_duration_remaining`


D1353L:
We decided not to merge this as it's overly cautious

PR #5392, opened 2026-04-22
feat(usage): Add filter_by_metric_code to current usage endpoint

murparreira:
## Context

 When a plan has multiple charges sharing the same billable metric, Lago auto-suffixes charge codes to enforce uniqueness (e.g. `advanced_support_plan` becomes `advanced_support_plan_2`). The `current_usage` endpoint already supports `filter_by_charge_code`, but that requires customers to first fetch the plan and inspect the suffixed codes, which aren't stable or human-readable. The `billable_metric.code` is already surfaced in the response, but there was no way to use it as a filter directly.

## Description

 Adds `filter_by_metric_code` as a new query parameter to `GET /customers/:id/current_usage`. When passed, it filters charges by `billable_metrics.code` instead of `charges.code`, letting customers use the stable metric code they already see in the response.  The join on `billable_metrics` is already present in the query, so no extra join is needed. When multiple charges share the same metric (the core use case), all of them are returned. A missing or foreign metric code returns a `404`, consistent with the existing `filter_by_charge_code` behaviour. The `full_usage` flag also works with this new filter since it routes through the same `has_charge_filter?` check.

### DIFF
--- a/spec/requests/api/v1/customers/usage_controller_spec.rb
+++ b/spec/requests/api/v1/customers/usage_controller_spec.rb
@@ -232,6 +232,67 @@ RSpec.describe Api::V1::Customers::UsageController do
         end
       end
 
+      context "when filter_by_metric_code is provided" do
+        let(:params) { {external_subscription_id: subscription.external_id, filter_by_metric_code: metric_a.code, apply_taxes: false} }
+
+        it "returns only charges for the given billable metric" do
+          subject
+
+          expect(response).to have_http_status(:success)
+          expect(json[:customer_usage][:charges_usage].count).to eq(1)
+
+          charge_usage = json[:customer_usage][:charges_usage].first
+          expect(charge_usage[:billable_metric][:code]).to eq(metric_a.code)
+          expect(charge_usage[:units]).to eq("14.0")
+          expect(charge_usage[:amount_cents]).to eq(14_000)
+        end
+      end
+
+      context "when filter_by_metric_code matches multiple charges sharing the same metric" do
+        let(:charge_a2) do
+          create(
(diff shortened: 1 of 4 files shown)


spec/requests/api/v1/customers/usage_controller_spec.rb
```diff
@@ -232,6 +232,67 @@
...
+      context "when filter_by_metric_code belongs to another organization" do
+        let(:other_metric) { create(:billable_metric) }
+        let(:params) { {external_subscription_id: subscription.external_id, filter_by_metric_code: other_metric.code} }
+
+        it "returns not found" do
+          subject
+
+          expect(response).to be_not_found_error("charge")
+        end
+      end
```
groyoh, line 297: Add a `full_usage: true` + `filter_by_metric_code` context mirroring the existing `filter_by_charge_code` coverage at lines 307–345.

spec/requests/api/v1/customers/usage_controller_spec.rb
```diff
@@ -232,6 +232,67 @@
...
+
+        before { charge_a2 }
+
+        it "returns all charges for the metric" do
+          subject
+
+          expect(response).to have_http_status(:success)
+          expect(json[:customer_usage][:charges_usage].count).to eq(2)
+          expect(json[:customer_usage][:charges_usage].map { |c| c[:billable_metric][:code] }).to all(eq(metric_a.code))
+        end
```
groyoh, line 275: Could you maybe verify the amount for the two charges to answer we don't have regression when there's multiple charges ?

groyoh (approved):

PR #5396, opened 2026-04-23
Feat/cs admin panel

annvelents:
## Context

Implement an admin panel

Note: admin panel is protected by license and user email pattern
also, users need to have cs_admin flag to access the panel


app/services/admin/organizations/update_service.rb
```diff
@@ -9,7 +9,7 @@ def initialize(organization:, params:)
         @organization = organization
         @params = params
 
-        super
+        super()
```
vincent-pochet: Can you maybe add a small comment explaining this change?

vincent-pochet (approved):

app/services/admin/organizations/update_service.rb
annvelents: should not be here 🕵️ 
removed, thank you!

PR #5397, opened 2026-04-23
fix(stripe): handle non-card payment methods in FetchDefaultPaymentMe…

mariohd:
## Summary                                                                                                                                                                                                         
                                                                                                                                                                                                                     
  - Fix `NoMethodError` in `FetchDefaultPaymentMethodService` when the Stripe payment method type is not `card` (e.g. `link`)                                                                                      
  - Branch on `pm.type == "card"` so card attributes (`last4`, `brand`, `exp_month`, `exp_year`) are only accessed for card payment methods                                                                          
  - For non-card types, store only `{type: pm.type}` instead of a hash full of nil card fields                                             
  - Add test scenario with a realistic Stripe `link` payment method response                                                                                                                                         
                                                                                                                                                                                                                     
  ## Test plan                                                                                                                                                                                                       
                                                                                                                                                                                                                     
  - [x] Card payment method: creates payment method with full card details (last4, brand, expiration)                                                                                                                
  - [x] Non-card payment method (link): creates payment method with only `{type: "link"}`                                                                                                                            
  - [x] No provider_customer_id: returns without creating payment method                 
  - [x] No payment method found on Stripe: returns without creating payment method                                                                                                                                   
  - [x] Existing payment method: returns existing without creating duplicate

### DIFF
--- a/app/services/payment_providers/stripe/customers/fetch_default_payment_method_service.rb
+++ b/app/services/payment_providers/stripe/customers/fetch_default_payment_method_service.rb
@@ -48,10 +48,10 @@ module PaymentProviders
 
           PaymentMethods::CardDetails.new(
             type: pm.type,
-            last4: pm.card&.last4,
-            brand: pm.card&.display_brand,
-            expiration_month: pm.card&.exp_month,
-            expiration_year: pm.card&.exp_year,
+            last4: pm.try(:card)&.last4,
+            brand: pm.try(:card)&.display_brand,
+            expiration_month: pm.try(:card)&.exp_month,
+            expiration_year: pm.try(:card)&.exp_year,
             card_holder_name: nil,
             issuer: nil
           ).to_h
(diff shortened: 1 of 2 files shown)


app/services/payment_providers/stripe/customers/fetch_default_payment_method_service.rb
```diff
@@ -48,10 +48,10 @@ def payment_method_details(payment_method_id:)
 
           PaymentMethods::CardDetails.new(
```
mariohd: I'm wondering if we need to have this `CardDetails` hash for `link` paymentMethod.

In case for `link`, maybe having a `{ type: pm.type }` would be enough? @domenicofalco

domenicofalco: @mariohd yes, having `link` only on `{ type: pm.type }` as response, it won't be breaking changes for the frontend. so feel free to apply this change specifically for `link`. And also the UI, is kinda prepared to accomodate `type` only

mariohd: Awesome, thanks for checking this! It makes things easier not storing "useless" data.

lovrocolic (approved):

toommz (approved):
You might have to rebase on master for the CI to pass 🤔

aquinofb (approved):
LGTM!

PR #5398, opened 2026-04-23
fix(stripe-webhooks): Stripe webhook jobs errors for parallel processing

osmarluz:
## Context

Concurrent writers for the same `payment_intent.succeeded` (parallel webhook workers, or `PaymentProviders::Stripe::Payments::CreateService` racing with the webhook) both try to insert a Payment. The losing worker's `save!` hits the unique index on `(provider_payment_id, payment_provider_id)` and raises `ActiveRecord::RecordNotUnique`. The existing rescue swallowed the error but had already assigned the unsaved `Payment.new` to `result.payment`, so `PaymentIntentSucceededService` enqueued `SetPaymentMethodAndCreateReceiptJob` with it and crashed with `ActiveJob::SerializationError`.

Simply returning an empty result on rescue would silence the error but skip the enqueue entirely — leaving the persisted Payment without `provider_payment_method_id` / `provider_payment_method_data`, since `SetPaymentMethodAndCreateReceiptJob` is the only path that sets them.

## Description

- Set `result.payment` / `result.payable` only after `payment.save!` succeeds.
- In the `RecordNotUnique` rescue, re-fetch the Payment by `provider_payment_id` and assign the persisted row, so the caller still enqueues `SetPaymentMethodAndCreateReceiptJob` against it (the job is idempotent).
- Add a regression spec that lets the real Postgres unique index raise `RecordNotUnique` and asserts the rescue's re-fetch returns the persisted Payment.

### DIFF
--- a/app/services/payment_requests/payments/stripe_service.rb
+++ b/app/services/payment_requests/payments/stripe_service.rb
@@ -71,6 +71,14 @@ module PaymentRequests
       rescue ActiveRecord::RecordInvalid => e
         result.record_validation_failure!(record: e.record)
       rescue ActiveRecord::RecordNotUnique
+        # NOTE: Another writer (a parallel webhook worker, or PaymentProviders::Stripe::Payments::CreateService)
+        #       committed the Payment first. Hand the persisted row back so the
+        #       caller can still enqueue downstream side effects (e.g. SetPaymentMethodAndCreateReceiptJob).
+        payment = Payment.find_by(provider_payment_id: stripe_payment.id)
+        if payment
+          result.payment = payment
+          result.payable = payment.payable
+        end
         result
       rescue BaseService::FailedResult => e
         result.fail_with_error!(e)
--- a/app/services/payment_requests/payments/stripe_service.rb
+++ b/app/services/payment_requests/payments/stripe_service.rb
@@ -50,15 +50,15 @@ module PaymentRequests
           return result
         end
 
-        result.payment = payment
-        result.payable = payment.payable
-
         processing = status == "processing"
         payment.status = status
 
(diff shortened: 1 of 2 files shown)


annvelents (approved):
Sounds good. So we can delete the jobs, right?

brunomiguelpinto (approved):

app/services/payment_requests/payments/stripe_service.rb
```diff
@@ -71,6 +71,14 @@ def update_payment_status(organization_id:, status:, stripe_payment:)
       rescue ActiveRecord::RecordInvalid => e
         result.record_validation_failure!(record: e.record)
       rescue ActiveRecord::RecordNotUnique
+        # NOTE: Another writer (a parallel webhook worker, or PaymentProviders::Stripe::Payments::CreateService)
+        #       committed the Payment first. Return the persisted row so the
+        #       caller can still enqueue downstream side effects (e.g. SetPaymentMethodAndCreateReceiptJob).
```
annvelents, line 76: aaaaah... nice!

annvelents: but
in case it happens, do the side effects need to be triggered?

annvelents (approved):

app/services/payment_requests/payments/stripe_service.rb
annvelents: talked with @osmarluz, it can be `PaymentIntentSucceededService`, which creates a payment if not found, and then runs `SetPaymentMethodAndCreateReceiptJob`, which will not create the receipt if we do not return the payment in the previous job

PR #5399, opened 2026-04-23
feat(graphql): expose downgrade_plan_date and previous relations on Subscription

ansmonjol:
## Context

Related ticket: [ISSUE-1845 — Wrong downgrade date displayed when downgrade is pending](https://linear.app/getlago/issue/ISSUE-1845/wrong-downgrade-date-displayed-when-downgrade-is-pending)

This is **not** a bug fix in the API — the REST serializer already returns the correct `downgrade_plan_date`. It is a schema change that **exposes the attributes the front-end needs to render the correct messaging** on the subscription details page. The actual UI fix lives in the front-end PR.

## Problem

On a subscription with a pending downgrade, the front-end was rendering "Downgrading to X on DATE" from `nextSubscriptionAt`. The resolver for that field returns `next_subscription&.started_at || next_subscription&.subscription_at`, i.e. the pending sub's scheduling date, not the date the downgrade actually takes effect. So users saw e.g. `April 22, 2026` instead of `May 22, 2026`.

The correct value (`end of current billing period + 1.day`) already exists on `Subscription#downgrade_plan_date` and is shipped via the REST serializer, but had no GraphQL entry point.

## Changes

`Types::Subscriptions::Object` gains three fields:

- **`downgrade_plan_date`** (`ISO8601Date`) — delegates to the existing model method. Returns the effective downgrade date when a pending downgrade exists, otherwise `nil`.
- **`previous_plan`** (`Plan`) — shortcut mirroring the existing `next_plan`, resolves to `previous_subscription&.plan`. Used by the FE to render "Downgrading from [plan name]" on the pending sub view.
- **`previous_subscription`** (`Subscription`) — the full relation, so callers that need the previous sub's own fields (e.g. its `downgrade_plan_date`, which is the pending sub's activation date) can read them directly instead of overloading `downgrade_plan_date` with dual semantics.

`schema.graphql` / `schema.json` are regenerated (required by the `LagoApiSchema` matcher spec and by front-end codegen).

## Companion PR

Front-end: https://github.com/getlago/lago-front/pull/new/issue-1845 (linked once opened)

### DIFF
--- a/app/graphql/types/subscriptions/object.rb
+++ b/app/graphql/types/subscriptions/object.rb
@@ -32,11 +32,14 @@ module Types
       field :created_at, GraphQL::Types::ISO8601DateTime, null: false
       field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
 
+      field :downgrade_plan_date, GraphQL::Types::ISO8601Date
       field :next_name, String, null: true
       field :next_plan, Types::Plans::Object
       field :next_subscription, Types::Subscriptions::Object
       field :next_subscription_at, GraphQL::Types::ISO8601DateTime
       field :next_subscription_type, Types::Subscriptions::NextSubscriptionTypeEnum
+      field :previous_plan, Types::Plans::Object
+      field :previous_subscription, Types::Subscriptions::Object
 
       field :activity_logs, [Types::ActivityLogs::Object], null: true
       field :charges, [Types::Charges::Object], null: true
@@ -57,6 +60,10 @@ module Types
 
       def next_plan
         object.next_subscription&.plan
+      end
+
+      def previous_plan
+        object.previous_subscription&.plan
       end
 
       def next_name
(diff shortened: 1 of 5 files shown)


ancorcruz (approved):
Looks good — clean change, mirrors the existing next_* pattern, and the tests cover the key scenarios well.

One thing worth flagging (not blocking):

N+1 on previous_subscription / previous_plan in list contexts: The REST API already eager-loads these associations in SubscriptionIndex (.includes(:plan, previous_subscription: :plan, ...)), but the GraphQL resolvers don't. Specifically:

- Customers::SubscriptionsResolver already has a pre-existing N+1 at line 24 — it calls s.previous_subscription&.downgraded? on every subscription without eager-loading. This PR doesn't make that worse, but if the FE starts requesting previousPlan / previousSubscription in the same query it'll compound.
- SubscriptionsResolver (root query) also has no includes, so paginated lists would fire separate queries per item for these fields.

Not caused by this PR, but worth a follow-up to bring the GraphQL resolvers in line with the REST API's eager-loading.

PR #5408, opened 2026-04-25
fix(wallets): retry RefreshWalletJob on transient Nango/Anrok errors

murparreira:
## Context

`Customers::RefreshWalletJob` was dying permanently on any transient Nango/Anrok error that wasn't one of the two patterns already hard-coded in its rescue block (`function_runtime_out_of_memory` and `customerAddressCouldNotResolve`).
Transient conditions, Nango task-queue delays, orchestrator POST failures, response-body rate limits, all fell through to `result.service_failure!`, got wrapped as `validation_failure!(errors: {tax_error: [...]})` by `CustomerUsageService`, and re-raised as `BaseService::ValidationFailure`, which has no `retry_on` in the job. The 12-hour uniqueness lock then blocked re-enqueue until expiry.
We were seing a lot of occurrences on observability channels, with bursts of ~10 dead jobs per 15-second window during Anrok rate-limit spikes.

## Description

Introduced five new retryable error classes under `Integrations::Aggregator` (`TaskInProgressError`, `TaskExpiredError`, `OrchestratorFailureError`, `ServerContentionError`, `TimeoutError`) and added a canonical `retryable_errors` class method on `BaseService` that returns the full list.

Transient conditions are now raised before reaching `service_failure!`:
- `action_script_failure` HTTP errors (task in progress / expired / orchestrator failure) are detected via new predicates in `Aggregator::BaseService` and raised from the HTTP rescue block of all four tax invoice services.
- Response-body rate limits and resource contention are detected in `process_response` and `process_void_response` and raised as `ServerContentionError`.
- Network timeouts (`Net::ReadTimeout`, `Net::OpenTimeout`, `OpenSSL::SSL::SSLError`) are caught in the same rescue block and re-raised as `TimeoutError`.

`RefreshWalletJob` now uses `retry_on *Integrations::Aggregator::BaseService.retryable_errors` instead of the previous ad-hoc list, and its rescue block is simplified to only handle the one non-transient case (`customerAddressCouldNotResolve`). `PreviewService` uses the same list to fall back to zero-tax on any aggregator transient, replacing the previous explicit list of network error types.

### DIFF
--- a/app/services/integrations/aggregator/base_service.rb
+++ b/app/services/integrations/aggregator/base_service.rb
@@ -9,6 +9,14 @@ module Integrations
       REQUEST_LIMIT_ERROR_CODE = "SSS_REQUEST_LIMIT_EXCEEDED"
       BAD_GATEWAY_ERROR = "502 Bad Gateway"
 
+      def self.retryable_errors
+        [
+          BadGatewayError, RequestLimitError, OutOfMemoryError,
+          TaskInProgressError, TaskExpiredError, OrchestratorFailureError,
+          ServerContentionError, TimeoutError
+        ]
+      end
+
       def initialize(integration:, options: {})
         @integration = integration
         @options = options
@@ -177,6 +185,18 @@ module Integrations
       def bad_gateway_error?(http_error)
         http_error.error_code.to_s == "502" ||
           http_error.error_body.include?(BAD_GATEWAY_ERROR)
+      end
+
+      def task_in_progress_error?(http_error)
+        code(http_error) == "action_script_failure" && message(http_error).include?("is in progress")
+      end
+
+      def task_expired_error?(http_error)
+        code(http_error) == "action_script_failure" && message(http_error).include?("expired")
+      end
+
+      def orchestrator_failure_error?(http_error)
+        code(http_error) == "action_script_failure" && message(http_error).include?("/v1/immediate failed")
       end
 
       def parse_response(response)
--- a/app/services/integrations/aggregator/base_service.rb
+++ b/app/services/integrations/aggregator/base_service.rb
@@ -8,6 +8,9 @@ module Integrations
       BASE_URL = "https://api.nango.dev/"
       REQUEST_LIMIT_ERROR_CODE = "SSS_REQUEST_LIMIT_EXCEEDED"
       BAD_GATEWAY_ERROR = "502 Bad Gateway"
+      TASK_IN_PROGRESS_PATTERN = /\bTask\b.*\bis in progress\b/.freeze
+      TASK_EXPIRED_PATTERN = /\bTask\b.*\bexpired\b/.freeze
+      ORCHESTRATOR_FAILURE_PATTERN = %r{POST https?://nango-orchestrator-svc\.nango/v1/immediate failed}.freeze
 
       def self.retryable_errors
         [
@@ -188,15 +191,15 @@ module Integrations
       end
 
       def task_in_progress_error?(http_error)
-        code(http_error) == "action_script_failure" && message(http_error).include?("is in progress")
+        code(http_error) == "action_script_failure" && TASK_IN_PROGRESS_PATTERN.match?(message(http_error))
       end
 
       def task_expired_error?(http_error)
-        code(http_error) == "action_script_failure" && message(http_error).include?("expired")
+        code(http_error) == "action_script_failure" && TASK_EXPIRED_PATTERN.match?(message(http_error))
       end
 
       def orchestrator_failure_error?(http_error)
-        code(http_error) == "action_script_failure" && message(http_error).include?("/v1/immediate failed")
+        code(http_error) == "action_script_failure" && ORCHESTRATOR_FAILURE_PATTERN.match?(message(http_error))
       end
 
       def parse_response(response)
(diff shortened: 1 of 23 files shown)


groyoh (commented):
@murparreira I believe we're lacking quite a few tests on this one.

Also @annvelents @mikeh-lago I believe we'll want/need some review on both your squads as well.

mikeh-lago:
> @murparreira I believe we're lacking quite a few tests on this one.
> 
> Also @annvelents @mikeh-lago I believe we'll want/need some review on both your squads as well.

@osmarluz Nango is a hot topic these days !

aquinofb (approved):
LGTM!

mariohd (approved):
Seems good

app/services/integrations/aggregator/base_service.rb
```diff
@@ -179,6 +187,18 @@ def bad_gateway_error?(http_error)
           http_error.error_body.include?(BAD_GATEWAY_ERROR)
       end
 
+      def task_in_progress_error?(http_error)
+        code(http_error) == "action_script_failure" && message(http_error).include?("is in progress")
+      end
+
+      def task_expired_error?(http_error)
+        code(http_error) == "action_script_failure" && message(http_error).include?("expired")
+      end
```
groyoh, line 199: ⚠️ `task_expired_error?` fires on any message containing the substring `"expired"`, but Nango could wrap multiple issues with `expired` in the message. Please check the actual errors we've come across and use a proper regex.

Same for the other errors.

Furthermore, the tests should rely on this exact error and not some generic ones.

app/services/integrations/aggregator/taxes/base_service.rb
```diff
@@ -109,6 +107,10 @@ def tax_breakdown(breakdown, taxes_to_pay)
           end
         end
 
+        def server_contention_error?(message)
+          message.include?("API limit") || message.include?("resource contention")
```
tiagolupepic, line 111: do we should move these strings to a constant? just to keep consistent

app/services/integrations/aggregator/taxes/base_service.rb
```diff
@@ -61,13 +62,10 @@ def process_response(body)
...
             code, message = retrieve_error_details(body["failedInvoices"].first["validation_errors"])
 
-            # Temp fix for the API limit issue
-            message = "API limit" if message == "Internal server error: resource contention."
-
-            unless message.include?("API limit")
-              deliver_tax_error_webhook(customer:, code:, message:) if customer.persisted? # Do not send this webhook in preview mode
-            end
+            raise Integrations::Aggregator::OutOfMemoryError if message.include?(OUT_OF_MEMORY_ERROR)
+            raise Integrations::Aggregator::ServerContentionError, message if server_contention_error?(message)
```
groyoh, line 66: I'm not sure we have proper tests for those.

spec/jobs/customers/refresh_wallet_job_spec.rb
```diff
@@ -97,6 +99,76 @@
...
+          before do
+            allow(Customers::RefreshWalletsService).to receive(:call).with(customer:).and_raise(Integrations::Aggregator::TimeoutError)
+          end
+
+          it "raises the error and retries the job" do
+            assert_performed_jobs(6, only: [described_class]) do
+              expect do
+                described_class.perform_later(customer)
+              end.to raise_error(Integrations::Aggregator::TimeoutError)
+            end
```
groyoh: We can use parameterized tests here to DRY this code a bit.

spec/services/integrations/aggregator/taxes/invoices/void_service_spec.rb
```diff
@@ -169,6 +169,45 @@
...
+        let(:body) { {error: {code: "action_script_failure", message: "Task has expired"}}.to_json }
+
+        it "raises TaskExpiredError" do
+          expect { service_call }.to raise_error(Integrations::Aggregator::TaskExpiredError)
+        end
+      end
+
+      context "when it is an orchestrator failure error" do
+        let(:error_code) { 500 }
+        let(:body) { {error: {code: "action_script_failure", message: "Call to /v1/immediate failed"}}.to_json }
```
groyoh: As per my previous command, this is actually:
```suggestion
        let(:body) { {error: {code: "action_script_failure", message: "POST http://nango-orchestrator-svc.nango/v1/immediate failed"}}.to_json }
```

Please double check the actual errors from Sentry.

groyoh (commented):
@murparreira please do a proper review of the code coverage. It seems that there might still be plenty of untested cases (timeouts, bad gateway, etc.). 

I believe you could DRY some of those code as there are the same across multiple services.

app/services/integrations/aggregator/taxes/base_service.rb
groyoh: Tiago's earlier comment about extracting these to constants doesn't look resolved. Other sibling matchers (`BAD_GATEWAY_ERROR`, `REQUEST_LIMIT_ERROR_CODE`, `OUT_OF_MEMORY_ERROR`, and the new `TASK_*_PATTERN`/`ORCHESTRATOR_FAILURE_PATTERN`) all live as constants, so the two literals here are the odd ones out. Promoting them (e.g. `API_LIMIT_ERROR` / `RESOURCE_CONTENTION_ERROR`) keeps the matching surface in one place and matches the rest of the file.

groyoh (approved):
I would suggest you document the actual errors which can be produce by Nango in the code. Could be useful for the future.

PR #5409, opened 2026-04-27
fix(payment_receipts): retry mailer when PDF files are missing

brunomiguelpinto:
## Context

`PaymentReceiptMailer` was crashing in production with:

```
ActionView::Template::Error: No route matches {}
```

`SendEmailJob`: the receipt email was being rendered before the linked invoice PDFs had finished generating. The Slim template at `app/views/payment_receipt_mailer/created.slim:41` calls `link_to invoice.number, invoice.file_url`, and `Invoice#file_url` returns `nil` when the ActiveStorage attachment isn't present yet — that nil flows into `url_for({})`, which has no route to match in a mailer context (no request).

The receipt PDF is generated synchronously by `PaymentReceiptMailer#ensure_pdf`, but the *invoice* PDFs are produced by `Invoices::GeneratePdfService` from a separate async job. When the receipt mailer runs before that job completes, the template blows up. Manually retrying the same Sidekiq job succeeded — confirming the fix is to defer rendering until the files exist.

Closes [ISSUE-1832](https://linear.app/getlago/issue/ISSUE-1832/add-guard-and-retry-for-missing-receipt-files)

## Description

- New `PaymentReceipts::FilesNotReadyError` defined in `app/services/payment_receipts/files_not_ready_error.rb`, following the existing convention used by `Invoices::Payments::ConnectionError` / `RateLimitError`.
- `PaymentReceiptMailer#ensure_pdf` now raises this error when either the receipt's own PDF or any linked invoice PDF is not yet attached. Multi-invoice `PaymentRequest` payables are covered too.
- `SendEmailJob` registers `retry_on PaymentReceipts::FilesNotReadyError, wait: :polynomially_longer, attempts: 8`. Polynomial backoff gives the async invoice-PDF jobs roughly an hour to catch up before the email is discarded.
- customers either get a complete email with working PDF links, or no email at all — never a broken one.

### DIFF
--- a/app/jobs/send_email_job.rb
+++ b/app/jobs/send_email_job.rb
@@ -37,6 +37,7 @@ class SendEmailJob < ActionMailer::MailDeliveryJob
   retry_on LagoHttpClient::HttpError, wait: :polynomially_longer, attempts: 6
   retry_on Net::ReadTimeout, wait: :polynomially_longer, attempts: 6
   retry_on Net::SMTPServerBusy, wait: :polynomially_longer, attempts: 25
+  retry_on PaymentReceipts::FilesNotReadyError, wait: :polynomially_longer, attempts: 8
 
   after_discard { |job, error| job.log(error:) }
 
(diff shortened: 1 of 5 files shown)


annvelents (approved):
🚀

PR #5410, opened 2026-04-27
feat(presentation_group_keys): Update graphql resolvers and mutations

tiagolupepic:
This commit updates the graphql resolvers to include the presentationBreakdowns.

Also, it's changing some mutations like "create plan" and "update plan" to receive the charge and chargeFilter properties containing presentation_group_keys.


annvelents (approved):
LGTM

PR #5411, opened 2026-04-27
feat(presentation_group_keys): Update serializers to include presentation breakdowns

tiagolupepic:
This commit updates the serializers of usage, fee and invoices to include the breakdowns.

We're including all presentation breakdowns associated to fees in invoice, in order to do that, this commit 
introduces a presentation builder, this class is responsible to group the breakdowns based on `presentation_by`. 
We avoid displaying the same `presentation_by` twice and instead we're grouping and sum them together.

Serializers are being updated:
- Charge ( Usage / Projected )
- Fee
- Invoice


spec/serializers/v1/customers/charge_usage_serializer_spec.rb
```diff
@@ -81,12 +84,67 @@
...
+          "filters" => [],
+          "presentation_breakdowns" => []
         }
       ]
     )
   end
 
+  context "when contains presentation breakdowns" do
+    let(:presentation_breakdowns) do
+      presentation_breakdown_class = Struct.new(:presentation_by, :units, keyword_init: true)
```
annvelents: Can we have a factory for presentation_breakdowns and use them here?

spec/serializers/v1/customers/charge_usage_serializer_spec.rb
```diff
@@ -81,12 +84,67 @@
...
+
+      [
+        presentation_breakdown_class.new(presentation_by: {"card_type" => "visa"}, units: "7"),
+        presentation_breakdown_class.new(presentation_by: {"card_type" => "visa"}, units: "1"),
+        presentation_breakdown_class.new(presentation_by: {"country" => "pt"}, units: "3")
+      ]
+    end
+
+    let(:usage) do
```
annvelents, line 121: I am not sure if I can find where this usage is used 👀

annvelents: found, nvm 😅

annvelents: hm, but I see no difference in this usage with usage from the outer scope. Do we need to redefine it?

spec/serializers/v1/customers/charge_usage_serializer_spec.rb
```diff
@@ -157,7 +217,8 @@
           grouped_by: {"card_type" => "visa"},
           charge_filter:,
           charge_filter_id: charge_filter.id,
-          pricing_unit_usage:
+          pricing_unit_usage:,
+          presentation_breakdowns: presentation_breakdowns
```
annvelents:
```suggestion
          presentation_breakdowns:
```

spec/serializers/v1/customers/charge_usage_serializer_spec.rb
```diff
@@ -44,7 +45,8 @@
         aggregation_type: billable_metric.aggregation_type,
         grouped_by: {"card_type" => "visa"},
         charge_filter: nil,
-        pricing_unit_usage:
+        pricing_unit_usage:,
+        presentation_breakdowns: presentation_breakdowns
```
annvelents:
```suggestion
        presentation_breakdowns:
```

spec/serializers/v1/customers/charge_usage_serializer_spec.rb
tiagolupepic: yes, I removed the Struct and replaced with FactoryBot presentation_breakdown

spec/serializers/v1/customers/charge_usage_serializer_spec.rb
tiagolupepic: I reviewed these specs, removed the duplicated usage and added a scenario where contains a usage with two fees, one when the grouped_by is empty and other with a value.

spec/serializers/v1/customers/projected_charge_usage_serializer_spec.rb
```diff
@@ -72,7 +73,8 @@
         aggregation_type: billable_metric.aggregation_type,
         grouped_by: {"card_type" => "visa"},
         charge_filter: nil,
-        pricing_unit_usage:
+        pricing_unit_usage:,
+        presentation_breakdowns: presentation_breakdowns
```
annvelents:
```suggestion
        presentation_breakdowns:
```

spec/serializers/v1/customers/projected_charge_usage_serializer_spec.rb
```diff
@@ -214,16 +255,17 @@
           grouped_by: {"card_type" => "visa"},
           charge_filter:,
           charge_filter_id: charge_filter.id,
-          pricing_unit_usage:
+          pricing_unit_usage:,
+          presentation_breakdowns: presentation_breakdowns
```
annvelents:
```suggestion
          presentation_breakdowns:
```

spec/serializers/v1/customers/projected_charge_usage_serializer_spec.rb
```diff
@@ -341,7 +383,8 @@
           grouped_by: nil,
           charge_filter: charge_filter_1,
           charge_filter_id: charge_filter_1.id,
-          pricing_unit_usage:
+          pricing_unit_usage:,
+          presentation_breakdowns: presentation_breakdowns
```
annvelents:
```suggestion
          presentation_breakdowns:
```

spec/serializers/v1/customers/projected_charge_usage_serializer_spec.rb
```diff
@@ -360,22 +403,23 @@
           grouped_by: nil,
           charge_filter: charge_filter_2,
           charge_filter_id: charge_filter_2.id,
-          pricing_unit_usage:
+          pricing_unit_usage:,
+          presentation_breakdowns: presentation_breakdowns
```
annvelents:
```suggestion
          presentation_breakdowns:
```

annvelents (approved):
LGTM, marked small nitpicks :)

spec/serializers/v1/customers/presentation_breakdown_builder_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+        ]
+      )
+    ]
+  end
+
+  it "returns one entry per breakdown with stringified units" do
+    expect(result).to eq([
+      {presentation_by: {"cloud" => "aws"}, units: "1.2"},
+      {presentation_by: {"cloud" => "gcp"}, units: "3"},
+      {presentation_by: {"cloud" => "aws"}, units: "0.3"}
```
rsempe, line 34: is it normal to have `aws` twice here? 🤔

app/serializers/v1/invoice_serializer.rb
```diff
@@ -93,6 +93,7 @@ def fees
             :customer,
             :charge,
             :billable_metric,
+            :presentation_breakdowns,
```
rsempe, line 96: do we have to include `:presentation_breakdowns` in past usage as well to avoid N+1 query?

app/serializers/v1/fee_serializer.rb
```diff
@@ -56,7 +56,8 @@ def serialize
         refunded_at: model.refunded_at&.iso8601,
         amount_details: model.amount_details,
         self_billed: model.invoice&.self_billed || false,
-        pricing_unit_details:
+        pricing_unit_details:,
+        presentation_breakdowns: V1::Customers::PresentationBreakdownBuilder.call([model], filter: V1::Customers::PresentationBreakdownBuilder::ALL)
```
rsempe: Not sure to see the real interest of having the `customer` namespace?

rsempe: Not really fan of having a `PresentationBreakdownBuilder`.
Why not using a serializer for this instead?

```
module V1
  class PresentationBreakdownSerializer < ModelSerializer
    def serialize                                                                                         
      {
        presentation_by: model.presentation_by,                                                           
        units: model.units.to_s                                      
      }
    end
  end
end
```

Then you can do `model.presentation_breakdowns.map { |b| V1::PresentationBreakdownSerializer.new(b).serialize }`

wdyt?

tiagolupepic: going to add the PresentationBreakdownSerializer as you suggested, but keep the `Customers:: PresentationBreakdownBuilder` to wrap the logic to filter or not the fees by group_by + call the serializer. I don't want to spread the logic of filtering fees by group_by in many places

app/serializers/v1/invoice_serializer.rb
tiagolupepic: good catch! I'll add the eager load there

spec/serializers/v1/customers/presentation_breakdown_builder_spec.rb
tiagolupepic: yes. I put in this way to make it clear the expected behaviour is not sum when have multiple fees with repeated presentation_by

app/serializers/v1/fee_serializer.rb
tiagolupepic: done!

tiagolupepic: addressed in another comment below

rsempe (approved):

PR #5412, opened 2026-04-27
feat(payment-gated-subs): fire deferred invoice side effects on payment success

ancorcruz:
## Context

When a subscription is gated by a payment activation rule, the invoice is created in :open status and the standard invoice.created side effects (webhook, activity log, document generation, accounting/CRM sync, analytics tracking) are deliberately skipped — outside actors should not see the invoice until payment has been collected.

## Description

Fire the deferred side effects in Payment::ResolveService#handle_success after FinalizeService transitions the invoice to :finalized and ResolveSubscriptionStatusService activates the subscription. Side effects are wrapped in an after_commit block so they only run if the surrounding transaction commits.

### DIFF
--- a/spec/services/subscriptions/activation_rules/payment/resolve_service_spec.rb
+++ b/spec/services/subscriptions/activation_rules/payment/resolve_service_spec.rb
@@ -99,10 +99,28 @@ RSpec.describe Subscriptions::ActivationRules::Payment::ResolveService do
       expect(Utils::ActivityLog).to have_produced("invoice.created").with(invoice)
     end
 
-    it "enqueues GenerateDocumentsJob" do
+    it "enqueues GenerateDocumentsJob with notify false" do
       result
 
       expect(Invoices::GenerateDocumentsJob).to have_been_enqueued.with(invoice:, notify: false)
+    end
+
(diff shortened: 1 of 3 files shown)


spec/services/subscriptions/activation_rules/payment/resolve_service_spec.rb
```diff
@@ -87,6 +87,72 @@
...
+      expect(SendWebhookJob).to have_been_enqueued.with("invoice.created", invoice)
+    end
+
+    it "produces an invoice.created activity log" do
+      result
+
+      expect(Utils::ActivityLog).to have_produced("invoice.created").with(invoice)
+    end
+
+    it "enqueues GenerateDocumentsJob" do
```
osmarluz: @ancorcruz very small thing here, but it might make sense to include the scenario where `should_deliver_email?` returns `true`

osmarluz (approved):
🚀

spec/services/subscriptions/activation_rules/payment/resolve_service_spec.rb
ancorcruz: added, thanks

PR #5414, opened 2026-04-27
  feat(rake): add receipes:events:delete task for soft-deleting PG ev…

mariohd:
## Summary                                                                                                                                                                                                       
                                                                                                                                                                                                                     
  - Add `receipes:events:delete` rake task to soft-delete PG events by organization and time range                                                                                                                 
  - Interactive flow: prompts for organization ID (validates and confirms by name), from/to timestamps, and final confirmation with event count
  - Restricted to organizations using PostgreSQL events store (rejects Clickhouse orgs)                                                        
  - Extract reusable `TaskPrompt` module (`ask`, `confirm!`, `ask_for_organization`, `ask_for_timestamp_range`) into `lib/tasks/rake_helpers.rb`                                                                     
                                                                                                                                                                                                                     
  ## Test plan                                                                                                                                                                                                       
                                                                                                                                                                                                                     
  - [x] Aborts when organization is not found                                                                                                                                                                        
  - [x] Aborts when user does not confirm the organization                                                                                                                                                         
  - [x] Aborts when organization uses Clickhouse events store
  - [x] Aborts when from_timestamp is after to_timestamp     
  - [x] Finishes gracefully when no events match the time range                                                                                                                                                      
  - [x] Soft-deletes only events within the time range when user confirms
  - [x] Aborts without deleting when user declines final confirmation

### DIFF
--- a/lib/tasks/receipes.rake
+++ b/lib/tasks/receipes.rake
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require_relative "../task_prompt"
+
+# rubocop:disable Rails/Output,Rails/Exit
+namespace :receipes do
+  namespace :events do
+    desc "Soft-delete PG events for an organization within a time range"
+    task delete: :environment do
+      organization = TaskPrompt.ask_for_organization
+      abort "This task only supports organizations using PostgreSQL events store." unless organization.postgres_events_store?
+
+      from_time, to_time = TaskPrompt.ask_for_timestamp_range
+
+      events = Event.where(organization_id: organization.id)
+        .from_datetime(from_time)
+        .to_datetime(to_time)
+
+      count = events.count
+
+      if count.zero?
+        puts "No events found in the given time range. Nothing to delete."
+        next
+      end
+
+      puts "\nThis will soft-delete #{count} events from \"#{organization.name}\" " \
+        "from #{from_time.utc} to #{to_time.utc} (inclusive)."
+      TaskPrompt.confirm!("Continue? (y/n): ")
+
+      events.update_all(deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
+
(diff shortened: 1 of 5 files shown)


lib/tasks/receipes/events.rake
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require_relative "../task_prompt"
+
+# rubocop:disable Rails/Output,Rails/Exit
+namespace :receipes do
+  namespace :events do
```
vincent-pochet, line 9: I like the `receipes` namespace,  but I would suggest to rename this file `events.rake` and the task itself `delete_in_range`.

It will probably open the door for a wider cleanup in the rake tasks as many of them are currently outdated.

vincent-pochet (approved):
Nice one with the `TaskPrompt` 👍

lib/tasks/receipes.rake
```diff
@@ -0,0 +1,36 @@
...
+      if count.zero?
+        puts "No events found in the given time range. Nothing to delete."
+        next
+      end
+
+      puts "\nThis will soft-delete #{count} events from \"#{organization.name}\" " \
+        "from #{from_time.utc} to #{to_time.utc} (inclusive)."
+      TaskPrompt.confirm!("Continue? (y/n): ")
+
+      events.update_all(deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
```
vincent-pochet: Since we can have a lot of events for a single organization, maybe we should batch the update. What do you think?

vincent-pochet (approved):

lib/tasks/receipes.rake
mariohd: Makes total sense!

vincent-pochet (approved):

PR #5415, opened 2026-04-27
feat(presentation_group_keys): Update DailyUsages to also include presentation breakdowns

tiagolupepic:
This commit changes the compute_diff_service to also make the diff when the presentation_breakdowns is available in charge_usage.


rsempe (approved):
👍

rsempe:
I would push for having all `DailyUsages::ComputeDiffService` tests  in `spec/services/daily_usages/compute_diff_service_spec.rb` instead to be consistent inside the codebase (and not having extracted spec inside `compute_diff_service_presentation_breakdowns_spec`) - but that's not a blocker ;)

PR #5420, opened 2026-04-27
fix(charges): Fix N+1 with charge_filter in Fees::ChargeService

tiagolupepic:
This commit fixes a N+1 when a fee calls the relation of charge_filter.

The issue is happening because the new created object in memory is getting the charge_filter.id. The new fee object doesn't know that charge_filter is already loaded in memory. Instead of associating the id, we're changing the code to associate the loaded object, having this, AR will not load again the data from database.

This is mainly happening in current usage which calculates the data on the fly and create Fee objects in memory.


annvelents (approved):
Nice! 🚀

PR #5421, opened 2026-04-28
fix: skip flaky clickhouse tests

lovrocolic:
## Description

Previous `skip` command hasn't work as intended so we had to change it and move to the example.

### DIFF
--- a/spec/services/events/stores/clickhouse/clean_duplicated_service_spec.rb
+++ b/spec/services/events/stores/clickhouse/clean_duplicated_service_spec.rb
@@ -2,9 +2,8 @@
 
 require "spec_helper"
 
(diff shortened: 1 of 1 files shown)


vincent-pochet (approved):
👍 

I have a proper fix for the test in https://github.com/getlago/lago-api/pull/5416

toommz (commented):
You are missing some indentation and maybe an `end` right?

lovrocolic:
> You are missing some indentation and maybe an `end` right?

Actually I just added new metadata attribute:)

toommz:
> > You are missing some indentation and maybe an `end` right?
> 
> Actually I just added new metadata attribute:)

Oh yes, I missed the added comma 👍

PR #5422, opened 2026-04-28
fix(lifetime_usage): Fix clock query performances

jdenquin:
- Move global query to a per organization query to optimize index usage
- New query with a simple organization_id filter is now done in 0,500ms
- Remove `recalculate_current_usage` from `LifetimeUsage` model and from the query filters, its not used anymore

### DIFF
--- a/spec/jobs/clock/refresh_lifetime_usages_job_spec.rb
+++ b/spec/jobs/clock/refresh_lifetime_usages_job_spec.rb
@@ -8,7 +8,7 @@ describe Clock::RefreshLifetimeUsagesJob, job: true do
   describe ".perform" do
     let(:organization) { create(:organization) }
     let(:lifetime_usage1) { create(:lifetime_usage, organization:, recalculate_invoiced_usage: true) }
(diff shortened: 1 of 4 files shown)


spec/jobs/clock/refresh_lifetime_usages_job_spec.rb
```diff
@@ -8,20 +8,17 @@
   describe ".perform" do
     let(:organization) { create(:organization) }
     let(:lifetime_usage1) { create(:lifetime_usage, organization:, recalculate_invoiced_usage: true) }
-    let(:lifetime_usage2) { create(:lifetime_usage, organization:, recalculate_current_usage: true) }
-    let(:lifetime_usage3) { create(:lifetime_usage, organization:, recalculate_invoiced_usage: false, recalculate_current_usage: false) }
+    let(:lifetime_usage3) { create(:lifetime_usage, organization:, recalculate_invoiced_usage: false) }
```
vincent-pochet: Since we removed `lifetime_usage2`, maybe we could rename this one

vincent-pochet (approved):

PR #5423, opened 2026-04-28
fix(invoices): Fix invoice query status filtering performances

jdenquin:
- Use `with_status` by default to avoid `where status IN()` duplicates that makes PG plan terrible
- Concatenate visible status with query filters
- Remove `visible` scope for customer_id filtering since the `with_status` will be applied as a AND in the outer scope


D1353L (approved):
LGTM 🚀

ancorcruz (approved):
LGTM!

vincent-pochet (approved):

PR #5425, opened 2026-04-28
feat(speedup): Use subscription loop to filter events in order to use…

mariohd:
Use subscription loop to speedup the process of event queries so it uses `idx_events_for_distinct_codes` index.

### DIFF
--- a/lib/tasks/recipes/events.rake
+++ b/lib/tasks/recipes/events.rake
@@ -2,7 +2,7 @@
 
 require_relative "../../task_prompt"
 
-BATCH_SIZE = 1000
+BATCH_SIZE = 2000
 
 # rubocop:disable Rails/Output,Rails/Exit
 namespace :recipes do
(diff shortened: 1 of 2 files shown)


lib/tasks/recipes/events.rake
```diff
@@ -14,32 +14,50 @@ namespace :recipes do
 
       from_time, to_time = TaskPrompt.ask_for_timestamp_range
 
-      events = Event.where(organization_id: organization.id)
-        .from_datetime(from_time)
-        .to_datetime(to_time)
+      subscriptions = organization.subscriptions.pluck(:external_id)
```
vincent-pochet: Since this can potentially be a lot of valus, maybe we should just check the number of distinct external_id?

lib/tasks/recipes/events.rake
```diff
@@ -14,32 +14,50 @@ namespace :recipes do
...
 
-      deleted = 0
+      total_deleted = 0
 
       # rubocop:disable Rails/SkipsModelValidations
-      events.in_batches(of: BATCH_SIZE) do |batch|
-        batch.update_all(deleted_at: Time.current)
-        deleted += BATCH_SIZE
-        print "\rDeleted #{[deleted, count].min}/#{count} events..."
+      subscriptions.each_with_index do |external_id, index|
```
vincent-pochet, line 33: I would suggest to loop over distinct external_id using a find_each or in_batch

vincent-pochet (approved):

lib/tasks/recipes/events.rake
mariohd: oh, makes total sense!!

lib/tasks/recipes/events.rake
mariohd: the `subscriptions` here is an Array of strings with the external ids. 
If we move to `find_each` or `in_batches` we would lose the counters feedback (not a super real problem/concern).

lib/tasks/recipes/events.rake
```diff
@@ -14,32 +14,50 @@ namespace :recipes do
 
       from_time, to_time = TaskPrompt.ask_for_timestamp_range
 
-      events = Event.where(organization_id: organization.id)
-        .from_datetime(from_time)
-        .to_datetime(to_time)
+      subscriptions = organization.subscriptions.distinct.pluck(:external_id)
```
groyoh, line 17: This could potentially be millions each entries 😅 I guess it's fine for now but it could be dangerous.

groyoh: Should we maybe check only active subscriptions ? 🤔

groyoh (approved):

lib/tasks/recipes/events.rake
mariohd: I thought on this as well and checked for their subs. 
They have around ~16k total subs and ~8k active subs. 

The problem of having only active is that if a sub status become different than `active` we will not destroy the events and it may result in strange scenarios for `terminated` for example, _I think_. 

 ```
lago-api(prod)> organization.subscriptions.pluck(:external_id).count
=> 16833
lago-api(prod)> organization.subscriptions.active.pluck(:external_id).count
=> 8073
lago-api(prod)> organization.subscriptions.distinct.pluck(:external_id).count
=> 16029
```

I checked some subs external_ids and they had no events, so we would skip some subs anyway. 
That's why I kept for all distinct external id subs.

aquinofb (approved):

PR #5426, opened 2026-04-28
fix(clickhouse): Speedup events enriched deduplication

vincent-pochet:
## Description

This PR follows https://github.com/getlago/lago-api/pull/5416

It applies the same two-pass deduplication strategy to the `Events::Stores::ClickhouseEnrichedStore`

### DIFF
--- a/app/services/events/stores/clickhouse_enriched_store.rb
+++ b/app/services/events/stores/clickhouse_enriched_store.rb
@@ -83,7 +83,6 @@ module Events
 
         query = Arel::Table.new(:events_enriched_expanded)
         query = query.order(arel_table[:timestamp].desc, arel_table[order_column.to_sym]) if ordered
-        query = apply_arel_grouped_by_values(query) if grouped_by_values?
 
         ctes["events"] = query.project(select).to_sql
         ctes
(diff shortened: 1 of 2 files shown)


spec/services/events/stores/clickhouse/clean_duplicated_service_spec.rb
toommz, line 19: Is it something we might want to have for each spec blocks with the `:clickhouse` tag?

app/services/events/stores/clickhouse_enriched_store.rb
```diff
@@ -61,77 +64,90 @@ def events_cte_queries_with_deduplication(force_from: false, ordered: false, sel
...
+      # Code-based fallback for recurring charges.
+      # First pass: groups events on the smaller dedup key (no charge_id/charge_filter_id)
+      # because fallback events come from a previous subscription's charges, so charge_ids are irrelevant.
+      # Matching/ignored filters and grouped_by_values are deferred to the second pass to keep
+      # the dedup grouping based on the base columns only (org, code, subscription, timestamp, transaction).
+      def latest_enriched_fallback_sql(from_datetime:)
+        <<~SQL.squish
+          SELECT #{DEDUP_FALLBACK_KEY_COLUMNS.join(", ")}, max(enriched_at) AS max_enriched_at
+          FROM events_enriched_expanded
+          WHERE #{code_based_fallback_where_sql(from_datetime:, include_grouped_by_values: false, include_filters: false)}
```
toommz, line 129: As we don't include filters on pass 1, could it be the case that during pass 2 we miss some rows because of the filtering? I think it can only be the case if event enrichment can change properties. Here is a schema to illustrate that:

Image

But it might not be a problem if the considered source of trust is the latest enrichment 🤔

vincent-pochet: @toommz
For now, that's actually the goal. We only want to consider the latest event, that's why I moved the filtering on the second pass. Otherwise, the "same" event (transaction_id/timestamp tuple) would be considered twice...

On the long run, we are discussing a change in the idempotency approach of the clickhouse pipeline to match what we currently have in Postgres (no rewriting of events). It would solve all these "duplication" issues.

spec/services/events/stores/clickhouse/clean_duplicated_service_spec.rb
vincent-pochet: Not necessarily, here we are forcing clickhouse to not deduplicate because we want to test the cleanup behavior. It's heavy and slow down the tests and mostly not useful in most cases.

app/services/events/stores/clickhouse_enriched_store.rb
toommz: Ok makes sense! Thanks for the explanation

toommz (approved):

PR #5435, opened 2026-04-29
chore(ci): alert when schema change breaks lago-front type-checking

ansmonjol:
## Summary

Adds a post-merge GitHub Actions workflow that surfaces schema changes which break `lago-front` type-checking, the moment they land on `main` — instead of the front team finding out when their next PR's codegen workflow fails (which is what happened with [getlago/lago-front#3442](https://github.com/getlago/lago-front/pull/3442) earlier today).

## How it works

1. **Trigger:** `push` to `main` **with `paths: ['schema.graphql']`** — fires only when a merge actually changes the schema. Most merges don't, so this is cheap.
2. **Steps:** check out lago-api + lago-front main → regenerate front's types using the new `schema.graphql` (file path, no API boot) → run `pnpm tsc`.
3. **Node version:** inherited from `lago-front/package.json` via `node-version-file`, so this workflow tracks any node bump on the front side without manual upkeep here (i'll improve in a followup PR tho)
4. **On failure:** the workflow turns red. GitHub sends the standard failed-workflow email to the commit author and to anyone subscribed to this repo's Actions notifications — that's the alert.

## Why post-merge, not pre-merge

We weighed three concrete shapes before landing here: Pre-merge blocking,Pre-merge, informational and Post-merge alert (this PR)

Picking between these comes down to a single question: **does the project care more about API velocity, or about strict main coherence?** and I prefer to be alerted soon then fix vs having to disable alerts or find workaround each time we build something with both FE and API

### DIFF
--- a/.github/workflows/front-compatibility.yml
+++ b/.github/workflows/front-compatibility.yml
@@ -22,6 +22,16 @@ on:
     branches: [main]
     paths:
       - 'schema.graphql'
+  # Manual dispatch for testing: lets you point the check at any
+  # lago-front ref. Useful for verifying both directions — a branch you
+  # expect to pass, and one you've intentionally broken (e.g. by removing
+  # an enum case from useActivityLogsInformation.ts) to see the failure.
+  workflow_dispatch:
+    inputs:
+      front_ref:
+        description: 'lago-front ref to type-check against (branch, tag, or SHA)'
+        required: false
+        default: 'main'
 
 jobs:
   front-typecheck:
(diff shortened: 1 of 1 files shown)


domenicofalco (approved):

groyoh:
@ansmonjol could you add a workflow dispatch trigger to be able to test it easily? Because right now, I'm not fully sure what's expected to happen.

ansmonjol:
Added `workflow_dispatch` in [20d8d39](https://github.com/getlago/lago-api/pull/5435/commits/20d8d392c). You can trigger it from the Actions tab once this PR's branch (or main) has the workflow.

The dispatch takes an optional `front_ref` input (default: `main`):

- Run with default → expects pass against current front main.
- Point it at a branch with intentionally broken front code (e.g. revert one of the enum cases from `src/hooks/activityLogs/useActivityLogsInformation.ts` on a throwaway branch) → expects failure with the same `TS2739` / `TS2345` errors a real schema break would produce.

The `push` trigger continues to always check against `main` regardless of the input — the input only applies to manual runs.

I'll merge this PR so we can test it in the actions

PR #5436, opened 2026-04-29
fix: handle n+1 issue in wallet serializer

lovrocolic:
## Context

`GET /api/v1/wallets triggers` an N+1 query pattern: each wallet on the page issues additional round-trips to RDS for `customers`, `recurring_transaction_rules`, `billable_metrics` (via wallet_targets), `invoice_custom_sections`, and `item_metadata`.

## Description

This PR eager loads associated wallet records. For recurring transaction rules small helper method is added because `.active` scope used to disable loaded records and we still had to hit DB for each recurring rule. With the new approach recurring rules are filtered in memory.

### DIFF
--- a/app/serializers/v1/wallet_serializer.rb
+++ b/app/serializers/v1/wallet_serializer.rb
@@ -43,10 +43,16 @@ module V1
 
     def recurring_transaction_rules
       ::CollectionSerializer.new(
-        model.recurring_transaction_rules.active,
+        active_recurring_transaction_rules,
         ::V1::Wallets::RecurringTransactionRuleSerializer,
         collection_name: "recurring_transaction_rules"
       ).serialize
+    end
+
+    def active_recurring_transaction_rules
+      model.recurring_transaction_rules.select do |rule|
+        rule.active? && (rule.expiration_at.nil? || rule.expiration_at > Time.current)
+      end
     end
 
     def limitations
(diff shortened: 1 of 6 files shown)


app/serializers/v1/wallet_serializer.rb
```diff
@@ -43,12 +43,18 @@ def serialize
...
-        model.recurring_transaction_rules.active,
+        active_recurring_transaction_rules,
         ::V1::Wallets::RecurringTransactionRuleSerializer,
         collection_name: "recurring_transaction_rules"
       ).serialize
     end
 
+    def active_recurring_transaction_rules
+      model.recurring_transaction_rules.select do |rule|
+        rule.active? && (rule.expiration_at.nil? || rule.expiration_at > Time.current)
```
groyoh: I'm wondering whether we should override `active?` to check for expiration

groyoh (approved):
Any chance you can add bullet to the controller test to ensure we have no N+1 regression there?

aquinofb (approved):
LGTM!

app/serializers/v1/wallet_serializer.rb
lovrocolic: TBH I would rather keep existing rails generated method, but I added new model method called `currently_active?` which is testable and used in serializer. wdyt?

groyoh (approved):

PR #5437, opened 2026-04-29
Multi be foundations

mariohd:
## Summary                                                                                                                                                                                                         
                                                                                                                                                                                                                     
  - Add optional `billing_entity_id` column to `subscriptions` and `wallets` tables with indexes and foreign keys                                                                                                    
  - Replace `has_one :billing_entity, through: :customer` on Subscription with a direct `belongs_to :billing_entity, optional: true`
  - Add `belongs_to :billing_entity, optional: true` to Wallet                                                                                                                                                       
  - Both models fall back to `customer.billing_entity` when their own `billing_entity_id` is nil                                                                                                                     
                                              
  ## Test plan                                                                                                                                                                                                       
                                                                                                                                                                                                                     
  - [x] Subscription returns its own billing entity when set                                                                                                                                                         
  - [x] Subscription falls back to customer billing entity when nil                                                                                                                                                  
  - [x] Wallet returns its own billing entity when set                                                                                                                                                               
  - [x] Wallet falls back to customer billing entity when nil

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -11,8 +11,8 @@ class Subscription < ApplicationRecord
   belongs_to :previous_subscription, class_name: "Subscription", optional: true
   belongs_to :organization
   belongs_to :payment_method, optional: true
+  belongs_to :billing_entity, optional: true
 
-  has_one :billing_entity, through: :customer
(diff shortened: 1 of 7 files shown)


aquinofb (approved):
LGTM!

lovrocolic (approved):
🚀

PR #5440, opened 2026-04-30
fix(clickhouse): Deadletter queue timestamp parsing

vincent-pochet:
## Description

This PR fixes the parsing of `event.timestamp` in the `events_dead_letter_mv` materialized view.

The timestamp is a unix timestamp stored as a string and finally parsed into a datetime. With some values, for example "3600.0", the Datetime conversion was failing because Clickhouse was not considering the value as a timestamp, but as date that it was then failing to parse.

The new logic adds a fallback on a Float64 conversion before converting into a Datetime, and finally on the event `ingested_at` value as a last resort, to avoid failing the pipeline completly since the `events_dead_letter` table is expecting a not null value for timestamp


groyoh (approved):

aquinofb (approved):
LGTM!

mariohd (approved):

PR #5441, opened 2026-04-30
  feat(payment-gated-subs): wire gated activation into Subscriptions::CreateService

ancorcruz:
## Context

PR #5370 introduced status-based branching in `Subscriptions::ActivateService` (pending → evaluate rules and either gate or activate; incomplete → resume after payment). However, `Subscriptions::CreateService` was still using a flat linear flow that activated subscriptions unconditionally after creation, so subscriptions created with `activation_rules` via REST or GraphQL never actually entered a gated state. This PR closes that gap.

## Description

Refactor `CreateService#handle_subscription` into three explicit branches keyed on `subscription_at` vs today in the customer's timezone:

- `handle_today_subscription` creates the subscription as `:pending`, applies the activation rules, and dispatches to `ActivateService.call!`. ActivateService's `activate_from_pending` evaluates the rules and either gates the subscription (transition to `:incomplete`, bill, fire `subscription.incomplete`) or activates it immediately when the rules evaluate to `:not_applicable` (e.g. trial without pay-in-advance fixed charges).
- `handle_past_subscription` activates immediately with `subscription_at` as `started_at`, fires `subscription.started`, and ignores activation rules.
- `handle_future_subscription` creates the subscription as `:pending` with rules attached as `:inactive`, deferring evaluation to the existing pending-subscription clock job that fires `ActivateService` on the activation date.

The dispatch comparison switches from sub-second time (`> Time.current`, `< Time.current`) to date-in-customer-timezone. A `subscription_at` falling on the same calendar date as today (in the customer's timezone) lands on the "today" branch regardless of time of day. This handles date-only API payloads (parsed as midnight, which would otherwise read as "past" the moment any time has elapsed) and aligns with the "today / future / past" buckets being conceptually about days, not moments. Practical effect: a subscription scheduled for later today (e.g. 5pm when it's currently 3pm) activates immediately rather than waiting; subscriptions scheduled for a future date still go to `:pending` and activate when that date arrives.

Two stale helper methods (`should_be_billed_today?` and `fixed_charges_billed_today?`) were left orphaned by the refactor — billing dispatch now lives in `ActivateService.bill_subscription` — and are removed.

### Skip charges flag at subscription start

Routing today subscriptions through `ActivateService` surfaced a latent asymmetry from #5370: `gate_subscription` passed `skip_charges: true` to `bill_subscription`, but the sibling `activate_with_side_effects` path did not. The old inline billing in `CreateService` always passed `skip_charges: true` for `:subscription_starting`, so the asymmetry was masked while no caller exercised `activate_with_side_effects` through the starting flow. With the new dispatch it does, and recurring non-invoiceable fees started being created at activation time — there is no usage history yet to bill, so this produced spurious baseline fees and broke several pre-existing scenario tests covering recurring fees on upgrade and downgrade, pay-in-advance charges with taxes, advance-charge dates, and advance-charge regeneration.

Pass `skip_charges: true` from `activate_with_side_effects` to align with `gate_subscription` and the old `CreateService` semantics. Resolves the recurring-fee scenario regressions in the same change.

## Out of scope

- Zero-amount activation shortcut for gated subscriptions (separate dependent PR).
- VIES and provider-tax-failure end-to-end scenarios — these depend on a small `TransitionToFinalStatusService` adjustment that's planned for the follow-up PR.

### DIFF
--- a/app/services/subscriptions/activate_service.rb
+++ b/app/services/subscriptions/activate_service.rb
@@ -4,9 +4,10 @@ module Subscriptions
   class ActivateService < BaseService
     Result = BaseResult[:subscription]
 
-    def initialize(subscription:, timestamp: Time.current)
+    def initialize(subscription:, timestamp: Time.current, during_creation: false)
       @subscription = subscription
       @timestamp = timestamp
+      @during_creation = during_creation
       super
     end
 
@@ -26,7 +27,7 @@ module Subscriptions
 
     private
 
-    attr_reader :subscription, :timestamp
+    attr_reader :subscription, :timestamp, :during_creation
 
     def activate_from_pending
       ActivationRules::EvaluateService.call!(subscription:)
@@ -82,6 +83,11 @@ module Subscriptions
     def notify_started
       SendWebhookJob.perform_later("subscription.started", subscription)
       Utils::ActivityLog.produce(subscription, "subscription.started")
+
+      # Skip Hubspot UpdateJob when activating during subscription creation —
+      # CreateService fires Hubspot::CreateJob after this, which captures the
+      # active state and avoids a redundant Update that would race with Create.
+      return if during_creation
 
       if subscription.should_sync_hubspot_subscription?
         Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob.perform_later(subscription:)
--- a/app/services/subscriptions/activate_service.rb
+++ b/app/services/subscriptions/activate_service.rb
@@ -74,7 +74,7 @@ module Subscriptions
       )
 
       after_commit do
-        bill_subscription
+        bill_subscription(skip_charges: true)
         notify_started
       end
     end
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -119,17 +119,6 @@ module Subscriptions
       plan.yearly_amount_cents < current_subscription.plan.yearly_amount_cents
     end
 
-    def should_be_billed_today?(sub)
-      sub.active? && sub.subscription_at.today? && plan.pay_in_advance? && !sub.in_trial_period?
-    end
-
-    def fixed_charges_billed_today?(sub)
-      return false if !(sub.active? && sub.started_at.today?)
-      return false if sub.fixed_charges.pay_in_advance.none?
-
-      !sub.plan.pay_in_advance? || sub.in_trial_period?
-    end
-
     def create_subscription
       new_subscription = Subscription.new(
         organization_id: customer.organization_id,
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -148,50 +148,16 @@ module Subscriptions
         new_subscription.payment_method_id = params[:payment_method][:payment_method_id] if params[:payment_method].key?(:payment_method_id)
       end
 
-      if new_subscription.subscription_at > Time.current
-        new_subscription.pending!
-        apply_activation_rules(new_subscription)
-      elsif new_subscription.subscription_at < Time.current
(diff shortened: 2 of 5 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -148,57 +137,48 @@ def create_subscription
...
-        after_commit do
-          if fixed_charges_billed_today?(new_subscription)
-            Invoices::CreatePayInAdvanceFixedChargesJob.perform_later(
-              new_subscription,
-              new_subscription.started_at + 1.second
-            )
-          end
-        end
+      if new_subscription.should_sync_hubspot_subscription?
+        after_commit { Integrations::Aggregator::Subscriptions::Hubspot::CreateJob.perform_later(subscription: new_subscription) }
```
vincent-pochet, line 153: In the case of `handle_today_subscription` it seems that both this service and the `ActivateService` are calling `Integrations::Aggregator::Subscriptions::Hubspot::CreateJob.perform_later`. Not sure if it's an issue, but I guess it might be good to double check

vincent-pochet (approved):

app/services/subscriptions/create_service.rb
ancorcruz: Good catch @vincent-pochet — confirmed real, and new in this PR.

Pre-PR, today + active subs activated inline (`mark_as_active!`) and the `after_commit` only fired `Hubspot::CreateJob` with the sub already `:active`. Post-PR they route through `ActivateService.activate_with_side_effects`, which also enqueues `Hubspot::UpdateJob` via `notify_started`. Both jobs end up in the same commit; `UpdateJob` runs first, fails (no record yet), retries until `CreateJob` creates it.

Other `ActivateService` callers (clock-activated future subs, post-payment activation for gated subs) are fine — the Hubspot record was created earlier in their flows, so `UpdateJob` is correct.

Pushed `ae52e4b26`: added a `during_creation` flag to `ActivateService` that skips the `UpdateJob` in `notify_started`. `CreateService.handle_today_subscription` passes it; the other call sites keep the default. `CreateJob` runs after the transaction commits, so the subscription is already `:active` and a single create captures the right state. Tests on both specs.

app/services/subscriptions/activate_service.rb
```diff
@@ -74,7 +75,7 @@ def activate_with_side_effects
       )
 
       after_commit do
-        bill_subscription
+        bill_subscription(skip_charges: true)
```
osmarluz, line 78: @ancorcruz just a heads up that the `app/services/subscriptions/activate_all_pending_service.rb` behavior will change because of this. Before this change, that path billed with `skip_charges: false` and now it bills with `skip_charges: true`

ancorcruz: Thanks for flagging @osmarluz. Confirmed: only fresh-sub activation paths (today via `CreateService`, future-dated via the clock) hit `activate_with_side_effects`. Both are subs with no usage history at activation time, so `skip_charges: true` is correct — there's nothing to bill beyond the subscription fee.

Upgrade and downgrade flows are unaffected — upgrades go through `PlanUpgradeService` (its own `BillSubscriptionJob` with `:upgrading` + dedicated `BillNonInvoiceableFeesJob`), and downgrades activate via `TerminateService.terminate_and_start_next` which marks the next sub active directly. Neither touches `ActivateService`. All 19 upgrade/downgrade scenario specs (incl. `recurring_fee_upgrade_spec` / `recurring_fee_downgrade_spec`) green on this branch.

The previous clock-path default (`skip_charges: false`) was creating a 0-amount baseline recurring fee at activation time — the CalculateFeesService comment around `should_create_recurring_non_invoiceable_fees?` already calls out that these zero-fees are undesirable and patches the upgrade case explicitly. The change here generalizes that intent to all `:subscription_starting` flows.

PR #5443, opened 2026-04-30
fix(payment-gated-subs): Modify validation service to ensure a default payment method exists

osmarluz:
## Context

The payment-rule activation validator previously accepted any provider-linked customer even when no payment method had been registered, so a `pending` subscription could enter activation with nothing to charge against. This PR tightens the validator to mirror the resolution logic in `PaymentMethods::DetermineService` and reject cases that would resolve to `nil` at charge time. Distinct error codes are emitted per failure mode so API consumers can act on the specific issue.

## Description

- Resolve the effective `payment_method_type` / `payment_method_id` by overlaying override params on the subscription per-key, matching what `Subscriptions::CreateService` / `Subscriptions::UpdateService` do — the validator now sees a single source of truth
- Reject activation when the resolved payment method would be `nil` at charge time (no override id, no subscription `payment_method_id`, no customer default)
- Move errors from `field: :payment_method` to `field: :customer`, with one of: `no_linked_payment_provider`, `manual_payment_method_invalid_for_payment_activation_rules`, `payment_method_not_found` (new — id was given but the lookup missed), or `no_default_payment_method` (new — no id and no customer default)
- `manual_payment_method_invalid_for_payment_activation_rules` is only emitted when the effective type is actually `"manual"`

### DIFF
--- a/app/services/subscriptions/activation_rules/payment/validate_service.rb
+++ b/app/services/subscriptions/activation_rules/payment/validate_service.rb
@@ -26,12 +26,54 @@ module Subscriptions
         end
 
         def valid_payment_method?
-          return true if args[:payment_method].present? && args[:payment_method][:payment_method_type] == PaymentMethod::PAYMENT_METHOD_TYPES[:provider]
-          return true if args[:payment_method].blank? && args[:subscription]&.payment_method_type == PaymentMethod::PAYMENT_METHOD_TYPES[:provider]
-          return true if args[:payment_method].blank? && args[:subscription].nil? && args[:customer]&.payment_provider.present?
+          return true if payment_provider? && payment_method_available?
 
-          return add_error(field: :payment_method, error_code: "no_linked_payment_provider") if args[:payment_method].blank? && args[:subscription].nil?
-          add_error(field: :payment_method, error_code: "invalid_for_payment_activation_rules")
+          add_error(field: :payment_method, error_code: failure_error_code)
+        end
+
+        def payment_provider?
+          return custom_payment_method_type_provider? if custom_payment_method?
+          return subscription_payment_method_type_provider? if args[:subscription]
+
+          customer_payment_provider?
+        end
+
+        def payment_method_available?
+          return custom_payment_method_id? || customer&.default_payment_method.present? if custom_payment_method?
+          return args[:subscription].payment_method_id.present? || customer&.default_payment_method.present? if args[:subscription]
+
+          customer&.default_payment_method.present?
+        end
+
+        def failure_error_code
+          return "customer_has_no_linked_payment_provider" if !custom_payment_method? && args[:subscription].nil? && !customer_payment_provider?
+          return "customer_has_no_default_payment_method" if payment_provider?
+
+          "manual_payment_method_invalid_for_payment_activation_rules"
+        end
+
+        def custom_payment_method?
+          args[:payment_method].present?
+        end
+
+        def custom_payment_method_type_provider?
+          args[:payment_method][:payment_method_type] == PaymentMethod::PAYMENT_METHOD_TYPES[:provider]
+        end
+
+        def custom_payment_method_id?
+          custom_payment_method? && args[:payment_method][:payment_method_id].present?
+        end
+
+        def subscription_payment_method_type_provider?
+          args[:subscription].payment_method_type == PaymentMethod::PAYMENT_METHOD_TYPES[:provider]
+        end
+
+        def customer_payment_provider?
+          args[:customer]&.payment_provider.present?
+        end
+
+        def customer
+          @customer ||= args[:customer] || args[:subscription]&.customer
         end
       end
     end
--- a/app/services/subscriptions/activation_rules/payment/validate_service.rb
+++ b/app/services/subscriptions/activation_rules/payment/validate_service.rb
@@ -26,54 +26,51 @@ module Subscriptions
         end
 
         def valid_payment_method?
-          return true if payment_provider? && payment_method_available?
+          if effective_payment_method_type.present?
+            return true if effective_payment_method_type_provider? && resolved_payment_method.present?
+          elsif customer_payment_provider? && resolved_payment_method.present?
+            return true
+          end
 
-          add_error(field: :payment_method, error_code: failure_error_code)
+          add_error(field: :customer, error_code: failure_error_code)
         end
 
-        def payment_provider?
-          return custom_payment_method_type_provider? if custom_payment_method?
-          return subscription_payment_method_type_provider? if args[:subscription]
-
-          customer_payment_provider?
+        def effective_payment_method_type_provider?
+          effective_payment_method_type == PaymentMethod::PAYMENT_METHOD_TYPES[:provider]
         end
 
-        def payment_method_available?
-          return custom_payment_method_id? || customer&.default_payment_method.present? if custom_payment_method?
-          return args[:subscription].payment_method_id.present? || customer&.default_payment_method.present? if args[:subscription]
+        def effective_payment_method_type
+          return args[:payment_method][:payment_method_type] if args[:payment_method]&.key?(:payment_method_type)
 
-          customer&.default_payment_method.present?
+          args[:subscription]&.payment_method_type
+        end
+
+        def effective_payment_method_id
+          return args[:payment_method][:payment_method_id] if args[:payment_method]&.key?(:payment_method_id)
+
+          args[:subscription]&.payment_method_id
+        end
+
+        def resolved_payment_method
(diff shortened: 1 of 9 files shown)


app/services/subscriptions/activation_rules/payment/validate_service.rb
```diff
@@ -26,12 +26,54 @@ def valid_timeout_hours?
...
+          return args[:subscription].payment_method_id.present? || customer&.default_payment_method.present? if args[:subscription]
+
+          customer&.default_payment_method.present?
+        end
+
+        def failure_error_code
+          return "customer_has_no_linked_payment_provider" if !custom_payment_method? && args[:subscription].nil? && !customer_payment_provider?
+          return "customer_has_no_default_payment_method" if payment_provider?
+
+          "manual_payment_method_invalid_for_payment_activation_rules"
```
ancorcruz: We still have use `no_linked_payment_provider` error code in some services...

* generate_checkout_url_service.rb,
* invoices/payments/generate_payment_url_service.rb,
* payment_requests/payments/generate_payment_url_service.rb,
* customers/generate_checkout_url_service.rb

we better keep the original naming to avoid breaking contracts or add inconsistent error codes across services.



@domenicofalco are we using these error codes in FE already?

app/services/subscriptions/activation_rules/payment/validate_service.rb
```diff
@@ -26,12 +26,54 @@ def valid_timeout_hours?
...
+        def custom_payment_method?
+          args[:payment_method].present?
+        end
+
+        def custom_payment_method_type_provider?
+          args[:payment_method][:payment_method_type] == PaymentMethod::PAYMENT_METHOD_TYPES[:provider]
+        end
+
+        def custom_payment_method_id?
+          custom_payment_method? && args[:payment_method][:payment_method_id].present?
```
ancorcruz: this only checks the payment_method_id is provided, but it does not ensures it belongs to the customer as DetermineService does with `customer.payment_methods.find_by(id: ...)`.

As this is a hard requirement, it should also ensures the payment method exists and belongs to the customer.

app/services/subscriptions/activation_rules/payment/validate_service.rb
```diff
@@ -26,12 +26,54 @@ def valid_timeout_hours?
...
-          return true if args[:payment_method].blank? && args[:subscription]&.payment_method_type == PaymentMethod::PAYMENT_METHOD_TYPES[:provider]
-          return true if args[:payment_method].blank? && args[:subscription].nil? && args[:customer]&.payment_provider.present?
+          return true if payment_provider? && payment_method_available?
 
-          return add_error(field: :payment_method, error_code: "no_linked_payment_provider") if args[:payment_method].blank? && args[:subscription].nil?
-          add_error(field: :payment_method, error_code: "invalid_for_payment_activation_rules")
+          add_error(field: :payment_method, error_code: failure_error_code)
+        end
+
+        def payment_provider?
```
ancorcruz: this method name is misleading... it reads as "is a payment provider configured?" but actually means "the effective payment method type is :provider".

Could we rename it as `provider_payment_method_type?` or `effective_payment_method_type_provider?` or similar.

app/services/subscriptions/activation_rules/payment/validate_service.rb
```diff
@@ -26,12 +26,54 @@ def valid_timeout_hours?
...
+        end
+
+        def failure_error_code
+          return "customer_has_no_linked_payment_provider" if !custom_payment_method? && args[:subscription].nil? && !customer_payment_provider?
+          return "customer_has_no_default_payment_method" if payment_provider?
+
+          "manual_payment_method_invalid_for_payment_activation_rules"
+        end
+
+        def custom_payment_method?
```
ancorcruz: this also reads as opposite as default.. but actually means "override params were supplied in the request". qhat about `payment_method_override?` or `override_payment_method?`. This would be more honest.

spec/services/subscriptions/activation_rules/payment/validate_service_spec.rb
```diff
@@ -61,16 +67,36 @@
...
+
+        context "when customer has no default payment method" do
+          it "is invalid with customer_has_no_default_payment_method error" do
+            expect(validate_service).not_to be_valid
+            expect(result.error.messages[:payment_method]).to eq(["customer_has_no_default_payment_method"])
+          end
+        end
+      end
+
+      context "when payment_method params include a payment_method_id" do
```
ancorcruz, line 93: lets also include another test case for payment method params includeing a payment method id but belonging to another customer (or one that does not exists).

```
  context "when payment_method_id refers to a non-existent payment method" do
    let(:payment_method_params) { {payment_method_type: "provider", payment_method_id: "00000000-0000-0000-0000-000000000000"} }

    it "is invalid" # currently passes, which is the bug
  end
```

spec/services/subscriptions/activation_rules/payment/validate_service_spec.rb
```diff
@@ -79,16 +105,36 @@
...
           let(:subscription) { create(:subscription, customer:, plan:, organization:, payment_method_type: "provider") }
 
-          it { is_expected.to be_valid }
+          context "when customer has a default payment method" do
+            before { create(:payment_method, customer:, organization:) }
+
+            it { is_expected.to be_valid }
+          end
+
+          context "when subscription has its own payment_method_id" do
```
ancorcruz, line 151: could also a subscription has a payment method id that is deleted or soft deleted?

ancorcruz (commented):
great improvement.

I have left some comments for you consideration

app/services/subscriptions/activation_rules/payment/validate_service.rb
osmarluz: @ancorcruz the problem for me is that `no_linked_payment_provider` is attached to the `base` field for most of those cases, which is confusing in my opinion. I'd say an improvement here would be changing the field to `customer`, because in the end everything derives from the customer. That way we'd have error messages that are clearer and we get rid of the `customer_has_` prefix:

```json
{
  "status": 422,
  "error": "Unprocessable Entity",
  "code": "validation_errors",
  "error_details": {
    "customer": ["no_linked_payment_provider"]
  }
}
```

```json
{
  "status": 422,
  "error": "Unprocessable Entity",
  "code": "validation_errors",
  "error_details": {
    "customer": ["no_default_payment_method"]
  }
}
```

```json
{
  "status": 422,
  "error": "Unprocessable Entity",
  "code": "validation_errors",
  "error_details": {
    "customer": ["manual_payment_method_invalid_for_payment_activation_rules"]
  }
}
```

What do you think?

osmarluz: @ancorcruz regarding your question about the frontend, this won't be used there because we won't allow the user to select `Activate on successful payment` if there's no provider payment method available for the customer

image

app/services/subscriptions/activation_rules/payment/validate_service.rb
osmarluz: @ancorcruz yeah, I thought about this as well and it makes sense. Ideally this would be added to the `app/services/payment_methods/validate_service.rb`, so it's not only validated when we have activation rules in scope. Ill add it here for now and we can check about moving it to `app/services/payment_methods/validate_service.rb` later

app/services/subscriptions/activation_rules/payment/validate_service.rb
osmarluz: @ancorcruz the last line effectively checks the payment provider presence. I think we can break this into 2 methods so it's unambiguous

spec/services/subscriptions/activation_rules/payment/validate_service_spec.rb
osmarluz: @ancorcruz yes, that's possible. We don't update the subscriptions once the payment method is soft deleted

ancorcruz: I guess my comment was in the line of payment_method_id referenced a payment_method that is already deleted/soft-deleted when this service is called. For example, the client has the id cached in their system and include it on the request even it was already deleted.

app/services/subscriptions/activation_rules/payment/validate_service.rb
ancorcruz: sounds right.

spec/services/subscriptions/activation_rules/payment/validate_service_spec.rb
osmarluz: @ancorcruz I considered this scenario as well and we'll raise an error if the payment_method_id is not found, either coming from the params or already saved in a subscription

ancorcruz (approved):
LGTM, thanks!

PR #5444, opened 2026-04-30
feat(data-pipeline): Add the usage views to the data pipeline

murparreira:
## Context

We need the `usage_monitoring_triggered_alerts` and `usage_monitoring_alert_thresholds` views.

## Description

Create both views with scenic.


ancorcruz (approved):
LGTM

PR #5445, opened 2026-04-30
feat(presentation_group_keys): Clear presentation_breakdowns for adjusted fees

D1353L:
## Context
Once the fees are generated and associated with an Invoice calling the services, fees can be adjusted, which means, users can change some attributes that directly affect the presentation_breakdowns. 

If the units changed, the breakdown is not valid anymore and should be cleared.

## Description

- `AdjustedFees::EstimateService` - clears `presentation_breakdowns` on the estimated fee unconditionally.
- `Fees::ChargeService` - when an adjusted fee applies, skips breakdown building only when units changed. Refactored the in-scope adjusted-fee check into a single `applicable_adjusted_fee` helper used by both `init_fee` and the breakdown gate.
- `Invoices::RegenerateFromVoidedService` - copies breakdowns from the voided fee onto the duplicated fee only when units aren't being adjusted.

Test coverage added across all three services for the units-changed, units-unchanged, amount-only, display-name-only, and fixed-charge paths, plus a regression test for the case where
`fee_params[:units]` is omitted by the GraphQL caller.

## Testing
2 Main UI scenarios:
- Edit fee units on Draft invoice.
- Edit fee units when regenerating a voided invoice.

### DIFF
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -168,10 +168,28 @@ module Fees
         fee = init_fee(amount_result, properties:, charge_filter:)
         next if fee.nil?
 
-        build_breakdowns_for_fee(fee:, breakdowns:)
+        unless adjusted_units_overridden?(amount_result:, fee:, charge_filter:)
+          build_breakdowns_for_fee(fee:, breakdowns:)
+        end
 
         fee
       end.compact
+    end
+
+    def adjusted_units_overridden?(amount_result:, fee:, charge_filter:)
+      return false unless applicable_adjusted_fee(amount_result:, charge_filter:)
+
+      amount_result.units != fee.units
+    end
+
+    def applicable_adjusted_fee(amount_result:, charge_filter:)
+      return nil if current_usage
+      return nil unless invoice&.draft?
+
+      adjusted = adjusted_fee(charge_filter:, grouped_by: amount_result.grouped_by)
+      return nil if adjusted.nil? || adjusted.adjusted_display_name?
+
+      adjusted
     end
 
     def build_breakdowns_for_fee(fee:, breakdowns:)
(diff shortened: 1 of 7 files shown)


app/services/fees/charge_service.rb
```diff
@@ -165,15 +165,28 @@ def fees_from_charge_model_result(charge_model_result, properties:, charge_filte
         # TODO: check if this is still needed as we now skip certain zero units fees
         next if current_usage && charge_filter && amount_result.units.zero? && !with_zero_units_filters
 
-        fee = init_fee(amount_result, properties:, charge_filter:)
+        adjusted = applicable_adjusted_fee(amount_result:, charge_filter:)
+        fee = init_fee(amount_result, properties:, charge_filter:, adjusted:)
         next if fee.nil?
 
-        build_breakdowns_for_fee(fee:, breakdowns:)
+        if adjusted.nil? || amount_result.units == fee.units
```
tiagolupepic, line 172: this logic can be delegated to the Fee instead. The adjusted service builds a Fee within adjusted_fee relation. So, you can have a method on `Fee#adjusted_units?` (or something similar), just need to double check to not accidentally introduce a N+1 when hitting the fee.adjusted_fee.

It's more or less what we have in `adjusted_display_name?` method.

app/services/fees/charge_service.rb
```diff
@@ -165,15 +165,28 @@ def fees_from_charge_model_result(charge_model_result, properties:, charge_filte
...
+          build_breakdowns_for_fee(fee:, breakdowns:)
+        end
 
         fee
       end.compact
     end
 
+    def applicable_adjusted_fee(amount_result:, charge_filter:)
```
tiagolupepic, line 180: could you move the comments from lines 214-215?

spec/services/adjusted_fees/estimate_service_spec.rb
```diff
@@ -429,4 +429,110 @@
       end
     end
   end
+
+  describe "presentation_breakdowns" do
```
tiagolupepic: nitpick.. this describe is not necessary, the context below is self explanatory

spec/services/adjusted_fees/estimate_service_spec.rb
```diff
@@ -429,4 +429,110 @@
...
+          expect(result.fee.presentation_breakdowns).to be_empty
+        end
+      end
+
+      context "when keeping units the same" do
+        let(:params) { {fee_id: charge_fee.id, units: 5} }
+
+        it "returns the fee without presentation_breakdowns" do
+          expect(charge_fee.units).to eq(result.fee.units)
+          expect(result.fee.presentation_breakdowns).to be_empty
```
tiagolupepic, line 524: do we should keep the presentation_breakdowns for this case?

D1353L: confirmed - shouldn't keep

app/services/fees/charge_service.rb
D1353L: This is a good idea but unfortunately these 2 approaches don't work as expected:
1. `adjusted_fee.adjusted_units?` - a flag that shows whether only units passed in request params.
2. `adjusted_fee.units != fee.units` - always false as InitFromAdjustedChargeFeeService copies units value from adjusted_fee to a new fee.

spec/services/adjusted_fees/estimate_service_spec.rb
D1353L: moved these tests to the existing contexts above

app/services/fees/charge_service.rb
D1353L: expanded the comment to reference this method. I think the current place is good as it explains why init_fee has 2 paths

tiagolupepic (approved):
:shipit:

PR #5448, opened 2026-05-04
fix(charge-filters): Cascade filter changes via `ChargeFilters::CascadeJob` on bulk updates

rsempe:
Extend the filter-level cascade introduced in #5357 to the bulk update paths: filters are stripped from the legacy job's params and dispatched individually via `ChargeFilters::CascadeJob`.

### DIFF
--- a/app/services/charge_filters/cascade_dispatcher.rb
+++ b/app/services/charge_filters/cascade_dispatcher.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module ChargeFilters
+  module CascadeDispatcher
+    module_function
+
+    # Diffs `before` against `after` and enqueues one ChargeFilters::CascadeJob
+    # per change. `before` and `after` are arrays of:
+    #   { values: {"key" => [...]}, properties: {...} | nil, invoice_display_name: "..." }
+    # `values` and `properties` must be string-keyed.
+    def call(charge:, before:, after:)
+      before_by_values = before.index_by { |f| f[:values] }
+
+      after.each do |new_filter|
+        existing = before_by_values.delete(new_filter[:values])
+
+        next if existing &&
+          existing[:properties] == new_filter[:properties] &&
+          existing[:invoice_display_name] == new_filter[:invoice_display_name]
+
+        ChargeFilters::CascadeJob.perform_later(
+          charge.id,
+          existing ? "update" : "create",
+          new_filter[:values],
+          existing&.dig(:properties),
+          new_filter[:properties],
+          new_filter[:invoice_display_name]
+        )
+      end
+
+      before_by_values.each_value do |old|
+        ChargeFilters::CascadeJob.perform_later(
+          charge.id,
+          "destroy",
+          old[:values],
+          old[:properties],
+          nil,
+          old[:invoice_display_name]
+        )
+      end
+    end
+  end
+end
--- a/app/services/charges/cascade_updatable.rb
+++ b/app/services/charges/cascade_updatable.rb
@@ -10,31 +10,62 @@ module Charges
       return unless cascade_updates
       return unless charge.children.exists?
 
+      # Charge-level cascade only — filters are stripped from `params` and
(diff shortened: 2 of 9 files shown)


spec/services/charges/update_service_spec.rb
```diff
@@ -413,15 +413,15 @@
...
 
-          expect(Charges::UpdateChildrenJob).to have_received(:perform_later).with(
-            params: hash_including("charge_model", "properties", "filters"),
-            old_parent_attrs: hash_including("id" => charge.id),
-            old_parent_filters_attrs: array_including,
-            old_parent_applied_pricing_unit_attrs: anything
-          )
+          expect(Charges::UpdateChildrenJob).to have_received(:perform_later) do |args|
+            expect(args[:params]).to include("charge_model", "properties")
+            expect(args[:params]).not_to have_key("filters")
```
annvelents, line 421: 👍

annvelents (approved):
LGTM!

app/services/charge_filters/cascade_dispatcher.rb
```diff
@@ -0,0 +1,43 @@
...
+      before_by_values = before.index_by { |f| f[:values] }
+
+      after.each do |new_filter|
+        existing = before_by_values.delete(new_filter[:values])
+
+        next if existing &&
+          existing[:properties] == new_filter[:properties] &&
+          existing[:invoice_display_name] == new_filter[:invoice_display_name]
+
+        ChargeFilters::CascadeJob.perform_later(
```
tiagolupepic, line 21: out of curiosity, since we're dispatching two background jobs and we don't have any guarantee they're be processed in the same order, is there a problem if the "destroy" option process first?

app/services/charges/cascade_updatable.rb
```diff
@@ -10,31 +10,50 @@ def trigger_cascade(old_filters_attrs, old_parent_attrs: nil, old_applied_pricin
...
+        {
+          values: f[:values].deep_stringify_keys,
+          properties: f[:properties]&.deep_stringify_keys,
+          invoice_display_name: f[:invoice_display_name]
+        }
+      end
+
+      charge.filters.reset
+      after = charge.filters.includes(values: :billable_metric_filter).unscope(:order).find_each.map do |filter|
+        {
```
tiagolupepic, line 35: perhaps we can use `.pluck` here and avoid instantiate AR objects in memory?

app/services/charges/cascade_updatable.rb
```diff
@@ -10,31 +10,50 @@ def trigger_cascade(old_filters_attrs, old_parent_attrs: nil, old_applied_pricin
...
-            values: f.to_h
-          }
-        end
+        properties: charge.properties
       }
     end
 
     def capture_old_filters_attrs
-      charge.filters.map { |f| {id: f.id, properties: f.properties} }
+      charge.filters.includes(values: :billable_metric_filter).map do |f|
```
tiagolupepic, line 54: ditto, `.pluck` might be an option

tiagolupepic (approved):
just left some comments to understand the context and they're not a blocker

:shipit:

app/services/charge_filters/cascade_dispatcher.rb
rsempe: i don't see any issues related to this, each job is keyed on a distinct `filter_values` tuple:
- "destroy" is only sent for values present in before but not in after
- "create" is only sent for values present in after but not in before

app/services/charges/cascade_updatable.rb
rsempe: as building values requires `filter.to_h` i would let this as it is for simplicity (with pluck we'd need a join on `charge_filter_values + billable_metric_filters`)

app/services/charges/cascade_updatable.rb
rsempe: same here, `f.to_h` is annoying here if we want to use `pluck`

PR #5450, opened 2026-05-04
fix(netsuite-integration): wait 5 minutes between Net::ReadTimeout retries for invoices

osmarluz:
## Context

`Integrations::Aggregator::Invoices::CreateJob` POSTs to Nango to create the NetSuite invoice. Lago's HTTP client has a 60-second read timeout, while Nango's NetSuite action can take longer than that to commit. When `Net::ReadTimeout` fires on Lago's side, the upstream Nango call keeps running and may still create the NetSuite invoice.

ActiveJob's `:polynomially_longer` then schedules the first retry only ~3-5 seconds after the timeout. That retry POSTs to Nango while the original write is still committing in NetSuite, the optimistic lock raises `RCRD_HAS_BEEN_CHANGED` (HTTP 424), retries exhaust, and the job dies. Lago never reaches the response-handling path that would persist an `IntegrationResource`, so the invoice can exist in NetSuite with no link in Lago.

Nango's NetSuite action times out upstream at 5 minutes. Waiting 5 minutes between retries guarantees the original call has settled (succeeded or definitively failed) before we retry, which eliminates the in-flight race that produces `RCRD_HAS_BEEN_CHANGED`.

## Description

- Replace `retry_on Net::ReadTimeout` with a custom `rescue_from(Net::ReadTimeout)` block. ActiveJob's `wait:` Proc receives only `executions`, not the job instance, so a Proc cannot branch on which integration the invoice belongs to. `rescue_from` is the canonical workaround when the retry decision depends on the job's instance state.
- Add `READ_TIMEOUT_WAIT_BY_PROVIDER_KEY` mapping each provider key to its wait strategy. NetSuite waits 5 minutes (matching Nango's upstream action timeout); other integrations that reach this job (Xero, via `accounting_kind`) fall through to `:polynomially_longer`.
- Reuse `executions_for` and `determine_delay` so the per-exception execution counter, attempts limit, and jitter behave identically to a normal `retry_on`.
- Cover both branches in specs: a NetSuite invoice schedules the next attempt within ~5 minutes (jitter tolerance), a Xero invoice schedules the first retry under one minute (polynomial backoff).

The durable fix is idempotent find-or-create on the Nango RESTlet, planned separately. With idempotency in place, retries against an already-committed record return success instead of `RCRD_HAS_BEEN_CHANGED`. This PR is the in-Lago mitigation until that ships.

### DIFF
--- a/app/jobs/integrations/aggregator/invoices/create_job.rb
+++ b/app/jobs/integrations/aggregator/invoices/create_job.rb
@@ -6,19 +6,43 @@ module Integrations
       class CreateJob < ApplicationJob
         include ConcurrencyThrottlable
 
+        # NOTE: NetSuite waits longer to avoid racing in-flight Nango calls; others use polynomial backoff.
+        READ_TIMEOUT_WAIT_BY_PROVIDER_KEY = {
+          "netsuite" => 5.minutes
+        }.freeze
+
         queue_as "integrations"
 
         unique :until_executed, on_conflict: :log
 
         retry_on LagoHttpClient::HttpError, wait: :polynomially_longer, attempts: 3
-        retry_on Net::ReadTimeout, wait: :polynomially_longer, attempts: 6
         retry_on RequestLimitError, wait: :polynomially_longer, attempts: 100
         retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
         discard_on BaseService::NonRetryableFailure
 
+        rescue_from(Net::ReadTimeout) do |error|
+          attempts = 6
+          executions_count = executions_for([Net::ReadTimeout])
+          raise error if executions_count >= attempts
+
+          wait_strategy = READ_TIMEOUT_WAIT_BY_PROVIDER_KEY.fetch(integration_provider_key, :polynomially_longer)
+
+          retry_job(
+            wait: determine_delay(seconds_or_duration_or_algorithm: wait_strategy, executions: executions_count),
+            error: error
+          )
+        end
+
         def perform(invoice:)
+          @invoice = invoice
           result = Integrations::Aggregator::Invoices::CreateService.call(invoice:)
           result.raise_if_error!
+        end
+
+        private
+
+        def integration_provider_key
+          @invoice&.customer&.integration_customers&.accounting_kind&.first&.integration&.provider_key
         end
       end
     end
--- a/app/jobs/integrations/aggregator/invoices/create_job.rb
+++ b/app/jobs/integrations/aggregator/invoices/create_job.rb
@@ -8,7 +8,7 @@ module Integrations
 
         # NOTE: NetSuite waits longer to avoid racing in-flight Nango calls; others use polynomial backoff.
         # 6 minutes covers Nango's 5-minute upstream NetSuite action timeout with a safety margin.
-        READ_TIMEOUT_WAIT_BY_PROVIDER_KEY = {
+        DELAY_BY_PROVIDER_KEY = {
           "netsuite" => 6.minutes
         }.freeze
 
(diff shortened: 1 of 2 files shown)


app/services/integrations/aggregator/base_service.rb
```diff
@@ -77,7 +78,10 @@ def throttle_key
       end
 
       def http_client
```
lovrocolic, line 79: Can't Nango handle this use case by doing `create_or_update` on second call, when we retry the failed job? I guess it is not ideal to hold worker thread that much 🤔

osmarluz: @lovrocolic yes, the find or create logic will be implemented on Nango through the PR [Find or create netsuite invoice](https://github.com/Nango-Collaborations/nango-lago-collab/pull/62), but we also need to make the read timeout a bit longer so we make sure the retries are less frequent and we don't cause duplicates on Netsuite

osmarluz: @lovrocolic we have reduced the `read_timeout` to 130 seconds after testing. 5 minutes was way too long and it was getting cut off by Nango's load balancer at 150 seconds with a gateway timeout

app/jobs/integrations/aggregator/invoices/create_job.rb
```diff
@@ -6,20 +6,44 @@ module Invoices
       class CreateJob < ApplicationJob
         include ConcurrencyThrottlable
 
+        # NOTE: NetSuite waits longer to avoid racing in-flight Nango calls; others use polynomial backoff.
+        READ_TIMEOUT_WAIT_BY_PROVIDER_KEY = {
+          "netsuite" => 5.minutes
```
ancorcruz: with ActiveJob retry jitter of +/- 45s the job retry could still be within the 5 minutes window of Nango's NetSuite action. Perhaps we should increase the wait time a little bit, for example 6 minutes, to avoid this?

app/jobs/integrations/aggregator/invoices/create_job.rb
```diff
@@ -6,20 +6,44 @@ module Invoices
...
         retry_on LagoHttpClient::HttpError, wait: :polynomially_longer, attempts: 3
-        retry_on Net::ReadTimeout, wait: :polynomially_longer, attempts: 6
         retry_on RequestLimitError, wait: :polynomially_longer, attempts: 100
         retry_on BaseService::ThrottlingError, wait: :polynomially_longer, attempts: 25
         discard_on BaseService::NonRetryableFailure
 
+        rescue_from(Net::ReadTimeout) do |error|
+          attempts = 6
+          executions_count = executions_for([Net::ReadTimeout])
+          raise error if executions_count >= attempts
```
ancorcruz: you can just `raise` to bubble up the same exception.

Also, shall we run `instrument :retry_stopped, error: error` before raisin the error when attempts are exhausted to ensure logs, metrics, Sentry, etc are notified properly?



```suggestion
          if executions_count >= attempts
            instrument :retry_stopped, error:
            raise
          end
```

app/jobs/integrations/aggregator/invoices/create_job.rb
```diff
@@ -6,20 +6,44 @@ module Invoices
...
+          @invoice = invoice
           result = Integrations::Aggregator::Invoices::CreateService.call(invoice:)
           result.raise_if_error!
         end
+
+        private
+
+        def integration_provider_key
+          @invoice&.customer&.integration_customers&.accounting_kind&.first&.integration&.provider_key
+        end
```
ancorcruz: instead of adding an instance variable and access it from the rescue_from block we could still use arguments for this inside the rescue handler `arguments.first[:invoice]`

ancorcruz (approved):
Great approach, I have left a few comments, something else to consider is that `executions_for` and `determine_delay` are not public API of ActiveJob and intended for internal use of `retry_on`. So if ActiveJob internals change (for example in a Rails upgrade) this job will be broken without notice. It does not hurt to makes things easier in the future adding a comment in the code for future reference flagging the dependency.

app/jobs/integrations/aggregator/invoices/create_job.rb
osmarluz: @ancorcruz the jitter is only additive, but it makes sense to increase it so we don't have the risk of retrying at exactly the 5 minute mark. I'll change it

lovrocolic (approved):

PR #5451, opened 2026-05-04
fix(invoices): Prevent duplicate PullTaxesAndApplyJob

D1353L:
## Context

PullTaxesAndApplyJob had no uniqueness lock, causing duplicated tax rows for the same invoice and invoice number shift.

## Description

Added uniqueness for Invoices::ProviderTaxes::PullTaxesAndApplyJob


annvelents:
Unfortunately I'm having some doubts in this solution because a draft invoice might have been edited and latest PullTaxes has more "correct" taxes, so it might wrong to just log the second job

PR #5453, opened 2026-05-04
fix(api): Ability to use params without filter_by prefix to reflect the rest of the API

murparreira:
## Context

We have different param naming convention in the `current_usage` endpoint. Some have the prefix `filter_by_`, not reflecting the pattern on the rest of the API. Add support for conventional ones, and mark the prefixed ones as deprecated.

## Description

Introduce unprefixed aliases for the `filter_by_*` query parameters on the `GET /customers/:customer_id/current_usage` endpoint (`charge_id, charge_code, billable_metric_code, group`), consistent with the naming convention used across the rest of the API.

The old `filter_by_*` names remain supported for backward compatibility.


vincent-pochet (approved):

groyoh (approved):
I believe you can add tests for the usage filter.

PR #5454, opened 2026-05-04
Fix(inovice-preview): add possibility to preview applied_credits application

annvelents:
## Context

When a possibility to add wallet targets was implemented, it was not reflected in invoices preview service.

## Description

Extracted from existing `Credits::AppliedPrepaidCreditsService` a part that calculates future wallet transactions and reused the new service in Invoices::Preview service

### DIFF
--- a/app/services/credits/applied_prepaid_credits_service.rb
+++ b/app/services/credits/applied_prepaid_credits_service.rb
@@ -2,11 +2,10 @@
 
 module Credits
   class AppliedPrepaidCreditsService < BaseService
-    Result = BaseResult[:prepaid_credit_amount_cents, :wallet_transactions, :wallet_transactions_preview]
+    Result = BaseResult[:prepaid_credit_amount_cents, :wallet_transactions]
 
-    def initialize(invoice:, preview: false)
+    def initialize(invoice:)
       @invoice = invoice
-      @preview = preview
 
       super(nil)
     end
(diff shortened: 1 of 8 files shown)


app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -181,28 +107,19 @@ def create_wallet_transaction(wallet, amount_cents)
...
         scope.in_application_order
       end
     end
+
+    def wallets_already_applied?
+      return false unless invoice
+      return false unless invoice.persisted?
+
+      WalletTransaction.exists?(invoice_id: invoice.id, wallet_id: wallets.map(&:id))
+    end
```
rsempe, line 123: can we remove it or avoid the duplication here?

app/services/credits/applied_prepaid_credits_service.rb
```diff
@@ -20,47 +22,14 @@ def call
...
-                fee_key: fee_key,
-                amount_cents: transaction_amount
-              }
-            end
-            total_amount_cents = wallet_fee_transactions.sum { |t| t[:amount_cents] }
-            next if total_amount_cents <= 0
-
-            wallet_transaction = create_wallet_transaction(wallet, total_amount_cents)
+          # per each wallet we create a single wallet transaction. wallets_transaction_amounts is a hash with wallet and total transaction amount
+          wallets_transaction_amounts = CalculateApplicableWalletTransactionsService.call!(invoice:).wallet_transactions
```
rsempe:
```suggestion
          wallets_transaction_amounts = AllocatePrepaidCreditsService.call!(invoice:).wallet_transactions
```

? 🤷‍♂️ 😅

rsempe (approved):

app/services/credits/applied_prepaid_credits_service.rb
annvelents: removed it from Calculate service as  we're not creating anything and "calculation" doesn't hurt even if they already have applied credits
and left it in the service that creates the credits

app/services/credits/applied_prepaid_credits_service.rb
annvelents: `AllocatePrepaidCreditsByWalletsService` 😁 
I think I like it this way!
(because we are also allocating credits to fees, and these are sum of multiple wallets, so without this I feel confused)

PR #5455, opened 2026-05-04
ING-32 perf(events): Add index on (organization_id, transaction_id)

mariohd:
## Summary                                                                                                                                                                                                         
                                                                                                                                                                                                                   
  - Add partial index `(organization_id, transaction_id) WHERE deleted_at IS NULL` on the events table                                                                                                               
  - Fixes `GET /api/v1/events/:id` taking 30+ seconds on Postgres-backed orgs with large event volumes                                                                                                             
  - Same index benefits the GraphQL `event(transactionId:)` resolver which uses the identical query pattern
                                              
  ## Test plan                            
                                                                                                                                                                                                                     
  - [ ] Run migration on staging and verify index creation with `\d events`                                                                                                                                          
  - [ ] EXPLAIN ANALYZE `SELECT * FROM events WHERE organization_id = '<org_id>' AND transaction_id = '<txn_id>' AND deleted_at IS NULL LIMIT 1` confirms index scan instead of sequential/filter scan               
  - [ ] `GET /api/v1/events/:id` responds in < 100ms for large PG-backed orgs

### DIFF
--- a/db/migrate/20260504134804_add_events_organization_transaction_id_index.rb
+++ b/db/migrate/20260504134804_add_events_organization_transaction_id_index.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddEventsOrganizationTransactionIdIndex < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :events,
+      [:organization_id, :transaction_id],
+      where: "deleted_at IS NULL",
+      name: "index_events_on_organization_id_and_transaction_id",
+      algorithm: :concurrently,
+      if_not_exists: true
+  end
+end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -85,11 +85,12 @@ end
 #
 # Indexes
 #
-#  idx_events_billing_lookup                       (external_subscription_id,organization_id,code,timestamp) WHERE (deleted_at IS NULL)
-#  idx_events_for_distinct_codes                   (external_subscription_id,organization_id,timestamp) WHERE (deleted_at IS NULL)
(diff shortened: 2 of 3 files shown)


toommz (approved):

vincent-pochet (approved):

aquinofb (approved):

db/migrate/20260504134804_add_events_organization_transaction_id_index.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddEventsOrganizationTransactionIdIndex < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :events,
+      [:organization_id, :transaction_id],
```
groyoh, line 8: We could save some spaces by only including `transaction_id` since we should have very few identical transaction ids across orgs/subscriptions.

mariohd: I kinda agree with you, but that would require us to add a loop in the query (either Ruby or Database side). 

e.g
**we query by org_id and transaction_id**: the index for transaction_id is used, but we have no guarantee that all transaction_ids are from the same org we are looking, so the database will have to check all tuples for their orgs. It's an (INDEX LOOKUP + LOOP in "K" tuples); O(Log n + K), while having the composite index is just O(Log n). 

Since events is a massive table, I prefer to have just the O(Log n). 

WDYT?

groyoh: To be fair, I believe the K to be close to 1 in most cases. Both work to be honest

PR #5458, opened 2026-05-05
[ING-25] fix(daily-usage): create daily usage based on ingested units and not based on zero amount cents

lovrocolic:
## Context

Currently, the daily usage record is skipped if the current usage amount is zero. However, we should always generate a daily usage record if any usage has been ingested. The only use case where we don't want a daily usage record is when no usage has been ingested at all.

## Description

This PR handles the case described above.

### DIFF
--- a/app/services/daily_usages/compute_service.rb
+++ b/app/services/daily_usages/compute_service.rb
@@ -19,9 +19,9 @@ module DailyUsages
         return result
       end
 
-      if current_usage.total_amount_cents.positive?
-        current_usage.fees = current_usage.fees.select { |f| f.units.positive? }
+      current_usage.fees = current_usage.fees.select { |f| non_empty_fee?(f) }
 
+      if current_usage.fees.any?
         daily_usage = DailyUsage.new(
           organization: subscription.organization,
           customer: subscription.customer,
@@ -94,6 +94,10 @@ module DailyUsages
 
     def usage_date
       @usage_date ||= date_in_timezone - 1.day
+    end
+
+    def non_empty_fee?(fee)
+      fee.units.positive? || fee.amount_cents.positive? || fee.events_count.to_i.positive?
     end
   end
 end
--- a/app/services/daily_usages/compute_service.rb
+++ b/app/services/daily_usages/compute_service.rb
@@ -19,7 +19,7 @@ module DailyUsages
         return result
       end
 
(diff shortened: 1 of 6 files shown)


app/services/daily_usages/compute_service.rb
```diff
@@ -95,5 +95,9 @@ def date_in_timezone
     def usage_date
       @usage_date ||= date_in_timezone - 1.day
     end
+
+    def non_empty_fee?(fee)
+      fee.units.positive? || fee.amount_cents.positive? || fee.events_count.to_i.positive?
```
groyoh: Maybe this could be a method on `Fee` ?

groyoh (approved):
I believe we're missing quite a few test branches here.

app/services/daily_usages/compute_service.rb
```diff
@@ -19,9 +19,9 @@ def call
         return result
       end
 
-      if current_usage.total_amount_cents.positive?
-        current_usage.fees = current_usage.fees.select { |f| f.units.positive? }
+      current_usage.fees = current_usage.fees.select { |f| non_empty_fee?(f) }
```
toommz: nit: instead of using a one letter variable name, I would rather use `_1` or `it` (or keep a variable but name it `fee`)

toommz (approved):

app/services/daily_usages/compute_service.rb
lovrocolic: Yep, moving it there

app/services/daily_usages/compute_service.rb
lovrocolic: It became model method so the call is not the same anymore:)

groyoh (approved):
I believe we're still lacking the test cases where we have a `min_amount_cents` in which case we could have 0 units and 0 events but a positive `amount_cents`.

lovrocolic:
> I believe we're still lacking the test cases where we have a `min_amount_cents` in which case we could have 0 units and 0 events but a positive `amount_cents`.

Hmm, TBH I think that all cases are covered. For `non_zero?` model method we have tests for all cases, including the case when `amount_cents` is positive (in practice that could be due to `min_amount_cents`). In the service layer we tests both branches, when `non_zero?` is true and false.

PR #5459, opened 2026-05-05
fix(bullet): Fix Bullet ignoring N+1 in specs

vankiru:
## Context

Bullet ignores anything created after `Bullet.start_request` and flags it as "impossible." Because we register the Bullet hook at the config level and due to how RSpec orders hooks, it ends up at the very top.

Basically, anything created inside examples gets skipped, which makes the current Bullet tests kinda useless – they're not actually testing what they're supposed to:

```ruby
Config.before -> bullet
# Everything created after this gets ignored by Bullet
Parent.before
Example.before
```

For referrence: https://github.com/flyerhzm/bullet/issues/427

## Description

This PR moves the Bullet hook further down the chain, so it's registered at the example group level. That way, Bullet can actually see everything:

```ruby
Config.before
Parent.before
Example.before -> bullet
```

Also merged `:with_bullet` and `:bullet` into just `:bullet` to keep things simpler.

## Requests vs Resolvers specs

All the request specs with Bullet work as expected and raise an error if we remove associations from `includes`. But all the other specs using Bullet – like graphql resolvers or services – don't actually check anything and never raise errors.

### Requests specs
 Request specs are integration tests, so they go through the full Rails stack, including routing and middleware. That means they also go through the Bullet rack middleware, which turns Bullet on and properly tracks "possible" and "impossible" associations. If it detects an N+1 query, it raises an error directly in the controller before the execution even gets back to the spec, so all the `expect`s are skipped. 

This also means that Bullet settings configured on the RSpec level don't actually affect request specs.

You can see in the callstack for `invoices_request` specs that the error is raised on the Rails level:
```
...
# /usr/local/bundle/gems/bullet-8.1.0/lib/bullet.rb:210:in 'Bullet.perform_out_of_channel_notifications'
# /usr/local/bundle/gems/bullet-8.1.0/lib/bullet/rack.rb:45:in 'Bullet::Rack#call'
...
-> Here we start going through the rails stack
# ./spec/support/api_helper.rb:6:in 'ApiHelper#get_with_token'
# ./spec/requests/api/v1/invoices_controller_spec.rb:357:in 'block (4 levels) in <top (required)>'
```

### Resolvers & other specs

Resolver specs are more like unit tests and don't go through the Rails stack, so here the Bullet settings from `spec_helper` do take effect and all the things described in the PR happen. 

If Bullet is configured properly to catch N+1s in specs, you can see in the callstack that the error is raised from the Bullet after hook, while all the `expect`s in the test still run as normal:

```
...
# /usr/local/bundle/gems/bullet-8.1.0/lib/bullet.rb:210:in 'Bullet.perform_out_of_channel_notifications'
-> Here is where the bullet hooks are defined 
# ./spec/spec_helper.rb:216:in 'block (3 levels) in <top (required)>'
# ./spec/spec_helper.rb:221:in 'block (2 levels) in <top (required)>'
# ./spec/spec_helper.rb:197:in 'block (3 levels) in <top (required)>'
```

### DIFF
--- a/app/graphql/resolvers/credit_notes_resolver.rb
+++ b/app/graphql/resolvers/credit_notes_resolver.rb
@@ -40,6 +40,7 @@ module Resolvers
       includes = [
         :customer,
         :error_details,
+        :metadata,
         {invoice: :billing_entity},
         items: {
           fee: [
--- a/spec/graphql/resolvers/credit_notes_resolver_spec.rb
+++ b/spec/graphql/resolvers/credit_notes_resolver_spec.rb
@@ -345,66 +345,74 @@ RSpec.describe Resolvers::CreditNotesResolver do
     end
   end
 
-  it "does not trigger N+1 queries for metadata", bullet: {unused_eager_loading: false} do
-    credit_notes = create_list(:credit_note, 3, customer:)
-    credit_notes.each do |credit_note|
-      create(:item_metadata, owner: credit_note, organization:, value: {"foo" => "bar"})
+  context "with N+1 queries detection on metadata", bullet: {unused_eager_loading: false} do
+    before do
+      credit_notes = create_list(:credit_note, 3, customer:)
+      credit_notes.each do |credit_note|
+        create(:item_metadata, owner: credit_note, organization:, value: {"foo" => "bar"})
+      end
     end
 
-    execute_graphql(
-      current_user: membership.user,
-      current_organization: organization,
-      permissions: required_permission,
-      query: <<~GQL
-        query {
-          creditNotes(limit: 5) {
-            collection {
-              id
-              metadata { key value }
+    it "does not trigger N+1 queries for metadata" do
+      execute_graphql(
+        current_user: membership.user,
+        current_organization: organization,
+        permissions: required_permission,
+        query: <<~GQL
+          query {
+            creditNotes(limit: 5) {
+              collection {
+                id
+                metadata { key value }
+              }
             }
           }
-        }
-      GQL
-    )
+        GQL
+      )
+    end
   end
 
-  it "does not trigger N+1 queries for items and fees", bullet: {unused_eager_loading: false} do
-    subscription = create(:subscription, customer:, organization:)
+  context "with N+1 queries detection on items and fees", bullet: {unused_eager_loading: false} do
+    before do
+      subscription = create(:subscription, customer:, organization:)
 
-    3.times do
-      inv = create(:invoice, customer:, organization:)
-      sub_fee = create(:fee, invoice: inv, subscription:, organization:)
-      charge_fee = create(:charge_fee, invoice: inv, subscription:, organization:)
+      3.times do
+        inv = create(:invoice, customer:, organization:)
+        sub_fee = create(:fee, invoice: inv, subscription:, organization:)
+        charge_fee = create(:charge_fee, invoice: inv, subscription:, organization:)
 
-      cn = create(:credit_note, customer:, invoice: inv, organization:)
-      create(:credit_note_item, credit_note: cn, fee: sub_fee)
-      create(:credit_note_item, credit_note: cn, fee: charge_fee)
-      create(:credit_note_applied_tax, credit_note: cn, organization:)
+        cn = create(:credit_note, customer:, invoice: inv, organization:)
+        create(:credit_note_item, credit_note: cn, fee: sub_fee)
+        create(:credit_note_item, credit_note: cn, fee: charge_fee)
+        create(:credit_note_applied_tax, credit_note: cn, organization:)
+      end
     end
 
-    execute_graphql(
-      current_user: membership.user,
-      current_organization: organization,
-      permissions: required_permission,
-      query: <<~GQL
-        query {
-          creditNotes(limit: 5) {
-            collection {
-              id
-              appliedTaxes { id taxRate }
-              items {
+    it "does not trigger N+1 queries for items and fees" do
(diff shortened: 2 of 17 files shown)


vincent-pochet (approved):

groyoh:
Nice catch! This is so counter intuitive 😅

Based on the issue, I guess we could still have issues with tests were the models are created within the tests themselves, right ?

groyoh:
@vankiru I did some tests locally by removing the `.includes` and bullet does raise an error in those tests:

```ruby
  1) Api::V1::InvoicesController GET /api/v1/invoices behaves like an invoice index endpoint with N+1 query detection does not trigger N+1 queries on invoice associations
     Failure/Error: get(path, params:, headers:)

     Bullet::Notification::UnoptimizedQueryError:
       user: root
       GET /api/v1/invoices
       USE eager loading detected
         Invoice => [:billing_entity]
         Add to your query: .includes([:billing_entity])
       Call stack
         /app/app/models/invoice.rb:172:in 'Enumerable#map'
         /app/app/models/invoice.rb:172:in 'Invoice.preload_offset_amounts'
         /app/app/controllers/concerns/invoice_index.rb:62:in 'InvoiceIndex#invoice_index'
         /app/app/controllers/api/v1/invoices_controller.rb:53:in 'Api::V1::InvoicesController#index'
         /app/spec/support/api_helper.rb:6:in 'ApiHelper#get_with_token'
         /app/spec/requests/api/v1/invoices_controller_spec.rb:294:in 'block (4 levels) in <top (required)>'
         /app/spec/support/shared_examples/invoice_index.rb:477:in 'block (3 levels) in <top (required)>'
         /app/spec/spec_helper.rb:221:in 'block (2 levels) in <top (required)>'
         /app/spec/spec_helper.rb:213:in 'block (3 levels) in <top (required)>'
         /app/spec/spec_helper.rb:212:in 'block (2 levels) in <top (required)>'
     Shared Example Group: "an invoice index endpoint" called from ./spec/requests/api/v1/invoices_controller_spec.rb:293
     # /usr/local/bundle/gems/uniform_notifier-1.18.0/lib/uniform_notifier/raise.rb:19:in 'UniformNotifier::Raise._out_of_channel_notify'
     # ...
     # ./spec/spec_helper.rb:212:in 'block (2 levels) in <top (required)>'
     # /usr/local/bundle/gems/webmock-3.26.2/lib/webmock/rspec.rb:39:in 'block (2 levels) in <main>'
```

Could you specify in which specific case it was not working please ?

vankiru:
@groyoh Good point! All the request specs with Bullet work as expected and raise an error if we remove associations from `includes`. But all the other specs using Bullet – like graphql resolvers or services – don't actually check anything and never raise errors. For example, `spec/graphql/resolvers/invoices_resolver_spec.rb`.

It took me a while to figure out why this happens, but here's the reason:

#### 1. Requests specs
 Request specs are integration tests, so they go through the full Rails stack, including routing and middleware. That means they also go through the Bullet rack middleware, which turns Bullet on and properly tracks "possible" and "impossible" associations. If it detects an N+1 query, it raises an error directly in the controller before the execution even gets back to the spec, so all the `expect`s are skipped. 

This also means that Bullet settings configured on the RSpec level don't actually affect request specs.

You can see in the callstack for `invoices_request` specs that the error is raised on the Rails level:
```
...
# /usr/local/bundle/gems/bullet-8.1.0/lib/bullet.rb:210:in 'Bullet.perform_out_of_channel_notifications'
# /usr/local/bundle/gems/bullet-8.1.0/lib/bullet/rack.rb:45:in 'Bullet::Rack#call'
...
-> Here we start going through the rails stack
# ./spec/support/api_helper.rb:6:in 'ApiHelper#get_with_token'
# ./spec/requests/api/v1/invoices_controller_spec.rb:357:in 'block (4 levels) in <top (required)>'
```

#### 2. Resolvers & other specs

Resolver specs are more like unit tests and don't go through the Rails stack, so here the Bullet settings from `spec_helper` do take effect and all the things described in the PR happen. 

If Bullet is configured properly to catch N+1s in specs, you can see in the callstack that the error is raised from the Bullet after hook, while all the `expect`s in the test still run as normal:

```
...
# /usr/local/bundle/gems/bullet-8.1.0/lib/bullet.rb:210:in 'Bullet.perform_out_of_channel_notifications'
-> Here is where the bullet hooks are defined 
# ./spec/spec_helper.rb:216:in 'block (3 levels) in <top (required)>'
# ./spec/spec_helper.rb:221:in 'block (2 levels) in <top (required)>'
# ./spec/spec_helper.rb:197:in 'block (3 levels) in <top (required)>'
```

groyoh (approved):
Damn, nice catch 👏

groyoh:
Could you update the PR description with this information please ?

vankiru:
> Could you update the PR description with this information please ?

Updated 👍

vankiru:
> Based on the issue, I guess we could still have issues with tests were the models are created within the tests themselves, right ?

Creating objects inside the test should generally work fine, but I noticed there can be issues with certain factories that assign associations through other associations, or when using Bullet inside service specs, since Bullet registers everything loaded from the db during the test and that can lead to false positives.

For example, the `credit_note` factory assigns organization from customer or invoice and loads them from the db, which ended up being registered by Bullet in `CreditNotesResolver` spec. I just moved the creation of credit notes outside the test into a `before` hook and it worked fine.

Another example I saw was `SendWebhookJob`, which loads some stuff internally – I just stubbed it in the test.

PR #5460, opened 2026-05-05
[ING-39] feat(multi-billing-entity): allow editing customer billing entity

lovrocolic:
## Context

Currently, a customer can be associated with only one billing entity.

## Description

The `multiple billing entity` feature allows customers to be associated with billing objects that belong to different billing entities.

This PR relaxes the rule for updating the billing entity on a customer. The billing entity on the customer is the default value, which can be overridden on a specific billing object such as a subscription or wallet. However, the goal is to allow users to edit this data at any time, which could potentially affect any future invoices.

### DIFF
--- a/app/services/customers/update_service.rb
+++ b/app/services/customers/update_service.rb
@@ -97,7 +97,7 @@ module Customers
       #       external_id,
       #       account_type,
       #       billing_entity_id (gated by editable? unless multi_entity_billing flag is enabled)
-      if args.key?(:billing_entity_code) && (organization.feature_flag_enabled?(:multi_entity_billing) || customer.editable?)
+      if args.key?(:billing_entity_code) && allow_billing_entity_update?
         customer.billing_entity = billing_entity
       end
 
@@ -214,6 +214,10 @@ module Customers
       return true if metadata.count <= ::Metadata::CustomerMetadata::COUNT_PER_CUSTOMER
 
       false
+    end
+
+    def allow_billing_entity_update?
+      organization.feature_flag_enabled?(:multi_entity_billing) || customer.editable?
     end
 
     def assign_premium_attributes(customer, args)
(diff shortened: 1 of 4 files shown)


app/services/customers/update_service.rb
```diff
@@ -96,10 +96,13 @@ def call
       # NOTE: Some fields are not editable if customer is attached to subscriptions:
       #       external_id,
       #       account_type,
-      #       billing_entity_id
+      #       billing_entity_id (gated by editable? unless multi_entity_billing flag is enabled)
+      if args.key?(:billing_entity_code) && (organization.feature_flag_enabled?(:multi_entity_billing) || customer.editable?)
```
vincent-pochet: Maybe you could extract the `(organization.feature_flag_enabled?(:multi_entity_billing) || customer.editable?)` in a method. It would ease the readability :)

app/services/customers/upsert_from_api_service.rb
```diff
@@ -46,7 +46,9 @@ def call
       ActiveRecord::Base.transaction do
         original_tax_values = customer.slice(:tax_identification_number, :zipcode, :country).symbolize_keys
 
-        customer.billing_entity = billing_entity if new_customer || (customer.editable? && params.key?(:billing_entity_code))
+        if new_customer || (params.key?(:billing_entity_code) && (organization.feature_flag_enabled?(:multi_entity_billing) || customer.editable?))
```
vincent-pochet: Same comment with the second part of the condition

vincent-pochet (approved):

toommz (approved):

aquinofb (approved):

app/services/customers/update_service.rb
aquinofb: maybe smth like:
```ruby
args.key?(:billing_entity_code) && allow_billing_entity_update?
```

mariohd (approved):

PR #5461, opened 2026-05-05
feat(payment-gated-subs): activate gated subs on zero-amount invoice

ancorcruz:
## Context

When a payment-gated subscription's invoice ends up with zero total_amount_cents (e.g. a 100% coupon, or a zero-priced plan), the payment activation rule cannot be resolved by collecting payment — there is nothing to charge. Without a shortcut, the subscription stays incomplete indefinitely because no payment webhook ever arrives.

 ## Description

Add a shortcut in Invoices::SubscriptionService and Invoices::CreatePayInAdvanceFixedChargesService that runs after fees are calculated. When the invoice is gated and totals to zero, mark the payment rule as satisfied via Payment::EvaluateService and delegate to ResolveSubscriptionStatusService, which activates the subscription through the standard ActivateService.activate_from_incomplete branch. After this, the subscription is no longer gated and the invoice proceeds through the regular non-gated zero-amount finalization, fires the standard side effects, and the no-op Payments::CreateService.call_async that already short-circuits for zero-amount invoices.

Drop the memoization on subscription_gated? in SubscriptionService so the predicate reflects the post-shortcut state when re-evaluated later in the call.

Add scenario coverage for end-to-end activation without any Stripe interaction, plus service-level tests asserting the rule satisfies and the subscription activates.

### DIFF
--- a/app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
+++ b/app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
@@ -47,6 +47,8 @@ module Invoices
         create_applied_prepaid_credit if should_create_applied_prepaid_credit?
         Invoices::ApplyInvoiceCustomSectionsService.call(invoice:)
 
+        skip_payment_gating_for_zero_amount if subscription.gated? && invoice.total_amount_cents.zero?
+
         invoice.payment_status = invoice.total_amount_cents.positive? ? :pending : :succeeded
         Invoices::TransitionToFinalStatusService.call(invoice:)
         invoice.save!
@@ -84,6 +86,14 @@ module Invoices
 
     attr_reader :subscription, :timestamp, :customer, :organization
     attr_accessor :invoice
+
+    def skip_payment_gating_for_zero_amount
+      Subscriptions::ActivationRules::Payment::EvaluateService.call!(
+        rule: subscription.activation_rules.payment.sole,
+        status: :satisfied
+      )
+      Subscriptions::ActivationRules::ResolveSubscriptionStatusService.call!(subscription:)
+    end
 
     def fixed_charge_events
       @fixed_charge_events ||= subscription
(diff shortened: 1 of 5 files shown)


app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
```diff
@@ -47,6 +47,8 @@ def call
         create_applied_prepaid_credit if should_create_applied_prepaid_credit?
         Invoices::ApplyInvoiceCustomSectionsService.call(invoice:)
 
+        skip_payment_gating_for_zero_amount if subscription.gated? && invoice.total_amount_cents.zero?
```
osmarluz: @ancorcruz it doesn't matter right now, but just to be more precise and avoid issues with the `subscription.activation_rules.payment.sole` call in the future

```suggestion
        skip_payment_gating_for_zero_amount if subscription.payment_gated? && invoice.total_amount_cents.zero?
```

app/services/invoices/subscription_service.rb
```diff
@@ -36,6 +36,8 @@ def call
         )
         Invoices::ApplyInvoiceCustomSectionsService.call(invoice:)
 
+        skip_payment_gating_for_zero_amount if subscription_gated? && invoice.total_amount_cents.zero?
```
osmarluz: @ancorcruz same idea here, with a new method

```suggestion
        skip_payment_gating_for_zero_amount if subscription_payment_gated? && invoice.total_amount_cents.zero?
```

osmarluz (approved):
Just left 2 suggestions, but they can be addressed separately if you want

app/services/invoices/create_pay_in_advance_fixed_charges_service.rb
ancorcruz: you are right, done it. thanks.

PR #5401, opened 2026-04-24
feat(presentation_group_keys): Change stores and aggregators to support breakdowns query

tiagolupepic:
This commit changes the aggregations services and stores, PostgresStore and ClickhouseStore in order to perform a query and return the grouped units based on grouped_by + presentation_by keys.

Before, we initially changed the SumService and PostgresStore#grouped_by only to have the initial step, the changes consisted in add the ability to the method to receive arguments, an array of keys and pass it down to the query. 
Having this, we can reuse the grouped_by to perform different queries with different keys.

This commit is part of presentation group keys initiative and it applies the same previous changes done in this PR https://github.com/getlago/lago-api/pull/5355 to another services and aggregators.

We're also touching in ClickhouseEnrichedStore, but just to keep the method compatible, the changes of queries will be addressed in another PR.

This commit is part of these other commits and belongs to presentation
group keys functionality:
- https://github.com/getlago/lago-api/pull/5332
- https://github.com/getlago/lago-api/pull/5335
- https://github.com/getlago/lago-api/pull/5351
- https://github.com/getlago/lago-api/pull/5355
- https://github.com/getlago/lago-api/pull/5402

PR #5465, opened 2026-05-06
feat(presentation_group_keys): Change ClickhouseEnrichedStore to support presentation_by keys

tiagolupepic:
This commit changes the ClickhouseEnrichedStore to support presentation group keys

Continuing the development of presentation group keys, we're changing the last store ClickhouseEnriched to
also support presentation group keys.
The previous changes were done [in this PR](https://github.com/getlago/lago-api/pull/5401) we're passing it down the unique keys (pricing group keys + presentation group keys) and also changed the Postgres + Clickhouse Store and skipping the ClickhouseEnriched.

Since the enriched store rows in clickhouse are generated by event processor and requires an extra effort, we're making this change in two steps:

First, in this commit we're making the feature available to work with `sorted_properties` column where contains the grouped (pricing keys) and also the presentation group keys. Having this, we're making the feature available to customers that are using the presentation group keys with enriched store. 
We're making a fallback to use the sorted_grouped_by when the grouped_by doesn't contain the presentation group keys. With this, we're still taking advantage of the added column. When contains more columns, we go to `sorted_properties`. 

> The behaviour of deciding what column to use `sorted_grouped_by/sorted_properties` is not happening for `grouped_unique_count` and `grouped_weighted_sum`. To avoid adding more complexity, these queries are always using the `sorted_properties`. 

In a second step, we're going to change the flow of event processor to also enrich the event containing the presentation group keys and then rebuild the events. After this step is done, we're going to visit again the store and change to use the new column added, avoiding these extra conditionals to use or not the sorted_properties.

### DIFF
--- a/app/services/events/stores/clickhouse/weighted_sum_query.rb
+++ b/app/services/events/stores/clickhouse/weighted_sum_query.rb
@@ -55,9 +55,12 @@ module Events
           :with_ctes,
           :charges_duration,
           :events_cte_queries,
-          :grouped_arel_columns,
           :grouped_by_columns,
           to: :store
+
+        def grouped_arel_columns
+          store.grouped_arel_columns(store.grouped_by)
+        end
 
         def group_names
           _, names = grouped_arel_columns
@@ -139,7 +142,7 @@ module Events
               arel_table[:timestamp].as("timestamp"),
               arel_table[:decimal_value].as("difference")
             ],
-            deduplicated_columns: %w[decimal_value]
+            deduplicated_columns: store.grouped_by.present? ? %w[decimal_value sorted_properties] : %w[decimal_value]
           )
 
           events_data = <<-SQL
--- a/app/services/events/stores/clickhouse/weighted_sum_query.rb
+++ b/app/services/events/stores/clickhouse/weighted_sum_query.rb
@@ -56,11 +56,8 @@ module Events
           :charges_duration,
           :events_cte_queries,
           :grouped_by_columns,
+          :grouped_arel_columns,
           to: :store
-
-        def grouped_arel_columns
-          store.grouped_arel_columns(store.grouped_by)
-        end
 
         def group_names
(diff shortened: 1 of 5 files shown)


app/services/events/stores/clickhouse/weighted_sum_query.rb
```diff
@@ -139,7 +142,7 @@ def grouped_events_cte_sql(initial_values)
               arel_table[:timestamp].as("timestamp"),
               arel_table[:decimal_value].as("difference")
             ],
-            deduplicated_columns: %w[decimal_value]
+            deduplicated_columns: store.grouped_by.present? ? %w[decimal_value sorted_properties] : %w[decimal_value]
```
vincent-pochet: This change means that if we rely on `grouped_by` (and not only on `presentation_by`),we will always dedup based on the `sorted_properties`. This will kill the performance advantage we are gaining from the pre-computed column. Is there a way to apply the same logic from other aggregation and rely on the `sorted_properties` only when we strictly need it? (the same comment apply for the unique_count_query as well)

vincent-pochet (commented):
Looks good, only one comments around the possible drawback on performance for `weighted_sum` and `unique_count`

app/services/events/stores/clickhouse/weighted_sum_query.rb
tiagolupepic: thanks for the heads up! I put a comment in the PR description regarding this. Consider that will affect the performance, I addressed the concern in another commit https://github.com/getlago/lago-api/pull/5465/commits/305004f6fe4ab9949d361f5206fa3cedcfe116ab. 
Now, it's only going to use the `sorted_properties` for the presentation breakdowns.

vincent-pochet: Great! Well done

vincent-pochet (approved):

PR #5467, opened 2026-05-06
feat(multi-billing-entity): allow setting subscription billing entity

aquinofb:
## Summary

Today every new subscription inherits its billing entity from the customer record, with no way to override it on a per-subscription basis. Multi-entity workflows need an explicit binding so that the same logical customer can hold subscriptions across different billing entities, each invoiced under its own legal identity.

This change teaches `Subscriptions::CreateService` to accept an explicit billing entity reference: `billing_entity_id` on the GraphQL path and `billing_entity_code` on the REST path. When neither is provided, the new `subscriptions.billing_entity_id` column persists as `NULL`, which is the load-bearing semantic for "no explicit binding, inherit from `customer.billing_entity` at billing time"; this preserves the property that later changes to a customer's default flow through to any subscription that never bound explicitly. When a reference is provided but unknown, the service fails with `not_found_failure!(resource: "billing_entity")`.

The REST permit list now allows `billing_entity_code` on `POST /api/v1/subscriptions` so the value reaches the service. The whole behavior is gated behind the `multi_entity_billing` feature flag; with the flag off the param is silently ignored, which keeps the surface forward-compatible for clients that start sending it before the rollout completes.

The service contract intentionally accepts both shapes already, so the GraphQL input addition can ship as a small follow-up PR with no further service changes.

## Try it locally

1. Pick an organization in your dev environment and enable the flag:
   ```ruby
   org = Organization.find_by(name: "...")
   org.enable_feature_flag!(:multi_entity_billing)
   ```
2. Create a second billing entity on that org so you have something to point at:
   ```ruby
   FactoryBot.create(:billing_entity, organization: org, code: "acme_eu")
   ```
3. Create a subscription bound to that entity via REST:
   ```bash
   curl -X POST http://localhost:3000/api/v1/subscriptions \
     -H "Authorization: Bearer <api_key>" \
     -H "Content-Type: application/json" \
     -d '{"subscription":{"external_customer_id":"<existing>","plan_code":"<existing>","external_id":"sub_eu_1","billing_entity_code":"acme_eu"}}'
   ```
   Expect the response to succeed and `Subscription.find_by(external_id: "sub_eu_1").billing_entity.code` to return `"acme_eu"`.
4. Repeat step 3 without `billing_entity_code` (use `external_id: "sub_inherit_1"`) and verify `billing_entity_id` persists as `NULL`.
5. Repeat step 3 with `billing_entity_code: "does_not_exist"` and expect a 404 response with `code: "billing_entity_not_found"`.
6. Disable the flag (`org.disable_feature_flag!(:multi_entity_billing)`), POST again with a valid `billing_entity_code`, and verify the column persists as `NULL` (param silently ignored).

### DIFF
--- a/app/controllers/api/v1/subscriptions_controller.rb
+++ b/app/controllers/api/v1/subscriptions_controller.rb
@@ -167,6 +167,8 @@ module Api
           .permit(
             :external_customer_id,
             :plan_code,
+            :billing_entity_code,
+            :billing_entity_id,
             :name,
             :external_id,
             :billing_time,
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -129,7 +129,8 @@ module Subscriptions
         external_id:,
         billing_time: billing_time || :calendar,
         ending_at: params[:ending_at],
-        progressive_billing_disabled: params[:progressive_billing_disabled] || false
+        progressive_billing_disabled: params[:progressive_billing_disabled] || false,
+        billing_entity: resolve_billing_entity
       )
 
       if params.key?(:payment_method)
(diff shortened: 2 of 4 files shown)


app/controllers/api/v1/subscriptions_controller.rb
```diff
@@ -167,6 +167,8 @@ def create_params
           .permit(
             :external_customer_id,
             :plan_code,
+            :billing_entity_code,
+            :billing_entity_id,
```
aquinofb, line 171: Even though we're not mass-assigning them, I needed to add these to the permitted parameters because we call the service with `create_params.to_h`, and that strips all non-permitted params.

vincent-pochet (approved):

app/services/subscriptions/create_service.rb
lovrocolic, line 98: Do you think we should extend this service to validate if passed billing entity identifier is correct? TBH I am not sure if this is necessary since we will filter out invalid records but just raising it

lovrocolic, line 189: I think we need to update downgrade flow as well

lovrocolic, line 185: Can we also check if we need to update upgrade flow

mariohd (approved):

lovrocolic (approved):

app/services/subscriptions/create_service.rb
aquinofb: Great catch!

aquinofb: yeah, I think it's alright, given that we will resolve it later on in `.resolve_billing_entity`.

PR #5469, opened 2026-05-06
feat(payment-gated-subs): tax-pending guards and tax-service shortcuts

ancorcruz:
## Context

Built on top of #5461 (zero-amount activation shortcut for SubscriptionService and CreatePayInAdvanceFixedChargesService). Two loose ends from that work plus two scenarios that have been pending since #5375 needed closing:

1. The gated invoice guards in TransitionToFinalStatusService and the zero-amount activation shortcut both treat `total_amount_cents` as "the invoice is zero" without distinguishing between "tax has computed and the result is zero" and "tax is still pending and totals haven't been computed yet". With Anrok or VIES in the mix, the second case happens regularly, and the literal 0 reading would either prematurely finalize a gated invoice or fire the activation shortcut while the tax flow was still in progress.
2. The same zero-amount activation shortcut needed to be applied in PullTaxesAndApplyService and FinalizePendingViesInvoiceService for parity. A free / coupon-zeroed gated subscription with a tax provider integration (Anrok) or pending VIES check would otherwise wait on the payment chain to resolve a payment that will never come.
3. The VIES and provider-tax-failure end-to-end scenarios for gated subscriptions were placeholders since #5375 — `pending` blocks with `raise "not implemented"` — and could now be implemented since the prerequisite wiring is in place.

## Description

**Tax-pending guards**

- `TransitionToFinalStatusService`: extend the gated guard to also keep the invoice `:open` when `tax_pending?`. Totals are not yet computed in that state — falling through to the standard finalize path would treat the invoice as zero-amount and finalize prematurely.
- `Invoices::SubscriptionService` and `Invoices::CreatePayInAdvanceFixedChargesService`: add `!tax_pending?` to the zero-amount activation shortcut condition. The shortcut is meant for invoices whose total is genuinely zero after tax computation, not for invoices waiting on tax resolution.

**Zero-amount activation shortcut in tax services**

- `Invoices::ProviderTaxes::PullTaxesAndApplyService`: after taxes are applied, if the resulting total is zero on a gated invoice, mark the payment activation rule as satisfied and resolve the subscription status. Subscription transitions to active before `TransitionToFinalStatusService` runs, so the invoice falls through to the standard zero-amount finalize/close path — same pattern as in the other invoice services.
- `Invoices::FinalizePendingViesInvoiceService`: same shortcut applied after VIES resolution and tax application.

**End-to-end scenarios**

- Gated subscription with pending VIES check: customer with a pending VIES record → invoice goes `:open` with `tax_status: :pending` → VIES resolves via `ViesCheckJob` → `FinalizePendingViesInvoiceService` applies taxes and triggers payment for the gated case → Stripe success webhook activates the subscription.
- Gated subscription with provider tax failure: Anrok stub returns the failure response → invoice goes `:failed` → `Invoices::RetryService` resets to `:open` and re-enqueues the tax fetch → Anrok stub re-stubbed to succeed → taxes applied, payment triggered → Stripe success webhook activates.

## Tests

- `transition_to_final_status_service_spec.rb`: new context covering gated + tax_pending → keeps `:open`.
- `subscription_service_spec.rb` and `create_pay_in_advance_fixed_charges_service_spec.rb`: new contexts inside the existing gated blocks asserting the zero-amount shortcut does not fire while tax is pending and the invoice stays `:open`.
- `pull_taxes_and_apply_service_spec.rb` and `finalize_pending_vies_invoice_service_spec.rb`: new contexts inside the existing gated blocks constructing zero-amount fees (and a zero-tax Anrok body for the provider tax case) so the shortcut fires through the real flow. Assert rule satisfied + subscription active.
- `payment_gated_activation_spec.rb`: the two `pending` placeholders from #5375 are replaced with full implementations.

## Out of scope

- Late payment refund routing in `Payment::ResolveService` (deferred to comming PR alongside the timeout/expire mechanism it depends on).

### DIFF
--- a/app/services/invoices/finalize_pending_vies_invoice_service.rb
+++ b/app/services/invoices/finalize_pending_vies_invoice_service.rb
@@ -29,6 +29,9 @@ module Invoices
 
         invoice.payment_status = invoice.total_amount_cents.positive? ? :pending : :succeeded
         invoice.tax_status = "succeeded"
+
+        skip_payment_gating_for_zero_amount if invoice.subscription_gated? && invoice.total_amount_cents.zero?
+
         Invoices::TransitionToFinalStatusService.call(invoice:)
 
         invoice.save!
@@ -59,6 +62,17 @@ module Invoices
     private
 
     attr_reader :invoice
+
+    def skip_payment_gating_for_zero_amount
+      gated = invoice.subscriptions.find(&:gated?)
+      return unless gated
+
+      Subscriptions::ActivationRules::Payment::EvaluateService.call!(
+        rule: gated.activation_rules.payment.sole,
+        status: :satisfied
+      )
+      Subscriptions::ActivationRules::ResolveSubscriptionStatusService.call!(subscription: gated)
+    end
 
     def customer
       @customer ||= invoice.customer
--- a/app/services/invoices/finalize_pending_vies_invoice_service.rb
+++ b/app/services/invoices/finalize_pending_vies_invoice_service.rb
@@ -30,7 +30,7 @@ module Invoices
         invoice.payment_status = invoice.total_amount_cents.positive? ? :pending : :succeeded
         invoice.tax_status = "succeeded"
 
-        skip_payment_gating_for_zero_amount if invoice.subscription_gated? && invoice.total_amount_cents.zero?
+        skip_payment_gating_for_zero_amount if invoice.subscription_payment_gated? && invoice.total_amount_cents.zero?
 
         Invoices::TransitionToFinalStatusService.call(invoice:)
 
@@ -64,7 +64,7 @@ module Invoices
     attr_reader :invoice
 
     def skip_payment_gating_for_zero_amount
-      gated = invoice.subscriptions.find(&:gated?)
+      gated = invoice.subscriptions.find(&:payment_gated?)
       return unless gated
 
       Subscriptions::ActivationRules::Payment::EvaluateService.call!(
--- a/app/services/invoices/provider_taxes/pull_taxes_and_apply_service.rb
+++ b/app/services/invoices/provider_taxes/pull_taxes_and_apply_service.rb
@@ -49,6 +49,9 @@ module Invoices
 
           invoice.payment_status = invoice.total_amount_cents.positive? ? :pending : :succeeded
           invoice.tax_status = "succeeded"
+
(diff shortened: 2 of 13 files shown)


app/services/invoices/provider_taxes/pull_taxes_and_apply_service.rb
```diff
@@ -49,6 +49,9 @@ def call
 
           invoice.payment_status = invoice.total_amount_cents.positive? ? :pending : :succeeded
           invoice.tax_status = "succeeded"
+
+          skip_payment_gating_for_zero_amount if invoice.subscription_gated? && invoice.total_amount_cents.zero?
```
osmarluz: @ancorcruz not urgent, but like #5461 it would be good to have a `invoice.subscription_payment_gated?` here. This method doesn't exist so we can handle this is the future no to add extra context for now

app/services/invoices/finalize_pending_vies_invoice_service.rb
```diff
@@ -29,6 +29,9 @@ def call
 
         invoice.payment_status = invoice.total_amount_cents.positive? ? :pending : :succeeded
         invoice.tax_status = "succeeded"
+
+        skip_payment_gating_for_zero_amount if invoice.subscription_gated? && invoice.total_amount_cents.zero?
```
osmarluz: @ancorcruz same idea from [the other comment](https://github.com/getlago/lago-api/pull/5469/changes#r3200813477) here

spec/scenarios/subscriptions/payment_gated_activation_spec.rb
```diff
@@ -220,29 +220,112 @@ def simulate_stripe_webhook(status:)
...
+      subscription = customer.subscriptions.sole
+      expect(subscription).to be_incomplete
+      expect(subscription.activation_rules.sole).to be_pending
+
+      invoice = subscription.invoices.sole
+      expect(invoice).to be_failed
+      expect(invoice.tax_status).to eq("failed")
+
+      # Stage 2: Re-stub Anrok to succeed, then retry. Invoice goes :open with taxes
+      # applied; FinalizePendingViesInvoiceService triggers payment for the gated case.
```
osmarluz: @ancorcruz just a small typo for this scenario:


```suggestion
      # applied; PullTaxesAndApplyService triggers payment for the gated case.
```

osmarluz (approved):
All good, just a couple of small non-critical things 🚀

PR #5474, opened 2026-05-07
[ING-121] feat(clickhouse): Make clickhouse the default event store when enabled

vincent-pochet:
## Description

This PR makes sure that every new organization created on an instance with `LAGO_CLICKHOUSE_ENABLED` env variable will set `clickhouse_events_store` to `true`.

### DIFF
--- a/app/services/organizations/create_service.rb
+++ b/app/services/organizations/create_service.rb
@@ -2,6 +2,8 @@
 
 module Organizations
   class CreateService < BaseService
+    Result = BaseResult[:organization]
+
     def initialize(params)
       @params = params
       super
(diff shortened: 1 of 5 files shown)


app/services/organizations/create_service.rb
```diff
@@ -12,6 +14,11 @@ def call
         params.slice(:name, :document_numbering, :premium_integrations)
       )
 
+      if ActiveModel::Type::Boolean.new.cast(ENV["LAGO_CLICKHOUSE_ENABLED"])
```
groyoh: What about this ?

```suggestion
      if ActiveModel::Type::Boolean.new.cast(ENV["LAGO_CLICKHOUSE_ENABLED"]) && ENV["LAGO_DEFAULT_EVENT_STORE"] == 'clickhouse'
```

groyoh (approved):

PR #5246, opened 2026-03-27
misc(openstruct): Remove taxes result openstructs

vincent-pochet:
## Description

This pull request refactors the tax calculation results in the aggregator integrations, to the `OpenStruct` objects with explicit, typed data structures.
This change improves code clarity, type safety and is one step further in the process of enabling YJIT in the project

PR #5475, opened 2026-05-07
feat(payment-gated-subs): add Adyen payment cancel service

ancorcruz:
## Context

Payment-gated subscriptions need a way to release the open Adyen authorization when activation times out without the customer completing payment, so funds aren't held indefinitely on the PSP side.

## Description

Adds `PaymentProviders::Adyen::Payments::CancelService`, that calls Adyen's modifications API (`cancel_authorised_payment_by_psp_reference`) to cancel the authorised payment by its psp reference. Returns a successful result either way:

- On 4xx responses (typically 422 — payment in a non-cancelable state like already captured or cancelled) the service logs and treats the request as a successful no-op so the timeout/expiration flow does not block on PSP-side cleanup.
- `Adyen::ValidationError` raised by the gem is handled the same way.
- Faraday connection errors propagate as `Invoices::Payments::ConnectionError` so the caller can retry.

Unlike Stripe, Adyen's sync cancel response is an acknowledgment (`"received"`), not a final state — Adyen confirms the actual cancellation asynchronously via the `CANCELLATION` webhook. The Payment record is therefore left untouched in this leaf; the lifecycle transition belongs to the eventual webhook event.

### DIFF
--- a/app/services/payment_providers/adyen/payments/cancel_service.rb
+++ b/app/services/payment_providers/adyen/payments/cancel_service.rb
@@ -17,17 +17,29 @@ module PaymentProviders
             payment.provider_payment_id
           )
 
-          if adyen_result.status >= 400
-            # Best-effort cancel: the payment is in a non-cancelable state
-            # (already captured/cancelled, etc.). Log and treat as a successful
-            # no-op so the caller (timeout/expiration flow) does not block on
-            # PSP-side cleanup.
+          if adyen_result.status == 422
+            # Best-effort cancel only for the "modification cannot apply to
+            # current state" case. Adyen returns 422 for validation/state
+            # errors — most commonly "modification not allowed on transaction
+            # status" when the payment is already captured/cancelled or
+            # otherwise outside the cancellable lifecycle window. Log and
+            # treat as a successful no-op so the caller (timeout/expiration
+            # flow) does not block on PSP-side cleanup.
             Rails.logger.info(
               "Adyen payment not cancelable for payment #{payment.id}: " \
               "status=#{adyen_result.status} message=#{adyen_result.response["message"]}"
             )
             result.payment = payment
             return result
+          end
+
+          if adyen_result.status >= 400
+            # Other 4xx/5xx responses (401 auth, 403 forbidden, 404 not found,
+            # 5xx server errors) propagate so the caller can retry or surface
+            # the failure.
+            raise ::Adyen::AdyenError.new(
+              nil, nil, adyen_result.response["message"], adyen_result.response["errorType"]
+            )
           end
 
           # Adyen's sync cancel response is an acknowledgment ("received"), not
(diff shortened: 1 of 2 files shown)


app/services/payment_providers/adyen/payments/cancel_service.rb
```diff
@@ -0,0 +1,73 @@
...
+        def initialize(payment:)
+          @payment = payment
+          super
+        end
+
+        def call
+          adyen_result = client.checkout.modifications_api.cancel_authorised_payment_by_psp_reference(
+            {merchantAccount: payment.payment_provider.merchant_account},
+            payment.provider_payment_id
+          )
```
osmarluz, line 19: @ancorcruz We should add the idempotency key like we do for the `app/services/payment_providers/adyen/payments/create_service.rb`, no? [The documentation](https://docs.adyen.com/online-payments/cancel?programming_language=rb) also mentions this


```suggestion
          adyen_result = client.checkout.modifications_api.cancel_authorised_payment_by_psp_reference(
            {merchantAccount: payment.payment_provider.merchant_account},
            payment.provider_payment_id,
            headers: {"Idempotency-Key" => "payment-#{payment.id}"}
          )
```

app/services/payment_providers/adyen/payments/cancel_service.rb
```diff
@@ -0,0 +1,73 @@
...
+              "status=#{adyen_result.status} message=#{adyen_result.response["message"]}"
+            )
+            result.payment = payment
+            return result
+          end
+
+          if adyen_result.status >= 400
+            # Other 4xx/5xx responses (401 auth, 403 forbidden, 404 not found,
+            # 5xx server errors) propagate so the caller can retry or surface
+            # the failure.
```
osmarluz: @ancorcruz I think we can drop this comment. The code itself is pretty clear already. WDYT?

spec/services/payment_providers/adyen/payments/cancel_service_spec.rb
osmarluz, line 1: @ancorcruz we should replace the `OpenStruct` on the responses with other approach since we're getting rid of them on the project as a whole. One possible approach is defining value objects with `Data` in the same way it was done for #5246

osmarluz (changes requested):
Leaving the request to remove the `OpenStruct` references. Other comments are up for discussion

app/services/payment_providers/adyen/payments/cancel_service.rb
ancorcruz: Agreed — dropped in 363f257dd. The 422 swallow branch still carries the WHY comment; this one was just enumerating the leftovers.

spec/services/payment_providers/adyen/payments/cancel_service_spec.rb
ancorcruz: Done in 363f257dd — replaced with `AdyenResponse = Data.define(:status, :response)` following the #5246 pattern. Localized to this spec for now; the existing `spec/factories/payment_responses.rb` factories are still on `OpenStruct` and would benefit from the same migration as a separate cleanup.

osmarluz (approved):
🚀

PR #5476, opened 2026-05-07
feat(payments): handle Adyen CANCELLATION webhook event

ancorcruz:
## Context

The Adyen cancel modifications API only acknowledges the cancel synchronously (`"received"`) and confirms the actual cancellation asynchronously through the `CANCELLATION` webhook. Until now Adyen `CANCELLATION` events were not registered in `WEBHOOKS_EVENTS`, so the `HandleEventService` rejected them as invalid. As a result a successful Adyen cancel never propagated back to the Payment record on the Lago side, leaving its lifecycle status stale.

## Description

Register `CANCELLATION` in \`PaymentProviders::AdyenProvider::WEBHOOKS_EVENTS\` and route the event in \`HandleEventService\`:

- The event is identified by `originalReference` (the original payment's PSP reference, which we stored as \`Payment#provider_payment_id\`); `pspReference` on a `CANCELLATION` event is the modification's own id.
- Failed cancellations (\`success != "true"\`) leave the Payment record untouched.
- Successful cancellations look up the Payment, derive the right payments service from its `payable_type`, and dispatch through the existing \`update_payment_status\` pipeline with status \`"Cancelled"\`. \`determine_payment_status\` maps that into \`payable_payment_status :failed\` via \`FAILED_STATUSES\`, fanning out invoice and webhook side effects through the same path used by \`AUTHORISATION\` events.
- Events that reference a payment we don't track are ignored silently.


app/services/payment_providers/adyen/handle_event_service.rb
```diff
@@ -43,6 +43,20 @@ def call
...
+          return result if event["success"] != "true"
+
+          provider_payment_id = event["originalReference"]
+          payment = Payment.find_by(provider_payment_id:) if provider_payment_id
+          return result unless payment
+
+          metadata = {lago_payable_type: payment.payable_type}
+          update_result = payment_service_klass(metadata)
+            .new
+            .update_payment_status(provider_payment_id:, status: "Cancelled", metadata:)
```
osmarluz, line 58: @ancorcruz here we're using `Cancelled` here but on the method `update_payment_status` we set the status to `status = (event["success"] == "true") ? "succeeded" : "failed"`. Shouldn't here be lower-cased as well?

osmarluz (approved):
Approved with a pending question

app/services/payment_providers/adyen/handle_event_service.rb
ancorcruz: no, it is expected as `determine_payment_status` matches the status against the payment provider defined statuses ... `    FAILED_STATUSES = %w[Cancelled CaptureFailed Error Expired Refused].freeze`.

PR #5478, opened 2026-05-07
perf(billable_metrics): Add bulk update charge filters

D1353L:
## Context
The previous implementation discarded each `ChargeFilterValue` one row at a time via `discard!`, issued an existence check per row to decide whether to discard the parent `ChargeFilter`, and then discarded that parent per row when needed. Combined with PaperTrail audit inserts, that's roughly 3 SQL statements + 1 `versions` insert per cfv.

There is a performance issue on production when the organization has 48,952 plans, one billable_metric_filter with 16 values, and ~586,000 charge_filter_values. A single filter mutation could generate up to ~2.3M queries, timing out API request.

## Description
Replaced per-row `update!/discard!` with batched bulk SQL operations.

- **Batched charge_filter_values processing.** `discard_filter_values_in_batches` iterates affected `ChargeFilterValue`s via `in_batches(of: 1_000)`, partitioning each batch into "trim" (cfvs that retain at least one allowed value) and "discard" (cfvs whose values are entirely being removed).
- **Bulk trim.** `bulk_update_trimmed_filter_values` groups charge_filter_values by their resulting `values` array (the intersection with the new allowed set) and executes one `UPDATE … WHERE id IN (…)` per group.
- **Bulk discard charge_filter_values + emptied charge_filters.** `discard_filter_values_and_emptied_charge_filters` flips `deleted_at` on the cfv ids in one statement, then derives the affected `ChargeFilter` ids and discards those with no remaining kept cfvs via a second statement. The "which charge filters are now empty?" computation is done in Ruby with two small `pluck`s and set subtraction (`candidates - charge_filter_ids_with_kept_values`).
- **Pre-flight validation.** Added a guard that short-circuits with `validation_failure!(values: ["value_is_mandatory"])` when any `filter_param[:values]` is blank, preventing partial work that would have to rollback mid-transaction. This is consistent with the original behavior, where ActiveRecord::RecordInvalid was raised in a transaction, resulting in the same API response.

The trade-off matches the one already accepted in `BillableMetricFilters::DestroyAllService`, `ChargeFilters::DestroyService`, and `ChargeFilters::CascadeService`: bulk `update_all` skips PaperTrail callbacks, so individual cfv/cf discards no longer create `versions` rows. Audit history is retained for `BillableMetricFilter`.

## Tests

Expanded the spec from 8 to 27 examples, organized into:

- **Empty-params path** (discard everything): single filter, multiple filters, with/without dependent charge_filter_values, shared charge_filter across multiple BMFs.
- **New-filters path**: creation, refresh-job enqueue, deduplication of input values, validation failure on blank values (single param + one-of-many).
- **Existing-filters path**: value addition, no-op when same values, partial trim with cfvs that mix kept and removed values, mixed trim+discard in the same batch, charge_filter preservation when a sibling filter_value from another BMF is still kept, filter removal with cascade, mixed new/update/remove in one call.
- **Scoping isolation** (new top-level context): with a parallel `other_billable_metric` set up, asserts the bulk SQL doesn't touch records that are not related to the current batch.

## Performance testing
Preconditions:
- 1 billable metric
- 1 filter with 16 filter values
- N Plans using the billable metric.

Action: deleting 6 filter values via UI (GraphQL updateBillableMetric)

| Plans amount (N) | Original latency | Optimized latency | Reduction |
| ------------------ | --------------- | ------------------ | ---------- | 
| 200 | 14400ms (14.4s) | 339ms | 97.6% (42.5× faster) |
| 1000 | Timeout | 1s | N/A |

### DIFF
--- a/app/services/billable_metric_filters/create_or_update_batch_service.rb
+++ b/app/services/billable_metric_filters/create_or_update_batch_service.rb
@@ -120,37 +120,24 @@ module BillableMetricFilters
     end
 
     def bulk_discard_emptied_charge_filters_for(filter_value_ids)
-      charge_filter_ids_to_discard = emptied_charge_filter_ids_for(filter_value_ids)
-      return if charge_filter_ids_to_discard.empty?
-
-      ChargeFilter
-        .where(id: charge_filter_ids_to_discard, deleted_at: nil)
-        .unscope(:order)
-        .update_all(deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
-    end
-
-    def emptied_charge_filter_ids_for(filter_value_ids)
-      charge_filter_ids = charge_filter_ids_referenced_by(filter_value_ids)
-      return [] if charge_filter_ids.empty?
-
-      charge_filter_ids - charge_filter_ids_with_kept_values(charge_filter_ids)
-    end
-
-    def charge_filter_ids_referenced_by(filter_value_ids)
-      ChargeFilterValue
+      charge_filter_ids = ChargeFilterValue
         .with_discarded
         .where(id: filter_value_ids)
         .unscope(:order)
         .distinct
-        .pluck(:charge_filter_id)
-    end
+        .select(:charge_filter_id)
 
-    def charge_filter_ids_with_kept_values(charge_filter_ids)
-      ChargeFilterValue
-        .where(deleted_at: nil, charge_filter_id: charge_filter_ids)
+      return if charge_filter_ids.empty?
+
+      ChargeFilter
+        .where(id: charge_filter_ids, deleted_at: nil)
+        .where.not(
+          id: ChargeFilterValue
+            .where(deleted_at: nil, charge_filter_id: charge_filter_ids)
+            .select(:charge_filter_id)
+        )
         .unscope(:order)
-        .distinct
-        .pluck(:charge_filter_id)
+        .update_all(deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
     end
   end
 end
--- a/app/services/billable_metric_filters/create_or_update_batch_service.rb
+++ b/app/services/billable_metric_filters/create_or_update_batch_service.rb
@@ -2,6 +2,8 @@
 
 module BillableMetricFilters
   class CreateOrUpdateBatchService < BaseService
+    BATCH_SIZE = 1_000
+
     def initialize(billable_metric:, filters_params:)
       @billable_metric = billable_metric
       @filters_params = filters_params
@@ -17,6 +19,8 @@ module BillableMetricFilters
 
         return result
       end
+
+      return result.validation_failure!(errors: {values: ["value_is_mandatory"]}) if any_filter_params_values_blank?
 
       ActiveRecord::Base.transaction do
         filters_params.each do |filter_param|
@@ -35,7 +39,7 @@ module BillableMetricFilters
                   *deleted_values
                 )
 
-              filter_values.each { |filter_value| discard_filter_value(filter_value, new_values:) }
+              discard_filter_values_in_batches(filter_values, new_values:)
             end
           end
 
@@ -60,6 +64,12 @@ module BillableMetricFilters
 
     attr_reader :billable_metric, :filters_params
 
+    def any_filter_params_values_blank?
+      filters_params.any? do |filter_param|
+        filter_param[:values].blank?
+      end
+    end
+
     def discard_all_filters
       ActiveRecord::Base.transaction do
         billable_metric.filters.each { discard_filter(it) }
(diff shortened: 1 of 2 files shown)


rsempe (approved):

app/services/billable_metric_filters/create_or_update_batch_service.rb
```diff
@@ -60,33 +64,93 @@ def call
...
+        .pluck(:charge_filter_id)
+    end
 
-      filter_value.charge_filter.discard! unless filter_value.charge_filter.discarded?
+    def charge_filter_ids_with_kept_values(charge_filter_ids)
+      ChargeFilterValue
+        .where(deleted_at: nil, charge_filter_id: charge_filter_ids)
+        .unscope(:order)
+        .distinct
+        .pluck(:charge_filter_id)
```
tiagolupepic: A refactor can be done here. You can also compose the queries into a single one. Using pluck, you're going to perform the first query. Instead you can use also the `select` which will delegate the query to the DB. You'll ended having a single query `UPDATE ... WHERE id IN (SELECT ...)`. 

Something like this

```
      ChargeFilter
        .where(id: charge_filter_ids, deleted_at: nil)
        .where.not(
          id: ChargeFilterValue
            .where(deleted_at: nil, charge_filter_id: charge_filter_ids)
            .select(:charge_filter_id)
        )
        .unscope(:order)
        .update_all(deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
```

charge_filter_ids are built in the `partition` logic

app/services/billable_metric_filters/create_or_update_batch_service.rb
```diff
@@ -60,33 +64,93 @@ def call
...
-    def discard_filter_value(filter_value, new_values: [])
-      deleted_values = filter_value.values - new_values
+    def discard_filter_values_in_batches(filter_values, new_values: [])
+      filter_values.unscope(:order).in_batches(of: BATCH_SIZE) do |filter_value_batch|
+        values_to_trim, values_to_discard = filter_value_batch.partition { |fv| trimmable?(fv, new_values) }
 
-      if deleted_values.any?
-        values = filter_value.values - deleted_values
+        bulk_update_trimmed_filter_values(values_to_trim, new_values)
+        discard_filter_values_and_emptied_charge_filters(values_to_discard)
```
tiagolupepic, line 90: any chance to have only these two methods? the first one does the update and don't need to follow other method.

The second one can have three statements/queries only. wdyt? Just to avoid having multiple methods

app/services/billable_metric_filters/create_or_update_batch_service.rb
```diff
@@ -60,33 +64,93 @@ def call
...
+      discard_filter_values_in_batches(filter.filter_values)
+
       filter.discard!
     end
 
-    def discard_filter_value(filter_value, new_values: [])
-      deleted_values = filter_value.values - new_values
+    def discard_filter_values_in_batches(filter_values, new_values: [])
+      filter_values.unscope(:order).in_batches(of: BATCH_SIZE) do |filter_value_batch|
+        values_to_trim, values_to_discard = filter_value_batch.partition { |fv| trimmable?(fv, new_values) }
```
tiagolupepic, line 87: using partition is a good call for this!

spec/services/billable_metric_filters/create_or_update_batch_service_spec.rb
```diff
@@ -32,9 +37,30 @@
...
+          create(
+            :charge_filter_value,
+            charge_filter:,
+            billable_metric_filter: other_filter,
+            values: [other_filter.values.first]
+          )
+        end
+
+        it "discards both filters, all filter_values, and the shared charge_filter" do
+          expect { service }.to change { filter.reload.discarded? }.to(true)
```
tiagolupepic: nitpick.. you can also use the `change { ChargeFilter.discarded.count }.from(0).to(x)`

tiagolupepic: btw, my syntax is not accurate 😅

annvelents (approved):

app/services/billable_metric_filters/create_or_update_batch_service.rb
```diff
@@ -60,33 +64,93 @@ def call
...
-          return
-        end
+    def trimmable?(filter_value, new_values)
+      filter_value.values.intersect?(new_values)
+    end
+
+    def bulk_update_trimmed_filter_values(filter_values, new_values)
+      return if filter_values.empty?
+
+      filter_values.group_by { |fv| fv.values & new_values }.each do |result_values, group|
```
tiagolupepic, line 101: not a big thing. group_by will create a hash in memory and we're not using it. Probably a simple each in filter_values can solve the problem here

app/services/billable_metric_filters/create_or_update_batch_service.rb
```diff
@@ -60,33 +64,93 @@ def call
...
+
+    def emptied_charge_filter_ids_for(filter_value_ids)
+      charge_filter_ids = charge_filter_ids_referenced_by(filter_value_ids)
+      return [] if charge_filter_ids.empty?
+
+      charge_filter_ids - charge_filter_ids_with_kept_values(charge_filter_ids)
+    end
+
+    def charge_filter_ids_referenced_by(filter_value_ids)
+      ChargeFilterValue
```
tiagolupepic: one more detail, I think it worth to scope by organization_id these batch update/deletions. Just to avoid the scenario that a not well intentioned user send ids that doesn't belongs to their organization (IDOR)

tiagolupepic: we might need indexes there, btw.

spec/services/billable_metric_filters/create_or_update_batch_service_spec.rb
```diff
@@ -174,6 +305,148 @@
...
+            values: ["US"]
+          )
+        end
+
+        it "discards the filter, its filter_values, and the emptied charge_filter" do
+          service
+
+          expect(filter.reload).to be_discarded
+          expect(filter_value.reload).to be_discarded
+          expect(charge_filter.reload).to be_discarded
```
tiagolupepic, line 327: ditto, small nitpick. can you the `change { scope }` instead

app/services/billable_metric_filters/create_or_update_batch_service.rb
D1353L: this is intentional to group filter_values with the same data for several update_all. also memory usage is limited by batch size

app/services/billable_metric_filters/create_or_update_batch_service.rb
D1353L: billable_metric is already scoped by organization_id. shouldn't be IDOR in this case

tiagolupepic: yes, you're right!

PR #5480, opened 2026-05-08
[ING-44] feat: add core logic for supporting passed billing entity in one-off flow

lovrocolic:
## Context

## Description

This PR adds new logic which supports passing selected billing entity in one-off flow

### DIFF
--- a/app/services/invoices/create_one_off_service.rb
+++ b/app/services/invoices/create_one_off_service.rb
@@ -2,7 +2,7 @@
 
 module Invoices
   class CreateOneOffService < BaseService
-    def initialize(customer:, currency:, fees:, timestamp:, skip_psp: false, voided_invoice_id: nil, payment_method_params: nil, invoice_custom_section: {})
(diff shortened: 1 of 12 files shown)


app/services/invoices/create_one_off_service.rb
```diff
@@ -88,25 +93,43 @@ def call
...
       invoice_result.raise_if_error!
 
       @invoice = invoice_result.invoice
     end
 
+    def resolve_billing_entity
+      if multi_entity_enabled? && billing_entity_id.present?
+        @billing_entity = customer.organization.billing_entities.find_by(id: billing_entity_id)
+        result.not_found_failure!(resource: "billing_entity") if @billing_entity.nil?
+      elsif multi_entity_enabled? && billing_entity_code.present?
```
toommz, line 115: What if both code and id are passed, and don't match? Here we prioritize the id, which is a sound choice, but should we fail when given both?

toommz (approved):
LGTM.

Just a nitpick about the "stamp" wording in specs, but maybe that's a commonly used term in the domain

app/services/invoices/create_one_off_service.rb
lovrocolic: We could, but not sure if this is even possible scenario. `Code` is permitted from the API side and `id` is whitelisted from the GQL side. Is someone directly calls the service, `id` has priority.

PR #5482, opened 2026-05-08
feat(netsuite-integration): Apply find or create logic to invoice netsuite integration

osmarluz:
## Context

NetSuite enforces `tranid` uniqueness on invoices. When the original POST hits a `Net::ReadTimeout`, retries either collide with that uniqueness check (Nango's rejection body gets written into `IntegrationResource.external_id` or an error is thrown) or it duplicate the invoice on Netsuite. A manual re-sync today re-POSTs the same payload and fails the same way — we need to look up the existing NetSuite record first.

## Description

- `Integrations::Aggregator::Invoices::ReconcileService` looks the invoice up by `tranid` (`v1/netsuite/invoices/by-tranid`). NetSuite-only; returns early for other providers or when an `IntegrationResource` already exists. On a hit, it creates the `IntegrationResource` linking the Lago invoice to the NetSuite internal id.

- `Integrations::Aggregator::Invoices::CreateJob#perform` now accepts a `find_first:` kwarg. When `find_first || executions > 1`, the job calls `FindService` first and short-circuits the create if the invoice is already present in NetSuite. The auto-sync path defaults to `false` and keeps the existing `executions > 1`-only behavior, so first attempts pay no extra Nango round-trip.

- `Integrations::Aggregator::Invoices::CreateService` captures `find_first:` in its initializer and forwards it through `call_async` → `CreateJob.perform_later`.

- `Mutations::Integrations::SyncInvoice` calls `CreateService.call_async(invoice:, find_first: true)` so user-triggered re-syncs always look upstream before attempting a create — recovering stuck sync state without a duplicate POST.


app/jobs/integrations/aggregator/invoices/create_job.rb
```diff
@@ -41,7 +41,13 @@ class CreateJob < ApplicationJob
           )
         end
 
-        def perform(invoice:)
+        def perform(invoice:, find_first: false)
+          if find_first || executions > 1
```
osmarluz, line 53: This logic covers 2 scenarios:

- When the job is triggered after invoice finalization, `find_first` will be false (the default value) and we know for sure the invoice doesn't exist on Netsuite for the first execution, so we'll skip trying to find the invoice there. If the job fails, then we run the find for the upcoming attempts
- When the job is triggered manually through the UI, `find_first` will be true, because we cannot guarantee Netsuite hasn't synced successfully. It's common to see invoices synced on Netsuite without a corresponding `integration_resource` created

app/jobs/integrations/aggregator/invoices/create_job.rb
osmarluz, line 1: I plan to rename the job to `FindOrCreateJob` to reflect the new behavior, but I'm not doing this here to avoid increasing the amount of touched files

ancorcruz: maybe `Sync` is a good concept for the job name?

app/jobs/integrations/aggregator/invoices/create_job.rb
```diff
@@ -41,7 +41,13 @@ class CreateJob < ApplicationJob
...
+            find_result = Integrations::Aggregator::Invoices::FindService.call(invoice:)
+            find_result.raise_if_error!
```
ancorcruz:
```suggestion
            find_result = Integrations::Aggregator::Invoices::FindService.call!(invoice:)
```

spec/services/integrations/aggregator/invoices/find_service_spec.rb
```diff
@@ -0,0 +1,184 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Integrations::Aggregator::Invoices::FindService do
+  subject(:service_call) { described_class.call(invoice:) }
+
+  let(:service) { described_class.new(invoice:) }
```
ancorcruz:
```suggestion
```

not used

app/services/integrations/aggregator/invoices/reconcile_service.rb
```diff
@@ -0,0 +1,47 @@
...
+            organization_id: integration.organization_id,
+            integration:,
+            external_id:,
+            syncable_id: invoice.id,
+            syncable_type: "Invoice",
+            resource_type: :invoice
+          )
+
+          result.external_id = external_id
+          result
```
ancorcruz, line 34: Could this also return the `IntegrationResource` that has been found or created by the service?

app/services/integrations/aggregator/invoices/find_service.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Integrations
+  module Aggregator
+    module Invoices
+      class FindService < BaseService
```
ancorcruz: The name lies. It does three things: find upstream → create IntegrationResource → return external_id. "Find" implies a read-only lookup.

  - `ReconcileService` — captures "make local state match upstream"; matches the actual recovery semantics described in the PR.
  - `LinkExistingService` — explicit about creating the link record.
  - `FindOrLinkService` — explicit about both halves.

I'd lean `ReconcileService` because that's the intent: this is a recovery/reconciliation step, not a generic lookup. If you ever need a pure read (e.g., for a webhook handler or admin tool), you'd want a real `FindService` for that — keeping the name available is worth something.


Also, lets define the returned result by this service.

```Result = BaseResult[:external_id, :integration_resource])```

It would self-document the side effect (caller can see "this thing creates an `IntegrationResource` and hands it back"). Also, future callers (logging, telemetry, admin UI) get it for free.

File: spec/services/integrations/aggregator/invoices/find_service_spec.rb [hunk 6]
```diff
@@ -0,0 +1,184 @@
...
+  describe "#call" do
+    context "when the invoice is found upstream" do
+      before do
+        allow(lago_client).to receive(:get).with(headers:, body: request_body, content_type: "application/json").and_return("12345")
+      end
+
+      it "creates an IntegrationResource with the returned external_id" do
+        expect { service_call }.to change(IntegrationResource, :count).by(1)
+
+        integration_resource = IntegrationResource.order(created_at: :desc).first
```
ancorcruz: if service returns the integatration resource in the result this test will improve as we can do `results.integration_resource`...

app/jobs/integrations/aggregator/invoices/create_job.rb
ancorcruz: to ease the future reader understanding I would leave a comment like:

```
  # Look upstream before POSTing in two cases:
  # - `find_first: true`: caller (typically the manual SyncInvoice mutation) suspects
  #   the invoice is stuck because a prior POST may have landed on NetSuite without
  #   us recording the IntegrationResource. Always reconcile before retrying.
  # - `executions > 1`: any retry — a previous attempt contacted NetSuite (or may
  #   have) and the safest assumption is that the record might already exist there.
  #   Skips a duplicate POST that would otherwise fail NetSuite's tranid uniqueness.
```

app/services/integrations/aggregator/invoices/find_service.rb
```diff
@@ -0,0 +1,47 @@
...
+          return result unless provider == "netsuite"
+          return result if integration_resource
+
+          throttle!(:netsuite)
+
+          external_id = http_client.get(headers:, body: {tranid: invoice.number}, content_type: "application/json")
+
+          return result if external_id.blank?
+
+          IntegrationResource.create!(
```
ancorcruz:
```suggestion
          IntegrationResource.find_or_create!(
```

this avoids a race condition between find and create with a network call in between. I would apply the same for the CreateService. This is not a regression but improves the code anyway.

ancorcruz (approved):
Shipping this as-is solves the immediate stuck-invoice problem and the cost is contained. One thought for a follow-up though: the `find_first` kwarg threads control flow through three layers (mutation → service → job → service) and the job ends up encoding the policy via `find_first || executions > 1`. A structurally simpler factoring would be find-on-conflict inside `CreateService` itself — POST first, and if NetSuite/Nango returns a duplicate-tranid error, do the GET by-tranid as a recovery step and create the `IntegrationResource` from that response.

Benefits:
- No `find_first` kwarg, no job changes, no mutation changes.
- The extra Nango call is paid only in the exact case it's needed (duplicate-tranid conflict), not on every retry or every manual re-sync.
- Recovery logic lives next to the code that caused the divergence, so the next person debugging stuck syncs has one file to read.

The catch is that today Nango's rejection body apparently isn't structured enough for us to recognize the duplicate-tranid case reliably (per the PR description — "Nango's rejection body gets written into IntegrationResource.external_id"). If we can either parse that body or get Nango to expose a clearer signal, find-on-conflict becomes strictly better than find-first. Worth a ticket either way.

WDYT?

app/services/integrations/aggregator/invoices/find_service.rb
osmarluz: @ancorcruz makes sense, I'll change that

osmarluz:
> Shipping this as-is solves the immediate stuck-invoice problem and the cost is contained. One thought for a follow-up though: the `find_first` kwarg threads control flow through three layers (mutation → service → job → service) and the job ends up encoding the policy via `find_first || executions > 1`. A structurally simpler factoring would be find-on-conflict inside `CreateService` itself — POST first, and if NetSuite/Nango returns a duplicate-tranid error, do the GET by-tranid as a recovery step and create the `IntegrationResource` from that response.
> 
> Benefits:
> 
> * No `find_first` kwarg, no job changes, no mutation changes.
> * The extra Nango call is paid only in the exact case it's needed (duplicate-tranid conflict), not on every retry or every manual re-sync.
> * Recovery logic lives next to the code that caused the divergence, so the next person debugging stuck syncs has one file to read.
> 
> The catch is that today Nango's rejection body apparently isn't structured enough for us to recognize the duplicate-tranid case reliably (per the PR description — "Nango's rejection body gets written into IntegrationResource.external_id"). If we can either parse that body or get Nango to expose a clearer signal, find-on-conflict becomes strictly better than find-first. Worth a ticket either way.
> 
> WDYT?

@ancorcruz there's one challenge with this approach:

- It's not guaranteed that Netsuite will return a duplicate-tranid error if we POST again after the resource is already present there. We have seen instances where the resource just gets duplicated. I'll double check with @matiasa18. If there's a way to guarantee this, we can switch to this approach

spec/services/integrations/aggregator/invoices/find_service_spec.rb
osmarluz: @ancorcruz it's used by the shared example `it_behaves_like "throttles!", :netsuite`. I've moved it down only to contexts that need this

PR #5486, opened 2026-05-11
[ING-45] feat(multi-billing-entity): support custom billing entity for the invoice preview

lovrocolic:
## Context

## Description

This PR supports custom billing entity for the invoice preview flow.

### DIFF
--- a/app/services/invoices/preview/build_subscription_service.rb
+++ b/app/services/invoices/preview/build_subscription_service.rb
@@ -5,9 +5,10 @@ module Invoices
     class BuildSubscriptionService < BaseService
       Result = BaseResult[:subscriptions]
 
-      def initialize(customer:, params:)
+      def initialize(customer:, params:, billing_entity: nil)
         @customer = customer
         @params = params.presence || {}
+        @billing_entity = billing_entity
         super
       end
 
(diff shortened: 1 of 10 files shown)


aquinofb (approved):
LGTM

app/services/invoices/preview/build_subscription_service.rb
```diff
@@ -30,6 +31,7 @@ def build_subscription
           organization_id: organization.id,
           customer:,
           plan:,
+          billing_entity: effective_billing_entity,
```
aquinofb, line 34: if the billing entity is equal to the customer billing entity, we will pass `nil` here. Is this expected?

lovrocolic: I would say yes. `billing_entity` variable originally comes from the resolved input param. By definition, we want to assign billing entity from the input params to subscription (for non existing customer/subscription). If customer or subscription exist, we will use billing entity from the database.

Before this step, already assigned organization default billing entity to the customer if `billing_entity` is present.

If there is billing entity passed in the params it should be assigned to subscription, but only if it differs from the default one (which is already on customer). If customer holds the same value for billing entity, we should skip it.

PR #5029, opened 2026-02-12
feat(redis): Add support for Redis Sentinel with Sidekiq

groyoh:
## Context

To improve Redis availability, it is possible to rely on [Redis Sentinel](https://redis.io/docs/latest/operate/oss_and_stack/management/sentinel/#asking-sentinel-about-the-state-of-a-master).

## Description

This adds support for Redis Sentinel with the Sidekiq redis.

A `RedisConfigBuilder` was created so it can share configuration. The class builds a base config from environment variables and exposes a chainable `with_options` method for consumer-specific keys (`sidekiq`'s connection pool vs `redis-client` gems). This config does not currently handle the Redis cache and store configuration but could be updated to do so.

This fail-over behaviour was tested locally with a master + replica + 3 sentinels setup:

```ruby
lago-api(dev)> Redis.new.keys('*').count
=> 1325
lago-api(dev)> Sidekiq.redis { |r| r.keys('*').count }
=> 1325
lago-api(dev)> ActiveJob::Uniqueness.config.redlock_servers.first.call(:keys, '*').count
=> 1325
# Run `lago down redis`
lago-api(dev)> Redis.new.keys('*').count
(lago-api):5:in '<main>': getaddrinfo: Name or service not known (redis://redis:6379) (Redis::CannotConnectError)
/usr/local/lib/ruby/3.4.0/socket.rb:891:in 'block in Socket.tcp_with_fast_fallback': getaddrinfo: Name or service not known (redis://redis:6379) (RedisClient::CannotConnectError)
        ...
        from /usr/local/lib/ruby/3.4.0/socket.rb:712:in 'block (2 levels) in Socket.tcp_with_fast_fallback'
lago-api(dev)> Sidekiq.redis { |r| r.keys('*').count }
=> 1325
lago-api(dev)> ActiveJob::Uniqueness.config.redlock_servers.first.call(:keys, '*').count
=> 1325
```

PR #5489, opened 2026-05-11
ING-96 feat(cache): Add support for Redis Sentinel on the cache

mariohd:
Sidekiq has been talking Redis Sentinel through `Lago::RedisConfigBuilder` since #5029, but the Rails cache layer was still pinned to a single Redis URL. This PR brings the cache up to parity so operators running an HA Redis topology can route the cache through Sentinel too.

The bulk of the work is a new `#cache` method on `Lago::RedisConfigBuilder` parallel to the existing `#sidekiq`. The two methods share parameterized private helpers (`add_sentinels(config, sentinels:, master_name:, sentinel_password: nil)` and `add_password(config, password:)`), so the env-var contract for each consumer lives at the call site of its public method while the parse/assemble logic is shared. The cache method reads `LAGO_REDIS_CACHE_URL`, `LAGO_REDIS_CACHE_PASSWORD`, `LAGO_REDIS_CACHE_SENTINELS`, `LAGO_REDIS_CACHE_MASTER_NAME`, and `LAGO_REDIS_CACHE_SENTINEL_PASSWORD`.

The three environment files (`development.rb`, `staging.rb`, `production.rb`) now build the cache_store hash via the builder instead of inline. A small class method `Lago::RedisConfigBuilder.cache_enabled?` centralizes the gate that decides whether to configure the Redis cache_store at all — it returns true when either `LAGO_REDIS_CACHE_URL` or `LAGO_REDIS_CACHE_SENTINELS` is set, so a Sentinel-only deployment (where the URL is discovered dynamically from the cluster) successfully activates the cache. This also brings staging to parity with production — before this PR staging was URL-only.

## Before and after

|  | Before | After |
| --- | --- | --- |
| Cache Sentinel support | Not available | `LAGO_REDIS_CACHE_SENTINELS`, `_MASTER_NAME`, `_SENTINEL_PASSWORD` |
| Cache config location | Inline in each env file | Single builder + shared `cache_enabled?` gate |
| Builder structure | Sidekiq-only `#sidekiq` method | Parallel `#sidekiq` and `#cache`, parameterized private helpers |
| Staging cache gate | `URL` only | `URL` or `SENTINELS` (matches production) |
| Sentinel-only deployments | Cache silently inert | Activates correctly without a static URL |

## Test plan

- [ ] `bundle exec rspec spec/lib/lago/redis_config_builder_spec.rb`
- [ ] In a Sentinel-enabled env, set `LAGO_REDIS_CACHE_SENTINELS` (without `LAGO_REDIS_CACHE_URL`) and confirm the cache connects through Sentinel discovery.
- [ ] `Rails.application.config.cache_store` remains `:null_store` under `RAILS_ENV=test`.
- [ ] Manual: set `LAGO_REDIS_CACHE_SENTINEL_PASSWORD` and confirm the resulting config carries `sentinel_password`.

## Migration

No env var changes required for existing deployments. The new env vars (`LAGO_REDIS_CACHE_SENTINELS`, `_MASTER_NAME`, `_SENTINEL_PASSWORD`) are only consulted when set. Sidekiq's TLS behavior is unchanged.

**Behavior delta worth knowing**: staging previously only activated the Redis cache when `LAGO_REDIS_CACHE_URL` was set; it now also activates on `LAGO_REDIS_CACHE_SENTINELS` alone, matching production. Most staging operators have the URL set anyway and will see no difference; only Sentinel-only staging setups change behavior, and the change is "cache now works" rather than a regression.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -2,6 +2,7 @@
 
 require "active_support/core_ext/integer/time"
 require "opentelemetry/sdk"
+require "lago/redis_config_builder"
 
 Rails.application.configure do
   config.middleware.use(ActionDispatch::Cookies)
@@ -53,22 +54,16 @@ Rails.application.configure do
     config.cache_store = :mem_cache_store, ENV["LAGO_MEMCACHE_SERVERS"].split(",")
 
   elsif ENV["LAGO_REDIS_CACHE_URL"].present?
-    cache_store_config = {
-      url: ENV["LAGO_REDIS_CACHE_URL"],
-      ssl_params: {
-        verify_mode: OpenSSL::SSL::VERIFY_NONE
-      },
-      pool: {size: ENV.fetch("LAGO_REDIS_CACHE_POOL_SIZE", 5)},
-      error_handler: lambda { |method:, returning:, exception:|
-        Rails.logger.warn(exception.message)
+    cache_store_config = Lago::RedisConfigBuilder.new
+      .with_options(
+        pool: {size: ENV.fetch("LAGO_REDIS_CACHE_POOL_SIZE", 5)},
+        error_handler: lambda { |method:, returning:, exception:|
+          Rails.logger.warn(exception.message)
 
-        Sentry.capture_exception(exception, level: :warning)
-      }
-    }
-
-    if ENV["LAGO_REDIS_CACHE_PASSWORD"].present? && !ENV["LAGO_REDIS_CACHE_PASSWORD"].empty?
-      cache_store_config = cache_store_config.merge({password: ENV["LAGO_REDIS_CACHE_PASSWORD"]})
-    end
+          Sentry.capture_exception(exception, level: :warning)
+        }
+      )
+      .cache
 
     config.cache_store = :redis_cache_store, cache_store_config
   end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -53,7 +53,7 @@ Rails.application.configure do
   if ENV["LAGO_MEMCACHE_SERVERS"].present?
     config.cache_store = :mem_cache_store, ENV["LAGO_MEMCACHE_SERVERS"].split(",")
 
-  elsif ENV["LAGO_REDIS_CACHE_URL"].present? || ENV["LAGO_REDIS_CACHE_SENTINELS"].present?
+  elsif Lago::RedisConfigBuilder.cache_enabled?
     cache_store_config = Lago::RedisConfigBuilder.new
       .with_options(
         pool: {size: ENV.fetch("LAGO_REDIS_CACHE_POOL_SIZE", 5)},
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -53,7 +53,7 @@ Rails.application.configure do
   if ENV["LAGO_MEMCACHE_SERVERS"].present?
     config.cache_store = :mem_cache_store, ENV["LAGO_MEMCACHE_SERVERS"].split(",")
 
-  elsif ENV["LAGO_REDIS_CACHE_URL"].present?
+  elsif ENV["LAGO_REDIS_CACHE_URL"].present? || ENV["LAGO_REDIS_CACHE_SENTINELS"].present?
     cache_store_config = Lago::RedisConfigBuilder.new
       .with_options(
         pool: {size: ENV.fetch("LAGO_REDIS_CACHE_POOL_SIZE", 5)},
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -2,6 +2,7 @@
 
 require "active_support/core_ext/integer/time"
 require "opentelemetry/sdk"
+require "lago/redis_config_builder"
 
 Rails.application.configure do
   # Used for GraphiQL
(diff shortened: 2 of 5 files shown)


lib/lago/redis_config_builder.rb
```diff
@@ -29,44 +53,56 @@ def with_options(options)
...
     end
 
-    def sidekiq
-      redis_config = {
-        url:,
-        ssl_params: {
-          verify_mode: OpenSSL::SSL::VERIFY_NONE
-        }
-      }.compact
+    def build
```
groyoh: To be fair, I don't think we should have a generic config builder here. It makes it much hard to review whether the changes are safe or not. I believe we should have a first iteration with a `def cache` method, the same way we built the `def sidekiq` method. It's not beautiful but at least we're sure there's no impact on the Sidekiq config and it also us to test it independently as well.

Does that make sense ?

mariohd: Sure, that's not a problem for me. 

I can change to reflect that format

mariohd: @groyoh I did the changes, can you please review it again? 

Thanks!

config/environments/production.rb
```diff
@@ -53,22 +54,16 @@
     config.cache_store = :mem_cache_store, ENV["LAGO_MEMCACHE_SERVERS"].split(",")
 
   elsif ENV["LAGO_REDIS_CACHE_URL"].present?
```
groyoh: I believe this should check whether we have `LAGO_REDIS_CACHE_URL` or `LAGO_REDIS_CACHE_SENTINELS` present.

lib/lago/redis_config_builder.rb
```diff
@@ -43,6 +43,20 @@ def sidekiq
       redis_config.merge(extra_options)
     end
 
+    def cache
```
groyoh, line 61: The class-level RDoc (lines 6-21) still describes a sidekiq-only API and both `@example` blocks call `.sidekiq`. Now that `#cache` is part of the public surface, an extra sentence in the description and an `@example` for the cache consumer would help future readers find this path.

lib/lago/redis_config_builder.rb
```diff
@@ -56,17 +70,34 @@ def add_sentinels(config)
...
-    end
-
     def add_password(config)
       password = ENV["REDIS_PASSWORD"].presence
       return unless password
 
       config[:password] = password
     end
 
+    def add_cache_sentinels(config)
```
groyoh: `add_cache_sentinels` and `add_cache_password` are near-copies of `add_sentinels` and `add_password`, differing only in the env-var prefix plus the new `sentinel_password` handling. Parameterizing a shared helper (e.g. `add_sentinels(config, prefix:, with_sentinel_password: false)`) would keep the two consumers from drifting and centralize the parse rules.

It also surfaces a parity question worth calling out: the cache path supports `LAGO_REDIS_CACHE_SENTINEL_PASSWORD`, but the sidekiq path has no equivalent. If that asymmetry is intentional for now, mentioning it in the docstring would prevent the next contributor from assuming it is an oversight.

spec/lib/lago/redis_config_builder_spec.rb
```diff
@@ -155,6 +161,154 @@
     end
   end
 
+  describe "#cache" do
+    subject(:result) { builder.cache }
```
groyoh, line 165: The `#with_options` describe block further down (lines 312-340) only exercises the sidekiq chain. The development environment now relies on `Lago::RedisConfigBuilder.new.with_options(db: ...).cache`, so the `with_options(...).cache` path, particularly the `ssl_params` override case, is not covered by any spec. Mirroring the four existing `with_options` examples against `.cache` would close that gap.

groyoh (changes requested):

config/environments/production.rb
mariohd: I believe `LAGO_REDIS_CACHE_SENTINELS` is optional. The only really mandatory is the `LAGO_REDIS_CACHE_URL` which is always used. 

```
    def add_cache_sentinels(config)
      sentinels = ENV["LAGO_REDIS_CACHE_SENTINELS"].presence
      return unless sentinels

      config[:sentinels] = parse_sentinels(sentinels)
      config[:role] = :master
      config[:name] = ENV.fetch("LAGO_REDIS_CACHE_MASTER_NAME", "master").presence || "master"

      sentinel_password = ENV["LAGO_REDIS_CACHE_SENTINEL_PASSWORD"].presence
      if sentinel_password
        config[:sentinel_password] = sentinel_password
      end
    end
```

groyoh: If `LAGO_REDIS_CACHE_SENTINELS` is set, we actually shouldn't have `LAGO_REDIS_CACHE_URL` as it will conflict (sentinels are providing the master url). So I believe the current configuration doesn't actually work unless we add
```suggestion
  elsif ENV["LAGO_REDIS_CACHE_URL"].present? || ENV["LAGO_REDIS_CACHE_SENTINELS"].present? || 
```

mariohd: I will perform this change

config/environments/staging.rb
```diff
@@ -50,17 +51,18 @@
     config.cache_store = :mem_cache_store, ENV["LAGO_MEMCACHE_SERVERS"].split(",")
 
   elsif ENV["LAGO_REDIS_CACHE_URL"].present?
```
groyoh: We're missing the same check as production here. Maybe we should have a `Lago::RedisConfigBuilder.cache_enabled?`

groyoh (changes requested):

config/environments/staging.rb
mariohd: Good suggestion, I will add this method so it becomes clear to who's reading.

lib/lago/redis_config_builder.rb
```diff
@@ -31,37 +47,62 @@ def with_options(options)
...
         }
       }.compact
 
-      add_sentinels(redis_config)
-      add_password(redis_config)
+      add_sentinels(
+        redis_config,
+        sentinels: ENV["LAGO_REDIS_CACHE_SENTINELS"].presence,
+        master_name: ENV.fetch("LAGO_REDIS_CACHE_MASTER_NAME", "master").presence,
+        sentinel_password: ENV["LAGO_REDIS_CACHE_SENTINEL_PASSWORD"].presence
```
groyoh: I believe we can skip this for now as it is not present in the Sidekiq config.

groyoh (approved):

groyoh (approved):

PR #5491, opened 2026-05-11
chore(charge-filters): Remove dead filter cascade

rsempe:
## Context

PR #5448 extracted `ChargeFilters::CascadeDispatcher` and made every filter create / update / destroy flow through per-filter `ChargeFilters::CascadeJob`, including the bulk update path. That left a second, parallel cascade implementation inside `ChargeFilters::CreateOrUpdateBatchService` doing the same work — dead code by the time this PR lands.

  ## Description

Remove the legacy in-the-lock filter cascade. Behavior is unchanged: every filter mutation still goes through `ChargeFilters::CascadeJob`, and group-key cascading on customized child filters continues via `ChargeFilters::CascadeService#cascade_group_keys`.


app/services/charge_filters/create_or_update_batch_service.rb
```diff
@@ -107,27 +76,15 @@ def call
...
-    end
-
     def remove_all
       ActiveRecord::Base.transaction do
-        if cascade_updates
-          charge.filters.where(id: inherited_filter_ids).unscope(:order).find_each { remove_filter(it) }
-        else
-          charge.filters.each { remove_filter(it) }
-        end
+        charge.filters.each { remove_filter(it) }
```
tiagolupepic, line 87: still have an improvement here, if the scope is quite big. It's doing a O(N x M) and it will take a bit of time. You can change the query to update the ChargeFilterValue using a join with charge_id.

just for the record, if we a charge has a considerable amount of filters + values.

tiagolupepic (approved):
:shipit:

app/services/charge_filters/create_or_update_batch_service.rb
rsempe: good spot 👍 - will do it in a separate PR

PR #5492, opened 2026-05-12
feat(presentation_group_keys): Update usages to expose presentation breakdowns in filters

tiagolupepic:
This commit changes the API and GraphQL responses to include presentation breakdowns in filters

Continuing the work of the epic of presentation group keys, we're including the presentation breakdowns
in filters, for the API and GraphQL.
Since ChargeFilter doesn't have any presentation group keys because they're related more for the pricing,
they're only inheriting the presentation group keys from the Charge, this means, the filters are applied and we
compute the presentation breakdowns using the keys from the charge, having more consistent structure of 
charge and filters.

This commit also include some small refactors:
- Removed the OpenStruct from usage specs in favor of FactoryBot
- Removed mocks of Time and Date in favor of Timecop
- Moved the logic in PresentationBreakdownBuilder to Fee in two methods


app/graphql/types/customers/usage/charge_filter.rb
```diff
@@ -26,6 +27,10 @@ def pricing_unit_amount_cents
         def invoice_display_name
           object.charge_filter&.invoice_display_name
         end
+
+        def presentation_breakdowns
+          Types::Fees::PresentationBreakdownBuilder.call(object, filter: Types::Fees::PresentationBreakdownBuilder::ALL, filter_breakdown: V1::Customers::PresentationBreakdownBuilder::ALL)
```
D1353L: Mixing constants from different modules.
For consistency - `Types::Fees::PresentationBreakdownBuilder::ALL`

(same for projected_charge_filter.rb)

tiagolupepic: thanks! I changed that

D1353L (approved):
LGTM 👍

PR #5493, opened 2026-05-12
chore(payments): Improve payments index performances

jdenquin:
- Remove usage of `for_organization` scope that is doing joins and not using `organization_id`, asking for a full table scan each time.
- Use correct `organization_id` filter

### DIFF
--- a/app/queries/payments_query.rb
+++ b/app/queries/payments_query.rb
@@ -24,8 +24,26 @@ class PaymentsQuery < BaseQuery
 
   def base_scope
     Payment.where.not(customer_id: nil)
-      .for_organization(organization)
+      .where(organization:)
+      .where.not(payable_id: nil)
+      .where(visible_payable_condition)
       .ransack(search_params)
+  end
+
+  def visible_payable_condition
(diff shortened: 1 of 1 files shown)


app/queries/payments_query.rb
```diff
@@ -24,10 +24,28 @@ def filters_contract
...
+  def visible_payable_condition
+    ActiveRecord::Base.sanitize_sql_array([
+      <<~SQL.squish,
+        CASE payments.payable_type
+          WHEN 'Invoice' THEN EXISTS(
+            SELECT 1 FROM invoices
+            WHERE invoices.id = payments.payable_id
+            AND invoices.status IN (:visible_statuses)
+          )
+          ELSE TRUE
```
annvelents, line 43: aaah, actually, nice!!
heh, I guess we needed it before we had organization_id on the payment itself 🙈

annvelents (approved):
🚀

app/queries/payments_query.rb
jdenquin: yes exactly!

PR #5495, opened 2026-05-12
fix(invoices): race condition on void calls

endenis:
## Context

Two concurrent `POST /invoices/:id/void` requests could both pass the initial in-memory `invoice.voided?` check before either had committed, causing the full void path to run twice on the same invoice.

## Description

This fix wraps the void path in `invoice.with_lock` and moves the `invoice.voided?` check inside this lock.

The first request runs normally.

Any concurrent request reloads the row after acquiring the lock, sees it's already voided, and returns `not_voidable` without running side effects.

### DIFF
--- a/app/services/invoices/void_service.rb
+++ b/app/services/invoices/void_service.rb
@@ -18,7 +18,6 @@ module Invoices
 
     def call
       return result.not_found_failure!(resource: "invoice") unless invoice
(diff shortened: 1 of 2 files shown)


rsempe (approved):
👍

annvelents (approved):

tiagolupepic (approved):
:shipit:

PR #5498, opened 2026-05-12
feat(multi-billing-entity): move subscription

aquinofb:
## Summary

Operators can now move an active subscription to a different billing entity through a normal `PATCH /api/v1/subscriptions/:external_id` request, instead of having to terminate and recreate it. From the next billing cycle on, invoices for that subscription land under the new entity (using its number prefix and sequence); invoices already generated stay where they were.

The change is gated behind the `multi_entity_billing` feature flag, so single-entity organizations see no behavior change at all.

## What changed

| Concern | Before | After |
|---|---|---|
| Moving a subscription to another billing entity | Terminate and recreate | `PATCH` with `billing_entity_id` or `billing_entity_code` |
| Where invoice rows read their billing entity from | `customer.billing_entity` | `subscription.billing_entity` (falls back to customer's) |
| Where fees read their billing entity from | `subscription.customer.billing_entity_id` | `subscription.billing_entity_id` (falls back to customer's) |
| Subscriptions with `NULL` `billing_entity_id` (every row in prod today) | Stamped with customer's entity | Same; the column fallback keeps stamping byte-identical |
| Customer with multiple subs on different entities billed together | Not reachable | Returns a `mixed_billing_entities` validation failure; never produces a mixed invoice |
| Feature flag off | Param wasn't accepted | Param is silently ignored, matching the create-side convention |

## How the resolver works

`Subscriptions::UpdateService` resolves the entity inline. It accepts either an id or a code on the same request; if both are sent, the id wins. Unknown values return a 404 with `billing_entity_not_found`. Cross-organization references resolve to "not found" because the lookup is scoped to the subscription's organization.

For the read side, every fee and invoice writer now prefers `subscription.billing_entity_id || subscription.customer.billing_entity_id`. Because every existing subscription row has a `NULL` override, the fallback path is taken and the behavior is identical to today. Once an operator flips the column, the next invoice and its fees follow.

The one read site that intentionally stays on `subscription.customer.billing_entity` is the previous-period proration lookup in `Fees::FixedChargeService`, because legacy paid fees were written against the customer's entity and the query needs to match them.

## Out of scope

The `subscription.updated` webhook payload doesn't surface the new `billing_entity_id` yet; that lives with the serializer and OpenAPI work. Batched billing across mixed-entity subscriptions returns a validation error today; the proper fix is to split the batch upstream in `OrganizationBillingService` and is deferred.

## Test plan

- [ ] Seed an organization with `multi_entity_billing` in `feature_flags`, plus two billing entities `us` (default) and `eu`. Seed a customer under `us` and an active subscription on a monthly recurring plan.
- [ ] `PATCH /api/v1/subscriptions/:external_id` with `{ subscription: { billing_entity_code: "eu" } }`. Expect `200` and verify the `subscriptions` row's `billing_entity_id` now points at `eu`; the `customers` row is unchanged.
- [ ] Force the next billing cycle (move the clock to the calendar boundary or invoke `BillSubscriptionJob` directly for the subscription). Verify the resulting invoice's `billing_entity_id` is `eu`, its number draws from `eu`'s sequence and prefix, and any previously-generated invoices are untouched.
- [ ] Toggle `multi_entity_billing` off on the organization. `PATCH` the subscription again with `billing_entity_code: "us"`. Expect `200` and no change to the row.
- [ ] Re-enable the flag. `PATCH` with `billing_entity_id` set to a random UUID. Expect `404` with `code: "billing_entity_not_found"`. `PATCH` with `billing_entity_id` referencing an entity from a different organization. Expect the same.
- [ ] Create a second subscription on the same customer and `PATCH` it to `eu`, leaving the original on `us`. Trigger their next billing day. Expect a validation failure with `billing_entity: ["mixed_billing_entities"]` rather than a mixed invoice.

### DIFF
--- a/app/services/fees/build_pay_in_advance_fixed_charge_service.rb
+++ b/app/services/fees/build_pay_in_advance_fixed_charge_service.rb
@@ -98,7 +98,7 @@ module Fees
 
       Fee.new(
         organization:,
-        billing_entity_id: subscription.billing_entity_id || subscription.customer.billing_entity_id,
+        billing_entity_id: subscription.applicable_billing_entity_id,
         subscription:,
         fixed_charge:,
         amount_cents:,
@@ -141,7 +141,7 @@ module Fees
     def build_zero_amount_fee(boundaries)
       Fee.new(
         organization:,
-        billing_entity_id: subscription.billing_entity_id || subscription.customer.billing_entity_id,
+        billing_entity_id: subscription.applicable_billing_entity_id,
         subscription:,
         fixed_charge:,
         amount_cents: 0,
--- a/app/services/fees/build_pay_in_advance_fixed_charge_service.rb
+++ b/app/services/fees/build_pay_in_advance_fixed_charge_service.rb
@@ -98,7 +98,7 @@ module Fees
 
       Fee.new(
         organization:,
-        billing_entity_id: subscription.customer.billing_entity_id,
+        billing_entity_id: subscription.billing_entity_id || subscription.customer.billing_entity_id,
         subscription:,
         fixed_charge:,
         amount_cents:,
@@ -141,7 +141,7 @@ module Fees
     def build_zero_amount_fee(boundaries)
       Fee.new(
         organization:,
-        billing_entity_id: subscription.customer.billing_entity_id,
+        billing_entity_id: subscription.billing_entity_id || subscription.customer.billing_entity_id,
         subscription:,
         fixed_charge:,
         amount_cents: 0,
--- a/app/services/subscriptions/update_service.rb
+++ b/app/services/subscriptions/update_service.rb
@@ -65,8 +65,8 @@ module Subscriptions
         end
 
         if params.key?(:billing_entity_id) || params.key?(:billing_entity_code)
-          resolved = resolve_billing_entity
-          subscription.billing_entity = resolved if resolved
+          new_billing_entity = resolve_billing_entity
+          subscription.billing_entity = new_billing_entity if new_billing_entity
         end
 
         subscription.plan = handle_plan_override.plan if params.key?(:plan_overrides)
@@ -158,14 +158,11 @@ module Subscriptions
     def resolve_billing_entity
       return nil unless subscription.organization.feature_flag_enabled?(:multi_entity_billing)
 
-      attrs = if params[:billing_entity_id].present?
-        {id: params[:billing_entity_id]}
+      if params[:billing_entity_id].present?
+        subscription.organization.billing_entities.find(params[:billing_entity_id])
       elsif params[:billing_entity_code].present?
-        {code: params[:billing_entity_code]}
+        subscription.organization.billing_entities.find_by!(code: params[:billing_entity_code])
       end
-      return nil unless attrs
-
-      subscription.organization.billing_entities.find_by!(attrs)
     rescue ActiveRecord::RecordNotFound
       result.not_found_failure!(resource: "billing_entity").raise_if_error!
     end
--- a/app/services/subscriptions/update_service.rb
+++ b/app/services/subscriptions/update_service.rb
@@ -64,6 +64,11 @@ module Subscriptions
           subscription.payment_method_id = params[:payment_method][:payment_method_id] if params[:payment_method].key?(:payment_method_id)
         end
 
+        if params.key?(:billing_entity_id) || params.key?(:billing_entity_code)
+          resolved = resolve_billing_entity
+          subscription.billing_entity = resolved if resolved
+        end
+
         subscription.plan = handle_plan_override.plan if params.key?(:plan_overrides)
 
         if params.key?(:usage_thresholds)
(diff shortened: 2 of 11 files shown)


mariohd (approved):
Looks good

app/services/fees/build_pay_in_advance_fixed_charge_service.rb
```diff
@@ -98,7 +98,7 @@ def build_delta_fee(delta_units, boundaries)
 
       Fee.new(
         organization:,
-        billing_entity_id: subscription.customer.billing_entity_id,
+        billing_entity_id: subscription.billing_entity_id || subscription.customer.billing_entity_id,
```
toommz: It could be a method on `Subscription` as this logic is used multiple times in this PR

app/services/subscriptions/update_service.rb
```diff
@@ -64,6 +64,11 @@ def call
           subscription.payment_method_id = params[:payment_method][:payment_method_id] if params[:payment_method].key?(:payment_method_id)
         end
 
+        if params.key?(:billing_entity_id) || params.key?(:billing_entity_code)
+          resolved = resolve_billing_entity
```
toommz: nit; `resolved` is not really descriptive

app/services/subscriptions/update_service.rb
```diff
@@ -150,6 +155,21 @@ def process_subscription_at_change(subscription)
...
+      return nil unless subscription.organization.feature_flag_enabled?(:multi_entity_billing)
+
+      attrs = if params[:billing_entity_id].present?
+        {id: params[:billing_entity_id]}
+      elsif params[:billing_entity_code].present?
+        {code: params[:billing_entity_code]}
+      end
+      return nil unless attrs
+
+      subscription.organization.billing_entities.find_by!(attrs)
```
toommz:
```suggestion
      if params[:billing_entity_id].present?
        subscription.organization.billing_entities.find(params[:billing_entity_id])
      elsif params[:billing_entity_code].present?
        subscription.organization.billing_entities.find_by!(code: params[:billing_entity_code])
      end
```

spec/services/invoices/subscription_service_spec.rb
```diff
@@ -101,6 +101,91 @@
...
+
+        new_invoice = described_class.call(
+          subscriptions: [subscription],
+          timestamp: timestamp.to_i,
+          invoicing_reason: :subscription_periodic
+        ).invoice
+
+        expect(new_invoice.billing_entity_id).to eq(eu_entity.id)
+        expect(new_invoice.fees.subscription.first.billing_entity_id).to eq(eu_entity.id)
+        expect(past_invoice.reload.billing_entity_id).to eq(billing_entity.id)
```
toommz, line 161: Could it happen that running the service changes the past invoice billing entity?

spec/services/subscriptions/update_service_spec.rb
```diff
@@ -1109,5 +1109,108 @@
...
+
+          it "assigns the new billing entity to the subscription" do
+            update_service.call
+
+            expect(subscription.reload.billing_entity_id).to eq(new_billing_entity.id)
+          end
+
+          it "sends subscription.updated webhook" do
+            expect { update_service.call }.to have_enqueued_job_after_commit(SendWebhookJob).with("subscription.updated", subscription)
+          end
```
toommz, line 1134: Is it intended that this expectation is only present for this context, and that we don't assert that it does **not** send the webhook in some other cases?

app/services/fees/build_pay_in_advance_fixed_charge_service.rb
aquinofb: Extracted to `Subscription#applicable_billing_entity_id` in 3e92a89fa, matching the existing `applicable_*` naming on Customer. All four fee writers and the `mixed_billing_entities?` predicate now use it.

app/services/subscriptions/update_service.rb
aquinofb: Renamed to `new_billing_entity` in 3e92a89fa.

app/services/subscriptions/update_service.rb
aquinofb: Adopted in 3e92a89fa. Note that ing-119 extracts this same body into a shared `BillingEntityResolutionConcern`, so the simplification will need to be ported there when that branch consolidates the three callers (CreateService, PlanUpgradeService, and UpdateService). Flagged in the linked Linear ticket.

spec/services/invoices/subscription_service_spec.rb
aquinofb: No code path mutates `invoice.billing_entity_id` after creation; the column is stamped once in `Invoices::CreateGeneratingService` and never re-written. The assertion is here to lock in that invariant rather than because anything in this PR threatens it. If a future change introduced a re-stamp on the past invoice (e.g. some refresh job), this test would catch it.

spec/services/subscriptions/update_service_spec.rb
aquinofb: Good catch. Added the negative case in 3e92a89fa: when an unknown `billing_entity_id` triggers `not_found_failure!`, the webhook is not enqueued. The positive case covered the happy path; the negative one locks in that the failure short-circuits before `subscription.save!` and the post-commit enqueue.

toommz (approved):

PR #5500, opened 2026-05-12
fix(coupon): Coupon code column missing NOT NULL constraint in database

murparreira:
## Context

The coupons table allows NULL values for the code column, meaning a coupon can be created via the API without providing a code. This is inconsistent with our [public API documentation](https://getlago.com/docs/api-reference/coupons/create#body-coupon-code), which marks code as a required field and every other model with a code column (add_on, billable_metric, plan, billing_entity, tax, pricing_unit, charge, etc.) which all have NOT NULL at the DB level.

## Description

Add presence: true to the model and required: true in GraphQL, stops new NULLs from being created. 
Included a data migration to assign auto-generated codes to any existing NULL-code coupons and the NOT NULL clause on the code.


db/migrate/20260512155310_add_not_null_to_coupon_code.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddNotNullToCouponCode < ActiveRecord::Migration[8.0]
+  def up
+    safety_assured do
+      execute <<~SQL
+        UPDATE coupons SET code = 'coupon-' || LEFT(id::text, 8) WHERE code IS NULL;
```
groyoh: ⚠️ `LEFT(id::text, 8)` only takes the first 32 bits of the UUID. The partial unique index `index_coupons_on_organization_id_and_code` (where `deleted_at IS NULL`) will reject the migration if two kept NULL-code coupons in the same organization happen to share those 8 hex chars. The collision is rare but the failure mode is a hard migration crash in production with no obvious recovery path.

Using the full UUID removes the risk entirely:

```sql
UPDATE coupons SET code = 'coupon-' || id::text WHERE code IS NULL;
```

db/migrate/20260512155310_add_not_null_to_coupon_code.rb
```diff
@@ -0,0 +1,17 @@
...
+      SQL
+
+      change_column_null :coupons, :code, false
```
groyoh: The codebase pattern for adding `NOT NULL` to an existing column on a non-trivial table is a two-step rollout: first an `add_check_constraint ... validate: false` migration, then a follow-up that runs `validate_check_constraint` + `change_column_null` with `disable_ddl_transaction!`. See `db/migrate/20251229153718_add_code_not_null_check_constraint_to_charges.rb` and `db/migrate/20251229153734_validate_charges_code_not_null.rb`.

Running `change_column_null` directly (even inside `safety_assured`) takes an `ACCESS EXCLUSIVE` lock while Postgres scans every row to verify the constraint. `coupons` may be small enough today that this is unnoticeable, but aligning with the existing pattern keeps the rollout predictable across environments.

app/graphql/types/coupons/create_input.rb
```diff
@@ -7,7 +7,7 @@ class CreateInput < Types::BaseInputObject
 
       argument :amount_cents, GraphQL::Types::BigInt, required: false
       argument :amount_currency, Types::CurrencyEnum, required: false
-      argument :code, String, required: false
+      argument :code, String, required: true
```
groyoh, line 10: ⚠️ `Types::Coupons::UpdateInput` inherits from `CreateInput`, so flipping `code` to `required: true` here also makes it required on `UpdateCouponInput` (visible in the `schema.graphql` diff). Today `Coupons::UpdateService` only writes `code` when `params.key?(:code)` (see `app/services/coupons/update_service.rb:54`), so partial updates that omit `code` are valid against the service but will now be rejected by GraphQL before they reach it.

Override the argument in `UpdateInput` to keep it optional:

```ruby
class UpdateInput < Types::Coupons::CreateInput
  graphql_name "UpdateCouponInput"

  argument :id, String, required: true
  argument :code, String, required: false
end
```

db/migrate/20260512155310_add_not_null_to_coupon_code.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddNotNullToCouponCode < ActiveRecord::Migration[8.0]
+  def up
+    safety_assured do
+      execute <<~SQL
```
groyoh: Please make sure to separate data migration from schema migration.

db/migrate/20260512155310_add_not_null_to_coupon_code.rb
groyoh: This data migration should be done in batch to avoid a long-running transaction/lock on the database.

groyoh (changes requested):
Could you please also regression tests to ensure sending `nil` code will be failing both on the API and GraphQL ?

db/migrate/20260512155310_backfill_coupon_codes.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class BackfillCouponCodes < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def up
+    Coupon.unscoped.where(code: nil).find_in_batches(batch_size: 1000) do |batch|
+      Coupon.unscoped.where(id: batch.pluck(:id))
+        .update_all("code = 'coupon-' || id::text") # rubocop:disable Rails/SkipsModelValidations
+    end
```
groyoh, line 10: To confirm but I believe this also works.

```suggestion
    Coupon.unscoped.where(code: nil).in_batches.update_all("code = 'coupon-' || id::text")
```

groyoh (approved):
I believe we're still missing a non-regression test on the graphQL/api side.

PR #5501, opened 2026-05-13
feat(presentation_group_keys): change cached aggregation to support presentation breakdowns

tiagolupepic:
This commit changes the CachedAggregation model to support presentation breakdowns

Continuing the epic to introduce the presentation_group_keys, we're changing one more piece
related to fees.  

First, we're changing the model CachedAggregation to store the presentation breakdowns, 
we opt in to use a single jsonb column in cached_aggregations table, we just want to store the data 
and we're not going to perform any query in this column.

In order to keep it consistent, we're storing only the group related to presentation group keys, 
this means, we're removing the keys that belongs to `grouped_by` before assign to cached aggregation
presentation_breakdowns column.
This affects the `Fees::ChargeService` and `Charges::PayInAdvanceAggregationService`. In
the first service, we did a small refactor to avoid compute twice the presentation breakdowns by group_by.

Finally, we're changing the WeightedSum service to retrieve the presentation_breakdowns when 
the cached aggregation is available, with this flow fixed, we're making sure the recurring billing
are consistent with other services like SumService and etc.


db/migrate/20260508134715_add_presentation_breakdowns_to_cached_aggregations.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPresentationBreakdownsToCachedAggregations < ActiveRecord::Migration[8.0]
+  def change
+    add_column :cached_aggregations, :presentation_breakdowns, :jsonb, null: false, default: []
```
rsempe, line 5: should we create it before deploying the code change?

rsempe (approved):
lgtm

db/migrate/20260508134715_add_presentation_breakdowns_to_cached_aggregations.rb
tiagolupepic: since the migrations run before the image being deployed, I think it's fine to have it in the same PR

PR #5502, opened 2026-05-13
feat(payment-gated-subs): Delegate downgrade logic to its own service

osmarluz:
## Context

`Subscriptions::PlanUpgradeService` already owns the upgrade flow, but downgrades stayed inline in `Subscriptions::CreateService`. This PR brings the two paths into symmetry ahead of the payment-gating work for plan changes (Milestone 3). Pure refactor — no behavior change.

## Description

- `Subscriptions::PlanDowngradeService` (new) owns the downgrade flow, mirroring `Subscriptions::PlanUpgradeService`. Returns `result.subscription = current_subscription` — the load-bearing difference from upgrades, which return the new subscription.

- `Subscriptions::CreateService#downgrade_subscription` collapses to a one-line delegation to the new service.


spec/services/subscriptions/create_service_spec.rb
```diff
@@ -1353,6 +1353,28 @@
...
+          context "when subscription downgrade fails" do
+            let(:result_failure) do
+              BaseService::Result.new.validation_failure!(
+                errors: {billing_time: ["value_is_invalid"]}
+              )
+            end
+
+            before do
+              allow(Subscriptions::PlanDowngradeService)
+                .to receive(:call)
```
ancorcruz, line 1384:
```suggestion
                .to receive(:call!)
```

app/services/subscriptions/plan_downgrade_service.rb
```diff
@@ -0,0 +1,93 @@
...
+          Utils::ActivityLog.produce(current_subscription, "subscription.updated")
+        end
+      end
+
+      result.subscription = current_subscription
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
+    rescue BaseService::FailedResult => e
+      result.fail_with_error!(e)
```
ancorcruz: Is this actually reachable?

ancorcruz (approved):
Could we add a case covering the rescue ActiveRecord::RecordInvalid branch? in `spec/services/subscriptions/plan_downgrade_service_spec.rb`. Something like:
 
```
context "when creating the next subscription fails validation" do
  before do
    allow_any_instance_of(Subscription)
      .to receive(:save!)
      .and_raise(ActiveRecord::RecordInvalid.new(Subscription.new))
  end

  it "returns a record validation failure" do
  end
end
```

Right now the rescue branch is uncovered for a freshly-extracted service — would rather pin it down.

spec/services/subscriptions/create_service_spec.rb
osmarluz: @ancorcruz I'll change this on the `context "when subscription upgrade fails"` as well

app/services/subscriptions/plan_downgrade_service.rb
osmarluz: @ancorcruz no, we can remove it, thanks!

spec/services/subscriptions/create_service_spec.rb
osmarluz: @ancorcruz this actually needs to be `call` to allow the real `call!` on the service to run and raise `BaseService::FailedResult` properly

PR #5471, opened 2026-05-06
ING-42 feat(wallets): Support billing_entity_code on wallet creation

mariohd:
## Summary
  - Accept `billing_entity_code` and `billing_entity_id` params on wallet creation to assign a billing entity when the `multi_entity_billing` feature flag is enabled                                                
  - `billing_entity_id` takes precedence over `billing_entity_code` when both are provided
  - Returns `not_found` error for unknown billing entity code or ID; falls back to the customer's billing entity when neither is provided
  - Billing entity assignment is skipped entirely when the feature flag is disabled
  - Billing entity lookups are scoped to the current organization, preventing cross-organization assignment                                                                                                          
                                              
  ## Test plan                                                                                                                                                                                                       
  - [x] Service tests cover lookup by code, by ID, unknown code, unknown ID, precedence of ID over code, fallback to customer, and flag-disabled scenarios                                                           
  - [x] Service tests verify that billing entities from other organizations are rejected with not_found                                                                                                              
  - [x] Controller shared examples cover both `Api::V1::WalletsController` and `Api::V1::Customers::WalletsController` for `billing_entity_code` and `billing_entity_id`

PR #5503, opened 2026-05-13
feat(multi-billing-entity): support billing entity on subscription upgrade

aquinofb:
## Context

Multi-billing-entity support has been rolling out across the API a slice at a time. Subscription create landed in #5467 and wallets in #5471. The upgrade path was the next obvious gap: if you'd explicitly bound a subscription to a non-default entity (say `eu` instead of the customer's default `us`), upgrading the plan silently dropped that binding. The new subscription ended up with `billing_entity_id` NULL and quietly inherited the customer's default at billing time. The `billing_entity_id` and `billing_entity_code` params that already work on create were silently ignored on upgrade, so operators couldn't re-bind during an upgrade either.

The same gap was hiding inside pending-subscription promotion: when an operator raised a plan's amount above the previous one (via `Plans::UpdateService` or `Plans::UpdateAmountService`), the pending subscription got promoted through `PlanUpgradeService`, and the binding was lost the same way.

## Description

There are two moving parts. The resolution rule (read params, gated by `multi_entity_billing`, scoped to the customer's organization, not-found error on unknown id/code) is the same one `CreateService` already had inline. I extracted it into a small shared concern, `Subscriptions::Concerns::BillingEntityResolutionConcern`, so create and upgrade now share a single source of truth.

The new behaviour on upgrade is straightforward: resolve the entity from params if the flag is on, else carry over `current_subscription.billing_entity_id`. The carry-over reads the raw FK column rather than the `Subscription#billing_entity` reader, because the reader would otherwise materialise the customer's current default and freeze it into the new row, collapsing the "inherit at billing time" semantic. The carry-over runs regardless of the feature flag, so an explicit binding made while the flag was on survives even if the flag is later toggled off; on single-entity orgs every FK is NULL, so this branch is a no-op for them.

The two pending-promotion paths (`Plans::UpdateService` and `Plans::UpdateAmountService`) needed no code change. They already call `PlanUpgradeService`, and once upgrade does the carry-over, those inherit the behaviour. Two new specs confirm it end-to-end.

### Behaviour by scenario

| Scenario | Before | After |
|---|---|---|
| Flag ON, no param, current sub bound to `eu` | new sub `billing_entity_id` = NULL (binding lost) | new sub bound to `eu` |
| Flag ON, no param, current sub NULL | NULL (inherit customer default at read time) | NULL (inherit customer default at read time) |
| Flag ON, `billing_entity_code: us`, current sub bound to `eu` | param ignored, new sub NULL | new sub bound to `us` |
| Flag ON, unknown `billing_entity_code` | param ignored, new sub NULL | `not_found_failure(resource: "billing_entity")` |
| Flag OFF, current sub bound to `eu` | new sub NULL (binding lost) | new sub bound to `eu` (carry-over independent of flag) |
| Pending-promotion via `Plans::UpdateService` / `UpdateAmountService` | new sub NULL | new sub keeps previous binding |

## How to try locally

Spin up a customer on an org with `multi_entity_billing` enabled and at least two billing entities. Call them `eu` and `us`.

1. Create a subscription bound to `eu`:
   ```bash
   curl -X POST http://localhost:3000/api/v1/subscriptions \
     -H "Authorization: Bearer $API_KEY" -H "Content-Type: application/json" \
     -d '{"subscription":{"external_customer_id":"cust_1","external_id":"sub_1","plan_code":"starter","billing_entity_code":"eu"}}'
   ```
2. Upgrade the plan with no `billing_entity_*` in the body:
   ```bash
   curl -X POST http://localhost:3000/api/v1/subscriptions \
     -H "Authorization: Bearer $API_KEY" -H "Content-Type: application/json" \
     -d '{"subscription":{"external_customer_id":"cust_1","external_id":"sub_1","plan_code":"growth"}}'
   ```
3. Check the new active subscription's `billing_entity_id`. It now matches the previous one. Before this change it would have been NULL, and the next invoice would have been numbered under the customer's default entity.

To exercise the explicit-override path, replace the second body with `"billing_entity_code":"us"`. The new subscription binds to `us`, and the termination invoice for the previous subscription stays under `eu`.

The full test suite for the change:
```
  spec/services/subscriptions/plan_upgrade_service_spec.rb \
  spec/services/subscriptions/create_service_spec.rb \
  spec/services/plans/update_service_spec.rb \
  spec/services/plans/update_amount_service_spec.rb
```

226 examples, 0 failures.

### DIFF
--- a/app/services/subscriptions/plan_upgrade_service.rb
+++ b/app/services/subscriptions/plan_upgrade_service.rb
@@ -2,6 +2,8 @@
 
 module Subscriptions
   class PlanUpgradeService < BaseService
+    include Subscriptions::Concerns::BillingEntityResolutionConcern
+
     Result = BaseResult[:subscription]
 
     def initialize(current_subscription:, plan:, params:)
@@ -73,7 +75,14 @@ module Subscriptions
         subscription_at: current_subscription.subscription_at,
         billing_time: current_subscription.billing_time,
         ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at,
-        consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : current_subscription.consolidate_invoice
+        consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : current_subscription.consolidate_invoice,
+        # fallback_id uses the raw FK column, not Subscription#billing_entity (which would
+        # materialize customer.billing_entity, collapsing the "inherit at billing time" intent).
+        billing_entity: resolve_billing_entity(
+          customer: current_subscription.customer,
+          params:,
+          fallback_id: current_subscription.billing_entity_id
+        )
       )
 
       if params.key?(:payment_method)
--- a/app/services/subscriptions/plan_upgrade_service.rb
+++ b/app/services/subscriptions/plan_upgrade_service.rb
@@ -76,8 +76,6 @@ module Subscriptions
         billing_time: current_subscription.billing_time,
         ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at,
         consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : current_subscription.consolidate_invoice,
-        # fallback_id uses the raw FK column, not Subscription#billing_entity (which would
-        # materialize customer.billing_entity, collapsing the "inherit at billing time" intent).
         billing_entity: resolve_billing_entity(
           customer: current_subscription.customer,
           params:,
--- a/app/services/subscriptions/plan_upgrade_service.rb
+++ b/app/services/subscriptions/plan_upgrade_service.rb
@@ -65,6 +65,7 @@ module Subscriptions
     attr_reader :current_subscription, :plan, :params, :name
 
     def new_subscription_with_overrides
+      resolved_entity = resolve_billing_entity(customer: current_subscription.customer, params:)
       new_subscription = Subscription.new(
         organization_id: current_subscription.customer.organization_id,
         customer: current_subscription.customer,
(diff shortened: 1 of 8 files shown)


app/services/subscriptions/plan_upgrade_service.rb
```diff
@@ -72,7 +75,8 @@ def new_subscription_with_overrides
         previous_subscription_id: current_subscription.id,
         subscription_at: current_subscription.subscription_at,
         billing_time: current_subscription.billing_time,
-        ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at
+        ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at,
+        billing_entity_id: resolved_entity&.id || current_subscription.billing_entity_id
```
lovrocolic, line 80: Can you please check with the product team if this is intended? For me it makes sense to inherit billing entity from old subscription as a fallback. However, this is changing current behaviour. If customer change billing entity n the meantime, it will not be reflected

aquinofb: From my understanding, the carry-over reads the FK column `billing_entity_id`, not the reader `billing_entity` (which is `super || customer&.billing_entity`). For subs that were never explicitly bound, the FK is `NULL`, stays `NULL` on the new row, and the customer-default fallback still kicks in at read time, so a later customer change is reflected as before. The non-`NULL` case is an explicit operator override from #5467 that's meant to insulate the sub from customer-default changes.

If we drop the carry-over and let the new sub fall back to `NULL` for the explicitly-bound case, aren't we essentially putting back the regression this PR is fixing? Or, maybe did I miss something, or misunderstood your point? 🤔

lovrocolic: Probably you are right:)  Looks good 👌

lovrocolic (approved):

mariohd (approved):

PR #5505, opened 2026-05-13
POC(fixed-charges): feature skip plan override on units-only update

ancorcruz:
## Summary

- Reintroduce `subscription_fixed_charge_units_overrides` to record per-subscription units without cloning the plan (originally added in #4011, removed in #4318 and #4876 while waiting on the wider plan-override refactor).
- Branch `Subscriptions::UpdateOrOverrideFixedChargeService`: when the request only changes `units` (optionally with `apply_units_immediately`) **and** the subscription is not already on an overridden plan, write to the new table and skip the plan/fixed-charge override path. Any other field change, or a subscription already on an overridden plan, keeps the existing behaviour.
- Route `FixedChargeEvents::CreateService` through a new `FixedCharge#units_for(subscription)` accessor so emitted events carry the effective units.

## Background

Updating a fixed charge's units for one subscription today materialises a full plan override plus a fixed-charge override.

That's correct semantically but expensive in practice: organisations with many subscriptions on a shared plan accumulate one plan override per subscription, even though only `units` differs. This PoC offers a units-only fast path that avoids the plan override entirely while preserving the legacy path for any other field change.

## Out of scope for this PoC

Read sites still consume `fixed_charge.units` directly. They need to be switched to `fixed_charge.units_for(subscription)` in a follow-up before this can ship beyond PoC:
- `app/services/fees/fixed_charge_service.rb`
- `app/services/fees/build_pay_in_advance_fixed_charge_service.rb`
- `app/services/fees/init_from_adjusted_fixed_charge_fee_service.rb`
- `app/services/fixed_charge_events/aggregations/preview_aggregation_service.rb`

The scenario spec deliberately stops at the event level until those reads are wired.

### DIFF
--- a/app/controllers/api/v1/subscriptions/fixed_charges_controller.rb
+++ b/app/controllers/api/v1/subscriptions/fixed_charges_controller.rb
@@ -19,7 +19,8 @@ module Api
               ::V1::FixedChargeSerializer,
               collection_name: "fixed_charges",
               meta: pagination_metadata(fixed_charges),
-              includes: %i[taxes]
+              includes: %i[taxes],
+              subscription: subscription
             )
           )
         end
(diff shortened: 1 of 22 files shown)


ancorcruz:
POC closed.

PR #5506, opened 2026-05-13
feat(payment-gated-subs): Apply activation rules on upgrade of a subscription starting in the future

osmarluz:
## Context

First slice of Milestone 3 (Upgrade Path). `Subscriptions::PlanUpgradeService` currently ignores `params[:activation_rules]`; on an upgrade of a `starting_in_the_future?` subscription, rules should land on the pending subscription so `ActivateService` evaluates them when the activation date arrives — same lifecycle as a new subscription with rules.

## Description

- `Subscriptions::PlanUpgradeService` calls `Subscriptions::ActivationRules::ApplyService` against the current subscription in the `starting_in_the_future?` branch when `params[:activation_rules]` is provided.

### DIFF
--- a/app/services/subscriptions/plan_upgrade_service.rb
+++ b/app/services/subscriptions/plan_upgrade_service.rb
@@ -14,17 +14,17 @@ module Subscriptions
     end
 
     def call
-      if current_subscription.starting_in_the_future?
-        apply_activation_rules if params[:activation_rules]
-        update_pending_subscription
-
-        result.subscription = current_subscription
-        return result
-      end
-
-      new_subscription = new_subscription_with_overrides
-
       ActiveRecord::Base.transaction do
+        if current_subscription.starting_in_the_future?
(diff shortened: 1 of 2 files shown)


app/services/subscriptions/plan_upgrade_service.rb
```diff
@@ -15,6 +15,7 @@ def initialize(current_subscription:, plan:, params:)
 
     def call
       if current_subscription.starting_in_the_future?
+        apply_activation_rules if params[:activation_rules]
```
ancorcruz: this and the `update_pending_subscription` call should be wrapped within a transaction to ensure atomicity on the rules application and subscription update.

ancorcruz (approved):
I'm approving this PR but it should not be merged without first wrapping the apply_activation_rules call along with the update_pending_subscription call in a database transaction to ensure the changes are atomic.

PR #5507, opened 2026-05-13
[ING-128] feat(invoice-grouping): Subscription consolidation

lovrocolic:
## Context

Currently customer subscriptions are grouped per payment method, billing entity and currency when creating recurring invoice.

## Description

We are adding `consolidate_invoice` flag on subscription so that users can define if certain subscription should be consolidated or excluded from consolidation.

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -130,7 +130,7 @@ module Subscriptions
         billing_time: billing_time || :calendar,
         ending_at: params[:ending_at],
         progressive_billing_disabled: params[:progressive_billing_disabled] || false,
-        invoice_consolidation_enabled: params.key?(:invoice_consolidation_enabled) ? params[:invoice_consolidation_enabled] : true,
+        consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : true,
         billing_entity: resolve_billing_entity
       )
 
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -130,6 +130,7 @@ module Subscriptions
         billing_time: billing_time || :calendar,
         ending_at: params[:ending_at],
         progressive_billing_disabled: params[:progressive_billing_disabled] || false,
+        invoice_consolidation_enabled: params.key?(:invoice_consolidation_enabled) ? params[:invoice_consolidation_enabled] : true,
         billing_entity: resolve_billing_entity
       )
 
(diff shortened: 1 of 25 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -130,6 +130,7 @@ def create_subscription
         billing_time: billing_time || :calendar,
         ending_at: params[:ending_at],
         progressive_billing_disabled: params[:progressive_billing_disabled] || false,
+        consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : true,
```
mariohd, line 133: I believe this would also work

```suggestion
        consolidate_invoice: params.fetch(:consolidate_invoice, true),
```

app/services/subscriptions/create_service.rb
```diff
@@ -207,7 +208,8 @@ def downgrade_subscription
         status: :pending,
         billing_time: current_subscription.billing_time,
         ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at,
-        progressive_billing_disabled: params[:progressive_billing_disabled] || false
+        progressive_billing_disabled: params[:progressive_billing_disabled] || false,
+        consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : current_subscription.consolidate_invoice
```
mariohd: same here with proper changes

app/services/subscriptions/plan_upgrade_service.rb
```diff
@@ -72,7 +72,8 @@ def new_subscription_with_overrides
         previous_subscription_id: current_subscription.id,
         subscription_at: current_subscription.subscription_at,
         billing_time: current_subscription.billing_time,
-        ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at
+        ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at,
+        consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : current_subscription.consolidate_invoice
```
mariohd, line 76: same

mariohd (approved):
overall PR looks good, I just added some comments

aquinofb (approved):

app/services/subscriptions/create_service.rb
mariohd: actually this would not work, if someone sends `nil`, it will assign `nil` to the value, which is wrong

PR #5511, opened 2026-05-14
FacturX > Cii

mariohd:
The serializers and services under `app/serializers/e_invoices/factur_x/` and `app/services/e_invoices/*/factur_x/` were named after Factur-X, but they actually produce Cross Industry Invoice (CII) XML. Factur-X is the *hybrid output format* — a PDF/A-3 file with a CII XML stream embedded as an attachment — and is assembled one level up in `Invoices::GeneratePdfService#attach_facturx` (and its credit-note and payment-receipt equivalents), where the PDF is generated and the XML is embedded. The XML-producing namespace itself produces pure CII, so the directory and module names should reflect that.

This PR is a cosmetic rename. The classes, generated XML, env vars, schema, and operator-facing API are all unchanged — only file paths and Ruby constants move.

## What moved

|  | Before | After |
| --- | --- | --- |
| Component serializers | `app/serializers/e_invoices/factur_x/*` | `app/serializers/e_invoices/cii/*` |
| Per-document builders | `app/serializers/e_invoices/{invoices,credit_notes,payments}/factur_x/builder.rb` | `…/cii/builder.rb` |
| Per-document create services | `app/services/e_invoices/{invoices,credit_notes,payments}/factur_x/create_service.rb` | `…/cii/create_service.rb` |
| All spec mirrors | `spec/{serializers,services}/e_invoices/.../factur_x/*` | `…/cii/*` |
| Module names | `EInvoices::FacturX::*` | `EInvoices::Cii::*` |
| Helper key | `XMLHelper::NAMESPACES[:factur_x]` | `XMLHelper::NAMESPACES[:cii]` |
| PDF service callers (constant references only) | `EInvoices::*::FacturX::CreateService` | `EInvoices::*::Cii::CreateService` |

## What did not move

The `attach_facturx` method in `Invoices::GeneratePdfService`, `CreditNotes::GeneratePdfService`, and `PaymentReceipts::GeneratePdfService` keeps its name. That method's job — embedding CII XML inside a PDF/A-3 — is genuinely Factur-X, so the name is still accurate at the call site even though the inner serializer namespace is now `Cii`.

## Test plan

- [ ] `bundle exec rspec spec/serializers/e_invoices/` (every CII and UBL serializer spec — both load the shared `spec/support/xml_helper.rb`)
- [ ] `bundle exec rspec spec/services/e_invoices/` (the renamed `cii/create_service.rb` files)
- [ ] `bundle exec rspec spec/services/{invoices,credit_notes,payment_receipts}/generate_pdf_service_spec.rb` (PDF callers stubbing the renamed constants)
- [ ] `grep -rn 'FacturX\|factur_x' app/ spec/ lib/` returns no results
- [ ] `bundle exec rubocop` clean on the 44 changed files

## Migration

None. Pure internal rename with no behavior change, no schema change, no env-var change, and no public-API change. Operators see nothing.


app/serializers/e_invoices/cii/applicable_trade_tax.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module EInvoices
-  module FacturX
+  module Cii
```
aquinofb, line 4: Shall we add CII as an acronym?

aquinofb (approved):

app/serializers/e_invoices/cii/applicable_trade_tax.rb
mariohd: having the acronym will require to add a new initializer just for it. 
I dont think this is necessary for a "cosmetic" change.

PR #5514, opened 2026-05-15
feat(payment-gated-subs): add PSP payment cancel dispatcher

ancorcruz:
## Context

The timeout/expiration flow for incomplete payment-gated subscriptions needs to release the open PSP authorization on a best-effort basis. Each supported PSP (Stripe, Adyen, GoCardless) has its own cancel service; we need a single entry point that routes a Payment to the right one without callers needing to know which provider is in play.

 ## Description

Add PaymentProviders::CancelPaymentService — a dispatcher that takes a payment and routes by payment.payment_provider.type:

- PaymentProviders::StripeProvider     -> Stripe::Payments::CancelService
- PaymentProviders::AdyenProvider      -> Adyen::Payments::CancelService
- PaymentProviders::GocardlessProvider -> Gocardless::Payments::CancelService
- Other providers (Cashfree, Flutterwave, MoneyHash): silent no-op with
  an info log. These PSPs do not expose a cancel API; the eventual
  webhook (or manual reconciliation) is the lifecycle authority.

Three short-circuit guards return a successful no-op with the payment exposed on the result:

- payment.payment_provider.blank?
- payment.provider_payment_id.blank?
- payment.payable_payment_status == "succeeded"

The "succeeded" guard is protective: canceling a captured payment would either error at the PSP or, depending on provider, trigger an unintended refund. The dispatcher is best-effort by contract — these guards keep callers from needing pre-checks.

PaymentProviders::CancelPaymentJob is the async wrapper. It takes a payment and forwards to the dispatcher; the caller (the upcoming ExpireService) is responsible for picking the right payment from whatever scope it has.

Placement follows the existing convention. The create-side dispatcher PaymentProviders::CreatePaymentFactory lives directly under PaymentProviders::; this dispatcher sits alongside it. Service rather than Factory because the cancel services share a uniform payment: signature, so the Factory.new_instance().call! two-step is unnecessary.

### DIFF
--- a/app/services/payment_providers/cancel_payment_service.rb
+++ b/app/services/payment_providers/cancel_payment_service.rb
@@ -14,7 +14,7 @@ module PaymentProviders
 
       return result if payment.payment_provider.blank?
       return result if payment.provider_payment_id.blank?
-      return result if payment.payable_payment_status == "succeeded"
+      return result if payment.succeeded?
 
       case payment.payment_provider.type
       when "PaymentProviders::StripeProvider"
--- a/app/services/payment_providers/cancel_payment_service.rb
+++ b/app/services/payment_providers/cancel_payment_service.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module PaymentProviders
+  class CancelPaymentService < BaseService
+    Result = BaseResult[:payment]
+
+    def initialize(payment:)
+      @payment = payment
+      super
+    end
+
+    def call
+      result.payment = payment
+
+      return result if payment.payment_provider.blank?
+      return result if payment.provider_payment_id.blank?
+      return result if payment.payable_payment_status == "succeeded"
+
+      case payment.payment_provider.type
+      when "PaymentProviders::StripeProvider"
+        PaymentProviders::Stripe::Payments::CancelService.call!(payment:)
+      when "PaymentProviders::AdyenProvider"
+        PaymentProviders::Adyen::Payments::CancelService.call!(payment:)
+      when "PaymentProviders::GocardlessProvider"
+        PaymentProviders::Gocardless::Payments::CancelService.call!(payment:)
+      else
+        # Cashfree, Flutterwave, MoneyHash, and any future provider without a
+        # dedicated cancel service: nothing to do here. The eventual webhook
+        # (or reconciliation) is the lifecycle authority for the Payment.
+        Rails.logger.info(
+          "PaymentProviders::CancelPaymentService: PSP cancel not supported for " \
+          "#{payment.payment_provider.type} (payment #{payment.id}); skipping"
+        )
+      end
+
+      result
+    end
+
+    private
+
+    attr_reader :payment
+  end
+end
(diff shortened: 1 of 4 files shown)


app/services/payment_providers/cancel_payment_service.rb
```diff
@@ -0,0 +1,43 @@
...
+      @payment = payment
+      super
+    end
+
+    def call
+      result.payment = payment
+
+      return result if payment.payment_provider.blank?
+      return result if payment.provider_payment_id.blank?
+      return result if payment.payable_payment_status == "succeeded"
```
osmarluz: @ancorcruz since it's an enum, we can use the method here:


```suggestion
      return result if payment.payable_payment_status_succeeded?
```

app/services/payment_providers/cancel_payment_service.rb
```diff
@@ -0,0 +1,43 @@
...
+        PaymentProviders::Gocardless::Payments::CancelService.call!(payment:)
+      else
+        # Cashfree, Flutterwave, MoneyHash, and any future provider without a
+        # dedicated cancel service: nothing to do here. The eventual webhook
+        # (or reconciliation) is the lifecycle authority for the Payment.
+        Rails.logger.info(
+          "PaymentProviders::CancelPaymentService: PSP cancel not supported for " \
+          "#{payment.payment_provider.type} (payment #{payment.id}); skipping"
+        )
+      end
```
osmarluz, line 34: @ancorcruz we can simplify this a bit by using a constant and checking `payment.payment_provider.payment_type` instead of the STI type:

 ```ruby
CANCEL_SERVICE = {
  "stripe" => "PaymentProviders::Stripe::Payments::CancelService",
  "adyen" => "PaymentProviders::Adyen::Payments::CancelService",
  "gocardless" => "PaymentProviders::Gocardless::Payments::CancelService"
}.freeze

...

  payment_type = payment.payment_provider.payment_type
  service_class = CANCEL_SERVICE[payment_type]

  if service_class
    service_class.constantize.call!(payment:)
  else
    # Cashfree, Flutterwave, MoneyHash, and any future provider without a
    # dedicated cancel service: nothing to do here. The eventual webhook
    # (or reconciliation) is the lifecycle authority for the Payment.
    Rails.logger.info(
      "PaymentProviders::CancelPaymentService: PSP cancel not supported for " \
      "#{payment_type} (payment #{payment.id}); skipping"
    )
  end
```

- The string class names followed by a `constantize` are just to work around the fact that the classes don't exist. After they do, having the constants directly work.
- The logged message above is using `payment_type`, so the tests would need to be adjusted accordingly

WDYT?

osmarluz (approved):
Looks good, just left 2 suggestions

app/services/payment_providers/cancel_payment_service.rb
ancorcruz: It would read as `payment.succeeded?` because the enum is not prefixed.

app/services/payment_providers/cancel_payment_service.rb
ancorcruz: I'd leave it as it is as it follows our existing implementation patterns used in many other services.

PR #5515, opened 2026-05-15
[ING-68] feat/multi-entity): adapt dunning campaign flow for multi entity billing

lovrocolic:
## Context

## Description

This PR adapts dunning campaign flow for multi entity billing.

### DIFF
--- a/app/services/dunning_campaigns/bulk_process_service.rb
+++ b/app/services/dunning_campaigns/bulk_process_service.rb
@@ -43,7 +43,13 @@ module DunningCampaigns
         increment_per_currency_attempts(thresholds)
 
         thresholds.each do |threshold|
-          DunningCampaigns::ProcessAttemptJob.perform_later(customer:, dunning_campaign_threshold: threshold)
+          overdue_billing_entities_for(threshold.currency).each do |entity|
+            DunningCampaigns::ProcessAttemptJob.perform_later(
+              customer:,
+              dunning_campaign_threshold: threshold,
+              billing_entity: entity
+            )
+          end
         end
 
         if all_dunned_currencies_max_attempts_reached?
@@ -75,6 +81,12 @@ module DunningCampaigns
             .where(currency:)
             .find_by("amount_cents <= ?", amount_cents)
         end
+      end
+
+      def overdue_billing_entities_for(currency)
+        entity_ids = customer.invoices.non_self_billed.payment_overdue
+          .where(currency: currency).distinct.pluck(:billing_entity_id)
+        customer.organization.billing_entities.where(id: entity_ids)
       end
 
       def increment_per_currency_attempts(thresholds)
--- a/app/services/dunning_campaigns/bulk_process_service.rb
+++ b/app/services/dunning_campaigns/bulk_process_service.rb
@@ -84,8 +84,14 @@ module DunningCampaigns
       end
 
       def overdue_billing_entities_for(currency)
-        entity_ids = customer.invoices.non_self_billed.payment_overdue
-          .where(currency: currency).distinct.pluck(:billing_entity_id)
+        entity_ids = customer
+          .invoices
+          .non_self_billed
+          .payment_overdue
+          .where(currency: currency)
+          .where(ready_for_payment_processing: true)
+          .distinct
+          .pluck(:billing_entity_id)
         customer.organization.billing_entities.where(id: entity_ids)
       end
 
--- a/app/services/dunning_campaigns/process_attempt_service.rb
+++ b/app/services/dunning_campaigns/process_attempt_service.rb
@@ -44,6 +44,7 @@ module DunningCampaigns
       return false if customer.exclude_from_dunning_campaign?
 
       custom_campaign = customer.applied_dunning_campaign
+      # Dunning campaign is resolved at the customer level (from their own billing entity), not from the invoice's BE.
       default_campaign = customer.billing_entity.applied_dunning_campaign
 
       custom_campaign == dunning_campaign || (!custom_campaign && default_campaign == dunning_campaign)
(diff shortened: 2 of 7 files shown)


app/services/dunning_campaigns/process_attempt_service.rb
```diff
@@ -44,21 +44,25 @@ def applicable_dunning_campaign?
...
       custom_campaign == dunning_campaign || (!custom_campaign && default_campaign == dunning_campaign)
     end
 
     def dunning_campaign_threshold_reached?
-      overdue_invoices.sum(:total_amount_cents) >= dunning_campaign_threshold.amount_cents
+      overdue_invoices_in_currency.sum(:total_amount_cents) >= dunning_campaign_threshold.amount_cents
     end
 
-    def overdue_invoices
+    def overdue_invoices_in_currency
```
ancorcruz, line 57: If I'm not wrong this also should exclude the self-billed invoices from the overdue ones.

app/services/dunning_campaigns/process_attempt_service.rb
```diff
@@ -44,21 +44,25 @@ def applicable_dunning_campaign?
...
       custom_campaign = customer.applied_dunning_campaign
-      default_campaign = billing_entity.applied_dunning_campaign
+      default_campaign = customer.billing_entity.applied_dunning_campaign
 
       custom_campaign == dunning_campaign || (!custom_campaign && default_campaign == dunning_campaign)
     end
 
     def dunning_campaign_threshold_reached?
-      overdue_invoices.sum(:total_amount_cents) >= dunning_campaign_threshold.amount_cents
+      overdue_invoices_in_currency.sum(:total_amount_cents) >= dunning_campaign_threshold.amount_cents
```
ancorcruz, line 54: I'm a bit confuse here...

We compute all the overdue invoices of the customer across billing entities with a given currency to check if the threshold has been reached. However, we run the dunning campaign scoped to each billing entity + currency.... then a customer with overdue invoices in EUR, for example 10k in billing_entity_1 and 5k in billing_entity_2, given a dunning threshold of 12k EUR, it will run two dunning processes for this customer one for the 10k overdue in BE_1 and another one with 5k in BE_2. Is this right?

app/services/dunning_campaigns/process_attempt_service.rb
```diff
@@ -44,21 +44,25 @@ def applicable_dunning_campaign?
       return false if customer.exclude_from_dunning_campaign?
 
       custom_campaign = customer.applied_dunning_campaign
-      default_campaign = billing_entity.applied_dunning_campaign
+      default_campaign = customer.billing_entity.applied_dunning_campaign
```
ancorcruz, line 48: why consider the customer default campaign the one defined as one of its billing entities' default campaing instead of the given billing entity default campaign? Does this make sense?

ancorcruz (approved):
I understand reading the code that we run the dunning for the customer across all billing entities (with a single counter per currency) but we split the amounts billing by BE. is this intended?

If that is the case, looks all good except self billed invoices must be excluded from dunning (pre existing bug I'm afraid).

app/services/dunning_campaigns/bulk_process_service.rb
```diff
@@ -77,6 +83,12 @@ def applicable_dunning_campaign_thresholds
         end
       end
 
+      def overdue_billing_entities_for(currency)
+        entity_ids = customer.invoices.non_self_billed.payment_overdue
+          .where(currency: currency).distinct.pluck(:billing_entity_id)
+        customer.organization.billing_entities.where(id: entity_ids)
```
ancorcruz: should we filter invoices by `ready_for_payment_processing: true` as well? We do in `ProcessAttemptService#overdue_invoices`

app/services/dunning_campaigns/process_attempt_service.rb
```diff
@@ -44,21 +44,25 @@ def applicable_dunning_campaign?
...
 
       custom_campaign == dunning_campaign || (!custom_campaign && default_campaign == dunning_campaign)
     end
 
     def dunning_campaign_threshold_reached?
-      overdue_invoices.sum(:total_amount_cents) >= dunning_campaign_threshold.amount_cents
+      overdue_invoices_in_currency.sum(:total_amount_cents) >= dunning_campaign_threshold.amount_cents
     end
```
ancorcruz, line 55: we run this query for each BE of the customer, it is always the same query as it is across all BE, we can run this one level above and pass the result to the service saving a DB query for each BE.

app/services/dunning_campaigns/process_attempt_service.rb
ancorcruz: if this is intentional it would be great to add a quick comment as insight for the next reader...

lovrocolic: A dunning campaign is a customer-level concept — it's resolved as customer.applied_dunning_campaign || 
  customer.billing_entity.applied_dunning_campaign. The per-BE iteration only scopes invoices, not which campaign applies. Note BulkProcessService#applicable_dunning_campaign (line 71) uses the exact same resolution, so they're consistent.

app/services/dunning_campaigns/process_attempt_service.rb
lovrocolic: I would say yes. BE affects only the billing side, not the tresholds

app/services/dunning_campaigns/bulk_process_service.rb
lovrocolic: Applied it, thanks!

app/services/dunning_campaigns/process_attempt_service.rb
lovrocolic: Applied it!

app/services/dunning_campaigns/process_attempt_service.rb
lovrocolic: I improved it a bit with memoization, but IMO not worth passing result around since in most cases, BE is the same and in rare cases we would have max. few billing entities. Let me know if you still think we should improve it more

PR #5517, opened 2026-05-15
ING-139 feat(payments): Filter payments by currency

mariohd:
The payments listing in the front-end needs to be filterable by currency, but the GraphQL `payments` query and the underlying `PaymentsQuery` had no such filter. This PR adds it at both layers.

The change is mechanical: a `currency` argument on `Resolvers::PaymentsResolver`, a `currency` entry in `PaymentsQuery::Filters`, and a `filter_by_currency` helper that scopes the query to `where(amount_currency: filters.currency)`. The resolver's `resolve` method defaults `currency:` to `nil` so existing callers that don't pass it continue to work unchanged. The generated `schema.graphql` and `schema.json` are regenerated to reflect the new argument.

## Before and after

|  | Before | After |
| --- | --- | --- |
| Filter payments by currency through GraphQL | No | `payments(currency: "USD") { ... }` |
| `PaymentsQuery::Filters` shape | `[:invoice_id, :external_customer_id]` | `[:invoice_id, :external_customer_id, :currency]` |
| Existing filters (invoice, external customer, search) | Worked | Unchanged |

## Test plan

- [ ] `bundle exec rspec spec/queries/payments_query_spec.rb`
- [ ] `bundle exec rspec spec/graphql/resolvers/payments_resolver_spec.rb`
- [ ] Manual via GraphiQL: `payments(currency: "USD") { collection { id } }` returns only payments whose `amount_currency = "USD"`.

## Migration

None. The new argument is optional and not gated — existing callers see no change.

## Follow-up worth opening

`app/contracts/queries/payments_query_filters_contract.rb` lists `invoice_id` and `external_customer_id` but not `currency`. Adding `optional(:currency).maybe(:string)` would reject malformed currency strings at the validation layer instead of silently returning empty results. Skipped here to keep the PR focused.

### DIFF
--- a/app/graphql/resolvers/payments_resolver.rb
+++ b/app/graphql/resolvers/payments_resolver.rb
@@ -9,6 +9,7 @@ module Resolvers
 
     description "Query payments of an organization"
 
+    argument :currency, String, required: false
     argument :external_customer_id, ID, required: false
     argument :invoice_id, ID, required: false
     argument :limit, Integer, required: false
@@ -17,12 +18,13 @@ module Resolvers
 
     type Types::Payments::Object.collection_type, null: false
 
-    def resolve(page: nil, limit: nil, invoice_id: nil, external_customer_id: nil, search_term: nil)
+    def resolve(currency: nil, page: nil, limit: nil, invoice_id: nil, external_customer_id: nil, search_term: nil)
       result = PaymentsQuery.call(
         organization: current_organization,
         filters: {
           invoice_id:,
-          external_customer_id:
+          external_customer_id:,
+          currency:
         },
         search_term:,
         pagination: {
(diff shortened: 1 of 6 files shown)


app/graphql/resolvers/payments_resolver.rb
```diff
@@ -9,6 +9,7 @@ class PaymentsResolver < Resolvers::BaseResolver
 
     description "Query payments of an organization"
 
+    argument :currency, String, required: false
```
domenicofalco: should this be `Types::CurrencyEnum` ?

domenicofalco (approved):

app/graphql/resolvers/payments_resolver.rb
mariohd: Good catch! I will change

PR #5518, opened 2026-05-15
feat: add billable_metric.{created,updated,deleted} webhooks

mikeh-lago:
Emit webhook events when a billable metric is created, updated or deleted, reaching parity with plans and features.

- New Webhooks::BillableMetrics::{Created,Updated,Deleted}Service reusing V1::BillableMetricSerializer.
- BillableMetrics::CreateService / UpdateService / DestroyService now fire SendWebhookJob.perform_after_commit with the matching event.
- Register the three event types in WEBHOOK_SERVICES and in the webhook_event_types.yml config under a new 'Billable Metrics' category — which auto-flows into the GraphQL EventTypeEnum and EventCategoryEnum used by the frontend webhook subscription UI.
- schema.graphql / schema.json regenerated for the new enum values.
- Specs added for the three webhook services and for the routing in SendWebhookJob; CreateService / UpdateService / DestroyService specs now assert SendWebhookJob is enqueued.

### DIFF
--- a/app/jobs/send_webhook_job.rb
+++ b/app/jobs/send_webhook_job.rb
@@ -15,6 +15,9 @@ class SendWebhookJob < ApplicationJob
 
   WEBHOOK_SERVICES = {
     "alert.triggered" => Webhooks::UsageMonitoring::AlertTriggeredService,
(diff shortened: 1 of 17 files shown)


annvelents (approved):
LGTM!

PR #5519, opened 2026-05-15
feat(payment-gated-subs): expire incomplete subscriptions on timeout

ancorcruz:
## Context

A payment-gated subscription enters `incomplete` and waits for the payment to resolve. If the resolution never arrives (3DS abandoned, SEPA pending indefinitely, customer closes browser, mandate cancelled externally and never re-established), the subscription stays in that state indefinitely. There is no automatic cleanup, no PSP-side release of pending authorizations, and no visibility into stuck records.

This PR adds the timeout pathway that closes the gap.

## Description

A clock-driven chain that finds incomplete subscriptions whose activation rule has aged past its `expires_at` and transitions them to `canceled` with `cancelation_reason: :timeout`, closes the open invoice, and best-effort cancels the pending payment with the PSP.

### `Subscriptions::ActivationRules::ExpireService`

Core service. For a given incomplete subscription:

1. Acquires `subscription.with_lock`. Race protection against a payment webhook resolution landing concurrently.
2. Re-checks `subscription.incomplete?` post-lock and bails if a webhook resolved the subscription first.
3. Transitions the payment activation rule to `:expired` via `Payment::EvaluateService`.
4. Closes the open invoice.
5. Calls `ResolveSubscriptionStatusService`, which performs the `mark_as_canceled!` transition and fires the `subscription.canceled` webhook + activity log (existing wiring, unchanged).
6. Sets `cancelation_reason: :timeout`. Follows the established `Payment::ResolveService#handle_failure` pattern of caller-sets-reason; rule status alone doesn't disambiguate which actor expired the rule (today only timeout, tomorrow potentially manual force-expire).
7. Enqueues `PaymentProviders::CancelPaymentJob` for the most recent pending/processing payment on the invoice. Runs after the transaction commits.

### `Subscriptions::ActivationRules::ExpireIncompleteJob`

Thin async wrapper. Each clock tick enqueues one per expirable subscription so a slow expiration doesn't block others. Queue: `:billing` when `SIDEKIQ_BILLING` is enabled, `:default` otherwise — matches sibling `Subscriptions::TerminateJob`. `unique :until_executed` prevents duplicate enqueues if a clock tick fires before the previous tick's jobs have drained.

### `Clock::ExpireIncompleteSubscriptionsJob`

Periodic batch worker. Uses the existing `Subscription.expirable` scope (incomplete + activation_rule pending + past `expires_at`). Inherits from `ClockJob`, which already routes the `:clock_worker` / `:clock` queue.

Registered in `clock.rb` to run hourly at `*:20`, staggered between the existing `*:15` and `*:30` schedules. Sentry cron monitor under slug `lago_expire_incomplete_subscriptions`.

### `Subscriptions::ActivationRules::Payment::ResolveJob` queue alignment

The existing `ResolveJob` was on `queue_as "default"`. Updated to the same conditional `:billing` / `:default` pattern as the new `ExpireIncompleteJob` so both halves of the activation-rule machinery (success/failure resolution and timeout expiration) scale onto the same dedicated worker pool when `SIDEKIQ_BILLING` is enabled. No behavioral change when the env var is unset.

### E2E scenarios

Two new scenarios in `payment_gated_activation_spec.rb`:

- Gated subscription whose rule has aged past `expires_at`: clock → expire job → expire service → subscription canceled with `cancelation_reason: :timeout`, rule expired, invoice closed.
- Gated subscription still within the timeout window: clock runs but subscription remains incomplete, confirming the `Subscription.expirable` scope correctly excludes future-expiry rules.

### DIFF
--- a/app/services/subscriptions/activation_rules/expire_service.rb
+++ b/app/services/subscriptions/activation_rules/expire_service.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  module ActivationRules
+    class ExpireService < BaseService
+      Result = BaseResult[:subscription]
+
+      def initialize(subscription:)
+        @subscription = subscription
+        super
+      end
+
+      def call
+        subscription.with_lock do
+          # Race protection: a payment webhook may resolve the subscription
+          # concurrently. If it already did so by the time we acquired the
+          # lock, bail.
+          next unless subscription.incomplete?
+
+          payment_rule = subscription.activation_rules.payment.sole
+          Payment::EvaluateService.call!(rule: payment_rule, status: :expired)
+
+          invoice = subscription.invoices.open.subscription.sole
+          invoice.closed!
+
+          ResolveSubscriptionStatusService.call!(subscription:)
+          subscription.update!(cancelation_reason: :timeout) if subscription.canceled?
+
+          enqueue_psp_cancel(invoice)
+        end
+
+        result.subscription = subscription
+        result
+      end
+
+      private
+
+      attr_reader :subscription
+
+      def enqueue_psp_cancel(invoice)
+        payment = invoice.payments
+          .where(payable_payment_status: %w[pending processing])
+          .order(created_at: :desc)
+          .first
+        return unless payment
+
+        PaymentProviders::CancelPaymentJob.perform_after_commit(payment)
+      end
+    end
+  end
+end
--- a/app/services/subscriptions/activation_rules/expire_service.rb
+++ b/app/services/subscriptions/activation_rules/expire_service.rb
@@ -24,7 +24,7 @@ module Subscriptions
           invoice.closed!
 
           ResolveSubscriptionStatusService.call!(subscription:)
-          subscription.update!(cancelation_reason: :timeout) if subscription.canceled?
+          subscription.update!(cancelation_reason: :timeout)
 
           enqueue_psp_cancel(invoice)
         end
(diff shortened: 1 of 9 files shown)


spec/jobs/clock/expire_incomplete_subscriptions_job_spec.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Clock::ExpireIncompleteSubscriptionsJob, job: true do
+  subject { described_class }
```
osmarluz: @ancorcruz since the tests are using `described_class` directly we can remove the `subject`

app/services/subscriptions/activation_rules/expire_service.rb
```diff
@@ -0,0 +1,51 @@
...
+          next unless subscription.incomplete?
+
+          payment_rule = subscription.activation_rules.payment.sole
+          Payment::EvaluateService.call!(rule: payment_rule, status: :expired)
+
+          invoice = subscription.invoices.open.subscription.sole
+          invoice.closed!
+
+          ResolveSubscriptionStatusService.call!(subscription:)
+          subscription.update!(cancelation_reason: :timeout) if subscription.canceled?
```
osmarluz: @ancorcruz since we now for sure that `Payment::EvaluateService.call!(rule: payment_rule, status: :expired)` succeeded at this point, we could omit the `if subscription.canceled?`

app/services/subscriptions/activation_rules/expire_service.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      private
+
+      attr_reader :subscription
+
+      def enqueue_psp_cancel(invoice)
+        payment = invoice.payments
+          .where(payable_payment_status: %w[pending processing])
+          .order(created_at: :desc)
+          .first
```
osmarluz, line 48: @ancorcruz the tests do not cover the scenario where more than one payment exist for the invoice. It would be good to cover this. Also, if there is more than one payment, shouldn't we enqueue a `PaymentProviders::CancelPaymentJob` for each one?

osmarluz (approved):
I've left a few small comments

app/services/subscriptions/activation_rules/expire_service.rb
ancorcruz: There's a partial unique constraint that limits provider payments in pending/processing to one per invoice — so two pending provider payments can't coexist. I'll add a comment in the code to make it clear.

PR #5521, opened 2026-05-18
feat(observability): enrich ActiveJob job logs (ING-136)

groyoh:
## Context

The structured JSON entries from `ActiveJob::JsonLogSubscriber` did not carry enough metadata to map a job log back to its inputs. Several events (the error events in particular) logged neither the `arguments` nor the `queue`, and the execution count was missing or inconsistent. A failed job could be seen in the logs, but not traced to the row, customer, or invoice it ran against.

## Description

Every entry from `ActiveJob::JsonLogSubscriber` now carries the same job context: `job_id`, `queue`, and `arguments`. This applies across `enqueue`, `enqueue_all`, `enqueue_at`, `perform_start`, `perform`, `enqueue_retry`, `retry_stopped`, and `discard`, for the success, error, and aborted branches alike. Arguments follow the existing `log_arguments?` setting and are formatted the way Rails formats them, including `GlobalID` identifiers for record arguments.

The execution count is emitted under a consistent `retries` key on the terminal events (`perform` error, `retry_stopped`, `discard`). `enqueue_retry` keeps its own `execution` key since it counts the current attempt rather than the terminal total.

The payload building was factored into a few private helpers (`entry`, `job_context`, `exception_payload`) so each handler no longer repeats the `level`/`event`/`status`/`job` scaffolding and the trailing `to_json`.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -279,6 +279,12 @@ class Organization < ApplicationRecord
     self
   end
 
+  # Same intent as #organization: a uniform interface so callers walking an arbitrary
+  # object can ask for the organization_id without special-casing Organization itself.
+  def organization_id
+    id
+  end
+
   def maximum_wallets_per_customer
     max_wallets if events_targeting_wallets_enabled?
   end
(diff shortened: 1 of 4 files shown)


vincent-pochet (approved):
Nice

PR #5523, opened 2026-05-18
Fix(finalization): fix double invoice finalization

annvelents:
## Context

We have a case of race condition for pull_taxes_job:
drafted invoice schedules pull_taxes_and_apply_job
at the same time client schedules a finalization that refreshes invoice, moves invoices to pending schedules the job again

next, depending on the timing:
- if first taxes job gets back when draft invoice is refreshed, we're failing to delete applied taxes from the fees (because applied taxes were done in another thread), and fail to refresh draft invoice
- if the first job returns when invoice is already pending, it will progress the invoice status to finalized, but then second job will do the same - we result in double applied taxes and double progressed invoice number

## Description

- forbid to finalize invoice that is draft but pending taxes
- added a new webhook to identify that invoice is ready to be finalized

### DIFF
--- a/app/jobs/send_webhook_job.rb
+++ b/app/jobs/send_webhook_job.rb
@@ -21,6 +21,7 @@ class SendWebhookJob < ApplicationJob
     "invoice.paid_credit_added" => Webhooks::Invoices::PaidCreditAddedService,
     "invoice.generated" => Webhooks::Invoices::GeneratedService,
     "invoice.drafted" => Webhooks::Invoices::DraftedService,
+    "invoice.ready_to_finalize" => Webhooks::Invoices::ReadyToFinalizeService,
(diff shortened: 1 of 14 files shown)


annvelents:
TODO: check if we need to update FE to subscribe to this webhook
Talk with FE to edit the webhook creation form
migration to include this webhook for webhooks with `invoice.draft` included

brunomiguelpinto (approved):

annvelents:
@rsempe , do we need any changes on the CH regarding the new activity_log type?

rsempe (approved):
nice!

rsempe:
> do we need any changes on the CH regarding the new activity_log type?

No, nothing to do on the Clickhouse side 👍

PR #5527, opened 2026-05-18
feat(multi-billing-entity): expose billingEntityId on Subscription

aquinofb:
## Summary

The write side of multi-entity billing for subscriptions has already shipped: `createSubscription` accepts a `billingEntityId` and the row is persisted on `subscriptions.billing_entity_id`. The `Subscription` GraphQL output type, however, did not expose the stored id anywhere, so every `subscription` / `subscriptions` query came back without it. The frontend had no way to tell whether a subscription was bound to a non-default entity, and operators could not see it either.

This PR adds the missing read path with a single new field on `Types::Subscriptions::Object`:

```ruby
field :billing_entity_id, ID, null: true
```

The default GraphQL resolver reads the column directly. The field is nullable because `subscriptions.billing_entity_id` is nullable on legacy rows. When the column is null, the frontend gets `null` back and can decide whether to render the customer's default; when it is set, the frontend sees exactly which entity the subscription is bound to.

The choice to expose a scalar id rather than the full `BillingEntity` object is deliberate. On a list endpoint, returning the nested type would commit the resolver to load the `billing_entities` association the moment any client selects a non-id field, which is a latent N+1 risk on a hot path. The frontend already pre-fetches and caches the billing entities collection on page load, so the id alone is sufficient for display, and this matches the pattern adopted on the wallet output type in #5526.

## Before and after

|  | Before | After |
| --- | --- | --- |
| Read `billingEntityId` from a subscription via GraphQL | Not possible | `subscription(id: "...") { billingEntityId }` |
| Distinguish "subscription has its own entity" from "subscription uses customer's" | Not possible at the API level | `null` vs. UUID |
| Resolver load path | n/a | scalar column read, no association load |
| Other Subscription fields | unchanged | unchanged |

## Test plan

- [ ] Boot the dev stack and open GraphiQL.
- [ ] Run a `createSubscription` mutation passing a non-default `billingEntityId` (an entity other than the customer's default).
- [ ] Take the returned subscription id and query `subscription(id: "...") { billingEntityId }`. Confirm the returned id matches the entity passed to the mutation.
- [ ] For a legacy subscription where `subscriptions.billing_entity_id` is `NULL`, run the same query and confirm `billingEntityId` returns `null` (rather than falling back to the customer's entity).

## Migration

None. The new field is read-only and nullable. Existing GraphQL clients that do not query `billingEntityId` see no change in their responses; the only new behavior is that clients can now request it.

## Out of scope

- Exposing the full `BillingEntity` object on Subscription. Would require a different shape and a custom resolver decision around the model's `customer.billing_entity` fallback. The frontend only needs the id for this round.
- Filter argument on the `subscriptions` resolver.
- `UpdateSubscriptionInput.billingEntityId`.

### DIFF
--- a/app/graphql/types/subscriptions/object.rb
+++ b/app/graphql/types/subscriptions/object.rb
@@ -5,7 +5,7 @@ module Types
     class Object < Types::BaseObject
       graphql_name "Subscription"
 
-      field :billing_entity, Types::BillingEntities::Object, null: false
+      field :billing_entity, Types::BillingEntities::Object, null: true
       field :customer, Types::Customers::Object, null: false
       field :external_id, String, null: false
       field :id, ID, null: false
--- a/app/graphql/types/subscriptions/object.rb
+++ b/app/graphql/types/subscriptions/object.rb
@@ -5,7 +5,7 @@ module Types
     class Object < Types::BaseObject
       graphql_name "Subscription"
 
-      field :billing_entity, Types::BillingEntities::Object, null: true
+      field :billing_entity_id, ID, null: true
       field :customer, Types::Customers::Object, null: false
       field :external_id, String, null: false
       field :id, ID, null: false
--- a/app/graphql/types/subscriptions/object.rb
+++ b/app/graphql/types/subscriptions/object.rb
@@ -5,6 +5,7 @@ module Types
     class Object < Types::BaseObject
       graphql_name "Subscription"
 
+      field :billing_entity, Types::BillingEntities::Object, null: false
       field :customer, Types::Customers::Object, null: false
       field :external_id, String, null: false
       field :id, ID, null: false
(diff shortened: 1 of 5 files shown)


aquinofb:
Following up on feedback received out of band:

> I see this PR exposes the entire BillingEntity object. I think returning only the id would be enough (same approach Mario took on the wallet PR). The FE already pre-fetches the entities collection on page load, cache them and map the id locally. The main reason is on the subscriptions list, where this scales much better. Instead, with the full object we trigger an extra association degrading perf.

Switched to a scalar `billingEntityId: ID`, matching the pattern in #5526. The original full-type exposure came from reading the spec's example query

```
subscription(id: "...") { billingEntity { id code name } }
```

as a "client picks fields" cue and assuming the frontend would just select `{ id }`. The list endpoint N+1 risk and the wallet precedent are clearly the better framing. Applied:

- field changed to `field :billing_entity_id, ID, null: true`
- resolver spec added covering the bound-entity and legacy-row scenarios
- schema dumps regenerated
- PR title and body updated to reflect the scalar-id design

domenicofalco (approved):

mariohd (approved):

PR #5529, opened 2026-05-18
feat(presentation_group_keys): Display presentation breakdowns in invoices

tiagolupepic:
This commit changes the invoice slim templates to also display presentation breakdowns

Continuing the epic of presentation_group_keys, we're changing the templates to also include
the presentation breakdowns.
The main difference of API and GraphQL response and the presentation breakdowns displayed in invoice
is a new method. We're adding the `Charge#presentation_group_keys_values_displayed_in_invoice` to return the keys
that should be displayed in invoice.

In the invoice perspective, we're filtering the breakdowns and then displaying them.

We're also doing some small refactors:
- Adding FeeDisplayHelper.fee_title to use around when we need to display the title of each fee.
- InvoiceSubscriptionsHelper to wrap the logic of loading fees for subscriptions instead of having in the view


vincent-pochet (approved):
👍

app/views/templates/invoices/v4/charge.slim
```diff
@@ -465,6 +465,8 @@ html
       - if applied_invoice_custom_sections.present?
         == SlimHelper.render('templates/invoices/v4/_custom_sections', self)
 
+      == SlimHelper.render('templates/invoices/v4/_presentation_breakdowns', self, fees: fees.includes(:presentation_breakdowns))
```
D1353L:
```suggestion
      == SlimHelper.render('templates/invoices/v4/_presentation_breakdowns', self, fees: fees.includes(:presentation_breakdowns, :charge, :charge_filter))
```
to avoid N+1

PR #5533, opened 2026-05-19
feat(payment-gated-subs): Apply activation rules on upgrade of a subscription

osmarluz:
## Context

Part of the payment-gated subscriptions effort. Activation rules were already supported on subscription creation; this extends them to plan upgrades, so an upgrade can be gated until its rules (e.g. payment) resolve before the new plan takes effect.

## Description

- `Subscriptions::ActivateService` becomes the single entry point for all activations. It auto-detects an upgrade via `subscription.previous_subscription` (different plan, `yearly_amount_cents` ≥ the previous) and branches into a new `activate_for_upgrade` path that terminates the previous, marks the new active, emits fixed-charge events, fires `subscription.started`, and bills both with `:upgrading`.
- `Subscriptions::PlanUpgradeService#call` collapses to `pending!` → optional `ActivationRules::ApplyService` → `ActivateService.call!`. Gated upgrades now follow the same gating path used at subscription creation.
- `Subscriptions::TerminateService#cancel_next_subscription` early-returns when `upgrade: true`, so the upgrade target we just persisted isn't canceled mid-flow.


spec/services/subscriptions/plan_upgrade_service_spec.rb
```diff
@@ -133,6 +133,74 @@
...
+        it "does not fire upgrade-completion webhooks" do
+          result
+          expect(SendWebhookJob).not_to have_been_enqueued.with("subscription.terminated", subscription)
+          expect(SendWebhookJob).not_to have_been_enqueued.with("subscription.started", result.subscription)
+        end
+
+        it "does not enqueue immediate-upgrade billing" do
+          expect { result }.not_to have_enqueued_job(BillSubscriptionJob)
+            .with(anything, anything, hash_including(invoicing_reason: :upgrading))
+        end
```
ancorcruz: shall it assert the upgrade subscription gets billed?

spec/services/subscriptions/activate_service_spec.rb
```diff
@@ -380,4 +380,127 @@
...
+        expect(result.subscription.started_at).to eq(Time.current)
+      end
+    end
+
+    it "does not enqueue Hubspot::UpdateJob for the new subscription" do
+      result
+
+      expect(Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob)
+        .not_to have_been_enqueued.with(subscription: subscription)
+    end
```
ancorcruz: why not? Should we update Hubspot the subscription is now activated?

from POV if we send a webhook we should also communicate it to Hubspot.

spec/services/subscriptions/activate_service_spec.rb
```diff
@@ -380,4 +380,127 @@
...
+
+        expect(SendWebhookJob).to have_been_enqueued.with("subscription.incomplete", subscription)
+      end
+
+      it "does not enqueue a BillSubscriptionJob with :upgrading" do
+        result
+
+        expect(BillSubscriptionJob).not_to have_been_enqueued
+          .with(anything, anything, invoicing_reason: :upgrading)
+      end
```
ancorcruz: Are we missing an assertion on billing the gated subscription here? That one is key.

app/services/subscriptions/activate_service.rb
```diff
@@ -80,6 +95,45 @@ def activate_with_side_effects
...
+
+      EmitFixedChargeEventsService.call!(
+        subscriptions: [subscription],
+        timestamp: subscription.started_at + 1.second
+      )
+
+      after_commit do
+        SendWebhookJob.perform_later("subscription.started", subscription)
+        Utils::ActivityLog.produce(subscription, "subscription.started")
+      end
```
ancorcruz: should we use `#notify_started` instead?

app/services/subscriptions/activate_service.rb
```diff
@@ -80,6 +95,45 @@ def activate_with_side_effects
...
+        subscription: subscription.previous_subscription,
+        upgrade: true
+      )
+
+      subscription.mark_as_active!(timestamp)
+
+      EmitFixedChargeEventsService.call!(
+        subscriptions: [subscription],
+        timestamp: subscription.started_at + 1.second
+      )
```
ancorcruz: could we extract this to a method?

app/services/subscriptions/terminate_service.rb
```diff
@@ -123,6 +123,13 @@ def terminate_and_start_next(timestamp:)
     attr_reader :subscription, :async, :upgrade, :on_termination_credit_note, :on_termination_invoice
 
     def cancel_next_subscription
+      # In the upgrade flow, the next_subscription is the upgrade target we're
+      # transitioning to (created and persisted just before this terminate call),
+      # not a stale scheduled change. Canceling it would mark our new
+      # subscription as canceled mid-flow and leave `canceled_at` set on the
+      # row even after it gets re-activated.
```
ancorcruz: too long, what about a more to the point comment?

```# Upgrade path: next_subscription is the new sub we just persisted, not a stale scheduled change```

File: spec/services/subscriptions/activate_service_spec.rb [hunk 7]
```diff
@@ -380,4 +380,127 @@
       expect(SendWebhookJob).not_to have_been_enqueued
     end
   end
+
+  context "when subscription comes from an upgrade" do
```
ancorcruz, line 384: we are missing a sibling block here to exercise the `activate_from_incomplete` of gated subscriptions that are an plan upgrade. Something like...

```
  context "when an upgrade subscription is incomplete and rules are satisfied" do
    # previous_subscription: active
    # subscription: previous_subscription set, status: :incomplete,
    #               activation_rules all in a satisfied state (so EvaluateService
    #               in activate_from_pending isn't what we're testing — we want
    #               the call to land in activate_from_incomplete)
    #
    # Assert: previous terminated, BillSubscriptionJob enqueued with :upgrading
    # for [previous, subscription] (or [previous] if new plan is pay-in-arrears),
    # subscription.started webhook fires, no Hubspot::UpdateJob for the new sub,
    # no duplicate FixedChargeEvent for the new sub (it was emitted at gate time).
  end
```

ancorcruz (approved):
I'm worried about the gated + upgrade flow actually gets the previous subscription terminated... We should include in this PR a E2E scenario test to ensure the flow is ok. My concern is `ActivateService#activate_from_incomplete` only activates the subscription but not consider the upgrade branch as `#activate_from_pending` does... Is this left on purpose fo a next PR perhaps?

scenario test should live in `spec/scenarios/subscriptions/payment_gated_activation_spec.rb` and has something like...

```
  describe "plan upgrade with payment successful" do
    # 1) Create active subscription on a cheaper plan (no gating).
    # 2) Call create_subscription on a pricier plan with activation_rules.
    #    Assert: previous still active, new sub incomplete, invoice open,
    #            activation_rules.sole pending.
    # 3) simulate_stripe_webhook(status: "succeeded")
    #    Assert: previous terminated, new sub active, BillSubscriptionJob
    #            ran with :upgrading for [previous, new] (or [previous] for
    #            pay-in-arrears new plan), invoice finalized.
  end

  describe "plan upgrade with payment failure" do
    # Same setup; webhook returns "payment_failed".
    # Assert: new sub canceled, previous remains active and untouched.
  end

  describe "plan upgrade with timeout (rules unresolved)" do
    # Optional but valuable — covers what happens when nothing resolves
    # within timeout_hours and the gated upgrade ages out.
  end
```

osmarluz:
> I'm worried about the gated + upgrade flow actually gets the previous subscription terminated... We should include in this PR a E2E scenario test to ensure the flow is ok. My concern is `ActivateService#activate_from_incomplete` only activates the subscription but not consider the upgrade branch as `#activate_from_pending` does... Is this left on purpose fo a next PR perhaps?
> 
> scenario test should live in `spec/scenarios/subscriptions/payment_gated_activation_spec.rb` and has something like...
> 
> ```
>   describe "plan upgrade with payment successful" do
>     # 1) Create active subscription on a cheaper plan (no gating).
>     # 2) Call create_subscription on a pricier plan with activation_rules.
>     #    Assert: previous still active, new sub incomplete, invoice open,
>     #            activation_rules.sole pending.
>     # 3) simulate_stripe_webhook(status: "succeeded")
>     #    Assert: previous terminated, new sub active, BillSubscriptionJob
>     #            ran with :upgrading for [previous, new] (or [previous] for
>     #            pay-in-arrears new plan), invoice finalized.
>   end
> 
>   describe "plan upgrade with payment failure" do
>     # Same setup; webhook returns "payment_failed".
>     # Assert: new sub canceled, previous remains active and untouched.
>   end
> 
>   describe "plan upgrade with timeout (rules unresolved)" do
>     # Optional but valuable — covers what happens when nothing resolves
>     # within timeout_hours and the gated upgrade ages out.
>   end
> ```

@ancorcruz yes, this is going to be handled in a separate PR so it's not too much work on just one PR. It's a separate bullet point on the dive-in

spec/services/subscriptions/plan_upgrade_service_spec.rb
osmarluz: @ancorcruz sure, we can include that as well

spec/services/subscriptions/activate_service_spec.rb
osmarluz: @ancorcruz sure, we can include that as well

spec/services/subscriptions/activate_service_spec.rb
osmarluz: @ancorcruz as mentioned on [the other comment](https://github.com/getlago/lago-api/pull/5533#issuecomment-4497311958), I'll handle this scenario in a separate PR

spec/services/subscriptions/activate_service_spec.rb
osmarluz: @ancorcruz because we never enqueue this on the existing production logic. We just had:

```ruby
after_commit do
  SendWebhookJob.perform_later("subscription.started", new_subscription)
  Utils::ActivityLog.produce(new_subscription, "subscription.started")
end
```

Should we consider it a bug?

app/services/subscriptions/activate_service.rb
osmarluz: @ancorcruz yes, depending on what we decide for [the question above](https://github.com/getlago/lago-api/pull/5533#discussion_r3273903452)

spec/services/subscriptions/activate_service_spec.rb
ancorcruz: It was maybe a bug, I don't really know.

Does it makes sense to notify hubspot the subscription upgrade is being started? for me it does if we notify the previous subscription was terminated (because the upgrade).

We can ask product.

PR #5535, opened 2026-05-19
feat(invoice-preview): add minimum commitment fees

endenis:
## Context

Invoice previews were not including minimum commitment true-up fees.

The existing commitment fee code could not be reused without significant changes.

## Description

- Adds `Fees::Commitments::Minimum::CalculatePreviewFeeService` to compute the commitment fee
- Current period's fees are passed as parameters and historical fees are queried by date range
- Pay-in-advance plans reconcile the previous completed period, matching real invoicing behavior


app/services/fees/commitments/minimum/calculate_preview_fee_service.rb
```diff
@@ -0,0 +1,182 @@
...
+            fixed_charge_fees.sum(:precise_amount_cents) +
+            fixed_charge_in_advance_fees.sum(:precise_amount_cents)
+        end
+
+        def charge_fees
+          charge_fees_base.where(charge: {pay_in_advance: false})
+        end
+
+        def charge_in_advance_fees
+          charge_fees_base.where(charge: {pay_in_advance: true}, pay_in_advance: true)
```
endenis, line 102: We calculate this separately from `charge_fees` because we don't want to take here into account the recurring settlement fees (they have `charge: {pay_in_advance: true}, pay_in_advance: false`).

From what I understood, if they correspond to the current invoicing boundaries they should be already calculated in the `preview_fees_amount_cents`.

Image

app/services/fees/commitments/minimum/build_fee_base_service.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+module Fees
+  module Commitments
+    module Minimum
+      class BuildFeeBaseService < ::BaseService
```
tiagolupepic, line 6: I think this abstraction make sense. We have several services doing `Fee.new` and each of them contains the same logic and/or contains some similarities and they can be abstracted. But, we need to discuss what approach we're going to use it, if I need a Factory concept for example, inheritance or composition.

I'm not against the refactor, and I think it's a good initiative, my the idea is to just align and follow the same approach :)

app/services/fees/commitments/minimum/calculate_preview_fee_service.rb
```diff
@@ -0,0 +1,133 @@
...
+        end
+
+        def proration_coefficient
+          @proration_coefficient ||= days_total.positive? ? days_active / days_total.to_f : 1.0
+        end
+
+        def days_active
+          first_invoice_subscription = subscription.invoice_subscriptions
+            .where("from_datetime >= ?", dates_service.previous_beginning_of_period)
+            .order(Arel.sql("COALESCE(to_datetime, timestamp) ASC"))
```
tiagolupepic: any chance to move this to a `scope`?

app/services/fees/commitments/minimum/calculate_preview_fee_service.rb
```diff
@@ -0,0 +1,133 @@
...
+            .first
```
tiagolupepic: nitpick.. you can use `.limit(1)` + `.pluck` here since you just need a single column. 

it's a small detail and you can ignore it if you want

app/services/invoices/preview_service.rb
```diff
@@ -226,6 +227,37 @@ def add_fixed_charge_fees
       end
     end
 
+    def add_commitment_fees
+      subscriptions.each do |subscription|
```
tiagolupepic: I don't think this is a blocker and it's just a FYI. Calling the new service in this loop will cause a N+1 due the queries happening there. Right now, most of the invoices > 98% contains a single subscription only. I don't think it's the right moment to invest time here and make performance optimizations.

app/services/invoices/preview_service.rb
```diff
@@ -226,6 +227,37 @@ def add_fixed_charge_fees
...
+          to_datetime: boundaries.to_datetime,
+          charges_from_datetime: boundaries.charges_from_datetime,
+          charges_to_datetime: boundaries.charges_to_datetime,
+          timestamp: billing_time
+        )
+
+        preview_fees = invoice.fees.select { |f| f.subscription_id == subscription.id }
+        preview_fees_amount_cents = preview_fees.sum(&:amount_cents)
+        preview_fees_precise_amount_cents = preview_fees.sum(&:precise_amount_cents)
+
```
tiagolupepic: nitpick.. you can optimize this using `.inject({ total_amount_cents: 0, total_precise_amount_cents: 0})` and do the sum in a single loop.

spec/services/invoices/preview_service_spec.rb
```diff
@@ -774,6 +798,42 @@
...
+
+                commitment_fees = result.invoice.fees.select { |f| f.fee_type == "commitment" }
+                expect(commitment_fees.size).to eq(1)
+
+                commitment_fee = commitment_fees.first
+                expect(commitment_fee).not_to be_persisted
+                expect(commitment_fee.invoiceable_type).to eq("Commitment")
+                expect(commitment_fee.invoiceable_id).to eq(commitment.id)
+                expect(commitment_fee.amount_cents).to be > 0
+                expect(commitment_fee.amount_cents).to be <= 1_000_00
```
tiagolupepic: do we should match the exactly value here?

app/services/fees/commitments/minimum/calculate_preview_fee_service.rb
endenis: Done 👍

app/services/fees/commitments/minimum/calculate_preview_fee_service.rb
endenis: Good point, I used `pick` 👍

app/services/invoices/preview_service.rb
endenis: True, this can be optimized in the future

app/services/invoices/preview_service.rb
endenis: Done

spec/services/invoices/preview_service_spec.rb
endenis: Done, thank you

app/services/fees/commitments/minimum/build_fee_base_service.rb
endenis: I didn't go the factory route because after looking at the existing Factories in the codebase (`ChargeModels::Factory`, `BillableMetrics::AggregationFactory`, `Events::CommonFactory`) I saw that they all select a strategy/class at runtime based on some condition.

That pattern doesn't seem to be needed here because both `CreateService` and `CalculatePreviewFeeService` call the same `build_fee` logic.

Inheritance felt like an easy way to reduce code duplication. Composition or a java-style factory class can be more readable and maintainable long-term but I don't have a strong opinion in this specific case.

tiagolupepic: that make sense. I have more context in this [service](https://github.com/getlago/lago-api/blob/main/app/services/fees/create_pay_in_advance_service.rb#L59) for pay in advance and also [this one](https://github.com/getlago/lago-api/blob/main/app/services/fees/charge_service.rb#L265) both shares the same idea. 
I'm also don't have a strong opinion, but definitely we need a better way to handle this.

tiagolupepic (approved):
:shipit:

app/services/fees/commitments/minimum/create_service.rb
```diff
@@ -53,43 +29,9 @@ def call
...
-
         def invoice_has_minimum_commitment_fee?
           invoice.fees.commitment.where(subscription:).any? { |fee| fee.invoiceable.minimum_commitment? }
         end
-
-        def pay_in_advance_first_period?
-          subscription.plan.pay_in_advance? && !reconciliation_invoice_subscription
-        end
-
-        # Returns the invoice_subscription that represents the period being reconciled.
```
annvelents, line 69: Can we keep the comments, they are quite helpful?

endenis: Done, I've added these comments to the build_fee_base_service.rb file

app/services/invoices/preview_service.rb
```diff
@@ -226,6 +227,43 @@ def add_fixed_charge_fees
       end
     end
 
+    def add_commitment_fees
+      subscriptions.each do |subscription|
+        next unless subscription.plan.minimum_commitment
+
+        boundaries = boundaries(subscription)
```
annvelents: spoke offline with @endenis , we want to refactor the service, because now we create :
- all subscription fees for all subscriptions
- all charge fees for all subscriptions
- all fixed charge fees
- all minimum commitment fees

each time inside these methods we're going through the whole array of subscriptions, defining boundaries for each subscirption and only after this create the fees, while it will be better to:
- go though all subscriptions on the top level,
- define boundaries once for each subscription,
- create each type of fee if needed

endenis: @annvelents Thank you, I've applied these changes 👍

spec/services/fees/commitments/minimum/calculate_preview_fee_service_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+            subscription:,
+            invoice: prev_invoice,
+            from_datetime:,
+            to_datetime: prev_to_datetime,
+            timestamp: DateTime.parse("2024-01-01T10:00:00")
+          )
+          create(:fee, fee_type: :subscription, subscription:, invoice: prev_invoice)
+        end
+
+        it "reconciles the previous period, not the current advance period" do
```
annvelents: Can we also add a fee in the current billing period for this test?

annvelents (approved):

PR #5536, opened 2026-05-19
fix(payments): Use provider-reported amount in webhook-built Payments

rsempe:
When a Stripe Checkout Session (or equivalent on Adyen, Cashfree, Flutterwave, MoneyHash) completes against an invoice that has already been settled by another path (typically off-session auto-pay), the webhook handler manufactures a Lago Payment for the incoming PaymentIntent. The Payment row was built using `invoice.total_due_amount_cents`, which at webhook time is 0 if the invoice is already paid.

### DIFF
--- a/app/services/invoices/payments/adyen_service.rb
+++ b/app/services/invoices/payments/adyen_service.rb
@@ -14,9 +14,9 @@ module Invoices
         super
       end
 
-      def update_payment_status(provider_payment_id:, status:, metadata: {})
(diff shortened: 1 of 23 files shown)


annvelents (approved):
LGTM, my only question is: is there4 a possibility that paid amount is > than invoice.total_amount_cents, does it make sence to have any additional guards, reporting for this?

brunomiguelpinto (approved):
Tested Stripe flow all is working fine

PR #5538, opened 2026-05-19
fix(subscriptions): Validate pending taxes on termination

D1353L:
## Context
Terminating a pay-in-advance subscription whose latest invoice was stuck in tax-pending state (provider tax check or VIES validation) crashed with `ArgumentError ('pending' is not a valid status)`. The credit-note creation inherits the invoice status, but `CreditNote.status` only accepts `[draft, finalized]`.

## Description
- Added validation for pending taxes on subscription termination.
- Unblocked invoice preview endpoint by setting "finalized" credit note status.

NOTE: This is a targeted fix for this specific case. The root cause affects multiple areas and needs to be addressed with a comprehensive change.

### DIFF
--- a/app/services/credit_notes/create_service.rb
+++ b/app/services/credit_notes/create_service.rb
@@ -43,7 +43,7 @@ module CreditNotes
           reason:,
           description:,
           credit_status: "available",
-          status: invoice.voided? ? "finalized" : invoice.status # credit notes dont have void state
+          status: credit_note_status
         )
 
         if metadata_value
@@ -161,6 +161,15 @@ module CreditNotes
 
     def non_offset_amounts_present?
       credit_amount_cents.positive? || refund_amount_cents.positive?
+    end
+
+    # NOTE: credit notes only support draft/finalized; voided invoices map to
+    #       finalized, and previews are never persisted so finalized is safe.
+    def credit_note_status
+      return "finalized" if invoice.voided?
+      return "finalized" if context == :preview
+
+      invoice.status
     end
 
     # NOTE: issuing_date must be in customer time zone (accounting date)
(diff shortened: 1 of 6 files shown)


app/services/credit_notes/create_service.rb
```diff
@@ -163,6 +163,15 @@ def non_offset_amounts_present?
       credit_amount_cents.positive? || refund_amount_cents.positive?
     end
 
+    # NOTE: credit notes only support draft/finalized; voided invoices map to
+    #       finalized, and previews are never persisted so finalized is safe.
+    def credit_note_status
+      return "finalized" if invoice.voided?
```
tiagolupepic, line 169: nitpick.. you can have a single guard clause here

app/services/subscriptions/terminate_service.rb
```diff
@@ -224,5 +229,9 @@ def update_on_termination_actions!
 
       Subscriptions::UpdateService.call!(subscription:, params:)
     end
+
+    def blocked_by_pending_taxes?
+      subscription.last_subscription_fee&.invoice&.tax_pending?
```
tiagolupepic: nitpick.. having the operator `&` the result here can be `nil`. So, the question mark for this method returns 3 values, true, false and nil. Maybe we could be explicit 
`subscription.last_subscription_fee&.invoice&.tax_pending? || false`

tiagolupepic (approved):
left two comments and they're not a blocker.

:shipit:

app/services/subscriptions/terminate_service.rb
D1353L: Nice catch 👍

app/services/credit_notes/create_service.rb
D1353L: Decided to keep these guards separate as they are conceptually different

PR #5541, opened 2026-05-20
feat(payment-gated-subs): Apply activation rules on plan upgrades

osmarluz:
## Context

When a plan's amount is updated and a pending upgrade subscription exists, `Plans::UpdateService` and `Plans::UpdateAmountService` trigger `Subscriptions::PlanUpgradeService` to create a new active subscription. The activation rules attached to the pending subscription were dropped in this flow, so payment-gated upgrades silently lost their rules.

## Description

`process_pending_subscriptions` in both services now serializes the pending subscription's `activation_rules` into the `{type:, timeout_hours:}` shape and forwards them via the `params` of `Subscriptions::PlanUpgradeService`, which re-applies them to the new subscription through `Subscriptions::ActivationRules::ApplyService`.

### DIFF
--- a/app/services/plans/update_amount_service.rb
+++ b/app/services/plans/update_amount_service.rb
@@ -41,10 +41,18 @@ module Plans
         next unless subscription.previous_subscription
 
         if plan.yearly_amount_cents >= subscription.previous_subscription.plan.yearly_amount_cents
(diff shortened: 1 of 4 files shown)


ancorcruz (approved):
LGTM!

PR #5542, opened 2026-05-20
fix(invoices): Fix UUID search term

jdenquin:
- Only applies `id` search term when search params is matching `UUID` regex.
- Using `id_cont` will do `uuid::varchar` casting under the hood. This leads to PG not using pkey index, degrading performances with a full table scan.


groyoh (approved):

groyoh (commented):
I guess we want to limit regression here but, with this changes which prevents partial match, I wonder if it makes sense to have id in search_term at all given we have the `GET /api/v1/invoices/:id`.

PR #5543, opened 2026-05-20
feat(exports): Exports all invoices V3

murparreira:
## Context

There were some duplicated data for the invoices view

## Description

Remove redundant LEFT JOIN from the view, as the correlated subquery in the SELECT already handles fetching metadata.


iavorbotev (approved):
🚢

PR #5550, opened 2026-05-21
feat(customers): filter credit notes scalars

aquinofb:
## Context

The customer detail page's "Total amount available" summary card is driven by currency and billing-entity filters that should also scope the card's totals. Today the two scalar fields on `Types::Customers::Object` backing that card cannot be filtered, and `credit_notes_balance_amount_cents` sums `balance_amount_cents` across every finalized credit note regardless of currency, producing a meaningless integer for any customer with credit notes in more than one currency.

## Description

Add optional `currency` and `billing_entity_id` GraphQL arguments to `credit_notes_balance_amount_cents` and `credit_notes_credits_available_count` on `Types::Customers::Object`. When provided, the resolver applies the filters before the aggregate. Defaults of `nil` preserve existing behavior so current callers continue to work unchanged. A private `filtered_credit_notes` helper extracts the shared filter chain so both resolvers stay in sync.

The cross-currency total is fixed at the call site, not the resolver default: the redesigned customer detail view always passes the customer's sticky default currency, so the displayed total is grouped by exactly one currency.

| Before | After |
|---|---|
| `creditNotesBalanceAmountCents` always sums across all currencies (EUR cents + USD cents added as raw integers) | `creditNotesBalanceAmountCents(currency: USD, billingEntityId: "...")` returns only the matching subset |
| `creditNotesCreditsAvailableCount` always counts across all currencies and billing entities | Same field can be filtered to a single currency, a single billing entity, or both |
| No way to scope the summary card to the active UI filters | Frontend can pass the same currency + billing-entity values that drive the credit-notes table |

## How to try locally

Seed a customer with credit notes in multiple currencies and billing entities, then exercise the GraphQL query:

```graphql
query($customerId: ID!, $currency: CurrencyEnum, $billingEntityId: ID) {
  customer(id: $customerId) {
    creditNotesBalanceAmountCents(currency: $currency, billingEntityId: $billingEntityId)
    creditNotesCreditsAvailableCount(currency: $currency, billingEntityId: $billingEntityId)
  }
}
```

Verify:
- No args: returns the existing unfiltered aggregate (legacy behavior preserved).
- `currency: USD`: returns only the USD subtotal.
- `billingEntityId`: returns only credit notes whose invoice belongs to that billing entity.
- Both combined: returns the intersection.

The new behavior is also covered by specs:

```
bundle exec rspec spec/graphql/types/customers/object_spec.rb
```

### DIFF
--- a/app/graphql/types/customers/object.rb
+++ b/app/graphql/types/customers/object.rb
@@ -81,7 +81,10 @@ module Types
       field :credit_notes_balance_amount_cents,
         GraphQL::Types::BigInt,
         null: false,
-        description: "Credit notes credits balance available per customer"
+        description: "Credit notes credits balance available per customer" do
+        argument :billing_entity_id, ID, required: false
+        argument :currency, Types::CurrencyEnum, required: false
+      end
       field :credit_notes_balances,
         [Types::Customers::CreditNotesBalance],
         null: false,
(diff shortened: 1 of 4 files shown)


domenicofalco (dismissed):

domenicofalco (commented):
this may no logger apply. let me clarify in a bit

aquinofb:
No longer needed. Thanks, @domenicofalco

PR #5552, opened 2026-05-21
ING-47 feat(subscriptions): List by billing entity

mariohd:
## Context

Multi-entity customers already filter `GET /api/v1/invoices` and `GET /api/v1/credit_notes` by billing entity through a `billing_entity_codes` list parameter — an array of codes that each endpoint resolves against the organization's billing entities and forwards to the matching query under `billing_entity_ids`. Subscriptions were the remaining gap from task T20 of the multi-entity dive-in (ING-47).

The ticket originally proposed a singular `billing_entity_code: String` parameter on `GET /api/v1/subscriptions`, but the two sibling list endpoints already use the plural array form, so I deliberately departed from the ticket wording and adopted the plural here too. Keeping the three list endpoints aligned avoids introducing a third style on the public REST surface and lets callers pass one or multiple codes from the same URL shape they already use for invoices and credit notes. The single-code path remains the typical case; the plural shape simply doesn't close off the multi-code case for free.

## Description

`SubscriptionIndex` now resolves `params[:billing_entity_codes]` against `current_organization.all_billing_entities` and returns a `billing_entity` not-found error if any code is unknown — the same handshake `InvoiceIndex` and `CreditNoteIndex` already use. The resolved IDs are forwarded to `SubscriptionsQuery` under a new `billing_entity_ids` filter, which applies `where(billing_entity_id: …)` directly against the `subscriptions` table. The query intentionally bypasses the `Subscription#billing_entity` reader's fallback to `customer.billing_entity`, so only subscriptions explicitly stamped under the requested entity are returned — matching the ticket's "stamped under EU" acceptance criterion.

The new parameter is optional, all existing filters keep their behaviour, and the change is purely additive. The shared example `"a subscription index endpoint"` covers single-code, multi-code, and unknown-code scenarios so both `GET /api/v1/subscriptions` and `GET /api/v1/customers/:id/subscriptions` exercise the new surface, and `SubscriptionsQuery` is covered directly at the query level with single-id, multi-id, and blank cases.

`Subscription.ransackable_attributes` is left untouched: the new filter is a direct column predicate, not a ransack search term.

## API surface

```
GET /api/v1/subscriptions?billing_entity_codes[]=EU
GET /api/v1/subscriptions?billing_entity_codes[]=EU&billing_entity_codes[]=US
GET /api/v1/customers/:external_id/subscriptions?billing_entity_codes[]=EU
```

### DIFF
--- a/app/controllers/concerns/subscription_index.rb
+++ b/app/controllers/concerns/subscription_index.rb
@@ -5,9 +5,13 @@ module SubscriptionIndex
   extend ActiveSupport::Concern
 
   def subscription_index(external_customer_id: nil)
+    billing_entities = current_organization.all_billing_entities.where(code: params[:billing_entity_codes]) if params[:billing_entity_codes].present?
+    return not_found_error(resource: "billing_entity") if params[:billing_entity_codes].present? && billing_entities.count != params[:billing_entity_codes].count
+
     filters = params.permit(:plan_code, :overriden, :overridden, :currency, status: [])
     filters[:status] = ["active"] if filters[:status].blank?
     filters[:external_customer_id] = external_customer_id
+    filters[:billing_entity_ids] = billing_entities&.ids
     result = SubscriptionsQuery.call(
       organization: current_organization,
       pagination: {
--- a/app/queries/subscriptions_query.rb
+++ b/app/queries/subscriptions_query.rb
@@ -10,7 +10,8 @@ class SubscriptionsQuery < BaseQuery
     :overriden, # overriden is a legacy typo kept for backward compatibility
     :overridden,
     :exclude_next_subscriptions,
-    :currency
+    :currency,
+    :billing_entity_ids
   ]
 
   def call
@@ -27,6 +28,7 @@ class SubscriptionsQuery < BaseQuery
       SQL
     )
 
+    subscriptions = with_billing_entity_ids(subscriptions) if filters.billing_entity_ids.present?
     subscriptions = with_external_customer(subscriptions) if filters.external_customer_id
     subscriptions = with_plan_code(subscriptions) if filters.plan_code
     subscriptions = with_overridden(subscriptions) unless overridden_filter.nil?
@@ -95,6 +97,10 @@ class SubscriptionsQuery < BaseQuery
 
   def with_currency(scope)
     scope.joins(:plan).where(plans: {amount_currency: filters.currency})
+  end
+
+  def with_billing_entity_ids(scope)
+    scope.where(billing_entity_id: filters.billing_entity_ids)
   end
 
   def with_excluded_next_subscriptions(scope)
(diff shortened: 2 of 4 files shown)


app/controllers/concerns/subscription_index.rb
```diff
@@ -5,9 +5,13 @@ module SubscriptionIndex
   extend ActiveSupport::Concern
 
   def subscription_index(external_customer_id: nil)
+    billing_entities = current_organization.all_billing_entities.where(code: params[:billing_entity_codes]) if params[:billing_entity_codes].present?
+    return not_found_error(resource: "billing_entity") if params[:billing_entity_codes].present? && billing_entities.count != params[:billing_entity_codes].count
```
toommz, line 9: If duplicate codes are in the params, the conditional won't be true. I don't see a clear argument for returning an error when that's the case 🤔 

So maybe that could be enough:

```suggestion
    return not_found_error(resource: "billing_entity") if params[:billing_entity_codes].present? && billing_entities.count != params[:billing_entity_codes].uniq.count
```

app/controllers/concerns/subscription_index.rb
```diff
@@ -5,9 +5,13 @@ module SubscriptionIndex
   extend ActiveSupport::Concern
 
   def subscription_index(external_customer_id: nil)
+    billing_entities = current_organization.all_billing_entities.where(code: params[:billing_entity_codes]) if params[:billing_entity_codes].present?
```
toommz, line 8: Should we just already fetch the ids, so we won't have to `count` in db afterwards and then fetch ids on line `14`?

Like:


```suggestion
    billing_entities = current_organization.all_billing_entities.where(code: params[:billing_entity_codes]).ids if params[:billing_entity_codes].present?
```

toommz (approved):

app/controllers/concerns/subscription_index.rb
mariohd: that's a good point, but

if the user copy/pasted a duplicate? Let's say, 
he has BE code BE1, copy + paste
he has BE code BE2, copy ~+ paste~ (failed or forgot to paste, keep BE1 on clipboard)

the params would be `BE1, BE1` and the results will only look for `BE1` subscriptions.

IMO, if the user typed 2, we should expect 2 different billing entities.

app/controllers/concerns/subscription_index.rb
mariohd: that could potentially save 1 query statement but have an extra memory usage.
We would have first the `SELECT COUNT(1) FROM SUBSCRIPTIONS ...` and other query for `SELECT ID FROM SUBSCRIPTIONS ...`.

In any case, both options are the kinda same and irrelevant in terms of time/resource consumption.

PR #5553, opened 2026-05-21
feat(payment-gated-subs): Apply activation rules on downgrade of a subscription

osmarluz:
## Context

Downgrades create (or mutate) a pending subscription, but `activation_rules` passed in the API params were being ignored on this path — only `CreateService` applied them. Closes that gap as part of payment-gated subs (Milestone 3), keeping behavior consistent with the create flow.

## Description

- `Subscriptions::PlanDowngradeService` now delegates to `Subscriptions::ActivationRules::ApplyService` when `activation_rules` is present in params.
- Applied on both branches: the newly created pending next subscription (downgrade from an active sub) and the current pending subscription (downgrade from a pending sub).

### DIFF
--- a/app/services/subscriptions/plan_downgrade_service.rb
+++ b/app/services/subscriptions/plan_downgrade_service.rb
@@ -16,6 +16,7 @@ module Subscriptions
 
     def call
       if current_subscription.starting_in_the_future?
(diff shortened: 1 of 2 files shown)


ancorcruz (approved):
LGTM

PR #5555, opened 2026-05-21
misc(clickhouse): Define primary key for events_enriched table

vincent-pochet:
## Description

This PR sets the `primary_key` attribute for `events_enriched` and `events_enriched_expanded` tables.
These two tables does not define a primary key and this change allows us to use `in_batch` AR method with a `cursor` to loop over rows.

### DIFF
--- a/app/models/clickhouse/events_enriched.rb
+++ b/app/models/clickhouse/events_enriched.rb
@@ -3,6 +3,7 @@
 module Clickhouse
   class EventsEnriched < BaseRecord
     self.table_name = "events_enriched"
+    self.primary_key = [:organization_id, :code, :external_subscription_id, :timestamp]
   end
 end
 
--- a/app/models/clickhouse/events_enriched_expanded.rb
+++ b/app/models/clickhouse/events_enriched_expanded.rb
@@ -3,6 +3,14 @@
 module Clickhouse
   class EventsEnrichedExpanded < BaseRecord
     self.table_name = "events_enriched_expanded"
+    self.primary_key = [
+      :organization_id,
+      :code,
+      :external_subscription_id,
+      :charge_id,
+      :charge_filter_id,
+      :timestamp
+    ]
   end
 end
 
--- a/spec/models/clickhouse/events_enriched_expanded_spec.rb
+++ b/spec/models/clickhouse/events_enriched_expanded_spec.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Clickhouse::EventsEnrichedExpanded, clickhouse: true do
+  subject(:events_enriched_expanded) { create(:clickhouse_events_enriched_expanded) }
+
+  it "persists a record via the factory" do
+    expect(events_enriched_expanded).to be_persisted
+  end
+
+  describe ".table_name" do
+    it "is events_enriched_expanded" do
+      expect(described_class.table_name).to eq("events_enriched_expanded")
+    end
+  end
+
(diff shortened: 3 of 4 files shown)


app/models/clickhouse/events_enriched.rb
```diff
@@ -3,6 +3,7 @@
 module Clickhouse
   class EventsEnriched < BaseRecord
     self.table_name = "events_enriched"
+    self.primary_key = [:organization_id, :code, :external_subscription_id, :timestamp]
```
toommz, line 6: Could it be a problem that the ORDER BY clause on the table definition is not the same as the primary key defined in the Rails model? (`ORDER BY (organization_id, code, external_subscription_id, toDate(timestamp), timestamp, transaction_id) `)

Also, the primary key defined at the table level is `PRIMARY KEY (organization_id, code, external_subscription_id, toDate(timestamp))` (applying toDate on the timestamp). Here the definition doesn't do that. But that might not be a problem at all 🤷

app/models/clickhouse/events_enriched_expanded.rb
```diff
@@ -3,6 +3,14 @@
...
   class EventsEnrichedExpanded < BaseRecord
     self.table_name = "events_enriched_expanded"
+    self.primary_key = [
+      :organization_id,
+      :code,
+      :external_subscription_id,
+      :charge_id,
+      :charge_filter_id,
+      :timestamp
+    ]
```
toommz, line 13: Here it's the same "discrepancy" between the primary key definition and the table definition.

But still not sure on my side if it's a problem or a deliberate choice 😊

groyoh (approved):
Let's add a simple test to ensure no regression 🙏

app/models/clickhouse/events_enriched.rb
groyoh: Nice catch!

vincent-pochet: @toommz  @groyoh 

Actually, the ORDER BY clause of Clickhouse should be seen as some sort of index that defined how the records are sorted in the table. 
Since we are also defining a Primary key (optional in Clickhouse which would just use the ORDER BY otherwise), I wanted to exposes it this in the model.

Specifying this value in rails does not impact how data is requested or store on the table, it just tell rails that there is no `id` column and that every AR query that rely on ordering (like batched queries) should just use a different set of columns to do so.

Also, it's not  possible to define it that way:
```ruby
self.primary_key =  ["organization_id", "code", "external_subscription_id", "toDate(timestamp)"].map(&:to_sym) # (same without symboles conversion)
```
=> TypeError:
       ["organization_id", "code", "external_subscription_id", "charge_id", "charge_filter_id", "toDate(timestamp)"] is not a symbol nor a string

toommz: Got it! Thanks for the explanation. But does that mean that we can't rely on `.find` for those models? As we can have multiple entries having the same Clickhouse PK it would return records non-deterministically isn't it?

vincent-pochet: Indeed, because of the "eventual consistency" aspect of these tables, we cannot use `find` consistently on these models, it should go through `find_by` or `where().first`.
At any given times, multiple records could match the same "primary key"

toommz: Maybe in the future we could override `find` (and maybe others 🤔) to raise when accessed, but that's outside the scope of this PR

toommz (approved):

PR #5556, opened 2026-05-21
perf(plans): Improve filter-heavy plan updates

D1353L:
## Context

PUT /api/v1/plans/:code is unusably slow when a plan has charges with hundreds of ChargeFilters. On prod, a payload with 8 charges and ~900 filters took ~18s in total with ~10s in the DB. The cost was dominated by per-record overhead inside ChargeFilters::CreateOrUpdateBatchService: redundant validation lookups, PaperTrail bookkeeping on every save, and 3,400+ individual INSERTs.

## Description

All changes are in ChargeFilters::CreateOrUpdateBatchService — the single hot spot of the request:

- **Preload the in-memory parents** on every built or looked-up ChargeFilter/ChargeFilterValue. The with_discarded scope on belongs_to :charge/:charge_filter/:billable_metric_filter defeats Rails automatic inverse_of, so every save was re-SELECTing those parents during presence validation.
- **Use the preloaded values collection instead of find_or_initialize_by**. Each (filter, key) pair was running `filter.values.find_or_initialize_by(...)`, which issues a SQL query per call. Since the values collection is already preloaded, this becomes an in-memory find.
- **Disable PaperTrail for ChargeFilter and ChargeFilterValue during the batch**. PaperTrail's per-save Version Create + the version_limit COUNT query was running thousands of times per request. The parent Plan/Charge still get their own audit entries from their respective services; per-filter granularity is intentionally dropped for this batch path.
- **Bulk-insert new filters via insert_all!**. Pre-generate filter UUIDs so child ChargeFilterValue rows can be wired without a round-trip. AR instances are still built once for validate! to preserve the existing validation contract (and the ActiveRecord::RecordInvalid propagation that callers rely on), but the actual write becomes one INSERT per charge for filters and one for filter values. Matched existing filters keep the per-record save! path.
- **Match existing filters via a hash index instead of a linear scan**. The lookup `filters.find { |f| f.to_h.sort == filter_param[:values].sort }` ran once per filter_param over the full preloaded set. Replaced with filters_by_values_key[values_params.sort], an index built once on the preloaded set.
- **Memoize the billable-metric-filter lookup**. Each pair was issuing `charge.billable_metric.filters.find_by(key:)` with one round-trip per call. Replaced with billable_metric_filters_by_key[key], a single `index_by(&:key)` over the already-loaded set.
- **Enhanced tests**: filter ordering after bulk insert, soft-delete of the removed filter's values, mixed kept/new/removed flow, partial validation failure across filter params.

Performance testing on staging env:
- 8 charges
- 972 filters

Create path (PUT with new charges).

| Metric| Original latency | Optimized latency |
| ------------------ | --------------- | ------------------ |
| total duration | ~36.2 s | ~2.6 s |
| DB time | ~22.0 s | ~1.0 s |
| Ruby allocations | ~8.8 M | ~2.25 M |

Update path (PUT with existing charges).
| Metric| Original latency | Optimized latency |
| ------------------ | --------------- | ------------------ |
| total duration | ~11.04 s| ~2.08 s |
| DB time | ~2.98 s | ~0.70 s |
| Ruby allocations | ~9.28 M| ~1.20 M |

### DIFF
--- a/app/services/charge_filters/create_or_update_batch_service.rb
+++ b/app/services/charge_filters/create_or_update_batch_service.rb
@@ -24,48 +24,36 @@ module ChargeFilters
       # We only care about order when you have less than 100 filters.
       @touch = filters_params.size < 100
 
-      # NOTE: PaperTrail's per-save Version Create plus version_limit COUNT is the
-      #       dominant DB cost for large filter batches (two queries per record save).
-      #       Skip per-filter audit rows for this operation — the parent Plan/Charge
-      #       still gets its own audit entry from their respective services.
-      PaperTrail.request.disable_model(ChargeFilter)
-      PaperTrail.request.disable_model(ChargeFilterValue)
+      ActiveRecord::Base.transaction do
+        @new_filter_rows = []
+        @new_filter_value_rows = []
+        @new_filter_ids = []
 
-      begin
-        ActiveRecord::Base.transaction do
-          @new_filter_rows = []
-          @new_filter_value_rows = []
-          @new_filter_ids = []
+        filters_params.each do |filter_param|
+          values_params = filter_param[:values].transform_keys(&:to_s)
 
-          filters_params.each do |filter_param|
-            values_params = filter_param[:values].transform_keys(&:to_s)
+          # NOTE: since a filter could be a refinement of another one, we have to make sure
+          #       that we are targeting the right one
+          existing_filter = filters_by_values_key[values_params.sort]
 
-            # NOTE: since a filter could be a refinement of another one, we have to make sure
-            #       that we are targeting the right one
-            existing_filter = filters_by_values_key[values_params.sort]
+          properties = ChargeModels::FilterPropertiesService.call(
+            chargeable: charge,
+            properties: filter_param[:properties]&.deep_symbolize_keys&.except(:presentation_group_keys)
+          ).properties
 
-            properties = ChargeModels::FilterPropertiesService.call(
-              chargeable: charge,
-              properties: filter_param[:properties]&.deep_symbolize_keys&.except(:presentation_group_keys)
-            ).properties
-
-            if existing_filter
-              update_existing_filter(existing_filter, filter_param, values_params, properties)
-            else
-              accumulate_new_filter(filter_param, values_params, properties)
-            end
-          end
-
-          bulk_insert_new_filters
-
-          # NOTE: remove old filters that were not created or updated
-          charge.filters.where.not(id: result.filters.map(&:id)).unscope(:order).find_each do
-            remove_filter(it)
+          if existing_filter
+            update_existing_filter(existing_filter, filter_param, values_params, properties)
+          else
+            accumulate_new_filter(filter_param, values_params, properties)
           end
         end
-      ensure
-        PaperTrail.request.enable_model(ChargeFilter)
-        PaperTrail.request.enable_model(ChargeFilterValue)
+
+        bulk_insert_new_filters
+
+        # NOTE: remove old filters that were not created or updated
+        charge.filters.where.not(id: result.filters.map(&:id)).unscope(:order).find_each do
+          remove_filter(it)
+        end
       end
 
       result
@@ -84,8 +72,10 @@ module ChargeFilters
 
       filter.save! if filter.changed?
 
+      PaperTrail.request.disable_model(filter.class)
       # NOTE: Make sure updated_at is touched even if not changed to keep the right order.
       filter.touch if touch # rubocop:disable Rails/SkipsModelValidations
+      PaperTrail.request.enable_model(filter.class)
 
       values_params.each do |key, values|
         billable_metric_filter = billable_metric_filters_by_key[key]
@@ -100,7 +90,9 @@ module ChargeFilters
         filter_value.values = values
         filter_value.save! if filter_value.changed?
 
-        filter_value.touch if @touch # rubocop:disable Rails/SkipsModelValidations
+        PaperTrail.request.disable_model(filter_value.class)
+        filter_value.touch if touch # rubocop:disable Rails/SkipsModelValidations
+        PaperTrail.request.enable_model(filter_value.class)
       end
 
       result.filters << filter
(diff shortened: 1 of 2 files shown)


app/services/charge_filters/create_or_update_batch_service.rb
```diff
@@ -76,12 +61,136 @@ def call
...
+        # NOTE: Make sure updated_at is touched even if not changed to keep the right order.
+        filter.touch # rubocop:disable Rails/SkipsModelValidations
+        PaperTrail.request.enable_model(filter.class)
+      end
+
+      values_params.each do |key, values|
+        billable_metric_filter = billable_metric_filters_by_key[key]
+
+        # NOTE: existing filter was preloaded with values, so this in-memory find avoids a SELECT.
+        filter_value = filter.values.to_a.find { |v| v.billable_metric_filter_id == billable_metric_filter&.id }
```
tiagolupepic: you can use group_by in the filter values outside of this loop and do a O(1) here. something like:

```
# outside of the loop
filter_values_grouped_by = filter.values.group_by(...)
# inside of the loop
filter_values_grouped_by.fetch(billable_metric_filter&.id, nil)
```

app/services/charge_filters/create_or_update_batch_service.rb
```diff
@@ -21,50 +22,34 @@ def call
       return result.single_validation_failure!(field: :values, error_code: "value_is_mandatory") if empty_filter_values?
 
       # We only care about order when you have less than 100 filters.
-      touch = filters_params.size < 100
+      @touch = filters_params.size < 100
```
tiagolupepic: nitpick.. you can move this line to the initialize and add a attr_reader

app/services/charge_filters/create_or_update_batch_service.rb
```diff
@@ -76,12 +61,136 @@ def call
...
+    def update_existing_filter(filter, filter_param, values_params, properties)
+      filter.charge = charge
+      filter.organization = organization
+
+      filter.invoice_display_name = filter_param[:invoice_display_name]
+      filter.properties = properties
+
+      filter.save! if filter.changed?
+
+      if @touch
```
tiagolupepic: out of curiosity, is this should be called only when the filter has changed?

app/services/charge_filters/create_or_update_batch_service.rb
```diff
@@ -21,50 +22,34 @@ def call
       return result.single_validation_failure!(field: :values, error_code: "value_is_mandatory") if empty_filter_values?
 
       # We only care about order when you have less than 100 filters.
-      touch = filters_params.size < 100
+      @touch = filters_params.size < 100
 
       ActiveRecord::Base.transaction do
+        @new_filter_rows = []
+        @new_filter_value_rows = []
+        @new_filter_ids = []
```
tiagolupepic: ditto, any chance to move this to initialize and use attr_reader?

app/services/charge_filters/create_or_update_batch_service.rb
```diff
@@ -76,12 +61,136 @@ def call
...
+      # NOTE: insert_all skips refreshing updated_at, so set
+      #       monotonically-increasing timestamps to preserve the request order under the
+      #       model's `order(updated_at: :asc)` default scope.
+      now = Time.current
+      @new_filter_rows.each_with_index do |row, idx|
+        timestamp = now + (idx / 1_000_000.0)
+        row[:created_at] = timestamp
+        row[:updated_at] = timestamp
+      end
+      ChargeFilter.insert_all!(@new_filter_rows) # rubocop:disable Rails/SkipsModelValidations
```
tiagolupepic: I think/guess you can use `returning` here avoid the extra query below.
Also also use `record_timestamps` delegating this to the AR logic

app/services/charge_filters/create_or_update_batch_service.rb
D1353L: Good idea! Added this with `index_by`

app/services/charge_filters/create_or_update_batch_service.rb
D1353L: funny enough that Rubocop treated `touch` variable as a method and raised `Rails/SkipsModelValidations "Avoid using touch..."`. 
Renamed to `should_touch`

app/services/charge_filters/create_or_update_batch_service.rb
D1353L: According to the intent in the original logic and comments, filter should be touched if not changed to preserve order. As `.save!` refreshes updated_at , `.touch` should be called only in the opposite case.

Added an additional check to reduce an extra query:
`if should_touch && !filter.saved_changes?`

app/services/charge_filters/create_or_update_batch_service.rb
D1353L: Nice catch, added the `returning` here.

Regarding the `record_timestamps` - insert_all applies the same timestamp to all records. As filters are sensitive to order, I believe it's good to keep the custom timestamps (updated a comment to show intention)

app/services/charge_filters/create_or_update_batch_service.rb
tiagolupepic: ohh.. I see! I think we could add a spec to cover this path, wdyt? Just to avoid the next developer accidentally remove the conditional

app/services/charge_filters/create_or_update_batch_service.rb
tiagolupepic: > As filters are sensitive to order, I believe it's good to keep the custom timestamps (updated a comment to show intention)

Didn't know about this detail, you're right, keeping the order makes totally sense

tiagolupepic (approved):
Left a comment to cover the path of `touch`. The rest looks very good!

:shipit:

app/services/charge_filters/create_or_update_batch_service.rb
tiagolupepic: :old-man-yells-rubocop: hehe

PR #5557, opened 2026-05-21
fix(subscriptions): update existing charge on overridden plan in place (#5549)

### LINKED ISSUE #5549: Switching from `plan_overrides` PATCH to per-charge PUT produces duplicate billable charges, with no API path to clean them up
The new per-charge subscription endpoints from [#4959](https://github.com/getlago/lago-api/pull/4959) / [#4963](https://github.com/getlago/lago-api/pull/4963) and the [March 2026 changelog](https://docs.getlago.com/changelog/product) are presented as additive on top of the existing `plan_overrides` PATCH path. In practice the two paths can leave a subscription with **two charges sharing the same `billable_metric_id` and `code`**, and the public API offers no way to delete the extra one. Both charges then generate independent `Fee` rows over the same events.

Tested on `lago-api` `main` at `v1.46.0` (commit `9ef876a`). Premium license required for both code paths.

## Reproduction

1. `PATCH /subscriptions/{external_id}` with `plan_overrides.charges = [{billable_metric_id, properties, filters, ...}]` (no `id`). Repeat once or twice.
2. `PUT /subscriptions/{external_id}/charges/{code}` with new pricing.
3. `GET /subscriptions/{external_id}/charges` → two entries with the same `code` and `billable_metric_id`.
4. Finalize an invoice for the subscription → two `Fee` rows per event window.

## How the duplicate forms

The legacy PATCH path, when the integrator doesn't carry per-charge `id`s in the payload, creates a **top-level** charge on the overridden plan (`parent_id: nil`, base code) via `Plans::UpdateService#process_charges` → `Charges::CreateService`. Its own `sanitize_charges` then keeps that plan tidy on subsequent PATCHes, so the legacy path is self-consistent.

The PUT path then resolves that same top-level charge via `subscription.plan.charges.find_by(code: params[:code])` (`app/controllers/api/v1/subscriptions/charges_controller.rb:82-85`). `Subscriptions::UpdateOrOverrideChargeService` treats it as the "parent" (because its `parent_id` is `nil`) and looks for an existing override whose `parent_id` points at it. There is none — so the service falls into `create_charge_override` and `Charges::OverrideService` `dup`s the charge, creating a second one with the same `code` and `billable_metric_id`. Every subsequent PUT idempotently updates that new override, and the original top-level charge stays as a permanent orphan.

## Why both charges generate fees

The canonical fee path enumerates `subscription.plan.charges` directly, with no `.parents` scope, no `parent_id` filter, and no de-dup:

```ruby
# app/services/invoices/calculate_fees_service.rb:115-140
subscription.plan.charges
  .joins(:billable_metric)
  .where(invoiceable: true)
  .where.not(pay_in_advance: true, billable_metric: {recurring: false})
  .find_each { |charge| Fees::ChargeService.call!(invoice:, charge:, subscription:, ...) }
```

Same shape in `progressive_billing_service.rb:107-114` and `customer_usage_service.rb:69-85`. `Fees::ChargeService#already_billed?` keys uniqueness on `charge_id`, so two charges sharing the same metric produce two `Fee` rows.

## Why it can't be cleaned up via the public API

- `config/routes.rb:93` — `resources :charges, only: %i[index show update]`. No `destroy` on `/subscriptions/{external_id}/charges/{code}`. Confirmed by the documented methods in [`llms.txt`](https://docs.getlago.com/llms.txt).
- The plan-level `DELETE /plans/{code}/charges/{code}` resolves plans via `current_organization.plans.parents.find_by(code:)`, which only matches base plans (DB index is `UNIQUE WHERE parent_id IS NULL`). It can't reach charges on per-subscription overridden plans.
- The GraphQL `DestroyCharge` mutation has the same scope limitation and requires dashboard admin auth.

An API-accessible workaround was attempted: read the override charge's `lago_id` from `GET /subscriptions/{external_id}` (the charge in `plan.charges` whose `parent_id` is non-null), then send `PATCH /subscriptions/{external_id}` with `plan_overrides.charges = [{id: <that lago_id>, billable_metric_id, properties, filters, ...}]`. `Plans::UpdateService#sanitize_charges` does destroy the orphan as a side effect — the duplicate is gone after the PATCH.

However, the resulting state leaves the surviving override with a dangling parent reference, and **every subsequent `PUT /subscriptions/{external_id}/charges/{code}` returns HTTP 500**:

- `app/services/charges/destroy_service.rb:18` discards the orphan with `charge.discard!`, which sets `deleted_at`. The `has_many :children, dependent: :nullify` (`app/models/charge.rb:19`) only nullifies on a hard destroy, not on `discard!`, so the surviving override's `parent_id` continues to point at the now-deleted charge.
- `app/models/charge.rb:62` declares `default_scope -> { kept }`, and the `belongs_to :parent` on line 14 does not opt into `with_discarded` (unlike `:plan` and `:billable_metric` on lines 12-13). `surviving_override.parent` therefore resolves to `nil`.
- `app/services/subscriptions/concerns/charge_override_concern.rb:25-31` returns that `nil` from `find_parent_charge`. The caller in `app/services/subscriptions/update_or_override_charge_service.rb:42-43` then evaluates `target_plan.charges.find_by(parent_id: parent_charge.id)`, raising `NoMethodError: undefined method 'id' for nil`. The error is unhandled by the controller and surfaces as 500.

Empirically reproduced against `api.eu.getlago.com` on May 21 2026 — a single PATCH with the cleanup payload removed the duplicate, and the very next per-charge PUT against that subscription returned `{"status":500,"error":"Internal Server Error"}`. So the PATCH path is not a viable cleanup route in practice: it fixes the duplicate but bricks the per-charge PUT endpoint for that subscription.

## Suggested directions

Any one closes the gap:

1. Add `DELETE /subscriptions/{external_id}/charges/{code}` (and a matching GraphQL mutation) delegating to `Charges::DestroyService`, scoped to the subscription's effective plan.
2. Make `Subscriptions::UpdateOrOverrideChargeService` discover and replace pre-existing top-level charges with the same `billable_metric_id` on the overridden plan, so PUT is self-healing for subscriptions that came through the legacy path.
3. Document the duplicate hazard and a recommended one-shot cleanup in the [March 2026 changelog](https://docs.getlago.com/changelog/product) and the [Update subscription reference](https://www.getlago.com/docs/api-reference/subscriptions/update).

Happy to send a PR for option 1 or 2 if either direction sounds preferable.

nlenepveu:
## Context

Closes #5549.

When a subscription has been updated through the legacy `PATCH /subscriptions/{external_id}` `plan_overrides` path and is later updated through the new per-charge `PUT /subscriptions/{external_id}/charges/{code}` endpoint, the overridden plan ends up holding two charges with the same `billable_metric_id` and `code`. Both are enumerated by the canonical fee-creation path (`app/services/invoices/calculate_fees_service.rb#create_charges_fees`) with no de-duplication, so each event produces two `Fee` rows. The full source trace and a reproduction sequence on `api.eu.getlago.com` are in #5549.

Implements **option 2** from the issue, per @rsempe's [suggestion](https://github.com/getlago/lago-api/issues/5549#issuecomment-4508815865):

> `Subscriptions::UpdateOrOverrideChargeService` will detect when the resolved charge is already on the overridden plan and update it in place rather than creating a duplicate.

## Change

In `Subscriptions::UpdateOrOverrideChargeService#find_or_update_charge_override(target_plan)`, short-circuit when the resolved charge already lives on the overridden plan — update it in place rather than walking up to `find_parent_charge` and creating a new override on top.

```ruby
def find_or_update_charge_override(target_plan)
  return update_charge_override(charge) if charge.plan_id == target_plan.id

  parent_charge = find_parent_charge
  existing_override = target_plan.charges.find_by(parent_id: parent_charge.id)

  if existing_override
    update_charge_override(existing_override)
  else
    create_charge_override(parent_charge, target_plan)
  end
end
```

## Behaviour across the three scenarios

| Subscription state | Resolved `charge.plan_id` vs. `target_plan.id` | Behaviour |
| --- | --- | --- |
| Fresh — on the base plan, no overridden plan yet | `base_plan.id != overridden_plan.id` | Short-circuit doesn't fire. `ensure_plan_override` creates the overridden plan and populates it via `Charges::OverrideService`. Existing branch finds the freshly-created override via `parent_id` match and updates it. **Unchanged.** |
| Already overridden via per-charge PUT — `parent_id` points at base-plan charge | `overridden_plan.id == overridden_plan.id` | Short-circuit fires → updates the override in place. **Same effective behaviour as before**, different code path. |
| Already overridden via legacy `plan_overrides` PATCH — `parent_id: nil`, charge sits directly on the overridden plan via `Plans::UpdateService#process_charges` | `overridden_plan.id == overridden_plan.id` | Short-circuit fires → updates the legacy charge in place. **No second override created. Bug fixed.** |

## Tests

Existing spec contexts in `spec/services/subscriptions/update_or_override_charge_service_spec.rb` continue to pass; new context `"when the charge passed lives directly on the overridden plan with parent_id: nil"` covers scenario 3 explicitly:

- Asserts `Charge.count` does not change (no second override).
- Asserts the existing charge is updated in place (`parent_id` still nil, same `plan_id`, override params applied).
- Asserts the overridden plan has exactly one charge for the `billable_metric` after the call.

## Out of scope

- Cleanup of already-affected production subscriptions. That stays a Lago-side concern (DB / support ticket); this fix only prevents new duplicates from forming.
- The full `DELETE /subscriptions/{external_id}/charges/{code}` endpoint (option 1 from the issue). Left for a follow-up.

cc @rsempe — thanks for the quick triage on #5549.

### DIFF
--- a/app/services/subscriptions/update_or_override_charge_service.rb
+++ b/app/services/subscriptions/update_or_override_charge_service.rb
@@ -39,6 +39,15 @@ module Subscriptions
     attr_reader :subscription, :charge, :params
 
     def find_or_update_charge_override(target_plan)
+      # If the resolved charge already lives on the overridden plan, update
+      # it in place. Without this short-circuit, a charge created directly
+      # on the overridden plan by `Plans::UpdateService#process_charges`
+      # (the legacy `plan_overrides` PATCH path, where `parent_id` is nil)
+      # would end up with a SECOND override layered on top via
+      # `Charges::OverrideService` — same `billable_metric_id`, same `code` —
+      # and both charges would generate fees over the same events.
+      return update_charge_override(charge) if charge.plan_id == target_plan.id
+
       parent_charge = find_parent_charge
       existing_override = target_plan.charges.find_by(parent_id: parent_charge.id)
 
--- a/app/services/subscriptions/update_or_override_charge_service.rb
+++ b/app/services/subscriptions/update_or_override_charge_service.rb
@@ -39,13 +39,7 @@ module Subscriptions
     attr_reader :subscription, :charge, :params
 
     def find_or_update_charge_override(target_plan)
-      # If the resolved charge already lives on the overridden plan, update
-      # it in place. Without this short-circuit, a charge created directly
-      # on the overridden plan by `Plans::UpdateService#process_charges`
-      # (the legacy `plan_overrides` PATCH path, where `parent_id` is nil)
-      # would end up with a SECOND override layered on top via
-      # `Charges::OverrideService` — same `billable_metric_id`, same `code` —
-      # and both charges would generate fees over the same events.
+      # NOTE: If the resolved charge already lives on the overridden plan, update it in place.
       return update_charge_override(charge) if charge.plan_id == target_plan.id
 
       parent_charge = find_parent_charge
--- a/spec/services/subscriptions/update_or_override_charge_service_spec.rb
+++ b/spec/services/subscriptions/update_or_override_charge_service_spec.rb
@@ -127,13 +127,6 @@ RSpec.describe Subscriptions::UpdateOrOverrideChargeService do
       end
 
       context "when the charge passed lives directly on the overridden plan with parent_id: nil" do
-        # Reproduces the state left by `Plans::UpdateService#process_charges`
-        # (the legacy `plan_overrides` PATCH path): a top-level charge sits
-        # directly on the overridden plan with no parent, alongside the
-        # subscription's overridden plan. The service must update this charge
-        # in place rather than layering a second override on top of it,
-        # which would produce two charges with the same `billable_metric_id`
-        # and `code` — both contributing fees over the same events.
         let(:overridden_plan) { create(:plan, organization:, parent: plan) }
         let(:subscription) { create(:subscription, customer:, plan: overridden_plan) }
         let!(:charge) do
--- a/spec/services/subscriptions/update_or_override_charge_service_spec.rb
+++ b/spec/services/subscriptions/update_or_override_charge_service_spec.rb
@@ -126,6 +126,44 @@ RSpec.describe Subscriptions::UpdateOrOverrideChargeService do
         end
       end
 
+      context "when the charge passed lives directly on the overridden plan with parent_id: nil" do
+        # Reproduces the state left by `Plans::UpdateService#process_charges`
+        # (the legacy `plan_overrides` PATCH path): a top-level charge sits
+        # directly on the overridden plan with no parent, alongside the
+        # subscription's overridden plan. The service must update this charge
+        # in place rather than layering a second override on top of it,
+        # which would produce two charges with the same `billable_metric_id`
+        # and `code` — both contributing fees over the same events.
+        let(:overridden_plan) { create(:plan, organization:, parent: plan) }
+        let(:subscription) { create(:subscription, customer:, plan: overridden_plan) }
+        let!(:charge) do
+          create(:standard_charge, plan: overridden_plan, organization:, billable_metric:)
+        end
+
+        it "does not create a second override charge" do
+          expect { service.call }.not_to change(Charge, :count)
+        end
+
+        it "updates the existing top-level charge in place" do
+          result = service.call
+
+          expect(result.charge.id).to eq(charge.id)
+          expect(result.charge.parent_id).to be_nil
+          expect(result.charge.plan_id).to eq(overridden_plan.id)
+          expect(result.charge.invoice_display_name).to eq("Overridden Charge")
+          expect(result.charge.min_amount_cents).to eq(500)
+          expect(result.charge.properties).to eq({"amount" => "150"})
+        end
+
+        it "does not create a duplicate charge for the same billable metric on the overridden plan" do
+          service.call
+
+          charges_for_metric = overridden_plan.charges.where(billable_metric_id: billable_metric.id)
+          expect(charges_for_metric.count).to eq(1)
+          expect(charges_for_metric.first.id).to eq(charge.id)
+        end
+      end
+
       context "with tax_codes" do
         let(:tax) { create(:tax, organization:) }
         let(:params) do


app/services/subscriptions/update_or_override_charge_service.rb
```diff
@@ -39,6 +39,15 @@ def call
     attr_reader :subscription, :charge, :params
 
     def find_or_update_charge_override(target_plan)
+      # If the resolved charge already lives on the overridden plan, update
+      # it in place. Without this short-circuit, a charge created directly
+      # on the overridden plan by `Plans::UpdateService#process_charges`
+      # (the legacy `plan_overrides` PATCH path, where `parent_id` is nil)
+      # would end up with a SECOND override layered on top via
+      # `Charges::OverrideService` — same `billable_metric_id`, same `code` —
+      # and both charges would generate fees over the same events.
```
rsempe:
```suggestion
      # NOTE: If the resolved charge already lives on the overridden plan, update it in place.
```

spec/services/subscriptions/update_or_override_charge_service_spec.rb
```diff
@@ -126,6 +126,44 @@
...
       end
 
+      context "when the charge passed lives directly on the overridden plan with parent_id: nil" do
+        # Reproduces the state left by `Plans::UpdateService#process_charges`
+        # (the legacy `plan_overrides` PATCH path): a top-level charge sits
+        # directly on the overridden plan with no parent, alongside the
+        # subscription's overridden plan. The service must update this charge
+        # in place rather than layering a second override on top of it,
+        # which would produce two charges with the same `billable_metric_id`
+        # and `code` — both contributing fees over the same events.
```
rsempe:
```suggestion
```

rsempe (approved):
Thanks @nlenepveu for the contribution! 👏 😍 

Would be happy to merge it when the two minor comments are resolved (just about comments).

app/services/subscriptions/update_or_override_charge_service.rb
nlenepveu: Done — collapsed the comment block to the single `# NOTE:` line. The full rationale lives in the PR description and #5549. 👍

spec/services/subscriptions/update_or_override_charge_service_spec.rb
nlenepveu: Done — removed the comment block; the context description string already documents the scenario.

nlenepveu:
Thanks for the review @rsempe! 🙏 Both comments are addressed in `7e0357874` (comment-only changes, no behaviour change) — service comment collapsed to a single `# NOTE:` line, and the redundant spec comment removed. Ready whenever you are.

PR #5558, opened 2026-05-21
fix(subscriptions): prevent duplicate external_ids per organization

endenis:
Fixes BIL-44

## Context

Multiple active subscriptions with the same `external_id` could coexist within the same organization for different customers.

The model validation only ran on `:create` missing status changes (like `pending` => `active`).

There was also no database-level unique index.

## Description

In this PR:
- Subscription model validation `validate_external_id` is now set to run on updates too
- Added a service-level cross-customer check in `Subscriptions::CreateService`
- Simplified the "with multuple subscriptions" tests for `SubscriptionsController`. They were by accident creating two active subscriptions and validations now block that
- Did not add any unique index to the database. This should be discussed further and handled separately.


db/migrate/20260520140601_add_unique_index_on_subscriptions_org_external_id_active.rb
```diff
@@ -0,0 +1,25 @@
...
+  def up
+    safety_assured do
+      execute <<-SQL
+        CREATE UNIQUE INDEX CONCURRENTLY index_subscriptions_on_org_external_id_active
+        ON subscriptions (organization_id, external_id)
+        WHERE (
+          status = 1
+          AND (
+            created_at >= '2026-05-20 00:00:00'::timestamp without time zone
+            OR activated_at >= '2026-05-20 00:00:00'::timestamp without time zone
```
vincent-pochet: I guess we will have to check that no new duplicates have been created between `2026-05-20 00:00:00` and the release of this PR.

vincent-pochet (approved):
Looks good

db/migrate/20260520140601_add_unique_index_on_subscriptions_org_external_id_active.rb
endenis: Yes, you're right. Thank you

But I think that I will remove this index from the PR because I am afraid that it can create problems for those who self-host

vincent-pochet: @endenis It could be an issue for sure.
In the meantime, I think that having an index is key to protect ourself from duplicated `external_id`.

What are we planing to do with existing duplicated subscriptions?
A way to handle it would be to have a migration handling some sort of "cleanup" of the duplicated values before creating the index. Is it possible in this situation? (I don't have all the context regarding the initial issue)

endenis: I agree that without the unique index we are not fully safe from this problem.

I don't think that we can automatically clean up existing duplicate subscriptions 😕

One possible way around this limitation is to add a flag column like:
1. First migration adds a `legacy_uniqueness_exception` boolean column that is true by default. This migration also adds a unique index with the condition `where legacy_uniqueness_exception = false`.
2. Second migration sets the default value of the `legacy_uniqueness_exception` column to false => all new entries are covered from this point.
3. Third migration sets `legacy_uniqueness_exception` to false for all old entries that have no duplicates per organization. 

@vincent-pochet @annvelents What do yo think about this approach? It's way heavier than just adding a rails validation but it would enforce uniqueness without problems for self-hosted users.

vincent-pochet: That would work but for sure it's bit heavier... :sweat_smile: 

I'm just curious about the current situation. does it means that today we have multiple customers sharing the same `external_id` on active subscriptions?
If so it will cause billing issue (the usage from events will be shared by the customers), and will have to be fixed anyway (probably manually). Am I missing something?

endenis: > does it means that today we have multiple customers sharing the same external_id on active subscriptions?

Yes but it happens very rarely because there is a rails-level validation on creation. Validation is missing on update (like when a pending subscription with a duplicate external_id is updated to active status).

> If so it will cause billing issue

Currently it raises an error in `Events::PostProcessService`. Sidekiq job retries will fail and it will end up in the dead queue.
https://github.com/getlago/lago-api/blob/7aaec35c60665bd5488e76a0c99bc672dbedc37d/app/services/events/post_process_service.rb#L61

endenis: As discussed with Anna, we will merge this PR without the unique index and later consider further improvements.

PR #5561, opened 2026-05-21
test(fixed-charges): repro subscription endpoint billing bug

ancorcruz:
## Summary

- Failing scenario tests that capture a broken billing flow on `PUT /api/v1/subscriptions/:external_id/fixed_charges/:code` with `apply_units_immediately: true`.
- **Intentionally failing — this PR does NOT include the fix.** It documents the bug as executable evidence so the fix can be scoped and validated.

## The bug

When the dedicated subscription endpoint is called with new units + `apply_units_immediately: true`:

1. A plan override is correctly created with the new units (e.g. 15). ✓
2. The request emits **TWO `FixedChargeEvent` rows** on the override fixed_charge:
   - `units = 10` anchored at the **next-period boundary** (created by `Plans::OverrideService` cloning fixed_charges with `apply_units_immediately` defaulting to false).
   - `units = 15` anchored at "now" (created by the subsequent `update_fixed_charge_override` step).
3. No `Invoices::CreatePayInAdvanceFixedChargesJob` is dispatched, so there's no mid-period delta invoice.
4. At the next regular billing cycle, the aggregation reads the **stale `units = 10` event** sitting at the period boundary instead of the `units = 15` value the merchant intended.

The plan-level endpoint (`PUT /api/v1/plans/:code/fixed_charges/:code`) and the `plan_overrides` path on `Subscriptions::UpdateService` both behave correctly — single `units = 15` event, immediate-billing job dispatched.

## Impact

- **Mid-period uplift never charged** (delta invoice missing).
- **Next billing cycle (and subsequent ones) silently under-bill** the new units.
- For pay-in-advance fixed charges this is direct lost revenue, not just delayed billing.

## Failing tests (3 of 4 in the new describe block)

Located at the end of `spec/scenarios/fixed_charges/pay_in_advance_units_change_spec.rb`:

1. **`creates a single FixedChargeEvent on the override with units = 15 at the update time`** — proves the duplicate-event bug. Expected 1 event with units = 15 anchored at the update time; gets 2 events on the child fixed_charge (one stale at the next-period boundary).
2. **`generates a mid-period delta invoice for 5 units (15 - 10)`** — proves the missing immediate-billing dispatch. Expected 2 invoices (initial + delta); gets only the initial.
3. **`next regular billing cycle bills the new units (15), not the old (10)`** — proves the stale event corrupts next-cycle billing too. Expected next-period invoice at 15 units / \$150; gets 10 units / \$100.

The positive control test (`creates a plan override with the new units`) passes — confirming the override IS correctly applied at the plan/fixed_charge model level. Only the event emission and billing dispatch are broken.

## Test plan

- [ ] `lago exec api bundle exec rspec spec/scenarios/fixed_charges/pay_in_advance_units_change_spec.rb -e "when updating subscription fixed charge via the dedicated endpoint"` reports **4 examples, 3 failures** with the assertions described above.
- [ ] Follow-up PR implements the fix and re-runs; expect 0 failures.

### DIFF
--- a/spec/scenarios/fixed_charges/pay_in_advance_units_change_spec.rb
+++ b/spec/scenarios/fixed_charges/pay_in_advance_units_change_spec.rb
@@ -1403,4 +1403,81 @@ describe "Pay in advance fixed charge units change mid-period", :premium do
       end
     end
   end
+
+  # Bug repro: hitting the dedicated subscription-scoped endpoint
+  # PUT /api/v1/subscriptions/:external_id/fixed_charges/:code with
+  # apply_units_immediately: true silently skips the mid-period delta invoice.
+  # The plan-level endpoint and the plan_overrides path both dispatch
+  # Invoices::CreatePayInAdvanceFixedChargesJob; this one doesn't.
+  describe "when updating subscription fixed charge via the dedicated endpoint" do
+    let(:subscription_date) { DateTime.new(2024, 12, 1) }
+    let(:subscription) { customer.subscriptions.first }
+
+    before do
+      fixed_charge
+
+      travel_to subscription_date do
+        create_subscription(
+          {
+            external_customer_id: customer.external_id,
+            external_id: "sub_endpoint_#{customer.external_id}",
+            plan_code: plan.code,
+            billing_time: "calendar",
+            subscription_at: subscription_date.iso8601
+          }
+        )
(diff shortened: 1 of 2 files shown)


ancorcruz:
Closing this as the test proving the bug was incorporated in PR #5672 fixing the issue.

PR #5562, opened 2026-05-21
[ING-49] feat(payment-requests): filter by entity

mariohd:
## Context

Multi-entity customers need to scope the payment requests list to a specific billing entity. The `/customers`, `/invoices`, and `/wallets` list endpoints already support a `billing_entity_codes` filter; this brings `/payment_requests` in line.

## Description

Adds an optional `billing_entity_codes` array filter to:

- `GET /api/v1/payment_requests`
- `GET /api/v1/customers/:external_id/payment_requests`

Behavior:

- Codes are resolved to IDs against `current_organization.all_billing_entities`. Unknown codes return `404 billing_entity_not_found`, matching the pattern used by `/customers`, `/invoices`, and `/wallets`.
- The query filters via the PR's **invoices' `billing_entity_id`**, not via the customer's default. A payment request's effective billing entity is determined by the invoices it groups — multi-entity customers can have invoices under an entity that differs from the customer's default, so scoping through invoices is the correct semantic.
- Backwards compatible: the filter is optional and existing callers see no change.

### Why a subquery instead of `.joins(:invoices).distinct`?

```ruby
def with_billing_entity_ids(scope)
  scope.where(
    id: PaymentRequest.joins(:invoices)
      .where(invoices: {billing_entity_id: filters.billing_entity_ids})
      .select(:id)
  )
end
```

`DISTINCT` would collide with `apply_consistent_ordering`'s `ORDER BY` — Postgres rejects it with "ORDER BY expressions must appear in select list" when DISTINCT'd columns don't cover the order-by expressions. The subquery sidesteps that and naturally deduplicates payment requests whose multiple invoices fall under the same entity.

## How to test

Manual:

```
GET /api/v1/payment_requests?billing_entity_codes[]=EU
GET /api/v1/payment_requests?billing_entity_codes[]=EU&billing_entity_codes[]=US
GET /api/v1/payment_requests?billing_entity_codes[]=BOGUS   # → 404 billing_entity_not_found
```

A payment request appears in `?billing_entity_codes[]=EU` if **any** of its invoices is stamped under `EU`, regardless of the customer's default entity.

Automated:

```
  spec/queries/payment_requests_query_spec.rb \
  spec/requests/api/v1/payment_requests_controller_spec.rb \
  spec/requests/api/v1/customers/payment_requests_controller_spec.rb
```

44 examples, 0 failures. Tests cover single-code, multi-code, unknown-code (404), and multi-invoice-no-duplicates scenarios.

### DIFF
--- a/app/queries/payment_requests_query.rb
+++ b/app/queries/payment_requests_query.rb
@@ -2,11 +2,12 @@
 
 class PaymentRequestsQuery < BaseQuery
   Result = BaseResult[:payment_requests]
-  Filters = BaseFilters[:external_customer_id, :payment_status, :currency]
+  Filters = BaseFilters[:external_customer_id, :payment_status, :currency, :billing_entity_ids]
 
   def call
     payment_requests = PaymentRequest.where(organization:)
 
+    payment_requests = with_billing_entity_ids(payment_requests) if filters.billing_entity_ids.present?
     payment_requests = with_external_customer(payment_requests) if filters.external_customer_id
     payment_requests = with_payment_status(payment_requests) if filters.payment_status
     payment_requests = with_currency(payment_requests) if filters.currency
@@ -18,6 +19,10 @@ class PaymentRequestsQuery < BaseQuery
   end
 
   private
+
+  def with_billing_entity_ids(scope)
+    scope.joins(:customer).where(customers: {billing_entity_id: filters.billing_entity_ids})
+  end
 
   def with_external_customer(scope)
     scope.joins(:customer).where(customers: {external_id: filters.external_customer_id})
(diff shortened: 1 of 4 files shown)


app/queries/payment_requests_query.rb
```diff
@@ -19,6 +20,10 @@ def call
 
   private
 
+  def with_billing_entity_ids(scope)
+    scope.joins(:customer).where(customers: {billing_entity_id: filters.billing_entity_ids})
```
lovrocolic: It seems to me it makes more sense to filter by billing entity from the associated invoice? wdyt? fetching the one from customer could result in invalid results

mariohd: Sure! Indeed it makes more sense, I will change

lovrocolic (approved):

aquinofb (approved):

PR #5568, opened 2026-05-22
feat(recipes): add org churn rake tasks

aquinofb:
## Context

Closes [ING-220](https://linear.app/getlago/issue/ING-220/deactivate-arxus-staging-api-key).

We periodically need to deactivate API keys and terminate active subscriptions for organizations that have churned (See also `ISSUE-1738`). The procedure was previously runbook-only in [Notion](https://www.notion.so/getlago/Delete-or-expire-an-API-key-342ef63110d28126b1eeca1b8e1add23) and required copy-pasting console snippets — error-prone, undocumented in the codebase, and untested.

## Summary

Two new domain-separated operator-run rake tasks under `lib/tasks/recipes/`:

- `recipes:api_keys:expire_all` — expires every active API key for an organization, bypassing `ApiKeys::DestroyService`'s "last non-expiring key" guard (which is the right call when the org is churning). Each key gets per-value cache invalidation and a security log entry, mirroring `DestroyService`'s side-effects minus the customer-facing mailer.
- `recipes:subscriptions:terminate_all` — terminates active subscriptions via `Subscriptions::TerminateService` synchronously (preserving each sub's `on_termination_credit_note` and `on_termination_invoice` settings) and cancels pending subscriptions. Aborts loudly on the first service failure rather than silently continuing.

Each recipe prompts for the organization, lists the affected records and known side-effects, and requires explicit confirmation before mutating data. They are independent, so compromised-key rotation or mid-engagement subscription cleanup remain possible without one implying the other.

## Side effects worth flagging

- `Subscriptions::TerminateService` emits `subscription.terminated` webhooks, generates final invoices and credit notes per each sub's settings, and triggers Hubspot sync. For a churned customer with no listeners these are wasted but harmless. The confirmation banner surfaces this to the operator.
- The api_keys recipe deliberately skips the destroy mailer (operator action, not customer-initiated) but still emits a per-key `api_key.deleted` security log.

## Test plan

- [x] `lago exec api bundle exec rspec spec/lib/tasks/recipes/` — 24 examples, 0 failures
- [x] `bundle exec rubocop lib/tasks/recipes/*.rake spec/lib/tasks/recipes/*.rb` — clean
- [x] `bundle exec rake -T recipes` lists both new tasks with their descriptions
- [ ] Post-merge: Operator runs both recipes for churned orgs

### DIFF
--- a/lib/tasks/recipes/api_keys.rake
+++ b/lib/tasks/recipes/api_keys.rake
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+require_relative "../../task_prompt"
+
+# rubocop:disable Rails/Output,Rails/Exit
+namespace :recipes do
+  namespace :api_keys do
+    desc "Expire all active API keys for an organization (force-revoke, bypasses last-key guard)"
+    task expire_all: :environment do
+      organization = TaskPrompt.ask_for_organization
+
+      active_keys = organization.api_keys.active.to_a
+
+      if active_keys.empty?
+        puts "No active API keys found for #{organization.name}."
+        next
+      end
+
+      puts ""
+      puts "Found #{active_keys.size} active API key(s) for #{organization.name}:"
+      active_keys.each do |key|
+        puts "  - #{key.name} (••••#{key.value.last(4)})"
+      end
+      puts ""
+      puts "This will force-expire ALL of them, bypassing the 'last non-expiring key' guard."
+      puts "After this, the organization will have NO valid API keys."
+      puts "If the organization is churning, also run `rake recipes:subscriptions:terminate_all`."
+      TaskPrompt.confirm!("Continue? (y/n): ")
+
+      active_keys.each_with_index do |key, index|
+        prefix = "[#{index + 1}/#{active_keys.size}]"
+
+        key.touch(:expires_at) # rubocop:disable Rails/SkipsModelValidations
+        ApiKeys::CacheService.expire_cache(key.value)
+        Utils::SecurityLog.produce(
+          organization:,
+          log_type: "api_key",
+          log_event: "api_key.deleted",
+          resources: {name: key.name, value_ending: key.value.last(4)}
+        )
+
+        puts "#{prefix} Expired #{key.name} (••••#{key.value.last(4)})"
+      end
+
+      puts ""
+      puts "Done. #{active_keys.size} API key(s) expired."
+    end
+  end
+end
+# rubocop:enable Rails/Output,Rails/Exit
--- a/lib/tasks/recipes/api_keys.rake
+++ b/lib/tasks/recipes/api_keys.rake
@@ -30,14 +30,10 @@ namespace :recipes do
       active_keys.each_with_index do |key, index|
         prefix = "[#{index + 1}/#{active_keys.size}]"
 
-        key.touch(:expires_at) # rubocop:disable Rails/SkipsModelValidations
-        ApiKeys::CacheService.expire_cache(key.value)
-        Utils::SecurityLog.produce(
-          organization:,
-          log_type: "api_key",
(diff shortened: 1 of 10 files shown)


vincent-pochet (approved):
Great one 👍

lib/tasks/recipes/api_keys.rake
```diff
@@ -0,0 +1,50 @@
...
+        prefix = "[#{index + 1}/#{active_keys.size}]"
+
+        key.touch(:expires_at) # rubocop:disable Rails/SkipsModelValidations
+        ApiKeys::CacheService.expire_cache(key.value)
+        Utils::SecurityLog.produce(
+          organization:,
+          log_type: "api_key",
+          log_event: "api_key.deleted",
+          resources: {name: key.name, value_ending: key.value.last(4)}
+        )
```
groyoh: I'd suggest adding a `force: false` parameter to `ApiKeys::DestroyService` to bypasse the "last api key" check instead of ad-hoc code here.

lib/tasks/recipes/subscriptions.rake
```diff
@@ -0,0 +1,57 @@
...
+      TaskPrompt.confirm!("Continue? (y/n): ")
+
+      total = active_count + pending_count
+      processed = 0
+
+      organization.subscriptions.active.find_each do |subscription|
+        processed += 1
+        prefix = "[#{processed}/#{total}]"
+
+        result = Subscriptions::TerminateService.call(subscription:, async: false)
```
groyoh, line 35: This might take a while. Maybe we should do it async. What do you think @vincent-pochet  ?

lib/tasks/recipes/subscriptions.rake
```diff
@@ -0,0 +1,57 @@
...
+        if result.failure?
+          abort "#{prefix} Failed to terminate subscription #{subscription.id}: #{result.error}"
+        end
+
+        puts "#{prefix} Terminated #{subscription.id} (was active)"
+      end
+
+      organization.subscriptions.pending.find_each do |subscription|
```
groyoh, line 43: I believe you can use `Subscriptions::TerminateService` for pending as well.

groyoh (commented):
Well done!

Can you also create a `organizations:terminate` recipe which calls the other two recipes please ? I believe we should also ensure we destroy all existing webhook endpoints first to avoid unnecessary noise.

lib/tasks/recipes/api_keys.rake
aquinofb: Done in 74fd206; `ApiKeys::DestroyService` now takes `force: false` (bypasses the last-key guard and skips the mailer).

lib/tasks/recipes/subscriptions.rake
aquinofb: Kept sync intentionally for per-sub visibility + abort-on-first-failure. Matches `Customers::TerminateRelationsService` and `Plans::DestroyService` which also pass `async: false` for bulk-terminate. Happy to revisit if you'd rather async.

lib/tasks/recipes/subscriptions.rake
aquinofb: Done in 74fd206; pending subs now also go through `Subscriptions::TerminateService`.

aquinofb:
Added in 74fd206: `recipes:webhook_endpoints:destroy_all` + `recipes:organizations:terminate` composite (single confirmation, webhooks ~> subs ~> keys order).

aquinofb:
@vincent-pochet and @groyoh, would you like to do a second review? I wanted to add a dry run, but it would cost a lot, so I decided to keep it as is, given that we still have the confirm prompting.

groyoh (approved):

PR #5571, opened 2026-05-25
[ING-217] misc(clickhouse): Add task to enable clickhouse events store

vincent-pochet:
## Context

Today, migrating an organization from the Postgres event store to the Clickhouse one is a manual process that requires few steps to check if an organization can be safely migrated.

The check includes:
- The event count comparison
- The correct enrichment of all events
- The correct current usage for the top 5 subscriptions (by number of received events on the last 30 days)

## Description

This PR automates the checks and enables the clickhouse events store wtith the deduplication turned on.

### DIFF
--- a/lib/tasks/recipes/clickhouse.rake
+++ b/lib/tasks/recipes/clickhouse.rake
@@ -1,0 +1,147 @@
+# frozen_string_literal: true
+
+require_relative "../../task_prompt"
+
+namespace :recipes do
+  namespace :clickhouse do
+    desc "Check if an organization can be safely migrated to ClickHouse events store, and enable it if possible"
+    task enable_clickhouse_events_store: :environment do
+      Rails.logger.level = Logger::Severity::ERROR
+
+      usage_comparison_sample_window = 30.days
+
+      organization = TaskPrompt.ask_for_organization
+      abort "The organization is already using ClickHouse events store." if organization.clickhouse_events_store?
+
+      active_subscriptions_count = Subscription.active.where(organization_id: organization.id).count
+      puts "Organization has #{active_subscriptions_count} active subscription(s)."
+      sample_size_input = TaskPrompt.ask("Number of subscriptions to sample for usage comparison [default: 100]: ")
+      usage_comparison_sample_size = sample_size_input.empty? ? 100 : sample_size_input.to_i
+
+      puts "Step 1: Compare the number of events in Postgres and ClickHouse"
+      max_age = Time.current.beginning_of_hour
+
+      sub_ids = Event.where(organization_id: organization.id).select("DISTINCT(external_subscription_id)")
+      postgres_count = Event.with_discarded
+        .where(organization_id: organization.id)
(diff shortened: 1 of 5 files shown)


toommz (approved):

lib/tasks/recipes/clickhouse.rake
```diff
@@ -0,0 +1,155 @@
...
+require_relative "../../task_prompt"
+
+namespace :recipes do
+  namespace :clickhouse do
+    desc "Check if an organization can be safely migrated to ClickHouse events store, and enable it if possible"
+    task enable_clickhouse_events_store: :environment do
+      Rails.logger.level = Logger::Severity::ERROR
+
+      usage_comparison_sample_window = 30.days
+      usage_comparison_sample_size = 5
```
groyoh: Should we make these inputs ? I think we should compare a bit more than 5 subs.

lib/tasks/recipes/clickhouse.rake
```diff
@@ -0,0 +1,155 @@
...
+    apply_taxes: false
+  )
+  legacy_fees = legacy_result.usage&.fees || []
+
+  enriched_fees = []
+  ActiveRecord::Base.transaction do
+    organization.update!(
+      clickhouse_events_store: true,
+      clickhouse_deduplication_enabled: true
+    )
```
groyoh: This is a bit dangerous. 😅  Is there no way around this ? Could we maybe add a parameter to usage to force event store ?

groyoh (approved):

lib/tasks/recipes/clickhouse.rake
vincent-pochet: Yes, it's far from ideal. I'll check how I could inject the store in the process. Since it could be re-used for other use case it worth the effort

spec/services/events/stores/store_factory_spec.rb
```diff
@@ -58,4 +58,34 @@
...
+    end
+
+    it "restores a previous override when nested" do
+      described_class.with_override(store_class: Events::Stores::ClickhouseStore, deduplicate: true) do
+        described_class.with_override(store_class: Events::Stores::PostgresStore, deduplicate: false) do
+          expect(described_class.store_class(organization:)).to eq(Events::Stores::PostgresStore)
+        end
+        expect(described_class.store_class(organization:)).to eq(Events::Stores::ClickhouseStore)
+      end
+    end
```
toommz: I'm ok with this implementation, but I would rather have raised when nesting overrides, wdyt?

groyoh (approved):

spec/services/events/stores/store_factory_spec.rb
groyoh: That's a good idea.

vincent-pochet: It makes total sense 👍
I will update it

toommz (approved):

PR #5575, opened 2026-05-25
fix(past-usage): Add customer_usage key to past_usage serializer response

murparreira:
## Context

The response shape for the past_usage endpoint does not match our [current API spec](https://getlago.com/docs/api-reference/customer-usage/get-past).

## Description

Add customer_usage key to past_usage serializer response so it will match our API spec.


toommz (approved):

mariohd (approved):

groyoh (commented):
This looks scary as it is a breaking change. Is this really what we want (compared to updating the doc to match this) ?

aquinofb (approved):

PR #5576, opened 2026-05-26
feat (quotes): Add new `currency`, `start_date` and `end_date` fields

rinasergeeva:
Following the recent spec changes, this PR adds the following three fields to the Quote:

* `currency`
* `start_date`
* `end_date`

and exposes them via GraphQL.

### DIFF
--- a/db/migrate/20260525102114_add_currency_and_subscription_dates_to_quote_versions.rb
+++ b/db/migrate/20260525102114_add_currency_and_subscription_dates_to_quote_versions.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddCurrencyAndSubscriptionDatesToQuoteVersions < ActiveRecord::Migration[8.0]
+  def change
+    add_column :quote_versions, :currency, :string
+    add_column :quote_versions, :start_date, :date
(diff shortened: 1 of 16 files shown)


murparreira (approved):

db/migrate/20260525102114_add_currency_and_subscription_dates_to_quote_versions.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddCurrencyAndSubscriptionDatesToQuoteVersions < ActiveRecord::Migration[8.0]
+  def change
+    add_column :quote_versions, :currency, :string
```
toommz, line 5: Should we set it as `not null`? Or is it impossible to know at creation time which currency to use?

toommz (approved):

db/migrate/20260525102114_add_currency_and_subscription_dates_to_quote_versions.rb
rinasergeeva: I'm not sure about `currency` usage at this point to implement such a restriction. I guess we can introduce it later if it's needed?

PR #5577, opened 2026-05-26
include deleted invoices

brunomiguelpinto:
## Context

The `check_stripe_payment` console helper added in #5574 was using `Invoice.with_discarded` to resolve the and it should be unscoped


annvelents (approved):
🙈

PR #5578, opened 2026-05-26
feat(payment-gated-subs): add polymorphic refundable to refunds

ancorcruz:
## Context

Milestone 4 of payment-gated subscription activation introduces automatic refunds for the late-payment-after-timeout race condition. The refund flow needs to create Refund records for payments tied to closed invoices (not to credit notes), so the existing Refund schema needs to accept refundables other than CreditNote.

 ## Description

Extends the refunds table to support payment-only refunds:

- Adds a polymorphic refundable association (refundable_type + refundable_id) with a composite index, so a Refund can target a CreditNote (existing use) or another model such as Invoice (M4 use case).
- Drops the NOT NULL on credit_note_id so activation-expired refunds can be persisted without a credit note.
- Adds a reason string column to distinguish credit-note refunds from subscription-activation-expired refunds at the data layer.

Model and service changes that consume the new columns ship in follow-up PRs.

### DIFF
--- a/db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
+++ b/db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddPolymorphicRefundableToRefunds < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :refunds, :refundable, type: :uuid, polymorphic: true, index: {algorithm: :concurrently}
+
+    safety_assured do
+      change_column_null :refunds, :credit_note_id, true
+    end
+
+    add_column :refunds, :reason, :string
+  end
+end
--- a/db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
+++ b/db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
@@ -6,10 +6,13 @@ class AddPolymorphicRefundableToRefunds < ActiveRecord::Migration[8.0]
   def change
     add_reference :refunds, :refundable, type: :uuid, polymorphic: true, index: {algorithm: :concurrently}
 
+    add_column :refunds, :reason, :string
+
+    # Last in forward = first in reverse. If any row has credit_note_id
+    # NULL (M4 activation refunds), rollback aborts here before any other
+    # column is dropped, leaving the schema intact.
     safety_assured do
       change_column_null :refunds, :credit_note_id, true
     end
-
-    add_column :refunds, :reason, :string
   end
 end
--- a/app/models/refund.rb
+++ b/app/models/refund.rb
@@ -18,15 +18,18 @@ end
 #  id                           :uuid             not null, primary key
 #  amount_cents                 :bigint           default(0), not null
 #  amount_currency              :string           not null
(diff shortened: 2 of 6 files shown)


db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class AddPolymorphicRefundableToRefunds < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :refunds, :refundable, type: :uuid, polymorphic: true, index: {algorithm: :concurrently}
+
+    safety_assured do
+      change_column_null :refunds, :credit_note_id, true
+    end
```
toommz, line 16: Do we rollback database migrations sometimes in production? A failed deployment for example, or anything like that?

Because this seems safe in the up direction, but not really on the down side.

db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddPolymorphicRefundableToRefunds < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :refunds, :refundable, type: :uuid, polymorphic: true, index: {algorithm: :concurrently}
```
toommz, line 7: Just asking, is there a reason not to add the association in the model right now?

toommz (approved):

db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    safety_assured do
+      change_column_null :refunds, :credit_note_id, true
+    end
+
+    add_column :refunds, :reason, :string
+  end
```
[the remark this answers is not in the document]
ancorcruz: we are not going to do a backfill now.

db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
```diff
@@ -0,0 +1,15 @@
...
+class AddPolymorphicRefundableToRefunds < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :refunds, :refundable, type: :uuid, polymorphic: true, index: {algorithm: :concurrently}
+
+    safety_assured do
+      change_column_null :refunds, :credit_note_id, true
+    end
+
```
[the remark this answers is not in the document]
ancorcruz: added in a separate migration

db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
ancorcruz: just keeping the change small, next PR will include models and relationship changes.

db/migrate/20260526131452_add_polymorphic_refundable_to_refunds.rb
ancorcruz: We embrace non reversible migrations. For example this one can not be rolled back once we add new data using these new columns.

PR #5580, opened 2026-05-26
feat(payment-gated-subs): Cover activation of incomplete subscriptions coming from upgrades

osmarluz:
## Context

Builds on #5533, which gated subscription upgrades only from the from-pending path. A gated upgrade actually completes from the *incomplete* state, once payment resolves — this PR covers that path.

## Description

- `Subscriptions::ActivateService#activate_from_incomplete` now routes through `activate_for_upgrade` when the subscription has a `previous_subscription`, mirroring the pending path.
- `activate_for_upgrade` distinguishes from-incomplete: skips re-emitting fixed-charge events (already done at gating) and bills only the previous subscription with `:upgrading` (the new subscription was billed with `skip_charges` when gated).
- The billing list now flows as an explicit argument into `bill_upgrade_subscriptions` instead of through an instance variable.

### DIFF
--- a/app/services/subscriptions/activate_service.rb
+++ b/app/services/subscriptions/activate_service.rb
@@ -104,9 +104,10 @@ module Subscriptions
       subscription.mark_as_active!(timestamp)
 
       billable_subscriptions = [subscription.previous_subscription]
+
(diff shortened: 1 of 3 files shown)


osmarluz:
Replaced by #5586

PR #5582, opened 2026-05-26
[ING-235] feat(wallets): allow target rules to grant credits

groyoh:
## Context

Wallet recurring transaction rules of type `target` currently always create a paid top-up. Free credits are only available via fixed top-ups (the existing `granted_credits` decimal column). The limitation lives in the API itself, so it blocks both UI consumers and direct API consumers.

## Description

A new boolean column `grants_target_top_up` is added to `recurring_transaction_rules`. The column is nullable: target rules carry `true` or `false`, non-target rules carry `nil` (the flag does not apply). Model validations enforce this (`inclusion: [true, false]` when target, `exclusion: [true, false]` otherwise). A backfill migration sets `false` on existing target rules.

When `true` on a target rule, the gap-fill at top-up time grants credits (creates a `granted` wallet transaction) instead of paying for them. Fixed rules are unchanged: paid vs granted stays implicit via the existing `paid_credits` and `granted_credits` decimal amounts.

`Wallets::RecurringTransactionRules::CreateService` defaults new target rules to `grants_target_top_up: false` when not provided. `UpdateService` normalizes the payload on method transitions (switching to target without specifying defaults to false; switching to fixed forces nil). `ValidateService` rejects any non-nil value when the method isn't target and surfaces the existing `invalid_recurring_rule` error code.

`RecurringTransactionRule#compute_paid_credits` was updated to honor the flag. `#compute_granted_credits` was updated to compute the dynamic grant amount and now reads `wallet.credits_ongoing_balance` directly instead of taking it as a parameter (the two callers were passing exactly that value). The private helper was renamed from `compute_target_paid_credits` to `compute_target_top_up_amount` since both compute methods share it.

Granted gap-fills bypass `paid_top_up_min`. The limit exists to avoid processing tiny payments, and a granted top-up has no payment, so the floor has nothing to constrain. A wallet with a $50 paid minimum that grants a $1 gap will grant $1, not over-shoot the target.

### DIFF
--- a/app/models/recurring_transaction_rule.rb
+++ b/app/models/recurring_transaction_rule.rb
@@ -15,6 +15,8 @@ class RecurringTransactionRule < ApplicationRecord
     source: :invoice_custom_section
 
   validates :transaction_name, length: {minimum: 1, maximum: 255}, allow_nil: true
+  validates :grants_target_top_up, inclusion: {in: [true, false]}, if: :target?
+  validates :grants_target_top_up, exclusion: {in: [true, false]}, unless: :target?
 
   STATUSES = [
     :active,
@@ -70,7 +72,9 @@ class RecurringTransactionRule < ApplicationRecord
 
   def compute_paid_credits(ongoing_balance:)
     if target?
-      compute_target_paid_credits(ongoing_balance:)
+      return 0.0 if grants_target_top_up?
+
+      compute_target_top_up_amount(ongoing_balance:)
     else
       paid_credits
     end
@@ -78,6 +82,8 @@ class RecurringTransactionRule < ApplicationRecord
 
   def compute_granted_credits
     if target?
+      return compute_target_top_up_amount(ongoing_balance: wallet.credits_ongoing_balance) if grants_target_top_up?
+
       0.0
     else
       granted_credits
@@ -86,12 +92,15 @@ class RecurringTransactionRule < ApplicationRecord
 
   private
 
-  def compute_target_paid_credits(ongoing_balance:)
+  def compute_target_top_up_amount(ongoing_balance:)
     if ongoing_balance >= target_ongoing_balance
       return 0.0
     end
 
     gap = target_ongoing_balance - ongoing_balance
+
+    # NOTE: granted top-ups skip the paid_top_up_min limit since no payment occurs
+    return gap if grants_target_top_up?
 
     # NOTE: in case of target rule, we don't apply max because reaching target balance is the most important
     apply_min_top_up_limits(credit_amount: gap)
@@ -106,6 +115,7 @@ end
 #  id                                  :uuid             not null, primary key
 #  expiration_at                       :datetime
 #  granted_credits                     :decimal(30, 5)   default(0.0), not null
+#  grants_target_top_up                :boolean
 #  ignore_paid_top_up_limits           :boolean          default(FALSE), not null
 #  interval                            :integer          default("weekly")
 #  invoice_requires_successful_payment :boolean          default(FALSE), not null
--- a/app/services/wallets/recurring_transaction_rules/update_service.rb
+++ b/app/services/wallets/recurring_transaction_rules/update_service.rb
@@ -35,6 +35,8 @@ module Wallets
 
           recurring_rule = wallet.recurring_transaction_rules.active.find_by(id: lago_id)
 
+          normalize_grants_target_top_up!(rule_attributes, recurring_rule)
+
           if rule_attributes.key?(:invoice_custom_section)
             invoice_custom_section = {
               invoice_custom_section: rule_attributes.delete(:invoice_custom_section)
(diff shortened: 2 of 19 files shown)


app/models/recurring_transaction_rule.rb
```diff
@@ -15,6 +15,8 @@ class RecurringTransactionRule < ApplicationRecord
     source: :invoice_custom_section
 
   validates :transaction_name, length: {minimum: 1, maximum: 255}, allow_nil: true
+  validates :grants_target_top_up, inclusion: {in: [true, false]}, if: :target?
+  validates :grants_target_top_up, exclusion: {in: [true, false]}, unless: :target?
```
lovrocolic, line 19: Is it safe to add this validation before we are sure the backfill is done? Should the validation `allow_nil` for a release?

lovrocolic (approved):

app/services/wallets/recurring_transaction_rules/update_service.rb
```diff
@@ -97,6 +99,20 @@ def hash_recurring_rules
         @hash_recurring_rules ||= params.map { |m| m.to_h.deep_symbolize_keys }
       end
 
+      def normalize_grants_target_top_up!(rule_attributes, recurring_rule)
+        effective_method = rule_attributes[:method]&.to_s || recurring_rule&.method
+
+        if effective_method == "target"
+          if rule_attributes.key?(:grants_target_top_up)
+            rule_attributes[:grants_target_top_up] = ActiveModel::Type::Boolean.new.cast(rule_attributes[:grants_target_top_up])
```
toommz, line 107: We don't do the same `|| false` default assignment as in the create service, is it intended?

spec/factories/recurring_transaction_rules.rb
```diff
@@ -9,5 +9,9 @@
     interval { "monthly" }
     trigger { "interval" }
     transaction_name { "Recurring Transaction Rule" }
+
+    after(:build) do |rule|
+      rule.grants_target_top_up = false if rule.method.to_s == "target" && rule.grants_target_top_up.nil?
+    end
```
toommz, line 15: What about having traits taking care of initializing the object with the right values? My main argument to do so would be that it adds a bit of "logic" to the factory to do that, and I'm not sure that factories should have some logic that the bare activerecord objects do not have 🤔

toommz (approved):
Reviewed as far as I could regarding the logic, but I admit that `#compute_paid_credits` and `#compute_target_top_up_amount` are blackboxes to me.

app/services/wallets/recurring_transaction_rules/update_service.rb
groyoh: Yes. This is to avoid setting `false` when the setting was not changed. For instance, if it was already `true`, you wouldn't want it to change.

spec/factories/recurring_transaction_rules.rb
groyoh: Sure.

app/models/recurring_transaction_rule.rb
groyoh: I believe so yes. You shouldn't deploy a new version unless the migrations for that version already ran. So this code will be deployed AFTER the migration has run.

What could happen though is a race-condition caused by creation of recurring rule after migration has run and before the code gets deployed.

groyoh: Maybe you're right and this should be a two-step deploy.

lovrocolic: Oh sorry, somehow I thought for a sec that backfill is async, that's why I asked it.
You are right about the race condition, but probably not a blocker IMO

app/models/recurring_transaction_rule.rb
```diff
@@ -15,6 +15,8 @@ class RecurringTransactionRule < ApplicationRecord
...
```
groyoh, line 20: To add:

```suggestion
def grants_target_top_up?
  super || false
end
```

db/migrate/20260528175021_backfill_grants_target_top_up_on_target_rules.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class BackfillGrantsTargetTopUpOnTargetRules < ActiveRecord::Migration[8.0]
```
groyoh, line 3: Let's remove this and keep it to `nil` for now when not set and treat `nil` as `false`.

spec/models/recurring_transaction_rule_spec.rb
```diff
@@ -138,6 +178,30 @@
       it "returns zero" do
         expect(subject).to eq 0.0
       end
+
+      context "when grants_target_top_up is true" do
```
groyoh, line 182: To add some tests when `grants_target_top_up` is `nil`.

app/models/recurring_transaction_rule.rb
groyoh: I discuss with @aquinofb to handle your initial feedback @lovrocolic. It does make sense.

aquinofb:
Superseded by #5755, which carries this work forward with the approach from review: the validation allows `nil` (no backfill, existing rows treated as `false`), and the column migration is included as its own commit. Closing in favour of that PR.

PR #5583, opened 2026-05-27
fix(events): preserve event timestamp precision

tiagolupepic:
## Context

Pay-in-advance aggregation filters uses persisted events with a max timestamp from the serialized common event in order to perform the aggregations. Serializing only with .to_f can round microseconds down example: `2026-05-22 10:04:50.227587000 +0000` becomes `227586` usec after `Time.zone.at(timestamp.to_f)`, which can exclude the current event from grouped_sum query or other aggregation that depends on `max_timestamp`.

Screenshot 2026-05-27 at 11 34 05

## Description

Serialize an additional `iso8601(9)` timestamp so hash-based common events can be rebuilt with nanosecond precision while keeping the existing float timestamp for compatibility. Fall back to the `.to_f` path when the precise timestamp is missing or cannot be parsed.

### DIFF
--- a/app/models/events/common.rb
+++ b/app/models/events/common.rb
@@ -46,7 +46,22 @@ module Events
     end
 
     def as_json
-      super.tap { |j| j["timestamp"] = timestamp.to_f }
+      super.tap do |j|
+        j["timestamp"] = timestamp.to_f
+        j["timestamp_with_precision"] = timestamp.iso8601(9)
+      end
+    end
+
+    def self.timestamp_from_source(source)
+      timestamp = Time.zone.parse(source["timestamp_with_precision"])
+
+      if timestamp.present?
+        return timestamp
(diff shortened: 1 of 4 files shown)


vincent-pochet (approved):
Nice catch! 🚀

groyoh:
@tiagolupepic I believe using `.to_d` instead of `.to_f` might have also worked. We started to fix this issue a few months back but it got lost in between rebases and PR split.

tiagolupepic:
@groyoh tried with .to_d and AS TimeWithZone doesn't implement `.to_d`

```
NoMethodError: undefined method 'to_d' for an instance of ActiveSupport::TimeWithZone (NoMethodError)
```

Probably it needs to extend to make it work smoothly. But yes, using BigDecimal could work as well

PR #5585, opened 2026-05-27
feat(order-forms): add sign lifecycle REST

toommz:
## Roadmap Task

## Context

Once an order form has been generated for an approved quote version, the next step in its lifecycle is signature. There was no way to mark an order form as signed through the REST API, so the `generated → signed` transition could not be driven by clients.

## Description

Add a `mark_as_signed` lifecycle action to order forms.

### DIFF
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -37,7 +37,7 @@ module Api
           order_form:,
           signed_document: mark_as_signed_params[:signed_document],
           execution_mode: mark_as_signed_params[:execution_mode],
-          execution_date: mark_as_signed_params[:execution_date]
+          execute_at: mark_as_signed_params[:execute_at]
         )
 
         if result.success?
@@ -75,7 +75,7 @@ module Api
       end
 
       def mark_as_signed_params
-        params.fetch(:order_form, {}).permit(:signed_document, :execution_mode, :execution_date)
+        params.fetch(:order_form, {}).permit(:signed_document, :execution_mode, :execute_at)
       end
 
       def render_order_form(order_form)
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -30,6 +30,18 @@ module Api
         end
       end
 
+      def mark_as_signed
+        order_form = current_organization.order_forms.find_by(id: params[:id])
+
+        result = OrderForms::MarkAsSignedService.call(order_form:)
+
+        if result.success?
+          render_order_form(result.order_form)
+        else
+          render_error_response(result)
+        end
+      end
+
       def show
         order_form = current_organization.order_forms.find_by(id: params[:id])
         return not_found_error(resource: "order_form") unless order_form
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -35,7 +35,9 @@ module Api
 
         result = OrderForms::MarkAsSignedService.call(
           order_form:,
-          signed_document: mark_as_signed_params[:signed_document]
+          signed_document: mark_as_signed_params[:signed_document],
+          execution_mode: mark_as_signed_params[:execution_mode],
+          execution_date: mark_as_signed_params[:execution_date]
         )
 
         if result.success?
@@ -73,7 +75,7 @@ module Api
       end
 
       def mark_as_signed_params
-        params.fetch(:order_form, {}).permit(:signed_document)
+        params.fetch(:order_form, {}).permit(:signed_document, :execution_mode, :execution_date)
       end
 
       def render_order_form(order_form)
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -33,7 +33,10 @@ module Api
       def mark_as_signed
         order_form = current_organization.order_forms.find_by(id: params[:id])
 
-        result = OrderForms::MarkAsSignedService.call(order_form:)
+        result = OrderForms::MarkAsSignedService.call(
+          order_form:,
+          signed_document: mark_as_signed_params[:signed_document]
+        )
 
         if result.success?
           render_order_form(result.order_form)
@@ -67,6 +70,10 @@ module Api
           expires_at_from: params[:expires_at_from],
           expires_at_to: params[:expires_at_to]
         }
+      end
+
+      def mark_as_signed_params
+        params.fetch(:order_form, {}).permit(:signed_document)
       end
 
       def render_order_form(order_form)
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -75,7 +75,7 @@ module Api
       end
 
       def mark_as_signed_params
-        params.fetch(:order_form, {}).permit(:signed_document, :execution_mode, :execute_at)
+        params.permit(order_form: [:signed_document, :execution_mode, :execute_at]).fetch(:order_form, {})
       end
 
       def render_order_form(order_form)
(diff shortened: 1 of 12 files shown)


app/services/order_forms/mark_as_signed_service.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+module OrderForms
+  class MarkAsSignedService < BaseService
+    Result = BaseResult[:order_form, :order]
+
+    EXECUTION_MODES = %w[execute_in_lago order_only].freeze
```
toommz, line 9: Will live in `Order` when the model will be introduced

app/services/utils/base64_file.rb
toommz, line 1: Outside the scope of this PR but this behavior is spread in 2 or 3 other places, so could be worth using in those places too.

murparreira (approved):

app/services/order_forms/mark_as_signed_service.rb
```diff
@@ -0,0 +1,111 @@
...
+      super
+    end
+
+    activity_loggable(
+      action: "order_form.signed",
+      record: -> { order_form }
+    )
+
+    def call
+      return result.forbidden_failure! unless License.premium?
```
groyoh: `CreateService` gates on the shared `OrderForms::Premium#order_forms_enabled?`, which checks both `License.premium?` and `feature_flag_enabled?(:order_forms)`. Here only `License.premium?` is checked, so the `:order_forms` flag is enforced only by the controller's `ensure_feature_flag!`. A caller reaching this service from a job or another service would skip the flag. Including `OrderForms::Premium` and gating with `order_forms_enabled?(order_form.organization)` (after the nil check on line 25) would line this up with `CreateService`.

app/services/utils/base64_file.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module Utils
+  class Base64File
+    Decoded = Data.define(:io, :content_type)
+
+    def self.decode(data_uri)
+      metadata, data = data_uri.split(",", 2)
+      content_type = metadata.split(";").first.split(":").second
+
+      Decoded.new(io: StringIO.new(Base64.decode64(data)), content_type:)
```
groyoh, line 14: ⚠️ When `signed_document` is not a valid `data:<type>;base64,<payload>` string (e.g. a value with no comma), `split(",", 2)` leaves `data` as `nil` and `Base64.decode64(nil)` raises. That raise propagates out of `signed_document_blob`, reaches the bare `rescue` in `MarkAsSignedService#call` (line 53), and re-raises, so the request returns 500 instead of a 422. Validating the data-URI shape here (or in the service) and returning a `single_validation_failure!` on `:signed_document` keeps malformed input on the validation path. No spec covers a malformed data URI today.

app/services/order_forms/mark_as_signed_service.rb
```diff
@@ -0,0 +1,111 @@
...
+      order_form.assign_attributes(
+        status: :signed,
+        signed_at: Time.current
+      )
+
+      ActiveRecord::Base.transaction do
+        order_form.signed_document.attach(blob) if blob
+        order_form.save!
+
+        # TODO: Create the Order here using execution_mode/execute_at
```
groyoh, line 45: `execution_mode` and `execute_at` are permitted, validated, and returned as a 200 success, but no order is created yet (left as TODOs here). A client calling with `execution_mode: "execute_in_lago"` gets a success response while nothing executes, so the public contract is misleading until the order creation lands. Is the intent to ship the param surface ahead of the behavior? If not, gating these params behind the unimplemented path (or rejecting them for now) would avoid clients integrating against a silent no-op.

app/controllers/api/v1/order_forms_controller.rb
```diff
@@ -57,6 +74,10 @@ def index_filters
         }
       end
 
+      def mark_as_signed_params
+        params.fetch(:order_form, {}).permit(:signed_document, :execution_mode, :execute_at)
```
groyoh:
```suggestion
        params.expect(order_form: [:signed_document, :execution_mode, :execute_at])
```

app/models/order_form.rb
```diff
@@ -41,6 +49,13 @@ def self.ransackable_attributes(_ = nil)
     %w[number]
   end
 
+  def signed_document_url
+    return unless signed_document.attached?
+    return unless signed_document.blob&.persisted?
```
groyoh: In the other models, we seem to have another pattern 🤔 

```suggestion
    return if signed_document.blank?
```

Any thought ? Maybe we need a helper for this.

app/services/order_forms/mark_as_signed_service.rb
```diff
@@ -0,0 +1,111 @@
...
+      unless decoded.io.size < OrderForm::SIGNED_DOCUMENT_MAX_SIZE
+        result.single_validation_failure!(field: :signed_document, error_code: "file_too_large")
+        return
+      end
+
+      ActiveStorage::Blob.create_and_upload!(
+        io: decoded.io,
+        filename: order_form.number,
+        content_type: decoded.content_type
+      )
```
groyoh: Seems like a different pattern from: https://github.com/getlago/lago-api/blob/f4227dd7ff9e6f2b2a171c79ebbf4b9246f0390f/app/services/billing_entities/create_service.rb#L97-L101

app/services/order_forms/mark_as_signed_service.rb
toommz: The plan was to implement that in a follow-up PR ASAP, as we don't have the Order model yet, as we did for the Quote approval.

app/services/order_forms/mark_as_signed_service.rb
toommz: The problem with attach is that it uploads the file inside the transaction. As we are starting to be wary of long transactions I thought that it might be interesting to handle the upload as a separate step outside the tx that we can rollback if something goes wrong inside the main db transaction.

app/services/order_forms/mark_as_signed_service.rb
toommz: Clearly missed that. Done in 3a9f42b25a937a34ef2976a0222e2ffbee3aeab0

app/models/order_form.rb
toommz: 4e4bf05bf16a2b529cb5633eb1193f7cdde5283f

app/services/utils/base64_file.rb
toommz: a2fc3c2967572ac5ccdbf7fc64ad356647f08be2

app/controllers/api/v1/order_forms_controller.rb
toommz: a1d97734e696ac94791abe4141c6e7a3812264fc

app/services/order_forms/mark_as_signed_service.rb
groyoh: Does it delete the file on rollback, though ?

app/services/order_forms/mark_as_signed_service.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    activity_loggable(
+      action: "order_form.signed",
+      record: -> { order_form }
+    )
+
+    def call
+      return result.not_found_failure!(resource: "order_form") unless order_form
+      return result.forbidden_failure! unless order_forms_enabled?(order_form.organization)
+      return result.not_allowed_failure!(code: "not_signable") unless order_form.generated?
```
groyoh, line 28:
```suggestion
      return result.single_validation_failure!(error_code: "not_signable") unless order_form.generated?
```

app/services/order_forms/mark_as_signed_service.rb
groyoh: Seems like `.attach` uploads after commit so we should be fine to use it.

groyoh (approved):

app/services/order_forms/mark_as_signed_service.rb
toommz: Created batch ticket for this

app/services/order_forms/mark_as_signed_service.rb
toommz: Yeah I was wrong, my memory is starting to trick me 😅

PR #5588, opened 2026-05-27
perf(inovoices): Introduce Preloaders to preload calculations for a bunch of records

vankiru:
## Context

We have a bunch of calculation methods across the models that make requests to the database to calculate specific metrics, such as sum of values or checking whether specific records exist. For example, in `Fee#creditable_amount_cents` we sum all its credit note items' `amount_cents`:

```ruby
def creditable_amount_cents
  remaining_amount = amount_cents - credit_note_items.sum(:amount_cents)
  ...
end
```

This results in N+1 queries when we want to fetch multiple records for invoices, fees, etc.

For most of these cases we can't replace `sum(:field)` with its ruby equivalent `sum(&:field)` (which iterates over already-loaded records instead of querying the db), due to the high probability of having stale records when doing Usage & Fees calculations.

## Current Solutions

This issue was solved for `offset_amount_cents` by introducing a method `Invoice.preload_offset_amounts`. This method preloads `offset_amount_cents` for all records passed to it with a single query to the database and caches the preloaded values for each record in a separate instance variable.

## Preloaders

This PR introduces an extension of the `Invoice.preload_offset_amounts` in the form of Preloaders. A preloader is a class responsible for loading calculation values and caching them at the record level. 

Preloaders caching logic doesn't have any invalidation mechanism – its only purpose is to preload a bunch of calculations for API and GraphQL index responses. 

Preloaders live under the `preloaders/` directory and follow the naming convention `{Model}sPreloader`. For now we only have `InvoicesPreloader` for preloading values on invoices.

### Preloading Logic

To define the preloading logic, a preloader defines a method per calculation value you want to preload. The base preloader provides a `cache` method that sets the preloaded values onto the model.

```ruby
def preload_offset_amount_cents
  offset_amounts = CreditNote
    .where(invoice_id: record_ids)
    .finalized
    .group(:invoice_id)
    .sum(:offset_amount_cents)

  cache(records, :offset_amount_cents, offset_amounts)
end
```

### Preloader Cache

All preloaded values are stored in a `preloader_cache` hash, which can be added to a model by including the `PreloaderCache` concern. You can then use `preloader_cache` to read the preloaded value if it exists, or fall back to loading from the db:

```ruby
def offset_amount_cents
  preloader_cache[:offset_amount_cents] || credit_notes.finalized.sum(:offset_amount_cents)
end
```

### Usage

Usage of a preloader is straightforward – you just pass the records for which you want to preload data:

```ruby
InvoicesPreloader.new(invoices).call
```

This will load all the predefined preloaded values. If you only need to load specific values, you can pass them as a second argument to the preloader:

```ruby
InvoicesPreloader.new(invoices, :offset_amount_cents).call
```

### DIFF
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -3,6 +3,7 @@
 class Fee < ApplicationRecord
   include Currencies
   include Discard::Model
+  include PreloaderCache
 
   self.discard_column = :deleted_at
   self.ignored_columns += %w[duplicated_in_advance]
@@ -216,13 +217,18 @@ class Fee < ApplicationRecord
   end
 
   def creditable_amount_cents
-    remaining_amount = amount_cents - credit_note_items.sum(:amount_cents)
+    remaining_amount = amount_cents - credited_amount_cents
 
     if credit?
       return [remaining_amount, creditable_from_wallet_amount_cents].min
     end
 
     remaining_amount
+  end
+
+  def credited_amount_cents
+    preloader_cache[:credited_amount_cents] ||
+      credit_note_items.sum(:amount_cents)
   end
 
   def creditable_from_wallet_amount_cents
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -227,8 +227,7 @@ class Fee < ApplicationRecord
   end
 
   def credited_amount_cents
-    preloader_cache[:credited_amount_cents] ||
-      credit_note_items.sum(:amount_cents)
+    preloader_cache[:credited_amount_cents] || credit_note_items.sum(:amount_cents)
   end
 
   def creditable_from_wallet_amount_cents
--- a/app/controllers/concerns/invoice_index.rb
+++ b/app/controllers/concerns/invoice_index.rb
@@ -59,7 +59,6 @@ module InvoiceIndex
     )
 
     if result.success?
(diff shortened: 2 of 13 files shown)


app/models/fee.rb
```diff
@@ -225,6 +226,10 @@ def creditable_amount_cents
     remaining_amount
   end
 
+  def credited_amount_cents
+    preloader_cache[:credited_amount_cents] || credit_note_items.sum(:amount_cents)
```
tiagolupepic, line 230: out of curiosity, how this cache is invalidated if the records in credit_note_items changes or if any `amount_cents` is updated?

vankiru: This hash doesn't have any invalidation mechanism – its only purpose is to preload a bunch of calculations for API and GraphQL index responses. We just load a set of records, preload their associations and calculations, build the JSON/GraphQL response, send it to the client, and discard it afterward.

It's not supposed to be used with any usage/fees/invoices/etc calculations, and in those cases we just fall back to the way we currently handle them.

tiagolupepic: understood! thanks for the explanation :) 

Have you evaluated another option? For example we have `scenic` that uses ActiveRecord and provides associations. You can move the AR to custom queries as a view. The benefit, you don't need to maintain extra code and/or have concerns to keep the data memoized, AR will do that for you.

PR #5589, opened 2026-05-28
[ING-21] fix(events): clickhouse removal on metric deletion

vincent-pochet:
## Context

When a billable metric is deleted, `BillableMetrics::DeleteEventsJob` deletes associated events from ClickHouse tables (`events_raw` and `events_enriched`) using synchronous `.delete_all` calls.
For billable metrics with a large number of associated events, the ClickHouse `DELETE FROM` mutation takes longer than the 60-second `Net::HTTP` read timeout, causing a `Net::ReadTimeout` error.

## Description

This PR:
- Extracts the removal logic from the `BillableMetrics::DeleteEventsJobs` to a new `Events::DeleteForMetricService`
- Uses the ClickHouse async delete mutation approach when removing Clickhouse events so that the removals happens in background
- Batches the subscription selection to avoid memory issue for metrics attached to more thant 10_000 subscriptions

### DIFF
--- a/app/services/events/delete_for_metric_service.rb
+++ b/app/services/events/delete_for_metric_service.rb
@@ -1,0 +1,95 @@
+# frozen_string_literal: true
+
+module Events
+  class DeleteForMetricService < BaseService
+    Result = BaseResult
+
+    def initialize(billable_metric:)
+      @billable_metric = billable_metric
+      @deleted_at = billable_metric.deleted_at
+      super
+    end
+
+    def call
+      return result unless billable_metric.discarded?
+
+      clickhouse_enabled = ENV["LAGO_CLICKHOUSE_ENABLED"].present?
+
+      # The subscription is queried in batches to avoid memory pressure
+      # and cross-database subqueries: events is built to live on a dedicated
+      # database (so is clickhouse) while subscriptions/charges live on the primary database.
+      subscriptions_relation.in_batches(of: BATCH_SIZE) do |batch|
+        rows = batch.pluck(:id, :external_id)
+        next if rows.empty?
+
+        subscription_ids, external_subscription_ids = rows.transpose
+
+        delete_postgres_events(subscription_ids, external_subscription_ids)
+        delete_clickhouse_events(external_subscription_ids) if clickhouse_enabled
+      end
+
+      result
+    end
+
+    private
+
+    BATCH_SIZE = 10_000
+    CLICKHOUSE_TABLES = %w[events_raw events_enriched events_enriched_expanded].freeze
+    CLICKHOUSE_MUTATIONS_SYNC = "0" # Async
+
+    attr_reader :billable_metric, :deleted_at
+
+    delegate :code, :organization_id, to: :billable_metric
+
+    def subscriptions_relation
+      # Explicit SQL join bypasses Charge's `default_scope -> { kept }` so
+      # discarded charges are still considered when collecting subscriptions.
+      Subscription
+        .joins(:plan)
+        .joins("INNER JOIN charges ON charges.plan_id = plans.id")
+        .where("charges.billable_metric_id = ?", billable_metric.id)
+        .distinct
+    end
+
+    def delete_postgres_events(subscription_ids, external_subscription_ids)
+      # Delete events having an old-style `subscription_id`
+      Event.where(
+        organization_id: organization_id,
+        code: code,
+        subscription_id: subscription_ids
+      ).where("events.timestamp <= ?", deleted_at)
+        .in_batches.update_all(deleted_at:) # rubocop:disable Rails/SkipsModelValidations
+
+      # Delete events using the new `external_subscription_id`
+      Event.where(
+        organization_id: organization_id,
+        code: code,
+        external_subscription_id: external_subscription_ids
+      ).where("events.timestamp <= ?", deleted_at)
+        .in_batches.update_all(deleted_at:) # rubocop:disable Rails/SkipsModelValidations
+    end
+
+    # Delete clickhouse events using async mutations to avoid blocking or
+    # timeouts when the metric is attached to a large number of events.
+    def delete_clickhouse_events(external_subscription_ids)
+      CLICKHOUSE_TABLES.each do |table|
+        async_delete_clickhouse(table, external_subscription_ids)
+      end
+    end
+
+    def async_delete_clickhouse(table, external_subscription_ids)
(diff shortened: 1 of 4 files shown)


app/services/events/delete_for_metric_service.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+module Events
+  class DeleteForMetricService < BaseService
+    Result = BaseResult
+
+    def initialize(billable_metric:)
+      @billable_metric = billable_metric
+      @deleted_at = billable_metric.deleted_at || Time.current
```
toommz: nit but as we check `billable_metric.discarded?` we shouldn't have to set a default value here, right?

app/services/events/delete_for_metric_service.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      # The subscription is queried in batches to avoid memory pressure
+      # and cross-database subqueries: events is built to live on a dedicated
+      # database (so is clickhouse) while subscriptions/charges live on the primary database.
+      subscriptions_relation.in_batches(of: BATCH_SIZE) do |batch|
+        rows = batch.pluck(:id, :external_id)
+        next if rows.empty?
+
+        subscription_ids = rows.map(&:first)
+        external_subscription_ids = rows.map(&:last)
```
toommz:
```suggestion
        subscription_ids, external_subscription_ids = rows.transpose
```

Could save another 10K loops for full batches

app/services/events/delete_for_metric_service.rb
```diff
@@ -0,0 +1,93 @@
...
+
+        delete_postgres_events(subscription_ids, external_subscription_ids)
+        delete_clickhouse_events(external_subscription_ids) if clickhouse_enabled
```
toommz, line 28: I'm wondering if it could be interesting to do local (in pg) deletes inside a transaction, and if everything is ok then call Clickhouse outside of it?

EDIT: I think that how `subscription_relation` is built makes it easy to re-run the service on failures so it might not be interesting at the product level, if we can do that without a transaction let's keep it that way!

toommz (approved):
Nice, TIL about Clickhouse async mutations

app/services/events/delete_for_metric_service.rb
vincent-pochet: Nice one :+1: I was not aware of this method

app/services/events/delete_for_metric_service.rb
```diff
@@ -0,0 +1,93 @@
...
+        .joins("INNER JOIN charges ON charges.plan_id = plans.id")
+        .where("charges.billable_metric_id = ?", billable_metric.id)
+        .distinct
+    end
+
+    def delete_postgres_events(subscription_ids, external_subscription_ids)
+      # Delete events having an old-style `subscription_id`
+      Event.where(code: code, subscription_id: subscription_ids)
+        .where("events.timestamp <= ?", deleted_at)
+        .update_all(deleted_at:) # rubocop:disable Rails/SkipsModelValidations
```
groyoh: To avoid long-running transactions:

```suggestion
        .in_batches.update_all(deleted_at:) # rubocop:disable Rails/SkipsModelValidations
```

app/services/events/delete_for_metric_service.rb
```diff
@@ -0,0 +1,93 @@
...
+      Subscription
+        .joins(:plan)
+        .joins("INNER JOIN charges ON charges.plan_id = plans.id")
+        .where("charges.billable_metric_id = ?", billable_metric.id)
+        .distinct
+    end
+
+    def delete_postgres_events(subscription_ids, external_subscription_ids)
+      # Delete events having an old-style `subscription_id`
+      Event.where(code: code, subscription_id: subscription_ids)
```
groyoh: I'm not sure we want to do this. This does not seem to rely on any index.

app/services/events/delete_for_metric_service.rb
```diff
@@ -0,0 +1,93 @@
...
+        .where("events.timestamp <= ?", deleted_at)
+        .update_all(deleted_at:) # rubocop:disable Rails/SkipsModelValidations
+
+      # Delete events using the new `external_subscription_id`
+      Event.where(
+        organization_id: organization_id,
+        code: code,
+        external_subscription_id: external_subscription_ids
+      ).where("events.timestamp <= ?", deleted_at)
+        .update_all(deleted_at:) # rubocop:disable Rails/SkipsModelValidations
```
groyoh:
```suggestion
        .in_batches.update_all(deleted_at:) # rubocop:disable Rails/SkipsModelValidations
```

app/services/events/delete_for_metric_service.rb
```diff
@@ -0,0 +1,93 @@
...
+        async_delete_clickhouse(table, external_subscription_ids)
+      end
+    end
+
+    def async_delete_clickhouse(table, external_subscription_ids)
+      sql = ::Clickhouse::BaseRecord.sanitize_sql_array([
+        "ALTER TABLE #{table} DELETE " \
+          "WHERE organization_id = ? AND code = ? " \
+          "AND external_subscription_id IN (?) AND timestamp <= ? " \
+          "SETTINGS mutations_sync = #{CLICKHOUSE_MUTATIONS_SYNC}",
```
groyoh, line 85: I believe ALTER TABLE DELETE is async by default.

vincent-pochet: It is, but this "trick" allows us to test the deletion itself with specs, otherwise we just test "it executed some queries" and hope that it will do something

app/services/events/delete_for_metric_service.rb
vincent-pochet: @groyoh to be fair, it's the actual behavior: https://github.com/getlago/lago-api/blob/main/app/jobs/billable_metrics/delete_events_job.rb#L15

In the old days of the event ingestion, `external_subscription_id` was not mandatory on event, and we were assigning the single active customer's subscription id to the event.
We could probably skip this path now since we are not using these events anywhere  as the aggregation now fully relies on the external_id :thinking:

groyoh: Oh god, you're right. My bad. I'd say, let's double check the data and remove it.

app/services/events/delete_for_metric_service.rb
groyoh: Oh nice trick!

groyoh (approved):

PR #5591, opened 2026-05-28
feat(payment-gated-subs): Defer invoice creation on tax providers to success path

osmarluz:
## Context

For payment-gated subscriptions, `PullTaxesAndApplyService` was hitting the tax provider's `finalized_invoices` endpoint while the invoice was still `:open` waiting for payment, because the draft/finalized branch only checked `invoice.draft?`. That committed a finalized Anrok/Avalara transaction (and, for Avalara, an `IntegrationResource`) before the customer ever paid; if payment then failed, the provider was left out of sync with Lago's now-closed invoice.

## Description

While the gated invoice is still `:open`, route tax computation to `CreateDraftService` so the provider only sees a calculation, not a committed transaction. Commit via `CreateService` once payment succeeds and `Payment::ResolveService#handle_success` finalizes the invoice — enqueued through a new `Integrations::Aggregator::Taxes::Invoices::CreateJob`, gated on the customer having a tax integration.

No cleanup is needed on payment failure: drafts on Anrok/Avalara aren't committed transactions, so closing the invoice is enough.

### DIFF
--- a/app/jobs/integrations/aggregator/taxes/invoices/create_job.rb
+++ b/app/jobs/integrations/aggregator/taxes/invoices/create_job.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Integrations
+  module Aggregator
+    module Taxes
+      module Invoices
+        class CreateJob < ApplicationJob
+          queue_as "providers"
+
+          unique :until_executed, on_conflict: :log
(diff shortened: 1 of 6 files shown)


ancorcruz (approved):
LGTM :shipit:

PR #5596, opened 2026-05-28
fix(integrations): Xero line item parity with Lago invoice (INT-51, INT-52)

mikeh-lago:
## Context

Two Xero invoice line-item bugs live in the same `Xero` payload class. Closes [INT-51](https://linear.app/getlago/issue/INT-51/xero-invoice-line-items-missing-pricing-group-key-names-compared-to) and [INT-52](https://linear.app/getlago/issue/INT-52/fees-with-dollar0-amount-appear-on-lago-invoices-but-not-on-xero).

- **INT-51** — Xero line item descriptions were built from the invoice/charge-filter name only, dropping `fee.grouped_by` values. The synced Xero invoice diverged from the Lago invoice (e.g. `Storage usage • green` on Lago, `Storage usage` on Xero).
- **INT-52** — `$0` fees were missing from Xero invoices. `BasePayload#fees` (shared by all invoice integrations, last changed in #2656) drops zero-amount fees whenever a positive fee exists, since some vendors reject `$0` line items. Xero accepts them, so the filter created a mismatch. Xero already existed when #2656 landed and has inherited this filter — and the resulting `$0` mismatch — ever since.

## Description

- Add `Fee#grouped_by_display`, returning the ` • `-joined `grouped_by` values for charge fees. `FeeDisplayHelper.grouped_by_display` now delegates to it, so the Lago invoice template and the Xero payload share a single implementation rather than the service reaching into a view helper.
- The Xero payload appends `fee.grouped_by_display` to the line item description.
- The Xero payload overrides `#fees` to keep zero-amount fees, bypassing the `BasePayload` filter. Other integrations keep the inherited behavior.

### DIFF
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -159,6 +159,12 @@ class Fee < ApplicationRecord
     charge_filter&.display_name(separator:)
   end
 
+  def grouped_by_display
+    return "" if !charge? || grouped_by.values.compact.blank?
+
+    " • #{grouped_by.values.compact.join(" • ")}"
+  end
+
   def invoice_sorting_clause
     base_clause = "#{invoice_name} #{filter_display_name}".downcase
 
--- a/app/services/integrations/aggregator/invoices/payloads/xero.rb
+++ b/app/services/integrations/aggregator/invoices/payloads/xero.rb
@@ -12,6 +12,7 @@ module Integrations
           def item(fee)
             base_item = super
             base_item["item_code"] = base_item.delete("external_id")
+            base_item["description"] = "#{base_item["description"]}#{FeeDisplayHelper.grouped_by_display(fee)}"
 
             if fee.precise_unit_amount.round(2) != fee.precise_unit_amount
               base_item["units"] = 1
--- a/app/services/integrations/aggregator/invoices/payloads/xero.rb
+++ b/app/services/integrations/aggregator/invoices/payloads/xero.rb
@@ -12,7 +12,7 @@ module Integrations
           def item(fee)
             base_item = super
             base_item["item_code"] = base_item.delete("external_id")
-            base_item["description"] = "#{base_item["description"]}#{FeeDisplayHelper.grouped_by_display(fee)}"
+            base_item["description"] = "#{base_item["description"]}#{fee.grouped_by_display}"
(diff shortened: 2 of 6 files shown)


mikeh-lago:
Thanks for the review. Pushed [9d8686a](https://github.com/getlago/lago-api/commit/9d8686ace) addressing the actionable items; new commit sits on top so the two per-ticket commits stay separately revertible.

**Verified before changing anything:**
- `FeeDisplayHelper.grouped_by_display` is `def self.grouped_by_display(fee)` on a plain `FeeDisplayHelper` class — safe to call from a service.
- `BasePayload#fees` orders by `.order(created_at: :asc)` — Xero override matches, no order divergence.
- `BasePayload#fees` is `private` — visibility matches.
- `tax_adjusted_fee_items`: $0 fees contribute `0` to the tax sum and never satisfy `fee["taxes_amount_cents"] > remaining_taxes_amount_cents.to_i.abs` (0 can't be greater than a positive absolute). No `amount_cents` divisor exists.

**Changes in [9d8686a](https://github.com/getlago/lago-api/commit/9d8686ace):**
- Added a comment on `#fees` pointing at #2359.
- Wrapped all new specs in a parent `context "with a single billable metric charge"` that holds the organization / integration / customer / mapping / invoice setup — removes ~160 lines of duplication. (Left the pre-existing `precise_unit_amount` context untouched — out of scope.)
- INT-52 mix assertion now checks size, ordered descriptions (by `created_at`), and the $0 line item's `units` + `precise_unit_amount`. Replaces the loose `include` matcher.
- Added a multi-key grouped_by context (`{"region" => "eu", "tier" => "gold"}` → `"Storage usage • eu • gold"`) to lock in the format.
- Added a regression guard for the all-$0 invoice case (every fee `amount_cents: 0`).

**Acknowledged, not done in this PR:**
- Charge-filter + grouped_by interaction. Tracked as a separate hidden inconsistency in the PR description — out of scope.
- Anrok / NetSuite / Hubspot. Same `BasePayload#item` source; will file separate tickets once verified per integration.

Waiting on CI green before flipping out of draft.

ancorcruz (approved):
LGTM

PR #5599, opened 2026-05-29
feat(payment-gated-subs): Recredit on invoice cancellation due to payment failure or expiration

osmarluz:
## Context

Payment-gated subscription invoices live in `:open` while waiting for payment. During their creation they consume customer balances — applied coupons, credit-note balance, and prepaid wallet credits — via `Invoices::CalculateFeesService` and `Invoices::CreatePayInAdvanceFixedChargesService`. When the gated invoice later transitions to `:closed` because payment failed (`Subscriptions::ActivationRules::Payment::ResolveService`) or the activation rule timed out (`Subscriptions::ActivationRules::ExpireService`), those consumptions were not reversed: customers permanently lost balances on invoices that were never billed.

## Description

Both close paths now enqueue per-resource recredit jobs after the close transaction commits:
- `AppliedCoupons::RecreditJob`, `CreditNotes::RecreditJob`, and `WalletTransactions::RecreditJob` are thin wrappers around the existing recredit services, one job per `Credit` row or outbound `WalletTransaction`. Per-resource isolation keeps a failure in one re-credit from interfering with the others.
- Each job pre-checks the resource is still eligible (applied_coupon non-nil, credit_note non-voided, wallet active) so transient state changes between enqueue and execute don't trigger needless retries.
- `Credit#active` scope extended to exclude `:closed` invoices in addition to `:voided`, so `applied_coupon.remaining_amount` correctly reflects coupons released by a closed gated invoice (the existing recredit service is a no-op for `:once` coupons unless `Credit#active` lets the balance recompute).

### DIFF
--- a/app/jobs/applied_coupons/recredit_job.rb
+++ b/app/jobs/applied_coupons/recredit_job.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module AppliedCoupons
+  class RecreditJob < ApplicationJob
+    queue_as "default"
+
+    def perform(credit)
+      return if credit.applied_coupon.nil?
+
+      AppliedCoupons::RecreditService.call!(credit:)
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


ancorcruz (approved):
LGTM 🚢

PR #5607, opened 2026-06-01
feat(order-forms): create at quote version approval

toommz:
## Roadmap Task

## Context

An order form only ever exists on an approved quote version (one-to-one). `QuoteVersions::ApproveService` already flipped a version to `approved` but left a `# TODO: OrderForms::CreateService.call` placeholder, so approving a quote version never produced the order form the business expects.

## Description

Generate the order form as part of quote version approval.

### DIFF
--- a/app/services/order_forms/create_service.rb
+++ b/app/services/order_forms/create_service.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module OrderForms
+  class CreateService < BaseService
+    include OrderForms::Premium
+
+    Result = BaseResult[:order_form]
+
+    def initialize(quote_version:)
+      @quote_version = quote_version
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "quote_version") unless quote_version
+      return result.forbidden_failure! unless order_forms_enabled?(quote_version.organization)
+      return result.not_allowed_failure!(code: "quote_version_not_approved") unless quote_version.approved?
+
+      order_form = OrderForm.create!(
+        organization: quote_version.organization,
+        customer: quote_version.quote.customer,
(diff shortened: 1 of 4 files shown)


app/services/order_forms/create_service.rb
```diff
@@ -0,0 +1,36 @@
...
+        organization: quote_version.organization,
+        customer: quote_version.quote.customer,
+        quote_version:,
+        status: :generated
+      )
+
+      result.order_form = order_form
+      result
+    rescue ActiveRecord::RecordInvalid => e
+      result.record_validation_failure!(record: e.record)
```
vincent-pochet, line 29: Since we also have a unique index on the `quote_version`, I would suggest to also rescue `ActiveRecord::RecordInvalid`

```suggestion
    rescue ActiveRecord::RecordInvalid => e
      result.record_validation_failure!(record: e.record)
    rescue ActiveRecord::RecordInvalid
      result.single_validation_failure!(field: :quote_version_id, error_code: "value_already_exist")
```

vincent-pochet (approved):

app/services/order_forms/create_service.rb
toommz: @vincent-pochet Almost sure but you were talking about `ActiveRecord::RecordNotUnique` right?

murparreira (approved):

app/services/order_forms/create_service.rb
vincent-pochet: @toommz Ah of course, sorry for the confusion!

toommz: 1945aada805fa0627164eb8b6532871aba2ae487

PR #5608, opened 2026-06-01
fix(current-usage): Filter out non-zero fees from current usage flow

osmarluz:
## Context

`Customers::RefreshWalletJob` ends up in the dead set for customers with a tax integration whose subscription generates more than 1200 fees in the current period: the current-usage flow sends every fee to the tax provider (Anrok/Avalara), which rejects payloads above 1200 line items. Unlike end-of-month invoices, current usage never skipped fees with no units, amount, or events, so empty fees inflated the count.

Fixes INT-15.

## Description

Current usage now drops fees whose units, amount, and events count are all zero — the same rule already applied when persisting end-of-month invoice fees — using `Fee#non_zero?` in `CustomerUsageService`. This keeps the tax-provider payload under the limit and aligns current usage with the eventual invoice.

Since the filtering now happens at the source, the redundant `non_zero?` calls in `DailyUsages::ComputeService` and `FillHistoryService` are removed.

### ⚠️ Behavior change ⚠️

The filter removes a fee only when units **and** amount **and** events count are all zero. This changes the current **and projected** usage responses across the dashboard GraphQL, customer-portal GraphQL, and the REST `GET /customers/:id/usage`:

- **Zero-activity charge filters disappear.** Both the implicit catch-all filter and any defined filter that got no events in the period (e.g. a `cloud=gcp` filter with no `gcp` events) are no longer returned under `filters` / `groupedUsage`. Until now the default `with_zero_units_filters: true` kept these for transparency.
- **A charge with no usage at all is omitted entirely** from `chargesUsage` — there is no longer a `$0` placeholder line for a charge that hasn't been used yet this period.
- **Anything that recorded events is kept**, even when it resolves to 0 units and 0 amount (`events_count > 0`), so genuine "used but free" usage still shows. Note this is stricter than `with_zero_units_filters`, which only looked at units: a zero-unit fee that still carries amount or events is unaffected; only fully-empty fees are dropped.

This matches end-of-month invoices (which never contain these empty lines), but it is a visible change for any consumer that assumed an exhaustive, stable list of charges/filters in the usage payload — dashboards, the customer portal, and public API integrations.

### DIFF
--- a/app/services/daily_usages/compute_service.rb
+++ b/app/services/daily_usages/compute_service.rb
@@ -19,8 +19,6 @@ module DailyUsages
         return result
       end
 
-      current_usage.fees = current_usage.fees.select(&:non_zero?)
-
       if current_usage.fees.any?
         daily_usage = DailyUsage.new(
           organization: subscription.organization,
--- a/app/services/daily_usages/fill_history_service.rb
+++ b/app/services/daily_usages/fill_history_service.rb
@@ -43,8 +43,6 @@ module DailyUsages
             previous_daily_usage = nil
           end
 
-          usage.fees = usage.fees.select(&:non_zero?)
-
           if usage.fees.any?
             daily_usage = DailyUsage.new(
               organization:,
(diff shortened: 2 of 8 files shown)


osmarluz:
This is replaced by the approach on PR #5609

PR #5609, opened 2026-06-01
fix(tax-providers): Filter out zero amount fees from tax provider calls

osmarluz:
## Context

`Customers::RefreshWalletJob` ends up in the dead set for customers with a tax integration whose subscription generates more than 1200 fees in the current period: when computing current usage, every fee is sent to the tax provider (Anrok/Avalara), which rejects payloads above 1200 line items. Fees with a zero amount carry no tax, yet still count toward that limit.

## Description

When computing current usage with provider taxation, only taxable fees — those with a positive amount — are sent to the `CreateDraftService` request and the matching tax-application loop, keeping the payload under the provider's line-item limit. A `Fee#taxable?` predicate captures this: tax is a function of the amount, so a fee with zero amount owes no tax regardless of its units or events count, and keeps its default zero taxes.

The fees themselves are left untouched everywhere else — the current and projected usage responses, daily usage, lifetime usage, and alerts still include them exactly as before. Only the tax-provider request is trimmed.

Invoice-level tax totals are unchanged: an excluded fee has a zero taxable base, so it would have been taxed at zero anyway, and it contributes nothing to the invoice's tax amount or rate. The one observable difference is per-fee: an excluded zero-amount fee now comes back in the usage response with no applied taxes and a zero tax rate, instead of a non-zero tax rate.

### DIFF
--- a/app/services/invoices/customer_usage_service.rb
+++ b/app/services/invoices/customer_usage_service.rb
@@ -211,6 +211,11 @@ module Invoices
     def compute_amounts_with_provider_taxes
       invoice.fees_amount_cents = invoice.fees.sum(&:amount_cents)
 
+      # NOTE: Set the sub total so Invoices::ApplyProviderTaxesService prorates taxes_rate
+      #       by amount (like persisted invoices) instead of falling back to its zero-amount
+      #       count-based branch, which would dilute the rate with the excluded non-taxable fees.
+      invoice.sub_total_excluding_taxes_amount_cents = invoice.fees_amount_cents
+
       # NOTE: Only fees with a positive amount can incur tax, so non-taxable fees are
       #       excluded from the provider request. This also keeps the payload under the
       #       provider line-item limit (Anrok/Avalara reject payloads above 1200 items).
--- a/app/services/invoices/customer_usage_service.rb
+++ b/app/services/invoices/customer_usage_service.rb
@@ -211,13 +211,18 @@ module Invoices
     def compute_amounts_with_provider_taxes
       invoice.fees_amount_cents = invoice.fees.sum(&:amount_cents)
 
-      taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateDraftService.call(invoice:, fees: invoice.fees)
+      # NOTE: Zero fees are excluded from the tax provider request to stay under its
+      #       line-item limit (Anrok/Avalara reject payloads above 1200 items). They owe
+      #       no tax, so they keep their default zero taxes and stay in the usage response.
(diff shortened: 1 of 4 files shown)


brunomiguelpinto (approved):

app/services/invoices/customer_usage_service.rb
```diff
@@ -211,13 +211,24 @@ def compute_amounts_without_tax
     def compute_amounts_with_provider_taxes
       invoice.fees_amount_cents = invoice.fees.sum(&:amount_cents)
 
-      taxes_result = Integrations::Aggregator::Taxes::Invoices::CreateDraftService.call(invoice:, fees: invoice.fees)
+      # NOTE: Set the sub total so Invoices::ApplyProviderTaxesService prorates taxes_rate
+      #       by amount (like persisted invoices) instead of falling back to its zero-amount
+      #       count-based branch, which would dilute the rate with the excluded non-taxable fees.
+      invoice.sub_total_excluding_taxes_amount_cents = invoice.fees_amount_cents
```
groyoh, line 217: This feels a bit out of place. Is there no risk of side-effect with this change ?

groyoh: Oh I just saw that the `invoice` is not actually stored in this case. Nevermind.

groyoh (approved):

app/services/invoices/customer_usage_service.rb
osmarluz: @groyoh yes, that's right. The problem was that inside `Invoices::ApplyProviderTaxesService` we do the following:

```ruby
def pro_rated_taxes_rate(tax)
  tax_rate = tax.rate.is_a?(String) ? tax.rate.to_f * 100 : tax.rate

  fees_rate = if invoice.sub_total_excluding_taxes_amount_cents.positive?
    fees_amount_cents(tax).fdiv(invoice.sub_total_excluding_taxes_amount_cents)
  else
    # NOTE: when invoice have a 0 amount. The prorata is on the number of fees
    key = calculate_key(tax)
    indexed_fees[key].count.fdiv(invoice.fees.count)
  end

  fees_rate * tax_rate
end
```

Since `sub_total_excluding_taxes_amount_cents` wasn't set, `pro_rated_taxes_rate` fell into the else branch and prorated by fee count (`indexed_fees[key].count / invoice.fees.count`) instead of by amount. Only the taxed (non-zero) fees are in `indexed_fees[key]`, but `invoice.fees.count` includes the zero fees too. That ended up reducing the invoice's taxes_rate from the real value

PR #5612, opened 2026-06-01
chore(order-forms): remove marked_as_signed_by_user

toommz:
We don't need the `order_forms.marked_as_signed_by_user_id` column actually. This PR removes it and references to it.


db/migrate/20260601174030_remove_marked_as_signed_by_user_from_order_forms.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class RemoveMarkedAsSignedByUserFromOrderForms < ActiveRecord::Migration[8.0]
+  def change
+    safety_assured do
```
toommz, line 5: `safety_assured` as it is not used yet.

vincent-pochet (approved):

PR #5613, opened 2026-06-01
feat(payment-gated-subs): adapt refund model for expired subscriptions

ancorcruz:
## Context

Payment-gated subscription activation rules needs refund late payments that settle after a subscription activation timeout. The DB changes are now on main, and this PR wires the model and existing credit-note refund writers to the new fields.

## Description

- Add Refund model support for polymorphic refundable records and explicit refund reasons.
- Add factory coverage for credit-note refunds and subscription-activation-expired refunds.
- Update Stripe, Adyen, and GoCardless credit-note refund creation to set refundable and reason fields.

### DIFF
--- a/spec/models/refund_spec.rb
+++ b/spec/models/refund_spec.rb
@@ -5,7 +5,49 @@ require "rails_helper"
 RSpec.describe Refund do
   subject(:refund) { build(:refund) }
 
+  describe "enums" do
+    it do
+      expect(subject).to define_enum_for(:reason)
+        .backed_by_column_of_type(:string)
+        .with_values(described_class::REASONS)
+        .validating(allowing_nil: true)
+    end
+  end
+
   describe "associations" do
-    it { is_expected.to belong_to(:organization) }
+    it do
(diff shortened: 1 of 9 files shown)


spec/models/refund_spec.rb
```diff
@@ -5,7 +5,49 @@
...
+      expect(refund).to be_valid
+    end
+
+    it "allows activation-expired refunds without a credit note" do
+      activation_refund = build(:subscription_activation_expired_refund)
+      expect(activation_refund).to be_valid
+      expect(activation_refund.credit_note).to be_nil
+      expect(activation_refund.refundable).to be_closed
+      expect(activation_refund.reason).to eq("subscription_activation_expired")
+    end
```
osmarluz, line 51: @ancorcruz very minor, but since we have the `|| credit_note.present?` check we could also add something like:


```suggestion
    end
    
    it "is valid with an unpersisted credit note" do
      expect(build(:refund)).to be_valid
    end
```

osmarluz (approved):
LGTM! Just left a small comment on one test

PR #5615, opened 2026-06-02
feat(premium_integrations): add revenue_recognition

ancorcruz:
## Context

The revenue recognition feature needs to be gated as a premium integration so it can be enabled per organization ahead of the staging release.

 ## Description

Add `revenue_recognition` to `Organization::PREMIUM_INTEGRATIONS`, which automatically wires up the `with_revenue_recognition_support` scope and `revenue_recognition_enabled?` predicate, and exposes the value through the GraphQL premium integration type enum.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -153,6 +153,7 @@ class Organization < ApplicationRecord
     security_logs
     granular_lifetime_usage
     order_forms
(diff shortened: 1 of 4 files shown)


osmarluz (approved):
🚀

PR #5618, opened 2026-06-02
feat(wallets): Permit targeted wallet refresh on the RefreshWalletJob

jdenquin:
# Context

`Customers::RefreshWalletJob` always triggered a full wallet refresh for a customer, fanning out usage recomputation across all of the customer's active wallets. In cases where only specific wallets need refreshing (e.g. after a targeted credit grant), recomputing every wallet is unnecessary and costly.

`Customers::RefreshWalletsService` already supports a `target_wallet_ids` argument to scope the refresh. This change exposes that capability through the job.

# Description

  - Add an optional `wallet_ids` argument to `RefreshWalletJob#perform`
  - Keep the argument optional with a `nil` default so existing callers continue to trigger a full refresh unchanged
  - When `wallet_ids` is `nil`, the service refreshes all active wallets as before
  - Update the job spec
  - Add a spec context covering the targeted-refresh path (wallet ids provided)

# Notes

  - Backward compatible: no behavior change for existing single-argument callers
  - Service-level coverage for `target_wallet_ids` (both `nil` and set) already exists in `spec/services/customers/refresh_wallets_service_spec.rb`

### DIFF
--- a/app/jobs/customers/refresh_wallet_job.rb
+++ b/app/jobs/customers/refresh_wallet_job.rb
@@ -17,11 +17,11 @@ module Customers
     retry_on BaseService::TooManyProviderRequestsFailure, wait: :polynomially_longer, attempts: 25
     retry_on(*Integrations::Aggregator::BaseService.retryable_errors, wait: :polynomially_longer, attempts: 6)
 
-    def perform(customer)
+    def perform(customer, wallet_ids = nil)
       return unless customer.awaiting_wallet_refresh?
       return if customer.error_details.tax_error.exists?
 
-      Customers::RefreshWalletsService.call!(customer:)
+      Customers::RefreshWalletsService.call!(customer:, target_wallet_ids: wallet_ids)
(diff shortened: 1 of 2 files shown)


app/jobs/customers/refresh_wallet_job.rb
```diff
@@ -17,11 +17,11 @@ class RefreshWalletJob < ApplicationJob
     retry_on BaseService::TooManyProviderRequestsFailure, wait: :polynomially_longer, attempts: 25
     retry_on(*Integrations::Aggregator::BaseService.retryable_errors, wait: :polynomially_longer, attempts: 6)
 
-    def perform(customer)
+    def perform(customer, wallet_ids = nil)
```
groyoh:
```suggestion
    def perform(customer, wallet_ids: nil)
```

groyoh (approved):

vincent-pochet (approved):

PR #5620, opened 2026-06-02
feat(order-forms): add expire lifecycle

toommz:
## Roadmap Task

## Context

Order forms have an optional `expires_at`. We need expiry to be enforced automatically, without a user action or any API call, so that stale order forms transition to a terminal state on their own.

## Description

Adds time-based expiration of order forms, triggered entirely by an hourly cron (no REST or GraphQL endpoint is exposed).

The CI typing issue is handled in https://github.com/getlago/lago-front/pull/3616


app/jobs/order_forms/expire_order_form_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module OrderForms
+  class ExpireOrderFormJob < ApplicationJob
+    queue_as :default
```
groyoh, line 5: Please add uniqueness check.

app/models/order_form.rb
```diff
@@ -32,6 +32,8 @@ class OrderForm < ApplicationRecord
 
   validates :void_reason, presence: true, if: :voided?
 
+  scope :expirable, -> { generated.where.not(expires_at: nil).where("expires_at < ?", Time.current) }
```
groyoh: You might want to include customer/billing entity timezone as we do in other time-based checks.

app/services/order_forms/expire_service.rb
```diff
@@ -0,0 +1,43 @@
...
+      @order_form = order_form
+
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "order_form") unless order_form
+      return result.forbidden_failure! unless order_forms_enabled?(order_form.organization)
+
+      return result.not_allowed_failure!(code: "order_form_is_voided") if order_form.voided?
```
groyoh:
```suggestion
      return result.forbidden_failure!(code: "order_form_is_voided") if order_form.voided?
```

app/services/order_forms/expire_service.rb
```diff
@@ -0,0 +1,43 @@
...
+      return result.not_allowed_failure!(code: "order_form_is_signed") if order_form.signed?
```
groyoh:
```suggestion
      return result.forbidden_failure!(code: "order_form_is_signed") if order_form.signed?
```

app/services/order_forms/expire_service.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      if order_form.expired?
+        result.order_form = order_form
+        return result
+      end
+
+      order_form.assign_attributes(status: :expired, voided_at: Time.current, void_reason: :expired)
+
+      ActiveRecord::Base.transaction do
+        order_form.save!
```
groyoh, line 30: I believe we need to consider locking on the quote level to avoid data inconsistencies. Might not be require within this PR but something to keep in mind.

app/jobs/clock/expire_order_forms_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Clock
+  class ExpireOrderFormsJob < ClockJob
```
groyoh, line 4: Please add uniqueness check.

app/jobs/order_forms/expire_order_form_job.rb
toommz: d9badf93e2e05204e2a861b5cb92177e66b5d0db

app/services/order_forms/expire_service.rb
toommz: f7380fb64529475779c3dbc44df33ce3b6fe4af2

app/services/order_forms/expire_service.rb
toommz: f7380fb64529475779c3dbc44df33ce3b6fe4af2

app/services/order_forms/expire_service.rb
toommz: Let's keep that in mind yes, at the moment also I would say that as the service is only called from the job that has uniqueness check, it might not be a problem until this "one call place rule" holds true

app/jobs/clock/expire_order_forms_job.rb
toommz: d9badf93e2e05204e2a861b5cb92177e66b5d0db

app/models/order_form.rb
toommz: What about doing that at write time? When populating the `expires_at` column

groyoh: Could make sense but I don't think this is usual in the codebase.

app/services/order_forms/expire_service.rb
groyoh: I was thinking more about expiring and another action at the same time (signing, voiding, etc.).

groyoh (approved):

PR #5621, opened 2026-06-02
feat(order-forms): add void lifecycle REST endpoint

toommz:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/create-quotes-order-forms

## Context

Part of the order forms feature (not yet released). An order form needs a way to be cancelled manually, in addition to expiring automatically.

## Description

Adds a new endpoint to void an order form: `POST /api/v1/order_forms/:id/void`

- Only order forms in `generated` status (not yet `signed`) can be `voided`.
- A `voided` order form can no longer be `signed` by the customer.
- The parent quote version is automatically `voided` as well, so it cannot be reused. A new version must be created to issue a new order form.

Requires a premium license with the `order_forms` feature flag.

### DIFF
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -54,6 +54,17 @@ module Api
         render_order_form(order_form)
       end
 
+      def void
+        order_form = current_organization.order_forms.find_by(id: params[:id])
+        result = OrderForms::VoidService.call(order_form:)
+
+        if result.success?
+          render_order_form(result.order_form)
+        else
+          render_error_response(result)
+        end
+      end
+
       private
 
       def ensure_feature_flag!
--- a/app/controllers/api/v1/order_forms_controller.rb
+++ b/app/controllers/api/v1/order_forms_controller.rb
@@ -68,7 +68,7 @@ module Api
       private
 
       def ensure_feature_flag!
-        forbidden_error(code: "feature_not_available") unless current_organization.feature_flag_enabled?(:order_forms)
+        forbidden_error(code: "feature_unavailable") unless current_organization.feature_flag_enabled?(:order_forms)
       end
 
       def index_filters
--- a/app/services/order_forms/void_service.rb
+++ b/app/services/order_forms/void_service.rb
@@ -31,8 +31,7 @@ module OrderForms
       ActiveRecord::Base.transaction do
         order_form.save!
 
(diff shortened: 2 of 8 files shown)


spec/requests/api/v1/order_forms_controller_spec.rb
```diff
@@ -86,4 +86,59 @@
...
+      expect(response).to have_http_status(:ok)
+      expect(json[:order_form][:lago_id]).to eq(order_form.id)
+      expect(json[:order_form][:status]).to eq("voided")
+    end
+
+    context "without a premium license" do
+      it "returns forbidden" do
+        subject
+
+        expect(response).to have_http_status(:forbidden)
```
groyoh, line 298: The org here has the `order_forms` flag, so this 403 comes from the premium guard in `VoidService`. The assertion only checks the status though, so it would also pass if the 403 came from `ensure_feature_flag!`. Asserting `expect(json[:code]).to eq("feature_unavailable")` pins it to the premium path and mirrors the `feature_not_available` assertion in the flag-disabled context below.

app/services/order_forms/void_service.rb
```diff
@@ -0,0 +1,46 @@
...
+        status: :voided,
+        voided_at: Time.current,
+        void_reason: :manual
+      )
+
+      ActiveRecord::Base.transaction do
+        order_form.save!
+
+        # TODO: Cascade the void to the parent quote version
+        # (QuoteVersions::VoidService.call!(quote_version: order_form.quote_version, ...))
```
groyoh: Do you plan to do it in this PR ?

toommz: Yep, after expire is merged so I can rebase on the changes made to the `QuoteVersions::VoidService` in this other PR

spec/requests/api/v1/order_forms_controller_spec.rb
toommz: Done

app/controllers/api/v1/order_forms_controller.rb
groyoh: I figured this should be:

```suggestion
        forbidden_error(code: "feature_unavailable") unless current_organization.feature_flag_enabled?(:order_forms)
```

No point in having two different errors `feature_not_available` and `feature_unavailable`.

groyoh (approved):

PR #5623, opened 2026-06-02
[ING-253] fix(wallets): allow billing entity move on update

aquinofb:
## The problem

Moving a wallet between billing entities through the API used to silently fail: `PATCH /api/v1/wallets/:id` with `billing_entity_code` would return 200 OK but the wallet stayed on its old entity. Operators got no error to react to.

There was a second, deeper issue. Even after fixing the silent drop, every transaction on the wallet would instantly re-route to the new entity, including purchases that were already pending. If a customer started a top-up while the wallet was on `acme_us`, and the operator moved the wallet to `acme_eu` before the purchase settled, the invoice would come out under `acme_eu` -- not where the purchase began.

## The fix

The move now works: `PATCH` actually changes the wallet's entity, the response reflects the new value on the next `GET`, and listing wallets filtered by entity returns the moved one.

Each wallet transaction also remembers the entity it was created under. A pending purchase keeps its original entity through a wallet move. A new purchase tracks whatever entity the wallet is currently on. A refund inherits from the transaction it is refunding, so refunds always land on the same entity as the original purchase.

Transactions that existed before this change keep their previous behavior: their settle path falls through to the customer's entity, just like before. Nothing about how legacy data is processed changes.

## How to try locally

With `multi_entity_billing` enabled:

~~~ruby
acme_us = client.billing_entities.create(code: "acme_us", name: "ACME US")
acme_eu = client.billing_entities.create(code: "acme_eu", name: "ACME EU")

wallet = client.wallets.create(
  external_customer_id: customer.external_id,
  rate_amount: "1", currency: "USD", name: "demo",
  billing_entity_code: "acme_us"
)

# Pending paid_credit while wallet is on acme_us
client.wallet_transactions.create(wallet_id: wallet.lago_id, paid_credits: "5")

# Move the wallet
client.wallets.update({billing_entity_code: "acme_eu"}, wallet.lago_id)
client.wallets.find(wallet.lago_id).billing_entity_code  # => "acme_eu"

# Settling the pending transaction issues the invoice under acme_us.
# Any new paid_credit after this point tracks acme_eu.
~~~

### DIFF
--- a/app/controllers/concerns/wallet_actions.rb
+++ b/app/controllers/concerns/wallet_actions.rb
@@ -70,9 +70,10 @@ module WalletActions
       render(
         json: ::CollectionSerializer.new(
           result.wallets.includes(
-            :customer,
+            :billing_entity,
             :metadata,
             :billable_metrics,
+            {customer: :billing_entity},
             {applied_invoice_custom_sections: :invoice_custom_section},
             {recurring_transaction_rules: {applied_invoice_custom_sections: :invoice_custom_section}}
           ),
@@ -162,6 +163,8 @@ module WalletActions
       :invoice_requires_successful_payment,
       :paid_top_up_min_amount_cents,
       :paid_top_up_max_amount_cents,
+      :billing_entity_code,
+      :billing_entity_id,
       metadata: {},
       recurring_transaction_rules: [
         :lago_id,
--- a/app/controllers/concerns/wallet_actions.rb
+++ b/app/controllers/concerns/wallet_actions.rb
@@ -164,7 +164,6 @@ module WalletActions
       :paid_top_up_min_amount_cents,
       :paid_top_up_max_amount_cents,
       :billing_entity_code,
-      :billing_entity_id,
       metadata: {},
       recurring_transaction_rules: [
         :lago_id,
--- a/app/services/invoices/paid_credit_service.rb
+++ b/app/services/invoices/paid_credit_service.rb
@@ -67,7 +67,7 @@ module Invoices
         invoice_type: :credit,
         currency:,
         datetime: Time.zone.at(timestamp),
-        billing_entity: wallet_transaction.wallet.billing_entity || customer.billing_entity
+        billing_entity: wallet_transaction.billing_entity || wallet_transaction.wallet.billing_entity || customer.billing_entity
       )
       invoice_result.raise_if_error!
 
--- a/app/services/wallet_transactions/create_service.rb
+++ b/app/services/wallet_transactions/create_service.rb
@@ -28,6 +28,7 @@ module WalletTransactions
           :voided_invoice_id
         ),
         organization_id: wallet.organization_id,
+        billing_entity_id: billing_entity_id_for_snapshot,
         amount:,
         credit_amount:,
         metadata: transaction_params[:metadata] || [],
@@ -57,6 +58,10 @@ module WalletTransactions
       return nil unless transaction_params[:transaction_status]&.to_sym == :granted
 
       wallet_credit.amount_cents
+    end
+
+    def billing_entity_id_for_snapshot
+      transaction_params[:billing_entity_id] || wallet.billing_entity&.id
     end
   end
 end
--- a/db/migrate/20260602175438_add_billing_entity_to_wallet_transactions.rb
+++ b/db/migrate/20260602175438_add_billing_entity_to_wallet_transactions.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddBillingEntityToWalletTransactions < ActiveRecord::Migration[8.0]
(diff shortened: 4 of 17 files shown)


db/migrate/20260602175438_add_billing_entity_to_wallet_transactions.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddBillingEntityToWalletTransactions < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :wallet_transactions, :billing_entity, type: :uuid, null: true,
```
aquinofb, line 7: This is now required because once the wallet flips the billing entity, then we need to retain the original.

db/migrate/20260602175439_backfill_wallet_transactions_billing_entity.rb
```diff
@@ -0,0 +1,25 @@
...
+            INNER JOIN wallets ON wallets.customer_id = customers.id
+            WHERE wallets.id = wallet_transactions.wallet_id
+          )
+        )
+      SQL
+    end
+  end
+
+  def down
+    # irreversible
```
mariohd: IMO this is not 100% irreversible.

Since this should run only once, in the ideal scenario, removing all BE from wallet_transactions would be the `down` of this migration.

aquinofb: That's true. I assumed we'd always rollback both migrations together, so we'd remove the column anyway.

Great call!

app/services/invoices/paid_credit_service.rb
```diff
@@ -66,7 +66,8 @@ def create_generating_invoice
         customer:,
         invoice_type: :credit,
         currency:,
-        datetime: Time.zone.at(timestamp)
+        datetime: Time.zone.at(timestamp),
+        billing_entity: wallet_transaction.billing_entity
```
lovrocolic: can we please let this PR handle this: https://github.com/getlago/lago-api/pull/5528

Or at least we can first merge 5528 and then adapt this line in your PR. wdyt?

app/services/wallet_transactions/create_service.rb
```diff
@@ -58,5 +59,9 @@ def initial_remaining_amount_cents
 
       wallet_credit.amount_cents
     end
+
+    def billing_entity_id_for_snapshot
+      transaction_params[:billing_entity_id] || wallet.billing_entity&.id
```
lovrocolic, line 64: Don't we need permit new param in the controller?

app/services/invoices/paid_credit_service.rb
aquinofb: yeah, makes sense. Let me know when you merge it and then I'll rebase 🙌

app/controllers/concerns/wallet_actions.rb
```diff
@@ -162,6 +163,8 @@ def update_params
       :invoice_requires_successful_payment,
       :paid_top_up_min_amount_cents,
       :paid_top_up_max_amount_cents,
+      :billing_entity_code,
```
groyoh, line 166: I don't think that's used.

app/services/wallet_transactions/create_service.rb
aquinofb: Hey Lovro, great call. This is actually set by the [void service](https://github.com/getlago/lago-api/pull/5623/changes#diff-786f18b52b028d25817ec8b3d7a1d230b546fe5a0d3a799b8f3d008ac96e482cR36), that's why we prioritise it, but then fallback to the wallet's.

app/controllers/concerns/wallet_actions.rb
aquinofb: Great call! The `_entity_code` is used yes, and covered by this [spec](https://github.com/getlago/lago-api/pull/5623/changes#diff-f8d7fb892057d9c64d80e31e0a2cc8e6061ce40bea13e2cfa5ff9c78ec51c720R1151-R1181). On the other hand, the `_entity_id` param is a leftover that can be removed.

I left both because I noticed we're always falling back to one or another, e.g., [Wallets::UpdateService#billing_entity](https://github.com/getlago/lago-api/blob/main/app/services/wallets/update_service.rb#L201-L210).

wdyt?

groyoh: We should consistency with other endpoint and use only one I believe. The service usually handle both because API (code) vs GraphQL (id).

lovrocolic (approved):

app/controllers/concerns/wallet_actions.rb
aquinofb: True. I just removed the `*_id`.

PR #5624, opened 2026-06-03
feat(payment-gated-subs): add Stripe refund and PSP dispatcher

ancorcruz:
## Context

When a payment-gated subscription activation times out, the pending PSP payment is canceled on a best-effort basis. That cancel can fail (or be unsupported by the provider), in which case the payment may still settle successfully after the subscription has already been canceled. Without an automated refund path, the customer ends up charged for a subscription that never activated, requiring manual intervention.

## Description

- Add `PaymentProviders::RefundPaymentService`, a thin dispatcher that routes to the provider-specific refund service based on the payment's provider type. Mirrors the shape of the existing `PaymentProviders::CancelPaymentService`.
- Add `PaymentProviders::Stripe::Payments::RefundService` issuing full-amount refunds against the original payment intent, with Stripe-side idempotency keyed by payment id and Lago metadata attached for support traceability.
- Add `PaymentProviders::RefundPaymentJob` async wrapper.
- Idempotency is layered: the dispatcher short-circuits when a `Refund` row already exists for the payment (one refund per payment is the contract for full refunds), and the PSP idempotency key is the second line of defense.
- Stripe rejections (`InvalidRequestError`) persist a sentinel `Refund` row in `failed` state so the dispatcher does not silently reattempt. Recovery from a hard-failed refund is intentionally a manual operation.
- Adyen, GoCardless, Cashfree, Flutterwave, and Moneyhash fall through to a log line; their refund integrations land in follow-up PRs.

### DIFF
--- a/app/jobs/payment_providers/refund_payment_job.rb
+++ b/app/jobs/payment_providers/refund_payment_job.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module PaymentProviders
+  class RefundPaymentJob < ApplicationJob
+    queue_as do
+      if ActiveModel::Type::Boolean.new.cast(ENV["SIDEKIQ_PAYMENTS"])
+        :payments
+      else
+        :providers
+      end
+    end
+
+    def perform(payment, reason: nil)
+      PaymentProviders::RefundPaymentService.call!(payment:, reason:)
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


ancorcruz:
Closing this PR because we have decided not to move forward with the payment refund on late payment success for gated subscription at this time.

PR #5625, opened 2026-06-03
perf(subscriptions): Fix N+1 for Subscriptions in GraphQL

vankiru:
## Context
`SubscriptionsResolver` loads a few associations for subscriptions, which ends up causing N+1 queries.

## Description

This PR preloads all necessary associations in `SubscriptionsResolver` to avoid the N+1 queries.

### DIFF
--- a/app/graphql/resolvers/subscriptions_resolver.rb
+++ b/app/graphql/resolvers/subscriptions_resolver.rb
@@ -31,7 +31,7 @@ module Resolvers
       )
 
       result.subscriptions.preload(
-        :next_subscriptions,
+        {next_subscriptions: :plan},
         {customer: :billing_entity}
       )
     end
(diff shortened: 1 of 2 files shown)


tiagolupepic (approved):
:shipit:

app/graphql/resolvers/subscriptions_resolver.rb
```diff
@@ -30,7 +30,10 @@ def resolve(page: nil, limit: nil, plan_code: nil, status: nil, external_custome
         search_term:
       )
 
-      result.subscriptions
+      result.subscriptions.preload(
+        :next_subscriptions,
+        {customer: :billing_entity}
+      )
```
rsempe, line 36:
```suggestion
      result.subscriptions.preload(
        {next_subscriptions: :plan},
        {customer: :billing_entity}
      )
```

as `next_plan` is used [here](https://github.com/getlago/lago-api/blob/main/app/graphql/types/subscriptions/object.rb#L64)

rsempe (approved):

PR #5626, opened 2026-06-03
perf(customers): Fix N+1 for Customers in GraphQL

vankiru:
## Context
`CustomersResolver` loads a bunch of associations for customers, which ends up causing N+1 queries.

## Description

This PR preloads all necessary associations in `CustomersResolver` to avoid the N+1 queries.

### DIFF
--- a/app/graphql/resolvers/customers_resolver.rb
+++ b/app/graphql/resolvers/customers_resolver.rb
@@ -52,7 +52,7 @@ module Resolvers
         :hubspot_customer,
         :netsuite_customer,
         :salesforce_customer,
-        :xero_customer,
+        :xero_customer
       )
     end
   end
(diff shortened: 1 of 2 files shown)


tiagolupepic (approved):
:shipit:

app/graphql/resolvers/customers_resolver.rb
```diff
@@ -44,7 +44,16 @@ def resolve(search_term: nil, page: nil, limit: nil, metadata: nil, **filters)
         filters:
       )
 
-      result.customers
+      result.customers.preload(
```
rsempe, line 47: should we have to preload `provider_customer` as well?

https://github.com/getlago/lago-api/blob/main/app/graphql/types/customers/object.rb#L48

rsempe (approved):

app/graphql/resolvers/customers_resolver.rb
vankiru: For now, we don't show `provider_customer` on the Customers page, and the frontend isn't querying for it either, so that's why I didn't include it.

PR #5627, opened 2026-06-03
fix(daily-usages): compute subs with time-dependent charges

brunomiguelpinto:
## Context

`DailyUsages::ComputeAllService` only scheduled a `ComputeJob` for subscriptions whose `last_received_event_on` was recent. But some subscriptions accrue usage between billing boundaries *without* receiving new events:

- **prorated charges** and **weighted_sum aggregations** — their value changes daily (time-weighted), even with no events;
- **recurring billable metrics** — their value is carried over and stays constant within a period, changing only at the period rollover.

These were silently skipped, so their daily usage went stale until the next event arrived — and then the whole gap was rolled into a single day.

## Description

Schedule a `ComputeJob` for three disjoint sets of subscriptions, per organization, restricted to customers entering a new day in their timezone:

1. **Recent events** — `last_received_event_on >= yesterday` (unchanged).
2. **Daily time-dependent** — plans with a prorated charge or a weighted_sum aggregation; recomputed every day.
3. **Recurring rollover** — plans with a recurring billable metric; recomputed **once per period** (on the run following the period rollover, since the value is constant in between). The rollover calendar is computed by the new `Subscriptions::BillingDateQuery`, which shares its calendar/anniversary/timezone logic with `Subscriptions::OrganizationBillingService`.

The plan-based sets are selected via a `plan_id` **subquery (semi-join)** instead of a correlated `EXISTS`, which was catastrophic on plan-per-subscription organizations.

Each selection query runs **once per organization** (the ids are `pluck`ed and unioned into a `Set`), already-computed subscriptions are removed in a single dedup query, and the survivors are enqueued by primary key in bounded batches. This avoids re-executing the heavy selection query per `find_each` batch.

### Performance

All numbers measured in production on the worst-case day (start of month, when every calendar-monthly period rolls over), across **all organizations**.

| Approach | Time |
|---|---|
| **Final** — 3 semi-join legs + `pluck`/`Set` + dedup subtract | **~52s** |

> [!NOTE]
> This increases the daily `ComputeJob` volume, since time-dependent and recurring subscriptions are now scheduled even without events. Watch the `analytics` queue latency after deploy; the existing 30-min scheduling jitter spreads the load. Subscriptions opted out via `skip_daily_usage` are still excluded, and the `ComputeJob` unique lock / `ComputeService#existing_daily_usage` keep computation idempotent.

### DIFF
--- a/app/services/daily_usages/compute_all_service.rb
+++ b/app/services/daily_usages/compute_all_service.rb
@@ -69,7 +69,7 @@ module DailyUsages
               .where.not(id: subscription_ids_with_daily_usage)
               .where(skip_daily_usage: false)
               .where(
-                "last_received_event_on >= :yesterday ",
+                "last_received_event_on >= :yesterday OR #{TIME_DEPENDENT_USAGE_SQL}",
                 yesterday: timestamp.to_date - 1.day
               )
               .find_each do |subscription|
--- a/app/services/daily_usages/compute_all_service.rb
+++ b/app/services/daily_usages/compute_all_service.rb
@@ -67,8 +67,11 @@ module DailyUsages
             Subscription.where(customer_id: customer_ids)
               .active
               .where.not(id: subscription_ids_with_daily_usage)
-              .where("last_received_event_on >= ?", timestamp.to_date - 1.day)
               .where(skip_daily_usage: false)
+              .where(
+                "last_received_event_on >= :yesterday ",
+                yesterday: timestamp.to_date - 1.day
+              )
               .find_each do |subscription|
                 yield subscription
             end
@@ -76,5 +79,25 @@ module DailyUsages
         end
       end
     end
+
+    # Subscriptions whose usage can change between billing boundaries even without receiving new
+    # events: prorated charges, recurring billable metrics, and weighted_sum aggregations (which
+    # are time-weighted). These must be recomputed every day regardless of `last_received_event_on`,
+    # otherwise their daily usage would go stale until the next event arrives.
+    TIME_DEPENDENT_USAGE_SQL = <<~SQL.squish.freeze
+      EXISTS (
+        SELECT 1
+        FROM charges
+        JOIN billable_metrics ON billable_metrics.id = charges.billable_metric_id
+        WHERE charges.plan_id = subscriptions.plan_id
+          AND charges.deleted_at IS NULL
+          AND billable_metrics.deleted_at IS NULL
+          AND (
+            charges.prorated = TRUE
+            OR billable_metrics.recurring = TRUE
+            OR billable_metrics.aggregation_type = #{BillableMetric.aggregation_types[:weighted_sum_agg]}
+          )
+      )
+    SQL
   end
 end
--- a/app/services/daily_usages/compute_all_service.rb
+++ b/app/services/daily_usages/compute_all_service.rb
@@ -69,7 +69,7 @@ module DailyUsages
               .where.not(id: subscription_ids_with_daily_usage)
               .where(skip_daily_usage: false)
               .where(
-                "last_received_event_on >= :yesterday OR #{TIME_DEPENDENT_USAGE_SQL}",
+                "last_received_event_on >= :yesterday OR #{PLAN_HAS_TIME_DEPENDENT_CHARGE_SQL}",
                 yesterday: timestamp.to_date - 1.day
               )
               .find_each do |subscription|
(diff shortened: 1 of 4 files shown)


spec/services/daily_usages/compute_all_service_spec.rb
```diff
@@ -166,6 +166,58 @@
...
 
+    context "when the subscription has time-dependent usage but no recent events" do
+      let(:plan) { create(:plan, organization:) }
+      let(:billable_metric) { create(:sum_billable_metric, organization:, recurring: true) }
+      let(:subscriptions) do
+        create_list(:subscription, 5, customer:, plan:, last_received_event_on: timestamp.to_date - 5.days)
+      end
+
+      context "with a prorated charge" do
+        before { create(:standard_charge, plan:, billable_metric:, prorated: true) }
```
rsempe, line 177: can we use a billable metric not `recurring` here?

app/services/daily_usages/compute_all_service.rb
```diff
@@ -67,14 +67,37 @@ def each_subscription(&block)
...
           end
         end
       end
     end
+
+    # Subscriptions whose usage can change between billing boundaries even without receiving new
+    # events: prorated charges, recurring billable metrics, and weighted_sum aggregations (which
+    # are time-weighted). These must be recomputed every day regardless of `last_received_event_on`,
+    # otherwise their daily usage would go stale until the next event arrives.
+    TIME_DEPENDENT_USAGE_SQL = <<~SQL.squish.freeze
```
rsempe: what about `PLAN_HAS_TIME_DEPENDENT_CHARGE_SQL` instead for the naming?

rsempe (approved):
👍

app/services/daily_usages/compute_all_service.rb
brunomiguelpinto: sounds good

annvelents (approved):
Noice!

vincent-pochet (approved):
🙌

app/services/daily_usages/compute_all_service.rb
```diff
@@ -67,14 +67,41 @@ def each_subscription(&block)
...
+        JOIN billable_metrics ON billable_metrics.id = charges.billable_metric_id
+        WHERE charges.plan_id = subscriptions.plan_id
+          AND charges.deleted_at IS NULL
+          AND billable_metrics.deleted_at IS NULL
+          AND (
+            charges.prorated = TRUE
+            OR billable_metrics.recurring = TRUE
+            OR billable_metrics.aggregation_type = #{BillableMetric.aggregation_types[:weighted_sum_agg]}
+          )
+      )
```
groyoh: @brunomiguelpinto did you test the performance of the query with this new change? We already spent a lot of time to improve this query performance so I want to ensure we don't introduce any performance regression here.

brunomiguelpinto: yes did a bunch of tests in prod and it takes around 52s to go through all the subscriptions

brunomiguelpinto: when some queries are cached we can go even lower DailyUsages::ComputeAllService.new(timestamp: Time.zone.parse("2026-06-02 00:05:00")).call
=== grand=57943 elapsed=26.2s ===
=> #<BaseService::LegacyResult>

app/services/subscriptions/billing_date_query.rb
```diff
@@ -0,0 +1,189 @@
...
+  # on `timestamp`'s date, in the customer's applicable timezone.
+  #
+  # The calendar logic (calendar vs anniversary, every interval, end-of-month and leap-year edge
+  # cases, timezone) is intentionally identical to the periodic billing selection in
+  # `Subscriptions::OrganizationBillingService#billable_subscriptions`. It is expressed here as a
+  # single OR-ed WHERE so it can be composed onto any subscriptions scope.
+  #
+  # NOTE: the given scope must allow joining `:plan` and `customer: :billing_entity` (needed for the
+  #       timezone-aware date comparisons).
+  class BillingDateQuery < BaseService
```
vincent-pochet, line 14: Good for now, but as a next improvement we might try to make this service generic to use it in both DailyJob & Billing

vincent-pochet (approved):

app/services/subscriptions/billing_date_query.rb
brunomiguelpinto: yes that would be good

annvelents (approved):
LGTM!

PR #5628, opened 2026-06-03
ING 255/254 filter by single BE

mariohd:
## Context

`billing_entity_code` (scalar) was silently dropped on `GET /wallets`, `GET /subscriptions`, and `GET /payment_requests`. The endpoints returned `200 OK` with the full unfiltered list and no error. Only the bracketed `billing_entity_codes[]=…` form was honoured, even though the scalar form is the shape in the dive-in QA plan (S69) and the OpenAPI spec — so any workflow scoping by entity through that parameter was getting wrong data.

**Closes ING-255 and ING-254** (sister ticket flagging the same gap on `/payment_requests`).

## Fix

The bug lived only in the param-parsing layer — controllers/concerns inspected `:billing_entity_codes` but not `:billing_entity_code`. The query layer (`WalletsQuery` / `SubscriptionsQuery` / `PaymentRequestsQuery`) already accepts a list of IDs correctly. So the fix is the same one-line pattern in three places: permit both params, `Array.wrap` each, union, dedupe.

After this PR, all of these work the same way and produce the same result:

```
?billing_entity_code=acme_eu
?billing_entity_codes[]=acme_eu
?billing_entity_code=acme_eu&billing_entity_codes[]=acme_us   (union)
```

Unknown codes still return `404 billing_entity_not_found` in either form.

## Commits

| Commit | Subject | Where |
|---|---|---|
| `8b332cef4` | `feat(wallets): accept scalar entity code` | both wallets controllers |
| `b4179cff7` | `feat(subscriptions): accept scalar code` | `SubscriptionIndex` concern |
| `eb0b2ce30` | `feat(payment-requests): scalar code` | `PaymentRequestIndex` concern |

Each commit includes the matching spec additions (single code, unknown-code 404, scalar + bracketed union). The two concerns are shared between the top-level and customer-scoped controllers, so all six endpoints get the fix.

## How to test

```
  spec/requests/api/v1/wallets_controller_spec.rb \
  spec/requests/api/v1/customers/wallets_controller_spec.rb \
  spec/requests/api/v1/subscriptions_controller_spec.rb \
  spec/requests/api/v1/customers/subscriptions_controller_spec.rb \
  spec/requests/api/v1/payment_requests_controller_spec.rb \
  spec/requests/api/v1/customers/payment_requests_controller_spec.rb
```

### DIFF
--- a/app/controllers/api/v1/customers/wallets_controller.rb
+++ b/app/controllers/api/v1/customers/wallets_controller.rb
@@ -29,11 +29,15 @@ module Api
         end
 
         def index
-          permitted_params = params.permit(:currency, billing_entity_codes: [])
+          permitted_params = params.permit(:currency, :billing_entity_code, billing_entity_codes: [])
+          billing_entity_codes = (
+            Array.wrap(permitted_params[:billing_entity_codes]) +
+            Array.wrap(permitted_params[:billing_entity_code])
+          ).compact_blank.uniq.presence
           wallet_index(
             external_customer_id: customer.external_id,
             currency: permitted_params[:currency],
-            billing_entity_codes: permitted_params[:billing_entity_codes]
+            billing_entity_codes: billing_entity_codes
           )
         end
 
--- a/app/controllers/api/v1/wallets_controller.rb
+++ b/app/controllers/api/v1/wallets_controller.rb
@@ -28,10 +28,18 @@ module Api
       end
 
       def index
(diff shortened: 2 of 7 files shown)


mariohd:
Was decided the we are keeping only the `billing_entity_codes[]`. If we the user want to filter by a single BE, he can send just one e.g `billing_entity_codes[]=ACME`.

PR #5630, opened 2026-06-03
fix(payment_requests): Do not flip succeeded invoices back to failed

rsempe:
## Context

When a Stripe PaymentRequest payable PaymentIntent auto-cancels after ~24h, the payment_intent.canceled webhook is routed through PaymentIntentPaymentFailedService. The PR-payable guard only checks payment_request.payment_succeeded?, so if the linked invoice was moved to succeeded by another path (e.g. manual API retry on the invoice), the cascade flipped it back to failed and the clock job re-marked it overdue.

## Description

Add a per-invoice guard in update_invoices_payment_status across all PR payment provider services (stripe, adyen, cashfree, flutterwave, gocardless, moneyhash, create_service): skip the update when the invoice is already payment_succeeded and the incoming status is not :succeeded.

### DIFF
--- a/app/services/payment_requests/payments/adyen_service.rb
+++ b/app/services/payment_requests/payments/adyen_service.rb
@@ -121,6 +121,8 @@ module PaymentRequests
 
       def update_invoices_payment_status(payment_status:, deliver_webhook: true)
         payable.invoices.each do |invoice|
+          next if invoice.payment_succeeded? && !payment_status_succeeded?(payment_status)
+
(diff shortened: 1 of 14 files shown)


groyoh (approved):
I believe we could improve test coverage here. Only one service out of 7 is tested for this use case 😅

brunomiguelpinto (approved):

groyoh (approved):

PR #5631, opened 2026-06-03
fix(wallets): Fix wallet increase refresh to be async

jdenquin:
## Context

We were refreshing the wallet when doing an increase, to ensure returned `balance` and `ongoing_balance` were up to date if there was any usage coming. The wallet refresh sync slow down the wallet transaction creation process and response AND the returned wallet was not reloaded, so the wallet refresh was not reflected in the service response.

## Changes

- Refresh the wallet asynchronously after wallet increase, keeping the wallet targeting (added in https://github.com/getlago/lago-api/pull/5618)
- Update the refresh wallet job to run if a list of wallet is provided, even if the customer is not flagged as to be refreshed. We did not want to update the customer at this level to avoid any lock contention.

### DIFF
--- a/app/services/wallets/balance/increase_service.rb
+++ b/app/services/wallets/balance/increase_service.rb
@@ -33,9 +33,8 @@ module Wallets
 
         # we only need to update all wallets when there is usage applied. In case we're increasing balance of only one wallet,
         # only this wallet will be affected and needs recalculation
-        Customers::RefreshWalletsService.call(customer: wallet.customer, target_wallet_ids: [wallet.id])
-
         after_commit do
+          Customers::RefreshWalletJob.perform_later(wallet.customer, wallet_ids: [wallet.id])
           SendWebhookJob.perform_later("wallet.updated", wallet)
           UsageMonitoring::ProcessWalletAlertsJob.perform_later(wallet)
         end
--- a/app/services/wallets/balance/increase_service.rb
+++ b/app/services/wallets/balance/increase_service.rb
@@ -33,11 +33,9 @@ module Wallets
 
         # we only need to update all wallets when there is usage applied. In case we're increasing balance of only one wallet,
         # only this wallet will be affected and needs recalculation
-        after_commit do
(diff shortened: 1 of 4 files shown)


app/services/wallets/balance/increase_service.rb
```diff
@@ -33,9 +33,8 @@ def call
 
         # we only need to update all wallets when there is usage applied. In case we're increasing balance of only one wallet,
         # only this wallet will be affected and needs recalculation
-        Customers::RefreshWalletsService.call(customer: wallet.customer, target_wallet_ids: [wallet.id])
-
         after_commit do
+          Customers::RefreshWalletJob.perform_later(wallet.customer, wallet_ids: [wallet.id])
           SendWebhookJob.perform_later("wallet.updated", wallet)
           UsageMonitoring::ProcessWalletAlertsJob.perform_later(wallet)
         end
```
groyoh:
```suggestion
        Customers::RefreshWalletJob.perform_after_commit(wallet.customer, wallet_ids: [wallet.id])
        SendWebhookJob.perform_after_commit("wallet.updated", wallet)
        UsageMonitoring::ProcessWalletAlertsJob.perform_after_commit(wallet)
```

spec/services/wallets/balance/increase_service_spec.rb
```diff
@@ -38,11 +38,9 @@ def call_and_reload_wallet
...
     end
 
-    it "refreshes wallet ongoing balance" do
-      call_and_reload_wallet
-
-      expect(wallet.ongoing_balance_cents).to eq(1450)
-      expect(wallet.credits_ongoing_balance).to eq(14.5)
+    it "enqueues a RefreshWalletJob targeting the wallet" do
+      expect { create_service.call }
+        .to have_enqueued_job(Customers::RefreshWalletJob).with(wallet.customer, wallet_ids: [wallet.id])
```
groyoh:
```suggestion
        .to have_enqueued_job_after_commit(Customers::RefreshWalletJob).with(wallet.customer, wallet_ids: [wallet.id])
```

groyoh (approved):

PR #5636, opened 2026-06-04
[ING-292] perf(cache): skip zero-fee caching for all orgs

groyoh:
## Context

The `non_persistable_charge_cache_optimization` feature flag was introduced in #5096 to exclude zero unit/amount/event fees from the charge usage cache, shrinking the cached payload while keeping the API response shape unchanged through rehydration. The flag has been progressively enabled per organization and is now considered stable (supporting work in #5264 and #5404).

## Description

The per-organization feature flag check is removed from `Fees::ChargeService#skip_caching_of_non_persistable_fee?`, so the optimization now applies to every organization. The flag definition and the per-organization flag values are cleaned up separately, which keeps a rollback path: reverting this change restores the per-organization behavior.

Specs asserting the legacy flag-off behavior were removed or updated to describe the new default: zero fees are no longer written to the cache, the persistable-fee check loads adjusted fees once, and the aggregation factory is invoked a second time for hydration. The flag-gated contexts that asserted the optimized behavior now run against a plain organization without the flag.

The `[true, false]` feature flag loops were also removed from the current usage scenario specs, since both iterations now exercise the same code path. Each example runs once, halving that suite; no assertion was changed.


vincent-pochet (approved):

vincent-pochet:
:raised_hands:

toommz (approved):

PR #5640, opened 2026-06-04
feat(fc-units-override): add FixedCharge#units_for resolver

ancorcruz:
## Context

Read paths that consume a fixed charge's units in a subscription context (events, aggregation, fee creation, invoice preview, serializers, GraphQL types) need to honour the per-subscription override when one exists. A single accessor on FixedCharge lets callers drop in a subscription-aware lookup without rewriting every read site.

The accessor lands ahead of the read-path swaps so subsequent changes are mechanical one-liners. No callers yet; the resolver is a no-op while the override table is empty.

## Description

- New `FixedCharge#units_for(subscription)` returns the matching `Subscription::FixedChargeUnitsOverride#units` when a kept override row exists for the (subscription, fixed_charge) pair, and falls back to the fixed charge's own `units` otherwise.
- Lookup goes through the existing `subscription_units_overrides` association, so soft-discarded rows are excluded by the default scope on the override model.
- Spec covers the four shapes: no override, override present, discarded override falls back, and override for a different subscription falls back.

### DIFF
--- a/app/models/fixed_charge.rb
+++ b/app/models/fixed_charge.rb
@@ -51,6 +51,10 @@ class FixedCharge < ApplicationRecord
       units == fixed_charge.units
   end
 
+  def effective_units_for(subscription)
+    subscription_units_overrides.find_by(subscription:)&.units || units
+  end
+
   # When upgrading a subscription with fixed_charges paid_in_advance,
   # this exact charge might have already been paid at the beginning of billing period.
   # in case of prorating, we need to deduct the prorated amount (remaining of the billing_period)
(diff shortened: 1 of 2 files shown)


app/models/fixed_charge.rb
```diff
@@ -51,6 +51,10 @@ def equal_properties?(fixed_charge)
       units == fixed_charge.units
   end
 
+  def units_for(subscription)
```
tiagolupepic: I'd improve a bit the naming, make it clear the fallback to `self.units` when the relation with the subscription doesn't exist. wdyt?

rsempe: `effective_units_for`?

rsempe (approved):

PR #5641, opened 2026-06-04
feat(fc-units-override): resolve sub-aware units in preview aggregation

ancorcruz:
## Context

Invoice preview is the one fee-pipeline read site that derives a fixed charge's units directly from the parent record. The other paths already route units through either a FixedChargeEvent (sub-aware once events are emitted with the resolver) or an AdjustedFee row (already customer-specific), so they pick up overrides automatically. The preview path runs against an in-memory or freshly persisted subscription where no event has been emitted yet, so it has to ask the resolver itself.

## Description

`FixedChargeEvents::Aggregations::PreviewAggregationService` now reads units through `fixed_charge.units_for(subscription)` at every site that previously hit `fixed_charge.units`. The lookup is wrapped in a single `effective_units` helper so both the prorated and non-prorated branches share one read path.

Specs cover three new shapes on top of the existing preview cases: override returns its own value, prorated path scales the override units, and a discarded override falls back to the plan-level units.

### DIFF
--- a/app/services/fixed_charge_events/aggregations/preview_aggregation_service.rb
+++ b/app/services/fixed_charge_events/aggregations/preview_aggregation_service.rb
@@ -7,15 +7,19 @@ module FixedChargeEvents
         units = if fixed_charge.prorated?
           calculate_prorated_units
         else
-          fixed_charge.units
+          effective_units
         end
 
         result.aggregation = units
-        result.full_units_number = fixed_charge.units
+        result.full_units_number = effective_units
         result
       end
 
       private
+
+      def effective_units
+        @effective_units ||= fixed_charge.effective_units_for(subscription)
+      end
 
       def calculate_prorated_units
         from_date = from_datetime.to_date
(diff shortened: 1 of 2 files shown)


app/services/fixed_charge_events/aggregations/preview_aggregation_service.rb
```diff
@@ -7,27 +7,31 @@ def call
...
         result.aggregation = units
-        result.full_units_number = fixed_charge.units
+        result.full_units_number = effective_units
         result
       end
 
       private
 
+      def effective_units
+        fixed_charge.units_for(subscription)
```
tiagolupepic: you can memoize this to avoid hit 4 times the DB

rsempe (approved):

app/services/fixed_charge_events/aggregations/preview_aggregation_service.rb
ancorcruz: thanks, I'm adding memoization, anyway AR cache will skip the DB round-trip as query is identical. That memoization is great.

PR #5642, opened 2026-06-04
feat(fc-units-override): emit fixed-charge events with sub-aware units

ancorcruz:
## Context

FixedChargeEvent rows drive downstream billing. When a subscription carries a per-subscription units override, the parent fixed charge is still reused for emission and its plain `units` value no longer reflects what the subscription should be billed for. The event row itself has to capture the effective units for the subscription so the Simple and Prorated aggregators (which read straight from FixedChargeEvent#units) pick up the override automatically.

## Description

`FixedChargeEvents::CreateService` now builds the event with `fixed_charge.units_for(subscription)` instead of `fixed_charge.units`. The resolver's existing behaviour with a nil subscription (returning the plan-level value) keeps the pre-existing nil-subscription validation path intact.

Spec adds one context covering the override-present path; existing cases (default units, missing-association validation failure) continue to pass unchanged.

### DIFF
--- a/app/services/fixed_charge_events/create_service.rb
+++ b/app/services/fixed_charge_events/create_service.rb
@@ -17,7 +17,7 @@ module FixedChargeEvents
         organization:,
         subscription:,
         fixed_charge:,
-        units: fixed_charge.units,
+        units: fixed_charge.effective_units_for(subscription),
         timestamp:
       )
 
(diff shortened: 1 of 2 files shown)


tiagolupepic (approved):
:shipit:

rsempe (approved):

PR #5643, opened 2026-06-04
feat(fc-units-override): surface sub-aware units in REST and GraphQL

ancorcruz:
## Context

API responses are how operators and the front-end observe a fixed charge's units. With override rows being written and events emitting the resolved value, the read surfaces still expose the plan-level units — the override is invisible to anyone hitting the API. The REST serializer and the GraphQL Subscription type need to resolve units against the subscription whenever one is in scope, while plan-scope reads keep returning the plan-level value.

## Description

- `FixedCharge#units_for` short-circuits when subscription is nil and returns plan-level units without hitting the overrides table. This removes a wasted query from every call site that may not have a subscription on hand (controller index/show, the REST serializer, legacy invocations).
- `V1::FixedChargeSerializer` reads units through `model.units_for(options[:subscription])`. The subscription option is optional, so existing plan-scope serializations stay unchanged.
- `Api::V1::Subscriptions::FixedChargesController` forwards the subscription to the serializer in index, show, and update so every subscription-scoped REST response reflects the override when one exists.
- New `Subscription::FixedChargePresenter` (a SimpleDelegator) wraps a FixedCharge with subscription context and overrides `#units` to call the resolver. Used only by the GraphQL Subscription type, where the FixedCharge type's `units` field is the entry point and we can't push subscription context any deeper. Plan-scope GraphQL resolvers keep returning raw FixedCharge records.
- `Types::Subscriptions::Object#fixed_charges` wraps each association member with the presenter on the way out.

Specs cover the resolver's nil short-circuit, the serializer with and without an override, the REST index/show paths, the presenter's delegation behaviour, and a GraphQL query that fetches fixed_charges through the Subscription type.

### DIFF
--- a/app/controllers/api/v1/subscriptions/fixed_charges_controller.rb
+++ b/app/controllers/api/v1/subscriptions/fixed_charges_controller.rb
@@ -13,11 +13,6 @@ module Api
             .page(params[:page])
             .per(params[:per_page] || PER_PAGE)
 
-          effective_units_by_id = ::Subscription::FixedChargeUnitsOverride.units_map_for(
-            subscription:,
-            fixed_charges:
-          )
-
           render(
             json: ::CollectionSerializer.new(
               fixed_charges,
@@ -25,8 +20,7 @@ module Api
               collection_name: "fixed_charges",
               meta: pagination_metadata(fixed_charges),
               includes: %i[taxes],
-              subscription:,
-              effective_units_by_id:
+              effective_units_by_id: effective_units_map_for(fixed_charges)
             )
           )
         end
@@ -37,7 +31,7 @@ module Api
               fixed_charge,
               root_name: "fixed_charge",
               includes: %i[taxes],
-              subscription:
+              effective_units_by_id: effective_units_map_for([fixed_charge])
             )
           )
         end
@@ -55,7 +49,7 @@ module Api
                 result.fixed_charge,
                 root_name: "fixed_charge",
                 includes: %i[taxes],
-                subscription:
+                effective_units_by_id: effective_units_map_for([result.fixed_charge])
               )
             )
           else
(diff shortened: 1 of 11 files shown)


app/graphql/types/subscriptions/object.rb
```diff
@@ -113,6 +113,7 @@ def fixed_charges
         object.plan.fixed_charges
           .includes(:add_on, :taxes)
           .order(created_at: :asc)
+          .map { |fc| ::Subscription::FixedChargePresenter.new(fc, object) }
```
tiagolupepic: any chance to eager load the `subscription_units_overrides` relation? 
the method `effective_units_for` will do a N+1. 

One detail, not sure the `includes` or any other eager load works with `find_by`. Need to test it.

tiagolupepic (approved):
just left a comment regarding the N+1, the rest looks good!

:shipit:

app/graphql/types/subscriptions/object.rb
ancorcruz: fixed, it now preloads the overrides in one single query.

app/controllers/api/v1/subscriptions/fixed_charges_controller.rb
```diff
@@ -13,13 +13,20 @@ def index
...
           render(
             json: ::CollectionSerializer.new(
               fixed_charges,
               ::V1::FixedChargeSerializer,
               collection_name: "fixed_charges",
               meta: pagination_metadata(fixed_charges),
-              includes: %i[taxes]
+              includes: %i[taxes],
+              subscription:,
+              effective_units_by_id:
```
tiagolupepic: do we need both attributes? `subscription` and `effective_units_by_id`? I'm wondering if we can send only `effective_units_by_id`. 

It's the same for the `show` and `update` action. 

Also, I'm lacking a bit the context, does this cover the webhook as well? because it's using the same serializer there

ancorcruz: no override-related change needed for webhooks. Plan webhooks correctly return plan-level units. Subscription webhooks don't currently surface fixed_charges at all.

tiagolupepic (approved):
thanks for addressing the comments, looks great! 

:shipit:

PR #5644, opened 2026-06-04
feat(fc-units-override): preserve sub-level units on plan updates

ancorcruz:
## Context

When the plan-level units of a fixed charge change, every active subscription on the plan currently receives a FixedChargeEvent (and a pay-in-advance billing job when requested). Subscriptions that have already decoupled their units via a per-subscription override row are not on the plan default anymore — overwriting them from the plan-level update would silently revert per-customer commitments.

Amount and properties changes don't need filtering: they live on the plan-level fixed_charge record and don't trigger event emission, so they continue to broadcast to every subscription on the next billing cycle (multiplied by each subscription's effective units).

## Description

- New `Subscription.without_fixed_charge_units_override_for(fixed_charge)` scope filters out subscriptions that hold a kept override row for the given fixed charge. Soft-discarded rows stop excluding so a removed override behaves like no override at all.
- `FixedCharges::EmitEventsService` applies the scope in its iterate- all-plan-subscriptions branch. The single-subscription branch is untouched — callers passing a specific subscription stay in control.
- `FixedCharges::UpdateService` applies the same scope when dispatching `Invoices::CreatePayInAdvanceFixedChargesJob` so the mid-period delta invoice skips overridden subscriptions as well.

Specs cover the scope, the EmitEventsService skip on iteration, and the UpdateService dispatch skip.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -80,7 +80,10 @@ class Subscription < ApplicationRecord
   scope :starting_in_the_future, -> { pending.where(previous_subscription: nil) }
   scope :expirable, -> { incomplete.joins(:activation_rules).merge(Subscription::ActivationRule.expirable) }
   scope :without_fixed_charge_units_override_for, ->(fixed_charge) {
-    where.not(id: Subscription::FixedChargeUnitsOverride.where(fixed_charge:).select(:subscription_id))
+    overrides = Subscription::FixedChargeUnitsOverride
+      .where(fixed_charge:)
+      .where("subscription_fixed_charge_units_overrides.subscription_id = subscriptions.id")
+    where.not(overrides.arel.exists)
   }
 
   # NOTE: SQL query to get subscription_at into customer timezone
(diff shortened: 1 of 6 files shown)


rsempe:
@ancorcruz can you rebase?

ancorcruz:
@rsempe it was in progress 😄 , all the branches are now rebased and ready.

app/models/subscription.rb
```diff
@@ -79,6 +79,9 @@ class Subscription < ApplicationRecord
 
   scope :starting_in_the_future, -> { pending.where(previous_subscription: nil) }
   scope :expirable, -> { incomplete.joins(:activation_rules).merge(Subscription::ActivationRule.expirable) }
+  scope :without_fixed_charge_units_override_for, ->(fixed_charge) {
+    where.not(id: Subscription::FixedChargeUnitsOverride.where(fixed_charge:).select(:subscription_id))
```
tiagolupepic: there is a index `index_sub_fc_units_overrides_on_sub_id_and_fc_id` which contains the `subscription_id` and it has low cardinality. Passing the `subscription_id` will perform better

spec/services/fixed_charges/emit_events_service_spec.rb
```diff
@@ -238,5 +238,21 @@
         end
       end
     end
+
+    context "when one of the plan subscriptions has a per-subscription units override" do
```
tiagolupepic: there is a scenario covering the `incomplete` subscription. I think it worth to also cover the added scope `.without_fixed_charge_units_override_for` lines 81 and 103.

tiagolupepic (approved):
left a comment to improve a bit the specs.

:shipit:

PR #5649, opened 2026-06-04
invoice-preview: fix collapsed billing period on downgrade

mikeh-lago:
## What changed and why

When a downgrade is scheduled, `POST /invoices/preview` returned the pending (next) subscription with both `current_billing_period_started_at` and `current_billing_period_ending_at` set to the same value: the old subscription's end-of-day termination timestamp (`T23:59:59Z`). The fees on the same invoice already had the correct `from_date` and `to_date`, so the problem was only in how the subscription was serialized.

There were two causes, both in serialization. No fee or billing math changed.

1. `Invoices::Preview::FindSubscriptionsService#adjusted_subscription` set the next subscription's `started_at` to the rotation timestamp `end_of_period + 1.day`, which lands at the end of the day (`T23:59:59Z`). A subscription should start at the beginning of its first day, so it now uses `beginning_of_day` in the customer timezone (`T00:00:00Z`). This also fixes the wrong `started_at` value.

2. `V1::SubscriptionSerializer` and the GraphQL `Types::Subscriptions::Object` resolvers built `Subscriptions::DatesService` from `Time.current`. For a subscription that has not started yet, `Time.current` is before `started_at`, so `charges_from_datetime` and `charges_to_datetime` both fall back to `started_at` and the period collapses. They now use `Subscription#billing_reference_time` (`max(Time.current, started_at)`), so a subscription starting in the future reports its real first period. Nothing changes for subscriptions that have already started. This is also why the fees were already correct: they use the preview's `billing_time`, not `Time.current`.

The same `Time.current` issue affected the GraphQL `current_billing_period_*` resolvers (which the frontend reads) and `period_end_date` (it returned a period before a pending subscription's start). All three now go through `Subscription#billing_reference_time`. These are resolver changes only, with no GraphQL schema change, so there is no codegen.

### Result

```
started_at:                         2026-07-03T00:00:00Z   (was 2026-07-03T23:59:59Z)
current_billing_period_started_at:  2026-07-03T00:00:00Z   (was 2026-07-03T23:59:59Z)
current_billing_period_ending_at:   2026-08-02T23:59:59Z   (was 2026-07-03T23:59:59Z)
```

## Tests

- `spec/scenarios/invoices/preview_spec.rb`: end-to-end test that creates a subscription, schedules a downgrade, previews the invoice, and checks the pending plan's billing period matches its fee boundaries on the same invoice.
- `spec/requests/api/v1/invoices_controller_spec.rb`: request spec covering the preview pipeline for a scheduled downgrade.
- `spec/graphql/types/subscriptions/object_spec.rb`: resolver specs for `current_billing_period_*` and `period_end_date` when the subscription starts in the future.
- `spec/serializers/v1/subscription_serializer_spec.rb`: serializer example for the future-start case.
- `spec/models/subscription_spec.rb`: `#billing_reference_time` for the already-started, future-start, and nil `started_at` cases.
- `spec/services/invoices/preview/find_subscriptions_service_spec.rb`: updated the `started_at` assertion to `beginning_of_day`.

```bash
bundle exec rspec spec/scenarios/invoices/preview_spec.rb \
                  spec/requests/api/v1/invoices_controller_spec.rb \
                  spec/graphql/types/subscriptions/object_spec.rb \
                  spec/serializers/v1/subscription_serializer_spec.rb \
                  spec/models/subscription_spec.rb \
                  spec/services/invoices/preview/find_subscriptions_service_spec.rb
```

## Out of scope

The old subscription's `terminated_at` is left as is (`end_of_period + 1.day`). It feeds the preview `billing_time` and fee boundaries, which are already correct.

## Performance

No impact on hot paths. The changes only touch serialization at preview and read time, with no new queries, loops, or allocations on the event or billing-run path.

### DIFF
--- a/app/graphql/types/subscriptions/object.rb
+++ b/app/graphql/types/subscriptions/object.rb
@@ -85,7 +85,7 @@ module Types
       end
 
       def period_end_date
-        ::Subscriptions::DatesService.new_instance(object, Time.current)
+        ::Subscriptions::DatesService.new_instance(object, billing_reference_time)
           .next_end_of_period
       end
 
@@ -116,7 +116,18 @@ module Types
       end
 
       def dates_service
-        @dates_service ||= ::Subscriptions::DatesService.new_instance(object, Time.current, current_usage: true)
+        @dates_service ||= ::Subscriptions::DatesService.new_instance(object, billing_reference_time, current_usage: true)
+      end
+
+      # NOTE: For a subscription that has not started yet (e.g. a scheduled downgrade), `Time.current`
+      #       precedes `started_at`, so DatesService computes boundaries for the period containing "now"
+      #       — before the subscription exists — and collapses current_billing_period_* (and skews
+      #       period_end_date) onto `started_at`. Anchoring on the later of the two reports the real
+      #       first period. No-op for already-started subscriptions (started_at is in the past).
+      #       `compact` guards against a nil started_at (pending subscription with no start yet); it is
+      #       load-bearing — `max` raises on a nil element, so do not drop it.
+      def billing_reference_time
+        [Time.current, object.started_at].compact.max
       end
     end
   end
--- a/app/graphql/types/subscriptions/object.rb
+++ b/app/graphql/types/subscriptions/object.rb
@@ -116,8 +116,6 @@ module Types
       end
 
       def dates_service
-        # NOTE: anchored on `billing_reference_time` (not `Time.current`) so a not-yet-started
-        #       subscription reports its real first billing period instead of collapsing the bounds.
         @dates_service ||= ::Subscriptions::DatesService.new_instance(object, object.billing_reference_time, current_usage: true)
       end
     end
--- a/app/graphql/types/subscriptions/object.rb
+++ b/app/graphql/types/subscriptions/object.rb
@@ -85,7 +85,7 @@ module Types
       end
 
       def period_end_date
-        ::Subscriptions::DatesService.new_instance(object, billing_reference_time)
+        ::Subscriptions::DatesService.new_instance(object, object.billing_reference_time)
           .next_end_of_period
       end
 
@@ -116,18 +116,9 @@ module Types
       end
 
       def dates_service
-        @dates_service ||= ::Subscriptions::DatesService.new_instance(object, billing_reference_time, current_usage: true)
-      end
-
-      # NOTE: For a subscription that has not started yet (e.g. a scheduled downgrade), `Time.current`
-      #       precedes `started_at`, so DatesService computes boundaries for the period containing "now"
-      #       — before the subscription exists — and collapses current_billing_period_* (and skews
-      #       period_end_date) onto `started_at`. Anchoring on the later of the two reports the real
-      #       first period. No-op for already-started subscriptions (started_at is in the past).
-      #       `compact` guards against a nil started_at (pending subscription with no start yet); it is
-      #       load-bearing — `max` raises on a nil element, so do not drop it.
-      def billing_reference_time
-        [Time.current, object.started_at].compact.max
+        # NOTE: anchored on `billing_reference_time` (not `Time.current`) so a not-yet-started
+        #       subscription reports its real first billing period instead of collapsing the bounds.
+        @dates_service ||= ::Subscriptions::DatesService.new_instance(object, object.billing_reference_time, current_usage: true)
       end
     end
   end
--- a/app/services/invoices/preview/find_subscriptions_service.rb
+++ b/app/services/invoices/preview/find_subscriptions_service.rb
@@ -65,11 +65,8 @@ module Invoices
           .end_of_period + 1.day
       end
 
-      # NOTE: `rotation_date` is the old subscription's end-of-period timestamp shifted by a day,
-      #       so it lands on the end of the day (`T23:59:59`). The downgraded subscription actually
-      #       starts at the *beginning* of that day, so we normalize to the start of the day in the
-      #       customer timezone. Without this, `started_at` (and the serialized current billing period)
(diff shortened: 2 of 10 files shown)


mikeh-lago:
All four addressed in 0000ea2:

1. GraphQL parity. Extended the same `max(Time.current, started_at)` anchoring to `Types::Subscriptions::Object#current_billing_period_started_at` and `_ending_at`. These are resolver changes only, with no schema change, so no codegen.
2. `period_end_date` (line 87). Same underlying issue: `Time.current` plus `next_end_of_period` returns the period containing "now", which is before a pending subscription's start. Fixed it here too.
3. Full-pipeline coverage. Added a request spec that posts to `/invoices/preview` with a scheduled downgrade, plus GraphQL resolver specs for the future-start case.
4. `compact` comment. Clarified that it handles a nil `started_at`.

app/graphql/types/subscriptions/object.rb
```diff
@@ -116,7 +116,18 @@ def fixed_charges
...
+      # NOTE: For a subscription that has not started yet (e.g. a scheduled downgrade), `Time.current`
+      #       precedes `started_at`, so DatesService computes boundaries for the period containing "now"
+      #       — before the subscription exists — and collapses current_billing_period_* (and skews
+      #       period_end_date) onto `started_at`. Anchoring on the later of the two reports the real
+      #       first period. No-op for already-started subscriptions (started_at is in the past).
+      #       `compact` guards against a nil started_at (pending subscription with no start yet); it is
+      #       load-bearing — `max` raises on a nil element, so do not drop it.
+      def billing_reference_time
+        [Time.current, object.started_at].compact.max
       end
```
rsempe: can we avoid duplication of `billing_reference_time`?

spec/graphql/types/subscriptions/object_spec.rb
```diff
@@ -60,4 +60,50 @@
     expect(subject).to have_field(:activation_rules).of_type("[SubscriptionActivationRule!]!")
     expect(subject).to have_field(:cancelation_reason).of_type("CancelationReasonEnum")
   end
+
+  # BIL-97: for a subscription that has not started yet (e.g. a scheduled downgrade), the billing
```
rsempe: BIL-97 is not available publicly, can we delete it

rsempe:
Is it possible to remove all the AI-related noise from this PR (comments, for example)?

Adding a scenario test case would also be really helpful 🙏

mikeh-lago:
Done in 322e78d. Moved the helper onto the model as `Subscription#billing_reference_time` (`[Time.current, started_at].compact.max`). `V1::SubscriptionSerializer#dates_service` and `Types::Subscriptions::Object#dates_service` and `#period_end_date` now call it, so the logic lives in one place. Added a model spec covering the already-started, future-start, and nil `started_at` cases.

mikeh-lago:
Added in 33de029, in `spec/scenarios/invoices/preview_spec.rb`, context "when a downgrade is scheduled".

It runs the full flow: it creates an anniversary subscription through the API, schedules a real downgrade (a second `create_subscription` on a cheaper plan, so the pending `next_subscription` is set up by the actual service rather than by hand), then posts to `/invoices/preview` before the rotation date. On the pending plan it checks:

- `started_at` and `current_billing_period_started_at` are `2026-07-03T00:00:00Z`, and `current_billing_period_ending_at` is `2026-08-02T23:59:59Z`
- the two dates are not equal
- the period matches the subscription fee's `from_date` and `to_date` for the same plan on the same invoice

It is marked `transaction: false` to match `subscriptions/downgrade_spec.rb`.

app/graphql/types/subscriptions/object.rb
mikeh-lago: Done, moved it to the model as `Subscription#billing_reference_time` (be5fba1 and 322e78d). The serializer and both GraphQL resolvers now call it.

spec/graphql/types/subscriptions/object_spec.rb
mikeh-lago: Removed the ticket references from all comments in be5fba1.

mikeh-lago:
Cleaned things up: removed the internal ticket references from the comments, simplified the explanatory comments, and trimmed the PR description, title, and earlier comments. The scenario test is in (33de029). Both inline threads are resolved.

app/services/invoices/preview/find_subscriptions_service.rb
```diff
@@ -64,6 +64,14 @@ def rotation_date(subscription)
           .new_instance(subscription, Time.current, current_usage: true)
           .end_of_period + 1.day
       end
+
+      # rotation_date lands at the end of the day, but the new subscription starts at the beginning
+      # of that day in the customer timezone.
+      def next_subscription_started_at(subscription)
```
tiagolupepic, line 70: There is a bug here (it was before) and it's happening in the rotation_date. Since the service is looping in each subscription, the rotation_date will be memoized for the first subscription, the next subscription will get the wrong datetime.
The fix is to memoize by subscription.id instead

something like this

```
@rotation_date ||= {}
@rotation_date[subscription.id] = ...
```

rsempe: That's true but could be tackled in a future PR as it's not related to these changes

rsempe (approved):

app/services/invoices/preview/find_subscriptions_service.rb
mikeh-lago: Good catch, and since this one predates this PR I pulled it into its own ticket so it can be fixed and reviewed on its own: BIL-106 (https://linear.app/getlago/issue/BIL-106). It captures the per-subscription memoization fix and a regression test for multiple downgrades in one preview.

PR #5650, opened 2026-06-04
chore(clock): Isolate clock analytics under an env var if needed

jdenquin:
## Context

- We want to add a dedicated Redis instance for analytics jobs, to avoid impacting billing performances
- Sharding with Sidekiq is not that easy and recommended and need more conception https://github.com/sidekiq/sidekiq/wiki/Sharding
- Isolating all analytics jobs with an other `REDIS_URL` is the fastest and best option right now

## Changes

- Add a `LAGO_REDIS_ANALYTICS_ENABLED` env var
- Add a `clock_analytics.rb` file that only take cares of the dialy usage for now
- Move Daily Usage clock job on `analytics` queue so the analytics worker will take care of it, since it will be running on an isolated Redis instance, we can't use the default clock and clock worker, the job would never be performed.

### DIFF
--- a/app/jobs/clock/compute_all_daily_usages_job.rb
+++ b/app/jobs/clock/compute_all_daily_usages_job.rb
@@ -4,6 +4,16 @@ module Clock
   class ComputeAllDailyUsagesJob < ClockJob
     unique :until_executed, on_conflict: :log
 
+    queue_as do
+      if ActiveModel::Type::Boolean.new.cast(ENV["SIDEKIQ_CLOCK"])
+        :clock_worker
+      elsif ActiveModel::Type::Boolean.new.cast(ENV["LAGO_REDIS_ANALYTICS_ENABLED"])
+        :analytics
+      else
+        :clock
+      end
+    end
+
     def perform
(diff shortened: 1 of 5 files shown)


app/jobs/clock/compute_all_daily_usages_job.rb
```diff
@@ -4,6 +4,16 @@ module Clock
   class ComputeAllDailyUsagesJob < ClockJob
     unique :until_executed, on_conflict: :log
 
+    queue_as do
+      if ActiveModel::Type::Boolean.new.cast(ENV["SIDEKIQ_CLOCK"])
+        :clock_worker
+      elsif ActiveModel::Type::Boolean.new.cast(ENV["LAGO_REDIS_ANALYTICS_ENABLED"])
+        :analytics
```
vincent-pochet, line 11: To ensure that the clocks have priority over existing jobs in the queue, I think we should be pushed it to an "analytics_high_priority" queue instead.
The queue should be added to the `config/sidekiq/sidekiq_analytics.yml` as well

vincent-pochet (approved):

PR #5651, opened 2026-06-04
fix(events): Preserve timestamp precision

tiagolupepic:
## Context

Event creation converted received epoch timestamps with Float before passing them to Time.zone.at. Decimal fractions such as 1780586634.1 cannot be represented exactly as binary floats, so serialized timestamps using iso8601(9) could become values like 2026-06-04T15:23:54.099999904Z instead of the received 2026-06-04T15:23:54.100000000Z.

## Description

Use BigDecimal when converting event timestamps in single and batch event creation so decimal epoch fractions keep their received precision. Add regression coverage for timestamps ending in .1, .2, and .3 to assert the nanosecond-precision ISO8601 output.

### DIFF
--- a/app/services/events/create_batch_service.rb
+++ b/app/services/events/create_batch_service.rb
@@ -49,7 +49,7 @@ module Events
         event.external_subscription_id = event_params[:external_subscription_id]
         event.properties = event_params[:properties] || {}
         event.metadata = metadata || {}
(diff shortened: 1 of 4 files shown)


groyoh (approved):

rsempe (approved):
👍

PR #5652, opened 2026-06-04
feat(products): add the product catalog data model

rsempe:
**Stack:** #5652 → #5690 → #5691 → #5707 → #5774 → #5775 → #5823 → #5820 → #5776 → #5779 → #5842 → #5871 → #5836 — **1/13**, merges into `main`.

## Context

Foundational data model for the product catalog (billing v2). It replaces the tight coupling between plans and charges with a composable layer: products are priced by rate cards, attached to plans and subscriptions, and sequenced by rate phases.

Vocabulary: `ProductCategory` (optional grouping) → `Product` (the billable unit) → `ProductFilter` (a slice of a product) → `RateCard` (prices one product or one slice) → `Rate` (effective-dated price on a card)

## Description

- Add `product_categories` — an optional grouping. It has no pricing role; a product can exist without one.
- Add `products` (`usage` | `fixed`) with type-specific validation: usage requires a billable metric, fixed forbids one. `add_on_id`/`charge_id` are nullable migration bridges. Code is unique per organization.
- Add `product_filters` + `product_filter_values`: a filter is a named combination of billable-metric-filter key/values. Values combine with AND across keys and OR within a key; a NULL value matches any value of the key (key-only selection).
- Add `rate_cards`, carrying the billing-semantic fields (currency, billing timing, proration, paid-fee regrouping, pricing unit), scoped either to a whole product or to one of its filter slices.
- Add `rate_card_rates`: effective-dated rates on an append-only timeline. Status (pending/active/terminated) is derived from `effective_datetime`, never stored. Each rate carries a mandatory caller-provided `code`, unique per card.
- Add `plan_rate_cards` — the plan-side attachment, binding a rate card to a plan with a default `units`, one per plan/card pair.
- Add `subscription_rate_cards` — the runtime rows the billing engine scans, with `billing_anchor_date`, `next_billing_at` and a `started_at`/`ended_at` validity window. Rows are versioned: at most one live row per subscription/card pair.
- Add `rate_phases`: the sequencing layer under a plan or subscription attachment (exactly one parent, enforced by check constraint), addressed by a mandatory `code` unique per parent and ordered by `position`.
- Add `rate_overrides`: per-phase price replacements, reusing the rate enums. An override belongs to exactly one phase.
- Add optional `rate_card_rate_id` / `rate_override_id` references and a `product` fee type to `fees`. Both FKs are added `NOT VALID` and validated in follow-up migrations.

Includes models, migrations, factories and model specs for every new entity.

### DIFF
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -21,6 +21,9 @@ class Fee < ApplicationRecord
   belongs_to :organization
   belongs_to :billing_entity
   belongs_to :fixed_charge, -> { with_discarded }, optional: true
+  # Fees are historical records: they must resolve their pricing source
+  # even after the rate is discarded.
+  belongs_to :rate_card_rate, -> { with_discarded }, optional: true
 
   has_one :adjusted_fee, dependent: :nullify
   has_one :billable_metric, -> { with_discarded }, through: :charge
@@ -45,7 +48,7 @@ class Fee < ApplicationRecord
   monetize :unit_amount_cents, disable_validation: true, allow_nil: true, with_model_currency: :currency
 
   # TODO: Deprecate add_on type in the near future
-  FEE_TYPES = %i[charge add_on subscription credit commitment fixed_charge].freeze
+  FEE_TYPES = %i[charge add_on subscription credit commitment fixed_charge product].freeze
   PAYMENT_STATUS = %i[pending succeeded failed refunded].freeze
 
   enum :fee_type, FEE_TYPES
@@ -414,6 +417,7 @@ end
 #  original_fee_id                     :uuid
 #  pay_in_advance_event_id             :uuid
 #  pay_in_advance_event_transaction_id :string
+#  rate_card_rate_id                   :uuid
 #  subscription_id                     :uuid
 #  true_up_parent_fee_id               :uuid
 #
@@ -436,6 +440,7 @@ end
 #  index_fees_on_organization_id_and_created_at_and_id  (organization_id,created_at,id) WHERE (deleted_at IS NULL)
 #  index_fees_on_original_fee_id                        (original_fee_id)
 #  index_fees_on_pay_in_advance_event_transaction_id    (pay_in_advance_event_transaction_id) WHERE (deleted_at IS NULL)
+#  index_fees_on_rate_card_rate_id                      (rate_card_rate_id)
 #  index_fees_on_subscription_id                        (subscription_id)
 #  index_fees_on_true_up_parent_fee_id                  (true_up_parent_fee_id)
 #
@@ -450,6 +455,7 @@ end
 #  fk_rails_...  (invoice_id => invoices.id)
 #  fk_rails_...  (organization_id => organizations.id)
 #  fk_rails_...  (original_fee_id => fees.id)
+#  fk_rails_...  (rate_card_rate_id => rate_card_rates.id)
 #  fk_rails_...  (subscription_id => subscriptions.id)
 #  fk_rails_...  (true_up_parent_fee_id => fees.id)
 #
--- a/app/models/fee.rb
+++ b/app/models/fee.rb
@@ -24,6 +24,7 @@ class Fee < ApplicationRecord
   # Fees are historical records: they must resolve their pricing source
   # even after the rate is discarded.
   belongs_to :rate_card_rate, -> { with_discarded }, optional: true
+  belongs_to :rate_override, -> { with_discarded }, optional: true
 
   has_one :adjusted_fee, dependent: :nullify
   has_one :billable_metric, -> { with_discarded }, through: :charge
@@ -418,6 +419,7 @@ end
 #  pay_in_advance_event_id             :uuid
 #  pay_in_advance_event_transaction_id :string
 #  rate_card_rate_id                   :uuid
+#  rate_override_id                    :uuid
 #  subscription_id                     :uuid
 #  true_up_parent_fee_id               :uuid
 #
@@ -441,6 +443,7 @@ end
 #  index_fees_on_original_fee_id                        (original_fee_id)
 #  index_fees_on_pay_in_advance_event_transaction_id    (pay_in_advance_event_transaction_id) WHERE (deleted_at IS NULL)
 #  index_fees_on_rate_card_rate_id                      (rate_card_rate_id)
+#  index_fees_on_rate_override_id                       (rate_override_id)
 #  index_fees_on_subscription_id                        (subscription_id)
 #  index_fees_on_true_up_parent_fee_id                  (true_up_parent_fee_id)
 #
@@ -456,6 +459,7 @@ end
 #  fk_rails_...  (organization_id => organizations.id)
 #  fk_rails_...  (original_fee_id => fees.id)
 #  fk_rails_...  (rate_card_rate_id => rate_card_rates.id)
+#  fk_rails_...  (rate_override_id => rate_overrides.id)
 #  fk_rails_...  (subscription_id => subscriptions.id)
 #  fk_rails_...  (true_up_parent_fee_id => fees.id)
 #
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -28,6 +28,8 @@ class Plan < ApplicationRecord
   has_many :coupons, through: :coupon_targets
   has_many :invoices, through: :subscriptions
   has_many :usage_thresholds
+  has_many :applied_rate_cards, class_name: "PlanRateCard"
+  has_many :products, through: :applied_rate_cards
 
   has_many :applied_taxes, class_name: "Plan::AppliedTax", dependent: :destroy
   has_many :taxes, through: :applied_taxes
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -17,6 +17,8 @@ class Product < ApplicationRecord
   belongs_to :add_on, -> { with_discarded }, optional: true
   belongs_to :charge, -> { with_discarded }, optional: true
 
+  has_many :filters, class_name: "ProductFilter"
+
   enum :product_type, PRODUCT_TYPES, validate: true
 
   validates :name, presence: true
@@ -40,9 +42,11 @@ class Product < ApplicationRecord
   private
 
   def validate_billable_metric_presence
-    if usage? && billable_metric_id.blank?
+    has_billable_metric = billable_metric_id.present? || billable_metric.present?
+
+    if usage? && !has_billable_metric
       errors.add(:billable_metric, :blank)
-    elsif fixed? && billable_metric_id.present?
+    elsif fixed? && has_billable_metric
       errors.add(:billable_metric, :present)
     end
   end
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -3,6 +3,7 @@
 class Product < ApplicationRecord
   include PaperTrailTraceable
   include Discard::Model
+  include CatalogAttachable
 
   self.discard_column = :deleted_at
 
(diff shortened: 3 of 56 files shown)


db/migrate/20260604181531_create_plan_products.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class CreatePlanProducts < ActiveRecord::Migration[8.0]
+  def change
+    create_table :plan_products, id: :uuid do |t|
```
rsempe: I'm wondering if this table is still needed - Looking on it 👀

app/models/rate_card.rb
```diff
@@ -0,0 +1,85 @@
...
+    advance: "advance"
+  }.freeze
+
+  REGROUP_PAID_FEES = {
+    invoice: "invoice"
+  }.freeze
+
+  belongs_to :organization
+  belongs_to :product_item
+  belongs_to :product_item_filter, optional: true
```
endenis: Should we also add a validation that checks that `product_item_filter` belongs to the `product_item`?

db/migrate/20260604181531_create_plan_products.rb
rsempe: Deleted ✅

db/migrate/20260604180513_create_product_item_filter_values.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      t.datetime :deleted_at
+
+      t.timestamps
+
+      t.index :deleted_at
+      t.index [:product_item_filter_id, :billable_metric_filter_id, :value],
+        unique: true,
+        where: "deleted_at IS NULL",
+        name: "idx_pif_values_on_filter_metric_filter_and_value"
```
endenis: By default postgres will consider nulls as different values, so this index would allow multiple entries with the same `product_item_filter_id` and `billable_metric_filter_id` if the `value` is null.

To prevent this I suggest to add `NULLS NOT DISTINCT` to this unique index:

```suggestion
      t.index [:product_item_filter_id, :billable_metric_filter_id, :value],
        unique: true,
        where: "deleted_at IS NULL",
        nulls_not_distinct: true,
        name: "idx_pif_values_on_filter_metric_filter_and_value"
```

app/models/fee.rb
```diff
@@ -45,7 +47,7 @@ class Fee < ApplicationRecord
   monetize :unit_amount_cents, disable_validation: true, allow_nil: true, with_model_currency: :currency
 
   # TODO: Deprecate add_on type in the near future
-  FEE_TYPES = %i[charge add_on subscription credit commitment fixed_charge].freeze
+  FEE_TYPES = %i[charge add_on subscription credit commitment fixed_charge product_item].freeze
```
tiagolupepic: we need to include this type in the export_fees view as well

app/models/plan.rb
```diff
@@ -28,6 +28,8 @@ class Plan < ApplicationRecord
   has_many :coupons, through: :coupon_targets
   has_many :invoices, through: :subscriptions
   has_many :usage_thresholds
+  has_many :applied_rate_cards, class_name: "PlanRateCard"
```
tiagolupepic, line 31: out of curiosity, why the name is applied_rate_cards and not just rate_cards?

app/models/rate_card.rb
rsempe: The service can't produce a mismatch right now but I agree, let's add a model validation 👍

db/migrate/20260604180513_create_product_item_filter_values.rb
rsempe: Good catch 👍

app/models/fee.rb
rsempe: Agreed 👍 - Will do it separately

app/models/plan.rb
rsempe: Because the association returns the attachment rows (`PlanRateCard`: `units`, `rate phases`, one per `plan×card`), not the catalog rate cards. It's a different resource, here it's the card's configuration inside one plan

app/models/rate_card_rate.rb
```diff
@@ -0,0 +1,122 @@
...
+  validates :code, uniqueness: {scope: :rate_card_id, conditions: -> { where(deleted_at: nil) }}
+  validates :billing_interval_unit, presence: true
+  validates :effective_datetime, presence: true
+  validates :rate_model, presence: true
+  validates :min_amount_cents, numericality: {greater_than_or_equal_to: 0}
+  validates :billing_interval_count, numericality: {greater_than_or_equal_to: 1}
+
+  validate :validate_effective_datetime_is_appended
+  validate :validate_pricing_unit_conversion_rate
+
```
endenis, line 57: Should `RateCardRate` validate rate properties like it's done in the `RateOverride` class?

app/models/fee.rb
```diff
@@ -21,6 +21,8 @@ class Fee < ApplicationRecord
   belongs_to :organization
   belongs_to :billing_entity
   belongs_to :fixed_charge, -> { with_discarded }, optional: true
+  belongs_to :rate_card_rate, optional: true
+  belongs_to :rate_override, optional: true
```
endenis: Should these relations also include the `with_discarded` scope?

```suggestion
  belongs_to :rate_card_rate, -> { with_discarded }, optional: true
  belongs_to :rate_override, -> { with_discarded }, optional: true
```

app/models/rate_phase.rb
```diff
@@ -0,0 +1,78 @@
...
+  validates :code, presence: true
+  validates :code,
+    uniqueness: {scope: :plan_rate_card_id, conditions: -> { where(deleted_at: nil) }},
+    if: :plan_rate_card_id?
+  validates :code,
+    uniqueness: {scope: :subscription_rate_card_id, conditions: -> { where(deleted_at: nil) }},
+    if: :subscription_rate_card_id?
+  validates :position, presence: true, numericality: {greater_than: 0}
+  validates :rate_override_id, uniqueness: {conditions: -> { where(deleted_at: nil) }}, allow_nil: true
+
```
endenis, line 25: Should there also be a validation for the `billing_interval_cycle_count` column? (nil, greater or greater and equal to zero)

app/models/fee.rb
rsempe: Done ✅

app/models/rate_phase.rb
rsempe: Done ✅

app/models/rate_card_rate.rb
rsempe: It's done on this PR: https://github.com/getlago/lago-api/pull/5707

endenis (approved):

app/models/product.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def self.ransackable_attributes(_auth_object = nil)
+    %w[name code]
+  end
+
+  def invoice_name
+    invoice_display_name.presence || name
+  end
+
+  def attached_to_plan_or_subscription?
```
tiagolupepic: This method is used in some services (like `ProductItemFilters::UpdateService`), what is the business logic behind? 

Also, since both relations are using rate_cards, isn't better to ask that model instead? 
If I'm not wrong, we're repeating the logic in RateCard as well.

app/models/product_category.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def self.ransackable_attributes(_auth_object = nil)
+    %w[name code]
+  end
+
+  def invoice_name
+    invoice_display_name.presence || name
+  end
+
+  def attached_to_plan_or_subscription?
```
tiagolupepic: Same method here

app/models/product_filter_value.rb
```diff
@@ -0,0 +1,62 @@
...
+  # created_at keeps the values order stable across edits.
+  default_scope -> { kept.order(created_at: :asc) }
+
+  delegate :key, to: :billable_metric_filter
+
+  private
+
+  def validate_value_inclusion
+    return if value.blank?
+    return if billable_metric_filter&.values&.include?(value) # rubocop:disable Performance/InefficientHashSearch
```
tiagolupepic, line 28: you can convert here to a `Set` and will perform a O(1)

app/models/rate_card.rb
```diff
@@ -0,0 +1,94 @@
...
+  belongs_to :organization
+  belongs_to :product
+  belongs_to :product_filter, optional: true
+
+  has_many :rates, class_name: "RateCardRate"
+  has_many :plan_applied_rate_cards, class_name: "PlanRateCard"
+  has_many :subscription_applied_rate_cards, class_name: "SubscriptionRateCard"
+
+  enum :billing_timing, BILLING_TIMINGS, validate: true
+  enum :regroup_paid_fees, REGROUP_PAID_FEES, validate: {allow_nil: true}
```
tiagolupepic: I think we should not allow nil here, better to have a `NONE` option

app/models/rate_card_rate.rb
```diff
@@ -0,0 +1,122 @@
...
+  belongs_to :organization
+  belongs_to :rate_card
+
+  has_many :fees
+
+  # allow_nil + explicit presence: a missing value reads as value_is_mandatory,
+  # an unknown value as value_is_invalid.
+  enum :rate_model, RATE_MODELS, validate: {allow_nil: true}
+  enum :billing_interval_unit, BILLING_INTERVAL_UNITS, validate: {allow_nil: true}
```
tiagolupepic, line 41: same here, we should not allow nil, better to have a specific value for that.

db/migrate/20260604181826_create_subscription_rate_cards.rb
```diff
@@ -0,0 +1,34 @@
...
+      t.references :organization, null: false, foreign_key: true, type: :uuid
+      t.references :subscription, null: false, foreign_key: true, type: :uuid
+      t.references :rate_card, null: false, foreign_key: true, type: :uuid
+
+      t.date :billing_anchor_date, null: false
+      t.datetime :next_billing_at, null: false
+      t.datetime :started_at, null: false
+      t.datetime :ended_at
+
+      t.decimal :units, precision: 30, scale: 10
```
tiagolupepic: I had some issues in presentation breakdown defining this precision https://github.com/getlago/lago-api/pull/5367. It was a mismatch with fee units and it was rounding. 
I think it worth to double check and also the other tables (units).

endenis (approved):

app/models/product.rb
rsempe: Adding a `CatalogAttachable` concern 👍

app/models/rate_card.rb
rsempe: Good spot

app/models/product_filter_value.rb
rsempe: it is a false positive, the cop sees `.values.include?` and assumes a `Hash`, but `values` is an array column

app/models/rate_card_rate.rb
rsempe: `nil` is forbidden by the presence validations below (and `NOT NULL` in the schema)
`allow_nil` here only keeps a missing value on the presence error (`value_is_mandatory`) instead of **also** failing inclusion (`value_is_invalid`)

db/migrate/20260604181826_create_subscription_rate_cards.rb
rsempe: thanks - doing it

PR #5653, opened 2026-06-04
fix(void PB): credits fix for voiding PB invoices

annvelents:
## Context

When voiding a PB invoice, 2 wrong behaviours were spotted:
1) 2 credit notes were issued with weird amounts
2) progressive billing credits, applied on the current invoice, were not "recredited" back

## Description

- Refactored void service
- when calculating applicable credit amount only deduct credits, applied on not voided invoices

### DIFF
--- a/app/services/invoices/void_service.rb
+++ b/app/services/invoices/void_service.rb
@@ -102,7 +102,7 @@ module Invoices
         description: "Credit note created due to voided invoice #{invoice.id}",
         credit_amount_cents: credit_amount,
         refund_amount_cents: refund_amount,
-        items: credit_note_items_matching_total_amount(invoice:, total_amount_cents: requested_credit_note_total_amount_cents)
+        items: credit_note_items_matching_requested_amount(invoice:, requested_amount_cents: requested_credit_note_total_amount_cents)
       )
     end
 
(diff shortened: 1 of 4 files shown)


app/services/subscriptions/progressive_billed_amount.rb
```diff
@@ -59,7 +59,7 @@ def call
 
       result.to_credit_amount = invoice.fees_amount_cents
       result.to_credit_amount -= invoice.coupons_amount_cents
-      result.to_credit_amount -= invoice.progressive_billing_credits.sum(:amount_cents)
+      result.to_credit_amount -= invoice.progressive_billing_credits.active.sum(:amount_cents)
```
annvelents, line 62: note: 
```
# credit.rb
scope :active, -> { joins(:invoice).where.not(invoices: {status: :voided}) }
```
so the original PB invoice, applied on the original subscription invoice keeps the credit, but it's applied now on voided invoice, as result this progressive_billing_credit is available again

rsempe (approved):
LGTM

app/services/invoices/void_service.rb
```diff
@@ -85,48 +85,81 @@ def valid_credit_note_amounts?
...
+        description: "Credit note created due to voided invoice #{invoice.id}",
+        credit_amount_cents: credit_amount,
+        refund_amount_cents: refund_amount,
+        items: credit_note_items_matching_total_amount(invoice:, total_amount_cents: requested_credit_note_total_amount_cents)
+      )
+    end
+
+    def create_voided_credit_note_for_remaining_amount!
+      remaining_items = remaining_credit_note_items(invoice:)
+      estimated_credit_note = estimate_credit_note(invoice:, items: remaining_items)
```
tiagolupepic, line 111: any chance to reduce the amount of methods? 

`remaining_credit_note_items`, `estimate_credit_note` and `estimated_credit_note_total_amount_cents` are quite similar, I think they can be a single method instead.

tiagolupepic (approved):
just small comment and it's not a blocker

:shipit:

app/services/invoices/void_service.rb
annvelents: `estimate_credit_note` and `estimated_credit_note_total_amount_cents` - agree, used one method for them
`remaining_credit_note_items` - should live separately since it returns a hash with fees and the remaining amount we can credit back

PR #5655, opened 2026-06-05
perf(alerts): fix N+1 on billable_metric lookup in CreateBatchService

tiagolupepic:
Api::V1::Subscriptions::AlertsController#create with a batch payload was triggering one billable_metrics SELECT per alert (21 queries observed in production). Each iteration in CreateBatchService called CreateAlertService independently, which resolved the billable_metric_code or billable_metric_id with a separate find_by! query.

Fixes API-21J

Preload all needed billable metrics in a single query before the loop, collecting all codes and ids across the batch upfront. Each alert's params are then enriched with the resolved object so CreateAlertService hits the already-present params[:billable_metric] branch and skips the query. When a metric is not found in the preloaded set, the params are left unchanged so CreateAlertService still handles the not_found failure naturally.

### DIFF
--- a/app/services/usage_monitoring/alerts/create_batch_service.rb
+++ b/app/services/usage_monitoring/alerts/create_batch_service.rb
@@ -29,7 +29,7 @@ module UsageMonitoring
               create_result = CreateAlertService.call(
                 organization:,
                 alertable:,
-                params: alert_params.to_h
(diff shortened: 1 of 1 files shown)


D1353L (approved):
Looks good 👍

PR #5662, opened 2026-06-08
fix(customers): Improve customers query

jdenquin:
## Context

- The customer list search (`CustomersQuery`) was slow on large organizations despite having trigram GIN indexes on all six searchable columns (`name, firstname, lastname, legal_name, external_id, email`).

- The bottleneck was the query shape, not missing indexes. Ransack generated a single `OR` across the six ILIKE columns:

```SQL
  WHERE organization_id = ?
    AND (name ILIKE ? OR firstname ILIKE ? OR ... OR email ILIKE ?)
  ORDER BY created_at DESC LIMIT 20
```

- With this shape the planner overestimates how many rows match, abandons the trigram indexes, and instead walks an ordering index in `created_at DESC`, applying the `OR` as a per-row filter — scanning a large fraction of the org's customers before collecting a page. The GIN indexes go unused.

## Change

- Rewrote the search to a `UNION` of one branch per column. Each branch filters on a single indexed column, so the planner can serve it directly from that column's trigram GIN index; the results are unioned and the main scope is filtered by the resulting ids.

  - base_scope returns the plain org scope when there's no search term, otherwise `WHERE id IN (<union>)`.
  - Added `matching_ids_by_search` building the per-column `UNION`.
  - `with_deleted` is preserved: branches add `with_discarded` when set, and the outer wrapper uses `Customer.unscoped` so the id-only union isn't re-scoped on `deleted_at`.
  - Removed the now-unused search_params ransack helper.

### DIFF
--- a/app/queries/customers_query.rb
+++ b/app/queries/customers_query.rb
@@ -59,10 +59,6 @@ class CustomersQuery < BaseQuery
     Customer.where(organization:).where(id: matching_ids_by_search)
   end
 
-  # Rather than a single OR across the searchable columns (which the planner
-  # evaluates as a per-row filter and cannot serve from the per-column trgm GIN
-  # indexes), build one branch per column and UNION them so each branch uses its
-  # own index, then filter the main scope by the resulting ids.
   def matching_ids_by_search
     search_base = Customer.where(organization:)
     search_base = search_base.with_discarded if filters.with_deleted
--- a/app/queries/customers_query.rb
+++ b/app/queries/customers_query.rb
@@ -63,8 +63,10 @@ class CustomersQuery < BaseQuery
     search_base = Customer.where(organization:)
     search_base = search_base.with_discarded if filters.with_deleted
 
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):

groyoh (approved):

vankiru (approved):
Nice! 👍

app/queries/customers_query.rb
```diff
@@ -52,7 +54,23 @@ def filters_contract
...
+  end
+
+  def matching_ids_by_search
+    search_base = Customer.where(organization:)
+    search_base = search_base.with_discarded if filters.with_deleted
+
+    escaped_term = "%#{Customer.sanitize_sql_like(search_term)}%"
+
+    branches = SEARCHABLE_FIELDS.map do |field|
+      search_base.where("customers.#{field} ILIKE ?", escaped_term).select(:id)
```
annvelents, line 69: Nice! Very interesting solution!

annvelents (approved):

PR #5668, opened 2026-06-09
misc(daily_usages): Auto retry fill history jobs + specs

vincent-pochet:
## Description

This PR adds an auto retry handler for `ActiveRecord::ActiveRecordError` on `DailyUsages::FillHistoryJob`.
It also adds missing tests for the job class

### DIFF
--- a/app/jobs/daily_usages/fill_history_job.rb
+++ b/app/jobs/daily_usages/fill_history_job.rb
@@ -10,8 +10,14 @@ module DailyUsages
       end
     end
 
+    retry_on DailyUsages::RetryableError, wait: :polynomially_longer, attempts: 6
+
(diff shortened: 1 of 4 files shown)


app/jobs/daily_usages/fill_history_job.rb
```diff
@@ -10,6 +10,8 @@ class FillHistoryJob < ApplicationJob
       end
     end
 
+    retry_on ActiveRecord::ActiveRecordError, wait: :polynomially_longer, attempts: 6
```
groyoh: This could catch unwanted errors. Maybe we should explicitly catch, check the message and re-raise a specific retryable error in such case ?

groyoh (approved):
Could you add a test for the retry (we have it in other job tests) ?

PR #5669, opened 2026-06-09
feat(redis): Improve Sidekiq Redis client resilience

groyoh:
## Context

During an ElastiCache instance upgrade or failover the Redis primary is briefly unreachable. The connection drops while a replica is promoted, and the promoted node then replies `LOADING Redis is loading the dataset in memory` until it finishes loading its RDB. With Sidekiq's default single reconnect, enqueues and the worker fetch loop error out during this window.

Retry was previously handled at the Redlock level (`lib/redlock/client_patch.rb`), which only covered the ActiveJob uniqueness locks and retried lock acquisition on any error. Note that the `redis_timeout: 5` configuration for `activejob_uniqueness` did not actually worked since we configured the client directly within the initializer. This setting only works when passing the Redis url and therefore defaults to 1 second timeout.

## Description

`RedisConfigBuilder#sidekiq` derives a retry schedule from `LAGO_REDIS_SIDEKIQ_RETRY_WINDOW_SECONDS` (default 5 seconds): a series of quadratically increasing backoff intervals (0.1, 0.4, 0.9, ...) whose total stays within the window, each jittered by +-25% at boot so processes don't retry in lockstep. The same schedule feeds two mechanisms:

- `reconnect_attempts` for connection drops (redis-client's built-in retry).
- `RedisLoadingRetryMiddleware`, a redis-client middleware that retries commands rejected with a LOADING error (matched on `e.code == "LOADING"`) and logs a warning before each retry. The schedule is passed per-client through `custom[:loading_retry_attempts]`.

The builder owns the `middlewares` and `custom` wiring, so the Sidekiq and ActiveJob uniqueness clients pick it up from `#sidekiq` with no extra wiring. Retry is now centralized at the Redis connection layer, so the redundant Redlock retry patch is removed and Redlock's own `retry_count` is set to 0 to avoid compounding retries. The client `timeout` is lowered to 1 second now that the retry window absorbs transient errors (it was already the case for Redlock anyway).

One caveat on the Redlock path: after a lock acquisition fails, Redlock issues an unlock, so the connection-retry window runs a second time. A fully unreachable Redis can therefore block an enqueue for ~10s or more (two windows plus the per-attempt timeouts) before the error bubbles up.

## Reproduction

Force a real LOADING window against a local Redis by giving it a large dataset to reload on boot. `DEBUG POPULATE` requires Redis to start with `--enable-debug-command local` (the DEBUG command is gated by default on Redis 7), so the `redis` service has to be launched with that flag.

```
# Create ~10M keys instantly (bump the count if the load is too fast to catch).
lago exec redis redis-cli DEBUG POPULATE 10000000
# Persist them so they reload at startup.
lago exec redis redis-cli SAVE
# Restart so Redis loads the RDB asynchronously and replies LOADING during the window.
lago restart redis
# Within the window, commands are rejected:
lago exec redis redis-cli PING
#   => LOADING Redis is loading the dataset in memory
```

With the node loading, an enqueue landing in the window logs a warning and retries on the schedule, then succeeds once the node finishes loading. A primary that stays down past the window surfaces `RedisClient::CannotConnectError` after `reconnect_attempts` is exhausted.

Cleanup:

```
lago exec redis redis-cli FLUSHALL
lago exec redis redis-cli SAVE
```

## Manual test

Console session against a local Redis, taken down mid-test (job payloads, object dumps and backtraces stripped). The client config carries the schedule for both `reconnect_attempts` and the LOADING middleware:

```
Sidekiq connecting to Redis with options {..., timeout: 1,
  reconnect_attempts: [0.1, 0.3, 0.8, 1.6, 2.1],
  middlewares: [Lago::RedisLoadingRetryMiddleware],
  custom: {loading_retry_attempts: [0.1, 0.3, 0.8, 1.6, 2.1]}, ...}
```

Direct Sidekiq enqueue. Succeeds while Redis is up; once it is taken down, `reconnect_attempts` retries the connection across the window (each HELLO is a reconnect handshake), then surfaces the error:

```
> Clock::WebhooksCleanupJob.perform_later
07:59:10  LPUSH queue:clock        enqueue status=success

> Clock::WebhooksCleanupJob.perform_later   # Redis down
07:59:15  HELLO 3                  # reconnect attempt
07:59:16  HELLO 3
07:59:18  HELLO 3
07:59:20  HELLO 3
07:59:23  HELLO 3
07:59:24  enqueue status=error  RedisClient::CannotConnectError: Waited 1 seconds
```

ActiveJob uniqueness (Redlock) path. `retry_count: 0` means no extra Redlock retries on top, but Redlock issues an unlock after the failed lock, so the connection-retry window runs twice (once for the lock, once for the unlock) before the error surfaces wrapped. Here that is ~17s end to end:

```
> Clock::RefreshDraftInvoicesJob.perform_later
08:01:18  EVALSHA <lock>           enqueue status=success

> Clock::RefreshDraftInvoicesJob.perform_later   # Redis down
08:03:57  EVALSHA <lock>
08:03:58  HELLO 3                  # reconnect attempts for the LOCK
...
08:04:05  HELLO 3
08:04:06  HELLO 3                  # ...then again for the UNLOCK
...
08:04:13  HELLO 3
08:04:14  enqueue status=error  Redlock::LockAcquisitionError:
            RedisClient::CannotConnectError: Waited 1 seconds
```


lib/lago/redis_loading_retry_middleware.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      begin
+        yield
+      rescue RedisClient::CommandError => e
+        raise unless e.code == LOADING_CODE
+        raise if attempt >= attempts.size
+
+        interval = attempts[attempt]
+        Rails.logger.warn(
+          "Redis replied LOADING, retrying in #{interval}s (attempt #{attempt + 1}/#{attempts.size})"
```
groyoh: I might rework this to be hash-based rather than a string.

vincent-pochet (approved):

PR #5670, opened 2026-06-09
fix(billable_metrics): Serialize concurrent updates with row lock

D1353L:
## Context
Concurrent `PUT` requests updating the same billable metric could interleave and corrupt its filters.

- Two simultaneous requests each diffed incoming filters against stale state, producing duplicate filters or partially applied changes.
- Simultaneous same requests do heavy work, running expensive DB queries in parallel.
- Because the billable_metric save was outside the filter transaction, a failure in one half did not roll back the other - the update was not atomic.

## Description
- Added a pessimistic row lock on the billable metric, so concurrent updates of the same metric are serialized.
- Moved `billable_metric.save!` inside a transaction for atomicity -  any failure now rolls back.
- Switched the draft-invoice refresh in `CreateOrUpdateBatchService` from `perform_later` to `perform_after_commit`, so the refresh is only enqueued once the transaction has actually committed.

### DIFF
--- a/app/services/billable_metric_filters/create_or_update_batch_service.rb
+++ b/app/services/billable_metric_filters/create_or_update_batch_service.rb
@@ -55,7 +55,7 @@ module BillableMetricFilters
         end
       end
 
-      BillableMetricFilters::RefreshDraftInvoicesJob.perform_later(billable_metric.id)
+      BillableMetricFilters::RefreshDraftInvoicesJob.perform_after_commit(billable_metric.id)
 
       result
     end
--- a/app/services/billable_metrics/update_service.rb
+++ b/app/services/billable_metrics/update_service.rb
@@ -23,37 +23,37 @@ module BillableMetrics
         return result.forbidden_failure!
       end
 
-      billable_metric.name = params[:name] if params.key?(:name)
-      billable_metric.description = params[:description] if params.key?(:description)
+      billable_metric.with_lock do
+        billable_metric.name = params[:name] if params.key?(:name)
+        billable_metric.description = params[:description] if params.key?(:description)
 
-      ActiveRecord::Base.transaction do
         if params.key?(:filters)
           BillableMetricFilters::CreateOrUpdateBatchService.call(
             billable_metric:,
             filters_params: params[:filters].map { |f| f.to_h.with_indifferent_access }
           ).raise_if_error!
         end
-      end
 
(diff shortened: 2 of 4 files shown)


app/services/billable_metric_filters/create_or_update_batch_service.rb
```diff
@@ -55,7 +55,7 @@ def call
         end
       end
 
-      BillableMetricFilters::RefreshDraftInvoicesJob.perform_later(billable_metric.id)
+      BillableMetricFilters::RefreshDraftInvoicesJob.perform_after_commit(billable_metric.id)
```
tiagolupepic, line 58: check the specs, it's doing the assertions with `have_enqueued_job` and it should fail isn't?  Because we have `have_enqueued_job_after_commit`. Probably some is off there

app/services/billable_metrics/update_service.rb
```diff
@@ -23,37 +23,37 @@ def call
         return result.forbidden_failure!
       end
 
-      billable_metric.name = params[:name] if params.key?(:name)
-      billable_metric.description = params[:description] if params.key?(:description)
+      billable_metric.with_lock do
```
tiagolupepic, line 26: @endenis also used this approach in VoidService and works pretty nice. 
He added specs, https://github.com/getlago/lago-api/blob/main/spec/services/invoices/void_service_spec.rb#L221. I think we could follow the same idea, wdyt?

app/services/billable_metric_filters/create_or_update_batch_service.rb
D1353L: Checked this - `have_enqueued_job` asserts whether the job was enqueued at all. But `have_enqueued_job_after_commit` is better here as it asserts the timing - that the job was enqueued after commit

app/services/billable_metrics/update_service.rb
D1353L: Good idea 👍

app/services/billable_metric_filters/create_or_update_batch_service.rb
tiagolupepic: have_enqueued_job should exclude the ones that are enqueue after commit IMO. Otherwise we can introduce accidental bugs.

tiagolupepic (approved):
:shipit:

PR #5671, opened 2026-06-09
feat(payment-gated-subs): Restrict subscription Hubspot sync to active and terminated

osmarluz:
## Context

Subscriptions were synced to Hubspot regardless of status: a new subscription was pushed even while `pending` (future-dated) or `incomplete` (gated, awaiting payment), and the plan upgrade/downgrade "edit pending" paths synced pending subscriptions. With payment-gated subscriptions a subscription can sit `pending`/`incomplete` before it ever activates, and we don't want those reflected in Hubspot — a subscription should appear there only once it actually becomes active.

## Description

- Centralize the Hubspot create in `ActivateService#notify_started`, so the record is created when a subscription first transitions to active, regardless of the trigger (creation, scheduled activation, gated-payment success, upgrade/downgrade rotation).
- `CreateService` no longer syncs at create time. The only activation that bypasses `ActivateService` — backdated subscriptions in `handle_past_subscription` — creates the record there. The `during_creation` flag, which only existed to prevent a double create, is removed.
- Drop the Hubspot updates that targeted not-yet-active subscriptions: the `update_pending_subscription` paths in `PlanUpgradeService`/`PlanDowngradeService` (pending), and `cancel_next_subscription` in `TerminateService` (a never-activated pending downgrade, previously a no-op).


spec/services/subscriptions/create_service_spec.rb
```diff
@@ -138,6 +138,25 @@
         create_service.call
         expect(Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob).not_to have_been_enqueued
       end
+
+      context "when the subscription starts in the future" do
+        let(:subscription_at) { Time.current + 5.days }
+
+        it "does not sync to Hubspot while pending" do
+          create_service.call
+          expect(Integrations::Aggregator::Subscriptions::Hubspot::CreateJob).not_to have_received(:perform_later)
```
ancorcruz:
```suggestion
          expect(Integrations::Aggregator::Subscriptions::Hubspot::CreateJob).not_to have_been_enqueued```

to be consistent.

File: spec/services/subscriptions/create_service_spec.rb [hunk 2]
```diff
@@ -138,6 +138,25 @@
...
+          expect(Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob).not_to have_been_enqueued
+        end
+      end
+
+      context "when the subscription is backdated" do
+        let(:subscription_at) { Time.current - 5.days }
+
+        it "enqueues the Hubspot create job" do
+          create_service.call
+          expect(Integrations::Aggregator::Subscriptions::Hubspot::CreateJob).to have_received(:perform_later)
```
ancorcruz:
```suggestion
          expect(Integrations::Aggregator::Subscriptions::Hubspot::CreateJob).to have_been_enqueued```

File: spec/services/subscriptions/activate_all_pending_service_spec.rb [hunk 3]
```diff
@@ -115,10 +115,10 @@
...
 
-      it "enqueues Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob" do
-        allow(Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob).to receive(:perform_later)
+      it "enqueues Integrations::Aggregator::Subscriptions::Hubspot::CreateJob" do
+        allow(Integrations::Aggregator::Subscriptions::Hubspot::CreateJob).to receive(:perform_later)
         activate_service.call
-        expect(Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob)
+        expect(Integrations::Aggregator::Subscriptions::Hubspot::CreateJob)
           .to have_received(:perform_later).with(subscription: pending_subscription)
       end
```
ancorcruz, line 122:
```suggestion
      it "enqueues Integrations::Aggregator::Subscriptions::Hubspot::CreateJob" do
        activate_service.call
        expect(Integrations::Aggregator::Subscriptions::Hubspot::CreateJob).to have_been_queued.with(subscription: pending_subscription)
      end
```

ancorcruz (approved):
looks great.

PR #5672, opened 2026-06-09
fix(subscriptions): Fix UpdateOrOverrideFixedChargeService with apply_units_immediately

tiagolupepic:
Updating fixed-charge units through the subscription-scoped endpoint or GraphQL mutation created a subscription plan override in two steps. The plan clone first emitted a fixed charge event with the old units, then the dedicated fixed charge update emitted a second event with the new units.

This left a stale event at the next billing boundary and skipped the pay-in-advance delta invoice, causing under-billing for subscription fixed charge unit increases.

Pass fixed charge override params into the plan override creation so the cloned fixed charge is created with the target units from the start. Reuse the plan override concern for this path and avoid emitting a second event for newly cloned fixed charges.

Trigger the pay-in-advance fixed charge invoice job when subscription fixed charge units are applied immediately. Add model support for resolving a fixed charge parent and cover the invoice job dispatch conditions in service specs.

Fixes BIL-83

### DIFF
--- a/app/services/subscriptions/update_or_override_fixed_charge_service.rb
+++ b/app/services/subscriptions/update_or_override_fixed_charge_service.rb
@@ -10,6 +10,7 @@ module Subscriptions
       @subscription = subscription
       @fixed_charge = fixed_charge
       @params = params
+      @subscription_plan_parent_present = subscription&.plan&.parent_id.present?
 
       super
     end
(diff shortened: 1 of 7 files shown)


app/services/subscriptions/update_or_override_fixed_charge_service.rb
```diff
@@ -35,27 +39,26 @@ def call
...
+
+      {fixed_charges: [params.merge(id: parent_fixed_charge.id)]}
+    end
 
-    def find_or_create_fixed_charge_override(target_plan)
-      parent_fixed_charge = find_parent_fixed_charge
+    def find_or_create_fixed_charge_override(parent_fixed_charge, target_plan)
       existing_override = target_plan.fixed_charges.find_by(parent_id: parent_fixed_charge.id)
 
+      return existing_override.reload unless subscription_plan_parent_present
```
D1353L: `NoMethodError: undefined method 'reload' for nil` when passing `{ units: "-1" }` in params

tiagolupepic: good catch! I'll cover this scenario as well

tiagolupepic: Done in this commit https://github.com/getlago/lago-api/pull/5672/changes/8232af7494f0f7a408f6914c3fb2768700919c08. The issue is happening in [this service](https://github.com/getlago/lago-api/blob/main/app/services/plans/override_service.rb#L45) which is ignoring the validation and not raising an exception. What I did is to force the exception in this service instead.

D1353L (approved):
LGTM 👍

PR #5674, opened 2026-06-09
perf(subscriptions): Add GIN indexes for subscription text fields

vankiru:
## Context

Searching subscriptions with `search_term` can be pretty slow, especially for organizations with a large number of subscriptions. The `search_term` is currently used to search across three subscription fields: `id`, `external_id`, and `name`, on top of other customer and plan fields.

## Description

This PR adds three GIN indexes for those fields. The

## Benchmarks

Here are some benchmarks without index (before) and with index (after):

```sql
SELECT
	*
FROM
	subscriptions
WHERE
	subscriptions.organization_id = '11111111-2222-3333-4444-555555555555'
	AND subscriptions.id IN (
		SELECT
			id
		FROM
			subscriptions
		WHERE
			organization_id = '11111111-2222-3333-4444-555555555555'
			AND id::varchar ILIKE '%fffff396-e75a-4a09-91a0-6c7111536324%'
		UNION
		SELECT
			id
		FROM
			subscriptions
		WHERE
			organization_id = '11111111-2222-3333-4444-555555555555'
			AND name ILIKE '%fffff396-e75a-4a09-91a0-6c7111536324%'
		UNION
		SELECT
			id
		FROM
			subscriptions
		WHERE
			organization_id = '11111111-2222-3333-4444-555555555555'
			AND external_id ILIKE '%fffff396-e75a-4a09-91a0-6c7111536324%'
	)
```

Benchmarking was done with `pgbench` on a database containing **~1.2m** subscriptions. The benchmark used 10 clients and 5 threads, ran for 30 seconds against a warmed-up database. 

```
pgbench --protocol=prepared -c 10 -j 5 -T 30 -h localhost -U lago -d lago -n -f subscriptions.sql
```

Solution | TPS | Latency (ms)
-- | -- | --
without index (before) | 2.13 | 4693.26
with index (after) | 164.94 | 60.63

### DIFF
--- a/db/migrate/20260609133104_add_gin_indexes_for_subscription_fields.rb
+++ b/db/migrate/20260609133104_add_gin_indexes_for_subscription_fields.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddGinIndexesForSubscriptionFields < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :subscriptions, "organization_id, (id::varchar) gin_trgm_ops", using: :gin, algorithm: :concurrently, if_not_exists: true
+    add_index :subscriptions, "organization_id, name gin_trgm_ops", using: :gin, algorithm: :concurrently, if_not_exists: true
+    add_index :subscriptions, "organization_id, external_id gin_trgm_ops", using: :gin, algorithm: :concurrently, if_not_exists: true
+  end
+end
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -356,20 +356,23 @@ end
 #
 # Indexes
 #
-#  idx_on_organization_id_subscription_at_created_at_id        (organization_id,subscription_at DESC NULLS LAST,created_at DESC,id)
-#  index_subscriptions_on_billing_entity_id                    (billing_entity_id)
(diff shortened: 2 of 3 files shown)


db/migrate/20260609133104_add_gin_indexes_for_subscription_fields.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddGinIndexesForSubscriptionFields < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :subscriptions, "organization_id, (id::varchar) gin_trgm_ops", using: :gin, algorithm: :concurrently, if_not_exists: true
```
jdenquin, line 7: what about doing the same as invoices? checking if we have an UUID in the search with a regex? So we can just use the pkey index and avoid casting

jdenquin (approved):

db/migrate/20260609133104_add_gin_indexes_for_subscription_fields.rb
vankiru: Yep, I'll do the same for invoices once I'm done with the subscriptions

jdenquin: I already did it for invoices, it should be good!

PR #5681, opened 2026-06-09
[ING-309] fix(wallets): Enqueue wallet refresh on update

aquinofb:
## Context

We noticed slow pages on EU: wallet updates were taking more than 10s, up to ~62s. `Wallets::UpdateService` refreshes the customer wallets synchronously when a refresh-relevant attribute changes, and the refresh recomputes the usage of every active subscription, so the response time grows with the number of subscriptions of the customer.

## Changes

- Flag the customer as `awaiting_wallet_refresh` and enqueue `Customers::RefreshWalletJob` after commit, instead of calling `Customers::RefreshWalletsService` synchronously. Same pattern as the wallet increase flow (https://github.com/getlago/lago-api/pull/5631), without `wallet_ids` since these attributes can affect the allocation of all the wallets of the customer.
- The flag and the enqueue are done together inside the update transaction, so they are atomic: both happen on commit, or neither. It also keeps a single `needs_refresh?` evaluation, before `InvoiceCustomSections::AttachToResourceService` can save the wallet again and reset `saved_changes`.
- Move the `wallet.updated` webhook (now with `perform_after_commit`) and the invoice custom sections attach inside the transaction, to align with the create and terminate flows.

| | Before | Now |
|---|---|---|
| Wallet update request | Blocks until the refresh is done (up to ~62s) | Returns immediately |
| Wallets refresh | Synchronous, inside the request | `Customers::RefreshWalletJob`, after commit |
| `ongoing_balance` in the response | Up to date | Last known value, recomputed right after in background |

### DIFF
--- a/app/services/wallets/update_service.rb
+++ b/app/services/wallets/update_service.rb
@@ -68,7 +68,7 @@ module Wallets
           Customers::RefreshWalletJob.perform_after_commit(wallet.customer)
         end
 
-        InvoiceCustomSections::AttachToResourceService.call(resource: wallet, params:)
+        InvoiceCustomSections::AttachToResourceService.call!(resource: wallet, params:)
         SendWebhookJob.perform_after_commit("wallet.updated", wallet)
       end
 
--- a/app/services/wallets/update_service.rb
+++ b/app/services/wallets/update_service.rb
@@ -68,10 +68,9 @@ module Wallets
           Customers::RefreshWalletJob.perform_after_commit(wallet.customer)
         end
 
+        InvoiceCustomSections::AttachToResourceService.call(resource: wallet, params:)
         SendWebhookJob.perform_after_commit("wallet.updated", wallet)
       end
-
-      InvoiceCustomSections::AttachToResourceService.call(resource: wallet, params:)
 
       result.wallet = wallet
       result
--- a/app/services/wallets/update_service.rb
+++ b/app/services/wallets/update_service.rb
@@ -63,12 +63,14 @@ module Wallets
 
         update_metadata!
 
-        wallet.customer.flag_wallets_for_refresh if needs_refresh?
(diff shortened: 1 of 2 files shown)


toommz (approved):

mariohd (approved):

app/services/wallets/update_service.rb
```diff
@@ -62,11 +62,16 @@ def call
...
+
+        if needs_refresh?
+          wallet.customer.flag_wallets_for_refresh
+          Customers::RefreshWalletJob.perform_after_commit(wallet.customer)
+        end
+
+        SendWebhookJob.perform_after_commit("wallet.updated", wallet)
       end
 
       InvoiceCustomSections::AttachToResourceService.call(resource: wallet, params:)
```
groyoh: Can you move this into the transaction as well please ?

groyoh (approved):

app/services/wallets/update_service.rb
```diff
@@ -62,11 +62,15 @@ def call
...
 
-      InvoiceCustomSections::AttachToResourceService.call(resource: wallet, params:)
-      SendWebhookJob.perform_later("wallet.updated", wallet)
-      Customers::RefreshWalletsService.call(customer: wallet.customer) if needs_refresh?
+        if needs_refresh?
+          wallet.customer.flag_wallets_for_refresh
+          Customers::RefreshWalletJob.perform_after_commit(wallet.customer)
+        end
+
+        InvoiceCustomSections::AttachToResourceService.call(resource: wallet, params:)
```
groyoh:
```suggestion
        InvoiceCustomSections::AttachToResourceService.call!(resource: wallet, params:)
```

groyoh (approved):

PR #5682, opened 2026-06-10
[ING-310] fix(metadata): Reject non-string metadata values

groyoh:
## Context

`Metadata::ItemMetadata` is the shared metadata model for `CreditNote`, `Wallet`, and `Plan`. Its `value_correctness` validation gates the per-value string check on `val.present?`. Because `present?` is false for any blank value, `false`, empty arrays, and empty hashes skip the string check and are wrongly accepted, while `true`, `0`, numbers, and non-empty arrays/hashes are rejected (inconsistent). Through the REST API (`params.permit(metadata: {})`) this lets non-string JSON leaf values be persisted, which then pollute the `exports_*` warehouse views (they get surfaced as `"false"`/`"[]"`/`"{}"`). A DB CHECK constraint already blocks top-level non-object values, so the remaining gap is strictly these non-string leaf values.

## Description

This PR adds regression specs that assert a metadata leaf value of `false`, an empty array, or an empty hash is rejected for all three owner types. The specs are intentionally pushed while red so the failure is visible in CI before the fix lands. `nil` and empty-string leaf values stay valid.

The fix follows in a second commit, replacing the `present?` guard with an explicit type check (`val.nil? || (val.is_a?(String) && ...)`) so only `nil` and strings up to the length limit pass.

## Test plan

- [ ] `spec/models/metadata/item_metadata_spec.rb` fails on the new false/empty-array/empty-hash cases before the fix
- [ ] `spec/models/metadata/item_metadata_spec.rb` is green after the fix, with the existing nil and empty-string cases still valid
- [ ] `spec/db/views/exports_credit_notes_spec.rb` stays green

### DIFF
--- a/app/models/metadata/item_metadata.rb
+++ b/app/models/metadata/item_metadata.rb
@@ -31,7 +31,8 @@ module Metadata
           errors.add(:value, "key '#{key}' must be a String up to #{MAX_KEY_LENGTH} characters")
         end
 
-        unless val.nil? || (val.is_a?(String) && val.length <= MAX_VALUE_LENGTH)
+        valid_value = val.nil? || (val.is_a?(String) && val.length <= MAX_VALUE_LENGTH)
+        unless valid_value
           errors.add(:value, "value for key '#{key}' must be empty or a String up to #{MAX_VALUE_LENGTH} characters")
         end
       end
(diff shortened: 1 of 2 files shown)


toommz (approved):

groyoh:
Thanks @toommz !

PR #5683, opened 2026-06-10
feat: Finance assistant

stephenlago99:
Context

  The data team built a standalone Finance Agent service (text-to-SQL over analytics data) that answers natural-language questions about an organization's financial data via a POST /ask endpoint. To expose it in the Lago app, the front-end needs a first-party API to call — the agent service itself isn't reachable from
  the browser and shouldn't be.

  This PR adds the GraphQL endpoint that proxies questions to that service, scoped to the current organization. It's consumed by the companion front-end PR, which integrates the finance assistant as a second agent in the AI panel.

  Description

  New mutation: askFinanceAssistant

  - Input: question (required) and an optional sessionId to continue an existing agent session (conversation context lives in the agent service, keyed by session).
  - Returns a FinanceAssistantAnswer: explanation, results (markdown table), sessionId, sessionExpired, and the optional generated sqlQuery.
  - Gated like the existing AI conversations feature: authenticated user + current organization, ai_conversations:create permission, and a premium license (unauthorized otherwise).

  FinanceAssistant::AskService

  - Proxies the question to the agent service at LAGO_FINANCE_ASSISTANT_URL (POST {url}/ask) using the shared LagoHttpClient::Client (open timeout 5s, read timeout 60s), passing the org scope via an X-Organization-Id header. Returns a forbidden_failure (feature_unavailable) when the env var isn't configured, so the
  feature is cleanly off by default.
  - Validates the upstream response shape before returning it: a 2xx body that is missing any of the keys backing the type's non-nullable fields (or isn't a JSON object) becomes a finance_assistant_invalid_response service failure instead of leaking a GraphQL InvalidNullError to the client.
  - Error mapping: non-2xx responses surface the agent's detail message when present (finance_assistant_error); JSON parse failures and unexpected exceptions are mapped to service failures as well.

  Tests

  - Mutation spec (auth/org/permission shared examples, premium gating, happy path), type spec, and service specs covering the happy path, missing URL, upstream error with detail, malformed 2xx body, and empty 2xx body.

  Notes for reviewers

  - The HTTP call is synchronous in the resolver (up to ~65s worst case with the agent's LLM round-trip). This mirrors the proxy nature of the endpoint but is an outlier vs. AiConversations::Create (job + subscription); if usage grows, moving to the async pattern is the known follow-up.
  - Requires LAGO_FINANCE_ASSISTANT_URL to be set in the API environment for the feature to be active.


app/services/finance_assistant/ask_service.rb
```diff
@@ -0,0 +1,75 @@
...
+      end
+
+      result.answer = body
+      result
+    rescue LagoHttpClient::HttpError => e
+      result.service_failure!(
+        code: "finance_assistant_error",
+        message: e.json_message["detail"].presence || e.message,
+        error: e
+      )
```
ancorcruz, line 39: result_error (in ExecutionErrorResponder) has no when BaseService::ServiceFailure case, so both finance_assistant_error and finance_assistant_invalid_response hit the else branch → error: "Internal error", status: 500, and only the `code` survives. The detail we extract here (e.json_message["detail"]) never reaches the client, and an upstream 422 (e.g. "Question cannot be empty") surfaces as a 500.

Can we map upstream 4xx to a validation/unprocessable error so the message is conveyed — e.g. validation_failure! / single_validation_failure!, which result_error already forwards with `details`? Otherwise the detail extraction is effectively dead code from the frontend's perspective.

spec/graphql/mutations/finance_assistant/ask_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+        "results" => "| Month | MRR |",
+        "sql_query" => "select * from invoices",
+        "session_id" => session_id,
+        "session_expired" => false
+      }
+    end
+  end
+
+  before do
+    allow(FinanceAssistant::AskService).to receive(:call).and_return(service_result)
```
ancorcruz, line 51: The service is stubbed to always succeed here, so the `result_error(result)` branch of the resolver is never tested. A context stubbing a failure result and asserting the GraphQL error payload would have caught the 500/dropped-message issue above. Worth adding.

spec/services/finance_assistant/ask_service_spec.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe FinanceAssistant::AskService do
```
ancorcruz, line 5: A few branches aren't covered:
- Success **without** `sql_query` — it's nullable and deliberately excluded from REQUIRED_ANSWER_KEYS, but every success test includes it, so the "optional field" behavior is unverified.
- `session_id` **omitted** — request_body drops it when blank, but the spec always passes a UUID, so the new-conversation case (the nil branch) is never exercised.
- 2xx with non-JSON body → the rescue JSON::ParserError branch (the empty-body test becomes "{}" and parses fine, so it doesn't hit this).
- Network failure / timeout → the generic rescue => e → finance_assistant_error (e.g. .to_timeout), relevant given the 5s/60s timeouts.
- HttpError without a detail key → the || e.message fallback.

app/services/finance_assistant/ask_service.rb
```diff
@@ -0,0 +1,75 @@
...
+      {"X-Organization-Id" => organization.id}
+    end
+
+    def http_client
+      LagoHttpClient::Client.new(
+        "#{finance_assistant_url.chomp("/")}/ask",
+        open_timeout: 5,
+        read_timeout: 60
+      )
+    end
```
ancorcruz, line 70: Only `X-Organization-Id` is sent — no shared secret or token. Org isolation rests entirely on the agent service being network-internal and scoping all generated SQL to that header. Worth confirming that's enforced server-side; a shared auth header would harden it cheaply. (Likely out of scope for this PR, but flagging the trust boundary since the agent runs generated SQL.)

app/services/finance_assistant/ask_service.rb
```diff
@@ -0,0 +1,78 @@
...
+      @question = question
+      @session_id = session_id
+
+      super()
+    end
+
+    def call
+      return result.forbidden_failure! if finance_assistant_url.blank?
+
+      response = http_client.post_with_response(request_body, headers)
```
ancorcruz, line 22: This blocks a web thread for the full open_timeout(5) + read_timeout(60) on the agent's LLM round-trip. Fine under the Helm defaults (40 threads/pod) for normal use, but a slow/hung agent makes 60s holds pile up on the single pod — and the CPU-based HPA won't scale out on blocked I/O.

Before real traffic, could we add a cheap bound — a concurrency cap so these calls can't consume all threads, or a tighter read_timeout if the agent's p99 is well under 60s? Full async (the `AiConversations::Create` job + subscription pattern) stays the eventual fix.

ancorcruz (approved):
Nice, clean addition — service follows all the conventions and the gating mirrors `AiConversations::Create` well. A few things to consider, one of which I think is a real correctness issue:

1.  Synchronous up-to-65s call holds a web thread. Not a blocker given the Helm defaults (4 workers × 10 threads = 40/pod), but a slow or hung agent makes 60s read-holds pile up on the single pod, and the CPU-based HPA won't scale out on blocked I/O — so I'd like a concurrency cap (or a tighter read timeout) before this sees real traffic. Left an inline note on the call site.
2. Error details never reach the client. A `ServiceFailure` falls through to the generic 500/"Internal error" branch in result_error, so the agent's detail message we carefully extract is dropped and a user-caused upstream 422 becomes a 500 to the frontend. Details below.
3. Test coverage — the error path isn't exercised at the GraphQL level, and a few optional-field branches are untested. List below.
4. A couple of smaller notes (service-to-service auth, the hand-maintained required-keys list).

ancorcruz (changes requested):
clicked approve by mistake on previous review

app/services/finance_assistant/ask_service.rb
ancorcruz: After checking the agent (lago-data/agent): user-facing failures come back as 200 + explanation, and every non-2xx is an operational error (bad API key, auth infra, FastAPI validation list, 500s) we shouldn't surface verbatim. So service_failure! → 500 (detail logged server-side) is the right call — I've left the mapping as-is. Instead I added boundary validation for a blank question so that case returns a clean error without a pointless round-trip.

spec/graphql/mutations/finance_assistant/ask_spec.rb
ancorcruz: Added — see the when the service fails context asserting a GraphQL error + null answer

spec/services/finance_assistant/ask_service_spec.rb
ancorcruz: All added: success without sql_query, session_id omitted, non-JSON 2xx body, timeout, and HttpError without a detail.

app/services/finance_assistant/ask_service.rb
ancorcruz: Addressed — requests now authenticate with X-LAGO-API-KEY, and the agent resolves the org from that key server-side (confirmed in lago-data/agent). Dropped the unused X-Organization-Id

app/services/finance_assistant/ask_service.rb
ancorcruz: Deferring this. Per Vlad, complex questions legitimately take 30–45s, so we can't tighten the timeout. A concurrency cap is disproportionate for now; the proper fix is the async job + subscription pattern (like AiConversations), which changes the FE contract. I wouldn't block this PR because of this.

ancorcruz (approved):
Approving ✅ — error handling matches the agent's real contract, blank-question validation avoids a wasted round-trip, auth + org scoping is via X-LAGO-API-KEY, and the new tests cover the gaps.

One to watch post-merge (not a blocker): `askFinanceAssistant` blocks a web thread for the whole agent round-trip (30–45s typical, 65s ceiling). Let's monitor /ask latency and in-flight concurrency once it's in real use — if it bites, the fix is the async job + subscription pattern (like AiConversations).

PR #5685, opened 2026-06-10
fix(integration-customers): Improve sanitization logic of integration customers

osmarluz:
## Context

Re-sending a customer's `integration_customers` block on `POST /api/v1/customers` could duplicate provider contacts. Stale-link removal was keyed only on the `id`, so a payload whose `id` didn't match the stored link — e.g. the all-zeros placeholder the Go client sent for an unset id — destroyed the existing link and immediately recreated it, adding a duplicate NetSuite/Anrok contact on every repeat.

## Description

`sanitize_integration_customers` now trusts a payload `id` only when it belongs to one of the customer's links; otherwise it resolves the existing link by integration type, so a missing or wrong `id` no longer orphans it. An entry pointing the same type at a different connection is still treated as a switch and removed. Matching avoids `external_customer_id`, which Anrok links lack until their first invoice sync.

### DIFF
--- a/app/services/integration_customers/create_or_update_batch_service.rb
+++ b/app/services/integration_customers/create_or_update_batch_service.rb
@@ -80,10 +80,47 @@ module IntegrationCustomers
     end
 
     def sanitize_integration_customers
-      updated_int_customers = integration_customers.reject { |m| m[:id].nil? }.map { |m| m[:id] }
-      not_needed_ids = customer.integration_customers.pluck(:id) - updated_int_customers
+      existing_ids = customer.integration_customers.ids
 
-      customer.integration_customers.where(id: not_needed_ids).destroy_all
+      kept_ids = integration_customers.filter_map do |params|
+        id = params[:id].presence
+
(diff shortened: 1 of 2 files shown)


app/services/integration_customers/create_or_update_batch_service.rb
```diff
@@ -80,10 +80,47 @@ def handle_update
...
+      else
+        customer.integration_customers.where.not(id: kept_ids).destroy_all
+      end
+    end
+
+    def existing_integration_customer_id(params)
+      return if params[:integration_type].blank?
+
+      integration_customer_type = IntegrationCustomers::BaseCustomer.customer_type(params[:integration_type])
+      integration_customer = customer.integration_customers.find_by(type: integration_customer_type)
```
lovrocolic, line 106: We are currently working on a feature where we will enable multiple integrations of the same type per customer. I can see conflict between this approach and the new feature.

osmarluz:
The existing logic exists for other resources as well, so for now it will be kept

PR #5686, opened 2026-06-10
[ING-252] fix(cache): expire on metric deletion

mariohd:
## Context

After a billable metric was deleted, recreated with the same code, and the plan's charge re-added — all before the async event-deletion job ran — `current_usage` kept serving the pre-deletion event count for the recreated charge. The cached value survived the eventual event purge, and the only way to clear it was manually resetting the subscription's cache.

**Closes ING-252.**

## Fix

`Events::DeleteForMetricService` now invalidates the charge-usage cache for each affected subscription after its events are deleted (both Postgres and Clickhouse paths). The async tail is the only safe hook — the stale value is written *between* BM deletion and event deletion, so invalidating at sync BM-delete time would be too early.

The orchestration lives in `Subscriptions::ChargeCacheService` as a new bulk class method:

```ruby
def self.expire_for_subscriptions(subscription_ids)
  Subscription
    .where(id: subscription_ids)
    .preload(plan: {charges: :filters})
    .find_each do |subscription|
      subscription.plan.charges.each do |charge|
        expire_for_subscription_charge(subscription:, charge:)
      end
    end
end

def self.expire_for_subscription(subscription)
  expire_for_subscriptions([subscription.id])
end
```

`expire_for_subscription` (singular) now funnels through the bulk path, so every existing caller (events post-process, clickhouse dup cleanup, customer-portal usage setup) picks up the same eager-loaded path for free.

`DeleteForMetricService` is just a delegator:

```ruby
def expire_charge_caches(subscription_ids)
  Subscriptions::ChargeCacheService.expire_for_subscriptions(subscription_ids)
end
```

## Why the eager loading matters

The previous shape of `expire_for_subscription` called `subscription.plan.charges.includes(:filters).find_each` for each subscription — 3 queries each (plan, charges, filters). For a BM deletion affecting N subscriptions that meant ~3N queries on top of the actual event deletion. The new `expire_for_subscriptions` preloads `plan: {charges: :filters}` once per `find_each` batch, so the cost is **4 DB queries per batch regardless of N**.

## Tests

`spec/services/events/delete_for_metric_service_spec.rb` — two new contexts under `with charge-usage cache invalidation`:

- **Spy** — asserts `Subscriptions::ChargeCacheService.expire_for_subscription_charge` is called for the affected (subscription, charge).
- **End-to-end (`cache: :memory`)** — pre-writes a value to the real cache key, runs the service, asserts the key is gone. Proves the round-trip end to end through `Rails.cache`, not just behaviour-stubs it.

`spec/services/subscriptions/charge_cache_service_spec.rb` — two new describe blocks:

- `.expire_for_subscriptions` — coverage across two subscriptions × two charges, plus a **query-budget assertion** (≤ 6 SQL queries via `ActiveSupport::Notifications`) that acts as a regression guard against a future refactor reintroducing the per-subscription re-query.
- `.expire_for_subscription` — delegation check confirming the singular method now funnels through the bulk path.

```
  spec/services/events/delete_for_metric_service_spec.rb \
  spec/services/subscriptions/charge_cache_service_spec.rb
# 17 examples, 0 failures
```

### DIFF
--- a/app/services/events/delete_for_metric_service.rb
+++ b/app/services/events/delete_for_metric_service.rb
@@ -83,9 +83,7 @@ module Events
     end
 
     def expire_charge_caches(subscription_ids)
-      Subscription.where(id: subscription_ids).find_each do |subscription|
-        Subscriptions::ChargeCacheService.expire_for_subscription(subscription)
-      end
+      Subscriptions::ChargeCacheService.expire_for_subscriptions(subscription_ids)
     end
 
     def async_delete_clickhouse(table, date_field, external_subscription_ids)
--- a/app/services/events/delete_for_metric_service.rb
+++ b/app/services/events/delete_for_metric_service.rb
@@ -26,6 +26,7 @@ module Events
 
         delete_postgres_events(subscription_ids, external_subscription_ids)
         delete_clickhouse_events(external_subscription_ids) if clickhouse_enabled
+        expire_charge_caches(subscription_ids)
       end
 
       result
@@ -78,6 +79,12 @@ module Events
     def delete_clickhouse_events(external_subscription_ids)
       CLICKHOUSE_TABLES.each do |table, date_field|
         async_delete_clickhouse(table, date_field, external_subscription_ids)
+      end
+    end
+
+    def expire_charge_caches(subscription_ids)
+      Subscription.where(id: subscription_ids).find_each do |subscription|
+        Subscriptions::ChargeCacheService.expire_for_subscription(subscription)
       end
     end
 
--- a/spec/services/subscriptions/charge_cache_service_spec.rb
+++ b/spec/services/subscriptions/charge_cache_service_spec.rb
@@ -58,7 +58,7 @@ RSpec.describe Subscriptions::ChargeCacheService do
       end
     end
 
-    it "loads plan, charges and filters in a constant number of queries", :aggregate_failures do
+    it "loads plan, charges and filters in a constant number of queries" do
       ids = [subscription.id, other_subscription.id]
 
       query_count = 0
--- a/spec/services/subscriptions/charge_cache_service_spec.rb
+++ b/spec/services/subscriptions/charge_cache_service_spec.rb
@@ -34,4 +34,53 @@ RSpec.describe Subscriptions::ChargeCacheService do
       expect(Rails.cache).to have_received(:delete).with(cache_service.cache_key)
     end
   end
+
+  describe ".expire_for_subscriptions" do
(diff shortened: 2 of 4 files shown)


app/services/events/delete_for_metric_service.rb
```diff
@@ -26,6 +26,7 @@ def call
 
         delete_postgres_events(subscription_ids, external_subscription_ids)
         delete_clickhouse_events(external_subscription_ids) if clickhouse_enabled
+        expire_charge_caches(subscription_ids)
```
toommz, line 29: I'm afraid adding this here will introduce a serious N+1 problem 🤔 The `Subscriptions::ChargeCacheService` is called for each sub, there can be a lot, and inside the service we do that:

```ruby
subscription.plan.charges.includes(:filters).find_each
```

mariohd: Good catch! 
Yes, totally valid point.

mariohd: Solved

vincent-pochet:
Not sure to understand the issue here. When we delete a billable metric, the related charges are soft deleted as well.
If you re-use the code in a BM and creates charges relying on this metric, the charge will be a new charge with a different id, since the cache key is using this charge_id, I'm not sure to get how it could conflict with previous charges?

mariohd:
@vincent-pochet you're totally right about that. 

What this PR does is just invalidate the cache, so the "old cached usage" gets removed.
This originally came from an issue that a client deleted a billable metric and recreated a new one with the same code, since the cache were considering this old metric, it appeared there in the subscription usage page, even if the new metric hasn't been used yet. They were complaining about that.

vincent-pochet:
@mariohd 
ohh ok, it makes sense. Thanks for clarifying!

spec/services/subscriptions/charge_cache_service_spec.rb
```diff
@@ -34,4 +34,53 @@
...
+
+      [subscription, other_subscription].each do |sub|
+        [charge, other_charge].each do |ch|
+          expect(described_class).to have_received(:expire_for_subscription_charge)
+            .with(subscription: having_attributes(id: sub.id), charge: having_attributes(id: ch.id))
+        end
+      end
+    end
+
+    it "loads plan, charges and filters in a constant number of queries", :aggregate_failures do
```
vincent-pochet:
```suggestion
    it "loads plan, charges and filters in a constant number of queries" do
```

This is now the default behavior

vincent-pochet (approved):

toommz (approved):

PR #5688, opened 2026-06-10
feat(payment-gated-subs): Handle billing of pay-in-advance fixed-charges emitted during the subscription incomplete period

osmarluz:
## Context

Subscriptions created with `activate_on_payment` stay `incomplete` until payment succeeds. While incomplete, pay-in-advance fixed-charge units can change (e.g. a plan update with `apply_units_immediately`): the resulting `FixedChargeEvent`s are emitted for the incomplete subscription but never billed, because immediate billing only runs for active subscriptions. On activation those deltas would be silently skipped.

## Description

Adds `Subscriptions::ActivationRules::BillFixedChargesDeltaService` and its job, enqueued from `Subscriptions::ActivateService` when a payment-gated subscription transitions from `incomplete` to `active`. The service collects the distinct pay-in-advance `FixedChargeEvent` timestamps from the incomplete window — excluding the creation event at `started_at + 1.second`, already covered by the gated invoice — and bills each one through `Invoices::CreatePayInAdvanceFixedChargesService`, producing one delta invoice per mid-incomplete plan update.

Timestamps are processed sequentially within the job, since each delta is computed against the fees already committed for the period, and already-billed timestamps are skipped so retries don't produce duplicate zero-amount invoices. The window is capped at the end of the gated period: events stamped on later periods belong to the periodic billing flows (cross-period catch-up lands separately with `BillMissedPeriodsService`).


app/services/subscriptions/activation_rules/bill_fixed_charges_delta_service.rb
```diff
@@ -0,0 +1,69 @@
...
+          next if already_billed?(timestamp)
+
+          invoice_result = Invoices::CreatePayInAdvanceFixedChargesService.call(
+            subscription:,
+            timestamp: timestamp.to_i
+          )
+          next if invoice_result.success? || tax_error?(invoice_result)
+
+          invoice_result.raise_if_error!
+        end
```
ancorcruz, line 26: This bills each fixed charge individually, then give N fixed charges to be billed, N invoices with their N payments will be issued.

from UX perspective this is terrible, however as this is an edge case we could improve it later.

app/services/subscriptions/activation_rules/bill_fixed_charges_delta_service.rb
```diff
@@ -0,0 +1,69 @@
...
+        return result unless subscription.fixed_charges.pay_in_advance.any?
+
+        delta_event_timestamps.each do |timestamp|
+          next if already_billed?(timestamp)
+
+          invoice_result = Invoices::CreatePayInAdvanceFixedChargesService.call(
+            subscription:,
+            timestamp: timestamp.to_i
+          )
+          next if invoice_result.success? || tax_error?(invoice_result)
```
ancorcruz, line 23: if there is tax error how the delta will be billed?

app/services/subscriptions/activation_rules/bill_fixed_charges_delta_service.rb
```diff
@@ -0,0 +1,69 @@
...
+      def initialize(subscription:)
+        @subscription = subscription
+        super
+      end
+
+      def call
+        return result unless subscription.fixed_charges.pay_in_advance.any?
+
+        delta_event_timestamps.each do |timestamp|
+          next if already_billed?(timestamp)
```
ancorcruz: this is a small N+1 query.... perhaps not a big deal as this is an edge. Anyway we can improve it fetching the already billed timestamps then check the the delta event timestamp is included in the collection instead of doing one query to DB for each timestamp.

ancorcruz (approved):
Looks good, left some comments.

app/services/subscriptions/activation_rules/bill_fixed_charges_delta_service.rb
osmarluz: @ancorcruz this follows the same pattern we have for `Invoices::CreatePayInAdvanceFixedChargesJob`; after a tax error happens and the invoices is marked with `tax_status` equal to `pending` we have two possible paths to finalize the invoice:

- Provider tax: `PullTaxesAndApplyJob`.
- VIES: the customer-level `ViesCheckJob`, on completion, sweeps all pending/open + tax-pending invoices and enqueues `FinalizePendingViesInvoiceJob`.

app/services/subscriptions/activation_rules/bill_fixed_charges_delta_service.rb
osmarluz: @ancorcruz good catch, I'll fix this

app/services/subscriptions/activation_rules/bill_fixed_charges_delta_service.rb
osmarluz: @ancorcruz yes, we can introduce changes to `Invoices::CreatePayInAdvanceFixedChargesService` to group them as an improvement

app/services/subscriptions/activation_rules/bill_fixed_charges_delta_service.rb
ancorcruz: there is a subtle difference, the invoices are not visible here (status: open)... if a tax error requires human intervention invoice will be blocked forever.

osmarluz: @ancorcruz when we enter this service the subscription was already marked as active and committed, so the invoices will be visible. We can monitor anyway to see how this behaves

PR #5690, opened 2026-06-10
feat(products): catalog guards, dedup and batching

rsempe:
## Context

Hardening on top of the catalog services (#6098) and GraphQL surface (#6099): the immutability rules, the review fixes and the list-performance work.

## Description

- Gate structural edits on attachment: a product, category or filter priced by a plan or subscription cannot change identity or be deleted (`attached_to_plan_or_subscription`); filter values freeze once a subscription bills through the filter (`attached_to_subscriptions`).
- Creating a filter on a fixed product fails with `not_allowed_for_product_type`.
- Reject duplicate filter value sets per product (`value_already_exist` on `values`) — two identical slices would be unresolvable for the billing engine; partial overlaps stay legal.
- Batch `filters_count`, `products_count` and `attached_to_plan_or_subscription` through dataloader sources — one grouped query per page instead of per row.

### DIFF
--- a/app/services/product_filters/update_service.rb
+++ b/app/services/product_filters/update_service.rb
@@ -22,26 +22,26 @@ module ProductFilters
         return resolved_values unless resolved_values.success?
       end
 
-      # NOTE: the code freezes as soon as the item is in a plan or subscription
-      if product_filter.attached_to_plan_or_subscription? &&
-          params.key?(:code) && params[:code]&.strip != product_filter.code
-        return result.single_validation_failure!(field: :code, error_code: "attached_to_plan_or_subscription")
-      end
+      product_filter.product.with_lock do
+        # NOTE: the code freezes as soon as the item is in a plan or subscription
+        if product_filter.attached_to_plan_or_subscription? &&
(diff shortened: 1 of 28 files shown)


app/services/product_filters/update_service.rb
```diff
@@ -20,10 +20,23 @@ def call
...
+          params.key?(:code) && params[:code]&.strip != product_filter.code
+        return result.single_validation_failure!(field: :code, error_code: "attached_to_plan_or_subscription")
+      end
+
+      if product_filter.attached_to_subscriptions? && params.key?(:values) && values_changed?
+        return result.single_validation_failure!(field: :values, error_code: "attached_to_subscriptions")
+      end
 
+      if params.key?(:values)
         values_validation = ProductFilters::ValidateValuesService.call(
```
endenis: I wonder if this check and the update should be done within a lock on product or if it is not really needed 🤔

endenis (approved):

app/services/product_filters/update_service.rb
rsempe: adding it you're right 👍

PR #5691, opened 2026-06-10
feat(products): add catalog REST endpoints

rsempe:
## Context

Public REST surface for catalog authoring, mirroring the GraphQL layer from #5690 and sharing the same services.

## Description

- Add REST endpoints for product categories, products and product filters (index, show, create, update, destroy), addressed by `code`.
- Support `search_term` on all three index endpoints.
- Expose `products_count` on the product category payload.
- Accept `item_type` as a single value rather than an array.
- Apply the same attachment gating as GraphQL, so a priced product cannot be structurally edited through the API.
- Move the catalog endpoints under `/api/v2`.

### DIFF
--- a/app/controllers/api/v2/products_controller.rb
+++ b/app/controllers/api/v2/products_controller.rb
@@ -3,6 +3,8 @@
 module Api
   module V2
     class ProductsController < Api::BaseController
+      ERROR_FIELDS = {billable_metric: :billable_metric_code, product_category: :product_category_code}.freeze
+
       def create
         if input_params[:product_category_code].present? && product_category.nil?
           return not_found_error(resource: "product_category")
@@ -95,7 +97,8 @@ module Api
         if result.success?
           render(
             json: ::CollectionSerializer.new(
-              result.products,
+              # Preloaded so filters_count reads the loaded association.
+              result.products.includes(:filters),
               ::V1::ProductSerializer,
               collection_name: "products",
               meta: pagination_metadata(result.products)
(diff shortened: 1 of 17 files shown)


app/controllers/api/v2/products_controller.rb
```diff
@@ -0,0 +1,170 @@
...
+          render_product(result.product)
+        else
+          render_item_error(result)
+        end
+      end
+
+      def update
+        product = current_organization.products.find_by(code: params[:code])
+
+        if update_params.key?(:product_category_code) && updated_product_category.nil?
```
endenis: Is there a way to detach a product from product category?
With graphql I can pass `product_category_id: null` but I don't see a way with this endpoint.

rsempe: True - fixed 👍

config/routes.rb
```diff
@@ -206,6 +206,13 @@
         get :json_public_key, on: :collection
       end
     end
+
+    namespace :v2 do
+      resources :products, param: :code, code: /.*/, only: %i[index show create update destroy] do
+        resources :filters, param: :code, code: /.*/, only: %i[index show create update destroy], controller: "products/filters"
```
endenis, line 212: No need to fix it now but a question for the future: do we plan to have any validation on characters acceptable to use in `code`? 

If product code is named `something/filters`, calling GET `products/{code}` would be => GET `products/something/filters`

endenis (approved):

config/routes.rb
rsempe: I totally agree 👍 
We don't have any validations right now for `code` but we can impose a format rule for the product catalog v2 
Creating a ticket (BIL-531) for this 👍

PR #5692, opened 2026-06-11
[ING-272] fix(wallets): Make wallet balance decrease refresh async

aquinofb:
## Context

When a wallet balance decreases (typically when voiding a wallet transaction), we refresh the customer's wallets synchronously, recomputing current usage for every active subscription before the request can finish. In the void path this even runs while holding the customer prepaid credit lock. The top-up path moved to an async refresh in #5631; this is the decrease counterpart ([ING-272](https://linear.app/getlago/issue/ING-272/make-wallet-balance-decrease-refresh-async)).

## Description

`Wallets::Balance::DecreaseService` now flags the customer for refresh and enqueues `Customers::RefreshWalletJob` after commit instead of calling `Customers::RefreshWalletsService` inline. `skip_refresh` keeps working and now skips both the flag and the enqueue.

The `include_generating_invoices` flag is dropped entirely: it was only needed when refreshing while applying credits to a generating progressive billing invoice, and that path runs its own refresh with `skip_refresh` since 05587e465. The job signature is untouched.

|  | Before | After |
|---|---|---|
| Refresh | Inline, in the request, inside the prepaid credit lock | Async on the job queue, after commit |
| Direct balance (`balance_cents`) | Updated inline | Updated inline (unchanged) |
| Ongoing balance | Up to date when the request returns | Eventually consistent, same as top-ups since #5631 |

## How to try locally

1. Create a customer with an active subscription, some usage, and a wallet with credits.
2. Void part of the wallet balance: `POST /api/v1/wallet_transactions` with `voided_credits`.
3. The request returns without recomputing usage; a `Customers::RefreshWalletJob` is enqueued and the ongoing balance updates once it runs.

### DIFF
--- a/app/services/wallets/balance/decrease_service.rb
+++ b/app/services/wallets/balance/decrease_service.rb
@@ -27,7 +27,7 @@ module Wallets
 
         unless skip_refresh
           wallet.customer.flag_wallets_for_refresh
-          Customers::RefreshWalletJob.perform_after_commit(wallet.customer, include_generating_invoices: true)
+          Customers::RefreshWalletJob.perform_after_commit(wallet.customer)
         end
 
         SendWebhookJob.perform_after_commit("wallet.updated", wallet)
--- a/app/services/wallets/balance/decrease_service.rb
+++ b/app/services/wallets/balance/decrease_service.rb
@@ -26,16 +26,12 @@ module Wallets
         )
 
         unless skip_refresh
-          Customers::RefreshWalletsService.call(
-            customer: wallet.customer,
-            include_generating_invoices: true
-          )
+          wallet.customer.flag_wallets_for_refresh
+          Customers::RefreshWalletJob.perform_after_commit(wallet.customer, include_generating_invoices: true)
         end
 
-        after_commit do
-          SendWebhookJob.perform_later("wallet.updated", wallet)
-          UsageMonitoring::ProcessWalletAlertsJob.perform_later(wallet)
-        end
+        SendWebhookJob.perform_after_commit("wallet.updated", wallet)
+        UsageMonitoring::ProcessWalletAlertsJob.perform_after_commit(wallet)
 
         result.wallet = wallet
         result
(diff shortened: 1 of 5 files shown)


vincent-pochet (approved):

app/services/wallets/balance/decrease_service.rb
```diff
@@ -26,16 +26,12 @@ def call
         )
 
         unless skip_refresh
-          Customers::RefreshWalletsService.call(
-            customer: wallet.customer,
-            include_generating_invoices: true
-          )
+          wallet.customer.flag_wallets_for_refresh
+          Customers::RefreshWalletJob.perform_after_commit(wallet.customer, include_generating_invoices: true)
```
groyoh: I believe we can remove `include_generating_invoices`. I discussed with @annvelents and this is actually no longer needed since https://github.com/getlago/lago-api/commit/05587e465513db63346ca35db39a201d271b22f2 as it is only needed when applying credits to `generating` progressive billing invoices.

groyoh (changes requested):
Also something to keep in mind in the future, when updating a worker with new parameters, we should ensure that we release the worker changes first. Otherwise, if the API gets deployed before the workers, the worker may receive parameter it cannot handle and will fail with `ArgumentError`.

app/services/wallets/balance/decrease_service.rb
aquinofb: That's true! For some reason, I relaxed and assumed the default value would prevent the new worker signature problem. Should have double-checked and tested it. On it.

aquinofb: ...and **great catch**! Because of this change to drop the flag, we no longer need to change the job signature 🤝

spec/services/wallet_transactions/create_from_params_service_spec.rb
```diff
@@ -73,11 +73,9 @@
       expect(wallet.reload.credits_balance).to eq(22.0)
     end
 
-    it "updates wallet ongoing balance based on granted and voided credits" do
-      subject
-
-      expect(wallet.reload.ongoing_balance_cents).to eq(2200)
-      expect(wallet.reload.credits_ongoing_balance).to eq(22.0)
+    it "enqueues a RefreshWalletJob to update the ongoing balance" do
+      expect { subject }
```
groyoh, line 81: We should assert the `awaiting_wallet_refresh` update.

groyoh (approved):

toommz (approved):

mariohd (approved):

PR #5693, opened 2026-06-11
feat(fc-units-override): units-only write branch on the dedicated endpoint

ancorcruz:
## Context

The subscription-scoped fixed_charge endpoint
`PUT /api/v1/subscriptions/:external_id/fixed_charges/:code` currently always clones the plan and creates a fixed_charge override even when the only change is the unit count. For seat-based plans where every customer signs up with a different seat count and adjusts it over time, this produces one duplicate Plan + one duplicate FixedCharge per customer per change. The override table introduced earlier exists to absorb these per-subscription unit edits without cloning anything; this commit wires the dedicated endpoint to that fast path.

## Description

`Subscriptions::UpdateOrOverrideFixedChargeService` now includes the new params-shape concern and branches inside one transaction:

- When the incoming params match the units-only shape (carrying only `units` and optionally `apply_units_immediately`) AND the subscription is not already on a cloned plan, the service writes a `Subscription::FixedChargeUnitsOverride` row for the (subscription, parent fixed_charge) pair and emits a FixedChargeEvent through `FixedCharges::EmitEventsService`. The override row is created or updated via `find_or_initialize_by`, so repeated edits update one row instead of stacking.
- When `apply_units_immediately` is set on a pay-in-advance fixed charge, `Invoices::CreatePayInAdvanceFixedChargesJob` is dispatched in an `after_commit` block so the mid-period delta invoice is generated after the override row is committed.
- Subscriptions already on a cloned plan, or requests carrying any non-units field (properties, tax_codes, invoice_display_name, etc.), fall through to the existing plan-cloning path unchanged.

The params-shape predicate lives in the shared
`FixedChargeUnitsOverrideDetectionConcern`; the cloned-plan guard stays in the service since it inspects subscription state.

Specs cover the new branch at the service level (override created / updated, no plan or fixed_charge clone, event emitted with override units, billing job dispatched for pay-in-advance + apply-immediately, no dispatch for pay-in-arrears, fall-through on cloned-plan subs). A scenario spec exercises the full HTTP path end-to-end and asserts the same outcomes against the real job pipeline.

### DIFF
--- a/app/services/subscriptions/update_or_override_fixed_charge_service.rb
+++ b/app/services/subscriptions/update_or_override_fixed_charge_service.rb
@@ -62,7 +62,7 @@ module Subscriptions
         timestamp:
       )
 
-      if apply_units_immediately && parent_fixed_charge.pay_in_advance?
+      if apply_units_immediately && parent_fixed_charge.pay_in_advance? && subscription.active?
         after_commit do
           Invoices::CreatePayInAdvanceFixedChargesJob.perform_later(subscription, timestamp)
         end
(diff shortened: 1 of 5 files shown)


ancorcruz:

PR #5694, opened 2026-06-11
misc(daily_usage): Improve fill_history task

vincent-pochet:
## Description

This PR improves the `daily_usages:fill_history` rake task:
- It makes it more interactive by asking the user for both the organization and the lower date boundary
- It adds more visual feedback, by displaying the organization ID and name, the number of affected subscription and the number of deleted DailyUsages
- It cleanup the existing DailyUsage before re-enqueing the `DailyUsages::FillHistoryJob`

### DIFF
--- a/lib/tasks/daily_usages.rake
+++ b/lib/tasks/daily_usages.rake
@@ -1,14 +1,19 @@
 # frozen_string_literal: true
 
+require_relative "../task_prompt"
+
+# rubocop:disable Rails/Output,Rails/Exit
 namespace :daily_usages do
   desc "Fill past daily usage"
-  task :fill_history, [:organization_id, :days_ago] => :environment do |_task, args|
-    abort "Missing organization_id\n\n" unless args[:organization_id]
-
+  task fill_history: :environment do
     Rails.logger.level = Logger::INFO
 
-    from_date = (args[:days_ago] || DailyUsage::DEFAULT_HISTORY_DAYS).days.ago.to_date
-    organization = Organization.find(args[:organization_id])
+    deletion_batch_size = 10_000
+
+    organization = TaskPrompt.ask_for_organization
+
+    default_from_date = DailyUsage::DEFAULT_HISTORY_DAYS.days.ago.to_date
+    from_date = TaskPrompt.ask_for_date("From date (YYYY-MM-DD)", default: default_from_date)
 
     subscriptions = organization.subscriptions
       .where(status: [:active, :terminated])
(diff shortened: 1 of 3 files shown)


toommz (approved):
LGTM, just one thing: the PR body only talks about some small changes, but the PR also introduces the deletion of daily usages, so could be great to explain why we do that now :)

lib/tasks/daily_usages.rake
```diff
@@ -1,23 +1,81 @@
...
+
+    # ----- delete existing daily_usages in batches -----
+    puts ""
+    puts "Deleting existing daily_usages..."
+    delete_start = Time.current
+    deleted_total = 0
+
+    DailyUsage
+      .where(organization_id: organization.id)
+      .where("usage_date >= ?", from_date)
```
groyoh: The deletion removes every daily_usage with `usage_date >= from_date` and no upper bound, but the enqueued `FillHistoryJob` rebuilds only up to yesterday (the service's `to` defaults to `Time.zone.yesterday`). A row dated today (or any future date) gets deleted here and never regenerated by the backfill. Bounding the deletion to the same window the job rebuilds, e.g. `.where("usage_date < ?", Date.current)`, keeps the two in sync.

lib/tasks/daily_usages.rake
```diff
@@ -1,23 +1,81 @@
...
       .where("terminated_at IS NULL OR terminated_at >= ?", from_date)
       .includes(customer: :organization)
 
+    # ----- recon (read-only) -----
+    sub_count = subscriptions.count
+    existing_rows = DailyUsage
+      .where(organization_id: organization.id)
+      .where("usage_date >= ?", from_date)
+      .where(subscription_id: subscriptions.select(:id))
+      .count
```
groyoh: This `DailyUsage.where(organization_id:).where("usage_date >= ?", from_date).where(subscription_id: subscriptions.select(:id))` chain is repeated verbatim in the deletion block below (lines 58-61). If one side is edited later, the "existing daily_usages (will be deleted)" recon count silently stops matching what the delete actually removes. Extracting it once, e.g. `usages_to_delete = DailyUsage.where(...)`, then using `usages_to_delete.count` for recon and `usages_to_delete.in_batches(...)` for the delete keeps the reported count and the deletion in lockstep.

lib/tasks/daily_usages.rake
vincent-pochet: @groyoh 
The usage is always computed one day after the stored `usage_date`, to consider the full day.
It means that the usage for `Date.current` will be computed tomorrow. From this perspective, it is expected to remove all records, the task will recompute everything up to yesterday, and the usage for the current will be computed automatically the next day.

Also, partially removing daily usage would potentially make the `usage_diff` invalid

lib/tasks/daily_usages.rake
vincent-pochet: @groyoh good catch indeed! It's fixes

PR #5695, opened 2026-06-11
feat(fc-units-override): units-only write branch on subscription creation

ancorcruz:
## Context

Subscription creation is the highest-volume code path that today overrides the plan and every fixed charge whenever `plan_overrides` is present, even if the only change is the unit count. For seat-based plans this produces one overridden Plan and N overridden FixedCharges per signup, which is unworkable at the scale Mistral is preparing for (~2000 subscriptions, every one starting with its own seat count). This commit routes units-only `plan_overrides.fixed_charges` requests on `POST /api/v1/subscriptions` through the override table instead.

## Description

`Subscriptions::CreateService` now includes the shared params-shape concern and branches inside `create_subscription`:

- When `plan_overrides` matches the units-only envelope shape (only the `fixed_charges` key, every entry carrying just `id`, `units`, and optionally `apply_units_immediately`) AND the target plan is not itself an override, the subscription is built on the parent plan (no plan override created), persisted early with status `pending` so it has an id, and one `Subscription::FixedChargeUnitsOverride` row is written per entry against the parent fixed_charge. The existing lifecycle paths (today / past / future) then run as normal and the events they emit pick up the override units through `effective_units_for`.
- Any other shape (mixed-field `plan_overrides`, non-units entry fields, target plan already an override) falls through to the existing `Plans::OverrideService` path unchanged.

`apply_units_immediately` is allowed by the predicate but intentionally not dispatched on creation. Unlike the mid-period delta case on the dedicated endpoint, the first invoice on a brand-new subscription already bills the override units through the standard activation flow, so adding `Invoices::CreatePayInAdvanceFixedChargesJob` here would either no-op or double-bill.

Specs replace the prior `plan_overrides` tests that documented the old plan-override behaviour on a units-only payload with assertions for the new branch (subscription on parent plan, override row written, events with override units), and add a context that re-establishes parity for the plan-override fall-through path (non-units-only payload).

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -3,6 +3,7 @@
 module Subscriptions
   class CreateService < BaseService
     include Subscriptions::Concerns::BillingEntityResolutionConcern
+    include Subscriptions::Concerns::FixedChargeUnitsOverrideDetectionConcern
 
     Result = BaseResult[:subscription, :payment_method]
 
@@ -131,7 +132,7 @@ module Subscriptions
       new_subscription = Subscription.new(
         organization_id: customer.organization_id,
         customer:,
-        plan: params.key?(:plan_overrides) ? override_plan(plan) : plan,
+        plan: target_plan_for_new_subscription,
         subscription_at:,
         name:,
         external_id:,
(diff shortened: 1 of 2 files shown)


ancorcruz:

PR #5696, opened 2026-06-11
feat(fc-units-override): units-only write branch on subscription update

ancorcruz:
## Context

`PUT /api/v1/subscriptions/:id` currently routes any `plan_overrides` through `Plans::OverrideService` (creating a new override plan) or `Plans::UpdateService` (mutating an existing override). When the only change is a units bump on existing fixed charges of a subscription that's still on the parent plan, neither operation is desirable: the plan-clone path duplicates the plan + every fixed_charge per subscription, and the existing-override path doesn't apply because the subscription isn't on an override plan yet. The override table absorbs these requests with one row per change.

## Description

`Subscriptions::UpdateService` now includes the shared params-shape concern and branches inside the existing transaction:

- When `plan_overrides` matches the units-only envelope shape (only the `fixed_charges` key, every entry carrying just `id`, `units`, and optionally `apply_units_immediately`) AND the subscription's plan is not itself an override, the service writes a `Subscription::FixedChargeUnitsOverride` row per entry via `find_or_initialize_by` (so repeated updates update one row rather than stacking), emits a FixedChargeEvent per entry through `FixedCharges::EmitEventsService`, and dispatches `Invoices::CreatePayInAdvanceFixedChargesJob` in an `after_commit` block when `apply_units_immediately` is set on a pay-in-advance fixed charge.
- Mixed-field `plan_overrides`, non-units entry fields, and updates against subscriptions already on an override plan all fall through to the existing `Plans::OverrideService` / `Plans::UpdateService` paths unchanged.

The cloned-plan guard (`subscription.plan.parent_id`) lives in the service alongside the predicate call; the concern stays a pure params-shape predicate.

Specs replace the prior tests that documented the old plan-cloning behaviour on units-only payloads (events on cloned FCs, two override fixed_charges) with assertions for the new branch (override row written, single event for the targeted FC at current time, no plan or fixed_charge clone). New contexts cover the override-update case, the pay-in-advance + apply_units_immediately dispatch, and the fall-through on subscriptions already on an overridden plan. Tests covering non-units payloads continue to assert the legacy clone behaviour unchanged.

### DIFF
--- a/app/services/subscriptions/update_service.rb
+++ b/app/services/subscriptions/update_service.rb
@@ -3,6 +3,7 @@
 module Subscriptions
   class UpdateService < BaseService
     include Subscriptions::Concerns::BillingEntityResolutionConcern
+    include Subscriptions::Concerns::FixedChargeUnitsOverrideDetectionConcern
 
     Result = BaseResult[:subscription, :payment_method]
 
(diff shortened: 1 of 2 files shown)


ancorcruz:

PR #5698, opened 2026-06-11
feat(payment-gated-subs): Handle billing of missed periods while the subscription was incomplete

osmarluz:
## Context

A payment-gated subscription can stay `incomplete` across one or more billing-period boundaries (the activation timeout is configurable and can be disabled). The billing clock only bills active subscriptions, so when payment finally succeeds, every period that elapsed while waiting is silently skipped — including pay-in-advance fixed-charge events scheduled for those periods.

## Description

Adds `Subscriptions::ActivationRules::BillMissedPeriodsService` and its job, enqueued from `Subscriptions::ActivateService` when a payment-gated subscription transitions from `incomplete` to `active`. The service replays the clock ticks the subscription missed: starting from the end of the period the gated invoice billed, it walks forward one billing period at a time and enqueues the same `BillSubscriptionJob` with `:subscription_periodic` the clock would have run at each boundary, so all existing boundary-scoped duplicate guards apply unchanged. The replayed periodic invoices also bill pay-in-advance fixed charges for their period, which is how events scheduled during the incomplete window get picked up.

Periods already billed are skipped via `InvoiceSubscription.matching?`, and yearly/semiannual plans with monthly-billed charges advance by their monthly split boundary. The catch-up runs only for first-time activations: for upgrades and downgrades the previous subscription remained active and billed until activation, and the new plan's current-period remainder is intentionally left to the next regular billing run.


ancorcruz (approved):
LGTM!

PR #5701, opened 2026-06-11
feat(orders): create order at order form signing

toommz:
## Roadmap Task

## Context

An Order is the immutable record of a signed commercial agreement. It is never created directly: it comes to life when an order form is marked as signed, and is later executed to provision billing in Lago (or exported as-is for external systems).

## Description

- Add the bare-minimum `Order` model and migration: `created`/`executed` statuses, `execution_mode` (`execute_in_lago` / `order_only`) and `execute_at` stored for audit, `OR-YYYY-NNNN` numbering, one order per order form.
- create an `Order` in `OrderForms::MarkAsSignedService` inside the signing transaction.

### DIFF
--- a/app/models/customer.rb
+++ b/app/models/customer.rb
@@ -65,6 +65,7 @@ class Customer < ApplicationRecord
   has_many :daily_usages
   has_many :quotes
   has_many :order_forms
+  has_many :orders
   has_many :wallets
   has_many :wallet_transactions, through: :wallets
   has_many :payment_provider_customers,
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+class Order < ApplicationRecord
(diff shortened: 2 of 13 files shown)


vincent-pochet (approved):
Looks great!
One small comment: the PR description is mentioning a new `Orders::CreateService` but the `MarkAsSignedService` is creating the order without relying on a service. I think it's good as it is, but maybe you should update the description accordingly.

toommz:
> Looks great! One small comment: the PR description is mentioning a new `Orders::CreateService` but the `MarkAsSignedService` is creating the order without relying on a service. I think it's good as it is, but maybe you should update the description accordingly.

@vincent-pochet oh yes good catch! I started with a CreateService, but it was really empty and also, creating an Order should not be sometihng happening from nowhere, so I decided to roll back to directly create the order. But of course forgot to update the PR body

PR #5704, opened 2026-06-11
[ING-253] feat(wallets): return billing_entity_code on wallet transactions

aquinofb:
## Context

Wallet transactions snapshot their billing entity since #5623, but the API doesn't return it anywhere.

## Changes

- Return `billing_entity_code` on wallet transaction payloads (index, show, consumptions/fundings and the webhooks). It returns the snapshotted entity, falling back to the wallet's one for transactions created before the snapshot existed -- the same entity the transaction settles under.
- Preload the associations on the index and consumption endpoints to avoid N+1 queries.

### DIFF
--- a/app/controllers/api/v1/wallet_transactions_controller.rb
+++ b/app/controllers/api/v1/wallet_transactions_controller.rb
@@ -104,8 +104,8 @@ module Api
 
         return render_error_response(result) unless result.success?
 
-        includes = (direction == :consumptions) ? %i[outbound_wallet_transaction] : %i[inbound_wallet_transaction]
-        preloads = {includes.first => [:billing_entity, {wallet: {customer: :billing_entity}}]}
+wallet_transaction_direction = (direction == :consumptions) ? :outbound_wallet_transaction : :inbound_wallet_transaction
+        includes = {wallet_transaction_direction => [:billing_entity, {wallet: {customer: :billing_entity}}]}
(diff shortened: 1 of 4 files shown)


mariohd (approved):

app/controllers/api/v1/wallet_transactions_controller.rb
```diff
@@ -105,11 +105,12 @@ def wallet_transaction_consumptions(direction:)
         return render_error_response(result) unless result.success?
 
         includes = (direction == :consumptions) ? %i[outbound_wallet_transaction] : %i[inbound_wallet_transaction]
+        preloads = {includes.first => [:billing_entity, {wallet: {customer: :billing_entity}}]}
```
toommz: nit but I find it strange to use two different but really close terms, and then eventually do `.includes(preloads)` while we have an `includes` variable at hands.

```suggestion
wallet_transaction_direction = (direction == :consumptions) ? :outbound_wallet_transaction : :inbound_wallet_transaction
        includes = {wallet_transaction_direction => [:billing_entity, {wallet: {customer: :billing_entity}}]}
```

toommz (approved):
One small nit you can omit, but LGTM

app/controllers/api/v1/wallet_transactions_controller.rb
toommz: Sorry @aquinofb it wasn't a "usable" suggestion, next time I'll use simple block codes, not suggestion when I know it's not 🙏

PR #5705, opened 2026-06-11
fix(subscription): add cancellation_reason column (expand step)

ancorcruz:
## Context

The `subscriptions.cancelation_reason` column shipped with a typo. The feature is currently used only by a development customer account, its data is small and will be migrated manually after releasing the 2nd PR. Renaming the column directly is unsafe because of the deploy race window: a single migration would either drop the old column before old pods stop running, or surface the new column before new code is rolled out.

## Description

First of three PRs implementing an expand/contract rename:

* New `subscription_cancellation_reasons` PG enum type and matching `subscriptions.cancellation_reason` column. Strictly additive.
* `CANCELATION_REASONS` constant renamed to `CANCELLATION_REASONS`; callers updated. The old `enum :cancelation_reason` declaration is preserved but switched to the renamed constant and prefixed to avoid predicate-method collisions with the new enum.
* No services, serializers or GraphQL fields change behaviour in this step. PR 2 will switch the application to read/write the new column and ignore the old one. PR 3 will drop the old column once self-hosted deployments have had time to roll through this and PR 2.

### DIFF
--- a/app/graphql/types/subscriptions/cancelation_reason_enum.rb
+++ b/app/graphql/types/subscriptions/cancelation_reason_enum.rb
@@ -3,7 +3,7 @@
 module Types
   module Subscriptions
     class CancelationReasonEnum < Types::BaseEnum
-      Subscription::CANCELATION_REASONS.each_key do |reason|
+      Subscription::CANCELLATION_REASONS.each_key do |reason|
         value reason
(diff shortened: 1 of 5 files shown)


osmarluz (approved):
🚀

PR #5706, opened 2026-06-11
perf(events): scope billing period filter to plan codes

jdenquin:
## Context

Events::BillingPeriodFilterService queries the event store for the distinct codes (or charge/filter pairs) received during the billing period, filtering only on organization_id and external_subscription_id. The ClickHouse events tables are sorted by (organization_id, code, external_subscription_id, ...), so without a code predicate ClickHouse cannot prune granules and scans the organization's entire event history. This query is issued for every usage and billing computation and is the single largest CPU consumer on ClickHouse Cloud.

## Description

Pass the plan's billable metric codes to distinct_codes and distinct_charges_and_filters in all event stores and filter on them. The result is unchanged: both code paths already intersect the event store result with the plan's charges, so events with codes outside the plan could never match a charge. With the code predicate, ClickHouse prunes by the (organization_id, code) sort key prefix and only reads the subscription's own data.

### DIFF
--- a/spec/services/events/billing_period_filter_service_spec.rb
+++ b/spec/services/events/billing_period_filter_service_spec.rb
@@ -552,6 +552,15 @@ RSpec.describe Events::BillingPeriodFilterService do
           expect(result.charges).to eq({})
         end
       end
+
+      it "scopes the event store query to the plan billable metric codes" do
+        event_store = instance_double(Events::Stores::PostgresStore, distinct_codes: [])
+        allow(Events::Stores::StoreFactory).to receive(:new_instance).and_return(event_store)
+
+        filter_service.call
+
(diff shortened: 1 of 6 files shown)


spec/services/events/billing_period_filter_service_spec.rb
```diff
@@ -1108,6 +1117,51 @@
...
+              value: 12,
+              decimal_value: 12.0,
+              charge:
+            )
+          end
+        end
+
+        before { enriched_events }
+
+        it "returns filtered charges" do
```
toommz: The example name seems a bit off as we are expecting an empty hash in the example body

toommz (approved):

annvelents (approved):

vincent-pochet (approved):
🚀

PR #5707, opened 2026-06-12
feat(products): add rate card GraphQL

rsempe:
## Context

Rate cards are where pricing lives: one card prices exactly one product, or one filter slice of it. Rates sit on the card as an effective-dated timeline, so a price change is an append rather than an edit.

## Description

- Types, enums and inputs for `RateCard` (billing timing, regroup), with mutations and resolvers behind the `rate_cards:*` permissions.
- List fields (`ratesCount`, the attachment lock signal) batched through the dataloader sources.
- Rate cards are exposed without nesting their rates: rates are fetched through their own query (`rateCardRates`), keyed by card.
- `regroup_paid_fees` is exposed as a nullable enum where a null input means `none` (the column is NOT NULL).
- Rate card added to the activity-log resource union.


app/serializers/v1/rate_card_serializer.rb
```diff
@@ -17,22 +17,22 @@ def serialize
         regroup_paid_fees: model.regroup_paid_fees,
         applied_pricing_unit_code: model.applied_pricing_unit_code,
         wallet_targetable: model.wallet_targetable,
+        rates_count: model.rates.size,
         created_at: model.created_at.iso8601,
         updated_at: model.updated_at.iso8601
       }
 
-      payload.merge!(rates) if include?(:rates)
```
endenis, line 24: The [app/services/utils/activity_log.rb](https://github.com/getlago/lago-api/blob/236bdb6bb36c0cc29551a99b6c73d9b727477895/app/services/utils/activity_log.rb#L13) file still references the `rates` option.

Is it better to use the `active_rate` in the activity log or to keep the `rates` option here?

app/graphql/types/rate_cards/object.rb
```diff
@@ -0,0 +1,50 @@
...
+      field :proration, Boolean, null: false
+      field :regroup_paid_fees, Types::RateCards::RegroupPaidFeesEnum, null: false
+
+      field :applied_pricing_unit_code, String, null: true
+      field :wallet_targetable, Boolean, null: true
+
+      # Lock signals for clients: deletion locks at any plan/subscription
+      # attachment, rate edits lock once a subscription bills the card.
+      field :attached_to_plan_or_subscription, Boolean, null: false
+      field :attached_to_subscriptions, Boolean, null: false, method: :attached_to_subscriptions?
```
endenis: This method does 2 database queries. We might need to optimize it later when rendering a list of rate cards.

app/graphql/types/rate_cards/object.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      field :product, Types::Products::Object, null: false
+      field :product_filter, Types::ProductFilters::Object, null: true
+
+      field :active_rate, Types::RateCardRates::Object, null: true
```
endenis, line 35: This field can also produce N+1 queries when rendering a list of rate cards

endenis (approved):

app/graphql/sources/attached_to_subscriptions.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Sources
+  # Batches RateCard#attached_to_subscriptions?: a card bills someone through
+  # a direct subscription attachment or through a plan that has subscriptions.
+  class AttachedToSubscriptions < GraphQL::Dataloader::Source
+    def fetch(ids)
+      attached = SubscriptionRateCard.where(rate_card_id: ids).distinct.pluck(:rate_card_id) |
```
tiagolupepic, line 8: perhaps we could use a scenic view for this?

app/serializers/v1/rate_card_serializer.rb
rsempe: Full rates are the only option that makes the audit diff show what actually changed for the activity log

app/graphql/types/rate_cards/object.rb
rsempe: Doing it

tiagolupepic (approved):
just a small comment and it's not a blocker

:shipit:

app/graphql/sources/attached_to_subscriptions.rb
rsempe: thanks! yeah interesting indeed, but not sure it's the correct timing as the semantics changed twice this week 😅 
for now it's two trivial indexed lookups, that's acceptable but yeah i'll keep in mind 👍

PR #5708, opened 2026-06-12
fix: prevent double-charge from open Stripe checkout sessions

brunomiguelpinto:
## Context

Lago has been double-charging invoices when a auto-charge runs against an invoice that also has an active hosted Stripe Checkout Session. Both succeed, producing two separate PaymentIntents with two different idempotency keys — so Stripe captures twice. 

- **Checkout-first then auto-charge**: the customer pays via the URL → Lago's auto-charge fires later and charges the saved card a second time.
- **Auto-charge mid-flight then checkout paid**: the off-session charge is already running at Stripe → the customer pays the still-live URL → second charge.

Stripe can't know the two PaymentIntents belong to the same invoice — only Lago can — and today nothing in the auto-charge path checks whether the invoice is already settled, nor kills open checkout sessions once it is. So both sides of the race stay open.

## Description

Two coordinated mechanisms close the two sides of the race, both treating the Lago invoice as the single source of truth.

### 1. Pre-confirm last-look — `Invoices::Payments::AlreadyPaidError`

`PaymentProviders::Stripe::Payments::CreateService` re-reads `invoice.reload.payment_succeeded?` immediately before `Stripe::PaymentIntent.create` — the tightest point before money moves, after the provider setup calls — and raises `AlreadyPaidError` if the invoice is already fully settled (`payment_succeeded?`, so partially-paid invoices still charge the remainder).

Both `Invoices::Payments::CreateService` and `PaymentRequests::Payments::CreateService` rescue it: skip silently (no error webhook, no retry, no failure email) and destroy the unused, never-charged pending `Payment` so it can't linger or be reused with a stale amount.

### 2. Post-settle expiry — `PaymentIntents::ExpireService` via `ExpireJob`

Hooked into `Invoices::Payments::StripeService#update_payment_status` on transition to `payment_status: :succeeded`. Cancels checkout sessions created mid-auto-charge that the last-look never saw. Enqueued **only when an active `PaymentIntent` exists**

> [!IMPORTANT]
> Scoped to **Stripe**. other providers will be on other pull request one per provider se we can easily test the flow first

### DIFF
--- a/app/services/invoices/payments/stripe_service.rb
+++ b/app/services/invoices/payments/stripe_service.rb
@@ -95,6 +95,8 @@ module Invoices
         )
 
         result
+      rescue ::Stripe::InvalidRequestError # the other ones are on the retry job
+        result
       end
 
       private
--- a/app/services/invoices/payments/stripe_service.rb
+++ b/app/services/invoices/payments/stripe_service.rb
@@ -39,7 +39,14 @@ module Invoices
         payment.error_code = stripe_payment.error_code if stripe_payment.error_code
         payment.save!
 
-        deliver_webhook if payable_payment_status.to_sym == :succeeded
+        if payable_payment_status.to_sym == :succeeded
+          deliver_webhook
+
+          # The invoice is paid lets expire/cancel any still-open checkout session
+          if PaymentIntent.active.exists?(invoice: payment.payable)
+            PaymentIntents::ExpireJob.perform_later(payment.payable)
+          end
+        end
 
         if status.to_s == "failed" && result.invoice.payments.excluding(result.payment).where(status: :requires_action).any?
           # We don't update the invoice status because it's likely the webhook of a failed payment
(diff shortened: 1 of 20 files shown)


app/services/invoices/payments/stripe_service.rb
```diff
@@ -70,12 +70,33 @@ def generate_payment_url(payment_intent)
...
+          {api_key: stripe_api_key}
+        )
+
+        return result unless session.status == "open"
+
+        ::Stripe::Checkout::Session.expire(
+          payment_intent.provider_session_id,
+          {},
+          {api_key: stripe_api_key}
+        )
```
rsempe, line 95: i would expect a `rescue ::Stripe::Something` here no?

rsempe (approved):
lgtm

app/services/invoices/payments/stripe_service.rb
brunomiguelpinto: yes you are right when it was already expired this can raise InvalidRequestError added it

PR #5709, opened 2026-06-12
fix(subscription): switch to cancellation_reason column

ancorcruz:
## Context

PR 1 added the correctly-named `cancellation_reason` column alongside the typo'd `cancelation_reason`. This is the middle step of the expand/contract rename: switch the application to the new column and make ActiveRecord forget the old one ever existed, so PR 3 can drop the column without any deploy race.

## Description

* Services (`Subscriptions::ActivationRules::ExpireService`, `Subscriptions::ActivationRules::Payment::ResolveService`), `V1::SubscriptionSerializer`, and the GraphQL `Subscription` object now read and write `cancellation_reason`.
* The `CancelationReasonEnum` GraphQL class is renamed to `CancellationReasonEnum`. The GraphQL field is exposed under `cancellationReason`. This is a breaking API rename, but only the development customer account consumes it today.
* `Subscription.ignored_columns` includes the old `cancelation_reason` so the schema cache no longer reflects it. The old `enum` declaration is removed. The DB column is left in place until PR 3.
* Tests updated accordingly.

### DIFF
--- a/app/graphql/types/subscriptions/cancellation_reason_enum.rb
+++ b/app/graphql/types/subscriptions/cancellation_reason_enum.rb
@@ -2,7 +2,7 @@
 
 module Types
   module Subscriptions
-    class CancelationReasonEnum < Types::BaseEnum
+    class CancellationReasonEnum < Types::BaseEnum
       Subscription::CANCELLATION_REASONS.each_key do |reason|
         value reason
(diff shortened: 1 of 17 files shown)


osmarluz (approved):
🚀

PR #5714, opened 2026-06-12
Bil 183 enable de e invoicing

mariohd:
## Context

The German e-invoicing project requires that DE billing entities go through the same structured-invoice pipeline France already uses. Today that pipeline is hard-gated by `BillingEntity::EINVOICING_COUNTRIES`, which contains only `"FR"`. This PR opens the gate for Germany and, while there, removes the six-way duplication of the eligibility predicate so future country/format work has one canonical place to read.

**Closes BIL-183.**

## Changes

Two commits, scoped tightly:

**`feat(einvoicing): Enable for Germany`** — adds `"DE"` to `BillingEntity::EINVOICING_COUNTRIES`. This single constant change activates both ZUGFeRD (Factur-X CII embedded in PDF) and the future XRechnung UBL work: both formats are gated by the same list. No serializer or builder code is touched — for ZUGFeRD the existing French pipeline already produces a compliant document, since the German guideline (`urn:cen.eu:en16931:2017`) is identical to France's. XRechnung-specific work lands in follow-up tickets.

**`refactor(einvoicing): Centralize gate`** — extracts the predicate

```ruby
einvoicing && EINVOICING_COUNTRIES.include?(country.try(:upcase))
```

into `BillingEntity#eligible_for_einvoicing?`. The expression was inlined in six `Generate*Service` files (invoice / credit note / payment receipt × PDF / XML). All six now delegate. Semantics are unchanged — nil-country and lowercase-country cases still handled.

## Tests

Updates to the existing specs cover the new shape:

- `spec/models/billing_entity_spec.rb` — corrected a misleading example description on the supported-country branch (the example was already asserting validity; the `it` string still said "is not valid"). The `validate_einvoicing` block continues to cover FR, BR, nil, and the `einvoicing: false` exemption.
- `spec/services/billing_entities/update_service_spec.rb` — adds a positive `when country is supported` context that enables einvoicing on a DE billing entity through the update service. The sibling `before { billing_entity.update(einvoicing: true) }` block and `einvoicing: true` override are hoisted into the parent `when enable einvoicing` context so all three sibling cases share the same setup.
- Every `Generate*Service` spec continues to exercise the gate through the real call path; behaviour for FR is preserved transitively.

Full run of the eight affected specs:

```
  spec/models/billing_entity_spec.rb \
  spec/services/billing_entities/update_service_spec.rb \
  spec/services/invoices/generate_{pdf,xml}_service_spec.rb \
  spec/services/credit_notes/generate_{pdf,xml}_service_spec.rb \
  spec/services/payment_receipts/generate_{pdf,xml}_service_spec.rb
# 150 examples, 0 failures
```

## Follow-up

`BillingEntity#eligible_for_einvoicing?` is currently covered transitively through the six service specs. A direct unit block on `spec/models/billing_entity_spec.rb` (FR / DE / lowercase / BR / nil / `einvoicing: false`) would lock the method's contract independently and survive future refactors of the service callers — happy to add as a small follow-up if you want it on this PR before merge.

### DIFF
--- a/app/services/credit_notes/generate_pdf_service.rb
+++ b/app/services/credit_notes/generate_pdf_service.rb
@@ -72,7 +72,7 @@ module CreditNotes
     end
 
     def should_generate_cii_einvoice_xml?
-      credit_note.billing_entity.einvoicing && BillingEntity::EINVOICING_COUNTRIES.include?(credit_note.billing_entity.country.try(:upcase))
+      credit_note.billing_entity.eligible_for_einvoicing?
     end
 
     def should_generate_pdf?
--- a/spec/models/billing_entity_spec.rb
+++ b/spec/models/billing_entity_spec.rb
@@ -241,7 +241,7 @@ RSpec.describe BillingEntity do
     context "with a supported country" do
       let(:country) { "fr" }
 
-      it "is not valid" do
+      it "is valid" do
         expect(billing_entity).to be_valid
       end
     end
--- a/app/models/billing_entity.rb
+++ b/app/models/billing_entity.rb
@@ -154,6 +154,10 @@ class BillingEntity < ApplicationRecord
       )
   end
 
+  def eligible_for_einvoicing?
+    einvoicing && EINVOICING_COUNTRIES.include?(country.try(:upcase))
+  end
+
   private
 
   def generate_document_number_prefix
--- a/app/models/billing_entity.rb
+++ b/app/models/billing_entity.rb
@@ -14,7 +14,7 @@ class BillingEntity < ApplicationRecord
     "payment_receipt.created"
   ]
 
-  EINVOICING_COUNTRIES = %w[FR].map(&:upcase)
+  EINVOICING_COUNTRIES = %w[FR DE].map(&:upcase)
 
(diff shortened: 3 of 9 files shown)


app/services/credit_notes/generate_pdf_service.rb
```diff
@@ -72,7 +72,7 @@ def cleanup_tempfiles(pdf_file, xml_file)
     end
 
     def should_generate_cii_einvoice_xml?
-      credit_note.billing_entity.einvoicing && BillingEntity::EINVOICING_COUNTRIES.include?(credit_note.billing_entity.country.try(:upcase))
+      credit_note.billing_entity.eligible_for_einvoicing?
```
annvelents, line 75: 👍

spec/models/billing_entity_spec.rb
```diff
@@ -241,7 +241,7 @@
     context "with a supported country" do
       let(:country) { "fr" }
 
-      it "is not valid" do
+      it "is valid" do
```
annvelents, line 244: 🙈

annvelents (approved):

rsempe (approved):

PR #5715, opened 2026-06-15
fix(subscriptions): avoid charge override N+1

tiagolupepic:
Creating subscriptions with plan overrides clones each plan charge. The cloning path reloaded the same plan, organization, billable metric, and pricing unit associations for every charge, which triggered N+1 queries.

Fixes API-248

Preload the associations needed when overriding plan charges and pass the overridden plan through charge override params. Reuse loaded associations while cloning charges and applied pricing units to avoid repeated lookups.

Add service specs covering the plan override handoff and charge override cloning when a plan object is provided.

### DIFF
--- a/app/services/charges/override_service.rb
+++ b/app/services/charges/override_service.rb
@@ -14,6 +14,9 @@ module Charges
 
       ActiveRecord::Base.transaction do
         new_charge = charge.dup.tap do |c|
+          c.organization = params[:plan].organization if params[:plan]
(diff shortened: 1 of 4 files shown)


rsempe (approved):

rsempe:
Should we have to do the same for **fixed charge** override?

tiagolupepic:
> Should we have to do the same for fixed charge override?

@rsempe, thanks for letting me know, address it in another PR!

PR #5717, opened 2026-06-15
feat(order-forms): file attachments upload

toommz:
## Roadmap Task

👉  https://getlago.canny.io/feature-requests/p/create-quotes-order-forms

## Context

Quotes' content can include attachments (images only at the moment), so the frontend needs a way to attach/upload medias to the quotes.

## Description

This PR adds a GraphQL mutation to upload images to a quote. The attachments are on the quote so can be shared between quote versions.


app/services/quotes/add_image_service.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    attr_reader :quote, :image
+
+    def image_blob
+      decoded = Utils::Base64File.decode(image)
+
+      if decoded.nil?
+        result.single_validation_failure!(field: :image, error_code: "invalid_format")
+        return
+      end
```
vincent-pochet, line 48: We are validating the image size using the model validation, witch is good, however it requires the file to be written to the storage before validating the size.

Claude is suggesting to add a prevalidation at this level, which I think makes sense:
```
  if decoded.io.size > IMAGE_MAX_SIZE
    result.single_validation_failure!(field: :image, error_code: "file_too_large")
    return
  end
```

vincent-pochet (approved):

app/services/quotes/add_image_service.rb
toommz: 8c88174af8551aed71968d3dbd8af8574daef5d4

PR #5718, opened 2026-06-15
misc(mailers): Auto retry on Net::OpenTimeout

vincent-pochet:
## Description
This PR adds auto retry handling to the `SendEmailJob` in case of `Net::OpenTimeout`.
The test coverage on all retryable error is also improved


rsempe (approved):
👍

toommz (approved):

PR #5361, opened 2026-04-17
feat(order-forms): add OrderForm GraphQL API

toommz:
## Roadmap Task

👉 Order Forms lifecycle

## Context

This is PR 2 of 2 in a stacked split. Stacked on the REST PR so the shared model, query, and factories are available. Requires a frontend dev to QA via the UI.

Stack:
- OrderForm REST (#5360)
- **#this** — OrderForm GraphQL

## Description

Add OrderForm read-only GraphQL API:

- `Types::OrderForms::Object` with customer, quote, and lifecycle fields
- `Types::OrderForms::StatusEnum`, `Types::OrderForms::VoidReasonEnum`
- `Resolvers::OrderFormResolver` (single) and `Resolvers::OrderFormsResolver` (collection with filters)
- Register `order_form` and `order_forms` on `QueryType`
- Regenerated `schema.graphql` and `schema.json`
- Resolver specs covering single lookup and each filter

PR #5720, opened 2026-06-15
feat(orders): REST API show and list endpoints

toommz:
## Roadmap Task

👉 Order Forms lifecycle

## Context

This is PR 3 of 4 in a stacked split. Stacked on the OrderForm GraphQL PR so all OrderForm code (REST + GraphQL + shared) is available.

Stack:
- OrderForm REST (#5360)
- OrderForm GraphQL (#5361)
- **#this** — Order REST
- Order GraphQL (stacked) #5724

## Description

Add Order read-only REST API plus the shared infrastructure for Order.


app/controllers/api/v1/orders_controller.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class OrdersController < Api::BaseController
+      def index
+        return forbidden_error(code: "feature_not_available") unless current_organization.feature_flag_enabled?(:order_forms)
```
vincent-pochet: It ssems that we are usually returning "feature_unavailable" (106 occurrences in the code base) rather than this "feature_not_available" (6 occurrences). I think it's best to align to most common case

app/controllers/api/v1/orders_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+              meta: pagination_metadata(result.orders)
+            )
+          )
+        else
+          render_error_response(result)
+        end
+      end
+
+      def show
+        return forbidden_error(code: "feature_not_available") unless current_organization.feature_flag_enabled?(:order_forms)
```
vincent-pochet: This guard is shared between `index` and `show` actions. Maybe we could reuse the pattern from OrderFormController with:
```
before_action :ensure_feature_flag!`
```

app/contracts/queries/orders_query_filters_contract.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Queries
+  class OrdersQueryFiltersContract < Dry::Validation::Contract
+    params do
+      optional(:status).maybe do
+        value(:string, included_in?: Order::STATUSES.keys.map(&:to_s)) |
```
vincent-pochet: Could we use `Order::STATUSES.values` as for `ORDER_TYPES`?

vincent-pochet (approved):

app/contracts/queries/orders_query_filters_contract.rb
toommz: All comments resolved in cdd785f8eaeadba5a5d837dbaadb13e501c86a72

PR #5721, opened 2026-06-15
feat(payment-gated-subs): push finalized invoice number to PSP after activation

ancorcruz:
## Context

When a payment-gated subscription activates, the underlying invoice is created in `open` state and carries no invoice number yet. The PSP payment we create at that moment ends up referencing a placeholder. Once the gating payment succeeds, the invoice transitions `open → finalized` and gets a real, human-readable number — but the PSP-side payment record never learns about it. Operator dashboards on Stripe, Adyen, and GoCardless keep showing the placeholder, which makes reconciliation against Lago's invoice list painful.

## Description

- After invoice finalization in `Subscriptions::ActivationRules::Payment::ResolveService#handle_success`, enqueue a new `PaymentProviders::UpdatePaymentReferenceJob` with the succeeded payment. The job dispatches to a provider-specific service based on the payment's provider type.
- Stripe (`PaymentProviders::Stripe::Payments::UpdateReferenceService`) calls `Stripe::PaymentIntent.update` to set the description and merge `lago_invoice_number` into metadata. Existing metadata keys are preserved by Stripe's per-key merge semantics.
- GoCardless (`PaymentProviders::Gocardless::Payments::UpdateReferenceService`) calls `client.payments.update` to set `lago_invoice_number` in the payment metadata.
- Adyen (`PaymentProviders::Adyen::Payments::UpdateReferenceService`) is a log-only no-op. Adyen treats `merchantReference` and `additionalData` on captured payments as immutable, and the Ruby SDK's `ModificationsApi` exposes no metadata-update surface. The service is wired in for shape consistency and documents the constraint inline; Adyen-side reconciliation continues to rely on the `lago_payment_id` set at payment creation.
- Cashfree, Flutterwave, and Moneyhash fall through to an info-level log noting the gap — their integrations have no metadata-update surface either.
- All updates run best-effort. Per-provider services catch PSP errors and log at warn level; the invoice is already finalized and the subscription is already active, so a failed reference update is presentation polish only and never blocks the activation flow.
- Guard clauses skip the call when `provider_payment_id` is blank, the payable is not an `Invoice`, or the invoice still has no number — keeping the contract narrow and easy to grep.


app/services/subscriptions/activation_rules/payment/resolve_service.rb
```diff
@@ -45,6 +45,12 @@ def handle_success
             Integrations::Aggregator::Invoices::Hubspot::CreateJob.perform_later(invoice:) if invoice.should_sync_hubspot_invoice?
             Integrations::Aggregator::Taxes::Invoices::CreateJob.perform_later(invoice:) if invoice.customer.tax_customer
             Utils::SegmentTrack.invoice_created(invoice)
+
+            # The PSP-side payment record carries a placeholder reference set
+            # at activation time (the invoice had no number yet). Now that the
+            # invoice is finalized and numbered, push the real number back to
+            # the PSP so operator dashboards reconcile cleanly. Best-effort.
+            PaymentProviders::UpdatePaymentReferenceJob.perform_later(succeeded_payment)
```
osmarluz, line 53: @ancorcruz when the invoice is a zero-amount one we never create a payment on `app/services/invoices/payments/create_service.rb`, so `succeeded_payment` can be `nil` here:


```suggestion
            PaymentProviders::UpdatePaymentReferenceJob.perform_later(succeeded_payment) if succeeded_payment
```

app/services/payment_providers/update_payment_reference_service.rb
```diff
@@ -0,0 +1,48 @@
...
+      result
+    end
+
+    private
+
+    attr_reader :payment
+
+    def delegate_to(service)
+      service.call!(payment:)
+      result
```
osmarluz: @ancorcruz the `result` is not needed here since we have it on line 36:


```suggestion
      service.call!(payment:)
```

app/services/payment_providers/gocardless/payments/update_reference_service.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        def call
+          result.payment = payment
+          return result if payment.provider_payment_id.blank?
+          return result unless payment.payable.is_a?(Invoice)
+          return result if invoice.number.blank?
+
+          client.payments.update(
+            payment.provider_payment_id,
+            params: {metadata: {lago_invoice_number: invoice.number}}
```
osmarluz: @ancorcruz does GoCardless merge new keys with the existing ones the same way Stripe does for the metadata? Asking because when the payment is created on GoCardless we send on `app/services/payment_providers/gocardless/payments/create_service.rb`:

```
metadata: {
  lago_invoice_id: invoice.id,
  lago_customer_id: invoice.customer_id,
  invoice_issuing_date: invoice.issuing_date.iso8601
}
```

spec/services/subscriptions/activation_rules/payment/resolve_service_spec.rb
osmarluz, line 1: @ancorcruz we should also have a context to cover the zero-amount invoice I mentioned

app/services/payment_providers/gocardless/payments/update_reference_service.rb
ancorcruz: Dropping the change for GoCardless, we never send the invoice number as reference in payment creation at first.

app/services/subscriptions/activation_rules/payment/resolve_service.rb
ancorcruz: If there is no payment this code won't be executed as zero-amount invoices does not go through the `ResolveService` at all. It does not reach `handle_success`. Rule evaluates directly to satisfied from `Invoices::Subscription` (line 43) service

spec/services/subscriptions/activation_rules/payment/resolve_service_spec.rb
ancorcruz: as mentioned before, we do not call ResolveService if there no payment, zero-amount invoices satisfies payment rule on rule evaluation stage.

osmarluz (approved):
🚀

PR #5722, opened 2026-06-15
fix(plans): Avoid duplicate plan.updated event from UI

rsempe:
## Context

The UI always submits **entitlements** as a non-null array in the `UpdatePlan` mutation, even when nothing entitlement-related changed. This caused both `Plans::UpdateService` and `Entitlement::PlanEntitlementsUpdateService` to emit a `plan.updated` webhook — and produce a `plan.updated` activity log — on every UI save: one with pre-save state and one with post-save state, with no way for consumers to tell which is authoritative.

On **create**, the entitlements service likewise emitted a spurious `plan.updated` on top of the `plan.created` event.

We don't have these issues on the API, because a plan and its entitlements cannot be updated in the same request.

## Description

Add a `send_webhook` option to `Entitlement::PlanEntitlementsUpdateService` (defaulting to `true`, so the standalone REST entitlement endpoints keep emitting their `plan.updated`). The flag gates **both** the webhook and the `plan.updated` activity log, since they represent the same event.

The `CreatePlan` and `UpdatePlan` mutations now call the entitlements service with `send_webhook: false`, so the plan event is emitted exactly once by the plan create/update service, while entitlements are still persisted (including clearing them when an empty array is sent).

### DIFF
--- a/app/graphql/mutations/plans/update.rb
+++ b/app/graphql/mutations/plans/update.rb
@@ -23,12 +23,15 @@ module Mutations
 
         return result_error(result) unless result.success?
 
+        # NOTE: send_webhook is false so we don't duplicate the plan.updated event
+        #       already emitted by Plans::UpdateService above.
         unless entitlements.nil?
           result = ::Entitlement::PlanEntitlementsUpdateService.call(
             organization: plan.organization,
             plan:,
             entitlements_params: Utils::Entitlement.convert_gql_input_to_params(entitlements),
-            partial: false
+            partial: false,
+            send_webhook: false
           )
         end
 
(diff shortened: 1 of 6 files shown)


vincent-pochet (approved):

tiagolupepic (approved):
:shipit:

app/graphql/mutations/plans/update.rb
```diff
@@ -23,12 +23,15 @@ def resolve(entitlements: nil, **args)
 
         return result_error(result) unless result.success?
 
+        # NOTE: send_webhook is false so we don't duplicate the plan.updated event
+        #       already emitted by Plans::UpdateService above.
         unless entitlements.nil?
```
annvelents, line 28: shall we also update it to `.empty?`?

annvelents: aaaaaahhhh
nope, we should not
we might be deleting the entitlements

annvelents (approved):

PR #5725, opened 2026-06-15
fix(events): count deduplicated events without JOIN to avoid ClickHouse OOM

brunomiguelpinto:
## Context

`Fees::CreatePayInAdvanceJob` is failing in production with ClickHouse `MEMORY_LIMIT_EXCEEDED` (Sentry API 243, 383 occurrences, escalating). The crash happens in `Events::Stores::ClickhouseStore#count`, reached through
`Charges::PayInAdvanceAggregationService` → `CountService#compute_aggregation` → `event_store.count` for a `count_agg` charge on a very large subscription.

`#count` builds the full two-pass deduplication CTE and runs `SELECT count() FROM events`. The `events` CTE performs an `INNER ANY JOIN` back to `events_enriched` whose only purpose is to fetch the `value` column of the latest enriched row per dedup key. For a count that column is never used, but the JOIN scans `events_enriched` a second time and materializes `value`, roughly doubling the memory of an already-large aggregation. Under worker
concurrency the per-query peaks stack up and cross the server-wide memory limit.

## Description

Count the distinct deduplication keys directly: a single `GROUP BY` over `DEDUP_KEY_COLUMNS` on `events_enriched`, wrapped in `SELECT count()`. No JOIN, no `value` materialization.

This is provably equivalent to the previous query for the unfiltered case: the inner grouping is identical to the `latest_enriched` CTE (same FROM, WHERE and GROUP BY), and the old `INNER ANY JOIN` returned exactly one row per `latest_enriched` row, so both queries equal the number of distinct dedup keys.

The original JOIN-based path is kept for filtered counts (`grouped_by_values`, `matching_filters`, `ignored_filters`) and the non-deduplicated case, so those behaviours are byte-for-byte unchanged. The filtered subsets are small and were never the source of the OOM, so there is no benefit in rewriting them and we avoid any risk to the "filter against the latest enriched row" semantics.

## Validation

Verified on the production subscription from the Sentry report (identical config, full history, ~10.8M events):

| metric       | current (JOIN) | proposed (no JOIN) |
|--------------|----------------|--------------------|
| count        | 10,822,296     | 10,822,296 (match) |
| server time  | 6,620 ms       | 997 ms (~6.6x)     |
| peak memory  | 4.7 GiB        | 3.6 GiB            |
| rows read    | 21,876,048     | 10,938,024 (½)     |
| bytes read   | 3.3 GiB        | 1.6 GiB (½)        |

The "rows read = exactly 2x" confirms the JOIN was re-scanning the table. The real win for API-243 is the combination of lower per-query memory and ~6.6x faster runtime, which removes the concurrent memory pressure that was tripping the server-wide limit.

### DIFF
--- a/app/services/events/stores/clickhouse_store.rb
+++ b/app/services/events/stores/clickhouse_store.rb
@@ -217,13 +217,41 @@ module Events
 
       def count
         Events::Stores::Utils::ClickhouseConnection.connection_with_retry do |connection|
-          sql = with_ctes(events_cte_queries(deduplicated_columns: %w[value]), <<-SQL)
+          connection.select_value(count_query).to_i
+        end
+      end
+
+      # Counting deduplicated events only needs the number of distinct dedup keys,
+      # not their latest enriched values. We therefore skip the `INNER ANY JOIN`
+      # performed by `events_cte_queries` (which materializes a column for every
+      # event and roughly doubles the memory of the dedup aggregation) and count
+      # the grouped keys directly. This avoids ClickHouse MEMORY_LIMIT_EXCEEDED on
+      # very large subscriptions.
+      #
+      # When the count is filtered (grouped_by_values or matching/ignored filters)
+      # we keep the JOIN-based path so the filter still applies to the latest
+      # enriched row per dedup key (identical semantics).
+      def count_query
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):

rsempe (approved):
👍

PR #5727, opened 2026-06-15
chore(invoices): Improve search query performances and add missing index

jdenquin:
- Add CTE to improve PG query plan
- Get back to `OR` for customer filtering, checked and it still uses right indexes
- Add missing index on `invoices.customer_id and invoices.organization_id`


annvelents (approved):
Nice!

PR #5728, opened 2026-06-15
perf(plans): Improve creating fixed-charge events

D1353L:
## Context
Updating a fixed charge's units on a plan (PUT /plans/{code}) emits a `FixedChargeEvent` for every active/incomplete subscription on the plan. This ran one subscription at a time through `FixedChargeEvents::CreateService`, and each iteration issued ~6 queries (plan, customer, billing entity, organization, units-override lookup, and the insert). On a plan with 10,000 subscriptions this meant ~60k SQL statements and >1 min for a single request.

## Description

- Replaced the per-subscription create loop with a batched write path
- Included `:plan` and `customer: :billing_entity` to prevent N+1

## Performance testing
- Plan with pay_in_advance fixed charge
- 20,000 active subscriptions

| Scenario | Original latency | Optimized latency |
| --------------- | ---------------  | ------------------ |
| Updating fixed charge units | >1min (timeout)  | 4.3s |


endenis (approved):
Looks good

PR #5730, opened 2026-06-16
[ING-330] misc(clickhouse): Returns grouped_sum with count in 1 query

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/5716 and https://github.com/getlago/lago-api/pull/5723

Today, all aggregation services (except CountService) perform two queries to retrieve data from the event store, a first one to get the actual aggregation and a second one to get the number of events involved in this result.

The current implementation is inherited from the initial Postgres store, and works well for this database engine. When the Clickhouse store, the same approach was used to keep the compatibility between the different implementation. At scale, this implementation is not ideal on the clickhouse side as it force the engine to retrieve twice the same set of records before performing the computation.

## Description

This PR introduce a new `GroupedAggregationResult` data structure, returned from the `grouped_sum` aggregation.
This structure returns for each grouped result, the `groups`, the `value` for the aggregation and `events_count` for the number of event.
Both Clickhouse and `Postgres` will query them in a single pass

The same approach will be added in next step on all other remaining aggregations

### DIFF
--- a/app/services/events/stores/clickhouse_enriched_store.rb
+++ b/app/services/events/stores/clickhouse_enriched_store.rb
@@ -532,13 +532,16 @@ module Events
         end
       end
 
-      def grouped_sum(columns = grouped_by)
+      def grouped_sum(columns = grouped_by, with_count: true)
+        count_select = with_count ? "count()" : "null"
+
         Utils::ClickhouseConnection.connection_with_retry do |connection|
           if columns == grouped_by
             sql = with_ctes(events_cte_queries(deduplicated_columns: %w[decimal_value]), <<-SQL)
               SELECT
                 sorted_grouped_by as groups,
-                sum(events.decimal_value) as value
+                sum(events.decimal_value) as value,
+                #{count_select} as events_count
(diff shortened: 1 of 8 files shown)


app/services/events/stores/clickhouse_enriched_store.rb
```diff
@@ -532,13 +532,16 @@ def sum(with_count: true)
         end
       end
 
-      def grouped_sum(columns = grouped_by)
+      def grouped_sum(columns = grouped_by, with_count: true)
+        count_select = with_count ? "toDecimal32(count(), 0)" : "null"
```
toommz: Why `toDecimal32`? Could it be not enough for some cases?

toommz (approved):

app/services/events/stores/clickhouse_enriched_store.rb
vincent-pochet: Good catch, there is no real reason for this cast. `count` is already returning an `UInt64`

PR #5732, opened 2026-06-16
[ING-312] misc(daily_usage): Allow fill_history task to ask for a list of subscriptions

vincent-pochet:
## Description

This PR is related to https://github.com/getlago/lago-api/pull/5694. It adds a new prompt to request an optional list of subscription to reprocess instead of all the active and terminated after the `from_date`

### DIFF
--- a/lib/task_prompt.rb
+++ b/lib/task_prompt.rb
@@ -26,6 +26,11 @@ module TaskPrompt
     organization
   end
 
+  def self.ask_for_subscription_ids
+    input = ask("Subscription IDs to refill (comma or space separated, leave blank to fill a whole organization): ")
+    input.split(/[\s,]+/).reject(&:empty?)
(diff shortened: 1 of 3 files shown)


rsempe (approved):

lib/task_prompt.rb
```diff
@@ -26,6 +26,11 @@ def self.ask_for_organization
     organization
   end
 
+  def self.ask_for_subscription_ids
+    input = ask("Subscription IDs to refill (comma or space separated, leave blank to fill a whole organization): ")
+    input.split(/[\s,]+/).reject(&:empty?)
+  end
```
toommz, line 32: Passing multiple IDs does not seem to be tested, might be a good idea to do so

toommz (approved):

PR #5734, opened 2026-06-16
feat(webhooks): Offload json toast to S3

jdenquin:


app/models/webhook.rb
```diff
@@ -15,19 +15,38 @@ def self.ransackable_attributes(_auth_object = nil)
...
 
-  # Up until 1.4.0, we stored the payload as a string. This method
-  # ensures that we can still read the old payloads.
-  # Webhooks created after 1.4.0 will have the payload stored as a hash.
-  # Webhooks are deleted after 90 days, so we can remove this method 90 days after every client has updated to 1.4.0.
+  def self.payload_storage
+    ActiveStorage::Blob.service
+  end
+
   def payload
```
vincent-pochet, line 22: Both `payload` and `response` are exposed in the webhook GraphQL type.
Since the front is actually consuming it directly, we should certainly sync with them on a way to change the current API contract:
- Remove the JSON fields from the main GraphQL type
- Add a new type and resolver exposing these JSON. It should be used to query the values for a **single** webhook.

Without it we are exposing ourself to a big performance issue

app/services/webhooks/send_http_service.rb
```diff
@@ -60,17 +60,17 @@ def retry_limit
 
     def mark_webhook_as_succeeded(response)
       webhook.http_status = response&.code&.to_i
-      webhook.response = response&.body.presence || {}
+      webhook.store_response(response&.body.presence || {})
```
vincent-pochet, line 63: If for some reasons, the calls to s3 fails, we will fail to mark the webhook as succeeded. 
We should probably add some error management around this logic.

We could maybe use the following approach:
Store temporary the response payload as JSON on the model and delegate the S3 handling to a new job responsible for the offload to s3 and the cleanup of the webhook record. What do you think?

vincent-pochet (approved):

PR #5736, opened 2026-06-17
fix(plans): Emit plan webhook after entitlements persist

rsempe:
When a plan is created or updated from the UI with entitlements, the single `plan.created`/`plan.updated` webhook was emitted by `Plans::CreateService` and `Plans::UpdateService`, which run before the mutation persists the entitlements through `Entitlement::PlanEntitlementsUpdateService`.


annvelents (approved):
Nice 🙌

PR #5737, opened 2026-06-17
[ING-316] feat(multi-billing-entity): backfill wallet transactions billing entity

aquinofb:
## Context

`billing_entity_id` was added to wallet transactions in #5623 and is set for new transactions, but older ones were left empty. Reads already fall back to the wallet's or customer's billing entity, so nothing is broken today. This just fills in the column so it can be relied on and queried directly.

## What it does

Fills the missing billing entity on existing wallet transactions, in the background and in small batches, so it stays safe on large datasets. We resolve it the same way new transactions do: the wallet's billing entity, falling back to the customer's.

It does not run on its own: you start it with a rake task when you are ready, and it can be safely re-run.

## Running it

```
bundle exec rake migrations:backfill_wallet_transactions_billing_entity
```

It runs on the `low_priority` queue, so make sure a worker is picking that up. It is light on the database; once it finishes, run `ANALYZE wallet_transactions`.

### DIFF
--- a/app/jobs/database_migrations/backfill_wallet_transactions_billing_entity_job.rb
+++ b/app/jobs/database_migrations/backfill_wallet_transactions_billing_entity_job.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 module DatabaseMigrations
+  # Post-release backfill: reads still fall back to the wallet/customer billing entity
+  # (Wallet#billing_entity), so this is safe to run whenever and is not an upgrade blocker today.
+  # When that fallback is removed, running this becomes mandatory and must be documented as a
+  # pre-upgrade step in the bridge-version migration guide.
   class BackfillWalletTransactionsBillingEntityJob < ApplicationJob
     queue_as :low_priority
     unique :until_executed
--- a/app/jobs/database_migrations/backfill_wallet_transactions_billing_entity_job.rb
+++ b/app/jobs/database_migrations/backfill_wallet_transactions_billing_entity_job.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module DatabaseMigrations
+  class BackfillWalletTransactionsBillingEntityJob < ApplicationJob
+    queue_as :low_priority
+    unique :until_executed
+
+    BATCH_SIZE = 1_000
+
+    def perform(batch_number = 1)
+      result = ActiveRecord::Base.connection.execute(<<~SQL.squish)
+        WITH batch AS (
+          SELECT wt.id,
+                 COALESCE(i.billing_entity_id, w.billing_entity_id, c.billing_entity_id) AS resolved_billing_entity_id
+          FROM wallet_transactions wt
+          JOIN wallets w ON w.id = wt.wallet_id
+          JOIN customers c ON c.id = w.customer_id
+          LEFT JOIN invoices i ON i.id = wt.invoice_id
+          WHERE wt.billing_entity_id IS NULL
+          LIMIT #{BATCH_SIZE}
+        )
+        UPDATE wallet_transactions wt
+        SET billing_entity_id = batch.resolved_billing_entity_id
+        FROM batch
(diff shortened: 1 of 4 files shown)


mariohd (approved):
Looks good

toommz (approved):

app/jobs/database_migrations/backfill_wallet_transactions_billing_entity_job.rb
```diff
@@ -0,0 +1,39 @@
...
+    queue_as :low_priority
+    unique :until_executed
+
+    BATCH_SIZE = 1_000
+
+    def perform(batch_number = 1)
+      result = ActiveRecord::Base.connection.execute(<<~SQL.squish)
+        WITH batch AS (
+          SELECT wt.id,
+                 COALESCE(i.billing_entity_id, w.billing_entity_id, c.billing_entity_id) AS resolved_billing_entity_id
```
lovrocolic: I think we only need `COALESCE(w.billing_entity_id, c.billing_entity_id)`

I guess we can also drop join with invoices in that case

app/jobs/database_migrations/backfill_wallet_transactions_billing_entity_job.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module DatabaseMigrations
+  class BackfillWalletTransactionsBillingEntityJob < ApplicationJob
+    queue_as :low_priority
+    unique :until_executed
+
+    BATCH_SIZE = 1_000
+
+    def perform(batch_number = 1)
```
groyoh, line 14: Is any a need to have this parameter since we already process the first batch ?

lib/tasks/migrations/wallet_transactions.rake
groyoh, line 1: Could you please add a test for this task ?

lib/tasks/migrations/wallet_transactions.rake
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+namespace :migrations do
+  desc "Backfill billing_entity_id on existing wallet transactions"
+  task backfill_wallet_transactions_billing_entity: :environment do
```
groyoh, line 5: Could we maybe use `perform_required_jobs` instead as it the rake task we try and re-use for upgrade when possible.

lib/tasks/migrations/wallet_transactions.rake
```diff
@@ -0,0 +1,52 @@
...
+      pp "- Make sure a Sidekiq worker is draining the low_priority queue"
+      DatabaseMigrations::BackfillWalletTransactionsBillingEntityJob.perform_later
+    else
+      pp "  -> Nothing to do"
+    end
+
+    stalled_checks = 0
+
+    while count > 0
+      sleep 30
```
groyoh: Is 30 seconds maybe a bit too much ?

app/jobs/database_migrations/backfill_wallet_transactions_billing_entity_job.rb
aquinofb: Good question! Purely to keep each re-enqueue unique for `until_executed`, it'd be deduped and stop after the first batch

lib/tasks/migrations/wallet_transactions.rake
aquinofb: Didn't know about that. Let me take a look

vincent-pochet (approved):

lovrocolic (approved):

lib/tasks/migrations/wallet_transactions.rake
aquinofb: Checked the convention internally. I'm keeping it standalone for this release (fallback covers reads, nothing depends on it yet). It becomes a documented mandatory step when we remove the fallback in a future bridge version.

PR #5739, opened 2026-06-17
[ING-320] misc (invoices): add migration for backfilling prepaid credits breakdown

lovrocolic:
## Context

Currently, we miss prepaid credits breakdown on historical invoices. Breakdown is only present on invoices created after the release of the feature.

## Description

This PR adds migration for `prepaid_granted_credit_amount_cents` and `prepaid_purchased_credit_amount_cents` bakfill on historical invoices

### DIFF
--- a/app/jobs/database_migrations/backfill_prepaid_credit_breakdown_job.rb
+++ b/app/jobs/database_migrations/backfill_prepaid_credit_breakdown_job.rb
@@ -45,14 +45,21 @@ module DatabaseMigrations
     end
 
     # Invoices the live code would compute a breakdown for, not yet filled:
-    # prepaid credit was applied, breakdown is empty, the customer is fully
-    # traceable, and there are consumption rows to aggregate.
+    #   * the invoice is settled (finalized or voided), not draft/failed/pending/etc.
+    #   * prepaid credit was applied and the breakdown is empty
+    #   * the customer is fully traceable
+    #   * there are consumption rows to aggregate
+    #   * those rows reconcile: total consumed == prepaid_credit_amount_cents.
+    #     A mismatch means the consumption ledger is inconsistent for this invoice,
+    #     so we skip it rather than write a wrong (and unbalanced) breakdown.
     def self.candidates(organization_id = nil)
       scope = Invoice
+        .where(status: %i[finalized voided])
         .where(prepaid_granted_credit_amount_cents: nil, prepaid_purchased_credit_amount_cents: nil)
         .where("prepaid_credit_amount_cents > 0")
         .where.not(customer_id: Wallet.where(traceable: false).select(:customer_id))
         .where(id: consumed_invoice_ids)
+        .where("prepaid_credit_amount_cents = (#{total_consumed_sql})")
       scope = scope.where(organization_id:) if organization_id
       scope
     end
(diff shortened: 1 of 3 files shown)


app/jobs/database_migrations/backfill_prepaid_credit_breakdown_job.rb
```diff
@@ -0,0 +1,110 @@
...
+          FROM wallet_transaction_consumptions c
+          JOIN wallet_transactions out_wt ON out_wt.id = c.outbound_wallet_transaction_id
+          JOIN wallet_transactions in_wt ON in_wt.id = c.inbound_wallet_transaction_id
+          WHERE out_wt.invoice_id = invoices.id
+            AND in_wt.transaction_status #{operator} #{GRANTED}
+        ), 0)
+      SQL
+    end
+
+    UPDATE_ASSIGNMENTS = [
```
toommz, line 99: Why not putting it at the top level?

toommz (approved):
I'm just not sure why we would want to pass an `organization_id` at some point 🤔 It adds some overhead to the code. Is this backfill supposed to be long lived and used for (only) one organization in the future?

lovrocolic:
> I'm just not sure why we would want to pass an `organization_id` at some point 🤔 It adds some overhead to the code. Is this backfill supposed to be long lived and used for (only) one organization in the future?

Original request was to perform backfill only for one client, but I made generic shape so that we can run it for any organization or for the all organizations. At the beginning we will trigger it only for one client who requested it.

app/jobs/database_migrations/backfill_prepaid_credit_breakdown_job.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+module DatabaseMigrations
+  # Backfills `prepaid_granted_credit_amount_cents` and
+  # `prepaid_purchased_credit_amount_cents` on invoices finalized before the
+  # wallet credit breakdown feature shipped (ING-320).
```
vincent-pochet: I think we should avoid adding "internal" references in the code. Maybe you can replace it with a link to the PR if its possible

app/jobs/database_migrations/backfill_prepaid_credit_breakdown_job.rb
```diff
@@ -0,0 +1,110 @@
...
+  #   * purchased = consumed from every other inbound transaction
+  #   * a column is written only when its amount is > 0, otherwise left nil
+  #   * only invoices whose customer is fully traceable are touched
+  #     (mirrors `customer.wallets.all?(&:traceable?)`)
+  #
+  # Selection uses ActiveRecord (readable gate); the write is a single set-based
+  # `update_all` per batch, so it scales (a few queries per batch instead of a
+  # few per invoice). The job re-enqueues itself until nothing is left. It is
+  # idempotent: filled invoices drop out of `candidates`, and the breakdown is
+  # recomputed deterministically from the same rows every time.
```
vincent-pochet: I think this part should be removed, it is just explaining the code in words and I think the code is self-explanatory enough 😉

app/jobs/database_migrations/backfill_prepaid_credit_breakdown_job.rb
```diff
@@ -0,0 +1,110 @@
...
+    def perform(organization_id = nil)
+      batch_ids = self.class.candidates(organization_id).limit(BATCH_SIZE).pluck(:id)
+
+      if batch_ids.empty?
+        Rails.logger.info("Finished backfilling prepaid credit breakdown")
+        return
+      end
+
+      Invoice.where(id: batch_ids).update_all(UPDATE_ASSIGNMENTS) # rubocop:disable Rails/SkipsModelValidations
+      self.class.perform_later(organization_id)
```
vincent-pochet: This job is re-enqueuing itself with the same argument, since it is marked as `unique :until_executed`, it will silently skip the enqueuing. I think you need to pass some extra argument, like a `batch_number` that will be incremented on each step

lib/tasks/migrations/backfill_prepaid_credit_breakdown.rake
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+# Backfills `prepaid_granted_credit_amount_cents` and
+# `prepaid_purchased_credit_amount_cents` on invoices finalized before the wallet
+# credit breakdown feature shipped (ING-320).
```
vincent-pochet: Same comment regarding the internal references

vincent-pochet (approved):

app/jobs/database_migrations/backfill_prepaid_credit_breakdown_job.rb
lovrocolic: It has to be below `bucket_sum_sql` definition

PR #5740, opened 2026-06-17
fix(events): Correct pay-in-advance fees for events with same timestamp

D1353L:
## Context

A customer reported incorrect amounts on `fee.created` webhooks when sending multiple pay-in-advance events for the same graduated charge within the same second.
Each pay-in-advance event is priced as a marginal cost: `amount(total_units) − amount(total_units − 1)`, where total_units comes from a count of events with `timestamp <= event.timestamp` (max_timestamp is the event's own timestamp). When N events share the same timestamp, they are all persisted before any fee is computed, so every concurrent fee calculation counts N and prices each event as the last unit of the batch.

Example: graduated pricing 0–2 units = 0€, 3+ = 2€, with 3 simultaneous events:

| | Event 1 | Event 2 | Event 3 | Total |
| --------------- | ---------------  | ------------------ | ------------------ | ------------------ |
| Expected | 0€  | 0€ | 2€ | 2€ |
| Actual (bug) | 2€ | 2€ | 2€ | 6€ |

## Description

In Events::Stores::PostgresStore, the pay-in-advance upper boundary is changed from timestamp <= T to a strict total order over events. Events sharing the boundary timestamp are tie-broken by ingestion order (created_at, id), so each event is assigned a distinct position

### DIFF
--- a/app/services/events/stores/postgres_store.rb
+++ b/app/services/events/stores/postgres_store.rb
@@ -420,21 +420,19 @@ module Events
         ).rows
       end
 
-      # NOTE: When aggregating in the context of a pay-in-advance event, the upper boundary
-      #       (max_timestamp) is the event's own timestamp. Multiple events can share this exact
-      #       timestamp (e.g. ingested in the same second), in which case they are tie-broken by
-      #       ingestion order (created_at, id) so that each one is assigned a distinct position.
-      #       Otherwise, all events sharing the timestamp would count each other and would all be
-      #       priced as the last unit of the batch.
+      # NOTE: For a pay-in-advance event, the upper boundary is the event's own timestamp.
+      #       Events sharing the same timestamp are tie-broken
+      #       by ingestion order (created_at, id) so each gets a distinct position. Otherwise
+      #       they would all count each other and be priced as the last unit of the batch.
       def apply_to_boundary(scope)
-        pay_in_advance_event_id = boundaries[:max_timestamp] && filters[:event]&.id
+        boundary_event = filters[:event] if boundaries[:max_timestamp]
 
-        if pay_in_advance_event_id
+        if boundary_event&.id
           scope.where(
             "events.timestamp < :to OR (events.timestamp = :to AND (events.created_at, events.id) <= " \
-            "(SELECT pia_event.created_at, pia_event.id FROM events pia_event WHERE pia_event.id = :pay_in_advance_event_id))",
+            "(SELECT boundary_event.created_at, boundary_event.id FROM events boundary_event WHERE boundary_event.id = :boundary_event_id))",
             to: applicable_to_datetime,
-            pay_in_advance_event_id:
+            boundary_event_id: boundary_event.id
           )
         else
           scope.to_datetime(applicable_to_datetime)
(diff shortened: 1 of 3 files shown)


app/services/events/stores/postgres_store.rb
```diff
@@ -389,6 +389,25 @@ def weighted_sum_breakdown(initial_value: 0)
...
+      #       Events sharing the same timestamp are tie-broken
+      #       by ingestion order (created_at, id) so each gets a distinct position. Otherwise
+      #       they would all count each other and be priced as the last unit of the batch.
+      def apply_to_boundary(scope)
+        boundary_event = filters[:event] if boundaries[:max_timestamp]
+
+        if boundary_event&.id
+          scope.where(
+            "events.timestamp < :to OR (events.timestamp = :to AND (events.created_at, events.id) <= " \
+            "(SELECT boundary_event.created_at, boundary_event.id FROM events boundary_event WHERE boundary_event.id = :boundary_event_id))",
```
vincent-pochet, line 433: This should be tested on a large dataset. Specially the `OR` close that could prevent PG from correctly using the index on timestamp.

vincent-pochet (approved):
Shouldn't we apply the same logic to the Clickhouse store?

app/services/events/stores/postgres_store.rb
D1353L: Good point, I'll check this 👍

D1353L: Tested with a dataset of 5M events locally - the index is used correctly and no performance degradation was noticed.

<details>
<summary>Explain Analyze</summary>

```
Aggregate  (cost=24782.18..24782.19 rows=1 width=8) (actual time=91.842..91.843 rows=1 loops=1)
  Buffers: shared hit=20010
  InitPlan 1 (returns $0,$1)
    ->  Index Scan using events_pkey on events boundary_event  (cost=0.43..8.45 rows=1 width=24) (actual time=0.026..0.027 rows=1 loops=1)
          Index Cond: (id = 'fde6dc2a-a562-499a-b85d-109a6239576d'::uuid)
          Buffers: shared hit=4
  ->  Bitmap Heap Scan on events  (cost=570.25..24755.10 rows=7455 width=0) (actual time=37.987..76.287 rows=500050 loops=1)
        Recheck Cond: (((external_subscription_id)::text = 'perf-pia-1782211218'::text) AND (organization_id = '11111111-2222-3333-4444-555555555555'::uuid) AND ((code)::text = 'perf_pia_code'::text) AND ("timestamp" >= '2026-05-19 09:40:18'::timestamp without time zone) AND (deleted_at IS NULL))
        Filter: (("timestamp" < '2026-06-23 09:40:18'::timestamp without time zone) OR (("timestamp" = '2026-06-23 09:40:18'::timestamp without time zone) AND (ROW(created_at, id) <= ROW($0, $1))))
        Heap Blocks: exact=12502
        Buffers: shared hit=20010
        ->  Bitmap Index Scan on idx_events_billing_lookup  (cost=0.00..568.38 rows=7455 width=0) (actual time=37.123..37.123 rows=500050 loops=1)
              Index Cond: (((external_subscription_id)::text = 'perf-pia-1782211218'::text) AND (organization_id = '11111111-2222-3333-4444-555555555555'::uuid) AND ((code)::text = 'perf_pia_code'::text) AND ("timestamp" >= '2026-05-19 09:40:18'::timestamp without time zone))
              Buffers: shared hit=7504
Planning Time: 0.100 ms
Execution Time: 91.897 ms
```
</details>

D1353L: > Shouldn't we apply the same logic to the Clickhouse store?

I'll prepare a separate PR

PR #5743, opened 2026-06-17
[ING-338] fix(events): batch CH delete per metric

mariohd:
## Context

When a billable metric is destroyed, Events::DeleteForMetricService walks every subscription that ever consumed it in pages of BATCH_SIZE (10_000) and issues an ALTER TABLE … DELETE per ClickHouse table, inlining the page's external_subscription_id list into the WHERE … IN (?) clause. At ~10k uuid-shaped ids the SQL string exceeds ClickHouse's default `max_query_size` (256 KiB) and the mutation fails with `Code: 62. DB::Exception: Max query size exceeded … (SYNTAX_ERROR)`. Reproduced on a metric tied to 86_814 subscriptions.

## Description

Introduces a CLICKHOUSE_BATCH_SIZE constant (BATCH_SIZE / 2) and slices the id list with each_slice before each ALTER TABLE … DELETE is emitted, so every CH query stays comfortably under the limit. The Postgres batch size stays at 10_000 — only ClickHouse needs the tighter slice because of its parser limit, and the PG throughput shouldn't pay for a CH-specific constraint. Mutations remain async (mutations_sync = 0), so the extra round-trips are enqueue-only and cheap.

Adds a spec context that stubs CLICKHOUSE_BATCH_SIZE to 1 with two subscriptions on the same charge, asserting the connection emits six DELETE statements (2 slices × 3 tables) instead of three. Lock prevents a regression to a single oversized query.

### DIFF
--- a/app/services/events/delete_for_metric_service.rb
+++ b/app/services/events/delete_for_metric_service.rb
@@ -35,6 +35,7 @@ module Events
     private
 
     BATCH_SIZE = 10_000
+    CLICKHOUSE_BATCH_SIZE = BATCH_SIZE / 2
     CLICKHOUSE_TABLES = {
       events_raw: :ingested_at,
       events_enriched: :enriched_at,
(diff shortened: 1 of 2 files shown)


app/services/events/delete_for_metric_service.rb
```diff
@@ -35,6 +35,7 @@ def call
     private
 
     BATCH_SIZE = 10_000
+    CLICKHOUSE_BATCH_SIZE = BATCH_SIZE / 2
```
vincent-pochet, line 38: To prevent too large queries, I guess we can even set this constant to `1_000`

vincent-pochet (approved):
🚀

toommz (approved):

app/services/events/delete_for_metric_service.rb
mariohd: Yeah, we can try to reduce this number even more in case this still happens. I believe 5_000 may be a good number.

PR #5745, opened 2026-06-18
test(quotes): expect 422 for non-approvable and non-voidable

toommz:
Fixing some failing specs that landed on `main`


osmarluz (approved):
🚀

PR #5749, opened 2026-06-18
perf(charge-filters): Match child filters per batch in cascade

rsempe:
`ChargeFilters::CascadeService` runs one job per changed filter. However, each job resolved its target by loading every child charge's entire filter set into Ruby (`find_child_filter`) and matching on `#to_h`, once per child. For an organization with many filters and many child subscriptions, a bulk restructure made this scale with children times filters squared.

Keep the per-filter job design, but resolve the matching child filter for a whole batch of children in two bounded queries instead of per child: narrow candidates via a value overlap in the database, then confirm the exact match in Ruby on that small set. This removes the per-child full filter load (memory) without introducing a query per child (N+1), so each job stays lightweight regardless of how many filters the charge has.

### DIFF
--- a/app/services/charge_filters/cascade_service.rb
+++ b/app/services/charge_filters/cascade_service.rb
@@ -15,21 +15,30 @@ module ChargeFilters
       super
     end
 
+    BATCH_SIZE = 1_000
+
     def call
-      charge.children
-        .joins(plan: :subscriptions)
-        .where(subscriptions: {status: %w[active pending]})
-        .distinct.find_each do |child_charge|
+      # NOTE: The cascade runs one job per changed filter
+      # Each job has a single target (filter_values), so the matching child filter
+      # is resolved for a whole batch of children in one query rather than loading
+      # every child's full filter set, which keeps each job lightweight.
+      child_ids.each_slice(BATCH_SIZE) do |ids|
+        matches = matching_child_filters(ids)
+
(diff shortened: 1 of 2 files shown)


app/services/charge_filters/cascade_service.rb
```diff
@@ -38,8 +47,38 @@ def call
...
     attr_reader :charge, :action, :filter_values, :old_properties, :new_properties, :invoice_display_name
 
-    def update_child_filter(child_charge)
-      child_filter = find_child_filter(child_charge)
+    def child_ids
+      @child_ids ||= charge.children
+        .joins(plan: :subscriptions)
+        .where(subscriptions: {status: %w[active pending]})
+        .distinct.pluck(:id)
+    end
```
toommz, line 55: Just wondering, why not remove the `pluck` here and doing the loading in batches (instead of `each_slice` on line `25`)? If the amount of ids can be really big it could help, if not let's keep it this way

toommz (approved):

vincent-pochet (approved):

app/services/charge_filters/cascade_service.rb
rsempe: thanks! `pluck` seems to be the lighter option here: the `join+distinct` runs once, then each batch is just a `WHERE id IN (…)` lookup. `find_in_batches` would re-run the `DISTINCT … JOIN subscriptions …` query every batch — more DB work. if ok, keeping it as-is for now 😅

PR #5753, opened 2026-06-18
feat(order-forms): lock when mutating quotes, order forms and orders

toommz:
This PR aims to protect the Quotes & Order forms features from concurrent state changes (for example, voiding a quote while someone else is approving the same quote).

I wrapped all the state changes in the feature services into a db advisory lock (`Quote::LockService`.


vincent-pochet (approved):
Nice one!

PR #5754, opened 2026-06-18
[ING-235] feat(wallets): add grants_target_top_up column

aquinofb:
## Context

Target recurring transaction rules always create a paid top-up. To let them grant free credits instead ([ING-235](https://linear.app/getlago/issue/ING-235)), the schema needs a flag. Shipping the column on its own, ahead of the code that reads it, lets it deploy first.

## Changes

- Add a nullable boolean `grants_target_top_up` to `recurring_transaction_rules`.

Nullable on purpose: target rules carry `true`/`false`, non-target rules stay `nil`. No backfill; absent values are read as `false`.

The behaviour that reads this column lands in #5755.

### DIFF
--- a/app/models/recurring_transaction_rule.rb
+++ b/app/models/recurring_transaction_rule.rb
@@ -113,6 +113,7 @@ end
 #  id                                  :uuid             not null, primary key
 #  expiration_at                       :datetime
 #  granted_credits                     :decimal(30, 5)   default(0.0), not null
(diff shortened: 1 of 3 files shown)


aquinofb:
Folded into #5755 — the migration is now the first commit there, since a single coherent PR is safe to ship (migrations run before app boot and the code tolerates nil). Keeping one PR per the ticket scope.

PR #5755, opened 2026-06-18
[ING-235] feat(wallets): allow target rules to grant credits

aquinofb:
## Context

Wallet recurring transaction rules of type `target` always create a paid top-up. Free credits were only reachable through fixed top-ups. This lets a target rule grant the gap-fill as free credits instead, over the REST API and the model layer ([ING-235](https://linear.app/getlago/issue/ING-235)).

## Changes

- Add a nullable boolean `grants_target_top_up` to `recurring_transaction_rules`. Target rules carry `true`/`false`; non-target rules stay `nil`.
- When `true` on a target rule, the gap-fill grants credits instead of paying for them, bypassing `paid_top_up_min` since no payment occurs.
- The validation allows `nil` so rows created before the column stay valid, and `nil` is read as `false`. No backfill.
- Create and update services default new target rules to `false` and normalise the flag on method transitions.
- Expose the flag on the REST create/update endpoints and the serializer (serialised as `false` when unset).

The migration that adds the column and the logic that reads it are kept as two commits here.

## Example

A target rule that tops a wallet back up to **$5** of credits every month (1 credit = $1 here):

```
method:                 target
interval:               monthly
target_ongoing_balance: 5
grants_target_top_up:   true
```

The customer has spent $3, so the ongoing balance is $2. The next monthly run fills the $3 gap:

| `grants_target_top_up` | Gap refill | Customer charged? |
| --- | --- | --- |
| `nil` / `false` _(before)_ | +$3 | Yes -- paid top-up |
| `true` _(after)_ | +$3 | No -- granted for free |

`nil` is what existing target rules carry (no backfill); the code reads it as `false`. The refill always tops up only to the target -- the flag just decides whether those credits are paid or free.

### DIFF
--- a/app/models/recurring_transaction_rule.rb
+++ b/app/models/recurring_transaction_rule.rb
@@ -113,6 +113,7 @@ end
 #  id                                  :uuid             not null, primary key
 #  expiration_at                       :datetime
 #  granted_credits                     :decimal(30, 5)   default(0.0), not null
+#  grants_target_top_up                :boolean
 #  ignore_paid_top_up_limits           :boolean          default(FALSE), not null
 #  interval                            :integer          default("weekly")
 #  invoice_requires_successful_payment :boolean          default(FALSE), not null
--- a/app/models/recurring_transaction_rule.rb
+++ b/app/models/recurring_transaction_rule.rb
@@ -15,6 +15,8 @@ class RecurringTransactionRule < ApplicationRecord
     source: :invoice_custom_section
 
   validates :transaction_name, length: {minimum: 1, maximum: 255}, allow_nil: true
+  validates :grants_target_top_up, inclusion: {in: [true, false]}, allow_nil: true, if: :target?
+  validates :grants_target_top_up, exclusion: {in: [true, false]}, unless: :target?
 
   STATUSES = [
     :active,
@@ -77,7 +79,9 @@ class RecurringTransactionRule < ApplicationRecord
 
   def compute_paid_credits(ongoing_balance:)
     if target?
-      compute_target_paid_credits(ongoing_balance:)
+      return 0.0 if grants_target_top_up?
+
+      compute_target_top_up_amount(ongoing_balance:)
     else
       paid_credits
     end
@@ -85,6 +89,8 @@ class RecurringTransactionRule < ApplicationRecord
 
   def compute_granted_credits
     if target?
+      return compute_target_top_up_amount(ongoing_balance: wallet.credits_ongoing_balance) if grants_target_top_up?
+
       0.0
     else
       granted_credits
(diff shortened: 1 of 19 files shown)


app/models/recurring_transaction_rule.rb
```diff
@@ -15,6 +15,8 @@ class RecurringTransactionRule < ApplicationRecord
     source: :invoice_custom_section
 
   validates :transaction_name, length: {minimum: 1, maximum: 255}, allow_nil: true
+  validates :grants_target_top_up, inclusion: {in: [true, false]}, allow_nil: true, if: :target?
```
aquinofb, line 18: Originally this came with a backfill migration; the open question was whether the validation was safe before the backfill ran. `allow_nil` solves it -- legacy `nil` rows stay valid and are treated as `false`, so no backfill is needed.

toommz: Surely stupid question, but right now, what we have in db will all return `false`, as they will all be `nil`, right? And it is intended, as no one never could create such a rule with `true` before this PR.

So why not just add a `default: false` in the migration and make it `not null`?

aquinofb: No no, this is actually a great question!

So, basically, the reason for keeping it without the default is that we need three states: target & enabled, target & disabled, and absent (non-targeted).

It's mainly for semantics: a `fixed` rule with a `nil` value means "this flag does not apply", while a `false` value means "this rule has a target top up, but it's not granted".

aquinofb: We use the `default: false` for the API, tho. However, having that in the DB could be misleading. If, in the future, we decide to backfill the data and set the column default to `false`, it's an additive change with zero impact on existing logic.

wdyt?

toommz: Oh ok, we need the third `nil` state 👍 The backfill would have been `set to false when null and is target`.

Now speaking about that, my memory starts to warm up (it took some time 😅) and I think Yohan worked on the same thing at some point => [PR](https://github.com/getlago/lago-api/pull/5582/changes)

aquinofb: Yes! I essentially took over that work and initially split it into two PRs: one to add the migration and backfill, and another for the logic. Then, I decided to fold all into a single PR, but didn't want to mess with that PR's history. So I just created a new one.

The code, now that I decided to keep a single PR, is 99.999% the same, but I applied the feedback 😅

toommz: Ok, and you don't want to backfill because of the historical comment on the previous PR about having the validation maybe running before the backfill is run?

aquinofb: No need to backfill because we changed the code to default to false, even though the historical data remains as is. We could backfill, but that wouldn't add much value for data persisted before this feature. Tho, we can backfill if we think it's useful, but that can happen at any time in the future.

toommz (approved):

app/models/recurring_transaction_rule.rb
```diff
@@ -77,29 +79,40 @@ def invoice_custom_section_params
...
+
       0.0
     else
       granted_credits
     end
   end
 
+  def grants_target_top_up?
+    super || false
+  end
```
groyoh: I don't think this is needed as Rails will cast `nil` into `false` in predicate methods.

groyoh (approved):

PR #5758, opened 2026-06-19
feat(payment-gated-subs): Support mid billing period downgrade from incomplete

osmarluz:
## Context

A payment-gated downgrade is gated to `incomplete` at the billing boundary while the previous plan keeps running and being billed. When the payment resolves mid-cycle, the previous (still-active, pay-in-advance) plan has to be terminated mid-cycle the same way an upgrade is: with a credit note for the unused prepaid days and its pay-in-arrears usage billed. Until now the downgrade activation simply marked the previous subscription terminated, so no credit note was issued and the customer kept paying for days they no longer used. This is restricted to downgrades coming from `incomplete`; regular downgrades are unchanged.

## Description

- `ActivateService#activate_for_downgrade`: when activating from `incomplete`, terminate the previous subscription through `TerminateService(rotation: true)` — issuing the credit note for unused days and billing arrears — instead of a bare `mark_as_terminated!`. The termination side effects (webhook, activity log, Hubspot sync) are left to `TerminateService` to avoid emitting them twice. The previous plan keeps being billed for the period while the downgrade is gated, so an invoice exists to credit when it is terminated.
- Add `EnsurePeriodInvoiceForCreditService`, called right before that termination, to verify the previous plan's pay-in-advance period invoice exists. It does not bill; when the invoice is missing it raises `MissingCreditableInvoiceError` so the payment-resolution job goes to the dead set for manual handling rather than terminating with nothing to credit. It no-ops for pay-in-arrears plans and when `on_termination_credit_note` is `skip`.
- Add `PayInAdvanceInvoiceIssuedService`, extracted from `TerminateService` and parameterized on a timestamp, so the verifier and `TerminateService`'s credit-note decision share the exact same "was the period invoice issued?" check.
- Rename the `upgrade:` flag to `rotation:` in `TerminateService` and `CreditNotes::CreateFromTermination`, since it now applies to both upgrades and downgrades.


app/services/subscriptions/activate_service.rb
```diff
@@ -108,11 +119,13 @@ def activate_for_downgrade
...
-        Utils::ActivityLog.produce(previous_subscription, "subscription.terminated")
+        unless from_incomplete
+          SendWebhookJob.perform_later("subscription.terminated", previous_subscription)
+          Utils::ActivityLog.produce(previous_subscription, "subscription.terminated")
 
-        if previous_subscription.should_sync_hubspot_subscription?
-          Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob.perform_later(subscription: previous_subscription)
+          if previous_subscription.should_sync_hubspot_subscription?
+            Integrations::Aggregator::Subscriptions::Hubspot::UpdateJob.perform_later(subscription: previous_subscription)
+          end
```
osmarluz, line 128: We have this check here now because ` Subscriptions::TerminateService ` (called from line 103 for a `from_incomplete` subscription) handles this logic

app/services/subscriptions/activate_service.rb
```diff
@@ -42,6 +42,8 @@ def gate_subscription
 
         SendWebhookJob.perform_later("subscription.incomplete", subscription)
         Utils::ActivityLog.produce(subscription, "subscription.incomplete")
+
+        bill_previous_subscription if subscription.payment_gated? && downgrade?
```
osmarluz, line 46: When the gated subscription comes from a downgrade we'll:

- Trigger the billing of the previous subscription
- When the subscription activates after a payment, we'll terminate the previous subscription using `Subscriptions::TerminateService` ( called from the `activate_for_downgrade` method below), which issues a credit note for the unconsumed period like we do for upgrades

app/services/subscriptions/activate_service.rb
```diff
@@ -95,7 +97,16 @@ def activate_for_downgrade
...
+      if from_incomplete
+        EnsurePeriodInvoiceForCreditService.call!(subscription: previous_subscription, timestamp:)
+
+        Subscriptions::TerminateService.call(
+          subscription: previous_subscription,
+          rotation: true
+        )
+      else
+        previous_subscription.mark_as_terminated!(timestamp)
+      end
```
osmarluz, line 109: We have 2 important points here:

- `EnsurePeriodInvoiceForCreditService` was added because the transition from `incomplete` to `active` can be really quick. In that case, we want to make sure the creditable invoice exists before `Subscriptions::TerminateService` runs
- If the downgrade is not coming `from_incomplete` we simply keep the existing logic and do `previous_subscription.mark_as_terminated!(timestamp)`. The previous subscription termination and activation of the downgrade happens at the begin of the new billing period and there's nothing to compensate for

app/services/subscriptions/ensure_period_invoice_for_credit_service.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    def call
+      return result unless subscription.plan.pay_in_advance?
+      return result if subscription.on_termination_credit_note_skip?
+      return result if creditable_invoice_issued?
+
+      raise MissingCreditableInvoiceError,
+        "subscription #{subscription.id} has no usable invoice for the period at #{timestamp.iso8601}: " \
+        "BillSubscriptionJob must finish successfully and produce the period invoice before " \
+        "Subscriptions::ActivationRules::Payment::ResolveJob is re-executed"
```
osmarluz, line 23: We run the logic we have on `Subscriptions::TerminateService` to decide if a credit note will be issued:

```ruby
def generate_credit_note_for_unconsumed_subscription?
  pay_in_advance? &&
    pay_in_advance_invoice_issued? &&
    on_termination_credit_note.in?(%i[credit refund offset])
end
```

and make it stricter by:

- Verifying the invoice is not voided
- Raising if a matching invoice is not found so we can monitor the dead jobs queue and retry after investigating

osmarluz:
For now we're not going forward with this solution to avoid billing the more expensive plan when an upgrade is in play. We'll monitor the subscriptions that end up in this scenario

PR #5759, opened 2026-06-19
misc(events): Retrieve and aggregate only matching filters

vincent-pochet:
## Context

Postgres usage pre-filtering included all of a charge's filters when its metric saw events, even filters that never matched a real event, leading to a lot of aggregation queries for filters with no usage at all. This is causing major performance issues for plans and subscriptions relying on a lot of charge filters.

## Description

- Add `PostgresStore#distinct_codes_and_property_combinations` to extract the distinct `[code, properties]` combinations seen in the period (the base store returns `nil`, so Clickhouse keeps the old "all filters" fallback).
- `BillingPeriodFilterService` now matches those combinations against charge filters and keeps only the matching ones. The unmatched combinations go to the default bucket. Recurring charges are still always included.
- `ChargeFilters::EventMatchingService` exposes the full `matching_charge_filters` list so the aggregation cascade assigns each event to its most specific filter.

### DIFF
--- a/app/services/charge_filters/event_matching_service.rb
+++ b/app/services/charge_filters/event_matching_service.rb
@@ -18,8 +18,12 @@ module ChargeFilters
         end
       end
 
-      # NOTE: An event could match multiple filters,
-      #       but we must take only the one matching the most properties
+      # NOTE: An event could match multiple filters. Expose the full list so callers
+      #       that need every candidate (e.g. usage pre-filtering) can let the
+      #       aggregation cascade assign the event to the most specific one.
+      result.matching_charge_filters = matching_filters
+
+      # NOTE: When a single filter must be picked, take the one matching the most properties
       result.charge_filter = matching_filters.max_by { |filter| filter.to_h.keys.size }
       result
     end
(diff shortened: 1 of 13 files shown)


app/services/charge_filters/event_matching_service.rb
```diff
@@ -29,8 +33,9 @@ def call
     attr_reader :charge, :event
 
     # NOTE: Exclude event properties not matching a billable metric filter
+    #       Use map over pluck so a preloaded filters association is reused without a query
     def applicable_event_properties
-      @applicable_event_properties ||= event.properties.slice(*charge.billable_metric.filters.pluck(:key))
+      @applicable_event_properties ||= event.properties.slice(*charge.billable_metric.filters.map(&:key))
```
toommz: Are we sure filters are preloaded here? We sure have the `Charge#filters` association cached, but not sure about `Charge#billable_metric->filters` 🤔

toommz: Ok so here that's the caller that is preloading, but other places calling this service might not. Do we have a pattern in the codebase to ensure that? For critical paths (as is the case here), raising if the association is not preloaded can be a sound tradeoff.

app/services/events/stores/postgres_store.rb
```diff
@@ -44,6 +44,29 @@ def distinct_charges_and_filters(codes: nil)
         scope.distinct.pluck(:charge_id, :charge_filter_id)
       end
 
+      # Returns the distinct [code, properties] combinations present in the events of the
+      # period. Only properties present in the filter_keys are considered, so the result holds
+      # only the dimensions that can be matched against charge filters.
+      # An empty hash represents the default (no filter) bucket.
+      def distinct_codes_and_property_combinations(codes:, filter_keys:)
+        scope = Event.where(external_subscription_id: subscription.external_id)
+          .where(organization_id: subscription.organization.id)
```
toommz:
```suggestion
          .where(organization_id: subscription.organization_id)
```

toommz (approved):
Looks good, but hard to tell from a reviewer point of view the impact on performances it will have. I can imagine that for organizations with lots of events it will help for sure.

PR #5760, opened 2026-06-19
fix: apply flat fee even when usage is zero

endenis:
## Context

For **Volume**, **Graduated**, **Graduated Percentage**, and **Prorated Graduated** charges the flat fee defined on the first tier was not applied when usage was zero.

Reported by the community in [getlago/lago#648](https://github.com/getlago/lago/issues/648).

## Description

1. Remove the `units.zero?` guards that prevented the first-tier flat fee from being applied on zero-usage periods. The per-unit amount naturally evaluates to `0` when there is no usage (`units × rate`), so removing the guards does not affect per-unit math.

2. `ProratedGraduatedService` is a bit different: its loop references a local `range` variable after it exits, so an early return with `result_with_flat_amount(0, 0, 0)` was added.

3. The fix in `graduated_service.rb` and `graduated_percentage_service.rb` is included because of the pay-in-advance edge case (see below)

## What changed

| Scenario | Before | After |
|---|---|---|
| First tier has flat fee, 0 usage | $0 ❌ | flat fee ✅ |
| First tier has flat fee, non-zero usage | flat fee + units ✅ | flat fee + units ✅ |
| First tier flat fee is $0, 0 usage | $0 ✅ | $0 ✅ |

⚠️ Before merging this PR we might need to change plans for some existing customers by adding an extra tier for them. This extra tier would be from zero to 1 unit and it would have flat fee set to zero.

## Pay-in-advance edge case

Pay-in-advance charges each event as a delta: cost(with event) − cost(without event). Before applying the fix for this case, on the first event with a $100 flat fee:

- cost(1 unit) = $102
- cost(0 units, exclude_event: true) = $100
- delta = $2 => the flat fee is never charged ❌

The fix forces to return 0 for the "without event" side at 0 units, using the `:exclude_event` flag already set by `ApplyPayInAdvanceChargeModelService`. With the fix:

- cost(1 unit) = $102
- cost(0 units, exclude_event: true) = $0
- delta = $102 ✅


app/services/charge_models/amount_details/range_graduated_percentage_service.rb
```diff
@@ -60,7 +56,6 @@ def units
           return to_value - (from_value.zero? ? 1 : from_value) + 1
         end
 
-        return to_value - from_value if to_value && total_units >= to_value
```
annvelents, line 63: why did we delete this line?

annvelents: it's already covered by lines 59-61

app/services/charge_models/graduated_service.rb
```diff
@@ -26,6 +26,10 @@ def adjacent_ranges?
     end
 
     def compute_amount
+      # First pay-in-advance event: delta = cost(1 unit) - cost(0 units, exclude_event: true).
+      # Avoid including the flat fee here so it does not get subtracted from the delta.
```
annvelents: can you please rephrase it to show that flat fee is excluded from 0-event

spec/scenarios/current_usage/by_charge_model/graduated_spec.rb
```diff
@@ -42,22 +42,22 @@
           # First call exercises the previously-crashing hydrate_non_persistable_fees path
           # for current usage when there are no persistable fees.
           fetch_current_usage(customer:)
-          expect(json[:customer_usage][:total_amount_cents]).to eq(0)
+          expect(json[:customer_usage][:total_amount_cents]).to eq(12_000)
```
annvelents: can we add here some line explaining 20% taxes? 🙈

annvelents (approved):
Looks good!
one small comment: can you please check that there are tests for two pay in advance fees:
that first includes the flat_fee, and the second - does not

endenis:
> Looks good! one small comment: can you please check that there are tests for two pay in advance fees: that first includes the flat_fee, and the second - does not

Yes, this test exists in the `spec/scenarios/pay_in_advance_charges_spec.rb` file.

PR #5762, opened 2026-06-22
fix(Clickhouse): Handle memory issues for pay in advance events

vincent-pochet:
## Context

Sometimes, a lot of `Events::PayInAdvanceJob` are failing in sequence with error similar to 

```
ActiveRecord::ActiveRecordError
Response code: 500: (ActiveRecord::ActiveRecordError)
Code: 241. DB::Exception: (total) memory limit exceeded: would use XXX GiB (attempt to allocate chunk of 0.00 B), current RSS: XXX GiB, maximum: XXX GiB. (MEMORY_LIMIT_EXCEEDED)
```

This kind of error happen when Clickhouse is receiving too many parallel intensive queries and is unable to scale up quickly.

Today the failing jobs end-up in the Sidekiq dead queue and have to be reprocessed manually.

## Description

This PR catches the errors, and add a retry logic on the `Events::PayInAdvanceJob`.

The retry uses a custom wait lambda that keeps the increasing backoff (`executions**4`) but adds an absolute random jitter (`RETRY_JITTER = 1..15 seconds`). The main goal is to give Clickhouse the time to recover and to avoid retrying all parallel failed attempt in a too close delayl

### DIFF
--- a/app/jobs/fees/create_pay_in_advance_job.rb
+++ b/app/jobs/fees/create_pay_in_advance_job.rb
@@ -4,7 +4,21 @@ module Fees
   class CreatePayInAdvanceJob < ApplicationJob
     queue_as :default
 
+    # Random jitter (in seconds) added to every retry so parallel failed jobs
+    # don't retry in lockstep and overload Clickhouse again at the same instant.
+    RETRY_JITTER = 1..15
+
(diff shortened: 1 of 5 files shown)


toommz (approved):
Nice 👍

PR #5764, opened 2026-06-22
feat(fc-units-override): wire write paths to per-sub units overrides

ancorcruz:
## Context

Seat-based pay-in-advance plans see frequent unit-count changes per customer. Every such change previously cloned the entire plan and produced a per-subscription FixedCharge, even when the only difference was the unit count. The earlier read-path work landed a new `Subscription::FixedChargeUnitsOverride` table and a resolver that applies per-sub overrides on top of the plan's fixed_charges. This PR wires the three write surfaces to that table so units-only changes no longer mint override plans.

## Description

Three write paths now detect units-only payloads and route them to the override table instead of cloning the plan:

- `POST /subscriptions` — units-only `plan_overrides.fixed_charges` writes override rows directly; the subscription stays on the parent plan.
- `PUT /subscriptions/:id` — same detection on the partial payload.
- `PUT /subscriptions/:id/fixed_charges/:code` — units-only params take the fast path; everything else (display name, properties, tax codes) continues through the legacy plan-override flow.

When a later change forces a full plan override (any non-units field, or a mixed `plan_overrides` payload), the existing override rows are promoted: their units are merged into the `fixed_charges` params handed to `Plans::OverrideService` so the resulting override plan is built with the customer's actual seat counts already in place. Without this, the override rows would orphan against the new override fixed_charges and the customer would snap back to plan-level units on the next billing cycle.

Shared write logic lives in `Subscriptions::FixedChargeUnitsOverrides::WriteService`: it writes the override row, emits the `FixedChargeEvent`, and dispatches the mid-period delta billing job when `apply_units_immediately` is set on a pay-in-advance fixed_charge. The dispatch is gated on `subscription.active?` so payment-gated subscriptions defer to the activation flow's catch-up billing.

Also fixed: the partial-update path on `Subscriptions::UpdateService`. When a subscription is on an override plan and the caller sends a partial `plan_overrides.fixed_charges` payload, the override plan's existing fixed_charges are re-hydrated and the caller's partials are overlaid by id. Without this, missing `properties` failed validation with a 422, unlisted fixed_charges were silently sanitized away, and unknown ids cascaded to a 500.

### DIFF
--- a/app/services/subscriptions/concerns/fixed_charge_units_override_promotion_concern.rb
+++ b/app/services/subscriptions/concerns/fixed_charge_units_override_promotion_concern.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  module Concerns
+    # Promotes a subscription's existing units override rows into the
+    # `fixed_charges` params passed to `Plans::OverrideService` so a
+    # subsequent change that requires a plan override (any non-units field)
+    # builds the override plan with the customer's actual seat counts
+    # already in place.
+    #
+    # Without this promotion the override rows would be orphaned by the
+    # plan override — they reference the parent fixed_charges, not the
+    # newly created override fixed_charges — and the customer would
+    # silently snap back to the plan-level units on the next billing
+    # cycle.
+    #
+    # The host service is expected to expose a `subscription` reader.
+    module FixedChargeUnitsOverridePromotionConcern
+      extend ActiveSupport::Concern
+
+      private
+
+      # Discards the subscription's override rows and returns a
+      # fixed_charges params array combining the caller's existing entries
+      # with a synthetic entry per discarded override (units carried
+      # forward). The caller's explicit entry wins when both an existing
+      # entry and an override row exist for the same fixed_charge.
+      # When the subscription has no override rows, returns the input
+      # unchanged.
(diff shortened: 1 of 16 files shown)


app/services/subscriptions/concerns/fixed_charge_units_override_promotion_concern.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+
+        overrides.each do |override|
+          params_by_id[override.fixed_charge_id] ||= {
+            id: override.fixed_charge_id,
+            units: override.units
+          }
+        end
+
+        overrides.each(&:discard!)
```
brunomiguelpinto, line 46: we are discarding, but the name lies about the effect. Consider renaming to something like promote_and_consume_units_overrides! or splitting the discard into another step

ancorcruz:
closing this to handle the review in smaller chuncks.

:pr: 1- https://github.com/getlago/lago-api/pull/5781
:pr: 2- https://github.com/getlago/lago-api/pull/5782
:pr: 3- https://github.com/getlago/lago-api/pull/5783
:pr: 4- https://github.com/getlago/lago-api/pull/5784
:pr: 5- https://github.com/getlago/lago-api/pull/5785
:pr: 6- https://github.com/getlago/lago-api/pull/5786

PR #5765, opened 2026-06-22
feat(payment-gated-subs): always serialize fields

ancorcruz:
## Context

The subscription serializer gated cancellation_reason, activated_at, and activation_rules behind the payment_gated_subscriptions feature flag. As the feature reaches public launch, these fields belong in the standard subscription payload for every consumer, and the published OpenAPI spec and regenerated client libraries expect them unconditionally rather than depending on a per-organization flag.

## Description

- Move cancellation_reason and activated_at into the core serialized payload so they are always present.
- Always serialize activation_rules, using CollectionSerializer in place of the inline map.
- Drop the payment_gated_subscriptions feature-flag branch and the spec example asserting the fields were hidden when the flag was disabled.

### DIFF
--- a/app/controllers/concerns/subscription_index.rb
+++ b/app/controllers/concerns/subscription_index.rb
@@ -23,7 +23,7 @@ module SubscriptionIndex
 
     if result.success?
       subscriptions = result.subscriptions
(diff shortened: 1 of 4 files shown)


osmarluz (approved):
🚀

PR #5774, opened 2026-06-24
feat(products): add rate card REST endpoints

rsempe:
## Description

- Add rate card REST endpoints (index, show, create, update, destroy), addressed by `code`.
- Support creating a card with its rates nested in a single call.
- Filter the list by product, filter slice or code, and expose the card's currently active rate.

### DIFF
--- a/app/controllers/api/v2/rate_cards_controller.rb
+++ b/app/controllers/api/v2/rate_cards_controller.rb
@@ -1,0 +1,149 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    class RateCardsController < Api::BaseController
+      def create
+        if create_params.key?(:product_filter_code) && product && product_filter.nil?
+          return not_found_error(resource: "product_filter")
+        end
+
+        result = ::RateCards::CreateService.call(
+          product:,
+          params: create_params
+            .except(:product_code, :product_filter_code)
+            .to_h.deep_symbolize_keys
+            .merge(product_filter_id: product_filter&.id)
+        )
+
+        if result.success?
+          render_rate_card(result.rate_card)
+        else
+          render_error_response(result)
+        end
+      end
+
+      def update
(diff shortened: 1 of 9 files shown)


app/controllers/api/v2/rate_cards_controller.rb
```diff
@@ -0,0 +1,153 @@
...
+            :billing_interval_count,
+            :billing_interval_unit,
+            :applied_pricing_unit_conversion_rate,
+            {rate_properties: {}}
+          ]
+        )
+      end
+
+      def update_params
+        params.require(:rate_card).permit(
```
rsempe, line 130: `code` is not present here - checking with product if it should be possible to update the code of a rate card when not linked to plan or subscription

rsempe: Updated - We prefer to allow it 👍

app/controllers/api/v2/rate_cards_controller.rb
```diff
@@ -0,0 +1,153 @@
...
+            product_code: params[:product_code],
+            product_filter_code: params[:product_filter_code]
+          }
+        )
+
+        if result.success?
+          render(
+            json: ::CollectionSerializer.new(
+              result.rate_cards.includes(:product, :product_filter, :rates),
+              ::V1::RateCardSerializer,
```
endenis, line 76: Should the `active_rate` be included when we serialize the collection? If yes, when adding it we have to be careful to not introduce N+1

endenis (approved):

app/controllers/api/v2/rate_cards_controller.rb
rsempe: Thanks - Let's try to avoid such nested resources on a collection - At least, I'm not adding it for now :)

PR #5778, opened 2026-06-24
feat(multi-entity): add logic to handle eu taxes when billing entity changes

lovrocolic:
## Context
Currently, it is possible to attach multiple billing entities to a customer. That said, a customer can have different billing entities assigned to different billing objects.

## Description
This PR handles the case when a customer's billing entity changes, which relates to the EU tax management system. In that case, we should either remove the old taxes from the customer (if the new billing entity is not EU-related) or recompute the taxes.


mariohd (approved):
Looks good

PR #5779, opened 2026-06-24
feat(api): add the /api/v2 namespace and pricing-engine guards

rsempe:
## Context

An organization is either on the product catalog or on legacy v1 pricing, never both. The engine is an organization-level property, so it must not be a per-plan payload choice.

## Description

The rule is enforced where it cannot be bypassed, as payload shape rather than blanket endpoint blocks:

- `Plans::CreateService` and `UpdateService` reject the legacy pricing fields (`interval`, `amount_cents`, `pay_in_advance`, `charges`, `fixed_charges`) for catalog organizations and plans with `legacy_billing_disabled`; everything else (name, description, currency…) stays editable, and catalog plans stay deletable.
- `Charge` and `FixedCharge` validate that their plan is not a catalog plan — the deepest layer, closing internal paths such as subscription plan overrides that no endpoint gate can see.
- The `ForbidsLegacyBilling` door-level 403 remains only on the pure-legacy charge, fixed-charge and charge-filter endpoints and mutations, where every action is illegal for a catalog org.
- `RequiresProductCatalog` gates the catalog surface on both APIs: v2 REST endpoints, catalog GraphQL mutations and resolvers (403 `feature_unavailable` without the flag).
- Each engine keeps its own strict GraphQL write shape, mirroring the REST v1/v2 split: `CreatePlanInput`/`UpdatePlanInput` stay unchanged for legacy, and catalog organizations get dedicated `createCatalogPlan`/`updateCatalogPlan` mutations with a catalog-only payload exposing the plan currency as `currency`. `destroyPlan` is shared. Rate cards attach to catalog plans only (`legacy_plan` otherwise).
- Add the `/api/v2` namespace, falling back to v1 for every path it does not define, kept exhaustive by a routing parity spec, plus a v2 plans endpoint with the same catalog-only payload.
- Derive `Plan#pricing_type` from the organization flag and remove it from the v1 payload.

Defaulting new organizations to the catalog moves to #5871, alongside the v2 subscription surface: enabling it here would let a fresh organization create catalog plans while its subscriptions still go through the ungated v1 endpoint into a runtime that cannot bill them.


config/routes.rb
```diff
@@ -217,10 +40,15 @@
       resources :rate_cards, param: :code, code: /.*/, only: %i[index show create update destroy] do
         resources :rates, param: :code, code: /.*/, only: %i[index show create update destroy], controller: "rate_cards/rates"
       end
-      resources :plans, param: :code, code: /.*/, only: [] do
+      resources :plans, param: :code, code: /.*/, only: %i[show create update] do
```
endenis: Does this mean that the `index` action returns a different response shape than `show/create/update`? Serializers in v1 and v2 are different 🤔

endenis (approved):

config/routes.rb
rsempe: Good catch 👍 Adding a v2 index!

PR #5780, opened 2026-06-24
fix(payment-gated-subs): Restrict subscription termination when next subscription is incomplete and upgrade flag is false

osmarluz:
## Context

`Subscriptions::TerminateService` terminated a subscription even when it had a `next_subscription` still in the `incomplete` state, silently canceling that scheduled change while it was awaiting payment/activation.

## Description

Reject the termination up front with a `not_allowed_failure` (code `next_subscription_incomplete`) when the subscription has an incomplete `next_subscription`. The check is skipped on the upgrade path, where the incoming subscription is the `next_subscription` and is expected to be incomplete.

### DIFF
--- a/app/services/subscriptions/terminate_service.rb
+++ b/app/services/subscriptions/terminate_service.rb
@@ -16,8 +16,8 @@ module Subscriptions
 
     def call
       return result.not_found_failure!(resource: "subscription") if subscription.blank?
-      return result.not_allowed_failure!(code: "subscription_incomplete") if subscription.incomplete?
-      return result.not_allowed_failure!(code: "next_subscription_incomplete") if !upgrade && subscription.next_subscription&.incomplete?
+      return result.single_validation_failure!(error_code: "subscription_incomplete") if subscription.incomplete?
+      return result.single_validation_failure!(error_code: "next_subscription_incomplete") if !upgrade && subscription.next_subscription&.incomplete?
 
       ActiveRecord::Base.transaction do
         if subscription.pending?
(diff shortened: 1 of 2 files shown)


app/services/subscriptions/terminate_service.rb
```diff
@@ -17,6 +17,7 @@ def initialize(subscription:, async: true, upgrade: false, on_termination_credit
     def call
       return result.not_found_failure!(resource: "subscription") if subscription.blank?
       return result.not_allowed_failure!(code: "subscription_incomplete") if subscription.incomplete?
+      return result.not_allowed_failure!(code: "next_subscription_incomplete") if !upgrade && subscription.next_subscription&.incomplete?
```
toommz: I think `not_allowed_failure` returns an `HTTP Method Not Allowed` error, which is not really what the error is about. We have some misuses of this method in the codebase (the line above your diff is a good example 😅).

Should be `single_validation_failure!` here.

toommz (approved):

app/services/subscriptions/terminate_service.rb
osmarluz: @toommz thanks, I'll change that

ancorcruz: true, we also use it on the line on top and that is not the 405 purpose. Thanks for raising this @toommz

ancorcruz (approved):
@osmarluz lets fix the line 19 as well in a separate PR.

osmarluz:
> @osmarluz lets fix the line 19 as well in a separate PR.

@ancorcruz I fixed that here already, should we proceed that way?

ancorcruz:
@osmarluz cool, thanks. 🚀

PR #5781, opened 2026-06-24
feat(fc-units-override): wire dedicated endpoint to per-sub units override

ancorcruz:
PUT /api/v1/subscriptions/:external_id/fixed_charges/:code previously always cloned the plan and created a fixed_charge override on every units change. Customers on seat-based plans saw one duplicate Plan plus one duplicate FixedCharge per per-customer edit, every edit.

When the inbound params only carry units (and optionally apply_units_immediately) and the subscription is on the parent plan, the service now writes one Subscription::FixedChargeUnitsOverride row keyed by (subscription, fixed_charge), emits a FixedChargeEvent against the parent fixed_charge with the override units, and — when apply_units_immediately is set on a pay-in-advance fixed_charge — dispatches Invoices::CreatePayInAdvanceFixedChargesJob after_commit. The dispatch is gated on subscription.active? so payment-gated subscriptions defer to the activation flow's catch-up billing rather than minting a delta invoice before payment is confirmed.

Everything else (display name, properties, tax codes, mixed payloads, override-plan subs) keeps going through the legacy plan-cloning path.

### DIFF
--- a/app/services/subscriptions/update_or_override_fixed_charge_service.rb
+++ b/app/services/subscriptions/update_or_override_fixed_charge_service.rb
@@ -37,7 +37,7 @@ module Subscriptions
     attr_reader :subscription, :fixed_charge, :params, :subscription_plan_parent_present
 
     def units_only_change?
-      return false if subscription.plan.parent_id
+      return false if subscription_plan_parent_present
 
       units_only_fixed_charge_params?(params)
     end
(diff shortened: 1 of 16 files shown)


app/services/subscriptions/update_or_override_fixed_charge_service.rb
```diff
@@ -41,6 +36,51 @@ def call
 
     attr_reader :subscription, :fixed_charge, :params, :subscription_plan_parent_present
 
+    def units_only_change?
+      return false if subscription.plan.parent_id
```
rsempe:
```suggestion
      return false if subscription_plan_parent_present
```

rsempe (approved):

brunomiguelpinto (approved):

PR #5782, opened 2026-06-24
feat(fc-units-override): wire subscription creation to per-sub units override

ancorcruz:
POST /api/v1/subscriptions with units-only plan_overrides.fixed_charges previously cloned the plan plus every fixed_charge for the new subscription. For seat-based signup flows that's one duplicate Plan plus N duplicate FixedCharges per signup, every signup — the highest-volume code path on the feature.

When the units-only predicate matches (plan_overrides contains only fixed_charges and each entry has only id, units, and optionally apply_units_immediately), Subscriptions::CreateService now attaches the subscription to the parent plan, writes one
Subscription::FixedChargeUnitsOverride row per entry, and emits a FixedChargeEvent against each fixed_charge with the override units.

Other plan_overrides shapes (mixed entries, amount_cents, charges, etc.) keep going through today's Plans::OverrideService.

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -3,6 +3,7 @@
 module Subscriptions
   class CreateService < BaseService
     include Subscriptions::Concerns::BillingEntityResolutionConcern
+    include Subscriptions::Concerns::FixedChargeUnitsOverrideDetectionConcern
 
     Result = BaseResult[:subscription, :payment_method]
 
@@ -131,7 +132,7 @@ module Subscriptions
       new_subscription = Subscription.new(
         organization_id: customer.organization_id,
         customer:,
-        plan: params.key?(:plan_overrides) ? override_plan(plan) : plan,
+        plan: target_plan_for_new_subscription,
         subscription_at:,
         name:,
         external_id:,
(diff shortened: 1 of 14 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -236,6 +243,34 @@ def override_plan(plan)
...
+        units_only_fixed_charges_plan_overrides?(params[:plan_overrides])
+    end
+
+    def create_fixed_charge_units_overrides(subscription)
+      params[:plan_overrides][:fixed_charges].each do |entry|
+        entry = entry.to_h.symbolize_keys
+
+        ::Subscription::FixedChargeUnitsOverride.create!(
+          subscription:,
+          fixed_charge: plan.fixed_charges.find(entry[:id]),
```
rsempe: It will probably raise a `RecordNotFound` exception here if `id` is not found, no?

We can probably use something like this:
```
fixed_charge = plan.fixed_charges.find_by(id: entry[:id])
return result.not_found_failure!(resource: "fixed_charge") unless fixed_charge
```

rsempe: Can you add a test with a not found fixed charge maybe as well?

ancorcruz: fixed, test added, thanks.

rsempe (approved):

PR #5783, opened 2026-06-24
feat(fc-units-override): wire subscription update to per-sub units override

ancorcruz:
PUT /api/v1/subscriptions/:external_id with units-only plan_overrides.fixed_charges previously cloned the subscription's plan on every units change, the same wasted cost the create path was paying on signup.

When the units-only predicate matches and the subscription is on the parent plan, Subscriptions::UpdateService now writes one Subscription::FixedChargeUnitsOverride row per entry, emits a FixedChargeEvent, and fires the mid-period delta billing job for pay-in-advance + apply_units_immediately combinations on active subs.

A separate but related fix on the legacy path: when a subscription is already on an override plan and the caller sends a partial plan_overrides.fixed_charges payload (e.g. units only), the previous handler dropped any unlisted fixed_charges and returned a 422 on missing properties. The handler now re-hydrates the override plan's fixed_charges into the payload, overlays the caller's partials by id, and rejects unknown ids with a not_found failure so the caller gets a 4xx instead of a 500.

### DIFF
--- a/app/services/subscriptions/update_service.rb
+++ b/app/services/subscriptions/update_service.rb
@@ -3,6 +3,7 @@
 module Subscriptions
   class UpdateService < BaseService
     include Subscriptions::Concerns::BillingEntityResolutionConcern
+    include Subscriptions::Concerns::FixedChargeUnitsOverrideDetectionConcern
 
     Result = BaseResult[:subscription, :payment_method]
 
@@ -73,7 +74,11 @@ module Subscriptions
           subscription.billing_entity = new_billing_entity
         end
 
-        subscription.plan = handle_plan_override.plan if params.key?(:plan_overrides)
+        if units_only_plan_overrides_change?
+          apply_units_only_plan_overrides
+        elsif params.key?(:plan_overrides)
+          subscription.plan = handle_plan_override.plan
+        end
 
         if params.key?(:usage_thresholds)
           UpdateUsageThresholdsService.call!(subscription:, usage_thresholds_params: params[:usage_thresholds], partial: false)
@@ -165,7 +170,7 @@ module Subscriptions
       if current_plan.parent_id
         Plans::UpdateService.call!(
           plan: current_plan,
-          params: params[:plan_overrides].to_h.with_indifferent_access
+          params: plan_update_params_with_full_fixed_charges(current_plan)
         )
       else
         Plans::OverrideService.call!(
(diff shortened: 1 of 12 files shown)


app/services/subscriptions/update_service.rb
```diff
@@ -176,6 +181,73 @@ def handle_plan_override
...
+    def plan_update_params_with_full_fixed_charges(plan)
+      payload = params[:plan_overrides].to_h.with_indifferent_access
+      return payload unless payload.key?(:fixed_charges)
+
+      fixed_charges_by_id = plan.fixed_charges.index_by(&:id)
+
+      overlays_by_id = payload[:fixed_charges].each_with_object({}) do |entry, acc|
+        entry_hash = entry.to_h.with_indifferent_access
+        id = entry_hash[:id]
+        result.not_found_failure!(resource: "fixed_charge") unless fixed_charges_by_id.key?(id)
```
rsempe: should we `raise_if_error!`?

app/services/subscriptions/update_service.rb
```diff
@@ -176,6 +181,73 @@ def handle_plan_override
...
+        override.save!
+
+        FixedCharges::EmitEventsService.call!(
+          fixed_charge:,
+          subscription:,
+          apply_units_immediately:,
+          timestamp:
+        )
+
+        if apply_units_immediately && fixed_charge.pay_in_advance?
```
rsempe: `if subscription.active?` maybe as well?

To be consistent with here:
https://github.com/getlago/lago-api/blob/main/app/services/subscriptions/update_service.rb#L96

app/services/subscriptions/update_service.rb
```diff
@@ -176,6 +181,73 @@ def handle_plan_override
...
+        params.key?(:plan_overrides) &&
+        units_only_fixed_charges_plan_overrides?(params[:plan_overrides])
+    end
+
+    def apply_units_only_plan_overrides
+      timestamp = Time.current.to_i
+
+      params[:plan_overrides][:fixed_charges].each do |entry|
+        entry = entry.to_h.symbolize_keys
+        fixed_charge = subscription.plan.fixed_charges.find(entry[:id])
```
rsempe: same `find_by` here?

app/services/subscriptions/update_service.rb
ancorcruz: yes, done it and added a test for this.

app/services/subscriptions/update_service.rb
ancorcruz: yes, done it and added a test for this.

app/services/subscriptions/update_service.rb
ancorcruz: yes, done it and added a test for this.

rsempe (approved):

PR #5784, opened 2026-06-24
feat(fc-units-override): promote sub-level overrides on plan-overridefallthrough

ancorcruz:
A subscription that already carries Subscription::FixedChargeUnitsOverride rows can later receive a change that forces a full plan override (any non-units field on the dedicated endpoint, or a mixed plan_overrides payload on subscription update). Without intervention the override rows would be orphaned by the override plan — they reference the parent fixed_charges, not the new override fixed_charges — and the customer would silently snap back to plan-level units on the next billing cycle.

Subscriptions::UpdateService and
Subscriptions::UpdateOrOverrideFixedChargeService now capture the existing override rows before falling through to
Plans::OverrideService, discard them, and merge the captured units into the fixed_charges params handed to the override service so the override plan is built with the customer's actual seat counts already in place. Promotion lives on the calling services so Plans::OverrideService stays generic.

### DIFF
--- a/app/services/subscriptions/concerns/fixed_charge_units_override_detection_concern.rb
+++ b/app/services/subscriptions/concerns/fixed_charge_units_override_detection_concern.rb
@@ -55,6 +55,39 @@ module Subscriptions
         (entry.keys - PLAN_OVERRIDES_FIXED_CHARGE_ALLOWED_KEYS).empty?
       end
 
+      # When a subscription that carries units override rows receives a change
+      # that requires a plan override (any non-units field), the override rows
+      # must be promoted into the resulting override plan or the customer
+      # silently snaps back to the plan-level units. Each calling service
+      # invokes this before falling through to `Plans::OverrideService`:
+      # discards the override rows on the subscription and returns a
+      # fixed_charges params array combining the caller's existing entries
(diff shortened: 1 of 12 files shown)


app/services/subscriptions/concerns/fixed_charge_units_override_detection_concern.rb
```diff
@@ -55,6 +55,39 @@ def units_only_fixed_charges_entry?(entry)
...
+      # must be promoted into the resulting override plan or the customer
+      # silently snaps back to the plan-level units. Each calling service
+      # invokes this before falling through to `Plans::OverrideService`:
+      # discards the override rows on the subscription and returns a
+      # fixed_charges params array combining the caller's existing entries
+      # with a synthetic entry per discarded override (units carried forward)
+      # so `Plans::OverrideService` builds the override plan with the
+      # customer's actual seat counts already in place. The caller's explicit
+      # params win when both an existing entry and an override row exist for
+      # the same fixed_charge.
```
rsempe: can we avoid such AI generated comments? at least some more digest.. 🙏

ancorcruz: done it. thanks

rsempe (approved):

PR #5786, opened 2026-06-24
refactor(fc-units-override): extract WriteService and split detection…

ancorcruz:
… concern

The dedicated subscription fixed_charge endpoint and the subscription update endpoint both wrote one Subscription::FixedChargeUnitsOverride row, emitted a FixedChargeEvent, and dispatched the mid-period billing job. The two services carried near-identical copies of that logic.

Extract the shared write into
Subscriptions::FixedChargeUnitsOverrides::WriteService. Both call sites now route through it.

The detection concern (params-shape predicates) had grown to include the promotion helper that captures override rows and merges them into Plans::OverrideService params. The name no longer matched the responsibility, and mixing a pure params-shape predicate set with a DB-mutating helper blurred the concern's boundary.

Move the promotion helper into
FixedChargeUnitsOverridePromotionConcern; the detection concern keeps only the predicates.

## Roadmap Task

## Context

## Description

### DIFF
--- a/app/services/subscriptions/concerns/fixed_charge_units_override_promotion_concern.rb
+++ b/app/services/subscriptions/concerns/fixed_charge_units_override_promotion_concern.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Subscriptions
+  module Concerns
+    # Promotes a subscription's existing units override rows into the
+    # `fixed_charges` params passed to `Plans::OverrideService` so a
+    # subsequent change that requires a plan override (any non-units field)
+    # builds the override plan with the customer's actual seat counts
+    # already in place.
+    #
+    # Without this promotion the override rows would be orphaned by the
+    # plan override — they reference the parent fixed_charges, not the
+    # newly created override fixed_charges — and the customer would
+    # silently snap back to the plan-level units on the next billing
+    # cycle.
+    #
+    # The host service is expected to expose a `subscription` reader.
+    module FixedChargeUnitsOverridePromotionConcern
+      extend ActiveSupport::Concern
+
+      private
+
+      # Discards the subscription's override rows and returns a
+      # fixed_charges params array combining the caller's existing entries
+      # with a synthetic entry per discarded override (units carried
+      # forward). The caller's explicit entry wins when both an existing
+      # entry and an override row exist for the same fixed_charge.
+      # When the subscription has no override rows, returns the input
+      # unchanged.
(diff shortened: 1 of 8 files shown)


app/services/subscriptions/concerns/fixed_charge_units_override_promotion_concern.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+
+        overrides.each do |override|
+          params_by_id[override.fixed_charge_id] ||= {
+            id: override.fixed_charge_id,
+            units: override.units
+          }
+        end
+
+        overrides.each(&:discard!)
```
brunomiguelpinto, line 46: promote_units_overrides_to_fixed_charges_params i would just change the name we are discarding stuff inside

ancorcruz: Thanks @brunomiguelpinto  — I'd like to keep the current name. My reasoning:

"Promote" describes the conceptual move (sub-level override row → plan-level override plan). The discard isn't a hidden side effect we're glossing over; it's a necessary consequence of promotion — keeping the row around after promoting would leave two sources of truth for the same units, and they'd drift by the next billing cycle.

The method's comment already opens with "Discards the subscription's override rows and returns…" so anyone reading the method sees the discard upfront, right where we'd want the implementation note to live.

Open to revisiting if you can point at a case where a reader could be genuinely misled — happy to brainstorm alternatives that preserve the intent if you have any suggestion.

app/services/subscriptions/fixed_charge_units_overrides/write_service.rb
```diff
@@ -0,0 +1,65 @@
...
+    # job after the surrounding transaction commits.
+    #
+    # The service is the shared building block for the two write surfaces
+    # that record units-only overrides mid-cycle: the dedicated subscription
+    # fixed_charge endpoint (`UpdateOrOverrideFixedChargeService`) and the
+    # subscription update endpoint (`UpdateService`). Subscription creation
+    # has different lifecycle constraints (events emit through the
+    # activation path) and writes override rows directly without this
+    # service.
+    #
```
rsempe: can we clean that non-digest comments?

app/services/subscriptions/concerns/fixed_charge_units_override_promotion_concern.rb
brunomiguelpinto: ok

brunomiguelpinto (approved):

PR #5787, opened 2026-06-24
[ISSUE-1965] fix(api_logs): allow PATCH http_method

mariohd:
## Summary

Adds `patch` to the `http_method` enum on the ClickHouse `api_logs` table so PATCH requests reaching `Utils::ApiLog#produce` are persisted instead of being rejected. ClickPipes was routing them to its dead letter queue on cloud, and the Kafka engine queue table was refusing them on self-hosted, so real PATCH endpoints in the API were silently absent from the logs.

## Changes

`app/models/clickhouse/api_log.rb` adds `patch: 5` to `HTTP_METHODS`. The GraphQL `Types::ApiLogs::HttpMethodEnum` derives its values from this constant, so it picks up `patch` automatically without an explicit edit.

`db/clickhouse_migrate/20260624144347_add_patch_to_api_logs_http_method.rb` is a new ClickHouse migration. It extends the enum on the main `api_logs` MergeTree via `ALTER TABLE … MODIFY COLUMN`, then drops and recreates `api_logs_queue` together with the `api_logs_mv` materialized view that bridges them — Kafka engine tables don't support `MODIFY COLUMN`, so a drop-and-recreate is the only available path. The underlying Kafka topic is untouched and the consumer group resumes from its committed offset on the next start, so no events are lost. The `down` path is symmetric: it ALTERs `api_logs` back to the original 4-value enum and recreates the queue + MV with the original enum signature. The `up`/`down` were both exercised against the test ClickHouse (migrate → rollback → migrate again).

`db/clickhouse_migrate/cloud/04_api_logs.sql` is updated so new cloud environments bootstrap with `'patch' = 5` baked in.

## Notes for review / ops

Existing cloud deployments still need the equivalent DDL (`ALTER TABLE api_logs MODIFY COLUMN …` and a queue/MV recreation) applied to the live ClickHouse to flip ClickPipes from "DLQ" to "ingest" for in-flight PATCH events. That's a one-time ops step outside this repo.

The migration uses `# frozen_string_literal: false` because the underlying `create_view` helper mutates the SQL string it receives — matching the pattern used by the original `create_api_logs_mv.rb` and other MV migrations in the same directory.

## Test plan

- [x] `bin/rails db:migrate:clickhouse RAILS_ENV=test` — migration applies cleanly.
- [x] `bin/rails db:rollback:clickhouse RAILS_ENV=test` — `down` reverts cleanly to the original enum.
- [x] Re-migrated to leave the test DB at the new state.
- [x] `rspec spec/models/clickhouse/api_log_spec.rb spec/services/utils/api_log_spec.rb spec/queries/api_logs_query_spec.rb spec/graphql/types/api_logs spec/graphql/resolvers/api_log{,s}_resolver_spec.rb spec/serializers/v1/api_log_serializer_spec.rb spec/requests/api/v1/api_logs_controller_spec.rb` — all green (86 examples, 0 failures including the two new PATCH-specific tests).

New tests added in this PR:

- `spec/services/utils/api_log_spec.rb` — a request double with `method_symbol: :patch` produces a Kafka payload that includes `"http_method":"patch"`.
- `spec/queries/api_logs_query_spec.rb` — a `clickhouse_api_log` created with `http_method: "patch"` round-trips through the extended ClickHouse enum and is selectable via the `http_methods: ["patch"]` filter. This is the end-to-end proof that the enum migration actually accepts the new value.
- `spec/graphql/types/api_logs/http_method_enum_spec.rb` — assertion updated to include `patch` in the exposed GraphQL enum values.

### DIFF
--- a/app/models/clickhouse/api_log.rb
+++ b/app/models/clickhouse/api_log.rb
@@ -14,7 +14,8 @@ module Clickhouse
       get: 1,
       post: 2,
       put: 3,
-      delete: 4
+      delete: 4,
+      patch: 5
     }.freeze
 
     private
--- a/db/clickhouse_migrate/20260624144347_add_patch_to_api_logs_http_method.rb
+++ b/db/clickhouse_migrate/20260624144347_add_patch_to_api_logs_http_method.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: false
+
+class AddPatchToApiLogsHttpMethod < ActiveRecord::Migration[8.0]
+  EXTENDED_HTTP_METHODS = {get: 1, post: 2, put: 3, delete: 4, patch: 5}.freeze
+  ORIGINAL_HTTP_METHODS = {get: 1, post: 2, put: 3, delete: 4}.freeze
+
+  def up
+    safety_assured do
+      modify_http_method_enum(EXTENDED_HTTP_METHODS)
+    end
+  end
+
+  def down
+    safety_assured do
+      modify_http_method_enum(ORIGINAL_HTTP_METHODS)
+    end
+  end
+
+  private
+
+  def modify_http_method_enum(values)
+    execute <<~SQL
+      ALTER TABLE api_logs
+        MODIFY COLUMN http_method #{enum_literal(values)};
+    SQL
+
+    # Kafka engine tables don't support ALTER MODIFY COLUMN, so we drop and
+    # recreate the queue together with the materialized view that bridges it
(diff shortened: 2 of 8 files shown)


vincent-pochet (approved):
Nice one 🚀! 

One thing however, the update path for ClickHouse cloud will be a bit different as it does not rely on the Kafka queue.

You can check for the dedicated migration in `/db/clickhouse_migrate/*`. We will probably have to perform a manual migration for this.

PR #5788, opened 2026-06-25
fix(exports): drop orphaned provider customers from exports_customers view

groyoh:
## Context

A Prequel synchronization of the `customers` model aborts on the BigQuery write step because the `exports_customers` view returns more than one row per customer. BigQuery's MERGE requires at most one source row per target row:

```
BigQuery Data Processing Failed [PRQL-FGWV]: Attempted to write data with duplicate rows to BigQuery: UPDATE/MERGE must match at most one source row for each target row, invalidQuery
```

The view LEFT JOINs `payment_provider_customers` filtering only on `deleted_at IS NULL`. `PaymentProviders::DestroyService` nullifies `payment_provider_id` via `update_all` instead of soft-deleting the row, so the orphaned row stays live (`deleted_at = NULL`). When the customer later connects a different payment provider, a fresh row is created. The customer now matches the join twice and the view emits two rows.

The model's uniqueness validation is scoped to `type` (`(customer_id, type)` unique where `deleted_at IS NULL`), so the duplicate happens across provider types: an orphaned row of one type plus a live row of another. The unique index blocks two live rows of the same type, so a same-type reconnection does not trigger this.

## Description

This adds an `exports_customers_v08` view that also requires `ppc.payment_provider_id IS NOT NULL` in the join, so orphaned provider customers are excluded and the view emits at most one row per customer. It ships through a migration following the existing `update_exports_customers_to_version_N` pattern.

Note that it may be be possible to have multiple provider customers of different type. This is out of scope.

### How to verify locally

```
lago exec api bundle exec rspec spec/db/views/exports_customers_spec.rb
```

The spec fails on `v07` (the orphan+live case returns two rows, the orphan-only case leaks the orphan's `provider_customer_id`) and passes on `v08`.

Ref: INT-137

### DIFF
--- a/db/views/exports_customers_v08.sql
+++ b/db/views/exports_customers_v08.sql
@@ -1,0 +1,53 @@
+SELECT
+  c.organization_id,
+  c.id AS lago_id,
+  c.billing_entity_id,
+  c.external_id,
+  c.account_type::text,
+  c.name,
+  c.firstname,
+  c.lastname,
+  c.customer_type::text,
+  c.sequential_id,
+  c.slug,
+  c.created_at,
+  c.updated_at,
+  c.deleted_at,
+  c.country,
+  c.address_line1,
+  c.address_line2,
+  c.state,
+  c.zipcode,
+  c.email,
+  c.city,
+  c.url,
+  c.phone,
+  c.legal_name,
+  c.legal_number,
+  c.currency,
+  c.tax_identification_number,
+  c.timezone,
+  COALESCE(c.timezone, o.timezone, 'UTC') AS applicable_timezone,
+  c.net_payment_term,
+  c.external_salesforce_id,
+  CASE c.finalize_zero_amount_invoice
+    WHEN 0 THEN 'inherit'
+    WHEN 1 THEN 'skip'
+    WHEN 2 THEN 'finalize'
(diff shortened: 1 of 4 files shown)


db/views/exports_customers_v08.sql
```diff
@@ -0,0 +1,53 @@
...
+  c.vat_rate,
+  COALESCE(c.invoice_grace_period, o.invoice_grace_period) AS applicable_invoice_grace_period,
+  c.document_locale,
+  ppc.provider_customer_id,
+  ppc.settings AS provider_settings,
+  '{}'::json AS metadata,
+  '[]'::json AS lago_taxes_ids
+FROM customers c
+LEFT JOIN organizations o ON o.id = c.organization_id
+LEFT JOIN payment_provider_customers ppc ON ppc.customer_id = c.id
```
vincent-pochet, line 51: As pointed by Copilot, we can still have multiple row per customer if a customer is attached to different payment providers.
Not sure if it's an issue, but it differs from the PR description

groyoh: It might be possible but it is not the case I'm trying to fix with this PR. I'm not too sure what's meant to happen when you change payment provider. This would require a deeper investigation.

vincent-pochet (approved):

PR #5789, opened 2026-06-25
feat(search): Add Meilisearch for Invoices

jdenquin:
## Description

  Adds a Meilisearch index for invoices and routes the GraphQL invoice list through it when enabled, with automatic
  fallback to Postgres.

**Indexing**
- Add invoice meilisearch configuration (indexable fields). Set `auto_index: false` to avoid usind AR callbacks.
- Index writes go tr
  - Index writes go through dedicated services and jobs on a `:meilisearch` queue:
    - `Invoices::SearchIndexService` / `RemoveFromSearchIndexService` (upsert / delete-by-id)
    - `Invoices::SearchIndexJob` (upserts if the invoice exists, else removes)
    - `Customers::ReindexInvoicesJob` (re-indexes a customer's invoices when denormalized fields change)
  - Indexing is hooked into the invoice lifecycle (`finalize`, `refresh_draft`, `subscription` for drafts, `update`,
  `void`) and customer `update` / `destroy`, all via `perform_after_commit`.

  **Querying**
  - `InvoicesQuery` gains a `meilisearch:` flag. When search is enabled and a term/filter is present, it resolves
  the page + exact total via Meilisearch (`meilisearch_invoices`); otherwise it uses the existing Postgres path. Any
  `Meilisearch::Error` falls back to Postgres so search never hard-fails.
  - `Lago::Meilisearch::Filter` centralizes filter-expression building (quoting/escaping, `IN`, comparisons) —
  mapping every existing invoice filter onto Meilisearch.
  - `InvoicesResolver` passes `meilisearch: true` and preloads associations whether the result is an AR relation or
  a Meilisearch result set.

  **Config & gating**
  - `MeilisearchClient` splits gating: `enabled?` (`LAGO_MEILISEARCH_URL` present) gates **indexing**;
  `search_enabled?` (also requires `LAGO_USE_MEILISEARCH`) gates the **read path**. This lets the index be
  populated/reindexed before search is served.
  - Initializer, dedicated Sidekiq worker config + start script, and rake tasks (`meilisearch:reindex_invoices`,
  `meilisearch:clear_invoices`).

  Ships dark: with the flags off, behavior is unchanged (Postgres path)

  **New env vars:** `LAGO_MEILISEARCH_URL`, `LAGO_MEILISEARCH_MASTER_KEY`, `LAGO_USE_MEILISEARCH`.
  
  ⚠️ We also need to add a new meilisearch dedicated worker


vincent-pochet (approved):
🚀

app/graphql/resolvers/invoices_resolver.rb
```diff
@@ -84,16 +85,20 @@ def resolve( # rubocop:disable Metrics/ParameterLists
 
       return result_error(result) unless result.success?
 
-      Invoice.preload_offset_amounts(
-        result.invoices.preload(
+      invoices = result.invoices
+
+      if invoices.respond_to?(:preload)
```
groyoh: Why is this conditional ? We should either always preload using `Preloader` or never preload.

spec/jobs/customers/reindex_invoices_job_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  it "runs on the meilisearch queue" do
+    expect(described_class.new.queue_name).to eq("meilisearch")
+  end
+
+  it "enqueues a reindex job for each of the customer's invoices" do
+    perform
+
+    expect(Invoices::SearchIndexJob).to have_received(:perform_later).with(invoice.id)
```
groyoh: Please use `have_enqueued_job` helper whenever possible.

lib/lago/meilisearch/filter.rb
```diff
@@ -0,0 +1,59 @@
...
+        ActiveModel::Type::Boolean.new.cast(value)
+      end
+
+      # Renders a value for an equality / IN expression: booleans bare, numerics
+      # bare, everything else as a quoted, escaped string.
+      def literal(value)
+        return value.to_s if value == true || value == false
+        return value.to_s if value.is_a?(Numeric)
+
+        %("#{value.to_s.gsub('"', '\\"')}")
```
groyoh: ⚠️ `literal` escapes the double quote but leaves backslashes untouched. Filter values here are user-controlled (`customer_external_id`, `currency`, metadata keys/values), and a value containing or ending in a backslash produces a malformed expression: `foo\` renders as `"foo\"`, where the trailing quote is now escaped so the string keeps consuming the following condition. Since the conditions are ANDed with `organization_id`, this stays within the org, but a crafted value can still broaden its own condition. Escaping backslashes before the quote closes this (escape `\` first, then `"`).

app/queries/invoices_query.rb
```diff
@@ -48,18 +64,123 @@ def call
...
+  def meilisearch_invoices
+    search_params = {
+      filter: meilisearch_filter,
+      sort: ["issuing_date:desc", "created_at:desc", "id:asc"],
+      page: current_page,
+      hits_per_page: per_page,
+      matching_strategy: "all"
+    }
+
+    search_params[:attributes_to_search_on] = ["number"] unless search_customers?
```
groyoh: The Postgres path matches an invoice by its exact UUID when the term looks like one (`matching_invoices` adds `search_base.where(id: search_term)` for `BaseQuery::UUID_REGEX`). The Meilisearch path only searches `number` (or the customer fields), and `id` is not in `searchable_attributes`, so pasting a full invoice UUID into search returns nothing here while the PG path finds it. Adding an `id` filter branch when the term matches `BaseQuery::UUID_REGEX` (or making `id` searchable) would keep the two paths equivalent.

spec/models/customer_spec.rb
```diff
@@ -1429,4 +1429,41 @@
...
     end
   end
+
+  describe "invoices search reindexing" do
+    let(:customer) { create(:customer) }
+
+    context "when Meilisearch is enabled" do
+      before do
+        customer
+        allow(Lago::Meilisearch::Client).to receive(:enabled?).and_return(true)
```
groyoh: Can we set `LAGO_MEILISEARCH_URL` in tests instead of mocking ?

app/models/invoice.rb
```diff
@@ -7,16 +7,49 @@ class Invoice < ApplicationRecord
...
+    attribute(:due_amount_cents) { total_amount_cents - total_paid_amount_cents }
+    attribute(:partially_paid) { total_amount_cents > total_paid_amount_cents && total_paid_amount_cents.positive? }
+    attribute(:payment_dispute_lost) { payment_dispute_lost_at.present? }
+    attribute(:customer_name) { customer.name if customer&.kept? }
+    attribute(:customer_firstname) { customer.firstname if customer&.kept? }
+    attribute(:customer_lastname) { customer.lastname if customer&.kept? }
+    attribute(:customer_legal_name) { customer.legal_name if customer&.kept? }
+    attribute(:customer_external_id) { customer.external_id if customer&.kept? }
+    attribute(:customer_email) { customer.email if customer&.kept? }
+    attribute(:subscription_ids) { invoice_subscriptions.pluck(:subscription_id).uniq }
```
groyoh: During `Invoice.reindex!` these attribute blocks run once per record: `invoice_subscriptions.pluck`, `invoice_settlements.distinct.pluck`, the `metadata` association, plus `customer` for the `customer_*` fields. That is ~4-5 queries per invoice, so a full reindex issues on the order of hundreds of thousands of queries for a large org. meilisearch-rails lets the reindex relation be scoped (https://github.com/meilisearch/meilisearch-rails/tree/c76d3104f9fbe26b57bad4ba571b73c86729bb72#eager-loading); eager-loading `:customer`, `:invoice_subscriptions`, `:invoice_settlements`, and `:metadata` there would keep the cost proportional to batches rather than rows.

spec/models/customer_spec.rb
```diff
@@ -1429,4 +1429,41 @@
...
+      end
+
+      it "enqueues an invoices reindex when a searchable field changes" do
```
groyoh: I would suggest to use a test table to test each searcheable fields and deleted_at.

spec/models/customer_spec.rb
```diff
@@ -1429,4 +1429,41 @@
       end
     end
   end
+
+  describe "invoices search reindexing" do
```
groyoh, line 1433: This test does not properly tests the "after commit" logic.

spec/models/invoice_spec.rb
```diff
@@ -2470,4 +2470,102 @@
...
+          "customer_id" => customer.id,
+          "customer_name" => nil,
+          "customer_external_id" => nil,
+          "customer_email" => nil
+        )
+      end
+    end
+  end
+
+  describe "search indexing" do
```
groyoh, line 2591: Same feedback as for customer.

spec/models/invoice_spec.rb
```diff
@@ -2470,4 +2470,102 @@
...
+      end
+    end
+
+    context "when the customer is discarded" do
+      before { customer.discard! }
+
+      it "omits the descriptive customer fields but keeps customer_id" do
+        expect(document).to include(
+          "customer_id" => customer.id,
+          "customer_name" => nil,
```
groyoh, line 2564: What about testing the other customer attributes ?

spec/models/invoice_spec.rb
```diff
@@ -2470,4 +2470,102 @@
...
+        payment_status: "pending",
+        total_amount_cents: 1000,
+        total_paid_amount_cents: 400,
+        issuing_date: Date.new(2026, 1, 15)
+      )
+    end
+
+    it "exposes the searchable, filterable and denormalized attributes" do
+      expect(document).to include(
+        "number" => "INV-001",
```
groyoh, line 2509: This is lacking some attributes, especially the ones which are transformed (e.g. `created_at`)

app/models/invoice.rb
```diff
@@ -7,16 +7,49 @@ class Invoice < ApplicationRecord
...
+
+    searchable_attributes %i[number customer_name customer_firstname customer_lastname
+      customer_legal_name customer_external_id customer_email]
+    filterable_attributes %i[organization_id billing_entity_id currency customer_id
+      customer_external_id invoice_type status payment_status payment_dispute_lost
+      payment_overdue self_billed issuing_date total_amount_cents due_amount_cents
+      partially_paid subscription_ids settlement_types metadata metadata_keys]
+    sortable_attributes %i[issuing_date created_at id]
+    typo_tolerance disable_on_attributes: %w[number customer_external_id customer_email]
+    pagination max_total_hits: 100_000
```
groyoh, line 41: This configuration is not tested (not sure if there's an easy way to test it though).

spec/queries/invoices_query_spec.rb
```diff
@@ -908,4 +908,116 @@
...
+    let(:search_results) do
+      Kaminari.paginate_array([ms_invoice_second, ms_invoice_first], total_count: 42).page(1).per(10)
+    end
+
+    before do
+      ms_invoice_first
+      ms_invoice_second
+      organization.enable_feature_flag!(:meilisearch)
+      allow(Lago::Meilisearch::Client).to receive(:search_enabled?).and_return(true)
+      allow(Invoice).to receive(:search).and_return(search_results)
```
groyoh, line 936: We should assert the search input in this tests with a comprehensive list of filter and their equivalent search param.

lib/lago/meilisearch/filter.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def eq(field, value)
+        "#{field} = #{literal(value)}"
+      end
+
+      def in_list(field, values)
+        "#{field} IN [#{Array(values).map { |value| literal(value) }.join(", ")}]"
+      end
+
+      def not_in(field, values)
```
groyoh: For consistency:
```suggestion
      def not_in_list(field, values)
```

lib/lago/meilisearch/filter.rb
```diff
@@ -0,0 +1,59 @@
...
+        "#{field} <= #{value}"
+      end
+
+      def boolean(value)
+        ActiveModel::Type::Boolean.new.cast(value)
+      end
+
+      # Renders a value for an equality / IN expression: booleans bare, numerics
+      # bare, everything else as a quoted, escaped string.
+      def literal(value)
```
groyoh, line 52: This should be private.

spec/queries/invoices_query_spec.rb
```diff
@@ -908,4 +908,116 @@
...
+      let(:search_term) { "Acme" }
+
+      before do
+        allow(Invoice).to receive(:search).and_raise(Meilisearch::CommunicationError.new("boom"))
+        allow(Sentry).to receive(:capture_exception)
+      end
+
+      it "falls back to the Postgres path and reports the error" do
+        expect(meili_result.invoices.map(&:id)).to include(ms_invoice_first.id, ms_invoice_second.id)
+        expect(Sentry).to have_received(:capture_exception)
```
groyoh: With `:sentry` metadata, you can do:

```suggestion
        expect(sentry_events).to include_sentry_event(exception: exception, message: message)
```

app/queries/invoices_query.rb
```diff
@@ -48,18 +64,123 @@ def call
...
+
+    ActiveRecord::Associations::Preloader.new(
+      records: invoices.to_a,
+      associations: [:customer, {file_attachment: :blob}, {xml_file_attachment: :blob}]
+    ).call
+
+    invoices
+  # NOTE: If any error happens, we fallback to PG search
+  rescue Meilisearch::Error => e
+    Sentry.capture_exception(e) if defined?(Sentry)
```
groyoh:
```suggestion
    Sentry.capture_exception(e)
```

lib/lago/meilisearch/client.rb
```diff
@@ -0,0 +1,25 @@
...
+    #
+    # Indexing and search settings live in the model's `meilisearch do … end` block
+    # (see Invoice).
+    #
+    # - `enabled?` answers "is Meilisearch configured for this env?" and gates
+    #   indexing (writing documents to the index).
+    # - `search_enabled?` additionally requires `LAGO_USE_MEILISEARCH` and gates the
+    #   query/read path. Keeping it separate lets the index be populated or
+    #   reindexed without impacting user search results until the flag is flipped.
+    class Client
```
groyoh: It is a bit confusing that this is called `Client` while there's no real client. Maybe move those method on the `Meilisearch` module ?

lib/lago/meilisearch/client.rb
```diff
@@ -0,0 +1,25 @@
...
+      def self.enabled?
```
groyoh: To avoid confusing:

```suggestion
      def self.indexing_enabled?
```

lib/lago/meilisearch/client.rb
```diff
@@ -0,0 +1,25 @@
...
+        ENV["LAGO_MEILISEARCH_URL"].present?
+      end
+
+      def self.search_enabled?
+        enabled? && ActiveModel::Type::Boolean.new.cast(ENV["LAGO_USE_MEILISEARCH"])
```
groyoh: To keep the same prefix for consistency:

```suggestion
        enabled? && ActiveModel::Type::Boolean.new.cast(ENV["LAGO_MEILISEARCH_SEARCH_ENABLED"])
```

lib/tasks/meilisearch.rake
groyoh, line 1: Could you add tests for these Rake tasks ?

spec/services/invoices/finalize_service_spec.rb
```diff
@@ -19,6 +19,17 @@
...
       end
+
+      context "when Meilisearch is enabled" do
+        before do
+          invoice
+          allow(Lago::Meilisearch::Client).to receive(:enabled?).and_return(true)
+        end
+
+        it "enqueues a search reindex for the invoice" do
+          expect { service.call }.to have_enqueued_job(Invoices::SearchIndexJob).with(invoice.id)
```
groyoh:
```suggestion
          expect { service.call }.to have_enqueued_job_after_commit(Invoices::SearchIndexJob).with(invoice.id)
```

spec/services/invoices/void_service_spec.rb
```diff
@@ -59,6 +59,17 @@
...
         end
+
+        context "when Meilisearch is enabled" do
+          before do
+            invoice
+            allow(Lago::Meilisearch::Client).to receive(:enabled?).and_return(true)
+          end
+
+          it "enqueues a search reindex for the invoice" do
+            expect { void_service.call }.to have_enqueued_job(Invoices::SearchIndexJob).with(invoice.id)
```
groyoh:
```suggestion
            expect { void_service.call }.to have_enqueued_job_after_commit(Invoices::SearchIndexJob).with(invoice.id)
```

config/initializers/meilisearch.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+MeiliSearch::Rails.configuration = {
+  meilisearch_url: ENV["LAGO_MEILISEARCH_URL"] || "http://localhost:7700",
+  meilisearch_api_key: ENV["LAGO_MEILISEARCH_MASTER_KEY"],
```
groyoh: Why not ?

```suggestion
  meilisearch_api_key: ENV["LAGO_MEILISEARCH_API_KEY"],
```

config/initializers/meilisearch.rb
```diff
@@ -0,0 +1,9 @@
...
+  active: ENV["LAGO_MEILISEARCH_URL"].present?,
+  per_environment: true,
```
groyoh, line 8: What does this mean ?

jdenquin: per_environment will create different index based on the env 
eg: `invoices_development, invoices_production`

app/graphql/resolvers/invoices_resolver.rb
jdenquin: added a preloader!

lib/lago/meilisearch/filter.rb
```diff
@@ -0,0 +1,60 @@
...
+    #
+    # Neither meilisearch-rails nor the meilisearch client provide a filter
+    # builder — they only forward the `filter:` value to the engine — so these
+    # helpers centralize the expression syntax (quoting/escaping, IN lists,
+    # comparisons) for reuse across queries.
+    module Filter
+      extend self
+
+      def eq(field, value)
+        "#{field} = #{literal(value)}"
```
groyoh, line 15: I'm wondering if we shouldn't use ActiveRecord's `sanitize_sql_for_conditions` instead of sanitizing manually.

groyoh (approved):
@jdenquin nice work! I'm wondering though, was https://github.com/getlago/lago-api/pull/5789#discussion_r3519600995 addressed ?

lib/lago/meilisearch/filter.rb
jdenquin: let me have a look!

jdenquin: `sanitize_sql_for_conditions` escape values for PostgreSQL's string syntax, but the string being built here is a Meilisearch filter expression, which has its own grammar

app/queries/invoices_query.rb
jdenquin: this has been added by https://github.com/getlago/lago-api/pull/5789/changes#diff-7da1f6ca966fcc604ca9ba657e2d08c55166c75ee2456f4f8e053a2d923bbaa6R82

```ruby
if search_term.present? && search_term.match?(BaseQuery::UUID_REGEX)
  query = ""
  search_params[:filter] << Lago::Meilisearch::Filter.eq("id", search_term)
else
  query = search_term.to_s
  search_params[:attributes_to_search_on] = ["number"] unless search_customers?
end
```

app/models/invoice.rb
jdenquin: mmm not really, I'll add more test in the CI to cover this model configuration

PR #5790, opened 2026-06-25
[ING-327][ING-335][ING-329] misc(clickhouse): Apply AggregationResult to unique_count query

vincent-pochet:
## Context

This PR follows https://github.com/getlago/lago-api/pull/5751

Today, all aggregation services (except CountService) perform two queries against the event store: one for the aggregation itself, and a second for the number of events involved in the result.

This implementation is inherited from the original Postgres store and works well for that engine. When the Clickhouse store was introduced, the same approach was kept for compatibility. At scale this is not ideal on the Clickhouse side, as it forces the engine to retrieve the same set of records twice, including a second pass over the expensive deduplication CTE.

This PR applies the single-pass approach, already introduced for `sum`,  `max`, and `last` to the `unique_count` aggregation.

## Description

`unique_count` and `grouped_unique_count` and their prorated counterparts now return the existing `AggregationResult` / `GroupedAggregationResult` structures carrying both the value and the events_count.

The key differences with already rewritten aggregation is that both `value` and `events_count` return the same value to keep the current behavior. This PR goal is mainly to use the common pattern for unique count even if it does not reduce the number of DB queries

The same approach will be applied to the remaining aggregations in follow-up steps.

### DIFF
--- a/app/services/billable_metrics/aggregations/unique_count_service.rb
+++ b/app/services/billable_metrics/aggregations/unique_count_service.rb
@@ -13,7 +13,7 @@ module BillableMetrics
       def compute_aggregation(options: {})
         return empty_result if should_bypass_aggregation?
 
-        aggregation = event_store.unique_count.ceil(5)
+        aggregation = event_store.unique_count.value.ceil(5)
 
         if options[:is_pay_in_advance] && options[:is_current_usage]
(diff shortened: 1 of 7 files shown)


toommz (approved):
LGTM 👍

PR #5792, opened 2026-06-25
fix(wallet): block zero-value inbound transactions [ING-109]

aquinofb:
## Context

When you add credits to a wallet, Lago converts them to money using the wallet's rate. If the amount you add is so small that it converts to zero money, the credits become useless:

- **Granted credits:** the credit count goes up, but the money balance doesn't. Spending is based on the money balance, so those credits can never be used.
- **Paid credits:** nothing happens at all, and it would mean charging an invoice for zero.

Today both are allowed. This change rejects them.

## What changed

If someone tries to add a positive amount of credits that converts to zero money, it's now refused with a clear error instead of silently creating a broken transaction. This is checked in the three places credits can be added:

- **Normal top-ups** (API, GraphQL, customer portal, automatic recurring top-ups): the request fails with a validation error.
- **Creating a wallet with a starting balance:** the wallet is not created and the user gets the same error right away.
- **Restoring credits after voiding an invoice:** this is skipped quietly, so voiding an invoice never breaks because of old data.

Credits that are exactly zero, and credits being spent (not added), are untouched.

## Examples

What happens when credits are added, for currencies with different numbers of decimals (checked against the real conversion code):

| Currency | Rate | Credits added | Becomes | Result |
|----------|-----:|--------------:|---------|--------|
| EUR (2 decimals) | 0.01 | 0.4 | EUR 0.00 | Refused |
| EUR (2 decimals) | 0.01 | 1.0 | EUR 0.01 | Allowed |
| JPY (0 decimals) | 1.0 | 0.4 | JPY 0 | Refused |
| JPY (0 decimals) | 1.0 | 1.0 | JPY 1 | Allowed |
| KWD (3 decimals) | 0.001 | 0.4 | 0.000 KWD | Refused |
| KWD (3 decimals) | 0.001 | 1.0 | 0.001 KWD | Allowed |
| any | any | 0 | nothing | Allowed (unchanged) |

We reuse Lago's existing money-conversion code so this works correctly for every currency. The note in the issue suggested multiplying by 100, but that only holds for currencies with 2 decimals (the last KWD row would have been wrongly refused).

## Tests

Added tests for the conversion check, the three entry points above, plus the existing transaction flow. All related tests pass.

### DIFF
--- a/app/services/wallet_transactions/recredit_service.rb
+++ b/app/services/wallet_transactions/recredit_service.rb
@@ -15,6 +15,9 @@ module WalletTransactions
 
       return result.not_allowed_failure!(code: "wallet_not_active") unless wallet.active?
 
+      # Only relevant for historical data: zero-rounding amounts are now blocked at creation,
+      # so this guards transactions created before that change. We return a success result so
+      # that voiding an invoice never fails because of such old transactions.
       return result if WalletCredit.rounds_to_zero?(wallet:, credit_amount: wallet_transaction.credit_amount)
 
       transaction_result = WalletTransactions::CreateFromParamsService.call(
--- a/app/services/wallet_transactions/recredit_service.rb
+++ b/app/services/wallet_transactions/recredit_service.rb
@@ -15,6 +15,8 @@ module WalletTransactions
 
       return result.not_allowed_failure!(code: "wallet_not_active") unless wallet.active?
 
+      return result if WalletCredit.rounds_to_zero?(wallet:, credit_amount: wallet_transaction.credit_amount)
+
       transaction_result = WalletTransactions::CreateFromParamsService.call(
         organization: customer.organization,
         params: {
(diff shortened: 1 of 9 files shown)


app/services/wallet_transactions/recredit_service.rb
```diff
@@ -15,6 +15,8 @@ def call
 
       return result.not_allowed_failure!(code: "wallet_not_active") unless wallet.active?
 
+      return result if WalletCredit.rounds_to_zero?(wallet:, credit_amount: wallet_transaction.credit_amount)
```
toommz, line 21: Why returning a valid result in this case? Not saying we shouldn't, but I'm curious

toommz (approved):

app/services/wallet_transactions/recredit_service.rb
groyoh, line 21: Maybe add a comment that this is needed only because of historical data where we might have had 0 amount transactions.

groyoh (approved):

app/services/wallet_transactions/recredit_service.rb
aquinofb: @groyoh replied in the comment below.

In a few words, the only way to reach this code is if the transactions were created before our zero-rounding fix.

PR #5799, opened 2026-06-25
fix(entitlement): clear privilege removal when re-adding at plan default

mikeh-lago:
When a previously removed subscription privilege is re-added with a value equal to the plan default, `SubscriptionEntitlementCoreUpdateService` discarded any subscription override but left the `SubscriptionFeatureRemoval` in place, so the privilege stayed hidden from the UI and API. Re-adding at a non-default value already cleaned up the removal; this makes the default-value branch do the same.

### DIFF
--- a/app/services/entitlement/subscription_entitlement_core_update_service.rb
+++ b/app/services/entitlement/subscription_entitlement_core_update_service.rb
@@ -106,9 +106,9 @@ module Entitlement
 
         if plan_val && value_is_the_same?(privilege.value_type, value, plan_val.value)
           sub_val&.discard!
-          subscription.entitlement_removals.where(privilege:).update_all(deleted_at: Time.zone.now) # rubocop:disable Rails/SkipsModelValidations
+          delete_all_privilege_entitlement_removals(privilege)
         elsif sub_val.nil?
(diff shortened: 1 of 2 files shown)


app/services/entitlement/subscription_entitlement_core_update_service.rb
```diff
@@ -106,6 +106,7 @@ def update_values_for_subscription(plan_entitlement, sub_entitlement)
 
         if plan_val && value_is_the_same?(privilege.value_type, value, plan_val.value)
           sub_val&.discard!
+          subscription.entitlement_removals.where(privilege:).update_all(deleted_at: Time.zone.now) # rubocop:disable Rails/SkipsModelValidations
```
vincent-pochet: Nitpicking here, but I think we could extract this into a small `delete_all_privilege_entitlement_removals` private method, that we could reuse on line 111 as well.

vincent-pochet (approved):

rsempe (approved):

app/services/entitlement/subscription_entitlement_core_update_service.rb
mikeh-lago: Good call — extracted into `delete_all_privilege_entitlement_removals(privilege)` and reused in both branches. Done in 1d2b1ec90.

PR #5800, opened 2026-06-26
feat(multiple-payment-methods): remove multiple payment method feature flag from the code

lovrocolic:
## Context

Multiple payment method feature is already used by cloud organizations.

## Description

This PR removes feature flag gates from the codebase.

### DIFF
--- a/app/services/invoices/payments/create_service.rb
+++ b/app/services/invoices/payments/create_service.rb
@@ -46,10 +46,8 @@ module Invoices
           payable_payment_status: "pending"
         )
 
-        if multiple_payment_methods_enabled?
-          payment.payment_method_id = determine_payment_method&.id
-          payment.save!
-        end
+        payment.payment_method_id = determine_payment_method&.id
+        payment.save!
 
         result.payment = payment
 
@@ -110,20 +108,12 @@ module Invoices
         @provider ||= invoice.customer.payment_provider&.to_sym
       end
 
-      def multiple_payment_methods_enabled?
-        customer.organization.feature_flag_enabled?(:multiple_payment_methods)
-      end
-
       def should_process_payment?
         return false if invoice.self_billed?
         return false if invoice.payment_succeeded? || invoice.voided?
         return false if current_payment_provider.blank?
 
-        if multiple_payment_methods_enabled?
-          current_payment_provider_customer&.provider_customer_id && determine_payment_method.present?
-        else
-          current_payment_provider_customer&.provider_customer_id
-        end
+        current_payment_provider_customer&.provider_customer_id && determine_payment_method.present?
       end
 
       def current_payment_provider
--- a/app/services/invoices/payments/create_service.rb
+++ b/app/services/invoices/payments/create_service.rb
@@ -125,7 +125,7 @@ module Invoices
       def provider_payment_method_pending_backfill?
         legacy_id = current_payment_provider_customer&.legacy_provider_method_id
         legacy_id.present? &&
-          current_payment_provider_customer.payment_methods.where(provider_method_id: legacy_id).none?
+          current_payment_provider_customer.payment_methods.with_discarded.where(provider_method_id: legacy_id).none?
       end
 
       def current_payment_provider
--- a/app/services/invoices/payments/create_service.rb
+++ b/app/services/invoices/payments/create_service.rb
@@ -113,7 +113,19 @@ module Invoices
         return false if invoice.payment_succeeded? || invoice.voided?
         return false if current_payment_provider.blank?
 
-        current_payment_provider_customer&.provider_customer_id && determine_payment_method.present?
+        current_payment_provider_customer&.provider_customer_id &&
+          (determine_payment_method.present? || provider_payment_method_pending_backfill?)
+      end
+
+      # NOTE: While the OSS payment methods backfill is still running
+      #       (rake migrations:backfill_*_payment_methods), a provider customer can have a
+      #       provider-side payment method/mandate with no matching PaymentMethod record yet.
+      #       In that window determine_payment_method is nil, but the payment can still be
+      #       processed since the provider service falls back to the provider-side method.
+      def provider_payment_method_pending_backfill?
+        legacy_id = current_payment_provider_customer&.legacy_provider_method_id
+        legacy_id.present? &&
+          current_payment_provider_customer.payment_methods.where(provider_method_id: legacy_id).none?
       end
 
       def current_payment_provider
--- a/app/services/payment_providers/stripe/payments/create_service.rb
+++ b/app/services/payment_providers/stripe/payments/create_service.rb
@@ -84,9 +84,6 @@ module PaymentProviders
           # NOTE: Retrieve list of existing payment_methods
           payment_method = customer_payment_methods.first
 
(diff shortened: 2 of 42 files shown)


toommz (approved):
Quick question, do we need to keep the flag in `app/config/feature_flags.yaml`? If yes, is it for the frontend to be updated before removing it?

lovrocolic:
> Quick question, do we need to keep the flag in `app/config/feature_flags.yaml`? If yes, is it for the frontend to be updated before removing it?

Hey, FE has also created their PR for removing the feature flag logic on their side.
First I removed flag from .yml file but then I realised that some specs that are related to feature flag model also fail if you do that. I will create additional PR where I will remove feature flag from the .yml file and where I will make feature_flag_specs more generic

groyoh:
> Quick question, do we need to keep the flag in app/config/feature_flags.yaml? If yes, is it for the frontend to be updated before removing it?

Indeed, this is a 3 step deploy:

* Deploy the removal of the feature flag usage on the backend and frontend
* Remove the feature flag from all the organizations
* Remove the feature flag from the `app/config/feature_flags.yaml`

app/services/invoices/payments/create_service.rb
```diff
@@ -108,20 +106,12 @@ def provider
...
         return false if invoice.self_billed?
         return false if invoice.payment_succeeded? || invoice.voided?
         return false if current_payment_provider.blank?
 
-        if multiple_payment_methods_enabled?
-          current_payment_provider_customer&.provider_customer_id && determine_payment_method.present?
-        else
-          current_payment_provider_customer&.provider_customer_id
-        end
+        current_payment_provider_customer&.provider_customer_id && determine_payment_method.present?
```
groyoh: ⚠️ With the flag gone, `should_process_payment?` now requires `determine_payment_method.present?` for every organization. `PaymentMethods::DetermineService` only resolves from Lago `PaymentMethod` rows (`customer.default_payment_method` / `customer.payment_methods`); it never falls back to the legacy `provider_customer.payment_method_id`. Any customer that still relies on that legacy field and has no `PaymentMethod` row will silently stop being auto-charged: this returns `false`, no payment is created, and the invoice just stays pending with no error.

The backfill rake tasks in `lib/tasks/migrations/payment_methods.rake` cover this, but they enqueue async jobs on the low_priority queue. If this merges/deploys before those jobs finish in every environment (self-hosted included), not-yet-backfilled customers fall into the silent-skip path. Worth gating the rollout on backfill completion, or keeping a temporary fallback to `provider_customer.payment_method_id` in `DetermineService` until backfill is confirmed done.

We should ensure that this change gets shipped in a follow-up release once the migration has completed. What do you think ?

app/services/payment_providers/stripe/payments/create_service.rb
```diff
@@ -88,12 +84,8 @@ def stripe_payment_method
...
           payment_method = customer_payment_methods.first
 
-          if invoice.organization.feature_flag_enabled?(:multiple_payment_methods)
-            # TODO: Double check
-            # payment.payment_method.update!(provider_method_id: payment_method&.id) if payment.payment_method
-          else
-            provider_customer.update!(payment_method_id: payment_method&.id)
-          end
+          # TODO: Double check
+          # payment.payment_method.update!(provider_method_id: payment_method&.id) if payment.payment_method
```
groyoh: This TODO should be handled (either uncommented or removed).

spec/services/payment_provider_customers/adyen_service_spec.rb
```diff
@@ -220,59 +220,45 @@
...
-          expect(payment_method.customer).to eq(customer)
-          expect(payment_method.payment_provider_customer).to eq(adyen_customer)
-          expect(payment_method.provider_method_id).to eq(payment_method_id)
-          expect(payment_method.provider_method_type).to eq("card")
-          expect(payment_method.is_default).to be(true)
+        it "does not create a new PaymentMethod" do
+          expect { preauthorise }.not_to change(PaymentMethod, :count)
         end
 
-        it "still updates adyen_customer payment_method_id for backward compatibility" do
```
groyoh, line 241: Why is this removed ?

spec/services/payment_providers/stripe/payments/create_service_spec.rb
```diff
@@ -117,34 +120,9 @@
       end
     end
 
-    context "when multiple payment methods are enabled" do
```
groyoh, line 120: Does the current test assert the same checks ?

aquinofb (approved):

mariohd (approved):
Looks correct.

As pointed by @groyoh 
We just need to ensure that when we remove the flag from app/config/feature_flags.yaml, we remove from all organizations, maybe in a migration...

spec/services/payment_provider_customers/adyen_service_spec.rb
lovrocolic: This is already tested in the line 210. This test was repeated for the case when flag is enabled

spec/services/payment_providers/stripe/payments/create_service_spec.rb
lovrocolic: yep, removed test is actually the same test as the one on the line 97

app/services/payment_providers/stripe/payments/create_service.rb
```diff
@@ -70,11 +70,7 @@ def handle_requires_action(payment)
         end
 
         def stripe_payment_method
-          payment_method_id = if invoice.organization.feature_flag_enabled?(:multiple_payment_methods)
-            payment&.payment_method&.provider_method_id
-          else
-            provider_customer.payment_method_id
-          end
+          payment_method_id = payment&.payment_method&.provider_method_id
```
groyoh, line 73: Don't we need the fallback here ?

app/services/payment_requests/payments/moneyhash_service.rb
```diff
@@ -122,11 +122,7 @@ def create_payment(provider_payment_id:, metadata:)
       end
 
       def moneyhash_payment_method_id
-        if organization.feature_flag_enabled?(:multiple_payment_methods)
-          customer.default_payment_method&.provider_method_id
-        else
-          customer.moneyhash_customer.payment_method_id
-        end
+        customer.default_payment_method&.provider_method_id
```
groyoh: Don't we need the fallback here ?

app/services/payment_providers/stripe/payments/create_service.rb
lovrocolic: Fallback is on the line 85 where latest payment method is directly fetched on the fly from the payment provider

app/services/payment_requests/payments/moneyhash_service.rb
lovrocolic: added just in case

app/services/invoices/payments/create_service.rb
lovrocolic: This is covered now. In all downstream services, we have fallbacks. At this specific place, we're doing a query that tries to find the corresponding payment method based on the legacy payment method. If none is found, we fall into the pending migration flow and still try to process the payment.

PR #5803, opened 2026-06-26
feat(orders): update order mutation

toommz:
Add a GraphQL mutation to let users edit an Order before execution. An order is mostly immutable, but we let users change those 2 settings linked to how we execute it:

- `execute_at` to schedule the execution at a future date
- `execution_mode` to set if we should handle the billing inside lago or not

### DIFF
--- a/app/services/orders/update_service.rb
+++ b/app/services/orders/update_service.rb
@@ -3,6 +3,7 @@
 module Orders
   class UpdateService < BaseService
     include OrderForms::Premium
+    include OrderForms::ExecutionSettingsValidation
 
     Result = BaseResult[:order]
 
@@ -23,7 +24,10 @@ module Orders
       Order.transaction do
         Quotes::LockService.call(quote: order.quote) do
           order.reload
-          next result.single_validation_failure!(field: :status, error_code: "not_editable") unless order.created?
+          unless order.created?
+            result.single_validation_failure!(field: :status, error_code: "not_editable")
+            result.raise_if_error!
+          end
 
           order.assign_attributes(params.slice(:execution_mode, :execute_at))
           order.save!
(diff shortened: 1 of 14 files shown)


app/services/orders/update_service.rb
```diff
@@ -0,0 +1,79 @@
...
+      return result.not_found_failure!(resource: "order") unless order
+      return result.forbidden_failure! unless order_forms_enabled?(order.organization)
+
+      validate_execution_settings
+      return result if result.failure?
+
+      Order.transaction do
+        Quotes::LockService.call(quote: order.quote) do
+          order.reload
+          next result.single_validation_failure!(field: :status, error_code: "not_editable") unless order.created?
```
groyoh: I believe this should raise instead to rollback if needed. It's not an issue now but it could be if it's wrapped by another transaction. The inner transaction fails but without rollback and the upper transaction may not rollback.

groyoh (approved):

app/services/orders/update_service.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def effective_execution_mode
+      params.key?(:execution_mode) ? params[:execution_mode] : order.execution_mode
+    end
+
+    def effective_execute_at
+      params.key?(:execute_at) ? params[:execute_at] : order.execute_at
+    end
+
+    def validate_execution_settings
```
groyoh, line 58: `validate_execution_mode` and `validate_execute_at` here are the same rules `OrderForms::MarkAsSignedService` already carries (its `validate_execution_settings`/`validate_execution_mode`/`validate_execute_at`). Two write paths now have to keep the execution-settings rules in sync by hand, so a tweak in one (e.g. a new constraint) can silently drift from the other. Extracting them into a small shared concern next to `OrderForms::Premium`, with the validators taking the mode/date values as arguments, would keep both paths aligned. The only delta is that this service feeds resolved `effective_*` / `params.key?` values while `MarkAsSignedService` feeds its named args.

PR #5809, opened 2026-06-26
fix(payment-providers): soft delete provider customers on destroy

ancorcruz:
## Context

When a payment provider is destroyed, its associated payment provider customers were only detached by nullifying their payment_provider_id. The records were left active, even though they no longer point to any provider and the provider itself is discarded.

## Description

Set deleted_at on the payment provider customers so they are soft deleted alongside the provider, in addition to clearing their payment_provider_id. This keeps the discarded provider and its customer records consistent.

### DIFF
--- a/app/services/customers/payment_provider_finder.rb
+++ b/app/services/customers/payment_provider_finder.rb
@@ -18,10 +18,10 @@ module Customers
         payment_provider_result.payment_provider
       end
 
-      def payment_provider_customer(customer)
-        return nil unless customer
+      def payment_provider_customer(payment)
+        return nil unless payment
 
-        PaymentProviderCustomers::BaseCustomer.with_discarded.find_by(customer_id: customer.id)
+        PaymentProviderCustomers::BaseCustomer.with_discarded.find_by(id: payment.payment_provider_customer_id)
       end
     end
   end
--- a/app/services/payment_providers/destroy_service.rb
+++ b/app/services/payment_providers/destroy_service.rb
@@ -14,7 +14,8 @@ module PaymentProviders
       customer_ids = payment_provider.customer_ids
 
       ActiveRecord::Base.transaction do
-        payment_provider.payment_provider_customers.update_all(payment_provider_id: nil, deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
+        now = Time.current
+        payment_provider.payment_provider_customers.update_all(updated_at: now, deleted_at: now) # rubocop:disable Rails/SkipsModelValidations
         payment_provider.discard!
 
         Customer.where(id: customer_ids).update_all(payment_provider: nil, payment_provider_code: nil) # rubocop:disable Rails/SkipsModelValidations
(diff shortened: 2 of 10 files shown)


app/services/payment_providers/destroy_service.rb
```diff
@@ -14,7 +14,7 @@ def call
       customer_ids = payment_provider.customer_ids
 
       ActiveRecord::Base.transaction do
-        payment_provider.payment_provider_customers.update_all(payment_provider_id: nil) # rubocop:disable Rails/SkipsModelValidations
+        payment_provider.payment_provider_customers.update_all(payment_provider_id: nil, deleted_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
```
groyoh: I'd suggest to also update `updated_at`:

```suggestion
        now = Time.current
        payment_provider.payment_provider_customers.update_all(payment_provider_id: nil, updated_at:  now, deleted_at: now) # rubocop:disable Rails/SkipsModelValidations
```

groyoh: I'm wondering if we should still set the `payment_provider_id` to `nil`. Since the provider is soft-deleted, I believe we can keep it.

ancorcruz: I'm applying this.

ancorcruz: yes, lets do that, I have did a quick scan on the code to ensure it won't break anything. looks safe.

groyoh (approved):
This will need thorough QA (or scenario) tests as I'm not sure what happens when we recreate a new provider/provider customer later on.

groyoh (commented):
Could you please add some tests cases for the refund changes ?

ancorcruz:
@groyoh Hey, could you take another look? I have added a scenario test to cover payment provider reconnection and fixed the payment refund using the same payment method used for the original payment (from the saved id in payment record) as we now could have more than one provider customer with an id but soft deleted.

app/services/customers/payment_provider_finder.rb
```diff
@@ -18,10 +18,10 @@ def payment_provider(customer)
         payment_provider_result.payment_provider
       end
 
-      def payment_provider_customer(customer)
-        return nil unless customer
+      def payment_provider_customer(payment)
+        return nil unless payment
 
-        PaymentProviderCustomers::BaseCustomer.with_discarded.find_by(customer_id: customer.id)
+        PaymentProviderCustomers::BaseCustomer.with_discarded.find_by(id: payment.payment_provider_customer_id)
```
groyoh, line 24: Could there be a case when `payment.payment_provider_customer_id` is `nil` ? Since it's not marked as "NOT NULL", I assume it could be the case but don't know enough this logic to know whether it's an issue.

groyoh (approved):

app/services/customers/payment_provider_finder.rb
ancorcruz: yes, the column is nullable, but every payment reaching the refund path is a succeeded provider payment that always has it set; and a `nil` would return nil safely rather than break.

PR #5820, opened 2026-06-30
feat(products): attach rate cards to plans

rsempe:
## Context

Attaching a rate card to a plan is what turns catalog pricing into an offer. The attachment is its own resource — it carries `units` and rate phases — so it is exposed as `applied_rate_cards` rather than as the cards themselves.

## Description

- Add `PlanRateCards` create/update/destroy services and the full REST resource (`/api/v2/plans/{code}/applied_rate_cards`). GraphQL exposes the collection and a `createPlanAppliedRateCard` mutation only — update and destroy will follow with the front work that needs them, rather than shipping unused schema.
- One card per pricing slice: a plan may hold several cards of the same product only when they cover different filter slices. A duplicate is rejected with `product_already_priced` or `product_filter_already_priced`. The check runs under a lock on the plan, since the rule spans several cards and no unique index can enforce it.
- Reject a card whose currency differs from the plan currency at configuration time (`currency_does_not_match`), freeze the plan currency once it holds attachments, and freeze a card's own currency once it is attached to a plan or a subscription.
- A subscribed plan is immutable: attaching, editing or removing a card is rejected with `plan_locked`. Price changes go through a new plan and a subscription migration.
- Create a default rate phase with each attachment.


app/services/rate_phases/create_service.rb
```diff
@@ -0,0 +1,38 @@
...
+    def call
+      parent = plan_rate_card || subscription_rate_card
+      return result.not_found_failure!(resource: "rate_phaseable") unless parent
+
+      rate_phase = RatePhase.create!(
+        organization: parent.organization,
+        plan_rate_card:,
+        subscription_rate_card:,
+        code: params[:code],
+        position: params[:position],
```
tiagolupepic, line 23: will the `position` validation be added in another PR?

tiagolupepic (approved):
one small comment and it's not a blocker

:shipit:

app/graphql/types/mutation_type.rb
```diff
@@ -26,6 +26,8 @@ class MutationType < Types::BaseObject
     field :destroy_plan, mutation: Mutations::Plans::Destroy
     field :update_plan, mutation: Mutations::Plans::Update
 
+    field :create_plan_applied_rate_card, mutation: Mutations::PlanAppliedRateCards::Create
```
endenis, line 29: The PR description speaks about create/update/destroy for GraphQL but only create is present.

Are the mutations for update and destroy planned in a separate PR? 🤔

app/services/plan_rate_cards/create_service.rb
```diff
@@ -0,0 +1,80 @@
...
+    def initialize(plan:, params:)
+      @plan = plan
+      @params = params.to_h.with_indifferent_access
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "plan") unless plan
+
+      # A plan with subscriptions is immutable: pricing changes go through a new
```
endenis, line 16: I suggest to add a check that the plan is not a legacy plan.
Something like `if plan.product_catalog?`

app/services/plan_rate_cards/update_service.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def call
+      return result.not_found_failure!(resource: "applied_rate_card") unless plan_rate_card
+
+      if plan_rate_card.plan.attached_to_subscriptions?
+        return result.single_validation_failure!(field: :plan, error_code: "plan_locked")
+      end
+
+      plan_rate_card.units = params[:units] if params.key?(:units)
+      plan_rate_card.save!
```
endenis, line 23: Is it a good idea to add some validations of units on the PlanRateCard model?
Something like:
```ruby
validates :units, numericality: {greater_than_or_equal_to: 0}, allow_nil: true
```

app/services/plan_rate_cards/create_service.rb
```diff
@@ -0,0 +1,80 @@
...
+        return result.single_validation_failure!(field: :plan, error_code: "plan_locked")
+      end
+
+      rate_card = organization.rate_cards.find_by(code: params[:rate_card_code])
+      return result.not_found_failure!(resource: "rate_card") unless rate_card
+
+      # Fees are billed in the card currency and the invoice in the plan
+      # currency; a mismatch must fail at configuration time, not on the
+      # first invoice.
+      if rate_card.currency != plan.amount_currency
```
endenis, line 28: Should we also include a similar check in `RateCards::UpdateService`? I think it can still change rate card's currency even when it is attached to a plan

app/graphql/types/mutation_type.rb
rsempe: yeah it's something extracted, update and destroy will come later, updating the description 👍

app/services/rate_phases/create_service.rb
rsempe: yes, i'll add it in [#5842](https://github.com/getlago/lago-api/pull/5842) 👍

app/services/plan_rate_cards/create_service.rb
rsempe: will add it with the feature-flag commit in #5776 👍

app/services/plan_rate_cards/update_service.rb
rsempe: yes doing it 👍

app/services/plan_rate_cards/create_service.rb
rsempe: done ✅

PR #5821, opened 2026-06-30
[ING-311] fix(data-api): Improve HTTP client retry on transient errors

groyoh:
## Context

`Api::V1::DataApi::UsagesController#index` calls `DataApi::UsagesService` synchronously, which fetches usage data from the Lago Data API over HTTP. When the Data API hits a transient database error (a closed connection), it returns a 5xx. `LagoHttpClient::Client` turns that into a `LagoHttpClient::HttpError`, which surfaces as a failed request to the caller. These errors are transient and succeed when the call is retried.

## Description

This adds an opt-in retry capability to `LagoHttpClient::Client`. A new `retry_on_transient_errors:` flag (default `false`, so existing callers are unaffected) makes the client retry a curated set of transient failures: connection-level exceptions (SSL, open and read timeouts, EOF, connection reset/refused) and transient 5xx responses (500, 502, 503, 504). Retries reuse the existing `MAX_RETRIES_ATTEMPTS` limit, with a short jittered pause between attempts. 4xx responses are never retried.

The flag is enabled for the DataApi client in `DataApi::BaseService`, so all DataApi services (usages, MRRs, prepaid credits, revenue streams) gain the retry, since they share the same backend and the same transient failure.

Rewriting the request loop also fixes a pre-existing issue: once an exception-based retry exhausted its attempts, the client raised a `NoMethodError` on a nil response instead of re-raising the original error. It now re-raises the original error.

### DIFF
--- a/lib/lago_http_client/lago_http_client/client.rb
+++ b/lib/lago_http_client/lago_http_client/client.rb
@@ -147,36 +147,38 @@ module LagoHttpClient
     def request(req, params = nil)
       attempt = 0
 
-      loop do
+      begin
         attempt += 1
-
-        begin
-          response = http_client.request(req, params)
-        rescue => e
-          raise unless retry_on_error?(e) && attempt < MAX_RETRIES_ATTEMPTS
-
-          backoff
-          next
-        end
-
-        if retry_on_status?(response.code.to_i) && attempt < MAX_RETRIES_ATTEMPTS
-          backoff
-          next
-        end
-
+        response = http_client.request(req, params)
         raise_error(response) unless RESPONSE_SUCCESS_CODES.include?(response.code.to_i)
+        response
+      rescue => e
+        if retryable?(e) && attempt < MAX_RETRIES_ATTEMPTS
+          backoff
+          retry
+        end
 
-        return response
+        raise
       end
     end
 
-    def retry_on_error?(error)
-      retries_on.include?(error.class) ||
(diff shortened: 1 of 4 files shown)


vincent-pochet (approved):

lib/lago_http_client/lago_http_client/client.rb
```diff
@@ -133,19 +147,42 @@ def raise_error(response)
     def request(req, params = nil)
       attempt = 0
 
-      response = begin
+      begin
         attempt += 1
-        http_client.request(req, params)
+        response = http_client.request(req, params)
+        raise_error(response) unless RESPONSE_SUCCESS_CODES.include?(response.code.to_i)
```
toommz, line 153: Really not important; We could also have checked if it is a `Net::HTTPSuccess`, the only thing to keep in mind being that `Net::HTTPSuccess` covers more 2xx codes than what we have in the `RESPONSE_SUCCESS_CODES` constant.

groyoh:
Thanks @vincent-pochet 🙏

toommz (approved):
Nice!

aquinofb (approved):

lib/lago_http_client/lago_http_client/client.rb
groyoh: I totally agree. The issue is that this was already in the previous code and I didn't want to break anything.

PR #5823, opened 2026-06-30
feat(products): support product-catalog plans

rsempe:
## Context

A plan on the product catalog prices through its rate cards, so the plan-level billing fields (interval, amount, pay-in-advance) no longer apply to it.

## Description

- Add `Plan#pricing_type` (`legacy` | `product_catalog`), defaulting to `legacy`. It is **not** accepted in any payload: the pricing engine is an organization-level choice, so the value is derived from the organization's `product_catalog` feature flag at plan creation (the derivation ships with the flag in a follow-up PR of this stack).
- Make `interval`, `amount_cents` and `pay_in_advance` optional for catalog plans, keeping them required for legacy ones.
- Return no billing period dates and no downgrade date for catalog subscriptions, instead of deriving them from a plan-level period.

### Acceptable breaking change on v1

A legacy plan created without `amount_cents` now returns `amount_cents: ["value_is_mandatory"]` instead of `["is not a number"]`

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -49,13 +49,22 @@ class Plan < ApplicationRecord
     semiannual
   ].freeze
 
-  enum :interval, INTERVALS, validate: true
+  PRICING_TYPES = {legacy: "legacy", product_catalog: "product_catalog"}.freeze
 
-  monetize :amount_cents
+  enum :pricing_type, PRICING_TYPES, validate: true
+  enum :interval, INTERVALS, validate: {allow_nil: true}
+
+  monetize :amount_cents, allow_nil: true
 
   validates :name, :code, presence: true
   validates :amount_currency, inclusion: {in: currency_list}
-  validates :pay_in_advance, inclusion: {in: [true, false]}
+  # Legacy plans price at the plan level; product-catalog plans price through
+  # their products, so the plan-level billing fields are optional for them.
+  # A blank interval keeps the historical "value_is_invalid" error (the enum used
+  # to reject nil as an out-of-range value) rather than a "value_is_mandatory" one.
+  validates :interval, presence: {message: "value_is_invalid"}, unless: :product_catalog?
+  validates :amount_cents, presence: true, unless: :product_catalog?
+  validates :pay_in_advance, inclusion: {in: [true, false]}, unless: :product_catalog?
   validate :validate_code_unique
 
   default_scope -> { kept }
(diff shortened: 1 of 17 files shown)


app/models/plan.rb
```diff
@@ -49,13 +49,22 @@ class Plan < ApplicationRecord
...
   validates :name, :code, presence: true
   validates :amount_currency, inclusion: {in: currency_list}
-  validates :pay_in_advance, inclusion: {in: [true, false]}
+  # Legacy plans price at the plan level; product-catalog plans price through
+  # their products, so the plan-level billing fields are optional for them.
+  # A blank interval keeps the historical "value_is_invalid" error (the enum used
+  # to reject nil as an out-of-range value) rather than a "value_is_mandatory" one.
+  validates :interval, presence: {message: "value_is_invalid"}, unless: :product_catalog?
+  validates :amount_cents, presence: true, unless: :product_catalog?
+  validates :pay_in_advance, inclusion: {in: [true, false]}, unless: :product_catalog?
```
endenis, line 67: Should we also allow these values to be null in [Types::Plans::Object](https://github.com/getlago/lago-api/blob/73c06304482e7910a6213f800288804d8ad16471/app/graphql/types/plans/object.rb#L11)?

db/migrate/20260630161816_add_pricing_type_to_plans.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddPricingTypeToPlans < ActiveRecord::Migration[8.0]
+  def change
+    create_enum :plan_pricing_type, %w[legacy product_catalog]
+    add_column :plans, :pricing_type, :enum, enum_type: :plan_pricing_type, default: "legacy", null: false
+
+    safety_assured do
+      change_column_null :plans, :interval, true
+      change_column_null :plans, :amount_cents, true
```
endenis, line 10: This change feels a bit scary because there is existing code that relies on these fields to be not null (like the [yearly_amount_cents](https://github.com/getlago/lago-api/blob/73c06304482e7910a6213f800288804d8ad16471/app/models/plan.rb#L120) method in plan that is used in other places)

app/models/subscription.rb
```diff
@@ -233,6 +233,10 @@ def downgrade_plan_date
     if next_subscription.active? && downgraded?
       return next_subscription.started_at&.to_date
     end
+    # A scheduled downgrade lands at the end of the current period, which
+    # product-catalog plans don't have: their rate cards each bill on their own
+    # cycle.
+    return if plan.product_catalog?
```
endenis: Should this check be done at the beginning of this method? There is an early return on line 234

app/models/plan.rb
rsempe: good catch

app/models/subscription.rb
rsempe: true!

db/migrate/20260630161816_add_pricing_type_to_plans.rb
rsempe: wdyt of https://github.com/getlago/lago-api/pull/5823/changes/3d205fa46810f0ef74c944aa8ddd84cd23a1d1fe ?

endenis (approved):

db/migrate/20260630161816_add_pricing_type_to_plans.rb
endenis: It covers well the v1 👍 
We will have to adapt all places that use the amount_cents for v2

PR #5824, opened 2026-06-30
feat(fc-units-override): drop premium gate on units-only write paths

ancorcruz:
## Context

Plan override is a premium feature. Fixed-charge units override is a separate feature that does not require premium — it isn't a plan override.

The three subscription write services premium-gated every request that carried a `plan_overrides` payload (or, on the dedicated endpoint, every request at all). The units-only write paths share those entry points, so they inherited a gate that doesn't belong to them.

## Description

Relax the premium gate so it only fires when the request actually routes to the plan-override branch. The units-only branch is reachable on any license.

- UpdateOrOverrideFixedChargeService gates premium on `!units_only_change?`. The not_found checks now come before the premium check, which is the standard 404-before-403 order.
- CreateService and UpdateService extend their existing `params.key?(:plan_overrides)` gate with `!units_only_plan_overrides_change?`.

Specs cover the new free-tier units-only success path on all three services alongside the existing premium-required plan-override failures.

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -31,7 +31,7 @@ module Subscriptions
         activation_rules: params[:activation_rules],
         subscription_type:
       )
-      return result.forbidden_failure! if !License.premium? && params.key?(:plan_overrides)
+      return result.forbidden_failure! if !License.premium? && params.key?(:plan_overrides) && !units_only_plan_overrides_change?
       return result.validation_failure!(errors: {external_customer_id: ["value_is_mandatory"]}) if params[:external_customer_id].blank? && api_context?
 
       # TODO: Remove check we stop supporting `plan_overrides.usage_thresholds`
--- a/spec/services/subscriptions/create_service_spec.rb
+++ b/spec/services/subscriptions/create_service_spec.rb
@@ -432,6 +432,33 @@ RSpec.describe Subscriptions::CreateService do
           expect(override.units).to eq(25)
         end
       end
+
+      context "when plan_overrides carries a fixed_charges change beyond units" do
+        let(:fixed_charge) { create(:fixed_charge, plan:, units: 5) }
+        let(:params) do
+          {
+            external_customer_id:,
+            plan_code:,
+            name:,
+            external_id:,
+            billing_time:,
+            subscription_at:,
+            subscription_id:,
+            plan_overrides: {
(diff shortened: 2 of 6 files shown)


spec/services/subscriptions/create_service_spec.rb
```diff
@@ -402,6 +402,36 @@
         expect(result).not_to be_success
         expect(result.error.code).to eq("feature_unavailable")
       end
+
+      context "when plan_overrides is a units-only fixed_charges payload" do
```
tiagolupepic, line 406: do we should cover as well the other path? 
at least an expectation to check when the additional condition is true. We could check the exception here. wdyt?

app/services/subscriptions/create_service.rb
```diff
@@ -31,7 +31,7 @@ def call
         activation_rules: params[:activation_rules],
         subscription_type:
       )
-      return result.forbidden_failure! if !License.premium? && params.key?(:plan_overrides)
+      return result.forbidden_failure! if !License.premium? && params.key?(:plan_overrides) && !units_only_plan_overrides_change?
```
tiagolupepic, line 34: we're checking the `params.key?(:plan_overrides)` in `units_only_plan_overrides_change` method. Do we still need it here?

ancorcruz: We still need it — removing it introduces a regression. `units_only_plan_overrides_change?` returns false in two different situations:

1. No plan_overrides payload at all (short-circuits on params.key?(:plan_overrides))
2. A plan_overrides payload that is a real override (not only units)

`!units_only_plan_overrides_change?` is therefore true in both. We only require premium license for case 2.

If we drop the explicit `params.key?(:plan_overrides)`, then a not premium license organization creating a subscription with no plan_overrides at all evaluates to !License.premium? && true → forbidden_failure!. That's a 403 on a perfectly valid not premium license request.

So the guard isn't duplicated logic — it's what distinguishes "no override" from "non-units override."

spec/services/subscriptions/create_service_spec.rb
ancorcruz: Added test for the other path, plan override with fixed charge overrides, not just units.

osmarluz (approved):
🚀

spec/services/subscriptions/create_service_spec.rb
tiagolupepic: thanks Ancor!

tiagolupepic (approved):
:shipit:

PR #5825, opened 2026-06-30
[ING-112] Cascade ongoing usage across wallets

aquinofb:
## Context

Lago consumes wallet credits through two paths that diverge for multi-wallet customers. Billing already cascades across wallets in priority order, but ongoing usage stopped at the first wallet, taking it negative instead of spilling onto the next. The negative ongoing balance is customer-visible (UI, API, threshold rules) and breaks the two-wallet (paid + free) setup.

Linear: https://linear.app/getlago/issue/ING-112

## Behavior

`Wallet B` = priority 1 (consumed first), `Wallet A` = priority 2. Ongoing balance per wallet after the given usage:

| Wallets (priority order)        | Usage | Before        | After                    |
|---------------------------------|-------|---------------|--------------------------|
| Single wallet, balance 50       | 30    | 20            | 20 (unchanged)           |
| Single wallet, balance 50       | 80    | −30           | −30 (unchanged)          |
| B 50 + A 150                    | 80    | B −30, A 150  | **B 0, A 120**           |
| B 50 + A 150                    | 250   | B −200, A 150 | B 0, **A −50**           |
| B 50 (threshold rule) + A 150   | 80    | B −30, A 150  | B −30, A 150 (unchanged) |

- **Single wallet** is unchanged: it still absorbs everything and may go negative.
- **Multiple wallets** now cascade in priority order -- a wallet fills to 0 before the next is touched (row 3, the headline case: matches what billing produces).
- When usage exceeds every wallet, the **last** wallet absorbs the overflow and goes negative (row 4) instead of the first -- nothing is silently dropped.
- A wallet with an **active threshold recurring rule** keeps today's behavior: it absorbs everything and goes negative so the rule can fire and refill it (row 5).

## Changes

- Add `Wallets::Balance::AllocateOngoingUsageByWalletsService`, mirroring the billing cascade (`Credits::AllocatePrepaidCreditsByWalletsService`): net usage per fee, distribute across active wallets in priority order, fill each to its balance, spill the overflow, and let the last applicable wallet go negative. An over-billed fee key reduces a per-currency budget the same way billing credits reduce the invoice total.
- A wallet with an active threshold-based recurring rule keeps the previous behavior (absorbs everything, may go negative) so the rule can fire and refill it.
- `RefreshWalletsService` runs the allocator once and feeds each wallet its precomputed amount. Because the cascade makes allocations interdependent, every refresh persists all wallets; the previous `target_wallet_ids` narrowing would leave non-target wallets stale.
- The old single-wallet assignment is replaced, which leaves `FindApplicableOnFeesService` and `BuildAllocationRulesService` without callers. They are deleted in https://github.com/getlago/lago-api/pull/5884 (~600 lines of pure deletion) to keep this one focused on the behavior change.

## Validation

The specs from this branch (they encode the expected cascade behavior) were run against `main` and against this branch -- same 46 examples:

| Codebase | Result |
|---|---|
| `main` (old allocation) | 46 examples, **10 failures** -- every multi-wallet cascade scenario |
| this branch | 46 examples, **0 failures** |

Per-scenario ongoing balances (each row is one failing-then-fixed spec):

| Scenario | On `main` | On this branch |
|---|---|---|
| B (50, priority 1) + A (150), usage 80 -- the ticket example | B **-30**, A 150 | B 0, A **120** |
| 4 wallets (10/20/30/40), each limited to 2 metrics, overlapping; usage 100 | W1 **-20**, W2 -10, W3 -10, W4 idle at 40 | W1 0, W2 -10, W3 -10, W4 **20** |
| 4 wallets, three limited to charge fees; charge usage 100 | W2 absorbs all: **-80**; W3, W4 idle | W2 0, W3 0, W4 **-10** |
| 4 unrestricted wallets; net usage 90 (10 already progressively billed) | W1 **-90**; W2, W3, W4 idle | cascades W2, W3, W4 -- all land at 0 |
| Wallet targeted to metric 1 (1000) + catch-all wallet; metric-1 usage 3000 | targeted **-2000**; catch-all untouched | targeted 0; catch-all absorbs the overflow (**-1000**) |
| Two wallets (20000 + 25000), usage 35000; billing deducts 20000 + 15000 | ongoing shows **35000 on wallet 1, 0 on wallet 2** -- diverges from billing | ongoing shows 20000 + 15000 -- **identical to what billing deducts** |

Preserved behavior passes on both codebases: single wallet going negative, threshold-rule wallets absorbing and refilling, and ongoing-balance alerts firing as the balance crosses into negative.

Edge cases pinned by the new allocator unit suite (the service does not exist on `main`, so no old-code column):

| Edge case | Behavior |
|---|---|
| Wallet with an active interval (non-threshold) recurring rule | cascades like a normal wallet -- only threshold rules absorb and refill |
| Threshold wallet that is not first in priority | higher-priority wallets fill first, then it absorbs the remainder and stops the cascade |
| Zero-balance wallet in the middle of the cascade | skipped, never forced negative |
| One metric over-billed by progressive billing (usage corrected down) | its excess offsets the other metrics, like billing credits offset the invoice |
| Fee already fully covered by progressive billing | nothing is allocated |
| Mixed-currency wallets | a fee is only absorbed by wallets in its currency |
| Fee targeting a wallet code that matches no wallet | allocated to no wallet, matching billing where no wallet covers it |
| Balance changed concurrently (stale in-memory value) | the cascade caps against the balance re-read from the database |

### DIFF
--- a/app/services/customers/refresh_wallets_service.rb
+++ b/app/services/customers/refresh_wallets_service.rb
@@ -31,8 +31,7 @@ module Customers
         )
       end
 
-      now = Time.current
-      Wallet.where(id: all_wallets.map(&:id)).update_all(updated_at: now, last_ongoing_balance_sync_at: now) # rubocop:disable Rails/SkipsModelValidations
+      Wallet.where(id: all_wallets.map(&:id)).touch_all(:last_ongoing_balance_sync_at) # rubocop:disable Rails/SkipsModelValidations
 
       customer.update!(awaiting_wallet_refresh: false)
 
--- a/app/services/customers/refresh_wallets_service.rb
+++ b/app/services/customers/refresh_wallets_service.rb
@@ -31,7 +31,8 @@ module Customers
         )
       end
 
-      Wallet.where(id: all_wallets.map(&:id)).update_all(last_ongoing_balance_sync_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
+      now = Time.current
+      Wallet.where(id: all_wallets.map(&:id)).update_all(updated_at: now, last_ongoing_balance_sync_at: now) # rubocop:disable Rails/SkipsModelValidations
 
       customer.update!(awaiting_wallet_refresh: false)
 
--- a/app/services/customers/refresh_wallets_service.rb
+++ b/app/services/customers/refresh_wallets_service.rb
@@ -2,68 +2,39 @@
 
 module Customers
   class RefreshWalletsService < BaseService
-    Result = BaseResult[:usage_amount_cents, :wallets, :allocation_rules]
+    Result = BaseResult[:wallets]
 
-    def initialize(customer:, include_generating_invoices: false, target_wallet_ids: nil)
+    def initialize(customer:, include_generating_invoices: false)
       @customer = customer
       @include_generating_invoices = include_generating_invoices
-      @target_wallet_ids = target_wallet_ids
 
       super
     end
 
     def call
-      usage_amount_cents = customer.active_subscriptions.map do |subscription|
-        invoice = ::Invoices::CustomerUsageService.call!(customer:, subscription:, usage_filters: UsageFilters::WITHOUT_PRESENTATION_FILTER).invoice
+      wallet_allocations = Wallets::Balance::AllocateOngoingUsageByWalletsService.call!(
+        customer:,
+        wallets: all_wallets,
+        current_usage_fees:,
+        draft_invoices_fees:,
+        progressive_billing_fees:,
+        pay_in_advance_fees:
+      ).wallet_allocations
 
-        billed_progressive_invoice_subscriptions = ::Subscriptions::ProgressiveBilledAmount
-          .call(subscription:, include_generating_invoices:)
-          .invoice_subscriptions
-
-        {
-          billed_progressive_invoice_subscriptions:,
-          invoice:,
-          subscription:
-        }
+      # The cascade makes every wallet's allocation depend on the others' balances, so all
+      # wallets must be persisted together; refreshing a subset would leave the rest stale.
+      all_wallets.each do |wallet|
+        Wallets::Balance::RefreshOngoingUsageService.call!(
+          wallet:,
+          ongoing_usage_amount_cents: wallet_allocations[wallet],
+          skip_single_wallet_update: true
+        )
       end
 
-      @allocation_rules = Wallets::BuildAllocationRulesService.call!(customer:).allocation_rules
-
-      # we need to get both: ALL fees and wallets_applicable_on_fees ({ fee_key => wallet_id, ... }) per each fee type
-      # to not rebuild wallet assigning per each fee when going per each separate wallet
-      usage_fees = usage_amount_cents.flat_map { |usage| usage[:invoice].fees }
-      wallets_applicable_on_usage_fees = assign_wallet_per_fee(usage_fees) # { usage_fee_key => wallet_id }
-
-      draft_invoice_fees = customer.invoices.draft.where.not(total_amount_cents: 0).includes(fees: :charge).flat_map(&:fees)
-      wallets_applicable_on_draft_fees = assign_wallet_per_fee(draft_invoice_fees) # { draft_fee_key => wallet_id }
-
-      progressive_billing_fees = usage_amount_cents.flat_map { |usage| usage[:billed_progressive_invoice_subscriptions].flat_map { it.invoice.fees } }
-      wallets_applicable_on_pb_fees = assign_wallet_per_fee(progressive_billing_fees) # { pb_fee_key => wallet_id }
-
-      pay_in_advance_fees = usage_amount_cents.flat_map { |usage| usage[:invoice].fees.select { |f| f.charge.pay_in_advance? } }
-      wallets_applicable_on_adv_fees = assign_wallet_per_fee(pay_in_advance_fees) # { adv_fee_key => wallet_id }
-
-      wallets_to_process = customer.wallets.active.includes(:recurring_transaction_rules)
-      wallets_to_process = wallets_to_process.where(id: target_wallet_ids) if target_wallet_ids
-      wallets_to_process.find_in_batches(batch_size: 100) do |wallets|
-        wallets.each do |wallet|
-          Wallets::Balance::RefreshOngoingUsageService.call!(
-            wallet:,
-            usage_amount_cents:,
-            skip_single_wallet_update: true,
-            current_usage_fees: applicable_fees(usage_fees, wallets_applicable_on_usage_fees, wallet),
-            draft_invoices_fees: applicable_fees(draft_invoice_fees, wallets_applicable_on_draft_fees, wallet),
-            progressive_billing_fees: applicable_fees(progressive_billing_fees, wallets_applicable_on_pb_fees, wallet),
-            pay_in_advance_fees: applicable_fees(pay_in_advance_fees, wallets_applicable_on_adv_fees, wallet)
-          )
-        end
-      end
-      wallets_to_process.update_all(last_ongoing_balance_sync_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
+      Wallet.where(id: all_wallets.map(&:id)).update_all(last_ongoing_balance_sync_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
 
       customer.update!(awaiting_wallet_refresh: false)
 
-      result.usage_amount_cents = usage_amount_cents
-      result.allocation_rules = allocation_rules
       result.wallets = customer.wallets.active.reload
       result
     rescue BaseService::FailedResult => e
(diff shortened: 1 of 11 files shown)


mariohd (approved):

brunomiguelpinto (approved):

brunomiguelpinto:
we should extract the allocation logic to be used in the 2 lanes to be easier to maintain

app/services/customers/refresh_wallets_service.rb
```diff
@@ -2,68 +2,39 @@
...
-            skip_single_wallet_update: true,
-            current_usage_fees: applicable_fees(usage_fees, wallets_applicable_on_usage_fees, wallet),
-            draft_invoices_fees: applicable_fees(draft_invoice_fees, wallets_applicable_on_draft_fees, wallet),
-            progressive_billing_fees: applicable_fees(progressive_billing_fees, wallets_applicable_on_pb_fees, wallet),
-            pay_in_advance_fees: applicable_fees(pay_in_advance_fees, wallets_applicable_on_adv_fees, wallet)
-          )
-        end
-      end
-      wallets_to_process.update_all(last_ongoing_balance_sync_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
+      Wallet.where(id: all_wallets.map(&:id)).update_all(last_ongoing_balance_sync_at: Time.current) # rubocop:disable Rails/SkipsModelValidations
```
groyoh:
```suggestion
      now = Time.current
      Wallet.where(id: all_wallets.map(&:id)).update_all(updated_at: now, last_ongoing_balance_sync_at: now) # rubocop:disable Rails/SkipsModelValidations
```

spec/scenarios/wallets/customer_wallets_balance_refresh_spec.rb
groyoh, line 1: Could you ask Claude to provide a more visually representative header for each scenario. Kind of like what we have for wallet traceability :

```rb
      # Customer has two top-ups ($30 granted, $50 granted), then invoice consumes $60:
      #
      # INBOUND                                 OUTBOUND
      # ┌─────────────────────────┐
      # │ TX1 (granted)           │             ┌─────────────────────────┐
      # │ amount: $30             │────$30─────▶│ TX3 (invoiced)          │
      # │ remaining: $0           │         ┌──▶│ amount: $60             │
      # └─────────────────────────┘         │   └─────────────────────────┘
      # ┌─────────────────────────┐         │
      # │ TX2 (granted)           │         │
      # │ amount: $50             │───$30───┘
      # │ remaining: $20          │
      # └─────────────────────────┘
      #
      # Join table:
      # │ TX1 │ TX3 │ $30 │  (first inbound consumed first - FIFO)
      # │ TX2 │ TX3 │ $30 │  (then second inbound)
```

spec/scenarios/wallets/customer_wallets_balance_refresh_spec.rb
```diff
@@ -92,13 +92,11 @@
...
-    # 1000 - 3000(from events) = -2000
-    # wallet 2
-    # 1000 - 500(from event) = 500
+    # bm1 usage (3000) fills wallet 1 to 0, then spills onto the unrestricted wallet 3.
+    # wallet 2 covers bm2 usage (500) -> 500 remaining.
     ##
     it "updates the correct ongoing balances for each wallet" do
-      expect_wallet(wallet, ongoing_usage: 3000, credits_usage: 30, ongoing: -2000, credits: -20)
+      expect_wallet(wallet, ongoing_usage: 1000, credits_usage: 10, ongoing: 0, credits: 0)
       expect_wallet(wallet2, ongoing_usage: 500, credits_usage: 5, ongoing: 500, credits: 5)
```
groyoh, line 109: Why don't we assert wallet 3 here when the comment clearly mentions it.

spec/scenarios/wallets/customer_wallets_balance_refresh_spec.rb
```diff
@@ -171,15 +169,14 @@
...
+      # bm1 usage (3000) fills wallet 1 to 0, spilling onto the unrestricted wallet 3.
       # wallet 2
       # fee 2 is not taken into account because of the wallet restrictions
       # 1000 - 500(from event) - 100(from invoice) ( 100 + 10(tax) - 10(already paid) )
       # = 400 # progressive billing invoice
       ##
       it "updates wallet ongoing balances including progressive billing invoice" do
-        expect_wallet(wallet, ongoing_usage: 3000, credits_usage: 30, ongoing: -2000, credits: -20)
+        expect_wallet(wallet, ongoing_usage: 1000, credits_usage: 10, ongoing: 0, credits: 0)
         expect_wallet(wallet2, ongoing_usage: 400, credits_usage: 4, ongoing: 600, credits: 6)
```
groyoh, line 203: Same here. No assertion on wallet 3.

spec/scenarios/wallets/customer_wallets_balance_refresh_spec.rb
```diff
@@ -213,14 +210,13 @@
...
-      # 1000 - 3000(from events) = -2000 #untouched
+      # bm1 usage (3000) fills wallet 1 to 0, spilling onto the unrestricted wallet 3.
       # wallet 2
       # 1000 - 500(from event) - 110(from draft invoice) + 70 (already paid)
       # = 540 # progressive billing invoice
       ##
       it "updates wallet ongoing balances including progressive billing invoice" do
-        expect_wallet(wallet, ongoing_usage: 3000, credits_usage: 30, ongoing: -2000, credits: -20)
+        expect_wallet(wallet, ongoing_usage: 1000, credits_usage: 10, ongoing: 0, credits: 0)
         expect_wallet(wallet2, ongoing_usage: 540, credits_usage: 5.4, ongoing: 460, credits: 4.6)
```
groyoh, line 246: Same here. No assertion on wallet 3.

spec/services/wallets/balance/allocate_ongoing_usage_by_wallets_service_spec.rb
```diff
@@ -0,0 +1,324 @@
...
+      end
+
+      it "lets the over-billed key offset the other keys, like billing credits offset the invoice" do
+        # net = 80 + (100 - 150) = 30, matching what billing would deduct from wallets
+        expect(result).to be_success
+        expect(result.wallet_allocations).to eq({wallet_b => 30, wallet_a => 0})
+      end
+    end
+
+    context "when a fee targets a wallet code that matches no wallet" do
```
groyoh, line 287: The missing-code case is covered here, but the matched-code overflow is not: a fee whose `target_wallet_code` matches a wallet with a balance smaller than the fee amount. Since `applicable_fee?` makes the coded wallet the only applicable one, it is also the last applicable wallet, so it absorbs the full amount and goes negative while the other wallets stay untouched. A spec pinning that outcome would protect the "targeted fees never spill onto other wallets" rule if the overflow handling changes later.

app/services/wallets/balance/allocate_ongoing_usage_by_wallets_service.rb
```diff
@@ -0,0 +1,145 @@
...
+      end
+
+      def wallet_meta(wallet)
+        threshold = threshold_wallet?(wallet)
+        {
+          wallet:,
+          targets: wallet.wallet_targets.filter_map { |wt| ["charge", wt.billable_metric_id] if wt.billable_metric_id },
+          types: wallet.allowed_fee_types,
+          threshold:,
+          # Re-read the balance so a concurrent pay-in-advance DecreaseService can't make us cap
```
groyoh: I'm not sure to follow why this would be needed. Shouldn't we use a lock if we want to prevent this issue ?

groyoh (commented):
I haven't been able to look deeply into the business logic but the code coverage looks sound so I'll trust that.

app/services/customers/refresh_wallets_service.rb
aquinofb: Great call! I just updated to use [touch_all](https://blog.saeloun.com/2019/08/27/rails-6-touch-all-method/).

app/services/wallets/balance/allocate_ongoing_usage_by_wallets_service.rb
aquinofb: That would increase the scope of this PR and make it even more complex and bigger, since I'd need to change DecreaseService to add the lock as well. wdyt? I can do it now or raise a follow up ticket

app/services/wallets/balance/allocate_ongoing_usage_by_wallets_service.rb
```diff
@@ -0,0 +1,145 @@
...
+          # against a stale value. Balance only, to keep the eager-loaded associations.
+          balance: threshold ? 0 : Wallet.where(id: wallet.id).pick(:balance_cents)
```
annvelents: a note here: Synthesia already has customers with 100 of wallets - these are organization-customers. They also are likely to send a lot of usages, so we'll be likely to need to refresh these customers' wallets often. Maybe we can reload balances all-in-once, wdyt?

```
def calculate_wallet_allocations
  net_amounts = net_usage_by_fee_key
  budgets = currency_budgets(net_amounts)
  balances = fresh_balances                       # single query, up front
  metas = wallets.map { |wallet| wallet_meta(wallet, balances) }
  # ...
end

def wallet_meta(wallet, balances)
  threshold = threshold_wallet?(wallet)
  {
    wallet:,
    targets: wallet.wallet_targets.filter_map { |wt| ["charge", wt.billable_metric_id] if wt.billable_metric_id },
    types: wallet.allowed_fee_types,
    threshold:,
    balance: threshold ? 0 : balances.fetch(wallet.id, 0)
  }
end

# Re-read balances from the DB so a concurrent pay-in-advance DecreaseService
# can't make us cap against stale in-memory values. One WHERE id IN (...) for
# all wallets instead of one query per wallet.
def fresh_balances
  Wallet.where(id: wallets.map(&:id)).pluck(:id, :balance_cents).to_h
end
```

annvelents (approved):
Amazing work on refactoring! 👏 🤩
left just one comment

app/services/wallets/balance/allocate_ongoing_usage_by_wallets_service.rb
aquinofb: Good catch! Applied in 7b36271fb. 

One query for all wallets instead of one per wallet. **It is much better than what I had:** a single statement reads every balance at the same instant, while the per-wallet reads could see different moments across wallets.

PR #5826, opened 2026-06-30
test(payment-gated-subs): drop flag from specs

ancorcruz:
## Context

Payment-gated subscriptions no longer rely on the payment_gated_subscriptions feature flag in application code — activation-rules behavior is driven entirely by the request payload. The specs were the only remaining place enabling the flag, so they had drifted out of date.

## Description

Remove the feature-flag setup from the subscriptions controller and subscription serializer specs: drop the "when feature flag is enabled" wrappers and the enable_feature_flag! / feature_flags setup, keeping the existing assertions. Restructure the serializer specs so each context reflects the subscription state it exercises — a canceled subscription for cancellation_reason and a subscription with activation rules for the rules array — and assert the always-present activated_at and empty activation_rules fields in the standard payload test.

### DIFF
--- a/spec/requests/api/v1/subscriptions_controller_spec.rb
+++ b/spec/requests/api/v1/subscriptions_controller_spec.rb
@@ -753,44 +753,40 @@ RSpec.describe Api::V1::SubscriptionsController, :premium do
         }
       end
 
-      context "when feature flag is enabled" do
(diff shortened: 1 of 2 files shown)


osmarluz (approved):
🚀

PR #5827, opened 2026-06-30
chore: Remove payment-gated-subscriptions FF

ancorcruz:
payment gated subscriptions feature is generally available. So, feature flag is not used anymore.


osmarluz (approved):
🚀

PR #5837, opened 2026-07-01
fix(jobs): Add missing `unique` guard to jobs

groyoh:
## Context

Clock jobs are scheduled by Clockwork on a fixed cadence and enqueued with `perform_later`. When the worker fleet is saturated, the clock queue drains slower than Clockwork enqueues, so copies of the same clock job stack up. Each copy of a fan-out job re-runs its full scan and enqueues one child job per record, so N stacked copies over M records produce up to N×M child enqueues. This can grow the Sidekiq queue by a very large factor. 14 of the 28 clock jobs already guarded against this with the `activejob-uniqueness` macro; the rest did not.

## Description

This adds `unique :until_executed, on_conflict: :log` to the 14 clock jobs and 3 event-triggered fan-out jobs that were missing it, matching the pattern already used across the codebase (for example `Clock::ActivateSubscriptionsJob`). The macro holds a Redis lock from enqueue until the job finishes, so a duplicate enqueue for the same key is dropped (and logged) while a copy is still pending or running.

The 12 clock fan-out drivers set `lock_ttl: 4.hours`. The default TTL (1 hour) is close to the hourly cadence of these jobs, so during a multi-hour worker backlog the enqueue lock can expire right as the next Clockwork tick fires, and a second copy still stacks up. A 4-hour TTL caps accumulation to one pending copy across a typical backlog window, and stays well under the 12 hours used by the heavy child jobs. The 2 daily cleanup jobs (`Clock::WebhooksCleanupJob`, `Clock::InboundWebhooksCleanupJob`) and the 3 event-triggered jobs keep the 1-hour default.

`Customers::ViesCheckJob` is the exception. It re-enqueues itself for the same customer from inside `perform` (the VIES retry backoff), so `until_executed` would keep the enqueue lock held during `perform` and drop the retry. It uses `until_and_while_executing` instead: the enqueue lock is released before `perform` runs (so the reschedule is allowed), while a separate runtime lock prevents two checks for the same customer from running at once. Because the retry is scheduled with a delay (at least 5 minutes), the original job has released the runtime lock by the time the retry runs.

Each job gets a duplicate-enqueue test through the existing `shared_examples "a unique job"`, and `ViesCheckJob` gets a regression test that exercises the full enqueue-to-perform lifecycle and fails if the strategy is switched back to `until_executed`.

Fixes LAGO-1673.

### DIFF
--- a/.github/workflows/migrations-test.yml
+++ b/.github/workflows/migrations-test.yml
@@ -19,6 +19,15 @@ jobs:
           POSTGRES_DB: lago
           POSTGRES_USER: lago
           POSTGRES_PASSWORD: lago
+      redis:
+        image: redis
+        ports:
+          - "6379:6379"
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
     env:
       RAILS_ENV: test
       DATABASE_URL: "postgres://lago:lago@localhost:5432/lago"
--- a/app/jobs/clock/inbound_webhooks_retry_job.rb
+++ b/app/jobs/clock/inbound_webhooks_retry_job.rb
@@ -2,6 +2,8 @@
 
 module Clock
   class InboundWebhooksRetryJob < ClockJob
+    unique :until_executed, on_conflict: :log
+
     def perform
       InboundWebhook.retriable.find_each do |inbound_webhook|
         InboundWebhooks::ProcessJob.perform_later(inbound_webhook:)
--- a/app/jobs/clock/inbound_webhooks_retry_job.rb
+++ b/app/jobs/clock/inbound_webhooks_retry_job.rb
@@ -2,7 +2,7 @@
 
 module Clock
   class InboundWebhooksRetryJob < ClockJob
-    unique :until_executed, on_conflict: :log
+    unique :until_executed, on_conflict: :log, lock_ttl: 4.hours
 
     def perform
       InboundWebhook.retriable.find_each do |inbound_webhook|
--- a/app/jobs/clock/retry_failed_invoices_job.rb
+++ b/app/jobs/clock/retry_failed_invoices_job.rb
@@ -2,6 +2,8 @@
 
 module Clock
   class RetryFailedInvoicesJob < ClockJob
+    unique :until_executed, on_conflict: :log
+
     def perform
       Invoice
         .failed
--- a/app/jobs/clock/retry_failed_invoices_job.rb
+++ b/app/jobs/clock/retry_failed_invoices_job.rb
@@ -2,7 +2,7 @@
 
 module Clock
   class RetryFailedInvoicesJob < ClockJob
-    unique :until_executed, on_conflict: :log
+    unique :until_executed, on_conflict: :log, lock_ttl: 4.hours
 
     def perform
       Invoice
--- a/app/jobs/customers/vies_check_job.rb
+++ b/app/jobs/customers/vies_check_job.rb
@@ -12,6 +12,9 @@ module Customers
     # perform runs and holds a separate runtime lock during execution. Releasing the
     # enqueue lock before perform is what lets schedule_retry re-enqueue this job for
     # the same customer from inside perform without being dropped.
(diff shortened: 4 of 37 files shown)


rsempe (approved):

groyoh:
Thanks @rsempe 🙏

app/jobs/clock/retry_failed_invoices_job.rb
```diff
@@ -2,6 +2,8 @@
 
 module Clock
   class RetryFailedInvoicesJob < ClockJob
+    unique :until_executed, on_conflict: :log, lock_ttl: 4.hours
```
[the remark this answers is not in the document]
groyoh: The 4-hour TTL is intentional. The default 1-hour TTL is close to the hourly cadence of these fan-out drivers, so during a multi-hour backlog the enqueue lock expires as the next Clockwork tick fires and a second copy still stacks up. 4 hours caps accumulation to one pending copy across a typical backlog window, and stays under the 12 hours used by the heavy child jobs. The crash-suppression window you note is the trade-off; capping stacking was the priority here. The PR description was updated to document this.

app/jobs/clock/inbound_webhooks_retry_job.rb
```diff
@@ -2,6 +2,8 @@
 
 module Clock
   class InboundWebhooksRetryJob < ClockJob
+    unique :until_executed, on_conflict: :log, lock_ttl: 4.hours
```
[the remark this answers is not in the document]
groyoh: The 4-hour TTL is intentional. The default 1-hour TTL is close to the hourly cadence of these fan-out drivers, so during a multi-hour backlog the enqueue lock expires as the next Clockwork tick fires and a second copy still stacks up. 4 hours caps accumulation to one pending copy across a typical backlog window, and stays under the 12 hours used by the heavy child jobs. The crash-suppression window you note is the trade-off; capping stacking was the priority here. The PR description was updated to document this.

app/jobs/customers/vies_check_job.rb
```diff
@@ -7,6 +7,16 @@ class ViesCheckJob < ApplicationJob
...
     queue_as :default
 
+    # until_and_while_executing takes the enqueue lock (keyed on the customer) so
+    # duplicate enqueues while one is pending are deduped, then releases it before
+    # perform runs and holds a separate runtime lock during execution. Releasing the
+    # enqueue lock before perform is what lets schedule_retry re-enqueue this job for
+    # the same customer from inside perform without being dropped.
+    # The retry is scheduled with a delay (RETRY_DELAYS starts at 5 minutes), so by
+    # the time it runs the current job has finished and released the runtime lock,
+    # and the retry is not dropped by the runtime guard either.
```
toommz, line 17: IMO this comment explains a documented option. Not saying I'm against a quick explanation of why this one uses another type of lock, but explaining the lock type may be a bit too much 🤷

.github/workflows/migrations-test.yml
```diff
@@ -19,6 +19,15 @@ jobs:
...
           POSTGRES_PASSWORD: lago
+      redis:
+        image: redis
+        ports:
+          - "6379:6379"
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
```
toommz, line 30: Just for my own knowledge, why do we need this now?

toommz (approved):

.github/workflows/migrations-test.yml
groyoh: Because of `db/migrate/20250415143607_enqueue_update_all_eu_taxes_job.rb` which enqueues jobs which now require uniqueness and thefore Redis. It was not needed before because it runs in test env and there with active job's `test` adapter, which is not using Redis.

app/jobs/customers/vies_check_job.rb
groyoh: It does more than explain the documented option. It explains WHY this option is necessary in this case and why it works. Without this, one may be tempted to with to a `until_executed` strategy which would not work.

PR #5838, opened 2026-07-02
[BIL-253] fix(fees): Speed up list endpoint for large organizations

groyoh:
## Context

`GET /api/v1/fees` could fail with a Postgres statement timeout for organizations that own a large share of the `fees` table. Both queries behind the endpoint (the pagination `COUNT(*)` and the page `SELECT ... ORDER BY created_at, id LIMIT`) could not use the single-column `index_fees_on_organization_id`: at low selectivity Postgres falls back to a full sequential scan of the table. On a large table that scan, and the sort that follows for the page query, exceed the statement timeout, so the request returns a 500.

## Description

This adds a partial composite index on `fees (organization_id, created_at, id) WHERE deleted_at IS NULL`. It matches the default `kept` scope, the `organization_id` filter, and the `created_at, id` ordering used by `FeesQuery`, so the page query becomes a bounded index scan (it stops after the page size instead of sorting the whole set) and the count becomes an index-only scan over just the organization's rows. The index is built with `algorithm: :concurrently` and `if_not_exists: true`.

The fees list endpoint now also caches its pagination count, the same way the invoices endpoint already does, by passing `key`, `organization_id`, and the request params to `pagination_metadata`. The `page` and `per_page` params are included so the existing last-page recount logic in `Pagination#_count_total` keeps working. Once the count is cached it stays off the request path for the cache TTL.

Fixes BIL-253.


vincent-pochet (approved):

groyoh:
Thanks @vincent-pochet 🙏

toommz (approved):
🚤

PR #5842, opened 2026-07-02
feat(products): add rate overrides on rate phases

rsempe:
## Context

A phase can replace the card's active rate with its own pricing: the rate override.

## Description

- `RateOverrides::CreateService` — creates the override with the same model/item/timing compatibility matrix as catalog rates; spend floors stay arrears-only; structural card fields (currency, billing timing, proration, …) are rejected with `not_overridable` instead of being silently dropped.
- Overrides attach through phase authoring: nested in whole-sequence replace, and per-phase create/update over REST and GraphQL (`rate_override` input; explicit null clears).
- Override lifecycle is transactional: replacing or clearing discards the superseded record, destroying a phase discards its override.
- REST and GraphQL expose `rate_override` on phases, preloaded/dataloaded.


app/services/rate_phases/update_service.rb
```diff
@@ -15,31 +15,46 @@ def initialize(rate_phase:, params:)
...
       # check or the blank slips past it and persists as an indefinite phase.
       if params.key?(:billing_interval_cycle_count)
         params[:billing_interval_cycle_count] = params[:billing_interval_cycle_count].presence
       end
 
-      if params.key?(:billing_interval_cycle_count) && params[:billing_interval_cycle_count].nil? && !last_phase?
-        return result.single_validation_failure!(field: :billing_interval_cycle_count, error_code: "indefinite_phase_must_be_last")
-      end
+      parent.with_lock do
+        if plan_locked?
```
endenis, line 25: Shouldn't `rate_properties` be still editable?
Or is the [Notion doc](https://app.notion.com/p/getlago/SPEC-Plan-v2-Plans-and-overrides-36def63110d280eca75ee8511e4f771f?source=copy_link#390ef63110d280f59e99e5e8eb68eb88) outdated?

endenis (approved):

app/services/rate_phases/update_service.rb
rsempe: yes the notion is not correct - updating it 👍

PR #5843, opened 2026-07-02
feat(products): rate_overrides table + create service

rsempe:
## Context

A rate phase can override the rate its card would otherwise apply for that stretch of the schedule (e.g. a free trial phase). Per [BE Dive-In 3](https://app.notion.com/p/36cef63110d281e19694f95b33a44e39), the overridden pricing lives in a dedicated `rate_overrides` table referenced by `rate_phases.rate_override_id` (`null` = use the card's active rate).

## Description

- **`rate_overrides` table + `RateOverride` model** — mirrors the *overridable* fields of a `rate_card_rate` (`rate_model`, `rate_properties`, `min_amount_cents`, `billing_interval_count`/`unit`, `pricing_unit_conversion_rate`) and deliberately omits the structural card fields (currency, billing timing, proration, …), which are always inherited from the card. Billing-interval fields are nullable to inherit the card's active rate. Reuses the existing `rate_card_rate_model` / `rate_card_rate_billing_interval_unit` PG enum types.
- **FK** — `rate_phases.rate_override_id`, shipped without a foreign key in the v1.a schema, now references `rate_overrides`.
- **`RateOverrides::CreateService`** — builds a `RateOverride` from pricing params in the context of a rate card; requires a `pricing_unit_conversion_rate` when the card carries a pricing unit, and surfaces model-level rate-property validation errors.

### Scope / follow-up
This is the storage + creation layer. **Wiring overrides into the rate phase replace flow** (accepting a `rate_override` object per phase in `RatePhases::ReplaceService` + the REST/GraphQL inputs, so the deferred `rate_override: null` in the previous PR becomes writable) is the next step.


rsempe:
Superseded: the rate_overrides table + FK moved into the data-model PR (#5652), and RateOverrides::CreateService moved into the rate-phase PR (#5842). Closing this branch.

PR #5844, opened 2026-07-02
feat(purchase_order_number) Change hubspot object to include PO number and invoice url

tiagolupepic:
This commit changes hubspot object to include two new fields.

Continuing the epic of purchase order number, we're sending the value to Hubspot.
In order to do that, we need to change the custom object of `LagoInvoices`. The steps are
quite simple, we're bumping the version to `2` and including the `lago_purchase_order_number`.

Also, we noticed the payload doesn't include the invoice_url. We refactored the `Netsuite` payload
to share the method and to be used in different payloads. 

The missing step, since the `DeployObjectService` it's called only when a service of Hubspot is created
and/or updated. We need to manually call via production console to push the new version.

### DIFF
--- a/app/services/integrations/hubspot/invoices/deploy_object_service.rb
+++ b/app/services/integrations/hubspot/invoices/deploy_object_service.rb
@@ -4,7 +4,7 @@ module Integrations
   module Hubspot
     module Invoices
       class DeployObjectService < Integrations::Aggregator::BaseService
-        VERSION = 1
+        VERSION = 2
 
         def action_path
           "v1/hubspot/object"
--- a/app/services/integrations/hubspot/invoices/deploy_object_service.rb
+++ b/app/services/integrations/hubspot/invoices/deploy_object_service.rb
@@ -84,6 +84,13 @@ module Integrations
                 searchableInGlobalSearch: true
               },
               {
+                name: "lago_invoice_purchase_order_number",
+                label: "Lago Purchase Order Number",
+                type: "string",
+                fieldType: "text",
+                searchableInGlobalSearch: true
+              },
+              {
                 name: "lago_invoice_issuing_date",
                 label: "Lago Invoice Issuing Date",
                 type: "date",
--- a/app/services/integrations/hubspot/invoices/deploy_object_service.rb
+++ b/app/services/integrations/hubspot/invoices/deploy_object_service.rb
@@ -4,7 +4,7 @@ module Integrations
   module Hubspot
     module Invoices
       class DeployObjectService < Integrations::Aggregator::BaseService
-        VERSION = 2
+        VERSION = 3
 
         def action_path
           "v1/hubspot/object"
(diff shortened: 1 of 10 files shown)


app/services/integrations/hubspot/invoices/deploy_object_service.rb
```diff
@@ -4,7 +4,7 @@ module Integrations
   module Hubspot
     module Invoices
       class DeployObjectService < Integrations::Aggregator::BaseService
-        VERSION = 1
+        VERSION = 2
```
osmarluz: @tiagolupepic

There's an issue with the rollout mechanism here: `DeployObjectService` and `DeployPropertiesService` both guard on and stamp the **same** `integration.invoices_properties_version` column, and this bump makes their `VERSION` constants collide (both = 2). That breaks things in both directions:

**New integrations:** `DeployObjectService` now stamps version 2 when it creates the object, which permanently short-circuits `DeployPropertiesService` (`return result if integration.invoices_properties_version == VERSION`). But the object payload here doesn't include `lago_invoice_total_due_amount` — that property only exists in `DeployPropertiesService`'s payload, which will never run anymore. Since the invoice payload unconditionally sends `"lago_invoice_total_due_amount"`, every new integration will sync invoices referencing a property that was never created in HubSpot.

**Existing integrations:** the planned manual console call will be a silent no-op. Any integration that has synced an invoice already has `invoices_properties_version == 2` (stamped by `DeployPropertiesService`), so the guard returns immediately. Even past the guard, when the object already exists `CustomObjectService` succeeds and we just save the object type id and return — the POST with the v2 schema only fires when the object doesn't exist. So there's no code path that adds the two new properties to an already-deployed object. That becomes urgent with #5845: once the payload starts sending `lago_invoice_purchase_order_number` / `lago_invoice_url`, existing integrations will send properties their object doesn't have, and HubSpot rejects records referencing unknown properties — so invoice sync breaks for existing customers, not just "fields stay empty".

**Suggestion:** use the same mechanism we used for `lago_invoice_total_due_amount` — add the two new properties to `DeployPropertiesService`'s payload and bump *its* VERSION to 3. That lazily migrates existing objects on their next invoice sync and removes the version collision. Keeping the new properties in `DeployObjectService`'s payload is still right (new objects get them at creation), but its property list should also gain `lago_invoice_total_due_amount` so newly created objects are complete without depending on the properties service running afterwards.

tiagolupepic: > add the two new properties to DeployPropertiesService's payload

I'm not sure if I agree with this change. We don't need to change the properties, only the object. I think idea of the comment is that integration has double responsibility. 
First one, covers the properties, second responsibility, also supports the object.

The silly difference is this code in `DeployObjectService`

```
if integration.invoices_properties_version == VERSION && integration.invoices_object_type_id.present?
  return result
end
```

it seems the integration can have a `invoices_object_type_id` for object only and the properties doesn't. This is not clear.

I'll move the version to 3 to make sure we'll push correctly the object to Hubspot.

osmarluz (approved):
🚀

PR #5845, opened 2026-07-02
feat(purchase_order_number) Add PO number to Xero and Hubspot payload

tiagolupepic:
This commit updates the connections of Xero and Hubspot

Continuing the PO number epic and considering the new custom object is already in place
[in this PR](https://github.com/getlago/lago-api/pull/5844)
We're updating the payload of Hubspot to include `lago_purchase_order_number` and `lago_invoice_url`.

Also, we're adding the `reference` to Xero payload. Since Xero doesn't support specific column for PO number,
we're using the `reference` which is a text field with a limit of 255 chars that matches exactly what we
need for the invoice purchase_order_number.

### DIFF
--- a/app/services/integrations/aggregator/invoices/payloads/hubspot.rb
+++ b/app/services/integrations/aggregator/invoices/payloads/hubspot.rb
@@ -28,7 +28,8 @@ module Integrations
                   "lago_invoice_total_amount" => total_amount,
                   "lago_invoice_total_due_amount" => total_due_amount,
                   "lago_invoice_subtotal_excluding_taxes" => subtotal_excluding_taxes,
-                  "lago_invoice_file_url" => invoice.file_url
+                  "lago_invoice_file_url" => invoice.file_url,
+                  "lago_invoice_url" => invoice_url
                 }
               }
             }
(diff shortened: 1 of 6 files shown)


app/services/integrations/aggregator/invoices/payloads/hubspot.rb
```diff
@@ -17,6 +17,7 @@ def create_body
                 "properties" => {
                   "lago_invoice_id" => invoice.id,
                   "lago_invoice_number" => invoice.number,
+                  "lago_invoice_purchase_order_number" => invoice.purchase_order_number,
```
osmarluz, line 20: @tiagolupepic 

There's an issue with how the new HubSpot property handles invoices without a PO number:

`"lago_invoice_purchase_order_number" => invoice.purchase_order_number` sends an explicit JSON `null` whenever the invoice has no PO number — which is the common case, since `purchase_order_number` is only set on one-off invoices (and the model normalizes blanks to `nil`). HubSpot's CRM v3 API rejects null property values with `400 "Property ... cannot be set to null"`, so every subscription invoice sync would fail on both create and update once the property exists in the portal. It's also the only nullable value in these hashes today — the other nullable field, `file_url`, is guarded by an explicit raise before the body is built.

**Suggestion:** omit the key when the value is nil (or send `""`, which is HubSpot's documented way to clear a property), e.g. build the properties hash and `.compact` it, or merge the key conditionally. Since `create_body` and `update_body` now duplicate the same properties hash, this could also be a good moment to extract a shared private `invoice_properties` method and handle the nil case in one place.

osmarluz (approved):
🚀

PR #5846, opened 2026-07-02
perf(charges): Do not compute charge model for unused filters

jdenquin:
## Context

  On subscriptions with a large number of charge filters, computing current usage is slow even when the
  billing-period pre-filtering (`Events::BillingPeriodFilterService`) works correctly. A production trace showed a
  `LifetimeUsages::RecalculateAndCheckJob` taking ~27s for a subscription with **zero events in the period**:
  `Fees::ChargeService` still iterates every filter (~1,550 here), paying one sequential Redis cache lookup (~9ms
  each, ~15.6s total) plus a bypassed aggregation per filter, only to end up hydrating the same zero-units fee in
  memory.

  ## Description

  - In `Fees::ChargeService`, skip the cache round-trip and the bypassed aggregation for filters that
  `filtered_aggregations` reports as having no usage, and hydrate the zero-units fee directly (same code path and
  output as before).
  - The skip is scoped to `current_usage` only:
    - Invoicing keeps the full path, as adjusted fees on draft invoices can target filters without usage, and
  true-up fees are unaffected.
    - Recurring metrics never skip, mirroring `should_bypass_aggregation?`, as usage carries over from previous
  periods.
  - All charge models return a zero amount for zero units (flat/fixed amounts only apply with usage), so no fee
  amount can change.

  Added scenario specs covering the pre-filtering end to end: zero events, partial filter usage with a catch-all
  bucket, and a recurring metric with usage carried over from a previous period.

### DIFF
--- a/app/services/charge_filters/event_matching_service.rb
+++ b/app/services/charge_filters/event_matching_service.rb
@@ -14,9 +14,8 @@ module ChargeFilters
     def call
       # NOTE: Find all filters matching event properties
       matching_filters = filters.select do |filter|
-        filter.to_h.all? do |key, values|
-          applicable_event_properties.key?(key) &&
-            (applicable_event_properties[key].to_s.in?(values) || values == [ChargeFilterValue::ALL_FILTER_VALUES])
+        filter.to_h_with_all_values.all? do |key, values|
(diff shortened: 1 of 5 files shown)


vincent-pochet (approved):
🚀

groyoh (approved):

PR #5855, opened 2026-07-03
feat(superset): add guest token mutation

ancorcruz:
## Context

Analytics dashboards embed Apache Superset through an iframe, authorized by a short-lived guest token (5 minutes by default). While a dashboard stays open the embed SDK must renew that token before it expires. The only way to obtain one was the `supersetDashboards` query, which re-authenticates, lists every dashboard, ensures each embedded config and mints a token for all of them — far too heavy to run on a recurring timer just to refresh a single token.

## Description

Add a `createSupersetGuestToken` mutation that mints one fresh guest token for a single dashboard, scoped to the current organization through a row-level security clause. The clause is always built server-side from the authenticated organization, so the dashboard id argument cannot widen data access; the mutation requires the `analytics:view` permission. It is a mutation rather than a query because each call issues a new, non-idempotent token that must never be served from a client cache.

Extract the shared Superset API plumbing (configuration checks, admin authentication, CSRF handling and header builders) into an `Auth::Superset::Client` concern, reused by both the existing dashboards service and the new `Auth::Superset::GuestTokenService`.


app/services/auth/superset/guest_token_service.rb
```diff
@@ -0,0 +1,81 @@
...
+        @http_client = nil
+
+        super()
+      end
+
+      def call
+        ensure_superset_configured
+        return result unless result.success?
+
+        auth_result = authenticate_with_api
```
osmarluz, line 29: @ancorcruz does Superset enforce any rate limits that would make this fail if we have multiple requests being fired very close to one another? It's good to keep an eye on this once we have multiple organizations using the analytics dashboard

osmarluz (approved):

app/services/auth/superset/guest_token_service.rb
ancorcruz: Good catch, worth watching. Each mint currently does a 3-request handshake against Superset (login → csrf → guest_token), and since the embed refreshes the token roughly every 5 min per open analytics session, that runs periodically per active user. The existing dashboards query has done login-on-every-call since December without hitting limits, so at our current handful of analytics orgs we're fine — but it won't scale linearly.

Worth flagging that all these requests come from the Lago API servers' IPs, not the customers', so if Superset throttles per-IP it's effectively one shared bucket across all orgs rather than per-tenant.

None of it blocks this PR, but as a follow-up I'd like to cache the admin login/CSRF server-side so a refresh becomes a single request (also makes the mutation faster), add a retry if Superset ever throttles us, and check with the team on whether request throttling is enabled at Superset and whether we can lengthen the guest-token lifetime so we refresh less often.

PR #5857, opened 2026-07-03
customer: discard provider customer on payment provider removal (API path)

mikeh-lago:
Recreates the API-side fix from the accidentally-closed #5833 (its branch was deleted). On payment-provider removal via the API upsert path, discards the old provider customer and clears payment_provider_code, mirroring the existing payment-methods discard. API path only; the UI path is handled in #5853.


app/services/customers/upsert_from_api_service.rb
```diff
@@ -270,6 +270,8 @@ def handle_api_billing_configuration(customer, new_customer)
       customer.save!
 
       if old_provider_customer && billing.key?(:payment_provider) && billing[:payment_provider].nil?
+        old_provider_customer.discard!
+        customer.update!(payment_provider_code: nil)
         discard_payment_methods(old_provider_customer.payment_methods)
       end
```
osmarluz, line 278: Avoid the extra DB trip by setting the `payment_provider_code` to `nil` before saving the customer, like shown below:


```suggestion
      removing_provider = old_provider_customer && billing.key?(:payment_provider) && billing[:payment_provider].nil?
      customer.payment_provider_code = nil if removing_provider

      customer.save!

      if removing_provider
        old_provider_customer.discard!
        discard_payment_methods(old_provider_customer.payment_methods)
      end
```

osmarluz (approved):

Issue #5481, opened 2026-05-08
[BUG]: migration always tries to create DB


jerbob92:
**Describe the bug**
When running the migrations it will always try to create the DB first, even if it already exists, which fails when the Lago DB user does not have access to create DB's.

**To Reproduce**
Try to run `./scripts/migrate.sh` with a DB user that does not have admin rights.

This is probably because it [always executes](https://github.com/getlago/lago-api/blob/main/scripts/migrate.sh#L8) `bundle exec rake db:create` before running `bundle exec rails db:migrate`.

**Expected behavior**
It should not fail on DB creation, when the DB creation already exists.

**Screenshots**
-

**Additional context**
-

**Version**
- Lago: v1.45.0

groyoh:
@jerbob92 thanks for reporting this. We'll have a look at what we can do to resolve this issue.

vincent-pochet:
This issue has been fixed with https://github.com/getlago/lago-api/pull/6192
Thanks again for reporting it

PR #5858, opened 2026-07-03
Fix migration script for existing databases

sudhanshu112233shukla:
## Summary
- Replace the unconditional `db:create` + `db:migrate` sequence in `scripts/migrate.sh` with `rails db:prepare`.
- Keep the existing role seeding and optional organization seeding behavior unchanged.

## Why
`scripts/migrate.sh` currently always runs `db:create` before migrations outside staging. That fails for deployments where the Lago database already exists but the Lago database user does not have permission to create databases.

`db:prepare` is the Rails task intended for this case: it migrates an existing database and only attempts setup/create behavior when the database is missing.

Fixes getlago/lago-api#5481.

### DIFF
--- a/scripts/migrate.sh
+++ b/scripts/migrate.sh
@@ -5,8 +5,7 @@ if [ "$RAILS_ENV" == "staging" ]
 then
   bundle exec rake db:prepare
 else
-  bundle exec rake db:create
-  bundle exec rails db:migrate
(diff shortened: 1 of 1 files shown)


vincent-pochet:
The issue has been addressed with https://github.com/getlago/lago-api/pull/6192. Closing this PR

PR #5859, opened 2026-07-06
[STA-9] misc: Define more service Results

vincent-pochet:
## Context

This PR is part of the long standing epic required to enable YJIT on Lago by removing all remaining usage of `OpenStruct`

## Description

The goal of this changes is to define the `Result` for a lot of new service class to avoid using the `LegacyResult` (an `OpenStruct`) that will be definitely removed when all services will be migrated to the new approach.

### DIFF
--- a/app/services/api_keys/create_service.rb
+++ b/app/services/api_keys/create_service.rb
@@ -2,6 +2,8 @@
 
 module ApiKeys
   class CreateService < BaseService
+    Result = BaseResult[:api_key]
+
     def initialize(params)
       @params = params
       super
--- a/app/services/applied_coupons/lock_service.rb
+++ b/app/services/applied_coupons/lock_service.rb
@@ -2,6 +2,8 @@
 
 module AppliedCoupons
   class LockService < BaseService
+    Result = BaseResult
+
     def initialize(customer:)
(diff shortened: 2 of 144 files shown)


app/services/commitments/fetch_invoices_service.rb
groyoh, line 1: I believe neither this file nor its children are tested. Could you either add tests for those to test the result change or remove from this PR  ?

app/services/fees/destroy_service.rb
groyoh, line 1: I believe this file is not tested. Could you either add tests for it to test the result change or remove from this PR  ?

app/services/integrations/salesforce/invoices/sync_service.rb
groyoh, line 1: I believe this file is not tested. Could you either add tests for it to test the result change or remove from this PR ?

app/services/credit_notes/lock_service.rb
groyoh, line 1: I believe this file is not tested. Could you either add tests for it to test the result change or remove from this PR ?

groyoh (approved):

rsempe (approved):

PR #5860, opened 2026-07-06
feat(quotes): validate one-off quote versions

toommz:
## Context

Implements the validation part for one-off quotes.

## Description

Validation for `one_off` quotes is split in two layers, composed by `QuoteVersions::Validators::OneOffValidator` and dispatched by order type via `QuoteVersions::Validators.for`:

- **Structural** (`OneOff::StructuralValidator`): validates the `billing_items` shape against a JSON Schema (new `json_schemer` dependency).
- **Business** (`OneOff::BusinessValidator`): rules needing DB or context.

Validation runs at three write points: `Quotes::CreateService`/`QuoteVersions::CreateService` and `QuoteVersions::UpdateService` at `:update` scope, `QuoteVersions::ApproveService` at `:approve` scope

Errors are returned as pointer-keyed codes so the frontend can map them to a specific line:

```json
{
  "billing_items.addons.0.payload.units": ["invalid_value"],
  "billing_items.addons.1.id": ["add_on_not_found"],
  "currency": ["value_is_mandatory"]
}
```

### DIFF
--- a/app/services/quote_versions/validators.rb
+++ b/app/services/quote_versions/validators.rb
@@ -5,7 +5,7 @@ module QuoteVersions
     def self.for(result, quote_version:, scope:)
       case quote_version.quote.order_type
       when Quote::ORDER_TYPES[:one_off]
-        OneOffService.new(result, quote_version:, scope:)
+        OneOffValidator.new(result, quote_version:, scope:)
       end
     end
   end
--- a/app/services/quote_versions/validators.rb
+++ b/app/services/quote_versions/validators.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module QuoteVersions
+  module Validators
+    def self.for(result, quote_version:, scope:)
+      case quote_version.quote.order_type
+      when Quote::ORDER_TYPES[:one_off]
+        OneOffService.new(result, quote_version:, scope:)
+      end
+    end
+  end
+end
--- a/app/services/quote_versions/validators/one_off/structural_validator.rb
+++ b/app/services/quote_versions/validators/one_off/structural_validator.rb
@@ -3,7 +3,7 @@
 module QuoteVersions
   module Validators
     module OneOff
-      class StructuralService
+      class StructuralValidator
         attr_reader :errors
 
         def initialize(billing_items:, scope:)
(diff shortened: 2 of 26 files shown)


app/services/quotes/create_service.rb
```diff
@@ -42,6 +42,8 @@ def call
       result
     rescue ActiveRecord::RecordInvalid => e
       result.record_validation_failure!(record: e.record)
+    rescue BaseService::ValidationFailure => e
+      e.result
```
vincent-pochet: To respect the return contract, maybe you should do:
```suggestion
      result.fail_with_error!(e)
```

app/services/quote_versions/validators/one_off/structural_validator.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module QuoteVersions
+  module Validators
+    module OneOff
+      class StructuralValidator
```
vincent-pochet: I'm wondering if by inheriting from `BaseValidator`, and maybe extending it if needed, we could reduce the duplicated `@errors` or `add_error()` definitions and enforce the validation pattern

app/services/quote_versions/validators.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module QuoteVersions
+  module Validators
+    def self.for(result, quote_version:, scope:)
```
vincent-pochet, line 5: I like the `for` name for this factory. We are usually using `new_instance` for similar pattern, but I always found the name far from pretty. I think we should push for `for`!

app/services/quotes/create_service.rb
toommz: 3b3de2401eee4eda7e7d39206d7ad422baa7ee04

app/services/quote_versions/validators/one_off/structural_validator.rb
toommz: Totally! Done in 9fe2db72bce2dfbcb09f198c70e629430a813628

app/services/quote_versions/validators.rb
toommz: Oh I didn't spot that we already have some in the codebase. That's why I used another naming. But yeah I always saw this `for` method name before, hence its usage here.

vincent-pochet (approved):
🚀 Nice one!

aquinofb (approved):

PR #5861, opened 2026-07-06
fix(events): Correct pay-in-advance fees for events with same timestamp ClickHouse

D1353L:
## Context
This is a ClickHouse version of the fix in PR #5740.

With a graduated charge (0–2 units = 0 EUR, 3+ = 2 EUR/unit) and 3 events at the same second, the expected fees are 0, 0, 200 cents; instead each fee is billed 200 cents, tripling the total. The fee amounts also feed fee.created webhooks, so customers receive incorrect amounts.

## Description

Tie-break events sharing the boundary timestamp by transaction_id, giving each event a distinct position in the aggregation.

### DIFF
--- a/app/services/events/stores/utils/clickhouse_sql_helpers.rb
+++ b/app/services/events/stores/utils/clickhouse_sql_helpers.rb
@@ -43,6 +43,25 @@ module Events
         def sql_condition(template, *values)
           ActiveRecord::Base.sanitize_sql_for_conditions([template, *values])
         end
+
+        # When the boundary is the pay-in-advance event's own timestamp, tie-break events
+        # sharing it by transaction_id so each gets a distinct position instead of all being
+        # priced as the batch's last unit. transaction_id trails every events sort key, so
+        # the comparison stays index-aligned. Any other boundary keeps the plain <=.
+        def upper_timestamp_boundary_sql(to_datetime, prefix: "")
+          boundary_transaction_id = (filters[:event]&.transaction_id if to_datetime == boundaries[:max_timestamp])
+
+          if boundary_transaction_id
+            sql_condition(
+              "(#{prefix}timestamp < ? OR (#{prefix}timestamp = ? AND #{prefix}transaction_id <= ?))",
+              to_datetime,
+              to_datetime,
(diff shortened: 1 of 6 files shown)


app/services/events/stores/utils/clickhouse_sql_helpers.rb
```diff
@@ -43,6 +43,25 @@ def decimal_literal(value)
...
+        # When the boundary is the pay-in-advance event's own timestamp, tie-break events
+        # sharing it by transaction_id so each gets a distinct position instead of all being
+        # priced as the batch's last unit. transaction_id trails every events sort key, so
+        # the comparison stays index-aligned. Any other boundary keeps the plain <=.
+        def upper_timestamp_boundary_sql(to_datetime, prefix: "")
+          boundary_transaction_id = (filters[:event]&.transaction_id if to_datetime == boundaries[:max_timestamp])
+
+          if boundary_transaction_id
+            sql_condition(
+              "(#{prefix}timestamp < ? OR (#{prefix}timestamp = ? AND #{prefix}transaction_id <= ?))",
```
groyoh, line 56: Transaction IDs aren't sorted. So shouldn't it be ?

```suggestion
              "(#{prefix}timestamp < ? OR (#{prefix}timestamp = ? AND #{prefix}ingested_at <= ?))",
```

D1353L: A few findings of my research:
1. If we receive a few events with the same timestamp, for graduated billing strict ordering doesn't matter. The idea is to have events in distinct positions and not process all of them as the last one. Ordering by `transaction_id` is enough for this.
2. This method is used only for `events_enriched` and `events_enriched_expanded` tables, which have `enriched_at` instead of `ingested_at`.
3. The `enriched_at` value isn't consistent as it gets re-stamped on re-enrichment - the same event will have a fresh value in a new record. This will reshuffle events and break strict ordering.
4. Ordering by `enriched_at` would add subquery complexity and possible performance impact. Barely worth it given the above.

WDYT?

vincent-pochet (approved):

PR #5864, opened 2026-07-06
fix: discard Idempotency::IdempotencyError for RecalculateAndCheckJob

lovrocolic:
## Context

`LifetimeUsages::RecalculateAndCheckJob` runs from two paths that don't share the Sidekiq uniqueness lock (the clock's perform_later and the inline perform_now from subscription activity), so two runs can execute concurrently for the same lifetime usage. When they do, both attempt the same progressive billing invoice; the loser raises `Idempotency::IdempotencyError` at the invoice-level idempotency guard because the winner already created it.

## Description

Add `discard_on Idempotency::IdempotencyError` to the job so the losing run is dropped instead of retried, stopping the dead-job accumulation.


toommz (approved):

mariohd (approved):

vincent-pochet (approved):
Great!

PR #5872, opened 2026-07-06
perf(events): add org_id + timestamp partial index

groyoh:
## Context

`GET /api/v1/events` lists events through `EventsQuery`, which runs `organization_id = ? ORDER BY timestamp DESC, transaction_id ASC` on the `events` table. No index covers this pattern: the `organization_id`-leading indexes are on other columns (`created_at`, `transaction_id`, `code`) and the remaining composite indexes are led by `external_subscription_id`. Postgres falls back to a sequential scan and an in-memory sort over all of an organization's events, and on high-volume organizations the query exceeds the 60s statement timeout with `ActiveRecord::QueryCanceled`.

## Description

This adds a partial index on `events`, built with `CREATE INDEX CONCURRENTLY`:

```sql
(organization_id, timestamp DESC) WHERE deleted_at IS NULL
```

The predicate matches the model's `default_scope` (Discard's `kept`), so all reads through `EventsQuery` qualify for the index.

The `transaction_id ASC` tiebreaker is intentionally not part of the index: ties on identical timestamps are resolved with an incremental sort on top of the presorted index scan, which is negligible (verified with `EXPLAIN`, the plan shows `Presorted Key: timestamp`).

Fixes ING-398.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -91,6 +91,7 @@ end
 #  index_events_on_organization_id                     (organization_id)
 #  index_events_on_organization_id_and_code            (organization_id,code)
 #  index_events_on_organization_id_and_created_at      (organization_id,created_at DESC) WHERE (deleted_at IS NULL)
+#  index_events_on_organization_id_and_timestamp       (organization_id,timestamp DESC) WHERE (deleted_at IS NULL)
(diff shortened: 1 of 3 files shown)


vincent-pochet (approved):

groyoh:
Thanks @vincent-pochet 👍

PR #5873, opened 2026-07-06
[STA-9] misc: Define more service Results

vincent-pochet:
## Context

Following https://github.com/getlago/lago-api/pull/5859, this PR is part of the long standing epic required to enable YJIT on Lago by removing all remaining usage of `OpenStruct`

## Description

The goal of this changes is to define the `Result` for a lot of new service class to avoid using the `LegacyResult` (an `OpenStruct`) that will be definitely removed when all services will be migrated to the new approach.

### DIFF
--- a/app/services/payment_requests/payments/flutterwave_service.rb
+++ b/app/services/payment_requests/payments/flutterwave_service.rb
@@ -6,8 +6,6 @@ module PaymentRequests
       include Customers::PaymentProviderFinder
       include Updatable
 
-      Result = BaseResult[:payable, :payment, :payment_url]
-
       def initialize(payable = nil)
         @payable = payable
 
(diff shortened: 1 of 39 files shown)


app/services/payment_requests/payments/flutterwave_service.rb
```diff
@@ -6,8 +6,6 @@ class FlutterwaveService < BaseService
       include Customers::PaymentProviderFinder
       include Updatable
 
-      Result = BaseResult[:payable, :payment, :payment_url]
```
vincent-pochet, line 9: This service does not match the "call" pattern for now and the result was added by mistake. It works but this result contains all possible returns from the different public methods defined in the class making it a bit misleading

toommz (approved):

PR #5874, opened 2026-07-07
feat(purchase_order_number) Expose purchase_order_number to graphql mutations and objects

tiagolupepic:
This commit adds the foundation of purchase_order_number

It's adding the `purchase_order_number` to some tables:
- wallets
- recurring_transaction_rules
- subscriptions
- wallet_transactions

Also, it's changing the Graphql schema to provide this new field to some objects and mutations.
For now, we're not tying the new field in each service to persist the data. This will be addressed
in the following PR's.


spec/factories/invoices.rb
```diff
@@ -45,6 +45,10 @@
       end
     end
 
+    trait :with_purchase_order_number do
+      purchase_order_number { "PO-123" }
```
toommz, line 49: Could it help to use a sequence for this? The problem I see is this "magic number" then leaks to the specs, while not carrying real meaning (it's just a "static random" string).


```suggestion
      sequence(:purchase_order_number) { |n| "PO-#{n.to_s.rjust(3, '0'}" }
```

tiagolupepic: I'd prefer to keep is static, because magic numbers for this will create flaky specs. I can eventually change this in the following PR's if we need it

endenis (approved):

PR #5877, opened 2026-07-07
feat(fc-units-override): Adapt find_already_paid_units method to consider already billed usage charges for migrated plans

osmarluz:
## Context

Plans migrated from usage-based recurring charges to pay-in-advance fixed charges (e.g. seats billed via a recurring sum metric) have their current billing period already paid through fees generated by the old usage charge. When the fixed charge units are then updated with `apply_units_immediately: true` before the next renewal, `find_already_paid_units` only counts `fixed_charge` fees, so the delta is computed against 0 and the customer is billed for all units instead of only the increase.

## Description

Behind the `fixed_charge_usage_delta_migration` organization feature flag, `Fees::BuildPayInAdvanceFixedChargeService` also counts fees from discarded recurring usage charges on the subscription's plan as already-paid units when they belong to the current billing period (matched on `charges_from_datetime` = the fixed charge period start, with true-up fees excluded).

The behavior is self-neutralizing: after a subscription's first renewal on the fixed charge, the old usage fees no longer match the current period. The flag can be removed once all migrated subscriptions have renewed.


app/services/fees/build_pay_in_advance_fixed_charge_service.rb
```diff
@@ -76,7 +76,32 @@ def find_already_paid_units(boundaries)
...
+      already_paid_units = existing_fees.sum(:units).to_d
+
+      if organization.feature_flag_enabled?(:fixed_charge_usage_delta_migration)
+        already_paid_units += superseded_usage_charge_units(boundaries)
+      end
+
+      already_paid_units
+    end
+
+    def superseded_usage_charge_units(boundaries)
```
osmarluz, line 88: **Context for this method**

It supports migrating a plan from usage-based recurring billing to fixed charges (e.g. seats billed via a recurring `sum_agg` charge → a pay-in-advance fixed charge with per-subscription units overrides).

During such a migration the old charge is discarded mid-period, which creates a transition gap: the current period's units were already paid through the old charge's fees (`fee_type: :charge`), but the delta computation above only counts `fixed_charge` fees. Without this, the first units update with `apply_units_immediately: true` after the migration would see `already_paid = 0` and bill the customer's full unit count again (e.g. 3 already-paid seats + 1 new seat → billed 4 instead of 1).

The behavior is self-neutralizing: after the subscription's first renewal on the fixed charge, the old fees belong to a past period and stop matching, so the method contributes 0. That's why it's gated behind the `fixed_charge_usage_delta_migration` feature flag and can be removed entirely once all migrated subscriptions have renewed once.

rsempe (approved):

ancorcruz (approved):
Looks good, however we have to bear in mind this only works if the subscription has one and only one fixed charge, or it will bill twice it, won't it?

Charge to Fixed Charge match is best guess.

osmarluz:
> Looks good, however we have to bear in mind this only works if the subscription has one and only one fixed charge, or it will bill twice it, won't it?
> 
> Charge to Fixed Charge match is best guess.

Yes, it's assuming that while a subscription is still in its migration window (hasn't renewed since the usage charge was discarded):

- No new pay-in-advance fixed charge is added to the plan and then given a units update with `apply_units_immediately: true` (plan-level or per-sub override)
- No new pay-in-advance fixed charge is created through the plan-update endpoint with `apply_units_immediately: true`

If either happens, it actually under-bills the new units. Adding a fixed charge without immediate-apply is safe: it first bills at renewal via the periodic path, which doesn't use this baseline.

PR #5878, opened 2026-07-07
fix(stripe-webhooks): Fix deadlock errors

osmarluz:
## Context

Stripe webhook jobs — `PaymentProviders::Stripe::HandleEventJob` (`customer.updated` / `setup_intent.succeeded`) and `PaymentProviders::Stripe::Customers::FetchDefaultPaymentMethodJob` (at customer creation) — were dead-lettering with `ActiveRecord::Deadlocked`. They all set the customer's default payment method through the shared `PaymentMethods::SetAsDefaultService`, which demotes the customer's other payment methods and promotes the target within one transaction. When two events for the same customer are processed concurrently, the transactions lock the same `payment_methods` rows in opposite order and deadlock; with `retry: 0` and no matching `retry_on`, the losing job dies and its update never lands.

## Description

- Serialize the default flip per customer: `SetAsDefaultService` now runs inside a `Customers::LockService` advisory lock (new `:payment_method` scope), so concurrent same-customer flips can no longer interleave.
- `UpdatePaymentMethodService` now writes the `StripeCustomer` settings and the default payment-method record in a single transaction, so a failure can't leave them out of sync.
- `HandleEventJob` and `FetchDefaultPaymentMethodJob` now `retry_on Customers::FailedToAcquireLock, ActiveRecord::Deadlocked`, so lock contention / deadlock retries instead of dead-lettering.


app/services/payment_methods/set_as_default_service.rb
```diff
@@ -18,8 +18,10 @@ def call
       end
 
       ActiveRecord::Base.transaction do
-        payment_method.customer.payment_methods.where.not(id: payment_method.id).update_all(is_default: false) # rubocop:disable Rails/SkipsModelValidations
-        payment_method.update!(is_default: true)
+        Customers::LockService.call(customer: payment_method.customer, scope: :payment_method) do
+          payment_method.customer.payment_methods.where.not(id: payment_method.id).update_all(is_default: false) # rubocop:disable Rails/SkipsModelValidations
```
ancorcruz: lets also touch the `updated_at` timestamp of the updated records.

app/services/payment_methods/set_as_default_service.rb
```diff
@@ -18,8 +18,10 @@ def call
       end
 
       ActiveRecord::Base.transaction do
-        payment_method.customer.payment_methods.where.not(id: payment_method.id).update_all(is_default: false) # rubocop:disable Rails/SkipsModelValidations
-        payment_method.update!(is_default: true)
+        Customers::LockService.call(customer: payment_method.customer, scope: :payment_method) do
```
ancorcruz, line 21: Non-blocking, but worth a conscious call: this service is also invoked synchronously outside the jobs — Mutations::PaymentMethods::SetAsDefault (app/graphql/mutations/payment_methods/set_as_default.rb:19) and Api::V1::Customers::PaymentMethodsController#create (app/controllers/api/v1/customers/payment_methods_controller.rb:29).

Now that the flip runs behind the advisory lock and SetAsDefaultService propagates FailedToAcquireLock (rather than returning a result failure), under same-customer contention those API paths would raise an unhandled exception → 500, where before they'd just succeed. The jobs recover via retry_on; the sync callers have no equivalent.

We can handle this by...
- Rescue BaseLockService::FailedToAcquireLock inside SetAsDefaultService and return result.service_failure!(...), so all callers get a graceful failure consistent with the service pattern (jobs can still map that to a retry if needed); or
- Leave as-is if we consider same-customer lock contention from the synchronous API negligible — just calling it out so it's an explicit decision rather than a side effect.

Either is fine by me; flagging so it's intentional.

app/services/customers/lock_service.rb
```diff
@@ -15,7 +15,7 @@ module Customers
   #   locking (lock_version), concurrent updates will raise StaleObjectError.
   #
   class LockService < BaseLockService
-    VALID_SCOPES = %i[prepaid_credit credit_note coupon].freeze
+    VALID_SCOPES = %i[prepaid_credit payment_method credit_note coupon].freeze
```
ancorcruz, line 18: There is a test gap in `spec/services/customers/lock_service_spec.rb` as it iterates over prepaid_credit credit_notes and coupon but not the newly added payment_method.

ancorcruz (approved):
Looks great, just left a few comments.

PR #5886, opened 2026-07-08
perf(events): drop redundant organization_id index

groyoh:
## Context

The `events` table has a single-column index on `organization_id`. That column is also the leading column of several composite indexes on the table (`index_events_on_organization_id_and_code`, `index_events_on_organization_id_and_created_at`, `index_events_on_organization_id_and_timestamp`, `index_events_on_organization_id_and_transaction_id`, and the unique `index_unique_transaction_id`), so lookups on `organization_id` alone stay covered without it. On a high-volume table, the extra index only adds write amplification and storage.

## Description

This adds an up-only migration dropping `index_events_on_organization_id` with `DROP INDEX CONCURRENTLY` (via `disable_ddl_transaction!`) to avoid locking the table. The `Event` model schema annotation was refreshed accordingly.

Note that `index_events_on_organization_id_and_code` is non-partial, so queries touching soft-deleted events by organization (for example `with_discarded` reads in rake tasks) keep index support after this drop. Dropping that index is a separate follow-up and will need its own coverage check.

Fixes ING-399

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -88,7 +88,6 @@ end
 #  idx_events_billing_lookup                           (external_subscription_id,organization_id,code,timestamp) WHERE (deleted_at IS NULL)
 #  idx_events_for_distinct_codes                       (external_subscription_id,organization_id,timestamp) WHERE (deleted_at IS NULL)
 #  index_events_on_created_at                          (created_at) WHERE (deleted_at IS NULL)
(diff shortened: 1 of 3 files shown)


toommz (approved):
🧹

jdenquin (approved):

aquinofb (approved):

PR #5887, opened 2026-07-08
feat(orders): execute OneOff orders

toommz:
## Description

Signed one-off orders now actually get billed. When the execution mode is `execute_in_lago`, Lago turns the order into a one-off invoice, either immediately at signing or at the scheduled `execute_at` date.

If billing fails, the order moves to a `failed` status with the `errors` recorded, and won't be retried automatically. Retrying is a deliberate user action. The execution outcome (when, how, resulting invoice) is exposed on the API and GraphQL so users can audit what happened to each order.

### Reviewer notes

Two changes touch the shared one-off invoicing path:

- post-creation jobs in `Invoices::CreateOneOffService` move to `perform_after_commit` because the service now runs inside the order execution transaction
- a `with_discarded_add_ons:` opt-in flag is added since add-ons can be discarded between signing and execution

### DIFF
--- a/app/services/invoices/create_one_off_service.rb
+++ b/app/services/invoices/create_one_off_service.rb
@@ -4,7 +4,7 @@ module Invoices
   class CreateOneOffService < BaseService
     Result = BaseResult[:invoice, :payment_method]
 
-    def initialize(customer:, currency:, fees:, timestamp:, skip_psp: false, voided_invoice_id: nil, payment_method_params: nil, invoice_custom_section: {}, billing_entity_id: nil, billing_entity_code: nil, purchase_order_number: nil)
+    def initialize(customer:, currency:, fees:, timestamp:, skip_psp: false, voided_invoice_id: nil, payment_method_params: nil, invoice_custom_section: {}, billing_entity_id: nil, billing_entity_code: nil, purchase_order_number: nil, with_discarded_add_ons: false)
       @customer = customer
       @currency = currency || customer&.currency
       @fees = fees
@@ -16,6 +16,7 @@ module Invoices
       @billing_entity_id = billing_entity_id
       @billing_entity_code = billing_entity_code
       @purchase_order_number = purchase_order_number
+      @with_discarded_add_ons = with_discarded_add_ons
 
       super(nil)
     end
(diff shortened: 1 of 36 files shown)


app/services/invoices/create_one_off_service.rb
```diff
@@ -76,11 +77,11 @@ def call
       return result if tax_deferred
 
       unless invoice.closed?
-        Utils::SegmentTrack.invoice_created(invoice)
-        SendWebhookJob.perform_later("invoice.one_off_created", invoice)
-        GenerateDocumentsJob.perform_later(invoice:, notify: should_deliver_email?)
-        Integrations::Aggregator::Invoices::CreateJob.perform_later(invoice:) if invoice.should_sync_invoice?
-        Integrations::Aggregator::Invoices::Hubspot::CreateJob.perform_later(invoice:) if invoice.should_sync_hubspot_invoice?
+        after_commit { Utils::SegmentTrack.invoice_created(invoice) }
```
toommz, line 80: `Utils::SegmentTrack.invoice_created` calls `perform_later` behind the hood, but changing the implementation is outside the scope of this PR. Wrapping it in an `after_commit` block here before doing the changes in `Utils::SegmentTrack` in another PR.

spec/services/orders/one_off/execute_service_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+  let(:quote) { create(:quote, organization:, customer:, order_type: :one_off) }
+  let(:quote_version) { create(:quote_version, :approved, quote:, organization:, currency: "EUR", billing_items:) }
+  let(:order_form) { create(:order_form, :signed, organization:, customer:, quote_version:) }
+  let(:order) { create(:order, organization:, customer:, order_form:, execution_mode:) }
+  let(:execution_mode) { :execute_in_lago }
+
+  # Add-ons are resolved by id only outside api context, and CurrentContext
+  # leaks across spec files (no global reset), so pin it.
+  before { CurrentContext.source = nil }
+
```
toommz, line 37: This should belong in `spec/spec_helper.rb` as:

```ruby
# spec/spec_helper.rb:162
# In config.before global hook

ActiveSupport::CurrentAttributes.reset_all
```

But outside the scope of this PR, as it might have some side effects to fix at the same time.

PR #5888, opened 2026-07-08
misc: remove multiple_payment_methods feature flag

lovrocolic:
## Context

Multiple payment method is deployed and enabled for all cloud and OSS users.

## Description

This PR removes feature flag that is not used anymore.


spec/models/feature_flag_spec.rb
toommz, line 1: 👍

toommz (approved):

groyoh (approved):

PR #5889, opened 2026-07-08
perf(features): Introduce materialized view for Features#subscriptions_count

vankiru:
## Context

The subscription count on the Features page is currently incorrect. Right now, it only counts subscriptions that get a feature through their plan. It doesn't include subscriptions that have the feature assigned directly on their Entitlements page, and it doesn't account for subscriptions where the feature has been removed.

## Description

This PR adds a materialized view that precomputes and stores the subscription count for every non-deleted feature that has at least one associated subscription.

The count is calculated in four steps for each feature:

1. Count subscriptions that get the feature through their plan. If a subscription has an overridden plan, it's counted towards the parent plan's features, since plan overrides can't have their own entitlements.
2. Count subscriptions that have the feature assigned directly on their Entitlements page. This step also removes duplicates when a subscription gets the same feature both through its plan and via a direct assignment.
3. Count subscriptions where the feature has been manually removed.
4. Calculate the final count as:

   `plan subscriptions + direct subscriptions - feature removals`

## Refreshing the Materialized View

On a database with around **2.5M subscriptions** and **4M entitlements**, refreshing the view takes about **3 seconds**. On databases roughly **10× larger**, it may take up to **30 seconds**.

To balance performance with data freshness, the view is refreshed every **10 minutes**. This helps avoid refresh-time spikes while keeping the counts reasonably up to date.

### DIFF
--- a/app/models/entitlement/feature.rb
+++ b/app/models/entitlement/feature.rb
@@ -27,8 +27,12 @@ module Entitlement
     def subscriptions_count
       return @subscriptions_count if defined?(@subscriptions_count)
 
-      base_scope = Subscription.joins(:plan).where(status: [:active, :pending])
-      base_scope.where(plan: plans).or(base_scope.where(plan: {parent: plans})).count
+      subscriptions_count = SubscriptionsCountQuery.call(
+        organization:,
+        filters: {feature_ids: [id]}
+      )
+
+      subscriptions_count[id]
     end
 
     def self.preload_subscriptions_count(organization, features)
--- a/app/models/entitlement/feature.rb
+++ b/app/models/entitlement/feature.rb
@@ -32,7 +32,7 @@ module Entitlement
         filters: {feature_ids: [id]}
       )
 
-      subscriptions_count[id]
+      subscriptions_count[id] || 0
     end
 
     def self.preload_subscriptions_count(organization, features)
--- a/clock.rb
+++ b/clock.rb
@@ -206,6 +206,10 @@ module Clockwork
       .perform_later
   end
 
+  every(10.minutes, "schedule:refresh_entitlement_features_subscription_count") do
+    Scenic.database.refresh_materialized_view(:entitlement_features_subscriptions_count, concurrently: true)
+  end
+
   # NOTE: Enable wallets and lifetime usage refresh from the events-processor
   if ENV["LAGO_REDIS_STORE_URL"].present? && ENV["LAGO_CLICKHOUSE_ENABLED"].present?
     every(10.seconds, "schedule:refresh_flagged_subscriptions") do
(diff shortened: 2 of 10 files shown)


app/models/entitlement/feature.rb
```diff
@@ -27,8 +27,12 @@ def self.ransackable_attributes(_auth_object = nil)
...
       return @subscriptions_count if defined?(@subscriptions_count)
 
-      base_scope = Subscription.joins(:plan).where(status: [:active, :pending])
-      base_scope.where(plan: plans).or(base_scope.where(plan: {parent: plans})).count
+      subscriptions_count = SubscriptionsCountQuery.call(
+        organization:,
+        filters: {feature_ids: [id]}
+      )
+
+      subscriptions_count[id]
```
D1353L: Here is a possibility that `nil` is returned for feature with no subscriptions. This breaks GraphQL contract `field :subscriptions_count, Integer, null: false`

vankiru: Fixed 👍

clock.rb
```diff
@@ -214,4 +214,8 @@ module Clockwork
         .perform_later
     end
   end
+
+  every(10.minutes, "refresh:entitlement_features_subscription_count") do
+    Scenic.database.refresh_materialized_view(:entitlement_features_subscriptions_count, concurrently: true)
```
D1353L, line 219: consider moving this to a background job as running this sync would block the clock process. also the pattern in this file is to use async jobs

spec/queries/entitlement/feature/subscriptions_count_query_spec.rb
```diff
@@ -5,57 +5,252 @@
...
+        create(:entitlement, feature:, plan:)
+
+        create(:subscription_feature_removal, feature:, subscription:)
+      end
+
+      it "does not include subscriptions for which the feature is removed" do
+        result = subject.call
+
+        expect(result).to eq({feature.id => 1})
+      end
```
D1353L, line 170: I've found this scenario with Codex. Please consider if this is valid and needs to be fixed (this test currently fails)
 
```suggestion
      end
      
      context "with a removal for a feature no longer provided by the plan" do
        let(:other_plan) { create(:plan, organization:) }
        let(:subscription) { create(:subscription, plan:) }

        before do
          create(:subscription, plan: other_plan)
          create(:subscription_feature_removal, feature:, subscription:)
          create(:entitlement, feature:, plan:, deleted_at: Time.current)
          create(:entitlement, feature:, plan: other_plan)
        end

        it "does not subtract the stale removal from other plan subscriptions" do
          result = subject.call

          expect(result).to eq({feature.id => 1})
        end
      end
```

PR #5894, opened 2026-07-09
misc(lock): Add BaseLockService

vincent-pochet:
## Context

While working on adding the last missing `Result` definition to the service layer, it appeared that the service responsible from locking some records/table for update where not following the standard `call` pattern:
- They are all using the same approach:
  - defining a lock key
  - locking the the record around the call to the provided block
- They are not returning any `Result` as the purpose is only to take a block to execute in a locked transaction

## Description

In order to avoid defining useless `Result`, this pull request introduce a new `BaseLockService` that does not inherit from `BaseService`.
This new base class adds all required methods and field necessary to define a Lock on a specific table or record, making it easier to add a new one.
All existing Lock service are updated to follow this new pattern, removing a lot of logic duplication.

Since all the observability/monitoring logic were wired in the `BaseService`, the middleware pattern is extracted from it in a composable `Middlewares::Chain` module that allow us to reuse this logic without redefining it.

`Customers::FailedToAcquireLock` error is also renamed into `BaseLockService::FailedToAcquireLock`


app/services/applied_coupons/lock_service.rb
```diff
@@ -1,34 +1,23 @@
 # frozen_string_literal: true
 
 module AppliedCoupons
-  class LockService < BaseService
-    Result = BaseResult
-
+  class LockService < BaseLockService
```
groyoh: We should probably replace this service calls with `Customers::LockService`.

app/services/credit_notes/lock_service.rb
```diff
@@ -1,32 +1,23 @@
 # frozen_string_literal: true
 
 module CreditNotes
-  class LockService < BaseService
+  class LockService < BaseLockService
```
groyoh: We should probably replace this service calls with `Customers::LockService`.

app/services/base_lock_service.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+# Base class for advisory-lock services.
+#
+# This is a thin wrapper around a PostgreSQL advisory lock: it
+# yields the caller's block while the lock is held and returns the block's value.
+# It does not use the `result` pattern, so it neither inherits from BaseService nor defines a `Result`.
```
groyoh: Shouldn't it reuse the `result` pattern instead to stay consistent ?

app/services/base_lock_service.rb
```diff
@@ -0,0 +1,51 @@
...
+    @timeout_seconds = timeout_seconds
+    @transaction = transaction
+  end
+
+  def call
+    lock_owner.with_advisory_lock!(
+      lock_key,
+      timeout_seconds:,
+      transaction:,
+      disable_query_cache: true
```
groyoh, line 29: This base now passes `disable_query_cache: true` for every lock owner, but `Customers::LockService` and `Quotes::LockService` previously called `with_advisory_lock!` without it (only `AppliedCoupons` and `CreditNotes` set it). So the locked blocks for prepaid-credit and quote operations now run with the AR query cache disabled, which changes read behavior inside those transactions.

I guess it's fine though as it's probably what we want.

groyoh (approved):

app/services/base_lock_service.rb
vincent-pochet: Thanks for pointing it out. I should have added a comment about it.
Indeed, I did some lookup for this and conclude that disabling the cache inside the locked transaction should be the default behavior to avoid issues on "staled records"

app/services/base_lock_service.rb
vincent-pochet: All these lock services just handle the given block without returning any thing. All callers manages their own result so their is no real point to initialize "ghost" results that will never get evaluated. That's why I wanted to introduce this default based service. 

If we want to keep a result, I can keep make a clear inheritance from `BaseService` and assign the result of the `with_advisory_lock!` as the service result. On the bright side it would remove the need for refactoring the middleware pattern. What do you think?

groyoh: I believe it would avoid confusion. Either that or let's not suffix it with `Service`. Both are fine with me. Not strong opinion. Feel free to merge as is as well if it's too bothersome.

app/services/credits/applied_coupons_service.rb
```diff
@@ -19,7 +19,7 @@ def call
       # We're not locking individual coupons as that might lead to deadlocks.
       # This will also keep the lock for the shortest time possible, otherwise
       # we'd have to wait for the transaction to either rollback / commit.
-      AppliedCoupons::LockService.new(customer:).call do
+      Customers::LockService.call(customer:, scope: :coupon) do
```
groyoh, line 22: One note here: if we update this in place there may be a time window where we wouldn't prevent concurrency on coupon and credit note as the old and new locks would have different keys.

I believe it could be OK.

That said, the proper solution would be to do a dual lock and then in a second release to remove the previous lock. Not sure if worth the hassle. Up to you.

groyoh (approved):

app/services/credits/applied_coupons_service.rb
vincent-pochet: The risk is very low. I'll deploy it on a "low usage" period to limit it at maximum

PR #5897, opened 2026-07-09
fix(clickhouse): Retry total_count failure on GET /events

vincent-pochet:
## Description

This PR adds the auto retry management from `Events::Stores::Utils::ClickhouseConnection` to the `total_count` used for pagination on the `GET /events endpoint`

This should reduce the number of occurrence of:
```
ActiveRecord::ActiveRecordError 
Response: end of file reached (ActiveRecord::ActiveRecordError)
```


groyoh (approved):
Ugly but I don't have a better solution right now 😄

PR #5898, opened 2026-07-09
fix: add small changes in pdf templates used for historical invoices

lovrocolic:
## Context

Old versions of PDF templates are not compatible anymore with latest code.

## Description

This PR fixs fee breakdown bug and adds some I18n guards so that historical invoice can be correctly generated.

### DIFF
--- a/app/views/templates/invoices/v1.slim
+++ b/app/views/templates/invoices/v1.slim
@@ -490,7 +490,7 @@ html
           h2.invoice-details-title.title-2.mb-24 = I18n.t('invoice.details', resource: subscription.invoice_name)
           .body-1 = I18n.t('invoice.subscription')
           .mb-24.body-3
-            | #{I18n.t('invoice.date_from')} #{I18n.l(invoice_subscription(subscription.id).from_datetime_in_customer_timezone&.to_date, format: :default)} #{I18n.t('invoice.date_to')} #{I18n.l(invoice_subscription(subscription.id).to_datetime_in_customer_timezone&.to_date, format: :default)}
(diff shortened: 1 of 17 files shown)


toommz (approved):

groyoh (approved):
Any chance you could add view/scenarios tests for these view changes ? Check existing view tests for v4.

vincent-pochet (approved):
Looks good! 

Do you think it's possible to add specs for these historical invoices templates the same way we have for v4 today?

app/views/templates/invoices/v1.slim
```diff
@@ -490,7 +490,7 @@ html
           h2.invoice-details-title.title-2.mb-24 = I18n.t('invoice.details', resource: subscription.invoice_name)
           .body-1 = I18n.t('invoice.subscription')
           .mb-24.body-3
-            | #{I18n.t('invoice.date_from')} #{I18n.l(invoice_subscription(subscription.id).from_datetime_in_customer_timezone&.to_date, format: :default)} #{I18n.t('invoice.date_to')} #{I18n.l(invoice_subscription(subscription.id).to_datetime_in_customer_timezone&.to_date, format: :default)}
+            | #{I18n.t('invoice.date_from')} #{DateHelper.format(invoice_subscription(subscription.id).from_datetime_in_customer_timezone&.to_date)} #{I18n.t('invoice.date_to')} #{DateHelper.format(invoice_subscription(subscription.id).to_datetime_in_customer_timezone&.to_date)}
```
groyoh, line 493: Wouldn't this display `From to` ?

lovrocolic: In theory it could (maybe just one value will be blank). I didn't want add extra complexity for historical invoices and templates that are not used anymore. The goal was mostly to reduce noise from errors raised in old templates (errors are related to test accounts only) and to render PDF successsfuly.

PR #5899, opened 2026-07-09
chore(db): Validate NOT VALID constraints

groyoh:
## Context

Eleven constraints were created as `NOT VALID` and never validated: ten foreign keys (on `wallet_transactions`, `wallets`, `subscriptions`, `payments`, `recurring_transaction_rules`, `adjusted_fees` and `invoice_subscriptions`) and the `payments_customer_id_null` check constraint.

The `invoice_subscriptions` foreign key on `regenerated_invoice_id` (`fk_rails_f435d13904`) was supposed to be validated by migration `20250716143358`, but that migration used the bare `validate_foreign_key :invoice_subscriptions, :invoices` form. The table has two foreign keys to `invoices`, so the already-valid `invoice_id` one was matched and the `regenerated_invoice_id` one stayed invalid.

## Description

Thirteen migrations were added, one operation each:

- Ten validate the foreign keys. The `invoice_subscriptions` call targets the constraint by column (`column: :regenerated_invoice_id`) to avoid the ambiguity described above.
- The last three validate `payments_customer_id_null`, convert `payments.customer_id` to a `NOT NULL` column, and drop the now redundant check constraint, following the same pattern as `20251229153734` for `charges.code`.

`db/structure.sql` and the `Payment` schema annotation were regenerated accordingly.

Fixes ING-427.


app/models/payment.rb
```diff
@@ -148,7 +148,7 @@ def payment_request_succeeded
 #  status                       :string           not null
 #  created_at                   :datetime         not null
 #  updated_at                   :datetime         not null
-#  customer_id                  :uuid
+#  customer_id                  :uuid             not null
```
vincent-pochet, line 151: The relation is marked as `optional: true`, I guess it should be updated as well

vincent-pochet (approved):
Looks good.

Only question: what risk can it brings to self-hosted instances if for some reason some constraint contain invalid data?

app/models/payment.rb
groyoh: You are right.

groyoh:
> Looks good.
> 
> Only question: what risk can it brings to self-hosted instances if for some reason some constraint contain invalid data?

Well they won't be able to migrate/upgrade. Nothing critical but it means that probably need to run a previous migration which was not run.

aquinofb (approved):
LGTM!

annvelents (approved):

lovrocolic (approved):

mariohd (approved):

brunomiguelpinto (approved):

app/models/payment.rb
groyoh: Fixed in cde88eb25: the association is no longer optional, so a missing customer surfaces as a presence validation error instead of a database error.

PR #5901, opened 2026-07-09
fix(subscriptions): Skip pay-in-advance event for terminated subscription

D1353L:
## Context

Pay-in-advance fee creation runs asynchronously from a job. Between enqueue and execution a subscription can be terminated, so by the time `Fees::CreatePayInAdvanceService` runs, the event no longer maps to an active subscription. The service assumed a subscription was always present and raised when it was `nil`, leaving the job to fail and retry indefinitely.

## Description

Skip event processing if there is no related active subscription

### DIFF
--- a/app/services/fees/create_pay_in_advance_service.rb
+++ b/app/services/fees/create_pay_in_advance_service.rb
@@ -50,6 +50,8 @@ module Fees
     private
 
     def skip_missing_subscription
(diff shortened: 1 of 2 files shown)


annvelents (approved):

annvelents:
already approved, but might worth to add a comment that subscription is linked from the event, where we're selecting an active subscription for the timestamp of the event

PR #5902, opened 2026-07-09
feat(purchase_order_number): Subscriptions upgrade/downgrade

D1353L:
## Context
During the upgrade/downgrade, it should be possible to edit, remove or keep the purchase order number.

## Description

This PR adds the new `purchase_order_number` field in subscriptions upgrade/downgrade flow.

### DIFF
--- a/app/services/subscriptions/plan_downgrade_service.rb
+++ b/app/services/subscriptions/plan_downgrade_service.rb
@@ -42,6 +42,7 @@ module Subscriptions
           ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at,
           progressive_billing_disabled: params[:progressive_billing_disabled] || false,
           consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : current_subscription.consolidate_invoice,
+          purchase_order_number: params[:purchase_order_number],
           billing_entity_id: current_subscription.billing_entity_id
         )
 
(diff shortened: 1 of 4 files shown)


app/services/subscriptions/plan_downgrade_service.rb
```diff
@@ -42,6 +42,7 @@ def call
           ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at,
           progressive_billing_disabled: params[:progressive_billing_disabled] || false,
           consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : current_subscription.consolidate_invoice,
+          purchase_order_number: params[:purchase_order_number],
```
rsempe: don't we have to check if `params.key?(:purchase_order_number)` to keep the value if not present? 🤔

app/services/subscriptions/plan_upgrade_service.rb
```diff
@@ -62,6 +62,7 @@ def new_subscription_with_overrides
         billing_time: current_subscription.billing_time,
         ending_at: params.key?(:ending_at) ? params[:ending_at] : current_subscription.ending_at,
         consolidate_invoice: params.key?(:consolidate_invoice) ? params[:consolidate_invoice] : current_subscription.consolidate_invoice,
+        purchase_order_number: params[:purchase_order_number],
```
rsempe: same here

app/services/subscriptions/plan_downgrade_service.rb
D1353L: fair point, we need to inherit it if not present

rsempe (approved):

PR #5903, opened 2026-07-09
[STA-7] misc: Add TypedResult concern for legacy services

vincent-pochet:
## Context

As we are migrating services to the standardized single-call + typed Result pattern, a number of legacy services still expose several public methods (login, register_user, authorize_url, …) instead of a single call, and return the deprecated LegacyResult (OpenStruct).
Refactoring each of these into proper single-call services is a large and dangerous effort, so we need an intermediate step that lets use typed results (in order to remove OpenStructs) before propertly refactoring these services.

## Description

The PR introduces a temporary `TypedResults` concern that gives multi-method services a class-level call entry point returning a real typed Result.

Targeted services will include the concern and declares a `RESULTS` constant mapping each public method to the `Result` it returns.
Calling `Service.call(:method_name, *args)` builds the service, initializes a typed Result for that method, and routes the call to the existing instance method, leaving the method bodies untouched. 

This new pattern is applied to `Auth::GoogleService` to validate the behavior. Other remaining services will follow

### DIFF
--- a/app/graphql/mutations/auth/google/accept_invite.rb
+++ b/app/graphql/mutations/auth/google/accept_invite.rb
@@ -13,7 +13,7 @@ module Mutations
         type Types::Payloads::RegisterUserType
 
         def resolve(code:, invite_token:)
-          result = ::Auth::GoogleService.new.accept_invite(code, invite_token)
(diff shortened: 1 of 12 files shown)


toommz (approved):
Nice workaround 👍

PR #5904, opened 2026-07-09
[STA-7] misc: Result for all remaining Auth services

vincent-pochet:
## Context
This PR follows https://github.com/getlago/lago-api/pull/5903/changes

## Description

It plugs `TypedResults` into the `UsersService` and rename `Auth::TokenService` into `Utils::AuthToken`

### DIFF
--- a/app/services/users_service.rb
+++ b/app/services/users_service.rb
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 
 class UsersService < BaseService
+  include TypedResults
+
+  RESULTS = {
+    login: BaseResult[:user, :token],
+    register: BaseResult[:user, :organization, :membership, :token],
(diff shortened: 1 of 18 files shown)


toommz (approved):

app/services/users_service.rb
```diff
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 class UsersService < BaseService
+  include TypedResults
```
groyoh, line 4: I would suggest updating `TypedResults` to use `send` instead of `public_send` and move the `logic`, `register`, etc. to be private method. This will prevent any accidental usage of those public methods and only rely on `.call`.

groyoh (approved):

app/services/users_service.rb
vincent-pochet: Good idea, I will update it!

PR #5905, opened 2026-07-09
feat(invoices): add deleted status to Invoice model

endenis:
## Context

_First step of the "Delete a draft invoice" feature._

Right now operators can't remove a draft invoice created by mistake (duplicate subscription, test data, config error).

We want to add a delete action for draft invoices. The invoice disappears should disappear as if it never existed: no AR entry, no email, no external sync. The subscription should be untouched and should keep billing from the next period.

## Description

This PR introduces the `deleted` status and makes sure it stays invisible everywhere.
It doesn't expose any way to delete an invoice yet.

## Implementation details

- **New `deleted` status** in the Invoicer model (`8`)
  - it is added to `INVISIBLE_STATUS`, not `GENERATED_INVOICE_STATUSES`),
  - it is used in the `draft => deleted` transition.
- **Scope fixes** so a deleted invoice stay hidden:
  - `Credit.active`,
  - `overdue_balance`,
  - `payment_requests`, 
  - `fixed_charge` and `subscription` fee count guards.
- **Data-export views** rewritten to exclude deleted invoices:
  - `exports_invoice_subscriptions` (v2),
  - `exports_fees` (v3),
  - `exports_invoices_taxes` (v3).
- `db/structure.sql` includes the invoices table moving in the file. pg_dump reorders it because the export views now depend on it.
## Tests

Model coverage for the status/scope/transition, the scope fixes (overdue balance, payment requests), and the three export views.


db/views/exports_fees_v03.sql
```diff
@@ -0,0 +1,122 @@
...
+LEFT JOIN subscriptions AS s ON f.subscription_id = s.id
+LEFT JOIN customers AS c ON s.customer_id = c.id
+LEFT JOIN charges AS ch ON f.charge_id = ch.id
+LEFT JOIN billable_metrics AS bm ON ch.billable_metric_id = bm.id
+LEFT JOIN add_ons AS ao ON f.add_on_id = ao.id
+LEFT JOIN plans AS p ON s.plan_id = p.id
+-- LEFT JOIN so fees with no invoice (e.g. recurring non-invoiceable fees) are kept.
+-- IS DISTINCT FROM keeps those NULL-invoice rows while excluding deleted invoices.
+LEFT JOIN invoices AS i ON i.id = f.invoice_id
+WHERE i.status IS DISTINCT FROM 8;
```
vincent-pochet:
```suggestion
WHERE i.status IS DISTINCT FROM 8; -- exclude deleted invoices
```

vincent-pochet (approved):
Good PR! 🚀 

Just one comment regarding the description: Could you remove the link to the dive-in?
Notion is an internal tool and is not publicly exposed. Maybe instead you could just add one or two lines exposing the end-goal of the feature

app/models/analytics/overdue_balance.rb
```diff
@@ -78,6 +78,7 @@ def query(organization_id, **args)
             WHERE i.organization_id = :organization_id
             AND i.self_billed IS FALSE
             AND i.payment_overdue IS TRUE
+            AND i.status != #{Invoice.statuses[:deleted]}
```
annvelents, line 81: ha!
Does it mean we include voided invoices in overdue? 😮 
that's actually an interesting point about overdue in general...
1) draft invoices will never be overdue, so when closing it should not be overdue either
2) I checked voiding invoices service, and there if invoice was overdue we do not set it back to not_overdue... which might result in them being included in the analytics...
Can you please check the voided invoices scenario?

app/models/invoice.rb
```diff
@@ -148,11 +148,16 @@ class Invoice < ApplicationRecord
     state :failed
     state :closed
     state :pending
+    state :deleted
 
     event :finalize do
       transitions from: :draft, to: :finalized
     end
 
+    event :mark_as_deleted do
```
annvelents, line 157: ahah
such a pity AR already has `delete` method 😅

app/models/analytics/overdue_balance.rb
endenis: @annvelents When invoices are marked as voided they are updated with `payment_overdue: false`:

https://github.com/getlago/lago-api/blob/1e76af5414499d9142214bb4dc8a5288b9edd4e8/app/models/invoice.rb#L607

Do you think it's better to do the same for deleted draft invoices?

annvelents: ah! nice! 😮‍💨 
no, no need to, because draft invoice will never be `payment_overdue`, but in case anything will ever happen to this status and it will be extended, it's better to keep the check in the service

annvelents (approved):

PR #5909, opened 2026-07-10
refactor(analytics): Resolve billing entity code in resolvers

groyoh:
## Context

Two analytics GraphQL requests that differ only by `billingEntityCode` could return each other's figures. `Analytics::OverdueBalance` and `Analytics::InvoiceCollection` filtered their SQL on that argument through a dedicated `billing_entities` join, but their cache keys only included `billing_entity_id`, so both requests shared one cache entry for up to 4 hours and the second request got the first one's numbers.

The analytics resolvers were also inconsistent: only `overdueBalances` and `invoiceCollections` accepted `billingEntityCode`, while `grossRevenues`, `mrrs` and `invoicedUsages` only accepted `billingEntityId`. The REST analytics controllers already resolve the code to an id before calling the models, so the GraphQL layer was the only caller pushing codes into the SQL.

## Description

Instead of patching the cache keys with the missing code argument, the GraphQL layer now resolves `billingEntityCode` to a `billing_entity_id` before calling the models, the same way the REST controllers do. A shared `BillingEntityArgsResolvable` concern is included in all five analytics resolvers. The lookup is scoped to the current organization's billing entities, returns a not-found error when the code matches no entity (previously an unknown code silently returned empty results), and rejects requests carrying both `billingEntityCode` and `billingEntityId` with a validation error. `grossRevenues`, `mrrs` and `invoicedUsages` gain the `billingEntityCode` argument that only `overdueBalances` and `invoiceCollections` had before.

The models are id-only again: the `billing_entity_code` SQL branch and its `billing_entities` join were removed from `Analytics::OverdueBalance` and `Analytics::InvoiceCollection`. Since `billing_entity_id` was already part of every cache key, the collision is gone by construction. Resolver specs now assert the entity scoping for all five queries.

This is the first PR of a stack split out of #5828.

Fixes ING-368

### DIFF
--- a/app/graphql/concerns/billing_entity_args_resolvable.rb
+++ b/app/graphql/concerns/billing_entity_args_resolvable.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module BillingEntityArgsResolvable
+  extend ActiveSupport::Concern
+
+  private
+
+  # Replaces the `billing_entity_code` argument with the matching `billing_entity_id`
+  # so analytics models only deal with ids.
+  # Mutates `args` and returns an execution error when the arguments are invalid, nil otherwise.
+  def resolve_billing_entity!(args)
+    if args[:billing_entity_code].present? && args[:billing_entity_id].present?
+      return validation_error(messages: {billing_entity_id: ["can't be present when billing_entity_code is provided"]})
+    end
+
+    code = args.delete(:billing_entity_code)
+    return nil if code.blank?
+
+    billing_entity = current_organization.billing_entities.find_by(code:)
+    if billing_entity.nil?
+      return not_found_error(resource: "billing_entity")
+    end
+
+    args[:billing_entity_id] = billing_entity.id
+    nil
+  end
+end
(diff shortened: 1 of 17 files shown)


app/graphql/concerns/billing_entity_args_resolvable.rb
```diff
@@ -0,0 +1,27 @@
...
+    code = args.delete(:billing_entity_code)
+    return nil if code.blank?
+
+    billing_entity = current_organization.billing_entities.find_by(code:)
+    if billing_entity.nil?
+      return not_found_error(resource: "billing_entity")
+    end
+
+    args[:billing_entity_id] = billing_entity.id
+    nil
```
toommz, line 25: Made me check if Ruby doesn't have a hash assignment method returning nil, but seems not 😅

toommz (approved):
LGTM

vincent-pochet (approved):

groyoh:
Thanks @toommz @vincent-pochet !

PR #5913, opened 2026-07-10
fix: refactor query for removing billable metric events

lovrocolic:
## Context

`BillableMetrics::DeleteEventsJob` keeps timing out on EU production with `ActiveRecord::QueryCanceled`, even after the `index_charges_on_billable_metric_id_all` index. With retry: 0, every failure lands in the dead set and the deleted metric's events are never cleaned up.

The cursor query in `Events::DeleteForMetricService` joined subscriptions → plans → charges with a `DISTINCT ... ORDER BY subscriptions.id LIMIT n` shape, letting the planner walk subscriptions_pkey across all 19.4M rows on an unreliable cross-join estimate. The join only existed to bypass Charge's kept default scope.

## Description

Removed the join and split the query to rely on single-column stats and existing indexes:

- Walk billable_metric.charges.with_discarded in batches — discarded charges are found via `index_charges_on_billable_metric_id_all`, no join needed.
- Query Subscription.where(plan_id: ...) in batches — cursor is now `WHERE plan_id IN (...) AND id > ? ORDER BY id LIMIT n`, served by `index_subscriptions_on_plan_id`.

Plans spanning two charge batches get processed twice; this is idempotent.

### DIFF
--- a/app/services/events/delete_for_metric_service.rb
+++ b/app/services/events/delete_for_metric_service.rb
@@ -15,18 +15,22 @@ module Events
 
       clickhouse_enabled = ENV["LAGO_CLICKHOUSE_ENABLED"].present?
 
-      # The subscription is queried in batches to avoid memory pressure
-      # and cross-database subqueries: events is built to live on a dedicated
-      # database (so is clickhouse) while subscriptions/charges live on the primary database.
-      subscriptions_relation.in_batches(of: BATCH_SIZE) do |batch|
-        rows = batch.pluck(:id, :external_id)
-        next if rows.empty?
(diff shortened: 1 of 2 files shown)


groyoh (approved):
Could you add a test case for:

> Plans spanning two charge batches get processed twice; this is idempotent.

groyoh:
Maybe, if not already the case, let's improve the test to ensure other unrelated events won't get deleted (other charges on the same subs/plans, events in other orgs, etc. Since we're talking about hard deletion on the Clickhouse side, we should be thorough with testing.

jdenquin (approved):

mariohd (approved):

PR #5915, opened 2026-07-10
[STA-6] misc(payment_provider_customer): Migrate to TypedResult

vincent-pochet:
## Context

The `TypedResults` concern was introduced to let legacy multi-method services return typed Results instead of the deprecated LegacyResult (OpenStruct), as an intermediate step before a full single-call refactor.

## Description

This PR applies the pattern to the `PaymentProviderCustomers` sub-services (Stripe, Adyen, Cashfree, Flutterwave, Gocardless, Moneyhash).

Each service now includes the concern, declares a `RESULTS` constant mapping every method to its typed Result, and exposes those methods as private — reached through `Service.call(:method_name, *args)`. The former per-instance constructor is dropped in favor of arguments passed to each method.

Call sites are updated accordingly: jobs use `call!`, `Factory#new_instance` is removed in favor of `Factory.for(provider_customer)`, and specs exercise the services through the `.call(:method)` entry point.

### DIFF
--- a/app/services/payment_provider_customers/factory.rb
+++ b/app/services/payment_provider_customers/factory.rb
@@ -2,11 +2,7 @@
 
 module PaymentProviderCustomers
   class Factory
-    def self.new_instance(provider_customer:)
-      service_class(provider_customer).new(provider_customer)
-    end
-
-    def self.service_class(provider_customer)
+    def self.for(provider_customer)
       case provider_customer&.class.to_s
       when "PaymentProviderCustomers::StripeCustomer"
         PaymentProviderCustomers::StripeService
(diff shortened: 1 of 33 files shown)


app/services/payment_provider_customers/factory.rb
```diff
@@ -2,11 +2,7 @@
 
 module PaymentProviderCustomers
   class Factory
-    def self.new_instance(provider_customer:)
-      service_class(provider_customer).new(provider_customer)
-    end
-
-    def self.service_class(provider_customer)
+    def self.for(provider_customer)
```
toommz, line 5: 😎

toommz (approved):

PR #5917, opened 2026-07-10
perf(activity-logs): add external id bloom filter indexes

groyoh:
## Context

Customer- and subscription-scoped activity log queries (GraphQL `CustomerActivityLogs`, REST `/api/v1/activity_logs`) filter the ClickHouse `activity_logs` table by `external_customer_id` / `external_subscription_id` and order by `logged_at DESC`. Neither column is part of the table's sort key, so ClickHouse has to read the whole organization partition to return a handful of rows. For a large organization (~100M rows) this takes a few seconds when queried directly and tens of seconds under load.

## Description

This adds `bloom_filter(0.001)` skip indexes with `GRANULARITY 1` on `external_customer_id` and `external_subscription_id`:

- two self-hosted migrations, one per index, each adding and materializing its index. ClickHouse DDL is not transactional, so keeping one index per migration means a failure leaves a clean state instead of a half-applied migration. Both use `IF NOT EXISTS` / `IF EXISTS` guards and a `down` that drops the index, so they are idempotent and reversible.
- the same `INDEX` declarations in the cloud DDL file, so new clusters get them at table creation.

A benchmark on a clone of the affected table showed rows read dropping from 100M+ to 300k and latency from 2.1s to 0.39s for the customer-scoped query, with identical results returned. No query or resolver change is needed; the indexes are transparent.

Note that `MATERIALIZE INDEX` runs as a background mutation: the migration returns immediately and the speedup on existing data applies once the mutation completes (visible in `system.mutations`).

The ClickHouse migration conventions used here (self-hosted vs cloud DDL, one concern per migration, explicit `up`/`down` with guards) were also documented in `AGENTS.md`.

Fixes ING-367

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -171,6 +171,12 @@ To create a webhook:
 - When adding a constraint with `validate: false` (`NOT VALID`), add a second migration right after it that validates the constraint (`validate_foreign_key` or `validate_check_constraint`), within the same PR. If validation is not possible yet (e.g. existing rows must be backfilled or fixed first), register the constraint in `db/not_valid_constraints.yml` with a `validate_by` deadline, and remove the entry once the constraint is validated. `spec/db/not_valid_constraints_spec.rb` enforces this.
 - When validating a foreign key on a table that has several foreign keys to the same target table, pass the `column:` option to `validate_foreign_key`, otherwise the wrong constraint may be validated silently.
 
+## Clickhouse migrations
+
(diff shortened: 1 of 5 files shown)


vincent-pochet (approved):

rsempe (approved):
`db/clickhouse_structure.sql` does not seem updated here

groyoh:
> db/clickhouse_structure.sql does not seem updated here

Nice catch!

groyoh:
Actually @rsempe I believe we don't have this file anymore 🤔

PR #5919, opened 2026-07-10
fix(fc-units-override): Add validation to prevent fixed charges updates for incomplete subscriptions

osmarluz:
## Context

A subscription gated on a payment activation rule (`incomplete` status) must not be editable until the gating invoice is paid. `Subscriptions::UpdateService` already rejects edits on incomplete subscriptions, but the dedicated fixed-charge edition endpoint (`updateSubscriptionFixedCharge` mutation and `PUT /api/v1/subscriptions/:external_id/fixed_charges/:code`) did not: the edition was applied and the emitted fixed-charge event was billed by `BillFixedChargesDeltaJob` right after activation, bypassing the payment gate.

## Description

Reject any fixed-charge edition on an incomplete subscription with a `subscription_incomplete` validation failure at the top of `Subscriptions::UpdateOrOverrideFixedChargeService#call`, covering both the GraphQL and REST entry points. Pending and active subscriptions (including formerly gated ones) remain fully editable.


ancorcruz (approved):
looks great.

PR #5925, opened 2026-07-10
fix(payment-gated-subs): Defer plan-level fixed charges updates to next billing cycle

osmarluz:
## Context

When a plan's fixed charge is edited with `apply_units_immediately` (update, create, or cascade) while a subscription on that plan is gated on payment (`incomplete`), the emitted fixed-charge event lands inside the gating window and `ActivationRules::BillFixedChargesDeltaJob` bills it right after the customer pays the gating invoice — the customer approves a payment for the original units and immediately owes a delta invoice for a change made while gated.

## Description

In `FixedCharges::EmitEventsService`, events for incomplete subscriptions are now always stamped at the next billing period start, even when `apply_units_immediately` is true. The change is deferred rather than dropped — periodic billing consumes these events, so skipping emission would freeze the old units forever. The gated customer keeps what they paid for through the first period, no delta invoice is issued at activation, and the new units take effect at renewal through the regular billing run. This covers all plan-level paths (fixed-charge update, create, plan cascade) as well as per-subscription overrides.


spec/scenarios/subscriptions/payment_gated_activation_spec.rb
```diff
@@ -618,7 +635,7 @@ def create_fixed_charge(plan, units, timestamp:, apply_units_immediately: true)
         end
     end
 
-    it "bills an override unit increase applied immediately as a delta invoice on activation" do
+    it "defers an override unit increase applied immediately to the next billing period" do
```
osmarluz: This behavior will be superseded by PR #5919

ancorcruz (approved):
looks great. 🚢

PR #5935, opened 2026-07-14
fix(payment-gated-subs): Skip billing of missed periods after activation of incomplete subscriptions

osmarluz:
## Context

When a payment-gated subscription stays incomplete past one or more billing boundaries, activation replayed every boundary crossed during the gating window (`BillMissedPeriodsService`): the customer who just approved the gating payment immediately received one full periodic invoice per elapsed period, paying for time the subscription was never active.

## Description

`BillMissedPeriodsService` is replaced by `BillCurrentPeriodService`: on activation, only the period in progress is billed, with a single periodic invoice backdated to the period's boundary so the subscription seamlessly rejoins the regular billing clock. The service computes that boundary directly from the current time (using the monthly split grid for yearly and semiannual plans with monthly-billed charges or fixed charges) instead of walking every boundary since `started_at`. If activation happens within the first period, nothing is billed since the gating invoice already covered it.

Periods fully elapsed while gated are intentionally forgiven and never billed. A fixed-charge change deferred into a skipped period (#5925) still takes effect on the first billed invoice, because fixed-charge aggregation carries the last event preceding the billed window.


app/jobs/subscriptions/activation_rules/bill_current_period_job.rb
```diff
@@ -2,7 +2,7 @@
 
 module Subscriptions
   module ActivationRules
-    class BillMissedPeriodsJob < ApplicationJob
```
tiagolupepic, line 5: There is a small issue renaming jobs. Imagine a scenario, we have the service running and publishing jobs using the name `BillMissedPeriodsJob` when the deployment happens, we might have the class name stored in Redis and when the new version of sidekiq pick the job, the class will not exist anymore and an error will be raised. 

The ideal scenario is to add a no op class `BillMissedPeriodsJob` just to avoid errors and then in another PR you remove the no op class

osmarluz: @tiagolupepic makes sense, but we don't have any incomplete subscriptions at the moment and this is an edge case of the activation process, so I renamed it here directly as there's a very low risk we'll be seeing this soon

ancorcruz (approved):
LGTM

PR #5937, opened 2026-07-14
fix(payments): scope checkout auto-payment delay to the org

brunomiguelpinto:
## Context

PR #5931 delays the first automatic payment for hosted-checkout customers so their manual checkout payment settles first, avoiding a double charge. It identified a "checkout customer" by whether that **customer** had ever generated a checkout URL (`PaymentIntent` on one of their invoices).

That misses a customer's **first** checkout invoice: at invoice-creation time the customer has no prior `PaymentIntent`, so the auto-payment is not delayed, races their manual checkout, and double-charges them.

Production confirmed this after the deploy — the remaining duplicates all had `prior_payment_intents = 0` (customers using checkout for the first time), with the two provider payments ~1–2 min apart (i.e. the delay never applied).

## Change

Decide the delay at the **organization** level instead of the customer level: if the organization has ever generated a checkout URL, delay the first automatic payment for **all** of its customers. This covers a customer's very first
invoice, which has no checkout signal of its own.

- Uses `payment_intents.organization_id` directly (indexed) — no join.
- Only the **first** attempt is delayed; retries no delay

### DIFF
--- a/app/services/invoices/payments/create_service.rb
+++ b/app/services/invoices/payments/create_service.rb
@@ -96,7 +96,7 @@ module Invoices
         return result unless provider
 
         forced_delay = nil
-        forced_delay = CHECKOUT_AUTO_PAYMENT_DELAY if defer_for_checkout_customer?
+        forced_delay = CHECKOUT_AUTO_PAYMENT_DELAY if defer_for_checkout_organization?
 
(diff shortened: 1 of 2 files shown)


tiagolupepic (approved):
:shipit:

annvelents (approved):

PR #5940, opened 2026-07-14
[ING-476] fix(lifetime_usage): retry wallet lock inline

mariohd:
## Context

When a subscription activity is processed, `UsageMonitoring::ProcessSubscriptionActivityService` runs the lifetime-usage recalculation inline via `RecalculateAndCheckJob.perform_now(lifetime_usage, current_usage:)`. `current_usage` is a `SubscriptionUsage` struct, which ActiveJob cannot serialize (it is only safe under `perform_now`, which does not serialize arguments).

The job retries wallet lock conflicts (`ActiveRecord::StaleObjectError`, `BaseLockService::FailedToAcquireLock`) through ActiveJob's `retry_on`. When a wallet optimistic-lock conflict happens during the inline run, `retry_on` tries to re-enqueue the job for a retry, which serializes the arguments and fails on the `SubscriptionUsage` struct. The result is a misleading `ActiveJob::SerializationError: Unsupported argument type: SubscriptionUsage` that masks the real underlying `StaleObjectError` on `Wallet`.

## Solution

Retry the lock conflict in-process on the inline path so it never reaches ActiveJob's `retry_on`. The async `perform_later` path (no `current_usage`) keeps relying on `retry_on`, so it does not hold a worker while waiting between attempts.

- Split `perform` into two paths: with `current_usage` present, wrap the work in an in-process retry (`MAX_LOCK_RETRY_ATTEMPTS`, `rand(0...MAX_LOCK_RETRY_DELAY)` backoff); otherwise call the services directly and let `retry_on` handle retries.
- On exhaustion, raise a dedicated `InlineLockRetryExhausted` error that is intentionally not listed in `retry_on`. It is raised with the cause chain broken (`cause: nil`) because `retry_on` also matches an exception's `cause` — otherwise the wrapped lock error would still trigger a retry and the same serialization failure. The original error is preserved in the message for observability.

## Testing

- New specs cover the inline `current_usage` path for both lock error classes: retries then resolves without raising `ActiveJob::SerializationError`, and exhausts to `InlineLockRetryExhausted` (never `SerializationError`).
- Existing `retry_on` (async) and `discard_on` behavior is unchanged and still covered.
- `RecalculateAndCheckJob` spec: 13 examples, 0 failures. `ProcessSubscriptionActivityService` spec: 12 examples, 0 failures. RuboCop clean on the changed files.

### DIFF
--- a/app/jobs/lifetime_usages/recalculate_and_check_job.rb
+++ b/app/jobs/lifetime_usages/recalculate_and_check_job.rb
@@ -2,6 +2,12 @@
 
 module LifetimeUsages
   class RecalculateAndCheckJob < ApplicationJob
+    # Raised when the inline (perform_now) invocation cannot resolve a lock conflict within
+    # MAX_LOCK_RETRY_ATTEMPTS. It intentionally is NOT a class listed in `retry_on`, so the
+    # lock conflict never triggers an ActiveJob retry (which would fail to serialize the
+    # non-serializable current_usage). The original lock error is preserved as `cause`.
+    class InlineLockRetryExhausted < StandardError; end
+
     queue_as do
       if ActiveModel::Type::Boolean.new.cast(ENV["SIDEKIQ_BILLING"])
         :billing_low_priority
@@ -25,6 +31,25 @@ module LifetimeUsages
 
     # NOTE: do not pass current usage with perform_later as it will be a huge JSON
     def perform(lifetime_usage, current_usage: nil)
+      # When invoked inline via perform_now, current_usage is a non-serializable
+      # SubscriptionUsage struct. A lock conflict must be retried in-process here:
+      # reaching ActiveJob's retry_on would try to re-enqueue the job and fail to
+      # serialize current_usage. The async perform_later path (current_usage nil) keeps
+      # relying on retry_on so it does not hold a worker while waiting between attempts.
+      if current_usage
+        perform_with_inline_lock_retry(lifetime_usage, current_usage)
+      else
+        process(lifetime_usage, current_usage:)
+      end
+    end
+
+    def lock_key_arguments
+      [arguments.first]
+    end
+
+    private
+
+    def process(lifetime_usage, current_usage:)
       LifetimeUsages::CalculateService.call!(lifetime_usage:, current_usage:)
 
       if lifetime_usage.organization.progressive_billing_enabled?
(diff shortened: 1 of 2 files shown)


toommz (commented):
Why not just call `.to_h` before enqueuing, and reconstruct the struct in the service when it is a `Hash`? It would save us from this inline retry logic 🤔 

I really think the real problem here is passing a non-serializable object to our job layer.

mariohd:
@toommz 
https://github.com/getlago/lago-api/pull/5940/changes#diff-70f317eb4d1bc8c374d8ee7af381d011551a621fc57cf8221bfd385211871576R32

``` # NOTE: do not pass current usage with perform_later as it will be a huge JSON```

The `current_usage` tends to be a super huge hash and if we serialize it, it may break the job schedule as well, so it has to be executed inline. 
Here it mentions about  `perform_later`, a `retry` is a form of `perform_later`, so we shouldn't serialize the params. 
In order to have the retries for inline execution, we need to catch the exception and run inline again; checking if the max attempts wasn't reached.

File: app/jobs/lifetime_usages/recalculate_and_check_job.rb [hunk 1]
```diff
@@ -25,15 +31,52 @@ class RecalculateAndCheckJob < ApplicationJob
...
+
+      begin
+        process(lifetime_usage, current_usage:)
+      rescue BaseLockService::FailedToAcquireLock, ActiveRecord::StaleObjectError => e
+        attempts += 1
+
+        if attempts < MAX_LOCK_RETRY_ATTEMPTS
+          sleep rand(0...MAX_LOCK_RETRY_DELAY)
+          retry
+        end
```
aquinofb, line 71: This could lock it for a long time (up to ~5 min), no? Could we give the inline path its own "retry budget" with fewer retry attempts?

mariohd: Actually, no. This could potentially let the time take around 6 minutes in the worst case scenario, 25 attempts with 16 seconds on each. 

This retry inline will be used only when the current usage is passed. But this part only gets triggered if the lock is not available. Since we can't re-schedule the job with the huge hash, we have to wait with the params in memory to retry. 

About reducing the attempts size, I'm ok with that, just used the existing constants for these procedure.

vincent-pochet: @mariohd Should we maybe reduce the number of attempt in that case?

An other approach could be to just "raise" and let the `UsageMonitoring::ProcessSubscriptionActivityService` (the caller with a `current_usage` argument)deal with the issue. What do you think?

mariohd: That's also a good approach! I will change to have this

PR #5944, opened 2026-07-15
feat(purchase_order_number): Add the new field to regenerate-from-voided invoice

D1353L:
## Context
When regenerating voided invoice, it should be possible to edit, remove or keep the purchase order number.

## Description
Adds an optional `purchaseOrderNumber` argument to the `regenerateFromVoided` mutation.
- Omitted → inherit the voided invoice's PO number.
- Provided (value or nil) → stamp that value on the regenerated invoice.


toommz (approved):

tiagolupepic (approved):
:shipit:

PR #5882, opened 2026-07-07
imp(agg): Bypass event agg when no filters are applied

jdenquin:
## Context

  While profiling customer usage computation (`Customers::RefreshWalletsService` → `Invoices::CustomerUsageService`), traces showed `ChargeFilters::MatchingAndIgnoredService` being called once per charge filter — over 1,500 times (~2.5s) per usage refresh on plans with many filters.

 Since #5846, aggregations for filters without usage in the period are bypassed (`bypass_aggregation`). The matching and ignored filters are only consumed by the event stores when building an events query, so computing them for bypassed aggregations is wasted work.

  ## Description

  Skip the computation of matching and ignored filters when the aggregation is bypassed:

  - `Fees::ChargeService#aggregation_filters` no longer calls `ChargeFilters::MatchingAndIgnoredService` when the aggregation is bypassed. Recurring metrics are excluded from the
  skip, as they always aggregate (matching `should_bypass_aggregation?` semantics).

  To make this safe, `per_event_aggregation` now returns an empty result when the aggregation is bypassed, instead of querying the event store:

  - `BillableMetrics::Aggregations::BaseService`
  - `BillableMetrics::ProratedAggregations::SumService`
  - `BillableMetrics::ProratedAggregations::UniqueCountService`

  Without this guard, some charge models (prorated graduated, percentage with per-transaction min/max) could still query the event store on a bypassed aggregator and would then miss the matching filters. This is also a small perf win on its own: those queries were guaranteed to return no events, since the billing period pre-filtering already established the filter had no usage.

PR #5945, opened 2026-07-15
chore(perf): Dedupe charge filter preload

jdenquin:
## Context
  
  When computing customer usage, the plan's charges and their filter tree (`charge_filters` + `charge_filter_values` with their large `properties` jsonb) are
  loaded from PostgreSQL twice per computation:

  1. `Invoices::CustomerUsageService#charges` preloads them for the fee loop
  2. `Events::BillingPeriodFilterService#non_recurring_charges_with_events` re-loads the exact same associations for the same plan

  On large plans (~1,600 charge filters) each load costs seconds, and `RefreshWalletsService` pays this per subscription.

  ## Description

  Load the charge/filter set once and share it:

  - `Events::BillingPeriodFilterService` accepts an optional `charges:` argument (pre-loaded charges with `billable_metric` and `filters: {values:
  :billable_metric_filter}`). When provided, `non_recurring_charges_with_events`, `recurring_event_charges_and_filters` and `current_recurring_charges` are
  resolved in memory instead of re-querying. Without it, behavior is unchanged, so `CalculateFeesService` and `ProgressiveBillingService` are unaffected.
  - `Invoices::CustomerUsageService` passes its memoized charges relation to the service; since it is the same object, the fee loop then reuses the loaded
  records.
  - `find_each` is replaced by `each` in the fee loop: `find_each` always re-queries in batches, which would have triggered the preload a second time.
  - The usage preload now includes `billable_metric: :filters`, required by `ChargeFilters::EventMatchingService` to stay query-free on pre-loaded charges.

### DIFF
--- a/app/services/events/billing_period_filter_service.rb
+++ b/app/services/events/billing_period_filter_service.rb
@@ -4,9 +4,13 @@ module Events
   class BillingPeriodFilterService < BaseService
     Result = BaseResult[:charges]
 
-    def initialize(subscription:, boundaries:)
+    # charges: optional pre-loaded collection of the plan charges, with :billable_metric
+    #          and filters: {values: :billable_metric_filter} associations loaded.
+    #          When provided, it is used instead of re-loading the same records from the database.
+    def initialize(subscription:, boundaries:, charges: nil)
       @subscription = subscription
       @boundaries = boundaries
+      @preloaded_charges = charges
       super
     end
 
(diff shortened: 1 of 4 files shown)


jdenquin:
with https://github.com/getlago/lago-api/pull/5882, we can again win around 50ms per subscription during a customer usage computation.
QA confirm it.

vincent-pochet (approved):

PR #5950, opened 2026-07-15
feat(purchase_order_number): Add the new field to invoices and credit notes filtering

D1353L:
## Context

It should be possible to filter invoices and credit notes by purchase order number. Filtering should be an exact match and case-insensitive.

## Description
- Added a btree index to support fast case-insensitive filtering.
- Added the new field to Meilisearch attributes.


app/models/invoice.rb
```diff
@@ -12,7 +12,7 @@ class Invoice < ApplicationRecord
   meilisearch(index_uid: "invoices", auto_index: false, auto_remove: true) do
     attribute :number, :organization_id, :billing_entity_id, :currency, :customer_id,
       :invoice_type, :status, :payment_status, :payment_overdue, :self_billed,
-      :total_amount_cents, :total_paid_amount_cents
+      :total_amount_cents, :total_paid_amount_cents, :purchase_order_number
```
jdenquin, line 16: this will need a complete invoice reindex on both cluster!

D1353L: as we currently have only 5 invoices per cluster with a PO number, do you think we can avoid a full reindex and backfill only those invoices?

`Invoice.where.not(purchase_order_number: nil).find_each { |i| Invoices::SearchIndexJob.perform_later(i.id) }`

vincent-pochet (approved):

app/models/invoice.rb
tiagolupepic: Tested `meilisearch` locally and we only need to reindex the affected invoices and we don't need to perform a full reindex.

app/models/invoice.rb
```diff
@@ -31,16 +32,16 @@ class Invoice < ApplicationRecord
     attribute(:metadata_keys) { metadata.map(&:key) }
 
     searchable_attributes %i[number customer_name customer_firstname customer_lastname
-      customer_legal_name customer_external_id customer_email]
+      customer_legal_name customer_external_id customer_email purchase_order_number]
```
tiagolupepic, line 35: @D1353L I'm adding the `purchase_order_number` here as well. Users can combine full text search (PO number) + filter (different values) OR use the filter (purchase_order_number)

PR #5953, opened 2026-07-16
feat(orders): execute one-off orders into billing

toommz:
## Why

Signed one-off orders need to be turned into invoices, on schedule, with a durable record of each execution attempt.

## What

- Add an `execution_record` JSONB column and a `failed` status to orders, plus an index for executable orders.
- Add `Orders::ExecuteService` / `Orders::OneOff::ExecuteService` to bill a one-off order's add-ons into a one-off invoice.
- Schedule due orders from the clock via `Clock::ExecuteScheduledOrdersJob` and `Orders::ExecuteOrderJob`.

## Stacked on

Based #5952 on `feat/invoices/one-off-discarded-add-ons` (review/merge that first).

PR #5954, opened 2026-07-16
feat(orders): expose and trigger order execution via GraphQL

toommz:
## Why

The dashboard needs to see an order's execution outcome and to trigger execution manually.

## What

- Expose the execution record on the order GraphQL type and the REST serializer.
- Add the `executeOrder` mutation, gated by a new permission.

## Stacked on

Based on #5953 `feat/orders/execution` (review/merge that first).

### DIFF
--- a/app/graphql/mutations/orders/execute.rb
+++ b/app/graphql/mutations/orders/execute.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module Mutations
+  module Orders
+    class Execute < BaseMutation
+      include AuthenticableApiUser
+      include RequiredOrganization
+
+      REQUIRED_PERMISSION = "orders:execute"
+
+      graphql_name "ExecuteOrder"
+      description "Execute an order"
+
+      input_object_class Types::Orders::ExecuteInput
+
+      type Types::Orders::Object
+
+      def resolve(**args)
+        order = current_organization.orders.find_by(id: args[:id])
+        result = ::Orders::ExecuteService.call(order:)
+
+        result.success? ? result.order : result_error(result)
+      end
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


app/graphql/types/orders/execution_record.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Types
+  module Orders
+    class ExecutionRecord < Types::BaseObject
+      graphql_name "OrderExecutionRecord"
+
+      field :errors, [String], null: false
+      field :executed_at, String, null: true
```
vincent-pochet: Should it be a `GraphQL::Types::ISO8601DateTime`?

app/graphql/mutations/orders/execute.rb
```diff
@@ -0,0 +1,26 @@
...
+      graphql_name "ExecuteOrder"
+      description "Execute an order"
+
+      input_object_class Types::Orders::ExecuteInput
+
+      type Types::Orders::Object
+
+      def resolve(**args)
+        order = current_organization.orders.find_by(id: args[:id])
+        result = ::Orders::ExecuteService.call(order:)
```
vincent-pochet, line 20: Unlike the job, this mutation does not handle:
- `BaseLockService::FailedToAcquireLock` -> not sure if it should/can be handled 🤔 
- Issues with tax providers when creating the invoice (in sync). -> The OneOff invoice creation already suffers from this issue so that might not be an issue here as well. I just wanted to point it out

vincent-pochet (approved):

app/graphql/types/orders/execution_record.rb
toommz: a6de7673df4266f20d8f058347bfdf74dcd0c643

app/graphql/mutations/orders/execute.rb
toommz: 0c503ec9bc21dbfa3e1cfed6a83cd8fea0e597b8

For the tax providers, it is done async right? Or did I miss something?

vincent-pochet: It is indeed. Sorry for the confusion

PR #5957, opened 2026-07-16
feat(invoices): add service for deleting draft invoices

endenis:
## Context

Part of the "Delete a draft invoice" feature. Builds on the `deleted` status added for [invoices](https://github.com/getlago/lago-api/pull/5905) and [credit notes](https://github.com/getlago/lago-api/pull/5946).

## Description

This PR adds the service layer that performs the deletion of draft invoices.
Exposing this service in API will be implemented in separate PRs.

## Details

- **`Invoices::DeleteService`** - performs the deletion inside a lock:
  - validates the invoice is `draft`
  - validates it hasn't been synced externally
  - soft-deletes any attached credit notes (rolls back if one can't be deleted)
  - transitions the invoice to `deleted`
  - writes an audit log and fires the `invoice.deleted` webhook
- **`CreditNotes::DeleteService`** - soft-deletes a credit note
  - only `draft` credit notes are deletable
- **`invoice.deleted` webhook** - a new event, mirroring `invoice.voided`

## Frontend

Requires a follow-up in the front (activity-log label for the new type): https://github.com/getlago/lago-front/pull/3929
This follow-up addresses the "Front compatibility" CI check that fails for this PR.


app/services/invoices/delete_service.rb
```diff
@@ -0,0 +1,58 @@
...
+        delete_result = CreditNotes::DeleteService.call(credit_note:)
+        next if delete_result.success?
+
+        result.not_allowed_failure!(code: "credit_note_not_deletable")
+        raise ActiveRecord::Rollback
+      end
+    end
+
+    def synced_externally?
+      invoice.integration_resources.exists?
```
tiagolupepic, line 54: out of curiosity, what is this integration_resources?

app/services/invoices/delete_service.rb
```diff
@@ -0,0 +1,58 @@
...
+        invoice.mark_as_deleted!
+      end
+
+      return result unless result.success?
+
+      result.invoice = invoice
+      SendWebhookJob.perform_later("invoice.deleted", result.invoice)
+
+      result
+    rescue AASM::InvalidTransition
```
tiagolupepic: is it necessary to rescue for invalid transitions?

tiagolupepic (approved):
looks good! just two comments to understand better the PR.

Be aware that you're updating the Graphql schema and you need to rebuild frontend after merge this PR.

:shipit:

app/services/invoices/delete_service.rb
annvelents: I'l answer here: if a customer has external integrations, and invoice should be synced to the external integration, we'll create an integration_resource for this sync. Usually it would have the external_integration_id, indicating invoice_id on the integration side

annvelents (approved):
LGTM!

spec/services/invoices/delete_service_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+        expect(result).not_to be_success
+        expect(result.error).to be_a(BaseService::MethodNotAllowedFailure)
+        expect(result.error.code).to eq("invoice_synced_to_external_system")
+        expect(invoice.reload).to be_draft
+      end
+    end
+
+    context "when the draft invoice has a credit note" do
+      let!(:credit_note) { create(:credit_note, :draft, invoice:, customer:) }
```
ancorcruz: Lets not introduce the usage of `let!` we prefer `let` + `before` block calling the variable to ensure creation.

spec/services/invoices/delete_service_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+        expect(result.error).to be_a(BaseService::MethodNotAllowedFailure)
+        expect(result.error.code).to eq("credit_note_not_deletable")
+        expect(invoice.reload).to be_draft
+        expect(credit_note.reload).to be_finalized
+      end
+
+      it "does not enqueue a webhook" do
+        expect { delete_service.call }.not_to have_enqueued_job(SendWebhookJob)
+      end
+    end
```
ancorcruz, line 106: Test coverage: the rollback test doesn't fully prove atomicity across sibling credit notes.

The `"when the draft invoice has a credit note that cannot be soft-deleted"` context uses a single finalized credit note, so it verifies the invoice stays draft but not that credit notes deleted earlier in the same cascade get unwound. Since `soft_delete_credit_notes!` iterates and deletes one at a time before hitting the failure, the all-or-nothing guarantee is exactly the part worth pinning down.

Consider adding a case with one draft + one finalized credit note and asserting the draft one is rolled back too:

```
context "when one of several credit notes cannot be soft-deleted" do
  let!(:deletable_credit_note) { create(:credit_note, :draft, invoice:, customer:) }
  let!(:blocking_credit_note) { create(:credit_note, status: :finalized, invoice:, customer:) }

  it "rolls back every credit note along with the invoice" do
    result = delete_service.call

    expect(result.error.code).to eq("credit_note_not_deletable")
    expect(invoice.reload).to be_draft
    expect(deletable_credit_note.reload).to be_draft
    expect(blocking_credit_note.reload).to be_finalized
  end
end
```

app/services/invoices/delete_service.rb
```diff
@@ -0,0 +1,58 @@
...
+    )
+
+    def call
+      return result.not_found_failure!(resource: "invoice") unless invoice
+
+      invoice.with_lock do
+        return result.not_allowed_failure!(code: "not_deletable") unless invoice.draft?
+        return result.not_allowed_failure!(code: "invoice_synced_to_external_system") if synced_externally?
+
+        soft_delete_credit_notes!
```
ancorcruz: Neither `Invoice` nor `CreditNote` includes `Discard::Model` — deletion is an enum value (uses enum magic method) to deleted, not a `deleted_at` discard. The code is internally consistent, so nothing to change but the method naming. Just flagging so `soft_delete_credit_notes!` isn't later mistaken for a discard/deleted_at call. Could we use a method name that actually express what is happening as this is a status update not a soft deletion?

app/services/invoices/delete_service.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    activity_loggable(
+      action: "invoice.deleted",
+      record: -> { invoice }
+    )
+
+    def call
+      return result.not_found_failure!(resource: "invoice") unless invoice
+
+      invoice.with_lock do
```
ancorcruz: `ActiveRecord::Rollback `won't actually roll back if this service is ever called inside an outer transaction.

`invoice.with_lock` opens a transaction with default options, so it joins a caller's transaction when one already exists rather than creating a savepoint. `ActiveRecord::Rollback` is only honored where there's a real transaction (or savepoint) to roll back to — inside a joined transaction it's caught and swallowed with no rollback SQL emitted. So if a future caller (e.g. a batch-delete endpoint) wraps `Invoices::DeleteService.call` in its own transaction and a credit note fails to delete, the credit notes already deleted in `soft_delete_credit_notes!` would ride along into the outer commit while the caller is handed a "failure" result — a partial, inconsistent state.

This isn't the pattern `VoidService` uses, despite the similar `with_lock` + guard-return shape: `VoidService` aborts via real exceptions (`RecreditService.call!`, `CreateService.call!`, `raise_if_error!`), which propagate through every transaction layer and stay atomic even when nested. This PR instead uses non-bang `.call` + `raise ActiveRecord::Rollback`, which only behaves correctly when it's the outermost transaction — which it is today (service-only PR), so there's no live bug, but the description says API exposure is coming and deletes tend to get batched.

Cleanest fix that keeps the current "clean failure result, no exception" contract is to give the lock its own savepoint:

```
invoice.transaction(requires_new: true) do
  invoice.lock!
  return result.not_allowed_failure!(code: "not_deletable") unless invoice.draft?
  return result.not_allowed_failure!(code: "invoice_synced_to_external_system") if synced_externally?

  soft_delete_credit_notes!
  invoice.mark_as_deleted!
end
```

`requires_new: true` makes it a SAVEPOINT, so `ActiveRecord::Rollback` unwinds the credit-note deletes without touching the outer transaction, and the method still returns a failure result. (Switching to `CreditNotes::DeleteService.call!` would also restore atomicity but changes the contract to "raises on failure," which the current specs and the future API layer wouldn't expect.)

Non-blocking for this PR, but worth tracking so it's handled before a transactional caller is added.

ancorcruz (changes requested):

app/services/invoices/delete_service.rb
endenis: Indeed, it's not necessary. I removed it.
Thank you 🙂

spec/services/invoices/delete_service_spec.rb
endenis: Done, thank you

spec/services/invoices/delete_service_spec.rb
endenis: You're right. I added a test for this (line 108)

app/services/invoices/delete_service.rb
endenis: I renamed it into `mark_credit_notes_as_deleted!` 👍

app/services/invoices/delete_service.rb
endenis: I changed the code to use `invoice.with_lock(requires_new: true) do ...` and I added a test to check that it does not rollback an outer transaction 👍

spec/services/invoices/delete_service_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+        ActiveRecord::Base.transaction do
+          other_invoice.update!(payment_overdue: true)
+          delete_service.call
+        end
+
+        # the surrounding transaction's change is kept...
+        expect(other_invoice.reload.payment_overdue).to be(true)
+        # ...while the service's own cascade was rolled back
+        expect(invoice.reload).to be_draft
+        expect(deletable_credit_note.reload).to be_draft
```
ancorcruz, line 141: Non-blocking — test hardening. This assertion (and the regression value of this whole test) depends on `deletable_credit_note` being processed before `blocking_credit_note`, but `mark_credit_notes_as_deleted!` iterates with `find_each`, which orders by primary key — and `credit_notes.id` is a random UUID. So the processing order is effectively random:

- deletable first → it's marked deleted, then the blocking one raises `ActiveRecord::Rollback`, unwinding it. The savepoint rollback is genuinely exercised. ✅
- blocking first → nothing was deleted before the failure, so there's nothing to roll back. The assertions still pass — but they'd pass against the pre-fix code too (no `requires_new:`), so the test isn't guarding the fix in that ordering.

Net effect: the test never flakes (always green), but ~half the runs don't actually validate the savepoint behavior it's meant to pin down. To make it deterministic, guarantee the deletable credit note is handled first — e.g. stub the relation to return an ordered array:

```
allow(invoice).to receive(:credit_notes).and_return(
  CreditNote.where(id: [deletable_credit_note.id, blocking_credit_note.id]).order(:created_at)
)
```

(or add a `deterministic .order(...)` to the `find_each` scope in the service, if a stable processing order is desirable there anyway). The production fix itself is correct — this is purely about making the test a reliable guard.

ancorcruz (approved):
Looks great. Thanks for addressing the feedback.

I have just left one more comment.

spec/services/invoices/delete_service_spec.rb
endenis: Fixed, thank you!

PR #5959, opened 2026-07-16
[ING-483] fix(cache): Ensure correct cache expiration with clickhouse

vincent-pochet:
## Context

Charge usage cache was invalidated eagerly, deleting entries on every event reception. With ClickHouse events are ingested asynchronously, so it leads to race condition depending how the cache was invalidated and rebuilt.

## Description

This PR adds a lazy cache validation strategy behind the `lazy_charge_usage_cache` feature flag.
Entries are stamped with the ingestion timestamp of the most recent event they were computed from, and recomputed at read time when a newer event exists, instead of being deleted on reception.

`Events::BillingPeriodFilterService` now returns the last-seen timestamp per charge/filter (`created_at` on Postgres, `enriched_at` on ClickHouse), exposed by the event store stores. The result becomes `{ charge_id => { filter_id => last_seen_at } }`.

Recurring charges carry over from previous periods and might not have in-period event, so they are seeded with the period start. This makes sure that the cache will be invalidated at the start of the period and lets any in-period event invalidate it and guards against ingestion lag.

### DIFF
--- a/app/config/feature_flags.yaml
+++ b/app/config/feature_flags.yaml
@@ -41,3 +41,7 @@ fixed_charge_usage_delta_migration:
 multi_connection:
   description: "Allow organizations to configure multiple connections of the same type per customer"
   owners: ["@lovrocolic", "@mariohd"]
+
+lazy_charge_usage_cache:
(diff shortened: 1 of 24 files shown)


annvelents (approved):
AWESOME! 🤩 🚀

PR #5961, opened 2026-07-16
refactor: replace direct OpenStruct usages with Data value objects

ancorcruz:
## Context

OpenStruct is banned in this codebase, but a handful of direct usages remained. This removes all of them except the deprecated `BaseService::LegacyResult < OpenStruct`, whose removal depends on migrating the ~89 services still using the default result and will be done separately.

## Description

- `FetchDraftInvoiceTaxes` mutation: replace the fake invoice/fee `OpenStructs` with `DraftInvoice/DraftFee` `Data` stand-ins that declare exactly the interface the Anrok/Avalara tax payloads read.
- `ApplyProviderTaxesService`: replace the internal fee-grouping struct with a `GroupingTax` `Data` value object.
- `device_info_benchmark.rb`: use a `Data` request and drop the `ostruct` require.
- `custom_aggregation.rake`: use a `Data` placeholder event.

Behavior is unchanged; unanticipated attribute reads now raise instead of silently returning `nil`.

### DIFF
--- a/app/graphql/mutations/integrations/fetch_draft_invoice_taxes.rb
+++ b/app/graphql/mutations/integrations/fetch_draft_invoice_taxes.rb
@@ -27,11 +27,33 @@ module Mutations
 
       private
 
-      # Note: We need to pass invoice object to the service that return taxes. This service should
-      # work with real invoice objects. In this case, it should also work with invoice that is not stored yet,
-      # because we need to fetch taxes for one-off invoice UI form.
+      # One-off invoices previewed from the UI are not persisted yet, so we build lightweight
+      # stand-ins exposing only the subset of the Invoice / Fee interface the tax payloads read.
+      # They fail loudly if a payload starts reading an attribute we did not anticipate.
+      DraftInvoice = Data.define(:issuing_date, :currency, :customer) do
+        def voided? = false
+      end
+
+      DraftFee = Data.define(:add_on_id, :item_id, :sub_total_excluding_taxes_amount_cents) do
+        def id = nil
+
+        def item_key = nil
+
+        def units = nil
+
+        def amount_cents = nil
+
+        def charge? = false
+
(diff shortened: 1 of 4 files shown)


toommz (approved):

vincent-pochet (approved):

app/graphql/mutations/integrations/fetch_draft_invoice_taxes.rb
```diff
@@ -27,11 +27,33 @@ def resolve(**args)
 
       private
 
-      # Note: We need to pass invoice object to the service that return taxes. This service should
-      # work with real invoice objects. In this case, it should also work with invoice that is not stored yet,
-      # because we need to fetch taxes for one-off invoice UI form.
+      # One-off invoices previewed from the UI are not persisted yet, so we build lightweight
+      # stand-ins exposing only the subset of the Invoice / Fee interface the tax payloads read.
+      # They fail loudly if a payload starts reading an attribute we did not anticipate.
+      DraftInvoice = Data.define(:issuing_date, :currency, :customer) do
```
tiagolupepic, line 33: just an idea, we started to add these PORO objects in each class and it's pretty convenient and easy to spot what the result we need to use. However, are we considering to share these PORO objects among services? We could put them in a single directory

ancorcruz: Good question. I'd keep these particular ones inline: DraftInvoice/DraftFee aren't reusable POROs — they're scoped stand-ins that each declare only the slice of the Invoice/Fee interface a specific tax payload reads, and fail loudly on anything unanticipated. Their value is being co-located with the contract they enforce.

That said, I agree a shared directory makes sense for POROs that genuinely get reused across services. Happy to establish that convention when we have a concrete case for it — this refactor just didn't produce one.

PR #5962, opened 2026-07-17
[STA-1]  misc(OpenStruct): Add missing Result on services

vincent-pochet:
## Description

This PR adds `Result` definition to replace `LegacyResult` in more services where is was missing.


toommz (approved):
🧹

PR #5966, opened 2026-07-17
fix(payments): scope checkout delay to manual wallet top-ups

brunomiguelpinto:
## Context

The checkout auto-payment delay (org-level, #5937) defers an invoice's first automatic payment for organizations that use hosted checkout, so the auto-payment can't race a customer's manual checkout and double-charge them.

That delay currently applies to **every** credit (wallet top-up) invoice. But automatic top-ups (`interval` / `threshold`) are triggered server-side —there's no customer checkout URL competing with them, so delaying them just postpones credit granting with no upside. All observed top-up double charges were `source = manual`.

## Change

- `defer_for_checkout_organization?` now skips **non-manual** wallet top-ups (`non_manual_wallet_topup?`). Only `manual` top-ups (which a customer can also pay via a checkout URL) are delayed. Whitelisting `manual` keeps any
  future server-triggered source excluded by default.

### DIFF
--- a/app/services/invoices/payments/create_service.rb
+++ b/app/services/invoices/payments/create_service.rb
@@ -138,7 +138,7 @@ module Invoices
       end
 
       def non_manual_wallet_topup?
-        invoice.credit? && invoice.wallet_transactions.where.not(source: :manual).exists?
+        invoice.credit? && invoice.wallet_transactions.inbound.where.not(source: :manual).exists?
       end
 
       def should_process_payment?
--- a/app/services/invoices/payments/create_service.rb
+++ b/app/services/invoices/payments/create_service.rb
@@ -127,17 +127,18 @@ module Invoices
         @provider ||= invoice.customer.payment_provider&.to_sym
       end
 
-      # Delay the first automatic payment for every customer of an organization that uses hosted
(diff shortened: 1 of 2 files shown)


app/services/invoices/payments/create_service.rb
```diff
@@ -127,19 +127,20 @@ def provider
...
-        # Payment-gated subscriptions activate synchronously, so charge now instead of deferring.
-        return false if invoice.subscription_payment_gated?
+        return false if invoice.subscription? && invoice.subscription_payment_gated?
+        return false if non_manual_wallet_topup?
 
         PaymentIntent.where(organization_id: invoice.organization_id).exists?
       end
 
+      def non_manual_wallet_topup?
+        invoice.credit? && invoice.wallet_transactions.where.not(source: :manual).exists?
```
annvelents: `wallet_transactions.inbound`
because if it's outbound it means that a wallet transaction was created to apply prepaid credits

annvelents: but it will not be manual!
nvm

annvelents (approved):

app/services/invoices/payments/create_service.rb
brunomiguelpinto: i will add it

PR #5970, opened 2026-07-17
feat(auth): add Microsoft Entra ID (OIDC) SSO

mikeh-lago:
Mirrors the existing Okta SSO flow with a generic OIDC path for Microsoft Entra ID: integration model, auth services (authorize, login, invite accept), integration CRUD services, GraphQL API and premium gating. Entra specifics: tenant-scoped v2.0 endpoints, Microsoft Graph userinfo, case-insensitive email check with preferred_username fallback.


app/services/auth/entra_id/authorize_service.rb
```diff
@@ -0,0 +1,55 @@
...
+      def call
+        check_invite(email) if invite_token.present?
+        check_entra_id_integration(email)
+
+        params = {
+          client_id: result.entra_id_integration.client_id,
+          response_type: "code",
+          response_mode: "query",
+          scope: "openid profile email",
+          redirect_uri: "#{ENV["LAGO_FRONT_URL"]}/auth/entra/callback",
```
toommz: We might want to use `Rails.application.config.lago_front_url` as it sets a sensible default, wdyt?

app/services/auth/entra_id/base_service.rb
```diff
@@ -0,0 +1,79 @@
...
+        response = token_client.post_url_encoded(params, {})
+        result.entra_id_access_token = response["access_token"]
+      end
+
+      # NOTE: Entra ID only returns the `email` claim when the account has a mail
+      #       attribute and the claim is configured, so we fall back to
+      #       `preferred_username` (the UPN). Comparison is case-insensitive as
+      #       Entra ID casing is not guaranteed to match the typed email.
+      def check_userinfo(email)
+        userinfo_client = LagoHttpClient::Client.new("https://graph.microsoft.com/oidc/userinfo")
```
toommz: Is it intended to have hardcoded the URL here?

app/services/auth/entra_id/authorize_service.rb
```diff
@@ -0,0 +1,55 @@
...
+          state:
+        }
+        result.url = URI::HTTPS.build(
+          host: result.entra_id_integration.host,
+          path: "/#{result.entra_id_integration.tenant_id}/oauth2/v2.0/authorize",
```
toommz, line 32: I'm far from being a security expert, but I would say that letting a user provided value that we don't validate be a part of a path can be risky 🤔

app/services/auth/entra_id/authorize_service.rb
mikeh-lago: This branch was just behind main so I'd missed it. Switched both Entra redirect URIs to Rails.application.config.lago_front_url and rebased the branch onto main so it picks it up. Thanks for the nudge.

app/services/auth/entra_id/base_service.rb
mikeh-lago: Azure AD exposes the OIDC userinfo through Microsoft Graph (graph.microsoft.com/oidc/userinfo), not the tenant login host, so it's the same endpoint for every commercial Entra tenant. I've pulled it into a named constant (same pattern as BASE_URL in the aggregator service).

app/services/auth/entra_id/authorize_service.rb
mikeh-lago: tenant_id and host both get interpolated into the authorize/token URL (host and path), so I added a validation restricting them to a safe URL segment ([a-zA-Z0-9.-]), which blocks slashes and other characters that could break out of the intended URL. Did it as a custom validate mirroring the existing domain_uniqueness rather than a format: validator, to match how this model already validates. Covered by two specs.

toommz (approved):

PR #5972, opened 2026-07-17
[ING-449] feat(integration_customers): add is_default and category

mariohd:
## Context

Groundwork for the multi-connection work on the customer: a connection will be marked as the default for its category, and routed per category. This adds the schema ahead of the routing and backfill logic.

## Changes

Add `is_default` (`boolean`, default `false`) and a nullable `category` column to `integration_customers`, backed by a new `integration_customer_category` enum (`tax`, `accounting`, `crm`). No-op and backward-compatible — no backfill, no index, and nothing reads the columns yet (`category` will be derived from the STI type in a later backfill).

## Testing

Migration applied and rolled back cleanly; `db/structure.sql` regenerated and model annotations updated. No behavioral change to test.

### DIFF
--- a/db/migrate/20260717163416_add_is_default_and_category_to_integration_customers.rb
+++ b/db/migrate/20260717163416_add_is_default_and_category_to_integration_customers.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddIsDefaultAndCategoryToIntegrationCustomers < ActiveRecord::Migration[8.0]
+  def change
+    create_enum :connection_category, %w[payment tax accounting crm]
+
+    add_column :integration_customers, :is_default, :boolean, default: false, null: false
+    add_column :integration_customers, :category, :connection_category
+  end
+end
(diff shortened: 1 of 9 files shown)


db/migrate/20260717163416_add_is_default_and_category_to_integration_customers.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddIsDefaultAndCategoryToIntegrationCustomers < ActiveRecord::Migration[8.0]
+  def change
+    create_enum :integration_customer_category, %w[tax accounting crm]
```
lovrocolic: we will also use this enum in billing_object_connections table. Can we maybe use some other neme?

lovrocolic:
It would be great if we can merge https://github.com/getlago/lago-api/pull/5971 first (so that merge conflicts are resolved in this PR)

lovrocolic (approved):

PR #5975, opened 2026-07-20
fix: add null-byte Sanitizing

annvelents:
## Context

When null-byte is sent as `customer.firstname` or `lastname`, customer creation crashes with 500 error

## Description

Added a null-byte-sanytiser that can be included in any models requirung null-byte sanytising


vincent-pochet (approved):
🚀

brunomiguelpinto (approved):
LGTM

PR #5978, opened 2026-07-20
fix(stripe-webhooks): Remove transaction from PaymentProviderCustomers::Stripe::UpdatePaymentMethodService

osmarluz:
## Context

#5878 wrapped the `StripeCustomer` settings write and the payment-method find-or-create in a single transaction to keep them in sync. However, `PaymentMethods::FindOrCreateFromProviderService` recovers from a concurrent creation of the same payment method (`ActiveRecord::RecordNotUnique`) by querying the winning record — and PostgreSQL rejects any statement issued inside a transaction that has already failed. When `HandleEventJob` raced `FetchDefaultPaymentMethodJob` at customer creation, the losing insert aborted the wrapping transaction and the recovery query raised `PG::InFailedSqlTransaction`, dead-lettering the job since `ActiveRecord::StatementInvalid` is not retried.

## Description

Remove the wrapping transaction so the recovery query runs on a healthy connection again. A failure between the settings write and the default flip is now reconciled by the job retries introduced in #5878, since the service is idempotent.

### DIFF
--- a/app/services/payment_provider_customers/stripe/update_payment_method_service.rb
+++ b/app/services/payment_provider_customers/stripe/update_payment_method_service.rb
@@ -17,22 +17,20 @@ module PaymentProviderCustomers
         return result.not_found_failure!(resource: "stripe_customer") unless stripe_customer
         return result.service_failure!(code: :deleted_customer, message: "Customer associated to this stripe customer was deleted") if deleted_customer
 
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):

ancorcruz (approved):
LGTM

PR #5979, opened 2026-07-20
feat(db): Avoid RDS Proxy connection pinning

binaryknot:
## Context

Lago's EU staging deployment is introducing AWS RDS Proxy in front of PostgreSQL for connection multiplexing and faster blue/green cutovers. However, RDS Proxy pins any connection where it detects a client-side session SET. ActiveRecord's PostgreSQLAdapter sends five SET commands during configure_connection and a sixth (SET SESSION timezone) via add_pg_decoders. With those pins in place, the proxy degrades to a 1:1 passthrough and connection multiplexing is lost.

AWS does not offer a session-pinning filter for PostgreSQL: the only available filter (EXCLUDE_VARIABLE_SETS) is MySQL-only. For PostgreSQL, the path to unpinned connections is to stop sending session SETs from the client and configure equivalent state via the proxy's initQuery.

This was re-opened to test in the lago-staging-eu-1 environment with a new RDS proxy specifically for Rails.

## Description

Adds Lago::RdsProxy::ConnectionPatch, a module prepended onto ActiveRecord::ConnectionAdapters::PostgreSQLAdapter when the DATABASE_VIA_RDS_PROXY environment variable is set. The patch replaces configure_connection with an equivalent that skips all session-level SET commands, and overrides reconfigure_connection_timezone as a no-op to prevent add_pg_decoders from emitting SET SESSION timezone.

Local dev, tests, and deployments without the flag are unaffected.

The proxy's initQuery must cover the six equivalent settings: client_encoding, client_min_messages, search_path, standard_conforming_strings, intervalstyle, timezone.

### DIFF
--- a/config/initializers/rds_proxy.rb
+++ b/config/initializers/rds_proxy.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+# Activates the RDS Proxy connection patch. See lib/lago/rds_proxy/connection_patch.rb
+# for what the patch does and why.
+
+if ENV["DATABASE_VIA_RDS_PROXY"].present?
+  require "lago/rds_proxy/connection_patch"
+
+  # The pg gem auto-sends SET client_encoding when default_internal is set,
+  # bypassing configure_connection. Reset to nil to suppress.
+  Encoding.default_internal = nil
+
+  ActiveSupport.on_load(:active_record) do
+    ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.prepend(
+      Lago::RdsProxy::ConnectionPatch
+    )
+  end
+end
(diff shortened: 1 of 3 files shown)


vincent-pochet (approved):
It looks globally good.
The only point of attention is that the specs are faking the connection adapter and not relying on the real PostgresSQLAdapter, we should at least add a scenario/an integration test validating the behavior with the real adapter otherwise we might not spot any change/issues in the integration with the adapter in future version of rails.

Pointed by Claude:
```
If proxy initQuery omits/mismatches any of the six settings, failures are silent and dangerous — 
wrong timezone or intervalstyle corrupts timestamp interpretation; wrong client_encoding corrupts strings. 
Code can't guard this. Ensure runbook + monitoring cover initQuery parity. 
Consider a boot-time assertion that queries SHOW timezone/SHOW client_encoding and warns on mismatch.
```

vincent-pochet (approved):

PR #5982, opened 2026-07-21
fix(payments-syncs): Support payments sync on async resolution paths

osmarluz:
## Context

`should_sync_payment?` gated on invoice finalization and the accounting integration, but not on the payment's status, and the sync job was only enqueued from the synchronous charge path and manual payments. That left the sync wired to the wrong moment, in both directions:

- Payments reaching `succeeded` through a provider webhook (3DS/SCA, async off-session confirmations, recovery of a payment missing in the database) were never synced, so the invoice appeared in NetSuite/Xero with no payment applied against it.
- Payments coming back `processing` from the charge call — Stripe 3DS retries, every GoCardless charge — were synced immediately, applying a payment against the invoice before the money settled. There is no update or delete path to correct that if the payment later fails.

## Description

`Payment#should_sync_payment?` now also requires the payment to have succeeded, so no call site can sync an unsettled payment, and `Integrations::Aggregator::Payments::CreateJob` is enqueued from the Stripe, Adyen and GoCardless webhook settlement paths, which is where those payments actually resolve.

Cashfree, Flutterwave and MoneyHash are left out because they are community-maintained. The new guard still applies to them: Cashfree and MoneyHash charges that sit at `pending` were being synced at charge time and no longer are, so those providers have no payment sync until the same enqueue is added to them.

Duplicated provider webhooks are already handled — `CreateJob` is declared `unique :until_executed`, and a job that runs once the sync has completed no-ops on the existing `payload.integration_payment` guard. Neither of those covers a retry after a timed-out POST: the aggregator payload carries no idempotency key, so a `Net::ReadTimeout` on a request NetSuite has already committed can create the payment twice. That behaviour is pre-existing and tracked separately.


app/services/integrations/aggregator/payments/create_service.rb
```diff
@@ -20,30 +20,35 @@ def call
           return result unless integration
           return result unless integration.sync_payments
           return result unless invoice.finalized?
-          return result if payload.integration_payment
 
-          throttle!(:netsuite, :xero)
+          LockService.call(payment:, timeout_seconds: 0, transaction: false) do
```
osmarluz, line 24: `timeout_seconds` is 0 here because we don't care about waiting and retrying (if the lock is taken, another process is already syncing this same payment, so this caller can safely bail and let the holder finish)

app/services/invoices/payments/adyen_service.rb
```diff
@@ -34,7 +34,10 @@ def update_payment_status(provider_payment_id:, status:, amount_cents: nil, meta
         payment.payable_payment_status = payable_payment_status
         payment.save!
 
-        deliver_webhook if payable_payment_status.to_sym == :succeeded
+        if payable_payment_status.to_sym == :succeeded
+          deliver_webhook
+          Integrations::Aggregator::Payments::CreateJob.perform_later(payment:) if payment.should_sync_payment?
```
ancorcruz, line 39: have you consider if this work for payment gated subscriptions? `should_sync_payment?` checks if invoice is finalized, won't it be 'open' for payment gated subs?

app/services/invoices/payments/gocardless_service.rb
```diff
@@ -26,7 +26,10 @@ def update_payment_status(provider_payment_id:, status:)
         payment.payable_payment_status = payable_payment_status
         payment.save!
 
-        deliver_webhook if payable_payment_status.to_sym == :succeeded
+        if payable_payment_status.to_sym == :succeeded
+          deliver_webhook
+          Integrations::Aggregator::Payments::CreateJob.perform_later(payment:) if payment.should_sync_payment?
```
ancorcruz, line 31: the same applies here.

app/services/invoices/payments/stripe_service.rb
```diff
@@ -42,7 +42,10 @@ def update_payment_status(organization_id:, status:, stripe_payment:, amount_cen
         payment.error_code = stripe_payment.error_code if stripe_payment.error_code
         payment.save!
 
-        deliver_webhook if payable_payment_status.to_sym == :succeeded
+        if payable_payment_status.to_sym == :succeeded
+          deliver_webhook
+          Integrations::Aggregator::Payments::CreateJob.perform_later(payment:) if payment.should_sync_payment?
```
ancorcruz, line 47: and here.

ancorcruz (approved):

app/services/invoices/payments/adyen_service.rb
osmarluz: Yes and for now we're not taking that into account because for a payment gated payment to be synced, the invoice would need to be synced as well, which is currently something we don't do because of the open status. Therefore right now it matches for the payment gated flow: neither the invoice nor the payment are synced to Netsuite

mikeh-lago (approved):

PR #5983, opened 2026-07-21
feat(stripe): Add consent collection setting

ancorcruz:
## Context

Stripe Checkout can require customers to accept a merchant's terms of service before paying, via `consent_collection[terms_of_service]`. Lago did not expose this option on Stripe connections.

## Description

Add a `require_terms_of_service_consent` accessor on the Stripe provider, stored in the existing settings JSONB and defaulting to `false`, so existing connections are unaffected. Extend `settings_accessors` with an optional `default:` value so an absent key reads as `false` instead of `nil`. Consuming the flag in the checkout flows comes in follow-up changes.

### DIFF
--- a/app/models/concerns/settings_storable.rb
+++ b/app/models/concerns/settings_storable.rb
@@ -4,10 +4,11 @@ module SettingsStorable
   extend ActiveSupport::Concern
 
   class_methods do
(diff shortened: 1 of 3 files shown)


osmarluz (approved):
🚀

PR #5990, opened 2026-07-21
[ING-480] feat(alerts): Reduce alerting interval

vincent-pochet:
## Context

The dedicated alerts worker processes two queues, `alerts_high_priority` and `alerts`. Wallet alert jobs were routed to the low-priority `alerts` queue, where they queued behind the bulk subscription-activity processing. Low-balance and depleted-credit alerts are latency-sensitive, so they should not wait on that backlog.

## Description

- Route `ProcessWalletAlertsJob` to `alerts_high_priority` (when `SIDEKIQ_ALERTS` is enabled) so wallet threshold alerts are processed ahead of usage-activity jobs.
- Switch the alerts worker queues from strict ordering to weighted (`alerts_high_priority` 2, `alerts` 1). The high-priority queue is favored but the `alerts` queue is still polled every cycle, so promoting wallet alerts cannot starve subscription-activity processing.


ancorcruz (approved):
LGTM

PR #5991, opened 2026-07-21
fix(payment-providers): backfill orphaned customers

ancorcruz:
## Context

Before PaymentProviders::DestroyService soft deleted its provider customers, it only nullified their payment_provider_id and left the rows live (deleted_at IS NULL). A later provider reconnection then created a second live provider customer for the same customer and type, which breaks the exports_customers view and violates the intended one-live-row invariant. The destroy service is now fixed, but rows orphaned before that fix are still live in existing databases.

## Description

Add a batched data migration that soft deletes the already orphaned provider customers (payment_provider_id IS NULL AND deleted_at IS NULL), setting deleted_at from updated_at, the moment they were nullified. This clears the accumulated duplicates so affected exports recover.


osmarluz (approved):
🚀

PR #5992, opened 2026-07-21
[ING-422] Stop auto top-up failing when the refill exceeds the max limit

aquinofb:
## Context

When a wallet's automatic top-up needs to refill more than the wallet's maximum paid top-up amount, the refill was computed by the recurring rule and then rejected by the transaction creation step. Nothing was created and nothing was surfaced: no transaction, no invoice, no webhook. Auto top-up silently stopped, the wallet kept running dry, and the refill was retried and failed on every balance refresh.

The two rules contradicted each other: the recurring rule intentionally lets a target refill go past the max, because reaching the target balance matters most, but transaction creation re-checked the max and rejected the amount.

## Changes

- Threshold and interval refills now create the transaction with the limit check disabled. The recurring rule has already computed the amount, and for a target rule it deliberately reaches past the maximum, so re-checking the limits at that point could only reject a refill that was meant to happen.
- The interval flow previously left the setting out of the payload entirely, so it defaulted to enforcing the limits. It now matches the threshold flow.

Rejecting a misconfigured fixed rule at wallet creation is not part of this change: `Wallets::RecurringTransactionRules::CreateService` already does it, inside the same transaction, so an out-of-limits rule fails the create today.

## Worth knowing when reviewing

The setting that disables the check turns off the minimum as well as the maximum, because the validator returns early before testing either bound. Target rules are unaffected: the recurring rule still clamps them up to the minimum before the transaction is created. A fixed rule whose paid amount sits below the wallet minimum would now go through instead of failing, but such a rule can no longer be configured, since both the wallet create and update paths reject it. Only rules stored before that validation existed can still be in that state.

### DIFF
--- a/app/services/wallets/threshold_top_up_service.rb
+++ b/app/services/wallets/threshold_top_up_service.rb
@@ -22,7 +22,8 @@ module Wallets
         invoice_requires_successful_payment: rule.invoice_requires_successful_payment?,
         metadata: rule.transaction_metadata,
         name: rule.transaction_name,
-        ignore_paid_top_up_limits: rule.ignore_paid_top_up_limits?,
+        # The rule already computed the amount; re-checking limits here would reject it.
+        ignore_paid_top_up_limits: true,
         purchase_order_number: rule.resolved_purchase_order_number
       }
 
--- a/app/services/wallets/threshold_top_up_service.rb
+++ b/app/services/wallets/threshold_top_up_service.rb
@@ -30,7 +30,7 @@ module Wallets
         invoice_requires_successful_payment: rule.invoice_requires_successful_payment?,
         metadata: rule.transaction_metadata,
         name: rule.transaction_name,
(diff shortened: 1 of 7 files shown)


app/services/wallets/threshold_top_up_service.rb
```diff
@@ -30,7 +30,7 @@ def call
         invoice_requires_successful_payment: rule.invoice_requires_successful_payment?,
         metadata: rule.transaction_metadata,
         name: rule.transaction_name,
-        ignore_paid_top_up_limits: rule.ignore_paid_top_up_limits?,
+        ignore_paid_top_up_limits: true,
```
toommz: Maybe this is intended, but hardcoding this to true also bypasses the configured maximum for legacy fixed rules whose ignore flag is false.

toommz (approved):

app/services/wallets/threshold_top_up_service.rb
aquinofb: Good catch! Applied.

PR #5993, opened 2026-07-22
fix(invoices): Lock invoice during refresh and finalize service

tiagolupepic:
This commit changes the RefreshDraftAndFinalizeService to lock invoices and avoid RecordNotUnique errors

The root cause of the errors is the unique constraint 
`index_invoice_subscriptions_on_invoice_id_and_subscription_id` having the lock
we avoid to raise such exception.

Fixes BIL-99

### DIFF
--- a/app/services/invoices/refresh_draft_and_finalize_service.rb
+++ b/app/services/invoices/refresh_draft_and_finalize_service.rb
@@ -13,11 +13,12 @@ module Invoices
       return result.not_found_failure!(resource: "invoice") if invoice.nil?
       return result.forbidden_failure! unless invoice.subscription?
 
-      return result unless invoice.draft?
-      return result.forbidden_failure!(code: "cannot_finalize_with_pending_taxes") if invoice.tax_pending?
-      drafted_issuing_date = invoice.issuing_date
+      invoice.with_lock do
+        result.invoice = invoice
+        return result unless invoice.draft?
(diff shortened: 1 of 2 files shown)


app/services/invoices/refresh_draft_and_finalize_service.rb
```diff
@@ -13,11 +13,12 @@ def call
...
-      drafted_issuing_date = invoice.issuing_date
+      invoice.with_lock do
+        result.invoice = invoice
+        return result unless invoice.draft?
+        return result.forbidden_failure!(code: "cannot_finalize_with_pending_taxes") if invoice.tax_pending?
+        drafted_issuing_date = invoice.issuing_date
 
-      ActiveRecord::Base.transaction do
         invoice.issuing_date = issuing_date
         refresh_result = Invoices::RefreshDraftService.call(invoice:, context: :finalize)
```
osmarluz, line 23: @tiagolupepic `Invoices::RefreshDraftService` is used on some other places as well (like `Invoices::RefreshDraftJob` and `Api::V1::InvoicesController#refresh`). It might make sense to move the lock one level down if we still see `RecordNotUnique` errors because of other flows

osmarluz (approved):
🚀

app/services/invoices/refresh_draft_and_finalize_service.rb
tiagolupepic: good point! I'll merge this PR and keep the eyes on that, specially in `refresh` endpoint.

PR #5994, opened 2026-07-22
feat(stripe): persist consent collection setting

ancorcruz:
Part of INT-163

## Context

The `require_terms_of_service_consent` flag exists on the Stripe provider but could not yet be set through the provider create/update flow.

## Description

Copy `require_terms_of_service_consent` from the arguments in `create_or_update`, mirroring the existing `supports_3ds` handling, so the setting can be turned on or off per connection.


osmarluz (approved):
🚀

PR #5995, opened 2026-07-22
feat(purchase_order_number): Restrict PO number editing on subscriptions

D1353L:
## Description

Requirement: the PO number can be edited any time while the subscription is in `pending` or `active` state.


endenis (approved):

ancorcruz (approved):
LGTM

PR #5997, opened 2026-07-22
[ING-452] feat(connections): backfill connections and add uniqueness indexes

mariohd:
## Context

Groundwork for the multi-connection work on the customer. The connection codes, categories, and default flags are now backfilled (previous change), so the routing keys can be enforced at the database level. This adds the uniqueness indexes, created concurrently so the change is safe to deploy without locking the tables.

## Changes

Add four unique indexes, created `concurrently`:

- `integration_customers (customer_id, category, code)` — one code per customer per category
- `integration_customers (customer_id, category) WHERE is_default` — one default per customer per category
- `payment_provider_customers (customer_id, code) WHERE deleted_at IS NULL` — one code per customer (active rows)
- `payment_provider_customers (customer_id) WHERE is_default AND deleted_at IS NULL` — one default payment connection per customer (active rows)

The indexes reflect the backfilled data and are effectively empty on rows that predate the routing columns (code/category NULL, is_default false), so creation is safe. The existing per-type constraints are kept for now and relaxed in a later change.

## Testing

Migration applied concurrently on dev and test; index definitions verified against the database. Model/backfill specs pass (63 examples, 0 failures) — the new constraints don't break existing behavior. Model annotations regenerated to reflect the new indexes.


db/migrate/20260722091941_add_connection_uniqueness_indexes.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class AddConnectionUniquenessIndexes < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
```
lovrocolic: Is it safe to have those indexes before the backfill is really done?

mariohd: Yes, it's safe, the records will not collide in the index

lib/tasks/multi_customer_connections.rake
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+namespace :multi_customer_connections do
+  desc "Backfill connection codes + category + mark existing connection is_default (ING-452). " \
```
lovrocolic: Can we please remove ticket references from the code (since it is OSS)?

app/services/multi_customer_connections/backfill_connection_codes_service.rb
```diff
@@ -0,0 +1,121 @@
...
+#   * marks the single existing connection of each category `is_default = true`
+#
+# Idempotent: codes/categories only touch NULL rows; defaults only apply when the
+# (customer, category) group has no default yet. Re-running is a no-op once complete.
+#
+# Safety: production data was verified to hold at most one connection per category
+# per customer. Should a group ever contain more than one, we skip it and count it
+# as a conflict instead of writing two defaults.
+module MultiCustomerConnections
+  class BackfillConnectionCodesService < BaseService
```
lovrocolic: Maybe the name is a bit misleading since we update more columns here

lovrocolic (approved):

vincent-pochet (approved):
Looks good 👍 

Tow comments maybe:
- Can you add a title to the PR ;)
- For index creation, I would suggest to define each index in a dedicated migration, it could help if anything goes wrong with one of the index as we are in non atomic mode (`disable_ddl_transaction`)

lovrocolic:
> Looks good 👍
> 
> Tow comments maybe:
> 
> * Can you add a title to the PR ;)
> * For index creation, I would suggest to define each index in a dedicated migration, it could help if anything goes wrong with one of the index as we are in non atomic mode (`disable_ddl_transaction`)

Mario addressed both comments so I will merge this one

PR #5999, opened 2026-07-22
feat(multi-connection): discard payment provider customer with dedicated mutation

lovrocolic:
## Context

Deleting a payment provider customer connection was only possible as a side-effect of a customer update. As part of multi-connections, we need a dedicated action to delete a specific connection.

## Description

Add a `destroyPaymentProviderCustomer` GraphQL mutation backed by `PaymentProviderCustomers::DestroyService`, which soft-deletes the connection and destroys its associated payment methods.


app/services/payment_provider_customers/destroy_service.rb
osmarluz, line 1: @lovrocolic when the payment provider customer is discarded through a customer update we also set the customer's `payment_provider` and `payment_provider_code` to `nil`. Is it something that needs to be done here as well?

lovrocolic: Hey @osmarluz! Those two fields should not be relevant anymore since we will need to handle multiple payment providers for the same customer and the fields are relevant only when we have one payment provider per customer.

We will keep those fields for some transient period just to keep backward compatibility.

mariohd (approved):

PR #6001, opened 2026-07-22
feat(stripe): expose consent setting in GraphQL

ancorcruz:
Part of INT-163

## Context

The `require_terms_of_service_consent` flag can be persisted on the Stripe provider but was not reachable through the GraphQL API.

## Description

Add the `require_terms_of_service_consent` argument to the Stripe create and update inputs and expose it as a field on the `StripeProvider` type, mirroring `supports_3ds`. Regenerate the schema dump accordingly.


osmarluz (approved):
🚀

PR #6002, opened 2026-07-22
feat(stripe): consent on setup checkout URL

ancorcruz:
Part of INT-163

## Context

The consent collection setting on the Stripe connection was stored and exposed but not yet applied to any checkout flow.

## Description

When require_terms_of_service_consent is enabled on the connection, add consent_collection[terms_of_service]=required to the setup-mode Checkout Session behind the customer checkout URL, so payers must accept the terms of service before saving a payment method. No change when the setting is off.


osmarluz (approved):
🚀

PR #6003, opened 2026-07-22
feat(stripe): consent on payment request URL

ancorcruz:
Part of INT-163

## Context

The consent collection setting applied to the customer checkout URL but not to the payment-request payment page.

## Description

When `require_terms_of_service_consent` is enabled on the connection, add `consent_collection[terms_of_service]=required` to the payment-mode Checkout Session behind the payment-request payment URL, so payers must accept the terms of service before paying. No change when the setting is off.


osmarluz (approved):
🚀

PR #6017, opened 2026-07-23
[ING-437] feat(connections): set connection default via GraphQL

mariohd:
## Context

  After a customer's connections are created, operators need to choose which one is the default for its category. This adds that ability over GraphQL for both integration and payment connections, mirroring the
  existing set-payment-method-as-default flow. The REST/API surface is intentionally left out and will be tackled in a follow-up.

  ## Changes

  Add `setIntegrationCustomerAsDefault` and `setPaymentProviderCustomerAsDefault` GraphQL mutations, each resolving the connection from a customer id and the connection code, guarded by the `customers:update`
  permission.
  
  Add the services behind them — `IntegrationCustomers::SetAsDefaultService` and `PaymentProviderCustomers::SetAsDefaultService` — which flip `is_default` on the target and clear it on the siblings, under a
  per-customer advisory lock. Integration clears only the connections of the same category (one default per category); payment clears all of the customer's connections (one default per customer). Only
  `is_default` is written.
  
  Expose `is_default` on the connection GraphQL types: add an `IntegrationCustomer` union over the existing per-provider integration types with `is_default` on each, and add `is_default` to the provider
  customer type. Register two new customer lock scopes and regenerate the GraphQL schema dump.

  Payment only handles provider connections here; the reserved `manual` connection and the connected/manual/not_connected status are deferred to a later change.

  ## Testing

  Service and mutation specs for both integration and payment, plus the schema spec. 23 examples, 0 failures.


app/graphql/mutations/integration_customers/set_as_default.rb
```diff
@@ -0,0 +1,29 @@
...
+      REQUIRED_PERMISSION = "customers:update"
+
+      argument :code, String, required: true
+      argument :customer_id, ID, required: true
+
+      type Types::IntegrationCustomers::Object
+
+      def resolve(**args)
+        customer = current_organization.customers.find_by(id: args[:customer_id])
+        integration_customer = customer&.integration_customers&.find_by(code: args[:code])
```
lovrocolic: code is unique per category, so there could be multiple records with the same code that falls into different categories

app/graphql/mutations/payment_provider_customers/set_as_default.rb
```diff
@@ -0,0 +1,29 @@
...
+      REQUIRED_PERMISSION = "customers:update"
+
+      argument :code, String, required: true
+      argument :customer_id, ID, required: true
+
+      type Types::PaymentProviderCustomers::Provider
+
+      def resolve(**args)
+        customer = current_organization.customers.find_by(id: args[:customer_id])
+        payment_provider_customer = customer&.payment_provider_customers&.find_by(code: args[:code])
```
lovrocolic: same as above

app/graphql/types/integration_customers/connection_category_enum.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  module IntegrationCustomers
+    class ConnectionCategoryEnum < Types::BaseEnum
```
vincent-pochet, line 5: Can you give it a name, for example:
```ruby
 graphql_name "IntegrationConnectionCategoryEnum"
```
Otherwise, it will be a generic one, hard to guess for the front

vincent-pochet (approved):
One future improvement, not directly related to this PR: All integration customers types share almost the same fields, it could be great to extract the common ones in a base type and inherit from it to define specific fields only

lovrocolic:
> One future improvement, not directly related to this PR: All integration customers types share almost the same fields, it could be great to extract the common ones in a base type and inherit from it to define specific fields only

Sounds good! I will do it in dedicated PR. Here I am using the same approach - https://github.com/getlago/lago-api/pull/6041 so dedicated PR will handle both PRs

app/graphql/types/integration_customers/connection_category_enum.rb
lovrocolic: Fixed

lovrocolic (approved):

PR #6018, opened 2026-07-24
feat(multi-connections): add support for destroy integration customer mutation

lovrocolic:
## Context
Removing an integration customer connection was only possible as a side-effect of a customer update. As part of multi-connections, we need a dedicated action to delete a specific connection.

## Description
Add a `destroyIntegrationCustomer` GraphQL mutation backed by `IntegrationCustomers::DestroyService`. Hard delete is kept for this first version, consistent with the current behavior.


app/services/integration_customers/destroy_service.rb
```diff
@@ -0,0 +1,26 @@
...
+    def initialize(integration_customer:)
+      @integration_customer = integration_customer
+
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "integration_customer") unless integration_customer
+
+      integration_customer.destroy!
```
D1353L, line 16: Is it expected to have a 500 error if integration_customer has linked invoice_connection? In this case `destroy!` raises `ActiveRecord::InvalidForeignKey`

lovrocolic: Hey Nikita! Thanks for the review! You are right here, however `invoice_connection` is new table and this mutation is also something that it's not going to be used in production until we finalise the whole feature.

The goal is to ship this mutation (and all others) ASAP so that FE team can work on top of it. We will most likely introduce soft-deletion for this resource since it can also happen on few other places - `CreateOrUpdateBatchService` and `Integrations::DestroyService`. We already defined this task but we will handle it a bit later.

D1353L: Nice! Looks good to me 👍

D1353L (approved):

mariohd (approved):

PR #6028, opened 2026-07-27
feat(multi-connection): Add GQL mutation for payment provider customer edition

lovrocolic:
## Context

Currently customers can be attached to only one connection of the same type. In order to change it, Lago is adding possibility to have multiple connections of the same type per customer.

## Description

This PR adds mutation to update payment provider customer. Currently it is possible to do it along with customer edition. This is the dedicated mutation, created only for this purpose.

### DIFF
--- a/app/services/payment_provider_customers/update_connection_service.rb
+++ b/app/services/payment_provider_customers/update_connection_service.rb
@@ -64,7 +64,7 @@ module PaymentProviderCustomers
         customer:,
         payment_provider_id: payment_provider_customer&.payment_provider&.id,
         params: provider_customer_params
-      ).call.raise_if_error!
+      ).call!
 
       customer.reload
     end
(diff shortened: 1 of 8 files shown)


app/services/payment_provider_customers/update_connection_service.rb
```diff
@@ -0,0 +1,72 @@
...
+      handle_provider_customer = provider_customer_params[:provider_customer_id].present?
+      handle_provider_customer ||= payment_provider_customer&.provider_customer_id.present?
+      return unless handle_provider_customer
+
+      PaymentProviders::CreateCustomerFactory.new_instance(
+        provider:,
+        customer:,
+        payment_provider_id: payment_provider_customer&.payment_provider&.id,
+        params: provider_customer_params
+      ).call.raise_if_error!
```
vincent-pochet:
```suggestion
      ).call!
```

vincent-pochet (approved):

PR #6029, opened 2026-07-27
fix(rds-proxy): route >16 KB queries around RDS Proxy via :direct role

IxDay:
## Summary

Route every known >16 KB query away from the RDS Proxy by wrapping the emitting call sites in `ApplicationRecord.connected_to(role: :direct) { ... }`. Introduce a `:direct` role on `ApplicationRecord` that aliases a second connection pool always pointing at the physical database.

Fixes RDS Proxy session-pinning on prod-us-1 for the two query classes visible in `pg_stat_statements` today:

| Emitter | Length | Calls | Where |
|---|---|---|---|
| `Subscriptions::OrganizationBillingService#billable_subscriptions` (18-way UNION CTE + timezone COALESCE branches) | **28,595 B** | 220,375 | clock-worker (`Clock::SubscriptionsBillerJob` → `OrganizationBillingJob`) |
| `Invoice` list eager-load (5 payment-provider join aliases + `id IN ($, …)` page of 100) | **16,200–16,700 B** | ~140k combined across 7 shapes | `lago-api` invoice index endpoints |
| Same shape on `CreditNote` list (`items → fee → …` + 5 payment-provider aliases) | approaches 16 KB with big pages | fewer calls, same failure mode | `lago-api` credit-note index endpoints |

RDS Proxy for PostgreSQL pins any client session that runs a statement over its non-tunable 16 KB threshold. During prod-us-1 load we saw `pinned/borrowed ≈ 1:1` and 100% of proxy pin events attributed to _"SQL query which exceeded the 16384 byte limit"_ — this PR is the app-side half of the fix.

## Env-var contract

| Env vars set | `:writing` (default) | `:direct` | Effect |
|---|---|---|---|
| `DATABASE_URL` only | `DATABASE_URL` | `DATABASE_URL` | No pooler; single pool; `:direct` is a no-op |
| `DATABASE_URL` + `DATABASE_PROXY_URL` | `DATABASE_PROXY_URL` | `DATABASE_URL` | Pooler default; `:direct` bypasses it |

Reads more naturally than the previous shape (where `DATABASE_URL` was the proxy URL and a separate env held the direct connection) — the canonical DB is the DB; the pooler is an optional layer in front. Non-proxied deployments don't need to know the pattern exists.

## Call sites routed direct

1. **`app/services/subscriptions/organization_billing_service.rb`** — wrap the `Subscription.find_by_sql` inside `billable_subscriptions` in `connected_to(role: :direct)`. Only the 28 KB CTE hits `:direct`; the surrounding service continues on the default pool.
2. **`app/controllers/concerns/invoice_index.rb`** — wrap the eager-load + serializer render block. Every `.includes(...)` expansion + `CollectionSerializer` traversal executes on `:direct` for the duration of one page render.
3. **`app/controllers/concerns/credit_note_index.rb`** — same shape as `InvoiceIndex`, same wrapping.

The two concern-based fixes cover every controller that includes them: `Api::V1::InvoicesController`, `Api::V1::Customers::InvoicesController`, `Api::V1::CreditNotesController`, `Api::V1::Customers::CreditNotesController`.

## Deploy sequencing

Needs a matching infra change in `lago-infrastructure`:
1. Every Rails workload has `DATABASE_URL` pointing at the direct RDS endpoint (today it's the proxy).
2. Workloads that should use the proxy also set `DATABASE_PROXY_URL` pointing at the RDS Proxy.

Safe to merge/deploy this PR first — with `DATABASE_PROXY_URL` unset, both roles resolve to `DATABASE_URL` and behaviour is identical to today. The infra layer already has the direct URL sealed as `database.direct.uri` from the earlier lago-infrastructure PR — the follow-up rewires the two secret keys under the new contract.

## Test plan

- [ ] `spec/services/subscriptions/organization_billing_service_spec.rb` passes locally
- [ ] `spec/models/application_record_spec.rb` passes locally (both roles resolve)
- [ ] Local boot without `DATABASE_PROXY_URL` → `:direct` and `:writing` resolve to the same DB, biller returns expected subscriptions
- [ ] Local boot with `DATABASE_PROXY_URL` set → normal traffic hits the pooler, biller + invoice/credit-note index endpoints hit `:direct`
- [ ] After the infra PR lands: on prod-us-1, `DatabaseConnectionsCurrentlySessionPinned` on `lago-production-rails-proxy` drops toward zero
- [ ] RDS Proxy pin log stops reporting the "16384 byte limit" reason for both query classes

### DIFF
--- a/app/controllers/concerns/credit_note_index.rb
+++ b/app/controllers/concerns/credit_note_index.rb
@@ -57,15 +57,23 @@ module CreditNoteIndex
     )
 
     if result.success?
-      render(
-        json: ::CollectionSerializer.new(
-          result.credit_notes,
-          ::V1::CreditNoteSerializer,
-          collection_name: "credit_notes",
-          meta: pagination_metadata(result.credit_notes),
-          includes: [:items, :applied_taxes, :error_details, {customer: [:integration_customers]}]
+      # Same shape as InvoiceIndex — five payment-provider join aliases on
+      # `customer` plus a deep `items → fee → …` chain push the rendered
+      # `.includes(...)` query past RDS Proxy's 16 KB per-statement pin
+      # threshold once a page of credit-note IDs is expanded into the
+      # `WHERE id IN (…)` clause. Route the query + serialization through
+      # `:direct` so the whole page render bypasses the pooler.
+      ApplicationRecord.connected_to(role: :direct) do
+        render(
+          json: ::CollectionSerializer.new(
+            result.credit_notes,
+            ::V1::CreditNoteSerializer,
+            collection_name: "credit_notes",
+            meta: pagination_metadata(result.credit_notes),
+            includes: [:items, :applied_taxes, :error_details, {customer: [:integration_customers]}]
+          )
         )
-      )
+      end
     else
       render_error_response(result)
     end
--- a/app/controllers/concerns/credit_note_index.rb
+++ b/app/controllers/concerns/credit_note_index.rb
@@ -57,12 +57,7 @@ module CreditNoteIndex
     )
 
     if result.success?
-      # Same shape as InvoiceIndex — five payment-provider join aliases on
-      # `customer` plus a deep `items → fee → …` chain push the rendered
-      # `.includes(...)` query past RDS Proxy's 16 KB per-statement pin
-      # threshold once a page of credit-note IDs is expanded into the
-      # `WHERE id IN (…)` clause. Route the query + serialization through
-      # `:direct` so the whole page render bypasses the pooler.
+      # Eager-load exceeds RDS Proxy's 16 KB pin threshold — bypass the pooler.
       ApplicationRecord.connected_to(role: :direct) do
         render(
           json: ::CollectionSerializer.new(
--- a/app/controllers/concerns/invoice_index.rb
+++ b/app/controllers/concerns/invoice_index.rb
@@ -61,39 +61,48 @@ module InvoiceIndex
     )
 
     if result.success?
-      invoices = Invoice.preload_offset_amounts(
-        result.invoices.includes(
-          :metadata,
-          :applied_taxes,
-          :billing_entity,
-          :applied_usage_thresholds,
-          customer: [
-            :billing_entity,
+      # The `.includes(...)` chain here (5 payment-provider join aliases plus
+      # billing_entity, metadata, integration_customers, applied_taxes,
+      # applied_usage_thresholds) renders to ~16.2–16.7 KB once ActiveRecord
+      # expands the `WHERE invoices.id IN ($, $, …)` clause against a page of
+      # 100 invoice IDs — crossing RDS Proxy's 16 KB per-statement pin
+      # threshold. Route the eager-load + serializer traversal through the
+      # `:direct` role so this whole page render bypasses the pooler.
+      ApplicationRecord.connected_to(role: :direct) do
+        invoices = Invoice.preload_offset_amounts(
+          result.invoices.includes(
             :metadata,
-            :stripe_customer,
-            :gocardless_customer,
-            :cashfree_customer,
-            :adyen_customer,
-            :moneyhash_customer,
-            {integration_customers: :integration}
-          ]
+            :applied_taxes,
+            :billing_entity,
+            :applied_usage_thresholds,
+            customer: [
+              :billing_entity,
+              :metadata,
+              :stripe_customer,
+              :gocardless_customer,
+              :cashfree_customer,
+              :adyen_customer,
+              :moneyhash_customer,
+              {integration_customers: :integration}
+            ]
+          )
         )
-      )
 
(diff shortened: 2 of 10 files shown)


aquinofb (approved):
LGTM

app/controllers/concerns/credit_note_index.rb
```diff
@@ -57,15 +57,23 @@ def credit_note_index(external_customer_id:)
...
-          ::V1::CreditNoteSerializer,
-          collection_name: "credit_notes",
-          meta: pagination_metadata(result.credit_notes),
-          includes: [:items, :applied_taxes, :error_details, {customer: [:integration_customers]}]
+      # Same shape as InvoiceIndex — five payment-provider join aliases on
+      # `customer` plus a deep `items → fee → …` chain push the rendered
+      # `.includes(...)` query past RDS Proxy's 16 KB per-statement pin
+      # threshold once a page of credit-note IDs is expanded into the
+      # `WHERE id IN (…)` clause. Route the query + serialization through
+      # `:direct` so the whole page render bypasses the pooler.
```
vincent-pochet: Can we get read of all these Claude comments? Or at least make them shorter (1 or 2 rows at most)?

app/controllers/concerns/invoice_index.rb
```diff
@@ -61,39 +61,48 @@ def invoice_index(customer_external_id: nil)
...
-          :applied_usage_thresholds,
-          customer: [
-            :billing_entity,
+      # The `.includes(...)` chain here (5 payment-provider join aliases plus
+      # billing_entity, metadata, integration_customers, applied_taxes,
+      # applied_usage_thresholds) renders to ~16.2–16.7 KB once ActiveRecord
+      # expands the `WHERE invoices.id IN ($, $, …)` clause against a page of
+      # 100 invoice IDs — crossing RDS Proxy's 16 KB per-statement pin
+      # threshold. Route the eager-load + serializer traversal through the
+      # `:direct` role so this whole page render bypasses the pooler.
```
vincent-pochet: Same comment regarding the comment

config/database.yml
```diff
@@ -79,16 +112,25 @@ staging:
...
+    # RDS Proxy for Postgres session-pins on prepared statements, so
+    # this must be off whenever the primary role goes through a pooler
+    # — derived from DATABASE_PROXY_URL presence (single source of truth).
+    prepared_statements: <%= ENV['DATABASE_PROXY_URL'].blank? %>
+    schema_search_path: <%= ENV.fetch('POSTGRES_SCHEMA', 'public') %>
+  direct:
     <<: *default
     url: <%= ENV['DATABASE_URL'] %>
     pool: <%= ENV.fetch('DATABASE_POOL', 10) %>
-    prepared_statements: <%= ENV.fetch('DATABASE_PREPARED_STATEMENTS', true) %>
```
vincent-pochet, line 85: I fear this is a breaking change for OSS users relying on it :/

app/models/application_record.rb
```diff
@@ -3,6 +3,33 @@
...
+
+  # In test env, Rails opens a distinct connection pool for `:direct`
+  # even though the config points at the same DB — separate pools mean
+  # separate PG sessions and separate transactions, and DatabaseCleaner
+  # only wraps the default (:writing) pool. Fixtures written via the
+  # default pool wouldn't be visible when the app switches to `:direct`,
+  # which would break every scenario spec that exercises the biller.
+  # Short-circuit the role swap so `:direct` queries land on the
+  # currently-active pool.
+  if Rails.env.test?
```
vincent-pochet: I would suggest to move this kind of condition in the spec configuration part, monkey patching the main path for test env is not ideal :/

app/models/invoice.rb
```diff
@@ -208,22 +208,33 @@ def self.ransackable_associations(_ = nil)
...
   # caching the result on each instance to avoid N+1 queries during serialization.
+  #
+  # `invoices.map(&:id)` iterates the relation, which is what materialises any
+  # chained `.includes(...)` — including the 5-payment-provider eager-load that
+  # invoice / credit-note index endpoints add on top. That materialised query
+  # renders to ~16.2–16.7 KB, above RDS Proxy's 16 KB per-statement pin
+  # threshold. Route both the eager-load fire AND the CreditNote batch through
+  # the `:direct` role so every caller (InvoiceIndex, CreditNoteIndex, GraphQL
+  # invoices resolvers, data exports) bypasses the pooler here — no need for
+  # each call site to remember the wrap.
```
vincent-pochet: Same comment regarding the comment

vincent-pochet (approved):
It looks okay for now, but I must say that it feels very "hacky" and very prone to error. Spreading specific routing config all over the code will be hard to maintain on the long run. 
I have nothing better to propose for now, but I really think that we will have to find an other way to address this.

vincent-pochet (approved):

PR #6035, opened 2026-07-27
fix(usage): Count exact duplicate charges only once

D1353L:
## Context
Concurrent plan updates and the async parent-to-child charge cascade can leave a plan with several live copies of the same charge. Current usage summed every copy, which had an impact on ongoing usage and triggered an erroneous threshold wallet top-up.

## Description

Deduplicate charges by their full billing definition before computing current usage: exact copies are counted once, charges differing in any billing-relevant attribute are preserved.
Invoice generation is not affected - this needs to be fixed in a follow-up PR.

### DIFF
--- a/app/services/invoices/customer_usage_service.rb
+++ b/app/services/invoices/customer_usage_service.rb
@@ -129,27 +129,35 @@ module Invoices
     # NOTE: Concurrent plan updates can transiently leave a plan with several copies of the
     #       same charge (only code/parent_id differ), which would multiply usage.
     #       Count exact copies once; charges differing in any billing attribute are preserved.
+    #       Filters are only compared between charges identical in everything else, and
+    #       since copies require a repeated billable metric, plans without one — the
+    #       common case — skip all signature work.
     def deduplicated_charges
-      charges.to_a.uniq { |charge| charge_definition_signature(charge) }
+      all_charges = charges.to_a
+      candidates = all_charges.group_by(&:billable_metric_id).values.reject(&:one?)
+      return all_charges if candidates.empty?
+
+      duplicates = candidates.flat_map do |charges_of_metric|
+        charges_of_metric
+          .group_by { |charge| charge_scalar_signature(charge) }
+          .values
+          .reject(&:one?)
+          .flat_map { |copies| copies - copies.uniq { |charge| charge_filters_signature(charge) } }
+      end
+
+      all_charges - duplicates
     end
 
-    def charge_definition_signature(charge)
+    def charge_scalar_signature(charge)
       [
-        charge.billable_metric_id,
-        charge.charge_model,
-        charge.properties,
-        charge.pay_in_advance,
-        charge.prorated,
-        charge.invoiceable,
-        charge.regroup_paid_fees,
-        charge.min_amount_cents,
-        charge.invoice_display_name,
-        charge.amount_currency,
-        charge.accepts_target_wallet,
+        charge.attributes.except("id", "code", "parent_id", "created_at", "updated_at", "deleted_at"),
         charge.taxes.map(&:id).sort,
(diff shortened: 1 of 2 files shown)


app/services/invoices/customer_usage_service.rb
```diff
@@ -120,12 +120,38 @@ def organization
...
+        charge.prorated,
+        charge.invoiceable,
+        charge.regroup_paid_fees,
+        charge.min_amount_cents,
+        charge.invoice_display_name,
+        charge.amount_currency,
+        charge.accepts_target_wallet,
+        charge.taxes.map(&:id).sort,
+        charge.applied_pricing_unit&.slice(:pricing_unit_id, :conversion_rate),
+        charge.filters.map { |f| [f.properties, f.values.map { |v| [v.billable_metric_filter_id, v.values] }.sort_by(&:to_s)] }.sort_by(&:to_s)
```
vincent-pochet: Maybe this could be enough:
```suggestion
        charge.filters.map(&:to_h)
```

vincent-pochet (commented):
Looks okay to me, but I have two questions:

- From the PR description: " Invoice generation is not affected", not sure to understand what differs in this path. Can you explain the difference?
- With  this change, the weight of the deduplication is applied on the current usage computation witch is today (by far) the most critical path of the application. To be honnest, I'm not sure of the new load this will add, specially on subscriptions with a lot of filters. Do you think this dedup could be done differently, before reaching the current usage computation path?

app/services/invoices/customer_usage_service.rb
D1353L: _Improvements in the last commits:_
3 steps of detecting duplicates with early exits - this preserves good performance for most cases.
`Check if same billable metric -> Same everything else -> Same filters -> Duplicate found`

> From the PR description: " Invoice generation is not affected", not sure to understand what differs in this path. Can you explain the difference?

The invoicing part keeps the old behavior with possible duplicates during a short window. If billing runs during that window, the invoice would include both fees.
The current PR fixes only the specific bug with customer usage. The root-cause fix for that path is preventing duplicates at write time, planned as a follow-up PR.

> With this change, the weight of the deduplication is applied on the current usage computation witch is today (by far) the most critical path of the application.

As parsing filters is the only potentially heavy part, added a comparison only between charges that are identical in everything else. If a duplicate is found, the dedup removes the duplicate fee computation - no performance impact in this case. The only possible performance impact is when 2 charges are equal in every field except filters.
Also filters are preloaded, meaning _no_ additional DB query is added.

> Do you think this dedup could be done differently, before reaching the current usage computation path?

Yes - fixing the root cause would be in the write path to avoid duplicates at all. This requires locking a plan, adding an additional uniqueness index, adding reconciliation as updates/deletes are async. This has wider scope and risk with current production data. Further investigation and a follow-up PR are needed.

> Maybe this could be enough: `charge.filters.map(&:to_h)`

Replaced with `charge.filters.map { |f| [f.properties, f.invoice_display_name, f.to_h.sort] }.sort_by(&:to_s)`

mariohd (approved):

D1353L:
This change breaks a valid scenario with 2 charges added intentionally:

1. Create a plan with 2 usage-metered charges with the same billable metric and amount.
2. Create a subscription with this plan.
3. Send 3 events using the subscription external_id and billable metric code.
4. Observe usage
**Expected**: usage is doubled (6 in total), and 2 rows are displayed (3 events per charge).
**Actual**: only one charge is displayed, and the total usage is 3.
5. Terminate subscription and check its invoice
**Expected** == **Actual**: 2 fees are displayed in the invoice with 3 units per fee.

The "fix" breaks usage display when a customer intentionally configured a plan with duplicate charges.

PR #6052, opened 2026-07-30
fix(subscriptions): Keep gated subs alive during 3DS

ancorcruz:
## Context

A payment-gated subscription is activated by its gating invoice being paid. When the activation charge lands on a card that requires 3D Secure, Stripe rejects the off-session attempt with `authentication_required` and Lago retries the charge on-session so the customer can authenticate.

That retry never got a chance to run. The rejection was recorded as a failed invoice payment before the retry was scheduled, and payment gating reads a failed payment as the terminal outcome of activation: the rule was rejected, the invoice closed and the subscription canceled with `payment_failed`. By the time the on-session retry ran, the subscription was already gone, so even a completed 3DS challenge could not activate it. The same happened a second way, through the `payment_intent.payment_failed` webhook Stripe emits for the rejected intent, which normally arrives before the retry is due.

For an ordinary invoice a failed payment status is transient bookkeeping that the retry overwrites, which is why this only broke gated subscriptions.

## Description

A failure that will be followed by another payment attempt no longer settles the invoice payment status: it stays pending, so gating leaves the activation rule untouched and the subscription waits for the customer to authenticate. Both the inline payment path and the Stripe webhook path respect this, since either can observe the rejection first.

Whether a rejection is followed by a retry is asked of the payment provider, which is where the knowledge of provider error codes and 3DS support belongs. Without 3DS support no retry follows, so the rejection stays terminal and the subscription is canceled as before.

### DIFF
--- a/app/models/payment_providers/base_provider.rb
+++ b/app/models/payment_providers/base_provider.rb
@@ -36,6 +36,10 @@ module PaymentProviders
 
       payment_status
     end
+
+    def retriable_authentication_failure?(_error_code)
+      false
+    end
   end
 end
 
--- a/app/models/payment_providers/stripe_provider.rb
+++ b/app/models/payment_providers/stripe_provider.rb
@@ -42,6 +42,10 @@ module PaymentProviders
     def payment_type
       "stripe"
     end
+
+    def retriable_authentication_failure?(error_code)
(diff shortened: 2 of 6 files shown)


osmarluz (approved):
🚀

PR #6067, opened 2026-08-03
[ING-536] fix(filters): Cascade charge filter changes to overridden plans

annvelents:
## Context

Removing a billable metric filter value leaves the dependent charge filters alive
on a shortened predicate. Deleting those filters, or editing the plan, did not
reach the charge filters on overridden plans — the parent was cleaned and the
override kept billing at the old price.

Reported on ING-536. What to do about the collapsed filters themselves  is not part of this PR.

## Description

- `CascadeDispatcher` groups by predicate and consumes one `before` filter per
  `after` one, so filters sharing a predicate are no longer dropped from the diff
- `CascadeService` reconciles every child filter on the predicate, not just one
- `DestroyService` broadcasts the live predicate: including already-discarded
  values described a predicate no child could match, so nothing cascaded

### DIFF
--- a/app/services/charge_filters/cascade_dispatcher.rb
+++ b/app/services/charge_filters/cascade_dispatcher.rb
@@ -8,36 +8,58 @@ module ChargeFilters
     # per change. `before` and `after` are arrays of:
     #   { values: {"key" => [...]}, properties: {...} | nil, invoice_display_name: "..." }
     # `values` and `properties` must be string-keyed.
+    #
+    # `values` is the filter's predicate, and it is what a job targets. Filters left on
+    # the same predicate by an older billable metric change are cascaded as one group.
     def call(charge:, before:, after:)
-      before_by_values = before.index_by { |f| f[:values] }
+      before_by_predicate = before.group_by { |filter| filter[:values] }
 
       after.each do |new_filter|
-        existing = before_by_values.delete(new_filter[:values])
+        previous_filters = before_by_predicate.delete(new_filter[:values]) || []
+        previous = previous_filters.first
 
-        next if existing &&
-          existing[:properties] == new_filter[:properties] &&
-          existing[:invoice_display_name] == new_filter[:invoice_display_name]
+        # Several filters on one predicate still need a job, even when the kept one is
+        # unchanged: the cascade is what discards the duplicates on the override
+        next if previous_filters.one? && unchanged?(previous, new_filter)
 
-        ChargeFilters::CascadeJob.perform_later(
-          charge.id,
-          existing ? "update" : "create",
+        enqueue(
+          charge,
+          previous ? "update" : "create",
           new_filter[:values],
-          existing&.dig(:properties),
+          previous&.dig(:properties),
           new_filter[:properties],
           new_filter[:invoice_display_name]
         )
       end
 
-      before_by_values.each_value do |old|
-        ChargeFilters::CascadeJob.perform_later(
-          charge.id,
+      # Predicates no filter in `after` claimed are the removed ones. A job targets a
+      # predicate and the cascade discards every child filter on it, so one job covers
+      # the group and any of its filters can carry the payload.
+      before_by_predicate.each_value do |removed_filters|
+        removed = removed_filters.first
+
+        enqueue(
+          charge,
           "destroy",
-          old[:values],
-          old[:properties],
+          removed[:values],
+          removed[:properties],
           nil,
-          old[:invoice_display_name]
+          removed[:invoice_display_name]
         )
       end
+    end
+
+    def unchanged?(previous, new_filter)
+      return false if previous.nil?
+
+      previous[:properties] == new_filter[:properties] &&
+        previous[:invoice_display_name] == new_filter[:invoice_display_name]
+    end
+
+    def enqueue(charge, action, values, old_properties, new_properties, invoice_display_name)
+      ChargeFilters::CascadeJob.perform_later(
+        charge.id, action, values, old_properties, new_properties, invoice_display_name
+      )
     end
   end
 end
--- a/spec/scenarios/plans/cascade_filters_over_metric_edits_spec.rb
+++ b/spec/scenarios/plans/cascade_filters_over_metric_edits_spec.rb
@@ -1,0 +1,399 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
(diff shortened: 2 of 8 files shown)


app/services/charge_filters/cascade_dispatcher.rb
```diff
@@ -8,36 +8,53 @@ module CascadeDispatcher
...
+    #
+    # `values` is the filter's predicate, and it is what a job targets. Filters left on
+    # the same predicate by an older billable metric change are cascaded as one group.
     def call(charge:, before:, after:)
-      before_by_values = before.index_by { |f| f[:values] }
+      before_by_predicate = before.group_by { |filter| filter[:values] }
 
       after.each do |new_filter|
-        existing = before_by_values.delete(new_filter[:values])
+        previous = before_by_predicate.delete(new_filter[:values])&.first
```
annvelents: will it make sense to have here
`previous_values = before_by_predicate.delete(new_filter[:values])`
and then 
`next if unchanged?(previous_values.first, new_filter) && previous_values.length == 1`

spec/scenarios/plans/create_and_update_filters_spec.rb
```diff
@@ -89,11 +89,11 @@
     end
 
     charge.reload
-    f1 = charge.filters.find_by(invoice_display_name: "f1")
-    expect(f1.to_h.keys).to match_array(%w[steps model_name])
-
-    f2 = charge.filters.find_by(invoice_display_name: "f2")
-    expect(f2.to_h.keys).to eq(%w[steps])
+    # f1 and f2 filtered on image_size=512x152, a value the metric no longer allows.
```
annvelents: I am not sure it's correct
1) 512x512 is still allowed, no?
2) f1 and f2 are still different because f2 is for llama-2,
so we should keep both still

spec/services/billable_metric_filters/create_or_update_batch_service_spec.rb
```diff
@@ -249,12 +249,14 @@
...
+        # Keeping it would drop the region condition and let the filter match every
+        # cloud=aws event at a price meant for US only
+        it "discards the charge_filter along with the values it still held" do
           service
 
           expect(shared_cfv_region.reload).to be_discarded
-          expect(shared_cfv_cloud.reload).not_to be_discarded
-          expect(shared_charge_filter.reload).not_to be_discarded
+          expect(shared_cfv_cloud.reload).to be_discarded
+          expect(shared_charge_filter.reload).to be_discarded
```
annvelents: let's not delete the charge_filter automatically on our side - it should be a decision of the customer

spec/services/charge_filters/destroy_service_spec.rb
```diff
@@ -76,6 +76,29 @@
...
+      context "when one of the values was already discarded" do
+        let(:tier_filter) do
+          create(:billable_metric_filter, billable_metric: charge.billable_metric, key: "tier", values: %w[gold])
+        end
+
+        before do
+          create(:charge_filter_value, charge_filter:, billable_metric_filter: tier_filter, values: ["gold"]).discard!
+        end
+
+        it "cascades the predicate the children still carry" do
```
annvelents: does it?
the discarded predicate is `tier: gold`, but cascaded is `{"card_location" => ["domestic"]`

app/services/charge_filters/cascade_dispatcher.rb
brunomiguelpinto: Covered every shared-predicate combination in cascade_dispatcher_spec.rb — all four fail without the change:

when one is kept unchanged → one update job
when one is kept and changed → one update job with the new properties
when none is kept → one destroy job
when both are kept → still cascades, so the override converges on one filter

app/services/charge_filters/cascade_dispatcher.rb
```diff
@@ -8,36 +8,58 @@ module CascadeDispatcher
...
+        # Several filters on one predicate still need a job, even when the kept one is
+        # unchanged: the cascade is what discards the duplicates on the override
+        next if previous_filters.one? && unchanged?(previous, new_filter)
 
-        ChargeFilters::CascadeJob.perform_later(
-          charge.id,
-          existing ? "update" : "create",
+        enqueue(
+          charge,
+          previous ? "update" : "create",
```
annvelents, line 27: why do we check only one and not go through all the previous_filters?

spec/services/charge_filters/cascade_dispatcher_spec.rb
```diff
@@ -88,6 +88,165 @@
...
+
+      let(:after) do
+        [
+          {values: us_values, properties: {"amount" => "15"}, invoice_display_name: "US"},
+          {values: eu_values, properties: {"amount" => "20"}, invoice_display_name: "EU"},
+          {values: ca_values, properties: {"amount" => "40"}, invoice_display_name: "CA"}
+        ]
+      end
+
+      it "enqueues one job per predicate, none dropped" do
```
annvelents: ?

spec/services/charge_filters/cascade_dispatcher_spec.rb
```diff
@@ -88,6 +88,165 @@
...
+        it "enqueues no destroy job" do
+          expect {
+            described_class.call(charge:, before:, after:)
+          }.not_to have_enqueued_job(ChargeFilters::CascadeJob).with(
+            charge.id, "destroy", any_args
+          )
+        end
+      end
+
+      context "when both are kept" do
```
annvelents, line 170: ?

annvelents (commented):
Let's add scenario tests with everything that might happen 
(10 filter values, some of them are removed, removed again, filters that are removed, changed, left, anything else that might happen)

spec/scenarios/plans/cascade_filters_over_metric_edits_spec.rb
```diff
@@ -0,0 +1,399 @@
...
+  # the parent and the override after each one.
+  it "keeps the override aligned with the plan across successive metric and plan edits" do
+    ctx = setup_plan_with_override
+    parent_charge = ctx[:parent_charge]
+    child_charge = ctx[:child_charge]
+
+    # STEP 0 — the override has just deep-copied the plan's filters, prices included.
+    # Parent and child both hold:
+    #
+    #   {type: [pages], model: [m1]}  @1
```
annvelents, line 117: nice! 😍  very understandable

PR #6069, opened 2026-08-03
feat/quotes/subscription creation validator

toommz:
## Context

`QuoteVersions::Validators.for` only knew about `one_off`, so it returned `nil` for
`subscription_creation` and both `UpdateService` and `ApproveService` let those payloads
through untouched. Anything the front end sent was accepted, and the first thing to look at
the payload was the execution flow, well after the quote had been approved and signed.

That is the wrong place to find out. A signed order that cannot execute rolls the whole
thing back and lands as failed, with no editable draft left to fix. So the validation stack
has to reject, at approval time, anything the downstream services would refuse:
`Subscriptions::CreateService`, `Commitments`/`Charges`/`FixedCharges::OverrideService`,
`AppliedCoupons::CreateService` and `Wallets::CreateService`.

## Description

**Shared scaffolding.** `OneOffValidator` and `OneOff::StructuralValidator` carried logic
that is not order-type specific: the structural-then-business ordering, `billing_items`
normalisation, and the JSON schema error to `billing_items.<pointer> => [error_code]`
flattening the API surfaces. Both move up into `BaseOrderTypeValidator` and
`BaseStructuralValidator`, leaving subclasses to name their two passes and their schemer.
The currency check moves into a `CurrencyValidation` concern that pulls `Currencies` in from
its own `included` hook, since `currency_list` is only ever defined on an includer and an
includer that forgot the line would lose the check to a `NoMethodError`.

**`subscription_creation` schema.** Two scopes over one literal document: `UPDATE_DEFINITION`
stays lenient so a draft can be saved half-filled, `APPROVE_DEFINITION` is derived from it and
narrows what a signable payload must carry. Keys are camelCase to match the FE payload,
unknown keys are rejected with `unsupported_key`, and the coupon type, coupon frequency and
recurring transaction rule trigger/interval/method enums are interpolated from `Coupon` and
`RecurringTransactionRule` rather than hand-copied, so adding an interval upstream does not
silently get rejected here.

Approve scope adds: at least one plan, `payload.code` on plans, `code` and `type` on coupons,
`amountCents` on a minimum commitment (a commitment carrying only a display name cannot be
created, so "no commitment" now means sending `null`), `paidCredits`/`grantedCredits`/
`rateAmount` on wallet credits, and a non-null `trigger` on each recurring rule. Commitment
and usage threshold amounts reject zero, matching the model validations.

**Business validator.** Everything the schema cannot express, one DB round trip per resource
kind:

- currency mandatory at approve, ISO 4217 when set
- start/end date mandatory at approve, end date strictly after start (`Subscriptions::ValidateService`
  requires it, and the quote dates are what a plan without its own falls back to)
- plan existence, plan currency against the deal currency, charge override `billableMetricCode`
  and fixed charge override `addOnCode` existence, fixed charge `units` against the same
  non-negative bound `FixedCharge` enforces
- coupon existence, fixed-amount coupon currency, snapshot `type` and matching amount field at
  approve (a coupon retyped since drafting changes the negotiated discount and should be
  flagged, not followed), and a duration on any coupon that ends up recurring. The effective
  frequency resolves through overrides, then payload, then the live coupon, mirroring how
  `AppliedCoupons::CreateService` falls back, so this cannot live in the schema
- wallet credit amounts, currency against the deal (`Wallets::CreateService` forces the customer
  currency unless `multi_currency` is on, so a mismatched credit is either dropped or funds a
  wallet nobody agreed to), expiration futureness, and per-rule `interval`/`thresholdCredits`/
  `targetOngoingBalance`/`grantsTargetTopUp`/credits, sign conventions matching
  `Wallets::RecurringTransactionRules::ValidateService`

Lookups use `with_discarded` so a soft-deleted plan or coupon quoted before deletion still
resolves.

### Dependencies

Stacked under #6075 (`feat(orders): execute subscription_creation orders`), which consumes the
payload this validates.

### DIFF
--- a/app/services/quote_versions/validators.rb
+++ b/app/services/quote_versions/validators.rb
@@ -2,6 +2,9 @@
 
 module QuoteVersions
   module Validators
+    # NOTE: subscription_amendment has no validator yet, so it returns nil and the callers
+    # (ApproveService, UpdateService) let it through. Orders::ExecuteService refuses that order
+    # type anyway, until the amendment validator lands.
     def self.for(result, quote_version:, scope:)
       case quote_version.quote.order_type
       when Quote::ORDER_TYPES[:one_off]
--- a/app/services/quote_versions/validators.rb
+++ b/app/services/quote_versions/validators.rb
@@ -6,6 +6,8 @@ module QuoteVersions
       case quote_version.quote.order_type
       when Quote::ORDER_TYPES[:one_off]
         OneOffValidator.new(result, quote_version:, scope:)
+      when Quote::ORDER_TYPES[:subscription_creation]
+        SubscriptionCreationValidator.new(result, quote_version:, scope:)
       end
     end
   end
--- a/app/services/quote_versions/validators/base_order_type_validator.rb
+++ b/app/services/quote_versions/validators/base_order_type_validator.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module QuoteVersions
+  module Validators
+    # Runs the structural pass before the business one so DB lookups only ever see a payload
+    # of the expected shape. Subclasses name the two passes for their order type.
+    class BaseOrderTypeValidator < ::BaseValidator
+      def initialize(result, quote_version:, scope:)
+        @quote_version = quote_version
+        @scope = scope
+
+        super
+      end
+
+      def valid?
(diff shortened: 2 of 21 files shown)


mariohd (approved):
Looks great

PR #6070, opened 2026-08-04
fix(payment-methods): Handle scenario where associated customer is discarded

osmarluz:
## Context

When a customer is deleted in Lago, the collection of its termination invoice can still run. If Stripe then reports the stored payment method as invalid, `PaymentProviderCustomers::Stripe::CheckPaymentMethodService` crashed with `NoMethodError: undefined method 'payment_methods' for nil` on `Invoices::Payments::CreateJob`: the rescue path resolves the payment method through `stripe_customer.customer`, which returns `nil` for discarded customers. This service was originally introduced in #3196 to rely only on the `stripe_customer` for this exact reason, and #4861 reintroduced the customer dependency.

## Description

Look up the stale `PaymentMethod` by the `customer_id` foreign key instead of going through the `customer` association. The lookup semantics are unchanged for kept customers, and for discarded ones the invalid payment method is now properly discarded and the payment fails with `value_is_invalid` instead of crash-looping the job.

### DIFF
--- a/app/services/payment_provider_customers/stripe/check_payment_method_service.rb
+++ b/app/services/payment_provider_customers/stripe/check_payment_method_service.rb
@@ -23,7 +23,10 @@ module PaymentProviderCustomers
         # NOTE: The payment method is no longer valid
         stripe_customer.update!(payment_method_id: nil)
 
-        payment_method = customer.payment_methods.find_by(provider_method_id: payment_method_id)
+        payment_method = PaymentMethod.find_by(
+          customer_id: stripe_customer.customer_id,
+          provider_method_id: payment_method_id
+        )
         PaymentMethods::DestroyService.call(payment_method:)
 
         result.single_validation_failure!(field: :payment_method_id, error_code: "value_is_invalid")
@@ -35,10 +38,6 @@ module PaymentProviderCustomers
 
       def api_key
         stripe_customer.payment_provider.secret_key
-      end
-
-      def customer
-        @customer ||= stripe_customer.customer
       end
     end
   end
--- a/app/services/payment_provider_customers/stripe/check_payment_method_service.rb
+++ b/app/services/payment_provider_customers/stripe/check_payment_method_service.rb
@@ -23,10 +23,7 @@ module PaymentProviderCustomers
         # NOTE: The payment method is no longer valid
         stripe_customer.update!(payment_method_id: nil)
 
-        payment_method = PaymentMethod.find_by(
-          customer_id: stripe_customer.customer_id,
-          provider_method_id: payment_method_id
-        )
+        payment_method = stripe_customer.payment_methods.find_by(provider_method_id: payment_method_id)
         PaymentMethods::DestroyService.call(payment_method:)
 
         result.single_validation_failure!(field: :payment_method_id, error_code: "value_is_invalid")
--- a/spec/services/payment_provider_customers/stripe/check_payment_method_service_spec.rb
+++ b/spec/services/payment_provider_customers/stripe/check_payment_method_service_spec.rb
@@ -73,6 +73,17 @@ RSpec.describe PaymentProviderCustomers::Stripe::CheckPaymentMethodService do
           expect(stripe_api_customer).to have_received(:retrieve_payment_method)
           expect(default_payment_method.reload.deleted_at).to be_present
         end
+
+        context "when customer is deleted" do
+          let(:customer) { create(:customer, :deleted, organization:) }
+
+          it "returns a failed result and discards payment method" do
+            result = check_service.call
+
+            expect(result).not_to be_success
+            expect(default_payment_method.reload.deleted_at).to be_present
+          end
+        end
       end
     end
 
--- a/spec/services/payment_provider_customers/stripe/check_payment_method_service_spec.rb
+++ b/spec/services/payment_provider_customers/stripe/check_payment_method_service_spec.rb
@@ -77,10 +77,15 @@ RSpec.describe PaymentProviderCustomers::Stripe::CheckPaymentMethodService do
         context "when customer is deleted" do
           let(:customer) { create(:customer, :deleted, organization:) }
 
+          before { stripe_customer.reload }
+
           it "returns a failed result and discards payment method" do
             result = check_service.call
 
             expect(result).not_to be_success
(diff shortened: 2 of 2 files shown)


app/services/payment_provider_customers/stripe/check_payment_method_service.rb
```diff
@@ -23,7 +23,10 @@ def call
         # NOTE: The payment method is no longer valid
         stripe_customer.update!(payment_method_id: nil)
 
-        payment_method = customer.payment_methods.find_by(provider_method_id: payment_method_id)
```
osmarluz, line 26: This was raising `NoMethodError: undefined method 'payment_methods' for nil` when the customer retrieved from `stripe_customer` was discarded

spec/services/payment_provider_customers/stripe/check_payment_method_service_spec.rb
```diff
@@ -73,6 +73,17 @@
...
           expect(default_payment_method.reload.deleted_at).to be_present
         end
+
+        context "when customer is deleted" do
+          let(:customer) { create(:customer, :deleted, organization:) }
+
+          it "returns a failed result and discards payment method" do
+            result = check_service.call
+
+            expect(result).not_to be_success
```
ancorcruz, line 87: shall we also check the failure `value_is_invalid` is returned as single validation failure? also the stripe_customer.payment_method_id was cleared? Currently it will pass on a not found failure.

I think this test passes even before the fix it is supposed to test.

The reason is that create(:stripe_customer, customer:) assigns the Customer object, so belongs_to :customer's target is already cached in memory — stripe_customer.customer returns the discarded record instead of `nil`, because no DB load ever happens and `default_scope { kept }` never applies. Adding `stripe_customer.reload` before the call makes it fail on main with `NoMethodError` at `check_payment_method_service.rb:26` and pass with the fix.

As written, this spec would not catch a revert of the PR. The pre-existing outer `context "when customer is deleted"` has the identical flaw.

Also, claude pointed me to the following finding...

**Wrong fix location, and the new lookup key is wrong**

The root cause is one missing association scope. I verified this directly:

```
app/models/payment_provider_customers/base_customer.rb:14:  belongs_to :customer          ← no scope
app/models/payment_method.rb:13:  belongs_to :customer, -> { with_discarded }
app/models/wallet.rb:8:            belongs_to :customer, -> { with_discarded }
app/models/payment.rb:10:          belongs_to :customer, -> { with_discarded }
app/models/invoice.rb:55:          belongs_to :customer, -> { with_discarded }
app/models/customer.rb:136:        default_scope -> { kept }

```
Every sibling model compensates for Customer's default_scope; BaseCustomer is the only one that doesn't. This PR patches one of seven call sites. The same nil-deref remains in set_as_default_service.rb:22, destroy_service.rb:39, setup_intent_succeeded_service.rb:81, moneyhash/payments/create_service.rb:88, and — silently, passing customer: nil so the payment method is never created — fetch_default_payment_method_service.rb:25. A third workaround for the same condition already exists next door in stripe/update_payment_method_service.rb:49 (Customer.unscoped.where(...)).

One association change fixes all of them. It would require rewriting the deleted_customer guard as customer.discarded?.

The replacement lookup is keyed on a non-unique pair. payment_methods has UNIQUE (payment_provider_customer_id, provider_method_id) WHERE deleted_at IS NULL — nothing enforces uniqueness on (customer_id, provider_method_id). A customer can hold methods under several provider customers (stripe + adyen + gocardless, or a discarded stripe connection whose methods weren't discarded). find_by has no ORDER BY, so Postgres may hand back the row belonging to a different provider customer: the wrong PaymentMethod gets discarded while the one on payment.payment_method stays active, and the next attempt fails identically.

BaseCustomer already declares has_many :payment_methods, foreign_key: :payment_provider_customer_id, so stripe_customer.payment_methods.find_by(provider_method_id:) matches the unique index, reuses the association, never touches customer, and drops the bare unscoped query.


Also pointed to **Pre-existing, but this fix makes it reachable**

These are on lines the PR didn't modify. Before the fix these paths crashed for discarded customers; now they run to completion, so latent behaviour becomes live. Your call whether they belong in this PR or a follow-up:

- The rescue treats every Stripe::InvalidRequestError as "payment method no longer valid." That class also covers No such customer: cus_x — quite likely for exactly the discarded-customer scenario targeted here — plus invalid params and wrong-account resources. Those now permanently clear payment_method_id and discard! a valid local PaymentMethod. Gating on e.code == "resource_missing" would narrow it.
- stripe_customer.update! is the first statement in the rescue and can itself raise. StripeCustomer validates provider_payment_methods presence with no backfill migration, so a legacy record with blank settings raises RecordInvalid inside the rescue — same infinite Invoices::Payments::CreateJob retry loop, different error class. The two writes also aren't in one transaction.
- PaymentMethods::DestroyService's result is dropped, so a nil lookup silently no-ops while the service still reports value_is_invalid.
- The customer_payment_methods.first fallback in payments/create_service.rb:87 is now reachable in more cases, which can charge a card other than one pinned via PaymentMethods::DetermineService.
- payments/create_service.rb:75 already holds the PaymentMethod object and passes only provider_method_id, forcing the re-query that creates the ambiguity above.

ancorcruz (approved):
the fix works for the one call site it touches, the test proving it is a no-op, and the actual bug is a missing `-> { with_discarded }` on `BaseCustomer` that leaves six other crash sites in place.

spec/services/payment_provider_customers/stripe/check_payment_method_service_spec.rb
osmarluz: @ancorcruz nice catches, thanks! I added the fix to tests and to the query to match the unique index as well. Regarding the suggestion to move the fix to `app/models/payment_provider_customers/base_customer.rb`, I'll keep the scope isolated to this service in particular, which we have concrete failures happening for now. Also, changing the behavior on the base class would cause change in behavior that we need to double check first. For example:

- `app/services/payment_providers/stripe/customers/fetch_default_payment_method_service.rb` would start creating payment methods for discarded customers
- `app/services/payment_providers/stripe/webhooks/setup_intent_succeeded_service.rb` would start processing setup intents for them.

I prefer exploring those separately not to increase the scope here

PR #6071, opened 2026-08-04
Chore(optimisation): rewrite BillableMetricFilters::RefreshDraftInvocies query to get rid of the joins

annvelents:
## Context

The job of refreshing invoices fails with timeout, having a separate query + subquery will help us not to join all the tables together

### DIFF
--- a/app/jobs/billable_metric_filters/refresh_draft_invoices_job.rb
+++ b/app/jobs/billable_metric_filters/refresh_draft_invoices_job.rb
@@ -7,12 +7,13 @@ module BillableMetricFilters
     def perform(billable_metric_id)
       billable_metric = BillableMetric.find_by(id: billable_metric_id)
       return unless billable_metric
+      plan_ids = billable_metric.plans.distinct.ids
+      return if plan_ids.empty?
 
       Invoice.draft
-        .joins(plans: [:billable_metrics])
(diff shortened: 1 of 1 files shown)


app/jobs/billable_metric_filters/refresh_draft_invoices_job.rb
```diff
@@ -7,12 +7,13 @@ class RefreshDraftInvoicesJob < ApplicationJob
...
+      plan_ids = billable_metric.plans.distinct.ids
+      return if plan_ids.empty?
 
       Invoice.draft
-        .joins(plans: [:billable_metrics])
-        .where(billable_metrics: {id: billable_metric.id})
-        .distinct
-        .in_batches do |batch|
+        .where(id: InvoiceSubscription
+                     .where(subscription_id: Subscription.where(plan_id: plan_ids).select(:id))
```
jdenquin, line 15: should we filter on active sub only?

annvelents: yes!

annvelents: actually, no
because if subscription is terminating / upgrading, we can generate an invoice that will be draft, when the subscription is already terminated

PR #6075, opened 2026-08-04
feat(orders): execute subscription_creation orders

toommz:
## Why

`Orders::ExecuteService` dispatched `one_off` only, so a `subscription_creation` quote could be drafted, approved, signed and turned into an order, then stopped with `order_type: ["unsupported_order_type"]`. This is the execution half of the spec whose validation half is #6069, mirroring how one_off shipped as #5860 then #5953.

Executing such an order means turning the frozen `billing_items` snapshot into real records: a subscription per `plans` entry with its negotiated overrides, an applied coupon per `coupons` entry, and a wallet per `walletCredits` entry.

## What

**Make the contract executable.** `Subscriptions::CreateService` needs fields the contract did not declare, `external_id` above all, which it does not generate outside a controller. The frontend already sends them inside the free-form plan `payload`, where nothing validated them, so `billingTime: "monthly"` or a stale `paymentMethodId` would only surface at execution and leave the order failed. They are now declared and checked at approve, along with the payment method belonging to the quote's customer, per-plan date coherence, wallet metric limitations resolving, and a single recurring rule per wallet since `Wallets::ValidateService` refuses more.

**Resolve charge overrides through the snapshot.** `billing_items` references charges by `billableMetricCode` while `Plans::OverrideService` keys on charge UUID, and the payload already carries both side by side. Resolving there pins the charge the approver was looking at rather than whatever the catalog holds at execution time. Approval now refuses a code with no snapshot entry, a code matching several, an id no longer on the plan, and a charge model that drifted, because `Plans::OverrideService` silently ignores an id it cannot find and `Charges::OverrideService` cannot switch models: either case would bill catalog pricing instead of the negotiated price. The executor re-checks the id against the live plan for the same reason.

**Add the executor.** `Orders::SubscriptionCreation::ExecuteService` creates subscriptions, then applied coupons, then wallets in one transaction under the quote's advisory lock, reusing the one_off executor's idempotency, `execution_mode` guard and durable failure trace. Subscriptions come first so the customer currency they settle is the one coupons and wallets reuse.

**Record what an execution produced.** `execution_record` gains `subscription_ids`, `applied_coupon_ids` and `wallet_ids`, exposed on `OrderExecutionRecord`. Nothing else linked an order to its output: `quotes.subscription_id` is only populated for amendments, and `invoice_id` stays null here because a subscription is invoiced later by `BillSubscriptionJob`, and only when the plan is paid in advance.

## Notes for the reviewer

- A charge or fixed charge discarded since the quote was approved now fails validation instead of passing. That is a behaviour change to the validator merged in #6069, and it is deliberate: the override would otherwise vanish silently.
- `ending_at` falls back to the quote's `end_date`, so every subscription created from a quote terminates at the contract end. If a quote's end date means the offer's validity rather than the contract term, this mapping should drop `ending_at` instead.
- Known gaps carried, not fixed: a discarded coupon or plan fails at execution rather than at approve, the customer currency mismatch still surfaces in `Customers::UpdateCurrencyService`, and an `subscriptionExternalId` colliding with a live subscription makes `Subscriptions::CreateService` reuse or upgrade it rather than create one.

## Stacked on

#6069, review and merge that first.

### DIFF
--- a/app/services/orders/subscription_creation/execute_service.rb
+++ b/app/services/orders/subscription_creation/execute_service.rb
@@ -135,7 +135,9 @@ module Orders
         payload = item["payload"] || {}
 
         {
-          external_id: payload["subscriptionExternalId"].presence || SecureRandom.uuid,
+          # localId is the quote line's own id, so a retry after a failed execution reuses the same
+          # external id instead of minting a new one.
+          external_id: payload["subscriptionExternalId"].presence || item["localId"].presence || SecureRandom.uuid,
           name: payload["subscriptionName"],
           billing_time: payload["billingTime"],
           subscription_at: subscription_datetime(payload["startDate"], quote_version.start_date),
--- a/app/services/orders/subscription_creation/execute_service.rb
+++ b/app/services/orders/subscription_creation/execute_service.rb
@@ -6,6 +6,8 @@ module Orders
     class ExecuteService < BaseService
       Result = BaseResult[:order]
 
+      CALENDAR_DATE = /\A\d{4}-\d{2}-\d{2}\z/
+
       def initialize(order:)
         @order = order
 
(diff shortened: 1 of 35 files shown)


app/services/orders/subscription_creation/execute_service.rb
```diff
@@ -0,0 +1,335 @@
...
+        date = calendar_date(value)
+        return value if date.nil?
+
+        date.in_time_zone(order.customer.applicable_timezone)
+      end
+
+      # A string that only looks like a date is left alone for Subscriptions::ValidateService to
+      # reject.
+      def calendar_date(value)
+        return value.to_date if value.is_a?(Date)
```
mariohd, line 84:
```suggestion
        return value if value.is_a?(Date)
```

mariohd (approved):
looks good, just a minor suggestion

PR #6079, opened 2026-08-05
feat(charge_filters): add code to identify filters

brunomiguelpinto:
## Context

A charge filter has no stable identity. It is known by its values, and those move: removing a value
from a billable metric trims it out of every filter using it, so two filters that were distinct can
collapse onto the same values and stop being distinguishable.

```
{region: [us], tier: [gold]}    @0.02     ->  {region: [us]}  @0.02
{region: [us], tier: [silver]}  @0.015    ->  {region: [us]}  @0.015
```

Propagating a plan's charge to its overrides pairs filters by those values, so once that happens it
can reprice the wrong one.

## Description

Charge filters get a `code`, derived from the values a filter is created with and frozen from then
on, so a filter keeps its identity even when its values move. An override inherits the code of the
filter it was copied from rather than deriving its own — that inheritance is what records the link
between the two.

A filter that exists **only** on the override, because the customer negotiated a segment the plan
does not price, derives its own code: there is nothing to inherit from, and it is not orphaned
either. If the plan later gains that predicate it derives the same code, so the two line up by
construction rather than by luck.

### Backfilling existing rows

One walk, over the charges belonging to a plan, a job per charge. Each job derives the codes for its
own filters and then hands its overrides on to a second job 

The unit of work is a charge on the way in and a parent on the way out, because every override of a
charge answers to the same set of codes.

### What stays NULL, on purpose

- **a charge holding two filters on one predicate** — whichever kept the code would be the one that
  bills from then on, and that is not a migration's call. Only the colliding filters are skipped;
  their siblings are still filled.
- **an override holding a filter its plan no longer has** — nothing left to link it to, either
  because the plan dropped it or because the parent charge was deleted and `dependent: :nullify` cut
  the link. Damage already done; no backfill undoes it.
- **a filter whose code a sibling already holds**, frozen when that sibling's own values were these —
  writing it again would break the unique index on `(charge_id, code)`.
- **organizations with nothing active or pending, and deleted plans** — not billing, not worth the work.

`rake filters:report_codes` reports what is left and why. Cleaning up the first kind and running the
upgrade task again converges, since a cleaned charge goes back to having filters without a code and
the walk selects it again.

### DIFF
--- a/app/jobs/database_migrations/backfill_charge_filter_codes_job.rb
+++ b/app/jobs/database_migrations/backfill_charge_filter_codes_job.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module DatabaseMigrations
+  class BackfillChargeFilterCodesJob < ApplicationJob
+    queue_as :low_priority
+    unique :until_executed, on_conflict: :log, lock_ttl: 6.hours
+
+    def perform(charge_id)
+      charge = Charge.kept.includes(:plan).find_by(id: charge_id)
+
+      return if charge.nil?
+      return unless charge.parent_id.nil? && charge.plan.parent_id.nil?
+
+      filters = charge.filters.includes(values: :billable_metric_filter).to_a
+      taken = filters.filter_map(&:code).to_set
+
+      # Grouped by the code the values derive rather than by the values, since the code is what can
+      # collide: it sorts, so `{region: [us, eu]}` and `{region: [eu, us]}` are the same
+      filters.group_by { ChargeFilter.generate_code(it.to_h) }.each do |code, group|
+        next unless group.one?       # skip duplicates
+        next if group.sole.code      # already have code
+        next if taken.include?(code) # already taken ( would generate duplicate )
+
+        group.sole.update_column(:code, code) # rubocop:disable Rails/SkipsModelValidations
+      end
+
+      return unless filters.any?(&:code)
+
+      BackfillChildChargeFilterCodesJob.perform_later(charge.id)
+    end
+
+    def lock_key_arguments
+      [arguments]
+    end
+  end
+end
--- a/app/jobs/database_migrations/backfill_child_charge_filter_codes_job.rb
+++ b/app/jobs/database_migrations/backfill_child_charge_filter_codes_job.rb
@@ -1,0 +1,100 @@
+# frozen_string_literal: true
+
+module DatabaseMigrations
+  class BackfillChildChargeFilterCodesJob < ApplicationJob
+    queue_as :low_priority
+    unique :until_executed, on_conflict: :log, lock_ttl: 6.hours
+
+    CHARGES_PER_BATCH = 1_000
+    IDS_PER_UPDATE = 10_000
+
+    # One row per charge filter. The fields are positional: they follow the `pluck` in `filters_of`,
+    Filter = Struct.new(:charge_id, :id, :code, :signature)
+
+    SIGNATURE_SQL = <<~SQL.squish
+      COALESCE(
+        jsonb_object_agg(
+          cfv.billable_metric_filter_id,
+          to_jsonb(ARRAY(SELECT unnest(cfv.values) ORDER BY 1))
+        ) FILTER (WHERE cfv.id IS NOT NULL),
+        '{}'::jsonb
+      )::text
+    SQL
+
+    def perform(parent_charge_id)
+      parent = Charge.kept.includes(:plan).find_by(id: parent_charge_id)
+      return if parent.nil?
+
+      return unless parent.parent_id.nil? && parent.plan.parent_id.nil?
+
+      to_copy = codes_to_copy(parent)
+      return if to_copy.empty?
+
+      parent.children.in_batches(of: CHARGES_PER_BATCH) do |children|
+        backfill(children, to_copy)
+      end
+    end
+
+    def lock_key_arguments
+      [arguments]
+    end
+
+    private
+
+    # The codes this parent can hand down (without duplicates)
+    def codes_to_copy(parent)
+      codes = {}
+
+      filters_of(parent.id).group_by(&:signature).each do |signature, filters|
+        next unless filters.one?
+        next if filters.sole.code.nil? # no code yet: nothing to hand down
+
+        codes[signature] = filters.sole.code
+      end
+
+      codes
+    end
+
+    # Walks what the parent has to hand down rather than the override's filters, so one the override
+    # has and the plan does not is never visited, and keeps no code.
+    def backfill(children, to_copy)
+      ids_by_code = Hash.new { |ids, code| ids[code] = [] }
+
+      filters_of(children).group_by(&:charge_id).each_value do |filters|
+        by_signature = filters.group_by(&:signature)
+        taken = filters.filter_map(&:code).to_set
+
+        to_copy.each do |signature, code|
+          next if taken.include?(code) # a filter here already holds it, from when its values were these
+
+          matching = by_signature[signature]
+          next unless matching&.one? # held twice here: neither can say which one is the copy
+          next if matching.sole.code # frozen at creation, and that is the one it keeps
+
+          ids_by_code[code] << matching.sole.id
+        end
+      end
+
+      write_codes(ids_by_code)
+    end
+
+    # One query for a whole batch of charges, filters and values and all
+    def filters_of(charge_scope)
+      ChargeFilter
+        .where(charge_id: charge_scope)
+        .joins("LEFT JOIN charge_filter_values cfv ON cfv.charge_filter_id = charge_filters.id AND cfv.deleted_at IS NULL")
+        .group("charge_filters.id")
+        .pluck(:charge_id, :id, :code, Arel.sql(SIGNATURE_SQL))
+        .map { Filter.new(*it) }
+    end
+
+    # One statement per code
+    def write_codes(ids_by_code)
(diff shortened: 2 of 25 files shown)


app/services/charge_filters/cascade_service.rb
```diff
@@ -120,6 +120,8 @@ def create_child_filter(child_charge, existing_filter)
             values:
           )
         end
+
+        child_filter.assign_code!
```
annvelents: chlid_filter should get the code from the parent, not generate it themselves

lib/tasks/filters.rake
```diff
@@ -23,4 +23,38 @@ namespace :filters do
...
+    processed = 0
+
+    scope = ChargeFilter.unscoped.where(code: nil)
+    scope = scope.where(organization_id: ENV["ORGANIZATION_ID"]) if ENV["ORGANIZATION_ID"].present?
+
+    # to_h_with_discarded, not to_h: discarded values are still on the row, so this rebuilds
+    # the values the filter was created with. That is what the code is meant to freeze, and it
+    # is what keeps filters that have since collapsed onto one predicate distinguishable.
+    scope.includes(values: :billable_metric_filter).find_each(batch_size:) do |charge_filter|
+      base_code = ChargeFilter.generate_code(charge_filter.to_h_with_discarded)
```
annvelents: maybe without discarded?

app/models/plan_rate_card.rb
annvelents, line 1: oops 🙈

app/services/database_migrations/enqueue_charge_filter_code_backfill_service.rb
```diff
@@ -0,0 +1,79 @@
...
+    # own it cannot tell a plan's own charge from an override that lost its parent
+    def next_charge_ids(cursor)
+      pending = ChargeFilter.unscope(:order)
+        .where(code: nil)
+        .where("charge_filters.charge_id = charges.id")
+
+      scope = Charge.kept
+        .joins(:plan)
+        .where(parent_id: nil, organization_id: billing_organization_ids)
+        .where(plans: {parent_id: nil})
```
annvelents: `.where(plans: {deleted_at: nil})`

app/jobs/database_migrations/backfill_charge_filter_codes_job.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      filters = charge.filters.includes(values: :billable_metric_filter).to_a
+
+      codes = filters.to_h { [it.id, it.code || ChargeFilter.generate_code(it.to_h)] }
+
+      # Only the filters that are not duplicated receive code
+      shared = codes.values.tally.select { |_, count| count > 1 }.keys.to_set
+
+      filters.each do |filter|
+        next if filter.code || shared.include?(codes[filter.id])
```
annvelents: let's log them

app/services/database_migrations/enqueue_child_charge_filter_code_backfill_service.rb
```diff
@@ -0,0 +1,79 @@
...
+    # what makes this re-runnable: a parent cleaned up by hand afterwards is picked up next time.
+    def next_parent_ids(cursor)
+      assigned = ChargeFilter.unscope(:order)
+        .where.not(code: nil)
+        .where("charge_filters.charge_id = charges.id")
+
+      scope = Charge.kept
+        .joins(:plan)
+        .where(parent_id: nil, organization_id: billing_organization_ids)
+        .where(plans: {parent_id: nil})
```
annvelents: where plan is not deleted

lib/tasks/upgrade_verification.rake
```diff
@@ -80,17 +59,37 @@ namespace :upgrade do
...
+      loop do
+        sleep 5
+
+        if backfill_jobs_running?(job)
+          pp "- Checking #{job.name}: 🔎"
+          pp "  -> #{Sidekiq::Queue.new(job.queue_name).size} jobs left 🧮"
+          next
         end
+
+        remaining = codeless_filters_that_can_be_filled(pass)
```
endenis: If I understood correctly, the enqueue logic and the completion check have completely different queries. Can there be a problem of this loop never stopping?

The query that enqueues filters the organizations with active subscriptions but I don't see anything like that inside the query in `parent_pending_codes_sql` 🤔

endenis (approved):

lib/tasks/upgrade_verification.rake
brunomiguelpinto: yes that can be an issue

brunomiguelpinto: changed it so the father after is done schedules do child job

lib/tasks/upgrade_verification.rake
```diff
@@ -80,19 +59,39 @@ namespace :upgrade do
...
+
+    DatabaseMigrations::EnqueueChargeFilterCodeBackfillService.call
+
+    codeless = -> { ChargeFilter.unscoped.where(code: nil, deleted_at: nil).count }
+    jobs = [DatabaseMigrations::BackfillChargeFilterCodesJob, DatabaseMigrations::BackfillChildChargeFilterCodesJob]
+    remaining = codeless.call
+
+    loop do
+      sleep 10
+
```
annvelents, line 86: let's also add a check that the queue size is < 1K, because new jobs will be scheduled only if there is a "space", but the queue might be busy with some other peocesses

lib/tasks/upgrade_verification.rake
```diff
@@ -80,19 +59,39 @@ namespace :upgrade do
...
+      next if jobs.any? { backfill_jobs_running?(it) }
```
brunomiguelpinto, line 87: check size of default queue if it's higher than the size we can skip put more jobs

annvelents (approved):

app/services/database_migrations/enqueue_charge_filter_code_backfill_service.rb
```diff
@@ -0,0 +1,81 @@
...
+        .order(:id)
+        .limit(batch_size)
+
+      scope = scope.where("charges.id > ?", cursor) if cursor
+
+      scope.pluck(:id)
+    end
+
+    # An organization with nothing active or pending should be skipped.
+    def billing_organization_ids
```
tiagolupepic, line 81: Probably it will be easier for the DB if you just interact in each organization instead.  Delegating this option/question to a guard clause in the line above.

app/services/database_migrations/enqueue_charge_filter_code_backfill_service.rb
```diff
@@ -0,0 +1,81 @@
...
+      pending = ChargeFilter.unscope(:order)
+        .where(code: nil)
+        .where("charge_filters.charge_id = charges.id")
+
+      scope = Charge.kept
+        .joins(:plan)
+        .where(parent_id: nil, organization_id: billing_organization_ids)
+        .where(plans: {parent_id: nil, deleted_at: nil})
+        .where(pending.arel.exists)
+        .order(:id)
```
tiagolupepic, line 72: just to double check, are they id sortable? otherwise would be better to use created_at and start from the past

app/jobs/database_migrations/backfill_child_charge_filter_codes_job.rb
```diff
@@ -0,0 +1,100 @@
...
+module DatabaseMigrations
+  class BackfillChildChargeFilterCodesJob < ApplicationJob
+    queue_as :low_priority
+    unique :until_executed, on_conflict: :log, lock_ttl: 6.hours
+
+    CHARGES_PER_BATCH = 1_000
+    IDS_PER_UPDATE = 10_000
+
+    # One row per charge filter. The fields are positional: they follow the `pluck` in `filters_of`,
+    Filter = Struct.new(:charge_id, :id, :code, :signature)
```
tiagolupepic: nitpicking.. use `Data` instead

lib/tasks/filters.rake
```diff
@@ -23,4 +23,34 @@ namespace :filters do
...
+  #     bills from then on, and that is not a migration's call.
+  #
+  #   - an override kept a filter its plan no longer has. There is nothing left to link it to —
+  #     the plan's filter is gone, or the parent charge was deleted and `dependent: :nullify` cut
+  #     the link. This is damage already done and no backfill undoes it.
+  #
+  # Cleaning up the first kind and running upgrade:perform_required_jobs again converges.
+  desc "Report the charge filters left without a code"
+  task report_codes: :environment do
+    scope = ChargeFilter.where(code: nil, deleted_at: nil)
```
annvelents, line 41: maybe.... we can also add a check if they belong to a charge that belongs to a plan with an active subscription 🙈

annvelents: but it will be a HUGE query 🙈

annvelents (approved):
Agree on @tiagolupepic's comments, but LGTM! 🚀

app/services/database_migrations/enqueue_charge_filter_code_backfill_service.rb
brunomiguelpinto: tested and with pagination on the id was quicker

app/services/database_migrations/enqueue_charge_filter_code_backfill_service.rb
brunomiguelpinto: yup good ideia changed it

PR #6080, opened 2026-08-05
[ING-502] feat(alerts): add alert resolution columns

aquinofb:
## Context

Lago alerts only fire when usage or a wallet balance crosses a limit; they never signal the return to normal. A customer who automatically blocks something on an alert has no automatic way to release that block. This adds the schema for a new `alert.resolved` webhook that a threshold can opt into.

Nothing here is reachable yet. The feature stays inactive until the API opt-in ships, so this is groundwork.

## Changes

- `notify_on` on alert thresholds. It defaults to notifying on trigger only, so existing alerts behave exactly as today.
- `kind` on alert events, distinguishing a real trigger from a recovery, and from a silent record written when a threshold already sits past its line at opt-in time.
- `in_alarm_thresholds` and `fully_resolved` on alert events, populated only on recovery rows.
- The three `triggered_alerts` associations are scoped to triggers, so readers whose name already promises triggers keep meaning that once recoveries exist. `Alert` also gains `all_triggered_alerts` for the full event stream, mirroring `billing_entities` and `all_billing_entities` on `Organization`.

Every column is added with a constant default, so these are metadata-only changes on PostgreSQL 15: no table rewrite, no long lock, and no back-fill.

Both alert serializers, the GraphQL threshold type and the export view all enumerate their fields explicitly, so none of the new columns reaches a customer-visible surface.

### DIFF
--- a/app/models/usage_monitoring/alert_threshold.rb
+++ b/app/models/usage_monitoring/alert_threshold.rb
@@ -2,11 +2,16 @@
 
 class UsageMonitoring::AlertThreshold < ApplicationRecord
   SOFT_LIMIT = 20
+  NOTIFY_ON_RESOLVED = "resolved"
 
   belongs_to :organization
   belongs_to :alert,
     foreign_key: "usage_monitoring_alert_id",
     class_name: "UsageMonitoring::Alert"
+
+  def notify_on_resolved?
+    notify_on.include?(NOTIFY_ON_RESOLVED)
+  end
 end
 
 # == Schema Information
@@ -16,6 +21,7 @@ end
 #
 #  id                        :uuid             not null, primary key
 #  code                      :string
+#  notify_on                 :string           default(["triggered"]), not null, is an Array
 #  recurring                 :boolean          default(FALSE), not null
 #  value                     :decimal(30, 5)   not null
 #  created_at                :datetime         not null
--- a/app/models/usage_monitoring/triggered_alert.rb
+++ b/app/models/usage_monitoring/triggered_alert.rb
@@ -2,6 +2,8 @@
 
 module UsageMonitoring
   class TriggeredAlert < ApplicationRecord
+    KINDS = {triggered: "triggered", resolved: "resolved", seeded: "seeded"}.freeze
+
     belongs_to :organization
     belongs_to :subscription, optional: true
     belongs_to :wallet, optional: true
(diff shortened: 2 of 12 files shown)


app/models/usage_monitoring/triggered_alert.rb
```diff
@@ -2,13 +2,17 @@
 
 module UsageMonitoring
   class TriggeredAlert < ApplicationRecord
+    KINDS = {triggered: "triggered", resolved: "resolved", seeded: "seeded"}.freeze
```
toommz, line 5: nitpick but it's strange to have a "triggered triggered alert"

app/models/usage_monitoring/alert_threshold.rb
```diff
@@ -2,11 +2,17 @@
 
 class UsageMonitoring::AlertThreshold < ApplicationRecord
   SOFT_LIMIT = 20
+  NOTIFY_ON_TRIGGERED = "triggered"
```
toommz: We don't seem to use this one, isn't it?

toommz (approved):

app/models/usage_monitoring/alert_threshold.rb
aquinofb: Good catch, you're right. It isn't used here, so I removed it from this PR.

It does get used, but in the follow-up that adds the `notify_on` validation rules, where it backs both the default and the "triggered is always required" check. Better for it to appear in the commit that first needs it than to sit here unused, so that's where the declaration lives now.

app/models/usage_monitoring/triggered_alert.rb
aquinofb: Agreed, it reads badly, and you're not the only one to trip on it.

I'd keep the values as they are though. They're deliberately the same words as the webhook event names, so `alert.triggered` and `alert.resolved` map one to one onto the `kind` of the row we store. We did consider something like `breach`/`recover` and dropped it for that reason: the delivery code would then have to translate between two vocabularies, and the payload would stop matching the row.

So the awkward part is really the model name. `TriggeredAlert` now holds rows that aren't triggers, because we reuse the same table for recoveries instead of adding a second one. Renaming it to something neutral like `AlertEvent` would be the honest fix, but it touches the table, both versions of the export view, the serializer, the webhook service, three associations and a good number of specs, so it doesn't belong in a schema PR that four branches are already stacked on.

Happy to open a follow-up for the rename if you think it's worth it. My instinct is yes, and sooner rather than later, since nothing outside Lago consumes these rows yet.

mariohd (approved):

PR #6081, opened 2026-08-05
feat(graphql): array-based customer payment connections

mariohd:
## Context

The customer connections UI is moving to an array-based shape as groundwork for multi-connection support, while behaviour stays one connection per type. Until now the GraphQL customer type only exposed singular connection fields (`providerCustomer`, `anrokCustomer`, …), which the new UI cannot consume. This delivers the GraphQL side only — the REST API is handled separately.

## Read side

The customer type now exposes `paymentProviderCustomers` and `integrationCustomers` lists, plus a customer-level `connectionStatus` (`not_connected | manual | connected`) derived from the default payment connection. The payment list always surfaces a "manual" option, synthesizing a placeholder when the customer has no persisted manual connection. The singular fields stay as deprecated shims returning the default.

## Write side

`Create`/`UpdateCustomerInput` accept a declarative `paymentProviderCustomers` array. A new `PaymentProviderCustomers::CreateOrUpdateBatchService` reconciles the customer's connections: connections omitted from the array (matched by `id`) are discarded, provider connections are created/updated, the reserved manual connection (null provider, `code "manual"`) is created on demand, and the element flagged `isDefault` becomes the customer default. The singular `providerCustomer` input still works; the array takes precedence when present.

## Notes

Ungated and backward-compatible, at most one connection per type for now. No migrations — the `code` / `is_default` columns already exist. The REST array and the integration-input fields (`code` / `is_default` / `category`) are out of scope and tracked separately.

### DIFF
--- a/app/graphql/types/customers/object.rb
+++ b/app/graphql/types/customers/object.rb
@@ -158,26 +158,6 @@ module Types
         object.active_subscriptions.count
       end
 
-      # The payment list always surfaces a "manual" option so the UI can select it. When the
-      # customer has no persisted manual connection, a non-persisted placeholder is prepended.
-      def payment_provider_customers
-        connections = object.payment_provider_customers.to_a
-        return connections if connections.any?(&:manual?)
-
-        [manual_connection_placeholder, *connections]
-      end
-
-      def manual_connection_placeholder
-        {
-          id: "#{object.id}-manual",
-          code: ::PaymentProviderCustomers::BaseCustomer::MANUAL_CODE,
-          is_default: false,
-          provider_customer_id: nil,
-          provider_payment_methods: nil,
-          sync_with_provider: nil
-        }
-      end
-
       def provider_customer # rubocop:disable GraphQL/ResolverMethodLength
         case object&.payment_provider&.to_sym
         when :stripe
--- a/app/graphql/types/customers/object.rb
+++ b/app/graphql/types/customers/object.rb
@@ -158,6 +158,26 @@ module Types
         object.active_subscriptions.count
       end
 
+      # The payment list always surfaces a "manual" option so the UI can select it. When the
+      # customer has no persisted manual connection, a non-persisted placeholder is prepended.
+      def payment_provider_customers
+        connections = object.payment_provider_customers.to_a
+        return connections if connections.any?(&:manual?)
+
+        [manual_connection_placeholder, *connections]
+      end
+
+      def manual_connection_placeholder
+        {
+          id: "#{object.id}-manual",
+          code: ::PaymentProviderCustomers::BaseCustomer::MANUAL_CODE,
+          is_default: false,
+          provider_customer_id: nil,
+          provider_payment_methods: nil,
+          sync_with_provider: nil
+        }
+      end
+
       def provider_customer # rubocop:disable GraphQL/ResolverMethodLength
         case object&.payment_provider&.to_sym
         when :stripe
--- a/app/graphql/types/customers/object.rb
+++ b/app/graphql/types/customers/object.rb
@@ -45,15 +45,31 @@ module Types
 
       field :billing_configuration, Types::Customers::BillingConfiguration, null: true
 
-      field :provider_customer, Types::PaymentProviderCustomers::Provider, null: true
+      field :payment_provider_customers, [Types::PaymentProviderCustomers::Provider], null: false
+      field :connection_status,
+        Types::PaymentProviderCustomers::ConnectionStatusEnum,
+        null: false,
+        method: :payment_connection_status,
+        description: "Payment connection status derived from the customer's default payment connection"
       field :subscriptions, [Types::Subscriptions::Object], resolver: Resolvers::Customers::SubscriptionsResolver
 
-      field :anrok_customer, Types::IntegrationCustomers::Anrok, null: true
-      field :avalara_customer, Types::IntegrationCustomers::Avalara, null: true
-      field :hubspot_customer, Types::IntegrationCustomers::Hubspot, null: true
-      field :netsuite_customer, Types::IntegrationCustomers::Netsuite, null: true
-      field :salesforce_customer, Types::IntegrationCustomers::Salesforce, null: true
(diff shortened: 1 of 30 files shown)


app/models/payment_provider_customers/base_customer.rb
```diff
@@ -11,6 +11,14 @@ class BaseCustomer < ApplicationRecord
 
     self.table_name = "payment_provider_customers"
 
+    MANUAL_CODE = "manual"
```
lovrocolic: There are minor changes proposed in another PR (453) so I guess we should also update this PR

app/graphql/types/payment_provider_customers/customer_input.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module Types
+  module PaymentProviderCustomers
+    class CustomerInput < Types::BaseInputObject
+      graphql_name "PaymentProviderCustomerInput"
+
+      argument :id, ID, required: false
+
+      argument :code, String, required: false
+      argument :is_default, Boolean, required: false
```
lovrocolic: I think we should also exclude `is_default` from this PR, similar as we did here - https://github.com/getlago/lago-api/pull/6153

lovrocolic:
For `integration_customers` we set `code` in `before_validation` hook. Here code is not set at all in api path. Can we please also tackle it here?

app/models/payment_provider_customers/base_customer.rb
mariohd: solved with a rebase

app/graphql/types/payment_provider_customers/customer_input.rb
```diff
@@ -0,0 +1,21 @@
...
+module Types
+  module PaymentProviderCustomers
+    class CustomerInput < Types::BaseInputObject
+      graphql_name "PaymentProviderCustomerInput"
+
+      argument :id, ID, required: false
+
+      argument :code, String, required: false
+      # NOTE: "lago_manual" flags the reserved null-provider connection; omit for provider connections.
+      argument :type, String, required: false
```
lovrocolic: why do we need this attribute?

app/services/payment_provider_customers/create_or_update_batch_service.rb
```diff
@@ -0,0 +1,87 @@
...
+  # is managed through the dedicated set-as-default mutation, not this array.
+  class CreateOrUpdateBatchService < BaseService
+    Result = BaseResult[:payment_provider_customers]
+
+    MANUAL_CODE = PaymentProviderCustomers::BaseCustomer::MANUAL_CODE
+
+    def initialize(customer:, payment_provider_customers:, new_customer:)
+      @customer = customer
+      @params_list = payment_provider_customers&.map { |c| c.to_h.deep_symbolize_keys }
+      @new_customer = new_customer
```
lovrocolic: I guess this attribute is unused? What was the idea behind it?

app/graphql/types/payment_provider_customers/customer_input.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      argument :payment_provider, Types::PaymentProviders::ProviderTypeEnum, required: false
+      argument :payment_provider_code, String, required: false
```
lovrocolic, line 15: `payment_provider` and `payment_provider_code` are not consumed on update. Do you think we should still use them somehow?

app/services/payment_provider_customers/create_or_update_batch_service.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+module PaymentProviderCustomers
+  # Declaratively reconciles a customer's payment provider connections from the GraphQL array:
+  # connections absent from the array (matched by id) are discarded, provider connections are
+  # created/updated, and the reserved manual connection is created on demand. The customer default
+  # is managed through the dedicated set-as-default mutation, not this array.
+  class CreateOrUpdateBatchService < BaseService
```
lovrocolic, line 8: From my understanding, the temporary goal is to merge this PR so that FE can start consuming it with only one element inside array.

Do you think that we should add temporary validation in this service that prevent having more than 1 element inside array? wdyt?

app/graphql/types/payment_provider_customers/customer_input.rb
mariohd: Yeah, you're right. 

The reason to have this is because this is shared with the create. In update they are not used, but in create they are.

app/services/payment_provider_customers/create_or_update_batch_service.rb
mariohd: Yeah, it's a good idea and a another layer of protection, the database already enforces just one, but we can add a new constraint limiting to be just one.

app/graphql/types/customers/object.rb
```diff
@@ -142,6 +158,26 @@ def active_subscriptions_count
         object.active_subscriptions.count
       end
 
+      # The payment list always surfaces a "manual" option so the UI can select it. When the
+      # customer has no persisted manual connection, a non-persisted placeholder is prepended.
+      def payment_provider_customers
+        connections = object.payment_provider_customers.to_a
+        return connections if connections.any?(&:manual?)
+
+        [manual_connection_placeholder, *connections]
```
domenicofalco: Since we haven't settled how manual will be handled (persisted row vs built on the fly) and it's out of M1 anyway, could we drop the placeholder from the response for now? We'd reintroduce it with the M2 default work once the design is decided.

One concrete reason beyond scope: the placeholder is a plain hash, so any query selecting `paymentMethods` under the array crashes — `undefined method 'organization' for an instance of Hash`, reproduced locally. Dropping it from M1 sidesteps that until the design call is made.

wdyt?

mariohd: Hey, got it, removed from this PR, can you please check it again?

domenicofalco: looks good! thanks

lovrocolic (approved):

PR #6082, opened 2026-08-05
[ING-504] fix(alerts): make wallet alert codes unique

aquinofb:
## Context

An alert `code` identifies an alert within its owner, and the subscription side enforces that with a unique index. The wallet equivalent was never created, so two wallet alerts on the same wallet can share a code.

That matters beyond tidiness. The REST endpoints look a wallet alert up by code for `show`, `update` and `destroy`, with no ordering, so where duplicates exist those calls can act on an arbitrary one of them. Deleting the wrong alert is possible.

An existing index already allows only one alert per wallet per type, and there are four wallet alert types, so duplicate codes can only appear across types:

```mermaid
graph LR
    W["wallet W"] --> A1["type: wallet_balance_amount<br/>code: low"]
    W --> A2["type: wallet_credits_balance<br/>code: low"]
    W --> A3["type: wallet_ongoing_balance_amount<br/>code: low"]
    R["GET / PUT / DELETE<br/>.../wallets/W/alerts/low"] --> F["find_by! wallet_id + code<br/>no ordering"]
    F -.-> A1
    F -.-> A2
    F -.-> A3
```

The create path already recognised the missing index by name, so that branch has been dead since it shipped, which is why nothing ever failed loudly.

## Changes

- Adds the missing unique index: one code per wallet among rows that are not soft deleted, created concurrently so writes are not blocked while it builds.
- Fixes the error reported when editing. The update path only recognised the subscription index, so an edit into a duplicate code reported a generic "alert already exists" against the record instead of naming the code field. Creating reported correctly, editing did not.
- Both paths now share one predicate for recognising a code collision. Two copies of that check drifting apart is what produced the bug above.

## Upgrading

Wallet alerts could share a code until now, so an existing instance may hold rows this index rejects. The migration resolves that itself rather than failing: where several alerts on one wallet share a code, the earliest keeps it and the others gain their alert type as a suffix, for example `low` and `low-wallet_credits_balance`. That suffix cannot collide, because one alert type per wallet is already unique.

Nothing is deleted, since either alert could be in use, and every rename is listed in the migration output with the wallet and both codes, so an operator can see what changed and update anything that addressed an alert by its old code:

```
-- Renamed 2 wallet alert(s) whose code was already used on the same wallet, suffixing the alert type.
   -> wallet 6bf7a1c4-3f8e-4a21-9c07-5d2b8e10af33 (main-wallet): low -> low-wallet_credits_balance
   -> wallet e2c9d05b-71a4-4f9d-8b16-0c3af5217d64: warn -> warn-wallet_ongoing_balance_amount
```

The wallet code appears when the wallet has one, since it is optional. A rollback drops the index but does not restore the previous codes, which is inherent to a data change.

Renaming is the least-bad of three options. Failing would block the upgrade until an operator ran manual SQL, and deleting would discard a working alert. A duplicate code was already broken: the endpoints below looked an alert up by code with no ordering, so they could act on either of them. Renaming makes that lookup well defined rather than taking anything away.

If a concurrent build fails partway, PostgreSQL leaves an index of that name marked invalid, and `IF NOT EXISTS` only checks the name. A plain retry then reports success while uniqueness is not actually enforced, so check `pg_index.indisvalid` before re-running.


app/services/usage_monitoring/create_alert_service.rb
```diff
@@ -99,7 +99,7 @@ def call
     rescue ActiveRecord::RecordInvalid => e
       result.record_validation_failure!(record: e.record)
     rescue ActiveRecord::RecordNotUnique => e
-      if e.message.include?("idx_alerts_code_unique_per_subscription") || e.message.include?("idx_alerts_code_unique_per_wallet")
+      if duplicate_code_error?(e)
```
toommz, line 102: 👍

db/migrate/20260805104301_add_alerts_code_unique_per_wallet_index.rb
```diff
@@ -0,0 +1,68 @@
...
+          ON c.wallet_id = a.wallet_id AND c.organization_id = a.organization_id
+        JOIN wallets w ON w.id = a.wallet_id
+        WHERE a.deleted_at IS NULL
+      )
+      UPDATE usage_monitoring_alerts a
+      SET code = a.code || '-' || a.alert_type::text
+      FROM duplicates d
+      WHERE d.id = a.id AND d.position > 1
+      RETURNING a.wallet_id, d.wallet_code, d.old_code, a.code AS new_code
+    SQL
```
toommz, line 82: I would just double check that this query (at least the "read" part) is fast enough to execute before in metabase or whatever. Looking at the plan too maybe

toommz (approved):

mariohd (approved):

db/migrate/20260805104301_add_alerts_code_unique_per_wallet_index.rb
aquinofb: Good call, and the answer is that we deliberately do not depend on it being fast during a release.

There is no index covering wallet_id and code today, so both the read that finds the duplicates and the unique build behind it are full passes over the table. On larger installs that does not belong in a deploy, so the rename and the index are done by hand ahead of the release: a temporary (wallet_id, code) index goes up first, which turns the count and the rename into indexed work and also gives the number of affected rows before anything is written, and it comes back down at the end. `index_state` then finds a valid index and returns early, so the migration itself is a no-op by the time the release runs.

Unattended installs have no such window, which is why the migration keeps both paths and why the rename is written to resolve duplicates rather than fail on them.

PR #6087, opened 2026-08-06
fix(invoice): display zero-unit flat fees in pdf

endenis:
## Context

Invoice PDFs were hiding charge and fixed-charge lines when their units were zero, even if they had a positive flat amount. The amount was still included in the invoice total, so the visible lines did not add up to the total.

This became more visible after [PR #5760](https://github.com/getlago/lago-api/pull/5760), which applies first-tier flat fees even when usage is zero. Those fees were calculated correctly but were still missing from the PDF.

## Description

Update the PDF fee filtering to include fees when either their units or amount is positive.
Fees with zero units and zero amount remain hidden.

Add tests for charge fees, fixed-charge fees, true-up fees, and the final PDF rendering.

### Visual comparison

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>

<tr>
<td>

Screenshot 2026-08-06 at 16 05 05

</td>

<td>

Screenshot 2026-08-06 at 16 08 52

</td>
</tr>
</table>


D1353L (approved):
LGTM 👍

annvelents (approved):
🚀🚀🚀

PR #6089, opened 2026-08-06
feat(multi-currency): make it generally available / Phase 1

mariohd:
## Context

Phase 1 of making multi-currency generally available, following the feature-flag removal runbook: remove all gating first, keeping the enabled behavior, before the flag definition and org values are dropped in later phases. The flag stays in feature_flags.yaml for now.

## Description

- update_currency_service: the customer currency is a default preference rather than a constraint; drop the mismatch/editability failures.
- preview_service: a customer/subscription currency mismatch no longer blocks the preview.
- organization_billing_service: always group subscriptions by their plan currency.
- wallets create_service: take the wallet currency from the param (falling back to the customer) and only set the customer currency when it is blank.
- Update the specs and drop the now-invalid disabled-behavior cases.

### DIFF
--- a/app/services/customers/update_currency_service.rb
+++ b/app/services/customers/update_currency_service.rb
@@ -17,28 +17,7 @@ module Customers
       return result if customer.currency == currency
 
       # Multi-currency: customer.currency becomes a default preference, not a constraint.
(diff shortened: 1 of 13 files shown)


aquinofb (approved):
LGTM

PR #6095, opened 2026-08-07
Chore(delete-bm): do not destroy events when discarding a billable metric

annvelents:
## Context

This job was introduced in 2023, however now it makes more sense to keep the events


jdenquin (approved):

brunomiguelpinto (approved):

brunomiguelpinto:
since the job is not used anymore should we remove it too ?

annvelents:
I would keep it at least until we have Vincent back...
but deleting the job will help us with retries - next time the worker tries to retry it after a deploy it won't find it anymore...

rsempe (approved):

PR #6098, opened 2026-08-07
feat(products): add product and category services

rsempe:
## Context

Service layer for `Product` and `ProductCategory`, two of the three catalog authoring entities introduced in #5652. Every write goes through a dedicated service so the REST and GraphQL surfaces (next slices) share identical behaviour and validation.

## Description

- Create/update/destroy services for `ProductCategory` and `Product`, with query objects and V1 serializers.
- Category destroy cascades product destroys; all soft-deletion cascades are explicit `discard_all!` chains.


app/queries/products_query.rb
```diff
@@ -0,0 +1,51 @@
...
+    products = with_product_types(products) if filters.product_types.present?
+
+    result.products = products
+    result
+  end
+
+  private
+
+  def base_scope
+    Product.where(organization:).ransack(search_params)
```
endenis, line 24: Should we include product_category and billable_metric here to avoid N+1 queries later? Or is it too early?

endenis (approved):

app/queries/products_query.rb
rsempe: Thanks - Adding them 👍

PR #6099, opened 2026-08-07
feat(products): add catalog GraphQL

rsempe:
## Context

GraphQL CRUD for the three catalog entities, calling the services from #6098 so REST (later in the stack) shares the same behaviour.

## Description

- Types, mutations and resolvers for `ProductCategory`, `Product` and `ProductFilter`, with permissions (`products:*`, `product_categories:*`, `product_filters:*`).
- Filter plans by product category, and products by a single `product_type`.
- Trim the type graph to what the FE consumes: drop `filters` from the product type and `products` from the category type.
- Eager load values in the product filters query to avoid N+1s.
- GraphQL validation errors keep the relation name (`billableMetric`, `productCategory`); REST (later in the stack) translates to the `_code` params it exposes.
- An explicit `values: null` on `updateProductFilter` fails with `value_is_mandatory` instead of crashing.

Note for review: `productsCount`/`filtersCount` resolve naively per row here; #5690 batches them through dataloader sources. Plans filter by a single `productCategoryId` (the plans page selector) while products filter by `productCategoryIds[]` + `withoutProductCategory` (the catalog list multi-select) — intentional asymmetry.


app/graphql/types/product_filter_values/input.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Types
+  module ProductFilterValues
+    class Input < BaseInputObject
+      graphql_name "ProductFilterValueInput"
+      description "ProductCategory item filter value input arguments"
```
endenis: hmmm not sure this description is up to date

endenis (approved):

app/graphql/types/product_filter_values/input.rb
rsempe: true! thanks

PR #6100, opened 2026-08-07
feat(products): add product filter services

rsempe:
## Context

Service layer for `ProductFilter`, the first of the three catalog authoring entities introduced in #5652. Every write goes through a dedicated service so the REST and GraphQL surfaces (later slices) share identical behaviour and validation.

## Description

- Create/update/destroy services for `ProductFilter`, with a query object and V1 serializer.
- Filter values are resolved by metric-filter key (`ResolveValuesService`) and validated (`ValidateValuesService`): values must belong to the product's metric, key-only wildcards cannot mix with specific values for the same key, and an empty values list is rejected.
- Filters are only allowed on metered products (`not_allowed_for_product_type` on fixed items).

### DIFF
--- a/app/services/product_filters/create_service.rb
+++ b/app/services/product_filters/create_service.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+module ProductFilters
+  class CreateService < BaseService
+    Result = BaseResult[:product_filter]
+
+    def initialize(product:, params:)
+      @product = product
+      @params = params.to_h.with_indifferent_access
+      super
+    end
+
+    activity_loggable(
+      action: "product_filter.created",
+      record: -> { result.product_filter }
+    )
+
+    def call
+      return result.not_found_failure!(resource: "product") unless product
+
+      unless product.usage?
+        return result.single_validation_failure!(field: :product, error_code: "invalid_product_type")
+      end
+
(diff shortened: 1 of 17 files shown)


app/services/product_filters/create_service.rb
```diff
@@ -0,0 +1,72 @@
...
+    def resolved_values
+      @resolved_values ||= ProductFilters::ResolveValuesService.call(product:, values_params: params[:values])
+    end
+
+    def create_values(product_filter)
+      resolved_values.values_params.each do |value_params|
+        product_filter.values.create!(
+          organization_id: product.organization_id,
+          billable_metric_filter_id: value_params[:billable_metric_filter_id],
+          value: value_params[:value]
```
endenis, line 67: What happens if after creation of a product filter the value gets removed from the billable metric filter? Can it produce some weird behavior or will everything continue to work as usual?

endenis (approved):

app/services/product_filters/create_service.rb
rsempe: Good spot, thanks @endenis 
I've created a ticket (BIL-528) for this - Not sure if we prefer to block the edition or accept and cascade as v1 does.

PR #6101, opened 2026-08-07
Fix(progressive-billing): rewrite source of charges for PB credits

annvelents:
## Context

When subscription is created, it can follow the original plan
if the original plan has progressive billing, it might get some progressive billing invoices
then subscription gets overridden, 
final subscription invoices is issued for new (children) charges
when we try to apply progressive billing credits, we're looking in the subscription invoices fees, matching charges of the progressive billing invoice
since charge_ids are not matching, we're not finding any matches and trying to create credit notes for all amount on the latest progressive billing invoice
but if there were multiple progressive billing invoices, final total progressively_billed amount can be higher than the latest progressive_billing invoice and we'll fail to issue the credit notes

## Description

when matching the invoice's fees to the progressive billing fees, not only check their charge_id, but also join them with charge and check charge.parent_id

### DIFF
--- a/app/services/credits/progressive_billing_service.rb
+++ b/app/services/credits/progressive_billing_service.rb
@@ -24,9 +24,15 @@ module Credits
         next unless progressive_billing_invoice
 
         progressive_billed_charge_ids = progressive_billing_invoice.fees.charge.pluck(:charge_id)
+        # A plan override creates a child charge, so a fee on this invoice can point at the child
+        # while the progressive billing invoice was billed on its parent. Both are matched, as the
+        # override can also predate the progressive billing invoices.
         invoice_charge_fees = invoice.fees.charge.where(subscription:).joins(:charge)
         total_charges_amount = invoice_charge_fees
-          .where("COALESCE(charges.parent_id, charges.id) IN (?)", progressive_billed_charge_ids)
+          .where(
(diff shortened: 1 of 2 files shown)


app/services/credits/progressive_billing_service.rb
```diff
@@ -81,5 +88,15 @@ def apply_credit_to_fees(progressive_billing_invoice)
...
     end
+
+    def matching_charge?(fee, progressive_fee)
+      fee.charge_id == progressive_fee.charge_id ||
+        fee.charge&.parent_id == progressive_fee.charge_id
+    end
+
+    def matching_charge_filter?(fee, progressive_fee)
+      fee.charge_filter_id == progressive_fee.charge_filter_id ||
+        fee.charge_filter&.code == progressive_fee.charge_filter&.code
```
annvelents, line 99: code to charge_filter is added in this pr: 
https://github.com/getlago/lago-api/pull/6079/changes/BASE..10789fdabc520e0749519b8a3df027cfbfa0a169#diff-662201afda3b7005c10e71a210add51d0ad199325e8c153c89f2f251cffbd37f

annvelents:
NoteL this PR was waiting to get also codes into charge_filters to distinguish charge_filters children

PR #6111, opened 2026-08-10
feat(products): add rate card services

rsempe:
## Context

Service layer for rate cards and their rates, the pricing objects attached to products. Every write goes through a dedicated service so the GraphQL surface (#5707) and REST (later in the stack) share identical behaviour and validation.

## Description

- Create/update/destroy services for `RateCard`, with a query object and V1 serializer.
- Append/update/destroy services for `RateCardRate` with the effective_from timeline rules: arrears canonicalized to midnight, one rate per day, appends only after the active rate.
- Rate card activity-log types (`rate_card.created/updated/deleted`) and the regenerated schema dumps that carry the new activity enum values.

### DIFF
--- a/app/serializers/v1/rate_card_rate_serializer.rb
+++ b/app/serializers/v1/rate_card_rate_serializer.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module V1
+  class RateCardRateSerializer < ModelSerializer
+    def serialize
+      {
+        lago_id: model.id,
+        effective_datetime: model.effective_datetime.iso8601,
+        status: model.status,
+        rate_model: model.rate_model,
+        rate_properties: model.rate_properties,
+        min_amount_cents: model.min_amount_cents,
+        billing_interval_count: model.billing_interval_count,
+        billing_interval_unit: model.billing_interval_unit,
+        applied_pricing_unit_conversion_rate: model.applied_pricing_unit_conversion_rate,
+        created_at: model.created_at.iso8601,
+        updated_at: model.updated_at.iso8601
+      }
+    end
+  end
+end
--- a/app/serializers/v1/rate_card_rate_serializer.rb
+++ b/app/serializers/v1/rate_card_rate_serializer.rb
@@ -5,6 +5,7 @@ module V1
     def serialize
       {
         lago_id: model.id,
+        code: model.code,
         effective_from: model.effective_from.iso8601,
         status: model.status,
         rate_model: model.rate_model,
--- a/app/serializers/v1/rate_card_rate_serializer.rb
+++ b/app/serializers/v1/rate_card_rate_serializer.rb
@@ -5,7 +5,7 @@ module V1
     def serialize
       {
         lago_id: model.id,
-        effective_datetime: model.effective_datetime.iso8601,
+        effective_from: model.effective_from.iso8601,
         status: model.status,
         rate_model: model.rate_model,
         rate_properties: model.rate_properties,
--- a/app/services/rate_card_rates/create_service.rb
+++ b/app/services/rate_card_rates/create_service.rb
@@ -20,10 +20,6 @@ module RateCardRates
     def call
       return result.not_found_failure!(resource: "rate_card") unless rate_card
 
-      if rate_card.attached_to_subscriptions?
-        return result.single_validation_failure!(field: :rate_card, error_code: "attached_to_subscriptions")
-      end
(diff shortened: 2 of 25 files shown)


app/services/rate_card_rates/create_service.rb
```diff
@@ -0,0 +1,49 @@
...
+    activity_loggable(
+      action: "rate_card.updated",
+      record: -> { rate_card },
+      condition: -> { emit_activity_log }
+    )
+
+    def call
+      return result.not_found_failure!(resource: "rate_card") unless rate_card
+
+      if rate_card.attached_to_subscriptions?
```
endenis: Wouldn't this prevent attaching rates with future `effective_from` to the currently used rate card?

app/services/rate_cards/update_service.rb
```diff
@@ -0,0 +1,67 @@
...
+    def assign_attributes
+      rate_card.name = params[:name] if params.key?(:name)
+      rate_card.description = params[:description] if params.key?(:description)
+      rate_card.currency = params[:currency] if params.key?(:currency)
+      rate_card.billing_timing = params[:billing_timing] if params.key?(:billing_timing)
+      rate_card.proration = params[:proration] if params.key?(:proration)
+      rate_card.display_on_invoice = params[:display_on_invoice] if params.key?(:display_on_invoice)
+      rate_card.regroup_paid_fees = params[:regroup_paid_fees] if params.key?(:regroup_paid_fees)
+      rate_card.applied_pricing_unit_code = params[:applied_pricing_unit_code] if params.key?(:applied_pricing_unit_code)
+      rate_card.wallet_targetable = params[:wallet_targetable] if params.key?(:wallet_targetable)
```
endenis, line 72: `CreateService` performs a check using `organization.events_targeting_wallets_enabled?`

Isn't this check needed in this service as well?

app/services/rate_cards/update_service.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    def assign_attributes
+      rate_card.name = params[:name] if params.key?(:name)
+      rate_card.description = params[:description] if params.key?(:description)
+      rate_card.currency = params[:currency] if params.key?(:currency)
+      rate_card.billing_timing = params[:billing_timing] if params.key?(:billing_timing)
+      rate_card.proration = params[:proration] if params.key?(:proration)
+      rate_card.display_on_invoice = params[:display_on_invoice] if params.key?(:display_on_invoice)
+      rate_card.regroup_paid_fees = params[:regroup_paid_fees] if params.key?(:regroup_paid_fees)
+      rate_card.applied_pricing_unit_code = params[:applied_pricing_unit_code] if params.key?(:applied_pricing_unit_code)
```
endenis, line 71: `CreateService` performs a check using `organization.pricing_units.exists?(code: params[:applied_pricing_unit_code])`

Isn't this check needed here as well?

app/serializers/v1/rate_card_rate_serializer.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module V1
+  class RateCardRateSerializer < ModelSerializer
+    def serialize
+      {
```
endenis, line 6: I think that this serializer should also render RateCardRate's code

app/services/rate_card_rates/create_service.rb
rsempe: Yes, once a subscription bills through a rate card, the whole timeline freezes, future rates included
Ensuring the behavior with the product right now

app/services/rate_cards/update_service.rb
rsempe: you're right, we plan to remove `wallet_targetable` on a future PR but adding it here

app/services/rate_cards/update_service.rb
rsempe: good spot 👍

app/serializers/v1/rate_card_rate_serializer.rb
rsempe: thanks!

app/services/rate_card_rates/create_service.rb
rsempe: After discussing with @michaelponrajah - you're right, we should be able to add a future rate
Adding this, thanks! 🤗

endenis (approved):

PR #6113, opened 2026-08-11
feat(netsuite-integration): Add functionality to void invoice on Netsuite

osmarluz:
## Context

When an invoice already synced to NetSuite is voided in Lago, the void never reaches NetSuite. The two systems drift apart and the invoice has to be voided manually in NetSuite.

## Description

Void the invoice in NetSuite when it is voided in Lago. `Invoices::VoidService` now enqueues a new `Integrations::Aggregator::Invoices::VoidJob` / `VoidService` pair that calls Nango's `void-invoice` action. The sync only runs when the invoice is voided, not self-billed, and belongs to a NetSuite integration with invoice sync enabled (`Invoice#should_sync_voided_invoice?`). If the invoice was never synced to NetSuite, the payload raises `BasePayload::Failure`, which the job retries with polynomial backoff before dead-queueing.

The void path reuses the error handling built for invoice creation:
- `retryable_error?` / `invalid_login_attempt_error?` move from `CreateService` to the shared invoices `BaseService`.
- The `Net::ReadTimeout` handling (6-minute delay for NetSuite to cover Nango's 5-minute upstream timeout, polynomial backoff for other providers) is extracted from `CreateJob` into a `ProviderDelayedRetry` concern included by both jobs.

### DIFF
--- a/app/jobs/concerns/integrations/aggregator/provider_delayed_retry.rb
+++ b/app/jobs/concerns/integrations/aggregator/provider_delayed_retry.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module Integrations
+  module Aggregator
+    module ProviderDelayedRetry
+      extend ActiveSupport::Concern
+
+      READ_TIMEOUT_ATTEMPTS = 6
+
+      # NOTE: NetSuite waits longer to avoid racing in-flight Nango calls; others use polynomial backoff.
+      # 6 minutes covers Nango's 5-minute upstream NetSuite action timeout with a safety margin.
+      DELAY_BY_PROVIDER_KEY = {
+        "netsuite" => 6.minutes
+      }.freeze
+
+      included do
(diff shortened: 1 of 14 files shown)


osmarluz:
> 
> HOLD — the NetSuite void path can silently select Xero for customers with multiple accounting connections.
> 
> * Scope `should_sync_voided_invoice?`, the void service's integration lookup, and the retry-provider lookup to the same NetSuite connection. The trigger currently accepts any accounting integration while the latter lookups use arbitrary `.accounting_kind.first`; with supported Xero + NetSuite connections, the service can return at the provider guard without ever voiding the NetSuite invoice.
> * Add a mixed Xero/NetSuite regression spec proving the NetSuite API call and six-minute timeout retry use the intended connection regardless of record ordering.

This is ok, it's mirroring the logic we currently have on `should_sync_invoice?` and I verified that one accounting connection per customer is the current production invariant

PR #6114, opened 2026-08-11
fix(payments): allow 3DS retry on activation payments

ancorcruz:
Part of INT-210

## Context

A payment-gated subscription activates only once its first invoice is paid. When the customer's card requires 3D Secure, Stripe rejects the off-session charge with `authentication_required`, and the payment can only succeed if it is retried on-session so the customer can complete the challenge.

That retry was governed by the Stripe connection's `supports_3ds` setting, which is off by default. With it off, the rejection was terminal: the invoice was marked as failed and the subscription was canceled seconds after creation — indistinguishable from a real decline, and with no path for the customer to authenticate or to subscribe with that card at all.

## Description

The activation payment of a payment-gated subscription now always gets the on-session retry, regardless of the connection setting, so the customer is offered the authentication challenge and the subscription stays incomplete while it is pending. The setting keeps governing recurring invoices, where escalating a silent off-session charge to an interactive one stays an opt-in.

The scoping lever is `Invoice#subscription_payment_gated?` (open invoice + incomplete gated subscription), so no new column or plan-level flag is needed: a recurring invoice of an active subscription can never match.

Both decision points now read the same policy, which matters because they race each other: the synchronous charge attempt (which drives the retry) and the `payment_intent.payment_failed` webhook for the rejected off-session intent (which would otherwise settle the invoice as failed and cancel the subscription before the retry runs).

Organizations that never enabled `supports_3ds` now receive `payment.requires_action` with the Stripe-hosted challenge URL for activation payments, and must pass it to their customer. If they ignore that webhook, the subscription sits incomplete until the activation timeout instead of being canceled immediately — a better failure mode, but a change for orgs that never opted in. Note `timeout_hours` defaults to `0`, which means no deadline.

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -86,7 +86,7 @@ class Payment < ApplicationRecord
   end
 
   def gated_subscription_activation?
-    payable.is_a?(Invoice) && payable.subscription_payment_gated?
+    payable.subscription_payment_gated?
   end
 
   def method_display_name
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -85,6 +85,10 @@ class Payment < ApplicationRecord
     payment_provider&.payment_type
   end
 
+  def gated_subscription_activation?
+    payable.is_a?(Invoice) && payable.subscription_payment_gated?
+  end
+
   def method_display_name
     return nil if provider_payment_method_data.blank?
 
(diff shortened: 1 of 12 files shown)


osmarluz (approved):
🚀

PR #6116, opened 2026-08-11
feat(payment_terms): Add value object and date math

D1353L:
## Context

We're replacing the legacy net_payment_term integer with a structured payment_term object supporting six new types:

term_type | Fields | Rule | Due date (invoice issued 2026‑07‑15)
-- | -- | -- | --
due_on_receipt | — | issue day | 2026‑07‑15
net | days | +N days | 2026‑08‑14
end_of_month | — | last day of month | 2026‑07‑31
net_end_of_month (US) | days | EOM then +N | 2026‑08‑30
days_end_of_month (EU) | days | +N then EOM | 2026‑08‑31
day_of_month | day_of_month (1–31), month_offset (default 1) | Dth of month+offset | 2026‑08‑15

## Description

- Added `PaymentTerm` value object with `due_date` calculations based on `payment_term` rule
- The `net_payment_term_alias` is applicable only for `due_on_receipt` and `net` terms. For all others it will be `nil`.

- For the `day_of_month` (MFI) term, 2 normalizations are introduced:
  - If `day_of_month` number is beyond the month's total days, set it to the last day of the month.
  _Example_: `day_of_month`: 31 -> in February, it will be 28 or 29.
  - If the date is in the past, move it to the next month.
  _Example_: current date 13/08, `day_of_month`: 3, `offset`: 0 -> move to the next month (03/09). Otherwise, because offset is 0, it would be a past date (03/08).

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/app/models/payment_term.rb
+++ b/app/models/payment_term.rb
@@ -2,7 +2,14 @@
 
 # Value object for a structured payment term, stored as jsonb tagged by term_type
 class PaymentTerm
-  TERM_TYPES = %w[due_on_receipt net end_of_month net_end_of_month days_end_of_month day_of_month].freeze
+  FIELDS_BY_TERM_TYPE = {
+    "due_on_receipt" => [],
+    "net" => ["days"],
+    "end_of_month" => [],
+    "net_end_of_month" => ["days"],
+    "days_end_of_month" => ["days"],
+    "day_of_month" => ["day_of_month", "month_offset"]
+  }.freeze
 
   attr_reader :term_type, :days, :day_of_month, :month_offset
 
@@ -19,8 +26,8 @@ class PaymentTerm
 
   def initialize(term_type:, days: nil, day_of_month: nil, month_offset: nil)
     @term_type = term_type.to_s
-    @days = days&.to_i
-    @day_of_month = day_of_month&.to_i
+    @days = days&.to_i if carries?("days")
+    @day_of_month = day_of_month&.to_i if carries?("day_of_month")
     @month_offset = normalized_month_offset(month_offset)
   end
 
@@ -42,6 +49,8 @@ class PaymentTerm
   end
 
   def due_date_for(issuing_date)
+    issuing_date = issuing_date.to_date
+
     case term_type
     when "due_on_receipt" then issuing_date
     when "net" then issuing_date + days
@@ -49,15 +58,20 @@ class PaymentTerm
     when "net_end_of_month" then issuing_date.end_of_month + days # US: EOM, then +N
     when "days_end_of_month" then (issuing_date + days).end_of_month # EU: +N, then EOM
     when "day_of_month" then day_of_month_due_date(issuing_date)
+    else raise ArgumentError, "unknown term_type: #{term_type}"
     end
   end
 
   private
 
+  def carries?(field)
+    FIELDS_BY_TERM_TYPE.fetch(term_type, []).include?(field)
+  end
+
   # Only day_of_month terms carry a month_offset.
   # If absent - next month (1) by default.
   def normalized_month_offset(month_offset)
-    if term_type == "day_of_month"
+    if carries?("month_offset")
       (month_offset || 1).to_i
     end
   end
@@ -75,6 +89,6 @@ class PaymentTerm
 
   # Clamp the configured day to the target month: day 31 → Sep 30 / Feb 28 (29 on leap years)
   def clamped_day_in_month(date)
-    Date.new(date.year, date.month, [day_of_month, date.end_of_month.day].min)
+    date.change(day: [day_of_month, date.end_of_month.day].min)
   end
 end
--- a/app/models/payment_term.rb
+++ b/app/models/payment_term.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+# Value object for a structured payment term, stored as jsonb tagged by term_type
+class PaymentTerm
+  TERM_TYPES = %w[due_on_receipt net end_of_month net_end_of_month days_end_of_month day_of_month].freeze
+
+  attr_reader :term_type, :days, :day_of_month, :month_offset
+
+  def self.from_h(hash)
+    hash = hash.to_h.with_indifferent_access
+
+    new(
+      term_type: hash[:term_type],
+      days: hash[:days],
+      day_of_month: hash[:day_of_month],
+      month_offset: hash[:month_offset]
+    )
+  end
+
+  def initialize(term_type:, days: nil, day_of_month: nil, month_offset: nil)
+    @term_type = term_type.to_s
+    @days = days&.to_i
+    @day_of_month = day_of_month&.to_i
+    @month_offset = normalized_month_offset(month_offset)
+  end
+
+  def to_h
+    {
+      "term_type" => term_type,
+      "days" => days,
+      "day_of_month" => day_of_month,
+      "month_offset" => month_offset
+    }.compact
+  end
+
+  # N for net, 0 for due_on_receipt, nil for the four types the integer cannot represent.
+  def net_payment_term_alias
+    case term_type
+    when "net" then days
+    when "due_on_receipt" then 0
+    end
+  end
+
+  def due_date_for(issuing_date)
+    case term_type
+    when "due_on_receipt" then issuing_date
+    when "net" then issuing_date + days
+    when "end_of_month" then issuing_date.end_of_month
+    when "net_end_of_month" then issuing_date.end_of_month + days # US: EOM, then +N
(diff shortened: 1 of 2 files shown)


app/models/payment_term.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+# Value object for a structured payment term, stored as jsonb tagged by term_type
+class PaymentTerm
+  TERM_TYPES = %w[due_on_receipt net end_of_month net_end_of_month days_end_of_month day_of_month].freeze
```
ancorcruz: the term types are defined but not used, whatever value received is accepted. Shouldn't https://github.com/getlago/lago-api/pull/6116/changes#diff-0d689a45f18f4300d470011ef73491bff6a13d8c596851a5a1ed7486c6f0a953R13 ensure the value is defined within the TERM_TYPES?

app/models/payment_term.rb
```diff
@@ -0,0 +1,80 @@
...
+  # N for net, 0 for due_on_receipt, nil for the four types the integer cannot represent.
+  def net_payment_term_alias
+    case term_type
+    when "net" then days
+    when "due_on_receipt" then 0
+    end
+  end
+
+  def due_date_for(issuing_date)
+    case term_type
```
ancorcruz, line 54: should have an `else` branch raising an exception as not supported term type?

app/models/payment_term.rb
```diff
@@ -0,0 +1,80 @@
...
+    when "due_on_receipt" then issuing_date
+    when "net" then issuing_date + days
+    when "end_of_month" then issuing_date.end_of_month
+    when "net_end_of_month" then issuing_date.end_of_month + days # US: EOM, then +N
+    when "days_end_of_month" then (issuing_date + days).end_of_month # EU: +N, then EOM
```
ancorcruz, line 59: `days` could have `nil` value because of setter `days&.to_i` then cause an exception here. We could remove the safe navigation operator and ensure days is at least zero. What do you think?

app/models/payment_term.rb
```diff
@@ -0,0 +1,80 @@
...
+    if due_date < issuing_date
+      clamped_day_in_month(due_date >> 1)
+    else
+      due_date
+    end
+  end
+
+  # Clamp the configured day to the target month: day 31 → Sep 30 / Feb 28 (29 on leap years)
+  def clamped_day_in_month(date)
+    Date.new(date.year, date.month, [day_of_month, date.end_of_month.day].min)
```
ancorcruz: same applies here, `day_of_month` could have a nil value because safe navigation operator then this code will raise an exception. 


also, in the codebase we use `date.change(day:)` perhaps we should use the same here.

app/models/payment_term.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  def initialize(term_type:, days: nil, day_of_month: nil, month_offset: nil)
+    @term_type = term_type.to_s
+    @days = days&.to_i
+    @day_of_month = day_of_month&.to_i
+    @month_offset = normalized_month_offset(month_offset)
+  end
+
+  def to_h
```
ancorcruz, line 34: `to_h` normalizes `month_offset` away for non-day_of_month types but keeps stale `days` and `day_of_month`, an asymmetry the spec claims not to have.

app/models/payment_term.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  # N for net, 0 for due_on_receipt, nil for the four types the integer cannot represent.
+  def net_payment_term_alias
+    case term_type
+    when "net" then days
+    when "due_on_receipt" then 0
+    end
+  end
+
+  def due_date_for(issuing_date)
```
ancorcruz, line 51: `due_date_for` never coerces its argument to a Date, so a Time/TimeWithZone means seconds for three branches and raises for another; the return class also varies by term type.

ancorcruz (approved):
good stuff. left some comments for you.

app/models/payment_term.rb
D1353L: Thanks for the good catches!

Changed this constant to represent the shape of each payment term - it's used in the following PR #6117 in PaymentTerm::ValidateService

app/models/payment_term.rb
D1353L: Updated: `days` and `day_of_month` are set only when applicable in the initialize method
```
@days = days&.to_i if carries?("days")
@day_of_month = day_of_month&.to_i if carries?("day_of_month")
```

app/models/payment_term.rb
D1353L: Good catch! Added `issuing_date = issuing_date.to_date`

app/models/payment_term.rb
D1353L: As the `days` param is required, it's better to fail loudly if it's `nil`. The validation with proper messages is implemented in the following PR #6117  (PaymentTerms::ValidateService)

app/models/payment_term.rb
D1353L: Added ✅

app/models/payment_term.rb
D1353L: Updated with `date.change(day: [day_of_month, date.end_of_month.day].min)`

app/models/payment_term.rb
```diff
@@ -0,0 +1,94 @@
...
+      "day_of_month" => day_of_month,
+      "month_offset" => month_offset
+    }.compact
+  end
+
+  # N for net, 0 for due_on_receipt, nil for the four types the integer cannot represent.
+  def net_payment_term_alias
+    case term_type
+    when "net" then days
+    when "due_on_receipt" then 0
```
tiagolupepic, line 47: I'd recommend to declare the other options here and raise an exception when the term_type is not defined.

```
when "end_of_month", "days_end_of_month"...
  nil
else
  raise ...
```

Having this, you're making sure this method will be visited when a new type is introduced.

app/models/payment_term.rb
```diff
@@ -0,0 +1,94 @@
...
+    issuing_date = issuing_date.to_date
+
+    case term_type
+    when "due_on_receipt" then issuing_date
+    when "net" then issuing_date + days
+    when "end_of_month" then issuing_date.end_of_month
+    when "net_end_of_month" then issuing_date.end_of_month + days # US: EOM, then +N
+    when "days_end_of_month" then (issuing_date + days).end_of_month # EU: +N, then EOM
+    when "day_of_month" then day_of_month_due_date(issuing_date)
+    else raise ArgumentError, "unknown term_type: #{term_type}"
```
tiagolupepic, line 61: this is a good approach!

app/models/payment_term.rb
```diff
@@ -0,0 +1,94 @@
...
+    if due_date < issuing_date
+      clamped_day_in_month(due_date >> 1)
+    else
+      due_date
+    end
+  end
+
+  # Clamp the configured day to the target month: day 31 → Sep 30 / Feb 28 (29 on leap years)
+  def clamped_day_in_month(date)
+    date.change(day: [day_of_month, date.end_of_month.day].min)
```
tiagolupepic, line 92: just an alternative and you don't need to change it. 

It's possible to also do using Date

```
Date.new(date.year, date.month, -1)
```

Ruby automatically identifies the last day of the month + year.

app/models/payment_term.rb
```diff
@@ -0,0 +1,94 @@
...
+  FIELDS_BY_TERM_TYPE = {
+    "due_on_receipt" => [],
+    "net" => ["days"],
+    "end_of_month" => [],
+    "net_end_of_month" => ["days"],
+    "days_end_of_month" => ["days"],
+    "day_of_month" => ["day_of_month", "month_offset"]
+  }.freeze
+
+  attr_reader :term_type, :days, :day_of_month, :month_offset
```
tiagolupepic, line 14: out of curiosity, why not use Data.define for this PaymentTerm?

app/models/payment_term.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def self.from_h(hash)
+    hash = hash.to_h.with_indifferent_access
+
+    new(
+      term_type: hash[:term_type],
+      days: hash[:days],
+      day_of_month: hash[:day_of_month],
+      month_offset: hash[:month_offset]
```
tiagolupepic, line 23: maybe I'm late to the party, would be good to have a version here? For example, `version: 1` 
just wondering if we change the `net` recipe or other term_type and somehow the previous stored payment terms doesn't have some attributes and we can keep the backward compability

PR #6122, opened 2026-08-11
feat(orders): Execute subscription amendments

toommz:
Step 3 of the CPQ execution engine: `subscription_amendment` under `execute_in_lago`.

An amendment restates one plan on a live subscription. Execution routes it through `Subscriptions::CreateService`, the entry point the API and the UI already use for a plan change, so it inherits Lago's own semantics in both directions: a raise or a repricing rotates the subscription now, keeping the `external_id` and the billing anchor, while a reduction keeps the target running and schedules the replacement for the next billing day (`Subscription#downgrade_plan_date`). The negotiated plan is built up front with `Plans::OverrideService` so the dispatch, which compares plan ids before prices, still rotates a repricing of the same catalog plan.

Both the approval gate and the executor inherit from their `subscription_creation` counterparts, since an amendment payload is a creation payload with extra constraints: a single plan, a mandatory ending date, and an active target subscription.

### DIFF
--- a/app/graphql/types/orders/execution_record.rb
+++ b/app/graphql/types/orders/execution_record.rb
@@ -12,6 +12,7 @@ module Types
 
       field :applied_coupon_ids, [ID], null: false
       field :subscription_ids, [ID], null: false
+      field :terminated_subscription_ids, [ID], null: false
       field :wallet_ids, [ID], null: false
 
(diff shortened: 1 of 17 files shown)


mariohd (approved):
Looks good

PR #6125, opened 2026-08-11
feat(products): add rate card rate GraphQL

rsempe:
## Context

GraphQL surface for the rates of a rate card — the effective-dated pricing entries whose rules live in #6112 and services in #6111.

## Description

- Types, enums and inputs for `RateCardRate` with the derived status, create/update/destroy mutations and a per-card `rateCardRates` query, behind the `rate_cards:*` permissions family.
- The `rate_cards` permissions and organization associations shared with the card surface (#5707).


app/graphql/types/rate_card_rates/input.rb
endenis, line 1: Is this file a duplicate of the `create_input.rb`?

app/queries/rate_card_rates_query.rb
```diff
@@ -0,0 +1,26 @@
...
+    rate_card_rates = rate_card_rates.order(effective_from: :desc)
+
+    result.rate_card_rates = rate_card_rates
+    result
+  end
+
+  private
+
+  def base_scope
+    RateCardRate.where(organization:)
```
endenis: When computing [status](https://github.com/getlago/lago-api/blob/eca2236a9cef2159c6aa3e726e72dbc2dbaadb67/app/models/rate_card_rate.rb#L75) for each RateCardRate we:
1. load a RateCard
2. run a query to check if there is a newer effective_from.

This can create a lot of queries when fetching a list of rates.

I am not sure what the best way to fix the second thing but the first one can be solved if we preload the RateCard here.

```suggestion
    RateCardRate.where(organization:).includes(:rate_card)
```

app/graphql/types/rate_card_rates/input.rb
rsempe: Not exactly, `RateCardRateInput` is the nested shape used by `createRateCard`'s rates, while `CreateInput` is the standalone mutation's input

Making `CreateInput` inherit from `Input` and add only `rate_card_id` 👍

app/queries/rate_card_rates_query.rb
rsempe: True - fixed 👍

endenis (approved):

PR #6126, opened 2026-08-11
Ing 454

mariohd:
## Context

This wires up the model layer for the multiple-connections-on-customer work. It builds on the schema and models that already landed in main and adds the remaining associations and helpers needed by later work.

## Description

`IntegrationCustomers::BaseCustomer`: remove the `only_one_tax_integration_per_customer` validation so a customer may connect more than one tax integration. Per-category `code` uniqueness and the partial-unique `is_default` per `(customer, category)` were already merged; the `*_kind` scopes and `tax_kind?` helper are kept, as they may now return several records and are still used elsewhere.

`Subscription`, `Wallet` and `RecurringTransactionRule`: add `has_many :billing_object_connections, as: :owner, dependent: :destroy` so billing objects own their polymorphic connections and cascade on deletion.

`Invoice`: add the per-category connection snapshots. Since the `invoice_connections` table carries `invoice_id`, the invoice exposes `has_many :invoice_connections` plus one scoped `has_one` accessor per category (`payment_connection`, `tax_connection`, `accounting_connection`, `crm_connection`). Add `manual_payment?`, which reads the payment snapshot rather than the customer's live provider and returns true when no provider customer was captured, meaning the invoice skips provider collection.

## Notes for reviewers

The ticket text predates the schema that actually landed, so two bullets read against an earlier design:

- `BillingObjectConnection` was described as validating a `connection_code` string. The merged model references connections by foreign key instead (`payment_provider_customer_id` / `integration_customer_id`, both optional), so there is no `connection_code` and no such validation.
- The invoice was described as `belongs_to` the four snapshot connections. The relationship is inverted because `invoice_connections` holds `invoice_id`, so the invoice side is `has_many` / scoped `has_one` rather than `belongs_to`. One snapshot per category is still enforced by the unique `(invoice_id, category)` index.

The `BillingObjectConnection` model, table and factory were already merged in main, so that part of the scope is not in this PR.

`manual_payment?` currently treats "no provider customer on the payment snapshot" as manual. If a reserved manual `payment_provider_customer` row is introduced later, this should switch to `payment_connection&.payment_provider_customer&.manual?`.

## Tests

All model specs pass: 466 examples, 0 failures across the touched models and the connection models.

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -82,6 +82,11 @@ class Invoice < ApplicationRecord
   has_many :plans, through: :subscriptions
   has_many :metadata, class_name: "Metadata::InvoiceMetadata", dependent: :destroy
   has_many :credit_notes
+  has_many :invoice_connections, dependent: :destroy
+  has_one :payment_connection, -> { where(category: :payment) }, class_name: "InvoiceConnection"
+  has_one :tax_connection, -> { where(category: :tax) }, class_name: "InvoiceConnection"
+  has_one :accounting_connection, -> { where(category: :accounting) }, class_name: "InvoiceConnection"
+  has_one :crm_connection, -> { where(category: :crm) }, class_name: "InvoiceConnection"
   has_many :progressive_billing_credits, class_name: "Credit", foreign_key: :progressive_billing_invoice_id
   has_many :invoice_settlements, foreign_key: :target_invoice_id
 
--- a/app/models/recurring_transaction_rule.rb
+++ b/app/models/recurring_transaction_rule.rb
@@ -8,6 +8,7 @@ class RecurringTransactionRule < ApplicationRecord
   belongs_to :organization
   belongs_to :payment_method, optional: true
 
+  has_many :billing_object_connections, as: :owner, dependent: :destroy
   has_many :applied_invoice_custom_sections,
     class_name: "RecurringTransactionRule::AppliedInvoiceCustomSection",
     dependent: :destroy
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -19,6 +19,7 @@ class Subscription < ApplicationRecord
   has_many :invoice_subscriptions
   has_many :invoices, through: :invoice_subscriptions
   has_many :integration_resources, as: :syncable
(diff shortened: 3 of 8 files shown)


app/models/invoice.rb
```diff
@@ -571,6 +576,12 @@ def allow_manual_payment?
     MANUALLY_PAYABLE_INVOICE_STATUS.include?(status.to_sym)
   end
 
+  # The invoice was issued under a manual (no-provider) payment connection, so it skips
+  # provider collection. Reads the snapshot, never the customer's live payment provider.
+  def manual_payment?
+    payment_connection&.payment_provider_customer_id.nil?
```
lovrocolic: why we need to check `payment_provider_customer_id`? Is this needed?

app/models/integration_customers/base_customer.rb
```diff
@@ -36,7 +36,6 @@ class BaseCustomer < ApplicationRecord
 
     validates :customer_id, uniqueness: {scope: :type}
     validates :code, uniqueness: {scope: %i[customer_id category]}, allow_nil: true
-    validate :only_one_tax_integration_per_customer, if: :tax_kind?
```
lovrocolic, line 39: I would suggest removing this validation later. For now we are just building core functionalities, but it is too early to unlock the feature. Maybe we could add feature flag guard, but I would rather skip it completely

lovrocolic: Is there any reason to remove validation at this stage?

app/models/invoice.rb
lovrocolic: On second look, I would exclude this method from this PR. We haven't performed any migration for invoice_connections so currently the method is not even behave correctly

app/models/integration_customers/base_customer.rb
mariohd: Hey, i've did the changes you commented. That makes sense! 

Thanks

app/models/invoice.rb
mariohd: Hey, i've did the changes you commented. That makes sense!

Thanks

lovrocolic (approved):

PR #6128, opened 2026-08-11
fix(charges): Skip existing child charges in cascade

D1353L:
## Context

Cascade jobs run at-least-once; a retry or concurrent plan update could materialize the same parent charge twice on a child plan, and every copy is billed. 

## Description

Skip a child plan when it already holds a non-deleted charge for the same parent charge, making the cascade idempotent. A discarded copy does not block re-creation.

Cleanup of existing duplicates and a unique index enforcing this at the database level are shipped separately in https://github.com/getlago/lago-api/pull/6129.

### DIFF
--- a/app/services/charges/create_children_service.rb
+++ b/app/services/charges/create_children_service.rb
@@ -18,6 +18,10 @@ module Charges
         # skip touching to avoid deadlocks
         Plan.no_touching do
           plan.children.where(id: child_ids).find_each do |child|
+            # NOTE: Cascade jobs are delivered at-least-once; a parent charge must
+            #       materialize at most once per child plan or usage is billed twice.
+            next if child.charges.exists?(parent_id: charge.id)
+
             create_params = if payload[:code].present?
               payload
             else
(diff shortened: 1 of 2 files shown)


app/services/charges/create_children_service.rb
```diff
@@ -18,6 +18,10 @@ def call
         # skip touching to avoid deadlocks
         Plan.no_touching do
           plan.children.where(id: child_ids).find_each do |child|
+            # NOTE: Cascade jobs are delivered at-least-once; a parent charge must
+            #       materialize at most once per child plan or usage is billed twice.
+            next if child.charges.exists?(parent_id: charge.id)
```
tiagolupepic, line 23: I have a bit of context in the create children, but having a big collection in the children, will do a N+1, no? 
Would be nice if we can avoid this.

tiagolupepic (approved):
just left a comment and it's not a blocker.

:shipit:

app/services/charges/create_children_service.rb
D1353L: Fair point. Although this service is called in an async job and the N+1 is limited to 20 indexed queries per job (batched), so the impact is limited

PR #6130, opened 2026-08-11
[ING-507] fix(alerts): lock the alert while evaluating

aquinofb:
## Context

Evaluating an alert and changing its configuration touch different tables, so they can run at the same time. Evaluation read the stored baseline and the set of thresholds without holding anything, so an edit landing in between could leave it comparing against a baseline that had moved, or against a set of thresholds only half replaced. The two sides also disagreed about what to lock: creating an alert held a lock on the wallet, the opposite order to the one evaluation needs.

## Changes

Evaluation now holds the alert while it works, and the crossing is computed inside that lock. Editing holds the same lock across the threshold rewrite. Only ever one lock is held, and it is always the alert.

Creation takes none at all: its whole write is one transaction and nothing can see the alert until the thresholds commit with it. Reading the wallet unlocked can leave the seeded baseline very slightly behind, which the next evaluation corrects, and that is a better trade than two locks taken in opposing orders.

Both delete paths now take the alert before its thresholds, and the bulk one takes them in a stable order. They used to do the reverse, which was harmless while a thresholds-only edit wrote nothing to the alert row, and would have become a deadlock the moment editing started locking that row first. The specs assert the order of the statements rather than merely that a lock was taken.

An evaluation that cannot compute a value now records the time it ran. Nothing reads that timestamp yet, but the period-rollover rule that follows keys off it.

One consequence worth knowing: the lock reloads the alert, so thresholds are always read fresh inside it. The two callers that walk a batch of alerts therefore no longer preload thresholds, because the lock was about to throw that work away for every alert in the batch.


app/services/usage_monitoring/process_lifetime_usage_alert_service.rb
```diff
@@ -27,7 +28,16 @@ def call
         usage_filters:
       )
 
-      ProcessAlertService.call(alert:, alertable: subscription, current_metrics: usage_for_charges_result.usage)
+      # The usage was filtered for one metric, and evaluation reloads the alert under its lock, so a metric
+      # changed in between would be measured against usage it does not appear in and read as no usage at all.
+      return result unless Alert.where(id: alert.id, billable_metric_id:).exists?
```
toommz, line 33: Should we put it before L22, so we don't compute usage if we will just early return?

toommz (commented):
I think we should reload the alerts in their locking blocks, because in the meantime something might have happened. WDYT?

app/services/usage_monitoring/process_lifetime_usage_alert_service.rb
aquinofb: Good question, but it has to stay after the usage build: that build is the slow part, so it is the window where a metric change or a deletion can actually land. Before it the alert has only just been loaded by the job, so the check would be re-reading a value read microseconds earlier and would catch nothing.

Two specs cover the cases it is there for, and both fail if it moves up: "when the alert's metric changes while the usage is being built" and "when the alert is deleted while the usage is being built".

The wasted-work worry is mostly handled a level up, since the alert is soft deleted out of the default scope and the job cannot deserialize a discarded one, so it fails at its boundary rather than reaching here.

The comment only explained what the check guards, not why it sits below the build, which is what made it read as misplaced. I have reworded it in a02cc89c8. Evaluation also re-checks the metric once it holds the lock now, so the narrow gap between this check and the lock is covered too.

vincent-pochet (approved):

toommz (approved):

PR #6138, opened 2026-08-12
feat(quotes): bound signing and execution dates

toommz:
## Context

Three quote dates are refused in the past by the execution flow: the subscription ending date (`end_date`, `plans[].payload.endDate`), a wallet `expirationAt` and a recurring rule `expirationAt`. They are only checked for futureness at approve, and nothing ties them to the dates that decide *when* execution happens, so a deal can be signed or scheduled past its own term and only blows up at execution as a failed order.

## Description

Adds `QuoteVersions::DealExpiration`, which folds those fields into the earliest date the deal must still be running, and bounds both write points against it:

- `expires_at` at approve (`OrderForms::CreateService`)
- `execute_at` at signing and on update, via the shared `ExecutionSettingsValidation` concern

Both fail with `after_deal_expiration`. The comparison is by date and strictly earlier, matching `valid_ending_at?`, which already fails on a same-day landing. `one_off` quotes carry none of these fields, so the check is a no-op for them.

Two known limits: an order with no `execute_at` is executed manually at any time, so the execution-time failure stays the backstop; and a short-lived wallet now constrains the signing window too.

### DIFF
--- a/app/services/order_forms/create_service.rb
+++ b/app/services/order_forms/create_service.rb
@@ -18,6 +18,10 @@ module OrderForms
       return result.single_validation_failure!(field: :quote_version, error_code: "not_approved") unless quote_version.approved?
       return result.single_validation_failure!(field: :expires_at, error_code: "invalid_date") unless valid_expires_at?
 
+      unless QuoteVersions::DealExpiration.covers?(quote_version, expires_at)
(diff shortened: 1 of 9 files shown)


aquinofb (approved):
LGTM

PR #6139, opened 2026-08-12
Fix(subscriptions): started_at for subscription_at in the past

annvelents:
## Context

We have a case where an organization handles subscriptions upgrades-downgrades manually. 
So they terminate a subscription and then create a new one with the same external_id;
depending on their internal logic, they send the same subscription_at as the previous subscirption, or subscription_at is now

however, there is already an invoice from the previous subscription for the existing billing period, so even though the subscription has subscription_at in the past, the past usage should not be shared (started_at) should be now - this can happen if sub_1 is terminated but not yet billed, sub_2 is already created and billing job picks it up

Also, having this change prevents us from double billing, if an organization already had an invoice for this customer and this external_id, having repeated started_at will result in double billing

## Description

Added a condition that checks if there is another terminated subscription with the same external_id with matching dates, if there is - started_at is previous subscription's terminated_at

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -203,24 +203,21 @@ module Subscriptions
     end
 
     # NOTE: Backdating normally means "this subscription really started then, bill it from then".
-    #       But when a previous subscription on the same external_id was terminated after that date,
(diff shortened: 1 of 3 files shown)


ancorcruz (approved):
🚀 LGTM

PR #6143, opened 2026-08-13
fix(auth): prevent cross-org account takeover via Okta SSO

ancorcruz:
Part of INT-203

## Context

Okta sign-in resolves the integration from the email domain, then exchanges the authorization code and fetches user identity from the host configured on that integration. The host is admin-configurable, and sign-in just-in-time (JIT) provisioned a membership for whatever email the host returned. An admin of any Okta-enabled organization could therefore point the host at a server they control and receive a session for an unrelated, existing user — a cross-organization account takeover.

Entra ID is unaffected: it fetches identity from a fixed Microsoft Graph endpoint rather than an admin-supplied host.

## Changes

- **Cross-org guard (the fix).** Reject Okta sign-in when the resolved email belongs to an existing user who is not an active member of the integration's organization. Brand-new users are still provisioned on the fly, and existing members keep signing in — so enabling Okta on an organization with existing members is unaffected. Adding an existing user to another organization now goes through an invite.
- **Case-insensitive userinfo email compare.** Okta does not guarantee the casing of the email it returns, so a user whose Okta email differed only in case from the address they typed was wrongly rejected. Aligns with Entra ID.
- **Host format validation (defense-in-depth).** The configured host is interpolated into the authorize/token/userinfo URLs, so restrict it to a bare host segment to prevent URL injection / SSRF. Only an explicitly configured host is validated; the derived `<organization_name>.okta.com` default and custom Okta domains such as `login.acme.com` remain valid.

If merged before #6146, existing users invited to Okta enabled organizations won't be able to accept the invite and join the organization. I'd still merge it as this fixes a security issue.

### DIFF
--- a/app/services/auth/okta/login_service.rb
+++ b/app/services/auth/okta/login_service.rb
@@ -20,6 +20,10 @@ module Auth
         query_okta_access_token
         check_userinfo(result.email)
 
+        if existing_user_outside_organization?
+          return result.single_validation_failure!(error_code: "user_does_not_belong_to_organization")
+        end
+
         find_or_create_user
         find_or_create_membership
 
@@ -40,6 +44,13 @@ module Auth
       private
 
       attr_reader :code, :state
+
+      def existing_user_outside_organization?
+        user = User.find_by(email: result.email)
+        return false if user.nil?
+
+        !user.memberships.active.exists?(organization_id: result.okta_integration.organization_id)
+      end
 
       def generate_token
         result.token = Utils::AuthToken.encode(user: result.user, login_method: Organizations::AuthenticationMethods::OKTA)
--- a/app/services/auth/okta/login_service.rb
+++ b/app/services/auth/okta/login_service.rb
@@ -27,7 +27,7 @@ module Auth
         find_or_create_user
         find_or_create_membership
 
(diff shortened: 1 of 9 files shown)


app/services/auth/okta/login_service.rb
endenis: Hmmm should this check only the organization of the `result.okta_integration` and not all active organizations of the user?

ancorcruz: fixed, also for Entra.

checks both, orga has integration enabled and user active organizations includes the organization.

app/services/auth/okta/login_service.rb
```diff
@@ -20,10 +20,14 @@ def call
         query_okta_access_token
         check_userinfo(result.email)
 
+        if existing_user_outside_organization?
```
endenis, line 23: This fixes the login service but the [Okta::AcceptInviteService](https://github.com/getlago/lago-api/blob/e941ecdfe286db4c79bac4e094616f67d12aafd0/app/services/auth/okta/accept_invite_service.rb#L15-L30) is still vulnerable to this attack

ancorcruz: thanks for raising,

I have added a guard clause to ensure that flow is now safe. Existing users will have to authenticate first with the changes you have made in #6146

endenis (approved):

PR #6146, opened 2026-08-13
fix(invitation): handle existing users

endenis:
## Context

Accepting an invitation does not handle existing users correctly.

## Description

- Add a flow for an email that already has an account. It requires the password of that account and only adds the membership.
- Authenticated users accept with the new `joinOrganization` mutation.
- `acceptInvite` keeps `email` as an optional ignored argument, so this backend fix can be deployed before the [frontend fix](https://github.com/getlago/lago-front/pull/4125).


app/models/user.rb
```diff
@@ -25,6 +25,10 @@ class User < ApplicationRecord
   def can?(permission, organization:)
     memberships.find { |m| m.organization_id == organization.id }&.can?(permission)
   end
+
+  def login_method_allowed?(login_method)
+    active_organizations.pluck(:authentication_methods).flatten.uniq.include?(login_method)
```
ancorcruz, line 30: if user belongs to two organizations;

org A allows password/google authentication methods, 
org B allows okta only (or whatever authentication method that is not listed in orga A allowed ones).

should user be able to login and have access to both organizations (A and B) when authenticating with a method that is only allowed by one of them?

endenis: @ancorcruz good question!

Right now it's already the case in backend on production: users can access both organizations even when only one of them allows password authentication. I am not looking into changing this behavior as part of this fix.

Long term we should consider limiting access by used sign in method in both backend and frontend.
The frontend already partially prevents this using `accessibleByCurrentSession`, but it does not cover every route and the backend does not enforce it.

PR #6150, opened 2026-08-14
Fix(members): make members searchable

annvelents:
## Context

Before we had all the list of members loaded at once for the billing entity. Now they are paginated, so the search should be on the BE

## Description

Added search and filter to membership_resolver

### DIFF
--- a/app/queries/invites_query.rb
+++ b/app/queries/invites_query.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+class InvitesQuery < BaseQuery
+  Result = BaseResult[:invites]
+  Filters = BaseFilters[:role_ids]
+
+  def call
+    invites = base_scope
+
+    invites = with_role_ids(invites) if filters.role_ids.present?
+
+    invites = paginate(invites)
+    invites = apply_consistent_ordering(invites)
+
+    result.invites = invites
+    result
+  end
+
+  private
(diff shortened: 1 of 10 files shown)


jdenquin:
do we have all the needed index?

app/queries/invites_query.rb
```diff
@@ -0,0 +1,39 @@
...
+    invites = apply_consistent_ordering(invites)
+
+    result.invites = invites
+    result
+  end
+
+  private
+
+  def base_scope
+    scope = organization.invites.pending
```
annvelents, line 22: note: `organization.invites.pending` will never be a big scope

app/queries/invites_query.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    return scope if search_term.blank?
+
+    scope.where("invites.email ILIKE ?", "%#{Invite.sanitize_sql_like(search_term)}%")
```
endenis, line 26: Should we also add a trigram index on `invites.emails`?

endenis (approved):

app/queries/invites_query.rb
annvelents: since base scope is pending invites, and it will never be a big table (because all invites are getting to be real users), I don't think it's actually needed

PR #6153, opened 2026-08-14
Ing 436

mariohd:
## Context

Support the `code` routing key and per-category default (`is_default`) on integration connections (tax / accounting / CRM), extending the existing declarative `integration_customers` array on the customer. The `code`, `is_default` and `category` columns and their indexes already landed; this wires the input, processing and output layers on top.

## Description

Accept `code` + `is_default` on each `integration_customers` element over both REST and GraphQL, store `is_default` on the connection row (partial-unique per `(customer, category)`), derive `category` from the STI type, and auto-derive `code` from the integration when absent. Ungated, one connection per type for now.

The change is split into four commits by concern:

- **Model** — `before_validation` callbacks on `IntegrationCustomers::BaseCustomer` derive `category` from the STI type and default `code` to the integration's code when none is given. Centralizing this on the model keeps every creation path consistent (declarative array, sync services, dedicated connection mutation) and lets callers omit both.
- **Service** — the integration customers `BaseService` reads `code` and `is_default`, and `CreateService` (both the link and sync paths) and `UpdateService` persist any explicit values. Derivation still happens on the model, so only explicit values flow through the services.
- **REST** — permit `code` and `is_default` on each element in the customers controller, and expose `code`, `is_default` and `category` in the integration customer serializer so the values round-trip.
- **GraphQL** — add `code` and `isDefault` to the shared `Types::IntegrationCustomers::Input` and regenerate the schema (GraphQL output types already expose all three).

## Decisions worth a reviewer's eye

- **Derivation lives on the model**, not in the services — DRY, and it also covers the provider-sync services (Netsuite/Xero/etc.) that create rows directly. The rules match the ING-452 backfill and the dedicated `CreateConnectionService` exactly (`code ← integration.code`, `category ← category_for(type)`).
- **No auto-defaulting of `is_default` in the array flow** — it honors an explicit value but does not auto-mark the first connection as default. That auto-defaulting belongs to the dedicated per-connection service, and Milestone 1 downstream routing still uses the type-based helpers (`tax_customer`, etc.), so it isn't needed here. Existing single connections were defaulted by the ING-452 backfill.
- **Presence validation deferred** — the scope calls for `code` presence validation "after backfill", so `code` stays `allow_nil` for now; auto-derivation ensures it is populated going forward.
- **Anrok/Salesforce array-update caveat** — `UpdateService` early-returns for those two types before saving, so `code`/`is_default` changes via the array don't apply to them (pre-existing behaviour); they would go through the dedicated connection mutation. Left as-is to avoid changing their sync flow.

## Tests

Model callbacks, `CreateService` honoring explicit `code`/`is_default`, and serializer output are covered. Broad regression across the integration services, customer create/upsert/update and the GraphQL customer mutations passes; Rubocop clean.

### DIFF
--- a/app/graphql/types/integration_customers/input.rb
+++ b/app/graphql/types/integration_customers/input.rb
@@ -7,10 +7,12 @@ module Types
 
       argument :id, ID, required: false
 
+      argument :code, String, required: false
       argument :external_customer_id, String, required: false
       argument :integration_code, String, required: false
       argument :integration_id, ID, required: false
       argument :integration_type, Types::Integrations::IntegrationTypeEnum, required: false
+      argument :is_default, Boolean, required: false
       argument :subsidiary_id, String, required: false
       argument :sync_with_provider, Boolean, required: false
       argument :targeted_object, Types::Integrations::Hubspot::TargetedObjectsEnum, required: false
--- a/app/graphql/types/integration_customers/input.rb
+++ b/app/graphql/types/integration_customers/input.rb
@@ -12,7 +12,6 @@ module Types
       argument :integration_code, String, required: false
       argument :integration_id, ID, required: false
       argument :integration_type, Types::Integrations::IntegrationTypeEnum, required: false
-      argument :is_default, Boolean, required: false
       argument :subsidiary_id, String, required: false
       argument :sync_with_provider, Boolean, required: false
       argument :targeted_object, Types::Integrations::Hubspot::TargetedObjectsEnum, required: false
--- a/app/services/integration_customers/create_service.rb
+++ b/app/services/integration_customers/create_service.rb
@@ -18,6 +18,8 @@ module IntegrationCustomers
       end
       return res if res&.error
 
+      assign_routing_attributes! if result.integration_customer
+
       result
     rescue ActiveRecord::RecordNotUnique
       # Avoid raising on race conditions when multiple requests are made at the same time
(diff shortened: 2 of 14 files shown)


app/graphql/types/integration_customers/input.rb
```diff
@@ -7,10 +7,12 @@ class Input < Types::BaseInputObject
 
       argument :id, ID, required: false
 
+      argument :code, String, required: false
       argument :external_customer_id, String, required: false
       argument :integration_code, String, required: false
       argument :integration_id, ID, required: false
       argument :integration_type, Types::Integrations::IntegrationTypeEnum, required: false
+      argument :is_default, Boolean, required: false
```
lovrocolic: based on latest designs, `is_default` is not a parameter. For setting as default, we will use dedicated mutation.

I would also remove it from the api side

app/services/integration_customers/create_service.rb
```diff
@@ -31,6 +33,16 @@ def call
 
     attr_reader :customer
 
+    # Honor an explicit routing code / default from the input; category and a fallback code are
+    # derived on the model. The connection is created by link/sync first, then restated here so
+    # both paths behave the same.
```
lovrocolic: Do we need those comments? IMO method is already readable

app/graphql/types/integration_customers/input.rb
lovrocolic: First connection os a type can be default, otherwise it should be false

app/services/integration_customers/update_service.rb
lovrocolic, line 18: I think we should move assignments above those checks. Currently new attributes cannot be assigned for Anrok or Salesforce

lovrocolic:
Can we please also add mutation spec? wdyt?

app/services/integration_customers/update_service.rb
mariohd: makes total sense!

app/services/integration_customers/update_service.rb
```diff
@@ -10,10 +10,13 @@ def initialize(params:, integration:, integration_customer:)
     def call
       result = super
       return result if result.error
+      return result.not_found_failure!(resource: "integration_customer") unless integration_customer
+
+      integration_customer.code = code if code.present?
+      integration_customer.save! if integration_customer.changed?
```
lovrocolic: Do we need `integration_customer.changed?`? Also in create service.

If we conditionally call save!, in some cases we would never call before_validation hooks

lovrocolic (approved):

PR #6154, opened 2026-08-14
feat(invoices): populate search_terms from the services

toommz:
2/4 of a stack improving invoice list search performance. Stacked on #6152.

Fills `search_terms` from the services that write the underlying data, rather than from model callbacks:

- `Invoices::RefreshSearchTermsService` at invoice creation, finalization and regeneration
- `Customers::RefreshInvoicesSearchTermsJob` from the three services that persist customer names or emails

From here every new or updated invoice is correct; existing rows stay NULL until the backfill.

Next in the stack: backfill, read switch behind a flag.

### DIFF
--- a/app/jobs/customers/refresh_invoices_search_terms_job.rb
+++ b/app/jobs/customers/refresh_invoices_search_terms_job.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Customers
+  class RefreshInvoicesSearchTermsJob < ApplicationJob
+    queue_as "default"
+    unique :until_executed
+
+    def perform(customer_id)
+      customer = Customer.with_discarded.find_by(id: customer_id)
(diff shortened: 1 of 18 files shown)


toommz:
> 
> HOLD: Persisted invoice search terms conflict with the established discarded-customer search behavior.
> 
> * Exclude discarded customer fields and refresh on both discard and restore, matching the existing PostgreSQL/Meilisearch behavior; add regression coverage for both transitions.
> * Add behavior coverage for the newly modified portal update, API upsert, and regenerated-invoice paths, including unchanged customer fields and the resolved purchase order number.

Yes but… The previous behavior was a bug. We even had a customer asking to be able to search by a soft-deleted customer name in the invoice list. Product team is aligned.

vincent-pochet (approved):

PR #6155, opened 2026-08-14
feat(invoices): backfill search_terms on existing invoices

toommz:
3/4 of a stack improving invoice list search performance. Stacked on #6154.

Fills `search_terms` on invoices that existed before the write path shipped. A migration enqueues one self-chaining job that walks `invoices.id`; a short batch means the walk is over, so the last job checks for any remaining NULL and, if it finds some, runs one more full pass before giving up and reporting.

Two deliberate deviations:

- `sidekiq_options retry: 5`, overriding `ApplicationJob`'s `retry: 0`. The chain is self-terminating, so a single unretried failure would silently end the whole backfill. Five puts a permanently broken job in the morgue after ~11 minutes rather than the ~4.5 hours a higher count would cost.
- `on_conflict: :log`, because a duplicate enqueue otherwise raises `JobNotUnique` and aborts the migration, which would fail the deploy.

Next in the stack: the read switch behind a feature flag.

### DIFF
--- a/app/jobs/database_migrations/backfill_invoices_search_terms_job.rb
+++ b/app/jobs/database_migrations/backfill_invoices_search_terms_job.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+module DatabaseMigrations
+  class BackfillInvoicesSearchTermsJob < ApplicationJob
+    queue_as :low_priority
+    unique :until_executed, on_conflict: :log, lock_ttl: 30.minutes
+
+    # ApplicationJob sets retry: 0. The chain is self-terminating, so one failed batch
+    # would silently end the whole backfill.
+    sidekiq_options retry: 5
+
+    BATCH_SIZE = 5_000
+    MAX_PASSES = 2
+
+    def perform(cursor = nil, pass = 1)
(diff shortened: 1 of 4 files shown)


db/migrate/20260814213522_enqueue_backfill_invoices_search_terms.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class EnqueueBackfillInvoicesSearchTerms < ActiveRecord::Migration[8.0]
```
jdenquin: I think it should be a rake task instead of a job in a migration.
It can block the whole deployment and lead to errors.
Thats better to handle it manually

toommz: @jdenquin 👍 Removed the migration and added a [task](https://github.com/getlago/lago-api/pull/6155/changes#diff-fcd3e59b9061170f41df59b94fc1ba79e69db4bd14ef91829db0bb76a1dc92ec)

vincent-pochet (approved):

PR #6156, opened 2026-08-14
feat(invoices): read the invoice search from search_terms behind a flag

toommz:
4/4 of a stack improving invoice list search performance. Stacked on #6155.

Adds the per-organization `invoice_search_terms` flag. When it is on, `InvoicesQuery` replaces the `matching_customers` / `matching_invoices` CTE union with a single `search_terms ILIKE` predicate. The CTE stays as the fallback for organizations that are not flipped, and an organization should only be flipped once the backfill has completed for it.

One intentional difference: `legal_name` and `purchase_order_number` become searchable, which aligns the Postgres path with the Meilisearch one.

> [!CAUTION]
> The feature flag should not be activated before the backfill in #6155 has been run!

### DIFF
--- a/app/queries/invoices_query.rb
+++ b/app/queries/invoices_query.rb
@@ -198,11 +198,27 @@ class InvoicesQuery < BaseQuery
   def base_scope
     scope = organization.invoices
     return scope if search_term.blank?
+    return search_terms_scope(scope) if search_terms_enabled?
 
     scope = scope.with(matching_customers: matching_customers) if search_customers?
     scope
       .with(matching_invoices: matching_invoices)
       .where("invoices.id IN (SELECT id FROM matching_invoices)")
+  end
+
+  # The customer portal passes a Customer as `organization`, and only Organization
+  # carries feature flags.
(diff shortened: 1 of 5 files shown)


app/queries/invoices_query.rb
```diff
@@ -198,13 +198,29 @@ def filters_contract
...
+    return search_terms_scope(scope) if search_terms_enabled?
 
     scope = scope.with(matching_customers: matching_customers) if search_customers?
     scope
       .with(matching_invoices: matching_invoices)
       .where("invoices.id IN (SELECT id FROM matching_invoices)")
   end
 
+  # The customer portal passes a Customer as `organization`, and only Organization
+  # carries feature flags.
```
vincent-pochet, line 210: Nice :sweat_smile:

vincent-pochet (approved):

app/queries/invoices_query.rb
toommz: Variable name typing always had its own flaws 😆

PR #6157, opened 2026-08-14
fix(charge_filters): cascade by filter code

brunomiguelpinto:
## Context

Charge filters were given a code to serve as their identity, precisely because their values move:
removing a value from a billable metric trims it out of every filter using it, so two filters that
were distinct collapse onto the same values and stop being distinguishable.

The cascade never used it. It still addressed the copies on overridden plans by their values, which
is the thing the code was introduced to replace, so the ambiguity remained: whichever filter
happened to sit on the predicate was the one repriced or discarded. And resolving a whole batch
through `index_by` silently kept one of the two, so deleting a filter on the plan left the duplicate
on the override behind.

## Description

The code now travels on every cascade job. `FilterCascadable` was dropping it, which meant the
single-filter REST and GraphQL surface ran entirely on the predicate no matter what the filters held.

Children are resolved by code, and the predicate is asked only about the ones the code could not
reach — so it is asked about none once every filter has one, and the legacy path is observably dead
before anyone deletes it.

### One rule per action

- **create** takes the widest match. Anything already on the predicate means there is nothing to add,
  and missing it would leave a duplicate. A filter that got there before the plan had one adopts the
  plan's code: this is the only moment the two can be said to be the same, and left unlinked it
  would be unreachable for good. A code of its own means the opposite — born on the override, nobody's
  copy — so it keeps it. Two filters on the predicate raise `DuplicatePredicate` instead: adopting
  into one picked by row order would settle which of the two bills from then on.
- **update** raises `MissingParentCode` rather than guess. Guessing wrong reprices a filter the
  customer negotiated, and doing nothing quietly leaves the override on the old price, which is the
  one failure nothing else in the system would surface. It stays quiet when the charge has no
  override with an active subscription, since a missing code costs nothing where there is nothing to
  cascade to.
- **destroy** does nothing without the code. Discarding the wrong filter takes billing config away
  from a customer irreversibly, whereas a filter left behind is already listed by the report below.

A filter created straight on an override needs none of this: `generate_code` is deterministic on the
values, so the plan derives the very same code when it later prices that segment and the two pair up
by construction rather than by adoption.

### What changes for filters still without a code

Two new exceptions, both raised inside the job after the user's request has already succeeded, and
`ApplicationJob` does not retry — so one occurrence is one dead job naming what to fix, not twenty.

- `MissingParentCode` — the plan's filter has no code, so a reprice no longer reaches the override.
  `rake filters:report_parent_codes` lists them ahead of time.
- `DuplicatePredicate` — an override holds two filters on the predicate. A plan gaining a filter
  fails for every override in that batch, not just the affected one. Cleaning the duplicate and
  retrying the job recovers it; the retry is idempotent, since copies already made are found by
  their code and skipped.

Both shrink as the codeless filters are cleaned up and the backfill is re-run, which is what
`upgrade:perform_required_jobs` does — a charge cleaned by hand goes back to having filters without
a code and the walk selects it again.

## Known issue, surfaced while reviewing this

> [!WARNING]
> **A price negotiated on an override is lost when the plan widens a filter's values.** Widening
> replaces the filter rather than editing it — `CreateOrUpdateBatchService` pairs the payload against
> the existing filters by their values, finds no match, and creates a new one while discarding the
> old. The replacement is created with the plan's price, so `filter_customized?` never gets to guard
> it and the override silently starts charging the plan's rate.
>
> ```
> child, negotiated   {region: [us]}       @2
> plan widens to      {region: [us, eu]}   @10
> child, after        {region: [us, eu]}   @10    ← the negotiated 2 is gone
> ```
>
> **This is not introduced here.** Verified against `main`, same failure: a changed predicate was a
> removal plus an addition there too. It is not fixed here either — it needs the plan payload to be
> able to name a filter by its `lago_id`, which the API already returns and which charges in the same
> payload already accept. Then widening becomes an update.
>
> It is pinned as a **pending spec** rather than left to a ticket, so it fails loudly the day someone
> fixes it, and lives where whoever edits the file will see it.

### Rake tasks

- `filters:report_parent_codes` — new. The plan's own filters with no code, on charges that have
  overrides, grouped by organization and by charge.
- `filters:report_codes` — every filter with no code, and the charges holding the most of them.

### DIFF
--- a/app/services/charge_filters/cascade_dispatcher.rb
+++ b/app/services/charge_filters/cascade_dispatcher.rb
@@ -33,9 +33,9 @@ module ChargeFilters
       before_by_code.each_value { enqueue_destroy(charge, it) }
     end
 
-    # The legacy path, for the filters the backfill left without a code. Two of them can share
-    # a predicate once a metric change shortened them onto it, and this keeps one — the same
-    # loss the cascade service takes on the same filters. Delete it once none are left.
+    # The legacy path, for the filters the backfill left without a code. Two of them can share a
+    # predicate once a metric change shortened them onto it, and this keeps one: nothing here can
+    # tell them apart. Delete it once no filter is left without a code.
     def diff_by_predicate(charge, before, after)
       before_by_values = before.index_by { it[:values] }
 
--- a/app/services/charge_filters/cascade_dispatcher.rb
+++ b/app/services/charge_filters/cascade_dispatcher.rb
@@ -9,38 +9,63 @@ module ChargeFilters
     #   { values: {"key" => [...]}, properties: {...} | nil, invoice_display_name: "...",
     #     code: "..." | nil }
     # `values` and `properties` must be string-keyed.
+    #
+    # A plan edit never gives a filter a code it did not have, so a filter is either coded on
+    # both sides of the diff or on neither, and the split cannot cut one in half.
     def call(charge:, before:, after:)
-      before_by_values = before.index_by { |f| f[:values] }
+      coded_before, plain_before = before.partition { it[:code].present? }
+      coded_after, plain_after = after.partition { it[:code].present? }
+
+      diff_by_code(charge, coded_before, coded_after)
+      diff_by_predicate(charge, plain_before, plain_after)
+    end
+
+    def diff_by_code(charge, before, after)
+      before_by_code = before.index_by { it[:code] }
 
       after.each do |new_filter|
-        existing = before_by_values.delete(new_filter[:values])
+        previous = before_by_code.delete(new_filter[:code])
+        next if previous && unchanged?(previous, new_filter)
 
-        next if existing &&
-          existing[:properties] == new_filter[:properties] &&
-          existing[:invoice_display_name] == new_filter[:invoice_display_name]
-
-        ChargeFilters::CascadeJob.perform_later(
-          charge.id,
-          existing ? "update" : "create",
-          new_filter[:values],
-          existing&.dig(:properties),
-          new_filter[:properties],
-          new_filter[:invoice_display_name],
-          new_filter[:code]
-        )
+        enqueue_change(charge, previous ? "update" : "create", new_filter, previous&.dig(:properties))
       end
 
-      before_by_values.each_value do |old|
-        ChargeFilters::CascadeJob.perform_later(
-          charge.id,
-          "destroy",
-          old[:values],
-          old[:properties],
-          nil,
-          old[:invoice_display_name],
-          old[:code]
-        )
+      before_by_code.each_value { enqueue_destroy(charge, it) }
+    end
+
+    # The legacy path, for the filters the backfill left without a code. Two of them can share
+    # a predicate once a metric change shortened them onto it, and this keeps one — the same
+    # loss the cascade service takes on the same filters. Delete it once none are left.
+    def diff_by_predicate(charge, before, after)
+      before_by_values = before.index_by { it[:values] }
+
+      after.each do |new_filter|
+        previous = before_by_values.delete(new_filter[:values])
+        next if previous && unchanged?(previous, new_filter)
+
+        enqueue_change(charge, previous ? "update" : "create", new_filter, previous&.dig(:properties))
       end
+
+      before_by_values.each_value { enqueue_destroy(charge, it) }
+    end
+
+    def unchanged?(previous, new_filter)
+      previous[:properties] == new_filter[:properties] &&
+        previous[:invoice_display_name] == new_filter[:invoice_display_name]
+    end
+
+    def enqueue_change(charge, action, filter, old_properties)
+      ChargeFilters::CascadeJob.perform_later(
+        charge.id, action, filter[:values], old_properties,
+        filter[:properties], filter[:invoice_display_name], filter[:code]
+      )
+    end
+
+    def enqueue_destroy(charge, filter)
+      ChargeFilters::CascadeJob.perform_later(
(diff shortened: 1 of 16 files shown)


app/services/charge_filters/cascade_service.rb
```diff
@@ -55,11 +55,24 @@ def child_ids
         .distinct.pluck(:id)
     end
 
+    # The code says which filter this is. The predicate is only a guess at it, and the
+    # legacy path until every filter has a code — so it is asked about the children the
+    # code could not reach, and about none once they all have one.
+    def child_filters_by_charge(batch_child_ids)
+      by_code = child_filters_holding_parent_code(batch_child_ids)
+      remaining = batch_child_ids.reject { by_code.key?(it) }
+      by_predicate = matching_child_filters(remaining)
```
brunomiguelpinto, line 84: should we do this or clean all duplicates first and drop this path ?

app/services/charge_filters/cascade_service.rb
```diff
@@ -79,6 +92,14 @@ def matching_child_filters(batch_child_ids)
         .index_by(&:charge_id)
```
brunomiguelpinto: i think i should left index but only select stuff without code ( this way we dont destroy something we already fixed ) becouse the this is not suposed to have more than one and if no code means its not fixed yet on the parent

app/services/charge_filters/cascade_dispatcher.rb
```diff
@@ -9,38 +9,63 @@ module CascadeDispatcher
...
-          charge.id,
-          existing ? "update" : "create",
-          new_filter[:values],
-          existing&.dig(:properties),
-          new_filter[:properties],
-          new_filter[:invoice_display_name],
-          new_filter[:code]
-        )
+        previous = before_by_code.delete(new_filter[:code])
+        next if previous && unchanged?(previous, new_filter)
```
annvelents, line 28: what will happen if a filter exists (with code assigned), and a new filter value has been added?

brunomiguelpinto: PARENT BEFORE:  id 8104ae39  code region_us_7ae73973      {region: [us]}
CHILD  BEFORE:              code region_us_7ae73973      {region: [us]}

PARENT AFTER:   id f3bdd042  code region_eu_us_d6cbc2d1   {region: [us, eu]}
CHILD  AFTER:               code region_eu_us_d6cbc2d1   {region: [us, eu]}

with_discarded: region_us_7ae73973 → deleted_at preenchido
SAME RECORD? false

ops it deletes/creates instead of update

spec/services/charge_filters/cascade_service_spec.rb
```diff
@@ -200,14 +270,120 @@
...
+
+        context "when it already has a code of its own" do
+          let(:existing) do
+            filter = create(:charge_filter, charge: child_charge, properties: {amount: "20"}, code: "its_own_code_1a2b3c4d")
+            create(:charge_filter_value, charge_filter: filter, billable_metric_filter: bm_filter, values: ["eu"])
+            filter
+          end
+
+          # Born on the override with an identity of its own, so it is nobody's copy
+          it "leaves the code alone" do
```
annvelents, line 327: but will it create a new filter then?

annvelents (approved):

annvelents (approved):
🚀  🐐

spec/services/charge_filters/cascade_service_spec.rb
```diff
@@ -200,14 +270,127 @@
...
+          expect { service }.not_to change { child_charge.filters.reload.count }
+        end
+
+        it "leaves the copy untouched" do
+          expect { service }.not_to change { copy.reload.updated_at }
+        end
+      end
+
+      # The lookup ran before the other job committed, so the insert is the first thing to see
+      # the conflict. Left to fail on purpose: the retry finds the copy and passes, and a
```
endenis: Hmmm does the CascaseJob actually retry on the `ActiveRecord::RecordNotUnique` error? I don't think it does 🤔

app/services/charge_filters/cascade_service.rb
```diff
@@ -55,11 +69,30 @@ def child_ids
...
+    # The code says which filter this is. The predicate is only a guess at it, and the
+    # legacy path until every filter has a code — so it is asked about the children the
+    # code could not reach, and about none once they all have one.
+    def child_filters_by_charge(batch_child_ids)
+      by_code = child_filters_holding_parent_code(batch_child_ids)
+
+      # Only a create still guesses, and it has to: anything already sitting on the predicate
+      # means there is nothing to add, and missing it would leave a duplicate. An update or a
+      # destroy acts on one filter, where acting on the wrong one is worse than not acting.
+      return by_code unless action == "create"
```
rsempe, line 81: can we have a child copy that never got a code yet?

rsempe (approved):

app/services/charge_filters/cascade_service.rb
brunomiguelpinto: Yes, structurally. The backfill only ever copies a code down from the parent, and it deliberately skips three cases: the override holds a filter the plan does not have, the plan's charge was deleted, or the predicate is held twice on either side.

but we cleaned all duplicates and now we dont have any child's without code so it will not be an issue now

spec/services/charge_filters/cascade_service_spec.rb
brunomiguelpinto: no the comment is wrong  fixing

PR #6160, opened 2026-08-16
[ING-504] fix(alerts): reject duplicate wallet codes

aquinofb:
> Split out of #6082, which is parked until the index build can be scheduled. This half needs no migration and can go in now.

## Context

A wallet alert code is meant to be unique per wallet. The API reference documents it that way, the subscription side is held to it by a unique index, and both the dashboard and the public API address a single alert by its code. Nothing enforces it for wallets, so the same code can be used twice on one wallet, and reading, updating or deleting by that code then reaches whichever of the two the database happens to return.

The index that would settle this is a full scan of a very large table, so it has to run as a scheduled operation rather than inside a deploy. That is what #6082 is waiting for. Meanwhile nothing stops new duplicates being created, and each one adds to the eventual cleanup.

## Changes

Creating or renaming a wallet alert now refuses a code already used on that wallet.

The check is deliberately narrow. A clash with an alert of the same type is left alone, because a unique index already refuses that and returns `alert_already_exists`, which callers may be relying on. Only a clash across two different types reaches the new rule, and that is exactly the case nothing catches today. Discarded alerts do not reserve their code, matching the partial index used on the subscription side.

This covers the create and update services, which is every path the API and the dashboard use. Two simultaneous writes can still slip a duplicate through, since only a unique index rules that out, so this reduces the problem rather than ending it. It also stops the eventual backfill from growing while the index waits for a window.


app/services/usage_monitoring/update_alert_service.rb
```diff
@@ -42,6 +42,10 @@ def call
       billable_metric = find_billable_metric_from_params!
       return result unless result.success?
 
+      if params.key?(:code) && wallet_alert_code_taken?(wallet_id: alert.wallet_id, code: params[:code], alert_type: alert.alert_type, excluding_id: alert.id)
```
toommz, line 45:
```suggestion
      if wallet_alert_code_taken?(wallet_id: alert.wallet_id, code: params[:code], alert_type: alert.alert_type, excluding_id: alert.id)
```

`#wallet_alert_code_taken?` already filters blank codes so let's let it do its work.

toommz (approved):

PR #6161, opened 2026-08-17
[ING-453] feat(connections): payment connection model helpers

mariohd:
## Context

The multiple-connections work needs the payment provider customer model to understand connection codes and a reserved manual (null-provider) connection, plus customer-level helpers to resolve the active payment connection and report its status. The schema (code, is_default) and its indexes are already in place; this adds the model behaviour on top.

## Description

On PaymentProviderCustomers::BaseCustomer, add a by_code scope, a manual? predicate identifying the reserved base-class row coded "manual", and a validation rejecting the reserved manual code on any provider-backed connection. Expose MANUAL_CODE and CONNECTION_STATUSES constants.

On Customer, add payment_connection, which resolves a connection by code and otherwise falls back to the default connection, and payment_connection_status, returning not_connected, manual or connected based on the default connection.

### DIFF
--- a/app/models/payment_provider_customers/base_customer.rb
+++ b/app/models/payment_provider_customers/base_customer.rb
@@ -52,19 +52,18 @@ module PaymentProviderCustomers
       get_from_settings("payment_method_id") || get_from_settings("provider_mandate_id")
     end
 
-    # A manual payment row is the reserved null-provider connection: a base-class row (no provider
-    # subclass) coded "manual".
+    # A manual payment row is the reserved null-provider connection (no payment provider) coded "manual".
     def manual?
-      instance_of?(BaseCustomer) && code == MANUAL_CODE
+      payment_provider_id.nil? && code == MANUAL_CODE
     end
 
     private
 
     # The "manual" code is reserved for the null-provider manual connection; a provider-backed
-    # connection (any subclass) cannot use it.
+    # connection cannot use it.
     def code_is_not_reserved
       return unless code == MANUAL_CODE
-      return if instance_of?(BaseCustomer)
+      return if manual?
 
       errors.add(:code, "value_is_reserved")
     end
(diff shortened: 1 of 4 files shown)


app/models/payment_provider_customers/base_customer.rb
```diff
@@ -41,6 +51,23 @@ def require_provider_payment_id?
     def legacy_provider_method_id
       get_from_settings("payment_method_id") || get_from_settings("provider_mandate_id")
     end
+
+    # A manual payment row is the reserved null-provider connection: a base-class row (no provider
+    # subclass) coded "manual".
+    def manual?
+      instance_of?(BaseCustomer) && code == MANUAL_CODE
```
lovrocolic: Do we need `instance_of?(BaseCustomer)`?

app/models/payment_provider_customers/base_customer.rb
```diff
@@ -41,6 +51,23 @@ def require_provider_payment_id?
...
+    end
+
+    private
+
+    # The "manual" code is reserved for the null-provider manual connection; a provider-backed
+    # connection (any subclass) cannot use it.
+    def code_is_not_reserved
+      return unless code == MANUAL_CODE
```
lovrocolic, line 65: can we use `manual?` method here?

app/models/payment_provider_customers/base_customer.rb
```diff
@@ -11,6 +11,14 @@ class BaseCustomer < ApplicationRecord
 
     self.table_name = "payment_provider_customers"
 
+    MANUAL_CODE = "manual"
```
lovrocolic: what do you think about naming it "lago_manual"? That way we are sure that it does not conflict with code that is generated by the users

mariohd: solved!

lovrocolic (approved):

PR #6164, opened 2026-08-17
feat(subscriptions): Allow cancelling incomplete subscriptions

ancorcruz:
## Context

A payment-gated subscription sits in `incomplete` while its activation payment is pending. Until now there was no way out of that state on demand: the subscription activated when the payment succeeded, or cancelled itself when the payment failed or the activation rule timed out. A merchant watching a subscription that will clearly never activate had nothing to do but wait.

Terminating one was refused outright with `subscription_incomplete`, and the REST endpoint could not even find it, since it only ever looked up `active` and `pending` records.

## Description

Let a merchant cancel an incomplete subscription, and cancel the pending payment intent with the provider on a best-effort basis — Stripe only allows cancellation in certain statuses, so this is attempted, not guaranteed.

Cancellation reuses the surface that already cancels pending subscriptions: a `DELETE` with `status=incomplete`, and the existing `terminateSubscription` mutation. Termination behaviours sent along with the request are ignored, because an incomplete subscription has never been billed for a period: there is nothing to invoice and nothing to credit. It emits `subscription.canceled` and records `cancellation_reason: manual`, distinguishing a deliberate cancellation from one Lago performed itself after a payment failure or a timeout.

The teardown this needs already existed for the timeout path, so it moves into a service both paths share. That service now takes the gating invoice whatever state the tax provider left it in. Previously it assumed the invoice was always open, which is false when a tax provider failure leaves it failed: cancelling such a subscription raised, reaching the merchant as a misleading not-found error.

## Drive-by fix

The same assumption made `Clock::ExpireIncompleteSubscriptionsJob` raise on a tax-failed gated subscription whose rule had expired. Expiry jobs do not retry, so those subscriptions were dead-lettering rather than cancelling. Sharing the fixed teardown resolves that too.

Closing a failed gating invoice also matters on its own: left alone, it can be retried and reopened against a subscription that is no longer activating.

Part of INT-214


app/services/invoices/subscription_service.rb
osmarluz: @ancorcruz the point raised by the agent review makes sense:

> The new outer transaction in Invoices::SubscriptionService absorbs its existing fee transaction. Because perform_call converts many exceptions into failed results, those errors no longer escape the outer transaction and partial fee/invoice writes can commit instead of rolling back. Preserve the prior failure rollback behavior and test an activation-billing failure.

One way of dealing with that is having a `require_new: true` so we get a new savepoint and restore the previous behavior in terms of what gets saved when a failure occurs:

```suggestion
      fee_result = ActiveRecord::Base.transaction(requires_new: true) do
```

osmarluz:
@ancorcruz this other point raised by the agent also makes sense:

>A cancellation can run after activation invoicing enqueues Invoices::Payments::CreateJob but before that job creates a payment. Invoices::Payments::CreateService does not reject the now-closed invoice and can still charge it. Make queued payment creation a no-op after cancellation and cover this ordering in a regression test.

`should_process_payment?` on `app/services/invoices/payments/create_service.rb` only refuses `payment_succeeded?` and `voided?` invoices , never `closed?` ones

app/services/subscriptions/terminate_service.rb
osmarluz, line 1: @ancorcruz this one is conditioned to the frontend behavior, but just raising that potentially a canceled subscription could be accepted by `app/graphql/mutations/subscriptions/terminate.rb` and there's not guard here as well against that

osmarluz:
@ancorcruz this one seems less likely to happen, but worth having a look:

> Closing a failed gating invoice does not synchronize with Invoices::RetryService; a concurrent retry can overwrite closed with open after cancellation. Coordinate retry with the same lock (or recheck under a lock) and add coverage for that race.

osmarluz (approved):
🚀

PR #6165, opened 2026-08-17
perf(invoices): cap the graphql invoice list count

toommz:
With the rows query fast, the pagination `COUNT` is what is left: ~16s on a 12M-invoice organization, because counting every match means finding every match, and a lossy trigram index has to visit the heap to verify each one.

The GraphQL invoice list now uses Kaminari `without_count` plus a count capped at 10,000, behind the existing per-organization `invoice_search_terms` flag. On 12M invoices, a term matching every row goes from 123s to 21ms.

The collection metadata gains two fields: `totalCountCapped`, for the frontend to render "10,000+" rather than a flat "10,000", and `hasNextPage`, which stays exact because `without_count` fetches one extra row, so paging continues past the cap. A result set landing exactly on the cap is reported as exact, since the count goes one past it.

REST keeps the exact count: integrations page it to completion, and it is already cached for 30 minutes per organization and filter set there.

Frontend PR: https://github.com/getlago/lago-front/pull/4148

### DIFF
--- a/app/graphql/mutations/invoices/finalize_all.rb
+++ b/app/graphql/mutations/invoices/finalize_all.rb
@@ -11,7 +11,7 @@ module Mutations
       graphql_name "FinalizeAllInvoices"
       description "Finalize all draft invoices"
 
-      type Types::Invoices::Object.collection_type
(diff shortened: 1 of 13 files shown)


annvelents (approved):
🚀

PR #6173, opened 2026-08-18
fix(billing): avoid overlaping on anniversary charges periods

vincent-pochet:
## Context

Charges periods for anniversary subscriptions did not tile: consecutive periods could overlap or leave a gap, and the period returned could slide depending on when it was asked for. Bounds are inclusive, so an overlapping day is billed twice and a gap day never billed. It happens when the subscription day has to be clamped to a shorter month — yearly on 29 Feb, quarterly and semiannual dated 29 or later. Calendar, weekly and monthly are unaffected. One production subscription is affected, on a non-production account.

The common cause is resolving a period from the billing date rather than from the anniversary that opens it. A yearly leap-day subscription resolved its base date a plain year back, landing one day before the clamped anniversary and walking the period back a further year, so two consecutive billing dates produced identical boundaries and the year in between was dropped as a duplicate invoice. Quarterly and semiannual anchored the next end of period on the billing month, which is not necessarily a billing month, so the period advanced a month at a time — that value prices prorated termination credit notes.

## Description

The anniversary day is now derived from the subscription and clamped to the target month, a non-billing month resolves back to an earlier billing month, and period ends are the day before the next anniversary, so periods tile by construction. The next end of period and the period duration are both read off those boundaries rather than recomputed from the billing date.

Resolving the period covering a date is now one strict comparison against the anniversary on or before it, for billing as much as for current usage. The one-day tolerance it used to have for month-end subscriptions was held in place by a spec whose two nested `mark_as_terminated!` calls made the inner one a no-op, so the tolerance stood in for a termination branch that was never reached; that spec now assigns `terminated_at` directly. The semiannual `next_end_of_period` expectations asserted the month-by-month walk and are corrected to the real period ends. A new scenario over the current usage endpoint fails five of its eight examples on `main`.

Affected boundaries move by a day. Quarterly and semiannual boundaries on actual billing days are unchanged, so those invoices need nothing. A yearly 29 Feb subscription previously closed its period on 28 Feb and the next period now opens on that day, so the transition invoice covers that day twice; with a single affected subscription there is nothing to backfill.

### DIFF
--- a/app/services/subscriptions/dates/yearly_service.rb
+++ b/app/services/subscriptions/dates/yearly_service.rb
@@ -126,18 +126,12 @@ module Subscriptions
         compute_to_date(compute_fixed_charges_from_date)
       end
 
+      # NOTE: the period is resolved from its own anniversary rather than re-walked from the billing
+      #       date, so there is a single derivation of the anniversary day.
       def compute_next_end_of_period
         return billing_date.end_of_year if calendar?
 
-        year = billing_date.year
-        month = subscription_at.month
-        day = subscription_at.day
-
-        # NOTE: we need the last day of the period, and not the first of the next one
-        result_date = build_date(year, month, day) - 1.day
-        return result_date if result_date >= billing_date
-
-        build_date(year + 1, month, day) - 1.day
+        compute_to_date(previous_anniversary_day(billing_date))
       end
 
       def compute_previous_beginning_of_period(date)
@@ -179,10 +173,6 @@ module Subscriptions
         # NOTE: a Feb 29 subscription has its anniversary on Feb 28 in a common year, so the raw
         #       subscription day would leave Feb 28 trailing the previous period.
         date.day < anniversary_day_in(date.year, subscription_at.month)
-      end
-
-      def anniversary_day_in(year, month)
-        [subscription_at.day, Time.days_in_month(month, year)].min
       end
     end
   end
--- a/app/services/subscriptions/dates/yearly_service.rb
+++ b/app/services/subscriptions/dates/yearly_service.rb
@@ -45,6 +45,16 @@ module Subscriptions
       end
 
       def compute_base_date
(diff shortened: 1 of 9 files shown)


jdenquin (approved):

app/services/subscriptions/dates/yearly_service.rb
```diff
@@ -141,11 +151,7 @@ def previous_anniversary_day(date)
       def compute_duration(from_date:)
         return Time.days_in_year(from_date.year) if calendar?
 
-        year = from_date.year
-        # NOTE: if after February we must check if next year is a leap year
-        year += 1 if from_date.month > 2
-
-        Time.days_in_year(year)
+        (compute_to_date(from_date).to_date + 1.day - from_date.to_date).to_i
```
jdenquin: It seems that we have a gap here. Since an upgraded subscription keeps the anniversary of the old it replaces.
We can end up in a case where the duration is lower than the effective duration.
For example : 
- initial sub start on 2024-01-01
- subscriptions is upgraded in 2024-10-01
- with the current code, duration will be 92, so each day proration is based on amount / 92. The subscription still started on 01-01 so we should prorate the new amount on the whole billing period (366)?

I think it's more a product question @sarkissianraffi @michaelponrajah

vincent-pochet: Good catch, but the upgrade invoice is not affected: `upgraded_amount` calls `single_day_price` without `optional_from_date`, so the denominator is the inherited anniversary period (366) and the 92 days are only the numerator.

It does break in `terminated_amount` and the trial branch of `full_period_amount`, which pass the boundary start, clamped to `started_at`. On an upgraded subscription that date is mid-period, so the denominator became 83 instead of 366. Fixed in da1e00fc by measuring the duration from the anniversary opening the period.

The monthly, quarterly and semiannual services have the same flaw (they already measured between boundaries before this PR). Will be addressed in a followup PR.

PR #6176, opened 2026-08-18
[ING-576] fix(wallets): back off after a declined top-up

aquinofb:
> Stacked on #6175 (ING-561), which is its base branch. Review that one first; this diff shows only the back-off.

## Context

When an automatic threshold top-up is declined, the credits are never granted, so the balance stays below the threshold and the rule fires again on the next ongoing balance refresh. Neither existing guard holds: the balance is still low, and the failed transaction leaves `scope :pending` as soon as `mark_as_failed!` runs. The result is an attempt every few seconds, then one per refresh, for as long as the customer keeps sending usage, and a wallet can keep that up for days.

No money moves, since every attempt is declined, but each one is a real authorisation against a card that is already failing, which is a good way to get it blocked by the issuer, and each one leaves an open credit invoice behind.

## Changes

- After a declined automatic top-up, the rule stops firing for an hour. The back-off is keyed on the outcome of the last automatic attempt rather than on the balance, which is what lets it bound a case the balance guards cannot see.
- It clears as soon as a paid top-up settles, since that is the one signal proving a payment can go through. A change of payment method is deliberately not treated as recovery: `PaymentMethods::DetermineService` resolves a threshold top-up through the rule, then the wallet, then the customer default, so reading the customer default here would unblock rules pinned to another method while leaving repointed ones stuck.
- Only rules that charge are paused. A rule that just grants credits, or a target rule that grants its top-up, cannot be declined, so pausing it would withhold credits for no protection.
- A declined manual top-up does not pause the rule, since only automatic attempts are retried without anyone asking, and free granted credits do not count as recovery.
- Nothing new is reported on the paused path: wallets on a dedicated worker refresh every few seconds, so a report per evaluation would flood, and the existing burst report already fires during the rapid phase that precedes the back-off.

Linear: [ING-576](https://linear.app/getlago/issue/ING-576)

## Before and after

Every row assumes the wallet is below its threshold and the customer keeps sending usage, so the rule is evaluated on each refresh.

| Scenario | Today on `main` | With this PR |
| -- | -- | -- |
| The card declines | a fresh attempt on every refresh, until the usage stops | one attempt, then nothing for an hour |
| An hour has passed since the decline | a fresh attempt on every refresh | one attempt, and the cycle repeats if it declines again |
| A paid top-up settles after the decline | a fresh attempt on every refresh | the pause lifts at once |
| The rule only grants credits, so nothing can decline | a fresh attempt on every refresh | keeps firing, never paused |
| Only free granted credits arrive | a fresh attempt on every refresh | the pause holds, since nothing proves the card works |
| A manual top-up is the one that declined | a fresh attempt on every refresh | the rule keeps firing, as before |

No money moves in any of these rows, since the attempts are declined. What changes is how many authorisations the card takes and how many open credit invoices are left behind.

### DIFF
--- a/app/services/wallets/threshold_top_up_service.rb
+++ b/app/services/wallets/threshold_top_up_service.rb
@@ -7,12 +7,14 @@ module Wallets
     BURST_TOP_UPS = 3
     BURST_WINDOW = 10.minutes
 
-    def initialize(wallet:)
+    def initialize(wallet:, state_changed: true)
       @wallet = wallet
+      @state_changed = state_changed
       super
     end
 
     def call
+      return result unless state_changed
       return result if rule.nil?
       return result if wallet.credits_ongoing_balance > rule.threshold_credits
       return result if (pending_transactions_amount + wallet.credits_ongoing_balance) > rule.threshold_credits
@@ -47,7 +49,7 @@ module Wallets
 
     private
 
-    attr_reader :wallet
+    attr_reader :wallet, :state_changed
 
     def rule
       @rule ||= wallet.recurring_transaction_rules.active.where(trigger: :threshold).first
--- a/app/services/wallets/threshold_top_up_service.rb
+++ b/app/services/wallets/threshold_top_up_service.rb
@@ -6,6 +6,7 @@ module Wallets
 
     BURST_TOP_UPS = 3
     BURST_WINDOW = 10.minutes
+    DECLINE_BACKOFF = 1.hour
 
     def initialize(wallet:, state_changed: true)
       @wallet = wallet
@@ -19,6 +20,9 @@ module Wallets
       return result if wallet.credits_ongoing_balance > rule.threshold_credits
       return result if (pending_transactions_amount + wallet.credits_ongoing_balance) > rule.threshold_credits
 
+      paid_credits = rule.compute_paid_credits(ongoing_balance: wallet.credits_ongoing_balance)
+      return result if paid_credits.positive? && backing_off_after_decline?
+
       burst_top_ups = top_ups_since(BURST_WINDOW)
       if burst_top_ups >= BURST_TOP_UPS
         report("Automatic wallet top-up burst", count: burst_top_ups)
@@ -26,7 +30,7 @@ module Wallets
 
       params = {
         wallet_id: wallet.id,
-        paid_credits: rule.compute_paid_credits(ongoing_balance: wallet.credits_ongoing_balance).to_s,
+        paid_credits: paid_credits.to_s,
         granted_credits: rule.compute_granted_credits.to_s,
         source: :threshold,
         invoice_requires_successful_payment: rule.invoice_requires_successful_payment?,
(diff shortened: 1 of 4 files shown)


app/services/wallets/threshold_top_up_service.rb
```diff
@@ -59,8 +61,26 @@ def pending_transactions_amount
...
+      last_decline_at = automatic_top_ups.failed.maximum(:failed_at)
+
+      return false if last_decline_at.nil?
+      return false if last_decline_at < DECLINE_BACKOFF.ago
+
+      !recovered_since?(last_decline_at)
+    end
+
+    def recovered_since?(timestamp)
+      PaymentMethod.default.where(customer_id: wallet.customer_id, updated_at: timestamp..).exists? ||
```
toommz: I think the Claude review is right, if we take a look at how the payment method is chosen in `app/services/payment_methods/determine_service.rb#63`:

```ruby
def determine_credit_payment_method
  wallet_transaction = invoice.wallet_transactions.first
  return nil unless wallet_transaction

  return nil if wallet_transaction.payment_method_type == "manual"

  if wallet_transaction.payment_method_id.present?
    return customer.payment_methods.find_by(id: wallet_transaction.payment_method_id)
  end

  if wallet_transaction.source.to_s.in?(%w[interval threshold])
    rule = wallet_transaction.wallet.recurring_transaction_rules.active.first
    return nil if rule&.payment_method_type == "manual"
    return customer.payment_methods.find_by(id: rule.payment_method_id) if rule&.payment_method_id.present?
  end

  wallet = wallet_transaction.wallet
  return nil if wallet.payment_method_type == "manual"

  if wallet.payment_method_id.present?
    return customer.payment_methods.find_by(id: wallet.payment_method_id)
  end

  customer.default_payment_method
end
```

It will pick up the payment method linked to the rule if present. Which could then end up in another failed payment 🤔

app/services/wallets/threshold_top_up_service.rb
```diff
@@ -6,6 +6,7 @@ class ThresholdTopUpService < BaseService
 
     BURST_TOP_UPS = 3
     BURST_WINDOW = 10.minutes
+    DECLINE_BACKOFF = 1.hour
```
toommz, line 9: Curious to know how 1 hour has been chosen.

toommz (approved):

app/services/wallets/threshold_top_up_service.rb
aquinofb: Nothing special. It gives the space to avoid ~12 and ~720 unnecessary attempts for standard and dedicated orgs, respectively. It also gives a decent window for a top-up to happen and recover when the top-up settles anyway.

app/services/wallets/threshold_top_up_service.rb
aquinofb: Applied by 2e08105cfbd884ef1bdc0898a83250808edf3141 — the payment method is no longer read at all. Recovery is now only a settled paid top-up, since which method a top-up uses depends on the rule, then the wallet, then the customer default.

PR #6177, opened 2026-08-18
[ING-560] fix(wallets): close the top-up gap in one charge

aquinofb:
## Context

A threshold rule using the `fixed` method always tops up the configured amount, whatever the size of the shortfall. When a wallet falls further behind than one increment, the rule answers a shortfall that is still open on the next refresh, and does so again, and again: one card charge per increment until it catches up.

The repeated runs are not a loop and do not come from a stale value. `Wallets::ThresholdTopUpService` is only reachable from `Wallets::Balance::RefreshOngoingUsageService`, which reloads the wallet and recomputes its ongoing balance on the line before writing, so every evaluation reads a current number and each top-up is a correct answer to what is still missing. Only the size of each top-up can reduce the number of charges.

Sizing from the ongoing balance alone also ignored two things that already close part of the gap: the credits the same transaction grants for free, and a top-up still on its way. The pending amount was read from the transaction amount as well, a currency figure, and compared against a threshold expressed in credits; those agree only when the wallet rate is one.

## Changes

- A `fixed` rule on a threshold trigger now tops up enough to clear the shortfall in one go, rounded up to a whole number of increments so the configured amount stays the unit the customer chose. This is the shape `compute_target_top_up_amount` already uses.
- The gap accounts for the granted credits and for pending top-ups, and pending credits are read from the credit column so the comparison holds at any wallet rate.
- A gap that divides exactly takes one more increment, so the wallet lands above the threshold rather than exactly on it. Landing on it left the wallet still crossed and triggered another top-up immediately; the customer pays the same either way, but as one charge.
- The amount is capped by the wallet's own maximum paid top-up unless the rule ignores those limits. Charging the maximum once is deliberate: refusing would leave the wallet absorbing all of the customer's usage with no credits to cover it, and any remainder stays visible through the existing burst report.
- Rules triggered by an interval keep paying their configured amount, as the shortfall has no meaning there, and so do rules with nothing to pay.

- A shortfall of exactly one increment rounds the same way, so it clears the threshold instead of landing on it and firing again.
- `WalletTransactions::CreateJob` deduplicated on `[organization_id, wallet_id, paid_credits, granted_credits]`, which stopped collapsing repeat evaluations once the amount started tracking the shortfall. Automatic top-ups now lock on the wallet and the source instead, so a second evaluation collapses onto the queued job whatever it computed. Manual top-ups keep locking on the amounts, since two deliberate top-ups of different sizes are distinct requests.

Linear: [ING-560](https://linear.app/getlago/issue/ING-560)

## Before and after

Threshold 10 credits, top-up 50, no granted credits and no wallet maximum unless the row says otherwise.

| Scenario | Today on `main` | With this PR |
| -- | -- | -- |
| Balance dips just under the threshold | one charge of 50 | one charge of 50 |
| Balance is exactly one increment short, 40 short | one charge of 50, landing on the threshold, then another | one charge of 100 |
| Balance is 429.8 short | nine charges of 50 | one charge of 450 |
| Shortfall is an exact multiple, balance 90 short | three charges of 50, the second landing exactly on the threshold | one charge of 150 |
| Balance 429.8 short with a 50 top-up already in flight | one charge of 50, repeated | one charge of 400 next to the 50 in flight |
| Rule also grants 50, balance 429.8 short | 50 paid plus 50 granted, repeated | 400 paid plus 50 granted, once |
| Wallet maximum is 100, balance 429.8 short | one charge of 50, repeated | one charge of 100, remainder left open and reported as a burst |
| Wallet rate is 0.1 and 1 credit is pending | pending read as 0.1, so the rule fires anyway | pending read as 1 credit, so the rule holds |
| Rule is triggered by an interval | one charge of 50 | one charge of 50 |

Total charged is the same in every row; the difference is how many card charges it takes, except where the pending or granted credits were being double-counted.

### DIFF
--- a/app/jobs/wallet_transactions/create_job.rb
+++ b/app/jobs/wallet_transactions/create_job.rb
@@ -5,6 +5,8 @@ module WalletTransactions
     queue_as "high_priority"
     unique :until_executed, on_conflict: :log
 
+    AUTOMATIC_SOURCES = %w[threshold interval].freeze
+
     # ActiveRecord::StaleObjectError is handled in WalletTransactions::CreateFromParamsService
 
     def perform(organization_id:, params:, unique_transaction: false)
@@ -21,12 +23,18 @@ module WalletTransactions
       unique_transaction = args[:unique_transaction]
 
       if unique_transaction
-        [
-          org_id,
-          params[:wallet_id],
-          params[:paid_credits],
-          params[:granted_credits]
-        ]
+        source = params[:source].to_s
+
+        if AUTOMATIC_SOURCES.include?(source)
+          [org_id, params[:wallet_id], source]
+        else
+          [
+            org_id,
+            params[:wallet_id],
+            params[:paid_credits],
+            params[:granted_credits]
+          ]
+        end
       else
         # Return a unique value for each job to effectively disable uniqueness
         # when unique_transaction is false
--- a/app/models/recurring_transaction_rule.rb
+++ b/app/models/recurring_transaction_rule.rb
@@ -129,7 +129,7 @@ class RecurringTransactionRule < ApplicationRecord
     return paid_credits if paid_credits.zero? || threshold_credits.nil?
 
     gap = threshold_credits - ongoing_balance - granted_credits - pending_credits
-    return paid_credits if gap <= paid_credits
+    return paid_credits if gap < paid_credits
 
     apply_max_top_up_limits(credit_amount: paid_credits * ((gap / paid_credits).floor + 1))
   end
--- a/app/models/recurring_transaction_rule.rb
+++ b/app/models/recurring_transaction_rule.rb
@@ -73,6 +73,14 @@ class RecurringTransactionRule < ApplicationRecord
     end
   end
 
+  def apply_max_top_up_limits(credit_amount:)
+    if ignore_paid_top_up_limits?
+      credit_amount
+    else
+      credit_amount.clamp(nil, wallet.paid_top_up_max_credits)
+    end
+  end
+
   def invoice_custom_section_params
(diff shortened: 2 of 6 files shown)


toommz (commented):
Could we keep the uniqueness key stable per wallet? `CreateJob` includes `paid_credits` in its lock key. Fixed rules previously always queued the configured amount, so repeated evaluations had the same `id` + `paid_credits` tuple as the lock key.

Now two balance refreshes can calculate different amounts, such as `450` and `500`, allowing both jobs to enqueue before either creates a `pending` transaction. Both could then charge. Maybe recalculating under a lock when the job executes would preserve the existing protection 🤔

It's not a situation that might happen a lot in the normal flow, but the dedicated worker flow has a 5s `RefreshWalletJob` rate.

aquinofb:
Right, and that protection was real even if it was accidental. Fixed in `c572812db`: automatic top-ups, threshold and interval, now lock on `[organization_id, wallet_id, source]` instead of the amounts, so a second evaluation collapses onto the job already queued whatever it computed. Manual top-ups keep locking on the amounts, since two deliberate top-ups of different sizes are distinct requests. Covered by three specs: the automatic key ignoring the amount, two different amounts sharing one key, and the manual key unchanged.

I did not move the sizing into the job. Recomputing under a lock is stricter, since the charge would always reflect the balance at execution time, but it turns the job's argument from an amount into a rule, with the payload compatibility that implies for a rolling deploy. The trade with a stable key is that if the deficit grows between enqueue and execution we charge the queued smaller amount and close the remainder on a later refresh, so it errs towards undercharging rather than double charging. Say the word if you would rather have the lock version in this PR.

toommz (approved):
LGTM after the worker lock key changes.

PR #6179, opened 2026-08-18
feat(multi-connection): add logic for grouping recurring invoices based on integration connec…

lovrocolic:
## Context

A customer can hold several connections per category (payment, tax, accounting, CRM), and an invoice can carry only one connection per category. Subscriptions routed to different connections therefore cannot share an invoice, so connection becomes an invoice-split dimension on top of billing entity, currency and payment method.

## Description

`Subscriptions::OrganizationBillingService` now splits each customer's subscription groups by their effective connections before invoices are generated. The effective connection of a subscription is resolved per category: an explicit `billing_object_connections` override pins a connection or skips the category entirely, and without a row the subscription falls back to the customer's default connection. Customers holding a single connection of a category resolve to it even when no default is flagged, so existing setups keep grouping exactly as before.

All four categories are resolved in one pass and compared as a single key, which partitions the groups identically to four sequential passes at a quarter of the work. The customer-level lookups are cached per customer and category, and the overrides are read through a new `billing_object_connections` association on Subscription.

The whole step is gated by the `multi_connection` feature flag: with the flag off, the billing pipeline is unchanged.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -19,6 +19,7 @@ class Subscription < ApplicationRecord
   has_many :invoice_subscriptions
   has_many :invoices, through: :invoice_subscriptions
   has_many :integration_resources, as: :syncable
+  has_many :billing_object_connections, as: :owner, dependent: :destroy
   has_many :fees
   has_many :applied_rate_cards, class_name: "SubscriptionRateCard"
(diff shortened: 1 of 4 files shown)


toommz (approved):
LGTM, besides the possible n+1 found by claude

PR #6182, opened 2026-08-18
fix(quotes): derive the deal term and report failures

toommz:
## Context

`quote_versions.start_date` and `end_date` duplicated dates the billing items already carry
(`plans[].payload.startDate` / `endDate`). Every validator, the deal-expiration fold and the
execution mapping had to reconcile a column with a payload key, and the frontend had to keep both
in sync. The billing items are now the single source of truth for the deal term.

Two QA returns from the end-to-end pass are addressed here: LAGO-1834 (the deal term) and
LAGO-1836 (failures that reached the UI as a 500 or as generic copy).

## Description

Drops both columns and everything downstream: the GraphQL arguments and fields, the REST serializer
keys, the quote-level date validations and the fallbacks in the execution services.

- `ComputeMentionVariablesService` derives the term from the plan billing items instead:
  `commercial_terms_start_date` is the earliest `payload.startDate` and
  `commercial_terms_term_duration` spans to the latest `payload.endDate`. An amendment plan need not
  carry a start date, since the replacement inherits the target's anniversary date, so it falls back
  to the amended subscription's own `subscription_at`. A `one_off` quote carries no plans, so its
  term comes from the add-on service period: the earliest `payload.fromDatetime` to the latest
  `payload.toDatetime`, reading the effective value so an override wins over the payload.
- `payload.startDate` becomes strictly mandatory at approve for `subscription_creation`: the
  quote-level fallback that made it optional is gone.
- `DealExpiration.earliest` folds plan and wallet dates only.

Breaking for the frontend: `startDate` / `endDate` disappear from `QuoteVersion`, `CreateQuoteInput`
and `UpdateQuoteVersionInput`, and the two keys leave the `quote.*` webhook bodies.

## Failure reporting

Lock errors on the quotes used to raise an exception on every mutation but
`ExecuteOrder`, so double-clicking Approve returned a 500. The six request-only services
(`QuoteVersions::Approve` / `Void` / `Clone`, `OrderForms::MarkAsSigned` / `Void`,
`Orders::Update`) now report it themselves as `422 base: ["concurrency_conflict"]`, which covers
GraphQL and REST from one place. `Orders::BaseExecuteService` deliberately keeps raising, because
`ExecuteOrderJob` relies on `retry_on BaseLockService::FailedToAcquireLock` to retry a scheduled
execution.

`QuoteVersions::ApproveService` and `Quotes::CreateService` also rescued only `ValidationFailure`
while their nested `call!` services can raise `NotFoundFailure` / `ForbiddenFailure`; both now
rescue `BaseService::FailedResult`.


mariohd (approved):
Looks good

PR #6185, opened 2026-08-19
feat(quotes): let a quote state its billing entity and deal currency

toommz:
Stacked on #6182.

Two things a quote could not say, and now can: **who issues the deal** and **what currency it is priced in**. Plus two QA returns that landed while this was open.

## Billing entity — LAGO-1835

A customer can hold subscriptions, wallets and invoices across several billing entities, but a quote could not name one: execution always fell back to the customer's default, so a deal for a subsidiary could not be quoted even though the direct API supports it.

Adds a nullable `quote_versions.billing_entity_id`, mirroring `currency`: negotiable per version, editable while drafting, and snapshotted into `mention_variables` at approval so an approved document and its issuer can never disagree. Blank keeps following the customer, resolved at billing time — the same semantic `Subscription` and `Wallet` already carry.

Execution fans it out to the three records that carry one: the subscription, the wallets, and the one-off invoice. The raw column is passed, not the resolved one, so `NULL` keeps inheriting dynamically.

**Amendments never name one.** The target subscription is already bound to an entity and the plan change carries that binding over, so pinning a different one would move a running subscription to another invoice-numbering series mid-life. The version instead *resolves* through the target — the document, the wallets and the plan change then all agree on the same issuer — and naming one explicitly is refused with `not_supported_for_order_type`.

That fallback is deliberately scoped to amendments. `quotes.subscription_id` is optional and only required for amendments, so another order type can carry one that its execution ignores; the document ignores it too rather than naming an issuer nothing bills under.

## Deal currency — LAGO-1840

Quoting a plan in a currency other than its own was rejected outright, even though `Plans::OverrideService` accepts `amount_currency` and reprices the plan it duplicates. The quote payload simply never exposed it. Same for coupons via `AppliedCoupons::CreateService`.

`overrides.amountCurrency` is now accepted on plan and coupon billing items, the deal-currency checks compare the *effective* currency rather than the catalog record's, and the value is passed through at execution.

Keeping that copy in the payload is what keeps the rendered quote honest — the preview reads `billing_items`, so a currency derived only at execution would show one thing on the document and bill another. The cost is that the copy can go stale, so `QuoteVersions::UpdateService` realigns it on **every** update rather than only when the currency changes: the payload is rewritten wholesale by whoever saves it, so a stale copy can arrive at any time and would otherwise sit there until it blocked approval. Realigning is idempotent, so a coherent payload is left untouched.

The override is set only while the catalog record differs, and removed once they agree — a stale override states a currency the deal no longer uses and fails validation exactly as a missing one did. Only that one key is ever touched, so a negotiated amount, a charge override or anything else the deal states survives untouched. **The figure is not converted**: a plan negotiated at 150000 USD becomes 150000 EUR. Re-serialising amounts across currencies is frontend follow-up work on LAGO-1838.

An emptied overrides is left as `{}` rather than having its key removed. Everything reading the payload treats missing, empty and nil alike, so the shape is chosen for the pages rendering the quote: they reach for `overrides.name`, and a key that disappeared takes them down.

Wallet credits are realigned the same way where they state a currency; one-off add-ons carry none, so nothing to do. Amendments refuse a currency change, for the same reason they refuse an entity.

## Also in here

- **LAGO-1839** — `QuoteVersions::CloneService` carried `mention_variables` into the cloned draft. The read path treats a filled column as "this version was approved" and serves the snapshot instead of recomputing, so a version cloned from a voided approval rendered the source's variables forever, whatever it was edited to. Pre-dates this branch; the entity work made it visible. The other half of that ticket is a frontend cache miss, tracked on LAGO-1838.
- **Start date no longer required** on a quoted plan, on the CPO's call. `Subscriptions::CreateService` defaults the subscription date to the moment execution runs, so a deal naming none simply starts then.
- Hardening from review: the realignment runs before the structural pass, on whatever JSON the caller sent, so anything not shaped as expected is left exactly as it arrived for the validator to report rather than raising a 500. And the new `billing_entity_code` on the REST serializer walks a fallback chain, so the quote and version list endpoints preload it — including the amended subscription's own chain — with a query-count guard that compares one row of each kind against three.


mariohd (approved):
Looks good

PR #6188, opened 2026-08-19
fix(subscription): fix race on lifetime usage creation

ansmonjol:
Two concurrent requests activating the same subscription (a manual update racing a scheduled job) could both find `lifetime_usage` blank and both try to build one, tripping the `index_lifetime_usages_on_subscription_id` unique constraint and raising `ActiveRecord::RecordNotUnique`. On the GraphQL mutation path this surfaced to the frontend as a 500 (Sentry FRONT-EU-EC).

`mark_as_active!` now creates the `LifetimeUsage` via `create_or_find_by!`, which retries the lookup on a unique violation instead of raising, so the loser of the race reuses the row the winner just inserted.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -123,16 +123,21 @@ class Subscription < ApplicationRecord
   def mark_as_active!(timestamp = Time.current)
     self.started_at ||= timestamp
     self.activated_at ||= timestamp
-    self.lifetime_usage ||= previous_subscription&.lifetime_usage || find_or_create_lifetime_usage!
+    self.lifetime_usage ||= previous_subscription&.lifetime_usage || existing_or_new_lifetime_usage
     self.lifetime_usage.recalculate_invoiced_usage = true
     active!
   end
 
-  # Two concurrent requests activating the same subscription (e.g. a manual update racing
-  # a scheduled job) can both find `lifetime_usage` blank and both attempt to build one,
-  # tripping the `index_lifetime_usages_on_subscription_id` unique index. `create_or_find_by!`
-  # runs the insert in a savepoint and falls back to the row the other request created.
-  def find_or_create_lifetime_usage!
+  # A brand new subscription can't race anyone for its lifetime usage row (nothing else
+  # references it yet), so it's built unsaved and persisted together with the subscription,
+  # same as before. An already-persisted subscription can: two concurrent requests activating
+  # it (e.g. a manual update racing a scheduled job) can both find `lifetime_usage` blank and
+  # both attempt to create one, tripping the `index_lifetime_usages_on_subscription_id` unique
+  # index. `create_or_find_by!` runs that insert in a savepoint and falls back to the row the
+  # other request created.
+  def existing_or_new_lifetime_usage
+    return build_lifetime_usage(organization:) unless persisted?
+
     LifetimeUsage.create_with(organization:).create_or_find_by!(subscription: self)
   end
 
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -120,10 +120,7 @@ class Subscription < ApplicationRecord
     billing_entity_id || customer&.billing_entity_id
   end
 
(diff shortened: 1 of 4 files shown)


ansmonjol:
Fixed in a8e4ba02b — the test was passing the same in-memory `subscription` object to the factory, which let Rails' automatic inverse-of wire the association cache and made the reader find the row directly, never reaching `create_or_find_by!`'s rescue branch. Now inserts the competing row via a separately-loaded `Subscription` instance so the acting object's cache stays genuinely stale. Verified red (raises `PG::UniqueViolation`) against main and green against the fix.

app/models/subscription.rb
```diff
@@ -123,11 +123,24 @@ def applicable_billing_entity_id
...
     active!
   end
 
+  # A brand new subscription can't race anyone for its lifetime usage row (nothing else
+  # references it yet), so it's built unsaved and persisted together with the subscription,
+  # same as before. An already-persisted subscription can: two concurrent requests activating
+  # it (e.g. a manual update racing a scheduled job) can both find `lifetime_usage` blank and
+  # both attempt to create one, tripping the `index_lifetime_usages_on_subscription_id` unique
+  # index. `create_or_find_by!` runs that insert in a savepoint and falls back to the row the
+  # other request created.
```
vincent-pochet: Can you remove this comment, I think it brings more confusion than actual explanation :D

spec/models/subscription_spec.rb
```diff
@@ -1076,6 +1057,43 @@
...
     end
+
+    context "when a concurrent request already created the lifetime usage" do
+      subject(:subscription) { create(:subscription, :pending) }
+
+      # Simulates the actual race: `subscription`'s association cache is memoized to nil
+      # (as it would be mid-request) before a *different* Ruby object inserts the row for
+      # the same row, via a separate `Subscription.find`. When `mark_as_active!` then tries
+      # to create its own, it hits the real `index_lifetime_usages_on_subscription_id`
+      # unique constraint instead of finding the row through its stale cache.
```
vincent-pochet: Classic claude over-explanation. `context` and `it` description are self-explanatory in this case, no need for a full paragraph to paraphrase it ;)

vincent-pochet (approved):

ansmonjol:
Fixed in 62909ecad. Valid catch — the race-recovery let the loser continue through activation and re-trigger `EmitFixedChargeEventsService`/`BillSubscriptionJob`/`CreatePayInAdvanceFixedChargesJob`. `mark_as_active!` now returns whether it actually performed the activation or found a concurrent request already had; `UpdateService#process_subscription_at_change` skips its billing side effects on a lost race. Added a service-level regression test (`update_service_spec.rb`) verifying jobs/events aren't duplicated — verified red against the pre-guard code (both fired twice), green with the fix.

Scoped to `UpdateService`, the exact path from the original Sentry crash — didn't touch `ActivateService`, which already guards on `subscription.active?` at entry and wasn't implicated.

app/models/subscription.rb
ansmonjol: Trimmed in f505fd538 — cut it to a one-liner on the method that actually does the racy work.

spec/models/subscription_spec.rb
ansmonjol: Trimmed in f505fd538 — down to one line.

vincent-pochet:
Good to go @ansmonjol :rocket:

app/services/subscriptions/update_service.rb
```diff
@@ -161,7 +161,8 @@ def process_subscription_at_change(subscription)
         )
       end
 
-      subscription.mark_as_active!(subscription.subscription_at)
+      # Skip billing side effects if a concurrent request already activated the subscription.
+      return unless subscription.mark_as_active!(subscription.subscription_at)
```
vincent-pochet: In case of concurrent request `subscription.mark_as_active!` could raise a `RecordInvalid: External value_already_exist` rather than returning false

PR #6194, opened 2026-08-20
fix(billing): prorate on the whole period

vincent-pochet:
> Follow-up of #6173, which fixes the same flaw in the yearly service. Not stacked on it: this branch is off `main` and the two touch distinct files.

## Context

The duration of a period is the denominator of the proration: a subscription fee bills a number of days at the price of a single day, the plan amount divided by the length of the whole period. That duration is derived from a date the callers choose, and the terminated and the trial fees pass the start of the invoice boundaries, which `DatesService#from_datetime` clamps to the start date of the subscription.

A subscription created by an upgrade inherits the anniversary of the one it replaces, so it starts in the middle of its first period, and that clamped date is not the beginning of the period. The window was then measured from it instead:

```
single_day_price = plan_amount / period_duration
period_duration  = period_end + 1 day - period_start

# calendar, started on 20 May 2024, period 01 Apr -> 30 Jun
before: 2024-06-30 + 1 day - 2024-05-20 = 42   # truncated to the rest of the period
after:  2024-06-30 + 1 day - 2024-04-01 = 91

# anniversary on the 15th, started on 20 May 2024, period 15 Apr -> 14 Jul
before: 2024-08-14 + 1 day - 2024-05-20 = 87   # whole window, walked from started_at
after:  2024-07-14 + 1 day - 2024-04-15 = 91
```

Both shapes shrink the denominator, so the price of a day is too high and the customer is overcharged on termination. Monthly drops to 21 days instead of 31, semiannual to 103 instead of 184.

## Description

Resolve the beginning of the period holding the given date and measure from there, so a date taken in the middle of a period still yields the duration of the whole period. `compute_previous_beginning_of_period` already answers this for both billing times, which is why the calendar early return of the monthly service could go with it.

Each interval spec gains a mid-period case, and a scenario reaches the termination invoice through the API on a plan whose amount makes a single day cost 1000 cents over the whole quarter, covering both shapes: terminating 32 days into the period billed 33471 cents on the upgrade and 69333 on the calendar subscription, where 32000 is due.

The weekly service is immune, its duration being a constant. The clamping itself is left alone, since the boundaries of the invoice are correct.

Termination and trial invoices of subscriptions started in the middle of a period become smaller, prorated on the whole period. Invoices already issued are untouched and nothing needs backfilling.

### DIFF
--- a/app/services/subscriptions/dates/monthly_service.rb
+++ b/app/services/subscriptions/dates/monthly_service.rb
@@ -28,11 +28,14 @@ module Subscriptions
         compute_to_date(compute_charges_from_date)
       end
 
+      # NOTE: `from_date` is not necessarily the beginning of the period: on a subscription resulting
+      #       from an upgrade, it is clamped to `started_at` while the anniversary is inherited from the
+      #       previous subscription. The duration is the one of the whole period, so it is measured from
+      #       the beginning of the period holding `from_date`.
       def compute_duration(from_date:)
-        return Time.days_in_month(from_date.month, from_date.year) if calendar?
+        period_start = compute_previous_beginning_of_period(from_date.to_date)
 
-        next_month_date = compute_to_date(from_date)
-        (next_month_date.to_date + 1.day - from_date.to_date).to_i
+        (compute_to_date(period_start).to_date + 1.day - period_start).to_i
       end
 
       alias_method :compute_charges_duration, :compute_duration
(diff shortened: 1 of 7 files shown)


vincent-pochet:
> 
> HOLD: Fix the same mid-period duration bug in `YearlyService` and add regression coverage. For an anniversary period starting January 15, 2024 with an upgraded subscription starting March 20, `compute_duration` still returns 365 days from the clamped date even though the full period has 366 days.

Addressed in https://github.com/getlago/lago-api/pull/6173

toommz (approved):

PR #6197, opened 2026-08-21
fix(email): decouple manual sends

anhtho-lago:
## Context

Customers may intentionally disable automatic document emails while still needing to send a finalized document manually to selected recipients. Foxglove uses this workflow for one-off invoices and is currently blocked because the manual endpoint rejects sends when the automatic scenario is disabled.

## Description

- Treat an explicit resend request as independent from automatic email scenarios
- Keep the existing finalized-status, premium-license, sender-email, and recipient validation
- Update the service spec to cover a manual invoice email when automatic emails are disabled

## Test plan

- `ruby -c app/services/emails/resend_service.rb`
- Updated `spec/services/emails/resend_service_spec.rb` (full execution delegated to CI because the local Lago Docker runtime is unavailable)

### DIFF
--- a/app/services/emails/resend_service.rb
+++ b/app/services/emails/resend_service.rb
@@ -88,6 +88,8 @@ module Emails
     def validation_errors
       errors = {}
       errors[:billing_entity] = ["must have email configured"] if billing_entity.email.blank?
+      errors[:from] = ["must have a sender email address configured"] if billing_entity.from_email_address.blank?
+      errors[:invoice] = ["must have a non-zero fees amount"] if zero_amount_invoice?
       errors[:to] = ["must have at least one recipient"] if recipients_to.empty?
 
       invalid_to = recipients_to.reject { |email| valid_email?(email) }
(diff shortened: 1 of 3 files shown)


app/services/emails/resend_service.rb
```diff
@@ -16,7 +16,6 @@ def call
       return result.not_found_failure!(resource: resource_type) unless resource
       return result.not_allowed_failure!(code: "#{resource_type}_not_finalized") unless valid_status?
       return result.forbidden_failure!(code: "premium_license_required") unless License.premium?
-      return result.not_allowed_failure!(code: "email_settings_disabled") unless email_settings_enabled?
```
vincent-pochet, line 19: Removing this guard will allow organization with no email config at all to "resend" an email, and would lead to error in the Sidekiq delivery job.
The idea is good but the service should enforce the validation of the email delivery configuration if we want to bypass the actual email settings

annvelents: @vincent-pochet , I've checked which values are required to send an email, it seems that only `billing_entity.email` and ` billing_entity.from_email_address` - added these into validations
there also might be a problem of missing `LAGO_SMTP_ADDRESS`, but I'm not sure it worth checking this env here

annvelents (approved):

vincent-pochet (approved):

PR #6208, opened 2026-08-24
feat: add taxes to rate cards

endenis:
## Context

_First step of the taxes support for Products and Plans._

Taxes for the product catalog are configured on Rate Cards. They can override taxes inherited from the Plan, Customer, or Billing Entity.

## Description

This PR adds tax assignments to Rate Cards and exposes them through the REST and GraphQL APIs.

It doesn't resolve taxes on product-catalog fees yet. This will be handled in future PRs.

## Implementation details

- **New `rate_cards_taxes` table** linking Rate Cards and Taxes.
- **Tax management on Rate Cards**:
  - `tax_codes` can be provided when creating or updating a Rate Card,
  - an empty list removes the Rate Card tax override,
  - an omitted or `null` value keeps the current assignments,
  - taxes can still be changed when the Rate Card is in use.
- **REST and GraphQL responses** now include the assigned taxes.
- Tax codes are resolved within the Rate Card organization.
- Tax assignments are replaced inside a transaction with a Rate Card lock.
- Removing a Tax also removes its Rate Card assignments.
- Tax changes are included in activity logs and PaperTrail history.

### DIFF
--- a/app/controllers/api/v2/rate_cards_controller.rb
+++ b/app/controllers/api/v2/rate_cards_controller.rb
@@ -74,11 +74,10 @@ module Api
         if result.success?
           render(
             json: ::CollectionSerializer.new(
-              result.rate_cards.includes(:product, :product_filter, :rates, :taxes),
+              result.rate_cards.includes(:product, :product_filter, :rates),
               ::V1::RateCardSerializer,
               collection_name: "rate_cards",
-              meta: pagination_metadata(result.rate_cards),
-              includes: %i[taxes]
+              meta: pagination_metadata(result.rate_cards)
             )
           )
         else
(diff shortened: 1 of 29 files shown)


app/controllers/api/v2/rate_cards_controller.rb
```diff
@@ -72,10 +72,11 @@ def index
         if result.success?
           render(
             json: ::CollectionSerializer.new(
-              result.rate_cards.includes(:product, :product_filter, :rates),
+              result.rate_cards.includes(:product, :product_filter, :rates, :taxes),
               ::V1::RateCardSerializer,
               collection_name: "rate_cards",
-              meta: pagination_metadata(result.rate_cards)
+              meta: pagination_metadata(result.rate_cards),
+              includes: %i[taxes]
```
rsempe: What do you think of removing `taxes` and keeping them on the single-card responses only?
We would like to avoid having nested resources on list endpoints for the v2.

endenis: @rsempe makes sense, I removed taxes from `index`

rsempe (approved):
👍

PR #6209, opened 2026-08-24
fix(events): address a single event by its dedup key

domenicofalco:
## Context

The developer Events tab cannot switch the detail pane between events that share a transaction ID. The API is what makes it impossible: `Query.event` resolves on a transaction ID alone, which is not an identity. The same transaction ID is legitimately allowed on different subscriptions, and on the ClickHouse store nothing enforces uniqueness at all, so the lookup returned an arbitrary match.

## Description

Give `Query.event` the arguments to address one specific event: `externalSubscriptionId`, `code` and `timestampMs`. Together with the organization they cover the identity aggregation already deduplicates on, so any row the list can show is a row this lookup can reach.

A transaction ID is now required. The other arguments only narrow the lookup, and without it a caller sending just an external subscription ID silently received that subscription's most recent event — a plausible wrong answer instead of an error. The argument stays nullable in the schema, so the frontend's existing `event(transactionId:)` query keeps working unchanged.

The timestamp needs a lossless companion field, because ISO8601 serializes at whole-second precision while the ClickHouse column is a `DateTime64(3)`, so a timestamp read from the API and sent back could not address a millisecond. `timestampMs` carries it as a `BigInt`. Matching goes through `toUnixTimestamp64Milli`, since comparing a formatted string against a `DateTime64` silently matches nothing. Postgres matches without it: `index_unique_transaction_id` already covers the tuple, and millisecond matching could not represent that column's microsecond precision anyway.

## What is deliberately not here

**Deduplicating the events list.** An earlier commit on this branch did it in `EventsResolver` and it was reverted in 2a31919c1. Measured against a synthetic organization with 3M `events_raw` rows:

| | before | with dedup |
|---|---|---|
| count | 18 ms | 1699 ms |
| page 1 | 45 ms | 2445 ms |
| page 50 | 81 ms | 1914 ms |

Without an explicit `max_memory_usage` the page query restarted the ClickHouse container. `ingested_at` is not in the table sort key, so the outer `LIMIT` used to allow a bounded top-N; a `LIMIT 1 BY` subquery removes that and forces every row of the organization to be read, sorted and materialized, with the count repeating it on each page load. Three million rows is roughly a day and a half for one real customer, and an unbounded ClickHouse count has already caused a production incident here.

Collapsing duplicate rows moves to the client, which already has to key rows and its Apollo cache on the same tuple and can do it over the page it has fetched. The trade: the total count keeps counting raw rows, so it can report one more event than the list displays.

**Narrowing `Api::V1::EventsController#show`.** It has the same defect and predates this change. `Query.event` is consumed only by our own frontend, so widening it is an internal contract change; `GET /api/v1/events/:id` is a documented public endpoint and changing it needs product sign-off, an OpenAPI update and SDK releases. Tracked separately.

**Ingestion.** ClickHouse accepting duplicate transaction IDs is intentional — non-blocking ingestion, duplicates resolved downstream — and was confirmed to the reporting customer in writing on 2026-08-20, with the docs corrected accordingly.

Fixes ING-594


app/graphql/resolvers/events_resolver.rb
```diff
@@ -27,5 +27,31 @@ def resolve(page: nil, limit: nil)
...
+    #
+    # LIMIT 1 BY has to live in a subquery: generated at the same level as Kaminari's
+    # LIMIT/OFFSET the two do not compose, and both the page and the total count come out wrong.
+    def deduplicated_clickhouse_events
+      base_sql = Clickhouse::EventsRaw
+        .where(organization_id: current_organization.id)
+        .to_sql
+
+      deduplicated_sql = "#{base_sql} ORDER BY ingested_at DESC " \
+                         "LIMIT 1 BY external_subscription_id, transaction_id, timestamp, code"
```
vincent-pochet: `LIMIT 1 BY` was tried in the past on `events_enriched` and lead to missed values. That's why in the event stores we moved to `argMax GROUP BY`

app/graphql/types/events/object.rb
```diff
@@ -15,6 +15,7 @@ class Object < Types::BaseObject
       field :deleted_at, GraphQL::Types::ISO8601DateTime, null: true
       field :received_at, GraphQL::Types::ISO8601DateTime, null: true, method: :created_at
       field :timestamp, GraphQL::Types::ISO8601DateTime, null: true
+      field :timestamp_ms, GraphQL::Types::BigInt, null: true
```
vincent-pochet, line 18: Rest API accepts input with millisecond precision and as such it makes sense to expose it here, I'm just wondering if it's not better to match the same format: `sssssss.mmm`. It's more a question that a change request ;)

domenicofalco:
Thanks — the inconsistency you point at was real, and the first commit is what introduced it. It is gone now, but not the way you proposed.

**Second bullet (`EventsQuery` / `/api/v1/events` duplicates): resolved by removing the GraphQL deduplication, not by adding it to REST.**

Benchmarked against a synthetic organization with 3M `events_raw` rows:

| | before | with dedup |
|---|---|---|
| count | 18 ms | 1699 ms |
| page 1 | 45 ms | 2445 ms |
| page 50 | 81 ms | 1914 ms |

Without an explicit `max_memory_usage` the page query restarted the ClickHouse container. The cause is structural: `ingested_at` is not in the table sort key, so the outer `LIMIT` used to permit a bounded top-N, and wrapping the scope in a `LIMIT 1 BY` subquery removes that — every row of the organization gets read, sorted and materialized, and the count repeats it on every page load.

Three million rows is about a day and a half for one real customer of ours, and an unbounded ClickHouse count has already caused a production incident here. So the deduplication is out of the resolver entirely (2a31919c1); the client collapses duplicates over the page it has already fetched, which costs ClickHouse nothing. GraphQL and REST now return the same raw rows again — no divergence left to reconcile.

**First bullet (`Api::V1::EventsController#show` narrowing): real, but deliberately not in this PR.**

You are right that it selects on organization + transaction ID alone and can return the wrong subscription's event. That predates this change and this PR does not make it worse.

Fixing it is not equivalent to the GraphQL change, though. `Query.event` is consumed only by our own frontend, so widening it is an internal contract change. `GET /api/v1/events/:id` is a documented public endpoint: changing which row it returns, or requiring extra parameters to disambiguate, is a customer-facing API change that needs product sign-off, an OpenAPI update and SDK releases. Slipping it into a bugfix PR for a debugging tab is the wrong vehicle.

Tracked separately rather than dropped.

app/graphql/resolvers/event_resolver.rb
```diff
@@ -7,15 +7,54 @@ class EventResolver < Resolvers::BaseResolver
...
+        clickhouse_event(transaction_id:, external_subscription_id:, timestamp_ms:, code:)
+      else
+        # NOTE: index_unique_transaction_id makes (organization_id, external_subscription_id,
+        # transaction_id) unique, so the timestamp adds nothing here — and matching it in
+        # milliseconds could not represent the column's microsecond precision anyway.
+        postgres_event(transaction_id:, external_subscription_id:)
+      end
+
+      event || not_found_error(resource: "event")
+    end
```
vincent-pochet: This is a bit too much business/query logic for a resolver. I would suggest to move it into a proper query or service

app/graphql/resolvers/event_resolver.rb
```diff
@@ -7,15 +7,54 @@ class EventResolver < Resolvers::BaseResolver
...
+    # The arguments mirror the key EventsResolver deduplicates the list on, so that every row
+    # the list can show is a row this lookup can address. `code` included: two rows differing
+    # only by it are separate events, both listed, and nothing else here separates them.
+    def clickhouse_event(transaction_id:, external_subscription_id:, timestamp_ms:, code:)
+      scope = Clickhouse::EventsRaw.where(organization_id: current_organization.id, transaction_id:)
+      scope = scope.where(external_subscription_id:) if external_subscription_id.present?
+      scope = scope.where(code:) if code.present?
+      # NOTE: timestamp is a DateTime64(3); comparing it against a formatted string
+      # silently matches nothing, so go through the millisecond representation.
+      scope = scope.where("toUnixTimestamp64Milli(timestamp) = ?", timestamp_ms) if timestamp_ms.present?
```
vincent-pochet: I would suggest to format the `timestamp_ms` as regular Time object, clickhouse via the ActiveRecord adapter is able to match it again a DateTime value and would give you the benefits of relying on the Clickhouse primary key definition for this table.

Also, I know the logic already rely on `Clickhouse::EventsRaw` but this is supposed to be a technical table, we should build on top of `Clickhouse::EventsEnriched` with already handles deduplication (eventually consistent)

app/graphql/resolvers/event_resolver.rb
domenicofalco: Agreed — moved into `EventQuery` in 22f88684f, next to the existing `EventsQuery`. The resolver is back to translating arguments and mapping the result.

app/graphql/resolvers/event_resolver.rb
domenicofalco: On the `Time` object: you were right, and it cost more than style — `toUnixTimestamp64Milli()` wrapped the column, so the sort key could not be used. Fixed in 22f88684f, with a spec on two rows 456ms apart.

On `EventsEnriched`: I would rather not, because I think it breaks this screen. Events whose billable metric cannot be resolved never reach `events_enriched` — `enrichment_service.go:40` routes them to the dead letter queue — and those are exactly what `matchBillableMetric` exists to surface. It also has no `ingested_at`, so `receivedAt` would change meaning.

let me know if you are fine staying on `events_raw`, or tell me the enriched store is meant to carry those events too and I will switch. This is the only thing I am holding merge on.

app/graphql/types/events/object.rb
domenicofalco: Kept integer milliseconds on purpose: this value is a lookup key the client has to send back byte-identical, and `1787563245.5` / `1787563245.500` are the same instant but different strings. `timestamp` is still exposed as ISO8601 for display.

let me know if you prefer matching the REST format and I will change it — one line on each side. Otherwise I leave it as is.

app/graphql/resolvers/event_resolver.rb
vincent-pochet: Good for me. Thank you for clarifying

vincent-pochet (approved):

PR #6210, opened 2026-08-24
feat(taxes): add product integration mappings

endenis:
## Context

_Part of the products and plans project._

Anrok and Avalara need an external item ID for each billed line.

The existing billing system stores this mapping on BillableMetric or AddOn. Product-catalog fees should use a mapping stored on Product.

## Description

- Add Product to the existing integration mapping system.
- Support creating, reading, updating, and deleting Product mappings through GraphQL.
- Scope Product and Integration lookups to the current organization.
- Prevent mappings from referencing another organization's Product or Integration.
- Batch integration mapping reads on Product lists.
- Keep existing BillableMetric and AddOn mappings unchanged.
- Add coverage for the existing integration fallback when a Product has no mapping.

No database migration is required because integration mappings already use a polymorphic association.

Frontend PR for compatibility => https://github.com/getlago/lago-front/pull/4177

## Testing

- Product mapping model and service specs
- GraphQL create, read, update, and delete specs
- Organization-isolation specs
- Product list query batching
- Integration fallback coverage
- GraphQL schema consistency

### DIFF
--- a/app/graphql/types/products/object.rb
+++ b/app/graphql/types/products/object.rb
@@ -38,10 +38,7 @@ module Types
       end
 
       def integration_mappings(integration_id: nil)
-        mappings = dataloader.with(Sources::ActiveRecordAssociation, :integration_mappings).load(object)
-        return mappings unless integration_id
-
-        mappings.select { |mapping| mapping.integration_id == integration_id }
+        dataloader.with(Sources::IntegrationMappingsByMappable, integration_id).load(object)
       end
     end
   end
--- a/app/graphql/types/products/object.rb
+++ b/app/graphql/types/products/object.rb
@@ -19,6 +19,9 @@ module Types
       field :product_type, Types::Products::ProductTypeEnum, null: false
 
       field :billable_metric, Types::BillableMetrics::Object, null: true
+      field :integration_mappings, [Types::IntegrationMappings::Object], null: true do
+        argument :integration_id, ID, required: false
+      end
       field :product_category, Types::ProductCategories::Object, null: true
 
       field :filters_count, Integer, null: false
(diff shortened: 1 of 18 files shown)


app/graphql/types/products/object.rb
```diff
@@ -33,6 +36,13 @@ def attached_to_plan_or_subscription
       def filters_count
         dataloader.with(Sources::CountByForeignKey, ProductFilter, :product_id).load(object.id)
       end
+
+      def integration_mappings(integration_id: nil)
+        mappings = dataloader.with(Sources::ActiveRecordAssociation, :integration_mappings).load(object)
+        return mappings unless integration_id
+
+        mappings.select { |mapping| mapping.integration_id == integration_id }
```
tiagolupepic: is it possible to delegate this filter to the dataloader (DB)?

tiagolupepic (approved):
just a small comment and it's not a blocker

:shipit:

spec/services/integrations/aggregator/base_payload_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    let!(:fallback_mapping) do
+      create(
+        :anrok_collection_mapping,
+        integration:,
+        organization:,
+        mapping_type: :fallback_item
+      )
+    end
+
+    it "returns the integration fallback when the Product has no mapping" do
```
rsempe, line 23: should we add a test when product has mapping?

rsempe (approved):
lgtm

app/graphql/types/products/object.rb
endenis: done, thank you

spec/services/integrations/aggregator/base_payload_spec.rb
endenis: Added the test, thank you 🙂

PR #6213, opened 2026-08-24
feat(products): author rate phase sequences on plan rate cards

rsempe:
## Context

Plan rate cards carry an ordered sequence of rate phases (e.g. 3 cycles of launch pricing, then a standard indefinite tail). This PR is the sequence machinery, the override pricing layer is a separate PR in #5842.

## Description

- `RatePhases::ReplaceService` — atomic whole-sequence authoring, used by nested `rate_phases` on plan rate card creation; validates contiguous positions and the indefinite-tail rule before writing.
- Single-phase REST endpoints under `/api/v2/plans/:plan_code/applied_rate_cards/:code/rate_phases` (index/create/update/destroy) with position renumbering, terminal-phase promotion on delete, and `plan_locked` once the plan has subscriptions.
- GraphQL: `createRatePhase` / `updateRatePhase` / `destroyRatePhase` mutations and the `RatePhase` object on `PlanAppliedRateCard`.


app/services/rate_phases/destroy_service.rb
```diff
@@ -0,0 +1,59 @@
...
+  # always keeps a coherent tail.
+  class DestroyService < BaseService
+    Result = BaseResult[:rate_phase]
+
+    def initialize(rate_phase:)
+      @rate_phase = rate_phase
+      super
+    end
+
+    def call
```
endenis, line 15: Should we lock the parent rate card before loading and updating the phase positions? 🤔 

Two concurrent requests could calculate from the same positions and leave gaps or duplicates

app/controllers/api/v2/plan_rate_cards/rate_phases_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+            plan&.applied_rate_cards&.joins(:rate_card)&.find_by(rate_cards: {code: params[:applied_rate_card_code]})
+          end
+        end
+
+        def find_rate_phase
+          plan_rate_card&.rate_phases&.find_by(code: params[:code])
+        end
+
+        def phase_params
+          params.require(:rate_phase).permit(:code, :position, :name, :billing_interval_cycle_count)
```
endenis, line 73: Not sure if it's worth changing: `:position` is used in the `create` but not in the `update`

endenis (approved):

app/services/rate_phases/destroy_service.rb
rsempe: Good catch 👍

app/controllers/api/v2/plan_rate_cards/rate_phases_controller.rb
rsempe: splitting it 👍

PR #6220, opened 2026-08-25
feat(graphql): provider identity on the connection

domenicofalco:
## Context

The `ProviderCustomer` type does not say which PSP the connection points at: the frontend reads the provider identity from the customer's top-level `paymentProvider` / `paymentProviderCode`, which only works while there is exactly one payment connection per customer. Lifting the max-1 cap (ING-435 / ING-442) breaks that — a single pair of customer-level fields cannot label N rows. Integration connections already carry their identity per row (`integrationType` / `integrationCode`).

## Description

Expose `paymentProvider` (ProviderTypeEnum) and `paymentProviderCode` on `ProviderCustomer`, read from the row:

- the provider enum is derived from the STI type (the demodulize pattern already used by `PaymentProviderCustomers::DestroyService#provider_type`), guarded so the base class resolves to `nil` rather than garbage;
- the code comes from the `payment_provider` association with nil-safety (deleted provider → `nil`, never an error);
- both fields nullable: the table can hold provider-less rows (STI base class), and they flow through this same type.

Purely additive — no migration, no behavior change for existing consumers (no existing operation selects the new fields), schema diff is the two fields only. Note for the array reader in #6081: resolving `paymentProviderCode` per row goes through the association, worth a preload once lists get real.

Fixes ING-602

### DIFF
--- a/app/graphql/types/payment_provider_customers/provider.rb
+++ b/app/graphql/types/payment_provider_customers/provider.rb
@@ -25,7 +25,7 @@ module Types
       end
 
       def payment_provider_code
-        object.payment_provider&.code
+        dataloader.with(Sources::ActiveRecordAssociation, :payment_provider).load(object)&.code
       end
     end
   end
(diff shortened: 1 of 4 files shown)


domenicofalco:
Applied in 93588f52f: `paymentProviderCode` now loads the association through `Sources::ActiveRecordAssociation`, so requests batch across all rows of an execution, and a request spec covers the multi-row path (two customers' connections resolved in a single query, each with its own provider).

lovrocolic (approved):

PR #6223, opened 2026-08-25
fix(stripe-customer-balance): implement auto payment retry for canceled intent with customer balance

lovrocolic:
## Context

For Stripe `customer_balance` (V-BAN) customers, a wire lands in the Stripe cash balance,
which is attached to the customer and not to an invoice. If the invoice's PaymentIntent was
canceled or failed, the funds have nothing to attach to — Lago never learns of them, and
Stripe later applies them to whichever invoice opens the next intent, marking an invoice paid
that the customer never funded.

## Description

Subscribe to Stripe's `customer_cash_balance_transaction.created` event. When a funding leaves
an unallocated balance and the oldest unpaid invoice's last payment attempt failed, open a new
intent on that invoice so Stripe applies the funds to the invoice they were meant for. Oldest
first, and only in the funded currency, since targeting a later invoice is the bug itself.

### DIFF
--- a/app/models/payment_providers/stripe_provider.rb
+++ b/app/models/payment_providers/stripe_provider.rb
@@ -19,6 +19,7 @@ module PaymentProviders
       charge.refund.updated
       customer.updated
       charge.dispute.closed
+      customer_cash_balance_transaction.created
     ].freeze
 
     PROCESSING_STATUSES = %w[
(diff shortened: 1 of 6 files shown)


db/structure.sql
toommz, line 1: I think this is not related to the current PR

toommz (approved):
The structure.sql changes have to be removed, but as the CI is blocking because of it I'm approving on my side to let you merge it when the CI is green 👍

mariohd (approved):

PR #6225, opened 2026-08-26
fix(charges): guard missing graduated percentage ranges

rsempe:
## Context

`Charges::Validators::GraduatedPercentageService#ranges` maps over `properties["graduated_percentage_ranges"]` before the `blank?` presence check can run, so a payload without the key raises `NoMethodError` instead of failing validation.


tiagolupepic (approved):
the changes looks good! just a heads up, there is a spec failing

:shipit:

PR #6228, opened 2026-08-26
chore(ruby): Upgrade to ruby 4.0.6

jdenquin:


.github/workflows/spec.yml
```diff
@@ -70,7 +70,7 @@ jobs:
       - name: Install Ruby and gems
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: "4.0.2"
```
vincent-pochet, line 73: I guess the same should be made on `.github/workflows/migrations-test.yml:57`

vincent-pochet (approved):

.github/workflows/spec.yml
jdenquin: yes, and that's why CI is failing!

PR #6235, opened 2026-08-26
feat(products): add dedicated contracts storage

rsempe:
## Context

The v2 runtime object gets dedicated resources instead of sharing the legacy `subscriptions` table.

## Description

- New `contracts` table: customer agreement with a nullable `plan_id` (the plan-less shape is native), `status` and `billing_time` PG enums, the billing anchor and lifecycle dates.
- `contract_rate_cards` takes over from `subscription_rate_cards`, attached to contracts. Per the [two-step drop policy](docs/dropping_columns_and_tables.md), the old table stays in place for one release with no remaining code reference; its drop ships in a dedicated follow-up after the next release.
- `rate_phases` now references `contract_rate_card_id` (column rename; it never held a value).
- `Contract` and `ContractRateCard` models, factories and specs; the rate-phase parent, GraphQL attachment sources and product catalog associations follow the rename.
- Plan and rate card attach-locks now count contracts; the subscription check stays for catalog plans subscribed through v1 before contracts existed.


app/models/contract.rb
```diff
@@ -0,0 +1,74 @@
...
+  }.freeze
+
+  BILLING_TIMES = {
+    calendar: "calendar",
+    anniversary: "anniversary"
+  }.freeze
+
+  belongs_to :organization
+  belongs_to :customer
+  belongs_to :plan, optional: true
```
endenis: Should relations to `customer` and `plan` use the `with_discarded` scope, like they do on Subscription?

app/models/contract.rb
```diff
@@ -0,0 +1,74 @@
...
+# == Schema Information
+#
+# Table name: contracts
+# Database name: primary
+#
+#  id                  :uuid             not null, primary key
+#  billing_anchor_date :date
+#  billing_time        :enum             default("calendar"), not null
+#  canceled_at         :datetime
+#  ending_at           :datetime
```
tiagolupepic: I'm wondering if this column would be used to have a sort of "scheduled" end of contract. Or a simple ended_at works for the propose which is ends the contract.

endenis (approved):

app/models/contract.rb
```diff
@@ -0,0 +1,74 @@
...
+#  name                :string
+#  started_at          :datetime
+#  starts_at           :datetime
```
tiagolupepic: out of curiosity, what is the difference between started_at and starts_at?

db/migrate/20260826235313_create_contract_rate_cards.rb
```diff
@@ -0,0 +1,32 @@
...
+class CreateContractRateCards < ActiveRecord::Migration[8.0]
+  def change
+    create_table :contract_rate_cards, id: :uuid do |t|
+      t.references :organization, type: :uuid, null: false, foreign_key: true, index: true
+      t.references :contract, type: :uuid, null: false, foreign_key: true, index: true
+      t.references :rate_card, type: :uuid, null: false, foreign_key: true, index: true
+
+      t.date :billing_anchor_date, null: false
+      t.timestamp :next_billing_at, null: false
+      t.timestamp :started_at, null: false
```
tiagolupepic: working in subscription rate cards and how the units changes in update service. This `started_at` is more a `effective_on` than a started_at. I see more the Contract#starts_at the source of truth of when the contract started than the contract rate card.

My suggestion is to keep this model with only a single date, `effective_on`. And/or to facilitate, copy the start_at + end_at from Contract here using ruby.

A Contract contains a start time, during the agreement, both parts accept the dates + conditions (units / plan /prices and etc).
What changes during the time are the conditions, for example, starting on 1st Sept the units is 10. The starting date of a contract doesn't change.

app/models/contract.rb
rsempe: good spot 👍

app/models/contract.rb
rsempe: `ending_at` is the scheduled end the customer agreed to, `terminated_at` is the actual end stamped when the contract terminates - it's the same at the subscription level

app/models/contract.rb
rsempe: `starts_at` is the agreed start (what `subscription_at` was on `Subscription`, renamed since "subscription_at" makes no sense on a contract)

`started_at` is the activation timestamp, when `pending → active`

do you have something better for the naming of `starts_at`?

tiagolupepic: I'd make a bit more specific then:

agreement_at -> the timestamp of agreement, when the "shake hands" happen 🤝
started_at -> the timestamp when the agreements are valid and the "charges" can be processed 
ended_at -> the last timestamp and after this the charges would stop
termination_at -> the moment when the customer told the organization to finish the contract
canceled_at -> ? I guess is this a specific flow?

wdyt?

db/migrate/20260826235313_create_contract_rate_cards.rb
rsempe: interesting - having a look 👀

rsempe: that's a fair point 👍 

what about splitting this into a `contract_rate_card_units` table? 🤔 
`contract_rate_cards` is now just the attachment (anchor, phases, its own start/end — a card can join mid-contract), and `units` moved to an append-only `contract_rate_card_units(units, effective_from)`?

app/models/contract.rb
rsempe: I agree to merge `starts_at` and `started_at`, it was something from the v1, but not sure we need the operational flip timestamp 👍 

- `agreement_at` => `created_at`
- `starts_at`, `started_at` => `started_at`
- `termination_at` => `terminated_at`
- `canceled_at`

tiagolupepic: looks good! the created_at will be shared with AR as well?

tiagolupepic (approved):
:shipit:

PR #6239, opened 2026-08-27
refactor(fees): introduce metered items

tiagolupepic:
## Context

This PR refactors `Fees::ChargeService` to introduce a clearer boundary between invoice fee computation and the source of metered pricing data.

The current service is tightly coupled to `Charge` and `ChargeFilter` inputs. That makes it harder to extend the same billing computation flow to future metered sources, especially product-catalog-backed objects, without adding more conditional logic directly inside `ChargeService`.

The goal of this refactor is to keep the existing legacy charge behavior unchanged while preparing the service API for additional metered inputs.

## What Changed

- Introduced `Fees::ChargeService::MeteredItem` as the main input object for metered fee calculation.
- Introduced `Fees::ChargeService::Options` to group execution flags and context-specific behavior.
- Introduced `Fees::ChargeService::Sources::Charge` as the legacy `Charge` / `ChargeFilter` adapter.
- Updated `Fees::ChargeService` to consume `metered_item` instead of receiving `charge`, `boundaries`, `charge_filter`, and properties directly.
- Updated invoice services to build `MeteredItem.from_charge(...)` before calling `Fees::ChargeService`.
- Added validation around `metered_item`, `options`, tax inputs, and supported contexts.
- Added unit coverage for `MeteredItem`, `Options`, and the charge-backed source.
- Updated existing charge and invoice specs to assert the new interface.

### Naming: `MeteredItem`

The new abstraction is called `MeteredItem` because it describes the billing behavior rather than the underlying persistence model.

A `Charge` is the current source, but the service is really computing fees for something metered. This keeps the naming aligned with the product terminology around fixed vs metered pricing and avoids baking the legacy `Charge` model into the new abstraction.

### Legacy Charge Support Only

This PR intentionally keeps the runtime scope limited to existing `Charge` and `ChargeFilter` behavior.

It does not yet wire product-catalog objects into fee computation. Instead, it creates the adapter shape needed for that future work while keeping this PR focused and safer to review.

### Source Adapter

`Fees::ChargeService::Sources::Charge` owns all charge-specific behavior, including:

- charge properties vs charge filter properties
- pricing structure construction
- period ratio calculation
- pricing group keys
- aggregation filters
- charge filter matching and ignored filters

This keeps `Fees::ChargeService` focused on fee computation rather than charge-model translation.

### Options Object

`Options` groups service flags that were previously separate keyword arguments, including:

- `context`
- `apply_taxes`
- `calculate_projected_usage`
- `with_zero_units_filters`
- `usage_filters`
- `skip_adjusted_fees`

This makes the call signature smaller and makes validation of execution options explicit.

### Deprecated Contexts

`Fees::ChargeService::CONTEXTS` now contains only active charge-service contexts.

`DEPRECATED_CONTEXTS` contains `:refresh` and `:draft`, which are still accepted because some invoice-level flows pass those contexts through charge fee calculation. They are preserved for compatibility but kept separate to make the distinction explicit.

### DIFF
--- a/app/services/fees/charge_service.rb
+++ b/app/services/fees/charge_service.rb
@@ -4,69 +4,51 @@ module Fees
   class ChargeService < BaseService
     Result = BaseResult[:fees, :cached_aggregations]
 
-    # optional params:
-    # | usage_filters - UsageFilters
-    #  - context (:current_usage, :invoice_preview, :recurring) - to be moved in usage_filters
-    # | results_adjustments - to be implemented as a class to pass the following inside:
-    #  - with_zero_units_filters
-    #  - calculate_projected_usage
-    #  - apply_taxes
-    #  - filtered_aggregations
     def initialize(
       invoice:,
-      charge:,
+      metered_item:,
       subscription:,
-      boundaries:,
-      context: nil,
       cache_middleware: nil,
       filtered_aggregations: nil,
-      apply_taxes: false,
-      calculate_projected_usage: false,
-      with_zero_units_filters: true,
-      usage_filters: UsageFilters::NONE,
-      skip_adjusted_fees: false,
+      options: nil,
       plan: nil,
       customer: nil
     )
       @invoice = invoice
-      @charge = charge
+      @metered_item = metered_item
       @subscription = subscription
-      @boundaries = boundaries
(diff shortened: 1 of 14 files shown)


rsempe (approved):
LGTM 👏

PR #6256, opened 2026-08-31
feat: resolve taxes for product fees

endenis:
## Context

_Part of the "products and plans" project._

Product fees did not resolve taxes from rate cards.

## Description

- Resolve taxes using Rate Card -> Plan -> Customer -> Billing Entity.
- Use the fee's billing entity for Product fees.
- Mark affected draft invoices for refresh when rate rard taxes change.
- Keep taxes on finalized invoice unchanged.

### DIFF
--- a/app/services/fees/apply_taxes_service.rb
+++ b/app/services/fees/apply_taxes_service.rb
@@ -64,6 +64,7 @@ module Fees
     def applicable_taxes
       # organization.taxes - are all taxes created on the organization
       return customer.organization.taxes.where(code: tax_codes) if tax_codes
+      return product_taxes if fee.product?
       return fee.add_on.taxes if fee.add_on? && fee.add_on.taxes.any?
       return fee.charge.taxes if fee.charge? && fee.charge.taxes.any?
       return fee.fixed_charge.taxes if fee.fixed_charge? && fee.fixed_charge.taxes.any?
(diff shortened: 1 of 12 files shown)


app/services/fees/apply_taxes_service.rb
```diff
@@ -74,7 +75,20 @@ def applicable_taxes
       return customer.taxes if customer.taxes.any?
 
       # billing_entity.taxes - are the default taxes applied on the billing entity
-      Tax.joins(:billing_entities_taxes).where(billing_entities_taxes: {billing_entity_id: customer.billing_entity_id})
+      billing_entity_taxes(customer.billing_entity_id)
+    end
+
+    def product_taxes
```
tiagolupepic, line 81: it's not the scope of this PR but the previous `applicable_taxes` is a bit problematic already. We perform the `ApplyTaxesService` and accidentally does a N+1 https://github.com/getlago/lago-api/blob/main/app/services/invoices/compute_amounts_from_fees.rb#L20. I'd add a task in Linear to tackle this and avoid continue with the same issue for product_taxes.

app/services/rate_cards/apply_taxes_service.rb
```diff
@@ -39,5 +44,13 @@ def call
     def taxes_by_code
       @taxes_by_code ||= rate_card.organization.taxes.where(code: tax_codes).index_by(&:code)
     end
+
+    def refresh_draft_invoices
```
tiagolupepic, line 48: question also for @rsempe @brunomiguelpinto do we need to apply the refresh for the `rate_override` as well?

rsempe: no extra refresh needed: overrides carry no taxes of their own (taxes are just on the card, right?)

rsempe:
Should we have to add `validates :rate_card_rate, presence: true, if: :product?` on `Fee`?

rsempe (approved):

app/services/rate_cards/apply_taxes_service.rb
tiagolupepic: I see, in this case what is the rule to apply taxes? I thought was because of the pricing in `properties`.

PR #6257, opened 2026-08-31
fix(stripe): skip refunds on disputed and already-refunded charges

toommz:
Fixes [ING-600](https://linear.app/getlago/issue/ING-600/stripe-refunds-fail-on-disputed-and-already-refunded-charges)

## Context

`CreditNotes::Refunds::StripeCreateJob` sent refunds Stripe can never accept, and since `ApplicationJob` sets `retry: 0` the job died straight into the dead queue. Two production incidents: a charge with an open dispute (`charge_disputed`, US) and a charge already refunded outside Lago (`charge_already_refunded`, EU).

Our only guard was `invoice.payment_dispute_lost_at?`, written solely by `charge.dispute.closed`, and nothing checked how much of the charge was still refundable.

## Description

Two independent layers:

- **Stripe pre-check** — the charge's remaining refundable amount is read before calling `Stripe::Refund.create`, and `charge_disputed` / `charge_already_refunded` now return without raising. This half needs no Stripe-side change and fixes both incidents on deploy.
- **Inbound Stripe dispute events** — we now subscribe to `charge.dispute.created` / `charge.dispute.updated` and record dispute state on the invoice (new `invoices.payment_refund_blocked_at`), so the common case costs no API call.

The column is named for its effect rather than the event, because it is not a dispute record: an open *inquiry* leaves it `nil` (Stripe still allows refunds there), and a *lost* dispute leaves it set (refunds stay blocked). It holds current state only: history of past disputes is not modelled, and a charge carrying two concurrent disputes is not ref-counted (the widened rescue is the backstop for both). Because nothing else can clear the flag, the dispute handler reads the dispute's current state from Stripe rather than trusting a payload that may have been replayed out of order, falling back to the payload if Stripe cannot be reached.

No new *outgoing* Lago webhook is added: `config/webhook_event_types.yml` is untouched and there is no `invoice.payment_dispute_opened` event. The column is internal state, and is deliberately not exposed in the v1 serializer, GraphQL, `InvoicesQuery`, the CSV export or the `exports_invoices` view.

Every skip path stays visible: the credit note is marked `failed`, `credit_note.provider_refund_failure` fires, and the activity log is produced, exactly as when Stripe rejects the refund today.

Two points worth a look in review:

- **The flag mirrors Stripe's `is_charge_refundable`, not the event name.** `charge.dispute.created` also fires for inquiries, where Stripe still accepts refunds and refunding is the normal way to resolve one. Gating on the event alone would silently block legitimate refunds. This is also why the flag is deliberately *not* cleared on a lost dispute.
- **The pre-check fails open**, skipping only on a positive signal, so it can never become a new failure path.

Also adds `retry_on` for rate-limit/connection errors to the refund job, which the invoice payment job already has.

## Deploy

A migration enqueues `PaymentProviders::Stripe::RefreshWebhookJob` for every provider with a secret key, mirroring `20250626175249_refresh_stripe_webhooks`, so existing Stripe connections pick up the new events on upgrade without anyone running a rake task (`rake stripe:refresh_registered_webhooks` remains available to re-run it). The pre-check keeps the guard correct until then, and for self-hosted installs that never run it.

### DIFF
--- a/app/jobs/credit_notes/refunds/stripe_create_job.rb
+++ b/app/jobs/credit_notes/refunds/stripe_create_job.rb
@@ -5,6 +5,14 @@ module CreditNotes
     class StripeCreateJob < ApplicationJob
       queue_as "providers"
 
+      # NOTE: safe to retry because stripe deduplicates on the idempotency key that
+      #       Stripe::Refund.create sends (the credit note id), so a replayed request returns
+      #       the original refund instead of issuing a second one. We enforce nothing on our
+      #       side, and stripe only honours that key for 24h; the backoff below tops out around
+      #       20 minutes, so raising `attempts` much further would void the guarantee.
+      retry_on ::Stripe::RateLimitError, wait: :polynomially_longer, attempts: 6
+      retry_on ::Stripe::APIConnectionError, wait: :polynomially_longer, attempts: 6
+
       def perform(credit_note)
         result = CreditNotes::Refunds::StripeService.new(credit_note).create
         result.raise_if_error!
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -553,6 +553,18 @@ class Invoice < ApplicationRecord
     save!
   end
 
+  # NOTE: mirrors stripe's `is_charge_refundable`: set while a dispute prevents a refund,
+  #       cleared as soon as the charge becomes refundable again.
+  def mark_refund_as_blocked!(timestamp = Time.current)
+    self.payment_refund_blocked_at ||= timestamp
+    save!
+  end
+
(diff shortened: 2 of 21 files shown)


toommz:
Superseded by a four-PR stack, same code re-cut into reviewable steps (the tip reproduces this branch's tree exactly):

- #6442 — `invoices.payment_refund_blocked_at` column
- #6443 — the refund fix, which is what closes both production incidents and deploys on its own
- #6444 — dispute tracking via `charge.dispute.created` / `.updated`
- #6445 — refreshing registered Stripe webhooks

The review history here is worth reading alongside #6444, which is where most of it applies.

PR #6259, opened 2026-08-31
refactor(events): resolve billing filters abstraction FilterTarget

tiagolupepic:
## Context

Billing-period filtering needs to support more than charge-backed targets in later product-catalog billing work. The existing service mixed charge event lookup, filter matching, recurring seeding, and caller-facing result shape in one class, which made the next target source harder to add safely.

## Description

Extract the charge-backed orchestration into `Events::BillingPeriodFilters::ChargesResolver` and keep Events::BillingPeriodFilterService as a thin generic service initialized with a resolver. Add a shared `FilterTarget` contract and generic matching services, while keeping ChargeFilters services as compatibility wrappers.

The result now exposes filter_targets keyed by target_key instead of charges keyed by charge id. Charge callers use `Charge#target_key and Events::BillingPeriodFilterService.for_charges(...).call!` so phase 1 remains charge-only and behavior-preserving.

Important decisions captured in the planning doc: use resolver terminology, keep direct factory methods instead of a ResolveService, use target_filter: for FilterTarget arguments, avoid ProductFilters-specific services, keep charge/product target_key ownership on models, and leave filter-index optimization to later phases.


app/services/events/billing_period_filter_service.rb
```diff
@@ -2,308 +2,35 @@
 
 module Events
   class BillingPeriodFilterService < BaseService
-    Result = BaseResult[:charges]
+    Result = BaseResult[:filter_targets]
+
+    def self.for_charges(subscription:, boundaries:, codes: nil, with_last_seen_at: true)
```
brunomiguelpinto: `for_charges(...).call!` hits the **instance** `call!`, which skips `call_with_middlewares` — the `LogTracerMiddleware`/`DatadogMiddleware` spans that the previous class-level `.call!` produced are gone on all four callers.

Fix: let the factory return the result — `call!(resolver: BillingPeriodFilters::ChargesResolver.new(subscription:, boundaries:, codes:, with_last_seen_at:))` — and drop the trailing `.call!` at the call sites.

brunomiguelpinto (approved):

PR #6262, opened 2026-08-31
feat(usage): serve usage from ClickHouse buckets [ING-586]

vincent-pochet:
> Stacked on #6408, review that one first.

## Context

Aggregating raw events dominates current usage: around ten seconds at the median and minutes in the
tail, paid by every consumer, from the usage endpoints to alerts and wallet refresh. The
`usage_buckets_15m` table now exists, but nothing decides which charges it can answer, and an
aggregation has no way to be handed a result it did not compute itself.

## Description

- Add `RealtimeUsage`, the one predicate for which charges the buckets can answer, and the gates: a
  premium license, the ClickHouse events store, `LAGO_REALTIME_USAGE_ENABLED`, the per-organization
  `realtime_usage` flag, and no event store override.
- Add `Events::Stores::UsageBucketSet`, one computation's totals indexed by charge and charge
  filter, held by the event store provider.
- Hand precomputed results to an aggregation through the options of `aggregate`, replacing the
  `injected_sum_result` accessors.

Billing is never served this way: it reads closed windows, where being wrong means a wrong invoice
rather than a stale reading. The events store gates too, because it and the buckets share one
stream while Postgres deduplicates `events` on insert and would disagree.

Nothing prefetches buckets yet, so this changes no behaviour.


vincent-pochet:
> 
> HOLD — the realtime-usage path is not connected to any production current-usage computation.
> 
>     * No production caller supplies `usage_buckets`/`current_usage` to the provider, calls `precomputed_options_for`, or merges those options into `aggregate`; enabling the new flags therefore still reads raw events. Wire bucket loading and the provider options through the current-usage call sites, with integration coverage proving supported count/sum charges use bucket totals and fallback cases still use the event store.

state prefetch is the follow-up PR, flags stay off until then. Dead-code-until-wired is the point of stacking.

app/services/events/stores/provider.rb
```diff
@@ -35,9 +37,41 @@ def deduplicate
...
+          precomputed_aggregation: usage_buckets.aggregation_result_for(charge_id:, charge_filter_id:),
+          precomputed_grouped_aggregations: usage_buckets.grouped_aggregation_results_for(charge_id:, charge_filter_id:)
+        }
+      end
+
+      def served_from_buckets?(charge:, boundaries:, filters: {})
+        return false if usage_buckets.blank?
+        return false unless current_usage
+        return false if boundaries[:max_timestamp].present?
+        return false unless RealtimeUsage.enabled?(organization)
```
toommz, line 54: I would have checked that first 🤔

toommz (approved):

app/services/events/stores/provider.rb
vincent-pochet: Yeah, good catch

PR #6263, opened 2026-08-31
feat(streaming): Kinesis streaming destinations (POC)

toommz:
## Context

POC for the "Streaming Kinesis for Synthesia" project (P1 + P2): let a single organization receive entity change events, the same payloads Lago already sends over HTTP webhooks, on an AWS Kinesis data stream in a customer-owned AWS account via cross-account `sts:AssumeRole`.

P0 (usage consumption records) is deliberately out of scope: those are materialized in the Go events processor, not here.

There is no API, no GraphQL and no UI. Destinations are seeded by an operator with a rake task.

## Description

A new `streaming_destinations` table follows the existing house pattern for provider-shaped records (`Integrations::BaseIntegration`, `PaymentProviders::BaseProvider`): one table, STI `type`, org-scoped `code`, `settings` jsonb via `SettingsStorable`, encrypted `secrets` via `SecretsStorable`, `Discard::Model`. No AWS-shaped columns, so a second transport would need no migration.

Delivery hangs off the existing webhook fan-out point rather than duplicating it. `Webhooks::BaseService#call` already builds a fully serialized payload and already filters per destination by event type, so the Kinesis path reuses that exact payload: one serialization of an entity change, not one per transport. No new serializers.

`webhook_endpoints` was intentionally not reused: it has `webhook_url NOT NULL`, a unique index on `(webhook_url, organization_id)`, a non-optional `Webhook belongs_to :webhook_endpoint`, and a documented `LIMIT` of 10.

### The two guard conditions that changed, and why

This is the part to review most carefully. Two guards previously short-circuited on "this organization has no webhook endpoints", and **both** had to widen. Changing only one makes the feature fail silently, with no error raised anywhere.

1. `Webhooks::BaseService#call` — was `return if current_organization.webhook_endpoints.none?`
2. `SendWebhookJob.perform_later` (the class-method override) — was `return if (webhook_id.nil? || webhook_id == UNDEFINED) && object.organization.webhook_endpoints.none?`

Both now consult a single predicate, `Organization#webhook_destinations?`, so they cannot drift. The predicate lives on `Organization` rather than in the service because the two call sites reach the organization by different paths (`current_organization` vs `object.organization`), so that is the only placement where one definition genuinely serves both.

`spec/services/webhooks/base_service_spec.rb` and `spec/jobs/send_webhook_job_spec.rb` each carry an explicit regression guard for the zero-endpoints-one-destination case.

Nothing else about the existing HTTP delivery path changed.

### Ordering: read this before writing a consumer

**Lago does not guarantee stream order for concurrent updates to the same object.** Concurrent Sidekiq workers can `PutRecord` two updates to the same object in either order, and Kinesis preserves *receive* order, not *commit* order. The partition key groups related records onto one shard; it does not order them.

Every streamed payload therefore carries an added **`emitted_at`** (ISO 8601, millisecond precision), stamped once per emission. **Consumers must use `emitted_at` for last-write-wins** and must not assume stream position implies recency.

One honest caveat: `emitted_at` is stamped when the webhook service runs, not when the database transaction committed. For two racing updates to the same object it is a much better ordering key than stream position, but it is not a true monotonic version, and neither is the object's own `updated_at`. This is most acute for wallets, whose balances mutate continuously server-side. A real version column is the correct fix and is out of scope for a POC.

Partition key defaults to the organization id. Three services override it to the customer's `external_id`, so a consumer can process one customer's records on one worker:

- `Webhooks::Subscriptions::UpdatedService`
- `Webhooks::Wallets::UpdatedService`
- `Webhooks::Customers::UpdatedService`

### Entitlements

Worth flagging, with a correction to how this was originally scoped: `subscription.updated` **does** carry entitlements, because `Webhooks::Subscriptions::UpdatedService` serializes `includes: %i[plan customer entitlements]`.

The real gap is narrower but still real: **no webhook event type fires when entitlements change on their own.** A consumer sees entitlement state only when something else touches the subscription, so it cannot reliably reconstruct per-subscription entitlements from this stream. Adding subscription-level entitlement events was explicitly out of scope here and is tracked separately.

### Delivery

`StreamingDestinations::DeliverJob#perform` stays provider-blind: it asks the destination for its `deliver_service_class` and calls it. No `case` on type.

`StreamingDestinations::Kinesis::DeliverService`:

- Rejects a record whose serialized payload exceeds 1 MB by failing the result with `record_too_large` and reporting to Sentry, rather than raising. Kinesis hard-rejects oversized records, so a retry storm helps nobody. The job deliberately does not raise on a failed result for the same reason; genuine transport errors arrive as exceptions and are retried.
- Caches the `Aws::AssumeRoleCredentials` object per destination across jobs, letting the SDK auto-refresh handle expiry. This is required, not an optimization: a fresh credentials object per job means one STS call per record, which hits STS throttling long before Kinesis notices any load.

## Deliberate deviations from the plan

Three, all small, flagged for review rather than buried:

1. **One migration, not the two-migration `validate_foreign_key` pair.** That convention exists for adding an FK to an existing populated table (as in the recent `quote_versions` pair). For a brand-new empty table the repo consistently uses inline `foreign_key: true` inside `create_table` (`create_rate_overrides`, `create_orders`, `create_billing_object_connections`, `create_invoice_connections`), and strong_migrations does not object. Following the pair here would have been cargo-culting.

2. **The credential cache key is `(id, role_arn, external_id, region)`, not the destination id alone.** Keying on the id alone would keep serving credentials for a role that had since been edited, a stale-credential bug needing a redeploy to clear. Still one STS call per destination.

3. **`standardrb --fix` was not run repo-wide.** All files in this branch are clean, but `origin/main` currently has 24 pre-existing offenses in unrelated files (`Performance/BigDecimalWithNumericArgument` and friends). Fixing them here would have dragged 24 unrelated files into this PR.

## Follow-ups deliberately deferred

- **Dedicated Sidekiq queue.** `DeliverJob` uses `queue_as :webhook` for now. A dedicated queue is the right end state for blast-radius isolation, so a backed-up stream cannot starve HTTP webhook delivery, but it needs a new process config under `config/sidekiq/` plus a deployment change in `lago-infrastructure`. Marked with a `TODO` in the job.
- **CRUD API / UI** for managing destinations. Seeded by rake task for the POC.
- **Delivery-history persistence.** No `Webhook`-equivalent rows for stream deliveries.

## Seeding

```
bundle exec rake "streaming_destinations:upsert_kinesis[<org_id>,<code>]" \
  STREAM_ARN=arn:aws:kinesis:eu-west-1:111122223333:stream/lago \
  REGION=eu-west-1 \
  ROLE_ARN=arn:aws:iam::111122223333:role/LagoStreamingWriter \
  EXTERNAL_ID=some-shared-secret \
  EVENT_TYPES=subscription.updated,wallet.updated
```

Idempotent on `(organization_id, code)`. Re-running updates only what is supplied, so a partial re-run does not blank correct settings. Omit `EVENT_TYPES` to subscribe to everything; pass `EVENT_TYPES="*"` to reset an existing destination back to all.

Verified by hand against the test database: create, then a region-only update preserving the other settings, then an `EVENT_TYPES="*"` reset.

## Testing

New: model specs (STI resolution, `subscribed?` across nil/matching/non-matching/empty, code uniqueness scoped to org and ignoring discarded records, settings round-tripping, `external_id` staying out of `secrets`), delivery service specs (`stub_responses: true`, asserting partition key and decoded record body, the oversized path failing without raising, and credentials not rebuilt on a second call for the same destination), job specs, fan-out specs, and one partition-key spec per overriding service.

`spec/services/webhooks` runs **119 examples, 0 failures**, confirming no existing webhook behaviour regressed.

Across `spec/models/streaming_destinations spec/services/streaming_destinations spec/services/webhooks spec/jobs`: 863 examples, 7 failures, all pre-existing and environmental on a local machine, none in webhook or streaming code:

- 4 × `SendEmailJob` — `SMTP From address may not be blank: nil`, missing local mail configuration; fails identically in isolation.
- 1 × `Clock::InboundWebhooksCleanupJob` — a stale `activejob_uniqueness` lock in the shared dev Redis (23 such stale locks present).
- 2 × `Clock::EventsValidationJob` — pass in isolation (3 examples, 0 failures); cross-spec lock pollution in that same Redis.

### DIFF
--- a/app/jobs/send_webhook_job.rb
+++ b/app/jobs/send_webhook_job.rb
@@ -98,11 +98,16 @@ class SendWebhookJob < ApplicationJob
   UNDEFINED = Object.new.freeze
   private_constant :UNDEFINED
 
-  # Override the default perform_later to avoid creating webhooks if no webhook endpoints are present.
+  # Override the default perform_later to avoid creating webhooks if the organization
+  # has nowhere to deliver them.
   #
   # This will prevent enqueueing jobs only to return early from the jobs.
+  #
+  # NOTE: shares Organization#webhook_destinations? with Webhooks::BaseService#call
+  #       on purpose. If the two guards drift, one of the transports goes silent
+  #       with no error raised anywhere.
   def self.perform_later(webhook_type, object, options = UNDEFINED, webhook_id = UNDEFINED)
-    return if (webhook_id.nil? || webhook_id == UNDEFINED) && object.organization.webhook_endpoints.none?
+    return if (webhook_id.nil? || webhook_id == UNDEFINED) && !object.organization.webhook_destinations?
 
     args = [webhook_type, object, options, webhook_id].filter { |arg| arg != UNDEFINED }
     super(*args)
--- a/app/models/streaming_destinations/base_destination.rb
+++ b/app/models/streaming_destinations/base_destination.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+module StreamingDestinations
+  # NOTE: Abstract STI parent, should not be instantiated directly
+  class BaseDestination < ApplicationRecord
+    include PaperTrailTraceable
+    include SecretsStorable
+    include SettingsStorable
+    include Discard::Model
+
+    self.discard_column = :deleted_at
+    default_scope -> { kept }
+
+    self.table_name = "streaming_destinations"
+
+    belongs_to :organization
+
+    validates :code, uniqueness: {conditions: -> { kept }, scope: :organization_id}
+    validates :name, presence: true
+
+    # A nil event_types subscribes the destination to every event type, matching
+    # WebhookEndpoint's behaviour.
+    def subscribed?(event_type)
+      return true if event_types.nil?
+
+      event_types.include?(event_type)
+    end
+
+    def deliver_service_class
+      raise NotImplementedError
+    end
+  end
+end
+
+# == Schema Information
+#
+# Table name: streaming_destinations
+# Database name: primary
+#
+#  id              :uuid             not null, primary key
+#  code            :string           not null
+#  deleted_at      :datetime
+#  event_types     :string           is an Array
+#  name            :string           not null
+#  secrets         :string
+#  settings        :jsonb            not null
+#  type            :string           not null
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#  organization_id :uuid             not null
+#
+# Indexes
+#
+#  index_streaming_destinations_on_deleted_at                (deleted_at)
+#  index_streaming_destinations_on_organization_id           (organization_id)
+#  index_unique_streaming_destinations_on_organization_code  (organization_id,code) UNIQUE WHERE (deleted_at IS NULL)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (organization_id => organizations.id)
+#
--- a/Gemfile
+++ b/Gemfile
@@ -105,6 +105,9 @@ gem "google-cloud-storage", require: false
 # stdlib json, so multi_json is no longer pulled in transitively and must be pinned.
 gem "multi_json", require: false
 
+# Streaming
+gem "aws-sdk-kinesis", require: false
+
 # Templating
 gem "slim"
 gem "slim-rails"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -122,7 +122,7 @@ GEM
     awesome_print (1.9.2)
     aws-eventstream (1.4.0)
     aws-partitions (1.1208.0)
-    aws-sdk-core (3.241.4)
+    aws-sdk-core (3.254.1)
       aws-eventstream (~> 1, >= 1.3.0)
       aws-partitions (~> 1, >= 1.992.0)
       aws-sigv4 (~> 1.9)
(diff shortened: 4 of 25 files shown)


app/jobs/send_webhook_job.rb
```diff
@@ -98,11 +98,16 @@ class SendWebhookJob < ApplicationJob
...
+  # has nowhere to deliver them.
   #
   # This will prevent enqueueing jobs only to return early from the jobs.
+  #
+  # NOTE: shares Organization#webhook_destinations? with Webhooks::BaseService#call
+  #       on purpose. If the two guards drift, one of the transports goes silent
+  #       with no error raised anywhere.
   def self.perform_later(webhook_type, object, options = UNDEFINED, webhook_id = UNDEFINED)
-    return if (webhook_id.nil? || webhook_id == UNDEFINED) && object.organization.webhook_endpoints.none?
+    return if (webhook_id.nil? || webhook_id == UNDEFINED) && !object.organization.webhook_destinations?
```
toommz: ⚠️

app/models/streaming_destinations/base_destination.rb
toommz: STI??? ⚠️

PR #6266, opened 2026-09-01
test(invoices): cover Invoice#recurring_breakdown

vincent-pochet:
## Context

`Invoice#recurring_breakdown` builds the prorated breakdown rendered on invoice PDFs for recurring
metrics, and is reachable only from `app/views/templates/invoices/*`, so no request spec touches it
either. Its existing coverage asserted that a period with no event returns an empty array, which
never reaches the dispatch between the two breakdown services nor the persisted-units path reading
events from before the billing window.

## Description

- Cover the `sum_agg` branch with an event before `charges_from_datetime` and one inside the period,
  asserting date, action, amount, duration and total duration of both items.
- Cover the `unique_count_agg` branch on the same window shape.
- Cover an unsupported aggregation type raising `NotImplementedError`.

The persisted item carries the whole period while the in-period one is prorated from its own date,
which is the part the empty-period case could not show:

```
duration = charges_to_datetime.to_date + 1 day - item_date

persisted (2023-05-01): 2023-05-31 + 1 day - 2023-05-01 = 31   # full period
in-period (2023-05-11): 2023-05-31 + 1 day - 2023-05-11 = 21
```

Spec only, and independent of #6261, which rewired this method onto `Events::Stores::Provider`
without touching `spec/models/invoice_spec.rb`. The unsupported-type case uses a weighted sum on a
non-prorated charge: `max_agg`, `latest_agg` and `count_agg` are refused by
`BillableMetric#validate_recurring`, and `prorated: true` is refused for weighted sum.

### DIFF
--- a/spec/models/invoice_spec.rb
+++ b/spec/models/invoice_spec.rb
@@ -1149,6 +1149,122 @@ RSpec.describe Invoice do
         expect(invoice.recurring_breakdown(fee)).to eq([])
       end
     end
+
+    context "with events on a prorated recurring charge" do
+      let(:organization) { create(:organization) }
+      let(:customer) { create(:customer, organization:) }
+      let(:invoice) { create(:invoice, organization:, customer:) }
+      let(:charge) { create(:standard_charge, organization:, plan: subscription.plan, billable_metric:, prorated: true) }
+
+      let(:subscription) do
+        create(
+          :subscription,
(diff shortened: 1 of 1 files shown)


toommz (approved):

tiagolupepic (approved):
:shipit:

PR #6267, opened 2026-09-01
feat(billing): Billing engine v2 date services

brunomiguelpinto:
Contract billing needs consistent dates across rate changes, phase overrides and long billing histories. This adds a rate-card schedule that produces billable segments and can resume near a requested date without walking every historical cycle.

### Behaviour

- `CycleWalker` owns traversal and resume. It stops at rate and phase transitions, recalculates the calendar when the interval changes, and reuses calendars between calls. Walking the complete history remains supported.
- `Schedule` splits cycles at rate changes and assigns each priced segment its own billing date: start for advance, end for arrears. Unpriced windows still consume cycle indices.
- Proration uses customer-local calendar days and the containing interval. Schedule windows have exclusive ends; `BillingSegment.inclusive_end` and `.exclusive_end` convert stored boundaries.
- `BuildScheduleService` reads `ContractRateCard`, resolves ordered rates and card or plan phases, and resumes from that attachment's latest persisted cycle. The last cycle is intentionally returned again, so the writer must deduplicate persisted segments. Separate attachments do not share billing history.
- Card effective and end dates use the customer's timezone. An inclusive card end date covers the whole final day; an earlier contract end or explicit termination instant takes precedence.
- Contract materialization seeds the initial billing clock from the schedule, including backdated contracts and phase interval overrides.

### Public schedule API

| Method | Purpose |
| --- | --- |
| `segments_due_by(timestamp)` | Return priced segments due by that instant. |
| `next_billing_at(after:)` | Find the next billing instant strictly after the given time. |
| `billing_at_covering(timestamp)` | Find the billing date of the segment being served, or the next priced segment. |
| `consumed_ratio(segment:, at:)` | Measure consumption against the original billed segment. |

### Integration with main

This PR is based directly on `main` and contains only the date engine and its contract integration. It uses the existing contract models and billing-segment table. Existing migrations, Ruby 4.0.6 and dependencies are unchanged.

A separate migration removes the unused `pricing_unit_id` and `rate_card_rate_id` indexes from `billing_segments`, with concurrent operations and a reversible rollback.

The invoice writer and processor are outside this PR.

### Validation

- 422 examples passed with Ruby 4.0.6 in a temporary API container and isolated database: 113 integration/model examples and 309 date/walker/resume examples.
- Coverage includes month ends, leap years, DST, rate and phase transitions, unpriced periods, termination, persisted history, contract materialization and equivalence between full and resumed walks.
- RuboCop: 32 files, no offenses.
- Migration up/down/up verified with an existing billing segment; both indexes were restored on rollback and the row was unchanged.

### DIFF
--- a/app/models/billing_segment.rb
+++ b/app/models/billing_segment.rb
@@ -5,6 +5,13 @@
 class BillingSegment < ApplicationRecord
   include Currencies
 
+  MICROSECOND = Rational(1, 1_000_000)
+
+  # Schedule windows have exclusive ends; the database overlap constraint uses inclusive ends.
+  def self.inclusive_end(instant) = instant - MICROSECOND
+
+  def self.exclusive_end(instant) = instant + MICROSECOND
+
   STATUSES = {
     pending: "pending",
     processing: "processing",
--- a/app/models/billing_segment.rb
+++ b/app/models/billing_segment.rb
@@ -124,8 +124,6 @@ end
 #  index_billing_segments_on_customer_id            (customer_id)
 #  index_billing_segments_on_invoice_id             (invoice_id)
 #  index_billing_segments_on_organization_id        (organization_id)
-#  index_billing_segments_on_pricing_unit_id        (pricing_unit_id)
-#  index_billing_segments_on_rate_card_rate_id      (rate_card_rate_id)
 #  index_billing_segments_on_rate_override_id       (rate_override_id)
 #
 # Foreign Keys
--- a/app/services/billing/calendar.rb
+++ b/app/services/billing/calendar.rb
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+module Billing
+  # A ruler: boundaries every `interval` from the start of the anchor day, in the customer's
+  # timezone. Anchored Feb 1 in "America/New_York", boundary 0 is Feb 1 05:00 UTC.
+  class Calendar
+    attr_reader :anchor_date, :interval
+
+    def initialize(anchor_date:, interval:, timezone:)
+      @anchor_date = anchor_date
+      @anchor = anchor_date.in_time_zone(timezone).beginning_of_day
+      @interval = interval
+      @timezone = timezone
+    end
+
+    # The half-open interval `timestamp` falls in, as [boundary, next boundary).
+    def interval_containing(timestamp)
+      index = index_of_interval(timestamp)
+
+      interval_with_index_starts_at(index)...interval_with_index_starts_at(index + 1)
+    end
+
+    # How many intervals separate two instants on this ruler.
+    #
+    # @example a monthly ruler anchored Jan 31, whose boundaries clamp to Feb 28
+    #   intervals_between(Feb 28, Mar 31)  # => 1
+    #   intervals_between(Feb 28, Mar 28)  # => 0   still inside the interval that opened Feb 28
+    def intervals_between(from, to)
+      index_of_interval(to) - index_of_interval(from)
+    end
+
+    # The boundary `steps` intervals after the one containing an instant.
+    #
+    # @example a monthly ruler anchored Jan 31
+    #   boundary_after(Feb 10, 2)  # => Mar 31
+    def boundary_after(timestamp, steps)
+      interval_with_index_starts_at(index_of_interval(timestamp) + steps)
+    end
+
+    # The first boundary at or after an instant — when a change landing mid-interval takes hold.
+    #
+    # @example a monthly ruler anchored Jan 1
+    #   boundary_at_or_after(Feb 1)   # => Feb 1   already a boundary
+    #   boundary_at_or_after(Feb 10)  # => Mar 1
+    def boundary_at_or_after(timestamp)
+      window = interval_containing(timestamp)
+
+      (window.begin == timestamp) ? window.begin : window.end
+    end
+
+    # The share of its containing interval that the half-open window [from, to) covers.
+    #
+    # @return [Float] denominator is the CONTAINING interval, never a fixed 30 days (decision #55)
+    #
+    # @example a 30-day interval anchored Jun 1
+    #   proration_ratio(Jun  1, Jul 1)  # => 1.0
+    #   proration_ratio(Jun 16, Jul 1)  # => 0.5
+    def proration_ratio(from, to)
+      raise ArgumentError, "window end #{to} precedes its start #{from}" if to < from
+
+      window = interval_containing(from)
+      raise ArgumentError, "window [#{from}, #{to}) crosses the boundary at #{window.end}" if to > window.end
+
+      Days.between(from, to, timezone:).fdiv(Days.between(window.begin, window.end, timezone:))
+    end
+
+    private
+
+    attr_reader :anchor, :timezone
+
+    # Which interval `timestamp` falls in — a position on this ruler, not a cycle number. The
+    # anchor is a reference day, not a start date, so instants before it have negative indices.
+    def index_of_interval(timestamp)
+      interval.steps_between(anchor, timestamp.in_time_zone(timezone))
+    end
+
+    def interval_with_index_starts_at(index)
+      @boundaries ||= {}
+      @boundaries[index] ||= interval.advance(anchor, index)
+    end
+  end
+end
--- a/app/services/billing/interval.rb
+++ b/app/services/billing/interval.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+module Billing
+  # A cadence: how far one billing period runs.
+  class Interval < Data.define(:count, :unit)
+    UNITS = %i[day week month year].freeze
+
+    # The cadence a rate bills on, with an override laid over it field by field.
+    #
+    # @param override [RateOverride, nil] either column alone wins
+    # @return [Interval]
+    #
+    # @example a `3 month` rate
+    #   from(rate)                                   # => 3 month
+    #   from(rate, override: unit week)              # => 3 week
+    #   from(rate, override: count 2)                # => 2 month
+    def self.from(rate, override: nil)
+      new(
+        count: override&.billing_interval_count || rate.billing_interval_count,
+        unit: override&.billing_interval_unit || rate.billing_interval_unit
+      )
+    end
+
+    def initialize(count:, unit:)
+      unit = unit.to_sym if unit.respond_to?(:to_sym)
+
+      raise ArgumentError, "unknown interval unit #{unit.inspect}, expected one of #{UNITS.join(", ")}" unless UNITS.include?(unit)
+      raise ArgumentError, "interval count must be a positive integer, got #{count.inspect}" unless count.is_a?(Integer) && count.positive?
+
+      super
+    end
+
+    # Move an instant by whole intervals.
+    #
+    # @param steps [Integer] may be negative
+    # @return [Time] month-end clamped
+    #
+    # @example monthly
+    #   advance(Jan 31, 1)   # => Feb 28   clamped, not Mar 3
+    #   advance(Jan 15, -2)  # => Nov 15 of the year before
+    def advance(timestamp, steps)
+      units = count * steps
+
+      case unit
+      when :day then timestamp + units.days
+      when :week then timestamp + units.weeks
+      when :month then timestamp + units.months
+      when :year then timestamp + units.years
+      end
+    end
+
+    # Whole intervals between two instants.
+    #
+    # @return [Integer] may be negative
+    #
+    # @example monthly, and the correction that earns this method
+    #   steps_between(Jan 31, Feb 15)  # => 0   the calendar changed month; a month has not passed
+    #   steps_between(Jan 31, Feb 28)  # => 1   here it has
+    #
+    # @example every 3 months, and weekly
+    #   steps_between(Jan 1, Aug 1)    # => 2   7 months hold two whole 3-month intervals
+    #   steps_between(Jun 1, Jun 20)   # => 2   weekly
+    def steps_between(from, to)
+      estimate = calendar_steps_between(from, to)
+
+      if advance(from, estimate) > to
+        estimate - 1
+      else
+        estimate
+      end
+    end
+
+    private
+
+    # Whole units by the calendar, then how many intervals of `count` fit in them.
+    #
+    # @return [Integer] over by at most one, which #steps_between corrects
+    #
+    # @example monthly
+    #   calendar_steps_between(Jan 31, Feb 15)  # => 1   the month number changed
+    def calendar_steps_between(from, to)
+      whole_units = case unit
+      when :day then (to.to_date - from.to_date).to_i
+      when :week then (to.to_date - from.to_date).to_i / 7
+      when :month then ((to.year - from.year) * 12) + (to.month - from.month)
+      when :year then to.year - from.year
+      end
+
+      whole_units / count
+    end
+  end
+end
--- a/app/services/billing/rate_cards/build_schedule_service.rb
+++ b/app/services/billing/rate_cards/build_schedule_service.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+module Billing
+  module RateCards
+    class BuildScheduleService < BaseService
+      class MismatchedPlanRateCard < StandardError; end
+
+      Result = BaseResult[:schedule]
+
+      def initialize(contract_rate_card:, plan_rate_card: nil, ends_at: nil)
+        @contract_rate_card = contract_rate_card
+        @plan_rate_card = plan_rate_card
+        @ends_at = ends_at
+        super
+      end
+
+      def call
+        rate_card = contract_rate_card.rate_card
+        rates = rate_card.ordered_rates.to_a
+
+        if rates.empty?
+          result.not_found_failure!(resource: "rate")
+        else
+          terms = Terms.new(timing: rate_card.billing_timing, prorated: rate_card.proration?)
+
+          result.schedule = Schedule.new(
+            rates:,
+            terms:,
+            phases:,
+            resume_at: contract_rate_card.billing_segments.maximum(:cycle_started_at),
+            starts_at: contract_rate_card.effective_date.in_time_zone(timezone),
+            ends_at: schedule_ends_at,
+            anchor_date: contract_rate_card.billing_anchor_date,
+            timezone:
+          )
+        end
+
+        result
+      rescue ArgumentError => error
+        result.service_failure!(code: "invalid_billing_schedule", message: error.message)
+      end
+
+      private
+
+      attr_reader :contract_rate_card, :plan_rate_card, :ends_at
+
+      def timezone
+        contract_rate_card.contract.customer.applicable_timezone
+      end
+
+      def schedule_ends_at
+        # Contract card dates are inclusive; the walker expects an exclusive instant.
+        card_end = contract_rate_card.ended_date&.next_day&.in_time_zone(timezone)
+
+        [ends_at, card_end, contract_rate_card.contract.ended_at].compact.min
+      end
+
+      def phases
+        rate_phases = ::ContractRateCards::ResolveRatePhasesService.call!(
+          contract_rate_card:,
+          plan_rate_card: resolved_plan_rate_card
+        ).rate_phases
+
+        configured = rate_phases.map do |rate_phase|
+          Phase.new(
+            code: rate_phase.code,
+            billing_interval_cycle_count: rate_phase.billing_interval_cycle_count,
+            rate_override: rate_phase.rate_override
+          )
+        end
+
+        if configured.last&.unbounded?
+          configured
+        else
+          configured + [Phase.default]
+        end
+      end
+
+      def resolved_plan_rate_card
+        if plan_rate_card.nil?
+          contract_rate_card.contract.catalog_plan&.applied_rate_cards
+            &.find { it.rate_card_id == contract_rate_card.rate_card_id }
+        elsif plan_rate_card.rate_card_id != contract_rate_card.rate_card_id
+          raise MismatchedPlanRateCard, "plan_rate_card #{plan_rate_card.id} prices rate card " \
+            "#{plan_rate_card.rate_card_id}, not #{contract_rate_card.rate_card_id}"
+        else
+          plan_rate_card
+        end
+      end
+    end
+  end
+end
--- a/app/services/billing/rate_cards/schedule.rb
+++ b/app/services/billing/rate_cards/schedule.rb
@@ -1,0 +1,112 @@
+# frozen_string_literal: true
+
+module Billing
+  module RateCards
+    class Schedule
+      def initialize(rates:, terms:, phases:, starts_at:, anchor_date:, timezone:, ends_at: nil, resume_at: nil)
+        if resume_at && resume_at < starts_at.in_time_zone(timezone).beginning_of_day
+          raise ArgumentError, "resume_at #{resume_at} precedes the card's start"
+        end
+
+        @rates = rates
+        @terms = terms
+        @timezone = timezone
+        @resume_at = resume_at
+        @walker = CycleWalker.new(rates:, phases:, starts_at:, anchor_date:, timezone:, ends_at:)
+      end
+
+      # A rate change can make a segment due before its cycle ends.
+      def segments_due_by(timestamp)
+        cycles = walker.walk_to(timestamp, from: resume_at)
+
+        billable_segments_of(cycles).select do |segment|
+          segment.billing_at <= timestamp
+        end
+      end
+
+      # Next billing instant strictly after the given time, or nil when billing has ended.
+      def next_billing_at(after:)
+        cycle = walker.resume(after)
+
+        while cycle
+          segment = billable_segments_of([cycle]).find do |candidate|
+            candidate.billing_at > after
+          end
+
+          if segment
+            return segment.billing_at
+          end
+
+          cycle = walker.advance
+        end
+
+        nil
+      end
+
+      # Bill the segment being served, or the first future one if pricing has not started.
+      def billing_at_covering(timestamp)
+        segment = segment_at(timestamp)
+
+        segment&.billing_at || next_billing_at(after: timestamp)
+      end
+
+      # Measure consumption against the original billed segment, whose end is exclusive.
+      def consumed_ratio(segment:, at:)
+        segment_days = Days.between(segment.started_at, segment.ended_at, timezone:)
+
+        if segment_days.zero?
+          1.0
+        else
+          consumed_until = at.clamp(segment.started_at, segment.ended_at)
+          consumed_days = Days.between(segment.started_at, consumed_until, timezone:)
+
+          consumed_days.fdiv(segment_days)
+        end
+      end
+
+      private
+
+      attr_reader :rates, :terms, :timezone, :resume_at, :walker
+
+      def segment_at(timestamp)
+        cycle = walker.resume(timestamp)
+
+        if cycle
+          billable_segments_of([cycle]).find do |segment|
+            (segment.started_at...segment.ended_at).cover?(timestamp)
+          end
+        end
+      end
+
+      def billable_segments_of(cycles)
+        cycles.flat_map do |cycle|
+          Segments.within(cycle, rates:).filter_map do |segment|
+            if segment.rate
+              build_billable_segment(cycle:, segment:)
+            end
+          end
+        end
+      end
+
+      def build_billable_segment(cycle:, segment:)
+        proration_ratio = if terms.prorated
+          cycle.calendar.proration_ratio(segment.started_at, segment.ended_at)
+        else
+          1.0
+        end
+
+        BillableSegment.new(
+          cycle_index: cycle.index,
+          cycle_started_at: cycle.started_at,
+          started_at: segment.started_at,
+          ended_at: segment.ended_at,
+          billing_at: terms.billing_at_for(segment),
+          rate: segment.rate,
+          rate_override: cycle.phase.rate_override,
+          proration_ratio:,
+          rate_phase_code: cycle.phase.code
+        )
+      end
+    end
+  end
+end
--- a/app/services/billing/segments.rb
+++ b/app/services/billing/segments.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Billing
+  module Segments
+    Segment = Data.define(:started_at, :ended_at, :rate)
+
+    # Split the half-open window at each rate change, keeping unpriced periods with rate: nil.
+    def self.within(window, rates:)
+      boundaries = [window.started_at] + rate_changes_inside(window, rates) + [window.ended_at]
+
+      boundaries.each_cons(2).map do |started_at, ended_at|
+        Segment.new(started_at:, ended_at:, rate: rate_at(rates, started_at))
+      end
+    end
+
+    def self.rate_changes_inside(window, rates)
+      rates.map(&:effective_from)
+        .select { |at| at > window.started_at && at < window.ended_at }
+        .uniq
+        .sort
+    end
+
+    def self.rate_at(rates, timestamp)
+      rates.select { |rate| rate.effective_from <= timestamp }.max_by(&:effective_from)
+    end
+
+    private_class_method :rate_changes_inside, :rate_at
+  end
+end
--- a/spec/services/billing/rate_cards/build_schedule_service_spec.rb
+++ b/spec/services/billing/rate_cards/build_schedule_service_spec.rb
@@ -1,0 +1,500 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Billing::RateCards::BuildScheduleService do
+  subject(:result) { described_class.call(contract_rate_card:, plan_rate_card:) }
+
+  let(:organization) { create(:organization) }
+  let(:customer) { create(:customer, organization:, timezone:) }
+  let(:timezone) { "UTC" }
+  let(:catalog_plan) { create(:catalog_plan, organization:) }
+  let(:contract) { create(:contract, organization:, customer:, catalog_plan:, started_at: Time.utc(2026, 1, 1), ended_at:) }
+  let(:rate_card) { create(:rate_card, organization:) }
+  let(:plan_rate_card) { nil }
+  let(:ended_at) { nil }
+
+  let(:contract_rate_card) do
+    create(
+      :contract_rate_card,
+      organization:, contract:, rate_card:,
+      billing_anchor_date: Date.new(2026, 1, 1),
+      effective_date: Date.new(2026, 1, 15),
+      next_billing_at: Time.utc(2026, 2, 1)
+    )
+  end
+
+  before do
+    create(
+      :rate_card_rate,
+      organization:, rate_card:,
+      effective_from: Time.utc(2025, 1, 1),
+      billing_interval_count: 1,
+      billing_interval_unit: "month"
+    )
+  end
+
+  it "builds a schedule when rates are available" do
+    expect(result).to be_success
+    expect(result.schedule).to be_a(Billing::RateCards::Schedule)
+  end
+
+  it "carries the card's anchor, timing and end onto the schedule" do
+    segment = result.schedule.segments_due_by(Time.utc(2026, 3, 1)).first
+
+    expect(result).to be_success
+    expect(segment.started_at...segment.ended_at).to eq(Time.utc(2026, 1, 15)...Time.utc(2026, 2, 1))
+    expect(segment.billing_at).to eq(Time.utc(2026, 2, 1))
+  end
+
+  # The service is what knows a card has been billed before, so the walk it hands back starts at
+  # the last billed cycle rather than replaying the card.
+  context "when the card has already been billed for some cycles" do
+    # The card's own first three cycles: anchored Jan 1 and attached Jan 15, the first is a stub
+    # running to the anchor's next boundary rather than a whole month.
+    before do
+      [[Time.utc(2026, 1, 15), Time.utc(2026, 2, 1)],
+        [Time.utc(2026, 2, 1), Time.utc(2026, 3, 1)],
+        [Time.utc(2026, 3, 1), Time.utc(2026, 4, 1)]].each do |opens_at, closes_at|
+        create(
+          :billing_segment,
+          organization:, contract:, customer:, contract_rate_card:,
+          rate_card_rate: rate_card.rates.sole,
+          cycle_started_at: opens_at, started_at: opens_at, billing_at: closes_at,
+          ended_at: BillingSegment.inclusive_end(closes_at)
+        )
+      end
+    end
+
+    it "resumes at the last billed cycle instead of walking from the card's start" do
+      cycles = result.schedule.segments_due_by(Time.utc(2026, 6, 1))
+
+      expect(cycles.map(&:cycle_index)).to eq([2, 3, 4])
+      expect(cycles.first.started_at).to eq(Time.utc(2026, 3, 1))
+    end
+
+    it "answers the same cycles the full walk would have" do
+      resumed = result.schedule.segments_due_by(Time.utc(2026, 6, 1))
+      contract_rate_card.billing_segments.destroy_all
+      full = described_class.call(contract_rate_card:).schedule.segments_due_by(Time.utc(2026, 6, 1))
+
+      expect(full.size).to be > resumed.size
+      expect(full.map { [it.cycle_index, it.started_at, it.ended_at] })
+        .to end_with(*resumed.map { [it.cycle_index, it.started_at, it.ended_at] })
+    end
+
+    it "does not carry billing history into a separate attachment of the same rate card" do
+      contract_rate_card.update!(ended_date: Date.new(2026, 6, 30))
+      successor = create(:contract_rate_card, organization:, contract:, rate_card:,
+        effective_date: Date.new(2026, 7, 1), billing_anchor_date: Date.new(2026, 1, 1))
+      schedule = described_class.call!(contract_rate_card: successor).schedule
+
+      expect(schedule.segments_due_by(Time.utc(2026, 8, 1)).map(&:cycle_index)).to eq([0])
+    end
+  end
+
+  context "when resuming the first cycle of an attachment" do
+    before do
+      create(:billing_segment, organization:, contract:, customer:, contract_rate_card:,
+        rate_card_rate: rate_card.rates.sole,
+        cycle_started_at: Time.utc(2026, 1, 15), started_at: Time.utc(2026, 1, 15),
+        ended_at: BillingSegment.inclusive_end(Time.utc(2026, 2, 1)), billing_at: Time.utc(2026, 2, 1))
+    end
+
+    it "resumes at midnight of the card's effective date" do
+      expect(result).to be_success
+      expect(result.schedule.segments_due_by(Time.utc(2026, 2, 1)).sole.cycle_started_at)
+        .to eq(Time.utc(2026, 1, 15))
+    end
+  end
+
+  context "with persisted advance segments" do
+    let(:product) { create(:product, :fixed, organization:) }
+    let(:rate_card) { create(:rate_card, organization:, product:, billing_timing: "advance", proration: true) }
+    let(:billed_segment) do
+      create(:billing_segment, organization:, contract:, customer:, contract_rate_card:,
+        rate_card_rate: rate_card.rates.sole, status: :done,
+        cycle_started_at: Time.utc(2026, 1, 15), started_at: Time.utc(2026, 1, 15),
+        ended_at: BillingSegment.inclusive_end(Time.utc(2026, 2, 1)), billing_at: Time.utc(2026, 1, 15))
+    end
+
+    it "resumes from the cycle start when the persisted segment is only its last slice" do
+      applied = create(:plan_rate_card, organization:, catalog_plan:, rate_card:)
+      create(:rate_phase, organization:, plan_rate_card: applied, code: "intro", billing_interval_cycle_count: 1)
+      new_rate = create(:rate_card_rate, organization:, rate_card:, effective_from: Time.utc(2026, 1, 20))
+      create(:billing_segment, organization:, contract:, customer:, contract_rate_card:,
+        rate_card_rate: new_rate, cycle_started_at: Time.utc(2026, 1, 15), started_at: Time.utc(2026, 1, 20),
+        ended_at: BillingSegment.inclusive_end(Time.utc(2026, 2, 1)), billing_at: Time.utc(2026, 1, 20))
+
+      segments = result.schedule.segments_due_by(Time.utc(2026, 2, 1))
+
+      expect(segments.map { |segment| [segment.cycle_index, segment.cycle_started_at, segment.rate_phase_code] })
+        .to eq([[0, Time.utc(2026, 1, 15), "intro"], [0, Time.utc(2026, 1, 15), "intro"], [1, Time.utc(2026, 2, 1), nil]])
+    end
+
+    it "measures consumption against the saved segment after the card terminates" do
+      persisted = billed_segment
+      contract.update!(ended_at: Time.utc(2026, 1, 20, 12))
+      segment = Billing::Segments::Segment.new(
+        started_at: persisted.started_at, ended_at: BillingSegment.exclusive_end(persisted.ended_at),
+        rate: persisted.rate_card_rate
+      )
+      schedule = described_class.call!(contract_rate_card:).schedule
+
+      consumed = [Time.utc(2026, 1, 20), Time.utc(2026, 1, 20, 12), Time.utc(2026, 2, 1)]
+        .map { |at| schedule.consumed_ratio(segment:, at:) }
+
+      # The original segment covers 17 days, even though the card has now ended inside it.
+      expect(consumed).to eq([5.fdiv(17), 6.fdiv(17), 1.0])
+      expect(schedule.next_billing_at(after: contract.ended_at)).to be_nil
+    end
+  end
+
+  # Effective dates are local calendar days, not UTC midnights.
+  context "with a customer timezone" do
+    let(:timezone) { "Europe/Paris" }
+
+    it "opens the first cycle at the start of the signing day there" do
+      cycle = result.schedule.segments_due_by(Time.utc(2026, 3, 1)).first
+
+      expect(cycle.started_at).to eq(Time.utc(2026, 1, 14, 23))
+    end
+  end
+
+  # The card's effective date and inherited anchor must use the same customer-local day.
+  context "when the signing instant falls on a different date where the customer is" do
+    let(:timezone) { "America/New_York" }
+
+    let(:contract_rate_card) do
+      create(
+        :contract_rate_card,
+        organization:, contract:, rate_card:,
+        billing_anchor_date: contract.effective_billing_anchor_date,
+        effective_date: contract.started_at.in_time_zone(timezone).to_date,
+        next_billing_at: Time.utc(2026, 2, 15)
+      )
+    end
+
+    let(:contract) do
+      create(:contract, organization:, customer:, catalog_plan:, started_at: Time.utc(2026, 1, 15, 2))
+    end
+
+    it "derives the anchor in the customer's calendar, not in UTC" do
+      expect(contract.effective_billing_anchor_date).to eq(Date.new(2026, 1, 14))
+    end
+
+    it "opens a single first cycle, not a one-day stub before the anchor" do
+      first = result.schedule.segments_due_by(Time.utc(2026, 4, 1)).first
+
+      expect(first.cycle_index).to eq(0)
+      expect(first.started_at).to eq(Time.utc(2026, 1, 14, 5))
+      expect(first.ended_at).to eq(Time.utc(2026, 2, 14, 5))
+    end
+  end
+
+  # The hint is a shortcut, not the only route. The termination and credit paths build a
+  # schedule from the card alone, and passing nil straight through made the resolver fall
+  # back to `nil&.rate_phases.to_a` — so every phase and every override the plan configured
+  # disappeared, with no error and no clue on the invoice.
+  describe "resolving the plan entry when the caller has no hint" do
+    let(:plan_rate_card) { create(:plan_rate_card, organization:, catalog_plan:, rate_card:, units: 5) }
+    let(:rate_override) { create(:rate_override, organization:, rate_properties: {"amount" => "1.00"}) }
+
+    before do
+      plan_rate_card
+      create(:rate_phase, organization:, plan_rate_card:, code: "intro",
+        billing_interval_cycle_count: 1, position: 1, rate_override_id: rate_override.id)
+    end
+
+    it "finds the plan entry itself when none is handed in" do
+      schedule = described_class.call(contract_rate_card:).schedule
+      first = schedule.segments_due_by(Time.utc(2026, 4, 1)).first
+
+      expect(first.rate_phase_code).to eq("intro")
+      expect(first.rate_override).to eq(rate_override)
+    end
+
+    it "answers the same whether the hint is handed in or looked up" do
+      with_hint = described_class.call(contract_rate_card:, plan_rate_card:).schedule
+      without = described_class.call(contract_rate_card:).schedule
+      codes = ->(s) { s.segments_due_by(Time.utc(2026, 4, 1)).map { |cycle| cycle.rate_phase_code } }
+
+      expect(codes.call(without)).to eq(codes.call(with_hint))
+    end
+
+    # A hint for a different card is a caller bug, and silently ignoring it prices the card
+    # at the base rate — the same invisible failure, arrived at from the other side.
+    it "refuses a hint that prices a different rate card" do
+      other = create(:plan_rate_card, organization:, catalog_plan:, rate_card: create(:rate_card, organization:))
+
+      expect { described_class.call(contract_rate_card:, plan_rate_card: other) }
+        .to raise_error(described_class::MismatchedPlanRateCard, /prices rate card/)
+    end
+  end
+
+  # The window is a fact about SERVICE; the price is a decision about BILLING. The schedule
+  # closes where service stopped, and the termination day is still paid in full — but by the
+  # day-ownership rule, not by stretching the boundary. Stretching it would record a window
+  # service never covered, and usage metering reads these boundaries to pick its events.
+  context "when the contract has ended" do
+    # Prorating, so the day count is visible in the ratio rather than flattened to 1.0. A fixed
+    # product, because proration on a usage product needs a recurring metric behind it.
+    let(:rate_card) do
+      create(:rate_card, organization:, proration: true, product: create(:product, :fixed, organization:))
+    end
+    let(:ended_at) { Time.utc(2026, 1, 20, 14, 30) }
+
+    def final_slice = result.schedule.segments_due_by(Time.utc(2026, 6, 1)).last
+
+    # The card opened Jan 15 and closed part-way through the 20th: six days begun, out of the
+    # 31 in the January interval the slice sits in.
+    it "closes where service stopped, and still bills that day whole" do
+      expect(final_slice.ended_at).to eq(ended_at)
+      expect(final_slice.proration_ratio).to eq(6.fdiv(31))
+    end
+
+    # The one instant where the two readings differ. Service ran for none of the 20th, so the
+    # 20th is not billed — 19 days, not 20. Rounding the boundary up to the 21st would charge
+    # a day the customer never entered.
+    context "when it ended on the first instant of a day" do
+      let(:ended_at) { Time.utc(2026, 1, 20) }
+
+      it "bills nothing for the day it ended on" do
+        expect(final_slice.ended_at).to eq(ended_at)
+        expect(final_slice.proration_ratio).to eq(5.fdiv(31))
+      end
+    end
+
+    # 2026-01-21 02:00 UTC is still the 20th in New York. The boundary is the raw instant
+    # either way; what the timezone decides is the day COUNT, and reading it in UTC would
+    # bill a day the customer never entered.
+    context "when the customer is west of UTC" do
+      let(:timezone) { "America/New_York" }
+      let(:ended_at) { Time.utc(2026, 1, 21, 2) }
+
+      it "counts the days where the customer is" do
+        expect(final_slice.ended_at).to eq(ended_at)
+        expect(final_slice.proration_ratio).to eq(6.fdiv(31))
+      end
+    end
+  end
+
+  context "when the card has an inclusive end date" do
+    let(:rate_card) { create(:rate_card, organization:, proration: true, product: create(:product, :fixed, organization:)) }
+
+    before { contract_rate_card.update!(ended_date: Date.new(2026, 1, 20)) }
+
+    it "covers the whole final day and stops at the next local midnight" do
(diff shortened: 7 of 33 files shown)


app/services/billing/schedules/build_service.rb
```diff
@@ -0,0 +1,90 @@
...
+            interval: Billing::Interval.from(base_rate, override: phase.rate_override),
+            code: phase.code,
+            override: phase.rate_override
+          )
+        end
+
+        return configured if configured.last && configured.last.cycle_count.nil?
+
+        configured + [tail_phase]
+      end
```
tiagolupepic: `Billing::Schedule::Phase` should be a contract (PORO) class from `ResolveRatePhasesService` instead, this class should not know how to decorate the object. Or as alternative, the Phase Object should belongs to this class.

app/services/billing/schedules/build_service.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      def tail_phase
+        Billing::Schedule::Phase.new(cycle_count: nil, interval: Billing::Interval.from(base_rate), code: nil, override: nil)
```
tiagolupepic: you can have this as `Billing::Schedule::Phase.default_with_rate(base_rate)`

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    # The boundary at position `index`, re-derived from the anchor so month-ends don't drift.
+    # @example anchored on Jan 31: at(1) is Feb 28, at(2) is Mar 31
+    def at(index)
+      interval.advance(anchor, index)
+    end
+
+    # @example a 30-day period: days_in(Jun 1...Jul 1) is 30
+    def days_in(period)
+      (period.end.to_date - period.begin.to_date).to_i
```
tiagolupepic: `period.days_in => integer` the method belongs to period instead. And you can avoid the `days_in` method.

app/services/billing/schedule.rb
```diff
@@ -0,0 +1,134 @@
...
+    # until the schedule ends. `interval` is already resolved: a phase overriding the card's
+    # cadence carries the override, one that doesn't carries the card's own. `code` and
+    # `override` are carried for the caller: a cycle knows which phase priced it.
+    Phase = Data.define(:cycle_count, :interval, :code, :override)
+
+    # One billing cycle. `due_at` is the instant it becomes billable, which is the only
+    # thing billing timing changes: the start of the cycle in advance, its end in arrears.
+    # `calendar` is the ruler it was measured on, so a caller can ask what share of a whole
+    # period any window inside it covers.
+    Cycle = Data.define(:index, :from, :to, :due_at, :phase, :calendar)
```
tiagolupepic: can we name the dates, `starts_on` and `ends_on` instead?

app/services/billing/schedule.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    # Walks the phases in order, collecting cycles until one falls due after `instant` or
+    # the schedule ends. Returns the collected cycles and the first one left out.
+    def walk(instant)
+      cursor = starts_at
+      index = 0
+      due = []
+      pending = nil
+
+      phases.each_with_index do |phase, position|
```
tiagolupepic: we could memoize this. it will save a lot of computation.

app/services/billing/segments.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Billing
+  # A cycle is billed as one fee unless a rate changes inside it. Each change part-way
+  # through opens a segment: a separate fee over a sub-window at its own price, without
+  # opening a new cycle — the segments of one cycle share its index.
+  module Segments
+    Segment = Data.define(:from, :to, :rate)
```
tiagolupepic: ditto. change the names of dates as `starts_on` and `ends_on`

app/services/billing/segments.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    def self.changes_inside(period, rates)
+      rates.map(&:effective_from)
+        .select { |at| at > period.begin && period.cover?(at) }
+        .uniq
+        .sort
+    end
+
+    def self.rate_at(rates, instant)
+      rates.select { |rate| rate.effective_from <= instant }.max_by(&:effective_from)
```
tiagolupepic: use bsearch here. Having a bit amount of rates, it will perform very badly.

app/services/billing/segments.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      starts.each_with_index.filter_map do |from, index|
+        rate = rate_at(rates, from)
+        next unless rate
+
+        Segment.new(from:, to: starts[index + 1] || period.end, rate:)
+      end
+    end
+
+    def self.changes_inside(period, rates)
```
tiagolupepic: both methods `changes_inside` and `rate_at` should be private, no?

db/migrate/20260901122946_create_billing_segments.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+class CreateBillingSegments < ActiveRecord::Migration[8.0]
+  def change
+    enable_extension "btree_gist"
+
+    create_enum :billing_segment_status, %w[pending processing done failed]
+
+    create_table :billing_segments, id: :uuid do |t|
```
annvelents, line 9: do we want to add an id of the cycle this segment belongs to?

app/services/billing/segments.rb
```diff
@@ -0,0 +1,46 @@
...
+      rates.map(&:effective_from)
```
tiagolupepic, line 19: same here, you could use sort + bsearch here and then slice the array after getting the first index occurrence

app/services/subscriptions/product_catalog/materialize_service.rb
```diff
@@ -41,6 +28,63 @@ def call
...
+        period = calendar_for(item, rate).period_at(first_billable_at)
+        opens_at = [period.begin, billing_starts_at].max
+
+        item.rate_card.advance? ? opens_at : period.end
+      end
+
+      def calendar_for(item, rate)
+        Billing::Calendar.new(
+          anchor_date: item.billing_anchor_date,
+          interval: Billing::Interval.new(count: rate.billing_interval_count, unit: rate.billing_interval_unit),
```
tiagolupepic: rate could expose as well the Billing::Interval object

rate.billing_interval => `Billing::Interval.new(count: self.billing_interval_count, unit: self.billing_interval_unit)`

app/services/subscriptions/product_catalog/materialize_service.rb
```diff
@@ -41,6 +28,63 @@ def call
...
+        return started_at unless rate
+
+        period = calendar_for(item, rate).period_at(first_billable_at)
+        opens_at = [period.begin, billing_starts_at].max
+
+        item.rate_card.advance? ? opens_at : period.end
+      end
+
+      def calendar_for(item, rate)
+        Billing::Calendar.new(
```
tiagolupepic: should Billing::Calendar have a method `initial_next_billing_at` that wraps the entire logic from here?

app/services/subscriptions/product_catalog/materialize_service.rb
```diff
@@ -41,6 +28,63 @@ def call
...
+      #
+      # Rate phases are NOT considered yet: a first phase carrying an interval override
+      # (QA plan PH3) would open on a different cadence than the one seeded here. Without a
+      # resolvable rate there is no interval at all, so fall back to started_at and let a
+      # later scheduler pass move the clock on.
+      def initial_next_billing_at(item)
+        rate = item.rate_card&.rate_active_at(started_at)
+        return started_at unless rate
+
+        period = calendar_for(item, rate).period_at(first_billable_at)
```
annvelents: let's have `segment` here, because technically these periods are segment

app/services/billing/interval.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+module Billing
+  # How long one billing cycle lasts: `count` calendar units of `unit`.
+  class Interval < Data.define(:count, :unit)
```
annvelents, line 8: add a comment:
` # for example Data(count: 2, unit: week) means interval `2 weeks``

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,70 @@
...
+    # this ruler, not a billing cycle number.
+    # @example anchored on Jul 15: index_at(Jul 2) is -1, the period [Jun 15, Jul 15)
+    def index_at(time)
+      interval.steps_between(anchor, time.in_time_zone(timezone))
+    end
+
+    # The half-open period `time` falls in, as [boundary, next boundary).
+    # @example anchored on Feb 1: period_at(Feb 10) is 2022-02-01...2022-03-01
+    def period_at(time)
+      index = index_at(time)
```
annvelents: cycle_index

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      at(index)...at(index + 1)
+    end
+
+    # Days in the whole period `time` falls in, whether or not the window covers it all.
+    # @example anchored on Jun 1 monthly: days_in_period_at(Jun 10) is 30
+    def days_in_period_at(time)
```
annvelents: days_in_period_covering(time)

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,70 @@
...
+    # The index of the period `time` falls in, negative before the anchor. A position on
+    # this ruler, not a billing cycle number.
+    # @example anchored on Jul 15: index_at(Jul 2) is -1, the period [Jun 15, Jul 15)
+    def index_at(time)
+      interval.steps_between(anchor, time.in_time_zone(timezone))
+    end
+
+    # The half-open period `time` falls in, as [boundary, next boundary).
+    # @example anchored on Feb 1: period_at(Feb 10) is 2022-02-01...2022-03-01
+    def period_at(time)
```
annvelents: period_covering?
(don't like "covering")

annvelents: period_enclosing

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,70 @@
...
+    # Days in the whole period `time` falls in, whether or not the window covers it all.
+    # @example anchored on Jun 1 monthly: days_in_period_at(Jun 10) is 30
+    def days_in_period_at(time)
+      days_in(period_at(time))
+    end
+
+    # Fraction of its period the half-open window [from, to) covers. The window must sit
+    # inside one period: measuring across a boundary would divide by the wrong length.
+    # @example in a 30-day period: [Jun 1, Jul 1) is 1.0, [Jun 28, Jul 1) is 0.1
+    def elapsed_ratio(from, to)
```
annvelents: proration?

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,70 @@
...
+      covered_days(from, to).fdiv(days_in(period))
+    end
+
+    private
+
+    attr_reader :anchor_date, :interval, :timezone
+
+    # The boundary at position `index`, re-derived from the anchor so month-ends don't drift.
+    # @example anchored on Jan 31: at(1) is Feb 28, at(2) is Mar 31
+    def at(index)
```
annvelents: interval_at(index)

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,70 @@
...
+    def period_at(time)
+      index = index_at(time)
+
+      at(index)...at(index + 1)
+    end
+
+    # Days in the whole period `time` falls in, whether or not the window covers it all.
+    # @example anchored on Jun 1 monthly: days_in_period_at(Jun 10) is 30
+    def days_in_period_at(time)
+      days_in(period_at(time))
```
annvelents: period_at(time).days

app/services/billing/schedule.rb
```diff
@@ -0,0 +1,134 @@
...
+      @ends_at = ends_at
+      @timing = timing.to_sym
+
+      validate!
+    end
+
+    # Every cycle billable by `instant` — the ones whose due_at has arrived. A cycle
+    # running over that instant but closing after it is not due yet, so it is left out.
+    # @example starting Jan 15, monthly on the 1st: [Jan 15, Feb 1), [Feb 1, Mar 1), ...
+    def cycles_due_by(instant)
```
annvelents: timestamp?

annvelents: time? to keep consistency with calendar

app/services/billing/schedule.rb
annvelents: are phases sorted?

app/services/billing/schedule.rb
```diff
@@ -0,0 +1,134 @@
...
+      Calendar.new(anchor_date: anchor, interval: phase.interval, timezone:)
+    end
+
+    # Clamped by where the walk stands and by the card's end — never by the instant the
+    # caller asked about. A cycle in force at that instant still runs to its own boundary.
+    def cycle_for(period, cursor, index, phase, calendar)
+      from = [period.begin, cursor].max
+      to = ends_at ? [period.end, ends_at].min : period.end
+
+      Cycle.new(index:, from:, to:, due_at: (timing == :advance) ? from : to, phase:, calendar:)
```
annvelents: It seems that we have PhaseCycles and Billing Cycles, and now it's not clear which one is meant 🙈

app/services/billing/schedule.rb
annvelents: phase_cycles_due_by

app/services/billing/schedule.rb
annvelents: do we need calendar for Cycle?

app/services/billing/segments.rb
```diff
@@ -0,0 +1,46 @@
...
+    # opens no segment. One landing on the end belongs to the next period, for the same
+    # reason a boundary belongs to the period it opens.
+    #
+    # Where no rate is in force yet the window is dropped rather than billed at nothing:
+    # billing starts at the first effective date and runs to the end of the period.
+    #
+    # @example [Sep 10, Oct 10) with a rate effective Sep 25
+    #   [Sep 10, Sep 25) at the old rate, [Sep 25, Oct 10) at the new one
+    def self.within(period, rates:)
+      starts = [period.begin] + changes_inside(period, rates)
```
annvelents: starts => segments_start_dates
changes_inside => dates_of_rate_changes_within_the_period => ...?

app/services/subscriptions/product_catalog/materialize_service.rb
```diff
@@ -41,6 +28,63 @@ def call
...
+        opens_at = [period.begin, billing_starts_at].max
+
+        item.rate_card.advance? ? opens_at : period.end
+      end
+
+      def calendar_for(item, rate)
```
brunomiguelpinto: **Two places now translate a card into date machinery, and they disagree.**

`BuildService` does the same job one directory over: anchor, timezone, floor the start to the customer's day. But it reads `card_started_at` and resolves rate phases, while this builds from `started_at` and ignores them.

Today they agree — at materialization there are no prior versions and no subscription-level phases — so this isn't a live bug. But it is the same rule written twice, and that is exactly the shape that produced the two money bugs we found in the engine branch (a missing `pricing_unit`, and a credit of 903 that should have been 452).

Either route this through `BuildService` (it already accepts `plan_rate_cards`, which is where a new subscription's phases live), or say in a comment why the seed deliberately does not.

app/services/subscriptions/product_catalog/materialize_service.rb
```diff
@@ -2,11 +2,6 @@
 
 module Subscriptions
   module ProductCatalog
-    # Materializes the plan's rate cards onto the subscription: one
-    # subscription_rate_card per plan_rate_card, carrying the billing
-    # lifecycle (anchor, clock, units). Pricing is not copied — a plan is
-    # immutable once it has subscriptions, so phases and rates resolve by
-    # reference through the plan entry.
     class MaterializeService < BaseService
```
brunomiguelpinto, line 9: The class comment was dropped in this diff — the one explaining that pricing is not copied because a plan is immutable once it has subscriptions, so phases and rates resolve by reference.

This PR makes the class harder to read, not easier: five of its private methods are now date logic. It needs that paragraph more than it did before, not less.

app/models/billing_segment.rb
```diff
@@ -0,0 +1,109 @@
...
+    done: "done",
+    failed: "failed"
+  }.freeze
+
+  belongs_to :organization
+  belongs_to :subscription
+  belongs_to :customer, -> { with_discarded }
+  belongs_to :subscription_rate_card, -> { with_discarded }
+  belongs_to :invoice, optional: true
+  belongs_to :rate_card_rate, -> { with_discarded }, optional: true
```
brunomiguelpinto: `rate_card_rate` is `optional: true` here and nullable in the migration, but every method below treats it as guaranteed — `billing_interval`, `currency`, `pricing_unit_conversion_rate` and `min_amount_cents` all call straight through it and would raise `NoMethodError` on a row without one.

The schema and the behaviour disagree. Either the column should be NOT NULL (is a segment without a rate ever a real row?) or the readers should handle the absence. The pattern predates this PR, but `billing_interval` is new and adds to it.

app/models/billing_segment.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+
+# A durable billing-period row: the outbox + retry unit of the new billing engine.
+# One per (subscription_rate_card, period). The scheduler inserts it (status
+# pending) while advancing the clock; the processor turns pending segments into an
+# invoice and marks them done. High-volume ledger — no soft delete, no PaperTrail.
+class BillingSegment < ApplicationRecord
+  STATUSES = {
```
brunomiguelpinto, line 8: Two of the four statuses are unreachable and `attempts` is never incremented — nothing in the codebase writes `processing` or `failed`.

Not a blocker for this PR (nothing writes the table yet at all), but worth knowing before someone relies on them: a segment that keeps failing stays `pending` and is retried forever, with no counter and no way to see it. Whoever wires the processor should either implement the states or drop them.

brunomiguelpinto (commented):
Read it end to end. The pure layer is the strongest part — `Interval`, `Calendar` and `Segments` have no dependencies, and having the QA scenarios pinned as value tables means an expectation can be checked against the plan rather than argued about. Three things I'd want addressed before this lands, one of them structural.

app/services/billing/schedule.rb
brunomiguelpinto: Good catch — no, and `Schedule` depends on the order in three places: the `each_with_index` walk, the `position.zero?` that decides which phase inherits the card anchor, and both validations (`phases.last`, `phases[..-2]`).

It happens to be correct today because the only builder sorts (`SubscriptionRateCards::ResolveRatePhasesService:30` does `phases.sort_by(&:position)`), but `Schedule` is public and an unsorted list bills the wrong cadences in the wrong windows without failing.

It also cannot sort defensively: `Phase` carries no `position`, the array order *is* the ordering. So I wrote the precondition down on `initialize` instead of pretending it is inferred.

app/models/billing_segment.rb
```diff
@@ -0,0 +1,109 @@
...
+  belongs_to :invoice, optional: true
+  belongs_to :rate_card_rate, -> { with_discarded }, optional: true
+  belongs_to :rate_override, -> { with_discarded }, optional: true
+  belongs_to :pricing_unit, optional: true
+
+  enum :status, STATUSES, validate: true
+
+  validates :billing_at, presence: true
+  validates :period_from, presence: true
+  validates :period_to, presence: true
```
tiagolupepic: can we rename the dates as well? 
`starts_on` and `ends_on`
`start_date` and `end_date`.

I don't have an strong option, just to make the names of dates consistent.

app/models/billing_segment.rb
```diff
@@ -0,0 +1,109 @@
...
+  validates :proration_ratio, numericality: {greater_than_or_equal_to: 0, less_than_or_equal_to: 1}
+
+  def rate
+    rate_override || rate_card_rate
```
tiagolupepic, line 33: we need one more extra validation, the billing_cycle can't be created if the rate_card_rate and rate_override are nil.

app/services/billing/interval.rb
brunomiguelpinto: Done, with the real constructor form since `Data(...)` is not how it is built:

```ruby
# How long one billing cycle lasts: `count` calendar units of `unit`.
# @example Interval.new(count: 2, unit: :week) is "every two weeks"
# @example Interval.new(count: 1, unit: :month) is "monthly"
```

app/services/subscriptions/product_catalog/materialize_service.rb
brunomiguelpinto: Two answers here, and the second is why it stays as it is.

The immediate one: `Billing::Interval.from(rate)` already exists and does exactly this, and `MaterializeService` was not using it. That whole `Calendar`/`Interval` construction is gone now — the service goes through `Billing::BuildScheduleService` like every other caller, so it no longer builds date machinery by hand at all.

The second: `rate.billing_interval` on the model cannot replace `Interval.from(rate, override:)`, because the override's two columns are independently nullable. An override that sets only `billing_interval_unit` has to take its `count` from the rate, so there is no whole interval on either side to prefer — it has to read the four columns. I have written that reason onto `Interval.from` so the next reader does not have to rediscover it.

app/services/billing/calendar.rb
brunomiguelpinto: Agreed, and it went further than moving: `days_in` is gone and so is `days_in_period_at`, because neither had a single caller — not in this PR and not in the branch that follows it. The engine prorates by ratio (`proration_ratio`), never by `amount / days × days_billed`, so the day count never leaves `Calendar`.

Its coverage did not go with it: the 28 / 29 / 30 / 31 / 90 / 14-day table moved under `#proration_ratio` and asserts that one day of an N-day period is `1/N`, which pins the same arithmetic through the surface that survives.

app/services/billing/calendar.rb
brunomiguelpinto: This one I pushed back on, because `cycle_index` would be actively wrong — and then took half of it.

It is not a cycle number. Anchored on Jan 1 with the card starting Jan 15, the period containing Jan 20 has index 0 while the card's cycle 0 is `[Jan 15, Feb 1)`; and asking about Dec 20 returns `-1`, which no cycle ever is. A cycle is clamped by the card's life and counted by its phase, neither of which this ruler knows about.

But you bounced off the name, which is evidence that `index_at` does not say *what* index. It is now **`period_index`** — the index of the period, paired with `period_containing` below.

app/services/billing/calendar.rb
brunomiguelpinto: Renamed, and I went with plain **`period_containing`** over `covering`/`enclosing`.

Your instinct was right about the reason, I think: `period_at` reads as "the period that starts at this instant", which is not what it returns. `containing` cannot be misread. Dropped the `?` because it returns a Range, not a boolean.

brunomiguelpinto: Same answer as the sibling comment: it is **`period_containing`** now. `enclosing` and `containing` say the same thing and `containing` is the plainer word.

app/services/billing/calendar.rb
brunomiguelpinto: Half yes. It is **`proration_ratio`** now — the name of the column it feeds (`billing_segments.proration_ratio`), so the number has one name from here to the database.

Not `proration?` though: it returns a `Float` (`0.4666…`), and a `?` promises a boolean. That is the sort of name that gets someone writing `if calendar.proration?`, which would be true for every non-zero window.

app/services/billing/calendar.rb
brunomiguelpinto: Countered rather than applied: it is **`boundary_at`**, not `interval_at`.

It does not return an interval — it returns the instant a boundary falls on, which is why its own doc said "The boundary at position `index`". An `Interval` in this layer is a *length* ("1 month"), with no position at all. `boundary_at` also matches the vocabulary the rest of the file already uses ("Boundary 0", "[boundary, next boundary)").

app/services/billing/calendar.rb
brunomiguelpinto: The method is gone — see the reply on `days_in` above. It had no caller anywhere, and the naming question went with it.

app/services/billing/calendar.rb
brunomiguelpinto: This is the same idea as Tiago's on `days_in`, and the answer is the same: rather than move the day count onto the period, it is deleted. Nothing read it. Introducing a `Period` value object to host `#days` would have been a wrapper built for a method with no consumers.

app/services/billing/segments.rb
brunomiguelpinto: Declining this one, and the same on `Schedule`. `_on` means `Date` in Rails, and these are `Time`s with an hour and a timezone — a segment can start at `Sep 25 14:00`.

But the deeper reason is the boundary rather than the type, and I have written it into `Schedule`'s doc since you raised it. This layer uses two schemes on purpose:

| | means | the column it becomes |
|---|---|---|
| `from` / `to` | a **window being billed**, half-open — `to` is where it stops covering, not the last moment it covers | `billing_segments.period_from` / `period_to`, the end written a microsecond earlier because that pair is inclusive (`tsrange(..., '[]')`) |
| `starts_at` / `ends_at` | a **record's own life**, nullable | `subscription_rate_cards.started_at` / `ended_at` |

`ends_at` on a half-open window would be the more dangerous lie of the two: a wrong type gets caught the first time someone passes a `Date`, a wrong boundary silently bills a day extra. That is exactly the bug this PR fixes — see the proration section in the description, where v1 charges 16 + 15 days of a 30-day month.

`Segment` now says so on itself.

app/services/billing/segments.rb
brunomiguelpinto: Measured before answering, because you may well be right at some scale — but the numbers say not here.

`rate_at` costs **1.7 µs** at 5 rates and **14 µs** at 60 (100k calls, in the container). It is called once per segment start, so one or two times per card per clock run: ~28 µs per card at 60 rates, against a run that does a DB insert per segment.

Two problems with the change itself, beyond the cost:

1. **bsearch needs sorted input**, and the contract here is explicitly that order does not matter — callers pass `rate_card.rates` however they have it. Adding an unchecked precondition is the kind of implicit contract we have just spent this PR making explicit elsewhere.
2. **If this ever does matter, it is not the line to fix.** `rate_changes_inside` runs the same number of times and does `map` + `select` + `uniq` + `sort` — O(n log n) with three arrays allocated. It is strictly more expensive than the linear scan you flagged.

Happy to revisit with a profile from a real run.

app/services/billing/segments.rb
brunomiguelpinto: Same measurement as on `rate_at`, and one detail specific to here: this method already **ends** with `.sort`. Sorting first so we can bsearch and slice is the same O(n log n) with more moving parts.

If profiling ever puts this on the critical path, the win is not bsearch — it is that `within` calls `rate_at` once per segment start after this has already walked the same list, so one pass could produce both. That is a rewrite worth doing with a profile in hand, not on suspicion.

app/services/billing/segments.rb
brunomiguelpinto: Both taken, with the types corrected.

`starts` is now **`segment_starts`**. Not `_dates` though — they are `Time`s, and an advance rate takes effect at an arbitrary instant, so `Sep 25 14:00` is a real value here.

`changes_inside` is now **`rate_changes_inside`**. Your instinct was right that it does not say *what* changes; `dates_of_rate_changes_within_the_period` says it three times, and the parameter is already called `window`.

Speaking of which, I renamed that parameter from `period` to `window` while I was here — `period` is a fifth noun that is not in the vocabulary, and what this actually receives is a **cycle**, which may have been clamped by the card's start or by a termination and so is not a whole calendar period.

app/services/billing/schedules/build_service.rb
brunomiguelpinto: Took the spirit of this, not the move — and the reason is one argument.

The translation needs **`base_rate`** as well as the phase: a phase overriding the cadence carries only the override, so the interval is `Interval.from(base_rate, override: phase.rate_override)`. `ResolveRatePhasesService` resolves *which phases apply* and has no rate to fold in; giving it one would hand it a second responsibility and an object it has no other use for.

The alternative — `Phase` belonging to this class — inverts the dependency the wrong way: `Schedule` takes `Phase` as its input and `Cycle` carries one, so the pure layer would end up depending on the ActiveRecord-touching service.

What was fair in the comment is that `#phases` was doing two things at once. The mapping is now its own named method, so the list-building reads on its own:

```ruby
def phases
  configured = rate_phases.map { |rate_phase| phase_for(rate_phase) }
  return configured if configured.last && configured.last.cycle_count.nil?
  configured + [default_phase]
end
```

and `#phase_for` carries the reason it lives here.

Worth saying plainly: translating ActiveRecord into values **is** this class's single responsibility — it is the only file in `Billing` that touches AR, which is what lets everything behind it be tested as tables of values.

app/services/billing/schedules/build_service.rb
brunomiguelpinto: Half taken. The name was the good half and it is in: `tail_phase` is now **`default_phase`**, because "tail" is wrong when it is the *only* phase — which happens whenever nothing is configured.

Not moving the factory onto `Phase` though. It would take an ActiveRecord rate to build `Interval.from(rate)`, pushing schema knowledge one level deeper into the pure layer for a single caller, and the method is one line either way. `Interval.from` already crosses that line, but it is forced to: the override's two columns are independently nullable, so there is no whole Interval on either side to compose. Here there is no such constraint.

The doc now says what it is and why it exists at all — a configured phase list may be bounded all the way through (`[intro(6), middle(3)]` is valid), and this is what follows it.

app/services/subscriptions/product_catalog/materialize_service.rb
brunomiguelpinto: Right that it did not belong here, wrong home though — it is **`Billing::Schedule#next_billing_at`** now, and this service asks for it:

```ruby
def initial_next_billing_at(subscription_rate_card, plan_rate_card)
  build = Billing::BuildScheduleService.call(subscription_rate_card:, plan_rate_card:)
  return started_at unless build.success?

  build.schedule.next_billing_at(Time.current)
end
```

`Calendar` would have been the wrong level: the rule is *"the first cycle not yet billed"*, which needs the phases, the timing and the card's end — a `Calendar` is one ruler and knows none of those. `Schedule` has all three, and the name matches the column it fills.

The whole hand-built `Calendar` and `Interval` are gone with it: the service went from 93 lines to 79 and now depends on one thing in `Billing` instead of two. And it fixed a bug the old code admitted to in a comment — *"Rate phases are NOT considered yet: a first phase carrying an interval override (QA plan PH3) would open on a different cadence than the one seeded here"*. Going through the schedule considers them, and there is a spec for it.

app/services/subscriptions/product_catalog/materialize_service.rb
brunomiguelpinto: Agreed, and it resolved itself: the local was called `period` because this service built its own `Calendar` and asked it for one. That whole block is gone — the service now goes through `Billing::BuildScheduleService` like every other caller, so there is no window variable here at all.

The word did get settled elsewhere while we were at it: `Billing::Segments.within` took a `period` argument and now takes a `window`, since what it actually receives is a **cycle** — possibly clamped by the card's start or by a termination, so not a whole calendar period.

app/services/subscriptions/product_catalog/materialize_service.rb
brunomiguelpinto: Fixed. The service goes through `Billing::BuildScheduleService` now, so there is one translation instead of two:

```ruby
- def calendar_for(item, rate)          # Calendar + Interval by hand
- def first_billable_at                 # [started_at, Time.current].max
- def billing_starts_at                 # beginning_of_day in the customer timezone
- def timezone                          # subscription.customer.applicable_timezone
- rate = item.rate_card&.rate_active_at(started_at)
- period = ...period_at(...); opens_at = [period.begin, billing_starts_at].max
- item.rate_card.advance? ? opens_at : period.end
+ build.schedule.next_billing_at(Time.current)
```

None of that was rewritten elsewhere — it stopped being necessary, because `Schedule` already clamps the first cycle to the card's start.

The disagreement I was worried about was real and worth more than the duplication: the old code read `started_at` and ignored phases, so a first phase with an interval override seeded the clock a whole cadence late (QA plan PH3). Its own comment admitted it. There is a spec now — a weekly intro phase gives `next_billing_at = Jan 22` where the old path gave `Feb 1`.

app/services/subscriptions/product_catalog/materialize_service.rb
brunomiguelpinto: Restored, and it earns its place more than before: five of the private methods were date logic when I wrote that, and now there are three methods left in the whole class.

```ruby
# Materializes the plan's rate cards onto the subscription: one subscription_rate_card
# per plan_rate_card, carrying the billing lifecycle (anchor, clock, units). Pricing is
# not copied — a plan is immutable once it has subscriptions, so phases and rates
# resolve by reference through the plan entry.
```

The spec that asserts it — *"does not copy the plan entry's phases: pricing resolves by reference"* — is still there, so the paragraph and the test say the same thing.

app/services/billing/segments.rb
brunomiguelpinto: yup already done

app/services/billing/segments.rb
brunomiguelpinto: yup going to check and change if its better

db/migrate/20260901122946_create_billing_segments.rb
```diff
@@ -0,0 +1,90 @@
...
+      t.references :pricing_unit, foreign_key: true, type: :uuid
+      t.jsonb :rate_properties, null: false, default: {}
+
+      # The instant this segment becomes billable — the boundary the scheduler picked it
+      # up on. Segments sharing a subscription and a billing_at are invoiced together.
+      t.datetime :billing_at, null: false
+
+      # The window this segment is billed for. Half-open in the engine, stored inclusive:
+      # period_to is the last instant covered, a microsecond before the window closes.
+      t.datetime :period_from, null: false
```
annvelents: just `from` / `to` ?

db/migrate/20260901122946_create_billing_segments.rb
```diff
@@ -0,0 +1,90 @@
...
+      # insert, never a double bill.
+      t.index [:subscription_rate_card_id, :period_from],
+        unique: true,
+        name: "index_billing_segments_on_product_and_period"
+
+      # Processor: find a subscription's pending segments for a billing moment.
+      t.index [:subscription_id, :billing_at, :status]
+
+      # Group a card's segments by the cycle they belong to, index-only.
+      t.index [:subscription_rate_card_id, :cycle_from]
```
annvelents: add also `t.index  :cycle_from` for search within the same cycle for different rate cards

app/models/billing_segment.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+
+# A durable billing-period row: the outbox + retry unit of the new billing engine.
+# One per (subscription_rate_card, period). The scheduler inserts it (status
+# pending) while advancing the clock; the processor turns pending segments into an
+# invoice and marks them done. High-volume ledger — no soft delete, no PaperTrail.
+class BillingSegment < ApplicationRecord
+  STATUSES = {
+    pending: "pending",
+    processing: "processing",
```
annvelents, line 10: let's keep it, as now it's useful for generating invoices to quickly find them

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+module Billing
+  # A ruler: boundaries every `interval` from the start of the anchor day, in the customer's timezone
+  class Calendar
+    def initialize(anchor_date:, interval:, timezone:)
+      @anchor_date = anchor_date
```
annvelents: `@anchor = anchor_date.in_time_zone(timezone).beginning_of_day`

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,89 @@
...
+      @interval = interval
+      @timezone = timezone
+    end
+
+    # Boundary 0: the start of the anchor day in the customer's timezone.
+    #
+    # @example anchored Feb 1 in "America/New_York"
+    #   anchor  # => 2022-02-01 05:00 UTC
+    def anchor
```
annvelents: don't need this :)

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,89 @@
...
+      @anchor ||= anchor_date.in_time_zone(timezone).beginning_of_day
+    end
+
+    # Which period `timestamp` falls in — a position on this ruler, not a cycle number.
+    #
+    # @example anchored Jul 15, monthly
+    #   period_index(Jul  2)  # => -1   the anchor is a reference day, not a start date
+    #   period_index(Jul 15)  # =>  0
+    #   period_index(Aug 20)  # =>  1
+    def period_index(timestamp)
```
annvelents: interval_index?

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,89 @@
...
+      interval.steps_between(anchor, timestamp.in_time_zone(timezone))
+    end
+
+    # The half-open period `timestamp` falls in, as [boundary, next boundary).
+    #
+    # @example anchored Feb 1, monthly
+    #   period_containing(Feb 10)  # => 2022-02-01 00:00...2022-03-01 00:00
+    def period_containing(timestamp)
```
annvelents: let's change periods to intervals to keep consistency of naming

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      boundary_at(index)...boundary_at(index + 1)
+    end
+
+    # Fraction of its period the half-open window [from, to) covers.
+    #
+    # @example anchored Jun 1, monthly — a 30-day period
+    #   proration_ratio(Jun  1, Jul 1)  # => 1.0
+    #   proration_ratio(Jun 28, Jul 1)  # => 0.1
+    def proration_ratio(from, to)
```
annvelents, line 29: just to think if we want to keep it in the calendar or in the segment, but calendar looks more correct, so most likely - discard this comment

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,89 @@
...
+    private
+
+    attr_reader :anchor_date, :interval, :timezone
+
+    # The boundary at position `index`
+    #
+    # @example anchored Jan 31, monthly
+    #   boundary_at(1)  # => Feb 28
+    #   boundary_at(2)  # => Mar 31
+    def boundary_at(index)
```
annvelents: start + steps
anchor + steps * intervales
beginning_at(steps)
lower_boundary_in(steps)
lower_boundary_at(steps)
beginning_boundary_at(steps)
...

the problem is that boundaries are always 2, so if we have just `boundary`, it's not clear which one is it

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,89 @@
...
+    def covered_days(from, to)
+      (opening_date(to) - opening_date(from)).to_i
+    end
+
+    # The first local date whose midnight is at or after `timestamp`. Compares against
+    # midnight rather than adding a day, so a DST transition cannot shift the answer.
+    #
+    # @example in "Europe/Paris"
+    #   opening_date(Jun 16 00:00 Paris)  # => Jun 16   already a local midnight
+    #   opening_date(Jun 16 00:00 UTC)    # => Jun 17   02:00 in Paris, so the 16th is gone
```
annvelents: add here an example with not midnight please

app/services/billing/segments.rb
```diff
@@ -0,0 +1,60 @@
...
+    # `effective_from`; order does not matter.
+    #
+    # @example [Sep 10, Oct 10) with rates effective Jan 1 (A) and Sep 25 (B)
+    #   within(...)  # => [Sep 10, Sep 25) at A, [Sep 25, Oct 10) at B
+    # @example a change landing exactly on the start is simply the rate in force
+    #   rates effective Jan 1 (A) and Sep 10 (C)  # => [Sep 10, Oct 10) at C, one segment
+    # @example no rate in force yet — that window is dropped rather than billed at nothing
+    #   a single rate effective Sep 25  # => [Sep 25, Oct 10), the first 15 days unbilled
+    def self.within(window, rates:)
+      segment_starts = [window.begin] + rate_changes_inside(window, rates)
```
annvelents, line 8: start_dates? 😅

spec/services/billing/rate_cards/build_schedule_service_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    it "takes the cadence from the rate in force when the window opens" do
+      cycle = result.schedule.cycles_due_by(Time.utc(2026, 3, 1)).first
+
+      expect(cycle.from...cycle.to).to eq(Time.utc(2026, 1, 15)...Time.utc(2026, 2, 1))
+    end
+
+    # The later rate is not ignored — it cuts the cycle into two priced windows instead.
+    it "bills the change as a segment rather than as a new cadence" do
+      cycle = result.schedule.cycles_due_by(Time.utc(2026, 3, 1)).first
+      segments = cycle.segments(rates: rate_card.rates.order(:effective_from))
```
annvelents, line 157: shall we have default order on the relation?

app/services/billing/calendar.rb
annvelents: interval_window?

app/services/billing/calendar.rb
annvelents: period_start_for_interval(index)

annvelents: interval_with_index_starts_at(index)

app/services/billing/schedule.rb
```diff
@@ -0,0 +1,233 @@
...
+      phases.each_with_index do |phase, position|
+        break if pending || ended?(cursor)
+
+        calendar = calendar_for(phase, position, cursor)
+        produced = 0
+
+        while phase.cycle_count.nil? || produced < phase.cycle_count
+          break if ended?(cursor)
+
+          period = calendar.period_containing(cursor)
```
annvelents: interval

app/services/billing/schedule.rb
```diff
@@ -0,0 +1,233 @@
...
+          end
+
+          due << cycle
+          cursor = period.end
+          index += 1
+          produced += 1
+        end
+      end
+
+      [due, pending]
```
annvelents: due is an array, pending is one record
[[due1, due2], pending]

annvelents: 1) why do we return all the due?
2) might it be needed to memoise pending?

app/services/billing/schedule.rb
```diff
@@ -0,0 +1,233 @@
...
+    # well as forward: a cycle already due but still running has not been billed, so it is
+    # what the clock owes next. Only once every due cycle has also closed does the answer
+    # move to the cycle ahead.
+    #
+    # @example advance, backdated to Dec 15 and asked on Mar 10: Mar 1, the cycle in force
+    # @example arrears, same card: Apr 1, since every closed cycle is already behind us
+    def next_billing_at(timestamp)
+      due, pending = walk(timestamp)
+
+      due.find { |cycle| cycle.to > timestamp }&.due_at || pending&.due_at
```
annvelents: `due.find { |cycle| cycle.to > timestamp }&.due_at` this should not be possible
and then we can only return pending and we don't need due

app/services/billing/schedule.rb
```diff
@@ -0,0 +1,233 @@
...
+    end
+
+    # `starts_at` is floored to the start of its day in `timezone`: a card attached at 12:34
+    # is billed for that whole day, the same rule that counts a day begun as a day used. The
+    # flooring is here so no caller can forget it — a schedule opening mid-day would lose
+    # the customer part of a day in silence.
+    #
+    # `ends_at` is deliberately not floored: a termination at 12:34 closes the cycle there,
+    # and the day it happened on is paid for by the day count, not by moving the boundary.
+    def initialize(anchor_date:, phases:, timezone:, starts_at:, timing:, ends_at: nil)
```
annvelents: shall we pass `rate_card` instead of :phases, :starts_at, :timing, to enhance that schedule is not per subscription, not per anything else, but per rate_card?

annvelents: or rename it into rate_card_schedule

app/services/billing/build_schedule_service.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    def base_rate
+      @base_rate ||= subscription_rate_card.rate_card.rate_active_at(starts_at)
+    end
+
+    # One phase per configured rate phase, and the card's own default to carry whatever they
+    # leave over — which is why a schedule always has at least one.
+    def phases
+      configured = rate_phases.map do |phase|
+        Billing::Schedule::Phase.new(
```
annvelents: why do we define a data object, can we just use AR model?
looks like we'll only need to define `def interval` on it, and that's it?

annvelents: also then we can remove phase from schedule and it also will be cleaner

app/services/billing/interval.rb
```diff
@@ -0,0 +1,88 @@
...
+  # @example Interval.new(count: 2, unit: :week)   is "every two weeks"
+  # @example Interval.new(count: 1, unit: :month)  is "monthly"
+  class Interval < Data.define(:count, :unit)
+    UNITS = %i[day week month year].freeze
+
+    # @example a monthly rate, and an override that only changes the unit
+    #   rate     = billing_interval_count: 1,   billing_interval_unit: "month"
+    #   override = billing_interval_count: nil, billing_interval_unit: "week"
+    #   Interval.from(rate, override:)  # => Interval(count: 1, unit: :week)
+    def self.from(rate, override: nil)
```
annvelents, line 15: can it also be created from rate_card?
because a rate_card should have rates with the same interval

app/services/billing/build_schedule_service.rb
```diff
@@ -0,0 +1,102 @@
...
+          cycle_count: phase.billing_interval_cycle_count,
+          interval: Billing::Interval.from(base_rate, override: phase.rate_override),
```
annvelents: `Billing::Iterval.from_rate_card(phase.rate_card)`

annvelents: if an override can have have a different interval, how do we do with rates with effective dates in the middle of the cycle?

annvelents: Image
so if in this example, rateB is an override, and it's 3 days (or 4 weeks 🤷 )

app/services/billing/build_schedule_service.rb
tiagolupepic: I'll not use AR model. You're mixing the persisted RatePhase which contains an id with the `default_phase` `Billing::Schedule::Phase.new(cycle_count: nil, code: nil, override: nil)` which doesn't contain an id. The next developer can accidentally persist/change such phase inside of any service.

app/services/billing/build_schedule_service.rb
tiagolupepic: I'm not following. How the rate_override can have an effective_date?

tiagolupepic: The previous implementation was first splitting the effective dates into Cycle and inside of each cycle the Periods were created using the override

brunomiguelpinto: yup @tiagolupepic  you are correct it was me that generated the image wrongly and created the confusion already added a spec that proves this i will update the  description and the image

app/services/billing/schedule.rb
brunomiguelpinto: no removed

app/services/billing/schedule.rb
brunomiguelpinto: It returns the card's cycles across all phases, not one phase's — the prefix would mislead.

app/services/billing/segments.rb
brunomiguelpinto: They're instants, not dates — the cuts land at 14:30 and 07:05 in the specs, an

app/services/billing/schedule.rb
brunomiguelpinto: via the namespace — it's Billing::RateCards::Schedule now, which says "rate card schedule" without stuttering.

app/services/billing/build_schedule_service.rb
brunomiguelpinto: RatePhase.new(position: nil, billing_interval_cycle_count: nil, code: nil) yes we produce this one has a default phase

app/services/billing/interval.rb
brunomiguelpinto: not quite — rate_cards has no interval columns at all; they live on rate_card_rates (and rate_overrides),

app/services/billing/schedule.rb
brunomiguelpinto: Right for arrears — there the branch can never fire, since a cycle is only in due because ended_at <= timestamp. In advance it's the opposite: due_at is the cycle's started_at, so a cycle can be billable and still running.

app/services/billing/schedule.rb
brunomiguelpinto: we return all of them on the /cycles and /bill payloads

db/migrate/20260901122946_create_billing_segments.rb
brunomiguelpinto: I'd leave it out: cards don't share cycles. billing_anchor_date is per subscription rate card

db/migrate/20260901122946_create_billing_segments.rb
brunomiguelpinto: t.datetime :started_at, null: false
t.datetime :ended_at, null: false
t.datetime :cycle_started_at, null: false

app/models/billing_segment.rb
brunomiguelpinto: yup added it

app/models/billing_segment.rb
brunomiguelpinto: Renamed — went with started_at / ended_at rather than the _on/_date forms, because these are datetime columns and a cut can land at 14:30.

app/services/billing/segments.rb
brunomiguelpinto: Checked it, and as written it comes out slower at every size — 0.55 vs 0.33 µs at n=1, 25.8 vs 23.4 at n=200. The reason is ordering: the current code selects first (O(n)) and sorts only what survives, which for a one-month window is zero or one entry. Sorting all n up front to bsearch for one hit pays more than the pass it saves.

app/services/billing/segments.rb
annvelents: segments_start_timestamps

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,95 @@
...
+    def interval_index(timestamp)
+      interval.steps_between(anchor, timestamp.in_time_zone(timezone))
+    end
+
+    # The boundary at position `index`
+    #
+    # @example anchored Jan 31, monthly
+    #   interval_with_index_start_at(1)  # => Feb 28
+    #   interval_with_index_start_at(2)  # => Mar 31
+    def interval_with_index_start_at(index)
```
annvelents: starts 🙈

app/models/billing_segment.rb
```diff
@@ -0,0 +1,122 @@
...
+# Indexes
+#
+#  billing_segments_no_overlapping_windows                       (organization_id, subscription_id, customer_id, subscription_rate_card_id, tsrange(started_at, ended_at, '[]'::text)) USING gist
+#  idx_on_subscription_id_billing_at_status_943e595a4d           (subscription_id,billing_at,status)
+#  idx_on_subscription_rate_card_id_cycle_started_at_23c80567e3  (subscription_rate_card_id,cycle_started_at)
+#  index_billing_segments_on_card_and_start                      (subscription_rate_card_id,started_at) UNIQUE
+#  index_billing_segments_on_customer_id                         (customer_id)
+#  index_billing_segments_on_invoice_id                          (invoice_id)
+#  index_billing_segments_on_organization_id                     (organization_id)
+#  index_billing_segments_on_pricing_unit_id                     (pricing_unit_id)
```
annvelents: is it needed?

app/models/billing_segment.rb
```diff
@@ -0,0 +1,122 @@
...
+#  index_billing_segments_on_rate_card_rate_id                   (rate_card_rate_id)
```
annvelents: do we need this one?

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,274 @@
...
+        index = 0
+        due = []
+
+        loop do
+          break if ended?(cursor)
+
+          rate = cadence_rate_at(cursor)
+          break unless rate
+
+          phase = phase_of(index)
```
annvelents: phase_by_index

app/services/billing/calendar.rb
```diff
@@ -0,0 +1,51 @@
...
+      Days.between(from, to, timezone:).fdiv(Days.between(window.begin, window.end, timezone:))
+    end
+
+    private
+
+    attr_reader :anchor, :interval, :timezone
+
+    # Which interval `timestamp` falls in — a position on this ruler, not a cycle number. The
+    # anchor is a reference day, not a start date, so instants before it have negative indices.
+    def interval_index(timestamp)
```
annvelents: index_of_interval?

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,274 @@
...
+        cursor = starts_at
+        anchor = anchor_date
+        cadence = nil
+        index = 0
+        due = []
+
+        loop do
+          break if ended?(cursor)
+
+          rate = cadence_rate_at(cursor)
```
annvelents: what does cadence stand for?

annvelents: can it be rate_at(cursor)?

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,274 @@
...
+      # Three exits: the card closed, the first undue cycle was reached, or no rate resolved —
+      # the last being unreachable while validate! requires a rate.
+      def walk_from(timestamp)
+        cursor = starts_at
+        anchor = anchor_date
+        cadence = nil
+        index = 0
+        due = []
+
+        loop do
```
annvelents: inside of this loop we're building cycle for cursor
maybe having it abstracted to a separate function will help

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,274 @@
...
+
+          rate = cadence_rate_at(cursor)
+          break unless rate
+
+          phase = phase_of(index)
+          interval = Interval.from(rate, override: phase.rate_override)
+          # When the cadence changes, billing re-anchors on the day it changes (LAGO-1766,
+          # "reading B, everything is glued"). The mode that would have kept the card's own
+          # anchor was built and reverted.
+          anchor = cursor.in_time_zone(timezone).to_date if cadence && cadence != interval
```
annvelents: if we don't change the cadence, anchor will be == cursor as well, so we can remove saving cadence, and just always calculate anchor

app/services/billing/rate_cards/schedule.rb
annvelents: cycle = create_cycle_for_cursor(cursor, current_cycle_index)

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,274 @@
...
+      # One loop, because the phase of cycle N is a function of N: phases carry their length, so
+      # there is no second loop to count cycles within a phase.
+      #
+      # Three exits: the card closed, the first undue cycle was reached, or no rate resolved —
+      # the last being unreachable while validate! requires a rate.
+      def walk_from(timestamp)
+        cursor = starts_at
+        anchor = anchor_date
+        cadence = nil
+        index = 0
```
annvelents: current_cycle_index

app/services/billing/rate_cards/schedule.rb
annvelents: but then we cannot use memoisation of the calendar 😖

annvelents: something = cursor.in_time_zone(timezone).to_date
anchor =  something if  days_between(anchor, something) % interval != 0

annvelents: what I do not fully like is that to support the memoisation we are sharing between loops two additional variables

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,274 @@
...
+          phase = phase_of(index)
+          interval = Interval.from(rate, override: phase.rate_override)
+          # When the cadence changes, billing re-anchors on the day it changes (LAGO-1766,
+          # "reading B, everything is glued"). The mode that would have kept the card's own
+          # anchor was built and reverted.
+          anchor = cursor.in_time_zone(timezone).to_date if cadence && cadence != interval
+          cadence = interval
+
+          calendar = calendar_for(anchor, interval)
+          window = calendar.interval_containing(cursor)
```
annvelents: is it window or interval? :)

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,274 @@
...
+          cursor = window.end
+          index += 1
+        end
+
+        [due, nil]
+      end
+
+      # Which phase cycle `index` belongs to. The last phase is always unbounded — validate!
+      # enforces it — so this always answers.
+      def phase_of(index)
```
annvelents: cycle_index?

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,274 @@
...
+      # The billing instant of the cycle covering an instant — what SEEDS the clock.
+      #
+      # @param timestamp [Time] the instant the cycle must cover
+      # @return [Time, nil] nil once the card has stopped billing
+      #
+      # Counts a cycle already due but still running, so an advance card is not skipped past
+      # the period it is still serving. #next_billing_at answers strictly after, and a cycle
+      # due now is not after now.
+      def billing_at_covering(timestamp)
+        due, pending = walk(timestamp)
```
annvelents: when walking, can we not start walking from started at, but start walking from the latest saved segment that is processed?
imagine, we have a subscription that is 2 years old with weekly billing, with several phases

app/services/billing/rate_cards/build_schedule_service.rb
```diff
@@ -0,0 +1,135 @@
...
+      # alone — must not thereby lose the plan's phases: passing nil straight through makes the
+      # resolver fall back to `nil&.rate_phases.to_a`, i.e. [], and every phase and every override
+      # the plan configured disappears with no error.
+      #
+      # So the hint is a shortcut, not the only route. A hint for a DIFFERENT rate card is a
+      # caller bug, not something to ignore.
+      def resolved_plan_rate_card
+        return plan_rate_card_of_subscription_plan if plan_rate_card.nil?
+
+        unless plan_rate_card.rate_card_id == subscription_rate_card.rate_card_id
```
annvelents: let's do it a direct logic
(if a!= b)

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,305 @@
...
+      def proration_ratio_of(cycle, window)
+        if terms.prorated
+          cycle.calendar.proration_ratio(window.started_at, window.ended_at)
+        else
+          1.0
+        end
+      end
+
+      # Segments are half-open; the requested range can include or exclude its end.
+      def overlaps?(window, range)
```
annvelents: segment? :)

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,305 @@
...
+        else
+          Days.between(segment.started_at, at, timezone:).fdiv(billed_days)
+        end
+      end
+
+      # Next billing instant strictly after the given time, or nil when billing has ended.
+      def next_billing_at(after:)
+        due, pending = walk(after, from: after)
+
+        billable_segments_of(due.last(1) + [pending].compact)
```
annvelents: why using segments, and not cycles?

annvelents: because the provided timestamp can be in the middle of the cycle when we're changing phase to pay in advance, so the next billing should happen in the middle of the cycle

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,305 @@
...
+
+      # Due listings start at resume_at; explicit queries start at the requested date.
+      # Cache both bounds so historical queries cannot replace resumed results.
+      def walk(timestamp, from: resume_at)
+        @walks ||= {}
+        @walks[[timestamp, from]] ||= walk_from(timestamp, from:)
+      end
+
+      # Collect due cycles and stop at the first pending cycle or the card's end.
+      def walk_from(timestamp, from:)
```
annvelents: walk_to(timestamp, from:)

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,308 @@
...
+          calendar = cycle.calendar
+          cycle_index += 1
+        end
+
+        [due, nil]
+      end
+
+      # Build the cycle at the cursor, clipped to the card's lifetime.
+      def cycle_at(cursor, cycle_index, calendar)
+        window, calendar, phase = window_at(cursor, cycle_index, calendar)
```
annvelents: cycle_boundaries, calendar, phase = cycle_somthing_at()

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,308 @@
...
+      # Reuse calendars across cycles and queries with the same anchor and interval.
+      def calendar_for(anchor, interval)
+        @calendars ||= {}
+        @calendars[[anchor, interval]] ||= Calendar.new(anchor_date: anchor, interval:, timezone:)
+      end
+
+      # Before the first rate takes effect, use its cadence to keep counting cycles.
+      # Segments.within omits the unpriced portion (decision #56).
+      def rate_for_interval(cursor)
+        Segments.rate_at(rates, cursor) || rates.min_by(&:effective_from)
```
annvelents: rate_covering(cursor) ?

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,308 @@
...
+          ended_at: ends_at ? [window.end, ends_at].min : window.end,
+          phase:,
+          calendar:
+        )
+      end
+
+      # Resolve the phase and calendar at a cycle start, then return its date window.
+      def window_at(cursor, cycle_index, calendar)
+        phase = phase_by_index(cycle_index)
+        interval = Interval.from(rate_for_interval(cursor), override: phase.rate_override)
```
annvelents: interval = Interval.from(rate_covering(cursor) (?), override: phase.rate_override)

app/services/billing/rate_cards/schedule.rb
```diff
@@ -0,0 +1,308 @@
...
+        calendar = calendar_at(cursor, interval, calendar)
```
annvelents: calendar = calendar_at(cursor, interval, prev_calendar: (?))
so we know that we're defining calendar by cursor and interval, but we can also send previous_calendar to be returned instead

annvelents (approved):
LGTM! 👏  ⭐ 🦸 🪨 🐐 🤘 🎊  🥇

PR #6268, opened 2026-09-01
feat(usage): route current usage to the buckets [ING-586]

vincent-pochet:
## Context

Current usage aggregates raw events on every read, so a large subscription is slow and the whole
cost of the read lands on ClickHouse. The pre-aggregated buckets and the store that answers from
them now exist, but nothing reaches for either.

## Description

- Fetch the buckets once per computation in the event store provider, and mint a bucket-backed
  store for every charge and filter they cover, leaving the rest on the events.
- Refuse a read narrower than a charge and its filter, a lifetime window, a window other than the
  prefetched one, a truncated window, a deduplicated organization and a billing segment: the
  buckets answer at none of those granularities.
- Ask the provider whether any answer could be precomputed before building an aggregator, then
  bypass the charge cache for the filters served that way.
- Make serving opt-in per caller.

- The usage endpoints, the customer portal, the wallet refresh and usage alerting serve current
  usage from the buckets, and stop caching the filters served that way. Daily usage and the
  invoice paths keep reading events.
- Serving still requires a premium license, `LAGO_REALTIME_USAGE_ENABLED`, the ClickHouse events
  store and the `realtime_usage` flag, so nothing changes until all four are on.


spec/services/daily_usages/compute_service_spec.rb
```diff
@@ -51,6 +51,16 @@
     context "when there is usage" do
       before { event }
 
+      context "with the usage service spied on" do
+        before { allow(Invoices::CustomerUsageService).to receive(:call).and_call_original }
+
+        it "computes the usage from the events, as the row it persists cannot be corrected later" do
```
toommz: Not sure to understand this example description 🤔

spec/services/fees/charge_service_spec.rb
```diff
@@ -4698,4 +4700,229 @@ def create_filter(amount:, values:)
...
+    end
+
+    it "bills what the store answers without reading a single event" do
+      result = charge_subscription_service.call
+      expect(result).to be_success
+
+      expect(result.fees.first).to have_attributes(units: 12, events_count: 3, amount_cents: 24_000)
+    end
+
+    it "bypasses the charge cache, whose staleness the precomputed store removes" do
```
toommz:
```suggestion
    it "bypasses the stale charge cache and uses the precomputed store instead" do
```

spec/services/fees/charge_service_spec.rb
```diff
@@ -4698,4 +4700,229 @@ def create_filter(amount:, values:)
...
+      allow(RealtimeUsage::FetchBucketsService).to receive(:call)
+        .and_return(RealtimeUsage::FetchBucketsService::Result.new.tap { it.usage_buckets = usage_buckets })
+    end
+
+    it "bills what the store answers without reading a single event" do
+      result = charge_subscription_service.call
+      expect(result).to be_success
+
+      expect(result.fees.first).to have_attributes(units: 12, events_count: 3, amount_cents: 24_000)
+    end
```
toommz: I don't really get how it ensures that we don't read a single event

toommz (approved):

PR #6270, opened 2026-09-02
feat(api): serve contracts on the v2 API

rsempe:
## Context

First runtime slice on the dedicated contracts storage (#6235): the catalog API records agreements as contracts instead of falling through to the legacy subscription flow. The applied-rate-card attach endpoints ship separately (coordinating with the units-versioning work).

## Description

- `POST /api/v2/contracts` creates the agreement: optional plan (a plan-less contract prices through directly attached cards), catalog plans only (`not_a_product_catalog_plan`), date validations, pending when the start is in the future, and **one live agreement per external id** — a duplicate among pending/active is rejected with `value_already_exists`; replacement flows will create their pending sibling explicitly when they exist.
- Creating a contract on a plan materializes one `contract_rate_card` per plan card (anchor, clock, units); pricing resolves by reference — phases are not copied.
- `GET /api/v2/contracts` and `/:external_id` serve the v2 shape with current and scheduled cards: grouped card counts on index (no per-row COUNT), status filters guarded against unknown PG-enum values, and external ids may contain dots (same constraint as v1).
- `contract` joins the ApiKey permission resources; missing locale entries for the contract window validations (`must_be_after_started_at`, `must_be_after_effective_date`) are added — they returned "Translation missing" before.

### DIFF
--- a/app/models/contract_rate_card.rb
+++ b/app/models/contract_rate_card.rb
@@ -24,6 +24,11 @@ class ContractRateCard < ApplicationRecord
 
   default_scope -> { kept }
 
+  # The window is day-grained and the end inclusive: the card still bills on
+  # its ended_date, charges stop after it. Ended attachments are history;
+  # upcoming ones stay visible.
+  scope :current_and_scheduled, -> { where("ended_date IS NULL OR ended_date >= ?", Date.current) }
+
   private
 
   def validate_effective_before_ended
--- a/app/models/contract_rate_card.rb
+++ b/app/models/contract_rate_card.rb
@@ -28,11 +28,18 @@ class ContractRateCard < ApplicationRecord
   # its ended_date, charges stop after it. Ended attachments are history;
   # upcoming ones stay visible. "Today" is the customer's day, not the
   # application's — same COALESCE chain as the v1 timezone SQL helpers.
+  #
+  # The plain date bound comes first: the timezone expression cannot use an
+  # index, which turned a similar comparison into a full-scan timeout in the
+  # termination clock (see #6086). Offsets span -12:00..+14:00, so any
+  # customer-local today is at least yesterday's date — the bound is a strict
+  # superset and the exact per-row check decides.
   scope :current_and_scheduled, -> {
     joins(contract: [:customer, :organization]).where(
-      "contract_rate_cards.ended_date IS NULL OR contract_rate_cards.ended_date >= " \
-      "(?::timestamptz AT TIME ZONE COALESCE(customers.timezone, organizations.timezone, 'UTC'))::date",
-      Time.current
+      "contract_rate_cards.ended_date IS NULL OR (contract_rate_cards.ended_date >= ? AND " \
+      "contract_rate_cards.ended_date >= " \
+      "(?::timestamptz AT TIME ZONE COALESCE(customers.timezone, organizations.timezone, 'UTC'))::date)",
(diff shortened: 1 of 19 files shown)


app/models/contract_rate_card.rb
```diff
@@ -24,6 +24,18 @@ class ContractRateCard < ApplicationRecord
...
   default_scope -> { kept }
 
+  # The window is day-grained and the end inclusive: the card still bills on
+  # its ended_date, charges stop after it. Ended attachments are history;
+  # upcoming ones stay visible. "Today" is the customer's day, not the
+  # application's — same COALESCE chain as the v1 timezone SQL helpers.
+  scope :current_and_scheduled, -> {
+    joins(contract: [:customer, :organization]).where(
+      "contract_rate_cards.ended_date IS NULL OR contract_rate_cards.ended_date >= " \
+      "(?::timestamptz AT TIME ZONE COALESCE(customers.timezone, organizations.timezone, 'UTC'))::date",
```
tiagolupepic: we had issues in another service because of this https://github.com/getlago/lago-api/pull/6086.

app/services/contracts/create_service.rb
```diff
@@ -0,0 +1,96 @@
...
+      # failing, so formats are rejected explicitly.
+      %i[billing_anchor_date started_at ended_at].each do |field|
+        if params[field].present? && !Utils::Datetime.valid_format?(params[field])
+          return result.single_validation_failure!(field:, error_code: "value_is_invalid")
+        end
+      end
+
+      # A window that already closed cannot be created: nothing would ever
+      # terminate it, leaving a zombie active contract.
+      if params[:ended_at].present? && Time.zone.parse(params[:ended_at].to_s) <= Time.current
```
tiagolupepic: Time.zone.parse is defined by the Rails app timezone. this should be validated agains the customer timezone, no?

app/services/contracts/create_service.rb
```diff
@@ -0,0 +1,96 @@
...
+
+      # One live agreement per external id. Replacement flows (upgrade,
+      # downgrade, renewal) will create their pending sibling explicitly when
+      # they exist; a blind second create is a mistake, not a replacement.
+      # Advisory under concurrency — the partial unique index closes the race.
+      if organization.contracts.where(status: %w[pending active], external_id: params[:external_id]).exists?
+        return result.single_validation_failure!(field: :external_id, error_code: "value_already_exists")
+      end
+
+      started_at = params[:started_at].present? ? Time.zone.parse(params[:started_at].to_s) : Time.current
```
tiagolupepic: ditto

app/services/contracts/materialize_rate_cards_service.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    private
+
+    attr_reader :contract
+
+    # Customer-local day of the start instant: the engine interprets card
+    # dates in the customer's timezone, so a UTC truncation would attach the
+    # card one day early or late around the customer's midnight.
+    def effective_date
```
tiagolupepic, line 45: 👌

app/models/contract_rate_card.rb
rsempe: oh interesting! did the same operation 👍

tiagolupepic (approved):
:shipit:

PR #6274, opened 2026-09-02
fix(exports): serialize the charge on fixed-charge fees

mikeh-lago:
## Context

Every `CASE f.fee_type` block in `exports_fees_v03` handled values 0 to 4 only. `Fee::FEE_TYPES` now has `fixed_charge` at index 5, so fixed-charge fees fell into the `ELSE` branches and were exported with the plan's code, name and description, `item_type` `'subscription'`, `lago_item_id` set to the subscription id, and `type` resolving to the literal `'unknown'`. `lago_fixed_charge_id` was not exposed at all, so the fee could not be attributed to the charge that produced it.

Reported by Foxglove across all 540 of their fixed-charge fees from June to August. On one invoice a `byos` add-on and 47 `developer_seats` carried byte-identical `item` payloads naming the plan, distinguishable only by units and amount.

Unlike the other Foxglove issues this one is not grace-period dependent. It affects every fixed-charge fee for every Data Pipeline customer, always.

## Change

`exports_fees_v04` adds a `WHEN 5` branch to each block, sourcing the item from the add-on behind the fixed charge, which is what `Fee#item_code`, `#item_name`, `#item_description`, `#item_id` and `#item_type` already do and what `V1::FeeSerializer` exports. `invoice_display_name` follows `Fee#invoice_name`, letting the fixed charge override the add-on. Date boundaries use the `fixed_charges_*` property keys per `Fee#date_boundaries`. `lago_fixed_charge_id` is exposed alongside `lago_charge_id`.

## Notes for review

**The `structure.sql` diff moves the `fixed_charges` table block earlier.** That is `pg_dump` reordering, not an edit: `exports_fees` now depends on `fixed_charges`, so the table has to be emitted before the view. Unavoidable.

**`item_type` is `'add_on'`, not `'fixed_charge'`.** The ticket asks for `'fixed_charge'` and `lago_item_id` = the charge id. I followed the API instead, where `item.type` is the fee type (`fixed_charge`) and `item.item_type` is the item's class (`AddOn`), because deviating would create a third convention for the same field alongside the API and the UI CSV export. The new `lago_fixed_charge_id` column gives the attribution the customer actually needs. Worth confirming with them before this ships.

**`lago_charge_id` stays null** on fixed-charge fees, since they carry `fixed_charge_id` instead. **`invoiceable` also stays null**, because it comes from `charges.invoiceable` and `fixed_charges` has no equivalent column, same as add-on, subscription and credit fees today.

**Fee type 6 (`product`) is deliberately untouched** and left on the existing fallback, tracked separately.

## Testing

Extends `spec/db/views/exports_fees_spec.rb` with a fixed-charge block covering the item payload, the new column, the display-name precedence and the date boundaries. The fee is built with a subscription so the plans join resolves: without it the old `ELSE` branch returned NULL rather than the plan's identity, which would have hidden the defect.

Refs INT-252

### DIFF
--- a/db/migrate/20260902120100_update_exports_fees_to_version_4.rb
+++ b/db/migrate/20260902120100_update_exports_fees_to_version_4.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class UpdateExportsFeesToVersion4 < ActiveRecord::Migration[8.0]
+  def change
+    update_view :exports_fees, version: 4, revert_to_version: 3
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1219,7 +1219,6 @@ DROP TABLE IF EXISTS public.groups;
 DROP TABLE IF EXISTS public.group_properties;
 DROP VIEW IF EXISTS public.flat_filters;
 DROP TABLE IF EXISTS public.fixed_charges_taxes;
-DROP TABLE IF EXISTS public.fixed_charges;
 DROP TABLE IF EXISTS public.fixed_charge_events;
 DROP VIEW IF EXISTS public.exports_wallets;
 DROP TABLE IF EXISTS public.wallets;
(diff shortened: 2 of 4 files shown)


annvelents (approved):

vincent-pochet (approved):

vincent-pochet:
@copilot resolve the merge conflicts in this pull request

PR #6275, opened 2026-09-03
fix(daily-usages): stop leaking the compute lock

vincent-pochet:
## Context

`DailyUsages::ComputeJob` locked with `until_executed`, which releases its lock only after a
successful `perform`. An execution that never completes — retries exhausted, worker killed
mid-perform — held the lock for its whole 12 hour TTL, and with `on_conflict: :log` every later
orchestrator tick for that key was dropped silently, so the row was never written.

## Description

- Lock with `until_and_while_executing`: the enqueue lock is released before `perform`, and a
  separate runtime lock covers the execution.
- Raise the enqueue lock TTL to 3 hours, outliving the three hourly ticks that share one lock key
  plus the enqueue jitter.
- Cap the runtime lock at 30 minutes, since a worker killed mid-perform strands it until it expires.
- Add a regression test that a duplicate job does not execute while the first one is still running.
- Drop the dead `ActiveRecord::RecordInvalid` rescue in `ComputeService`; `DailyUsage` declares no
  validations beyond its `belongs_to` presence checks, and real database failures raise
  `StatementInvalid`, which `retry_on` already matches.

Plain `until_executing` was rejected: `ComputeService` reads before inserting and `daily_usages` has
no unique index on `(subscription_id, usage_date)`, so nothing would keep two overlapping executions
apart. That index is not an option yet — production rows violate it through a separate, still-open
bug.

## Known limitation

The runtime lock is a lease, so it serialises the insert on a best-effort basis only: a worker killed
mid-`perform` strands it for its TTL, and a run outliving the TTL can overlap a later job. No lock
strategy or TTL removes that — the property actually wanted is idempotent persistence, via a unique
index on `(subscription_id, usage_date)` plus `rescue ActiveRecord::RecordNotUnique`, which would
also cover the writers this lock never touched (`FillFromInvoiceService`, `FillHistoryService`,
manual backfills).

Resolving that separate bug is the prerequisite for the index, and is the real fix. This PR is
scoped to removing the
12-hour stranded lock that silently dropped every subsequent tick.

### DIFF
--- a/app/jobs/daily_usages/compute_job.rb
+++ b/app/jobs/daily_usages/compute_job.rb
@@ -11,7 +11,14 @@ module DailyUsages
     end
 
     retry_on ActiveRecord::ActiveRecordError, wait: :polynomially_longer, attempts: 6
-    unique :until_executing, on_conflict: :log, lock_ttl: 2.hours
+    # The enqueue lock must outlive the three hourly ticks that share one lock key (customer-local
+    # 00/01/02) plus the enqueue jitter, so a job still queued dedups the later ticks. The runtime
+    # lock serialises ComputeService's check-then-insert, and stays short because a killed worker
+    # strands it for its whole TTL.
+    unique :until_and_while_executing,
+      on_conflict: :log,
+      lock_ttl: 3.hours,
+      runtime_lock_ttl: 30.minutes
 
     def perform(subscription, timestamp:)
       DailyUsages::ComputeService.call(subscription:, timestamp:).raise_if_error!
--- a/app/jobs/daily_usages/compute_job.rb
+++ b/app/jobs/daily_usages/compute_job.rb
@@ -11,10 +11,6 @@ module DailyUsages
     end
 
     retry_on ActiveRecord::ActiveRecordError, wait: :polynomially_longer, attempts: 6
-    # The enqueue lock must outlive the three hourly ticks that share one lock key (customer-local
-    # 00/01/02) plus the enqueue jitter, so a job still queued dedups the later ticks. The runtime
-    # lock serialises ComputeService's check-then-insert, and stays short because a killed worker
-    # strands it for its whole TTL.
     unique :until_and_while_executing,
       on_conflict: :log,
       lock_ttl: 3.hours,
--- a/app/jobs/daily_usages/compute_job.rb
+++ b/app/jobs/daily_usages/compute_job.rb
@@ -11,7 +11,7 @@ module DailyUsages
     end
 
     retry_on ActiveRecord::ActiveRecordError, wait: :polynomially_longer, attempts: 6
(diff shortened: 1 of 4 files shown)


vincent-pochet:
Addressed in 5e6abd8.

**Execution-time mutual exclusion.** The job now uses `until_and_while_executing` with
`runtime_lock_ttl: 30.minutes`. The enqueue lock is still released in `before_perform` (so retries
and later ticks are never dropped), and a separate runtime lock on the same
`[subscription_id, usage_date]` key covers the whole `perform`. To be precise about the mechanism: a
duplicate *enqueue* during execution still succeeds — it is the duplicate *execution* that is
aborted. Regression test: `spec/jobs/daily_usages/compute_job_spec.rb` "does not run a concurrent
job for the same subscription and date" enqueues and performs a second job from inside the first
job's `perform`, and asserts `ComputeService` runs once. It fails with `until_executing` (service
called twice) and on `main` with `until_executed`.

**Enqueue lock TTL.** Raised from 2 to 3 hours, covering the customer-local 00/01/02 tick window
plus the up-to-30-minute `ComputeAllService#job_wait_time` jitter. One correction to the premise:
now that the enqueue lock no longer spans `perform`, the TTL only has to cover the pending-in-queue
window. It still matters when the queue is backed up and a tick-1 job is still queued as ticks 2
and 3 fire, so the 3 hour value stands.

A unique index on `(subscription_id, usage_date)` was considered and deliberately deferred:
production currently holds rows that violate it, produced by a separate tracked bug (two writers
disagreeing on their existence checks at yearly-plan / monthly-charge billing boundaries). Adding
the index here would start raising in production.

The PR description's claim that deduplication never relied on this lock has been removed — with no
unique index, the lock was the only thing serialising the insert.

vincent-pochet:
Both points are correct, and neither is fixable by tuning this lock.

A killed worker strands the runtime lock, and `super_fetch` re-runs the orphan straight into it, where `on_conflict: :log` aborts it silently. But the enqueue lock is already free, the ticks are 60 minutes apart and the TTL is 30 minutes, so the next tick re-enqueues and succeeds. The residual case is a kill during the last tick of the window — against 12 hours of blocked ticks on `main`.

A run outliving the lease can also overlap the check-then-insert. That is unavoidable: the lock is doing two jobs, deduplicating enqueues and serialising an insert, and a lease can only do the second best-effort. A shorter TTL makes stranding cheaper and expiry likelier; a longer one does the reverse.

The property actually wanted is idempotent persistence — a unique index on `(subscription_id, usage_date)` with `rescue ActiveRecord::RecordNotUnique` — which would also cover the writers this lock never touched. Production rows violate that index today because of a separate open bug, so it cannot land here. Resolving that is the real fix; this PR is scoped to removing the 12-hour stranded lock.

toommz (approved):

PR #6276, opened 2026-09-03
fix(daily-usage): isolate per-organization failures

vincent-pochet:
## Context

`DailyUsages::ComputeAllService#call` walks every organization with revenue analytics support in one
unguarded loop, running three heavy selection queries each. A statement timeout on any of them
propagates out of `find_each` and aborts the whole run, so every organization ordered after the
failing one is never enqueued. Nothing recovers from that: the orchestrator only computes the
previous day, only inside the customer-local 0–2h window, and `Clock::ComputeAllDailyUsagesJob` runs
with `retry: 0`. One organization timing out silently costs all the following ones a full day of
`daily_usages` rows.

## Description

- Process each organization in isolation so a failure skips only that organization and the loop
  continues with the next one.
- Rescue `StandardError` only, so `ActiveRecord::StatementTimeout` is contained while `Interrupt`
  and `Sidekiq::Shutdown` still stop the run.
- Log the skipped organization and run timestamp, and capture the exception in Sentry with the same
  context, so a skip surfaces as an alert instead of as missing rows.
- Add a regression spec proving the other organizations are still enqueued when one raises.

The result stays successful: a skipped organization is not a failed tick, and with `retry: 0`
failing it would only hide the work that did succeed. Selection legs, timezone window, dedup and
enqueue batching are untouched.

### DIFF
--- a/app/services/daily_usages/compute_all_service.rb
+++ b/app/services/daily_usages/compute_all_service.rb
@@ -14,19 +14,11 @@ module DailyUsages
 
     def call
       Organization.with_revenue_analytics_support.find_each do |organization|
-        ids = Set.new
-
-        # Each leg runs its (cheap, indexed) selection query ONCE via pluck. We never iterate the
-        # heavy relation with find_each, which would re-execute the whole query per 1000-row batch.
-        ids.merge(event_subscriptions(organization).pluck("subscriptions.id"))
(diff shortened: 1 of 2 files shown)


tiagolupepic (approved):
:shipit:

PR #6279, opened 2026-09-03
feat(contracts): attach rate cards over REST

rsempe:
## Context

The contract runtime shipped its object and read/create surface, but a plan-less contract had no way to carry pricing — rate cards could only be attached to plans. This is the applied-rate-card surface of the contract runtime, reworking #5871's subscription attach flow onto the dedicated `contract_rate_cards` storage.

## What this adds

- **`POST/GET/PUT/DELETE /api/v2/contracts/:external_id/applied_rate_cards`** + nested **`rate_phases`**, addressed by contract external id and card code, mirroring the `plan_rate_cards` surface.
- **Pending-only authoring**: a pending contract's cards are freely editable; once the contract is `active`/terminated/canceled every write returns `contract_locked` (via `Contract#locked?`).
- **Attach rules** ported from #5871 onto the new storage: one card per pricing slice (`product_already_priced` / `product_filter_already_priced`), currency match against the contract's currency (its plan's, or the customer's for a plan-less contract), day-grained window / anchor / billing clock seeded from the contract.
- **Rate phase services generalized**: `RatePhases::{Replace,Update,Destroy}Service` now accept a `contract_rate_card` parent (Create already did), with a parent-aware lock that reports `plan_locked` or `contract_locked` per parent.
- **`contract_rate_card`** ApiKey permission resource.


rsempe:
Thanks — both points are valid.

**Rate-phase PUT coverage:** fixed in 27825cfc8. Added request specs for the contract rate-phase `update` action — a successful edit while pending, and the `contract_locked` rejection once the contract is active.

**Immediate-start plan-less contracts:** confirmed real. A contract that starts immediately is born `active`, a plan-less contract materialises no cards, and this attach surface rejects active contracts — so that specific shape has no pricing path. It is not blocking today (the v2 engine does not bill yet, and the pending flow works: create with a future `started_at`, attach cards while pending, activate via BIL-448), and the proper fix — atomic initial-pricing (nested `applied_rate_cards` on contract creation, bypassing the active lock only for birth authoring while preserving post-activation immutability) — reaches into the already-merged create surface and is its own capability. Tracked as BIL-641 rather than expanding this PR, keeping #6279 scoped to the pending-contract attach surface.

rsempe:
Confirmed real and fixed in 2723eb3ba. `Contract` has no default scope and the `live_external_id` unique index only covers pending/active, so terminated/canceled siblings reuse the external id as history — an unordered `find_by(external_id:)` could resolve to one of them and wrongly report `contract_locked` (or act on the wrong row).

Added `Contract.live` and `Contract.live_by_external_id` (pending/active, `order(started_at: :desc)`, deterministic — the same ordered-and-status-filtered shape `ContractsController#show` uses), and both attach controllers now resolve through it. Added request coverage for a reused external id with a terminated sibling on both surfaces (applied_rate_cards create and the nested rate_phases), plus model specs for the scope and resolver. CI green, 69 examples across the touched specs.

rsempe:
Design note on the contract lookup (4e945bcfb): reverted `Contract.live_by_external_id` to the ordered form (`live.order(started_at: :desc)`) on purpose.

The earlier explicit pending-then-active resolution only diverged from the ordered lookup when a pending and an active contract coexist under one external id. That state is not reachable today — `Contracts::CreateService` rejects a second live contract on the same external id, and the replacement/handover flow (BIL-637) that would legitimately create it is not built yet. And even for that future case the ordered lookup resolves correctly: a pending contract is future-dated (that is why it is pending), so newest-first already prefers it over the active one. The explicit branching only mattered in a degenerate stale-pending state that should not occur.

Kept the other half of that commit — the card lookups scope to `current_and_scheduled` — because an ended attachment sharing an open one's rate-card code is a real, reachable ambiguity (directly, and once units-versioning lands). That fix and its regression test stand.

Happy to revisit the explicit form alongside BIL-637 if the coexistence state becomes real.

app/services/contract_rate_cards/create_service.rb
```diff
@@ -0,0 +1,113 @@
...
+        rate_cards: {
+          product_id: rate_card.product_id,
+          product_filter_id: rate_card.product_filter_id
+        }
+      )
+    end
+
+    def contract_currency
+      contract.plan&.amount_currency || contract.customer.currency || organization.default_currency
+    end
```
tiagolupepic: this method and its conditionals should belong to contract. Something like `contract.currency_with_fallback`

app/controllers/api/v2/contract_rate_cards/rate_phases_controller.rb
```diff
@@ -0,0 +1,154 @@
...
+module Api
+  module V2
+    module ContractRateCards
+      class RatePhasesController < Api::BaseController
+        include Api::RequiresProductCatalog
+
+        def index
+          return not_found_error(resource: "applied_rate_card") unless contract_rate_card
+
+          render_rate_phases(contract_rate_card.rate_phases.includes(:rate_override).order(:position))
```
tiagolupepic: we have this `order(:position)` in several places and is hard to remember. This should be defined by default in the Rails relationship.

app/services/contract_rate_cards/destroy_service.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def initialize(contract_rate_card:)
+      @contract_rate_card = contract_rate_card
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "applied_rate_card") unless contract_rate_card
+
+      if contract_rate_card.contract.locked?
```
tiagolupepic: `locked?` mislead a bit with https://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html. From my understanding here, the lock option is related to something mutable where the `status == 'pending'`

app/services/contract_rate_cards/update_service.rb
```diff
@@ -0,0 +1,42 @@
...
+    def initialize(contract_rate_card:, params:)
+      @contract_rate_card = contract_rate_card
+      @params = params.to_h.with_indifferent_access
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "applied_rate_card") unless contract_rate_card
+
+      if contract_rate_card.contract.locked?
```
tiagolupepic: ditto

app/services/rate_phases/create_service.rb
```diff
@@ -87,8 +88,13 @@ def build_override
       ).raise_if_error!.rate_override
     end
 
-    def plan_locked?
-      plan_rate_card.present? && plan_rate_card.plan.attached_to_subscriptions?
+    def lock_error_code
```
tiagolupepic: we're repeating this lock_error_code in different services. do we should extract it?

app/models/contract.rb
```diff
@@ -31,6 +31,17 @@ class Contract < ApplicationRecord
...
 
+  LIVE_STATUSES = %w[pending active].freeze
+
+  # The contract currently in effect for an external id. Exactly one can be
+  # live (the partial unique index covers pending/active), while terminated
+  # and canceled siblings accumulate as history under the same external id.
+  scope :live, -> { where(status: LIVE_STATUSES) }
+
+  def self.live_by_external_id(external_id)
+    live.order(started_at: :desc).find_by(external_id:)
```
tiagolupepic: The comment above says "Exactly one can be live," but the partial unique index is on `(organization_id, external_id, status)` scoped to `status IN (pending, active)` — so a `pending` **and** an `active` contract sharing an `external_id` can both satisfy `.live` at the same time. When that happens this lookup relies on `order(started_at: :desc)` to pick a winner, and `started_at` is nullable (`timestamp without time zone`, no NOT NULL), so the result is nondeterministic if two live siblings both have a null `started_at`.

The commit history (`fix: resolve attach lookups deterministically` → `revert: restore ordered live contract lookup`) suggests this was already contentious. Worth either tightening the invariant (a tie-breaker like `created_at`/`id` in the ORDER BY) or correcting the comment to reflect that pending+active can coexist and which one this method intends to return.

app/services/contract_rate_cards/create_service.rb
```diff
@@ -0,0 +1,113 @@
...
+
+    # The card starts with the contract, in the customer's timezone: the
+    # engine reads card dates as customer-local days, so a UTC truncation
+    # would attach it a day early or late around the customer's midnight.
+    def effective_date
```
tiagolupepic: **Duplicated Code.** This `effective_date` helper (and its comment) is byte-for-byte identical to `Contracts::MaterializeRateCardsService#effective_date`, and the `create!` above seeds the same lifecycle triple (`effective_date` / `billing_anchor_date: contract.effective_billing_anchor_date` / `next_billing_at: contract.started_at`) that `MaterializeRateCardsService` also builds. Both services encode the single rule "a card starts with the contract, in the customer's timezone."

Extracting that seeding (e.g. a shared builder or a method on `Contract` returning the default lifecycle attributes) would keep the timezone rule in one place; today a change to how a card inherits the contract's clock has to be made twice and can silently drift between the attach path and the materialize path.

app/queries/contract_rate_cards_query.rb
```diff
@@ -0,0 +1,31 @@
...
+    contract_rate_cards = apply_consistent_ordering(contract_rate_cards)
+
+    result.contract_rate_cards = contract_rate_cards
+    result
+  end
+
+  private
+
+  def base_scope
+    ContractRateCard.where(organization:)
```
endenis: Is the index supposed to include ended cards?

The show endpoint and the contract serializer use `current_and_scheduled`, so an ended card can appear here but return 404 when fetched by code.

app/services/contract_rate_cards/create_service.rb
rsempe: agreed - let's just use `contract.currency`

app/controllers/api/v2/contract_rate_cards/rate_phases_controller.rb
rsempe: removed - we already have `has_many :rate_phases, -> { order(:position) }`

app/services/contract_rate_cards/destroy_service.rb
rsempe: agreed - renamed to `!contract.editable?` 👍

app/services/contract_rate_cards/create_service.rb
rsempe: extracted to `Contract#default_rate_card_lifecycle`

app/queries/contract_rate_cards_query.rb
rsempe: good spot - added `current_and_scheduled`

app/controllers/api/v2/contract_rate_cards_controller.rb
```diff
@@ -0,0 +1,137 @@
...
+          )
+        else
+          render_error_response(result)
+        end
+      end
+
+      private
+
+      def find_contract
+        @find_contract ||= current_organization.contracts.live_by_external_id(params[:contract_external_id])
```
tiagolupepic: `||=` memoizes falsy values poorly here: `live_by_external_id` legitimately returns `nil` for an unknown/terminated external id, and since `nil ||= ...` never caches, every call re-runs the query. `create` calls `find_contract` twice (line 9 guard + line 12 arg) and `index` twice (line 57 + line 65), so a missing-contract request issues the lookup twice. Consider memoizing with `defined?`/an explicit sentinel, or resolving the contract once into a local.

tiagolupepic (approved):
one last comment and it's good to go

:shipit:

PR #6283, opened 2026-09-03
refactor(usage): build the provider once per plan [ING-631]

vincent-pochet:
> Stacked on #6268, review that one first.

## Context

Every charge of a usage computation built its own event store provider, so the store class, the
deduplication decision and the prefetched usage buckets were resolved once per charge. The
prefetched set travelled through `Fees::ChargeService`, which never read it — it only handed it to
the provider it then built.

## Description

- Build one provider per usage computation and give it to every charge of the plan.
- Replace the `usage_buckets` argument of `Fees::ChargeService` with the provider holding them.
- Move the aggregation window onto `BillingPeriodBoundaries`, so the charges period has one
  definition rather than three.
- Refuse an injected provider scoped to another subscription or window, at construction rather than
  at the first read.
- Keep reading events for a projected usage read, whose projection re-aggregates at presentation
  time and would otherwise render below the usage it projects.

Pay-in-advance aggregation still builds its own: its window ends at the event being billed, not at
the end of the period. The invoice, preview and progressive billing paths keep one provider per
charge, since they have no buckets to share yet.


vincent-pochet:
Closing: most of this branch has landed elsewhere, and the branch itself never moved past the
merged stack, so it cannot be rebased cheaply.

Already on `main` or in #6268: routing current usage to the buckets, the per-call-site opt-in, the
premium licence gate, the cache bypass through `ChargeCacheMiddleware`, building one provider per
usage computation, and dropping the bucket coverage probe. The provider API this branch was written
against (`subscription:`, `store_for(charge:)`, `precomputed_options_for`) has since been replaced
by `billing_context` / `metered_item` / `UsageBucketStore`.

The two commits that had not landed are re-cut on top of #6268:

- #6439 — serve every current usage window
- #6440 — keep projected reads off the buckets

Note for #6291: it is based on this branch, so it needs a new base before it can go anywhere.

PR #6284, opened 2026-09-03
feat(exports): expose deleted rows to the pipeline

annvelents:
## Context

The data pipeline syncs by upsert only and cannot propagate deletions, so anything Lago deletes stays in the customer's warehouse forever. `Invoices::RefreshDraftService#reset_invoice_values` hard deletes every `fee`, `invoice_subscription` and `invoices_taxes` row on a draft invoice and recreates them with fresh UUIDs (`fees_taxes` cascades along with the fee, and the credit note refresh does the same to `credit_notes_taxes`). All five tables are exported, and finalization refreshes every draft at least once, so every grace-period invoice leaves a dead copy of its rows in the consumer's warehouse next to the replacement — with nothing to identify which is which.

Adding `deleted_at` to `fees` and soft deleting was the alternative, and is what DAT-487 asked for. It was rejected: `Analytics::MRR`, `Analytics::InvoicedUsage` and `Analytics::GrossRevenue` all query `fees` in raw SQL with no `deleted_at` filter, and the finalization refresh discards fees on an invoice that then becomes `status = 1`, so discarded fees would silently inflate MRR, gross revenue and invoiced usage for exactly the grace-period customers this is meant to help.

## Description

- `record_deletions` — `(organization_id, record_table, record_id, deleted_at)`, indexed on `(organization_id, deleted_at)`, `deleted_at` and `updated_at`.
- A single `record_deletion()` trigger function behind `AFTER DELETE` constraint triggers on `fees`, `fees_taxes`, `invoice_subscriptions`, `invoices_taxes` and `credit_notes_taxes`. `TG_TABLE_NAME` fills `record_table`, so one function covers all five and anything added later. A trigger rather than an app-level insert because an export guarantee has to hold for delete paths nobody has written yet.
- The triggers are `DEFERRABLE INITIALLY DEFERRED`, so a tombstone is written as the transaction commits rather than when the row is deleted — see the ordering note below.
- `exports_record_deletions` view exposing `table_name` and `lago_record_id`.
- `Clock::RecordDeletionsCleanupJob`, daily at 01:20, batched delete of tombstones older than one month.

`fees` keeps exactly its current meaning, so no analytics or reporting query changes behaviour.

### Ordering against an incremental cursor

An immediate trigger stamps the tombstone when the row is deleted. `Invoices::RefreshDraftService` deletes the fees and then keeps recalculating in the same transaction, so that stamp can sit the whole transaction behind the commit that makes the row visible. A second transaction deleting and committing inside that window is exported first, the cursor advances past the earlier stamp, and the slower transaction's tombstone is never read.

Reproduced on Postgres 15 with two sessions, times relative to the slow transaction's `BEGIN`: its `DELETE` ran at `+0.46s` but its tombstone was stamped `+0.00s`, while the fast transaction's tombstone was stamped `+1.03s` and visible from `+1.03s`. The slow tombstone only became visible at `+4.5s`, permanently behind any cursor that had already reached `+1.03s`.

Deferring the triggers moves the stamp to commit time. Re-running the same scenario, the slow transaction's tombstone came out `2.9s` **after** the fast one's, despite its `DELETE` having run about `3s` **before** — ordered with the commits rather than with the deletes. What remains is the interval between the deferred trigger running and the commit becoming visible, which is microseconds and does not grow with transaction duration, so a small fixed replay window covers it rather than one bounded by an enforced transaction-duration limit.

Costs of deferring: transition tables are not available to constraint triggers, so the function inserts per row rather than once per statement, and the after-trigger queue holds one entry per deleted row until commit.

### Follow-ups, not in this PR

- Prequel has to be configured to sync `exports_record_deletions`, with a replay window rather than a strictly forward-only cursor. The view exposes `created_at`, `updated_at` and `deleted_at`, equal on every row with two of them indexed, so whichever cursor column is configured behaves the same.
- Consumers anti-join this feed rather than filtering a `deleted_at` on fees, which is a query change on their side.
- Three exported tables are hard deleted outside the invoicing path and are deliberately not tracked here: `usage_monitoring_alert_thresholds`, `item_metadata`, `integration_customers`.
- Recomputing draft fees in place instead of destroying and recreating them would remove the need for tombstones on this path entirely. Tractable for `fixed_charge` and `subscription` fees, which have a natural key; a `CalculateFeesService` refactor for usage charge fees.

## Tests

- `spec/db/triggers/record_deletions_spec.rb` — a tombstone per tracked table, one row per record on a bulk delete, nothing on rollback, triggers installed on exactly `TRACKED_TABLES` and all deferred, timestamps stamped at commit, an end-to-end `Invoices::RefreshDraftService` run, and a two-connection example asserting a late committer is stamped after a tombstone that committed before it. The file runs with `transaction: false`, since deferred triggers never fire inside a rolled back transaction.
- `spec/db/views/exports_record_deletions_spec.rb`, `spec/models/record_deletion_spec.rb`, `spec/jobs/clock/record_deletions_cleanup_job_spec.rb`, and the `schedule:clean_record_deletions` example in `spec/clockwork_spec.rb`.
- `spec/db`, `spec/clockwork_spec.rb`, the model and cleanup-job specs, both `refresh_draft_service_spec.rb` files and `spec/models/fee_spec.rb` all pass. Rubocop clean.

### DIFF
--- a/db/migrate/20260903164412_create_record_deletions.rb
+++ b/db/migrate/20260903164412_create_record_deletions.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class CreateRecordDeletions < ActiveRecord::Migration[8.0]
+  def change
+    create_table :record_deletions, id: :uuid do |t|
+      t.references :organization, type: :uuid, null: false, foreign_key: true, index: false
+
+      t.string :record_table, null: false
+      t.uuid :record_id, null: false
+
+      # Defaults are required because the only writer is the record_deletion()
+      # trigger, which inserts outside of ActiveRecord.
+      t.datetime :deleted_at, null: false, default: -> { "CURRENT_TIMESTAMP" }
+      t.timestamps default: -> { "CURRENT_TIMESTAMP" }
+
+      t.index %i[organization_id deleted_at], name: "idx_lookup_on_record_deletions"
+      t.index :deleted_at, name: "idx_retention_on_record_deletions"
+      t.index :updated_at, name: "idx_sync_cursor_on_record_deletions"
+    end
+  end
+end
--- a/db/migrate/20260903164412_create_record_deletions.rb
+++ b/db/migrate/20260903164412_create_record_deletions.rb
@@ -8,8 +8,7 @@ class CreateRecordDeletions < ActiveRecord::Migration[8.0]
       t.string :record_table, null: false
       t.uuid :record_id, null: false
 
-      # Defaults are required because the only writer is the record_deletion()
-      # trigger, which inserts outside of ActiveRecord.
+      # Defaulted in the database: the record_deletion() trigger inserts outside ActiveRecord.
       t.datetime :deleted_at, null: false, default: -> { "CURRENT_TIMESTAMP" }
       t.timestamps default: -> { "CURRENT_TIMESTAMP" }
 
--- a/db/migrate/20260903164414_add_record_deletion_triggers.rb
+++ b/db/migrate/20260903164414_add_record_deletion_triggers.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+class AddRecordDeletionTriggers < ActiveRecord::Migration[8.0]
+  TRACKED_TABLES = %w[fees fees_taxes invoice_subscriptions invoices_taxes].freeze
+
+  def up
+    safety_assured do
+      execute <<~SQL
+        CREATE OR REPLACE FUNCTION record_deletion() RETURNS trigger
+        LANGUAGE plpgsql AS $$
+        BEGIN
+          INSERT INTO record_deletions (organization_id, record_table, record_id, deleted_at)
+          SELECT deleted_rows.organization_id, TG_TABLE_NAME, deleted_rows.id, statement_timestamp()
+          FROM deleted_rows;
+
+          RETURN NULL;
+        END;
+        $$;
+      SQL
+
+      TRACKED_TABLES.each do |table|
+        execute <<~SQL
+          CREATE TRIGGER record_deletions_on_#{table}
+          AFTER DELETE ON #{table}
+          REFERENCING OLD TABLE AS deleted_rows
+          FOR EACH STATEMENT
+          EXECUTE FUNCTION record_deletion();
+        SQL
+      end
+    end
+  end
+
+  def down
+    safety_assured do
+      TRACKED_TABLES.each do |table|
+        execute "DROP TRIGGER IF EXISTS record_deletions_on_#{table} ON #{table};"
+      end
+
+      execute "DROP FUNCTION IF EXISTS record_deletion();"
+    end
+  end
+end
--- a/db/migrate/20260903164414_add_record_deletion_triggers.rb
+++ b/db/migrate/20260903164414_add_record_deletion_triggers.rb
@@ -8,9 +8,14 @@ class AddRecordDeletionTriggers < ActiveRecord::Migration[8.0]
       execute <<~SQL
         CREATE OR REPLACE FUNCTION record_deletion() RETURNS trigger
         LANGUAGE plpgsql AS $$
+        DECLARE
+          -- All three timestamps come from one value: leaving created_at/updated_at to the
+          -- column defaults would stamp them with the transaction start, which can be far
+          -- behind the delete and behind the pipeline cursor by the time the row commits.
+          deleted_time timestamp := statement_timestamp() AT TIME ZONE 'UTC';
         BEGIN
-          INSERT INTO record_deletions (organization_id, record_table, record_id, deleted_at)
-          SELECT deleted_rows.organization_id, TG_TABLE_NAME, deleted_rows.id, statement_timestamp()
+          INSERT INTO record_deletions (organization_id, record_table, record_id, deleted_at, created_at, updated_at)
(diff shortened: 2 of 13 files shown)


mikeh-lago:
**[P2] Timestamp cursors can permanently miss late-committing tombstones**

In `db/migrate/20260903164414_add_record_deletion_triggers.rb:15`, `statement_timestamp()` still precedes transaction commit. `Invoices::RefreshDraftService` deletes fees before recalculating them within the same transaction. For example, transaction A deletes at t1 and keeps calculating; transaction B deletes at t2 and commits; the export reads B and advances its timestamp cursor to t2; A then commits with a tombstone stamped t1. A forward-only incremental sync will never read A's tombstone, leaving the original warehouse duplicate. Using the same statement timestamp for all three columns does not close this race.

Before relying on this feed, establish a sync strategy that handles late commits (for example, a replay window backed by an enforced transaction-duration limit) and add a two-connection regression test for this ordering.

Validation on `11647559`: 155 examples, 0 failures across the four new spec files plus `refresh_draft_service_spec.rb` and `fee_spec.rb`. Those tests do not cover this concurrency scenario.

annvelents:
Confirmed, and it was worse than described — `statement_timestamp()` lands at the transaction's start, not at the delete.

Reproduced on PG 15 with two sessions, times relative to the slow transaction's `BEGIN`:

- slow: `DELETE` at `+0.46s`, tombstone stamped `+0.00s`, only visible at `+4.5s`
- fast: tombstone stamped and visible at `+1.03s`

A cursor that reached `+1.03s` never reads the slow tombstone. You were also right that using one statement timestamp for all three columns changed nothing here.

Fixed in 9ad9c01db by making the triggers `DEFERRABLE INITIALLY DEFERRED`, so the tombstone is written as the transaction commits. Re-running the same scenario, the slow transaction's tombstone comes out 2.9s **after** the fast one's, despite its `DELETE` running about 3s **before** — ordered with the commits rather than the deletes.

This doesn't make the race impossible; nothing short of commit timestamps does. But what remains is the interval between the deferred trigger firing and the commit becoming visible, which is microseconds and does not grow with transaction duration. So the replay window becomes a small fixed value instead of one bounded by an enforced transaction-duration limit — I think that removes the prerequisite, but the sync config is yours: this still needs a replay window rather than a strictly forward-only cursor.

Costs of deferring, for the record: constraint triggers can't use transition tables, so the function inserts per row rather than once per statement, and the after-trigger queue holds one entry per deleted row until commit.

The two-connection regression test is in `spec/db/triggers/record_deletions_spec.rb` under "two transactions deleting concurrently". The file moved to `transaction: false`, since deferred triggers never fire inside a rolled-back transaction. 191 examples, 0 failures across it, the other three new spec files, both `refresh_draft_service_spec.rb` files and `fee_spec.rb`.

db/migrate/20260903164412_create_record_deletions.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.string :record_table, null: false
+      t.uuid :record_id, null: false
+
+      # Defaulted in the database: the record_deletion() trigger inserts outside ActiveRecord.
+      t.datetime :deleted_at, null: false, default: -> { "CURRENT_TIMESTAMP" }
+      t.timestamps default: -> { "CURRENT_TIMESTAMP" }
+
+      t.index %i[organization_id deleted_at], name: "idx_lookup_on_record_deletions"
+      t.index :deleted_at, name: "idx_retention_on_record_deletions"
+      t.index :updated_at, name: "idx_sync_cursor_on_record_deletions"
```
toommz: Simple question, why not let active record name the indexes?

spec/db/triggers/record_deletions_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+      FROM pg_trigger AS t
+      JOIN pg_class AS c ON c.oid = t.tgrelid
+      WHERE t.tgname LIKE 'record_deletions_on_%'
+        AND NOT t.tgisinternal
+        AND t.tgdeferrable
+        AND t.tginitdeferred
+    SQL
+
+    expect(installed).to match_array(RecordDeletion::TRACKED_TABLES)
+  end
```
toommz, line 26: Nice 👍 I was thinking about keeping it up to date from the top of the file to review list. It does the job

toommz (approved):

db/migrate/20260903164412_create_record_deletions.rb
annvelents: hah, that's a very good point 🙈 better let AR, yes

db/migrate/20260903164414_add_record_deletion_triggers.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class AddRecordDeletionTriggers < ActiveRecord::Migration[8.0]
+  TRACKED_TABLES = %w[fees fees_taxes invoice_subscriptions invoices_taxes credit_notes_taxes].freeze
+
+  def up
+    safety_assured do
+      execute <<~SQL
+        CREATE OR REPLACE FUNCTION record_deletion() RETURNS trigger
```
tiagolupepic, line 9: it's a good solution using a trigger, simple and straightforward. There is a small downside because we're delegating the responsibility to the database. 
We could evaluate a better option to this using CDC instead or the new tooling used in real time solution

tiagolupepic (approved):
:shipit:

PR #6285, opened 2026-09-03
feat(wallets): consume the refresh trigger topic [ING-632]

vincent-pochet:
## Context

Wallet ongoing balances refresh on a five-minute sweep of a Redis flag, so a balance trails its
usage by up to five minutes. The pipeline already emits a trigger per enriched event at a p50 of
603 ms, so that interval is now the latency left.

## Description

- Consume the realtime usage trigger topic, behind `LAGO_KAFKA_REALTIME_USAGE_TRIGGERS_TOPIC`,
  and refresh the wallets inline: throughput comes from the topic's partitions and Karafka's
  concurrency, not from a job queue.
- Collapse a poll batch into one refresh per customer, filtered through the sweep's scope in one
  query.
- Hold a customer until every one of its subscriptions has buckets at the trigger's watermark,
  pausing the partition rather than sleeping.
- Judge a trigger's age once, on first delivery; leave one older than 30 seconds to the sweep.
- Stop waiting after a five-second grace: the topic is keyed by customer, so a longer wait delays
  everyone else on the partition.
- Floor the watermark read two days below the trigger's own watermark, the table's partition key,
  and skip tombstoned buckets.
- Extract the job's refresh body into `Customers::RefreshWalletService` so both lanes share it,
  under a new `wallet_refresh` customer lock scope the consumer takes without waiting.

The sweep stays authoritative, and already covers every customer this lane walks away from:
ingestion flags them. The refresh carries the wallet ids, so it needs no
`awaiting_wallet_refresh` bookkeeping of its own.


vincent-pochet:
Superseded by #6453, which restarts from the POC branch on top of current main.

PR #6288, opened 2026-09-04
feat(catalog): create the catalog_plans storage

rsempe:
## What this is

The v2 catalog gets its **own** plan table, so a product-catalog plan stops being a `pricing_type: product_catalog` row on the shared `plans` table (carrying nil `interval`/`amount_cents`/`pay_in_advance`).

## Changes

- `catalog_plans` table — the plan **envelope only**: `name`, `code`, `invoice_display_name`, `description`, `currency`, `organization_id`, timestamps, `deleted_at`. No legacy pricing columns, and **no self-reference** (catalog plans don't use plan parent/child overrides — the dive-in's open question, resolved).
- `CatalogPlan` model — paper-trailed, soft-deletable (`default_scope { kept }`), validates name/code presence and currency inclusion.
- Partial unique index on `(organization_id, code) WHERE deleted_at IS NULL` — code unique per org among kept rows.

## Not in scope

No `product_category_id` (plans link to categories through products, not a column) and no `pending_deletion` yet — added later if the catalog destroy flow needs it.

### DIFF
--- a/app/models/catalog_plan.rb
+++ b/app/models/catalog_plan.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+# The product-catalog plan: a named, priced envelope whose pricing lives in
+# its attached rate cards, not in legacy plan columns. It is the v2 catalog's
+# own table — the legacy `plans` table is untouched and never shared.
+class CatalogPlan < ApplicationRecord
+  include PaperTrailTraceable
+  include Currencies
+  include Discard::Model
+
+  self.discard_column = :deleted_at
+
+  belongs_to :organization
+
+  validates :name, presence: true
+  validates :code, presence: true
+  validates :currency, inclusion: {in: currency_list}
+
+  default_scope -> { kept }
+end
+
+# == Schema Information
+#
+# Table name: catalog_plans
+# Database name: primary
+#
+#  id                   :uuid             not null, primary key
+#  code                 :string           not null
+#  currency             :string           not null
+#  deleted_at           :datetime
+#  description          :string
+#  invoice_display_name :string
+#  name                 :string           not null
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#  organization_id      :uuid             not null
+#
+# Indexes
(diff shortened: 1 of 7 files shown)


rsempe:
Thanks. On the two points:

1. **Code uniqueness validation** — already present in the reviewed commit (89a15f915): `validates :code, presence: true, uniqueness: {scope: :organization_id, conditions: -> { where(deleted_at: nil) }}`, with specs covering same-organization rejection, cross-organization reuse, and the code freeing up once the holder is discarded (plus a `save(validate: false)` test keeping the DB-index guarantee for the concurrency path). Looks like this flag was raised against the pre-fix state.

2. **`Organization#catalog_plans`** — added in 77fd289b8, alongside the other org-owned catalog associations (`products`, `rate_cards`, …), with the association assertion in the organization spec. Good catch.

brunomiguelpinto (approved):

endenis (approved):

tiagolupepic (approved):
:shipit:

app/models/catalog_plan.rb
```diff
@@ -0,0 +1,47 @@
...
+  include Currencies
+  include Discard::Model
+
+  self.discard_column = :deleted_at
+
+  belongs_to :organization
+
+  validates :name, presence: true
+  validates :code, presence: true, uniqueness: {scope: :organization_id, conditions: -> { where(deleted_at: nil) }}
+  validates :currency, inclusion: {in: currency_list}
```
tiagolupepic: Minor consistency nit vs. the closest sibling model: `RateCard` validates currency as `validates :currency, presence: true, inclusion: {in: currency_list, allow_nil: true}`, whereas here it's `inclusion: {in: currency_list}` only. Since the column is `NOT NULL`, a missing currency should ideally surface as a `:blank`/presence error rather than an `:inclusion` error. Consider matching `RateCard` (`presence: true, inclusion: {in: currency_list, allow_nil: true}`) so error semantics stay consistent across catalog models.

PR #6289, opened 2026-09-04
fix(taxes): skip zero-amount fees in the Anrok tax payload

annvelents:
## Context

Anrok and Avalara reject tax payloads above 1200 line items, so an invoice carrying many zero-amount charge fees fails tax computation outright. A fee with no amount incurs no tax, yet still counts toward that limit.

#5609 fixed this for current usage only, in `Invoices::CustomerUsageService`. Six other call sites still sent every fee: finalization and draft refresh via `PullTaxesAndApplyService`, `PreviewService`, `Taxes::Invoices::CreateJob`, `ApplyProviderTaxesToStandaloneFeesService`, and the `FetchDraftInvoiceTaxes` mutation.

## Changes

- The filter moves into `Taxes::Invoices::BaseService` — the constructor every invoice tax call already goes through — so one place covers all seven call sites and both providers. It reuses `Fee#taxable?` from #5609 rather than adding a second definition of "taxable"; `DraftFee` gains the same predicate off the sub-total it carries, since it exposes no amount by design.
- `CreateService` and `CreateDraftService` skip the provider entirely when nothing is taxable, reporting `result.fees = []`. Empty rather than `nil` is deliberate: each consumer keeps taking its provider branch and reaches zero taxes on its own, so no call site needs its own guard. `VoidService` / `NegateService` are untouched — their payloads carry only the invoice id and must still reach the provider for invoices reported before this change.
- `Fees::ApplyProviderTaxesService` returns early when a fee has no entry in the response, since five callers loop over every invoice fee and match by `item_id` / `item_key`.

This also means `pro_rated_taxes_rate`'s count-based branch can no longer be reached by an all-zero invoice — there are no applicable taxes to prorate — so #5609's `sub_total_excluding_taxes_amount_cents` pre-set remains the answer for the usage path and nothing needs changing here.

## Side effect worth reviewing

Amounts are unchanged everywhere; excluded fees had a zero taxable base. Rates read differently in two cases:

- A zero-amount fee gets no `Fee::AppliedTax` row, so its rate shows as `0.0%` on the PDF, `taxrate` 0 in the Netsuite/Xero sync, and no `fees_taxes` row in the data pipeline.
- An invoice whose fees are all zero gets no `Invoice::AppliedTax` row and `taxes_rate: 0`, where it previously carried the jurisdiction rate against a zero base.

## Tests

Zero-fee exclusion and the all-zero skip on `CreateService` (Anrok and Avalara contexts) and `CreateDraftService`, asserted against the exact request body; the same two cases on the GraphQL mutation; the nil-`fee_taxes` guard; and in `PullTaxesAndApplyService`, an end-to-end case where the zero fee is absent from the request and ends untaxed while invoice totals hold, plus an all-zero invoice that finalizes with zero taxes and no provider call.


brunomiguelpinto (approved):
only small nitpicks about 0 fees 

should we still send invoice if all fees are 0

app/services/fees/apply_provider_taxes_service.rb
```diff
@@ -14,6 +14,8 @@ def initialize(fee:, fee_taxes:)
     def call
       result.applied_taxes = []
       return result if fee.applied_taxes.any?
+      return result if fee_taxes.nil? && !fee.taxable?
+      return unreported_fee_failure if fee_taxes.nil?
```
osmarluz, line 18: `fee_missing_from_tax_response` has no handling path in any of its three consumers.

In the finalization path, `Invoices::ComputeAmountsFromFees` calls this service with `call!`, so the failure raises `BaseService::FailedResult` inside `PullTaxesAndApplyService`'s transaction. It is then caught by the method-level `rescue BaseService::FailedResult => e; e.result` — which, unlike the `unless taxes_result.success?` branch further up the same method, never calls `create_error_detail`, never sets `tax_status = "failed"`, and never calls `notify_ready_to_finalize`. The invoice stays in `tax_pending` indefinitely, and since `PullTaxesAndApplyJob` also uses `call!` and has no `retry_on` covering `FailedResult`, the job dead-queues with no error detail for anyone to act on.

In `Invoices::PreviewService#apply_provider_taxes` the same failure escapes `rescue BaseService::ThrottlingError, *retryable_errors`, so a preview 500s instead of falling back to `apply_zero_tax`.

Worth deciding what should happen on an incomplete provider response and wiring that into the `PullTaxesAndApplyService` failure branch and the preview rescue, rather than introducing a failure code that only surfaces as a dead job.

app/services/integrations/aggregator/taxes/invoices/void_service.rb
```diff
@@ -36,6 +37,14 @@ def call
 
           private
 
+          # NOTE: An invoice with no taxable fee never reached the provider, so it holds no
+          #       transaction to void and would answer with an error that lands on the customer
+          #       as a tax webhook. Invoices finalized before that skip existed were reported and
+          #       carry the tax rows from that response, so they are still voided.
+          def never_reported?
```
osmarluz, line 45: `never_reported?` uses "the invoice has no tax rows" as a proxy for "the invoice never reached the provider", and that proxy doesn't hold for invoices finalized before this change.

`Taxes::BaseService#tax_breakdown` maps an empty `tax_breakdown` array to `[]`, so an invoice that *was* reported to Anrok for a tax-exempt or unresolved-address customer comes back with no `Invoice::AppliedTax` rows. Voiding one of those now returns success without ever calling `void_invoices`, leaving an open transaction at the provider that gets filed.

The comment above the method assumes pre-skip invoices always "carry the tax rows from that response", which is the case this misses. A real "was reported" marker would be safer than inferring it from the tax rows — an `IntegrationResource` row, or a flag persisted when the create call actually goes out.

app/services/integrations/aggregator/taxes/invoices/base_service.rb
```diff
@@ -18,6 +18,22 @@ def initialize(invoice:, fees: nil)
 
           delegate :customer, to: :invoice, allow_nil: true
 
+          # NOTE: Only fees with a positive amount can incur tax, so non-taxable fees are
+          #       excluded from the provider request. This keeps the payload under the provider
+          #       line-item limit (Anrok and Avalara reject payloads above 1200 items).
+          #       Excluded fees are absent from the response and keep their zero taxes.
+          def taxable_fees
+            @taxable_fees ||= fees.select(&:taxable?)
```
osmarluz: Two separate problems with `taxable?` as the filter predicate.

**`positive?` drops negative fees as well as zero ones.** `Fee#taxable?` is `amount_cents.positive?`, and nothing validates `amount_cents >= 0` — `Fee` only validates non-negativity on `units` and `events_count`, and `Fees::OneOffService` doesn't constrain `unit_amount_cents`. A one-off invoice with a +10_000 add-on line and a -2_000 discount line previously sent both and was taxed on 8_000; it now sends only the +10_000 line, so the customer is taxed on the full amount and the negative line also loses its `Fee::AppliedTax` row. `!amount_cents.zero?` is the safe predicate here. The new `DraftFee#taxable?` has the same issue.

**The predicate and the payload disagree on which attribute matters.** `Fee#taxable?` keys off `amount_cents`, but the payload sends `sub_total_excluding_taxes_amount_cents` (`payloads/anrok.rb:58`). So a fee with `amount_cents` 5000 fully covered by coupons has a taxable base of 0 and is still sent as a line item — meaning an invoice fully covered by coupons can still exceed the 1200-item limit this PR is meant to fix. Meanwhile the new `DraftFee#taxable?` *does* use `sub_total_excluding_taxes_amount_cents`, so this one `taxable_fees` call filters real fees and draft fees by different criteria, and the two will drift as either definition changes. Worth converging on a single definition of "taxable".

osmarluz (commented):
Review of the zero-amount fee filter. One finding below that couldn't be attached inline (the file isn't part of this diff); the rest are inline comments.

### `Invoices::ApplyProviderTaxesService#pro_rated_taxes_rate` now dilutes `taxes_rate`

`app/services/invoices/apply_provider_taxes_service.rb:104-116`

Filtering the provider request shrinks `indexed_fees` but leaves `invoice.fees.count` unchanged, so the count-based branch divides a filtered numerator by an unfiltered denominator.

The PR description says this branch "can no longer be reached by an all-zero invoice" — that part is true, but it isn't only reachable that way. It fires whenever `invoice.sub_total_excluding_taxes_amount_cents` is zero, which includes coupon-covered and progressive-billing-zeroed invoices that still have a mix of taxable and zero-amount fees.

Concretely: fee A with `amount_cents` 1000 fully covered by a 1000-cent coupon, plus 3 zero-amount charge fees, jurisdiction rate 20%. `ComputeAmountsFromFees` sets the invoice sub total to 0, so the count branch runs:

- before: all 4 fees carried a `Fee::AppliedTax` row -> `4.fdiv(4)` -> `taxes_rate` **20.0**
- after: only A is in `indexed_fees` -> `1.fdiv(4)` -> `taxes_rate` **5.0**

The denominator needs to be the set of fees that were actually reported, not `invoice.fees.count`.

app/services/integrations/aggregator/taxes/invoices/base_service.rb
osmarluz: @annvelents just to give you more context: on the `app/services/invoices/customer_usage_service.rb` implemented on #5609 the fees are zeroed on `app/services/fees/charge_service.rb`:

```ruby
# Prevent trying to create a fee with negative units or amount.
if amount_result.units.negative? || amount_result.amount.negative?
  amount_result.amount = amount_result.unit_amount = BigDecimal(0)
  amount_result.full_units_number = amount_result.units = BigDecimal(0)
end
```

That's why the `taxable?` in that scenario is sufficient

osmarluz (approved):
🚀

PR #6290, opened 2026-09-04
feat(billing): add billing segments

tiagolupepic:
## Context

Billing needs a durable representation of priced contract rate card slices so future processing can persist the exact rate snapshot and period boundaries used for invoicing.

## Description

Add the BillingSegment model, table, constraints, indexes, factory, and model specs. Link billing segments to contracts and contract rate cards, persist the cycle start as cycle_started_at, and enforce period, proration, and rate presence invariants at both model and database level.


brunomiguelpinto (approved):
LGTM

annvelents (approved):

PR #6292, opened 2026-09-04
feat(billing): add metered segment source

tiagolupepic:
## Context

Metered fee computation is being prepared to support product-catalog billing through BillingSegment, the renamed durable billing-cycle record. BillingSegment stores the pricing snapshot and period data that should be used when computing metered fees instead of reading mutable rate properties at runtime.

## Description

Add a BillingSegment-backed MeteredItem source and a pricing-structure factory entry point that read charge model, pricing properties, currency, proration, and boundaries from the segment. Add inclusive segment duration support and keep the legacy Charge source compatible with the expanded MeteredItem interface. Cover the new adapter contract with focused specs using in-memory objects where persistence is unnecessary.

### DIFF
--- a/app/models/billing_segment.rb
+++ b/app/models/billing_segment.rb
@@ -39,7 +39,7 @@ class BillingSegment < ApplicationRecord
   end
 
   def duration_in_days
-    (ended_at.to_date - started_at.to_date).to_i + 1
+    Utils::Datetime.date_diff_with_timezone(started_at, ended_at, customer.applicable_timezone)
   end
 
   def pricing_unit_conversion_rate
--- a/app/models/billing_segment.rb
+++ b/app/models/billing_segment.rb
@@ -1,0 +1,134 @@
+# frozen_string_literal: true
+
+# A durable priced slice of a contract rate card billing cycle. A whole cycle has one
+# segment; a rate change inside the cycle creates several segments sharing cycle_started_at.
+class BillingSegment < ApplicationRecord
+  include Currencies
+
+  STATUSES = {
+    pending: "pending",
+    processing: "processing",
+    done: "done",
+    failed: "failed"
+  }.freeze
+
+  belongs_to :organization
+  belongs_to :contract
+  belongs_to :customer, -> { with_discarded }
+  belongs_to :contract_rate_card, -> { with_discarded }
+  belongs_to :invoice, optional: true
+  belongs_to :rate_card_rate, -> { with_discarded }, optional: true
+  belongs_to :rate_override, -> { with_discarded }, optional: true
+  belongs_to :pricing_unit, optional: true
+
+  enum :status, STATUSES, validate: true, prefix: true
+
+  validates :billing_at, presence: true
+  validates :cycle_started_at, presence: true
+  validates :currency, presence: true, inclusion: {in: currency_list, allow_nil: true}
+  validates :started_at, presence: true
+  validates :ended_at, presence: true
+  validates :proration_ratio, numericality: {greater_than_or_equal_to: 0, less_than_or_equal_to: 1}
+
+  validate :validate_rate_presence
+  validate :validate_period_bounds
+  validate :validate_cycle_bounds
+
+  def rate
+    rate_override || rate_card_rate
+  end
+
+  def pricing_unit_conversion_rate
+    if rate_override
+      rate_override.pricing_unit_conversion_rate
+    else
+      rate_card_rate.applied_pricing_unit_conversion_rate
+    end
+  end
+
+  def min_amount_cents
+    if rate_override
+      rate_override.min_amount_cents
+    else
+      rate_card_rate.min_amount_cents
+    end
+  end
+
+  private
+
+  def validate_rate_presence
+    if rate_card_rate || rate_override
+      return
+    end
+
+    errors.add(:base, :rate_card_rate_or_rate_override_required)
+  end
+
+  def validate_period_bounds
+    if started_at.blank? || ended_at.blank? || started_at <= ended_at
+      return
+    end
+
+    errors.add(:ended_at, :must_be_after_started_at)
+  end
+
+  def validate_cycle_bounds
+    if cycle_started_at.blank? || started_at.blank? || cycle_started_at <= started_at
+      return
+    end
+
+    errors.add(:cycle_started_at, :must_be_before_started_at)
+  end
+end
+
+# == Schema Information
+#
+# Table name: billing_segments
+# Database name: primary
+#
+#  id                    :uuid             not null, primary key
+#  billing_at            :datetime         not null
+#  currency              :string           not null
+#  cycle_started_at      :datetime         not null
+#  ended_at              :datetime         not null
+#  proration_ratio       :decimal(30, 10)  default(1.0), not null
+#  rate_properties       :jsonb            not null
+#  started_at            :datetime         not null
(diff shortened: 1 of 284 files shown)


app/services/fees/charge_service/sources/billing_segment.rb
```diff
@@ -0,0 +1,137 @@
...
+        def ignored_filters
+          []
+        end
+
+        def matching_and_ignored_filters
+          BaseResult[:matching_filters, :ignored_filters].new.tap do |result|
+            result.matching_filters = matching_filters
+            result.ignored_filters = ignored_filters
+          end
+        end
```
tiagolupepic, line 87: this is to keep compatibility. The code will be added in another PR

app/services/fees/charge_service/sources/billing_segment.rb
```diff
@@ -0,0 +1,137 @@
...
+        def currency
+          Money::Currency.new(billing_segment.currency)
+        end
+
+        def pricing_structure
+          ChargeModels::PricingStructure.from_billing_segment(billing_segment)
+        end
+
+        def period_ratio
+          billing_segment.proration_ratio
```
tiagolupepic, line 55: `period_ratio` here returns the statically persisted `billing_segment.proration_ratio`, but the sibling `Sources::Charge#period_ratio` computes a *live* ratio from the boundaries and `Time.current` (`days_passed.fdiv(charges_duration)`). Since `MeteredItem` treats both sources polymorphically and callers use `selected_metered_item.period_ratio` for current-usage proration (see `charge_service.rb`), the two implementations mean the same interface method yields live-vs-frozen proration depending on the source. If that divergence is intentional (segment proration is pre-computed), a short comment would help; otherwise this is a behavioral inconsistency worth resolving.

Also note `proration_ratio` is already delegated to `billing_segment` above, so `period_ratio` is a thin re-exposure of the same value.

app/services/fees/charge_service/sources/billing_segment.rb
```diff
@@ -0,0 +1,145 @@
...
+        def matching_filters
+          product_filter&.to_h || {}
+        end
+
+        def ignored_filters
+          []
+        end
+
+        def matching_and_ignored_filters
+          BaseResult[:matching_filters, :ignored_filters].new.tap do |result|
```
tiagolupepic: [P3] Define the filter result type once

`BaseResult[...]` creates a new anonymous class, so this allocates a class on every call rather than just a result instance. It also means otherwise identical results cannot compare equal, because `BaseResult#==` requires the same class. Define a reusable result type once, following `ChargeFilters::MatchingAndIgnoredService::Result`, and instantiate it here.

app/services/fees/charge_service/sources/billing_segment.rb
```diff
@@ -0,0 +1,145 @@
...
+
+        delegate :organization_id,
+          :rate,
+          :rate_card_rate,
+          :rate_override,
+          :pricing_unit,
+          :proration_ratio,
+          to: :billing_segment
+
+        delegate :charge, to: :product
```
brunomiguelpinto: Implement the remaining MeteredItem source methods

`MeteredItem` delegates both `dynamic?` and `charge_id` to its source, but this source implements neither. Delegating `charge` here does not also expose those methods.

brunomiguelpinto (commented):
One source-interface issue reproduced with this PR applied to the current main branch.

app/services/fees/charge_service/sources/billing_segment.rb
```diff
@@ -0,0 +1,145 @@
...
+          {
+            free_units_per_events: properties["free_units_per_events"].to_i,
+            free_units_per_total_aggregation: BigDecimal(properties["free_units_per_total_aggregation"] || 0),
+            is_current_usage: current_usage,
+            is_pay_in_advance: pay_in_advance?
+          }
+        end
+
+        def accepts_target_wallet?
+          rate_card.wallet_targetable?
```
rsempe, line 101: FYI i will delete this one in a future PR

app/services/fees/charge_service/sources/billing_segment.rb
rsempe: This resolves to the legacy charge link, which is nil for catalog products (`Product belongs_to :charge, optional: true`). `init_metered_items_fees` starts with `metered_item.charge.filters.any?`, and four sites build `charge_id:` from `metered_item.charge.id`. Probably wants `charge_id: nil` plus rate-card-based fee identification rather than borrowing the v1 association.

app/services/fees/charge_service/sources/billing_segment.rb
```diff
@@ -0,0 +1,145 @@
...
+        # NOTE: Product-catalog presentation groups will move to product/plan data once that feature is supported.
+        def presentation_group_keys_values
+          []
+        end
+
+        def matching_filters
+          product_filter&.to_h || {}
+        end
+
+        def ignored_filters
```
rsempe, line 80: Hardcoding this to `[]` risks double billing: it's what excludes events already claimed by sibling filters (`charge_service.rb:501-503`). Fine for a filter-scoped card, but an unscoped card sharing a product with scoped ones would re-aggregate their events.

rsempe (approved):

app/services/fees/charge_service/sources/billing_segment.rb
brunomiguelpinto:
```
if metered_item.dynamic? # ← on this check
  compute_precise_total_amount_cents(options:)
end
```

brunomiguelpinto (commented):
The billing schedule now merged into main also changes the day-counting rule this method should follow.

app/models/billing_segment.rb
```diff
@@ -38,6 +38,10 @@ def rate
     rate_override || rate_card_rate
   end
 
+  def duration_in_days
+    Utils::Datetime.date_diff_with_timezone(started_at, ended_at, customer.applicable_timezone)
+  end
```
brunomiguelpinto, line 54: For a June 1 00:00 → July 1 00:00 cycle in UTC, with a rate change on June 16 at 09:30, this method returns **16 + 15 = 31 days** across the two segments. The helper rounds each segment's elapsed duration up independently.

The merged `Billing::Days` rule assigns each day to the segment containing that day's local midnight. June 16 therefore belongs to the first segment, giving **16 + 14 = 30 days**, 

```ruby
def duration_in_days
  Billing::Days.between(
    started_at,
    BillingSegment.exclusive_end(ended_at),
    timezone: customer.applicable_timezone
  )
end
```

tiagolupepic: thanks! I'll update this

app/services/fees/charge_service/sources/billing_segment.rb
tiagolupepic: I'll address this in another PR, after this PR https://github.com/getlago/lago-api/pull/6296 is merged

app/services/fees/charge_service/sources/billing_segment.rb
tiagolupepic: done!

app/services/fees/charge_service/sources/billing_segment.rb
tiagolupepic: is not going to live in rate_card anymore?

PR #6294, opened 2026-09-04
refactor(events): pass event contexts instead of subscription

tiagolupepic:
## Context

This change unblocks the initial BillingSegment refactor needed to support BillingPeriodFilterService. BillingSegment filtering queries raw events through contract data, while the existing event stores and billable metric aggregators assumed every event context was a Subscription.

## Description

Introduce Events::Stores::EventContext as the explicit object passed into event stores and aggregators. The stores now ask the context for external ids, organization, customer, duration, subscription history, and date calculations instead of depending directly on a subscription.

Contract-backed contexts currently support the raw event lookup surface needed by BillingSegment, and intentionally raise NotImplementedError for subscription-specific methods that are not defined yet: invoice_subscriptions, previous_subscription, previous_subscription_id, previous_subscription_id?, next_subscription, upgraded?, downgraded?, id, and charges_duration_at. Those contract behaviours will be addressed in a separate PR.

This refactor is necessary to unblock products and plans (contract) https://github.com/getlago/lago-api/pull/6296

### DIFF
--- a/app/models/concerns/billing_period_date_diff.rb
+++ b/app/models/concerns/billing_period_date_diff.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module BillingPeriodDateDiff
+  extend ActiveSupport::Concern
+
+  def date_diff_with_timezone(from_datetime, to_datetime)
+    number_of_days = Utils::Datetime.date_diff_with_timezone(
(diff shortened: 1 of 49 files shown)


rsempe (approved):
lgtm!

PR #6295, opened 2026-09-04
feat(usage): make bucket serving observable [ING-589]

vincent-pochet:
> Stacked on #6291, review that one first.

## Context

_Part of the Realtime Usage project._

The served path delegates to the events store silently by design, so a lagging or broken pipeline
looks exactly like usage merely being slow. Aggregation is the cost, so the win equals coverage:
one delegated high-traffic charge keeps a subscription slow however well the rest is served.

## Description

- Count every current usage lookup as served or delegated, tagged with the reason, from the event store provider — the only place that knows both.
- Keep the reasons coarse: eight labels an operator would act on differently, rather than one per branch. A failed ClickHouse read shares a label with the callers that decline the prefetch.
- Report the freshness of the buckets behind a served read, taken as the most recent write: the oldest reports the age of the billing period.
- Memoize the outcome per (charge, filter), so the cache bypass and the aggregation options count one lookup between them.

Metrics are emitted only for organizations the gate is on for, and none carries an organization,
subscription or charge id. Alert rules and the pipeline's liveness loopback live outside this
repository.

### DIFF
--- a/app/services/events/stores/provider.rb
+++ b/app/services/events/stores/provider.rb
@@ -5,6 +5,9 @@ module Events
     # Mints the event store instances used by a single usage or billing computation, and
     # holds the pre-aggregated usage buckets it may serve count and sum from.
     #
+    # It is also the only place that knows whether a (charge, filter) was served or delegated
+    # and why, so it is the only place that reports it.
+    #
     # Scoped to one computation at one window: build one per request or job.
     class Provider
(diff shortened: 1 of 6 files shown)


vincent-pochet:
Closing: ING-589 has been rewritten. Observability of served-versus-delegated does not answer the question we actually need answered before enabling the flag anywhere — whether the bucket path returns the same usage as the events store. That comparison ships first, as a rake task. Metrics can come back as their own ticket once the two paths are known to agree.

PR #6296, opened 2026-09-04
feat(events): filter billing segments

tiagolupepic:
## Context

Billing segment fee computation needs the billing period prefilter to work with product-catalog billing work items and product filters, while preserving the existing charge-backed path.

ChargeFilter treats ALL_FILTER_VALUES as the values configured on the billable metric filter, not any event carrying the key. Product filters must follow the same rule so equivalent configurations do not select different usage across the two engines.

## Description

Add a billing segment resolver and source adapter to the billing period filter service. Key segment-backed filter targets by product, support key-only product filter values in the shared matcher, and cover the new behavior with focused specs plus a charge filter regression.

Expand nil product filter selections to configured metric values while preserving nil in the raw representation. Remove unrestricted nil matching and align overlap resolution with charge filters.

Add regression coverage for rejecting unconfigured, null, and empty event values, and for creating all-values selections with an explicit nil value.


app/services/events/billing_period_filters/billing_segments_resolver.rb
```diff
@@ -0,0 +1,108 @@
...
+        @billable_metric_filter_keys ||= BillableMetricFilter
+          .where(billable_metric_id: target_segments.map { |segment| segment.contract_rate_card.product.billable_metric_id })
+          .distinct
+          .pluck(:key)
+      end
+
+      def event_store
+        @event_store ||= Events::Stores::StoreFactory.new_instance(
+          organization:,
+          subscription: contract,
```
tiagolupepic, line 73: this is wrong and depends on this PR https://github.com/getlago/lago-api/pull/6294

app/services/events/billing_period_filters/charges_resolver.rb
```diff
@@ -2,22 +2,14 @@
...
       def initialize(subscription:, boundaries:, codes: nil, with_last_seen_at: true)
         @subscription = subscription
         @boundaries = boundaries
         @codes = codes
         @with_last_seen_at = with_last_seen_at
       end
 
-      def filter_targets
-        if organization.pre_filter_events?
-          filter_targets_from_pre_enriched_events
```
tiagolupepic, line 15: This is deprecated. I'm removing the pre_filter_events option (code) in this PR and keeping only the column in `organizations` table

app/services/events/billing_period_filters/base_resolver.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+module Events
+  module BillingPeriodFilters
+    class BaseResolver
+      def filter_targets
+        combinations = event_values_with_history do |**options|
+          event_store.distinct_codes_and_property_combinations(filter_keys: billable_metric_filter_keys, **options)
```
brunomiguelpinto, line 8: **[P1] Preserve prefilter support for `ClickhouseEnrichedStore`**

This now always calls `distinct_codes_and_property_combinations`, but `ClickhouseEnrichedStore` does not implement it: it inherits the `BaseStore` implementation returning `[]`. The previous charge resolver used `distinct_charges_and_filters` when `pre_filter_events` was enabled, which this store does support. For organizations using the enriched ClickHouse aggregation path, removing that branch silently drops non-recurring charges from the prefilter and causes customer usage to return zero despite existing events.

Reproduced in Docker with a persisted enriched ClickHouse event, a non-recurring count metric and a standard charge priced at `2`, explicitly selecting `ClickhouseEnrichedStore` through `StoreFactory.with_override`: the old store query finds the event, but this PR returns `filter_targets = {}` and `Invoices::CustomerUsageService` succeeds with 0 events / 0 cents. Restoring only the charge resolver from `main` makes the same scenario return 1 event / 200 cents. The three focused examples give 2 failures with this PR and 0 failures with the `main` resolver.

Could we keep a compatible enriched-store path, or implement `distinct_codes_and_property_combinations` in `ClickhouseEnrichedStore` with the required filtering and timestamp semantics before switching to this shared resolver? A regression spec exercising the real enriched store through the charge prefilter/customer usage would cover this.

brunomiguelpinto (commented):
Confirmed an enriched ClickHouse regression with a focused integration test; details inline.

app/services/events/billing_period_filters/base_resolver.rb
brunomiguelpinto: Not used so no issue

rsempe (approved):

PR #6297, opened 2026-09-04
feat(streaming): store where an organization's events should be streamed

toommz:
First of four stacked PRs adding Kinesis streaming destinations.

## Context

Delivery is driven by config: an organization that wants its events streamed has a destination row, and one that has none gets nothing delivered. This is the table and the models that hold it.

## Description

`streaming_destinations` follows the STI convention already used by `Integrations::BaseIntegration` and `PaymentProviders::BaseProvider`: a base model with `SettingsStorable` and `SecretsStorable`, and a `KinesisDestination` subtype holding `stream_arn`, `region`, `role_arn` and `partition_key` in settings. A role ARN is infrastructure config rather than a credential, since holding it grants nothing unless the workload identity is already allowed to assume it, so it lives in `settings` and `secrets` stays empty for this type.

`event_types` is its own array column rather than a settings key, so it is shared across every destination type and can be indexed. Lookup goes through `for_event`, which queries array containment and is the only form that works: `find_by(event_types: "...")` compares the whole array against one string, and Postgres rejects it outright. There is a spec pinning that, because a lookup that silently matches nothing means delivery never happens and nothing raises.

## The overlap rule is enforced in the model, not the database

Two destinations for one organization must never claim the same event type, or `for_event` becomes ambiguous. Postgres cannot express that constraint on this column: an exclusion constraint needs a GiST operator class for `varchar[] && varchar[]`, which does not exist, and GIN backs neither exclusion constraints nor unique indexes.

```
ERROR:  data type character varying[] has no default operator class for access method "gist"
ERROR:  access method "gin" does not support exclusion constraints
```

So it is a model validation, and `insert_all` or `update_column` can still create an overlap. GIN is still the right index for the containment lookup and is present.

### DIFF
--- a/app/models/streaming_destinations/base_destination.rb
+++ b/app/models/streaming_destinations/base_destination.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module StreamingDestinations
+  class BaseDestination < ApplicationRecord
+    include PaperTrailTraceable
+    include SecretsStorable
+    include SettingsStorable
+
+    self.table_name = "streaming_destinations"
+
+    EVENT_TYPES = %w[customer_usage.refreshed].freeze
+
+    belongs_to :organization
+
+    validates :event_types, presence: true
+    validate :event_types_are_known
+    validate :event_types_not_already_claimed
+
+    scope :for_event, lambda { |organization, event_type|
+      where(organization:).where("event_types @> ARRAY[?]::varchar[]", event_type)
+    }
+
+    private
+
+    def event_types_are_known
+      return if event_types.blank?
+      return if (event_types - EVENT_TYPES).empty?
+
+      errors.add(:event_types, :inclusion)
+    end
+
+    def event_types_not_already_claimed
+      return if organization_id.nil? || event_types.blank?
+
+      claimed = BaseDestination.where(organization_id:)
+      claimed = claimed.where.not(id:) if persisted?
+
+      return unless claimed.where("event_types && ARRAY[?]::varchar[]", event_types).exists?
+
+      errors.add(:event_types, :taken)
+    end
+  end
+end
+
+# == Schema Information
+#
+# Table name: streaming_destinations
+# Database name: primary
+#
+#  id              :uuid             not null, primary key
+#  event_types     :string           default([]), not null, is an Array
+#  secrets         :string
+#  settings        :jsonb            not null
+#  type            :string           not null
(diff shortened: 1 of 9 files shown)


toommz:
> 
> HOLD — event ownership is not enforced atomically, so the lookup is not reliably singular.
> 
> * `event_types_not_already_claimed` is only a pre-write query; concurrent creates or updates can both pass and make `.for_event` return multiple destinations. Enforce the invariant atomically and add regression coverage.
> * Add the missing `already_claimed` validation translation. Record validation failures expose `record.errors.messages`, so this currently produces a missing-translation message.
> * Add coverage for `Organization#streaming_destinations` and the new model's PaperTrail behavior to match the repository's model-spec conventions.

About the `event_types_not_already_claimed`, it won't be a high load table. Organizations won't create lots of streaming destinations at the same time, and they probably will do it once and then let the data flow. So 👎 

About the other parts, 👀

vincent-pochet (approved):

app/models/streaming_destinations/base_destination.rb
```diff
@@ -0,0 +1,67 @@
...
+
+module StreamingDestinations
+  class BaseDestination < ApplicationRecord
+    include PaperTrailTraceable
+    include SecretsStorable
+    include SettingsStorable
+
+    self.table_name = "streaming_destinations"
+
+    EVENT_TYPES = %w[customer_usage.refreshed].freeze
```
tiagolupepic: I'd put versions here already and make it mandatory. like `customer_usage.v1.refreshed`, I checked other PR's and since the idea is to not have schema in this "destination" just pure JSON, it's very easy to someone change the JSON and break the consumers because a key is missing or is wrong (nil).

For example, right now, we have the `current_usage` tied to charge/charge_filter. In the following months we'll have products and plans and the usage will be tied to rate/rate_override and other dependencies.

So, we'll need another stream (destination) to this new usage and the v2 will fit perfectly.

toommz: @tiagolupepic That's a really strong point! I added commits in each PR related to this. I also removed the `schema_version` in the payload envelop as it is now "encoded" in the event name.

tiagolupepic (approved):
:shipit:

app/models/streaming_destinations/base_destination.rb
toommz: Resolving myself as you approved

PR #6298, opened 2026-09-04
feat(streaming): deliver current usage to kinesis on wallet refresh

toommz:
Stacked on #6297.

## Context

A customer needs each of their own customers' current usage within about a minute of their wallet refreshing, so they can gate locally instead of depending on our real-time alerts. HTTP webhooks are the wrong fit: they persist a row per delivery and retry until they succeed, which does not suit an event that can fire every 40 seconds per customer, and a retry can resend a stale snapshot after a newer one already went out.

## Description

Wallet refresh is the trigger. `RefreshWalletsService` enqueues with `perform_after_commit`, gated on the existence of a "streaming destination".

`DeliverEventJob` dispatches through an event-type registry rather than hardcoding the one event that exists today, mirroring how webhooks map event types to services. The delivery path is transport-agnostic in the same spirit: it resolves the destination through `StreamingDestinations::BaseDestination.for_event`, across every destination type, and asks that destination for its own producer and partition key. Nothing outside `KinesisDestination` and `KinesisProducer` names Kinesis, so a second transport is a new subtype rather than a change to the delivery service.

Looking up across types is also what the data model already assumes: `event_types_not_already_claimed` queries the base class, so it rejects a second destination for one organization and event type regardless of type. A Kinesis-only lookup would have contradicted that, and an organization whose destination happened to be another type would have gone silently undelivered while the service still returned success. It runs on a dedicated `streaming` queue: the worker consuming it needs an AWS identity the general workers do not have.

Serialization goes through a dedicated `CustomerUsageSerializer`.

The producer fails fast rather than retrying hard: SDK defaults allow up to 240s per record against a dead endpoint while holding a Sidekiq thread, so timeouts are pinned on the Kinesis and the STS client alike. Assumed credentials are memoized per process per role ARN, since rebuilding them per delivery would turn a 10,000 job burst into 10,000 STS calls.

## Behaviour worth calling out in review

- **Usage goes out on the invoicing path as well as the clock path**, because `RefreshWalletsService` fires on both and the hook does not distinguish them. Narrowing it later changes what the consumer receives.
- **The payload carries `credits` alongside `amount_cents` and `currency`**, converted via `WalletCredit` through the first active wallet in the usage currency, and null when none matches.
- **Nothing is delivered without a destination row.** There is no feature flag, so the gate is a single existence query on every wallet refresh.

### Preconditions, both silent when unmet

- The organization needs the `granular_lifetime_usage` premium integration and a premium license, or every full-usage delivery is refused and the current-period event keeps working.
- Full usage is refused for any subscription whose plan carries a prorated charge. Those subscriptions never receive it, and still receive the current-period event.

## Deliberate, not oversights

- An overlapping delivery for the same customer is dropped, not queued behind: best-effort latest-state, not at-least-once.
- Delivery failures are logged and abandoned rather than retried; the next wallet refresh supersedes them.
- `credits` uses one wallet's rate, so multi-wallet splits and `wallet_targets` are not reflected.
- Full usage aggregates from `subscription.started_at` on every delivery rather than incrementally, and its cache is invalidated by any newer event, so a continuously ingesting customer pays the cold cost each time.


app/serializers/event_destinations/customer_usage_serializer.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module EventDestinations
+  class CustomerUsageSerializer < ModelSerializer
+    def serialize
+      {
+        from_datetime: model.from_datetime,
+        to_datetime: model.to_datetime,
```
tiagolupepic: use the iso8601 format here

app/serializers/event_destinations/customer_usage_serializer.rb
```diff
@@ -0,0 +1,50 @@
...
+        credits: credits,
+        charges_usage: charges_usage
+      }
+    end
+
+    private
+
+    def wallet
+      options[:wallet]
+    end
```
tiagolupepic, line 26: maybe would be a good option to also include the wallet.id, wdyt? 
just to avoid the consumers start wondering why the credits is nil. They'll not suspect if the wallet_id is also nil.

app/services/customers/refresh_wallets_service.rb
```diff
@@ -45,6 +47,16 @@ def call
 
     attr_reader :customer, :include_generating_invoices
 
+    def deliver_current_usage
+      event_type = EventDestinations::CustomerUsage::RefreshedService::EVENT_TYPE
+
+      return unless StreamingDestinations::BaseDestination
+        .for_event(customer.organization, event_type)
+        .exists?
```
tiagolupepic: nitpick.. you can also have some sort of `StreamingDestinations::BaseDestination.customer_usage_refresh_service?(organization)` => true/false

app/services/event_destinations/customer_usage/refreshed_service.rb
```diff
@@ -0,0 +1,104 @@
...
+      def envelope(subscription, usage)
+        {
+          event_id: SecureRandom.uuid_v7,
+          event_type: EVENT_TYPE,
+          object_type: OBJECT_TYPE,
+          organization_id: customer.organization_id,
+          customer_external_id: customer.external_id,
+          subscription_external_id: subscription.external_id,
+          version:,
+          customer_usage: serialized_usage(usage)
```
tiagolupepic, line 89: not sure the correct place or the envelope or only the serialied_usage. But this can be a `binary`, you'll save a lot of data traveling around.

app/serializers/event_destinations/customer_usage_serializer.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module EventDestinations
+  class CustomerUsageSerializer < ModelSerializer
+    def serialize
```
tiagolupepic, line 5: just an idea. Debezium also uses a pure JSON by default and each message contains a schema in JSON, we could evaluate to have the same here

tiagolupepic: one option is to use an abstraction on top of dry-validation and the abstraction also generates the schema for the payload

app/services/event_destinations/kinesis_producer.rb
```diff
@@ -0,0 +1,97 @@
...
+    private
+
+    attr_reader :destination
+
+    def log_prefix
+      "[streaming] destination=#{destination.id} stream=#{destination.stream_arn}"
+    end
+
+    def client
+      CLIENTS.compute_if_absent([destination.role_arn, destination.region]) do
```
tiagolupepic: using concurrent map for this is a good call!

app/models/streaming_destinations/kinesis_destination.rb
```diff
@@ -12,6 +14,19 @@ class KinesisDestination < BaseDestination
     validates :region, presence: true
     validates :role_arn, presence: true
     validates :partition_key, inclusion: {in: PARTITION_KEYS}
+
+    def producer
+      EventDestinations::KinesisProducer.new(destination: self)
+    end
+
+    def partition_key_for(customer:, subscription:)
```
tiagolupepic: are you going to use the subscription here? 
spoiler, we're introducing `contract` which is part of new product and plans

app/services/event_destinations/kinesis_producer.rb
```diff
@@ -0,0 +1,97 @@
...
+          client: Aws::STS::Client.new(region: destination.region, **client_timeouts)
+        )
+      end
+    end
+
+    def client_timeouts
+      {
+        http_open_timeout: HTTP_OPEN_TIMEOUT,
+        http_read_timeout: HTTP_READ_TIMEOUT,
+        retry_limit: RETRY_LIMIT
```
tiagolupepic: **[P2] Pin the retry mode to enforce the fail-fast policy**

`retry_limit` only applies to the AWS SDK's `legacy` retry mode. Both clients still inherit their mode from AWS configuration, so a deployment using `AWS_RETRY_MODE=standard` and `AWS_MAX_ATTEMPTS=20` can make 20 attempts despite `RETRY_LIMIT = 1`. Repeated read timeouts and backoff can then occupy streaming workers well beyond the approximately one-minute delivery objective; adaptive mode can also wait for client-side capacity outside the HTTP timeouts. Explicitly set `retry_mode: "legacy"` alongside this limit, or pin `standard` with `max_attempts: 2`, for both Kinesis and STS. Add coverage with nonlegacy AWS environment settings so the test verifies the effective retry policy rather than only `config.retry_limit`.

app/services/event_destinations/kinesis_producer.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+require "aws-sdk-kinesis"
+
+module EventDestinations
+  class KinesisProducer
```
tiagolupepic: **[P3] Keep the transport adapter outside the service directory or use the service contract**

`AGENTS.md` requires classes under `app/services/` to inherit `BaseService`, end in `Service`, declare a `Result`, and expose an argument-free `call`. This class instead exposes `produce` and returns an AWS response or `nil`, bypassing the shared service middleware as well. If it is intentionally a low-level transport adapter, move it outside `app/services/`; otherwise adopt the documented service contract. Either option can preserve the deliberate log-and-abandon delivery behavior.

tiagolupepic (approved):
:shipit:

PR #6300, opened 2026-09-04
docs(streaming): document destinations and how to create one by hand

toommz:
Stacked on #6299.

## Context

Destinations are provisioned by hand: no rake task, no admin CRUD, no GraphQL, because nobody is asking for self-service yet and a handful of rows a year is manageable. That only works if the procedure is written down. Otherwise it lives in one person's head, and the first person to create a row without that context points a destination at the wrong stream.

That failure is not theoretical. A customer's production and development environments are often two organizations on the same database, and their stream ARNs differ only by an account id. Crossing them sends one environment's usage to the other's consumer, and nothing in the system reports it: both destinations look healthy and both emit `outcome=delivered`.

## Description

`docs/streaming_destinations.md` covers what has to exist before you touch the console, how to create the row, what each setting means, how to verify it before real usage flows, and how delivery actually behaves, including the parts worth knowing before promising anything to a consumer: best-effort latest state rather than at-least-once, the `version` ordering token, and the fact that failures never surface as failed jobs.

Verification is a console one-liner rather than tooling: enqueue `DeliverEventJob`, then read the streaming worker's logs. The doc says to enqueue rather than run the delivery inline, because the job is pinned to the `streaming` queue and only that worker holds an AWS identity allowed to assume the destination role. A delivery run in a console process fails on credentials even when everything else is configured correctly.

## Not in this PR

End-to-end delivery against a real stream is still unproven, and no shard count has been derived from a measured burst.


toommz:
Closing: the in-repo docs are being dropped from this stack. The content lives outside the repo instead.

PR #6304, opened 2026-09-04
feat(catalog): attach the plan ecosystem to catalog plans

rsempe:
## What this is

The shared plan ecosystem — coupon targets, plan taxes, entitlements — must be able to point at a **catalog plan**, not only a legacy `Plan`. Additive: existing rows keep their `plan_id`.

## Changes

Nullable `catalog_plan_id` (+ index + a NOT VALID → validated FK) added to `coupon_targets`, `plans_taxes`, `entitlement_entitlements` — the `rate_phase` XOR pattern, real Postgres FKs.

- **plans_taxes**: `plan_id` relaxed to nullable; unique `(catalog_plan_id, tax_id)` index mirroring the legacy `(plan_id, tax_id)` one (also serves the FK lookup, so no standalone column index). Model validates exactly one of plan / catalog_plan.
- **entitlement_entitlements**: unique partial `(entitlement_feature_id, catalog_plan_id)` index mirroring the plan one; the DB `exactly_one_parent` check swapped from `plan XOR subscription` to `num_nonnulls(plan, catalog_plan, subscription) = 1`, and the model validation counts all three.
- **coupon_targets**: model validates a target points at a legacy **or** a catalog plan, never both (billable-metric targets unaffected).

### DIFF
--- a/app/models/plan/applied_tax.rb
+++ b/app/models/plan/applied_tax.rb
@@ -6,9 +6,23 @@ class Plan
 
     include PaperTrailTraceable
 
-    belongs_to :plan
+    belongs_to :plan, optional: true
+    belongs_to :catalog_plan, optional: true
     belongs_to :tax
     belongs_to :organization
+
+    validate :exactly_one_plan
+
+    private
+
+    # A tax applies to exactly one plan, legacy or catalog.
+    def exactly_one_plan
+      has_plan = plan_id.present? || plan.present?
+      has_catalog_plan = catalog_plan_id.present? || catalog_plan.present?
+      return if has_plan ^ has_catalog_plan
+
+      errors.add(:base, :exactly_one_plan_required)
+    end
   end
 end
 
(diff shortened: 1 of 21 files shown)


app/models/plan/applied_tax.rb
```diff
@@ -6,9 +6,23 @@ class AppliedTax < ApplicationRecord
...
+
+    private
+
+    # A tax applies to exactly one plan, legacy or catalog.
+    def exactly_one_plan
+      has_plan = plan_id.present? || plan.present?
+      has_catalog_plan = catalog_plan_id.present? || catalog_plan.present?
+      return if has_plan ^ has_catalog_plan
+
+      errors.add(:base, :exactly_one_plan_required)
```
tiagolupepic, line 24: [P2] Add translations for the new validation error symbols

Neither `exactly_one_plan_required` nor `single_plan_target` (introduced in `CouponTarget`) is defined in `config/locales/en.yml`. Rails interprets these symbols as translation keys, so invalid parent combinations produce a `Translation missing` message when `errors.messages` is read, rather than the intended error code. `BaseResult#record_validation_failure!` forwards those messages directly. Add both keys under `en.activerecord.errors.messages` and assert the resulting messages in the specs; the current `errors.where` assertions only check error types and miss this.

tiagolupepic (approved):
the PR looks good! just one small comment regarding translations.

:shipit:

endenis (approved):

PR #6314, opened 2026-09-07
feat(events): recover missing pay-in-advance fees

vincent-pochet:
Part of ING-606

## Context

`Events::CreateService` commits the event row, then enqueues `Events::PostProcessJob`. The two are not transactional and the rescue clauses cover only ActiveRecord errors, so when Kafka or Redis is down the row is committed and the job never runs: the event counts in usage but no pay-in-advance fee is created. Reposting the same `transaction_id` returns `value_already_exist` forever, and nothing replays post-processing over persisted Postgres events.

## Description

- Add `events:recover_pay_in_advance_fees`, reporting only unless `DRY_RUN=false`.
- Scope detection to one organization and a `created_at` window, keyed on `(transaction_id, charge_id)` like the fee duplication guard indexes.
- Re-enqueue `Events::PayInAdvanceJob` per recoverable event, oldest first.
- Report, rather than replay, events with no subscription covering their timestamp, a voided fee, or only partly billed charges.
- Do not repair fees created after the gap: those were priced without the missing cached aggregations.

Detection cannot reuse `already_processed?`: its scope filters `invoice_id: nil` and misses invoiced fees. Replay enters at `PayInAdvanceJob` because `PostProcessJob` raises `RecordNotUnique` in `Events::EnrichService` under `postgres_enriched_events`, and `PostProcessService` swallows that as success without creating the fee.

### DIFF
--- a/lib/tasks/events.rake
+++ b/lib/tasks/events.rake
@@ -115,4 +115,249 @@ namespace :events do
   ensure
     Karafka.producer.close if reprocess
   end
+
+  # Recovers pay-in-advance fees for events that were persisted but never post-processed, because
+  # Kafka or Redis was down when they were ingested. Only the fee is recovered.
+  #
+  # Beware: on an invoiceable charge the replay also finalizes an invoice dated at the event
+  # timestamp, emails it and starts a payment. The dry-run report says how many are affected.
+  #
+  # Usage:
+  #   lago exec api bundle exec rails events:recover_pay_in_advance_fees \
+  #     ORGANIZATION_ID=<uuid> FROM=2026-08-22T00:00:00Z TO=2026-08-24T00:00:00Z [DRY_RUN=false]
+  #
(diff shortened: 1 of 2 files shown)


vincent-pochet:
@claude
- `custom_agg` is legacy and not used in production anymore (will be retired completely soon)
- transaction_ids are not re-used on multiple subscriptions

vincent-pochet:
> 
> HOLD — live recovery can create incorrect pay-in-advance fees for stateful aggregations.
> 
>     * Ordering the scan only orders job enqueueing; the parent jobs and their fee/invoice jobs can execute concurrently and out of order. Cached aggregation state is order-sensitive, so a sequence with decrements can be over-billed when later events finish first. Serialize recovery per subscription/charge in the intended event order (or refuse non-count aggregations), and add a regression spec that performs a multi-event job chain with decrements and verifies the resulting fee units and amounts.

We know for a fact that the targeted events are not decreasing the result. (Decrease in a pay in advance scenario is dangerous anyway)

jdenquin (approved):

PR #6315, opened 2026-09-07
feat(catalog): own catalog plans on the v2 API (REST + GraphQL)

rsempe:
## Context

The v2 catalog-plan surface (REST + GraphQL) created and read legacy `Plan` rows filtered to `pricing_type: "product_catalog"`. The product catalog now owns a dedicated `CatalogPlan` table, so both surfaces should operate on it.

## Description

- **REST**: `Api::V2::PlansController` (index/show/create/update) repointed onto `CatalogPlan` via `CatalogPlans::{Create,Update}Service` + `CatalogPlansQuery`, serialized by `V2::CatalogPlanSerializer`. Native `currency` column removes the amount_currency remap. Seed creates a `CatalogPlan`.
- **GraphQL**: `createCatalogPlan`/`updateCatalogPlan` repointed onto the same services and return a dedicated `Types::CatalogPlans::Object` (graphql_name `CatalogPlan`) exposing only the fields a catalog plan carries — `currency`, not `amountCurrency` — so no `Types::Plans::Object` field ever resolves against a `CatalogPlan`.
- **Webhooks + activity logs**: emits `catalog_plan.created`/`catalog_plan.updated` (new `Webhooks::CatalogPlans` services, `V1::CatalogPlanSerializer`, `SendWebhookJob` + `webhook_event_types.yml` + `ACTIVITY_TYPES` entries) and activity logs from the services, with `CatalogPlan` registered in the activity-log `RESOURCE_TYPES` and the `ActivityLogs::ResourceObject` union.

## Deferred to the pricing-rename PR

Rate cards still attach to legacy plans here, so `applied_rate_cards_count` is `0`, the currency lock is off, and the nested `applied_rate_cards` route 404s for a plan created here — deliberately owned by a request spec.


app/models/entitlement/entitlement.rb
tiagolupepic: TIL: didn't know about this option to pass the object in #count method.

app/models/plan/applied_tax.rb
tiagolupepic: this logic is quite similar in `RatePhase#validate_exactly_one_parent` you can extract to a ActiveModel validator if you want

app/models/coupon_target.rb
tiagolupepic: ditto

app/services/catalog_plans/create_service.rb
```diff
@@ -0,0 +1,40 @@
...
+        name: args[:name],
+        code: args[:code],
+        description: args[:description],
+        invoice_display_name: args[:invoice_display_name],
+        currency: args[:currency]
+      )
+      catalog_plan.save!
+
+      result.catalog_plan = catalog_plan
+      SendWebhookJob.perform_after_commit("plan.created", catalog_plan) if send_webhook
```
tiagolupepic, line 30: are we using the same webhook name here? shouldn't be `catalog_plan.created`?

tiagolupepic (approved):
just a couple of comments and they are not a blocker

:shipit:

app/services/catalog_plans/create_service.rb
rsempe: no i think we should keep the same webhook, end users always see `plan.*` `catalog_plan` is the internal model/DB name only

PR #6316, opened 2026-09-07
feat(catalog): rename the pricing tables to catalog_plan

rsempe:
## Context

`plan_rate_cards` and `contracts` only ever pointed at product-catalog plans, which now live in `catalog_plans`. Their `plan_id` referenced the shared `plans` table, so both tables must be repointed. This is a **rename, not a dual parent**: a legacy plan is subscribed through `subscriptions`, never through a contract or a rate card.

## Description

- Two-step migration adds `catalog_plan_id` to both tables (concurrent index, NOT VALID FK then validated), relaxes the old `plan_rate_cards.plan_id` NOT NULL. `plan_id` stays a dead column, dropped in a follow-up.
- `PlanRateCard`/`Contract` drop `belongs_to :plan` for `belongs_to :catalog_plan`; `CatalogPlan` gains `applied_rate_cards` + `contracts` and `attached_to_contracts?`.
- Every internal reader repointed (services, queries, controllers, serializers, GraphQL) while **the public surface stays `plan`/`plan_code`/`plan_id`** — `catalog_plan` is the DB/model name only.
- Rate-card locking moves off subscriptions onto contracts; Part 1's deferrals are restored (real `applied_rate_cards_count`, currency lock, seed rate-card wiring) across REST and GraphQL.

Legacy `Plan` rate-card code paths (`Plan#applied_rate_cards` and its consumers, plus `PlansQuery#with_product_category` / the GraphQL `productCategoryId` argument) are now dead after the rename and retired in this PR.

### DIFF
--- a/app/services/catalog_plans/update_service.rb
+++ b/app/services/catalog_plans/update_service.rb
@@ -19,6 +19,13 @@ module CatalogPlans
     def call
       return result.not_found_failure!(resource: "plan") unless catalog_plan
 
+      # The currency is fixed once rate cards price against it: fees bill in the
+      # card currency and the invoice in the plan currency, so a change would
+      # desync already-attached cards.
+      if params.key?(:currency) && params[:currency] != catalog_plan.currency && catalog_plan.applied_rate_cards.exists?
+        return result.single_validation_failure!(field: :currency, error_code: "not_editable_with_applied_rate_cards")
+      end
+
       catalog_plan.name = params[:name] if params.key?(:name)
       catalog_plan.code = params[:code] if params.key?(:code)
       catalog_plan.description = params[:description] if params.key?(:description)
--- a/app/services/catalog_plans/update_service.rb
+++ b/app/services/catalog_plans/update_service.rb
@@ -19,10 +19,12 @@ module CatalogPlans
     def call
       return result.not_found_failure!(resource: "plan") unless catalog_plan
 
-      # The currency is fixed once rate cards price against it: fees bill in the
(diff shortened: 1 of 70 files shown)


app/services/catalog_plans/update_service.rb
```diff
@@ -19,6 +19,13 @@ def initialize(catalog_plan:, params:, send_webhook: true)
     def call
       return result.not_found_failure!(resource: "plan") unless catalog_plan
 
+      # The currency is fixed once rate cards price against it: fees bill in the
+      # card currency and the invoice in the plan currency, so a change would
+      # desync already-attached cards.
+      if params.key?(:currency) && params[:currency] != catalog_plan.currency && catalog_plan.applied_rate_cards.exists?
```
tiagolupepic: [P1] Preserve the currency lock for contracts with direct rate cards

Checking only `catalog_plan.applied_rate_cards` misses contracts that reference an empty catalog plan and carry directly attached rate cards. Create a pending contract referencing a USD catalog plan, attach a USD card through `ContractRateCards::CreateService`, then update the plan's currency to EUR: this guard allows the update, so `Contract#currency` becomes EUR while the contract's card remains USD. This bypasses the currency invariant enforced when attaching the card. Previously, `Plans::UpdateService` prevented currency assignment once the plan had contracts through `attached_to_subscriptions?`. Also reject currency changes when the catalog plan has contracts, and add coverage for a contract with direct cards but no plan-level cards.

tiagolupepic (approved):
:shipit:

PR #6319, opened 2026-09-07
fix(events): fail closed when the enqueue fails

vincent-pochet:
Part of ING-606

## Context

`Events::CreateService` commits the event row and only then enqueues `Events::PostProcessJob`. The two are not transactional, so when Redis is down the row is committed and the job never runs: no pay-in-advance fee is created, and reposting the same `transaction_id` answers `value_already_exist` forever. `Events::CreateBatchService` has the same shape, and one bulk push failure strands the whole batch.

## Description

- Remove the event, or the whole batch, when its post-processing cannot be enqueued, so the caller's retry succeeds instead of being refused as a duplicate.
- Move the enqueue ahead of the Kafka produce, so a failed enqueue publishes nothing downstream.
- Scope the `ArgumentError` rescue to the timestamp parse: `ActiveJob::SerializationError` is an `ArgumentError`, so an enqueue failure was reported as an invalid timestamp.

Hard-deleted rather than discarded: `index_unique_transaction_id` has no `deleted_at` predicate, so a discarded event keeps refusing the retry. Enqueuing inside the transaction is worse: the job is pushed before the commit, so every event would race its own commit instead of only the rare accepted-then-failed enqueue.

- Aggregation reads `events` directly, so a stranded event still billed pay-in-arrears charges correctly. Rejecting it drops that usage unless the caller retries.
- Narrows the window rather than closing it: a process killed between commit and cleanup still strands the event.

### DIFF
--- a/app/services/events/create_batch_service.rb
+++ b/app/services/events/create_batch_service.rb
@@ -68,8 +68,10 @@ module Events
         return if result.errors.any?
       end
 
-      KafkaProducerService.call!(events: result.events, organization:)
+      # Enqueued before producing to Kafka so that a failed enqueue leaves nothing behind
+      # downstream either.
       enqueue_post_process_jobs if organization.postgres_events_store?
+      KafkaProducerService.call!(events: result.events, organization:)
     end
 
     def bulk_insert_events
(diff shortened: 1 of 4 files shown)


toommz (approved):
Nice catch 👏

lovrocolic (approved):

PR #6322, opened 2026-09-07
fix(billing): skip no-op fixed charge unit updates

annvelents:
## Context

One of our organizations creates a subscription with override of units on fixed charge, and immediately sends another request, updating the subscription to the same number of units. 
both of these requests issue a fixed_charge_event and start pay in advance invoicing
we do have uniqueness on PayInAdvanceFixedChargeJob, but it's on combination of subscription + timestamp, while the timestamps are slightly different, so uniqueness allows both jobs to be executed.
The jobs start to be processed in parallel and don't see each other's fees when the guard checks against double billing, as result both of them succeed

## Details

**Current change** - do not issue the same fixed charge event if units are not changed from the previous fixed_charge_event by timestamp (not created_at). Looked in the cases in the DB and it will save us from most of the cases, while later we can work on a better solution

Two constraints that shaped it:

- Compare at the event's timestamp, not now.  Deferred changes are stamped at `next_billing_period`. With units at 100 and a deferred change to 150 pending, a deferred update back to 100 must still emit — comparing against current units would suppress it and wrongly bill 150 next period.
- **No prior event always emits.** At activation the first event establishes the baseline; suppressing it would leave the subscription with no events at all, aggregating to 0 and never billing.

Skipping events is safe for billing: both fixed-charge aggregations carry forward the last event *before* the period (`FixedChargeEvents::Aggregations::BaseService#events_in_range`), so a period containing zero events still bills from the last known units. Gaps are by design — a subscription whose seat count never changes emits no events for years and must still bill every period.

## Scope

This fixes the no-op trigger only. The underlying race — two concurrent `CreatePayInAdvanceFixedChargesJob` runs each reading a stale "already paid" total under READ COMMITTED — is **not** addressed here and is logged as another ticket, which will be addressed later

Classifying the over-billed periods in the affected org over 90 days, by whether the duplicated events carried the same units:

| Category | Periods | Excess units |
|---|---|---|
| No-op duplicate events — fixed here | 123 | 366 |
| Genuinely different units — BIL-661 | 5 | 35 |

~96% of observed over-billing, and the entire reported incident, is the no-op case.

### DIFF
--- a/app/services/fixed_charges/emit_events_service.rb
+++ b/app/services/fixed_charges/emit_events_service.rb
@@ -4,6 +4,8 @@ module FixedCharges
   class EmitEventsService < BaseService
     Result = BaseResult[:fixed_charge_events]
 
+    BATCH_SIZE = 1_000
+
     def initialize(fixed_charge:, subscription: nil, apply_units_immediately: false, timestamp: Time.current.to_i)
       @fixed_charge = fixed_charge
       @subscription = subscription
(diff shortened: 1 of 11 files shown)


tiagolupepic (approved):
:shipit:

PR #6329, opened 2026-09-08
fix(taxes): group charge fees in provider calls

annvelents:
## Context

Grouping fees by charge_id before sending them to Anrok and Alavara

## Description

Anrok and Avalara reject tax payloads above 1200 line items. A charge split by charge filters or by `grouped_by` produces one fee per combination, so a single charge can occupy dozens of those items and an invoice with heavily filtered charges fails tax computation outright. Dropping zero-amount fees (#6289) was not enough on its own

## Changes

- **One line item per charge**, on every path that reports fees to a provider, for both Anrok and Avalara, invoices and credit notes. `Taxes::Invoices::ChargeFeeGroup` stands in for a `Fee` and sums what its members report, so neither payload builder needs a notion of grouping. Non-charge fees still travel individually, and a charge with a single taxable fee keeps its own line so its identifier at the provider is unchanged.
- **The response is expanded back to one result per fee** in `Taxes::Invoices::BaseService#process_response`, so all four places that match a response entry to a fee are untouched. Every charge fee's result carries its `charge_id`, grouped or not.
- **One allocation rule.** `Taxes::Allocation` splits a provider figure over shares by weight, settling the difference on the last share. `ChargeFeeGroup` uses it per jurisdiction across a charge's fees; the aggregator uses it per jurisdiction within a solo line. So every fee books an allocation of what the provider returned, and `Invoices::ApplyProviderTaxesService` just sums what the fees hold — no branch on how a fee was reported.
- **Fees are sorted before grouping** — `Invoice#fees` has no `ORDER BY`, so finalization and the later void refund would otherwise post the same charge's lines in a different sequence, and the fee standing in for an untaxable invoice would not be stable.

## Known limitation

`Fees::ApplyProviderTaxesService` derives `taxes_base_rate` from `tax_breakdown.first` and applies it to every breakdown item, so a reduction the provider applied to one jurisdiction is smeared across the others. That is pre-existing and still affects the precise columns; it deserves its own change, since fixing it moves tax for every provider-taxed customer.

## Tests

New specs cover the allocation (including a jurisdiction amount that does not divide evenly, special taxation types, and coupons cancelling a group), grouping across two subscriptions, payload line ordering, and the credit-note payloads.


annvelents:
Addressing both pre-review comments — no code change needed for either.

**On the `Data` initializers (`sha=85c6e881c`).** Bare `super` in Ruby forwards the method's bound parameter values, including ones that took their default — not only the caller's original keywords. This is the documented `Data` idiom:

```ruby
Measure = Data.define(:amount, :unit) do
  def initialize(amount:, unit: "meters") = super
end
```

Verified against the real class rather than a reproduction:

```
$ ruby -e 'require_relative "app/services/integrations/aggregator/taxes/tax_result.rb"
T = Integrations::Aggregator::Taxes::TaxResult
p T.new(item_key: "k", item_id: "i", item_code: "c", amount_cents: 100, tax_amount_cents: 10, tax_breakdown: [])'

#<data TaxResult item_key="k", item_id="i", item_code="c", amount_cents=100, tax_amount_cents=10, tax_breakdown=[], charge_id=nil>
```

No `ArgumentError`. The comment was also written against `85c6e881c`, which still carried `charge_tax_amount_cents` and `allocated_amount_cents`; both have since been removed, so `TaxResult` has a single optional member and `TaxBreakdownItem` no longer overrides `initialize` at all.

**On the grouped reconciliation (`sha=2fbca7e9`).** `Fees::ReconcileGroupedProviderTaxesService` has been deleted. Each fee now books an allocation of the figure the provider returned for its line, so there is no compute-then-patch step for a residual cent to land in the wrong place. On the example given — a `1.4` + `5.2` breakdown against a 7¢ total — the allocation yields `[1, 6]`, which is the remainder-based result the comment asked for.

endenis (approved):

vincent-pochet (approved):

app/services/fees/apply_provider_taxes_service.rb
```diff
@@ -35,21 +44,19 @@ def call
         )
         fee.applied_taxes << applied_tax
 
-        tax_amount_cents = (fee.sub_total_excluding_taxes_amount_cents * taxes_base_rate * tax_rate).fdiv(100)
         tax_precise_amount_cents = (fee.sub_total_excluding_taxes_precise_amount_cents * taxes_base_rate * tax_rate).fdiv(100.to_d)
 
-        applied_tax.amount_cents = tax_amount_cents.round
+        applied_tax.amount_cents = amounts[index]
```
osmarluz: `amount_cents` now comes from the provider allocation, but `precise_amount_cents` on the next line is still derived from `taxes_base_rate(tax_breakdown.first)`. On a multi-jurisdiction fee the two columns no longer agree.

That matters downstream: `EInvoices::Invoices::Common#taxes` builds the Factur-X/UBL tax totals from `taxes_precise_amount_cents`, so the e-invoice would report a different tax total than the invoice itself and than the amount actually charged.

The "known limitation" note in the PR description covers the base-rate smear itself, but not this new divergence between the two columns.

app/services/integrations/aggregator/taxes/invoices/base_service.rb
```diff
@@ -28,6 +28,48 @@ def taxable_fees
             @taxable_fees ||= fees.select(&:taxable?).presence || Array(fees.first)
```
osmarluz: The PR description says fees are sorted before grouping, but no ordering was actually added here and `Invoice#fees` has no `ORDER BY`.

Two consequences: the `Array(fees.first)` stand-in picks an arbitrary fee to represent the invoice, and it can be a different one on the finalize call than on the void-refund report; and the `Allocation` residue tie-break becomes non-deterministic between calls on the same invoice.

The credit-note payload added in this PR does order its items (`credit_note.items.order(created_at: :asc)`) — the invoice path should match.

app/services/integrations/aggregator/taxes/credit_notes/payloads/base_payload.rb
```diff
@@ -0,0 +1,46 @@
...
+            #       providers accept.
+            def charge_grouped_items
+              ChargeGroup.by_charge(credit_note.items.order(created_at: :asc)) do |item|
+                item.fee.charge_id if item.fee.charge?
+              end
+            end
+
+            # NOTE: Fee#item_id is the billable metric for a charge fee, so two charges over the
+            #       same metric would share one identifier within a payload.
+            def cn_item_id(items)
```
osmarluz, line 24: `cn_item_id` returns `charge_id` for every charge fee, including single-item credit notes that are never grouped. So the identifier sent to the provider changes for those too, not only for the grouped case the PR describes as the change.

If Anrok or Avalara key line history on this identifier, that is a silent change in behaviour for credit notes that this PR otherwise leaves alone. Worth confirming it is intended — and if it is, the PR description should say so.

osmarluz (commented):
Reviewed the full diff. The core design holds up: every consumer of `Taxes::Result#fees` is correctly fed by `split_group_taxes` (`ComputeAmountsFromFees` matches on `item_id == fee.id`, `PreviewService`/`CustomerUsageService` on `item_key`, `ApplyProviderTaxesToStandaloneFeesService` on `item_key` with an `item_id` fallback, and the `FetchDraftInvoiceTaxes` mutation returns them raw). `ChargeFeeGroup` implements everything the payloads call on a `Fee`, `Allocation` is sound for positive, negative, fractional and zero totals, and the credit-note response is never consumed for amounts, so grouping there is safe. Touched specs are green (556 examples, 0 failures).

Three things worth addressing before merge, inline.

app/services/integrations/aggregator/taxes/invoices/base_service.rb
annvelents: added sorting

PR #6331, opened 2026-09-08
fix(catalog): pass seed plan attributes as a positional hash

rsempe:
## Context

The product-catalog seed (`db/seeds/30_product_catalog.rb`) calls `CatalogPlans::CreateService` with **keyword** arguments, but the service takes its attributes as a **single positional hash** (it mirrors the controller's permitted params: `initialize(args, send_webhook: true)`). `BaseService.call!` forwards to `new(*, **)`, so keywords reach `initialize` with zero positional args and Ruby 3 raises `ArgumentError: wrong number of arguments (given 0, expected 1)`.

## Fix

Wrap the attributes in an explicit hash so they arrive as the positional argument the service expects, with a comment noting the convention.


domenicofalco (approved):

tiagolupepic (approved):
:shipit:

PR #6332, opened 2026-09-08
feat(account-tree): add table and model setup for usage_attribution_types

lovrocolic:
## Context

First piece of the account tree config foundation: an organization declares
once which dimensions usage can be attributed to below a customer (user, team,
department, api_key, model). Instances are not declared here - they are
discovered lazily from events in a later step.

## Description

This PR adds new `usage_attribution_types` table and  `UsageAttributionType` model with required validations.

### DIFF
--- a/app/models/usage_attribution_type.rb
+++ b/app/models/usage_attribution_type.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+class UsageAttributionType < ApplicationRecord
+  include Discard::Model
+
+  self.discard_column = :deleted_at
+
+  ROLES = {
+    hierarchical: "hierarchical",
+    flat: "flat"
+  }.freeze
+
+  belongs_to :organization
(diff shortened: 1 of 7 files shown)


app/models/usage_attribution_type.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    while node && visited.exclude?(node)
+      return true if node.id == id
+
+      visited << node
+      node = node.parent
+    end
+
+    false
+  end
```
toommz, line 53: I don't know the intended metrics for this table, if we have short chains and not too much new additions to the table it should be fine. My hunch is it won't be a high-churn table and so it is ok.

toommz (approved):

app/models/usage_attribution_type.rb
lovrocolic: yeah I guess it should be fine. For example the tree for one organization could be: department -> team -> users. I guess there won't be more than few levels in totals.

PR #6333, opened 2026-09-08
feat(catalog): expose catalogPlan and catalogPlans on GraphQL

rsempe:
## Context

The v2 catalog plan surface already had the `CatalogPlan` GraphQL type plus `createCatalogPlan` / `updateCatalogPlan` mutations, but **no query resolvers** — the legacy `PlansResolver` never exposed a product-catalog view, so catalog plans could not be read through GraphQL. This closes that gap (the dedicated v2 plan surface).

## Description

Add two catalog-gated query resolvers backed by the `CatalogPlan` model, mirroring the legacy plan resolvers and the `RequiresProductCatalog` doctrine (same `Contract` vs `Subscription` split):

- **`catalogPlans`** (collection) — reuses `CatalogPlansQuery` (name/code search, pagination, `applied_rate_cards` counts preloaded), returns `CatalogPlanCollection`.
- **`catalogPlan`** (single) — scoped to `current_organization.catalog_plans`, returns a not-found `plan` error when missing.

Both require `plans:view` and the `product_catalog` feature flag; the legacy `Plan` type/resolvers are untouched.

### DIFF
--- a/app/graphql/resolvers/catalog_plan_resolver.rb
+++ b/app/graphql/resolvers/catalog_plan_resolver.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Resolvers
+  class CatalogPlanResolver < Resolvers::BaseResolver
+    include RequiresProductCatalog
+    include AuthenticableApiUser
+    include RequiredOrganization
+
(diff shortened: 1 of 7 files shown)


tiagolupepic (approved):
:shipit:

PR #6334, opened 2026-09-08
feat(contracts): add billing settings

tiagolupepic:
## Context

Billing segment processing needs contract-level billing settings and access to the invoices created for a contract.

## Description

Add optional billing entity and payment method associations, and expose distinct invoices through billing segments. Add normalized purchase order numbers and invoice consolidation enabled by default. Include indexed references, foreign key validation migrations, and model specs.


app/models/contract.rb
```diff
@@ -71,6 +78,14 @@ def currency
     catalog_plan&.currency || customer.currency || organization.default_currency
   end
 
+  def billing_entity
```
rsempe: Overriding the association reader desyncs it from the column: `contract.billing_entity.id` won't match `contract.billing_entity_id` when no override is set, `includes(:billing_entity)` preloads nil and then N+1s through the customer, and `joins(:billing_entity)` silently drops every contract without an override.

Since `applicable_billing_entity_id` already exists, an `applicable_billing_entity` counterpart would keep the association plain and the fallback explicit at call sites.

rsempe (approved):

PR #6340, opened 2026-09-08
feat(catalog): update a pending contract

rsempe:
## Context

A contract is authored while **pending**: its window, plan and billing settings can still be corrected before it goes live. Once `active` the agreement is locked — the same rule its attached rate cards already follow. Until now only creation and reads were exposed (BIL-640).

## Description

Adds `Contracts::UpdateService`, exposed over both surfaces:

- **REST** — `PUT /api/v2/contracts/:external_id`
- **GraphQL** — `updateContract` mutation

Editable fields: `name`, `started_at`, `ended_at`, `billing_anchor_date`, `billing_time`, `plan_code`.

- Changing `plan_code` re-materialises the new plan's rate cards onto the contract (discards the old applied rate cards, re-runs `Contracts::MaterializeRateCardsService`); clearing it leaves a plan-less contract.
- Dates are read through the customer timezone, mirroring create; malformed formats are rejected explicitly and an already-past `ended_at` is refused.
- Editing a non-pending contract fails with `contract_locked`.

**No billing side-effects.** Out of scope: status/lifecycle mutations (`terminateContract` et al.) — those live in their own services.

## Testing

- `spec/services/contracts/update_service_spec.rb`
- `spec/requests/api/v2/contracts_controller_spec.rb` (PUT block)
- `spec/graphql/mutations/contracts/update_spec.rb`
- Schema dump (`schema.graphql`/`schema.json`) regenerated for the new mutation; `lago_api_schema_spec` green.

### DIFF
--- a/app/services/contracts/update_service.rb
+++ b/app/services/contracts/update_service.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+module Contracts
+  # Edits a pending contract's authoring fields. Pending-only: once the
+  # agreement is active it is locked, the same rule rate-card edits follow.
+  # Changing the plan re-materialises its rate cards onto the contract. No
+  # billing side-effects — lifecycle transitions live in their own services.
+  class UpdateService < BaseService
+    include CustomerTimezone
+
+    Result = BaseResult[:contract]
+
+    def initialize(contract:, params:)
+      @contract = contract
+      @params = params.to_h.with_indifferent_access
+      super
+    end
+
+    def call
+      return result.not_found_failure!(resource: "contract") unless contract
+      return result.single_validation_failure!(field: :contract, error_code: "contract_locked") unless contract.editable?
+
+      if params[:plan_code].present? && catalog_plan.nil?
+        return result.not_found_failure!(resource: "plan")
+      end
+
+      # Date columns: a malformed value would silently cast to nil instead of
+      # failing, so formats are rejected explicitly.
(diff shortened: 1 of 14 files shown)


app/services/contracts/update_service.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      if params[:plan_code].present? && catalog_plan.nil?
+        return result.not_found_failure!(resource: "plan")
+      end
+
+      # Reject malformed dates explicitly — they would otherwise cast to nil.
+      %i[billing_anchor_date started_at ended_at].each do |field|
+        if params[field].present? && !Utils::Datetime.valid_format?(params[field])
+          return result.single_validation_failure!(field:, error_code: "value_is_invalid")
+        end
```
tiagolupepic, line 39: [P2] Reject boolean dates instead of silently clearing the end date

REST strong parameters permit `false`, but `false.present?` is false, so `{"contract":{"ended_at":false}}` bypasses both date validation and the past-end check. The assignment below still runs because the key exists, and the `ended_at` reader converts `false` to `nil` through `.presence`. A pending contract with a future end therefore becomes open-ended and the request succeeds, rather than rejecting the malformed date as promised. Validate explicitly supplied non-null date values, preserving intentional `null` clearing, and add a regression test asserting that `false` returns a validation error without changing the existing end.

tiagolupepic (approved):
:shipit:

PR #6344, opened 2026-09-09
[ING-466] feat(wallets): accept connections param

mariohd:
## Context

ING-466. Billing objects need to route themselves to one of a customer's connections. The override table, the resolution cascade and the connection codes already shipped; nothing wrote the rows yet. This is the first write path (dive-in T18c + T18d), and it is additive — nothing reads `effective_*_connection` in production yet.

## Description

`POST/PATCH /wallets` and each nested `recurring_transaction_rules` entry accept a `connections` object keyed by category:

```json
"connections": {
  "payment":    { "behavior": "skip" },
  "accounting": { "code": "netsuite_sub2" },
  "crm":        { "behavior": "inherit" }
}
```

A `code` pins that connection, `skip` short-circuits the category so it never falls back, `inherit` drops the override, and an omitted category changes nothing. `inherit` is params-only: resolution reads a missing row as inheritance, so it destroys the row.

Validation and persistence live in two new services, polymorphic in the owner, so ING-462 and the one-off/preview surfaces can reuse them.

Behind the `multi_connection` flag. Connections sent while the flag is off are refused rather than ignored; an unresolvable code is a 422. The existing `payment_method` is untouched. Serializer exposure and OpenAPI are separate dive-in items.

**Reviewer note:** a rule matched by `lago_id` keeps its overrides; a rule sent without one is replaced, so the new rule gets fresh rows while the superseded rule keeps its own — rules are soft-terminated, so `dependent: :destroy` never fires. Both cases are asserted in the specs.

## Testing

53 new examples: unit, request (running against both `/wallets` and `/customers/:external_id/wallets`), and the reconciliation cases above. Regression across the wallet suites (640) plus the non-REST callers — GraphQL mutations, `Orders::SubscriptionCreation::ExecuteService`, wallet metadata (115) — all green. Rubocop clean.

### DIFF
--- a/app/services/wallets/create_service.rb
+++ b/app/services/wallets/create_service.rb
@@ -20,6 +20,7 @@ module Wallets
       result.payment_method = payment_method
 
       return result unless valid?
+      return result.forbidden_failure! if connections_requested? && organization_flag_disabled?(:multi_connection)
 
       code = params[:code]
 
@@ -86,6 +87,10 @@ module Wallets
 
         if params[:invoice_custom_section].present?
           InvoiceCustomSections::AttachToResourceService.call(resource: wallet, params:)
+        end
+
+        if connections_requested?
+          BillingObjectConnections::AttachToResourceService.call!(resource: wallet, params:)
         end
 
         billable_metrics.each do |bm|
@@ -161,6 +166,16 @@ module Wallets
 
     def organization_flag_enabled?(flag)
       customer.organization.feature_flag_enabled?(flag)
+    end
+
+    def organization_flag_disabled?(flag)
+      !organization_flag_enabled?(flag)
+    end
+
+    def connections_requested?
+      return true if params[:connections].present?
+
+      Array(params[:recurring_transaction_rules]).any? { |rule| rule[:connections].present? }
     end
 
     def valid?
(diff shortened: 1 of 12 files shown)


app/services/wallets/recurring_transaction_rules/create_service.rb
lovrocolic: In case of `FailedResult` (from AttachToResourceService) we are just returning result from this layer and `AttachToResourceService` has completely different `result` object with attached error which is swallowed.

app/services/wallets/create_service.rb
```diff
@@ -20,6 +20,7 @@ def call
       result.payment_method = payment_method
 
       return result unless valid?
+      return result.forbidden_failure! if connections_requested? && organization_flag_disabled?(:multi_connection)
```
lovrocolic, line 22: can we maybe move feature flag check before other validations?

lovrocolic:
I think we also miss few small changes since this PR is covering `billing_object_connections` configuration:
- `has_many :billing_object_connections, foreign_key: :integration_customer_id, dependent: :destroy` in /integration_customers/base_customer.rb
- remove `billing_object_connections` on discarding payment provider customer

wdyt?

lovrocolic (approved):

PR #6348, opened 2026-09-09
feat: produce due billing segments

brunomiguelpinto:
`main` has the billing calendar and it has `BillingSegments::ProcessService`, which turns stored segments into invoices. Nothing created the rows in between. This adds that.

Four single-purpose units and an orchestrator that composes them:

| | |
| --- | --- |
| `ContractRateCard.due_for_billing` | which cards owe something at a given instant |
| `BillingSegments::MissingBillableSegmentsService` | of what they owe, what is not stored yet |
| `BillingSegments::CreateService` | writes those slices as rows, in one statement |
| `ContractRateCards::AdvanceBillingClockService` | moves the clock, never backwards |
| `BillingSegments::ScheduleService` | one customer, one lock, one transaction |

Two things worth knowing before reading the diff.

**What is already stored is decided by overlap, not by an equal start.** A rate added inside a billed cycle re-splits it and gives the later piece the rate's own date, so a start-only test calls it new — which trips the exclusion constraint and stops that customer billing on every later tick. This branch had that bug.

**An exhausted schedule blanks the clock**, so the column, the GraphQL field and the REST payload become nullable. Presence is still validated on create, so null means finished and never malformed.

**Merging this invoices nobody.** `ProcessService` has no trigger on `main` and this does not add one; the rows sit `pending` until someone wires it. Before that happens, settle the non-prorated double charge described in the comments below.


app/services/billing/rate_cards/schedule.rb
```diff
@@ -16,11 +16,26 @@ def initialize(rates:, terms:, phases:, starts_at:, anchor_date:, timezone:, end
...
-          segment.billing_at <= timestamp
+        cycles.flat_map do |cycle|
+          segments = billable_segments_of([cycle])
+
+          if billing_from
+            # A saved arrears clock can point to the original cycle end even
+            # after termination shortens it. Select against the unshortened
+            # pricing windows, then persist the actual service boundaries.
+            full_cycle = cycle.with(ended_at: cycle.calendar.interval_containing(cycle.started_at).end)
+            eligible_starts = billable_segments_of([full_cycle]).filter_map do |segment|
```
tiagolupepic, line 30: convert this to a Set `.to_set` to avoid O(MxN) in the line 35

app/services/billing_segments/schedule_service.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+module BillingSegments
+  # Persists the calendar's due slices. The clock is both the lower billing
+  # bound (backdated contracts can join the current period) and the next wake-up.
+  class ScheduleService < BaseService
+    Result = BaseResult[:billing_segments]
+
+    def initialize(customer:, timestamp: Time.current)
```
tiagolupepic, line 9: better to use a range here, having a single timestamp we don't cover the option to generate segments for past dates/segments.

app/services/billing_segments/schedule_service.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      # nil means the schedule is exhausted, so this card is no longer due.
+      card.update!(next_billing_at: schedule.next_billing_at(after: timestamp))
+      scheduled
+    end
+
+    def persist_segment(card, segment)
+      # Resuming replays the last cycle, which can contain several segments.
+      # Keep existing snapshots and statuses intact, including on a stale-clock retry.
+      if card.billing_segments.exists?(started_at: segment.started_at)
```
tiagolupepic, line 64: Could we handle the known billing-segment overlap constraint as a structured service failure?

persist_segment only skips existing rows with the same started_at. If an existing segment overlaps the candidate but starts at a different instant, create! raises ActiveRecord::StatementInvalid wrapping PG::ExclusionViolation. Neither rescue in call handles that exception.

The transaction still rolls back correctly, but the caller receives an exception instead of a failed Result. The legacy `BillingCycles::ScheduleService` explicitly handles its known period conflicts and returns `overlapping_periods`.

Could we rescue the specific billing_segments_no_overlapping_periods constraint violation, return a structured failure, and add a regression for a different-start overlap? Unrelated database errors should continue to propagate.

app/services/billing_segments/schedule_service.rb
```diff
@@ -0,0 +1,100 @@
...
+        return
+      end
+
+      card.billing_segments.create!(
```
tiagolupepic, line 68: I think we can use a import! here and hit the DB into a single shot. Using batches as well

app/services/billing_segments/schedule_service.rb
```diff
@@ -0,0 +1,100 @@
...
+        organization: customer.organization,
+        customer:,
+        contract: card.contract,
+        cycle_started_at: segment.cycle_started_at,
+        started_at: segment.started_at,
+        ended_at: BillingSegment.inclusive_end(segment.ended_at),
+        billing_at: segment.billing_at,
+        rate_card_rate: segment.rate,
+        rate_override: segment.rate_override,
+        rate_properties: (segment.rate_override || segment.rate).properties,
```
tiagolupepic, line 78: move this to segment.

```
segment.properties
```

app/services/billing_segments/schedule_service.rb
```diff
@@ -0,0 +1,100 @@
...
+        currency: card.rate_card.currency,
+        pricing_unit: pricing_unit(card),
```
tiagolupepic, line 80: `pricing_unit(card)` runs `customer.organization.pricing_units.find_by(code:)` once per persisted segment, but the result depends only on the card, not the segment. On a catch-up run that persists several segments for the same card (e.g. the delayed-run and phase-transition cases), this fires the same query repeatedly. Consider resolving the pricing unit once per card in `schedule_card` and passing it into `persist_segment`, so the lookup (and its `pricing_unit_not_found` failure) happens a single time per card.

app/services/billing_segments/schedule_service.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    private
+
+    attr_reader :customer, :timestamp
+
+    def due_cards
+      ContractRateCard.joins(:contract)
+        .where(organization_id: customer.organization_id, next_billing_at: ..timestamp)
+        .where(contracts: {customer_id: customer.id, status: %w[active terminated], started_at: ..timestamp})
+        .includes(:rate_card, contract: :customer)
```
tiagolupepic, line 47: isn't missing to load the rate_phases here?

app/services/billing_segments/schedule_service.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    def schedule_card(card)
+      schedule = Billing::RateCards::BuildScheduleService.call!(contract_rate_card: card).schedule
```
tiagolupepic, line 51: there is a small issue here. If a customer has multiple contract rate cards with configured rate_card in each of them plus rate card rates. But a single one does not have a single rate_card_rates, the entire schedule will fail due this error https://github.com/getlago/lago-api/blob/billing-segments-scheduler/app/services/billing/rate_cards/build_schedule_service.rb#L22

Basically, it's all or nothing for a single customer. We should evaluate this to ignore that missing one

app/services/billing_segments/schedule_service.rb
```diff
@@ -0,0 +1,100 @@
...
+      result.record_validation_failure!(record: error.record)
+    rescue BaseService::FailedResult => error
+      result.fail_with_error!(error)
+    end
+
+    private
+
+    attr_reader :customer, :timestamp
+
+    def due_cards
```
tiagolupepic, line 43: From AI:

```
The scheduler preloads the rate card and contract/customer, but the builder additionally loads rates, phases, overrides, and potentially plan-applied cards. Candidate segments each trigger an existence check; pricing-unit lookup repeats for each new segment.

With C cards, P phases per card, and S candidate segments, these paths can generate O(CP + S) read queries alongside required writes.

Consider preloading the builder's association graph, fetching existing starts in bulk, and resolving pricing units once per distinct code.
```

app/services/billing/rate_cards/schedule.rb
brunomiguelpinto: Moot now: this line was part of the \`billing_from:\` bound added to \`Schedule\`, which the rebuild removed — \`schedule.rb\` is identical to \`main\` again. The equivalent lookup lives in \`BillingSegments::MissingSlicesService\` and does use a \`Set\`.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Not taken, and I want to give the reason rather than just close it.

The range in \`BillingCycles::ScheduleService\` exists so that "backdated advance subscriptions bill the cycle due at subscription start, not a one-day today window". This calendar has no such window: \`walk_to(timestamp, from: nil)\` calls \`resume(nil)\`, which starts at the card's \`effective_date\`, so a backdated card already produces every cycle from its start. The lower bound is structural — \`starts_at\`, or \`resume_at\` from the stored segments — not a parameter, and passing a past \`timestamp\` works today.

If you still want an explicit window I will add it, but I did not want to add a parameter that duplicates a bound the walker already has.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Done. \`ScheduleService\` rescues \`ActiveRecord::StatementInvalid\`, checks \`error.cause&.message || error.message\` for \`billing_segments_no_overlapping_periods\`, and returns \`single_validation_failure!(error_code: "overlapping_periods", field: :billing_segment)\`. Anything else re-raises. Regression covers a stored segment overlapping the candidate with a *different* start, and asserts nothing is written.

Same-start replays are not this path: they are expected, and \`MissingSlicesService\` filters them out before the write.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Deferred rather than refused, and with numbers.

A customer holds few cards, and the measured cost is 9 queries per additional card, of which 2 are the inserts themselves. The scale here lives in the fan-out — one job per customer — not in bulk-writing one customer.

Batching would also mean giving up the model validations on the write path. That is a smaller loss than it looks, because the table already has CHECK constraints mirroring all of them except the currency inclusion, so I am happy to switch to \`insert_all(unique_by: [:contract_rate_card_id, :started_at])\` when there is a measurement that asks for it — it would let Postgres absorb the replay filter too.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Done — \`Billing::BillableSegment#properties\` returns \`(rate_override || rate).properties\`, and the writer just reads \`billable_segment.properties\`.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Done. The query runs once per customer — `pricing_units_by_code` memoizes `organization.pricing_units.index_by(&:code)` — and the card's unit is resolved from that hash once per card, outside the write loop. The writer takes it as an argument and queries nothing.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Done: `.includes(:rate_card, {rate_phases: :rate_override}, contract: [:customer, {catalog_plan: {applied_rate_cards: :rate_phases}}])`.

Two things measuring turned up. `:organization` was missing, so I added it — and then the batch write made it dead, because the row takes `organization_id` rather than the association and nothing else reads it. It is gone again, one query lighter per run.

The one preloading cannot fix is `RateCard#ordered_rates`: it is `rates.order(:effective_from)`, and `.order` on a loaded association issues a fresh query, so the rates lookup stays. Changing that means changing `RateCard`, which is outside this PR.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Done. \`ScheduleService\` calls the builder non-bang and skips a card whose failure is \`rate\` not found, so one unpriced card no longer takes the customer's whole run with it. Its clock is left untouched, so it is picked up again once it has a rate.

Deliberately narrow: only that failure is skipped. \`invalid_billing_schedule\` still propagates, because a genuinely broken schedule should be loud rather than silently never billed. Two specs cover it — the priced card still bills, and the unpriced card's clock does not move.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Measured rather than argued: 9 queries per additional card, 2 of them the inserts.

- Preloading the builder's graph: done, and \`:organization\` was missing and cost one query per card.
- Existing starts in bulk: done — \`MissingSlicesService\` issues one \`WHERE started_at IN (…)\` per card instead of an existence check per candidate.
- Pricing units once per code: done, memoized once per customer.

Two remain and both need changes outside this PR: \`RateCard#ordered_rates\` re-queries because it calls \`.order\` on the association, and \`card.billing_segments.create!\` reloads the card once per segment because \`belongs_to :contract_rate_card\` carries a \`with_discarded\` scope, so Rails cannot validate presence from the in-memory object.

brunomiguelpinto:
Answering the pre-review HOLD, since it survives the rebuild and is the one finding I did not make go away.

The verdict is correct: \`due_for_billing\` selects on the saved \`next_billing_at\`, so a schedule change that makes a slice due *earlier* is not discovered by the run. A rate added effective Feb 15 on a card whose clock reads Mar 1 splits the February cycle, and the Feb 15 slice is written on the Mar 1 tick — late, though carrying its own Feb 15 billing date, so nothing is lost or mispriced.

I scoped it instead of fixing it, for a reason I want on the record. Discovering it in the producer would mean building every card's schedule on every tick, which is exactly the work the clock exists to avoid. So pulling the clock back belongs to whatever mutates the schedule — and that path must write \`next_billing_at\` directly, because \`AdvanceBillingClockService\` only ever moves forward, deliberately: a replayed run must not rewind a clock and bill a period twice.

What I did do is stop it being a lurking surprise. Two specs pin it: nothing is written at the new billing time, and the slice is written on the tick the clock does reach. The same reasoning is in the class comment, next to the code that causes it.

The termination half of the finding is scoped the same way and for the same reason: a terminated contract's final arrears period is produced by the termination path, through the same \`CreateService\`.

app/models/contract_rate_card.rb
```diff
@@ -44,6 +45,20 @@ class ContractRateCard < ApplicationRecord
...
+  # customer-local cut, not a plain `ended_date IS NULL`.
+  #
+  # That cut is against Time.current, not the given instant: a run for a past instant
+  # leaves out a card that has ended since. A card already ended is the termination
+  # path's business, and so is a terminated contract's final arrears period — a run
+  # driven by a clock cannot tell a finished schedule from one still to come.
+  scope :due_for_billing, ->(timestamp) {
+    current_and_scheduled
+      .where(next_billing_at: ..timestamp)
+      .where(contracts: {status: :active, started_at: ..timestamp})
```
tiagolupepic: better to move this status to Contract model. put into a constant `BILLABLE_STATES`, wdyt?

brunomiguelpinto: yes good ideia

app/jobs/clock/create_billing_segments_job.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def perform
+      Customer.where(id: due_customer_ids).find_each do |customer|
+        BillingSegments::ScheduleJob.perform_later(customer)
+      end
+    end
+
+    private
+
+    def due_customer_ids
```
tiagolupepic, line 21: review these queries. You're doing two, it's possible to use a single one and change the BillingSegments::ScheduleJob accept the customer_id instead.

app/services/billing_segments/schedule_service.rb
```diff
@@ -0,0 +1,103 @@
...
+      build = Billing::RateCards::BuildScheduleService.call(contract_rate_card: card)
+
+      if build.failure?
+        return report_unschedulable(card, build.error)
+      end
+
+      schedule = build.schedule
+      pricing_unit = pricing_unit_of(card)
+
+      missing = MissingBillableSegmentsService.call!(contract_rate_card: card, schedule:, timestamp:)
```
tiagolupepic, line 59: what is the propose of this service?

app/models/contract_rate_card.rb
brunomiguelpinto: Done — `Contract::BILLABLE_STATUSES`

brunomiguelpinto: Done — `Contract::BILLABLE_STATUSES`, next to the `LIVE_STATUSES` already there, and the scope reads `contracts: {status: Contract::BILLABLE_STATUSES, ...}`.

I spelled it `BILLABLE_STATUSES` rather than `BILLABLE_STATES` only to match `STATUSES` and `LIVE_STATUSES` in the same file — say the word if you prefer yours.

It also gave the boundary somewhere to be written down, which the scope could not carry: a terminated contract is excluded because its final arrears period belongs to the termination path, not because termination means nothing is owed.

app/jobs/clock/create_billing_segments_job.rb
brunomiguelpinto: Right on both counts, fixed.  One query.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: It answers one question: of what the calendar says a card owes, which pieces are not stored yet. It exists because the calendar hands back pieces a previous run already wrote, by design.

A schedule resumes from `billing_segments.maximum(:cycle_started_at)` — the *start* of the last stored cycle — and the walker stays on a cycle until its end is passed, so the last cycle is always replayed whole. It has to be: a cycle can be half written (an arrears cycle split by a rate change bills one piece on one run and the other on the next), and resuming from the cycle's end instead would skip the unwritten piece for good.

Measured in, and out:

```
already stored     in, from the calendar         out
(nothing)          Jan 1-Feb 1, Feb 1-Feb 15     both
Feb 1-Feb 15       Feb 1-Feb 15, Feb 15-Mar 1    Feb 15-Mar 1
Feb 1-Mar 1        Feb 1-Feb 15, Feb 15-Mar 1    (nothing)
```

Row three is why the test is overlap and not an equal start, and it is the bug this branch had: a rate added inside an already-billed cycle re-splits it, and the later piece starts at the rate's date rather than the cycle's. An equal-start test called it new. An advance card that had billed February in full, then took a rate effective February 10, stopped billing permanently — the write tripped the exclusion constraint, the run wrote nothing and never advanced the clock, and every later tick failed the same way.

It could live inside `Schedule` — the calendar already takes a persistence-derived input in `resume_at`. What keeps it out is the shape of the query: its window is derived from the candidates, so it can only run once the calendar has answered. Inside, it would have to read every period ever stored for the card.

tiagolupepic: > It could live inside Schedule instead, but that would put a database read into the pure calendar,

I see, make sense. Maybe it worth to review the `billing_segments.maximum(:cycle_started_at)`. It looks like we have two paths to follow and we could avoid it. 

Also, would be nice to create a `request` spec to cover such scenario.

app/services/billing_segments/schedule_service.rb
tiagolupepic: > A customer holds few cards 

I hope don't haha. Since we already know the charges we have, it would be nice to use insert instead. I'll create a task to tackle this later and not in this PR

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Measured, since it is easy to be wrong about. A February cycle cut on the 15th, arrears, the first piece already written, asked at March 1:

```
resume_at = the stored cycle's start (Feb 1), what we do  ->  Feb 1, Feb 15
resume_at = the stored piece's end (Feb 15)               ->  Feb 1, Feb 15
resume_at = the cycle's end (Mar 1)                       ->  (nothing)
```

No resume point avoids the replay: the walker resolves to a cycle, so any instant inside one replays it whole, and the only one that does not loses the Feb 15 piece. A scalar cannot express a hole either — with only the second half of February stored, a high-water mark of March 1 drops the unstored first half, which the overlap test keeps.

Scenario spec added: `spec/scenarios/billing_segments/produce_and_invoice_spec.rb`, from the clock job through to the invoice.

brunomiguelpinto:
Correcting my own answer to the pre-review HOLD above, because the design it describes has been superseded.

I wrote that the mutating path should pull `next_billing_at` back and let the producer write on its next tick. That is wrong, and the reason is in #6422: `Events::PayInAdvanceBillingSegmentResolver` looks up a segment whose period covers the event's timestamp. Between a rate change committing and the producer running, the old segment no longer covers that instant and the new one does not exist yet, so a pay-in-advance event arriving in that window finds nothing — and `handle_pay_in_advance` returns early rather than deferring, so the fee is never created and nothing says so.

The fix is not a wider clock or a lock. A reader cannot see half a change: if the mutation closes the current segment and writes its replacements in one transaction, MVCC gives every reader either the old state or the new one, never a gap. So the mutating path writes the segments itself, inline, and `BillingSegments::ScheduleService` composes for that — its transaction nests in the caller's, and the advisory lock is held to the outer commit.

What stays true for this PR: the clock-driven producer does not discover a change made before a card's saved clock, and two specs pin that it writes the piece on the next tick it does reach, carrying its own earlier billing date. What changes is who closes the window, and how.

The termination half is unchanged and takes the same route: it writes its own final arrears period through `CreateService`.

app/services/billing/rate_cards/schedule.rb
brunomiguelpinto: changed the code this no longer applies

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: the walker already starts at the card's `effective_date`, so a past `timestamp`
produces every cycle from there.

But you were right that past dates were broken: the instant only governed the clock, while the
cut that drops ended cards ran against `Time.current`, losing the unbilled periods of a card
that had ended since. `current_and_scheduled` now takes that instant and `due_for_billing`
passes it, so the scope is a pure function of what you ask for.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Done, and the case you describe stopped reaching the database anyway: the filter before the write now tests overlap rather than an equal start, so a stored segment that covers a candidate from a different instant is treated as settled and never written.

app/services/billing/rate_cards/schedule.rb
brunomiguelpinto: The code you flagged is gone — it was the `billing_from:` branch this PR added to `Schedule`, and the rebuild dropped it.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Not a range — the walker already starts at the card's `effective_date`, so a past `timestamp` produces every cycle from there.

But you were right that past dates were broken: the instant only governed the clock, while the cut that drops ended cards ran against `Time.current`, losing the unbilled periods of a card that had ended since. `current_and_scheduled` now takes that instant and `due_for_billing` passes it, so the scope is a pure function of what you ask for.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: you were right, changed to be an insert all

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: Fixed, though not by skipping: a card whose rate card has no rate is no longer selected — `due_for_billing` keeps only priced ones, because a card with no price owes nothing. Its clock is untouched, so it bills from the moment it gets a price.

app/services/billing_segments/schedule_service.rb
brunomiguelpinto: All three done, measured at 6 queries per additional card — two of them the batched insert and its returning.

The graph is preloaded, the stored periods come back in one windowed query per card rather than a check per candidate, and pricing units resolve once per customer. What preloading cannot reach is `RateCard#ordered_rates`: `.order` on a loaded association re-queries, so fixing it means changing `RateCard`.

app/services/billing_segments/missing_billable_segments_service.rb
```diff
@@ -0,0 +1,62 @@
...
+        return result
+      end
+
+      settled = settled_periods(due)
+
+      # Overlap, not an equal start: row 3 above is what an equal-start test gets wrong.
+      result.billable_segments = due.reject do |billable_segment|
+        period = billable_segment.started_at...billable_segment.ended_at
+
+        settled.any? { |settled_period| settled_period.overlaps?(period) }
```
tiagolupepic, line 35: Any intersection is treated as though the whole candidate were settled. For example, a stored Feb 5–20 period makes a Feb 1–Mar 1 candidate disappear; the clock then advances and permanently skips the uncovered days. Reject only candidates fully covered by stored periods. A partial overlap should instead produce the documented `overlapping_periods` failure or be split safely.

app/jobs/billing_segments/schedule_job.rb
```diff
@@ -0,0 +1,21 @@
...
+        :default
+      end
+    end
+
+    unique :until_executed, on_conflict: :log, lock_ttl: 30.minutes
+
+    retry_on BaseLockService::FailedToAcquireLock, attempts: MAX_LOCK_RETRY_ATTEMPTS, wait: random_lock_retry_delay
+
+    def perform(customer_id)
+      ScheduleService.call!(customer: Customer.find(customer_id))
```
tiagolupepic, line 18: `due_for_billing` joins through `Contract#customer`, whose association includes discarded customers, while this lookup applies `Customer`'s kept-only default scope. Because customer deletion does not terminate contracts, discarded customers with active contracts remain due indefinitely, but every hourly job raises `ActiveRecord::RecordNotFound`. Exclude discarded customers from the due query, or deliberately load them with `with_discarded` if they should still be scheduled.

app/models/contract_rate_card.rb
```diff
@@ -35,15 +36,26 @@ class ContractRateCard < ApplicationRecord
...
 
+  scope :due_for_billing, ->(timestamp) {
+    current_and_scheduled(timestamp)
+      .where(next_billing_at: ..timestamp)
+      .where(contracts: {status: Contract::BILLABLE_STATUSES, started_at: ..timestamp})
+      # Only a priced card owes anything: a period with no price is not one to be paid for.
+      .where(rate_card_id: RateCardRate.select(:rate_card_id))
+      # And only a card with a window: one that starts after its contract ends has no period
+      # to owe anything in, which is an ordinary outcome of bringing a termination forward.
+      .where("contracts.ended_at IS NULL OR contracts.ended_at > contract_rate_cards.effective_date")
```
tiagolupepic: `effective_date` is a customer-local date, but this comparison casts it to midnight in the database timezone. For example, a UTC+14 card effective Jan 1 actually starts Dec 31 10:00 UTC, so a contract ending Dec 31 20:00 UTC has a valid window but is excluded here. Negative offsets can conversely admit contracts that ended before the local start and make schedule construction fail. Compare `ended_at` with the effective date converted through the customer's applicable timezone.

app/services/billing_segments/missing_billable_segments_service.rb
brunomiguelpinto: this will be address in another pr

tiagolupepic (approved):
:shipit:

PR #6349, opened 2026-09-09
feat(billing): scaffold fixed segment processing

tiagolupepic:
## Context

Product-catalog billing needs an invoice-processing boundary for priced contract segments without duplicating the shared monetary calculations. This establishes a fixed-product processing skeleton, not a complete production billing runtime.

## Description

Define BillingSegments::ProcessService as the customer-scoped orchestrator. It locks processing per customer, selects pending fixed-product segments, and groups them by local billing date, consolidation setting, currency, billing entity, payment method, and purchase order. It creates generating invoices, persists computed fees and segment completion atomically, and retries finalization of existing generating invoices without rebuilding fees. Usage segments remain pending.

Add BillingSegments::Fees::ComputeService to build unsaved product fees and linked minimum true-ups from segment pricing properties and contract rate-card units. Reuse Fees::AmountsService for monetary calculations and move minimum proration and fiat-to-pricing-unit normalization into BillingSegment.

Rename the calculator value object to AppliedPricingUnit and expose separate factories for pricing units with conversion rates and existing applied pricing units. Update caller references and retain explicit empty-state handling. Link invoices to distinct contracts through billing segments.

Add coverage for fixed-product pricing, precision, minimums, invoice grouping, locking, rollback, finalization retries, and leaving usage segments untouched.

## Remaining work

Fee model and schema support for the contract and product-pricing references must be completed before the fixed-product flow works end to end. The compute specs currently fail because Fee does not accept the contract attribute.

Metered fee creation, event prefiltering, and Fees::ChargeService integration are intentionally not part of ProcessService. Scheduler/job wiring and a due-date cutoff are not included in this change; pending segments are currently selected without checking billing_at. Consolidation opt-out currently groups by segment rather than rate-card cycle.


app/services/billing_segments/process_service.rb
```diff
@@ -0,0 +1,122 @@
...
+            []
+          end
+        end.each do |fee|
+          fee.invoice = invoice
+          fee.billing_entity = invoice.billing_entity
+          fee.save!
+        end
+        invoice.fees.reload
+
+        Invoices::ComputeAmountsFromFees.call!(invoice:)
```
rsempe, line 95: Sequencing note rather than a defect here: `Fees::ApplyTaxesService#applicable_taxes` has no `product?` branch on `main`, so these fees skip the rate card's own `taxes` and fall through to customer / billing-entity defaults — and `finalize_generating_invoices` then finalizes the invoice with them.

`feat/BIL-586/resolve-taxes-on-fees` already adds it (`return product_taxes if fee.product?`, reading `fee.rate_card_rate.rate_card.taxes`). Worth calling out in Remaining work that scheduler wiring must not land before BIL-586.

rsempe (commented):
lgtm

rsempe (approved):

PR #6350, opened 2026-09-09
feat(streaming): send usage since subscription start for the plans that need it

toommz:
Stacked on #6299. Closes ING-675.

## Context

Synthesia does not consume current-period usage today. They call the usage endpoint with the `full_usage` flag for their enterprise customers and without it for self-served ones, so P0 as it stands does not replace their polling for the enterprise half.

They need both scopes: every subscription gets current-period usage, and subscriptions on plans that are not excluded additionally get usage since `subscription.started_at`.

## Description

A second event type, `customer_full_usage.refreshed.v1`, delivered alongside the first. Both carry `object_type: "customer_usage"` and the same payload shape; only the window differs. The two records for one subscription are independent, with no shared `version` and no ordering guarantee, so a consumer dedups per `(organization, customer, subscription, event_type)`.

Selection is `customer_full_usage_excluded_plan_codes` on the destination, read through `event_types_for(subscription)`. The list is an exclusion, not an inclusion: an absent or empty list excludes nobody, so a new plan nobody has configured receives the complete data rather than silently dropping to less.

Full usage is computed in **one call carrying every charge id**. `has_charge_filter?` only checks presence, so an array of all charge ids satisfies the guard, and a spec asserts this directly since a per-charge loop is the regression to prevent. Measured at 70 charges: 73.7 ms batched against 728 ms looped.

The service is duplicated rather than sharing a base class with the current-period one. Extracting a parent would mean refactoring code already reviewed and approved in #6298, which is a worse trade than the duplication.

## Preconditions, both silent when unmet

- The organization needs the `granular_lifetime_usage` premium integration and a premium license, or every full-usage delivery is refused and logged as `outcome=skipped` while the current-period event keeps working.
- Full usage is refused for any subscription whose plan carries a prorated charge. Those subscriptions never receive it, and still receive the current-period event.

## Not in this PR

Cold cost is unmeasured. Full usage carries its own cache key the wallet refresh never warms, and it aggregates from `subscription.started_at`, so its cost grows with subscription age. That needs measuring on a mature subscription against real event history before any cadence or worker sizing is committed to, and it may lead to debouncing the event or computing it incrementally.

### DIFF
--- a/app/jobs/deliver_event_job.rb
+++ b/app/jobs/deliver_event_job.rb
@@ -3,7 +3,16 @@
 class DeliverEventJob < ApplicationJob
   queue_as :streaming
 
-  unique :until_and_while_executing, on_conflict: :log, lock_ttl: 10.minutes
+  ON_CONFLICT = lambda do |job|
+    EventDestinations::DeliveryLogger.emit(
+      :superseded,
+      event_type: job.arguments.first,
+      customer_id: job.arguments.second.try(:id),
+      lock_key: job.lock_key
+    )
+  end
+
+  unique :until_and_while_executing, on_conflict: ON_CONFLICT, lock_ttl: 10.minutes
 
   EVENT_SERVICES = {
     EventDestinations::CustomerUsage::RefreshedService::EVENT_TYPE =>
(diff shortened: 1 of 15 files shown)


toommz:
Superseded: the full usage event now ships in #6298, so it does not depend on the delivery logging PR.

PR #6360, opened 2026-09-10
feat(organization): Ignore pre_filter_events column

vincent-pochet:
## Context

`pre_filter_events` was added on `organizations` as an exploratory flag and was never enabled. It is going away, and the column can only be dropped once no running instance selects it anymore.

## Description

- Add `pre_filter_events` to `Organization.ignored_columns` so the column is no longer selected before it is removed.

The migration dropping the column comes in a follow-up PR, once this one is deployed everywhere.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -8,7 +8,7 @@ class Organization < ApplicationRecord
   include HasFeatureFlags
   include Organizations::Sluggable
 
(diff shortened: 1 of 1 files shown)


toommz (approved):

tiagolupepic (approved):
thanks!

:shipit:

PR #6361, opened 2026-09-10
[ING-504] fix(alerts): report duplicate wallet alert codes on the code field

aquinofb:
## Context

An alert code identifies an alert within its owner. The subscription side enforces that with a unique index and the wallet side does not, so two wallet alerts on the same wallet can share a code today.

The update path only recognised the subscription index when translating a uniqueness violation, so an edit into a duplicate code reported a generic "alert already exists" against the record instead of naming the code field. Creating reported correctly, editing did not.

## Changes

- Both paths now share one predicate for recognising a code collision. Two copies of that check drifting apart is what produced the difference.
- Creating or renaming a wallet alert rejects a code already used on the same wallet before any insert or update is attempted, rather than relying on the database to refuse it.

## Why this is split out

This carries no migration. It is split out of #6082 so the application understands the constraint before the constraint exists: the unique index is built by hand ahead of the deploy on large installs, and until this ships a customer renaming a wallet alert to a taken code would get the wrong error.

Merge order: this PR, then the index build, then #6082, then #6160.


aquinofb:
This PR ships without the index on purpose. It is split out of #6082, which adds `idx_alerts_code_unique_per_wallet` together with the migration that renames existing duplicates.

The reason for the split is ordering. On larger installs the index is built by hand ahead of the deploy rather than inside the migration, which leaves a period where the database enforces uniqueness before the application recognises it. Until this PR is deployed, `UpdateAlertService` only matches the subscription index when translating a uniqueness violation, so renaming a wallet alert to a code already taken reports a generic "alert already exists" against the record instead of `value_already_exist` on the code field. Deploying this first closes that window, and the pre-check also stops fresh duplicates being written while the index is building.

On the race: the pre-check is not atomic and is not meant to be. Wallet alert codes have no unique index today, so this change does not widen the existing window, it narrows it. Atomic enforcement arrives with the index in #6082, and the specs that exercise the constraint-backed path live there as well, since they need the index to exist.

Merge order: this PR, then the manual index build, then #6082, then #6160.

toommz (approved):

PR #6362, opened 2026-09-10
perf(plans): Optimize subscription and invoice counts

annvelents:
Plans with many overrides can load slowly because active-subscription and draft-invoice counts scan records beyond the relevant plans.

Sum active-subscription counts per override within SQL, allowing PostgreSQL to use the existing `(plan_id, status)` index. Scope draft-invoice counts to the plan's organization and remove the duplicate subscription join, while retaining distinct invoice counting and excluding discarded overrides.

Validation:
- 11 focused model examples passed for active-subscription and draft-invoice counts, including statuses, empty results, unrelated plans, discarded overrides, and shared invoices.
- RuboCop passed for both changed files.
- `EXPLAIN ANALYZE` on the EU read replica confirmed index-only scans using the existing subscription index. No migration is required.

Commands:
```sh
lago exec api bundle exec rspec spec/models/plan_spec.rb --example '#active_subscriptions_count' --example '#draft_invoices_count'
lago exec api bundle exec rubocop app/models/plan.rb spec/models/plan_spec.rb
```

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -123,9 +123,13 @@ class Plan < ApplicationRecord
 
   def active_subscriptions_count
     count = subscriptions.active.count
-    return count unless children
+    active_count_per_child = Subscription.active
+      .where(Subscription.arel_table[:plan_id].eq(Plan.arel_table[:id]))
+      .select("COUNT(*)")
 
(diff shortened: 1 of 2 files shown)


vincent-pochet (approved):
👍

toommz (approved):

PR #6363, opened 2026-09-10
misc(clickhouse): remove enriched store migration

vincent-pochet:
## Context

An automated pipeline migrating organizations from the ClickHouse event store to the enriched one was scaffolded but never finished: the organization-level orchestrator was left as an empty stub and nothing ever created a migration record. The state machine is unreachable, its only inbound edge coming from within the pipeline itself, so both backing tables are empty.

## Description

- Remove the `EnrichedStoreMigration` and `EnrichedStoreSubscriptionMigration` models, along with the `Organization` association.
- Remove the three orchestration jobs and the subscription orchestrator, wait-for-enrichment and comparison services.
- Remove the `enriched_events:compare` rake task, which the comparison service backed.
- Remove the matching factories and specs.

The tables and enum types stay for now and are dropped in a follow-up, once this is deployed everywhere. Re-enrichment and the enriched store itself are untouched: they are used outside of this pipeline.


toommz (approved):
:'(

PR #6364, opened 2026-09-10
misc(clickhouse): remove enriched events store

vincent-pochet:
> Stacked on #6363, review that one first.

## Context

The enriched events store was an alternative aggregation implementation reading from `events_enriched_expanded`, gated behind the `enriched_events_aggregation` feature flag. The flag was never enabled for any organization, so the store was only ever exercised from its own specs.

One of its methods was reachable regardless of the flag: the ClickHouse store delegated `distinct_charges_and_filters` to it, because that lookup genuinely needs the expanded table.

## Description

- Remove the `enriched_events_aggregation` feature flag and the `StoreFactory` branch selecting the enriched store.
- Move `distinct_charges_and_filters` into `ClickhouseStore`, now its only caller.
- Remove `Events::Stores::ClickhouseEnrichedStore` and its spec.
- Regenerate the GraphQL schema dumps, since the feature flag enum is built from the flag definitions.

Organizations may still carry the flag string in `feature_flags`. It is inert — `feature_flag_enabled?` returns false and `CurrentOrganizationType` filters on `FeatureFlag.valid?` — and `FeatureFlag.sanitize!` clears it. The enrichment pipeline and the enriched events tables are untouched: `ClickhouseStore` keeps reading from them.


vincent-pochet:
Split into #6368 (stop routing to the enriched store), #6369 (remove the feature flag) and #6370 (remove the store itself). The enriched store migration half already landed in #6363.

PR #6366, opened 2026-09-10
misc(db): drop the enriched store migration tables

vincent-pochet:
> Stacked on #6365, review that one first.

## Context

The tables backing the enriched store migration pipeline are no longer read or written by any code, and they never held a row: nothing ever created a migration record, so dropping them cannot lose data.

## Description

- Drop `enriched_store_subscription_migrations`, then `enriched_store_migrations`.
- Drop the `enriched_store_sub_migration_status` and `enriched_store_migration_status` enum types.

Merge only once #6363 is deployed everywhere, so that no running instance still resolves the models.


toommz (approved):
🧹

PR #6367, opened 2026-09-10
Ing 462

mariohd:
## Context

ING-462. The subscription half of per-object connection routing (dive-in T18b). ING-466 added the write path for wallets plus the shared validation and persistence services; this reuses them unchanged.

## Description

`POST/PATCH /subscriptions` accept a `connections` object keyed by category, with the same semantics as the wallet endpoints:

```json
"connections": {
  "payment":    { "behavior": "skip" },
  "accounting": { "code": "netsuite_sub2" },
  "crm":        { "behavior": "inherit" }
}
```

A `code` pins that connection, `skip` short-circuits the category so it never falls back, `inherit` drops the override, and an omitted category changes nothing.

**No new services.** ING-466's two are polymorphic in the owner, so this PR is only params, one validator method, and two write paths — 261 lines, most of it specs.

Behind the `multi_connection` flag. Connections sent while the flag is off are refused rather than ignored; an unresolvable code is a 422. The existing top-level `payment_method` is untouched.

### Reviewer note — the downgrade case

`Subscriptions::CreateService` attaches `invoice_custom_section` with `unless downgrade?`. I deliberately did **not** copy that guard. Connections are routing config in the same family as `payment_method`, which both `PlanUpgradeService` and `PlanDowngradeService` already carry onto the new subscription — so skipping them on a downgrade would silently discard the operator's choice. Upgrades and downgrades are each pinned by a spec.

## Testing

12 new examples: 7 request (4 POST, 3 PUT) and 5 service, the latter covering plain create, unresolvable code, flag off, and the upgrade/downgrade behaviour above. Regression: 2001 examples across the subscription request and service suites, plus 156 for the non-REST callers — GraphQL mutations, `Orders::SubscriptionCreation::ExecuteService`, `Orders::SubscriptionAmendment::ExecuteService`. All green, Rubocop clean.

### DIFF
--- a/app/services/subscriptions/create_service.rb
+++ b/app/services/subscriptions/create_service.rb
@@ -28,12 +28,14 @@ module Subscriptions
         subscription_at:,
         ending_at: params[:ending_at],
         payment_method: params[:payment_method],
+        connections: params[:connections],
         activation_rules: params[:activation_rules],
         subscription_type:,
         consolidate_invoice: params[:consolidate_invoice],
         consolidate_invoice_provided: params.key?(:consolidate_invoice)
       )
       return result.forbidden_failure! if !License.premium? && params.key?(:plan_overrides)
+      return result.forbidden_failure! if connections_requested? && organization_flag_disabled?(:multi_connection)
 
       if params.key?(:plan_overrides) && plan.organization.product_catalog_enabled?
         return result.single_validation_failure!(field: :plan_overrides, error_code: "legacy_billing_disabled")
@@ -83,6 +85,10 @@ module Subscriptions
           end
           InvoiceCustomSections::AttachToResourceService.call(resource: subscription, params:) unless downgrade?
 
+          if connections_requested?
+            BillingObjectConnections::AttachToResourceService.call!(resource: subscription, params:)
+          end
+
           result.subscription = subscription
         end
       end
@@ -112,6 +118,14 @@ module Subscriptions
       result.payment_method = payment_method
 
       Subscriptions::ValidateService.new(result, **args).valid?
+    end
+
+    def connections_requested?
+      params[:connections].present?
+    end
+
+    def organization_flag_disabled?(flag)
+      !customer.organization.feature_flag_enabled?(flag)
     end
 
     def handle_subscription
--- a/app/controllers/api/v1/subscriptions_controller.rb
+++ b/app/controllers/api/v1/subscriptions_controller.rb
@@ -180,6 +180,12 @@ module Api
               :payment_method_type,
               :payment_method_id
             ],
+            connections: [
+              payment: [:behavior, :code],
+              tax: [:behavior, :code],
+              accounting: [:behavior, :code],
+              crm: [:behavior, :code]
+            ],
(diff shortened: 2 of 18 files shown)


app/services/subscriptions/create_service.rb
```diff
@@ -83,6 +85,10 @@ def call
           end
           InvoiceCustomSections::AttachToResourceService.call(resource: subscription, params:) unless downgrade?
 
+          if connections_requested?
+            BillingObjectConnections::AttachToResourceService.call!(resource: subscription, params:)
```
lovrocolic, line 89: We should guard this call `unless downgrade?`

lovrocolic:
@mariohd Can you please follow the design we used for `InvoiceCustomSections::AttachToResourceService`? In this PR we miss support for downgrade case and potentially for upgrade case. Can you please double check?

app/services/subscriptions/create_service.rb
mariohd: Done, and thanks for pushing on this — the downgrade case was a real bug.

**Design.** The caller now reads exactly like the ICS line above it:

```ruby
InvoiceCustomSections::AttachToResourceService.call(resource: subscription, params:) unless downgrade?
BillingObjectConnections::AttachToResourceService.call!(resource: subscription, params:) unless downgrade?
```

I dropped the `connections_requested? &&` part — the service already returns early when the key is absent, same as ICS. Kept `call!` on purpose: an unresolvable code should fail the request rather than silently drop routing.

**Downgrade.** Genuinely broken, as you suspected. `PlanDowngradeService` returns `current_subscription`, not the new pending one, so attaching in the caller was writing routing onto the subscription that's still billing today while the downgraded one got nothing. The guard alone would have dropped it entirely, so the attach also moved into `PlanDowngradeService` next to the ICS call, on `new_sub`.

**Upgrade.** Checked both branches — already correct, no change needed. The normal path returns `new_subscription`, and the `starting_in_the_future?` path returns the pending subscription it just updated in place. Both are the right target.

**One thing worth knowing** (pre-existing, not from this PR): the future-dated *downgrade* only works because `downgrade?` is re-evaluated *after* `handle_subscription` has run, and `update_pending_subscription` mutates `current_subscription.plan` in place — so `plan.id == current_subscription.plan.id` becomes true and `downgrade?` flips to false, letting the caller attach. Correct outcome, accidental mechanism, and ICS rides on the same accident. Anyone who memoizes `downgrade?` or hoists it above `handle_subscription` silently breaks both. Happy to harden it in a follow-up if you'd rather not leave it load-bearing.

Added regression specs for all four paths, including the two future-dated ones.

app/services/subscriptions/plan_downgrade_service.rb
```diff
@@ -61,6 +61,10 @@ def call
 
         InvoiceCustomSections::AttachToResourceService.call(resource: new_sub, params:)
 
+        if params[:connections].present?
```
lovrocolic: This guard is not needed if I am not wrong

lovrocolic:
@mariohd I guess we also need support in the serializer

PR #6380, opened 2026-09-11
perf(usage): skip last_seen_at without lazy cache

vincent-pochet:
## Context

The charge usage cache is invalidated in one of two ways. With `lazy_charge_usage_cache` enabled, an entry keeps the ingestion watermark it aggregated and is validated at read time against the most recent event. Without the flag, the entry is expired when the event is received and the watermark is never read.

That watermark is the `MAX(enriched_at)` aggregate of the billing period filter query, so requesting it reads the column for every event matching the subscription and the metric.

## Description

- Gate the watermark on lazy validation rather than on the charge cache alone, so organizations relying on eager expiration stop requesting it.
- Add `ChargeCacheService.lazy_validation_enabled?` as the single predicate, shared by the full usage gate and by the eager expiration that is its complement.
- Request the watermark in the invoice preview under that same condition, where it was previously always asked for.
- Cover both gates in the customer usage and preview specs, which previously asserted the watermark was always requested.

Measured against a week of production query logs, `enriched_at` is roughly five percent of what this query reads, so this is cleanup rather than a material reduction in cluster pressure. The bulk is the `properties` map, untouched here.

### DIFF
--- a/app/services/events/post_process_service.rb
+++ b/app/services/events/post_process_service.rb
@@ -82,7 +82,7 @@ module Events
     end
 
     def expire_cached_charges
-      return if organization.feature_flag_enabled?(:lazy_charge_usage_cache)
+      return if Subscriptions::ChargeCacheService.lazy_validation_enabled?(organization)
       return if active_subscription.nil?
       return unless billable_metric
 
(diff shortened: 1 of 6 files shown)


vincent-pochet:
Closing as `lazy_charge_usage_cache` will become the default. This change will become useless

PR #6381, opened 2026-09-11
refactor(product): rename product_type usage to metered

rsempe:
## Context

The `usage` product type is better named `metered` — it matches how the value is described elsewhere and avoids collision with the many unrelated "usage" concepts in billing.

## Description

Renames the `product_type` PostgreSQL enum value `usage` → `metered` via `rename_enum_value` (metadata-only `ALTER TYPE ... RENAME VALUE`, no table rewrite). `fixed` is unchanged.

- `Product::PRODUCT_TYPES` mapping
- `metered?` predicate call sites (`product.rb`, `rate_card.rb`, `product_filters/create_service.rb`)
- GraphQL `ProductTypeEnum` (derives from `PRODUCT_TYPES`, now exposes `metered`)


tiagolupepic (approved):
:shipit:

PR #6382, opened 2026-09-11
fix(usage): minimize ignored charge filters [ING-637]

vincent-pochet:
Part of ING-637

## Context

Usage for the default bucket of a charge excludes every event already billed by one of the charge's other filters, one `NOT (... OR ...)` clause per filter with the values inlined as literals. The predicate therefore grows without bound with the number of filters on the charge.

On a charge carrying 2892 filters it reaches ~700 kB, past the 256 kB ClickHouse parses by default, and the aggregation fails with `Max query size exceeded`. Current usage, daily usage and wallet refresh all break for those subscriptions.

## Description

- Rewrite the ignored filters into a smaller list describing the same set of events, in `Events::BillingPeriodFilters::MinimizeIgnoredFiltersService`.
- Drop keys without values, which the event stores already skip when building the SQL.
- Merge clauses differing on a single key by taking the union of that key's values.
- Drop a clause whose events are all matched by another, more generic, one.
- Index clauses by the values they allow on a pivot key, so absorption compares a handful of candidates rather than every pair.

```
charge with 2892 filters, 164 models x 3 types x optional zone and tier

before: 2892 clauses, 698_094 bytes   # over the 262_144 ClickHouse parses
after:      2 clauses,   9_156 bytes
```

Charges whose filters do not overlap are left untouched. This lowers the predicate rather than raising the ClickHouse ceiling, which is worth doing separately.

- Several existing expectations now list fewer ignored filters. Each dropped clause was already covered by another, so the excluded events are unchanged.

### DIFF
--- a/app/services/events/billing_period_filters/matching_and_ignored_service.rb
+++ b/app/services/events/billing_period_filters/matching_and_ignored_service.rb
@@ -21,7 +21,7 @@ module Events
           end
         end
 
-        result.ignored_filters = children.map do |child_filter|
+        ignored_filters = children.map do |child_filter|
           child = target_filter.filter_values(child_filter).dup
 
           if child.keys.sort == result.matching_filters.keys.sort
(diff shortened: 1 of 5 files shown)


annvelents (approved):
WOW, it is so beautiful! 😍

PR #6384, opened 2026-09-11
feat(account-tree): add update flow for usage attribution types

lovrocolic:
## Context

Account tree feature allows tracking usage under custom types like teams or users.

## Description

This PR adds support for `usage_attribution_type` update flow with GQL.


toommz (approved):

toommz:
> 
> HOLD: Types become editable again after their last attribution value is soft-deleted.
> 
> * `attributed?` uses the default-scoped `usage_attribution_values.exists?`, so discarded historical values are ignored. Include discarded values in this check and add a regression spec proving `code`, `attribution_key`, `role`, and `parent_id` remain frozen once any attribution has existed.

@lovrocolic I would rather update the UNIQUE INDEX to exclude discarded rows rather than check for discarded in the service, but you know better than me what is the best product choice here :)

lovrocolic:
> > **Automated pre-review** (advisory, not a required check) — verdict: **HOLD** · CI green
> > HOLD: Types become editable again after their last attribution value is soft-deleted.
> > 
> > * `attributed?` uses the default-scoped `usage_attribution_values.exists?`, so discarded historical values are ignored. Include discarded values in this check and add a regression spec proving `code`, `attribution_key`, `role`, and `parent_id` remain frozen once any attribution has existed.
> 
> @lovrocolic I would rather update the UNIQUE INDEX to exclude discarded rows rather than check for discarded in the service, but you know better than me what is the best product choice here :)

@toommz We actually went the other way on purpose in #6342 — I had the partial index in an early version and dropped it before pushing. The idea was that (customer_id, usage_attribution_type_id, value) should be one row forever: when a value we've seen before shows up again in events we undiscard the existing row rather than insert a second one, so `last_seen_at`, `parent_id` and its children stay attached to a single record.
Note the types table is deliberately the opposite — both its unique indexes are WHERE deleted_at IS NULL — because deleting a type should free its code for reuse.

Issue #5890, opened 2026-07-08
[FEAT]: Add Tunisian Dinar (TND) currency support


soufiene-slimi:
**Is your feature request related to a problem? Please describe.**
TND (Tunisian Dinar) is a standard ISO 4217 currency that is not currently in the supported currencies list. Tunisian users are unable to use Lago with their local currency.

**Describe the solution you'd like**
Add `TND: "Tunisian Dinar"` to the `ACCEPTED_CURRENCIES` hash in `app/models/concerns/currencies.rb` and regenerate the GraphQL schema files.

**Describe alternatives you've considered**
Manually overriding the currency validation on a per-request basis, but this is error-prone and inconsistent with Lago's design.

**Additional context**
TND uses 3 decimal places (1000 millimes = 1 dinar), which is already accounted for in the frontend serialization (`CURRENCIES_WITH_3_DECIMALS` in `serializeAmount.ts`). The `money-rails` gem natively supports TND. Only the backend currency whitelist needs updating.

Corresponding changes will also be needed in `lago-openapi` (`src/schemas/Currency.yaml`).

PR #6389, opened 2026-09-13
feat(currencies): add Tunisian Dinar (TND) currency support (#5890)

Tyagiquamar:


vincent-pochet (approved):

vincent-pochet:
Thanks for the contribution @Tyagiquamar! 🙏

The currency list isn't the only place TND has to land: `schema.graphql` and `schema.json` are checked in and compared against the live schema in `spec/graphql/lago_api_schema_spec.rb`, so a new `CurrencyEnum` value has to be regenerated there too (`bundle exec rake graphql:schema:dump`) or CI fails.

I've opened #6424 with your change plus the regenerated schemas, keeping you as co-author. Closing this one in favour of it.

PR #6393, opened 2026-09-14
feat(catalog): create product metered fees from billing segments

tiagolupepic:
Extend Fees::ChargeService so a metered product's BillingSegment can be priced into product-backed fees, without going through a legacy Charge.

- MeteredItem gains a from_billing_segment source and drives buckets through source.pricing_buckets, so filter expansion, true-up target and filter association are resolved polymorphically by the source rather than assuming charge-shaped data.
- Segment-backed fees persist fee_type: :product with the product as invoiceable and set product_filter_id (never charge_filter_id). Legacy charge filters are not expanded, adjusted-fee lookup is skipped, and no charge-shaped CachedAggregation is written.
- Segment items use a nil cache path instead of building a charge cache middleware from fake boundaries.
- Fees::CreateTrueUpService supports segment-backed true-ups, reading the prorated minimum, proration ratio and pricing-unit conversion from the segment snapshot while leaving charge-backed true-ups unchanged.
- Add fees.product_filter_id (nullable, indexed, FK validated in a paired migration) and the discarded-aware Fee#product_filter association.

Billing-state idempotency is owned by BillingSegment persistence, so fee-level already_billed? is bypassed for segments. Store/enrichment and cache identity by product_id remain out of scope and are tracked by TODOs.


brunomiguelpinto (approved):

app/services/fees/charge_service.rb
```diff
@@ -293,19 +297,24 @@ def init_fee(amount_result, selected_metered_item:, adjusted:)
...
+        fee_type: selected_metered_item.fee_type,
+        invoiceable: selected_metered_item.invoiceable,
+        rate_card_rate: selected_metered_item.rate_card_rate,
+        rate_override: selected_metered_item.rate_override,
+        product_filter: selected_metered_item.product_filter,
         units:,
         total_aggregated_units: amount_result.total_aggregated_units || units,
-        properties: selected_metered_item.filtered_for_charge_boundaries,
+        # TODO: Review which fee properties billing segments should expose.
+        properties: selected_metered_item.billing_segment ? {} : selected_metered_item.filtered_for_charge_boundaries,
```
brunomiguelpinto, line 317: With properties: {} the fee has no start and end date, so the invoice line shows no period. The dates already exist — Sources::BillingSegment#boundaries builds them from the segment's started_at/ended_at — they're just being dropped here

tiagolupepic: Indeed, I'll create a task and we can address later, the current names are tied to "charge" name, not sure if we want to keep them.

PR #6395, opened 2026-09-14
feat(rate-cards): reject backdated rates

brunomiguelpinto:
Adds a validation so a rate card rate cannot be created, or moved, with an effective date in the past. It was possible before: the append-only rule only checks placement against the currently active rate, so the first rate on a card, or any date after an already-past active rate, went through.

Nothing already billed changes — a billing segment snapshots its price when it is scheduled, and issued invoices are untouched.

The day is measured in the organization's timezone rather than the server's

The check sits in the create and update services, the only paths user input reaches.


app/services/utils/datetime.rb
```diff
@@ -52,6 +52,15 @@ def self.future_date?(datetime)
       parsed_date&.future? || false
     end
 
+    # Days, not instants: a value landing anywhere inside today is not past.
+    def self.before_today?(datetime)
+      return datetime < Time.current.beginning_of_day if datetime.is_a?(ActiveSupport::TimeWithZone)
+      return false unless valid_format?(datetime, format: :any)
+
+      parsed_date = Time.zone.parse(datetime.to_s)
```
tiagolupepic: we should consider customer timezone here, no? give an example.

UTC is Sep 15th 3am
BRT is Sep 14th 10pm

if the user send a valid effective_from using BRT after 10pm, this will be invalid. Another option, is to use the same timezone sent by the user in effective_from.

brunomiguelpinto: yup good catch added

db/seeds/30_product_catalog.rb
```diff
@@ -87,15 +87,13 @@
...
-  RateCardRates::CreateService.call!(
-    rate_card:,
-    params: {
-      code: "rate_1",
-      effective_from: 1.month.ago.beginning_of_day,
-      rate_model: "standard",
-      rate_properties: {amount: "0.01"},
-      billing_interval_unit: "month"
-    }
+  rate_card.rates.create!(
```
tiagolupepic: any reason to not rely on service here? We might bypass some validations

tiagolupepic (approved):
just a comment and it's not a blocker

:shipit:

db/seeds/30_product_catalog.rb
brunomiguelpinto: No good reason — reverted,

PR #6396, opened 2026-09-14
feat(catalog): wire metered billing through ProcessService

tiagolupepic:
This commit introduces the metered items compute Fee in ProcessService

- Remove fixed-only filter from billing_segments scope to include metered
- Partition segments by product_type using PRODUCT_TYPES constants
- Add attach_fixed_fees and attach_metered_fees methods
- Call ChargeService for metered segments with pre-loaded filter targets
- Add specs for count/sum aggregations, graduated pricing, minimum amounts
- Add mixed fixed+metered segments spec


tiagolupepic:
> build_invoice groups segments across contracts but resolves event filters using only the first contract. The resolver queries that contract's external ID, so ChargeService can bypass aggregation for another contract's segment. Resolve filters in each segment's contract context and add a regression spec with two consolidated metered contracts that both have usage.

Refactored the event filters to use the contract scope. 
- Removed the contract from_billing_segments!. 
- BillingSegmentsResolver is responsible to group by contract and define the boundaries (from_datetime / to_datetime) per contract
- Changed the queries of codes per contract (recurrence and not recurrence)
- Changed the target_key to use contract id scope. During the billing, the customer can have two different contracts (pricing) to the same product X. Using the previous target_key might create collisions because the timestamps will be put in the the same bucket and `ChargeService` will receive wrong data to filter

tiagolupepic:
> A valid recurring weighted_sum_agg reaches WeightedSumService#latest_value, which calls billing_context.previous_subscription_id?; Billing::Context raises NotImplementedError for contracts, rolling back the invoice group.
> Advance rate cards using latest_agg, max_agg, or weighted_sum_agg are also accepted by current validations, but AggregationFactory explicitly raises during finalization. Restrict unsupported configurations or implement contract-compatible behavior, and cover these cases in the process-service specs.

Both options are going to be addressed soon in another ticket. It's part of `Billing::Context`

brunomiguelpinto (approved):
can you do just a small check 

the `filter_target_for` memoization in `charges_resolver.rb:40-43`. Memoizing by id is only safe when there's one object per id, and here the same recurring charge arrives from two queries with different preloads — `current_recurring_targets` (line 62, `includes(:filters)`) and `targets_with_events` (line 45, `includes(billable_metric: :filters, filters: {values: :billable_metric_filter})`). Since kwargs in `BaseResolver#filter_targets` evaluate left to right, `result: recurring_event_filter_targets` runs first and the poorly preloaded object wins the cache, so `Sources::Charge#filters` returns `charge.filters` without the values and every `to_h_with_all_values` hits the DB — an N+1 on `CalculateFeesService`/`CustomerUsageService`/`PreviewService` that the specs can't catch since it isn't a correctness failure. Just reverting to `main`'s version fixes it;

tiagolupepic:
> the filter_target_for memoization in charges_resolver.rb:40-43. Memoizing by id is only safe when there's one object per id, and here the same recurring charge arrives from two queries with different preloads — current_recurring_targets (line 62, includes(:filters)) and targets_with_events (line 45, includes(billable_metric: :filters, filters: {values: :billable_metric_filter})). Since kwargs in BaseResolver#filter_targets evaluate left to right, result: recurring_event_filter_targets runs first and the poorly preloaded object wins the cache, so Sources::Charge#filters returns charge.filters without the values and every to_h_with_all_values hits the DB — an N+1 on CalculateFeesService/CustomerUsageService/PreviewService that the specs can't catch since it isn't a correctness failure. Just reverting to main's version fixes it;

Fixed using the `compare_by_identity` in hash memoization. Having the same Charge id loaded in recurring or targets_with_events will ended having different FilterTarget objects. 

This doesn't fix the double query in `charges` table and its dependencies, but fixes the initial issue. This problem deserves a proper investigation using data evidences to avoid regressions

PR #6397, opened 2026-09-14
fix(rate-cards): cut rate days in the org's zone

brunomiguelpinto:
A rate on an arrears card is canonicalized to the start of its day, but the day was cut in the application zone rather than the organization's. An organization west of UTC therefore priced from a boundary belonging to someone else.

Measured, for an organization in `America/Sao_Paulo`:

| sent | stored before | stored after |
| --- | --- | --- |
| `2026-12-01T17:00:00Z` | Dec 1 00:00 UTC — Nov 30 21:00 locally | Dec 1 03:00 UTC — Dec 1 00:00 locally |
| `2026-12-01` | Dec 1 00:00 UTC — Nov 30 21:00 locally | Dec 1 03:00 UTC — Dec 1 00:00 locally |

The rest of the codebase already decides the day locally and stores the resulting instant — `Subscriptions::DatesService` writes `in_time_zone(customer.applicable_timezone).beginning_of_day.utc`, and `Billing::Calendar` anchors the same way. A census of every day truncation under `app/` found this the only place cutting a billing day without a timezone.

## Changes

- Read an incoming effective date as belonging to the organization, so a value carrying no offset of its own means that organization's day. A value that names an offset keeps the instant it named. Added as `Utils::Datetime.in_zone`.
- Cut the day in the same zone when canonicalizing an arrears rate.

## Notes for review

- **Both halves are needed together, and either one alone is wrong.** Parsing stays in the application zone until the model runs, so the raw intent of a bare date is already gone by then: cutting the day in the organization's zone at that point moves `2026-09-14` from a western organization onto the *13th* rather than correcting it. Changing only the parse leaves the arrears truncation cutting in the wrong zone. This is why they are one change.
- **Why the organization and not the customer.** A rate card has no customer — it reaches them only through `contract_applied_rate_cards`, a `has_many`, so at write time there may be none or hundreds across different zones, and the column holds one instant. `Organization#timezone` resolves through the default billing entity, which is the same source `Customer#applicable_timezone` falls back to.
- **Advance cards are unaffected by the truncation** — they keep full instants, as before — but they do gain the parse, so a bare date now means the organization's midnight rather than the application's.
- `Organization#timezone` hops to the default billing entity, so the model reads it once per validated rate. That is invisible for ordinary writes; a future bulk rate import should pass the zone in rather than resolve it per record.

## Not in this PR

Existing rows keep the dates they have. Rewriting them would move boundaries on periods already billed, which is a data decision rather than a code one.

## Validation

- 488 RSpec examples passing in the API container, covering the rate card services and models, the REST v2 and GraphQL write paths, the rate queries, contract materialization and the billing date engine.
- New coverage for a bare date, a value with its own offset, a value already carrying a zone, unreadable input, an unknown zone name, and the arrears cut for an organization west of UTC.
- RuboCop passed for all changed Ruby files.


brunomiguelpinto:
Closing: measured, and the change does not alter a single billing outcome.

A rate boundary at UTC midnight and one at the organization's midnight snap to the same cycle boundary through `boundary_at_or_after`, and produce identical proration ratios — `Billing::Days` counts days by local-midnight ownership, and both instants sit in a window containing no local midnight.

So the cost was real and the benefit was not: parsing incoming dates in a business timezone is a pattern that exists nowhere else in this codebase (the v1 engine takes the instant as given and applies the timezone on read), and it added a timezone lookup per record inside a `before_validation`.

The genuine user-facing problem — an organization west of UTC having its own today rejected — is handled in #6395 with a day of slack on the bound, which no timezone on earth can exceed.

PR #6401, opened 2026-09-14
[ING-683] feat(wallets): expose connections

mariohd:
## Context

ING-683. ING-466 (REST) and ING-467 (GraphQL) added the write path for per-object connections on wallets and recurring transaction rules, but nothing reads it back. ING-527 needs that routing to render the wallet's "External apps" tab, and ING-667 covers only the invoice snapshot. Branched from `main` — the read depends on nothing in those two PRs.

## Description

Both surfaces report **every** category with the behaviour the object holds and the code of the connection actually in effect:

```graphql
connections { category behavior code }
# behavior: specific | skip | inherit
```

```json
"connections": {
  "payment":    {"behavior": "specific", "code": "stripe_us"},
  "tax":        {"behavior": "inherit",  "code": "anrok_main"},
  "crm":        {"behavior": "skip",     "code": null}
}
```

`inherit` means the object makes no choice of its own, and the `code` then shows the customer default that applies. That pairing is the point: a client can render the drawer state *and* a connection chip from one payload, without reimplementing the cascade. The REST shape mirrors the object already accepted on create/update.

One derivation feeds both — `ConnectionResolvable#connection_routing`, which loads overrides once rather than per category. `Subscription` includes the same concern, so ING-463's read half is already done at the model layer.

The wallet index eager-loads the overrides plus the customer connection lists the inherited path resolves through; without that the existing N+1 guard fails.

## Two calls worth a reviewer's eye

**Wallet webhooks gain the key.** `wallet.created`, `wallet.updated`, `wallet.terminated` and `wallet.depleted_ongoing_balance` all share `V1::WalletSerializer`, so their payloads now carry `connections` and run the resolution queries. It's additive and consistent with `payment_method`, which is also unconditional and also in those payloads — but the alternative is to gate it behind `include?(:connections)` so only the REST endpoints opt in. Happy to switch.

**The read is not gated behind `multi_connection`**, contrary to the ticket text. The resolution cascade is ungated by design (ING-456 shipped "ungated, no behaviour change"), and with the flag off the read reports the customer default, which is accurate — gating would hide true information rather than protect anything.

## Before merging

The ticket asked for the payload shape to be agreed with FE (ING-527 / ING-528) first, and that hasn't happened. If the tab needs the stored override *separately* from the resolved connection, `ConnectionRouting` changes shape — cheap now, a breaking change once it's merged and codegen'd.

## Testing

481 examples, 0 failures. New coverage: the cascade at model level across wallet, rule and subscription (15), a real GraphQL query on the wallet and its nested rules, the REST show endpoint for inherit / specific / skip, and the rule serializer. The GraphQL query matters specifically — `category` is a String against an enum built from Symbol keys, which would raise at query time if they didn't coerce. The index N+1 guard and the schema-dump spec both pass; Rubocop clean.


app/controllers/concerns/wallet_actions.rb
```diff
@@ -73,9 +73,13 @@ def wallet_index(external_customer_id:, currency:, billing_entity_codes: nil)
             :billing_entity,
             :metadata,
             :billable_metrics,
-            {customer: :billing_entity},
+            :billing_object_connections,
+            {customer: [:billing_entity, :payment_provider_customers, :integration_customers]},
```
lovrocolic, line 77: Do we really need `payment_provider_customers` and `integration_customers`?

app/models/concerns/connection_resolvable.rb
```diff
@@ -30,6 +37,32 @@ def effective_crm_connection
...
+  # loaded once rather than per category, so `includes(:billing_object_connections)` on a
+  # collection keeps this to one query.
+  def connection_routing
+    overrides = billing_object_connections.index_by(&:category)
+
+    CATEGORIES.each_value.map do |category|
+      override = overrides[category]
+
+      connection = if override.nil?
+        customer_default_connection(category)
```
lovrocolic, line 51: Probably we are hitting n+1 here

lovrocolic (approved):

app/models/concerns/connection_resolvable.rb
lovrocolic: If we cannot solve n+1 in this ticket, can we at least skip this logic if feature flag is not turned on and open a new ticket for this specific problem

mariohd: solved!

app/controllers/concerns/wallet_actions.rb
mariohd: addressed! Thanks!

app/models/concerns/connection_resolvable.rb
```diff
@@ -89,9 +89,9 @@ def customer_default_connection(category)
     return unless customer
 
     if category == CATEGORIES[:payment]
-      customer.payment_connection
+      customer.payment_provider_customers.detect(&:is_default?)
     else
-      customer.integration_connection(category)
```
lovrocolic, line 61: This does not work right now since we haven't ran migrations for code and `is_default`. Should we do it before merging?

lovrocolic: Maybe we should fetch first one (if there is only one connection) and use `is_default` in case of more connections

mariohd: Good catch — without the backfill `is_default` is `false` on every row, so the read would return `null` for everyone.

Went with your suggestion: a single connection is taken as the default, and `is_default` only decides when there's more than one. The max-1-per-type cap is still in place (model validation + unique index), so before the backfill there's always exactly one candidate; after it, the flag takes over. No migration needed.

Several connections with none flagged still resolve to `nil` instead of picking one arbitrarily — same as the previous `find_by`.

PR #6402, opened 2026-09-14
fix(payments): cancel abandoned 3DS challenges

brunomiguelpinto:


app/models/payment.rb
```diff
@@ -36,6 +49,12 @@ class Payment < ApplicationRecord
 
   delegate :billing_entity, to: :customer
 
+  scope :abandoned_at_authentication, lambda {
+    where(payable_type: "Invoice", status: "requires_action", payable_payment_status: :processing)
+      .where(created_at: ..AUTHENTICATION_ABANDON_PERIOD.ago)
```
tiagolupepic: This uses the row's `created_at` as the start of the 3DS abandonment window, but payment rows can be reused across attempts while they are still in `payable_payment_status = pending` (`Invoices::Payments::CreateService` does a `find_or_create_by!` on that status before Stripe updates the row). If an old pending row is retried today and only then moves to `requires_action`, the hourly scan will treat it as already older than 24h and cancel a fresh authentication challenge almost immediately. The cutoff should be based on when the payment entered `requires_action` (for example the `updated_at` from that transition or a dedicated timestamp), or the retry path needs to create a fresh row for the challenge.

brunomiguelpinto:
Superseded by #6416, which rebuilds this from the investigation rather than from the first diagnosis.

What carried over: the abandonment window measured from `updated_at` rather than `created_at` (thanks @tiagolupepic — the reused-pending-row case now has a regression spec), skipping payments that gate a subscription activation, and the payment factory's next action carrying a `type`.

What changed, and why:

- **Only `redirect_to_url`.** Classifying every stuck intent against Stripe found no `use_stripe_sdk` at all, and the stored next action matched the live one every time — the staleness this assumed does not show up.
- **No separate check service.** `PaymentProviders::Stripe::Payments::RetrieveService` just reads the intent; whether a failure means "do nothing" or "retry" is decided by the service choosing to cancel. Permanent errors are swallowed, rate limits and dropped connections reach the job's backoff — swallowing everything meant a rate limit silently skipped a payment.
- **Scope names describe the redirect**, not an "authentication challenge" we cannot prove from local data.
- **Payments whose provider was deleted are excluded when selecting**, since nothing can cancel them and they would otherwise be retried hourly forever.
- **No index.** Whether the scan needs one is measurable and has not been measured.

Closing this one.

PR #6403, opened 2026-09-14
feat(catalog): enforce a slug-safe code on v2 catalog objects

rsempe:
## Context

No Lago model validates code format, so every code-keyed v2 route carries a greedy `code: /.*/` constraint. Verified consequence: a product coded `something/filters` is unreachable on member routes because `GET /api/v2/products/something/filters` resolves to the *filters index* of product `something`; other slashed codes shadow-collide with current or future nested segments. The catalog tables are empty in production until v2 ships — **the only window** to add a format rule for free.

## Description

- New `CatalogCodeFormat` concern (`validates :code, format: {with: CODE_FORMAT}, allow_blank: true`) included in **every code-keyed v2 catalog model**: Product, ProductCategory, ProductFilter, RateCard, RateCardRate, **CatalogPlan, RatePhase**.
- `allow_blank` leaves the empty case to each model's existing presence validation (no double error).
- The failure is a standard `code` field error (`value_is_invalid`), so REST and GraphQL surface it on the field the caller sent.
- **No data migration** — tables are empty.


endenis (approved):

spec/support/shared_examples/catalog_code_format.rb
```diff
@@ -0,0 +1,33 @@
...
+    it "allows a slug-safe code" do
+      record = build(factory, code: "valid_code-1.2")
+      record.valid?
+      expect(record.errors.where(:code, :invalid)).to be_empty
+    end
+
+    it "rejects a code containing a slash" do
+      record = build(factory, code: "a/b")
+      record.valid?
+      expect(record.errors.where(:code, :invalid)).to be_present
```
tiagolupepic, line 16: I think this could be more sensitive and validate any error in `code`. `record.errors[:code]` it does validate any error and not only the invalid one.

app/models/concerns/catalog_code_format.rb
```diff
@@ -0,0 +1,19 @@
...
+# member-route params, so restricting them to a URL-safe charset keeps every
+# code addressable and avoids collisions with nested route segments. The
+# leading negative lookahead rejects an all-dot code (`.`, `..`, …): those are
+# path-segment specials that clients and proxies normalize away, so they are
+# not reliably addressable. allow_blank leaves the empty case to the presence
+# validation.
+module CatalogCodeFormat
+  extend ActiveSupport::Concern
+
+  CODE_FORMAT = /\A(?!\.+\z)[a-zA-Z0-9_\-.]+\z/
```
tiagolupepic, line 14: Image

these options with many dots and text are allowed as well?

tiagolupepic (approved):
minor comments

:shipit:

app/models/concerns/catalog_code_format.rb
rsempe: for now.. yes.. we'll see with the product team

PR #6407, opened 2026-09-15
feat(contracts): expose billing, invoicing & payment settings

rsempe:
## Context

The finalized Create/Edit contract design has **Contract / Invoicing / Payment settings** sections, but the create/update mutations only accepted plan/dates/name — even though the model columns already exist.

## Description

Exposed on contract **create & update** (GraphQL inputs + services) and on the **`Contract` object** read side:

- **Billing entity** — `billingEntityId` (ID), an explicit override; omitted inherits the customer's (`applicable_billing_entity`), matching the invoices/quotes/subscriptions/wallets convention.
- **Invoicing settings** — `consolidateInvoice`, `purchaseOrderNumber`.
- **Payment settings** — `paymentMethodId` (scoped to the contract's customer) and `paymentMethodType` (reuses `PaymentMethodTypeEnum` — provider/manual).

The two NOT NULL columns (`consolidate_invoice`, `payment_method_type`) keep their DB defaults when omitted. An unknown `billingEntityId` or `paymentMethodId` returns a not-found failure. Update remains pending-only (unchanged lock).


app/graphql/types/contracts/object.rb
```diff
@@ -22,6 +22,17 @@ class Object < Types::BaseObject
       field :started_at, GraphQL::Types::ISO8601DateTime, null: true
       field :terminated_at, GraphQL::Types::ISO8601DateTime, null: true
 
+      # Settings. billing_entity and payment_method are the explicitly-set
+      # overrides (nil = inherit from the customer); the others carry the
+      # stored value.
+      field :billing_entity, Types::BillingEntities::Object, null: true
```
tiagolupepic, line 28: **N+1 on contract lists.** `billing_entity` (line 28) and `payment_method` (line 32) are `belongs_to` associations exposed as plain fields with no batching. Contracts are served as a paginated collection via `ContractsResolver` (`type Types::Contracts::Object.collection_type`), so querying either of these fields on a list fires one query per contract — O(n) queries. This file already avoids exactly this for its other associations with `dataload_association :customer` (line 9) and the batched `plan` loader. Suggest adding `dataload_association :billing_entity, :payment_method` for consistency.

app/services/contracts/update_service.rb
```diff
@@ -87,6 +105,23 @@ def catalog_plan
       @catalog_plan = params[:plan_code].present? ? organization.catalog_plans.find_by(code: params[:plan_code]) : nil
     end
 
+    def billing_entity
```
tiagolupepic: **Duplicated Code across sibling services.** `billing_entity` (108), `payment_method_params` (115) and `payment_method` (119) are byte-for-byte identical to the same private methods in `Contracts::CreateService` (lines 113/120/125), and the two `not_found_failure!` guards + the `result.payment_method = payment_method` / `PaymentMethods::ValidateService` block are duplicated in both `#call`s as well. The same setting-assignment rules are also expressed twice but *differently* — create bundles them in `optional_settings`, update inlines them (lines 70‑73). This is the sibling-service duplication the repo tends to extract (cf. the shared `CustomerTimezone` concern already included here): consider a shared `Contracts::` concern or model-level resolvers so the rule lives in one place.

tiagolupepic (approved):
:shipit:

PR #6408, opened 2026-09-15
feat(usage): add the usage buckets table [ING-586]

vincent-pochet:
## Context

The RisingWave pipeline pre-aggregates events into 15-minute usage buckets and upserts them into
ClickHouse. The API has no table to receive them and no model to read them, so the pipeline has
nowhere to write and the read path cannot be built on top.

## Description

- Add `usage_buckets_15m` as a self-hosted ClickHouse migration and as the ClickHouse Cloud DDL.
- Partition by month of the bucket, so merges never range over the whole history and retention is a
  `DROP PARTITION` rather than a TTL mutation.
- Version the `ReplacingMergeTree` on a materialized `ver`, so the sink can re-upsert a bucket
  freely and the row written last wins.
- Add `Clickhouse::UsageBucket`, read-only because the sink owns every write, reading with `FINAL`
  by default.

15 minutes is the granularity that makes every timezone's day boundary land on a bucket wall, since
every real UTC offset is a multiple of 15 minutes. `FINAL` is not optional: versions collapse only
on merge, so a plain read double-counts a bucket the sink has just re-upserted.

Nothing reads the table yet, so this changes no behaviour.


toommz (approved):

db/clickhouse_migrate/cloud/10_usage_buckets_15m.sql
```diff
@@ -0,0 +1,24 @@
...
+(
+    `bucket` DateTime64(3),
+    `organization_id` String,
+    `subscription_id` String,
+    `customer_id` String,
+    `plan_id` Nullable(String),
+    `code` String,
+    `target_wallet_code` Nullable(String),
+    `charge_id` String,
+    `charge_filter_id` String,
```
tiagolupepic, line 11: we're adding product and plans and soon we need to include the product_id and product_filter_id in this table. Any chance to make this polymorphic already? This will facilitate our migration

PR #6409, opened 2026-09-15
perf(streaming): produce current usage from the wallet refresh

toommz:
The wallet refresh already computes current usage, then `DeliverEventJob` computed it again in another process. It now reuses what it has and produces inline when running on a worker, falling back to the job otherwise. Full usage is unchanged.

A spec asserts the envelope is identical either way.


toommz:
> 
> HOLD: `Sidekiq.server?` does not establish that the process has the streaming worker's AWS identity.
> 
> * `RefreshWalletJob` runs on `wallets`, `low_priority`, or `dedicated_wallets`, while only the `streaming` worker is identity-bearing; use an explicit capability guard (or keep delivery on `DeliverEventJob`) and test the actual wallet-worker path so hot wallet jobs do not probe STS once per destination/process before falling back.

I added a retry logic, if the current pod is not allowed to use Kinesis it will raise, catch the error and enqueue a job on the right pod.

app/services/customers/refresh_wallets_service.rb
```diff
@@ -48,13 +48,63 @@ def call
     attr_reader :customer, :include_generating_invoices
 
     def deliver_streaming_events
+      streamed_event_types.each do |event_type|
+        if event_type == StreamingDestinations::BaseDestination::CURRENT_USAGE_EVENT_TYPE && produce_inline?(event_type)
```
tiagolupepic, line 52: it's doing a N+1 in produce_inline?(event_type) method, no?

app/services/customers/refresh_wallets_service.rb
```diff
@@ -48,13 +48,70 @@ def call
...
+    end
+
+    # Only a worker holds the AWS identity the producer needs, and only until it proves otherwise:
+    # a process that has already failed to obtain credentials stops paying the STS timeout and
+    # hands the delivery to the streaming worker instead.
+    def produce_inline?(event_type)
+      return false unless Sidekiq.server?
+
+      # An event type is claimed by a single destination, so the first match is the only one.
+      destination = active_destinations.find { it.event_types.include?(event_type) }
```
tiagolupepic: I know you have a few events for now and this is not problematic. You can avoid such O ( N x M ) in memory using a `Set` O(1)

tiagolupepic (approved):
one last comment and I'd address before merge it. In the future if we start having many events it will slow down the refresh wallet. 

:shipit:

PR #6411, opened 2026-09-15
feat(streaming): report usage per wallet instead of naming one

toommz:
`wallet_id` named a single wallet picked by currency, which could not describe a customer whose charges draw on several.

It is replaced by a `wallets` array carrying each wallet's own share of the ongoing usage, as the wallet refresh allocated it, so wallet targets, thresholds and the cascade are already applied. The top-level `credits` goes with it, since it was that arbitrary wallet's rate.

The figures are customer scoped, not per subscription.


toommz:
> 
> HOLD — This makes an incompatible schema change to both existing `.v1` streaming events.
> 
> * `CustomerUsageSerializer` removes `credits` and `wallet_id`, and the spec explicitly requires their absence. Preserve those fields while adding `wallets`, or publish and wire a new event version; otherwise existing consumers will break.

The contract never was consumed or used, it's still beta, so this is v1 in fact atm.

tiagolupepic (approved):
:shipit:

PR #6412, opened 2026-09-15
feat(streaming): omit charges with no usage from the payload

toommz:
A plan carries a charge per feature, so a customer using a handful of them shipped a long tail of zeroes on every record. They are now omitted, which buys headroom under the Kinesis 1MB cap.

A charge is kept when it has units or an amount, so a free allowance or a charge billed without units still appears.


app/serializers/event_destinations/customer_usage_serializer.rb
```diff
@@ -38,14 +38,24 @@ def wallets_usage
       end
     end
 
+    # A plan carries a charge per feature, so a customer using a handful of them would otherwise
+    # ship a long tail of zeroes on every record. Dropping them keeps headroom under the 1MB cap.
     def charges_usage
-      model.fees.group_by(&:charge_id).map do |_charge_id, fees|
+      model.fees.group_by(&:charge_id).filter_map do |_charge_id, fees|
```
tiagolupepic: you can do this using a single loop using the non_zero? method https://github.com/getlago/lago-api/blob/main/app/models/fee.rb#L319-L321.

tiagolupepic (approved):
:shipit:

PR #6414, opened 2026-09-15
refactor(billing): Use metered items for pay-in-advance

tiagolupepic:
## Context

Pay-in-advance jobs and services reconstructed charge, event, boundary, and filter state independently. That made the flow difficult to extend to billing segments and left ActiveJob payload compatibility scattered across jobs.

## Description

- Make MeteredItem the canonical input for pay-in-advance aggregation, fee creation, and invoice charge creation.
- Add PayInAdvanceArguments to bridge legacy serialized charge/event arguments while preserving job lock keys.
- Add an ActiveJob serializer for Charge and BillingSegment sources, including source-specific attributes and event round trips.
- Require MeteredItem events to be wrapped in Events::Common and resolve event-specific boundaries and pricing buckets.
- Add schedule segment lookup and update service, scenario, and focused specs.

Next steps are to pass BillingSegment through the full pay-in-advance flow, introduce the BillableSegment source where needed, and cover segment-specific boundary, filtering, and fee behavior.


app/jobs/fees/create_pay_in_advance_job.rb
```diff
@@ -22,22 +22,29 @@ def self.retry_wait(executions)
 
     unique :until_executed, on_conflict: :log
 
-    def perform(charge:, event:, billing_at: nil)
-      result = Fees::CreatePayInAdvanceService.call(charge:, event:, billing_at:)
+    def perform(charge: nil, metered_item: nil, event: nil, billing_at: nil)
```
tiagolupepic, line 25: Keeping the charge and event here to avoid break jobs during the deployment. Persisted job schemas in Redis will contain only charge + event, if keep only metered_item, the deployed job will not know how to parse them.

app/jobs/serializers/metered_item_serializer.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module ActiveJob
+  module Serializers
+    class MeteredItemSerializer < ObjectSerializer
```
tiagolupepic: ActiveJob doesn't know how to decode Data (PORO) classes.

app/jobs/serializers/metered_item_serializer.rb
```diff
@@ -0,0 +1,56 @@
...
+          {
+            "source_type" => "billing_segment",
+            "billing_segment" => source.billing_segment,
+            "product_filter" => source.product_filter
+          }
+        else
+          raise ArgumentError, "Unsupported MeteredItem source: #{source.class}"
+        end
+
+        super(payload.merge("event" => metered_item.event&.as_json))
```
tiagolupepic: [P1] Serialize nested records and boundary values through ActiveJob

`ObjectSerializer#serialize` only adds the serializer marker; it does not recursively serialize the payload. This leaves raw ActiveRecord objects (`charge`, filters, and billing segments) and time values inside the job arguments. After Sidekiq's JSON round trip, these are no longer the model/time instances that `deserialize` passes directly into the sources, so newly enqueued pay-in-advance jobs cannot execute correctly. Serialize and deserialize the nested values through `ActiveJob::Arguments` so records use GlobalIDs and boundary types survive. The round-trip spec should include JSON encoding/decoding between serialization and deserialization; its current in-memory round trip retains the original Ruby objects and misses this failure.

app/services/fees/charge_service/sources/charge.rb
```diff
@@ -37,7 +37,14 @@ def filter_association
           :charge_filter
         end
 
-        def pricing_buckets
+        # Pay-in-advance events must use only their matching filter; returning every
+        # configured filter would create duplicate fees for a single event.
+        def pricing_buckets(event: nil)
+          if event
+            matching_filter = ChargeFilters::EventMatchingService.call(charge:, event:).charge_filter
+            return [with_filter(matching_filter)]
```
tiagolupepic: [P1] Preserve filter exclusions for the default pricing bucket

When a charge has configured filters but the event matches none, `matching_filter` is `nil`. Returning `with_filter(nil)` makes `PayInAdvanceAggregationService#aggregation_filters` skip computing `matching_filters` and `ignored_filters`. Previously, `init_charge_filter_fee` supplied an unsaved `ChargeFilter` for this case, which excluded events belonging to the configured buckets. Without those exclusions, default-bucket aggregation and free-unit calculations can include usage from separately priced buckets, changing fees and estimates. Use `with_default_filter` when configured filters exist but none matches, and cover a sequence containing both matching and unmatched events.

app/jobs/serializers/metered_item_serializer.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      def deserialize(payload)
+        event = payload["event"] && Events::CommonFactory.new_instance(source: payload["event"])
+
+        case payload["source_type"]
+        when "billing_segment"
+          Fees::ChargeService::MeteredItem.from_billing_segment(payload["billing_segment"], event:)
```
tiagolupepic: [P2] Restore the billing segment's selected product filter

The serializer writes `source.product_filter`, but this reconstruction ignores it: `from_billing_segment` initializes the source with `product_filter: nil`. Consequently, a metered item selected with `with_filter(...)` loses its filter even in an in-memory round trip, and subsequent `matching_and_ignored_filters` resolves the empty/default filter instead. Restore the serialized product filter when rebuilding the source and add a round-trip example with a non-nil selected filter; the existing example only exercises the nil case.

brunomiguelpinto (approved):
One operational note for the deploy, not a change request. This release both introduces ActiveJob::MeteredItemSerializer and switches the enqueue side to it:

- Fees::CreatePayInAdvanceJob.perform_later(charge:, event: event.as_json)
+ Fees::CreatePayInAdvanceJob.perform_later(metered_item: metered_item_for(charge))

On main the worker is perform(charge:, event:, billing_at: nil) — both required. So during the rolling window, a job enqueued by an updated pod and picked up by a worker that has not rolled yet fails twice over: safe_constantize cannot find ActiveJob::MeteredItemSerializer, and even past that the keywords are missing. ApplicationJob sets sidekiq_options retry: 0, so those go straight to the morgue. Same for Invoices::CreatePayInAdvanceChargeJob.

tiagolupepic:
@brunomiguelpinto regarding your comment. 

It's addressed already. Check this comment https://github.com/getlago/lago-api/pull/6414#discussion_r4016753932.

For now, the job is make `charge` and `event` optional to support the rollout process.

PR #6416, opened 2026-09-15
fix(payments): cancel abandoned provider redirects

brunomiguelpinto:
## Context

When a provider sends an end customer off site to complete a payment and they never come back, the intent stays awaiting that customer forever. Stripe does not expire a PaymentIntent created through its API — Checkout Sessions expire on their own, this path does not use them — so the only reset is a webhook that requires the merchant to cancel, by hand, an intent Lago created and owns.

The invoice is then held at `ready_for_payment_processing = false`, which refuses retries, dunning, payment requests and the Pay invoice button at once. And `index_payments_on_payable_id_and_payable_type` is unique over pending and processing payments, so the stuck row also occupies the invoice's only payment slot: unlocking alone would just make the next charge collide with the index. Cancelling the payment is the only way back.

## Scope

A payment is recovered once the customer has had a day to come back, and only while it went stale within the last month. Everything older is left untouched. -> this will be address after this pr 

So: nothing new joins the backlog from here on, and what is already in it comes out separately — **including the invoices that prompted the report**. Releasing those needs the affected accounts asked first, and is tracked apart from this PR.

## Description

An hourly clock job hands abandoned payments to a service that cancels them at the provider and lifts the lock. It reuses `PaymentProviders::CancelPaymentService`, whose Stripe branch already cancels with `cancellation_reason: :abandoned` and already treats `payment_intent_unexpected_state` as a no-op, so the cancel call is its own atomic state check. The definitive status is left to Stripe's webhook; this only lifts the lock, so that a webhook which never arrives cannot leave the invoice stuck.


app/services/invoices/payments/cancel_abandoned_service.rb
```diff
@@ -0,0 +1,88 @@
...
+          "#{payment.provider_payment_id} for payment #{payment.id}: #{e.class.name}"
+        )
+        false
+      end
+
+      def abandoned_at_redirect?
+        payment.status == "requires_action" &&
+          payment.processing? &&
+          payment.provider_payment_data&.dig("type") == "redirect_to_url" &&
+          payment.updated_at.between?(RECOVERY_WINDOW.ago, ABANDONED_PERIOD.ago)
```
rsempe: `updated_at` is when the row was last written, not when the customer was redirected, can it be an issue?

app/services/invoices/payments/cancel_abandoned_service.rb
```diff
@@ -0,0 +1,88 @@
...
+      # The local columns cannot tell a card apart from a redirect-based alternative method: the
+      # payment method is unknown on most rows, and `provider_payment_method_data` is only ever
+      # written by the `payment_intent.succeeded` webhook, so a payment stuck before success has it
+      # empty by construction. Only cards do 3DS, so the provider is asked before anything is
+      # cancelled. The answer also catches an intent that moved on without us.
+      def confirmed_by_provider?
+        return false unless payment.payment_provider_type == "stripe"
+
+        live = ::PaymentProviders::Stripe::Payments::RetrieveService.call!(payment:)
+        live.status == "requires_action" && live.payment_method_type == "card"
```
rsempe: generated by AI:

This excludes the case the Context names as the motivation: merchant cancelled the intent by hand, webhook never arrived. Stripe says `canceled`, so `cancellable?` returns false and the invoice stays locked — and since the local row still looks `processing`/`requires_action`, the clock re-selects it every hour and burns a `PaymentIntent.retrieve` per payment indefinitely, with no path out.

`RetrieveService` already has the live status: treating `canceled` / `requires_payment_method` as grounds to sync the row and unlock would close that loop.

brunomiguelpinto: Goood catch changing it

app/jobs/clock/cancel_abandoned_payments_job.rb
```diff
@@ -0,0 +1,36 @@
...
+            .perform_later(payment)
+        end
+      end
+    end
+
+    private
+
+    # Narrowed to what the service can ever act on
+    def candidates
+      service = Invoices::Payments::CancelAbandonedService
```
tiagolupepic: nitpick.. you can have Invoices::Payments::CancelAbandonedService.range or window_range and build the range inside of such class instead.

app/services/invoices/payments/cancel_abandoned_service.rb
```diff
@@ -0,0 +1,91 @@
...
+        return false unless payment.payment_provider_type == "stripe"
+        return false unless abandoned_at_redirect?
+        return false if payable.payment_succeeded? || payable.voided? || payable.closed?
+        # Cancelling would land a failed payment status on the invoice, and that resolves the
+        # activation through Invoices::UpdateService. That flow has its own window and its own clock.
+        !payment.gated_subscription_activation?
+      end
+
+      def abandoned_at_redirect?
+        payment.status == "requires_action" &&
```
tiagolupepic: I'd move this logic to a method in payment instead. Something like `payment.abandoned_on_redirect?(....)` you can wrap all the logic in the model and with unit testing validate each state.

app/jobs/clock/cancel_abandoned_payments_job.rb
brunomiguelpinto: yup

app/services/invoices/payments/cancel_abandoned_service.rb
brunomiguelpinto: sounds good

app/services/invoices/payments/cancel_abandoned_service.rb
brunomiguelpinto: The write that sets `requires_action` is the one that comes back from Stripe, which is why `updated_at` is the closest thing 
created at we didn't even reach stripe

app/services/invoices/payments/cancel_abandoned_service.rb
brunomiguelpinto: Fixed. When the provider reports a status in its own `FAILED_STATUSES`, the row is brought in line with what it says and the invoice released — no cancellation attempted, and the loop closes instead of burning a retrieve every hour.

The webhook machinery is deliberately not reused for it. `update_payment_status` routes through `authentication_retry_pending?`, which holds the invoice back whenever the error code is an authentication failure and the provider supports 3DS — true of this entire population by definition. Going through it would have left every one of these invoices exactly as locked as before.

Three specs cover it, including that a payment the provider reports as `succeeded` is left alone: only its webhook can settle a payment.

tiagolupepic (approved):
minor comments

:shipit:

app/services/invoices/payments/cancel_abandoned_service.rb
brunomiguelpinto: Didn't move it — this is only true for Stripe. For the other providers the same check has no meaningful answer, so on `Payment` it wouldn't make sense.

brunomiguelpinto: didnt move because this is only true for stripe for other providers this will not be the same answer no in the model would not make sense

rsempe (approved):

PR #6419, opened 2026-09-16
feat(contracts): add list filters and search

rsempe:
## Context

The contract list only filtered by customer, plan, external id and status. The product needs to filter by **billing entity** and by whether a contract **has rate overrides**, and to **search** contracts the same way subscriptions are searched.

## Description

Extends `ContractsQuery` with three inputs, exposed on **both** the GraphQL resolver and the REST `index` so the single query serves both surfaces:

- **`billing_entity_ids`** — matches a contract's own billing entity or the one inherited from its customer (the same fallback `SubscriptionsQuery` uses).
- **`has_rate_overrides`** — selects contracts by whether any phase of any attached rate card carries a rate override.
- **search** — mirrors the subscription free-text search: a `UNION` of single-table branches over the contract name + external id, the plan name + code, the customer, and the contract id when the term is a UUID, so each branch uses its own trigram index instead of a cross-table `OR`.

Adds the supporting `gin_trgm` indexes on `contracts` (org_id, name / external_id) and `catalog_plans` (org_id, name / code) so every search branch is indexed rather than a sequential scan.


annvelents (approved):

tiagolupepic (approved):
:shipit:

PR #6421, opened 2026-09-16
feat(account-tree): add attribution_labels to events_enriched

lovrocolic:
## Context

Account tree feature allows tracking usage under custom types like teams or users.

## Description

This PR adds the `attribution_labels` column to the ClickHouse `events_enriched` table.

**NOTE**: The column is storage only, so the read path can be built and tested against it. Declaring the field on the Kafka queue table and projecting it in the materialized view lands with the events-processor change that starts emitting it.


toommz (approved):

db/clickhouse_migrate/cloud/02_events_enriched.sql
```diff
@@ -10,7 +10,8 @@ CREATE TABLE default.events_enriched
     `enriched_at` DateTime64(3) DEFAULT now64(3),
     `value` Nullable(String),
     `decimal_value` Nullable(Decimal(38, 26)) DEFAULT toDecimal128OrZero(value, 26),
-    `precise_total_amount_cents` Nullable(Decimal(40, 15))
+    `precise_total_amount_cents` Nullable(Decimal(40, 15)),
+    `attribution_labels` Map(String, String)
```
vincent-pochet: Since the table already exists in production, I think you should provide an update migration instead of updating the creation script itself

vincent-pochet (approved):

db/clickhouse_migrate/cloud/02_events_enriched.sql
lovrocolic: Updated

vincent-pochet (approved):

PR #6422, opened 2026-09-16
feat(fees): support products and plans pay-in-advance

tiagolupepic:
## Context

Pay-in-advance events only handled legacy charges. Collecting product-catalog billing segments therefore had no standalone fee or invoice processing path.

## What Changed

- Add the collecting billing-segment status and its database enum value.
- Resolve eligible catalog billing segments during event post-processing and pay-in-advance execution.
- Dispatch each metered item to the standalone-fee or invoice job according to its invoiceability.
- Build segment-backed pay-in-advance fees from a contract billing context, including product, rate-card, filter, currency, and tax inputs.
- Make billing-segment MeteredItem construction keyword-only and update serialization and callers.
- Add coverage for resolution, dispatch, contract-backed fees, and local and provider tax application.

### Billing Segment Eligibility

Only collecting segments for live contracts, advance metered rate cards, the event billable metric, and the event time window are processed. Product-catalog processing remains disabled unless the organization feature flag is enabled.

### Contract-Backed Fees

Billing-segment fees do not require a legacy subscription or charge. They use Billing::Context from the segment contract, preserve product pricing metadata, and pass the context customer to tax services.

### Legacy Charge Support

Legacy charge-backed pay-in-advance behavior remains on the same dispatch flow and continues to skip events without a subscription billing context.

## New Abstractions

Events::PayInAdvanceBillingSegmentResolver centralizes the catalog-segment eligibility query so post-processing and job dispatch use identical selection rules.


app/services/events/pay_in_advance_billing_segment_resolver.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    private
+
+    attr_reader :event
+
+    delegate :billable_metric, :organization, to: :event
+
+    def billing_segments
+      BillingSegment.status_processing
```
tiagolupepic, line 29: @brunomiguelpinto we need to persist the `BillingSegment` with status `processing` in the Schedule during the `ScheduleProcess`. This applies only for advance billing_timing. 

Having this option, we avoid having an aditional job/service just to change the status from pending -> processing.

brunomiguelpinto: yes i agree

PR #6427, opened 2026-09-17
feat(usage): read the pre-aggregated buckets [ING-586]

vincent-pochet:
## Context

The realtime usage pipeline writes 15-minute buckets of pre-aggregated usage into ClickHouse, and
`Events::Stores::UsageBucketSet` already knows how to answer an aggregation out of them. Nothing
fills that set, so the buckets are written and never read.

## Description

- Add `RealtimeUsage::FetchBucketsService`, reading one subscription's buckets over one window in a
  single grouped query and building the set out of them.
- Serve a window only when it opens on a bucket boundary or on `subscription.started_at`, and refuse
  a truncated one.
- Return no set when the ClickHouse read fails, leaving the caller to read events.

A window opening mid-bucket would count the whole bucket, except on the very first period, where the
pipeline attributes no event to the subscription before `started_at`. The rescue is scoped to the
read alone: the error list is wide enough to swallow a bug of our own anywhere else.

The service has no caller yet. The store that answers from the set, and the routing that reaches
for it, are the next two PRs of the stack.


app/services/realtime_usage/fetch_buckets_service.rb
```diff
@@ -0,0 +1,130 @@
...
+    # to the read rather than to the whole service, as `READ_ERRORS` is wide enough to swallow a
+    # bug of our own anywhere else.
+    def rows
+      return @rows if defined?(@rows)
+
+      @rows = Events::Stores::Utils::ClickhouseConnection.with_retry { fetch_rows }
+    rescue *READ_ERRORS => e
+      Rails.logger.warn("Realtime usage bucket prefetch failed: #{e.class} #{e.message}")
+      Sentry.capture_exception(e)
+      @rows = nil
```
toommz:
```suggestion
      @rows = []
```

As we will try to iterate on it later 🤔

toommz (approved):

app/services/realtime_usage/fetch_buckets_service.rb
vincent-pochet: Good catch indeed.
I updated the service to return an error in this case. The idea will be to fall-back on the "classic aggregation" if we do not manage to fetch the data from clickhouse

PR #6431, opened 2026-09-17
[ING-511] feat(alerts): accept and return notify_on over REST (merge after #6373)

aquinofb:
## Context

Thresholds have carried `notify_on` since the columns landed, and both the dashboard schema and the validation rules followed. REST never accepted it: strong parameters dropped the key before it reached the service, so a customer had no way to opt a threshold in to resolved notifications, and the rules that were already written could never fire on the surface customers actually integrate with.

## Changes

- `notify_on` is permitted on create, update and batch, for both the wallet and the subscription alert endpoints.
- Reads return it alongside the rest of the threshold, so a client can confirm what it set rather than writing blind. The same serializer feeds the `alert.triggered` webhook.

No new validation. The rules shipped with the dashboard schema and live in the shared service, so opening REST is what finally lets them run: `triggered` is always required, values must be known transitions, and `resolved` needs a unique code and is refused on recurring thresholds.

## Compatibility

Omitting the field behaves exactly as before and falls to the column default. The one behaviour change for an existing caller is that a payload carrying an unknown `notify_on` value used to be silently stripped and now returns a validation error, which is the point of accepting the field.

## Not in scope

Threshold and alert `state` on reads, the other half of ING-511. It depends on the definition of `stale`, which is open on ING-546 because wallets have no billing period. Splitting it keeps this PR to the opt-in itself.


toommz (approved):

aquinofb:
**Do not merge this before #6373.**

This is the launch switch. It is the one PR in the set that is not inert: once it ships, a customer can `POST` a threshold with `notify_on: ["triggered", "resolved"]` and get a 200 back.

Checked against `main` today, nothing can honour that yet:

- `alert.resolved` is not in `config/webhook_event_types.yml`
- no `alert_resolved` webhook service and no `resolved_alert` serializer exist
- the only trace of recovery is the `fully_resolved` column annotation on `TriggeredAlert`; the detection engine is still in #6131

So merging this first lets a customer opt in to a feature that silently does nothing, with no error to tell them, until the rest of the chain lands.

Safe order: #6130, then #6131, then #6373, then this one. Everything before it is unreachable by customers by construction, which is why the ordering only matters here.

PR #6432, opened 2026-09-17
feat(catalog): add applied rate card mutations

rsempe:
## Context

The contract overview's rate cards tab has add and remove actions with no GraphQL backing: only `createPlanAppliedRateCard` existed, with no contract equivalent and no destroy mutation for either plan or contract applied rate cards.

## Description

Adds `createContractAppliedRateCard`, `destroyContractAppliedRateCard`, and `destroyPlanAppliedRateCard` for plan parity, each wiring an existing service so the immutability rules hold unchanged: cards attach to or leave only a pending contract, and a plan card is removable only while the plan has no contracts. The contract mutation is keyed by `externalId` and resolves the editable contract, like `updateContract`; both destroys are organization-scoped.


tiagolupepic (approved):
:shipit:

PR #6438, opened 2026-09-18
refactor(billing): use advance billing contexts

tiagolupepic:
## Summary

- pass `Billing::Context` objects through advance-charge billing instead of exposing subscription-specific inputs
- return pending billing contexts with successful unattached fees directly from `AdvanceChargesService`
- rename `CreateAdvanceChargesInvoiceSubscriptionService` to `CreateAdvanceChargesInvoiceService`
- expose purchase order numbers and plan IDs through `Billing::Context`; contract contexts return `nil` for `plan_id`
- resolve currency, billing entity, plan, and invoice grouping data through billing contexts
- stop persisting `skip_charges: true` on advance-charge invoices

## Billing context flow

Advance-charge invoice creation now receives `billing_contexts_with_fees` and `all_billing_contexts`. Contexts are selected and grouped before reaching the creation service, so the service no longer accepts subscription collections. It uses context attributes for organization and subscription identity and unwraps `billing_context.subscription` only when the existing subscription date service is needed to calculate boundaries.

## Why `skip_charges` can be removed here

`skip_charges` remains necessary for subscription activation and free-trial billing. In those flows, `Invoices::CalculateFeesService` reads the flag to suppress invoiceable charge fees and recurring non-invoiceable advance-charge fees.

`Invoices::AdvanceChargesService` does not call `CalculateFeesService`. It creates an invoice, attaches existing successful standalone fees, aggregates their amounts, and immediately finalizes or closes the invoice. No fee-generation path reads the flag.

The other readers do not affect advance-charge invoices:

- draft refresh only supports subscription invoices
- generating-invoice retry only selects subscription invoices
- GraphQL `acceptNewChargeFees` remains false because advance-charge invoice subscriptions use `in_advance_charge_periodic`, rather than `subscription_periodic` or `subscription_terminating`

Removing the argument therefore only leaves the persisted column at its default `false`; it does not generate additional charges. The service spec continues to assert the exact number and amounts of the attached fees.


app/services/invoices/advance_charges_service.rb
```diff
@@ -4,21 +4,21 @@ module Invoices
   class AdvanceChargesService < BaseService
     Result = BaseResult[:invoice]
 
-    def initialize(initial_subscriptions:, billing_at:)
-      @initial_subscriptions = initial_subscriptions
+    def initialize(billing_contexts:, billing_at:, metered_items: [])
```
rsempe, line 7: `metered_items` is assigned and exposed but not used, right?

app/services/invoices/create_advance_charges_invoice_service.rb
```diff
@@ -17,13 +17,13 @@ def initialize(invoice:, timestamp:, subscriptions_with_fees:, all_subscriptions
...
+      boundaries = calculate_boundaries(latest_billing_context)
 
-      subscriptions_with_fees.each do |subscription|
+      billing_contexts_with_fees.each do |billing_context|
         invoice.invoice_subscriptions << InvoiceSubscription.create!(
-          organization: subscription.organization,
+          organization: billing_context.organization,
           invoice:,
-          subscription:,
+          subscription_id: billing_context.subscription_id,
```
rsempe: Passing the id instead of the object leaves the association cold, and the caller does `is.subscription.fees` right after — one extra query per invoice_subscription. `subscription: billing_context.subscription` avoids it.

app/services/invoices/advance_charges_service.rb
tiagolupepic: yes. Not used yet. it will be used with contracts (in another PR).

brunomiguelpinto (approved):

PR #6439, opened 2026-09-18
refactor(usage): serve every current usage window [ING-587]

vincent-pochet:
> Stacked on #6268, review that one first.

## Context

The bucket fetch only served a window whose start fell on a 15-minute wall, or on the subscription
start. A billing period rarely opens on a bucket boundary, so almost every current usage read fell
back to the events store and the buckets served nothing.

The guard was there to avoid counting a bucket that straddles two billing periods. It cannot happen:
the pipeline attributes an event to a subscription only within that subscription's lifetime, so an
upgrade, a downgrade or a termination landing mid-bucket still routes each event to the subscription
it belongs to.

## Description

- Floor the window start and top the window end to whole buckets, and serve every window.
- Drop the alignment check, the subscription start escape hatch and the `max_timestamp` refusal,
  together with the nil window the caller had to handle.

A frozen window is still refused per charge by the provider, which is where the rest of the serving
gates live.


vincent-pochet:
The only mid-bucket start an ongoing period can have is the timezone-change one, and it is worth one second.

`Subscriptions::DatesService#charges_from_datetime` opens the period at `previous_charges_to_datetime + 1.second`, and that previous value is an `end_of_day` (`…23:59:59.999999`), so the start sits `0.999999s` past a bucket wall and flooring pulls the wall back in. What it adds is the second between the two periods: a hole the `+ 1.second` correction leaves, not an overlap, billed by no invoice. Nothing already billed in the previous period is counted twice.

The other unaligned starts are safe: the first period (`from == started_at`, and an event is attributed to a subscription only within its lifetime), a termination (the end is ceiled, and later events belong to another subscription), and truncated windows, already refused by `Events::Stores::Provider#may_precompute_charge?`.

Codified rather than guarded, in `spec/services/realtime_usage/fetch_buckets_service_spec.rb` ("when the window starts one second after the previous period"). The real defect is the boundary arithmetic, which leaves the hole it means to close with or without buckets — TODO added there.

app/services/realtime_usage/fetch_buckets_service.rb
toommz: Didn't catch that during last review, but the name is a bit misleading. Not blocking at all but I was a bit surprised that you were doing maths on times and sizes in `floor_to_bucket` 😅 Turned out this constan is a "time" value too

toommz (approved):

app/services/realtime_usage/fetch_buckets_service.rb
vincent-pochet: ahh, good catch indeed, I'll rename it into `BUCKET_DURATION` to make it clearer

PR #6453, opened 2026-09-21
feat(wallets): refresh wallets from usage triggers [ING-632]

vincent-pochet:
> Replaces #6285, restarted from the proof-of-concept branch and replayed onto main.

## Context

Wallet ongoing balances refresh on a five-minute sweep of a Redis flag, so a balance trails its
usage by up to five minutes. The pipeline already emits a trigger per enriched event, so that
interval is the latency left.

## Description

- Consume the realtime usage trigger topic behind `LAGO_KAFKA_REALTIME_USAGE_TRIGGERS_TOPIC` and refresh inline: throughput comes from the topic's partitions, not a job queue.
- Collapse a poll batch into one refresh per customer, skipping the customers with no wallet.
- Hold a customer, up to five seconds, until its subscriptions' usage buckets reach the watermark the trigger carries.
- Leave a trigger older than thirty seconds to the sweep.
- Skip the organizations realtime usage is not served for, whose buckets no read path would use.
- Poll the buckets `unscoped`: any row version at the watermark proves the epoch landed, so the model's `final` default scope is not worth paying on every cycle of the wait.

The sweep stays authoritative and covers every customer this lane walks away from. Not covered
yet, to settle while iterating: a lock against a concurrent sweep refresh, a consume deadline, one
batched watermark read rather than one per subscription, and ClickHouse error handling.


vincent-pochet:
Pre-review points, on `4ba075145`.

Fixed:
- a timed-out bucket wait no longer refreshes, so the customer stays flagged for the sweep
- stale watermarks get one check instead of none, without spending the batch's deadline on a sleep
- `CONSUME_DEADLINE` (2 min) keeps a batch under `max.poll.interval.ms`
- the refresh is rescued per customer, so one failure no longer replays the batch
- payloads with no `subscription_id` are dropped from the watermarks

Left: the `tax_error` detail is not created here (the guard was dropped, realtime usage excludes the tax integrations), the `streaming` queue's inflow now scales with the trigger rate, and the sweep can still refresh a customer concurrently.

toommz (approved):

